PR #367, opened 2026-04-16
add risk-query component integration test

### LINKED ISSUE #362: Component integration test: risk-query
The integration test should follow the test pattern from this report https://github.com/HarperFast/harper-qe-logbook/blob/main/releases/v5/risk-query.mdx

ldt1996:
closes #362 
It automates the manual tests from https://github.com/HarperFast/harper-qe-logbook/blob/main/releases/v5/risk-query.mdx

I did not include two cases:
- auth test:  the component doesn't require auth in dev mode so the no-auth check passes when it shouldn't. Need to figure out the right config to enforce auth in CI.
- TTL eviction: testing record expiry means setting a short TTL, inserting a record, and waiting for it to expire. Doable but adds wait time to the test. 

Would love some suggestions for these two cases.

### DIFF
--- a/integrationTests/components/risk-query.test.ts
+++ b/integrationTests/components/risk-query.test.ts
@@ -7,25 +7,19 @@
 import { suite, test, before, after } from 'node:test';
 import { strictEqual, ok, deepStrictEqual } from 'node:assert/strict';
 
-import { startHarper, teardownHarper, type ContextWithHarper } from '../utils/harperLifecycle.ts';
+import { startHarper, teardownHarper, sendOperation, type ContextWithHarper } from '../utils/harperLifecycle.ts';
 
 suite('Component: risk-query', (ctx: ContextWithHarper) => {
 	before(async () => {
 		await startHarper(ctx);
 
 		// Deploy risk-query from GitHub
-		const response = await fetch(ctx.harper.operationsAPIURL, {
-			method: 'POST',
-			headers: { 'Content-Type': 'application/json' },
-			body: JSON.stringify({
-				operation: 'deploy_component',
-				project: 'risk-query',
-				package: 'https://github.com/HarperFast/risk-query',
-				restart: true,
-			}),
+		const body = await sendOperation(ctx.harper, {
+			operation: 'deploy_component',
+			project: 'risk-query',
+			package: 'https://github.com/HarperFast/risk-query',
+			restart: true,
 		});
-		strictEqual(response.status, 200);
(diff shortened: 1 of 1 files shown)


integrationTests/components/risk-query.test.ts
```diff
@@ -0,0 +1,177 @@
...
+import { strictEqual, ok, deepStrictEqual } from 'node:assert/strict';
+
+import { startHarper, teardownHarper, type ContextWithHarper } from '../utils/harperLifecycle.ts';
+
+suite('Component: risk-query', (ctx: ContextWithHarper) => {
+	before(async () => {
+		await startHarper(ctx);
+
+		// Deploy risk-query from GitHub
+		const response = await fetch(ctx.harper.operationsAPIURL, {
```
kriszyp: As mentioned in other PR, might use `sendOperation` here.

kriszyp (approved):
Looks great.
(It seems like TTL expiration of like 5 seconds wouldn't slow the tests down too much, but not sure if that would troublesome.)

integrationTests/components/risk-query.test.ts
ldt1996: done

ldt1996:
> Looks great. (It seems like TTL expiration of like 5 seconds wouldn't slow the tests down too much, but not sure if that would troublesome.)

Looking again at the tests we already have, TTL eviction is covered in `integrationTests/server/storage-reclamation.test.ts` and `unitTests/resources/caching.test.js`, so I think it's safe to skip it at the component layer ?

integrationTests/components/risk-query.test.ts
```diff
@@ -0,0 +1,171 @@
...
+		const deadline = Date.now() + 30_000;
+		while (true) {
+			try {
+				const check = await fetch(`${ctx.harper.httpURL}/RisqTable/`);
+				if (check.status === 200) break;
+			} catch {
+				// server not yet accepting connections
+			}
+			if (Date.now() > deadline) throw new Error('Timed out waiting for risk-query to be ready after deploy');
+			await new Promise((resolve) => setTimeout(resolve, 250));
```
cb1kenobi, line 35: Nothing wrong with this code, but we've been adopting this pattern if you wanna give it a try:

```
import { setTimeout as delay } from 'node:timers/promises';
```

```suggestion
			await delay(250);
```

cb1kenobi (approved):

PR #368, opened 2026-04-17
Add version support to spawn process management

kriszyp:
When a higher version number is passed in spawn options, the existing process is killed and a new one is started. The version is stored alongside the PID in the pid file. This allows callers to force a process restart when the underlying configuration or code changes.

Including this in a patch as it will likely to be important for our work on spawning local language models.

### DIFF
--- a/security/jsLoader.ts
+++ b/security/jsLoader.ts
@@ -763,7 +763,7 @@ function isProcessRunning(pid: number): boolean {
  */
 function parsePidFile(content: string): { pid: number; version: number } {
 	const lines = content.trim().split('\n');
-	const pid = parseInt(lines[0], 10);
+	const pid = Number.parseInt(lines[0], 10);
 	const version = lines.length > 1 ? parseInt(lines[1], 10) : 0;
 	return { pid, version };
 }
@@ -788,7 +788,7 @@ function acquirePidFileLock(
 					const { pid: existingPid, version: existingVersion } = parsePidFile(pidContent);
 
 					if (!isNaN(existingPid) && isProcessRunning(existingPid)) {
-						// If a higher version is requested, kill the existing process and re-acquire
+						// If the version isn't the one we want, kill the existing process and re-acquire
 						if (requestedVersion != null && requestedVersion !== existingVersion) {
 							try {
 								process.kill(existingPid);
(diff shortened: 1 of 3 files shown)


security/jsLoader.ts
```diff
@@ -761,23 +761,54 @@ function isProcessRunning(pid: number): boolean {
...
 				try {
 					const pidContent = readFileSync(pidFilePath, 'utf-8');
-					const existingPid = parseInt(pidContent.trim(), 10);
+					const { pid: existingPid, version: existingVersion } = parsePidFile(pidContent);
 
 					if (!isNaN(existingPid) && isProcessRunning(existingPid)) {
-						// Valid process is running, return its PID immediately
-						return existingPid;
+						// If a higher version is requested, kill the existing process and re-acquire
+						if (requestedVersion != null && requestedVersion > existingVersion) {
```
dawsontoth: In case the higher version has some breaking change, but still launches successfully, being able to kill that upgraded version and go back would be nice too, yeah?
```suggestion
						if (requestedVersion != null && requestedVersion !== existingVersion) {
```

kriszyp: My concern with this was that maybe there would be intertwining of old and new threads spawning. But I think this is a silly concern.

security/jsLoader.ts
```diff
@@ -761,23 +761,54 @@ function isProcessRunning(pid: number): boolean {
  * Acquires an exclusive lock using the PID file itself (synchronously with busy-wait)
  * Returns 0 if lock was acquired (need to spawn new process), or the existing PID if process is running
  */
-function acquirePidFileLock(pidFilePath: string, maxRetries = 100, retryDelay = 5): number {
+function parsePidFile(content: string): { pid: number; version: number } {
+	const lines = content.trim().split('\n');
+	const pid = parseInt(lines[0], 10);
```
cb1kenobi:
```suggestion
	const pid = Number.parseInt(lines[0], 10);
```

security/jsLoader.ts
```diff
@@ -761,23 +761,54 @@ function isProcessRunning(pid: number): boolean {
...
 				// File exists - check if it contains a valid running process
 				try {
 					const pidContent = readFileSync(pidFilePath, 'utf-8');
-					const existingPid = parseInt(pidContent.trim(), 10);
+					const { pid: existingPid, version: existingVersion } = parsePidFile(pidContent);
 
 					if (!isNaN(existingPid) && isProcessRunning(existingPid)) {
-						// Valid process is running, return its PID immediately
-						return existingPid;
+						// If a higher version is requested, kill the existing process and re-acquire
```
cb1kenobi:
```suggestion
						// If the version isn't the one we want, kill the existing process and re-acquire
```

cb1kenobi (approved):

PR #369, opened 2026-04-17
add ecommerce-template component integration test

### LINKED ISSUE #366: Component integration test: ecommerce-template
The integration test should follow the test pattern from this report https://github.com/HarperFast/harper-qe-logbook/blob/main/releases/v5/harper-ecommerce-template.mdx

ldt1996:
closes #366 

It automates the manual tests from https://github.com/HarperFast/harper-qe-logbook/blob/main/releases/v5/harper-ecommerce-template.mdx

It fails currently at deploy, as it should.

VM sandbox blocks the component from loading modules outside its application folder, and the Next.js build step isn't allowed to run without shell command permissions and a prebuilt config. 
These are the same issues I ran into during v5 testing.

**Local verification with workarounds:**
Ran harper from the component dir with `dependencyContainment: false` and a prebuilt Next.js bundle:
```
✔ homepage returns 200
✔ /products page returns 200
✔ /products/[id] page returns 200
✔ GET /Product/ returns array of seeded products
✔ GET /Product/:id returns complete product record
✔ GET /Traits/ returns trait data
✔ POST /Product/ creates a new product
✔ PUT /Product/:id replaces the record
✔ DELETE /Product/:id removes the record
✔ invalid product ID returns 404
✔ invalid route returns 404
✔ string product ID returns 404
✔ REST API nonexistent product returns 404
✖ GraphQL query returns products (this is a known bug: [nextjs#36](https://github.com/HarperFast/nextjs/issues/36))
```

It's actually reassuring to see how it catches all bugs caught via manual tests


integrationTests/components/ecommerce-template.test.ts
```diff
@@ -0,0 +1,166 @@
...
+import { suite, test, before, after } from 'node:test';
+import { strictEqual, ok, deepStrictEqual } from 'node:assert/strict';
+
+import { startHarper, teardownHarper, type ContextWithHarper } from '../utils/harperLifecycle.ts';
+
+suite('Component: ecommerce-template', (ctx: ContextWithHarper) => {
+	before(async () => {
+		await startHarper(ctx);
+
+		const response = await fetch(ctx.harper.operationsAPIURL, {
```
kriszyp: We do have a `sendOperation` function exported from harperLifecycle.ts to make this a lot easier.

kriszyp (commented):
This looks great!

I think we do have CI test failure due to:
`Could not find a production build in the '.next' directory. Try building your app with 'next build' before starting the production server.`

ldt1996:
@kriszyp Afaik, The NextJS failure is the same build issue we discussed with Ethan 2 weeks ago (I think). `npx next build` hits the RocksDB lock during SSG, so CI can't produce a build. The only way to bypass it is to skip the test.

When I ran it locally to see if the rest of tests work, I used `prebuilt: true`

integrationTests/components/ecommerce-template.test.ts
ldt1996: Thank you, updating

cb1kenobi:
> hits the RocksDB lock during SSG

What's this?

ldt1996:
> > hits the RocksDB lock during SSG
> 
> What's this?

statis site generation, what I meant is that NextJS it prerenders pages at build time

cb1kenobi:
> > > hits the RocksDB lock during SSG
> > 
> > 
> > What's this?
> 
> statis site generation, what I meant is that NextJS it prerenders pages at build time

Yes, I know what SSG is, but I don't know what "hits the RocksDB lock" is. That would only happen if you have another process trying to open the same database like when you have a orphan Harper process running (no pid file).

kriszyp:
Yeah, when the nextjs plugin runs the build, nextjs spawns a new process to do the SSG build, and that process can try to access the database, and as you might expect, that fails. However, the error is (not sure if you are seeing that):
```
Error: IO error: While lock file: /user.../hdb/database/system/LOCK: Resource temporarily unavailable opening database
```
The resolution to that is probably to implement https://github.com/HarperFast/rocksdb-js/issues/490 and then implement support for falling back to read-only mode in Harper, so that nextjs can do the build. Or convince nextjs to run in a worker thread or something (Ethan was looking into that). I think we considered this to not be that high of priority because most customers deploy from a CI in GH.

kriszyp:
Added https://github.com/HarperFast/nextjs/issues/37 to track it.
Let's disable the test for now.
I did notice that you referenced https://github.com/HarperFast/nextjs/issues/36. Do you think that is _caused_ by https://github.com/HarperFast/nextjs/issues/37 (is 37 the real issue)?

cb1kenobi:
> The resolution to that is probably to implement https://github.com/HarperFast/rocksdb-js/issues/490 and then implement support for falling back to read-only mode in Harper, so that nextjs can do the build.

I'm very eager to work on issue 490!

kriszyp (commented):
Thanks for this — the assertions are genuinely solid (real CRUD round-trips with follow-up GETs, DELETE asserting `=== true` then a 404, not a smoke test).

What's blocking me is CI: all four `Integration Tests 1/4` shards fail deterministically in the `before` hook with "Timed out waiting for ecommerce-template to be ready after deploy." From the Node 22 log (run `26468726590`), the root cause is `npx next build` colliding on the running instance's LMDB `LOCK` (`Resource temporarily unavailable`) — the build exits 1, the component never serves, and the 60s readiness poll times out, cascading all 13 tests.

Question before this lands: **should this still be failing?** I thought we'd fixed the Next.js adapter to support running the build against a live instance — did that fix not cover this path, or has this PR drifted since (it's ~62 days stale, last touched 2026-05-26)? If the adapter now handles it, this may just need a rebase; if not, the suite should self-skip when the build can't complete, so a required check doesn't sit permanently red on the shared branch and mask real regressions in that shard.

Happy to approve once CI is green (or consciously skipped). Thanks!

— Reviewed by Claude (claude-opus-4-8)

PR #370, opened 2026-04-17
Add loadComponent config option for conditional package loading

kriszyp:
Supports 'always' (default), 'if-installed' (skip silently if not found), and 'dev-only' (skip unless process.env.DEV_MODE is set).


dawsontoth (approved):
I can update the config schema file - I'm assuming we got it copied into this repo but I never checked it!

cb1kenobi (approved):

PR #371, opened 2026-04-17
Fix multiple HNSW bugs causing intermittent search quality issues

kriszyp:
- Only replace entry point when new node level is strictly higher, not equal, to avoid replacing well-connected entry points with unconnected new nodes
- Use passed distance function for entry point in searchLayer instead of always using the instance default, which mixed distance metrics
- Prevent orphaning nodes at level 0 when pruning excess connections
- Update stale distances in reverse connections when a node's vector changes

### DIFF
--- a/resources/indexes/HierarchicalNavigableSmallWorld.ts
+++ b/resources/indexes/HierarchicalNavigableSmallWorld.ts
@@ -130,8 +130,8 @@ export class HierarchicalNavigableSmallWorld {
 			// Generate random level for this new element
 			const level = oldNode.level ?? Math.min(Math.floor(-Math.log(Math.random()) * this.mL), MAX_LEVEL);
 			let currentLevel = entryPoint.level;
-			if (level >= currentLevel) {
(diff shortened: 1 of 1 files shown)


resources/indexes/HierarchicalNavigableSmallWorld.ts
```diff
@@ -130,8 +130,8 @@ export class HierarchicalNavigableSmallWorld {
 			// Generate random level for this new element
 			const level = oldNode.level ?? Math.min(Math.floor(-Math.log(Math.random()) * this.mL), MAX_LEVEL);
 			let currentLevel = entryPoint.level;
-			if (level >= currentLevel) {
-				// if we are at this level or higher, make this the new entry point
+			if (level > currentLevel) {
+				// if we are at a higher, make this the new entry point
```
heskew:
```suggestion
				// if we are at a higher level, make this the new entry point
```

heskew (approved):

dawsontoth (approved):

PR #373, opened 2026-04-17
allow module imports from component dir

ldt1996:
Module imports from components broke with `Can not load module outside
of allowed paths` when running from source.

The sandbox only checked against the component directory, but modules also need access
to `PACKAGE_ROOT`. On top of that, macOS resolves `/va`r to `/private/var` through a symlink, so the real path never matched the allowed path.

### DIFF
--- a/components/Application.ts
+++ b/components/Application.ts
@@ -187,7 +187,9 @@ export async function extractApplication(application: Application) {
 			try {
 				packResult = JSON.parse(stdout.slice(stdout.indexOf('[')));
 			} catch (err) {
-				throw new Error(`Failed to parse npm pack output for ${application.packageIdentifier}: ${err.message}\nstdout: ${stdout}`);
+				throw new Error(
+					`Failed to parse npm pack output for ${application.packageIdentifier}: ${err.message}\nstdout: ${stdout}`
+				);
 			}
 			if (!Array.isArray(packResult) || typeof packResult[0]?.filename !== 'string') {
 				throw new Error(`Unexpected npm pack output for ${application.packageIdentifier}:\n${stdout}`);
--- a/components/ApplicationScope.ts
+++ b/components/ApplicationScope.ts
@@ -21,10 +21,10 @@ export class ApplicationScope {
 	server: Server;
 	mode?: 'native' | 'vm' | 'vm-current-context' | 'compartment'; // option to set this from the scope
 	dependencyLoader?: 'native' | 'app' | 'auto'; // option to set this from the scope
-	verifyPath?: string;
+	allowedPaths?: string[];
 	config: any;
 	moduleCache: any; // used by the loader to retain a cache of modules, type is an internal detail of the loader
-	constructor(name: string, resources: Resources, server: Server, isInternal = false, verifyPath?: string) {
+	constructor(name: string, resources: Resources, server: Server, isInternal = false) {
 		this.logger = forComponent(name, !isInternal);
 
 		this.resources = resources;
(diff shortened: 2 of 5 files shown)


ldt1996:
Ran into this issue while writing the integration test for redirector

kriszyp (commented):
The sandbox is preventing you from accessing modules outside of your component directory, and it sounds like it succeeded. Why do you need to access modules in the package root?

ldt1996:
> The sandbox is preventing you from accessing modules outside of your component directory, and it sounds like it succeeded. Why do you need to access modules in the package root?

I ran into this

`Could not load component 'jsResource' for application 'redirector' due to: Failed to load resource module /Users/lavinia/harper/components/redirector/src/resources/index.js: Can not load module outside of allowed paths`

The symlink from symlinkHarperModule points to P`ACKAGE_ROOT`, which is outside the component dir, and the sandbox blocks it. Adding `PACKAGE_ROOT` to the allowed paths lets the component follow the symlink that harper created.

The `realpathSync` is because macOS resolves `/var` to `/private/var`, which causes the path check to fail even for files inside the component dir.

kriszyp:
You symlinked the redirector to /private/var? I don't understand.

cb1kenobi:
> You symlinked the redirector to /private/var? I don't understand.

On macOS, `/var` is symlinked to `/private/var` as is `/tmp` is symlinked to `/private/tmp`.

Do we need to call `realpathSync()`?

ldt1996:
@kriszyp @cb1kenobi Let me try to better explain my thought process.

Let's say the allowed directory is `/tmp/harper-test/components/redirector`. When a module imports a file, the sandbox checks if the file's path starts with `/tmp/harper-test/components/redirector`. Because on macOS `/tmp` is a symlink to `/private/tmp`, the check fails. The file's resolved path would be `/private/tmp/harper-test/components/redirector/<file>`.
I used `realpathSync` on both the allowed path and the module path so the comparisson is always between real paths.

The scope is narrow if we think only about macOS, but it could happen anywhere there's a symlink in the path. Is there an alternative to `realpathSync` ?

kriszyp:
I was using `realpathSync` precisely _because_ of symlinks. Without that, there can ambiguously different paths for the same file/folder. Are we comparing a canonical/real path with a non-canonical path somewhere? I suspect we need to use `realpathSync` _more_.

ldt1996:
@kriszyp `checkAllowedModulePath` was the mismatch in this case. Happy to look for other places that might have the same issue if you want.

ldt1996:
@kriszyp you're right, `realpathSync` alone is enough

.github/workflows/create-release.yaml
```diff
@@ -43,7 +43,7 @@ jobs:
           echo "Package.json version: v${{ steps.package-version.outputs.packageVersion }}"
           test "${{ steps.tag-version.outputs.tagVersion }}" == "v${{ steps.package-version.outputs.packageVersion }}"
       - name: Notify release in progress in Slack
-        uses: slackapi/slack-github-action@91efab103c0de0a537f72a35f6b8cda0ee76bf0a # v2.1.1
+        uses: slackapi/slack-github-action@af78098f536edbc4de71162a307590698245be95 # v3.0.1
```
heskew, line 46: Are these needed for the fix? 👀 🙂

ldt1996: I  pulled from `v5.0` branch, the target was main before, and I needed to change it

ldt1996: should be gone now

kriszyp (approved):
Great, thank you!

PR #375, opened 2026-04-17
Add migrateOnStart option to migrate LMDB databases to RocksDB

kriszyp:
Adds a new `storage.migrateOnStart` config option that, when enabled, reads all LMDB databases on startup and copies their data into RocksDB format, then backs up the original .mdb files. This enables a one-shot migration path from LMDB to RocksDB storage engines.

Adds an integration test that verifies data integrity after migration from a 4.x (LMDB) Harper instance to v5 RocksDB storage.

Note: audit store history is not migrated due to fundamental format differences between LMDB and RocksDB audit stores; a fresh audit store is created automatically on first open.

### DIFF
--- a/bin/copyDb.ts
+++ b/bin/copyDb.ts
@@ -440,7 +440,13 @@ async function copyDbToRocks(sourceRootStore, sourceDatabase: string, targetPath
 								outstandingWrites = 0;
 							}
 						} catch (error) {
-							console.error('Error migrating record', typeof key === 'symbol' ? 'symbol' : key, 'from', sourceDatabase, error);
+							console.error(
+								'Error migrating record',
+								typeof key === 'symbol' ? 'symbol' : key,
+								'from',
+								sourceDatabase,
+								error
+							);
 						}
 					}
 				} else {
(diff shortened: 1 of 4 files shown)


DavidCockerill (approved):

Devin-Holland:
This requires ~2x the disk space assuming the mdb isn't inflated space compared to what it's actually using correct?

bin/copyDb.ts
```diff
@@ -278,3 +281,208 @@ export async function copyDb(sourceDatabase: string, targetDatabasePath: string)
...
+		console.log('migrated database ' + sourceDatabase + ' to RocksDB');
+	} finally {
+		transaction.done();
+		targetRootStore.close();
+	}
+
+	async function copyDbiToRocks(sourceDbi, targetDbi, isPrimary, transaction) {
+		let recordsCopied = 0;
+		let skippedRecord = 0;
+		let retries = 10000000;
```
cb1kenobi: This seems huge. What would even trigger a retry? `getRange()` throwing? Why would that happen? Corruption?

cb1kenobi (approved):

bin/copyDb.ts
kriszyp: Yes, we had done this to try to tolerate corruption errors. When there is a failure it picks up at the last key position (it doesn't restart the iteration).

PR #380, opened 2026-04-18
parse npm pack output as JSON

### LINKED ISSUE #379: `deploy_component` fails to parse `npm pack` stdout when lifecycle scripts run
### Bug Summary

`npm pack` stdout is parsed as a filename via `stdout.trim()`. When the packed template defines a `prepare` script, npm prepends lifecycle output to stdout and the tarball path resolves to nonsense, throwing ENOENT before extraction.

https://github.com/ldt1996/template-early-hints

### Steps to Reproduce

1. Deploy a template with a `prepare` script to Harper via `deploy_component`:
 `  curl -X POST <harper>/ -d '{"operation":"deploy_component","project":"early-hints","package":"https://github.com/ldt1996/template-early-hints","restart":true}'`
2. See error: ENOENT on a path containing the lifecycle script output
3. Reproduced on both local and Fabric stage (us-lax-1.v5-testing)

### Expected Behavior

Deploy succeeds. `npm pack` runs `prepare` per npm convention, `dist/ ` gets built into the tarball, harper extracts and loads the component normally.

### Actual Behavior

Deploy fails with:

`error: "ENOENT: no such file or directory, open '/path/to/components/> template-early-hints@2.0.0 prepare\n> tsc\ntemplate-early-hints-2.0.0.tgz'"`

### Platform

Harper 5.0.1, both local and Fabric cluster

### Console Errors

_No response_

### Screenshots or Videos

_No response_

### Additional Context

_No response_

ldt1996:
closes #379

### DIFF
--- a/components/Application.ts
+++ b/components/Application.ts
@@ -188,9 +188,12 @@ export async function extractApplication(application: Application) {
 				throw new Error(`npm pack produced no JSON output for ${application.packageIdentifier}:\n${stdout}`);
 			}
 
-			const [{ filename }] = JSON.parse(stdout.slice(jsonStart)) as [{ filename: string }];
-			tarballPath = join(parentDirPath, filename);
-			// Create a Readable from the tarball
+			const packResult = JSON.parse(stdout.slice(jsonStart));
+			if (!Array.isArray(packResult) || typeof packResult[0]?.filename !== 'string') {
+				throw new Error(`Unexpected npm pack output for ${application.packageIdentifier}:\n${stdout}`);
+			}
+
+			tarballPath = join(parentDirPath, packResult[0].filename);
 			tarball = createReadStream(tarballPath);
 		}
 	}
--- a/components/Application.ts
+++ b/components/Application.ts
@@ -183,12 +183,12 @@ export async function extractApplication(application: Application) {
 				throw new Error(`Failed to download package ${application.packageIdentifier}: ${stderr}`);
 			}
 
-			const jsonStart = stdout.indexOf('[');
-			if (jsonStart === -1) {
-				throw new Error(`npm pack produced no JSON output for ${application.packageIdentifier}:\n${stdout}`);
+			let packResult: Array<{ filename: string }>;
+			try {
+				packResult = JSON.parse(stdout.slice(stdout.indexOf('[')));
+			} catch (err) {
+				throw new Error(`Failed to parse npm pack output for ${application.packageIdentifier}: ${err.message}\nstdout: ${stdout}`);
 			}
-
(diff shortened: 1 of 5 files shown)


components/Application.ts
```diff
@@ -172,14 +172,24 @@ export async function extractApplication(application: Application) {
...
+			if (code !== 0) {
+				throw new Error(`Failed to download package ${application.packageIdentifier}: ${stderr}`);
+			}
+
+			const jsonStart = stdout.indexOf('[');
+			if (jsonStart === -1) {
+				throw new Error(`npm pack produced no JSON output for ${application.packageIdentifier}:\n${stdout}`);
+			}
+
+			const [{ filename }] = JSON.parse(stdout.slice(jsonStart)) as [{ filename: string }];
```
kriszyp: Can this return an object or something with an error? (something other than a filename?) is it worth checking for that?

kriszyp (commented):
This looks great, I like this approach.

components/Application.ts
ldt1996: good point, added a shape check https://github.com/HarperFast/harper/pull/380/changes/6a4943414f28d757ac15a8ea139aa4792846fbef
do you think this would do?

components/Application.ts
```diff
@@ -172,15 +172,28 @@ export async function extractApplication(application: Application) {
...
+				application.name,
+				'npm',
+				['pack', '--json', application.packageIdentifier],
+				parentDirPath
+			);
+			if (code !== 0) {
+				throw new Error(`Failed to download package ${application.packageIdentifier}: ${stderr}`);
+			}
+
+			const jsonStart = stdout.indexOf('[');
```
cb1kenobi: Do we need to do this? Assuming a success exit code, the output should be only valid JSON.

components/Application.ts
```diff
@@ -172,15 +172,28 @@ export async function extractApplication(application: Application) {
...
+			if (jsonStart === -1) {
+				throw new Error(`npm pack produced no JSON output for ${application.packageIdentifier}:\n${stdout}`);
+			}
+
+			const packResult = JSON.parse(stdout.slice(jsonStart));
```
cb1kenobi: Consider wrapping `JSON.parse()` in a try/catch so we can prepend some context to the error message.

components/Application.ts
ldt1996: fair, folded it into the slice. still keeping --json tho, lifecycle output can land on stdout before the array

components/Application.ts
ldt1996: done

cb1kenobi (approved):

kriszyp (approved):
Great, thank you

PR #388, opened 2026-04-20
Tolerate missing role on user

kriszyp:


DavidCockerill (approved):

cb1kenobi (approved):

Devin-Holland (approved):

heskew (approved):
Seems correct, but not caught early because user is (often) `any` typed? User interface has the role as optional.

PR #391, opened 2026-04-21
support `HARPER_CLI_*` env vars

ldt1996:
UPDATE:

After some discussions, the agreement is that we'll prefer `HARPER_` prefixed env vars , but we accept both versions

fixes `HdbError: Harper properties file at path /Users/lavinia/.nvm/versions/node/v24.14.1/lib/node_modules/harper/dist/hdb_boot_properties.file does not exist`

### DIFF
--- a/bin/cliOperations.js
+++ b/bin/cliOperations.js
@@ -84,9 +84,12 @@ async function cliOperations(req) {
 			password: req.password || target.password || process.env.HARPER_CLI_PASSWORD || process.env.CLI_TARGET_PASSWORD,
 			rejectUnauthorized: req.rejectUnauthorized,
 		};
+		console.error(`Connecting to ${target.protocol}//${target.hostname}:${target.port}`);
+
 	} else {
 		// if we aren't doing a targeted operation (like deploy), we initialize the config and verify that local harper
 		// is running and that we can communicate with it.
+		console.error('Connecting to local Harper instance');
 		initConfig();
 		if (!getHdbPid()) {
 			console.error('Harper must be running to perform this operation');
--- a/bin/cliOperations.js
+++ b/bin/cliOperations.js
@@ -85,7 +85,6 @@ async function cliOperations(req) {
 			rejectUnauthorized: req.rejectUnauthorized,
 		};
 		console.error(`Connecting to ${target.protocol}//${target.hostname}:${target.port}`);
-
 	} else {
 		// if we aren't doing a targeted operation (like deploy), we initialize the config and verify that local harper
 		// is running and that we can communicate with it.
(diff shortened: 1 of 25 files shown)


bin/cliOperations.js
```diff
@@ -63,7 +63,7 @@ function buildRequest() {
  */
 async function cliOperations(req) {
 	if (!req.target) {
-		req.target = process.env.CLI_TARGET;
+		req.target = process.env.CLI_TARGET || process.env.HARPER_CLI_TARGET;
```
cb1kenobi: What are these `HARPER_CLI_*` env vars? Why aren't the `CLI_TARGET*` env vars sufficient?

ldt1996: It began with a user setting `HARPER_CLI_TARGET` in a GH Actions workflow. We discussed on the slack thread, and Kris confirmed we want to move toward `HARPER_`-prefixed env vars

cb1kenobi: K, just checking. We should add a ticket these new env vars.

cb1kenobi (approved):

kriszyp (approved):
This is good, but were you going to add some logging to indicate if the CLI was connecting to local vs a target?

ldt1996:
@kriszyp That is correct. Finally added it. I overengineered it, then dropped all that, now it's 2 lines

bin/cliOperations.js
```diff
@@ -73,20 +73,22 @@ async function cliOperations(req) {
 			try {
 				target = new URL(`https://${req.target}:9925`);
 			} catch {
-				throw error; // throw the original error
```
kriszyp, line 76: Why you no like comments?

bin/cliOperations.js
```diff
@@ -73,20 +73,22 @@ async function cliOperations(req) {
...
 			protocol: target.protocol,
 			hostname: target.hostname,
 			port: target.port,
-			username: req.username || target.username || process.env.CLI_TARGET_USERNAME,
-			password: req.password || target.password || process.env.CLI_TARGET_PASSWORD,
+			username: req.username || target.username || process.env.HARPER_CLI_USERNAME || process.env.CLI_TARGET_USERNAME,
+			password: req.password || target.password || process.env.HARPER_CLI_PASSWORD || process.env.CLI_TARGET_PASSWORD,
 			rejectUnauthorized: req.rejectUnauthorized,
 		};
+		console.error(`Connecting to ${target.protocol}//${target.hostname}:${target.port}`);
```
kriszyp, line 87: It seems like this is not an error... Is this to keep stdout cleaner with the response data? (that's fine with me, if that's the reason)

ldt1996: yep, it's to keep stdout cleaner

bin/cliOperations.js
ldt1996: I removed it by mistake

ldt1996: I don't have beef with it tho

PR #392, opened 2026-04-21
Adjust scoring

kriszyp:


DavidCockerill (approved):

security/keys.js
```diff
@@ -351,7 +351,7 @@ function certExtensions() {
 async function createCertificateTable(cert, caCert) {
 	await setCertTable({
 		name: getThisNodeName(),
-		uses: ['https', 'wss'],
```
Devin-Holland, line 354: Should this one be `uses: ['replication']`?

Devin-Holland (commented):
We're testing this out in 4.7 off the testing branch today. Will let you know if we come up with any more changes.

Devin-Holland (approved):

PR #393, opened 2026-04-22
Add retry backoff and fix index removal in transactions

kriszyp:
Implements exponential backoff for transaction retries (up to 100 attempts) to reduce contention conflicts. Fixes index removal to pass primary key through options due to DBI interface constraints. Updates eviction logic to use transactions for RocksDB index cleanup instead of ifVersion.

### DIFF
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -95,6 +95,8 @@ export class DatabaseTransaction implements Transaction {
 		if (this.open !== TRANSACTION_STATE.OPEN) return; // can not start a new read transaction as there is no future commit that will take place, just have to allow the read to latest database state
 
 		this.transaction = new RocksTransaction(this.db.store);
+		harperLogger.warn('Created new transaction to read', this.transaction.id);
+
 		if (this.timestamp) {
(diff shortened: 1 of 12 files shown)


resources/DatabaseTransaction.ts
```diff
@@ -295,6 +297,15 @@ export class DatabaseTransaction implements Transaction {
...
 							// for future transactions
 							this.retries++;
+							if (this.retries > 2) {
+								if (this.retries > MAX_RETRIES) {
+									throw new ServerError(
+										`After ${MAX_RETRIES} retries, unable to commit transaction, transaction is in conflict with ongoing writes`
+									);
+								}
+								// start delaying, back off to try to space out transactions and avoid excessive conflicts
+								return delay(this.retries).then(() => this.commit({ transaction }));
```
cb1kenobi: Not a very sophisticated backoff delay. Maybe use a quadratic backoff?

```suggestion
								return delay(Math.max(1, Math.min(0.6 * Math.pow(this.retries, 2), 64))).then(() => this.commit({ transaction }));
```

cb1kenobi (approved):

PR #394, opened 2026-04-22
Skip node hostname prompt when env var not present

DavidCockerill:
Skip the node hostname prompt and default its value to null when installing Harper with the basic required prompt override env vars set (rootpath, admin username, admin password) but no hostname specified.


kriszyp (approved):
Looks great, thank you!

cb1kenobi (approved):
I swear I've run into this. Thank you for fixing this.

PR #396, opened 2026-04-23
Expose RocksDB stats in system information

### LINKED ISSUE #70: Expose RocksDB statistics/metrics
### Feature Summary

With https://github.com/HarperFast/rocksdb-js/pull/292, we should have access to a lot of RocksDB statistics. In https://github.com/HarperFast/harper/blob/main/utility/environment/systemInformation.js#L243-L279, we should branch for RocksDB, and provide a good set of database stats/metrics, like block cache usage, mem table usage, read latency, write stalls, compactions and anything else that might be informative (but not noisy).

### Problem This Solves

Gives us deeper insight into RocksDB performance, issues.

### Proposed Solution

do it.

### User Stories

_No response_

### Alternatives Considered

_No response_

### Priority/Impact

High - critical for user success

### Examples or References

_No response_

### Additional Context

_No response_

No, just reporting the issue

cb1kenobi:
Add RocksDB to `getMetrics()` and `getTableSize()`. Ported the `systeminformation.js` file to TypeScript, defined types for everything, and fixed the test.

You can manually test by running:
```bash
curl -X POST http://localhost:9925 -u "username:password" -H "Content-Type: application/json" --data-raw '{"operation":"system_information"}' | jq
```

Fixes #70


kriszyp (commented):
This looks great, nice work!
One thing that might be nice to do is to separate table and database level stats. The transaction log size and block cache capacity are not specific to tables, for example, they are shared by the tables in database, so there is no distinction in these stats between tables (well, block cache capacity is actually shared across all databases, but not table specific).

kriszyp (approved):
So many great stats! (we might need to prune at some point, but this looks amazing for now). Also, I was curious: it looks like this is now returning histogram stats. I thought that histogram stats were supposed to be explicitly enabled because there is database overhead involved in collecting them. Is that true?

Ethan-Arrowood (approved):
cool! nice work updating to typescript too.

Issue #395, opened 2026-04-22
Support named ordering of middleware


kriszyp:
Currently, we can only order middleware by the order in which the middleware (HTTP listener) was registered, or a big clumsy override of `runFirst`. This is not sufficiently granular control over middleware ordering. 
We need components be able to specify relative ordering, or dependencies in their position. So rather than just always loading the authentication handler first, the REST component should be able to specify authentication  as a dependency. And then the ordering will just ensure that authentication  goes before REST. And if a config has static before REST (which we could/should recommend in our template), than the natural ordering enforce by the dependency is static -> auth -> rest (generally the preferred order), but if you have REST before static, then you get auth -> rest -> static, which honors the order of the config.
So in addition to `runFirst` (and we should probably deprecate that), we should have:
`before` - The name of the middleware that the new middleware entry must run before
`after` - The name of the middleware that the new middleware entry must run after
One of the primary challenges is in "naming" the middleware (2 hard problems...). We probably want to actually start creating a scope `server` with an `http` listener that is prebound to the plugin's name. And then we could use that to reference middleware (with maybe an option to explicitly self-name).

kriszyp:
Shipped via #397 (declarative before/after middleware ordering and routing), merged 2026-05-12.

— Claude (Opus 4.8), issue-backlog triage on Kris's behalf

PR #397, opened 2026-04-23
Middleware ordering and routing

kriszyp:
### Problem

Harper's HTTP middleware stack had two coarse ordering mechanisms: registration order and the `runFirst` flag (a boolean override to jump to the front). This made it impossible for a component like REST to declaratively state "I depend on authentication" ? the ordering was implicit and config-order-sensitive.

There was also no way to scope middleware to specific URL paths or virtual hosts. All components registered globally and were responsible for filtering requests themselves.

### Changes

**`before`/`after` dependency-based ordering** (`Server.ts`, `middlewareChain.ts`)\
Components can now declare `before: 'name'` or `after: 'name'` in their `server.http()` options. A Kahn's algorithm topological sort uses the original registration index as a tiebreaker, so config file order is preserved when no explicit constraints exist between two entries. `runFirst` is deprecated but continues to work.

**Automatic component naming** (`Scope.ts`)\
The `scope.server` proxy now injects the plugin's name (and its `urlPath`/`host` config) into every `server.http/request/ws/upgrade` call. Plugins get named for free ? no extra code needed.

**REST declares its auth dependency** (`REST.ts`)\
REST now carries `after: 'authentication'`, ensuring auth always runs before REST regardless of config order.

**Static no longer forces itself to the front** (`static.ts`)\
`runFirst: true` is removed. Static now participates in normal ordering; its `before: 'authentication'` default (configurable) keeps the preferred `static ? auth ? rest` chain when they appear in that order in config.

**Per-route middleware chains** (`middlewareChain.ts`)\
Components configured with `urlPath` or `host` get their own isolated middleware chain. The dispatcher routes requests to the best-matching chain (host+path > host > path; longer prefix wins ties). Unmatched requests fall to the default chain. Sub-route chains automatically pull in transitive `after` dependencies from any route ? so if REST on `/api` declares `after: 'authentication'`, auth is included in the `/api` chain even though auth registered on the default route. Auth runs once per request (requests hit exactly one chain).

**Pure utility module + 38 unit tests** (`middlewareChain.ts`, `unitTests/server/middlewareChain.test.js`)\
All chain-building logic lives in a dependency-free module, directly importable in tests without mocking. Tests cover topoSort (ordering, cycle detection, first/last name semantics), dependency resolution, route matching, and full `makeCallbackChain` integration scenarios.

Addresses #395 and #416

## Documentation

- [Document middleware host and path routing (HarperFast/documentation#595)](https://github.com/HarperFast/documentation/pull/595)


Ethan-Arrowood (approved):
great work! 

do any other built-in plugins need ordering updates? I noticed you modified static and rest, but should something like authentication be set first? What about other server middlewares like the websocket stuff in rest, or the rest ones in graphql? Or is that all follow up?

server/middlewareChain.ts
```diff
@@ -0,0 +1,227 @@
...
+ */
+export function resolveDeps(entries: HttpEntry[], nameToEntry: Map<string, HttpEntry>): HttpEntry[] {
+	const included = new Set(entries);
+	let changed = true;
+	while (changed) {
+		changed = false;
+		for (const entry of [...included]) {
+			if (entry.after) {
+				const dep = nameToEntry.get(entry.after);
+				if (dep && !included.has(dep)) {
```
heskew, line 117: AI Review [High]: Trailing slash in urlPath causes match failure. If urlPath is '/api/', this becomes startsWith('/api//'), failing for '/api/foo'. Suggested fix: Normalize urlPath (remove trailing slash) before comparison. 🤖

server/middlewareChain.ts
```diff
@@ -0,0 +1,227 @@
...
+					changed = true;
+				}
+			}
+		}
+	}
+	return [...included];
+}
+
+/**
+ * Returns true when `request` satisfies the route's host and urlPath constraints.
```
heskew, line 137: AI Review [High]: stripPrefix returns malformed paths if prefix has trailing slash. e.g., /api/foo becomes foo instead of /foo. Also, this captures stale path state if middleware modifies the request later. 🤖

server/Server.ts
```diff
@@ -67,7 +67,18 @@ export interface UpgradeOptions {
 }
 
 export interface HttpOptions extends ServerOptions {
+	/** @deprecated Use `before` or `after` for explicit ordering instead */
```
heskew, line 59: AI Review [Low]: UpgradeOptions and WebSocketOptions were not updated to include these new fields, leading to TS errors for those listener types. 🤖

server/middlewareChain.ts
```diff
@@ -0,0 +1,227 @@
...
+			if (!nameToFirst.has(name)) nameToFirst.set(name, i);
+			nameToLast.set(name, i);
+		}
+	}
+
+	// successors[i] = list of indices that must come after i
+	const successors: number[][] = Array.from({ length: n }, () => []);
+	const inDegree = new Int32Array(n);
+	const addEdge = (from: number, to: number) => {
+		successors[from].push(to);
```
heskew, line 40: AI Review [Low]: Silent fallback on dependency cycles. Server.ts does not provide an onCycle callback, resulting in non-deterministic ordering if cycles exist. 🤖

heskew (commented):
These changes will be neat. Picking at it with gemini and only finding some edge cases...

----
## AI Review: Middleware ordering and routing

I have completed a deep review of this PR. The implementation of Kahn's algorithm for topological sorting and the new routing dispatcher is well-structured and provides a significant improvement over the legacy `runFirst` boolean.

However, I identified several issues related to trailing slash handling and state consistency in the `stripPrefix` proxy that should be addressed before merging.

🤖 Generated by Gemini CLI

cb1kenobi (approved):
Gave this a test with my app and things seemed to good. Nice work!

heskew (approved):

PR #400, opened 2026-04-23
More carefully cache transaction logs when we do not (yet) have acces…

kriszyp:
…s node ids

### DIFF
--- a/resources/RocksTransactionLogStore.ts
+++ b/resources/RocksTransactionLogStore.ts
@@ -122,14 +122,14 @@ export class RocksTransactionLogStore extends EventEmitter {
 	getEntry() {
 		throw new Error('Not implemented');
 	}
-	addLogToMaps(logName: string, log: TransactionLog) {
-		const nodeId = ((globalThis as any).server?.replication?.getIdOfRemoteNode?.(logName, this) ?? 0) as number;
+	addLogToMaps(log: TransactionLog) {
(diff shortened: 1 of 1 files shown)


DavidCockerill (approved):

resources/RocksTransactionLogStore.ts
```diff
@@ -122,39 +122,53 @@ export class RocksTransactionLogStore extends EventEmitter {
...
-		const nodeId = ((globalThis as any).server?.replication?.getIdOfRemoteNode?.(logName, this) ?? 0) as number;
+	addLogToMaps(log: TransactionLog) {
+		let logName: string = log.name;
 		if (this.nodeLogs) {
+			const nodeId = ((globalThis as any).server?.replication?.getIdOfRemoteNode?.(logName, this) ?? 0) as number;
 			this.nodeLogs![nodeId] ??= log;
 		}
 		this.updates++;
 		this.logByName.set(logName, log);
-		return nodeId;
```
cb1kenobi, line 132: `ensureLogExists()` returns `return this.addLogToMaps(log);`, so removing this could potentially affect whatever calls `ensureLogExists()`... which currently looks like nothing. 🙃

Consider either reverting this or update `ensureLogExists()` to not explicitly return a value;.

cb1kenobi (approved):

PR #403, opened 2026-04-24
Move nodeIdMapping from pro version to core

kriszyp:


kriszyp:
I tried eagerly loading replication https://github.com/HarperFast/harper-pro/pull/96/changes/6fffc73c247e65ff26c91cf60ee60c61ee16c10c, but that failed (https://github.com/HarperFast/harper-pro/actions/runs/24856733779)
I tried to updated transaction log caching to avoid caching until `getRemoteNodeId` is available: https://github.com/HarperFast/harper/pull/400
But that seems rather complicated and weird that we have special logic before a dependency is loaded and there are ids that can't be mapped until it is loaded.
This PR moves the `nodeIdMapping.ts` module into core, where it is really already pretty heavily used. Which I think is a little more sensible solution.

cb1kenobi (approved):

cb1kenobi:
LGTM!

PR #408, opened 2026-04-25
Add withNodeAdapter() to Request for Node middleware adapter pattern

kriszyp:
## Summary

- Adds `request.withNodeAdapter(handler)` to the `Request` class, which invokes `handler` with a Node.js-compatible `IncomingMessage`/`ServerResponse` pair adapted from the current Harper request, and returns a `Promise` that resolves to a Harper response once the handler writes status and headers.
- The `IncomingMessage` proxies the current `Request` state (method, url, headers — including any middleware mutations) while delegating body streaming to the underlying socket.
- The `ServerResponse` captures writes into a `PassThrough` body stream and resolves the promise as soon as headers are available, so the result can be returned directly from a Harper middleware handler.
- Propagates async handler rejections and post-flush `destroy()` errors to the appropriate consumers (promise or body stream).

## Usage

```js
server.http((request, next) =>
  request.withNodeAdapter((req, res) => someNodeMiddleware(req, res))
);
```

## Test plan

- [x] `nodeRequest` reflects middleware-mutated method, url, and headers
- [x] `writeHead` with object, array-of-pairs, and flat alternating-array formats
- [x] `setHeader`/`getHeader`/`hasHeader`/`removeHeader`
- [x] `setHeader` with `string[]` preserves multi-value headers (Set-Cookie)
- [x] `statusCode` set directly is captured correctly
- [x] `headersSent` kept in sync by all flush paths (write/end/flushHeaders/writeHead)
- [x] Body streaming: single chunk, multiple writes, empty body, callback overloads
- [x] `finish` event forwarded from body PassThrough to nodeResponse
- [x] `destroy()` before headers rejects the response promise
- [x] `destroy()` with no error rejects with a descriptive message
- [x] `destroy()` after flush propagates error to body consumer, not the promise
- [x] Async handler rejection before headers rejects the response promise
- [x] Async handler rejection after `end()` does not double-reject


server/serverHelpers/Request.ts
```diff
@@ -103,6 +105,171 @@ export class Request {
...
+	 * by middleware) while delegating body reading to the underlying stream. The ServerResponse captures
+	 * status, headers, and body, resolving the returned promise as soon as headers are available with a
+	 * streaming body that can be piped back through the Harper middleware chain.
+	 *
+	 * Example:
+	 *   server.http((request, next) =>
+	 *     request.sendNodeRequestResponse((req, res) => someNodeMiddleware(req, res))
+	 *   );
+	 */
+	sendNodeRequestResponse(
```
cb1kenobi: I'm not a super huge fan of the function name. It's verbose. "send" implies this sends something, but in actuality, it returns a `Response` object. Consider something more generic/fitting like `handleRequest`.

cb1kenobi (approved):

server/serverHelpers/Request.ts
kriszyp: I think it should have "Node" in it right? it's already request, in request handler, the point of the method is to provide the Node.js API for consumers that need it?
* handleRequestAsNode?
* provideNodeRequestResponse?
* dispatchAsNode?

cb1kenobi: As discussed on Slack, `withNodeAdapter()` sounds gooder!

PR #409, opened 2026-04-26
Use PID-based paths consistently in unit test utilities

kriszyp:
Previously the unit tests were spamming the logs with errors about being unable to find the primary key due to the system database changing paths mid-stream, resulting in out-of-sync system tables.

All test helpers now resolve test dbs to the same `unitTests/envDir/<pid>/` directory, preventing system database path conflicts when tests switch between the fixed envDir/schema path and the PID-scoped path. Teardown now removes only the PID directory, so concurrent test runs don't interfere with each other.

The system database path is protected.

### DIFF
--- a/unitTests/testUtils.js
+++ b/unitTests/testUtils.js
@@ -20,7 +20,8 @@ const MOCK_ARGS_ERROR_MSG =
 const UNIT_TEST_DIR = __dirname;
 const ENV_DIR_NAME = 'envDir';
 const ENV_DIR_PATH = path.join(UNIT_TEST_DIR, ENV_DIR_NAME);
-const BASE_SCHEMA_PATH = path.join(ENV_DIR_PATH, 'schema');
+const PID_DIR_PATH = path.join(ENV_DIR_PATH, process.pid.toString());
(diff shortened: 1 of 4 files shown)


unitTests/testUtils.js
```diff
@@ -10,7 +10,7 @@ const { table: ensure_table, resetDatabases } = require('#src/resources/database
 const terms = require('#src/utility/hdbTerms');
 const harperBridge = require('#js/dataLayer/harperBridge/harperBridge');
 const { isMainThread } = require('node:worker_threads');
-const { getDatabases } = require('#src/resources/databases');
+const { getDatabases, databases } = require('#src/resources/databases');
```
cb1kenobi, line 13: What's the difference between `getDatabases` and `databases`?

cb1kenobi (approved):
This looks great!

Ethan-Arrowood (approved):

unitTests/testUtils.js
kriszyp: `getDatabases` makes sure all the databases are loaded. In this case I only wanted to prevent a change to existing loaded databases, so that's why I used `databases` in the system check.

PR #410, opened 2026-04-26
Record caching

kriszyp:
Introduces CachingRocksDatabase, a RocksDatabase subclass that layers a
per-instance WeakLRUCache on top of the process-wide VerificationTable.
Cache freshness is checked via entry.version (Harper's localTime timestamp)
rather than a txnId, so reads hitting a fresh VT slot skip disk entirely.

Primary RocksDB stores now use CachingRocksDatabase instead of RocksDatabase.

DatabaseTransaction creates transactions with coordinatedRetry: true, so
IsBusy conflicts are signalled as RETRY_NOW_VALUE (immediate retry, no
backoff) rather than an ERR_BUSY rejection. The old quadratic-backoff retry
handler is removed.


resources/DatabaseTransaction.ts
```diff
@@ -9,7 +9,8 @@ import { CONFIG_PARAMS } from '../utility/hdbTerms.ts';
 import { convertToMS } from '../utility/common_utils.ts';
 import { when } from '../utility/when.ts';
 import { setTimeout as delay } from 'node:timers/promises';
-import { Transaction as RocksTransaction, type Store as RocksStore } from '@harperfast/rocksdb-js';
+import { Transaction as RocksTransaction, type Store as RocksStore, constants } from '@harperfast/rocksdb-js';
+const RETRY_NOW_VALUE = constants.RETRY_NOW_VALUE;
```
cb1kenobi, line 13: **Low (hardening): unguarded module-load capture of `RETRY_NOW_VALUE`.**

A normal successful `transaction.commit()` resolves to `undefined`, and the coordinated-retry check is `commitResult === RETRY_NOW_VALUE` (line 320). If the imported sentinel is ever `undefined` — e.g. rocksdb-js resolves to a build without this constant / version skew below 2.3.0 — the check becomes `undefined === undefined`, so *every* successful commit is misread as a retry signal: it increments `retries`, re-commits, and after 40 spurious rounds throws `ServerError`. The effect is that all writes fail, not that caching silently misbehaves. The `^2.3.0` pin makes this low-probability, but the correctness of the whole commit path hinges on this one import being defined; a load-time assertion turns a mysterious all-writes-fail into an immediate, clear boot error.

Suggested fix:
```suggestion
const RETRY_NOW_VALUE = constants.RETRY_NOW_VALUE;
if (RETRY_NOW_VALUE == null) throw new Error('@harperfast/rocksdb-js is missing constants.RETRY_NOW_VALUE; coordinatedRetry requires >= 2.3.0');
```

—
_Generated by Barber AI_

cb1kenobi: Eh, this is weak. Ignore.

PR #418, opened 2026-04-28
fix: Specify default dev node version

dawsontoth:
I added two lock files, because nvm doesn't respect .node-version, but they're identical. Renovate will suggest updates to both of them together, I believe, so it shouldn't be a burden for us.

This lets us use a consistent version of NPM and Node locally. We can switch around to other versions, but this gives a stable base when doing releases.

I didn't touch the matrix testing strategies that use multiple versions of node intentionally.

I also updated the lock because we've had drift.


.node-version
Ethan-Arrowood, line 1: fnm actually supports nvmrc (https://github.com/Schniz/fnm#features) so we can just stick to that unless others need .node-version 🤷‍♂️

Ethan-Arrowood (approved):

.node-version
```diff
@@ -0,0 +1 @@
+24.15.0
```
kriszyp, line 1: Renovate is configured to keep this updated to LTS?

kriszyp (approved):
Looks good, thank you

.node-version
dawsontoth: Yup!

dawsontoth: https://github.com/HarperFast/studio/pull/1176 for example

.node-version
dawsontoth: It should be fine to have both, doesn't hurt anything IMO 🤷

PR #421, opened 2026-04-28
chore: add workflow to sync package-lock.json when it drifts from pac…

kriszyp:
…kage.json


.github/workflows/format-check.yml
```diff
@@ -24,5 +28,16 @@ jobs:
       - name: Install dependencies
         run: npm install
 
-      - name: Run format check
-        run: npm run format:check
+      - name: Apply formatting
+        run: npm run format:write
+
+      - name: Commit fixes if needed
```
cb1kenobi: You may wanna borrow what we do in `rocksdb-js`: https://github.com/HarperFast/rocksdb-js/blob/main/.github/workflows/formatting.yml#L54-L64

kriszyp: Great steer, thank you! Adopted the rocksdb-js pattern across the board:

- **GitHub App token + PR-based fix** instead of direct auto-push. `sync-lockfile.yml` (and the new `fix-formatting.yml`) now generate a token via `actions/create-github-app-token` (`GH_APP_CLIENT_ID`/`GH_APP_PRIVATE_KEY`) and open a PR via `peter-evans/create-pull-request` — same as `rocksdb-js/formatting.yml` and this repo's own `update-rocksdb-js.yml`. App-token-authored commits re-trigger CI (default `GITHUB_TOKEN` ones don't), so the sync/format PRs actually run checks.
- **format-check.yml reverted to a read-only `format:check` gate.** The broken inline stage-write-push is gone; formatting auto-fix now lives in the scheduled `fix-formatting.yml` that opens a PR (mirrors `formatting.yml`). This also sidesteps the fork-PR write-token problem entirely, since the fix job never runs on `pull_request` events.
- Pinned to the same action SHAs this repo already uses (create-github-app-token v3.2.0, create-pull-request v8.1.1), gated PR creation on actual drift, and restored `npm ci` in lint-code.

— KrAIs

cb1kenobi (approved):
YES! THANK YOU FOR THIS PR!!!!

PR #415, opened 2026-04-28
Reproduce and fix hnsw race #386

kriszyp:
Reproduction and fix for https://github.com/HarperFast/harper/issues/386
Depends on https://github.com/HarperFast/rocksdb-js/pull/531
This is marked as ready for review, although it is not ready to merge.

PR #423, opened 2026-04-29
Handle error in outstandingCommit promise

cb1kenobi:
If a transaction fails due to `ERR_BUSY`, the retry mechanism returns a new promise from the error handler attached to the `commitResolution` promise. However the `finally()` on the `outstandingCommit` starts a new chain and the original rejected promise propagates the error to `outstandingCommit`. This error has already been handled, but we need to explicitly squetch it so that we don't get an unhandled rejection.

Supercedes https://github.com/HarperFast/harper/pull/415


kriszyp:
maybe rebase to main to fix those update.test.js/sql errors, and let's see how the unit tests look.

kriszyp (approved):
Looks great, nice catch! Ironically I just fixed this same exact type of error recently here https://github.com/HarperFast/harper-pro/commit/fea048cdbde8ba225f2f4200843d2ccdb216e832#diff-e707e24c4bc1646c60ac7da292c9707439ee0083f363f0433fda2f851c2b20d1R316

Ethan-Arrowood (approved):

PR #429, opened 2026-04-30
fix: watch dotfile parent directories in static file watcher

### LINKED ISSUE #196: harper dev does not serve static files when app path contains a dotfile directory
### Bug Summary

When running harper dev . from a directory whose path contains a dotfile directory (e.g. .claude), the static extension fails to serve files — all requests return 404. Other extensions (jsResource, fastifyRoutes, graphqlSchema, loadEnv) work correctly from the same path.

_No response_

### Steps to Reproduce

1. Create a Harper app with a `static` extension in `config.yaml`:
```
static:
  files: 'site/**'
  urlPath: 'app'
  index: true
  extensions: ['html']
```
2. Place HTML files in the `site/` directory
3. Run `harper dev .` from a path containing a dotfile directory, e.g.:
`/Users/me/my-app/.hidden-dir/worktrees/my-branch/`
4. Request `http://localhost:9926/app/` → 404
5. Request `http://localhost:9926/HealthCheck` (jsResource) → 200 ✅

### Expected Behavior

Static files are served at `/app/` regardless of the app's directory path.

### Actual Behavior

Static files return 404 when the path includes a dotfile directory. Other extensions load fine.

### Platform

Harper 4.7.20 (also reproduced on 4.7.3 and 4.7.19), Node.js v24.11.0, macOS

### Console Errors

_No response_

### Screenshots or Videos

_No response_

### Additional Context

Workaround: Run `harper dev .` from a path that doesn't contain dotfile directories (e.g. the main repo root rather than a git worktree under a hidden directory).

- File permissions are all standard (755 dirs, 644 files) — not a permissions issue
- `loadEnv` loads correctly from the same path (env conflict warnings confirm Harper finds the right directory)
- Fresh Harper install (removed data directory and reinstalled) — same result
- The same app code works perfectly when deployed to Harper Fabric
- The same app works with `harper dev .` from the main repo path (`/Users/me/my-app/`) which has no dotfile directories
- This was discovered using git worktrees located under `.claude/worktrees/`, a common pattern with Claude Code

No, just reporting the issue

kriszyp:
## Summary

- Add `dot: true` to chokidar options in `EntryHandler.ts`
- Without this, chokidar silently skips traversing parent directories that start with a dot (e.g. `.claude/worktrees/`), so the static extension's file map is never populated and every static request returns 404

## Purpose

Fixes #196 — `harper dev` fails to serve static files when the app path contains a dotfile directory (e.g. git worktrees under `.claude/worktrees/`, a common Claude Code pattern). Other extensions were unaffected because they don't rely on chokidar for their initial file discovery.

## Test plan

- [ ] Run `harper dev .` from a path containing a dotfile directory (e.g. `/path/.hidden/my-app/`) with a `static` extension configured
- [ ] Verify static files are served correctly at the configured `urlPath`
- [ ] Verify existing static serving still works from non-dotfile paths
- [ ] Verify dotfiles within the watched directory are still excluded (the custom `ignored` function handles exclusions, `dot: true` only affects parent-path traversal)

## Attention

The change is minimal (one line). The main question is whether `dot: true` has any unintended side effects — specifically, whether it could cause the watcher to pick up dotfiles within the `files` glob that should remain ignored. The custom `ignored` function already correctly gates on `allowedBases`, so only paths matching the configured `files` pattern will be watched regardless of `dot`.

### DIFF
--- a/components/EntryHandler.ts
+++ b/components/EntryHandler.ts
@@ -190,6 +190,7 @@ export class EntryHandler extends EventEmitter<EntryHandlerEventMap> {
 				cwd: this.#component.directory,
 				persistent: false,
 				followSymlinks: false,
+				dot: true,
 				ignored: (path) => {
 					const normalizedPath = path.replace(/\\/g, '/');
 					const normalizedBases = allowedBases.map((base) => base.replace(/\\/g, '/'));


components/EntryHandler.ts
```diff
@@ -190,6 +190,7 @@ export class EntryHandler extends EventEmitter<EntryHandlerEventMap> {
 				cwd: this.#component.directory,
 				persistent: false,
 				followSymlinks: false,
+				dot: true,
```
cb1kenobi, line 193: I think Claude is hallucinating. There is no `dot` option: https://github.com/paulmillr/chokidar/blob/main/src/index.ts.

`glob` has a `dot` option, but neither `chokidar` or this code use `glob`.

cb1kenobi (commented):
I'm not sure this is a valid issue. I have the exact opposite problem. Harper will absolutely restart the workers if any files in the `.next` folder are changed. This is actually one of my pain points. I needed to hack my local Harper to add `.next` to https://github.com/HarperFast/harper/blob/8333594f373172fe8a42f58a1cf71425e82f4590/server/threads/manageThreads.js#L562 so that it wouldn't restart when Next.js' hot module reloader recompiled a page.

kriszyp:
You're right to push back, @cb1kenobi — and it turns out the fix is wrong for a more fundamental reason: chokidar 4 accepts `dot` as an option but never reads it. The option is a no-op, so this PR would change nothing at runtime.

Your observation about `.next` being watched in `manageThreads.js` also shows that chokidar 4 has no inherent problem traversing dotfile directories — so the original diagnosis was wrong.

I'm going to close this PR while the actual root cause is investigated. The symptom is real (static 404s when the app path contains a dotfile ancestor), but the cause isn't in the `dot` option.

kriszyp:
Well, at least Claude automatically acknowledged the mistake (this is the real Kris Zyp here).

PR #430, opened 2026-04-30
fix: handle scalar root config values in OptionsWatcher

kriszyp:
## Summary

- `OptionsWatcher` threw `CannotSetPropertyError` whenever a plugin's YAML config value was a scalar (string, number, boolean) at the root level and the config file changed
- Fixed `#setValue` to handle `keys=[]` (root replacement) by directly assigning `#scopedConfig` instead of trying to index into it as a parent object
- Added `isDeepStrictEqual` guard in `#merge`'s non-Config shortcut path to prevent spurious `change` events when the chokidar `ready` event re-reads an unchanged scalar

## Root Cause

Two issues combined:
1. `#merge` short-circuits at the top level when either value is not a `Config` object, calling `#setValue([], newValue)`. `#setValue` had no handling for an empty key array — it traversed nothing, then checked `typeof obj !== 'object'` on the scalar `#scopedConfig` itself, throwing `CannotSetPropertyError` with a blank property name.
2. Without the equality guard, any file-watch re-read of a scalar config would also trigger a spurious `change` event.

## Test Plan

- [x] 3 new regression tests added: scalar→scalar, scalar→object, object→scalar
- [x] All 23 unit tests passing (`npx mocha unitTests/components/OptionsWatcher.test.js`)
- [x] Cross-model review (Gemini) confirmed fix is correct and edge cases are covered
- [x] Format and lint clean on changed files

## Attention

- The `keys=[]` path in `#setValue` is new behavior — callers that pass empty keys now get root-level replacement with a `change` event emitting `([], value, value)`. Listeners expecting a non-empty key path for all changes should be verified if they handle this case.
- Object→scalar transitions now emit a single `change([], scalar)` instead of per-key `undefined` events for the old object's keys. This is consistent with how type changes (e.g., array↔object) already worked.

### DIFF
--- a/components/OptionsWatcher.ts
+++ b/components/OptionsWatcher.ts
@@ -178,7 +178,9 @@ export class OptionsWatcher extends EventEmitter<OptionsWatcherEventMap> {
 		// First, ensure current and new config values are Config objects (not null, undefined, or a primitive)
 		if (!this.#isConfig(currentConfigValue) || !this.#isConfig(newConfigValue)) {
 			// If either is not a config, then just set as there is no need to diff/merge
-			this.#setValue(prevKeys, newConfigValue);
+			if (!isDeepStrictEqual(currentConfigValue, newConfigValue)) {
+				this.#setValue(prevKeys, newConfigValue);
+			}
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


kriszyp:
I am not entirely sure what kind of config leads to this, just saw the error.

cb1kenobi (approved):
I literally just ran into this problem. Appreciate you fixing this!

PR #433, opened 2026-04-30
yield every 100 entries in subscription catchup

harper-joseph:
## Summary

Reworks `Table.ts` subscribe replay logic to fix latent races at the boundary between historical replay and live-event delivery, add periodic event-loop yielding under load, and tighten consistency across the four replay shapes (`startTime` / `count` / `!omitCurrent` / `non-collection`).

## Bugs fixed

**Pre-subscribe event leak via `waitForDrain` yields.** Under realistic load (async-iterator consumer with `setImmediate` between events), the IIFE's sync prologue can fill `subscription.queue` past `EVENT_HIGH_WATER_MARK`, triggering `await waitForDrain()`. During that yield, `'committed'` callbacks fire for writes that committed before `subscribe()` was called but whose notification was still in flight. The original listener gate didn't filter these because `subscription.startTime` was `0` for `count` and `!omitCurrent`. Manifestation: occasional ghost events (e.g. delete events for keys not in current state, duplicates between count history and the listener path). Fixed by gating the listener with the latest record-time the cursor saw — captured in the audit log's own time domain so it works on both lmdb and rocksdb (see "Time-domain correctness" below).

**`subscription.startTime` undefined on rocksdb.** `auditRecord.localTime` is only populated on lmdb (set in the wrapped `auditStore.getRange`). The startTime branch's `subscription.startTime = auditRecord.localTime` left it undefined on rocksdb, making the listener gate a no-op. Fixed with the `localTime ?? version` fallback used elsewhere.

**`request.listener` attached after IIFE start.** The `if (request.listener) subscription.on('data', request.listener)` line was at the bottom of `subscribe()`, *after* the IIFE was invoked. Consumers passing `request.listener` got the same queue-fills-before-listener-attaches behavior that hits anyone using `on('data', ...)` after `await subscribe()`. Moved before the IIFE invocation.

## Time-domain correctness (count / !omitCurrent gate)

The audit log's record timestamps come from different sources on different backends: lmdb uses `getNextMonotonicTime()` (JS-side via `performance.now()`), but rocksdb uses native `transaction.getTimestamp()` from C++. These are NOT comparable — a JS-side gate value can be numerically smaller than a rocksdb-side audit timestamp from a strictly-earlier commit, making the gate filter incorrect.

To gate correctly on both backends, both branches now use a value drawn from the audit log itself (the cursor's own observed timestamps):

- **count**: after iteration, `subscription.startTime = history[0].localTime ?? history[0].version` (the latest record cursor saw).
- **!omitCurrent**: track `cursorMaxTime` across all primaryStore entries — including deletion tombstones (the update happens *before* the `if (!value) continue` so deleted-key timestamps still raise the gate). After iteration, `subscription.startTime = cursorMaxTime`.

In addition, `pendingRealTimeQueue` is filtered at the end of each branch to drop any in-flight pre-subscribe events the listener queued while `subscription.startTime` was still `0`:

```js
pendingRealTimeQueue = pendingRealTimeQueue.filter(
    (event) => (event.localTime ?? event.version) > cursorMaxTime
);
```

## Yielding under load

Added `REPLAY_YIELD_INTERVAL = 100`. Each replay loop calls `await rest()` (`new Promise(setImmediate)`) every 100 records via a shared `recordsSinceYield` counter. This:

- Lets `'committed'` callbacks fire periodically so `notifyFromTransactionData` can advance `lastTxnTime` during long replays
- Lets lmdb's `snapshot: false` cursor renew via the `setTimeout(0)` `resetReadTxn` chain
- Avoids event-loop starvation when replaying many records to a fast consumer
- Prevents the queue from filling past `EVENT_HIGH_WATER_MARK` before the consumer's listener attaches, sidestepping a latent `IterableEventQueue` bug where `on('data', ...)` drains the queue inline without emitting `'drained'` (so `waitForDrain()` would hang)

## Branch-by-branch changes

| Branch | Strategy |
|---|---|
| `isCollection && startTime` | New `dropDuringReplay = true; pendingRealTimeQueue = null` so the listener drops during replay; forward audit-log iteration with `snapshot: false` catches the live tail. No catchup sweep — the renewId chain (lmdb) and `_lastCommittedPosition` mmap (rocksdb) keep the cursor at or past `lastTxnTime`. `subscription.startTime` updated per-record with `localTime ?? version` fallback. Obsolete TODO removed. |
| `isCollection && count` | Cursor reverse-iterates audit log into `history`, sends oldest-first. After iteration, `subscription.startTime = history[0].localTime ?? history[0].version` (audit-log time domain, both-backend safe). `pendingRealTimeQueue` filtered to drop pre-subscribe in-flight events that leaked through while `startTime` was 0. |
| `isCollection && !omitCurrent` | Cursor key-order-iterates primaryStore and sends current state. Tracks `cursorMaxTime` across all entries (including deletion tombstones — update happens *before* `if (!value) continue`). After iteration, `subscription.startTime = cursorMaxTime` and `pendingRealTimeQueue` is filtered the same way. **Retained-message semantics**: a key updated during cursor iteration may legitimately reach the subscriber twice; both deliveries land at the same final state, so no dedupe. |
| non-collection | `subscription.startTime = localTime ?? entry?.version` set *before* the do-while (was after). |

The shared `pendingRealTimeQueue` queue-drain at the IIFE's tail now checks `if (pendingRealTimeQueue)` to handle the `null` set by the startTime branch.

## Listener change

Adds `if (dropDuringReplay) return;` early-out at the top of the `addSubscription` listener, used only by the startTime branch.

## Tests

New `unitTests/resources/subscriptionReplay.test.js` — 32 cases on lmdb, 23 + 9 pending on rocksdb (count branch's `start: 'z'` is a pre-existing rocksdb incompatibility).

Coverage:

- Per-branch happy-path: each replay shape delivers expected events in correct order
- Concurrent writes during replay: started before/during/after subscribe, all delivered exactly once (or, for `!omitCurrent`, each key reaches the subscriber with the correct final value — duplicates allowed)
- Race conditions: `subscribe()` while writes are in flight, both on a fresh database (first subscription) and after many existing subscriptions have advanced shared `lastTxnTime`
- Edge cases: `previousCount` larger than total, empty database, deleted (tombstoned) keys, future `startTime`, single-version records, subscribe-to-non-existent-record, mid-replay cleanup
- Both backends where applicable
- Chronological ordering: strict on lmdb; relaxed on rocksdb for in-flight tests where adjacent commits with sub-ms timestamps can land in the audit log out of order (backend property, not a subscription bug — every event still arrives exactly once)


resources/Table.ts
```diff
@@ -2702,14 +2727,24 @@ export function makeTable(options) {
...
+							subscription!.startTime = auditRecord.localTime ?? auditRecord.version; // update so we don't double send
 						}
+						// No catch-up sweep needed. With snapshot:false (lmdb), notifyFromTransactionData
+						// calls resetReadTxn before iterating, which bumps renewId; on the cursor's next
+						// .next() it renews to a fresh txn whose snapshot is at least as recent. With
+						// rocksdb, the audit-log iterator re-reads `_lastCommittedPosition` on each next()
+						// (live tail). Either way, at loop exit subscription.startTime is at or past
+						// lastTxnTime, and the gate in notifyFromTransactionData handles the handoff
+						// once dropDuringReplay flips back.
+						dropDuringReplay = false;
```
kriszyp: Maybe this should be in a `finally` in case of an error?

kriszyp (approved):
Looks great, nice work!

PR #440, opened 2026-05-01
Bump main to 5.1.0

cb1kenobi:
Bumping `main` to next minor.


heskew:
Would you want to bump the lock version while you're in here?

heskew (approved):

cb1kenobi:
> Would you want to bump the lock version while you're in here?

Good catch! Bumped!

dawsontoth (approved):

PR #445, opened 2026-05-01
fix: always set ITC originator from threadId; guard readerList with optional chaining

kriszyp:
## Summary

- **ITC originator (critical):** `sendItcEvent` had a `!isMainThread` guard that skipped stamping `originator` on outgoing messages when called from the main thread. Workers' `schemaHandler` calls `validateEvent`, which returns early on a missing originator — silently dropping the schema sync on all 8 workers. Removed the guard so `threadId` is always stamped (0 for main thread, which passes the `isEmpty` check).
- **readerList optional chaining (non-critical):** `getLMDBStats` called `.readerList()` unconditionally; some table types expose no such method, causing a recurring `[notify]`-level error during metrics collection every ~9 hours. Added `?.` to guard both the call and the subsequent `.split()`.

## Test plan

- [ ] Run `npm run test:unit:server` — all 292 tests pass including 3 new `sendItcEvent` tests
- [ ] Run `npm run test:unit:utility` — 12 systemInformation tests pass
- [ ] Verify in production that the ~9-hour `readerList is not a function` error stops appearing
- [ ] Verify that schema changes triggered from the main thread (monitoring job / scheduled schema op) are received and applied on all worker threads

### DIFF
--- a/utility/environment/systemInformation.ts
+++ b/utility/environment/systemInformation.ts
@@ -613,8 +613,8 @@ function getLMDBStats(table: Table, dbStats: DBStats): void {
 		const { root: _root, ...stats } = table.primaryStore.rootStore.getStats();
 		Object.assign(dbStats, stats);
 		dbStats.readers = table.primaryStore.rootStore
-			.readerList()
-			.split(/\n\s+/)
+			.readerList?.()
+			?.split(/\n\s+/)
 			.slice(1)
 			.map((line) => {
 				const [pid, thread, txnid] = line.trim().split(' ');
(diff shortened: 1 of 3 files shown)


utility/environment/systemInformation.ts
```diff
@@ -613,8 +613,8 @@ function getLMDBStats(table: Table, dbStats: DBStats): void {
 		const { root: _root, ...stats } = table.primaryStore.rootStore.getStats();
 		Object.assign(dbStats, stats);
 		dbStats.readers = table.primaryStore.rootStore
-			.readerList()
-			.split(/\n\s+/)
+			.readerList?.()
```
cb1kenobi, line 616: This should only be called from the LMDB code path. Is it being called somehow from a RocksDB path?!

cb1kenobi (approved):

PR #446, opened 2026-05-01
return a promise that resolves once the initial load of entries is co…

Ethan-Arrowood:
…mplete for graphqlSchema


heskew (approved):

kriszyp (approved):
Do you want this in a patch? (we can just add a `patch` label now, if you do).

PR #448, opened 2026-05-01
Only ignore nested node_modules dirs

cb1kenobi:
Fixes a regression introduced in #383 where plugins loaded from `node_modules` (like `@harperfast/nextjs`) couldn't watch their own files because the `ignored` function in `EntryHandler.ts` was filtering out any path containing `/node_modules`.

The fix was to change the check to only ignore nested `node_modules` directories.


Ethan-Arrowood (approved):

kriszyp (commented):
The PR looks good. But you really want this as a patch, even though it is a fix for a regression on a PR that wasn't a patch?

cb1kenobi:
> The PR looks good. But you really want this as a patch, even though it is a fix for a regression on a PR that wasn't a patch?
Whoops, didn't notice #383 was not a patch. Label removed!

kriszyp (approved):

PR #450, opened 2026-05-01
Read-only mode

### LINKED ISSUE #404: Read-only mode
To complement https://github.com/HarperFast/rocksdb-js/issues/490, add read-only mode support for Harper. Obviously this means opening RocksDB in read-only mode (LMDB actually has a read-only mode as well). From my quick tests, the most obvious additional thing is to disable analytics (analytics/write.ts has a nice big switch for this, I think). I'd guess we probably want to control this with an env variable, but other possibilities include checking for having a harper parent process, or trying write mode with fallback to read-only mode. The known "use case" is for when NextJS launches build processes (and we aren't directly making the spawn call, it is somewhere in the NextJS code).

cb1kenobi:
Adding support for a `HARPER_READONLY` environment variable and `--readonly` CLI flag to open RocksDB databases in read-only mode and avoids the dreaded `Error: IO error: While lock file: /Users/chris/harper/database/data/LOCK: Resource temporarily unavailable opening database /Users/chris/harper/database/data`.

Also exports a new `flushDatabases()` function on the public API.

Fixes https://github.com/HarperFast/harper/issues/404. Also addresses https://github.com/HarperFast/nextjs/pull/41.


heskew (commented):
Submitting inline feedback from Deep Review.

heskew:
## AI Review: Read-only mode implementation

....

🤖 _Generated by Gemini CLI_

kriszyp (commented):
Cool, this is great!

> Data Integrity (Crash Recovery):

I believe this is actually a really important thing here. RocksDB _often_ has much/most of its data just in memory. Using read-only mode successfully will actually usually require that you flush data and _then_ start the new harper process. I suppose this is actually executed in the nextjs plugin, but noting here. But maybe this might drive a need for having a public `flush()` method (and I guess maybe it kind of a top-level piece of functionality, ideally you want to flush everything before a new process because it is difficult to guess what the new process might access).

resources/databases.ts
```diff
@@ -33,6 +33,35 @@ import { RocksIndexStore } from './RocksIndexStore.ts';
...
+ * Read-only mode can be enabled via:
+ * - READONLY environment variable (truthy value)
+ * - --readonly CLI flag
+ * - storage.readOnly config setting
+ */
+let _isReadOnlyMode: boolean | undefined;
+export function isReadOnlyMode(): boolean {
+	if (_isReadOnlyMode !== undefined) return _isReadOnlyMode;
+	// Check environment variable
+	if (process.env.READONLY && process.env.READONLY !== '0' && process.env.READONLY !== 'false') {
```
Ethan-Arrowood: I mentioned this in another PR but I would love if we can have this prefixed. Like `HARPER_READONLY` or something like that.

kriszyp (approved):
Great!

heskew:
Filed two follow-ups for read-only-mode gaps surfaced in an earlier Gemini review (intentionally out of scope for this PR):

- #474 — proactive 405 at the Resource API layer instead of a 500 from the storage engine
- #475 — guard filesystem-backed management ops (component deploys, config writes, roles/users) so the read-only contract extends past RocksDB

heskew (approved):

PR #455, opened 2026-05-02
fix: reject legacy v4 'schema' field in DDL ops with 400 error

kriszyp:
## Summary

- Adds `rejectLegacySchemaField()` helper in `dataLayer/schema.js` that throws a `400 ClientError` when `schema` is provided without `database`
- Applied to all DDL entry points: `dropTable`, `dropSchema`, `createSchemaStructure`, `createTable`, `dropAttribute`, `createAttribute`
- 7 new unit tests covering rejection behavior and replication safety

## Purpose

In v5 the `database` field replaced v4's `schema` field for identifying the database in DDL operations. A caller using the legacy `schema` field received a 200 success response but the operation failed silently cluster-wide — `Promise.allSettled` in `replicateOperation` absorbs remote failures and the success message is always written over the response. There was no signal that anything went wrong.

## Design decision worth reviewing

**Check placement:** After Joi validation, before `transformReq()`. This is intentional — `transformReq` sets `req.schema = req.database` on the originating node before replication, so replicated calls always have **both** fields set; `rejectLegacySchemaField` only fires when `schema` is present *without* `database`, which uniquely identifies the legacy external API case.

**Skipped existing tests:** The `describe.skip('Test schema module')` block contains test constants (`DROP_TABLE_OBJECT_TEST`, etc.) that use `schema` without `database`. These would fail if unskipped, but updating them is a separate cleanup. They were already skipped before this change.

## Attention

- The `dropSchema` Joi schema uses `.or('database', 'schema')` for backwards compat; `rejectLegacySchemaField` then provides the more specific error. No change to the Joi schema was needed.
- Other operations (search, insert, delete) also call `transformReq` with `schema`/`database` but are out of scope — this was a DDL-specific report.

### DIFF
--- a/dataLayer/schema.ts
+++ b/dataLayer/schema.ts
@@ -19,6 +19,12 @@ import { transformReq } from '../utility/common_utils.ts';
 import { server } from '../server/Server.ts';
 import { cleanupOrphans } from '../resources/blob.ts';
 
+function rejectLegacySchemaField(obj: any) {
+	if (obj.schema && !obj.database) {
+		throw new ClientError("'database' is required. Use 'database' instead of the deprecated 'schema' field.");
+	}
+}
+
 const DB_NAME_CONSTRAINTS = Joi.string()
 	.min(1)
 	.max(commonValidators.schema_length.maximum)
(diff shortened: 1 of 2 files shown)


cb1kenobi (approved):

cb1kenobi (approved):
Is this considered a breaking change?

kriszyp:
Closing — rejecting the legacy v4 `schema` field outright is a breaking change for any v4-era client still sending it, and the upside (a cleaner 400) doesn't justify breaking those callers. If we want to nudge off the legacy field we should do it as a deprecation warning rather than a hard reject. — Claude (Opus 4.8), on behalf of Kris

PR #457, opened 2026-05-02
feat: upgrade msgpackr to v2 and use structon for random-access struct encoding

kriszyp:
## Summary

- Upgrades `msgpackr` from `1.11.9` to `^2.0.1`
- Adopts the `structon` package (the intended replacement for msgpackr v1's `randomAccessStructure` feature)
- `RecordEncoder` now extends `createStructon(Encoder)` instead of plain `Encoder`
- Removes `this.randomAccessStructure = true` from `OpenDBIObject` (the option was removed in msgpackr v2)

## Purpose

`msgpackr` v1's built-in `randomAccessStructure` option was removed in v2. The author extracted this functionality into the separate `structon` package. This change adopts structon to maintain random-access struct encoding (which enables efficient in-place field reads without full deserialization).

A bug in structon's `onLoadedStructures` was also fixed (in the structon repo): it previously only reloaded typed structures when loading shared structures from the DB, but not named structures. This caused `isCompatible` to always fail (named structure count mismatch), triggering an infinite encode retry loop. The fix reloads named structures into `this.structures`, unfreezes array elements (needed because `freezeData: true` froze them, but msgpackr lazily adds a `read` property), and clears transitions to force msgpackr to rebuild and update `lastNamedStructuresLength`.

## Test plan

- [x] `npm run test:unit:resources` — 336 passing, 5 pre-existing failures (blob test + 4 vector index tests, pre-existing before this change)
- [ ] Integration tests

## Attention areas

- The TypeScript cast `createStructon(Encoder) as typeof Encoder` is needed because `createStructon`'s return type needs to be an exact `typeof Encoder` for subclassing to work cleanly in TypeScript. The runtime behavior is correct — structon does return a subclass of the input.
- The `structon` dependency uses `file:../../structon` (local sibling directory) for now. Once structon 1.0.1 is published to npm (with the `onLoadedStructures` fix), this should be updated to `^1.0.1`.
- The structon fix is in a separate PR in the structon repo (feat/onloaded-named-structures-fix branch).


Ethan-Arrowood (approved):
if tests pass this is good with me!

cb1kenobi (approved):

PR #463, opened 2026-05-04
perf(hnsw): MinHeap for candidates, bisect for results, pre-compute query magnitude

kriszyp:
## Summary

This roughly doubles the performance of HNSW queries with cosine similarity. From my tests:
* 15-20% gains come from MinHeap insertion (should apply regardless of distance function)
* Remaining comes from pre-normalizing vector magnitude

Future improvement is to pre-normalize vectors into the database, is a little more involved.

Claude's summary:

- Replace `candidates.sort()` + `candidates.shift()` with a binary **MinHeap** — O(log n) push/pop instead of O(n log n) sort + O(n) shift per graph traversal step
- Replace `results.sort()` + `results.splice()` with **binary insertion** (`bisectInsert`) + `pop()` — maintains sorted order at O(log n) comparisons instead of re-sorting after every node
- **Pre-compute query vector magnitude** once at the top of `searchLayer` for cosine distance, eliminating one extra `sqrt` + inner summation loop per neighbor visit
- Add `benchmarks/hnsw-search.js`: standalone benchmark that builds an in-memory index and reports latency, throughput, nodes/query, and recall@10 vs brute force

## Purpose

Hot path optimization for `searchLayer`, which is called once per level per query and once per level per indexed insert. The sort-per-iteration pattern was the main algorithmic waste; the pre-computed magnitude eliminates repeated scalar work for the dominant (cosine) distance metric.

All changes are in-memory only — no data format changes, fully backward compatible with existing databases.

## Test plan

- [x] All 341 existing HNSW unit tests pass (`npm run test:unit:resources`)
- [x] Benchmark runs and reports recall@10 ≥ 85% at 2000×384 defaults
- [x] Cross-model review (Gemini) — verified heap invariants, bisect correctness, cosine math, and semantic equivalence to original traversal

## Review attention

- **`MinHeap.pop()`** when `data.length === 1`: `data.pop()` empties the array, then `if (this.data.length > 0)` is false so sinkDown is skipped — correct but non-obvious
- **Dimension mismatch in optimized cosine**: if `neighbor.vector.length > queryVector.length`, the inner loop reads `queryVector[i] === undefined` → `NaN`. The original had the same assumption. HNSW only stores consistent-dimension vectors, but worth noting.
- `furthestDistance` is captured once per outer iteration and held fixed through the inner neighbor loop — intentional, matches original behavior

## Benchmark usage

```
cd core && npm run build
node benchmarks/hnsw-search.js                   # 2000 × 384 dims × 50 queries
node benchmarks/hnsw-search.js 5000 768 100      # realistic embedding workload
```

To compare against a previous implementation, checkout the old branch, build, run the benchmark, then switch branches and compare.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (except the first part)

### DIFF
--- a/benchmarks/hnsw-search.js
+++ b/benchmarks/hnsw-search.js
@@ -1,0 +1,157 @@
+'use strict';
+/**
+ * HNSW searchLayer performance benchmark — measures search latency, throughput,
+ * nodes visited per query, and recall@K against brute-force ground truth.
+ *
+ * Run (from core/ directory, after npm run build):
+ *   node benchmarks/hnsw-search.js [N_VECTORS] [DIMS] [N_QUERIES]
+ *
+ * Examples:
+ *   node benchmarks/hnsw-search.js            # defaults: 2000 x 384 x 50
+ *   node benchmarks/hnsw-search.js 5000 768   # realistic embedding workload
+ */
+
+const { performance } = require('node:perf_hooks');
+const { HierarchicalNavigableSmallWorld } = require('#src/resources/indexes/HierarchicalNavigableSmallWorld');
+const { cosineDistance } = require('#src/resources/indexes/vector');
+
+const N_VECTORS = parseInt(process.argv[2]) || 2_000;
+const DIMS = parseInt(process.argv[3]) || 384;
+const N_QUERIES = parseInt(process.argv[4]) || 50;
+const TOP_K = 10;
+
+// ---------------------------------------------------------------------------
+// Minimal in-memory store — enough interface for HierarchicalNavigableSmallWorld
+// ---------------------------------------------------------------------------
+
+class MemoryStore {
+	constructor() {
+		this._map = new Map();
+		this.encoder = { useFloat32: null };
+	}
+	_key(k) {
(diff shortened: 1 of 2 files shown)


kriszyp:
**Additional commit: invMag caching for sqrt-free cosine hot path**

Added `invMag` (= `1/|vector|`) as a field on stored nodes. When present, `searchLayer`'s cosine distance inner loop becomes:

```
dot(queryVector, v) * invMagA * invMagB   (no sqrt)
```

vs. the previous:

```
dot(queryVector, v) * invMagA / sqrt(sum(v[i]²))
```

Also eliminates the second pass over `b` for `magBSq` — previously the combined loop did two multiply-accumulates per element; the fast path does one.

**Backward compatible:** nodes without `invMag` (legacy data) fall back to computing `sqrt(magBSq)` on the same code path. Migration is lazy — any write to a node sets `invMag` going forward.

> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

cb1kenobi (approved):

heskew (approved):

PR #469, opened 2026-05-05
fix(search): handle null/missing primary-store entries from HNSW results

sleekmountaincat:
## Problem

Two bugs in the `searchByIndex` custom-index path (`search.ts`), both triggered when HNSW search results are loaded from the primary store:

### Bug 1 — Transaction context not forwarded to `getEntry`

`getEntry(key)` was called without a transaction, so it read from a committed RocksDB snapshot. A record written earlier in the same transaction was invisible to a same-transaction vector search, returning no results when results were expected.

**Root cause:** `context?.transaction` is a `DatabaseTransaction` wrapper. Passing it directly to `getEntry` threw `TypeError: Invalid transaction` because `getTxnId` expects the raw RocksDB transaction (`options.transaction.id`). The correct call, consistent with `Table.ts:1591`, is `context?.transaction?.getReadTxn?.()`.

### Bug 2 — Null `loadedEntry` produced a silent partial entry

When a primary-store record was absent (deleted, expired, or HNSW node outliving its record), `getEntry` returned `null`. The spread `{ ...otherProps, ...null }` is valid JS and silently produces a stub with only `$distance` — no key, no value — that propagated downstream and could cause crashes (e.g. `ordered-binary` throwing "Unable to serialize object as a key").

**Fix:** return `SKIP` when `loadedEntry` is falsy. Also added a `key == null` guard for HNSW nodes missing a `primaryKey`.

## Changes

- [`resources/search.ts`](resources/search.ts) — 3-line fix in `searchByIndex`'s custom-index map callback
- [`unitTests/resources/vectorIndex.test.js`](unitTests/resources/vectorIndex.test.js) — new `describe` block with two tests:
  - *skips a deleted record instead of returning a broken partial entry* — validates no ghost or partial entries leak through
  - *write-then-search within the same transaction sees the written record* — fails without the fix, passes with it

## Test results

```
9 passing
0 failing
```

Without the fix: both new tests fail. With fix: all 9 pass.

### DIFF
--- a/resources/search.ts
+++ b/resources/search.ts
@@ -358,7 +358,9 @@ export function searchByIndex(
 				if (typeof entry === 'object' && entry) {
 					const { key, ...otherProps } = entry;
 					if (key == null) return SKIP; // primaryKey missing from HNSW node — skip rather than crash
-					const loadedEntry = Table.primaryStore.getEntry(key, { transaction: context && Table._readTxnForContext(context) });
+					const loadedEntry = Table.primaryStore.getEntry(key, {
+						transaction: context && Table._readTxnForContext(context),
+					});
 					if (!loadedEntry) return SKIP; // record was deleted/expired or not yet visible
 					Object.freeze(loadedEntry?.value);
 					recordRead(loadedEntry);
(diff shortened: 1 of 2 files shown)


kriszyp (approved):
Looks great!

PR #471, opened 2026-05-05
Set correct RocksDB size

cb1kenobi:
The `db_size` property returned from `describe_database` was reporting the wrong size for a RocksDB database.


heskew (approved):

kriszyp (approved):
Looks good!

PR #472, opened 2026-05-05
Update read_log to handle any log name

DavidCockerill:


validation/readLogValidator.js
```diff
@@ -44,20 +43,23 @@ function validateDatetime(value, helpers) {
...
+		return helpers.message(`'log_name' '${value}' is invalid.`);
+	}
+
+	const logName = ext === '.log' ? value : `${value}.log`;
 	const logPath = getConfigPath(hdbTerms.HDB_SETTINGS_NAMES.LOG_PATH_KEY);
-	const logName = value === undefined ? hdbTerms.LOG_NAMES.HDB : value;
 	const readLogPath =
 		logName === hdbTerms.LOG_NAMES.INSTALL
 			? path.join(INSTALL_LOG_LOCATION, hdbTerms.LOG_NAMES.INSTALL)
 			: path.join(logPath, logName);
```
heskew: The validator's `INSTALL_LOG_LOCATION` (line 13: `path.resolve(__dirname, '../logs')`) and `readLog.js`'s (`path.join(PACKAGE_ROOT, 'logs')`) resolve to different directories on a typical install — the validator's `existsSync` check passes/fails on a different path than what the reader actually opens. While here, both files now compute `logName` the same way too; a tiny shared helper would prevent future drift.

validation/readLogValidator.js
```diff
@@ -44,20 +43,23 @@ function validateDatetime(value, helpers) {
 }
 
 function validateReadLogPath(value, helpers) {
-	const processLogName = _.invert(hdbTerms.LOG_NAMES);
-	if (processLogName[value] === undefined) {
+	if (path.basename(value) !== value) {
```
heskew: On POSIX, `path.basename('..\\hdb.log') === '..\\hdb.log'` (backslash isn't a separator), so this passes. Harmless on POSIX (`\` is a literal char), but if a request reaches a Windows peer via `replicateOperation`, `\` *is* a separator there and the path could resolve outside `logPath`. Consider rejecting any value containing `/` or `\`, or requiring `path.posix.basename(value) === value && path.win32.basename(value) === value`.

validation/readLogValidator.js
```diff
@@ -44,20 +43,23 @@ function validateDatetime(value, helpers) {
...
 		return helpers.message(`'log_name' '${value}' is invalid.`);
 	}
 
+	const ext = path.extname(value);
+	if (ext && ext !== '.log') {
+		return helpers.message(`'log_name' '${value}' is invalid.`);
+	}
+
+	const logName = ext === '.log' ? value : `${value}.log`;
```
heskew, line 54: `path.extname` treats leading-dot strings as having no extension, so a few inputs produce surprising files: `'.log'` → `.log.log`, `'.config'` → `.config.log`, `'..'` → `...log`, `''` → `.log`. None escape `logPath`, but the operator gets a file they didn't ask for and the "does not exist" error echoes the *input*, not the resolved name. Worth tightening the regex (e.g. `/^[A-Za-z0-9_.-]+$/`, forbid `.` / `..`).

validation/readLogValidator.js
```diff
@@ -44,20 +43,23 @@ function validateDatetime(value, helpers) {
...
 	const readLogPath =
 		logName === hdbTerms.LOG_NAMES.INSTALL
 			? path.join(INSTALL_LOG_LOCATION, hdbTerms.LOG_NAMES.INSTALL)
 			: path.join(logPath, logName);
 
-	let exists = fs.existsSync(readLogPath);
-	if (exists) {
+	if (fs.existsSync(readLogPath)) {
 		return null;
 	}
```
heskew, line 60: `Joi.custom(fn)` uses the return value as the new validated value, so `return null` makes Joi's validated object `log_name: null`. Today `validationWrapper.validateBySchema` only reads `result.error`, so `request.log_name` isn't mutated — but if the wrapper ever switches to `Object.assign(object, result.value)`, `readLog.js`'s `request.log_name === undefined` check (line 43) becomes false and `path.extname(null)` throws. Pre-existing pattern; easy to fix while here: `return value;`.

unitTests/validation/readLogValidator.test.js
```diff
@@ -99,16 +99,42 @@ describe('Test readLogValidator module', () => {
...
+		sandbox.stub(fs, 'existsSync').returns(true);
+
+		const test_read_log_object = {
+			operation: 'read_log',
+			log_name: 'clustering_connector',
+		};
+
+		const result = read_log_validator(test_read_log_object);
+		expect(result).to.be.undefined;
+	});
```
heskew, line 116: Stubbing `existsSync` to return `true` unconditionally means this test doesn't verify the validator computes `<TEST_LOG_DIR>/clustering_connector.log` — exactly the path-resolution invariant that would catch validator/reader drift. Capture the stub and assert: `expect(existsStub.firstCall.args[0]).to.equal(path.join(TEST_LOG_DIR, 'clustering_connector.log'));`. A parallel test for `log_name: 'install'` would also exercise the `INSTALL_LOG_LOCATION` branch.

heskew (commented):
Drive-by deep review (the v5.0 branch doesn't carry the Claude review workflow yet). 5 findings, 2 worth addressing before merge:

- **`validation/readLogValidator.js`** — its `INSTALL_LOG_LOCATION` resolves to a different directory than `utility/logging/readLog.js`'s, so an `install`/`install.log` request can pass validation and then ENOENT in the reader (or vice versa).
- **`unitTests/validation/readLogValidator.test.js`** — the new happy-path test stubs `existsSync` unconditionally and doesn't assert the resolved path, so it can't catch the class of regression above.

Three smaller hardening notes inline (Windows backslash on POSIX, leading-dot `extname` edge cases, Joi `custom()` returning `null`). (🤖)

kriszyp (commented):
Looks good, but can you rebase against `main`?

DavidCockerill:
rebase messed up

PR #476, opened 2026-05-05
Copy database in binary mode

cb1kenobi:
I created an database using LMDB storage and then inserted a bunch of rows. Next I ran `harper copy-db testdb testdb2` and it errored with:
```
[main/0] [error]: Reading audit entry error RangeError: Error reading float64: Offset is outside the bounds of the DataView at position 4
```

When opening the source and target databases, set the encoding to binary so that we completely bypass decoding/encoding and do a true byte copy.

### DIFF
--- a/bin/copyDb.ts
+++ b/bin/copyDb.ts
@@ -200,9 +200,18 @@ export async function copyDb(sourceDatabase: string, targetDatabasePath: string)
 			await copyDbi(sourceDbi, targetDbi, isPrimary, transaction);
 		}
 		if (sourceAuditStore) {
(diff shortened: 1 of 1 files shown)


kriszyp (commented):
The RocksDB record and LMDB records don't really have the same binary format. The RocksDB record starts with the version timestamp (because we decided to not include that natively like lmdb-js does), and then LMDB typically has encoding of id/timestamp reference into the audit log. So I wouldn't think this would work.
I would think the LMDB audit log error probably needs to be addressed regardless.

cb1kenobi:
Abandoning this.

PR #477, opened 2026-05-06
fix(blob): retry bytes() up to 3x when lock-free but file is incomplete

kriszyp:
## Summary

Adds retry logic in `bytes()` when `tryLock` returns `true` but the blob file is still incomplete — a state that, given the locking invariant, can only occur after a process crash.

## The locking invariant

In `writeBlobWithStream`, the lock is acquired **before** the file is created or any data written, and released only **after** both content and the final size header are fully written to disk. Therefore in normal operation, a reader always sees one of two consistent states:
- Lock held → write in progress (reader waits via callback)
- Lock free → file is complete (or doesn't exist)

**Lock-free + incomplete file can only happen after a process crash** that cleared in-memory lock state while leaving a partial file on disk.

## What the retry does and doesn't do

**What it does:** After detecting a crash-incomplete file, retries up to 3× at 100ms intervals (300ms total) before throwing, instead of throwing immediately.

**What it does NOT do:** Recover from the crash. After a crash, replication re-sends the blob data to a **new local fileId** — it never re-writes the old partial fileId. So all 3 retries read the same incomplete file and eventually throw `"Incomplete blob"` regardless.

The practical effect is a 300ms delay before the error surfaces instead of immediate failure. Serent-Canopy's 200ms application-level retry (which re-reads the LMDB record and gets the new fileId) is still required for actual recovery.

## Is this worth keeping?

Arguments for: avoids immediate throw in edge cases where there might be a race not yet fully understood; gives a brief window if any future scenario warrants it.

Arguments against: 300ms latency on an error path that will always fail in the known crash scenario; the retry is misleading about recovery capability.

Leaving this open for discussion — I can drop the retry entirely and just leave the error throw unchanged if that's preferred.

## Test

The test correctly simulates the crash state (file written then corrupted with DEFAULT_HEADER + partial content, `isSaving` awaited to ensure the lock is free), and verifies both the rejection and the ≥290ms timing of 3 retries.

### DIFF
--- a/.github/workflows/cherry-pick-patch.yml
+++ b/.github/workflows/cherry-pick-patch.yml
@@ -16,6 +16,7 @@ on:
 permissions:
   contents: write
   pull-requests: write
+  actions: write
 
 jobs:
   cherry-pick:
@@ -121,6 +122,7 @@ jobs:
           if git cherry-pick $PICK_FLAGS $PICK_SHAS; then
             git push origin "$RELEASE_BRANCH"
             echo "✅ Cherry-picked PR #$PR_NUMBER onto $RELEASE_BRANCH"
+            gh workflow run integration-tests.yml --ref "$RELEASE_BRANCH" --repo "$GITHUB_REPOSITORY"
           else
             echo "⚠️ Cherry-pick conflict — creating resolution branch"
             git cherry-pick --abort 2>/dev/null || true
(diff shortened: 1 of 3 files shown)


kriszyp:
I challenged Claude on this one, and after further research admitted it is not sure if this will really help. Probably will close.

PR #494, opened 2026-05-07
fix: Restore getRecordCount sampling early-exit on RocksDB

Devin-Holland:
## Summary

`Table.getRecordCount` reads `entryCount` from `primaryStore.getStats()`, which is an LMDB field name. On RocksDB stores `getStats()` returns keys like `rocksdb.estimate-num-keys` instead, so on every RocksDB table the lookup was `undefined`, `halfway = Math.floor(undefined / 2) = NaN`, and the `entriesScanned < halfway` guard on the 500ms early-exit was always false. The sampling estimator never engaged — every `describe_all` / `describe_table` silently iterated and decoded the entire table.

Reproduced live on a Harper Pro 5.0.10 instance with a 1.46M-row RocksDB table: `describe_all` consistently ~18-20s, returning exact counts every time. CPU profile during the call was dominated by `rocksdb-js next` and per-record msgpackr decoding inside the `getRecordCount` loop.

The fix branches on `isRocksDB` and reads `rocksdb.estimate-num-keys` via `getDBIntProperty`, the same pattern already used by `getSize` a few lines above.

## What changed

- `resources/Table.ts` — read `entryCount` correctly on RocksDB; accept `options.timeLimit` so the early-exit path is reachable deterministically from tests.
- `unitTests/resources/getRecordCount.test.js` — new regression suite. The `timeLimit: 0` test fails without the source fix and passes with it.

## Where to look

- The `?? 0` fallback in case `getDBIntProperty` ever returns null — keeps `halfway` numeric (0 means \"loop never breaks early on an empty table,\" same effective behavior as before for a 0-row table).
- `options.timeLimit` is an undocumented escape hatch added for testability. If we'd rather keep `getRecordCount`'s public surface untouched, I can move the test to use a different mechanism, but this seemed lowest-impact.
- I did not change the estimator math itself; the `entryCount` value also feeds the variance / CI calculation downstream, which was previously NaN-poisoned in the unreachable branch.

### DIFF
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -3252,8 +3252,10 @@ export function makeTable(options) {
 		}
 		static async getRecordCount(options?: any) {
 			// iterate through the metadata entries to exclude their count and exclude the deletion counts
-			const entryCount = primaryStore.getStats().entryCount;
-			const TIME_LIMIT = 1000 / 2; // one second time limit, enforced by seeing if we are halfway through at 500ms
+			const entryCount = isRocksDB
+				? (primaryStore.getDBIntProperty('rocksdb.estimate-num-keys') ?? 0)
+				: primaryStore.getStats().entryCount;
+			const TIME_LIMIT = options?.timeLimit ?? 1000 / 2; // one second time limit, enforced by seeing if we are halfway through at 500ms
 			const start = performance.now();
 			const halfway = Math.floor(entryCount / 2);
 			const exactCount = options?.exactCount;
--- a/unitTests/resources/getRecordCount.test.js
+++ b/unitTests/resources/getRecordCount.test.js
@@ -1,0 +1,48 @@
+require('../testUtils');
+const assert = require('node:assert/strict');
+const { setupTestDBPath } = require('../testUtils');
+const { table } = require('#src/resources/databases');
+const { setMainIsWorker } = require('#js/server/threads/manageThreads');
+
+describe('Table.getRecordCount', () => {
+	let RecordCountTable;
+
+	before(async function () {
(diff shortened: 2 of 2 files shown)


cb1kenobi (approved):
This looks good, but `search.ts` also calls `getStats().entryCount`: https://github.com/HarperFast/harper/blob/a4086b3c5e3c0f3db34527f87d2d2f5b0b381f0b/resources/search.ts#L1297

cb1kenobi:
Also here: https://github.com/HarperFast/harper/blob/main/resources/indexes/HierarchicalNavigableSmallWorld.ts#L715
and
https://github.com/HarperFast/harper/blob/main/resources/Table.ts#L3255

Devin-Holland:
> Also here: https://github.com/HarperFast/harper/blob/main/resources/indexes/HierarchicalNavigableSmallWorld.ts#L715 and https://github.com/HarperFast/harper/blob/main/resources/Table.ts#L3255

That last one is the one that's fixed.

Devin-Holland:
> This looks good, but `search.ts` also calls `getStats().entryCount`:
> 
> https://github.com/HarperFast/harper/blob/a4086b3c5e3c0f3db34527f87d2d2f5b0b381f0b/resources/search.ts#L1297

// use getStats for LMDB because it is fast path, otherwise RocksDB can handle fast path on its own
store.estimatedEntryCount = store.readerCheck ? store.getStats().entryCount : store.getKeysCount();
		
This one has a comment about RocksDB, is `store.readerCheck` causing it to fall into the else of the ternary?

kriszyp (approved):

PR #496, opened 2026-05-07
Use RocksDB's `estimate-num-keys` instead of non-existent `entryCount`

cb1kenobi:
Builds on https://github.com/HarperFast/harper/pull/494 and fixes the other instances that reference `entryCount`.

### DIFF
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -961,7 +961,7 @@ export function makeTable(options) {
 					database: databaseName,
 					auditSize:
 						auditStore instanceof RocksDatabase
-							? (auditStore.getDBIntProperty('rocksdb.estimate-num-keys') ?? 0)
+							? auditStore.getKeysCount()
 							: auditStore?.getStats().entryCount,
 					attributes,
 					recordCount: undefined,
@@ -3232,7 +3232,7 @@ export function makeTable(options) {
 		 */
 		static getSize() {
 			if (isRocksDB) {
-				return primaryStore.getDBIntProperty('rocksdb.estimate-live-data-size') ?? 0;
+				return primaryStore.getKeysCount();
 			}
 			const stats = primaryStore.getStats();
 			return (stats.treeBranchPageCount + stats.treeLeafPageCount + stats.overflowPages) * stats.pageSize;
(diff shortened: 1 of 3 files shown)


resources/Table.ts
```diff
@@ -3252,8 +3255,10 @@ export function makeTable(options) {
 		}
 		static async getRecordCount(options?: any) {
 			// iterate through the metadata entries to exclude their count and exclude the deletion counts
-			const entryCount = primaryStore.getStats().entryCount;
-			const TIME_LIMIT = 1000 / 2; // one second time limit, enforced by seeing if we are halfway through at 500ms
+			const entryCount = isRocksDB
+				? (primaryStore.getDBIntProperty('rocksdb.estimate-num-keys') ?? 0)
```
kriszyp: There is an `exactCount` option, and I don't think we should use the estimate when an `exactCount` is explicitly requested.

kriszyp (approved):
Honoring the exactCount might take a bit more work, so if you are eager to get this out in a patch (if so, make sure to label it), happy for follow work to come later (to a PR that is follow up work :) )

resources/Table.ts
```diff
@@ -3229,7 +3232,7 @@ export function makeTable(options) {
 		 */
 		static getSize() {
 			if (isRocksDB) {
-				return primaryStore.getDBIntProperty('rocksdb.estimate-live-data-size') ?? 0;
+				return primaryStore.getKeysCount();
```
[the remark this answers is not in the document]
cb1kenobi: Fixed.

resources/Table.ts
cb1kenobi: Done.

Instead of getting `rocksdb.estimate-num-keys`, we can call `primaryStore.getKeysCount()` without a start and end. Under the hood, `Database::GetCount()` returns `rocksdb.estimate-num-keys` when there is no start and end.

If we set the `start` to `null`, It sets the `Slice` to `<Buffer 00>` which is the null terminator, an empty string, and bypasses the estimate and performs the slow iteration path.

cb1kenobi: I created a PR to update the `getKeysCount()` docs: https://github.com/HarperFast/rocksdb-js/pull/547.

DavidCockerill (approved):

kriszyp (approved):
Nice!

PR #503, opened 2026-05-11
fix: backfill schemaDefined on replicas missing the flag

ldt1996:


cb1kenobi (approved):
I don't know much about this, but makes sense. It would be nice if it was easy to add a test that demonstrates this fixes things.

kriszyp:
What is the situation where this can occur? I thought line 907 in databases.ts consistently set this flag.
And reminder: please try to use /harper-engineering-guidelines skill to file PRs, this should automatically ensure there is an AI review prior to the PR being filed.

ldt1996:
@kriszyp I think that line just sets the local variable... from what I can tell it only gets persisted via the new table branch, and the existing table branch doesn't seem to rewrite the descriptor, so once a replica ends up on disk without the flag the defaulting probably can't reach back. My guess is that's what 4.7.x replicated=true deploys were producing

PR #516, opened 2026-05-12
fix: Lookup hostname in logs properly

dawsontoth:
This regressed at some point, and I'm not entirely sure why. But with this change, here's the change I see.

Before:
```
Hostname:           undefined <----
Worker Threads:     1
Root Path:          /Users/dawson/hdb
Debugging:          enabled: true
Logging:            level: info, location: /Users/dawson/hdb/log/hdb.log, stdout/err
Default:            HTTP (and WS): 9926, CORS: enabled for *
Operations API:     HTTP: 9925, CORS: enabled for *, unix socket: /Users/dawson/hdb/operations-server
                    http://undefined:9925/ <----
MQTT:               TCP: 1883, TLS: 8883, WS: 9926
REST:               HTTP: 9926, WS: 9926
                    http://undefined:9926/ <----
```

After:
```
Hostname:           localhost <----
Worker Threads:     1
Root Path:          /Users/dawson/hdb
Debugging:          enabled: true
Logging:            level: info, location: /Users/dawson/hdb/log/hdb.log, stdout/err
Default:            HTTP (and WS): 9926, CORS: enabled for *
Operations API:     HTTP: 9925, CORS: enabled for *, unix socket: /Users/dawson/hdb/operations-server
                    http://localhost:9925/ <----
MQTT:               TCP: 1883, TLS: 8883, WS: 9926
REST:               HTTP: 9926, WS: 9926
                    http://localhost:9926/ <----
```

Should we apply this elsewhere? I don't know, maybe? I use these log statements constantly to pop open a browser, that's why I noticed.

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -25,6 +25,7 @@ const keys = require('../security/keys.js');
 const { startHTTPThreads } = require('../server/threads/socketRouter.ts');
 const hdbInfoController = require('../dataLayer/hdbInfoController.js');
 const { isReadOnlyMode } = require('../resources/databases.ts');
+const { getThisNodeName } = require('../server/nodeName.ts');
 const hdbTerms = require('../utility/hdbTerms.ts');
 const { getHdbPid, isProcessRunning } = require('../utility/processManagement/processManagement.js');
 const { PACKAGE_ROOT } = require('../utility/packageUtils');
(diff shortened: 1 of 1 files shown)


cb1kenobi (approved):

kriszyp (approved):
Nice. I have definitely noticed this and was annoyed, but hadn't tracked down the fix. Thank you!

PR #517, opened 2026-05-12
Fix invalid `this` reference

cb1kenobi:
I hope this fixes the `TypeError: Cannot read properties of undefined (reading 'encoder') at <anonymous> (/home/harperdb/.npm-global/lib/node_modules/@harperfast/harper-pro/replication/replicator.ts:166:40)` error.


kriszyp (approved):
Looks good, nice catch

PR #518, opened 2026-05-12
fix(migrate): skip Symbol.for('structures') entries in copyDbiToRocks

ldt1996:
LMDB→RocksDB migration was copying `Symbol.for('structures')` entries from source DBIs into target primary column families. lmdb-js stores msgpackr's shared-structures dict at that symbol as a regular DBI entry, so `sourceDbi.getRange()` yielded it alongside real records. The copied entry sat in the same iterable keyspace as records, with a malformed metadata prefix — any post-migration unfiltered scan tripped over it. In production this broke `subscribeToNodeUpdates` with `Data read, but end of buffer not reached 109`.

Fix: skip symbol-keyed entries in `copyDbiToRocks`. The structures dict doesn't need to migrate — production rebuilds it on first encode via `RecordEncoder.saveStructures` → rootStore composite key.

## Test plan

- [x] Added `widgets` fixture (60 records, 6-field shape) to `integrationTests/upgrade/4.x-upgrade.test.ts` — enough for msgpackr to populate the source's shared-structures dict.
- [x] Added a `getKeys()` walk on the post-migration primary CF asserting no symbol keys remain. Verified: fails on `main`, passes with this fix.
- [ ] Preexisting unrelated flake: `downgrade and start` fails with `ECONNREFUSED 127.0.0.1:9925`. Reproduces on `main`.


bin/copyDb.ts
```diff
@@ -385,6 +385,16 @@ async function copyDbToRocks(sourceRootStore, sourceDatabase: string, targetPath
 		name: INTERNAL_DBIS_NAME,
 	});
 
+	const STRUCTURES_KEY = Symbol.for('structures');
```
cb1kenobi, line 388: `rocksdb-js` accepts a `sharedStructuresKey` option when opening the database. I wonder if you need to pass it into the `openRocksDb()` functions above. I ask because I don't know if Harper internally calls `getStructures()` and `saveStructures()` during the copy.

ldt1996: I'm not at my laptop rn, but will check in 1h

kriszyp (commented):
I am confused, it looks like you are editing the LMDB's `copyDb` function instead of the RocksDB's `migrateOnStart` function. `copyDb` should be when LMDB is the target, `migrateOnStart` should be used when RocksDB is the target. (Maybe the naming needs to be improved?)

ldt1996:
I think the naming needs improvements indeed, `copyDbToRocks` is what `migrateOnStart` calls afaik

ldt1996:
marking this as a draft cause it's not tested, and I need to adress Chris' comment as well

kriszyp (approved):

cb1kenobi (approved):

PR #527, opened 2026-05-13
feat: harper login and logout

dawsontoth:
This saves jwts, per target, in your ~/.harper directory under credentials.json. There are likely common bits that are already implemented that Junie missed, and wasted time recreating. It got a bit creative with the readline stuff too, I think that could be tighter, but the experience is decent.

Will follow up with a documentation PR and adjustments to create-harper if we like this pattern.

To verify this, I `npm link`'d harper, and went through various use cases.

```
git clone git@github.com:HarperFast/local-inference.git
cd local-inference
harper login
harper deploy
harper logout
```

### DIFF
--- a/bin/cliCredentials.ts
+++ b/bin/cliCredentials.ts
@@ -1,0 +1,133 @@
+import { getHomeDir } from '../utility/common_utils.ts';
+import fs from 'node:fs';
+import path from 'node:path';
+
+const ownerRWDenyAllOthers = 0o600;
+
+interface TargetedCredentials {
+	last_target: string | null;
+	targets: {
+		[target: string]: Tokens;
+	};
+}
+
+interface Tokens {
+	operation_token: string;
+	refresh_token: string;
+}
+
+/**
+ * Normalizes a target operations API URL to a canonical form (with trailing slash).
+ */
+export function normalizeTarget(target: string): string {
+	if (!target) return target;
+	let normalized = target;
+	if (!normalized.startsWith('http://') && !normalized.startsWith('https://')) {
+		normalized = 'https://' + normalized;
+	}
+	try {
+		const url = new URL(normalized);
+		if (!url.port && !normalized.includes(':', normalized.indexOf('://') + 3)) {
+			url.port = '9925';
+		}
+		normalized = url.toString();
+	} catch {
+		// If it's not a valid URL yet, we'll let it be handled later or it will fail
+	}
+	if (!normalized.endsWith('/')) {
+		normalized += '/';
+	}
+	return normalized;
+}
+
+/**
+ * Loads the JWT credentials from the ~/.harperdb/credentials.json file.
+ */
+export function loadCredentials(): TargetedCredentials {
+	const credentialsFile = getCredentialsFile();
+	try {
+		return JSON.parse(fs.readFileSync(credentialsFile, 'utf8'));
+	} catch (err) {
+		if (err.code !== 'ENOENT') {
+			throw new Error(`Error reading credentials file: ${err.message}`);
+		}
+	}
+	return {
+		last_target: null,
+		targets: {},
+	};
+}
+
+/**
+ * Saves the JWT credentials to the ~/.harperdb/credentials.json file.
+ */
+export function saveCredentials(target: string, tokens: Tokens): void {
+	if (!target) {
+		throw new Error('Target is required to save credentials.');
+	}
+
+	target = normalizeTarget(target);
+	const allCredentials = loadCredentials();
+
+	allCredentials.targets ||= {};
+	allCredentials.targets[target] = tokens;
+	allCredentials.last_target = target;
+
+	try {
+		fs.mkdirSync(getCredentialsDir(), { recursive: true });
+		fs.writeFileSync(getCredentialsFile(), JSON.stringify(allCredentials, null, 2), { mode: ownerRWDenyAllOthers });
+	} catch (err) {
+		throw new Error(`Error saving credentials file: ${err.message}`);
+	}
+}
+
+/**
+ * Deletes the credentials for a specific target or all if no target provided.
+ */
+export function clearCredentials(target: string): void {
+	const credentialsFile = getCredentialsFile();
+	if (target) {
+		target = normalizeTarget(target);
+		const allCredentials = loadCredentials();
+		if (allCredentials && allCredentials.targets) {
+			if (allCredentials.targets[target]) {
+				delete allCredentials.targets[target];
+			} else {
+				console.error(`No credentials found for ${target}`);
+				process.exit(1);
+			}
+
+			if (allCredentials.last_target === target) {
+				allCredentials.last_target = null;
+			}
+			try {
+				fs.writeFileSync(credentialsFile, JSON.stringify(allCredentials, null, 2), {
+					mode: ownerRWDenyAllOthers,
+				});
+				console.log(`Logged out from ${target}`);
+			} catch (err) {
+				throw new Error(`Error clearing credentials file: ${err.message}`);
+			}
+		} else {
+			console.error(`No credentials found for ${target}`);
+			process.exit(1);
+		}
+	} else if (fs.existsSync(credentialsFile)) {
+		try {
+			fs.unlinkSync(credentialsFile);
+			console.log('Logged out from all targets');
+		} catch (err) {
+			throw new Error(`Error clearing credentials file: ${err.message}`);
+		}
+	} else {
+		console.log('No credentials found to clear.');
+	}
+}
+
+function getCredentialsFile(): string {
+	return path.join(getHomeDir(), '.harperdb', 'credentials.json');
+}
+
+function getCredentialsDir(): string {
+	return path.join(getHomeDir(), '.harperdb');
+}
--- a/bin/cliOperations.ts
+++ b/bin/cliOperations.ts
@@ -1,5 +1,7 @@
 'use strict';
 
+import { loadCredentials, saveCredentials, normalizeTarget } from './cliCredentials.ts';
+import { isJWTExpired } from '../security/tokenAuthentication.ts';
 import * as envMgr from '../utility/environment/environmentManager.ts';
 envMgr.initSync();
 import * as terms from '../utility/hdbTerms.ts';
@@ -57,14 +59,31 @@ function buildRequest(): any {
 }
 
 /**
+ * Resolves the target URL from various sources.
+ * @param {Object} req The request object.
+ * @param {Object} allCredentials Stored credentials.
+ * @returns {string|null} The resolved target URL.
+ */
+function resolveTarget(req, allCredentials) {
+	return (
+		req.target ||
+		process.env.HARPER_CLI_TARGET ||
+		process.env.CLI_TARGET ||
+		(allCredentials && allCredentials.last_target)
+	);
+}
+
+/**
  * Using a unix domain socket will send a request to hdb operations API server
  * @param req
+ * @param skipResponseLog By default, the response is logged to the console. Set this to true to skip logging it, which can be useful for sensitive responses like login calls!
  * @returns {Promise<void>}
  */
-async function cliOperations(req: any) {
-	if (!req.target) {
-		req.target = process.env.HARPER_CLI_TARGET || process.env.CLI_TARGET;
-	}
+async function cliOperations(req: any, skipResponseLog = false) {
+	require('dotenv').config();
+
+	const allCredentials = loadCredentials();
+	req.target = normalizeTarget(resolveTarget(req, allCredentials));
 	let target;
 	if (req.target) {
 		try {
@@ -76,6 +95,7 @@ async function cliOperations(req: any) {
 				throw error;
 			}
 		}
+		const resolvedTarget = req.target;
 		target = {
 			protocol: target.protocol,
 			hostname: target.hostname,
@@ -83,8 +103,9 @@ async function cliOperations(req: any) {
 			username: req.username || target.username || process.env.HARPER_CLI_USERNAME || process.env.CLI_TARGET_USERNAME,
 			password: req.password || target.password || process.env.HARPER_CLI_PASSWORD || process.env.CLI_TARGET_PASSWORD,
 			rejectUnauthorized: req.rejectUnauthorized,
+			resolvedTarget,
 		};
-		console.error(`Connecting to ${target.protocol}//${target.hostname}:${target.port}`);
+		console.error(`Connecting to ${resolvedTarget}`);
 	} else {
 		// if we aren't doing a targeted operation (like deploy), we initialize the config and verify that local harper
 		// is running and that we can communicate with it.
@@ -110,6 +131,47 @@ async function cliOperations(req: any) {
 		options.headers = { 'Content-Type': 'application/json' };
 		if (target?.username) {
 			options.headers.Authorization = `Basic ${Buffer.from(`${target.username}:${target.password}`).toString('base64')}`;
+		} else if (allCredentials) {
+			let tokens = null;
+			let lookupKey = null;
+			if (target && allCredentials.targets) {
+				lookupKey = target.resolvedTarget;
+				tokens = allCredentials.targets[lookupKey] ?? null;
+			}
+
+			if (tokens?.operation_token) {
+				if (tokens.refresh_token && isJWTExpired(tokens.operation_token)) {
+					console.error('Operation token expired, attempting to refresh...');
+					try {
+						const refreshOptions = { ...options };
+						refreshOptions.headers = { ...options.headers, Authorization: `Bearer ${tokens.refresh_token}` };
+						const refreshResponse = await httpRequest(refreshOptions, {
+							operation: 'refresh_operation_token',
+						});
+						if (refreshResponse.statusCode === 200) {
+							const refreshData = JSON.parse(refreshResponse.body);
+							if (refreshData.operation_token) {
+								tokens.operation_token = refreshData.operation_token;
+								saveCredentials(lookupKey || target?.resolvedTarget, {
+									operation_token: tokens.operation_token,
+									refresh_token: tokens.refresh_token,
+								});
+								console.error('Operation token refreshed successfully.');
+								// Update the original request's authorization header with the new token
+								options.headers.Authorization = `Bearer ${tokens.operation_token}`;
+							}
+						} else if (refreshResponse.statusCode === 401) {
+							console.error('Refresh token expired or invalid. Please run harper login again.');
+							process.exit(1);
+						} else {
+							console.error(`Failed to refresh operation token: ${refreshResponse.statusCode}`);
+						}
+					} catch (refreshErr) {
+						console.error(`Error refreshing operation token: ${refreshErr.message}`);
+					}
+				}
+				options.headers.Authorization = `Bearer ${tokens.operation_token}`;
+			}
 		}
 		if (req.cborEncode) {
 			options.headers['Content-Type'] = 'application/cbor';
@@ -141,7 +203,13 @@ async function cliOperations(req: any) {
 			process.exit(1);
 		}
 
-		console.log(responseLog);
+		if (!skipResponseLog) {
+			console.log(responseLog);
+		}
+
+		if (target) {
+			responseData.resolvedTarget = target.resolvedTarget;
+		}
 
 		return responseData;
 	} catch (err) {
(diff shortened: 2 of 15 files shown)


kriszyp:
Should we put it in `.harperdb` since we already create and use that directory? (I know we eventually want to rename that one too, but for the time being seems like maybe it is preferable to have a single "hidden" directory?)

bin/cliCredentials.js
```diff
@@ -0,0 +1,112 @@
+'use strict';
+
+const path = require('node:path');
+const os = require('node:os');
+const fs = require('fs-extra');
+
+function getHomeDir() {
```
cb1kenobi: Can we use the `getHomeDir()` in `utility/common_utils.js`?

bin/cliOperations.ts
```diff
@@ -56,15 +57,29 @@ function buildRequest() {
...
+/**
+ * Resolves the target URL from various sources.
+ * @param {Object} req The request object.
+ * @param {Object} allCredentials Stored credentials.
+ * @returns {string|null} The resolved target URL.
+ */
+function resolveTarget(req, allCredentials) {
+	return (
+		req.target ||
+		process.env.HARPER_CLI_TARGET ||
```
cb1kenobi, line 70: Generally, env vars trump CLI params.

bin/logout.js
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Executes the logout command.
+ * @param {string} [target] Optional target to logout from.
+ */
+async function logout(target) {
+	clearCredentials(target);
+	process.exit(0);
+}
+
+module.exports = logout;
```
cb1kenobi: For consistency, should we "name" the export?

```suggestion
module.exports = { logout };
```

bin/harper.js
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 'use strict';
 
+require('dotenv').config();
```
cb1kenobi: Is this gonna cause weird side effects when Harper restarts?

bin/login.js
```diff
@@ -0,0 +1,192 @@
...
+		output: process.stdout,
+	});
+
+	try {
+		const { cliOperations } = require('./cliOperations.js');
+		const { saveCredentials } = require('./cliCredentials.js');
+		console.log(chalk.cyan('\nHarper login'));
+		console.log(
+			chalk.gray(
+				'Harper apps can be deployed to Fabric for free, where one runtime can house your app, database, cache and messaging.'
```
cb1kenobi: Oxford comma?
```suggestion
				'Harper apps can be deployed to Fabric for free, where one runtime can house your app, database, cache, and messaging.'
```

bin/cliCredentials.js
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Loads the JWT credentials from the ~/.harper/credentials.json file.
+ * @returns {Object|null} The credentials object or null if not found.
+ */
+function loadCredentials() {
+	const credentialsFile = getCredentialsFile();
+	if (fs.existsSync(credentialsFile)) {
+		try {
+			return fs.readJsonSync(credentialsFile);
```
cb1kenobi: I for one would love to see us remove `fs-extra` as a dependency.
```suggestion
			return JSON.parse(fs.readFileSync(credentialsFile));
```

bin/cliCredentials.js
```diff
@@ -0,0 +1,112 @@
...
+		} catch (err) {
+			console.error(`Error reading credentials file: ${err.message}`);
+			return null;
```
cb1kenobi: Technically you could just fall through.
```suggestion
```

bin/cliCredentials.js
```diff
@@ -0,0 +1,112 @@
...
+			allCredentials.targets = allCredentials.targets || {};
+			allCredentials.targets[target] = tokens;
+			allCredentials.last_target = target;
+		} else {
+			// Fallback for when target is not provided (shouldn't happen with new login)
+			Object.assign(allCredentials, tokens);
+		}
+
+		fs.ensureDirSync(getCredentialsDir());
+		fs.writeJsonSync(getCredentialsFile(), allCredentials, { spaces: 2, mode: 0o600 });
```
cb1kenobi:
```suggestion
		fs.mkdirSync(getCredentialsDir(), { recursive: true });
		fs.writeFileSync(getCredentialsFile(), JSON.stringify(allCredentials, null, 2), { mode: 0o600 });
```

bin/cliCredentials.js
```diff
@@ -0,0 +1,112 @@
...
+					// try without trailing slash just in case
+					const altTarget = target.replace(/\/$/, '');
+					delete allCredentials.targets[altTarget];
+				}
+
+				if (allCredentials.last_target === target || allCredentials.last_target === target.replace(/\/$/, '')) {
+					const remainingTargets = Object.keys(allCredentials.targets);
+					allCredentials.last_target = remainingTargets.length > 0 ? remainingTargets[0] : null;
+				}
+				fs.writeJsonSync(credentialsFile, allCredentials, { spaces: 2, mode: 0o600 });
```
cb1kenobi:
```suggestion
				fs.writeFileSync(credentialsFile, JSON.stringify(allCredentials, null, 2), { mode: 0o600 });
```

bin/cliOperations.js
```diff
@@ -110,6 +127,19 @@ async function cliOperations(req) {
...
+			let tokens = null;
+			if (target && allCredentials.targets) {
+				const lookupKey = target.resolvedTarget.endsWith('/') ? target.resolvedTarget : target.resolvedTarget + '/';
+				tokens =
+					allCredentials.targets[lookupKey] ?? allCredentials.targets[target.resolvedTarget.replace(/\/$/, '')] ?? null;
+			} else if (allCredentials.operation_token) {
+				tokens = allCredentials;
+			}
+
+			if (tokens && tokens.operation_token) {
```
cb1kenobi: Now I'm getting nitpicky
```suggestion
			if (tokens?.operation_token) {
```

cb1kenobi (approved):
This is awesome, nice job!

bin/cliCredentials.js
dawsontoth: yup yup

bin/logout.js
dawsontoth: love it, so shall it be

bin/cliOperations.ts
dawsontoth: humm, i'm not sure I agree with that! Let's discuss. In my mind, the cli params are more specific, more recently intentional from the user. I see the env stuff as good... defaults. Or good support when you want to inject secrets in GHA or something without them being in a param. So if we had a scenario like this:
```
HARPER_CLI_TARGET=foo
harper login bar
```
I'd expect "bar" to win. You'd want foo to win?

bin/harper.js
dawsontoth: ah, i didn't realize it added it out here, yeah... that's problematic

bin/login.js
dawsontoth: accepted

bin/cliCredentials.js
dawsontoth: Thou dost speak troth

bin/cliCredentials.js
dawsontoth: nice

bin/cliOperations.js
dawsontoth: we are coallescing on agreement slowly ;D

bin/cliOperations.ts
cb1kenobi: In my upside down world, I would think that the environment variable "foo" would win. However, Claude agrees with you.

I guess part of my internal struggle here is that I don't think I mix env vars and CLI options. I have used both at the same time, but for different settings. When it comes to the same setting, I guess I never poured enough thought into it.

Now that I've thought about it, I agree CLI options should trump environment variables. It's silly if they didn't. We can resolve this. Carry on!

bin/harper.js
dawsontoth: ok, i moved it to only happen when we're doing login / logout / or one of the "make a request to a remote operations api" CLI commands. That should maintain the existing behavior, and unlock this nice `harper login \n harper deploy` flow.

dawsontoth:
@kriszyp the harper-agent puts its stuff in that ~/.harper/ folder already, and I thought we had some other stuff that can get put in there. If that's where we want it to be eventually, might as well stick it there now, that's my thinking at least!

bin/logout.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+require('dotenv').config();
```
Ethan-Arrowood: Is env not being loaded by the main cli path?

bin/logout.js
Ethan-Arrowood, line 1: why not make new files TS?

bin/cliOperations.ts
```diff
@@ -56,15 +57,32 @@ function buildRequest() {
...
  * @param req
+ * @param skipResponseLog By default, the response is logged to the console. Set this to true to skip logging it, which can be useful for sensitive responses like login calls!
  * @returns {Promise<void>}
  */
-async function cliOperations(req) {
-	if (!req.target) {
-		req.target = process.env.HARPER_CLI_TARGET || process.env.CLI_TARGET;
-	}
+async function cliOperations(req, skipResponseLog = false) {
+	require('dotenv').config();
```
Ethan-Arrowood, line 83: I'm concerned about how many times this is required and executed. The CLI should read environment variables once upon invocation, regardless of the command being executed, no?

bin/cliCredentials.js
```diff
@@ -0,0 +1,131 @@
...
+		console.error(`Error clearing credentials file: ${err.message}`);
+	}
+}
+
+/**
+ * Decodes a JWT and returns its payload.
+ * @param {string} token The JWT token to decode.
+ * @returns {Object|null} The decoded payload or null if invalid.
+ */
+function decodeJWT(token) {
```
Ethan-Arrowood: Would it make more sense for this to come from the `security/tokenAuthentication.ts` ? Do we already have decode or isExpired methods?

bin/cliCredentials.js
```diff
@@ -0,0 +1,131 @@
...
+function loadCredentials() {
+	const credentialsFile = getCredentialsFile();
+	if (fs.existsSync(credentialsFile)) {
+		try {
+			return JSON.parse(fs.readFileSync(credentialsFile, 'utf8'));
+		} catch (err) {
+			console.error(`Error reading credentials file: ${err.message}`);
+		}
+	}
+	return null;
```
Ethan-Arrowood:
```suggestion
	try {
		return JSON.parse(fs.readFileSync(credentialsFile, 'utf8'));
	} catch (err) {
		if (err.code === 'EEXIST') return null;
		else console.error(`Error reading credentials file: ${err.message}`);
	}
```

Just read it; and if the error is does not exist, return null.

bin/cliCredentials.js
```diff
@@ -0,0 +1,131 @@
...
+			allCredentials.targets = allCredentials.targets || {};
+			allCredentials.targets[target] = tokens;
+			allCredentials.last_target = target;
+		} else {
+			// Fallback for when target is not provided (shouldn't happen with new login)
+			Object.assign(allCredentials, tokens);
+		}
+
+		fs.mkdirSync(getCredentialsDir(), { recursive: true });
+		fs.writeFileSync(getCredentialsFile(), JSON.stringify(allCredentials, null, 2), { mode: 0o600 });
```
Ethan-Arrowood: What is this mode value mean? Please add in a comment.

bin/cliCredentials.js
```diff
@@ -0,0 +1,131 @@
...
+		}
+	}
+	return null;
+}
+
+/**
+ * Saves the JWT credentials to the ~/.harper/credentials.json file.
+ * @param {Object} credentials The credentials object containing operation_token and refresh_token.
+ */
+function saveCredentials(credentials) {
```
Ethan-Arrowood: It would be great to have a proper type for `credentials`.

bin/cliCredentials.js
```diff
@@ -0,0 +1,131 @@
...
+			if (!target.endsWith('/')) {
+				target += '/';
+			}
+			const allCredentials = loadCredentials();
+			if (allCredentials && allCredentials.targets) {
+				if (allCredentials.targets[target]) {
+					delete allCredentials.targets[target];
+				} else {
+					// try without trailing slash just in case
+					const altTarget = target.replace(/\/$/, '');
```
Ethan-Arrowood: Will `target` always have a trailing slash at this point considering the `target += '/';` above? 

If so, this can be simplified to just `target.slice(0,-1)` so you don't have to do regex parsing (even if relatively simple).

bin/cliCredentials.ts
```diff
@@ -0,0 +1,131 @@
...
+					const altTarget = target.replace(/\/$/, '');
+					delete allCredentials.targets[altTarget];
+				}
+
+				if (allCredentials.last_target === target || allCredentials.last_target === target.replace(/\/$/, '')) {
+					const remainingTargets = Object.keys(allCredentials.targets);
+					allCredentials.last_target = remainingTargets.length > 0 ? remainingTargets[0] : null;
+				}
+				fs.writeFileSync(credentialsFile, JSON.stringify(allCredentials, null, 2), { mode: 0o600 });
+				console.log(`Logged out from ${target}`);
```
Ethan-Arrowood, line 107: I notice this file only uses `console.log`. Should it be using the harper logger?

bin/login.js
```diff
@@ -0,0 +1,189 @@
+'use strict';
+
+require('dotenv').config();
+
+const fs = require('node:fs');
+const path = require('node:path');
+const readline = require('node:readline/promises');
+const dotenv = require('dotenv');
```
Ethan-Arrowood: dotenv is required twice.

Ethan-Arrowood (changes requested):

bin/cliOperations.ts
Ethan-Arrowood: There is precedence for this: https://docs.harperdb.io/reference/v5/cli/commands#:~:text=Note%3A%20When%20used%20in%20conjunction%2C%20command%20line%20arguments%20override%20environment%20variables.%20See%20Configuration%20for%20a%20full%20list%20of%20configuration%20parameters.

CLI options trump environment variables.

bin/cliOperations.ts
dawsontoth: If you `harper dev`, it should _not_ invoke it, leaving it up to our config.yaml env loader. Only specific commands should trigger it.

bin/cliCredentials.ts
cb1kenobi: Eh, this doesn't feel like a thing that needs to be logged. It's just a message to stdout.

dawsontoth: Most of the bin scripts utilize console.log or console.error, not the logger. This doesn't log into harper, only for the user on their CLI.

dawsontoth:
@Ethan-Arrowood I shifted to TypeScript and shuffled a few things around per your recommendations. The dotenv configs are also clearer now, happening when login, logout or cliOperations are invoked. Those are the 3 cases where we'll be purely shuttling requests to a targeted instance, and not handling them locally.

Ethan-Arrowood (approved):
nice work!

dawsontoth:
The unit test failure isn't related to the changes made here.

kriszyp (approved):
This is an awesome feature and I love it. I will include my AI review of this because it seems like it has some good suggestions, but ultimately I approve of this (if you want `~/.harper`, that's ok with me too):

1. .harper vs .harperdb inconsistency — user-visible
bin/cliCredentials.ts writes to path.join(getHomeDir(), '.harperdb', 'credentials.json') (correct — matches terms.HDB_HOME_DIR_NAME = '.harperdb'), but the welcome banner in bin/login.ts tells the user the file lives in ~/.harper/credentials.json, and the PR description also says ~/.harper. Users following the printed instructions will look in the wrong place. Use terms.HDB_HOME_DIR_NAME everywhere and update the prompt text. Even better: hard-code neither — print the resolved path.

2. saveCredentials silently swallows write failures
saveCredentials catches its own errors and only console.errors. bin/login.ts:152 then prints Successfully logged in to ... and process.exit(0) regardless. If the home dir is unwritable, the user is told they're logged in but nothing was persisted. saveCredentials should throw (or return a boolean) and login should surface failure and non-zero exit.

3. cliOperations fallback branch writes malformed data
In saveCredentials:

} else {
    // Fallback for when target is not provided (shouldn't happen with new login)
    Object.assign(allCredentials, tokens);
}
This sprays operation_token / refresh_token onto the top-level TargetedCredentials object, violating the declared type and producing a file no other code knows how to read. Since the caller guarantees a target, just delete the branch (or throw) — don't leave a broken fallback as a trap for future reads.

4. clearCredentials only writes back when targets existed
If a user runs harper logout some-stale-target and that key isn't present, the function silently does nothing — no message, no error, success exit. At minimum print a "no credentials found for X" message and exit non-zero. The "altTarget without trailing slash" double-lookup is also a sign the data isn't being normalized on the way in (see point 5).

5. Centralize target normalization
normalizeTarget exists in bin/login.ts but saveCredentials, clearCredentials, and the cliOperations lookup each re-implement "append a slash, then also try without one." Pick one canonical form, normalize once at the entry point, and stop carrying the variant matching everywhere. The trailing-slash fallback chains (allCredentials.targets[lookupKey] ?? allCredentials.targets[target.resolvedTarget.replace(/\/$/, '')]) get harder to reason about as the surface grows.

Medium
6. Hand-rolled raw-mode password prompt
The 30-line setRawMode block in login.ts (the author already flagged this) doesn't handle Unicode well (char += data.toString('utf-8') accumulates byte-wise, breaking multi-byte chars split across chunks), doesn't disable echo via the normal Node pattern, and process.exit(1) on Ctrl+C skips the finally { rl.close() }. The conventional fix is a tiny writable wrapper passed as output that swallows writes after the prompt — readline then handles backspace, Ctrl+C, paste, and Unicode for free. If you keep the hand-rolled version, at least: decode with StringDecoder, restore the TTY before exiting, and clear password from memory after use.

7. Lockfile churn looks unintentional
The package-lock.json portion of the diff flips peer: true / dev: true flags on a large number of unrelated packages (smithy, eslint, typescript-eslint, @types/node, acorn, etc.) and adds an optional bufferutil dependency that doesn't appear motivated by login/logout. This looks like a lockfile regenerated under a different npm or Node version, not a deliberate dependency change for this PR. Worth regenerating against the project's canonical npm version and dropping the noise — otherwise it'll trigger spurious diffs for everyone else on rebase.

8. isJWTExpired import
bin/cliOperations.ts imports isJWTExpired from security/tokenAuthentication.ts. I couldn't find it in the current source — confirm it exists on the target branch (or is being added in this PR) and isn't a Junie hallucination. If it's new, make sure it's not double-checking expiry of an unverified JWT (i.e., decoding without verifying signature is fine for "should I bother refreshing", but worth a comment).

9. Refresh-on-expiry: no retry on the original op
If operation_token is expired and the refresh succeeds, the new token is saved — but the original request continues with options.headers.Authorization = Bearer ${tokens.operation_token}. That line is set after refresh, so it works on the happy path. However, if the refresh fails (network error, 401), the code still attaches the expired token below; the server will 401 and the user gets the raw error. Consider: if refresh fails, prompt "session expired, run harper login again" instead of letting the underlying call surface a generic 401.

10. cliOperations signature lost its type annotation
Was async function cliOperations(req: any), now async function cliOperations(req, skipResponseLog = false). Either restore the any or type properly — Harper enforces TypeStrip-clean type annotations on signatures.

11. require('dotenv').config() inside cliOperations
Every CLI op now re-parses .env from disk. The login and logout functions also each call dotenv.config(). Move this to a single early init point. Also: cliOperations.ts mixes import at the top with require('dotenv').config() inline — pick one style.

12. .env mutation on first login
The login flow appends HARPER_CLI_TARGET=... to a .env file in process.cwd(). This is surprising — harper login from any random directory will scribble in that directory's .env. If the user is in their app repo, fine; if they're in /, they get a stray .env. At minimum: only do this if a .env already exists, or prompt before creating one. Also, it doesn't check whether HARPER_CLI_TARGET is already set in .env under a different value — appending a second line silently shadows.

Minor
require('./cliOperations.js') from login.ts uses the .js extension while the rest of the new code uses .ts imports. Inconsistent under TypeStrip.
clearCredentials picks remainingTargets[0] as the new last_target — arbitrary; consider null instead so the next op falls back to env vars rather than a surprise different host.
process.exit(0) at the end of logout skips any pending async work; for this command it's fine, but exiting from a library-style function is a smell.
Comment "shouldn't happen with new login" in bin/cliCredentials.ts — if it shouldn't happen, throw; don't leave the malformed-write fallback.

PR #528, opened 2026-05-13
Investigate Integration Test Flakiness

kriszyp:
Adding verbose logging and increasing polling timeouts to investigate intermittent failures on redirector and early-hints.


dawsontoth:
someone forgot to update the lock file when they updated a dependency 😁

dawsontoth (approved):

kriszyp:
> someone forgot to update the lock file when they updated a dependency 😁

As if that's all it took. I have run npm install locally to update a lock file so many times. It never succeeds. Claude said it was hopeless, basically. I have given up. If you want try, go for it.

components/Application.ts
```diff
@@ -600,6 +600,10 @@ export function nonInteractiveSpawn(
 			env.GIT_SSH_COMMAND = gitSSHCommand;
 		}
 
+		if (process.platform === 'win32' && command === 'npm') {
+			command = 'npm.cmd';
```
cb1kenobi, line 607: Unrelated, yet related, we probably only want to set `shell: true` on Windows and if we're setting `shell: true`, then might as well set `windowsHide: true`:

```javascript
		const childProcess = spawn(command, args, {
			shell: process.platform === 'win32',
			cwd,
			env,
			stdio: ['ignore', 'pipe', 'pipe'],
			windowsHide: true,
		});
```

config/configUtils.js
```diff
@@ -97,7 +97,28 @@ function getConfigPath(param) {
 function atomicWriteFile(filePath, content) {
 	const tempPath = `${filePath}.${process.pid}.${threadId}.${randomBytes(4).toString('hex')}.tmp`;
 	fs.writeFileSync(tempPath, content);
-	fs.renameSync(tempPath, filePath);
+	let retries = 5;
```
cb1kenobi, line 100: Is this from a merge or something? I swear this retry logic was proposed in another PR.

config/configUtils.js
```diff
@@ -97,7 +97,28 @@ function getConfigPath(param) {
...
+				retries--;
+				// sleep synchronously to allow the reader to close the file
+				const start = Date.now();
+				while (Date.now() - start < 10) {}
+				continue;
+			}
+			// if it fails we should clean up the tmp file
+			try {
+				fs.unlinkSync(tempPath);
+			} catch (cleanupErr) {
```
cb1kenobi: Making the linter happier
```suggestion
			} catch {
```

integrationTests/apiTests/tests/15_customFunctionsAndComponents.mjs
```diff
@@ -1,3 +1,6 @@
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+const __dirname = dirname(fileURLToPath(import.meta.url));
```
cb1kenobi, line 3: With the integration tests running on Node 24, we can use `import.meta.dirname`.

cb1kenobi (approved):

Issue #135, opened 2026-02-06
A target (`RequestTarget`) should have parsed query properties in static REST methods


kriszyp:
Currently, the target is not fully parsed when REST static methods are called, which makes it harder to use. They should be parsed _before_ calling the REST static methods, for better usability since we are encouraging more use of the static methods.

PR #529, opened 2026-05-14
fix: prevent silent index gaps when per-record puts fail during schema migration

kriszyp:
## Summary

Fixes a silent-data-loss bug in `runIndexing` (resources/databases.ts): when a per-record `index.put` fails during a schema-migration backfill, the error was caught/logged and the loop continued, eventually marking the index as _complete_ with gaps. Queries against the new index silently returned fewer rows than `primaryStore` scans (SQL / ops API), which is exactly the serent-canopy issue #135 fingerprint.

## Root causes

1. **Per-record error handling**: both sync catch (line 1257) and async `when()` error handler (line 1276) logged and continued — the migration appeared to complete successfully with dropped records.
2. **`await lastResolution` bypass**: if the last put in the loop was itself a rejected promise, it threw past the `hadIndexingErrors` check to the outer catch, so the fix code was never reached.

## Fix

- Wrap `await lastResolution` in its own try-catch that sets `hadIndexingErrors = true`.
- When `hadIndexingErrors` is true: do NOT clear `indexingPID`, `isIndexing`, or `lastIndexedKey`. Set `indexingFailed = true` on the descriptor and persist it.
- Add `indexingFailed` to the condition in `table()` so the next call (including after restart with a new PID) detects it and re-triggers the backfill from `lastIndexedKey`.
- Also ensure any dbi opened while `indexingPID` is set inherits `isIndexing = true` (companion fix for the `resetDatabases()` race during an active migration).
- Make the `Object.defineProperty(attribute, 'dbi', ...)` call `configurable: true` for forward compatibility.

## Test

`unitTests/resources/schemaMigrationFragility.test.js` — F2 test now passes:
- Simulates ERR_BUSY rejections on every 10th index put.
- Verifies search throws 503 "not indexed yet" (not silently returns 45/50 rows).
- Verifies `resetDatabases()` detects `indexingFailed` and re-triggers a clean backfill.

## Review attention

- `table()` condition change (lines 1106/1115): adding `|| attributeDescriptor.indexingFailed` — verify this doesn't fire spuriously in normal operation.
- `try { await lastResolution } catch` wrapping (line ~1298): previously the outer catch caught this; the new path correctly routes it through `hadIndexingErrors`. Low risk, but worth a second look.
- The `if (attributeDescriptor?.indexingPID) dbi.isIndexing = true` line (after the migration detection block) is new. It ensures any dbi created by a concurrent `resetDatabases()` also sees `isIndexing = true` while the migration is running. Confirm it doesn't leave indexes stuck in `isIndexing` when migration was already complete (it shouldn't, since `indexingPID` is cleared in the success path before `indexing-finished` signals).

🤖 Generated with Claude Sonnet 4.6 (1M context)

### DIFF
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -1103,6 +1103,7 @@ export function table<TableResourceType>(tableDefinition: TableDefinition): Tabl
 				const dbi = openIndex(dbiKey, rootStore, attribute);
 				if (
 					changed ||
+					attributeDescriptor.indexingFailed ||
 					(attributeDescriptor.indexingPID && attributeDescriptor.indexingPID !== process.pid) ||
 					attributeDescriptor.restartNumber < workerData?.restartNumber
 				) {
@@ -1111,6 +1112,7 @@ export function table<TableResourceType>(tableDefinition: TableDefinition): Tabl
 					attributeDescriptor = attributesDbi.getSync(dbiKey);
 					if (
 						changed ||
+						attributeDescriptor.indexingFailed ||
 						(attributeDescriptor.indexingPID && attributeDescriptor.indexingPID !== process.pid) ||
 						attributeDescriptor.restartNumber < workerData?.restartNumber
 					) {
@@ -1124,14 +1126,21 @@ export function table<TableResourceType>(tableDefinition: TableDefinition): Tabl
 						if (hasExistingData) {
 							attribute.lastIndexedKey = attributeDescriptor?.lastIndexedKey ?? undefined;
 							attribute.indexingPID = process.pid;
+							delete attribute.indexingFailed; // clear failure flag for the new run
+							delete attribute.indexingAttempt; // reset attempt counter
 							dbi.isIndexing = true;
-							Object.defineProperty(attribute, 'dbi', { value: dbi });
+							Object.defineProperty(attribute, 'dbi', { value: dbi, configurable: true, enumerable: false });
 							// we only set indexing nulls to true if new or reindexing, we can't have partial indexing of null
 							attributesToIndex.push(attribute);
 						}
 					}
 					attributesDbi.put(dbiKey, attribute);
 				}
+				// If a migration is in progress (indexingPID set), any newly opened dbi must also
+				// reflect isIndexing = true. A resetDatabases() during an active runIndexing creates
+				// a new dbi object; without this, queries could use the new dbi (isIndexing = false)
+				// and return incomplete results while the backfill is still running.
+				if (attributeDescriptor?.indexingPID) dbi.isIndexing = true;
 				if (attributeDescriptor?.indexNulls && attribute.indexNulls === undefined) attribute.indexNulls = true;
 				dbi.indexNulls = attribute.indexNulls;
 				indices[attribute.name] = dbi;
@@ -1202,6 +1211,7 @@ async function runIndexing(Table, attributes, indicesToRemove) {
 			lastResolution = index.drop();
 		}
 		let interrupted;
+		let hadIndexingErrors = false;
 		const attributeErrorReported = {};
 		let indexed = 0;
 		const attributesLength = attributes.length;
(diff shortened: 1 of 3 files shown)


resources/databases.ts
```diff
@@ -1284,20 +1296,64 @@ async function runIndexing(Table, attributes, indicesToRemove) {
 				if (outstanding > MAX_OUTSTANDING_INDEXING) await lastResolution;
 				else if (outstanding > MIN_OUTSTANDING_INDEXING) await new Promise((resolve) => setImmediate(resolve)); // yield event turn, don't want to use all computation
 			}
+		}
+		// Await the last pending put. If it rejects, that's also an indexing error.
+		try {
+			await lastResolution;
```
cb1kenobi, line 1307: I'm starting to not like this pattern. If any of the non-last promises reject, they will be unhandled.

resources/databases.ts
```diff
@@ -1284,20 +1296,64 @@ async function runIndexing(Table, attributes, indicesToRemove) {
...
+					`On next restart the migration will be retried from the last checkpoint (indexingFailed=true). ` +
+					`Affected attributes: ${attributes.map((a) => a.name).join(', ')}`
+			);
+		} else {
 			// update the attributes to indicate that we are finished
 			for (const attribute of attributes) {
 				delete attribute.lastIndexedKey;
 				delete attribute.indexingPID;
+				delete attribute.indexingFailed;
+				delete attribute.indexingAttempt;
```
cb1kenobi: Is this being used?

resources/databases.ts
```diff
@@ -1124,14 +1126,21 @@ export function table<TableResourceType>(tableDefinition: TableDefinition): Tabl
 						if (hasExistingData) {
 							attribute.lastIndexedKey = attributeDescriptor?.lastIndexedKey ?? undefined;
 							attribute.indexingPID = process.pid;
+							delete attribute.indexingFailed; // clear failure flag for the new run
+							delete attribute.indexingAttempt; // reset attempt counter
```
cb1kenobi: Is this being used?

resources/databases.ts
```diff
@@ -1284,20 +1296,64 @@ async function runIndexing(Table, attributes, indicesToRemove) {
...
+				lastResolution = Table.dbisDB.put(attribute.key, attribute);
+				// Keep isIndexing = true on both the attribute.dbi and the currently-active dbi
+				// in Table.indices (which may differ if resetDatabases() ran during this pass).
+				attribute.dbi.isIndexing = true;
+				const activeDbi = Table.indices[attribute.name];
+				if (activeDbi) activeDbi.isIndexing = true;
+			}
+			await lastResolution;
+			logger.warn(
+				`Indexing of ${Table.tableName} encountered errors on some records — index will remain incomplete. ` +
```
cb1kenobi: Question, should be be avoiding mdashes in our logs? I know it's 2026 and just because we can doesn't necessarily mean we should. It could complicate log monitoring.

```suggestion
				`Indexing of ${Table.tableName} encountered errors on some records - index will remain incomplete. ` +
```

kriszyp:
Addressed all four review comments:

1. **Line 1302** (`await lastResolution`): Added a comment clarifying that this try-catch specifically handles the case where `lastResolution` itself rejects (the last put in the loop), which would otherwise bypass `hadIndexingErrors`. Acknowledged that the broader unhandled-rejection risk for non-last puts in multi-value attributes is a pre-existing issue in the `when()` pattern and out of scope for this fix.

2. **Lines 1130 & 1341** (`delete attribute.indexingAttempt`): Removed both lines. The `indexingAttempt` property was never set anywhere in the final code — it was a leftover from an abandoned in-process retry mechanism that was replaced with the simpler `indexingFailed` approach.

3. **Line 1331** (em dash): Changed `—` to `-` in the log message.

🤖 Generated by Claude Sonnet 4.6 (1M context)

cb1kenobi (approved):

PR #533, opened 2026-05-14
Fix flaky unit tests: timing races and LMDB segfault on Node v24

kriszyp:
## Summary

- **Skip LMDB tests on Node.js v24** — the LMDB native module segfaults under v24; all other suites (`main`, `apitests`, `resources`) still run on v24. Workaround until upstream adds v24 support.
- **Fix `DatabaseTransaction.addWrite()`** — when a transaction is already closed (`immediateCommit` path), `save()` returns a commit Promise that was being silently discarded. Callers awaiting `table.put()` were resolving before the write was actually durable, causing intermittent ordering failures.
- **Replace fixed-duration waits with polling loops** in `auditLog.test.js` (audit cleanup) and `crud.test.js` (analytics flush) — fixed delays were too short on loaded CI runners.
- **Stabilize `trackedTxns` baseline** in `txn-tracking.test.js` — stale async commit callbacks from prior tests caused intermittent off-by-one on the size measurement.
- **MQTT timing fixes** in `mqtt-test.mjs` — small delay between PUT and PATCH prevents out-of-order message delivery; increased disconnect delay 10ms → 50ms so the broker fully processes the disconnect before offline patches are queued.
- **Fix `EADDRINUSE` crash in worker threads** (`threadServer.js`) — the Node.js `listenOnPorts()` path was rejecting on EADDRINUSE (e.g. MQTT port 1883 already held by the main thread), causing workers to crash before sending `child_started`. The parent then hung on `Promise.all(workersReady)` for the full 500 s mocha timeout. The Bun path already handled this gracefully (resolve instead of reject); this fix brings the Node.js path into line. Fixes the consistently-failing **Multi-threaded cache updates** suite on Node.js v20.

## Context

Identified by reviewing two weeks of CI failures on `main`. All were timing races or environment-sensitivity issues, with two implementation-level fixes (DatabaseTransaction and the EADDRINUSE worker crash).

## Areas deserving attention

- **`DatabaseTransaction.addWrite()`** — the `immediateCommit` path is the most common `put()` path. The fix propagates `saveResult.then(() => operation)`. Verify callers expecting a synchronous `TransactionWrite` still work (they get one in the non-immediate-commit path).
- **`threadServer.js` EADDRINUSE handling** — resolving instead of rejecting on EADDRINUSE means the port silently isn't bound in that worker. For HTTP (SO_REUSEPORT), this is fine. For MQTT, which runs on a single port, only the main thread ends up holding it — which is the correct behaviour. No MQTT connections are lost, but this makes the assumption explicit.

_Generated by Claude Sonnet (agent)_

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -55,11 +55,4 @@ jobs:
         run: node --enable-source-maps ./dist/bin/harper.js install
 
       - name: Run tests
-        # LMDB's native module segfaults under Node.js v24; skip the LMDB variant until the
-        # upstream module is updated to support v24.  All other test suites still run on v24.
-        run: |
-          if [ "${{ matrix.node-version }}" = "24" ]; then
-            npm run test:unit:main && npm run test:unit:apitests && npm run test:unit:resources
-          else
-            npm run test:unit:all
-          fi
+        run: npm run test:unit:all
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -55,4 +55,11 @@ jobs:
         run: node --enable-source-maps ./dist/bin/harper.js install
 
       - name: Run tests
-        run: npm run test:unit:all
+        # LMDB's native module segfaults under Node.js v24; skip the LMDB variant until the
+        # upstream module is updated to support v24.  All other test suites still run on v24.
+        run: |
+          if [ "${{ matrix.node-version }}" = "24" ]; then
+            npm run test:unit:main && npm run test:unit:apitests && npm run test:unit:resources
+          else
+            npm run test:unit:all
+          fi
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -111,6 +111,7 @@ source tree — which permanently shadows git's config file. Every subsequent ag
 `fatal: unknown error occurred while reading the configuration files` and the cycle repeats.
 
 **Rules to prevent recurrence:**
+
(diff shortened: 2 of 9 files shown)


cb1kenobi (approved):
Yikes, LMDB segfaults on Node 24 sounds bad.

kriszyp:
**Update (agent follow-up):**

Two additional fixes pushed in `dcbd811f`:

1. **`DatabaseTransaction`: graceful retry when `this.writes` was cleared by a concurrent immediate-commit**
   The v22 test failure ("Transaction was abandoned after writing to the transaction log") traced to a race where the outer transaction's `rocksTxn.commit()` gets `ERR_BUSY`, but by retry time `this.writes = []` has been reset by a concurrent immediate-commit on the same `DatabaseTransaction` object. The retry then called `abort()` on a transaction that had writes. Fixed by catching the abort error in that empty-writes branch — the concurrent commit owns those writes, so the retry can safely exit. Verified with 20 consecutive local runs.

2. **`threadServer.js`: reverted the EADDRINUSE-swallowing change**
   As pointed out in review, all Node.js `listen()` calls already pass `reusePort: !isWindows`. If SO_REUSEPORT is set consistently, EADDRINUSE should not occur. Silently resolving EADDRINUSE could mask a real error (port already in use by another process). The Bun path keeps its equivalent guard because Bun's raw-socket listener does not honour the reusePort option. This means the multi-threaded Node v20 failure may resurface — if it does, we'll have actionable information about which server is binding without SO_REUSEPORT.

_Generated by Claude Sonnet (agent)_

PR #413, opened 2026-04-28
Eliminate TypeScript Errors and Convert foundational .js files to .ts

kriszyp:
This PR migrates the HarperDB core codebase to NodeNext TypeScript with strict typing and ESM
  compatibility. 

  Main Changes:
   - Strict NodeNext Resolution: All imports now use explicit .js extensions.
   - Circular Dependency Resolution: Refactored core dataLayer and sqlTranslator components to break
     initialization loops.
   - Lazy Initialization: Refactored static table and database mounting into lazy-loaded getters to
     prevent bootstrap order failures (fixing "path must be a string" errors).
   - LMDB & Transactions: Stabilized transaction logic to ensure zero double-closes and proper blob
     cleanup during aborts, preventing fatal C++ level crashes.
   - Ecosystem Integration: Successfully merged and resolved complex conflicts with origin/main,
     including new Bun-specific optimizations and routing changes.
   - Test Infrastructure: Restored unit test passing rate to >99%, bridging legacy JavaScript tests
     to the new transpiled TypeScript source.

  Validation:
  Verified with npm run build, npm run lint, and npm run test:unit:main. user.ts unit tests and core
  authentication logic are fully verified and passing.

PR #534, opened 2026-05-14
fix(ops-api): skip reusePort for operations API to fix macOS ENOTSUP

kriszyp:
## Summary

- Sets `server.noReusePort = true` on the Node http.Server backing the operations API inside `getHTTPServer()` when `usageType === 'operations-api'`
- `listenOnPorts()` now skips `reusePort` for servers marked with this flag

## Purpose

macOS users reported `ENOTSUP: operation not supported on socket :::9925` when starting Harper (see [#413 comment](https://github.com/HarperFast/harper/pull/413#issuecomment-4447632878)).

The operations API (default port 9925) only runs on the main thread. `reusePort` (`SO_REUSEPORT`) is needed for the application port so multiple worker threads can all bind to the same port. The operations API doesn't use multiple threads, so it never needed `reusePort`. On macOS, passing `reusePort: true` to `listen()` on a plain Node http.Server in this context fails with `ENOTSUP`.

## Design note

The flag is set on the server object stored in `SERVERS[port]` (created by `getHTTPServer`) — not on Fastify's internal http.Server (which is the cascade target). This is the object that `listenOnPorts()` iterates and calls `listen()` on.

Cross-model review (Gemini) confirmed the fix targets the right object after an initial version incorrectly set the flag on the Fastify server.

Generated by Claude Code (claude-sonnet-4-6).

### DIFF
--- a/security/keys.js
+++ b/security/keys.js
@@ -689,6 +689,10 @@ if (typeof globalThis.Bun === 'undefined') {
 		delete lessOptions.key;
 		delete lessOptions.cert;
 		let ctx = origCreateSecureContext(lessOptions);
+		if (typeof ctx.context?.setCert !== 'function') {
+			// setCert is a Node.js internal — not available in all environments; fall back to default
+			return origCreateSecureContext(options);
+		}
 		ctx.context.setCert(options.cert);
 		ctx.context.setKey(options.key, undefined);
 		return ctx;
--- a/server/http.ts
+++ b/server/http.ts
@@ -455,7 +455,8 @@ function getHTTPServer(port: number, secure: boolean, options: ServerOptions) {
 			server.isSecure = true;
 		}
 		registerServer(server, port);
-		if (isOperationsServer) server.noReusePort = true;
+		// macOS doesn't support SO_REUSEPORT on all socket types; operations API also doesn't need it
+		if (isOperationsServer || process.platform === 'darwin') server.noReusePort = true;
 
 		// Operations API domain socket connections bypass auth (equivalent to local access)
 		if (isOperationsServer && String(port).includes('/')) server.bypassLocalAuth = true;
@@ -655,7 +656,7 @@ function getBunHTTPServer(port: number, secure: boolean, options: ServerOptions)
 		// Store the config for Bun.serve() — will be started by threadServer.js listenOnPorts()
 		const config: any = {
 			fetch: fetchHandler,
-			reusePort: true,
+			reusePort: process.platform !== 'darwin' && process.platform !== 'win32',
 		};
 		if (secure) {
 			// TLS config for Bun
--- a/server/http.ts
+++ b/server/http.ts
@@ -455,6 +455,7 @@ function getHTTPServer(port: number, secure: boolean, options: ServerOptions) {
 			server.isSecure = true;
 		}
 		registerServer(server, port);
+		if (isOperationsServer) server.noReusePort = true;
 
 		// Operations API domain socket connections bypass auth (equivalent to local access)
 		if (isOperationsServer && String(port).includes('/')) server.bypassLocalAuth = true;
--- a/server/threads/threadServer.js
+++ b/server/threads/threadServer.js
@@ -257,9 +257,9 @@ function listenOnPorts() {
 				listen_on = {
 					host: port.slice(0, lastColon).replace(/[[\]]/g, ''),
 					port: +port.slice(lastColon + 1),
-					reusePort: !isWindows,
+					reusePort: !isWindows && !server.noReusePort,
 				};
-			else listen_on = { port: +port, host: '::', reusePort: !isWindows };
+			else listen_on = { port: +port, host: '::', reusePort: !isWindows && !server.noReusePort };
 			if (isNaN(listen_on.port)) continue;
 		} catch (error) {
 			harperLogger.error(`Unable to bind to port ${port}`, error);
--- a/server/threads/threadServer.js
+++ b/server/threads/threadServer.js
@@ -280,6 +280,7 @@ function listenOnPorts() {
 }
 
 async function listenOnPortsBun() {
+	const isMac = process.platform === 'darwin';
 	const bunServeConfigs = httpComponent.bunServeConfigs;
 	for (let port in bunServeConfigs) {
 		const config = bunServeConfigs[port];
@@ -302,7 +303,6 @@ async function listenOnPortsBun() {
 			} else {
 				portNumber = +port;
 			}
-			const isMac = process.platform === 'darwin';
 			const serveOptions = {
 				port: portNumber,
 				reusePort: !isWindows && !isMac,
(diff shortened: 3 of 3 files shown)


cb1kenobi (commented):
PR fails on macOS:
```
[main/0] [notify]: Harper installation was successful.
[main/0] [info]: Harper PID 39927
[main/0] [info]: Checking if HDB software has been updated
Debugger listening on ws://127.0.0.1:9229/479e2bfa-acd3-4ee4-be20-9e92b139f36f
For help, see: https://nodejs.org/en/docs/inspector
[main/0] [info]: All root applications loaded
[main/0] [info]: Domain socket listening on /Users/chris/harper/operations-server
[http/1] [error]: unhandledRejection Error: listen ENOTSUP: operation not supported on socket :::1883
    at Server.setupListenHandle [as _listen2] (node:net:1918:21)
    at listenInCluster (node:net:1997:12)
    at node:net:2206:7
    at process.processTicksAndRejections (node:internal/process/task_queues:89:21) {
  code: 'ENOTSUP',
  errno: -45,
  syscall: 'listen',
  address: '::',
  port: 1883
}
```

cb1kenobi:
Claude recommends these changes, which do fix the problem:
```
diff --git a/server/http.ts b/server/http.ts
index 6eb773305..c2c821c27 100644
--- a/server/http.ts
+++ b/server/http.ts
@@ -455,7 +455,8 @@ function getHTTPServer(port: number, secure: boolean, options: ServerOptions) {
                        server.isSecure = true;
                }
                registerServer(server, port);
-               if (isOperationsServer) server.noReusePort = true;
+               // macOS doesn't support SO_REUSEPORT on all socket types; operations API also doesn't need it
+               if (isOperationsServer || process.platform === 'darwin') server.noReusePort = true;

                // Operations API domain socket connections bypass auth (equivalent to local access)
                if (isOperationsServer && String(port).includes('/')) server.bypassLocalAuth = true;
@@ -655,7 +656,7 @@ function getBunHTTPServer(port: number, secure: boolean, options: ServerOptions)
                // Store the config for Bun.serve() — will be started by threadServer.js listenOnPorts()
                const config: any = {
                        fetch: fetchHandler,
-                       reusePort: true,
+                       reusePort: process.platform !== 'darwin' && process.platform !== 'win32',
                };
                if (secure) {
                        // TLS config for Bun
diff --git a/server/threads/threadServer.js b/server/threads/threadServer.js
index b239550ba..45accb464 100644
--- a/server/threads/threadServer.js
+++ b/server/threads/threadServer.js
@@ -302,9 +302,10 @@ async function listenOnPortsBun() {
                        } else {
                                portNumber = +port;
                        }
+                       const isMac = process.platform === 'darwin';
                        const serveOptions = {
                                port: portNumber,
-                               reusePort: !isWindows,
+                               reusePort: !isWindows && !isMac,
                                fetch: config.fetch,
                        };
                        if (portHostname) serveOptions.hostname = portHostname;
@@ -416,7 +417,7 @@ async function listenOnPortsBun() {
                                listening.push(
                                        new Promise((resolve, reject) => {
                                                server
-                                                       .listen({ port: portNum, host: rawHostname || '::' }, () => {
+                                                       .listen({ port: portNum, host: rawHostname || '0.0.0.0' }, () => {
                                                                resolve({ port });
                                                                harperLogger.trace('Listening on port ' + port, threadId);
                                                        })
@@ -463,6 +464,7 @@ function onSocket(listener, options) {
                        listener
                );
                SNICallback.initialize(socketServer);
+               socketServer.noReusePort = true;
                SERVERS[options.securePort] = socketServer;

                // Create a corresponding Unix Domain Socket mirror for the secure socket
@@ -495,6 +497,7 @@ function onSocket(listener, options) {
                        keepAlive: true,
                        keepAliveInitialDelay: 600,
                });
+               socketServer.noReusePort = true;
                SERVERS[options.port] = socketServer;
        }
        return socketServer;
```

kriszyp:
Applied all requested changes from the review:

- Expanded the macOS guard to cover all servers in `getHTTPServer()` (not just the ops API) via `process.platform === 'darwin'`
- Disabled `reusePort` in `getBunHTTPServer()` on macOS and Windows
- Disabled `reusePort` in `listenOnPortsBun()` on macOS
- Marked TLS and plain socket servers as `noReusePort = true` in `onSocket()`
- Changed raw socket listen fallback host from `'::'` to `'0.0.0.0'` to avoid IPv6-only binding on macOS

> Generated by Claude Code agent

kriszyp:
Applied: scoped the `0.0.0.0` host fallback to macOS only — Linux uses `::` (dual-stack) to preserve IPv4+IPv6 connectivity. Also moved `isMac` to function scope so it's accessible in both the Bun server loop and the raw socket loop.

> Generated by Claude Code agent

kriszyp:
Two separate issues in this report:

**`uv_ip6_addr` crash**: This is NOT from `node-unix-socket` NAPI calls. After the previous fix removed `reusePort` from raw socket binds, the `host: '::'` remained. Bun's `net.Server.listen()` compatibility layer internally calls `uv_ip6_addr` to parse the IPv6 address — which Bun hasn't implemented (see oven-sh/bun#18546). The fix already in this PR (changing `::` to `0.0.0.0` on macOS) addresses this by avoiding IPv6 address parsing entirely on macOS.

**`setCert` error**: The `ctx.context.setCert` Node.js-internal hack in `security/keys.js` is already gated by `if (typeof globalThis.Bun === 'undefined')`, so it shouldn't fire in Bun. However, just pushed a defensive fallback: if `setCert` is not available (e.g. Bun worker thread edge case or non-standard TLS context), it now falls back to `origCreateSecureContext(options)` instead of crashing.

> Generated by Claude Code agent

cb1kenobi:
This fixes the errors on both Node and Bun startup, but Bun will throw this whenever you access http://localhost:9926:
```
[http/1] [error]: TypeError: null is not an object (evaluating 'req.url')
    at handle (/Users/chris/projects/harper-doom/node_modules/.pnpm/vite@8.0.8_@types+node@25.6.0_yaml@2.8.3/node_modules/vite/dist/node/chunks/node.js:6108:32)
    at app (/Users/chris/projects/harper-doom/node_modules/.pnpm/vite@8.0.8_@types+node@25.6.0_yaml@2.8.3/node_modules/vite/dist/node/chunks/node.js:6053:8)
    at <anonymous> (unknown:15:70)
    at http (/Users/chris/projects/harper/dist/server/REST.js:71:24)
    at <anonymous> (/Users/chris/projects/harper/dist/server/REST.js:322:16)
    at authentication (/Users/chris/projects/harper/dist/security/auth.js:345:32)
    at async <anonymous> (/Users/chris/projects/harper/dist/server/http.js:571:48)
    at processTicksAndRejections (native:7:39)
    ```

cb1kenobi (approved):
Node.js seems happy! Bun, not so much.

PR #535, opened 2026-05-14
Update package-lock.json

kriszyp:


.github/workflows/lint-code.yaml
```diff
@@ -18,5 +18,5 @@ jobs:
         with:
           node-version-file: '.node-version'
           cache: 'npm'
-      - run: npm ci
+      - run: npm install --ignore-scripts
```
cb1kenobi, line 21: This is not going to honor the lock file. Can we do this?

```suggestion
      - run: npm ci --ignore-scripts
```

kriszyp: It doesn't work, it is not possible to create a lock file with our set of transitive dependencies derived from the `ws` dependency and the native dependencies that will satisfy `npm ci`.

cb1kenobi (approved):

.github/workflows/lint-code.yaml
kriszyp: But `npm ci` seems to be passing on main now. Pure black magic, I don't understand. (not gonna merge unless we need this)

PR #531, opened 2026-05-14
feat(deploy): live SSE progress for deploy_component

kriszyp:
> **Stacked on #530** — please review that one first; this PR will be retargeted to `main` after #530 lands.

## Summary

- Adds a `ProgressEmitter` and an `Accept: text/event-stream` path to the operations API for `deploy_component`.
- `deployComponent` and `Application.prepareApplication` emit `phase` events at the extract → install → load → replicate → restart boundaries; `install` events stream each line of `npm install`/`pnpm`/`yarn`/custom-command stdout & stderr live; the stream terminates with `done` (operation result) or `error`.
- The CLI requests SSE for `deploy_component` and renders a real upload progress bar (cli-progress in TTY, periodic text lines in non-TTY) followed by live phase and install events. Stdout still receives the final JSON/YAML result so existing CLI consumers don't break.
- Non-SSE callers see zero behavior change: the emitter is undefined on that path and every emission is optional-chained.

First slice of HarperFast/harper#526.

## Where to look

- `server/serverHelpers/progressEmitter.ts` — small pub-sub + the `createSSEResponseStream` helper. The interesting subtlety is the `subscribe()` snapshot before iteration (so listeners that unsubscribe themselves during dispatch don't shift indexes), and the swallowed-exception policy so a buggy listener can never break the operation.
- `server/serverHelpers/serverHandlers.js` — the new SSE branch in `handlePostRequest`. Headers (`Content-Type`, `Cache-Control`, `X-Accel-Buffering`) are set before the stream so reverse proxies and Fastify compression don't buffer the response.
- `components/operations.js` + `components/Application.ts` — emission points. `Application.prepareApplication` emits the `extract done → install start` boundary because `Application` itself doesn't know which phase comes next; the rest is in `deployComponent`. `installApplication` builds an `onLine` callback that forwards each complete stdout/stderr line as an `install` SSE event; threaded through all three install code paths (custom command, devEngines packageManager, npm fallback).
- `components/Application.ts` `nonInteractiveSpawn` — line-buffered `onLine` parameter so a chunk that splits mid-line doesn't fire a half-event; trailing partial lines flushed on close.
- `components/packageComponent.ts` — `getPackagedDirectorySize` mirrors `streamPackagedDirectory`'s skip predicates so the CLI's upload bar has a meaningful 100% target.
[...]

Issue #524, opened 2026-05-13
Support for deployments larger than 2GB


kriszyp:
## Problem

Payload-based `deploy_component` is capped at ~2 GB. The CLI (`bin/cliOperations.js`) calls `packageDirectory()`, which returns the tar+gzip as a `Buffer`, CBOR-encodes the whole request, and POSTs it. The server (`server/operationsServer.ts`, body limit `REQ_MAX_BODY_SIZE` ~1 GB) reads the body whole before dispatch. Node.js's hard 2 GB Buffer ceiling is the absolute upper bound, and the configured body limit is lower in practice.

Replication compounds the problem: `server.replication.replicateOperation()` relays operations as single payloads over the inter-node WebSocket. Multi-GB payloads exceed both buffer and frame-size limits there as well.

### 1. CLI — stream the package as `multipart/form-data`

`bin/cliOperations.js` switches `deploy_component` from CBOR-encoded body to `multipart/form-data`:

- One field part per operation property (`operation`, `project`, `package`, flags, etc.) serialized as JSON.
- One file part carrying the tar+gzip payload, piped from a streaming variant of `packageDirectory()`. No intermediate `Buffer`.
- Sent via Node `https` with `Transfer-Encoding: chunked` — no `Content-Length` precomputation.

Touches `bin/cliOperations.js`, `components/packageComponent.ts` (expose streaming variant), and `utility/common_utils.js` (multipart-aware `httpRequest`).

### 2. Operations API server — receive multipart and stream into extraction

`server/operationsServer.ts` accepts a multipart body when `Content-Type: multipart/form-data`:

- Field parts are decoded into the same `req` object shape the JSON/CBOR path produces.
- The file part is exposed as a `Readable` and passed straight into `deployComponent`, replacing `req.payload: Buffer` with `req.payload: Readable` in this path.
- The existing extraction pipeline in `components/Application.ts:217` (`gunzip-maybe` → `tar-fs.extract`) is already stream-native — it accepts the file part directly, eliminating the in-memory hop.
- `REQ_MAX_BODY_SIZE` is bypassed on the multipart deploy path (the limit only meaningfully constrains the small field parts; the file part is streamed and bounded by disk only).

Touches `server/operationsServer.ts`, `server/serverHelpers/serverHandlers.js`, `server/serverHelpers/serverUtilities.ts`, `components/operations.js`, `components/Application.ts`.

### 3. Replication — direct HTTPS relay instead of WebSocket-framed operation

[...]

PR #536, opened 2026-05-14
feat(deploy): stage streamed payloads to a temp file for replication

kriszyp:
> **Stacked on #531** — please review #530 and #531 first; this PR will be retargeted to \`main\` after both land.

## Summary

Origin-side payload staging needed for the direct-HTTPS peer relay (harper-pro slice 3b). When a \`deploy_component\` request will be replicated to peers, the origin now writes the streamed payload to a temp file before local extraction so it can re-stream the same bytes to each peer afterwards. Without this, the payload Readable is consumed once by local extract and gone by the time replication runs.

Third slice of HarperFast/harper#524.

## Where to look

- \`components/payloadStaging.ts\` — single-purpose utility that pipelines a Readable into a temp \`payload.tar.gz\` and returns a cleanup callback. The trade-off comment in the file header explains why we write-then-read instead of teeing in flight (simpler code, identical backpressure for extraction; disk speed isn't the bottleneck).
- \`components/operations.js\` — staging gate in \`deployComponent\`. Skipped when:
  - the deploy is package-identifier-based (\`req.package\`, no \`payload\` stream),
  - \`req.replicated === false\` (the operation is itself a replicated deploy that the peer received, so no further fan-out),
  - or there are no peers (\`server.nodes.length === 0\` — single-node deploys keep their zero-disk-copy property).
- Cleanup runs in a \`finally\` after \`server.replication.replicateOperation(req)\`, plus a defensive cleanup if \`prepareApplication\` fails before replication.

## Backward compatibility

- The staging is invisible to the operations API contract: \`req.payload\` is replaced internally with a \`createReadStream\` over the temp file before the \`Application\` is constructed, so every downstream code path (\`Application.extractApplication\`, the SSE phase emitter, anything reading \`application.payload\`) keeps seeing a \`Readable\` exactly as before.
- Single-node and package-identifier deploys: zero change in behavior, no temp file written.

## Test plan

- [x] \`npx mocha unitTests/components/payloadStaging.test.js\` — 5/5 pass (round-trip contents, cleanup removes file+dir, double-cleanup is safe, path-traversal sanitization keeps the staged dir under tmpdir, source-stream errors propagate).
- [x] \`npx mocha unitTests/server/fastifyRoutes/operations.test.js\` — 21/21 pass (existing deployComponent tests; the staging branch is only entered when nodes are present, which these tests don't simulate).
- [x] \`npx mocha unitTests/components/packageComponent.test.js unitTests/server/serverHelpers/multipartParser.test.js unitTests/bin/multipartBuilder.test.js unitTests/server/serverHelpers/progressEmitter.test.js unitTests/bin/sseConsumer.test.js\` — 32/32 pass (slice 1 + slice 2 unaffected).
- [ ] End-to-end replicated deploy — exercised in harper-pro's slice 3b integration test, not here.


kriszyp:
Pausing this PR — converting to draft.

This work is being rebuilt on top of the design in [#641](https://github.com/HarperFast/harper/issues/641): a replicated `hdb_deployment` system table that owns the deploy lifecycle, audit trail, and (because Harper's `BLOB_CHUNK` replication already supports chunked, back-pressured blob transfer) the payload delivery channel.

In the new design:
- The `ProgressEmitter` from this PR becomes one of two subscribers — the second being a `DeploymentRecorder` that writes the persistent record. SSE remains the live channel for the CLI; `get_deployment` becomes content-negotiated to serve the same stream to Studio.
- `_stagedPayloadPath` (#536) and the direct-HTTPS peer relay (harper-pro#146) both go away — peers read the payload directly from the replicated blob attribute on the row.

This work resumes as part of Slice B in #641 once Slice A (table + blob-backed multipart receive) lands.

— Claude

kriszyp:
Closing in favor of the deployment-tracking redesign tracked in #641, now in flight as Slices A (#655, merged), B1 (#657, merged), and B2 (#758).

The redesign supersedes this PR's entire approach:

- **No more staged payload temp file.** The payload lives as a `Blob` attribute on a replicated `hdb_deployment` row. Origin writes the multipart body straight into `row.payload_blob`; peers read it back via Harper's existing `BLOB_CHUNK` streamed-blob replication. Slice A landed the row write; Slice B2 lands the peer-side read.
- **No more `payloadStaging.ts`.** Slice A's `deploymentRecorder.ingestPayload` is the equivalent (hash + size tap, blob-backed instead of file-backed). It will be cleaned up by table reclamation rather than per-deploy cleanup.
- **No more direct-HTTPS peer relay (harper-pro#146).** Because the payload now ships via the row replication channel, the operations-API frame cap that #146 was working around no longer applies to deploy_component.

The two unresolved review threads on `payloadStaging.ts` and `operations.js` (temp-dir leak on pipeline failure; staged payload not cleaned on load-phase failure) are real bugs in this approach, but the file is being deleted as part of the redesign rather than fixed. Closing the threads by closing the PR.

🤖 Generated by Claude

PR #538, opened 2026-05-14
fix: defer and yield audit-store commit notifications

kriszyp:
## Summary

Three targeted changes to [`resources/transactionBroadcast.ts`](resources/transactionBroadcast.ts) so that audit-store `'committed'` events no longer monopolize the worker event loop during heavy commit traffic (e.g. replication backlog catch-up):

- **`activeCount` short-circuit** — track live subscriptions on `databaseSubscriptions`. When zero, `notifyFromTransactionData` skips the iteration entirely. The reusable rocksdb iterator preserves its position and resumes once a subscriber arrives.
- **`setImmediate`-deferred notification** — the `'committed'` listener no longer runs `notifyFromTransactionData` synchronously inside the commit microtask. A `notifyScheduled` flag coalesces a burst of commits in the same turn into one notify pass and lets I/O (replication sockets, blob streams, HTTP) interleave.
- **Batched yielding mid-iteration** — every `NOTIFY_BATCH_SIZE` (256) records the drain `setImmediate`s its continuation. `pendingTxnSubscribers` carries across the yield so `end_txn` still fires exactly once when the iterator truly drains.

Yielding is gated by an `allowYield` parameter. The same-thread `listenToCommits` aftercommit path explicitly does **not** set it, because that path holds the cross-thread `'thread-local-writes'` lock and must complete in one turn.

Also cleans up a pre-existing leak in `Subscription.end()` where `delete envSubscriptions[dbi]` referenced an undefined `dbi` property instead of `tableSubscriptions.tableId`, leaving empty table Maps in place.

## Why

While investigating worker unresponsiveness on us-central-1.pb-prod.gend during replication backlog catch-up, CPU profiling on a 100%-pinned worker thread showed `>94%` of samples in `rocksdb-js` `TransactionLog.next()`, driven by the merge-sort aggregator iterator in `RocksTransactionLogStore.getRange({})` reached via `commit → save → … → next()`. That iterator is created by `transactionBroadcast` and walked synchronously on every commit's `'committed'` event. During backlog catch-up the cumulative cost saturates the event loop, CDP eval to the worker times out, and a single Receiving peer (4.9h behind on `prerender`) stalls every other subscription on that worker.

## Where to look

- **`notifyFromTransactionData`** in [`resources/transactionBroadcast.ts`](resources/transactionBroadcast.ts) — the drain loop is now wrapped in `try { … } finally { … }`, with `yielded`/`allowYield` controlling whether the finally clears `notifyScheduled` (so any throw inside the drain can't permanently freeze future notifications).
- **Re-entrance.** `notifyScheduled` stays `true` for the full drain *including* yield/resume turns. The `'committed'` listener early-returns when it's set, so a new commit during a yielded drain never spawns a second concurrent iteration on the same shared iterator. The iterator is reusable for rocksdb (`auditStore.reusableIterable`) and recreated from `lastTxnTime` for LMDB on each yield-continuation.
- **Semantics change worth a careful look.** Subscribers' listeners now fire on the *next* event-loop turn rather than synchronously inside the commit microtask. The two new tests below cover delivery; if anything downstream assumes synchronous-within-commit notification I'd appreciate flagging it.
- **`Subscription.end()`** — the `delete envSubscriptions[dbi]` fix changes behavior only for cases where the last subscription for a table is removed; the empty Map was previously kept alive. No call sites depend on the leaked Map.

## Test plan

- [x] `unitTests/resources/subscriptionReplay.test.js` — full file (25 passing, 9 pending). Adds two new tests:
  - `delivers all live events when a single committed burst exceeds the notify batch size` (N=600 sequential puts, asserts all 600 ids delivered with no duplicate `(id,version)` pairs).
  - `resumes delivery after a subscribe/end/subscribe cycle with writes in between` (regression for the `activeCount` short-circuit).
- [x] `unitTests/resources/auditLog.test.js`, `transaction.test.js`, `caching.test.js`, `crud.test.js` — all passing.
- [x] Lint (`npm run lint:required`) — 0 errors (392 pre-existing warnings).
- [x] Formatted (`npm run format:write`).
- [ ] Worth verifying on a live Fabric node before merging by watching worker `recentELU.utilization` and CDP responsiveness during a real backlog catch-up.

Cross-model review (Gemini) flagged three concerns; all addressed in this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code) by Claude Opus 4.7 (1M context)

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -37,6 +37,7 @@ When adding a new commit-handler early-return path: reset `write.skipped = false
 When `table()` is called with an attribute newly marked `indexed: true` (or with any change that requires re-building the secondary index), `runIndexing` is launched asynchronously and `Table.indexingOperation` is set to its promise. While running:
 
 **In-flight state tracking (persisted to `attributesDbi`):**
+
 - `attribute.indexingPID = process.pid` — set at migration start; cleared on clean completion. On restart with a different PID, `indexingPID !== process.pid` triggers a re-migration.
 - `attribute.lastIndexedKey` — updated every 100 records as a resumable checkpoint. Cleared on clean completion; preserved on error so a retry starts from this key.
 - `attribute.indexingFailed = true` — set if any record's `index.put` errors during the backfill. `table()` checks this flag: a fresh call in the same or a new process re-triggers the backfill from `lastIndexedKey`.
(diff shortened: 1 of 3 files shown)


kriszyp:
Thanks for catching this — pushed [3b8ac02](https://github.com/HarperFast/harper/pull/538/commits) to address it.

Took the inline approach in the `'committed'` listener rather than moving `nextTransaction()` into `notifyFromTransactionData`'s prelude: when `activeCount === 0` we rotate the promise (gated on `auditStore.nextTransaction` so we don't allocate one for databases that have never had a waiter) and then return, so we still skip the expensive iteration. Added a regression test that calls `whenNextTransaction` directly with no per-key subscriber in scope and asserts the promise resolves after a put.

DavidCockerill (approved):

cb1kenobi (approved):

PR #539, opened 2026-05-14
fix(replay): skip corrupt audit entries instead of crash-looping on replay

kriszyp:
## Summary

Harden `replayLogs()` against corrupt audit entries left behind by an unclean shutdown. Skip entries whose header decode failed or whose record value is missing for write-path ops, log the total skip count once at the end.

## Why

A Harper Pro v5.0.16 prod node OOM'd, and on restart it spent ~3 minutes pinned at ~700% CPU iterating the unflushed RocksDB audit log. Profile against the running worker showed:

```
35% readAuditEntry @ auditStore.js:442
28% (garbage collector)
 6% Decoder @ auditStore.js:540
```

The loop was throwing the same TypeError on every iteration:

```
[main/0] [error]: Error decoding record Error: Unexpected end of MessagePack data
[main/0] [error]: Error writing from replay of log TypeError: Cannot read properties of undefined (reading 'cacheKey')
    at TableResource.validate (Table.js:2903:57)
    at DatabaseTransaction.save (DatabaseTransaction.js:182:37)
    at replayLogs (replayLogs.js:101:39)
```

The per-iteration `try/catch` was catching it but the loop kept going. The audit log on disk had thousands of entries where `auditRecord.getValue()` either threw (truncated msgpack) or returned `undefined` (no `HAS_RECORD`/`HAS_PARTIAL_RECORD` flag set). For record-requiring ops (`put`/`patch`/`message`/`invalidate`) the eventual `validate()` call dereferenced `record[attribute.name]` where `record === undefined`.

## What changed

- New pure helper `resources/replayLogsGuards.ts` exporting `classifyAuditEntryForReplay(type, tableId, hasRecord)` → `'corrupt-header' | 'missing-record' | null`, plus the `REQUIRES_RECORD` set it consults.
- `replayLogs()` calls the helper twice — once on the destructured header fields, once after `getValue()` — and increments a `skipped` counter on either rejection. The `getValue()` `try/catch` is now silent (`catch {}`); millions of identical decode errors per replay was not actionable information.
- One summary log line at the end: `Skipped N unrecoverable audit entries in <db> database during replay`.
- 5 unit tests covering corrupt-header, missing-record, happy path, no-value ops (`delete`/`relocate`/`structures`), and a lock on the `REQUIRES_RECORD` set so additions to the switch don't drift from the guard.

## Where to look

- **`replayLogs.ts` `getValue()` catch** — silently dropping the error is a deliberate trade-off. The error is the same per iteration (`Unexpected end of MessagePack data`); the count goes into the end-of-replay summary. If you want a single sample logged, easy to add.
- **`REQUIRES_RECORD` membership** — verified against the switch in `replayLogs.ts` and `Table.validate()`. `delete`/`relocate` operate on the existing DB entry, `structures` uses `getBinaryValue()`, so all three are safe to replay without a `record`. `invalidate` is included defensively per Gemini review — its current `_writeInvalidate` path doesn't crash on undefined record but would write an incomplete tombstone.
- **`hasRecord: boolean` signature** — chosen over passing the audit record itself so the helper has zero Harper deps and tests don't need the full module bootstrap (importing `replayLogs.ts` directly hits a `LMDBTransaction extends DatabaseTransaction` circular load).

## Testing

- `npx mocha unitTests/resources/replayLogs.test.js` — 5 passing
- `npm run build` — same pre-existing TS error count plus 1 (an extra `databaseName` reference of the same kind already present); dist still emits
- `npx oxlint resources/replayLogs.ts resources/replayLogsGuards.ts unitTests/resources/replayLogs.test.js` — 0 warnings, 0 errors
- `npx prettier --check ...` — clean
- Cross-model review (Gemini): positive; suggested swapping `catch (decodeError) { void decodeError; ... }` for optional-catch `catch { ... }`, applied.

## Risk

Low. The only behavioral change for the happy path is the new summary log line. The skip paths only trigger when the existing code was guaranteed to throw inside `validate()`.

## Test plan

- [ ] CI passes
- [ ] On a node with a corrupt audit log, observe `Skipped N unrecoverable audit entries…` log line instead of per-record `Error writing from replay of log` flood, and CPU returns to baseline within seconds rather than minutes

### DIFF
--- a/resources/replayLogsGuards.ts
+++ b/resources/replayLogsGuards.ts
@@ -1,0 +1,30 @@
+// Pure helpers for replayLogs (no Harper module dependencies, so unit tests can load
+// them without bootstrapping the full Resource / RocksDB / DatabaseTransaction graph).
+//
+// Background: a node that crashed unclean re-runs replayLogs against the unflushed audit
+// log on next boot. If any audit entry is corrupt or missing its record body, the loop
+// hits a TypeError inside Table.validate() ("Cannot read properties of undefined
+// (reading 'cacheKey')") and the per-iteration catch swallows it — but the loop keeps
+// running over potentially millions of entries, pinning CPU. These guards classify each
+// entry up front so the loop can skip cleanly.
+
+// Types whose replay requires a decoded record value. Missing records on any of these
+// crashes inside validate() on the first attribute dereference.
+export const REQUIRES_RECORD = new Set(['put', 'patch', 'message', 'invalidate']);
+
+/**
+ * Returns `null` if the entry is safe to replay, or a short reason string otherwise.
+ *
+ * @param type        `auditRecord.type` — `undefined` when readAuditEntry caught a header decode error
+ * @param tableId     `auditRecord.tableId` — `undefined` for the same reason
+ * @param hasRecord   `true` if `auditRecord.getValue(...)` produced a non-undefined value
+ */
+export function classifyAuditEntryForReplay(
+	type: string | undefined,
+	tableId: number | undefined,
+	hasRecord: boolean
+): 'corrupt-header' | 'missing-record' | null {
+	if (type === undefined || tableId === undefined) return 'corrupt-header';
+	if (REQUIRES_RECORD.has(type) && !hasRecord) return 'missing-record';
+	return null;
+}
--- a/resources/replayLogsGuards.ts
+++ b/resources/replayLogsGuards.ts
@@ -8,23 +8,35 @@
 // running over potentially millions of entries, pinning CPU. These guards classify each
 // entry up front so the loop can skip cleanly.
 
-// Types whose replay requires a decoded record value. Missing records on any of these
-// crashes inside validate() on the first attribute dereference.
-export const REQUIRES_RECORD = new Set(['put', 'patch', 'message', 'invalidate']);
+// Mirrors `HAS_RECORD` (16) | `HAS_PARTIAL_RECORD` (32) from auditStore.ts — the action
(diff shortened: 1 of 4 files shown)


resources/replayLogsGuards.ts
```diff
@@ -0,0 +1,30 @@
...
+ * @param type        `auditRecord.type` — `undefined` when readAuditEntry caught a header decode error
+ * @param tableId     `auditRecord.tableId` — `undefined` for the same reason
+ * @param hasRecord   `true` if `auditRecord.getValue(...)` produced a non-undefined value
+ */
+export function classifyAuditEntryForReplay(
+	type: string | undefined,
+	tableId: number | undefined,
+	hasRecord: boolean
+): 'corrupt-header' | 'missing-record' | null {
+	if (type === undefined || tableId === undefined) return 'corrupt-header';
```
cb1kenobi: Consider using integers instead of strings. Kinda like we do in rocks/lmdb.

cb1kenobi (approved):

resources/replayLogsGuards.ts
kriszyp: Good call — switched to bitmask check against `HAS_RECORD | HAS_PARTIAL_RECORD` (commit 794cd08e). The guard now takes `auditRecord.extendedType` (raw action int) and ANDs against `RECORD_BEARING_FLAGS = 16 | 32`. Drops the string-name Set entirely; tests updated to use the bit values with a lock test pinning the mask.

🤖 AI-generated reply

Issue #549, opened 2026-05-15
Windows: investigate `restart_service` crash blocking component-load test paths


heskew:
## Context

While trying to expand Windows integration test coverage (#525, #546, #547), the actual underlying blocker came into focus. Multiple integration test files have per-test skips for `restart_service` operations with the same explanation:

> `'Windows: Skipping restart_service to avoid crash'`

Found in:
- `integrationTests/apiTests/tests/23_blob.mjs` (2 occurrences)
- Likely elsewhere — `grep -rE "Windows.*Skipping restart_service|restart_service.*crash"` for the full set

These skips look surgical and isolated but they're load-bearing in suite flow: any test that needs Harper to pick up a newly-written component file depends on `restart_service` to actually load it. With the operation skipped, downstream tests fail with "schema not found" / "component not loaded" cascades.

This was the latent root cause behind #546/#547. The outer suite-level skip on `23_blob.mjs` (`{ skip: process.platform === 'win32' }`) is effectively a workaround for this — undocumented, but real.

See HarperFast/harper#547 (closed) for the cascade trace: the file-write portion of the blob suite passes on Windows, then "Confirm Blob schema and table created" fails because the schema isn't there, because the component was never loaded, because `restart_service` was skipped.

## Goal

Make `restart_service` work on Windows, OR replace the operation in the test paths with a Windows-compatible alternative that achieves the same effect (Harper picks up a newly-written component).

## Approach (suggested)

1. **Reproduce the crash locally on Windows.** Run a minimal test that calls `restart_service` (e.g., one of the currently-skipped tests in `23_blob.mjs`). Capture the actual crash signature — stack trace, exit code, hdb.log output.
2. **Identify the failure mode.** Likely candidates based on Kris's prior Windows fixes:
   - File-locking / EPERM during the restart sequence (`7823d97a` "workaround database drop crashes on windows" suggests this pattern is present)
   - Worker-process spawn ordering specific to Windows process semantics
   - Socket / pipe handling for IPC during the restart handshake
   - Database file handles not released cleanly across the restart
3. **Fix or work around.** Either fix the crash in Harper itself, or introduce a Windows-specific code path (e.g., separate-process restart, retry-with-backoff on EPERM, explicit file-handle release before restart).
[...]

PR #556, opened 2026-05-15
spike(#549): capture restart_service crash signature on Windows (DO NOT MERGE)

heskew:
**DO NOT MERGE** — diagnostic branch only.

## Purpose

Trigger the existing \`run-integration-apiTests-windows\` CI job against an unskipped \`23_blob.mjs\` so we capture the actual crash signature for #549. The crash was documented in commit \`0751644f\` ("'restart_service' operation for 'http_workers' in 23_blob.mjs causes the HarperDB process to crash") but no stack trace was ever preserved.

## What this branch does

Removes the three Windows skips in \`integrationTests/apiTests/tests/23_blob.mjs\`:

1. Outer suite-level skip on \`describe('23. Blob', { skip: process.platform === 'win32' }, ...)\`
2. Per-test \`t.skip\` inside the first "Restart Service: http workers and wait"
3. Per-test \`t.skip\` inside the second "Restart Service: http workers and wait"

Every other Windows skip elsewhere in the apiTests suite remains untouched. No source code change.

## What happens

1. CI fires the existing \`run-integration-apiTests-windows\` job
2. The blob suite deploys its component, then hits \`restart_service { service: 'http_workers' }\` while the blob component has open file handles
3. Harper crashes (per Kris's prior documentation)
4. The existing \`harper-integration-api-test-logs-windows\` artifact upload captures hdb.log + stdouts + stderrs

## After the run

1. Download the artifact
2. Attach the crash signature to #549
3. Close this PR without merging

The production fix for the underlying crash lives in \`lmdb-js\` native binding cleanup or in Harper's worker shutdown ordering — neither involves un-skipping these tests in this manner. This is purely a diagnostic capture.

## Note re: Kris's PR #555

#555 is migrating the apiTests suite to the new \`@harperfast/integration-testing\` framework on isolated per-test-file Harper instances. It does not touch \`23_blob.mjs\` in this first slice. Spike and #555 don't conflict at the file level.

### DIFF
--- a/integrationTests/apiTests/tests/23_blob.mjs
+++ b/integrationTests/apiTests/tests/23_blob.mjs
@@ -17,7 +17,7 @@ import { createBlobCustom } from '../utils/blob.mjs';
 import { exec } from 'node:child_process';
 import { timestamp } from '../utils/timestamp.mjs';
 
-describe('23. Blob', { skip: process.platform === 'win32' }, () => {
+describe('23. Blob', () => {
 	beforeEach(timestamp);
 
 	const blobId = randomInt(1000000);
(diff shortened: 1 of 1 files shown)


heskew:
Spike complete. Crash signature captured and attached to #549. Closing without merge as planned.

PR #577, opened 2026-05-18
fix: skip corrupt audit entries during iteration instead of throwing

### LINKED ISSUE #575: Audit decoder throws uncaught RangeError on corrupt entries during replication catch-up
## Summary

During replication catch-up, reading audit entries via `auditStore.getRange(...)` can throw `RangeError: Error reading float64: Offset is outside the bounds of the DataView` from `getFloat64` when the underlying audit DBI contains a corrupt entry. The throw originates in `dist/core/resources/auditStore.js:537` ("Reading audit entry error" log line precedes the throw).

## Field evidence

Observed in production on a 12-node fully replicated cluster:
- `uncaughtException RangeError: Offset is outside the bounds of the DataView at position 709224988` on multiple http workers and `[main/0]` on jp-osa and ap-west nodes, immediately preceded by `Reading audit entry error RangeError` from auditStore.
- The corruption is at a **fixed audit byte position** (709224988) — every restart hits it again during catch-up.
- Resulting symptom: outgoing replication for affected (peer, db) pairs gets silently stuck — open WebSocket, no further data flowing, `lastReceivedRemoteTime` frozen at the moment of failure.

## Why this matters now

`HarperFast/harper-pro#154` makes the pro-side replication connection retry on this kind of failure instead of dying silently. Without a core-side fix, that retry will loop on the same corrupt audit position forever (retry timer grows only ~0.4%/iter, so backoff is slow). The pro change converts "silent dead" into "noisy stuck" — escaping the stuck state still needs core to handle the corrupt entry.

## What core needs

Some combination of:

1. **Recover/skip the corrupt entry.** When `readAuditEntry` (or whatever surfaces the float64 read) detects out-of-bounds, log + skip + advance past the bad entry rather than throwing. The audit log is a write-ahead-style structure; skipping one entry is preferable to refusing to read any.
2. **Bound the throw to a recoverable error.** If skipping isn't safe, at least surface it through the iterator's normal error channel so callers can decide policy — currently the throw escapes the for-of in an async context and lands as `uncaughtException` on a different tick (this is what was observed live).
3. **Diagnose how the corruption happens.** Position 709224988 on multiple unrelated nodes is suspicious — could be deterministic from a crash-during-write, a checksum gap, or a partial truncation. Worth a focused investigation into whether audit writes are atomic across the relevant boundary.

## Suggested triage

- [ ] Reproduce by injecting a malformed audit entry at a known offset and exercising replication catch-up
- [ ] Determine which read path actually throws (`readAuditEntry`, `getFloat64` directly, decoder reset on bad metadata)
- [ ] Decide between skip-and-continue vs surface-via-iterator
- [ ] Once decided, add a regression test under `core/unitTests` for the corrupt-entry case

Filed as follow-up from HarperFast/harper-pro#154.

🤖 Filed by Claude (Opus 4.7) on behalf of @Devin-Holland

Devin-Holland:
## Summary

A corrupt audit entry encountered during `auditStore.getRange(...)` was throwing `RangeError: Offset is outside the bounds of the DataView` out of the for-of consumer in an async context, landing as `uncaughtException` on a later tick. Outgoing replication for the affected (peer, db) pair would silently stall, and because the bad byte offset is deterministic, every restart re-hit the same throw during catch-up.

This change makes the iterator skip past corrupt entries instead of throwing, so replication catch-up can drain past the bad offset and resume.

Closes #575.

## What was wrong

Three failure points, all reachable from the rocks-path stack trace in the issue:

1. **`RocksTransactionLogStore.getRange`'s `.map()` callback** read the rocks-side prelude (`structureVersion`/`previousResidencyId`/`previousVersion`) via raw `DataView.getUint32` / `getFloat64` *before* calling `readAuditEntry`. Those reads were not wrapped, so a short/corrupt prelude threw straight out through `iterable.map` → consumer's for-of → uncaught. The field stack trace pointed to `decoder.getUint32(0)` at `RocksTransactionLogStore.ts:294` — exactly this path.
2. **`readAuditEntry`** had a try/catch around the header decode, but a corrupt `length` field (e.g., a `0xff`-prefixed uint32) pushed `decoder.position` hundreds of megabytes past the end of the buffer before any check ran; the next `readFloat64` then threw with the bogus position in the message. The throw was still caught by the outer try, but the resulting log line had an opaque/misleading position number.
3. **Lazy `recordId` / `user` getters** on the returned record lived *outside* the outer try/catch — a header that decoded "successfully enough" but had a corrupt key region would throw on later property access.

There's also a parallel LMDB-side hole: `keyEncoder.readKey` called `dataView.getFloat64(start)` with no bounds check, so a truncated key buffer threw through lmdb-js's iterator.

## What changed

- `RocksTransactionLogStore.getRange`'s map callback wraps the entire per-entry decode in try/catch. On failure it yields a corrupt-marker record (`corrupt: true`, `version = timestamp`, `tableId`/`type`/etc. undefined) so iteration advances past the bad entry while preserving the log timestamp.
- `readAuditEntry` validates `length` fields against the buffer's `byteLength` before advancing `decoder.position`, fails fast inside the outer try, and returns a structurally complete sentinel (with all expected getters and methods) from the catch — so downstream consumers don't NPE when they call methods on a record that failed to decode.
- The lazy `recordId` and `user` getters each catch their own decode errors and return `undefined`.
- `keyEncoder.readKey` checks `start + 8 <= buffer.byteLength` before the float64 read and returns `NaN` instead of throwing.

Downstream consumers already skip records with no `tableId`/`type` — `classifyAuditEntryForReplay` flags them as `corrupt-header`, and `Table.ts` / `transactionBroadcast.ts` use `tableId !== tableId` guards that filter the sentinel cleanly. So no consumer-side changes were needed.

## Where to focus review attention

- **`RocksTransactionLogStore.ts` map callback shape** — the corrupt sentinel returned from the catch must be a valid `AuditRecord` enough that downstream consumers behave like they do for the lmdb-side `{}` sentinel today. I confirmed `transactionBroadcast.notifyFromTransactionData` and `replayLogs` both already handle the undefined-tableId case, but a second pair of eyes here would be useful.
- **`createCorruptAuditSentinel` field surface** — I copied the shape from the happy-path return, but if a consumer reads a field I missed (e.g., a `previousNodeId` getter elsewhere) the sentinel would surface `undefined` and that consumer might need a guard. Worth a once-over against existing callers of `auditStore.getRange`.
- **`length` validation bounds** (`length < 0 || decoder.position + length > buffer.byteLength`) — the `length < 0` guard handles a theoretical signed-overflow path; current `Decoder.readInt` returns unsigned values but the check is cheap insurance against future refactors.

The fix doesn't address the underlying *cause* of the corruption (suggested triage item 3 in the issue — possibly atomic-write boundary). That's a separate investigation; this PR is just about keeping the iterator alive when it happens.

## Test plan

- [x] Added 4 regression unit tests in `unitTests/resources/auditLog.test.js` covering:
  - oversized recordId length on a truncated buffer
  - mid-header length-field mutation that would push position past byteLength
  - lazy getter access on a corrupt body (asserts no throw on `.recordId` / `.user`)
  - happy-path round-trip to lock in the new guards don't regress valid entries
- [x] Existing `auditLog.test.js` suite still passes (13 tests)
- [ ] Full `npm run test:unit` (skipped here — pre-existing TypeScript build errors on `main` unrelated to this change make local full-suite runs noisy; CI will exercise the lmdb + rocks unit-test matrices)

🤖 Generated by Claude Opus 4.7

### DIFF
--- a/resources/RocksTransactionLogStore.ts
+++ b/resources/RocksTransactionLogStore.ts
@@ -293,9 +293,9 @@ export class RocksTransactionLogStore extends EventEmitter {
 			// throw a raw `RangeError: Offset is outside the bounds of the DataView` out
 			// through `iterable.map`, escape the for-of consumer, and land as an
 			// uncaughtException on a later tick — stalling outgoing replication at the
-			// failing offset on every catch-up attempt. On error, yield a corrupt-marker
-			// record with the timestamp preserved so iteration advances past the bad
-			// entry; downstream consumers already skip records with no `tableId`/`type`.
+			// failing offset on every catch-up attempt. On error, yield a sentinel record
+			// with the timestamp preserved so iteration advances past the bad entry;
+			// downstream consumers already skip records with no `tableId`/`type`.
 			try {
 				const decoder = new Decoder(data.buffer, data.byteOffset, data.byteLength);
 				(data as any).dataView = decoder;
@@ -326,7 +326,6 @@ export class RocksTransactionLogStore extends EventEmitter {
 					byteLength: data?.byteLength,
 				});
 				return {
-					corrupt: true,
 					version: timestamp,
 					endTxn,
 					type: undefined,
--- a/resources/RocksTransactionLogStore.ts
+++ b/resources/RocksTransactionLogStore.ts
@@ -5,6 +5,7 @@ import { Decoder, readAuditEntry, ENTRY_DATAVIEW, AuditRecord, createAuditEntry 
 import { isMainThread } from 'node:worker_threads';
 import { EventEmitter } from 'node:events';
 import { asBinary } from 'lmdb';
+import * as harperLogger from '../utility/logging/harper_logger.ts';
 
 if (!process.env.HARPER_NO_FLUSH_ON_EXIT && isMainThread) {
 	// we want to be able to test log replay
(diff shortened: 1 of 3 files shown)


kriszyp (approved):
Not really root cause, but I guess this helps with problems from becoming worse.

PR #580, opened 2026-05-18
cherry-pick: fix: prevent silent index gaps when per-record puts fail during schema migration (PR #529)

kriszyp:
Cherry-pick of [#529](https://github.com/HarperFast/harper/pull/529) to `v5.0`.

## Summary

- `runIndexing` (in `resources/databases.ts`) no longer marks a schema-migration backfill complete when per-record `index.put` calls fail. On any error during the backfill it now persists `indexingFailed = true` and keeps `indexingPID`, `isIndexing`, and `lastIndexedKey` set, so queries get `503 "not indexed yet"` instead of silently returning a partial result set.
- Wraps the final `await lastResolution` in its own try-catch so a rejected last put cannot jump past the `hadIndexingErrors` check into the outer catch.
- A subsequent `table()` call (same process, or after restart with a different PID) detects `indexingFailed = true` and re-triggers the backfill from `lastIndexedKey`.
- `Object.defineProperty(attribute, 'dbi', ...)` is made `configurable: true` so future retry mechanisms can reassign `dbi`.
- Adds `unitTests/resources/schemaMigrationFragility.test.js` with a repro for the silent-gap fingerprint (transient `ERR_BUSY` on every 10th put), verifying (a) post-failure search throws 503 and (b) restart re-runs a clean backfill.

## Cherry-pick notes

- The DESIGN.md docs commit (`docs: document runIndexing schema-migration internals`) was a no-op on `v5.0` — those notes are already present on the branch (carried in by an earlier cherry-pick). The commit was skipped.
- `resources/databases.ts` was the only meaningful merge conflict and resolved cleanly against the v5.0 surrounding context.

## Test plan

- [ ] Integration tests pass on `cherry-pick/v5.0/pr-529`.
- [ ] `unitTests/resources/schemaMigrationFragility.test.js` runs and passes.

---
🤖 Cherry-picked by [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -1087,7 +1087,6 @@ export function table<TableResourceType>(tableDefinition: TableDefinition): Tabl
 							attribute.lastIndexedKey = attributeDescriptor?.lastIndexedKey ?? undefined;
 							attribute.indexingPID = process.pid;
 							delete attribute.indexingFailed; // clear failure flag for the new run
-							delete attribute.indexingAttempt; // reset attempt counter
 							dbi.isIndexing = true;
 							Object.defineProperty(attribute, 'dbi', { value: dbi, configurable: true, enumerable: false });
 							// we only set indexing nulls to true if new or reindexing, we can't have partial indexing of null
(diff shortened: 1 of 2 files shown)


dawsontoth (approved):
The test failures look familiar for sinon getting updated, which I saw and adjusted on https://github.com/HarperFast/harper/pull/563 -- the lock file for v5 looks to have merge conflicts that were committed into it raw, which may explain the odd dependency installations, leading to the failures.

https://github.com/HarperFast/harper/blob/v5.0/package-lock.json

PR #605, opened 2026-05-19
fix: clean up TypeScript build errors surfacing from harper-pro

kriszyp:
## Summary

Type-safety fixes to make `tsc` clean when harper-pro builds against these core files (harper-pro pulls them in via shared `include` paths). All changes are type-only except one latent bug fix called out below.

## Files

- `server/serverHelpers/serverUtilities.ts` — added optional `parametersSchema?: any[]` to `OperationDefinition`. Several callers in harper-pro register operations with this property; it was already a runtime contract, just missing from the type.
- `security/keys.ts` — `createTLSSelector(type, mtlsOptions)` is invoked with one arg in multiple places (e.g. core's own `keys.ts:112` uses `undefined` explicitly). Made `mtlsOptions?` optional and annotated the return as `any` so the `initialize` property assigned via `as any` is callable on the returned function.
- `resources/transactionBroadcast.ts`
  - `subscriptions` local was typed `any[]` but is then annotated with `.tables` / `.key` properties. Loosened to `any` (runtime usage is already heterogeneous).
  - `Subscription.subscriptions` field loosened from `[]` to `any`.
  - `notifyFromTransactionData` second param made optional to match the existing single-arg callsite at line 280.
- `resources/replayLogs.ts`
  - `(rootStore as any).auditStore` / `(rootStore as any).databaseName` — augmentations on `RocksDatabase` that aren't on the base type.
  - `(rootStore.getRange(...) as any)` — destructuring fields not on the public `AuditRecord` type.
  - `Table as any` / `tableInstance: any` for the `_writeUpdate` / `_writePublish` / `_writeRelocate` / `_writeDelete` / `_writeInvalidate` / `.save()` calls.
  - **Latent bug fix:** the `Context` was built with `user: { name: username }`, but `User` requires `username`. Anything downstream reading `context.user.username` would find `undefined`. Changed to `user: { username }`. **Worth a look** — please confirm `username` is the intended property (it matches the `User` interface in `security/user.ts`).
- `server/serverHelpers/contentTypes.ts` ndjson handler — `.map(JSONParse)` passes `(value, index, array)` to `JSONParse`, and `JSON.parse`'s second arg is `reviver`. Receiving a number where a function is expected would throw `TypeError: reviver is not a function`. Wrapped in an arrow: `.map((line) => JSONParse(line))`.

## Where to focus review

1. The `username` fix in `replayLogs.ts` — pretty sure this was a real bug masked by `{ name: any }` being unconstrained, but it's the only behavior change in the diff.
2. The `OperationDefinition.parametersSchema` addition — `any[]` is loose; if there's an existing parameter-schema type elsewhere, happy to use it.
3. `createTLSSelector`'s return type widened to `any` — alternative would be to make the `initialize` extension part of the actual return type.

Everything else is `as any` casting that suppresses errors without changing behavior.

This unblocks a separate harper-pro PR that bumps the submodule pointer.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Opus 4.7).

### DIFF
--- a/resources/replayLogs.ts
+++ b/resources/replayLogs.ts
@@ -26,8 +26,8 @@ export function replayLogs(rootStore: RocksDatabase, tables: any): Promise<void>
 		let lastTimestamp = 0;
 		let writes = 0;
 		let skipped = 0;
-		const txnLog: RocksTransactionLogStore = rootStore.auditStore;
-		for (const auditRecord of txnLog.getRange({ startFromLastFlushed: true, readUncommitted: true })) {
+		const txnLog: RocksTransactionLogStore = (rootStore as any).auditStore;
+		for (const auditRecord of txnLog.getRange({ startFromLastFlushed: true, readUncommitted: true }) as any) {
 			const {
 				type,
 				tableId,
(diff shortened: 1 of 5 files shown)


kriszyp:
**CI status note** (from Claude Opus 4.7 agent):

The single failing job — `Integration API Tests (Windows, Node.js v24)`, failing on Blob tests — fails identically on `main`'s most recent Integration Tests run ([26076091280](https://github.com/HarperFast/harper/actions/runs/26076091280)). Not introduced by this PR. All Linux/Bun/Mac matrices pass.

Moving to Ready for Review.

Ethan-Arrowood (approved):
Not sure I would really count `any`'s as "fixes", but I digress. Let's make the build step happy!

dawsontoth (approved):

PR #579, opened 2026-05-18
Add pre-merge cherry-pick flow for patch-labeled PRs

kriszyp:
## Summary

Mirror of [HarperFast/harper-pro#151](https://github.com/HarperFast/harper-pro/pull/151) (plus the checkout-ref fix from [harper-pro#157](https://github.com/HarperFast/harper-pro/pull/157)), ported to core.

When a PR is labeled `patch`:
- Create/reset `cherry-pick/<release>/pr-<N>` off the release branch, cherry-pick the PR commits.
- On success, dispatch both `Integration Tests` and `Unit Test` workflows against that branch. Sticky comments on the original PR link the runs and (via `report-cherry-pick-tests.yml`) update with pass/fail when each completes.
- On conflict, commit the markers, push, and @-mention Claude to suggest a resolution patch (no auto-push).
- On `unlabeled`, tear down the branch.
- On `closed && merged`, re-pick from the final merge SHA and fast-forward into the release branch.

## Differences from harper-pro

- Dispatches both `integration-tests.yml` and `unit-test.yml` (core has both; harper-pro only has integration tests).
- `report-cherry-pick-tests.yml` listens for completion of both workflows and uses a per-workflow sticky marker so the two results don't overwrite each other.

## Where to focus review

- `.github/workflows/cherry-pick-patch.yml` — the main logic.
- `report-cherry-pick-tests.yml` — gated on `head_repository.full_name == github.repository` to block fork spoofing.

🤖 Generated by Claude Opus 4.7 (1M context).

PR #609, opened 2026-05-19
fix(itc): direct-route component status responses; drop wasteful broadcast

### LINKED ISSUE #574: ITC component status: direct response fails and broadcasts unnecessarily
The ITC `componentStatusRequestHandler` cannot always send a direct response to the originating thread, and falls back to broadcasting the response to all threads. This is wasteful, noisy, and risks orphaned entries in `awaitingResponses`.

## Observed

Fabric logs show:

```
[main/0] [trace]: ITC componentStatusRequestHandler received request: {
  type: 'component_status_request',
  message: { requestId: 1, originator: 1 },
  requestId: 10
}
[main/0] [debug]: Failed to send direct response to thread 1, falling back to broadcast
```

(Log level was downgraded `warn` → `debug` after the ticket was filed, but the underlying behavior is unchanged.)

## Root cause

The fallback fires from [server/itc/serverHandlers.js](server/itc/serverHandlers.js) when `threads.sendToThread(originatorThreadId, responseMessage)` returns `false`. `sendToThread()` in [server/threads/manageThreads.js](server/threads/manageThreads.js) returns `false` when the originator's port is no longer in `connectedPorts`.

Contributing factors:

1. **Race condition** — originator thread exits / crashes / disconnects between sending the status request and the handler sending the response. The port is removed from `connectedPorts`, so `sendToThread()` can't find it.
2. **Main-thread requests have no originator** — `sendItcEvent()` in [server/threads/itc.js](server/threads/itc.js) only stamps `originator = threadId` when `!isMainThread`. Main-thread-originated status requests always trigger the broadcast fallback.
3. **Broadcast fallback is wasteful** — all threads receive a response for a request they didn't make; no mechanism to scope the fallback.

## Acceptance criteria

- [ ] Race condition: drop the response silently (or log at trace) when the originator thread is gone. No broadcast in this case — the originator is already unreachable.
- [ ] Stamp `originator` for main-thread-sent ITC events too, so direct responses work for main-thread requests.
- [ ] If broadcast is genuinely needed (no originator info available), tag the response so non-originator threads can ignore it cheaply instead of going through the full `awaitingResponses` lookup.
- [ ] Verify `awaitingResponses` cleanup timer reaps any entries that never receive a response.

## Impact

- Unnecessary broadcast traffic when direct send fails.
- Potential orphaned requests in `awaitingResponses` map (5s timeout, 30s cleanup timer).
- Was previously noisy in logs (now `debug`, but still wasteful).

---

Tracked in Jira: [CORE-3049](https://harperdb.atlassian.net/browse/CORE-3049)
Jira fixVersions (v4.7, v5.0) are stale — retargeting to **V5.2**.

kriszyp:
## Summary

`componentStatusRequestHandler` fell back to **broadcasting** the status response to every thread whenever the originator port wasn't in `connectedPorts`. That fired on every main-thread-originated request and on every race where the originator exited mid-request — wasteful and noisy.

This PR makes direct routing actually work for main-thread originators and drops the broadcast fallback entirely.

### Changes
- `server/threads/itc.js` — `sendItcEvent` now always stamps `event.message.originator = threadId`. Removed the `!isMainThread` guard. Main thread's `threadId` is `0` by `worker_threads` convention.
- `server/threads/manageThreads.js`:
  - Workers set `parentPort.threadId = 0` so `connectedPorts.sendToThread(0, ...)` routes back to main.
  - `sendToThread` wraps `postMessage` in `try/catch` — a port closing between `find()` and `postMessage()` is now treated as unreachable instead of throwing.
- `server/itc/serverHandlers.js` — when `sendToThread` returns `false`, log at `trace` and drop. The `CrossThreadStatusCollector`'s own 5s timeout already handles missing replies, so a broadcast just costs every other thread a Map lookup for a response it didn't ask for.
- `components/status/crossThread.ts` — the 30s safety-net cleanup also clears `responseCheckers` (it previously only cleared `awaitingResponses`).
- `unitTests/server/itc/serverHandlers.test.js` — new tests for both the direct-send and silent-drop paths.

## Purpose

Closes #574 (CORE-3049). Acceptance criteria from the issue:
- [x] Race condition: drop silently (trace) when originator is gone; no broadcast.
- [x] Stamp originator for main-thread events.
- [x] Broadcast no longer needed — removed entirely; non-originator threads no longer see status responses.
- [x] `awaitingResponses` cleanup safety net audited; extended to also reap stale `responseCheckers`.

## Where to look

- **`parentPort.threadId = 0`** (`manageThreads.js`) — please double-check this doesn't collide with any Node behavior I haven't considered. The Node convention is that the main thread's `threadId` is `0`; `parentPort` is a `MessagePort`, not a `Worker`, so we're free to attach a `threadId` property. All existing call sites that compare `port.threadId === someId` will find it.
- **Removing the broadcast fallback** — non-originator threads previously ignored the broadcast cheaply (Map miss in `awaitingResponses`), so the *consequences* of broadcasting were small; but the volume could be high in fabric setups. After this change, a stuck/crashed originator just times out via the collector's 5s timeout instead of getting a useless reply blast.
- **`sendToThread` try/catch around `postMessage`** — silently swallows post failures and returns `false`. This is correct for the race the issue describes, but worth a sanity check that we don't want to surface a different error class here.

## Test plan

- [x] Existing unit tests (`unitTests/server/itc`, `unitTests/components/status`) — all 156 tests pass.
- [x] New unit tests covering direct-send and silent-drop paths in `serverHandlers.test.js`.
- [ ] Manual: bring up a multi-worker fabric instance and verify the previous `[debug]: Failed to send direct response to thread N, falling back to broadcast` log is gone.

## Notes

- Cross-model review: ran the diff through Gemini 0.42. Verdict was clean — no regressions or risks identified. The only suggestion (include `requestId` in the drop trace) is applied.
- Generated by Claude Opus 4.7 (1M context). I followed the [harper-engineering-guidelines](https://harperdb.atlassian.net/wiki/spaces/SD/pages/256704558/Engineering+Ethos+and+Coding+Guidelines) lifecycle.


kriszyp:
CI status: 33/36 checks green. The 2 failing checks are pre-existing Windows-only flakes that also fail on `main` itself (4 of the last 4 main integration runs hit `Integration API Tests (Windows, Node.js v24)`):

- `Integration API Tests (Windows, Node.js v24)` — `ECONNREFUSED 127.0.0.1:9925` in `28_transactionLogs.mjs` after the http_workers restart in `27_headerTests.mjs`. The fixed 15s wait in `restartServiceHttpWorkersWithTimeout` is not always enough on Windows; `headers.test.mjs` already uses the readiness-poll variant from `lifecycle.mjs` (added in #579) but the older `tests/*.mjs` suite still uses the timed-wait shim.
- `Integration Tests 4/4 (Windows, Node.js v24)` — `npm install --ignore-scripts` in the "Relink bin scripts" step exits 1 with no useful output (Windows runner infra flake).

All Linux Unit Tests, Linux Integration Tests, Build, Lint, Format, and Socket Security checks pass.

🤖 — Claude

server/threads/manageThreads.js
```diff
@@ -55,9 +55,13 @@ connectedPorts.onMessageByType = onMessageByType;
 connectedPorts.sendToThread = function (threadId, message) {
 	if (!message?.type) throw new Error('A message with a type must be provided');
 	const port = connectedPorts.find((port) => port.threadId === threadId);
-	if (port) {
+	if (!port) return false;
+	try {
 		port.postMessage(message);
 		return true;
+	} catch {
+		// Port may have closed between find() and postMessage() — treat as unreachable.
```
dawsontoth, line 63: 💡 Could we catch more specific errors here for the exceptions we expect?

unitTests/server/itc/serverHandlers.test.js
```diff
@@ -182,5 +182,45 @@ describe('Test hdbChildIpcHandler module', () => {
 			// Trace log confirms handler received and started processing the event
 			expect(log_trace_stub).to.have.been.called;
 		});
+
+		// Tests direct-response path: when sendToThread succeeds, log a trace and do not broadcast
```
dawsontoth: The comments before the unit tests don't seem particularly useful

dawsontoth (approved):

server/threads/manageThreads.js
kriszyp: Narrowed the catch in 871427dfe — `sendToThread` now only swallows `err.code === 'ERR_CLOSED_MESSAGE_PORT'` (the documented closed-port race) and rethrows everything else. That keeps `DataCloneError` serialization bugs and other unexpected failures visible instead of silently turning into `false`. See `server/threads/manageThreads.js:55-69`.

🤖 — Claude

unitTests/server/itc/serverHandlers.test.js
kriszyp: Good call — removed the leading comments above the two new `componentStatusRequestHandler` test cases in 871427dfe (`unitTests/server/itc/serverHandlers.test.js`). The `it(...)` names already describe what each one covers.

🤖 — Claude

PR #611, opened 2026-05-19
test(integration): restore Windows skips for restart_service tests (#549)

kriszyp:
## Summary

Restores the canonical `process.platform === 'win32'` skip pattern in two integration test files so Windows CI stops failing on `restart_service http_workers` (HarperFast/harper#549).

- `integrationTests/apiTests/tests/23_blob.mjs` — re-add the suite-level skip and the two per-test `t.skip(...)` guards that existed pre-`ac846f7`. The spike commit's "DO NOT MERGE" disclaimer was ignored when it landed on main.
- `integrationTests/apiTests/headers.test.mjs` — add Windows to the suite-skip predicate alongside the existing Bun skip. The new suite (added in `9517320`) restarts http_workers in `before()` and hit the same crash as soon as it shipped.

This is a workaround, not a fix for #549.

## Goal

Get the Windows integration jobs green again on `main`. As of [run 26104711599](https://github.com/HarperFast/harper/actions/runs/26104711599):

- `Integration Tests 2/4 (Windows, Node.js v24)` — fails in `headers.test.mjs` with `Probe /CookieTest did not become ready within 60000ms after restart_service`
- `Integration API Tests (Windows, Node.js v24)` — fails in `23_blob.mjs` after the restart; every subsequent test hits ECONNREFUSED on both ops API and REST ports

Both root-cause to the same #549 crash; this PR just restores the workaround.

## Where to look

- The diff is mechanically restoring previously-existing skips. The only piece worth a second look is the comment block I added at the top of `headers.test.mjs` documenting *why* we skip on Windows — feel free to trim if it reads too verbose.
- The two per-test `t.skip(...)` guards in `23_blob.mjs` are now dead code on Windows because the parent suite is also skipped. I kept them because that's the exact pre-spike shape; happy to drop them if you'd rather collapse to suite-only.

## Caveats

- This does **not** address #549. The underlying LMDB / native-binding cleanup collision on Windows' single-worker model is still open.
- I'm the author of `headers.test.mjs` — apologies for missing the Windows skip in the original migration; the legacy `27_headerTests.mjs` also lacks the skip but uses a local fixture tarball that side-steps the slow npm install (the legacy test still triggers the crash, just behind enough other Windows failures to not be the visible signal).

## Notes for reviewer

- Cross-model review (Claude Sonnet) on the diff: confirmed correct node:test syntax, no regressions for non-Windows runs, matches pre-spike pattern.
- `prettier --check` and `oxlint --quiet` clean on both modified files.

🤖 Generated by Claude Code (Opus 4.7). PR description and commit drafted by the agent; cross-model review used Sonnet (Gemini CLI lacked an API key in this environment).

### DIFF
--- a/integrationTests/apiTests/tests/23_blob.mjs
+++ b/integrationTests/apiTests/tests/23_blob.mjs
@@ -17,7 +17,7 @@ import { createBlobCustom } from '../utils/blob.mjs';
 import { exec } from 'node:child_process';
 import { timestamp } from '../utils/timestamp.mjs';
 
-describe('23. Blob', () => {
+describe('23. Blob', { skip: process.platform === 'win32' }, () => {
 	beforeEach(timestamp);
 
 	const blobId = randomInt(1000000);
@@ -122,11 +122,8 @@ describe('23. Blob', () => {
 			.expect(200);
 	});
 
-	it('Restart Service: http workers and wait', function () {
-		// SPIKE: per-test Windows skip removed deliberately to capture the
-		// restart_service crash signature via the existing
-		// `run-integration-apiTests-windows` CI job + log artifact upload.
-		// See issue #549. DO NOT merge — diagnostic branch only.
+	it('Restart Service: http workers and wait', function (t) {
+		if (process.platform === 'win32') return t.skip('Windows: Skipping restart_service to avoid crash');
 		return restartServiceHttpWorkersWithTimeout(testData.restartHttpWorkersTimeout);
 	});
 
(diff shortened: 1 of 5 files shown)


integrationTests/apiTests/tests/23_blob.mjs
```diff
@@ -122,11 +122,8 @@ describe('23. Blob', () => {
 			.expect(200);
 	});
 
-	it('Restart Service: http workers and wait', function () {
-		// SPIKE: per-test Windows skip removed deliberately to capture the
-		// restart_service crash signature via the existing
-		// `run-integration-apiTests-windows` CI job + log artifact upload.
-		// See issue #549. DO NOT merge — diagnostic branch only.
+	it('Restart Service: http workers and wait', function (t) {
+		if (process.platform === 'win32') return t.skip('Windows: Skipping restart_service to avoid crash');
```
Ethan-Arrowood: Is this necessary with the `describe` block level skip?

dawsontoth (approved):

integrationTests/apiTests/tests/23_blob.mjs
kriszyp: Good catch — agreed it's redundant. Dropped both per-test guards in d163e1f. Kept it as-is in the first commit only to mirror the exact pre-spike shape, but suite-level is cleaner. (Reply generated by AI.)

Issue #628, opened 2026-05-20
[Models] Phase 1 — Foundation: scope.models interfaces, hdb_model_calls, TestBackend


heskew:
## Scope

Foundation for the model-access API in #510:

- TypeScript interfaces (`ModelBackend`, `Models`, options types, `ModelCallResult`, `TokenUsage`).
- `Models` facade exposed as `scope.models` on the `Scope` class — internally reads `contextStorage.getStore()` (the existing ALS at `resources/Table.ts:3484`) to pick up per-request context (tenant, signal, app).
- Backend registry (config-driven `models.<embedding|generative>.<name>` → backend instance).
- `hdb_model_calls` system table + buffered writer, declared imperatively per existing system-table convention.
- In-memory `TestBackend` for tests.
- `opts.conversationId` accepted on `GenerateOpts` (inert until #511 lands; no observable behavior in this phase).

No real backends ship in this phase — interface plus test fixture only. Real backends land in Phases 2 (`ollama`), 3 (`openai`), and 8 (`anthropic`, `bedrock`).

## Why first

Every other phase depends on the interface, the facade, the backend registry, and the analytics writer. Sub-issues that build on this: Phase 2, 3, 4, 5, 8 (#510 sub-issues) and #612 (agent-loop orchestration). External issues that consume the interface: #511 (`opts.conversationId` activation).

## API surface

```typescript
// resources/models/types.ts

export interface Models {
  embed(input: string | string[], opts?: EmbedOpts): Promise<Float32Array[]>;
  generate(input: GenerateInput, opts?: GenerateOpts): Promise<GenerateResult>;
  generateStream(input: GenerateInput, opts?: GenerateOpts): AsyncIterable<GenerateChunk>;
}

export interface ModelBackend {
  readonly name: string;
  capabilities(): ModelCapabilities;
  embed?(input: string | string[], opts: BackendOpts<EmbedOpts>): Promise<ModelCallResult<Float32Array[]>>;
  generate?(input: GenerateInput, opts: BackendOpts<GenerateOpts>): Promise<ModelCallResult<GenerateResult>>;
  generateStream?(input: GenerateInput, opts: BackendOpts<GenerateOpts>): AsyncIterable<GenerateChunk>;
}

export interface ModelCapabilities {
  embed: boolean;
  generate: boolean;
  stream: boolean;
  tools: boolean;
  adapters: boolean;
}

export type EmbedOpts = {
  model?: string;
[...]

Issue #510, opened 2026-05-12
Add unified model-access API (scope.models)


kriszyp:
## Context

The shared GPU model work (FAB-502 embeddings, FAB-503 generative + LoRA) requires a stable in-runtime API that apps target.

Without a unified API, Fabric apps and Pro / self-hosted apps will fork — Fabric using a "magic" backend, Pro users wiring their own Ollama / OpenAI SDK. That breaks the source-portability claim ("the same runtime works in all contexts").

This issue proposes `scope.models` as the single API for embedding and LLM inference, with pluggable backends, an OpenAI-compatible gateway endpoint, schema-level `@embed` directives, and built-in token accounting.

## API surface

```ts
scope.models.embed(
  text: string | string[],
  opts?: EmbedOpts
): Promise<Float32Array[]>;

scope.models.generate(
  input: GenerateInput,
  opts?: GenerateOpts
): Promise<GenerateResult>;

scope.models.generateStream(
  input: GenerateInput,
  opts?: GenerateOpts
): AsyncIterable<GenerateChunk>;

type GenerateInput = string | Message[] | { messages, tools?, system? };

type GenerateOpts = {
  model?: string;              // logical name; resolves via config
  adapter?: string;            // LoRA adapter id (FAB-503)
  temperature?: number;
  maxTokens?: number;
  responseFormat?: 'text' | 'json' | { schema };
  tools?: ToolDef[];
  toolMode?: 'return' | 'auto';   // 'auto' runs tools in-process to completion
  conversationId?: string;        // auto-bind to ConversationResource (separate issue)
  signal?: AbortSignal;
};
```

## Configuration

```yaml
models:
  embedding:
    default:
      backend: fabric              # fabric | ollama | openai | bedrock | custom
      model: harper-embed-v1
  generative:
    default:
      backend: fabric
      model: llama-3.3-70b
      adapter: harper-tenant-v1   # optional; omit to use base model
    fast:
      backend: ollama
      host: localhost:11434
      model: llama-3.2-3b
```

Resolution: logical name (`'default'`) → `models.{embedding|generative}.<name>`. Apps reference logical names so the physical model can be swapped without code changes.

Each Harper instance has its own config file, so adapter (and model) are set statically per instance. App code can also pass `adapter:` in call opts to override for a specific call.

## Backend interface

```ts
interface ModelBackend {
  embed?(input, opts: BackendOpts): Promise<Float32Array[]>;
[...]

PR #635, opened 2026-05-20
feat: expose request.signal AbortSignal, aborted on client disconnect

### LINKED ISSUE #513: Expose request.signal (AbortSignal) for Resource methods
## Context

When a client disconnects mid-stream, the SSE / WebSocket plumbing already calls `iterator.return()` on the returned async iterator (`REST.ts` line 331 for WebSocket; the Node stream pipeline for SSE), so generator bodies see their `try/finally` run. That's sufficient for cleanup that lives inside the generator.

The gap: a generator body that needs to pass a cancellation signal to *external* work — most concretely `scope.models.generateStream(input, { signal })` (#510) — has no `AbortSignal` to forward. The model call keeps tokenizing on the upstream provider after the client has gone away, wasting tokens and GPU time.

Grep confirms no `signal` / `AbortSignal` / `AbortController` in `core/server/serverHelpers/Request.ts`. There's no exposed signal on the request object.

## Proposed change

1. Construct an `AbortController` per request in the HTTP/WebSocket entry points.
2. Abort it on client disconnect (`nodeResponse.on('close', ...)` at `http.ts` line 439; `ws.on('close', ...)` at `REST.ts` line 327 — both hooks already exist).
3. Expose `request.signal: AbortSignal` so Resource methods can read and forward it.

Use:

```ts
class Chat extends Resource {
  async *post(target, body, request) {
    yield* scope.models.generateStream(body.input, { signal: request.signal });
  }
}
```

The model-access API (#510) already declares `signal?: AbortSignal` in `GenerateOpts`; this issue plugs the wire that feeds it.

## Acceptance

- [ ] `request.signal` is a live `AbortSignal` on every Resource method invocation.
- [ ] Client disconnect (HTTP, SSE, WebSocket) aborts the signal.
- [ ] Forwarding the signal to a `fetch()` / generator cancels in-flight upstream work.
- [ ] Existing behavior unaffected for methods that don't read `signal`.

## Out of scope

- Server-side timeouts that abort the signal (separate concern; can layer on later).
- Per-method timeout configuration.

heskew:
## Summary

Closes #513. Constructs an `AbortController` per `Request` and exposes `request.signal` so Resource methods can forward cancellation to external work — most concretely `scope.models.generateStream(input, { signal })` planned for #510. Without a forwardable signal, a generator body whose client has already gone away keeps tokenizing upstream, wasting tokens / GPU time.

## What changed

- **Node HTTP/SSE path** (`server/serverHelpers/Request.ts`): hook `nodeResponse.on('close', ...)` from the `Request` constructor, guarded by `!writableFinished` so clean completion does NOT abort.
- **WebSocket-upgrade path** (`server/REST.ts`): the existing `ws.on('close')` hook now calls `request._abort?.()`. The `Request` is constructed with no response in `http.ts:929`, so the constructor also attaches `nodeRequest.socket.once('close')` as a defensive fallback (idempotent abort — comment in code explains).
- **Bun runtime**: `BunRequest.signal` returns the underlying Web Request's native signal, which `Bun.serve()` already aborts on HTTP client disconnect. WS-on-Bun behavior is flagged in a code comment as implementation-defined; ad-hoc plumbing can be added when a use case arrives.
- **`isAborted`** now reflects the live signal state (was a TODO returning `false`).
- **Constructor signature** loosened: `nodeResponse?` is now optional (matching the WS-upgrade call site `http.ts:929` which has always passed one argument). The duck-typed `on` guard preserves all existing test mocks that pass `{}` as the response.

## Where to look

- `server/serverHelpers/Request.ts:64-92` — the constructor wiring + `signal` / `_abort` getters. The `writableFinished` guard is the most non-trivial bit; rationale in comments.
- `server/REST.ts:340-347` — the one-line WS hook addition.
- `unitTests/server/serverHelpers/Request.test.js` (signal describe block) — 5 new unit tests covering initial state, close-before-finish, close-after-finish, WS-upgrade socket close, and explicit `_abort()`.

## Test plan

- [x] Unit tests pass: `mocha unitTests/server/serverHelpers/Request.test.js` (59 passing — 27 pre-existing + 5 new in signal describe + 27 in withNodeAdapter/etc. that pass under typestrip)
- [x] Lint clean: `oxlint server/serverHelpers/Request.ts server/REST.ts`
- [x] Format clean: `prettier --check ...`
- [x] TypeScript build: no new errors in changed files
- [ ] CI green (will monitor)

## Reviewer attention

This unblocks #628 (Phase 1 of #510's model-access API), which reads `(ctx as any)?.signal` from the ALS-bound `Request` to plumb cancellation into model backend calls. Once this merges, that cast can drop the `as any`.

The most reviewer-worthy decisions:

1. **`writableFinished` guard on close** — is this the right Node API to distinguish "client gave up" from "we finished sending"? I believe yes (standard pattern), but flagging.
2. **Dual-abort on WS path** — both the socket `once('close')` in the constructor (when no response) AND `_abort()` from `REST.ts` will fire. Intentional redundancy because abort is idempotent and the socket fallback also covers any future single-arg `new Request(...)` site. Comment in code explains.
3. **Bun WS signal** — left as "use the underlying Web Request signal" without explicit Bun-side abort plumbing. Comment flags this as implementation-defined and worth wiring if a real use case appears.

A cross-model review (Gemini CLI) was run on the diff; the two actionable comments it surfaced are now in-code documentation (no behavioral changes).

### DIFF
--- a/server/REST.ts
+++ b/server/REST.ts
@@ -340,6 +340,7 @@ export function handleApplication(scope: import('../components/Scope.ts').Scope)
 				recordActionBinary(!hasError, 'connection', 'ws', 'disconnect');
 				incomingMessages.emit('close');
 				if (iterator) iterator.return();
+				request._abort?.();
 			});
 			try {
 				await chainCompletion;
--- a/server/serverHelpers/Request.ts
+++ b/server/serverHelpers/Request.ts
@@ -27,6 +27,7 @@ interface IncomingMessage extends NodeIncomingMessage {
 export class Request {
 	#body: RequestBody | undefined;
 	#peerCertificate: any;
+	#abortController = new AbortController();
 	public _nodeRequest: IncomingMessage;
 	public _nodeResponse?: NodeServerResponse;
 	public method: string;
@@ -60,7 +61,7 @@ export class Request {
 	public lastModified?: number;
 	public lastRefreshed?: number;
 
-	constructor(nodeRequest: IncomingMessage, nodeResponse: NodeServerResponse) {
+	constructor(nodeRequest: IncomingMessage, nodeResponse?: NodeServerResponse) {
 		this.method = nodeRequest.method;
 		const url = nodeRequest.url;
 		this._nodeRequest = nodeRequest;
@@ -68,6 +69,26 @@ export class Request {
 		this.url = url;
 		this.headers = new RequestHeaders(nodeRequest.headers);
 		this.__harperRequestUpgraded = false;
+		// Abort the request's signal on premature client disconnect. nodeResponse 'close'
+		// also fires on clean completion; the writableFinished guard restricts to disconnect.
+		if (typeof nodeResponse?.on === 'function') {
+			nodeResponse.on('close', () => {
+				if (!nodeResponse.writableFinished) this.#abortController.abort();
+			});
+		} else if (typeof nodeRequest.socket?.once === 'function') {
+			// WebSocket-upgrade path: no response. The TCP socket is the only signal.
+			nodeRequest.socket.once('close', () => this.#abortController.abort());
+		}
+	}
+	get signal(): AbortSignal {
+		return this.#abortController.signal;
+	}
+	/**
+	 * Abort this request's signal. Used by transports (e.g. WebSocket) that need to
+	 * signal client-side cancellation independently of the Node response lifecycle.
+	 */
+	_abort(): void {
+		this.#abortController.abort();
 	}
 	get absoluteURL() {
 		return this.protocol + '://' + this.host + this.url;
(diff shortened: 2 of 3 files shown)


kriszyp (approved):
Looks great!

(It looks like you followed /harper-engineering-guidelines, and ran into something I have experienced, it doesn't seem eager to move to "Ready for review" after CI finishes; maybe it was discouraged by the integration test failure/flake. But I assume this is ready for review)

heskew:
> Looks great!
> 
> (It looks like you followed /harper-engineering-guidelines, and ran into something I have experienced, it doesn't seem eager to move to "Ready for review" after CI finishes; maybe it was discouraged by the integration test failure/flake. But I assume this is ready for review)

I had stacked another task on it and it decided it didn't want to wait for the /loop step. I figured I'd switch it over manually later, after seeing the ci run.

dawsontoth (approved):

PR #355, opened 2026-04-15
Run the operations API on the main thread

kriszyp:

Issue #299, opened 2026-03-30
Run operations API on main thread only


kriszyp:
With our modern usage of the application APIs and the operations API, I think it may make sense to move operations API to the main, separated for from application code, for better separation and reliability. 

At a basic level, I think this should actually be a relatively simple change, we should easily be able to conditionally load the operations API to the main thread. There are some nuances to consider though and potentially work through:
* We still want to support programmatic access to the operations API in application code. We could still load all the operations on worker thread for direct execution. Or send operations cross-thread to the main thread. Or some combination.
* deploy_component can not load the code to test it like we do now. The component status needs to be checked afterwards.
* We have operations openapi endpoint to expose information about application exports. We would probably have to query worker threads for this information.
* We have been moving to allow applications to register custom operations for better administrative integration into the studio. Obviously the thread separation is a challenge here. We could have a mechanism for user code to run some code on the main thread. Or the main thread could query worker threads for registered functions and perform operations cross-thread (most likely choice). Or we could say the operations API really should never interact with user code, to ensure stability.

PR #652, opened 2026-05-20
fix(openapi): query worker thread for REST OpenAPI spec on main thread

kriszyp:
## Summary

- Adds two new ITC event types (`RESOURCE_OPENAPI_REQUEST` / `RESOURCE_OPENAPI_RESPONSE`) following the existing `COMPONENT_STATUS_REQUEST` pattern
- Worker handler in `serverHandlers.js` generates the OpenAPI spec from its local resources and returns it directly to the requesting thread; skips if resources are empty (guards against job-type workers responding before app workers)
- `restOpenAPIHandler` in `operationsServer.ts` falls back to cross-thread query when `resources.size === 0` (i.e. running on main thread with worker threads); uses local resources directly in single-thread mode
- 5-second timeout with 503 error if no worker responds

## Purpose

Resolves the concern from [#299](https://github.com/HarperFast/harper/issues/299) introduced by moving the operations API to the main thread ([PR #355](https://github.com/HarperFast/harper/pull/355)): application resources are only registered on worker threads, so `/api/openapi/rest` was returning an empty spec when called on the main thread.

## What to review

- **`server/operationsServer.ts`**: `queryWorkerForOpenApi` / `attachOpenApiResponseListener` — the main-thread side of the query. The `requestId` counter and pending-request map follow the same pattern as `CrossThreadStatusCollector` in `components/status/crossThread.ts`.
- **`server/itc/serverHandlers.js`**: `resourceOpenApiRequestHandler` — runs on worker threads; guards with `resources.size === 0` so workers with no registered resources (e.g. job workers at startup) don't respond.
- **Design trade-off**: All workers with resources will respond (first one wins, rest are silently discarded). An alternative would be to target only one worker, but that requires maintaining worker ID state and handling worker exit races — not worth the complexity for a low-frequency endpoint.

## Test plan

- [ ] Unit tests added in `unitTests/server/itc/serverHandlers.test.js` covering validation, happy-path response, and unreachable-originator drop — all passing
- [ ] All 413 server unit tests pass
- [ ] Manual verification: start Harper with multiple workers, call `GET /api/openapi/rest` as a super_user, confirm non-empty spec is returned

_Generated by Claude Sonnet 4.6 (agent)_

### DIFF
--- a/server/itc/serverHandlers.js
+++ b/server/itc/serverHandlers.js
@@ -21,6 +21,7 @@ const serverItcHandlers = {
 	[hdbTerms.ITC_EVENT_TYPES.SCHEMA]: schemaHandler,
 	[hdbTerms.ITC_EVENT_TYPES.USER]: userHandler,
 	[hdbTerms.ITC_EVENT_TYPES.COMPONENT_STATUS_REQUEST]: componentStatusRequestHandler,
+	[hdbTerms.ITC_EVENT_TYPES.RESOURCE_OPENAPI_REQUEST]: resourceOpenApiRequestHandler,
 };
 
 /**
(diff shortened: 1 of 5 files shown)


server/itc/serverHandlers.js
```diff
@@ -155,4 +156,46 @@ async function componentStatusRequestHandler(event) {
...
+			hdbLogger.error(validate);
+			return;
+		}
+
+		hdbLogger.trace(`ITC resourceOpenApiRequestHandler received request:`, event);
+
+		const { resources } = require('../../resources/Resources.ts');
+		if (!resources || resources.size === 0) {
+			// This thread has no registered resources — don't respond so another worker can.
+			return;
```
dawsontoth: Is there ever a case we have 0 resources across everything, and it hangs for the 5 seconds then times out?

dawsontoth (approved):

server/itc/serverHandlers.js
kriszyp: Removed the guard entirely. The worker now always calls `generateJsonApi` (even with an empty resources map) and sends back the response, so the caller gets a spec immediately rather than hanging for 5 seconds. — Claude

server/itc/serverHandlers.js
```diff
@@ -155,4 +156,42 @@ async function componentStatusRequestHandler(event) {
...
+		if (validate) {
+			hdbLogger.error(validate);
+			return;
+		}
+
+		hdbLogger.trace(`ITC resourceOpenApiRequestHandler received request:`, event);
+
+		const { resources } = require('../../resources/Resources.ts');
+		const { generateJsonApi } = require('../../resources/openApi.ts');
+		const openapi = generateJsonApi(resources, event.message.serverHttpURL);
```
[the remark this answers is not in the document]
kriszyp: Agreed. Restored the guard: workers with `resources.size === 0` return early so only an app worker with real resources replies. If no worker has resources, the caller gets a 503 after the timeout, which is more honest than an empty spec. Tests updated to verify the guard fires when resources is empty and to use a properly-wrapped `{isError: true}` entry to exercise the send/drop paths. — Claude

PR #669, opened 2026-05-21
chore: remove install README copy

### LINKED ISSUE #25: Move (or remove) the installation root README file to source code
### Feature Summary

Move (or completely remove) the README file that is created during installation. I personally never even realized it existed until I accidentally broke some path resolution logic and installation started failing. It has some light information in it, all of which is well covered by our docs. I want us to heavily consider just removing this file and including a docs link in the CLI help command or something like that. But if it must stick around, then can we at least store the relatively simple contents as a string in source so we can avoid the FS op and path resolution complexity?

I understand if the file was more complex we may want to retain it as markdown source so we could leverage like markdown dev tools (linting, formatting, link checking, etc.)

### Problem This Solves

Ref #16, similar justification to #24, we are incurring unnecessary costs by reading this README file during installation.

### Proposed Solution

- Remove the file

Alternative:
- Move file contents to a JS string

### User Stories

_No response_

### Alternatives Considered

_No response_

### Priority/Impact

Low - nice to have

### Examples or References

_No response_

### Additional Context

_No response_

No, just reporting the issue

tatakaisun:
Summary
Closes #25

Removes the install-generated root README.md path by deleting the packaged static/README.md source file and removing the install-time copy step.

Why
The install README duplicates information already covered by the public docs and requires an extra filesystem copy/path resolution step during installation.

Changes
Removed the static/README.md file that was copied into the install root.
Removed the copySync(PACKAGE_ROOT/static/README.md, <root>/README.md) step from mountHdb.
Added the docs link to harper help so users still have a discoverable documentation path.
Updated the mount_hdb unit test setup to drop the obsolete copySync stub.
Testing
npm run build
node --conditions=typestrip .\node_modules\mocha\bin\mocha unitTests\utility\mount_hdb.test.js
node .\dist\bin\harper.js help
npx prettier --check bin\harper.ts utility\mount_hdb.ts unitTests\utility\mount_hdb.test.js
git diff --check
Reference search for the old install README path and removed filesystem-structure content
Known Issues
None found.


Ethan-Arrowood (approved):
I'm in favor. And implementation looks good to me. But may need to consider if others agree with the principle of the change.

bin/harper.ts
```diff
@@ -18,6 +18,8 @@ Usage: harperdb [command]
 
 With no command, harper will simply run Harper (in the foreground)
 
+Documentation: https://docs.harperdb.io/learn
```
heskew: nit: /learn isn't docs?

Ethan-Arrowood: Good point. Could just send them to the base route `/` or `/reference` instead.

kriszyp (commented):
I thought this README.md was to help provide highly contextual information for this directory. If you are `cwd`'ed in `~/harper`, and wondering what all the folders are for, this immediately tells you. That would take a while to figure out from sifting through docs, I would think, and not particularly convenient if you are just `ls`ing around in a terminal?

Ethan-Arrowood:
I think the argument previously discussed (maybe on a call) was that how realistic is it that someone navigates to a harper installation directory... without knowing its harper? 

A. If they have no idea what Harper is, then the better question is what are they doing cd'ing around a machine with it installed
B. If they know what Harper is, and are cd'ing around an install, why wouldn't they navigate to our docs to better understand it? I don't think this file is doing all that much (its a few lines describing some directories and has no actionable follow up information). 

All that said, I'm happy to keep the file. Then lets make it actually useful. Include links to docs and whatever. Beyond that, do we need this to be a file? Can it just be a string we _write_ to a file? That goes back to the static file debate we've had before that I think landed with static files are generally okay.

kriszyp:
> If they know what Harper is, and are cd'ing around an install, why wouldn't they navigate to our docs to better understand it?

There are people that like terminals (CLIs are still wildly popular). And `--help` is generally mandatory UX (expecting a terminal user to read the docs and not have some level command line contextual guidance is generally considered poor UX; I think that is somewhat analogous here).

> Then lets make it actually useful.

:+1:

Ethan-Arrowood:
@tatakaisun instead of removing the file; can you update it to include more information such as links to common Harper resources?

tatakaisun:
Thanks for the feedback. I’ll update the PR to keep the install README and make it more useful by adding links to common Harper resources and clearer contextual guidance.

static/README.md
```diff
@@ -1,13 +1,14 @@
...
-- `harper/keys` - This folder contains the private keys (and can also have certificates) for your PKI/TLS.
-- `harper/components` - This folder contains editable components that are stored in Harper. This folder is intended for components that are edited and canonically stored on this server. The standard approach for components is that they are deployed from npm or GitHub and installed to the `hdb/components` directory, then symlinked to the `node_modules/` folder for dependency resolution purposes.
-- `harper/log` - This folder contains the log file for Harper
-- `harper/backup` - This folder contains backup copies for files when they are modified. When harper-config.yaml is modified through the API, previous versions are stored here.
+- `harper-config.yaml` - Local configuration read by Harper at startup and updated when settings change through the API.
+- `database/` - Default location for database storage files.
+- `components/` - Editable local components stored on this server.
+- `keys/` - Private keys and certificates used for PKI/TLS.
+- `log/` - Harper log output.
+- `backup/` - Backup copies of files Harper updates, such as previous `harper-config.yaml` versions.
```
cb1kenobi: nit: I enjoy when lists of directories are sorted alphabetically.

cb1kenobi (approved):
I'm cool with this.

tatakaisun:
Addressed — sorted the directory list alphabetically in `static/README.md`.

PR #747, opened 2026-05-22
feat(models): @embed directive + write-time hook + auto-HNSW (Phase 5 of #510)

### LINKED ISSUE #632: [Models] Phase 5 — @embed directive + write-time hook + auto-HNSW + model tracking
## Scope

Schema-level RAG (Retrieval-Augmented Generation) without app code. A developer marks a field with `@embed(source: "<source-field>", model: "<logical-model-name>")` and Harper auto-fires the embedding on every write, stores it on the record, and indexes it via HNSW (Hierarchical Navigable Small World) — no application-level hooks needed.

What ships:

- `@embed(source, model)` schema directive — new parsing branch in `resources/graphql.ts`.
- `setEmbedAttribute(name, embedder)` registration API on the Table class, mirroring the existing `setComputedAttribute` pattern at `resources/Table.ts:3438`.
- Write-time hook integration via the existing `write.before` / `write.beforeIntermediate` pre-commit callback pattern (`resources/Table.ts:4582` and call sites at 1345 / 1939 / 3026 / 4368).
- Automatic HNSW indexing on the `@embed`-decorated field — defaults to cosine distance; explicit `@indexed` on the same field overrides defaults.
- Per-table embedding-model tracking metadata, with invalidation when the model changes (mirrors `@computed`'s `version`-driven reindex at `resources/graphql.ts:131-145`).
- Replicated-write predicate: hook fires only on originating writes (`!ctx.replicateFrom && !ctx.alreadyLogged`) — receivers store the originator's embedding, don't re-compute.

## Why this is meaningful

Three lines of schema replace what would otherwise be a custom embedding pipeline:

```graphql
type Document @table @export {
  id: ID @primaryKey
  content: String
  embedding: Vector @embed(source: "content", model: "default")
}
```

Behavior: on write/update of `content`, embedding is computed (sync; see open decision below), stored in `embedding`, indexed via HNSW. No app-side ingest pipeline, no separate embedding service, no consistency gap between record and vector.

### Directive parsing

Add a new branch to `processGraphQLSchema()` in `resources/graphql.ts:117` alongside the existing `@computed` handler:

```typescript
} else if (directiveName === 'embed') {
  const args = directive.arguments;
  property.embed = {
    source: args.find(a => a.name.value === 'source')?.value.value,
    model: args.find(a => a.name.value === 'model')?.value.value,
  };
  // If model changes between deploys, this triggers invalidation; same as @computed's version tracking
  property.embedModelVersion = property.embed.model;
}
```

Add `'embed'` to `server.knownGraphQLDirectives` at `resources/graphql.ts:11-27` so the unknown-directive warning at line 167 stays clean.

### Registration API

Mirror `setComputedAttribute` (existing precedent at `resources/Table.ts:3438`):

```typescript
class Table {
  static setEmbedAttribute(attributeName: string, embedder: (record: any) => Promise<Float32Array>) {
    // Stores the embedder function on attribute metadata, like setComputedAttribute does
    // Embedder signature accepts the full record so source-field changes are visible
  }
}
```

At schema load, after `@embed` is parsed, the framework calls `setEmbedAttribute(name, defaultEmbedder)` where `defaultEmbedder` is:

```typescript
async (record) => {
  const sourceValue = record[embedConfig.source];
  if (sourceValue == null) return null;
  const vectors = await scope.models.embed(sourceValue, {
    model: embedConfig.model,
    inputType: 'document',                  // @embed always produces document embeddings
  });
  return vectors[0];
}
```

The developer can override via `setEmbedAttribute(name, customEmbedder)` if they need different logic (e.g. multi-field concatenation, custom preprocessing).

### Write-time hook integration

In each write path that currently wires blob pre-commit (`resources/Table.ts:1345, 1939, 3026, 4368`), add a pass for `@embed`-decorated attributes:

```typescript
// pseudocode at each put/patch/create call site
if (!ctx.replicateFrom && !ctx.alreadyLogged) {
  for (const attr of this.embedAttributes) {
    write.beforeEmbed = chainCallback(write.beforeEmbed, async () => {
      const sourceValue = record[attr.embed.source];
      if (sourceValueChanged(sourceValue, existingEntry)) {
        record[attr.name] = await attr.embedder(record);
      }
    });
  }
}
```

The transaction processor awaits `write.beforeEmbed` (or chains it onto `write.beforeIntermediate`) before commit. Failures propagate as normal pre-commit failures.

### Auto-HNSW indexing

When `@embed` is parsed, the framework automatically attaches `@indexed(type: "HNSW")` to the same attribute if no explicit `@indexed` directive is present. Default distance metric: cosine. Explicit `@indexed` parameters on the same field override the defaults entirely.

Rationale: makes the issue's "indexed via HNSW. No app code needed." statement literally true with one directive.

### Model tracking + invalidation

Table metadata records the model name that produced the stored vectors. When the table is loaded and the schema's current `@embed(model: ...)` differs from the recorded model, vectors are marked stale and re-embedded — mirrors the `@computed` version-tracking pattern at `resources/graphql.ts:131-145`.

Two re-embed modes:
- **Sync** (default): block at table load until re-embed completes. Fine for small tables; blocking for large.
- **Background**: schedule via the existing job infrastructure at `server/jobs/`; mark vectors as stale until the job catches up. Configurable.

### Replicated-write predicate

When a record arrives via replication, the originating node has already computed the embedding; the receiver should just store. The signal exists today: `resources/replayLogs.ts:50` synthesizes Context with `alreadyLogged: true`; REST sets `request.replicateFrom` when explicitly suppressed. Hook predicate: **`!ctx.replicateFrom && !ctx.alreadyLogged`**.

## Open decision (resolve in this issue or in PR review)

**Default execution mode for `@embed`: sync-by-default vs queued-by-default.**

- **Sync mode** (recommended default): commit blocks until the embedding callback resolves. Fits the existing `write.before` pattern with zero new infrastructure. Write latency includes the model round-trip.
- **Queued mode**: record commits without embedding, an async task computes the vector and back-fills via a second write. Requires additional plumbing on top of the existing job infrastructure (`server/jobs/`). Isolated, medium-sized.

Recommendation: ship **sync as the default**, with `queued` available as a per-table or per-app config opt-in. Apps that need write-latency guarantees can flip the toggle.

Decision needed before this phase implementation begins.

## Files

| Path | Change |
|---|---|
| `resources/graphql.ts` | extended — parse `@embed` directive; add `'embed'` to `knownGraphQLDirectives` |
| `resources/Table.ts` | extended — `setEmbedAttribute` API; wire write-time hook into existing pre-commit pattern; model-tracking metadata |
| `resources/models/embedHook.ts` | new — default embedder factory + queued-mode scheduler |
| `test/embed.test.ts` | new — directive parsing, sync write, replicated-write skip, model-change invalidation |

## Acceptance criteria

- [ ] `@embed(source, model)` directive parsed from GraphQL schemas; the attribute is registered with an embedder.
- [ ] Writes to a record fire the embedder, producing a vector stored in the attribute, indexed via HNSW.
- [ ] `@embed`-decorated attributes are automatically HNSW-indexed when no explicit `@indexed` is present.
- [ ] `setEmbedAttribute(name, embedder)` allows component authors to override the default embedder.
- [ ] Replicated writes skip the embedder — receivers store the originator's embedding (`!ctx.replicateFrom && !ctx.alreadyLogged` predicate verified by test).
- [ ] Model change invalidates existing vectors and triggers re-embed (sync at table load by default; queued mode configurable).
- [ ] Per-call accounting flows through `hdb_model_calls` (because the embedder calls `scope.models.embed()`; Phase 1's writer handles it). Attribution shows the originating user / tenant.
- [ ] Sync-by-default execution mode confirmed (per open decision above).
- [ ] Documented: declaring `@embed`, supplying a custom embedder, configuring queued mode, model-change behavior.
- [ ] CI green (unit + integration + 3 Node versions).

## Out of scope

- Custom embedding model training — fine-tuning lives at the application layer.
- Schema directive registry — `@embed` ships hardcoded alongside the existing directives; registry pattern is a follow-up if more model-aware directives surface.
- Cross-model migration tooling (`harper kb reembed --from X --to Y`) — separate concern; not blocking #510.
- Conversation-resource embedding wiring — that's #511's territory (its `turns` table declares its own `@embed`).
- Multimodal embeddings (image, audio) — text-only in v1; capability extension when backends support it.

## Stacks on

- **Phase 1** (#628) — needs `scope.models.embed()` and `hdb_model_calls` for accounting.
- **Any one backend** (Phase 2 #629 or Phase 3 #630) — needs something that actually produces vectors.

## Branch & PR conventions

- Branch: `feat/models-embed-directive`
- PR base: `main` (after Phase 1 and at least one backend phase merge).
- Closes this issue via `Closes #<self>`; references #510 via `Tracking: #510`.

## Smoke test

```graphql
# Schema:
type Document @table @export {
  id: ID @primaryKey
  content: String
  embedding: Vector @embed(source: "content", model: "default")
}
```

```bash
# Configure models.embedding.default to point at a real backend (Ollama or OpenAI)
# Deploy schema, then write a record:

curl -X POST http://localhost:9926/Document/ \
  -H 'Content-Type: application/json' \
  -d '{"id": "doc1", "content": "harper is a database"}'

# Expected: 201; record persisted.
# Verify: GET /Document/doc1 → embedding is populated with a real Float32Array.
# Verify: SELECT * FROM system.hdb_model_calls WHERE method = 'embed' ORDER BY $createdtime DESC LIMIT 1 — shows the embed call.

# Vector search:
curl -X POST http://localhost:9926/Document/ \
  -H 'Content-Type: application/json' \
  -d '{"operation": "search_by_value", "search_attribute": "embedding", "search_value": <a query vector>}'

# Expected: doc1 returned, ordered by cosine similarity.

# Replicated-write skip: on a replica, the embedding written by the originating node persists as-is;
# the embedder does NOT fire again. Verify by checking that no new analytics.model_call rows appear on the replica.
```

## Tracking

Part of #510. Sub-issue 5 of 6.

---

heskew:
## Summary

Phase 5 of #510. An `@embed` schema directive that, on write, embeds a source field into a vector attribute and HNSW-indexes it — no app code.

```graphql
type Document @table @export {
  id: ID @primaryKey
  content: String
  embedding: [Float] @embed(source: "content", model: "default")
}
```

Three lines replace a custom embedding pipeline.

## What ships

- **Parser** (`resources/graphql.ts`): an `@embed(source, model)` branch. Both args are required and must be string literals — a malformed directive fails schema install with a `ClientError(400)`. When the attribute has no explicit `@indexed`, HNSW indexing is auto-attached. The model is recorded as the attribute version so a model change triggers a reindex.
- **Override API** (`resources/Table.ts`): `Table.setEmbedAttribute(name, embedder)` lets a component author replace the default embedder (e.g. multi-field concatenation or a different `inputType`).
- **Write-time hook** (`resources/models/embedHook.ts`, `resources/Table.ts`): on a write whose payload includes the source field, the embedder runs and writes the vector onto the record *before* the write commits (sync-by-default). It also wires the cache-fill (`getFromSource`) path so caching tables embed on source fetch.
- **Storage shape**: vectors are stored as a plain `number[]`. Harper's record encoder does not round-trip typed arrays, so any embedder output (incl. `Float32Array`) is flattened at the boundary. Declaring the attribute `[Float]` reuses the existing array type and HNSW path — no new primitive type.
- **Replication-receiver skip**: the embedder is skipped when the write already carries the vector — cluster-replication receive (`isNotification`), REST `x-replicate-from: none`, or audit-log replay — so receivers store the originator's vector verbatim and the cluster stays in one embedding space.
- **Error sanitization**: backend URLs / model IDs / API-key tails stay in server logs; the caller sees only `Failed to compute embedding for attribute "<name>"`.

## Design notes

- **Sync-by-default** per Kris on #632 ("overwhelmingly the mode that should and would be used") — fits the existing pre-commit write path with no new infrastructure.
- **No `Vector` primitive.** A vector column is just `[Float]`; HNSW already indexes a plain `number[]`. (Reviewed and dropped in favor of reusing `[Float]`.)
- **Model-change handling is partial.** A model change reindexes the *existing* stored vectors — it does not re-embed source fields through the new model. New writes use the new model; existing rows keep their vectors until rewritten. Full re-embed backfill is tracked in #750.

## Known limitation

Proactive **source-subscribe** pushes into a caching table set `isNotification`, so they currently skip embedding (the predicate can't yet distinguish a cluster-replication receive from an external source push). The lazy `getFromSource` path and direct writes embed normally. Closing the subscribe path is folded into the derived-caching work (#750).

## Tests

- **Unit** (`unitTests/resources/models/embedHook.test.js`): default embedder, the replication-skip signals, source-presence semantics (embed on present, clear on null, skip on omit), typed-array → `number[]` normalization, CRDT-op skip, and error sanitization.
- **Integration** (`integrationTests/server/embed-directive.test.ts`): end-to-end against an in-test fake Ollama — schema install + auto-HNSW, POST/PUT/PATCH embed behavior, replication-receiver skip, and the caching-table path.

## Out of scope (follow-ups)

- **Derived caching table** (#750): low-latency writes, re-embed-on-model-change backfill, and the source-subscribe embed gap above.
- Cross-model migration tooling and multimodal embeddings.

Closes #632. Part of #510.


resources/Table.ts
```diff
@@ -4740,6 +4800,15 @@ export function coerceType(value: any, attribute: any): any {
...
 				return new Date(+value); // epoch ms number
+			case 'Vector':
+				// JSON-typed input arrives as a plain Array<number>; the storage layer
+				// expects Float32Array (what HNSW and the embedder produce). Pass typed
+				// arrays through, coerce JSON arrays, reject anything else.
+				if (value === null || value === 'null') return null;
+				if (value instanceof Float32Array) return value;
+				if (Array.isArray(value)) return Float32Array.from(value);
+				if (ArrayBuffer.isView(value)) return new Float32Array((value as ArrayBufferView).buffer);
+				throw new SyntaxError();
```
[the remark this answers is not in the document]
heskew: Applied — switched to `Float32Array.from(value)` which iterates element values rather than reinterpreting bytes. Float64Array / subarray inputs now convert correctly. Pushed at b1088047.

resources/Table.ts
```diff
@@ -3383,6 +3412,16 @@ export function makeTable(options) {
...
 				const relationship = attribute.relationship;
 				const computed = attribute.computed;
+				// `@embed` directive: register the default embedder if no override is
+				// already in place. This is one-time setup (populating userEmbedders),
+				// not a runtime resolver, so it lives outside the if/else-if chain
+				// below — `@embed` fields ALSO get auto-HNSW indexing, and the chain's
+				// `customIndex.propertyResolver` branch needs to fire for them.
+				// Component authors override the default via
+				// `Table.setEmbedAttribute(name, customEmbedder)` after schema load.
+				if (attribute.embed && !this.userEmbedders[attribute.name]) {
```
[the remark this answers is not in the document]
heskew: Applied — `setEmbedAttribute` now tracks user-set names in `static userSetEmbedders: Set<string>`; defaults refresh on every `updatedAttributes()` call, user overrides are preserved. Pushed at b1088047.

resources/models/embedHook.ts
```diff
@@ -0,0 +1,203 @@
...
+			const sourceValue = record[sourceKey];
+			if (sourceValue == null) {
+				record[attr.name] = null;
+				continue;
+			}
+			const embedder = userEmbedders[attr.name];
+			if (!embedder) continue;
+			let vector;
+			try {
+				vector = await embedder(record);
```
kriszyp: Sequential? Why not in parallel?

resources/models/embedHook.ts
```diff
@@ -0,0 +1,203 @@
...
+			anySourcePresent = true;
+			break;
+		}
+	}
+	if (!anySourcePresent) return undefined;
+	return async (): Promise<void> => {
+		for (const attr of embedAttributes) {
+			const sourceKey = attr.embed?.source;
+			if (!sourceKey) continue;
+			if (!Object.prototype.hasOwnProperty.call(record, sourceKey)) continue;
```
kriszyp: So paranoid. We have prototype pollution protection (freezing). (not a blocker, just a nit)

resources/Table.ts
```diff
@@ -4740,6 +4819,17 @@ export function coerceType(value: any, attribute: any): any {
 					return date;
 				}
 				return new Date(+value); // epoch ms number
+			case 'Vector':
```
kriszyp: This is a clever idea. But does it really mean only Float32? Or should this also allow a Float64 and Float16? (seems like it should). (I would actually love to move to using Float16 in HNSW)

kriszyp (changes requested):
It seems like we are entirely missing support for embeddings with `getFromSource` writes. And using embeddings with caching tables actually seems like one of the most compelling uses cases.

My Gemini review caught a number of things:

  ### 1. Unhandled Promise in  TableResource.update  leads to Transaction Commits before Embed resolves — CRITICAL (Confirmed)                                                                                                             
                                                                                                                                                                                                                                           
  File: Table.ts                                                                                                                                                                                                                     
  Domain: concurrency                                                                                                                                                                                                                      
  What: The  update(updates, fullUpdate)  instance method of  TableResource  calls  this._writeUpdate(...)  which returns a  Promise  when  @embed  is active. However,  update()  completely ignores this promise and returns  this       
  synchronously.                                                                                                                                                                                                                           
  Why it matters: Because  update()  returns synchronously, the calling application will immediately proceed to run  await instance.save() . However, the asynchronous embedder operation is still pending, meaning the write has not been 
  registered in the transaction's write list yet. The transaction will either commit without saving the updated record (causing silent data loss) or crash with a "Transaction already closed" error when the embedder finally resolves and
  tries to invoke  addWrite()  on a committed transaction.                                                                                                                                                                                 
  Code:                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                           
    // lines 1243-1247                                                                                                                                                                                                                     
    return when(loading, () => {                                                                                                                                                                                                           
        this.#changes = updates;                                                                                                                                                                                                              
        this._writeUpdate(id, this.#changes, false);                                                                                                                                                                                          
        return this;                                                                                                                                                                                                                          
    });                                                                                                                                                                                                                                    
    ...                                                                                                                                                                                                                                    
    // line 1251-1252                                                                                                                                                                                                                      
    this._writeUpdate(id, this.#changes, fullUpdate);                                                                                                                                                                                      
    return this;                                                                                                                                                                                                                           
                                                                                                                                                                                                                                           
  Suggested fix:                                                                                                                                                                                                                           
  Wrap both invocations of  _writeUpdate  inside the  update  method with Harper's  when()  helper so that the promise is properly propagated back to the caller when an asynchronous hook is active:                                      
                                                                                                                                                                                                                                           
    return when(loading, () => {                                                                                                                                                                                                           
        this.#changes = updates;                                                                                                                                                                                                              
        return when(this._writeUpdate(id, this.#changes, false), () => this);                                                                                                                                                                 
    });                                                                                                                                                                                                                                    
    ...                                                                                                                                                                                                                                    
    return when(this._writeUpdate(id, this.#changes, fullUpdate), () => this);                                                                                                                                                             
  ──────                     

  ### 2.  coerceType  for Vector returns  Float32Array  which is mangled by the Record Encoder — HIGH (Confirmed)                                                                                                                          
                                                                                                                                                                                                                                           
  File: Table.ts                                                                                                                                                                                                                     
  Domain: data-integrity                                                                                                                                                                                                                   
  What:  coerceType  converts incoming  Vector  attribute values into a  Float32Array . However, as documented in  createDefaultEmbedder , Harper's record encoder ( updateAndFreeze ) does not cleanly serialize typed arrays, instead    
  mangling them into  {0: val1, 1: val2, ...}  objects in the database.                                                                                                                                                                    
  Why it matters: While the default embedder circumvents this by casting its result back to  Array<number> , any user-supplied vectors (e.g., replication receiver paths, programmatic writes, or manual API uploads) will pass through    
  coerceType  and be coerced to a  Float32Array . These will get stored in the database in a mangled object format, corrupting the vector column and breaking subsequent HNSW indexing and search operations.                              
  Code:                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                           
    case 'Vector':                                                                                                                                                                                                                         
        if (value === null || value === 'null') return null;                                                                                                                                                                                  
        if (value instanceof Float32Array) return value;                                                                                                                                                                                      
        if (Array.isArray(value)) return Float32Array.from(value);                                                                                                                                                                            
        if (ArrayBuffer.isView(value)) return Float32Array.from(value as any);                                                                                                                                                                
        throw new SyntaxError();                                                                                                                                                                                                              
                                                                                                                                                                                                                                           
  Suggested fix:                                                                                                                                                                                                                           
  Coerce vectors to standard JavaScript arrays ( number[] ) instead of  Float32Array s, ensuring they bypass record-encoder mangling while remaining compatible with HNSW:                                                                 
                                                                                                                                                                                                                                           
    case 'Vector':                                                                                                                                                                                                                         
        if (value === null || value === 'null') return null;                                                                                                                                                                                  
        if (value instanceof Float32Array) return Array.from(value);                                                                                                                                                                          
        if (Array.isArray(value)) return value.map(Number);                                                                                                                                                                                   
        if (ArrayBuffer.isView(value)) return Array.from(value as any);                                                                                                                                                                       
        throw new SyntaxError();                                                                                                                                                                                                              
  ──────          

  ### 3. Model Change via Schema Reload causes Dimension Mismatches or Mixed-Model Vectors in HNSW — MEDIUM (Confirmed)                                                                                                                    
                                                                                                                                                                                                                                           
  File: graphql.ts and embedHook.ts (comments)                                                                                                                                                                                       
  Domain: data-integrity / config                                                                                                                                                                                                          
  What: Modifying the  @embed(model: "...")  argument in a schema file updates the attribute version, triggering a schema reload and an HNSW reindex. However, existing rows are not automatically re-embedded through the new model.      
  Why it matters: If a user changes their embedding model (e.g., swapping a 384-dimension model for a 1536-dimension model), existing rows retain their old-dimension vectors. Loading these mixed-dimension vectors into the same HNSW    
  index                                                                                                                                                                                                                                    
  will either trigger a crash during index construction or severely pollute the embedding space, returning garbage query results.                                                                                                          
  Suggested fix:                                                                                                                                                                                                                           
                                                                                                                                                                                                                                           
  1. Document this operational hazard explicitly.                                                                                                                                                                                          
  2. Consider adding validation to block schema updates that change model definitions on populated tables unless the vector attribute is cleared, or plan a background job pathway to invalidate and backfill the vector column when a     
  model version change is detected.
  ──────
  ### 4. Malformed  @embed  Directives Fail Silently during Schema Deployment — LOW (Confirmed)
  
  File: graphql.ts 
  Domain: api / config
  What: When a malformed  @embed  directive is encountered (e.g., omitting  source  or  model , or supplying non-string literals), the schema compiler logs via  console.error  and skips registering the  @embed  property metadata, but  
  continues compiling.
  Why it matters: Because no exception is thrown, the overall schema installation succeeds with a  200 OK . The field is silently registered as a regular, un-embedded  Vector  attribute. Users receive no indication that their  @embed  
  directive was discarded unless they inspect the server console logs.
  Code:
  
    if (!embedDefinition.source || !embedDefinition.model) {
        console.error(
            `@embed on "${property.name}" requires both "source" and "model" arguments, at`,
            directive.loc
        );
    }
  
  Suggested fix:
  Throw a descriptive  ClientError  to halt compilation and fail the schema deployment loudly, returning a helpful error message to the client:
  
    if (!embedDefinition.source || !embedDefinition.model) {
        throw new ClientError(
            `@embed on "${property.name}" requires both "source" and "model" arguments`,
            400
        );
    }
  ──────
  ### 5. Dependency on  globalThis.logger  in  embedHook.ts  — LOW (Likely)
  
  File: embedHook.ts 
  Domain: config
  What: Failure logging in  embedHook.ts  relies on resolving the logger dynamically via  (globalThis as any).logger  rather than using standard module imports.
  Why it matters: Relying on  globalThis  for logger references deviates from ES module standards, bypasses static type checks, and can lead to unhandled runtime exceptions if the logger isn't populated on  globalThis  in specific     
  worker threads.
  Code:
  
    const logger = (globalThis as any).logger;
    logger?.error?.(`Embedder for attribute "${attr.name}" failed:`, err);
  
  Suggested fix:
  Import the standard Harper logger statically:
  
    import { logger } from '#src/utility/logging/logger.ts';

resources/Table.ts
heskew: Two parts to your question:

**Input/coercion side** (`coerceType`): this is not limited to `Float32Array`. The explicit `Float32Array` check is just a fast path / documentation signal; the `ArrayBuffer.isView` fallback converts typed-array views with `Array.from`, so `Float64Array`, `Float16Array` on Node versions that expose it, and other numeric typed arrays are accepted. Small caveat: this is the query/PK coercion path; write-time `@embed` normalization happens in `embedHook.normalizeVector`, because `Table.validate()` does not currently have a `Vector` case.

**F16 in HNSW**: agreed that Float16 is worth pursuing. The HNSW index store is currently forced to Float32 (`HierarchicalNavigableSmallWorld.ts:106-108`, including the existing "we would actually like to use float16 if it were available" comment). `msgpackr` exposes `useFloat32`, but not a comparable `useFloat16` encoder option in the version pinned here. Moving the HNSW index to F16 would need new encoder support or a custom ext type. That seems separate from this PR and a good follow-up against #510 / #750.

One storage distinction: the HNSW *index* store is not the same as the primary row store. The primary record store uses the Harper `RecordEncoder`, which extends the default msgpackr `Encoder` without `structuredClone`. This repo pins `msgpackr@1.11.12`; with default encoding, typed arrays do not round-trip: `Float32Array(N)` decodes as a `Buffer` instead of a typed array. `structuredClone: true` helps for some typed arrays like `Float32Array` / `Float64Array`, but not `Float16Array` in my local probe, so a Vector-specific ext type would likely be the more complete row-storage fix. That is why this PR flattens vectors to `Array<number>` at the boundary today.

---

🤖 Posted by Claude on behalf of Nathan (and fixed by codex)

resources/DatabaseTransaction.ts
```diff
@@ -202,21 +202,46 @@ export class DatabaseTransaction implements Transaction {
...
+			// the txn completions so the whole txn waits.
+			const validateResult: any = operation.validate?.(txnTime);
+			if (validateResult?.then) {
+				return validateResult.then((res: any) => {
+					if (res === false) {
+						operation.commit = () => {}; // noop if we try again
+						return;
+					}
+					return runBeforeAndCommit();
+				});
```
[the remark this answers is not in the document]
heskew: Stale — the entire validate-time refactor (Phase 1/3/4 commit pipeline, async-validate plumbing in `DatabaseTransaction.save` / `LMDBTransaction.commit`) was reverted in `e0233037`. The `@embed` hook is back to running at `_writeUpdate`-time, which both reviewers had approved in an earlier round; the codex blocker that drove the validate-time refactor (instance-mutation `update(id, {}); row.x = ...; save()` not picking up `@embed`) is now a documented known limitation in `embedHook.ts`, with the proper fix tracked as a follow-up at the resource layer.

---

resources/Table.ts
```diff
@@ -3739,31 +3828,30 @@ export function makeTable(options) {
 	function checkValidId(id) {
 		switch (typeof id) {
 			case 'number':
-				if (isNaN(id)) throw new ClientError('Invalid primary key of NaN', 400);
 				return true;
```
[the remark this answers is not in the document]
heskew: Fixed in `441bcf46` — restored Chris's `d10b483b` `checkValidId` block verbatim: `isNaN` guard reinstated on the numeric branch, all four invalid-PK throw sites converted back to `ClientError(_, 400)`. The regression was a side-effect of the validate-time refactor revert's `git checkout 05f19e60 -- resources/Table.ts`, which inadvertently rolled the file back past Chris's fix. Verified locally: `unitTests/apiTests/undefinedIdInUrl-test.mjs` 18/18 pass.

---

resources/Table.ts
```diff
@@ -3739,31 +3842,30 @@ export function makeTable(options) {
...
 			default:
-				throw new ClientError('Invalid primary key type: ' + typeof id, 400);
+				throw new Error('Invalid primary key type: ' + typeof id);
 		}
 		// otherwise it is difficult to determine if the key size is too large
 		// without actually attempting to serialize it
 		const length = writeKey(id, TEST_WRITE_KEY_BUFFER, 0);
-		if (length > MAX_KEY_BYTES) throw new ClientError('Primary key size is too large: ' + id.length, 400);
+		if (length > MAX_KEY_BYTES) throw new Error('Primary key size is too large: ' + id.length);
 		return true;
```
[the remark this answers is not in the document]
heskew: Fixed in `441bcf46` — see the parent thread for context.

---

resources/graphql.ts
```diff
@@ -143,6 +157,57 @@ async function processGraphQLSchema(gqlContent, urlPath, filePath, resources) {
...
+								// operator signal. Halt schema processing so the component install
+								// surfaces the error instead of returning 200 OK on a broken
+								// schema. Other directive validations in this parser (e.g. dup
+								// primary key at line ~119) only log and continue — the difference
+								// here is that the failure mode is "looks-correct schema that
+								// never embeds anything," which is harder to diagnose than a
+								// missing primary key.
+								const loc = directive.loc;
+								console.error(`@embed on "${property.name}" requires both "source" and "model" arguments, at`, loc);
+								throw new Error(
```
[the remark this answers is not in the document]
heskew: Fixed in `b0cf804c` — `@embed` validation throws `ClientError(..., 400)` and `ClientError` is imported alongside the existing utility-error imports. Aligns malformed-directive schema-install responses with the rest of Harper's error contract.

---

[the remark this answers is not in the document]
heskew: Fixed in `b0cf804c` — see the parent thread.

---

resources/Table.ts
```diff
@@ -3566,6 +3657,28 @@ export function makeTable(options) {
...
+		/**
+		 * Override the embedder that fires on writes to an `@embed`-decorated
+		 * attribute (Phase 5 of #510). The default embedder calls
+		 * `Models.embed(record[source], { model, inputType: 'document' })`. Pass a
+		 * custom function when you need multi-field concatenation, custom
+		 * preprocessing, or a different inputType — the embedder receives the
+		 * full record (post-merge with the existing entry for PATCH) and returns
+		 * the vector to store at `attribute_name`.
+		 */
+		static setEmbedAttribute(attribute_name: string, embedder: Embedder): void {
```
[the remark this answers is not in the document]
heskew: Fixed in `b0cf804c` — the JSDoc now says the embedder receives the WRITE PAYLOAD (fields present in the PUT/PATCH body), NOT the post-merge record, and points readers at the known-limitation section in `embedHook.ts`. The behavior fix (running embed against the post-merge record) is the same resource-layer follow-up tracked there.

---

resources/models/embedHook.ts
heskew: Addressed in `47bd103c` — the per-attribute embedder loop in `buildEmbedBefore` is now `Promise.all(embedAttributes.map(...))` (`resources/models/embedHook.ts:221`). Typical schemas have one `@embed` per table so the parallelism rarely matters in practice, but multi-`@embed`-per-table no longer serializes embedder HTTP roundtrips. Resolving since this thread is marked outdated; please reopen if my read is off.

---

resources/models/embedHook.ts
heskew: Addressed in `47bd103c` — switched to `sourceKey in record` (`resources/models/embedHook.ts:209` and `:225`); we lean on Harper's existing freeze-based prototype-pollution protection rather than the defensive `hasOwnProperty.call`. Resolving since this thread is marked outdated.

---

resources/Resource.ts
```diff
@@ -107,9 +107,29 @@ export class Resource<Record extends object = any> implements ResourceInterface<
 	static put = transactional(
 		function (resource: any, query: RequestTarget, request: Context, data: any) {
 			if (Array.isArray(data) && resource.#isCollection && resource.constructor.loadAsInstance !== false) {
+				const resourceClass = resource.constructor;
+				// For tables with `@embed` attributes, sequence per-element resource.put() calls
+				// so a payload with duplicate ids commits in submission order rather than
+				// embedder-resolution order. Each `resource.put` triggers the embedder hook in
+				// `_writeUpdate`; `Promise.all` over independent resources would otherwise let
+				// embedder timing decide the winner. Non-`@embed` tables stay parallel.
```
kriszyp: What is the reason for putting this in Resource.ts? I don't think there is such a thing as a "Non-`@embed` tables" (you can't declare custom resources in a graphql schema, and seems presumptuous to apply this logic to a custom resource).

resources/Table.ts
kriszyp: How is this different than `Float[]` then? Should we just move all the coercion logic in there? (use that for vectors)

resources/Table.ts
```diff
@@ -3566,6 +3657,35 @@ export function makeTable(options) {
 			}
 			this.userResolvers[attribute_name] = resolver;
 		}
+		/**
+		 * Override the embedder that fires on writes to an `@embed`-decorated
+		 * attribute (Phase 5 of #510). The default embedder calls
```
kriszyp: Avoid transient information, keep comments concise and focused on required knowledge to understand code.

resources/Table.ts
```diff
@@ -3383,6 +3461,19 @@ export function makeTable(options) {
...
 				const computed = attribute.computed;
+				// `@embed` directive: register the default embedder. This is one-time
+				// setup (populating userEmbedders), not a runtime resolver, so it lives
+				// outside the if/else-if chain below — `@embed` fields ALSO get
+				// auto-HNSW indexing, and the chain's `customIndex.propertyResolver`
+				// branch needs to fire for them. Component authors override the default
+				// via `Table.setEmbedAttribute(name, customEmbedder)` after schema load.
+				// `userSetEmbedders` tracks the override names so we can refresh defaults
+				// on an in-place `Table.attributes.splice()` schema reload at
+				// `databases.ts:940` (which calls `updatedAttributes()` again on the
```
kriszyp: Fragile reference, keep concise.

kriszyp (commented):
Good progress. I don't think we necessarily want this in Resource.ts, and am still skeptical of the new `Vector` type.
Also some of the comments are aimed at Claude's comment generation tactics. I tend to think its training has been influenced by more-comments-more-better mentality, but I would tend to believe that future agents' tokens are much better spent on reading code itself, than long commentaries on the code.

resources/Table.ts
heskew: yes! sigh... :)

heskew:
Thanks Kris — all three addressed in `5e5ce295`:

- **`Vector` type dropped.** You were right it's redundant — HNSW already indexes plain `[Float]` columns, so the declaration is now `embedding: [Float] @embed(...)`. Removed the primitive and its `coerceType` case. Also sidesteps the Float64/Float16 question — everything's `number[]` now; F16-in-HNSW is a separate follow-up.
- **Out of `Resource.ts`.** Removed the `@embed` special-casing from the base class, along with the duplicate-id sequencing it existed to serve — that payload-order guarantee was speculative, so it and its test are gone. The async embed hook is still awaited before commit via `when()` on the put/patch/update paths.
- **Comments.** Full pass — no more file:line refs, commit hashes, or process narrative. `embedHook.ts` went 286→136 lines. Point taken on tokens better spent reading the code.

−377 lines net. Codex review clean; unit + integration suites green.

---

kriszyp (approved):
This is going to be awesome, good job.

kriszyp:
Heads-up on a friendly seam where this meets the HNSW int8 work in #894.

#894 makes the **index** store vectors as int8 for fast graph traversal (the per-node-visit vector decode was 54–71% of search/insert CPU; quantizing the index gave ~4.9× search and a ~3× smaller index). This PR's `Vector` type stores the **record** vector as `Float32Array` — which turns out to be the ideal substrate for the planned over-fetch + rerank step (recompute exact distances from the record after int8 navigation, to recover exact `$distance` and recall): a `Float32Array` record makes that rerank decode cheap.

One small coordination point, no conflict (#894 is index-only): after this lands, `HierarchicalNavigableSmallWorld.index()` may receive a `Float32Array` rather than a `number[]`. `quantizeInt8` already works on it at runtime (it's array-like), but we should widen the param type to `number[] | Float32Array` when the two meet. I'll track it on the rerank issue.

Nice work on the `@embed` flow — three lines replacing a pipeline is a great DX win.

— Claude

PR #761, opened 2026-05-23
test(integration): port delete operations tests to self-contained framework

kriszyp:
## Summary

- Ports `8_deleteTests.mjs` (76 tests) to the new per-suite isolated Harper instance pattern in `integrationTests/apiTests/delete.test.mjs`
- Each suite starts its own Harper instance via `startHarper`/`teardownHarper`
- `before()` creates all required fixtures: `northnwd.employees`, `dev.rando` (4 pre-seeded records), and numeric-string schemas `123.4`/`1123.1` that the legacy suite inherited from `1_environmentSetup.mjs`
- Uses `awaitJob`/`getJobId` from `utils/operations.mjs` for async job polling

## Attention areas

- The `delete_files_before` tests depend on a `insertTimestamp` variable captured between two insert batches — this replaces the legacy `testData.insert_timestamp` mutable shared state. The timing is tight: we capture the ISO timestamp immediately before the second batch, which should be sufficient but is worth verifying on slow CI runners.
- The `setTimeout` delays from the legacy file are preserved (27s total). These were in the original for disk-async reasons; I left them rather than risk flakiness, but they could be reduced in a follow-up if CI confirms they're unnecessary.
- Codex (P1) and Gemini both flagged the missing numeric-string schema fixtures — fixed in the second commit.

🤖 Generated by Claude Sonnet 4.6

### DIFF
--- a/integrationTests/apiTests/delete.test.mjs
+++ b/integrationTests/apiTests/delete.test.mjs
@@ -52,6 +52,14 @@ suite('Delete operations', (ctx) => {
 				records: [{ id: 987654321 }, { id: 987654322 }, { id: 987654323 }, { id: 987654324 }],
 			})
 			.expect(200);
+
+		// Numeric-string schemas/tables used by the "Drop number" tests.
+		// In the legacy suite these were created by 1_environmentSetup.mjs;
+		// here we create them in before() so the suite is fully self-contained.
+		await client.req().send({ operation: 'create_schema', schema: '123' }).expect(200);
+		await client.req().send({ operation: 'create_table', schema: '123', table: '4', primary_key: 'id' }).expect(200);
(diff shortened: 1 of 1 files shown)


kriszyp:
## CI Status ✅

All integration tests pass (including our delete operation tests in shard 3/4 and 4/4 on all platforms: Node.js v22/v24/v26, Bun, and Windows).

**Failing checks are pre-existing flakes affecting the entire repo (not caused by this PR):**
- **Unit Tests (v22/v24/v26)**: `logRotator.test.js` timeout — failing on \`main\` and all other open PRs too
- No integration test failures on our changes

Ready for review and merge whenever convenient! 🎉

— Claude

Issue #621, opened 2026-05-19
[MCP] Stdio CLI (harper mcp) with UDS-first, network fallback


kylebernhardy:
**Scope.** `harper mcp` CLI subcommand: a full MCP Streamable HTTP client over stdin/stdout, defaulting to a local UDS connection (no credentials needed) and falling back to HTTPS when a remote URL is given.

**Design reference.** Sections "stdio (`harper mcp`)" and "Client Configuration & Invocation → 1. Desktop / IDE clients (stdio via `harper mcp`)" in #465.

**Acceptance criteria**
- [ ] New `harper mcp` subcommand of the existing Harper CLI; implementation under `cli/mcp/`.
- [ ] Two connection modes:
  - **Local UDS (default).** Connects to `operationsApi_network_domainSocket` (or the application-port UDS for `--profile application` when configured). Uses the same pattern as `bin/cliOperations.ts:119-131`. **No credentials required** — gated by filesystem permissions.
  - **Network HTTPS.** Used when `--url` / `HARPER_URL` points off-box. Credentials required (`HARPER_TOKEN`, `HARPER_AUTH`, or `HARPER_USER`/`HARPER_PASS`).
- [ ] Resolution order: `--socket` → auto-detect local UDS from Harper config → `--url` → `HARPER_URL`. Clear error if none resolves.
- [ ] CLI implements a full MCP Streamable HTTP client: stdin JSON-RPC → upstream POST/GET; parses `application/json` and `text/event-stream`; emits messages to stdout; persists `Mcp-Session-Id`; opens GET SSE channel after `initialize`.
- [ ] Strips `Authorization` and credential env vars from any error output it writes.
- [ ] CLI flags: `--profile`, `--socket`, `--url`, `--mount-path`, `--protocol-version`, `--insecure`.
- [ ] `harper mcp print-config --client claude-desktop --profile application` emits a paste-ready JSON block for the named MCP client.
- [ ] `harper mcp doctor` runs `initialize` against the resolved upstream and prints pass/fail per step (TLS, auth, Origin, version, tool count).
- [ ] Unit + integration tests for both modes; integration test must verify zero-credential local-UDS happy path.

**Out of scope.** OAuth device-flow auth (v1.1). Profile switcher / saved-config file (skipped per #465).

**Stacks on.** #614 (transport). Can run in parallel with #615–#620.

**Branch & PR conventions**
- Branch: `feat/mcp-stdio-cli`
- PR base: `main` (after #614 merges).
[...]

Issue #620, opened 2026-05-19
[MCP] Audit logging + per-session rate limiting


kylebernhardy:
**Scope.** Per-tool audit logging and per-session rate limiting. Both required by the MCP spec (§server/tools Security Considerations).

**Design reference.** Sections "Safety & Observability" and "Rate Limiting (v1, required by spec)" in #465.

**Acceptance criteria**
- [ ] Every `tools/call` invocation is written to Harper's existing audit log channel with: timestamp, profile, session id, tool name, argument summary (with PII-redaction hook), authenticated user id/role, result status (`ok` / `isError` / protocol error), duration.
- [ ] Rate limiter (token bucket) per `(session, tool)`:
  - `mcp.<profile>.rateLimit.perToolPerSecond` (default 10 ops, 25 app)
  - `mcp.<profile>.rateLimit.perToolBurst` (default 20 ops, 50 app)
  - `mcp.<profile>.rateLimit.sessionConcurrency` (default 25 ops, 50 app)
  - `mcp.<profile>.rateLimit.sessionPerSecond` (default 100 ops, 200 app)
- [ ] When a limit fires, the call returns `result.isError = true` with `kind: "rate_limited"` — **not** a JSON-RPC error.
- [ ] Documented as session-scoped: this is not a global REST/operations rate limit. Out-of-scope global limiting is referenced in the Rate Limiting section of #465.
- [ ] Unit tests cover audit-log shape, burst behavior, concurrency clamp, and the `isError` mapping. No new `sinon` or `rewire` (per `AGENTS.md:137`).

**Out of scope.** Global REST/operations rate limiting (not a Harper feature today; not in this design).

**Stacks on.** #615 (tool registry).

**Branch & PR conventions**
- Branch: `feat/mcp-audit-ratelimit`
- PR base: `main` (after #615; can be parallel with #617 / #618 / #621).
- Closes this issue; references #465.

**Smoke test**
```bash
# Send 30 search_product calls in a tight loop.
# Expected: first ~20 succeed; remainder return isError=true with kind=rate_limited.
# Audit log shows all 30 entries with correct status.
```

**Tracking.** Part of HarperFast/harper#465. Sub-issue #8 of 11.

kriszyp:
I am not sure what  "Harper's existing audit log channel" is, but sounds good if you can find one.

Rul1an:
[...]

Issue #619, opened 2026-05-19
[MCP] listChanged notifications + per-session bookkeeping


kylebernhardy:
**Scope.** Per-session bookkeeping for active MCP sessions, plus the two listChanged notification flows: `notifications/tools/list_changed` when a role mutates and `notifications/resources/list_changed` when the schema mutates.

**Design reference.** Section "`tools/list_changed` session bookkeeping" in #465.

**Acceptance criteria**
- [ ] In-memory session table keyed by `Mcp-Session-Id`: `{ user, role, profile, sseStream? }`. Idle sessions terminated per `mcp.session.idleTimeoutSeconds`.
- [ ] Subscribe to Harper's existing role-cache-invalidation events (`security/user.ts`); on event, recompute `tools/list` for affected sessions, diff against the cached list, send `notifications/tools/list_changed` over each session's SSE channel.
- [ ] Subscribe to Harper's existing schema-reload events (`create_schema` / `create_table` / `drop_table` / GraphQL reload); on event, recompute `resources/list` for affected sessions and send `notifications/resources/list_changed`.
- [ ] **Never broadcast.** Notifications are computed per-session so users never learn about schema/role changes for objects they cannot see.
- [ ] Sessions reaching idle timeout return HTTP 404 on subsequent requests (per the transport spec from #614).
- [ ] Unit + integration tests: open two sessions with different roles; mutate one role; verify only the affected session receives `notifications/tools/list_changed`.

**Out of scope.** Resumability via `Last-Event-ID` (v2). `resources/subscribe` (v2). Cross-process session sharing (v2).

**Stacks on.** #615, #616, #617, #618 (needs both profiles' tool lists and the resources capability).

**Branch & PR conventions**
- Branch: `feat/mcp-listchanged-sessions`
- PR base: `main` (after dependencies merge).
- Closes this issue; references #465.

**Smoke test**
```bash
# Open SSE on session A; alter_role for the user attached to session A in another session B.
# Expected: SSE event "notifications/tools/list_changed" arrives on session A's GET /mcp stream.
# Session A's next tools/list returns the new filtered set.
```

**Tracking.** Part of HarperFast/harper#465. Sub-issue #7 of 11.

kylebernhardy:
Landed in PR #856 (consolidated PR-1 `feat/mcp-tools`) — bundled with #617, #618, #619, #620, #622 per @kriszyp's review-load preference. See umbrella #465 for the full delivery table.

Issue #616, opened 2026-05-19
[MCP] Resources capability: resources/list, read, templates/list (https:// + harper://)


kylebernhardy:
**Scope.** Implement the MCP `resources/` capability: `resources/list`, `resources/read`, `resources/templates/list`. App-exported resources use their canonical `https://host:port/<path>` URIs (resolved in-process via `Resources.getMatch(url)`, **no outbound HTTP**); synthetic resources use the `harper://` scheme (`harper://schema/{db}/{table}`, `harper://openapi`, `harper://about`, `harper://operations`).

**Design reference.** Sections "URI schemes: `https://` (default for app) and `harper://` (synthetic only)" and "MCP `resources/` capability" in #465.

**Acceptance criteria**
- [ ] `resources/list` enumerates resources the authenticated user can access (RBAC-filtered against the role permission walk used in #615).
- [ ] `resources/read` for an `https://...` URI dispatches through `Resources.getMatch(url)` in-process; never makes an outbound HTTP request.
- [ ] `harper://schema/{db}/{table}` reads `Table.attributes`, filters by the user's `attribute_permissions`.
- [ ] `harper://openapi` returns the existing `generateJsonApi()` output (`resources/openApi.ts`), RBAC-filtered.
- [ ] `harper://about` returns server version, profile, enabled features.
- [ ] `harper://operations` (operations profile only) returns the user-filtered operations catalog with JSON Schemas.
- [ ] `resources/templates/list` advertises both URI templates (the `https://...` template and `harper://schema/{database}/{table}`).
- [ ] Server advertises `resources.listChanged: true` in `initialize` capabilities. `subscribe` is NOT advertised in v1.
- [ ] Unit + integration tests cover both URI schemes and RBAC filtering.

**Out of scope.** `resources/subscribe` (v2). Schema-change notifications (`resources/list_changed`) ship in #619.

**Stacks on.** #614 (transport).

**Branch & PR conventions**
- Branch: `feat/mcp-resources`
- PR base: `main` (after #614 merges; can be parallel with #615).
- Closes this issue; references #465.

**Smoke test**
```bash
curl -sS -X POST http://localhost:9925/mcp ... \
  -d '{"jsonrpc":"2.0","id":3,"method":"resources/read","params":{"uri":"harper://schema/shop/product"}}'
# Expected: contents[0] with mimeType application/json and a JSON body containing
# attributes, primaryKey, relationships, with attribute_permissions applied.
```

[...]

PR #763, opened 2026-05-24
[MCP] Streamable HTTP transport

### LINKED ISSUE #614: [MCP] Streamable HTTP transport: session, lifecycle, Origin, contentTypes reuse

kylebernhardy:
## Summary

Replaces the #613 503 stub with a real MCP Streamable HTTP transport per spec [rev 2025-06-18](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#streamable-http). Wires the application-profile registration that #613 deferred. Transport plumbing only — tools/resources/listChanged/rate-limit/audit stay out of scope (sub-issues #615-#620). Unknown JSON-RPC methods return `-32601 Method-not-found` so the dispatcher is ready for them to plug in.

Closes #614
Tracking: #465 (sub-issue #2 of 11)

## What this PR lands

- **`components/mcp/transport.ts`** — framework-agnostic core. Single entry `handleMcpRequest(normRequest)`. Implements: Origin validation, Mcp-Session-Id lifecycle, MCP-Protocol-Version enforcement, status code mapping (200/202/400/403/404/405), JSON-RPC envelope parse + standard error codes.
- **`components/mcp/adapters/fastify.ts`** — operations-port adapter. Maps Fastify `(request, reply)` to the normalized shape. Now passes the already-parsed JSON body straight through (no stringify/re-parse).
- **`components/mcp/adapters/harperHttp.ts`** — application-port adapter. Maps Harper's HTTP handler `(request, next)` to the normalized shape. Returns `{status, headers, body}` for serialization through `server/serverHelpers/contentTypes.ts`.
- **`components/mcp/session.ts`** — wraps a new `system.mcp_session` system table. **No per-thread Map, no custom timer or sweep loop** — Harper's native `Table.setTTLExpiration({expiration, eviction})` handles idle eviction. Every `table.put` updates record `version`, which gives sliding-window idle semantics for free. Default `idleTimeoutSeconds = 1800` (30 min) when omitted.
- **`components/mcp/lifecycle.ts`** — `handleInitialize` (protocol negotiation + session create) and `handleInitialized` (sets `initialized: true`). Supports `2025-06-18` (preferred) and `2025-03-26` (backcompat).
- **`components/mcp/jsonrpc.ts`** — JSON-RPC 2.0 envelope parse + builders + error codes.
- **`components/mcp/index.ts`** — rewritten. Exports `registerMcpProfile(...)` (Fastify entry for ops) and `handleApplication(scope)` (Trusted Resource Plugin entry for app). Single module-level guard prevents the application registration from firing twice.
- **`components/componentLoader.ts`** — adds `mcp: mcpComponent` to `TRUSTED_RESOURCE_PLUGINS`.
- **`resources/databases.ts`** — adds `mcp_session` to `NON_REPLICATING_SYSTEM_TABLES`.
- **`server/operationsServer.ts`** — switches MCP route's preValidation from `authHandler` to `authAndEnsureUserOnRequest` (puts user on `req.hdb_user` instead of `req.body.hdb_user`, which would otherwise contaminate the JSON-RPC envelope).
- **85 unit tests + 8 integration tests.**

## Architecture: shared core + native streaming per adapter

The operations port runs Fastify; the application port runs Harper's own HTTP handler abstraction. Spec-critical logic lives in `transport.ts`; each adapter passes streaming through its native idiom. Both adapters route SSE through the existing serializer at `server/serverHelpers/contentTypes.ts:128-162`. No new Fastify dependency in the component (the adapter is what touches Fastify).

## Where to put attention

- **`components/mcp/transport.ts`** — every spec MUST lives here. The handler-side narrowing pattern (`if (parsed.ok !== true) { const fail = parsed as {...} }`) is intentional and commented: tsconfig is non-strict so TS can't narrow discriminated unions from `!== true` without help.
- **`components/mcp/session.ts`** — TTL is configured once at component init from `mcp.session.idleTimeoutSeconds` (default 1800s) + 60s eviction window. Live reconfig isn't supported in v1 (operator restart picks up the new value).
- **`server/operationsServer.ts` preValidation swap** — `authHandler` mutates `req.body`; for the JSON-RPC envelope we need it untouched, so we use `authAndEnsureUserOnRequest` which puts the authenticated user on `req.hdb_user` instead.

## Cross-model review (Gemini 0.42.0)

Completed. Findings applied:

| # | Finding | Disposition |
|---|---|---|
| 1 | Origin validation missed the CORS enablement flag and `'*'` wildcard | **Fixed** — now mirrors `security/auth.ts:65` pattern (check `HTTP_CORS`/`OPERATIONSAPI_NETWORK_CORS`, honor `'*'`). |
| 2 | `void touchSession()` could resurrect a concurrently-deleted session | **Fixed** — awaited. ~1ms hot-path cost vs the race risk; clear win. |
| 3 | Fastify adapter re-stringified an already-parsed JSON body | **Fixed** — `parseMessage` now accepts string or object. |
| 4 | Session id logged at full length on user-mismatch warn | **Fixed** — masked to first 8 chars. |
| 5 | No try/catch around `handleMcpRequest` (uncaught Harper-table errors → opaque 500) | **Fixed** — wrapped, returns `-32603 Internal Error` on failure. |
| 6 | Type casts on discriminated unions are noisy | **Held** — non-strict tsconfig prevents narrowing; cast is required. Commented in code. |
| 7 | Header-normalization DRY between adapters | **Held** — ~6 LOC each, different input types; extraction's overhead beats the dedup gain. |

## Verification

```bash
npm run build       # clean
npm run lint:required # clean
npx mocha unitTests/components/mcp/**/*.test.js
# Expected: 85 passing

# Manual smoke (with mcp.operations: {} in config):
curl -sS -X POST http://localhost:9925/mcp \
  -H 'Origin: http://localhost' \
  -H 'Accept: application/json, text/event-stream' \
  -H 'Authorization: Bearer ...' \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"smoke","version":"0"}}}' -i
# Expected: 200, Mcp-Session-Id response header, body with serverInfo + capabilities.
```

## Test plan

- [x] 85 targeted unit tests pass locally
- [x] Build clean for files in this PR
- [x] Lint clean
- [x] Cross-model review (Gemini) findings applied
- [ ] CI green across the full matrix (Node 20/22/24, Bun, Windows shards)
- [ ] Manual smoke against a local Harper
- [ ] Regression: operations API endpoints unchanged when `mcp:` absent

## Out of scope (deferred to other sub-issues)

- Tools (`tools/list`, `tools/call`) — #615/#617/#618
- Resources (`resources/list`, `read`) — #616
- `listChanged` notifications via GET /mcp SSE channel — #619
- Audit logging + rate limiting — #620
- Stdio CLI (`harper mcp`) — #621
- `Last-Event-ID` resumability — deferred to MCP v2

## Notes

- **Sessions are intentionally not browser-bound.** The MCP spec's 404-on-unknown-session-id mechanism is the universal recovery path: restart, idle eviction, manual termination, cross-thread miss — all spec-equivalent to the client.
- **Multi-instance deployments:** because sessions are LMDB-backed (via the Harper table), any thread can find any session — round-robin LBs without sticky affinity are fine. (This was the chief operational concern weighed against pure in-memory storage.)
- Generated by an AI agent (Claude Opus 4.7) following the Harper Engineering Process & Guidelines lifecycle.

### DIFF
--- a/components/mcp/index.ts
+++ b/components/mcp/index.ts
@@ -45,6 +45,8 @@ interface FullConfig {
 
 interface FastifyLikeHost {
 	post: (path: string, ...rest: unknown[]) => unknown;
+	get: (path: string, ...rest: unknown[]) => unknown;
+	delete: (path: string, ...rest: unknown[]) => unknown;
 }
 
 export interface RegisterMcpProfileArgs {
@@ -69,10 +71,18 @@ export function registerMcpProfile({ profile, host, config, routeOptions }: Regi
 	ensureSessionTable();
 	const mountPath = profileConfig.mountPath ?? DEFAULT_MOUNT_PATH;
 	const handler = createFastifyHandler(profile);
-	if (routeOptions) {
-		host.post(mountPath, routeOptions, handler);
-	} else {
-		host.post(mountPath, handler);
+	// Register POST, GET, and DELETE on the same mount path. The transport
+	// core decides whether to handle (POST initialize / JSON-RPC dispatch),
+	// return 405 with an accurate `Allow` header (GET always in v1, DELETE
+	// when `mcp.session.allowClientDelete` is false), or process (DELETE
+	// when enabled). Without explicit GET/DELETE routes Fastify would
+	// short-circuit to its built-in 404 before the transport runs.
+	for (const method of ['post', 'get', 'delete'] as const) {
+		if (routeOptions) {
+			host[method](mountPath, routeOptions, handler);
+		} else {
+			host[method](mountPath, handler);
+		}
 	}
 	harperLogger.info(`MCP ${profile} profile registered at ${mountPath}`);
 }
--- a/components/mcp/index.ts
+++ b/components/mcp/index.ts
@@ -1,16 +1,35 @@
 /**
  * Native MCP (Model Context Protocol) server component for Harper.
  *
- * Foundation PR (#613): exports a presence-gated registration hook used by
- * the operations and HTTP host servers. The hook installs a placeholder
- * route that returns HTTP 503 with body `{ error: 'mcp_not_implemented',
- * profile }` until the Streamable HTTP transport lands in #614. A profile
- * is enabled when its sub-block exists in config (matches Harper's
- * `replication` convention — no explicit `enabled` flag). Tracking: #465.
+ * #614 replaces the foundation stub from #613 with a real Streamable HTTP
+ * transport. Two entry points:
+ *
+ *   - `registerMcpProfile({profile:'operations', host, config, routeOptions})`
+ *     called from `server/operationsServer.ts` (Fastify-side gate).
+ *
+ *   - `handleApplication(scope)` invoked by the component loader when the
+ *     root config contains a top-level `mcp:` block. Registers the
+ *     application-profile handler on the HTTP port iff `mcp.application` is
+ *     present.
+ *
+ * Profile presence drives enablement, matching Harper's `replication`
+ * convention (no `enabled` flag). See #465 for the umbrella design.
  */
 import harperLogger from '../../utility/logging/harper_logger.ts';
+import { getConfigObj as realGetConfigObj } from '../../config/configUtils.js';
+import { createFastifyHandler } from './adapters/fastify.ts';
+import { createHarperHttpHandler } from './adapters/harperHttp.ts';
+import { ensureSessionTable } from './session.ts';
+import type { McpProfile } from './transport.ts';
 
-export type McpProfile = 'operations' | 'application';
+// Indirection so tests can swap the config source.
+let getConfigObj: () => unknown = realGetConfigObj as () => unknown;
+export function _setGetConfigObjForTest(fn: () => unknown): void {
+	getConfigObj = fn;
+}
+export function _restoreGetConfigObj(): void {
+	getConfigObj = realGetConfigObj as () => unknown;
+}
 
 interface McpProfileConfig {
 	mountPath?: string;
@@ -20,28 +39,26 @@ interface FullConfig {
 	mcp?: {
 		operations?: McpProfileConfig;
 		application?: McpProfileConfig;
+		session?: unknown;
 	};
 }
 
-interface FastifyLike {
+interface FastifyLikeHost {
 	post: (path: string, ...rest: unknown[]) => unknown;
 }
 
 export interface RegisterMcpProfileArgs {
 	profile: McpProfile;
-	host: FastifyLike;
+	host: FastifyLikeHost;
 	config: FullConfig;
-	/** Route-level options forwarded to the host's `post(path, options, handler)` 3-arg form (e.g., Fastify `preValidation`). */
 	routeOptions?: Record<string, unknown>;
 }
 
 const DEFAULT_MOUNT_PATH = '/mcp';
 
 /**
- * Register the MCP profile on its host server when enabled in config.
- *
- * The stub responder is intentionally minimal — sub-issue #614 replaces it
- * with the real Streamable HTTP transport without changing this gate.
+ * Fastify-side registration. Used by `server/operationsServer.ts` for the
+ * operations profile. Idempotent through the host's own route table.
  */
 export function registerMcpProfile({ profile, host, config, routeOptions }: RegisterMcpProfileArgs): void {
 	const profileConfig = config?.mcp?.[profile];
@@ -49,9 +66,9 @@ export function registerMcpProfile({ profile, host, config, routeOptions }: Regi
 		harperLogger.trace(`MCP ${profile} profile not configured, skipping registration`);
 		return;
 	}
-
+	ensureSessionTable();
 	const mountPath = profileConfig.mountPath ?? DEFAULT_MOUNT_PATH;
-	const handler = createStubHandler(profile);
+	const handler = createFastifyHandler(profile);
 	if (routeOptions) {
 		host.post(mountPath, routeOptions, handler);
 	} else {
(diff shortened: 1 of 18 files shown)


kylebernhardy:
**CI revealed a harness-only config issue** — pushed `22a876b73` with the fix.

The first CI run had 4 failing integration shards, all on the new MCP tests. Root cause: the @harperfast/integration-testing harness passes test config via the `HARPER_SET_CONFIG` env var. That env var is applied through `applyConfigLayer → flattenObject` in `config/harperConfigEnvVars.ts:176`, which **skips empty objects** because they produce no flat paths. So `{ mcp: { operations: {} } }` got dropped entirely — Harper booted with no MCP config, and the `/mcp` route was never registered (hence the 404s).

**This affects only the test harness path, not real users.** Production YAML loading preserves empty objects, so presence-based enablement still works when an operator writes `mcp:\n  operations: {}` in `harperdb-config.yaml`. Verified by reading the YAML → configDoc path.

**Fix:** the integration test now passes an explicit `mountPath` so `flattenObject` has something to walk. Comment in the test points at the harness limitation as Harper-wide cleanup worth a separate issue.

components/mcp/transport.ts
```diff
@@ -0,0 +1,272 @@
...
+function handleGet(): NormResponse {
+	// v1 does not offer a GET SSE channel. #619 will replace this with a
+	// real server-push stream for `listChanged` notifications.
+	return { status: 405, headers: { Allow: 'POST, DELETE' } };
+}
+
+async function handleDelete(request: NormRequest): Promise<NormResponse> {
+	const allow = env.get(CONFIG_PARAMS.MCP_SESSION_ALLOWCLIENTDELETE);
+	if (allow !== true) {
+		return { status: 405, headers: { Allow: 'POST, GET' } };
```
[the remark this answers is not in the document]
kylebernhardy: Good catch — applied in 1878dd08d. Centralized in `currentlyAllowedMethods()`: `POST` always; `DELETE` only when `mcp.session.allowClientDelete === true`; `GET` never (until #619 brings the server-push SSE channel). Updated all three 405 paths (`handleGet`, `handleDelete` disabled, top-level unsupported-method fallback). Added unit tests covering Allow values for both configs.

kylebernhardy:
**CI flake on `b561d0099`** — re-running the failed shard.

Job: `Integration Tests 3/4 (Bun)` failed on the same `csv_file_load` test in `integrationTests/apiTests/terminology.test.mjs:481` that flaked on PR #649. The CSV-load job stayed `IN_PROGRESS` past the 30s timeout instead of completing. Same Harper-wide flake pattern; not MCP-related (this test doesn't touch any MCP code).

Other 34 checks ✅. The fix in `b561d0099` (register POST/GET/DELETE on the Fastify route so the transport's 405 paths are reachable) resolved the remaining MCP test failures.

Rerunning just the Bun shard.

components/mcp/index.ts
```diff
@@ -20,61 +39,87 @@ interface FullConfig {
...
+	const config = getConfigObj() as FullConfig | undefined;
+	if (!config?.mcp?.application) {
+		harperLogger.trace('MCP application profile not configured, skipping registration');
+		return;
+	}
+	applicationStarted = true;
+	ensureSessionTable();
+	const mountPath = config.mcp.application.mountPath ?? DEFAULT_MOUNT_PATH;
+	const handler = createHarperHttpHandler('application');
+	scope.server.http(handler, { urlPath: mountPath, after: 'authentication' });
```
kriszyp, line 118: Wow, didn't take long for this (brand) new feature to get used.

resources/databases.ts
```diff
@@ -82,6 +82,7 @@ export const NON_REPLICATING_SYSTEM_TABLES = [
 	'hdb_session_will',
 	'hdb_job',
 	'hdb_info',
+	'mcp_session',
```
kriszyp, line 85: Wait, we don't have to prefix with `hdb_` anymore?!? This is fine, `hdb_` is certainly antiquated, no longer our name, and not clear what prefixing tables with hdb_ within a database that is already separate from users, really means.

kriszyp (approved):
This looks good. A couple of things to acknowledge:
* I believe this is putting both application agent sessions and system agent sessions in the same table. These are very different agents for different users. I think that is ok, I can't think of any problems with that, but just noting that as a potential concern.
* Acknowledging that sessions are in a non-replicating table: intentionally not in memory to support requests across any threads, but not replicating, so does not support requests that might be routed to another node in the same session. I think that's fine, it would be unusual for a session to switch nodes (DNS expiration, GTM rebalancing), but does seem possible. Again, noting as a potential concern.
Good job, I approve merging.

kylebernhardy:
> * Acknowledging that sessions are in a non-replicating table: intentionally not in memory to support requests across any threads, but not replicating,

Yes The MCP spec regarding session management is no where near as stringent as MQTT and dropping sessions is expected and handled on the client side. From the spec: 
```
The server MAY terminate the session at any time, after which it MUST respond to requests containing that session ID with HTTP 404 Not Found.
```

Initial development there was thinking around in memory data maps with TTL but like you said that is not manageable across threads, adds overhead and could cause memory pressure to any thread. Instead we get all of this for free with a Harper table.

resources/databases.ts
kylebernhardy: i made the decision with the same realization you made :)

PR #780, opened 2026-05-25
Make RocksDB memory config configurable (block cache + WriteBufferManager)

kriszyp:
Exposes RocksDB memory tuning knobs as Harper config params so deployments can override the hardcoded defaults. Adds four new params, all under `storage.rocks.*`:

| Param | What it controls |
|---|---|
| `storage.rocks.blockCacheSize` | Bytes for the shared block cache. Defaults to 25% of constrained memory. |
| `storage.rocks.writeBufferManagerSize` | Bytes the process-wide WriteBufferManager caps total memtable memory at (0 = disabled, the default). Structural cap on memtables + OptimisticTransactionDB maintain-window history. |
| `storage.rocks.writeBufferManagerCostToCache` | Charge memtable bytes against the block cache so both subsystems share accounting. Visible via a single `rocksdb.block-cache-usage` metric. |
| `storage.rocks.writeBufferManagerAllowStall` | Stall writes once the WBM cap is hit (hard cap) vs. allow brief overshoot with aggressive flushing (soft cap). |

**Why**: On shared Fabric hosts, each container's RocksDB memory footprint accumulates. For a canopy-style customer with 14 tables, the per-CF maintain window alone holds ~1.8 GB of anonymous memory by default. Today there's no way to bound this without rebuilding harper-pro. These params let host-manager (and other operators) configure both subsystems without code changes.

**Implementation**: Both reads happen inside `openRocksDatabase()` so env/CLI overrides applied after module load are respected. Values are passed to `RocksDatabase.config()` via conditional spread — unset env vars don't surface unknown options to older bindings of `@harperfast/rocksdb-js` (forward-compatibility).

**Dependency**: The WBM options are no-ops until [HarperFast/rocksdb-js#584](https://github.com/HarperFast/rocksdb-js/pull/584) lands and harper-pro updates its dependency. The `blockCacheSize` param works today.

**Cross-model review findings addressed**:
- Gemini (first round): module-load timing — moved reads from module-level constants into `openRocksDatabase()` so config changes after module load are picked up.
- Codex: same timing concern — same fix.
- Input validation: `> 0` guard on `blockCacheSize` handles `0`/negative/`NaN`.
- Gemini (WBM additions): string coercion handled by Harper's existing envGet (parses `'true'`/`'false'` → boolean, `'123'` → number).

Generated by Claude. — *Claude Sonnet 4.7*


resources/databases.ts
```diff
@@ -149,15 +149,20 @@ export const databaseEventsEmitter = new EventEmitter<DatabaseWatcherEventMap>()
...
 	if (isReadOnlyMode()) {
 		options.readOnly = true;
 	}
-	RocksDatabase.config({ blockCacheSize: MEMORY_FOR_ROCKS_DB });
+	// Read block cache size lazily so env/CLI overrides applied after module load are respected.
+	// Falls back to 25% of constrained (cgroup) memory when not configured.
+	const configuredBlockCacheSize = envGet(CONFIG_PARAMS.STORAGE_ROCKS_BLOCKCACHESIZE);
+	const blockCacheSize =
+		configuredBlockCacheSize > 0
+			? configuredBlockCacheSize
```
cb1kenobi, line 173: Is `configuredBlockCacheSize` still a string here or does `envGet()` coerce it to a number?

kriszyp: Sorry, deleting those previous comments induced by a combination of insufficient coffee and excessive AI. tl;dr: yes, `configuredBlockCacheSize` should be a number (that's how it should from YAML). And we added type checking, so if it isn't, we don't use it.

cb1kenobi (approved):

Issue #618, opened 2026-05-19
[MCP] Application profile: tool generation over Resources registry


kylebernhardy:
**Scope.** Application MCP profile. Walk the `Resources` registry and generate per-resource MCP tools (`get_*`, `search_*`, `create_*`, `update_*`, `delete_*`). Derive input JSON Schema from `Table.attributes`. Implement tool-name sanitization and collision rules.

**Design reference.** Sections "Application MCP" and "Tool-list filtering" in #465.

**Acceptance criteria**
- [ ] Walk `resources/Resources.ts` registry; for each exported resource (table or custom Resource), publish only the verbs the class implements (class-level introspection from #615 — `prototype.method !== Resource.prototype.method`).
- [ ] Generated tool set per resource: `get_<r>`, `search_<r>`, `create_<r>`, `update_<r>`, `delete_<r>` (each only when the verb is implemented and the user's role grants the corresponding `read/insert/update/delete` flag).
- [ ] Input schemas for table-backed tools derived from `Table.attributes` (types, nullable, primary key) at registration time.
- [ ] `search_*` input schema includes optional `limit` (capped at `searchMaxResults`) and opaque `cursor`; responses include `nextCursor` when more rows exist.
- [ ] Tool-name sanitization: replace `/` and `.` with `_`, drop other invalid chars; on collision, prefix with database segment, then hash if still colliding. The canonical resource path appears in the tool description.
- [ ] Tools delegate to the existing `Resource.get/post/...` invocation path so `transactional()` and instance-level `allowRead/Create/Update/Delete` enforce RBAC at call time. **Restricted-attribute writes must be rejected at runtime**, not just absent from the input schema — include a test covering this.
- [ ] `attribute_permissions` narrow the JSON Schema's allowed properties.
- [ ] Tools dispatch returns `structuredContent` plus a serialized text content block for backwards compat (per MCP §server/tools).
- [ ] Unit + integration tests for: per-resource tool generation, RBAC filtering of the tool list, schema correctness, isError on restricted-write rejection, search pagination round-trip.

**Out of scope.** SQL (explicit non-goal). Custom Resource opt-in for non-verb methods (#622). listChanged (#619). Audit + rate limiting (#620).

**Stacks on.** #615 (tool registry).

**Branch & PR conventions**
- Branch: `feat/mcp-app-profile`
- PR base: `main` (after #615 merges; can be parallel with #617).
- Closes this issue; references #465.

**Smoke test**
```bash
[...]

Issue #617, opened 2026-05-19
[MCP] Operations profile: tool generation over OPERATION_FUNCTION_MAP


kylebernhardy:
**Scope.** Operations MCP profile. Walk `OPERATION_FUNCTION_MAP` and generate one MCP tool per operation in the v1 conservative `allow` list. Each tool delegates to the existing operation handler so RBAC enforcement is unchanged.

**Design reference.** Sections "Operations MCP (admin)", "Config (`harperdb-config.yaml`) → operations", and "Compliance Audit Summary" in #465.

**Acceptance criteria**
- [ ] Walk `OPERATION_FUNCTION_MAP` (`server/serverHelpers/serverUtilities.ts:330-450`); for each operation name matching `mcp.operations.allow` and not matching `mcp.operations.deny`, register an MCP tool.
- [ ] Default `allow` list: `describe_*`, `list_*`, `search_*`, `get_*`, `system_information`, `read_log`, `read_audit_log`. Destructive operations require explicit opt-in even for super_user.
- [ ] Hand-curated JSON Schemas for each allowed operation's params (committed alongside the operation registration).
- [ ] Per-operation MCP annotations: `readOnlyHint: true` for describe/list/get/read; the few destructive ops (`drop_*`, `delete_*`, `restart`, `set_configuration`, etc.) — when opted in via `allow` — carry `destructiveHint: true`.
- [ ] Tool dispatch delegates to the existing operation handler in `OPERATION_FUNCTION_MAP`. The MCP layer does **not** re-implement `verifyPerms`; existing checks run as today.
- [ ] Class-level permission filter (in `tools/list`) uses `verifyPerms`-equivalent class-level predicates (`super_user`, `structure_user`, `cluster_user`, `operations` allowlist on the role).
- [ ] Unit tests for: tool list filtered per role; tool call delegating correctly; allow/deny glob honored; isError on permission denial at call time.
- [ ] Integration test runs `describe_all` end-to-end over MCP.

**Out of scope.** Application-side tools (#618), listChanged (#619), audit + rate limiting (#620). No SQL exposure on this profile (explicit non-goal in #465).

**Stacks on.** #615 (tool registry).

**Branch & PR conventions**
- Branch: `feat/mcp-ops-profile`
- PR base: `main` (after #615 merges; can be parallel with #618).
- Closes this issue; references #465.

**Smoke test**
```bash
curl -sS -X POST http://localhost:9925/mcp ... \
[...]

PR #781, opened 2026-05-25
[MCP] Tool registry + RBAC-aware tools/list

### LINKED ISSUE #615: [MCP] Tool registry + class-level introspection + RBAC-aware tools/list filtering
**Scope.** Build the tool registry, JSON Schema input plumbing, and the RBAC-aware `tools/list` filter. No real tools yet — this PR ships the framework; the operations and application profiles plug into it in #617 and #618.

**Design reference.** Sections "Tool-list filtering" and "Safety & Observability → Error mapping" in #465.

**Acceptance criteria**
- [ ] Tool registry abstraction (`addTool`, `removeTool`, `listTools(user)`) with per-session caching of the filtered list.
- [ ] Class-level verb introspection reuses the pattern at `resources/openApi.ts:149-153` (`prototype.method !== Resource.prototype.method`).
- [ ] User-level RBAC walk reuses the pattern at `dataLayer/schemaDescribe.ts:29-49` (direct walk of `user.role.permission[db].tables[table]`). **Does not use `Resource.allowRead/Create/Update/Delete` — those are instance methods bound to a record id (`resources/Resource.ts:413-427`).**
- [ ] `tools/list` paginates via opaque `cursor` / `nextCursor`; page size capped by `mcp.<profile>.maxTools`.
- [ ] `tools/call` dispatches to a registered tool's handler; unknown tool → JSON-RPC `-32601`; invalid args → `-32602`; tool-execution error returns `result.isError = true` (NOT a JSON-RPC error).
- [ ] Tool annotations (`readOnlyHint`, `destructiveHint`, `idempotentHint`, `openWorldHint`) flow through.
- [ ] Unit tests cover: filter for super_user, structure_user, read-only role, anonymous; pagination; isError mapping; unknown-tool error.

**Out of scope.** No operations API integration (#617), no Resources registry integration (#618), no listChanged (#619), no rate limiting (#620).

**Stacks on.** #614 (transport + lifecycle).

**Branch & PR conventions**
- Branch: `feat/mcp-tool-registry`
- PR base: `main` (after #614 merges).
- Closes this issue; references #465.

**Smoke test**
```bash
# Register a stub tool in test, then:
curl -sS -X POST http://localhost:9925/mcp ... \
  -d '{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}'
# Expected: { result: { tools: [<stub>], nextCursor?: ... } }, RBAC-filtered to caller's role.
```

**Tracking.** Part of HarperFast/harper#465. Sub-issue #3 of 11.

kylebernhardy:
## Summary

Stacks on #763 (transport). Adds the tool registry framework + RBAC-aware `tools/list` / `tools/call` dispatchers per MCP §server/tools (rev 2025-06-18). **No real tools register here yet** — the operations profile (#617) walks `OPERATION_FUNCTION_MAP`, the application profile (#618) walks the `Resources` registry. Both will call `addTool(def)` to publish.

Closes #615
Tracking: #465 (sub-issue #3 of 11)
Base: `feat/mcp-transport` (rebase to `main` after #763 merges)

## What this PR lands

- **`components/mcp/toolRegistry.ts`** — new module with:
  - Registry CRUD (`addTool`, `removeTool`, `getTool`).
  - `listTools(args)` — profile + `visibleTo(user)` filter, opaque cursor pagination, per-profile `maxTools` cap (default 200), per-session pagination cache for stable paging.
  - `clearSessionCache(sessionId)` — wired into `session.deleteSession` so the cache doesn't outlive the session.
  - RBAC helpers ready for #617/#618:
    - `isSuperUser(user)`
    - `hasClassLevelVerbs(prototype, resourcePrototype)` — mirrors `resources/openApi.ts:149-153`.
    - `userTablePermissions(user, db, table)` — mirrors `dataLayer/schemaDescribe.ts:29-49`. Does **not** use `Resource.allowRead/Create/Update/Delete` (those are per-record, not class-level).
    - `canRoleInvokeOperation(user, op)` — role-level (`super_user` / `structure_user` / `cluster_user` / `role.operations` allowlist).
- **`components/mcp/transport.ts`** — adds `dispatchToolsList` and `dispatchToolsCall`. `NormRequest` now carries `userObject?: AuthedUser` so tool RBAC sees the full role tree.
- **Adapters** — thread the full user object through (`request.hdb_user` on Fastify, `request.user` on Harper-HTTP).
- **42 new tests** (130 passing total: 88 from #614 baseline + 42 here).

## Spec-conformance notes

- **Pagination** uses an opaque base64url-encoded cursor (`{offset:N}`). Bad cursors decode to offset 0. Per MCP §server/utilities/pagination.
- **Error mapping** distinguishes protocol vs tool-execution errors per the spec's emphatic note:
  - `tools/call` with unknown name → JSON-RPC `-32601`.
  - `tools/call` with missing/non-string `params.name` → JSON-RPC `-32602`.
  - Handler throws → JSON-RPC **success** with `result.isError: true` and `{kind: "harper_error", message}` content. Stack stays in the server log; only the message goes to the wire.
- **Annotations** (`readOnlyHint`, `destructiveHint`, `idempotentHint`, `openWorldHint`) flow through to the public descriptor.

## Where to put attention

- **`toolRegistry.ts:sessionListCache`** — process-memory `Map<sessionId, {tools}>`. Cleared on `addTool` / `removeTool` (any registry mutation drops all caches) and on `deleteSession` (per-session). Cache eviction at session expiry is **not** instrumented yet because Harper's TTL eviction on `system.mcp_session` happens in the data layer, not as a per-key callback. For v1 this is acceptable — sessions expire at `idleTimeoutSeconds` (default 30 min) and the cache leak is bounded by per-thread session count × ~5KB. If this becomes a real issue, the fix is a TTL-coupled cache or a periodic sweep.

- **`canRoleInvokeOperation`'s `SCHEMA_STRUCTURE_OPERATIONS` / `CLUSTER_OPERATIONS`** — seeded with canonical operations. #617 will expand the structure set against `OPERATION_FUNCTION_MAP`. Drift between this seed and the real ops map could silently hide tools from users that have the role-level privilege. Worth a note in #617 to keep in sync.

- **Input-schema validation** — `tools/call` does not validate `arguments` against the tool's `inputSchema` in this PR. Individual handlers can validate (or `#617`/`#618` can add a centralized JSON Schema validator). Deferred intentionally to keep this PR focused on the framework.

## Cross-model review (Gemini 0.42.0)

Completed. Findings applied:

| # | Finding | Disposition |
|---|---|---|
| 1 | `sessionListCache` leaked past session deletion | **Fixed** — `clearSessionCache` exported + invoked from `session.deleteSession`. |
| 2 | Centralized inputSchema validation in `tools/call` | **Held for #617/#618** — real tools will validate; framework stays minimal. |
| 3 | Add mid-flow pagination invalidation test | **Added** — registry mutation between pages; cursor still works via recompute path. |
| 4 | `kind: "harper_error"` is the only error variant | **Held for #617/#618** — real tools will refine into `validation_error`, `permission_denied`, etc. |
| 5 | TS discriminated-union casts are noisy | **Held** — non-strict tsconfig; commented in code (same as #614). |
| 6 | Hardcoded operations sets risk drift with `OPERATION_FUNCTION_MAP` | **Noted in code** — #617 will sync against the real map. |

## Verification

```bash
npm run build       # clean
npm run lint:required # clean
npx mocha unitTests/components/mcp/**/*.test.js
# Expected: 130 passing
```

## Out of scope (deferred to other sub-issues)

- Real tool registrations — operations API (#617), Resources registry (#618).
- `notifications/tools/list_changed` over a server-push SSE channel (#619).
- Audit logging + per-session rate limiting (#620).
- Stdio CLI (#621).
- `Last-Event-ID` resumability — MCP v2.

## Stacking

This PR is based on `feat/mcp-transport` (PR #763). Once #763 merges, I'll rebase this to `main`. Reviewing the diff against `feat/mcp-transport` shows only the tools work — ~1000 LOC, 1 commit.

Generated by an AI agent (Claude Opus 4.7) following the Harper Engineering Process & Guidelines lifecycle.

### DIFF
--- a/components/mcp/toolRegistry.ts
+++ b/components/mcp/toolRegistry.ts
@@ -1,0 +1,324 @@
+/**
+ * MCP tool registry (#615).
+ *
+ * Holds the framework for `tools/list` and `tools/call` per MCP §server/tools.
+ * No real tools register here yet — the operations profile (#617) walks
+ * `OPERATION_FUNCTION_MAP`, and the application profile (#618) walks the
+ * `Resources` registry. Both call `addTool(def)` to publish.
+ *
+ * Filtering follows the two-step pattern documented in #465:
+ *   1. Class-level verb introspection (which tools are *publishable*)
+ *      reuses the prototype-comparison helper from `resources/openApi.ts:149-153`.
+ *   2. User-level RBAC walk (which publishable tools are *visible*) reuses
+ *      the direct `user.role.permission[db].tables[table]` walk from
+ *      `dataLayer/schemaDescribe.ts:29-49`. Resource.allow* methods are
+ *      *not* used — they're per-record instance methods.
+ *
+ * The schema-level filter here is a UX optimization, not a security boundary.
+ * Runtime enforcement still runs at tool-call time.
+ */
+import type { McpProfile } from './transport.ts';
+
+export interface ToolAnnotations {
+	title?: string;
+	readOnlyHint?: boolean;
+	destructiveHint?: boolean;
+	idempotentHint?: boolean;
+	openWorldHint?: boolean;
+}
+
+export interface ToolContent {
+	type: 'text' | 'image' | 'audio' | 'resource';
+	text?: string;
+	data?: string;
+	mimeType?: string;
+}
+
+export interface ToolResult {
+	content: ToolContent[];
+	isError?: boolean;
+	structuredContent?: unknown;
+}
+
+/** Public shape sent to clients on `tools/list`. */
+export interface ToolDescriptor {
+	name: string;
+	description: string;
+	inputSchema: object;
+	annotations?: ToolAnnotations;
+}
+
+/** Authenticated user object as Harper builds it (subset we touch). */
+export interface AuthedUser {
+	username?: string;
+	role?: {
+		role?: string;
+		permission?: {
+			super_user?: boolean;
+			structure_user?: boolean;
+			cluster_user?: boolean;
+			operations?: string[];
+			[database: string]:
+				| boolean
+				| string[]
+				| undefined
+				| {
+						describe?: boolean;
+						tables?: {
+							[table: string]:
+								| undefined
+								| {
+										read?: boolean;
+										insert?: boolean;
+										update?: boolean;
+										delete?: boolean;
+										describe?: boolean;
+										attribute_permissions?: unknown;
+								  };
+						};
+				  };
+		};
+	};
+}
+
+export interface ToolCallContext {
+	user: AuthedUser;
+	profile: McpProfile;
+	sessionId: string;
+}
+
+/**
+ * Internal tool definition. `handler` and `visibleTo` are not sent to the
+ * client; only the public `ToolDescriptor` fields are.
+ */
+export interface ToolDef extends ToolDescriptor {
+	profile: McpProfile;
+	visibleTo: (user: AuthedUser) => boolean;
+	handler: (args: unknown, context: ToolCallContext) => Promise<ToolResult> | ToolResult;
+}
+
+const registry = new Map<string, ToolDef>();
+
+/** Per-session pagination cache. Invalidated when a fresh `tools/list` (no cursor) call recomputes. */
+const sessionListCache = new Map<string, { tools: ToolDescriptor[] }>();
+
+export function addTool(def: ToolDef): void {
+	if (!def?.name) throw new Error('addTool: name is required');
+	registry.set(def.name, def);
+	// New tool → drop pagination caches; next list call rebuilds.
+	sessionListCache.clear();
+}
+
+export function removeTool(name: string): boolean {
+	const existed = registry.delete(name);
+	if (existed) sessionListCache.clear();
+	return existed;
+}
+
+export function getTool(name: string): ToolDef | undefined {
+	return registry.get(name);
+}
+
+/**
+ * Drop the pagination cache entry for a session. Called by
+ * `session.deleteSession` so the cache doesn't outlive the session that
+ * created it. Without this, every session that ever paged tools/list
+ * leaves an orphan entry until the process restarts.
+ */
+export function clearSessionCache(sessionId: string): void {
+	sessionListCache.delete(sessionId);
+}
+
+/** Test seam: drop all registrations. */
+export function _resetRegistryForTest(): void {
+	registry.clear();
+	sessionListCache.clear();
+}
+
+export interface ListToolsArgs {
+	user: AuthedUser;
+	profile: McpProfile;
+	sessionId: string;
+	cursor?: string;
+	limit: number;
+}
+
+export interface ListToolsResult {
+	tools: ToolDescriptor[];
+	nextCursor?: string;
+}
+
+/**
+ * Filter the registry to tools the user can see on this profile, then page.
+ *
+ * Cursor encoding: base64 of `{offset:N}`. Opaque to clients per MCP
+ * §server/utilities/pagination. A fresh call (no cursor) recomputes and
+ * caches the filtered list; subsequent paged calls reuse the cache so
+ * pagination is stable even if the registry mutates between pages.
+ */
+export function listTools(args: ListToolsArgs): ListToolsResult {
+	const { user, profile, sessionId, cursor, limit } = args;
+	if (limit < 1) throw new Error('listTools: limit must be >= 1');
+
+	let cached = sessionListCache.get(sessionId);
+	let offset = 0;
+	if (cursor) {
+		offset = decodeCursor(cursor);
+		if (!cached) {
+			// Cache evicted (likely registry change). Treat as a fresh call from
+			// the cursor's offset — best effort. The client may see the next
+			// page from a slightly different list, which is acceptable per
+			// MCP's eventual-consistency stance on listChanged.
+			cached = { tools: computeFilteredList(user, profile) };
+			sessionListCache.set(sessionId, cached);
+		}
+	} else {
+		cached = { tools: computeFilteredList(user, profile) };
+		sessionListCache.set(sessionId, cached);
+	}
+
+	const slice = cached.tools.slice(offset, offset + limit);
+	const next = offset + slice.length;
+	return {
+		tools: slice,
+		nextCursor: next < cached.tools.length ? encodeCursor(next) : undefined,
+	};
+}
+
+function computeFilteredList(user: AuthedUser, profile: McpProfile): ToolDescriptor[] {
+	const out: ToolDescriptor[] = [];
+	for (const def of registry.values()) {
+		if (def.profile !== profile) continue;
+		if (!def.visibleTo(user)) continue;
+		out.push({
+			name: def.name,
+			description: def.description,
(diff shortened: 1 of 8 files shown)


components/mcp/transport.ts
```diff
@@ -201,6 +213,79 @@ function handleGet(): NormResponse {
...
+	const params =
+		'params' in message ? (message.params as { name?: unknown; arguments?: unknown } | undefined) : undefined;
+	const name = typeof params?.name === 'string' ? params.name : undefined;
+	if (!name) {
+		return jsonResponse(200, buildError(messageId, ERROR_CODES.INVALID_PARAMS, 'tools/call requires params.name'));
+	}
+	const tool = getTool(name);
+	if (!tool || tool.profile !== request.profile) {
+		return jsonResponse(200, buildError(messageId, ERROR_CODES.METHOD_NOT_FOUND, `Unknown tool: ${name}`));
+	}
```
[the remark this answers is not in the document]
kylebernhardy: Good catch on the misleading wording — applied **Option B** in 557c274c4.

The pass-through is intentional per #465 "Tool-list filtering":

> The schema-level filtering in Steps 1–3 is a UX optimization (so the LLM only sees what's relevant) — it is **not** a security boundary. v1 includes a verification test where the role's tool list omits `update_product`, the LLM nonetheless sends an `update_product` call, and the runtime predicate rejects it with an `isError` tool result.

Runtime enforcement is `transactional()` + per-record `allow{Read,Create,Update,Delete}` in the handler — inherited from `OPERATION_FUNCTION_MAP` by #617 and from `Resources` by #618. The framework deliberately passes through so an LLM that hallucinates a name gets a clean `isError` from the runtime predicate (the verification-test shape from the design doc).

Disposition:
- Rewrote the security-model paragraph in `toolRegistry.ts` to make the layering explicit — the prior wording "Runtime enforcement still runs at tool-call time" didn't say *where* and read like a framework-layer guarantee.
- Added a `tools/call` test with `visibleTo: () => false` that asserts the handler is still reached. Documents the intentional pass-through so a future defense-in-depth change to add visibleTo to the call path is caught and discussed.

components/mcp/toolRegistry.ts
```diff
@@ -0,0 +1,335 @@
+/**
+ * MCP tool registry (#615).
```
kriszyp: I don't think we want PRs referenced in permanent code. That's what blame is for. I have scolded Claude for this noise before. Will update skills to try to avoid this in the future.

components/mcp/toolRegistry.ts
```diff
@@ -0,0 +1,335 @@
...
+
+/** Authenticated user object as Harper builds it (subset we touch). */
+export interface AuthedUser {
+	username?: string;
+	role?: {
+		role?: string;
+		permission?: {
+			super_user?: boolean;
+			structure_user?: boolean;
+			cluster_user?: boolean;
```
kriszyp: Why are you still here :/

kylebernhardy:
**CI flake on `5ad8154ac`** — re-running the failed shard.

Job: `Unit Test (Node.js v24)` failed on a single test in `unitTests/resources/subscriptionReplay.test.js` — *Subscription replay > startTime branch > delivers writes that land during replay without duplicates*: `AssertionError: missing concurrent id 2029`. Race-condition assertion ("writes that **land during** replay"), single-platform, other Node versions and Bun all green. Same flake shape as the `query.test.js` / `terminology.test.mjs` patterns we've seen before — not MCP-related (this file doesn't touch any MCP code).

Re-running just the Node 24 unit shard.

components/mcp/toolRegistry.ts
```diff
@@ -0,0 +1,335 @@
...
+ * The implementation here is intentionally conservative — only flags that
+ * grant operations globally short-circuit. Per-operation per-target checks
+ * are evaluated at call time by Harper's existing `verifyPerms`.
+ */
+export function canRoleInvokeOperation(user: AuthedUser, operation: string): boolean {
+	if (isSuperUser(user)) return true;
+	const perm = user?.role?.permission;
+	if (!perm) return false;
+	if (perm.structure_user && SCHEMA_STRUCTURE_OPERATIONS.has(operation)) return true;
+	if (perm.cluster_user && CLUSTER_OPERATIONS.has(operation)) return true;
```
kriszyp: Is that really permission we have?

components/mcp/toolRegistry.ts
```diff
@@ -0,0 +1,335 @@
...
+	attribute_permissions?: unknown;
+}
+
+/**
+ * Walk `user.role.permission[database].tables[table]`. Mirrors the pattern
+ * at `dataLayer/schemaDescribe.ts:29-49`. Returns `null` when the user has
+ * no entry at this scope (treat all verbs as denied). Super-user short-
+ * circuits to all-true.
+ */
+export function userTablePermissions(user: AuthedUser, db: string, table: string): TablePermissions | null {
```
kriszyp: This isn't used anywhere!?
(except in the tests 😆 )

kriszyp (approved):
A few things that should probably be cleaned up, but I approve. Good work.

components/mcp/toolRegistry.ts
kylebernhardy: Applied in `e494921a3`. Stripped `#615`/`#617`/`#618`/`#465` from the header docblock and the three helper docblocks below it. Kept the path-anchored "Mirrors `resources/openApi.ts:149-153`" / "`dataLayer/schemaDescribe.ts:29-49`" references — those describe behavior and stay accurate as PR numbers age out.

Saved the preference to memory so future MCP work avoids the noise.

components/mcp/toolRegistry.ts
kylebernhardy: Good catch — removed in `e494921a3`. Verified `ROLE_TYPES_ENUM` (`utility/hdbTerms.ts:87-89`) only declares `super_user`, and `security/impersonation.ts:159` forcibly sets `cluster_user: false` on impersonated requests, so it's never a role-grantable permission. Dropped it from the `AuthedUser` interface here and from the `canRoleInvokeOperation` branch / `CLUSTER_OPERATIONS` set below.

components/mcp/toolRegistry.ts
kylebernhardy: You're right — deleted in `e494921a3` (along with its `TablePermissions` companion interface and the matching `describe('userTablePermissions', …)` test block). It was a forward-looking helper for #617/#618, and #616 ended up inlining its own walk in `resources.ts` rather than depending on this export, so the symbol was genuinely orphaned. When #617 actually needs it I'll factor it back out of the inlined #616 version.

components/mcp/toolRegistry.ts
kylebernhardy: Removed in `e494921a3`. You're right — `cluster_user` is internal: `security/impersonation.ts:159` forces it to `false`, and `ROLE_TYPES_ENUM` (`utility/hdbTerms.ts:87-89`) only declares `super_user`. Dropped both the branch and the `CLUSTER_OPERATIONS` set, plus the corresponding test case.

PR #837, opened 2026-05-27
Fix jsLoader to support ESM-only packages

### LINKED ISSUE #826: jsLoader: pure-ESM packages fail to resolve (ERR_PACKAGE_PATH_NOT_EXPORTED)
## Problem

`jsLoader.ts`'s `resolveModule` function uses `createRequire(referrer).resolve(specifier)` (the CJS resolver) to turn a module specifier into a file path. This works for CJS packages and dual-mode packages that include a `"require"` condition in their `exports` map, but fails with `ERR_PACKAGE_PATH_NOT_EXPORTED` for pure-ESM packages whose `exports` map only has `"import"` conditions.

Example: `@octokit/app@15` exports:

```json
{
  ".": {
    "node":    { "types": "...", "import": "./dist-node/index.js" },
    "default": { "types": "...", "import": "./dist-node/index.js" }
  }
}
```

Node's CJS `require.resolve()` sees the exports map but finds no `require` condition and throws:

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in
  .../node_modules/@octokit/app/package.json
    at exportsNotFound (node:internal/modules/esm/resolve:313:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:603:13)
    at resolveExports (node:internal/modules/cjs/loader:646:36)
    at Function._findPath (node:internal/modules/cjs/loader:713:31)
    at Function._resolveFilename (node:internal/modules/cjs/loader:1351:27)
    at Function.resolve (node:internal/modules/helpers:145:19)
    at resolveModule (harper/security/jsLoader.ts:190:51)
    at linker (harper/security/jsLoader.ts:373:23)
```

This surfaces as a `ResourceLoadError` when any Harper application imports a pure-ESM npm package.

## Root cause

`jsLoader.ts` line 190:

```typescript
const resolved = createRequire(resolveReferrer).resolve(specifier);
```

The CJS resolver does not understand the `"import"` condition in an exports map. Pure-ESM packages (increasingly common — `@octokit/*`, `chalk`, `got`, `sindresorhus/*`, etc.) only publish `"import"` conditions and have no `"require"` fallback.

## Fix

Catch `ERR_PACKAGE_PATH_NOT_EXPORTED` and fall back to `import.meta.resolve()`, which uses the ESM resolver and correctly handles `"import"` conditions. `import.meta.resolve()` is stable and synchronous since Node.js 20.6.

See the linked PR for the one-liner change.

Ethan-Arrowood:
Fixes #826.

Pure-ESM packages — those with an `exports` map containing only `"import"` conditions and no `"require"` — cause `createRequire().resolve()` to throw `ERR_PACKAGE_PATH_NOT_EXPORTED` in Harper's VM module linker. This is because the CJS resolver doesn't match the `"import"` condition.

## Changes

Added two module-level helpers in `security/jsLoader.ts`:

- **`walkExportsConditions`**: recursively walks an exports-map entry (string/array/object) against a condition priority list, returning the first matching path string.
- **`resolveESMPackageExports`**: walks the filesystem up from the referrer directory to locate the package root, reads `package.json`, determines the correct exports map entry for the subpath, and resolves it with `['import', 'node', 'default']` conditions. Returns an absolute `file://` URL.

In `resolveModule`, an `ERR_PACKAGE_PATH_NOT_EXPORTED` catch block now calls `resolveESMPackageExports` and returns the resolved URL, which `createModule` then loads via dynamic `import()`. If resolution fails (unknown format), the original error is rethrown.

**Exports map shapes handled**: bare string, subpath map (`{ ".": ... }`), root conditions map (`{ "import": ..., "require": ... }`), nested conditions (`{ "default": { "import": ... } }`), arrays.

## Testing

New regression test `pure-ESM package resolution` in `unitTests/security/jsLoader/jsLoader.test.js` uses a fixture package at `fixtures/esm-only-test/node_modules/pure-esm-pkg/` with `"type": "module"` and a `{ ".": { "default": { "import": "./index.js" } } }` exports map. All 339 security unit tests pass.

*Generated by Claude Sonnet 4.6*

### DIFF
--- a/security/jsLoader.ts
+++ b/security/jsLoader.ts
@@ -143,6 +143,83 @@ function parseJsonModule(source: string, url: string): any {
 }
 
 /**
+ * Walk an exports-map entry (string, array, or conditions object) with the given
+ * condition priority list and return the first matching path string, or null.
+ */
+function walkExportsConditions(entry: unknown, conditions: readonly string[]): string | null {
+	if (typeof entry === 'string') return entry;
+	if (Array.isArray(entry)) {
+		for (const e of entry) {
+			const r = walkExportsConditions(e, conditions);
+			if (r !== null) return r;
+		}
+		return null;
+	}
+	if (entry !== null && typeof entry === 'object') {
+		for (const cond of conditions) {
+			const val = (entry as Record<string, unknown>)[cond];
+			if (val !== undefined) {
+				const r = walkExportsConditions(val, conditions);
+				if (r !== null) return r;
+			}
+		}
+	}
+	return null;
+}
+
+/**
+ * Resolve a bare package specifier to a file:// URL using the package's exports map
+ * with ESM import conditions. Used as a fallback when createRequire().resolve() throws
+ * ERR_PACKAGE_PATH_NOT_EXPORTED for pure-ESM packages (exports map with only "import"
+ * conditions and no "require").
+ */
+function resolveESMPackageExports(specifier: string, fromDir: string): string | null {
+	const isScoped = specifier.startsWith('@');
+	const parts = specifier.split('/');
+	const packageName = isScoped ? `${parts[0]}/${parts[1]}` : parts[0];
+	const subpathParts = parts.slice(isScoped ? 2 : 1);
+	const subpath = subpathParts.length > 0 ? './' + subpathParts.join('/') : '.';
+
+	let dir = fromDir;
+	while (true) {
+		const pkgRoot = join(dir, 'node_modules', packageName);
+		let pkgJson: any;
+		try {
+			pkgJson = JSON.parse(readFileSync(join(pkgRoot, 'package.json'), 'utf-8'));
+		} catch {
+			const parent = dirname(dir);
+			if (parent === dir) return null;
+			dir = parent;
+			continue;
+		}
+
+		if (!pkgJson.exports) return null;
+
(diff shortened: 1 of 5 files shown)


kriszyp (approved):
Looks good, good catch!

security/jsLoader.ts
```diff
@@ -142,6 +142,83 @@ function parseJsonModule(source: string, url: string): any {
...
+/**
+ * Resolve a bare package specifier to a file:// URL using the package's exports map
+ * with ESM import conditions. Used as a fallback when createRequire().resolve() throws
+ * ERR_PACKAGE_PATH_NOT_EXPORTED for pure-ESM packages (exports map with only "import"
+ * conditions and no "require").
+ */
+function resolveESMPackageExports(specifier: string, fromDir: string): string | null {
+	const isScoped = specifier.startsWith('@');
+	const parts = specifier.split('/');
+	const packageName = isScoped ? `${parts[0]}/${parts[1]}` : parts[0];
```
cb1kenobi, line 179: What's the point of splitting the `specifier` only to join it? Is there a scenario where `specifier != packageName`? If `specifier` is a file url, then `parts[0]` is gonna be `file:`, which is not the package name.

cb1kenobi (approved):

PR #844, opened 2026-05-28
fix: persist indexingFailed in runIndexing outer catch

### LINKED ISSUE #843: runIndexing: outer catch does not persist indexingFailed, leaving wedged isIndexing after clear() crash
## Summary

When `runIndexing()` in `core/resources/databases.ts` encounters an unhandled error (e.g., `clear()` failing because a RocksDB column family was dropped), the outer `catch` only logs the error and returns. It **does not** write `indexingFailed = true` back to the persisted attribute descriptor.

This means:

- `indexingPID` in the DB descriptor still holds the now-dead process PID (written before `runIndexing` was called)
- `isIndexing` stays `true` on the in-memory dbi
- `indexingFailed` is **never set**, so the failure state is invisible
- On every subsequent restart, `indexingPID !== process.pid` triggers a re-run which may encounter the same error → infinite loop
- Searches on the affected attribute 503 indefinitely with no operator-visible failure signal

## Contrast with the inner error path

The `hadIndexingErrors` path (lines 1329–1352) correctly persists `indexingFailed = true` and keeps `isIndexing = true` so queries return 503 rather than partial results. The outer `catch` must do the same.

## Observed in production

Cluster `mynpga`/`jeff-cerro` (5.0.21, RocksDB). A `@computed(version:1→2)` bump on `Goat.searchTokens` raced a concurrent `drop_table`, causing `clear()` to fail with "Column family dropped during compaction". `isIndexing` was left stuck on west for 23 days. The cluster self-healed only when west restarted and the CF was cleanly recreated — there was no operator-visible failure state and no recovery path without a restart that happened to land clean.

## Affected code

`core/resources/databases.ts` — the outer `catch` at approximately line 1374:

```typescript
} catch (error) {
    logger.error('Error in indexing', error);
    // ← indexingFailed not persisted here
}
```

## Fix

In the outer `catch`, mirror the inner `hadIndexingErrors` block: set `attribute.indexingFailed = true`, write the descriptor to `Table.dbisDB`, and keep `isIndexing = true`. This ensures the failed state is persistent across restarts.

## Related

- rocksdb-js: `clear()` on a dropped CF should return OK (companion fix)

kriszyp:
## Summary

- `runIndexing`'s outer `catch` now persists `indexingFailed = true` for each attribute before exiting
- Uses `Promise.all` over all descriptor puts to ensure all writes are awaited
- Logs a warning if the secondary persist attempt itself fails (best-effort)

## Context

When an error fires before the record scan begins (e.g. `clear()` throws `kColumnFamilyDropped` due to a concurrent `drop_table`), the outer catch previously only logged the error. `indexingPID` is written to the descriptor *before* `runIndexing` is called, so on restart the PID mismatch re-triggers `runIndexing`, which fails again — an infinite silent loop where all searches return `503 not indexed yet`.

The fix mirrors the `hadIndexingErrors` inner path: write `indexingFailed = true` into the descriptor so the next startup sees it, clears the stale PID, and triggers a fresh rebuild from a known failure state.

Closes #843

> Generated by Claude Sonnet 4.6

### DIFF
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -1396,6 +1396,24 @@ async function runIndexing(Table, attributes, indicesToRemove) {
 		}
 	} catch (error) {
 		logger.error('Error in indexing', error);
+		// Persist indexingFailed so the next restart re-triggers the rebuild from an
+		// explicitly failed state rather than silently looping. Without this,
+		// indexingPID (written before runIndexing was called) stays in the descriptor
+		// but indexingFailed is never set, leaving isIndexing stuck with no recovery
+		// signal. Mirrors the hadIndexingErrors path. harper#843
+		try {
+			const puts: Promise<unknown>[] = [];
+			for (const attribute of attributes) {
+				attribute.indexingFailed = true;
+				puts.push(Table.dbisDB.put(attribute.key, attribute));
+				attribute.dbi.isIndexing = true;
+				const activeDbi = Table.indices[attribute.name];
+				if (activeDbi) activeDbi.isIndexing = true;
+			}
+			await Promise.all(puts);
+		} catch (persistError) {
+			logger.warn('Failed to persist indexing failure state', persistError);
+		}
 	}
 }
 
--- a/unitTests/resources/schemaMigrationFragility.test.js
+++ b/unitTests/resources/schemaMigrationFragility.test.js
@@ -237,6 +237,78 @@ describe('schema-migration fragility: stale index entry from concurrent write du
 	});
 });
 
+describe('schema-migration fragility: outer catch does not persist indexingFailed when clear() throws (F4)', () => {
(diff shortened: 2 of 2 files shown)


kriszyp:
Addressing the gemini-code-assist suggestion: null guards for `Table.dbisDB`/`attribute.dbi` are not added because (1) the entire recovery block is already wrapped in `try/catch(persistError)` so any `TypeError` would be caught and logged, and (2) the mirrored `hadIndexingErrors` inner path uses the same pattern without guards — adding them only in the outer catch would be asymmetric. Both fields are required by table initialization. — Claude Sonnet 4.6

kriszyp:
/gemini this should be fine running on multiple threads, right?

cb1kenobi (approved):

PR #849, opened 2026-05-28
Fix circular dependency causing jsLoader tests to fail in isolation

Ethan-Arrowood:
## Summary

- `blob.ts` had a static top-level `import { HAS_BLOBS } from './auditStore.ts'` that created a cycle: `DatabaseTransaction → blob → auditStore → RecordEncoder → analytics/write → databases → Table → LMDBTransaction → DatabaseTransaction`
- When `jsLoader.test.js` ran in isolation, `DatabaseTransaction` was still mid-evaluation when `LMDBTransaction` tried to extend it, causing `TypeError: Class extends value undefined`
- Fix: move the `HAS_BLOBS` import inside `cleanupOrphans()` using `await import()` — this function is an infrequently-called cold-path for orphan blob cleanup, so the lazy import has no performance impact

The full-test-glob happened to load other files first that resolved `DatabaseTransaction` before the cycle was hit, masking the bug when running `unitTests/security/**/*.js`. Running `jsLoader.test.js` alone exposed it cold.


kriszyp (approved):
Looks good! Despite what Gemini may think!

Ethan-Arrowood:
I think the CI failures are flake?

kriszyp:
> I think the CI failures are flake?

Yes, I think so. Been working on trying to improve this, moving the API tests to your testing framework.

kriszyp:
Do you know if this issue affects v5.0? I see you made this a patch, but have we seen this occur in any v5.0 releases?

Ethan-Arrowood:
You cannot run the `jsLoader.test.js` in isolation on the `v5.0` branch so I guess, yes? This isn't exactly something that might affect end users so I suppose it doesn't need to be backported.

cb1kenobi (approved):

Issue #623, opened 2026-05-19
[MCP] E2E tests, docs, and migration from external addon


kylebernhardy:
**Scope.** Final v1 PR: end-to-end test suite using an off-the-shelf MCP SDK, user-facing documentation, sample MCP client configs, and the deprecation path for the external `HarperFast/mcp-server` addon.

**Design reference.** Sections "Verification", "Migration from the external addon", and "Client Configuration & Invocation" in #465.

**Acceptance criteria**
- [ ] New integration suite under `test/integration/mcp/` driven by `@modelcontextprotocol/sdk` (TypeScript SDK).
  - Operations profile: `initialize` → `tools/list` → `tools/call` for `describe_all`, `add_user`, `create_schema`, with varied roles.
  - Application profile: same flow for `get_*`, `search_*`, `create_*`, `update_*`, `delete_*` on a seeded table.
  - Pagination round-trip on `search_*`.
  - `tools/list_changed` fires after `alter_role` in another session.
  - Restricted-attribute write rejected.
  - Both Streamable HTTP and `harper mcp` stdio (UDS) connections exercised.
- [ ] Public docs: a new MCP section under existing Harper docs covering: enabling the feature, the two profiles, RBAC behavior, `harper mcp` CLI usage, Claude Desktop / Cursor / Zed config examples, and the migration mapping from the external addon.
- [ ] `HarperFast/mcp-server` repo: deprecation README pointing to the native integration; archive the repo on this PR's merge.
- [ ] Release notes drafted for inclusion in Harper's next changelog.
- [ ] `npm run test:integration:mcp` registered as a script and added to the integration shard rotation in CI.

**Out of scope.** Shipping defaults that auto-enable MCP (`static/defaultConfig.yaml` continues to omit the `mcp:` block — MCP stays opt-in via block presence). v1.1 work: OAuth, scoped MCP tokens, rate-limit telemetry.

**Stacks on.** #617, #618, #619, #620, #621, #622 (all v1 functional work).

**Branch & PR conventions**
- Branch: `feat/mcp-e2e-docs-migration`
- PR base: `main` (after the rest of v1 merges).
- Closes this issue; references #465. **Closing this issue closes the umbrella tracking on #465.**

**Smoke test**
```bash
npm run test:integration:mcp
# Expected: green across the new MCP suite.
# Plus: claude-desktop loaded with the doc-provided config can list and invoke
# at least one tool against a local Harper.
```

[...]

Issue #622, opened 2026-05-19
[MCP] Custom Resource opt-in via static mcpTools declaration


kylebernhardy:
**Scope.** Let component authors expose non-verb methods on custom `Resource` classes as MCP tools via an opt-in static declaration. Reflection alone never publishes arbitrary methods.

**Design reference.** Section "Application MCP → For **custom Resources**" in #465.

**Acceptance criteria**
- [ ] Support a static `mcpTools` declaration on Resource subclasses with shape:
  ```ts
  static mcpTools = [
    { method: 'recommendSimilar', name: 'recommend_similar', description: '...', inputSchema: { ... } },
  ];
  ```
- [ ] Each entry registers an MCP tool that calls the named method with the parsed arguments.
- [ ] Authoring the JSON Schema by hand is required; **TS-type reflection into JSON Schema is out of scope** (explicit in #465).
- [ ] Methods named via `mcpTools` are RBAC-checked at call time using whatever predicates the Resource itself enforces; the MCP layer does not invent new RBAC for them.
- [ ] Verb-style tools (`get_*` / `search_*` / etc.) from #618 continue to be auto-published independently of `mcpTools`.
- [ ] Component-author docs page added (under existing components docs path) showing the opt-in pattern.
- [ ] Unit + integration tests with a test fixture Resource that declares `mcpTools`.

**Out of scope.** `@mcpTool` decorator syntax (mentioned in #465 as "or `@mcpTool` decorator" — defer to a v1.1 ergonomic pass). Anything that walks TS types.

**Stacks on.** #618 (application profile).

**Branch & PR conventions**
- Branch: `feat/mcp-custom-tools`
- PR base: `main` (after #618 merges).
- Closes this issue; references #465.

**Smoke test**
```ts
class Recommendations extends Resource {
  async recommendSimilar({ productId, limit }) { /* ... */ }
  static mcpTools = [{ method: 'recommendSimilar', name: 'recommend_similar', description: '...', inputSchema: {...} }];
}
// tools/list should include recommend_similar; tools/call routes to recommendSimilar.
```

**Tracking.** Part of HarperFast/harper#465. Sub-issue #10 of 11.

kriszyp:
> RBAC-checked at call time using whatever predicates the Resource itself enforces;

This is correct.

kylebernhardy:
[...]

PR #856, opened 2026-05-29
[MCP] v1 tools surface + listChanged + audit/rate-limit (consolidated: #617, #618, #619, #620, #622)

kylebernhardy:
## Summary

Consolidated PR-1 of the MCP v1 finish-line, per Kris's preference for fewer, larger reviewable PRs (his #788 review comment). Closes **five** sub-issues of #465 in one branch:

- **#617** — Operations profile: tool generation over `OPERATION_FUNCTION_MAP` with hand-curated JSON Schemas + allow/deny config.
- **#618** — Application profile: verb tool generation over the `Resources` registry, with input schemas derived from `Table.attributes`. **Also honors `exportTypes` (#788 kriszyp review)** — `mcp.exportTypes` opt-out + `http: false` mirroring.
- **#619** — `notifications/{tools,resources}/list_changed` over a real GET-SSE channel, per-session bookkeeping, diff-and-suppress on the per-session visible set.
- **#620** — Audit logging + per-`(session, tool)` token-bucket rate limiting.
- **#622** — Custom Resource opt-in via `static mcpTools = [...]` declaration.

A second consolidated PR (`feat/mcp-distribution`, planned next) will follow with #621 (stdio CLI) + #623 (E2E + docs + addon migration).

## Where to put attention

**Highest-risk surface** — `components/mcp/tools/application.ts`. This is ~600 LOC and ties together verb-presence introspection, schema derivation, `exportTypes` gating, name sanitization + collision recovery, and dispatch into Harper's static `Resource.get/put/patch/delete/post/search` methods. Worth a careful read.

**Look at this carefully:**
- `components/mcp/tools/application.ts` — dispatch path correctness, especially how `RequestTarget` is populated for each verb.
- `components/mcp/listChanged.ts` — the per-session diff is the security boundary that prevents users from learning about objects they can't see. Verify the diff logic correctly suppresses unchanged sets.
- `components/mcp/transport.ts` — the GET `/mcp` SSE channel open path (`handleGet`). Lifecycle, error paths, header validation.
- `components/mcp/rateLimit.ts` — token-bucket math + the `release()` contract; the lazy idle-prune was added to address a gemini-spotted leak.
- `components/mcp/audit.ts` — redaction depth was bumped from 5 → 10 with overflow-redaction after gemini caught a leak path.
- `server/serverHelpers/serverUtilities.ts` — single-line change: `OPERATION_FUNCTION_MAP` now exported (was module-private). Harper already exposes it for mutation via `server.registerOperation`, so the surface is no wider than today.
- `server/itc/serverHandlers.js` — added `schemaHandler.addListener` mirroring the existing `userHandler.addListener` pattern. Named exports added so the MCP layer can subscribe by handler reference.

**Less risky:**
- `components/mcp/tools/operations.ts` — straightforward delegator to `chooseOperation` + `processLocalTransaction`. Schemas are hand-curated data.
- `components/mcp/tools/schemas/derive.ts` — pure data transformation; well-covered by `derive.test.js` (10 tests).

## Implications on other functionality

- The `Allow` header on `/mcp` now lists `POST, GET` (and optionally `DELETE`). GET is no longer a 405 since #619.
- Components that register custom Resources can now opt out of MCP exposure via `exportTypes: { mcp: false }`.
- Tool registrations from `OPERATION_FUNCTION_MAP` and the `Resources` registry happen at component boot — order matters slightly (Resources must be populated before `handleApplication` fires; this is already the convention).

## Verification

- **Build / lint / format**: all clean.
- **Unit tests**: 251 passing across the MCP suite (up from 178 before this PR).
- **Integration tests**: extended `integrationTests/server/mcp.test.ts` with `tools/list` happy path, `tools/call describe_all` end-to-end, and unknown-tool `-32601`. The application-profile tools are exercised by the existing CI's Harper boot.
- **Cross-model review**: gemini review pass run on the full diff before opening this PR; two findings (audit redaction depth, rate-limit state leak) addressed in the final commit. Other findings were either spec-correct or already mitigated.

## Note

🤖 Generated with [Claude Code](https://claude.com/claude-code). Reviewer: kriszyp.

### DIFF
--- a/components/mcp/rateLimit.ts
+++ b/components/mcp/rateLimit.ts
@@ -106,11 +106,35 @@ interface SessionState {
 	inFlight: number;
 	config: RateLimitConfig;
 	profile: 'operations' | 'application';
+	lastSeen: number;
 }
 
 const sessions = new Map<string, SessionState>();
 
+// Belt-and-braces against state leaks: sessions that get TTL-evicted from the
+// system.mcp_session table never reach deleteSession() in this process, so
+// `clearSessionRateState` is never called for them. Prune any session that
+// hasn't admitted a call in this many ms on every getOrCreate. The threshold
+// is generously above the default idle timeout (1800s) — well-behaved live
+// sessions never get pruned by accident.
+const IDLE_PRUNE_MS = 60 * 60 * 1000; // 1 hour
+const PRUNE_INTERVAL_MS = 5 * 60 * 1000; // run at most every 5 minutes
+let lastPruneAt = 0;
+
+function pruneIdleSessions(): void {
+	const t = now();
+	if (t - lastPruneAt < PRUNE_INTERVAL_MS) return;
+	lastPruneAt = t;
+	const cutoff = t - IDLE_PRUNE_MS;
+	for (const [id, s] of sessions) {
+		if (s.inFlight === 0 && s.lastSeen < cutoff) {
+			sessions.delete(id);
+		}
+	}
+}
+
 function getOrCreate(sessionId: string, profile: 'operations' | 'application'): SessionState {
+	pruneIdleSessions();
 	let s = sessions.get(sessionId);
 	if (!s) {
 		const config = configFor(profile);
@@ -120,8 +144,11 @@ function getOrCreate(sessionId: string, profile: 'operations' | 'application'): 
 			inFlight: 0,
 			config,
 			profile,
+			lastSeen: now(),
 		};
 		sessions.set(sessionId, s);
+	} else {
+		s.lastSeen = now();
 	}
 	return s;
 }
--- a/components/mcp/rateLimit.ts
+++ b/components/mcp/rateLimit.ts
@@ -1,0 +1,175 @@
+/**
+ * Per-session, per-tool rate limiting for `tools/call`.
+ *
+ * Four configurable limits per profile (operations / application):
+ *   - perToolPerSecond:   sustained per-tool rate (token bucket refill)
+ *   - perToolBurst:       per-tool burst capacity (token bucket size)
+ *   - sessionConcurrency: max in-flight tool calls per session
+ *   - sessionPerSecond:   sustained per-session rate across all tools
+ *
+ * Limit hits surface as `result.isError = true` with `kind: 'rate_limited'`
+ * (NOT a JSON-RPC error) per the MCP spec's tools-call convention. The LLM
+ * sees and adapts; the protocol envelope stays clean.
+ *
+ * State is in-memory per worker process. Buckets are evicted lazily when
+ * a session's record is removed (#619 cleanup) or after they've been idle
+ * past the idle eviction threshold. Multi-process coordination isn't
+ * attempted in v1 — the limits are per-worker.
+ */
+import * as env from '../../utility/environment/environmentManager.ts';
+import { CONFIG_PARAMS } from '../../utility/hdbTerms.ts';
+
+export interface RateLimitConfig {
+	perToolPerSecond: number;
+	perToolBurst: number;
+	sessionConcurrency: number;
+	sessionPerSecond: number;
+}
+
+const DEFAULTS: Record<'operations' | 'application', RateLimitConfig> = {
+	operations: { perToolPerSecond: 10, perToolBurst: 20, sessionConcurrency: 25, sessionPerSecond: 100 },
+	application: { perToolPerSecond: 25, perToolBurst: 50, sessionConcurrency: 50, sessionPerSecond: 200 },
+};
+
+const CONFIG_KEYS = {
+	operations: {
+		perToolPerSecond: CONFIG_PARAMS.MCP_OPERATIONS_RATELIMIT_PERTOOLPERSECOND,
+		perToolBurst: CONFIG_PARAMS.MCP_OPERATIONS_RATELIMIT_PERTOOLBURST,
+		sessionConcurrency: CONFIG_PARAMS.MCP_OPERATIONS_RATELIMIT_SESSIONCONCURRENCY,
+		sessionPerSecond: CONFIG_PARAMS.MCP_OPERATIONS_RATELIMIT_SESSIONPERSECOND,
+	},
+	application: {
+		perToolPerSecond: CONFIG_PARAMS.MCP_APPLICATION_RATELIMIT_PERTOOLPERSECOND,
+		perToolBurst: CONFIG_PARAMS.MCP_APPLICATION_RATELIMIT_PERTOOLBURST,
+		sessionConcurrency: CONFIG_PARAMS.MCP_APPLICATION_RATELIMIT_SESSIONCONCURRENCY,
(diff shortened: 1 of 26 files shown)


components/mcp/rateLimit.ts
```diff
@@ -0,0 +1,202 @@
...
+	}
+	if (!state.sessionRate.tryConsume()) {
+		return { allowed: false, reason: 'session_rate' };
+	}
+	let toolBucket = state.perTool.get(toolName);
+	if (!toolBucket) {
+		toolBucket = new TokenBucket(state.config.perToolPerSecond, state.config.perToolBurst);
+		state.perTool.set(toolName, toolBucket);
+	}
+	if (!toolBucket.tryConsume()) {
```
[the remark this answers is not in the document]
kylebernhardy: Fixed in `eb3138646`. Went with the restructure (your "more principled" option) rather than the swap or the refund.

Added `TokenBucket.hasToken()` (refill + check, no deduct). `tryAdmit` now peeks both buckets first and only consumes when both have capacity. Symmetric — neither bucket can drain the other on a denial path.

Regression test asserts the exact scenario: `sessionPerSecond=5` + `perToolBurst=1`, hammer one throttled tool 11 times — first succeeds, next 10 denied with `per_tool` reason, session-rate untouched. Then exhaust the remaining 4 session-rate tokens on different tools and confirm the 5th attempt denies with `session_rate`.

components/mcp/tools/schemas/operations.ts
```diff
@@ -0,0 +1,245 @@
...
+	type: 'object',
+	additionalProperties: true,
+	description: 'Free-form arguments — Harper validates server-side and returns a structured error if invalid.',
+};
+
+/**
+ * Map of operation name → input schema. Lookup misses fall back to
+ * `PERMISSIVE_SCHEMA`. Keys mirror `OPERATIONS_ENUM` values.
+ */
+export const OPERATION_INPUT_SCHEMAS: Record<string, object> = {
```
kriszyp, line 33: Ugh, @claude please make sure to file a follow-up ticket for refactoring operations API registration in a way where the schema can validate and be introspected as valid JSON Schema.

components/mcp/tools/application.ts
```diff
@@ -0,0 +1,609 @@
+/**
+ * Application-profile tool generation. Walks Harper's `Resources` registry
+ * and registers verb tools (`get_*`, `search_*`, `create_*`, `update_*`,
+ * `delete_*`) for each exported Resource that:
+ *   1. Has a `'mcp'` exportType not explicitly set to `false`.
+ *   2. Has `'http'` exportType not explicitly set to `false` (Kris's review
+ *      on #788: the application MCP surface mirrors the public REST surface,
```
kriszyp: Avoid temporal/development comments (issues fine, not PRs).

components/mcp/tools/operations.ts
```diff
@@ -0,0 +1,268 @@
...
+	_opMapOverride = m;
+}
+export function _setChooseOperationForTest(fn: ChooseOperation | undefined): void {
+	_chooseOperationOverride = fn;
+}
+export function _setProcessLocalTransactionForTest(fn: ProcessLocalTransaction | undefined): void {
+	_processLocalTransactionOverride = fn;
+}
+
+function loadServerUtilities():
```
kriszyp, line 53: Hoping the latest TypeScript mitigates the need for this, should file a ticket for circular dependency workaround cleanup.

components/mcp/resources.ts
```diff
@@ -310,6 +312,30 @@ function enumerateAppHttpResources(): ResourceDescriptor[] {
...
+	return types.mcp !== false;
+}
+
+/**
+ * The application MCP surface mirrors the public REST surface. A Resource
+ * that operators explicitly hid from HTTP (`exportTypes.http === false`)
+ * should not be advertised as an `https://...` URI here either — even if
+ * the same Resource is still exposed via, say, WebSockets or GraphQL.
+ */
+function isHttpExposed(entry: { exportTypes?: unknown }): boolean {
```
kriszyp: I think we could just go with the `mcp` exportTypes flag, don't need to mirror http that strongly.

kriszyp (approved):
## Approve ✅

Solid, well-structured work — clear module boundaries, consistent test seams, and genuinely strong coverage (251 unit + 3 integration). The prior review fixes are all correctly landed: the token-bucket peek-then-consume avoids draining the unrelated bucket on denial, `release()` is in a `finally`, and the depth-10 audit redaction with overflow-redaction closes the buried-credential path. Operation injection is handled cleanly (args spread first, `operation`/`hdb_user` set after). Nice work. 🎉

Approving for merge. A few non-blocking follow-ups worth tracking — none is an auth hole, but they're worth a look before this surface gets heavy use. (See inline comments for the schema-introspection, circular-dep, exportTypes, and temporal-comment notes.)

### Recommended follow-ups

1. **`sessionRegistry` leak on dropped SSE connections (medium).** `unregisterSession` is only reached via `deleteSession` or session supersede. As the rate-limiter's own comment notes, TTL-evicted sessions never reach `deleteSession` — which is why `rateLimit.ts` grew an `IDLE_PRUNE_MS` sweep. `sessionRegistry` has no equivalent and no disconnect hook on the returned `sseIterable`, so a client that opens `GET /mcp` and drops the connection leaves its `RegisteredSession` + `IterableEventQueue` resident until a DELETE arrives. Same leak class just fixed for rate limits — suggest mirroring the idle-prune (or wiring an on-close from the SSE consumer).

2. **`listChanged` recomputes against a stale user snapshot (medium).** `registerSession` captures `effectiveUser(request)` once at GET time, and `onUserChange` recomputes each session's visible set against that frozen `record.user`. When a role's permissions change, the diff runs against the *old* permissions, so the event that should fire `notifications/tools/list_changed` may compute "no change" and suppress it. Actual `tools/call` re-authorizes, so this is notification accuracy rather than an auth bypass — but it undercuts #619's intent for the user-change channel. Consider re-resolving the user inside the handler.

3. **`get_*` default-allow glob breadth (verify).** Worth confirming nothing sensitive matches the `get_*` glob in the default read-only allow list (e.g. `get_configuration`, if it exists, surfacing config/secrets). `verifyPerms` still gates server-side, so defense-in-depth holds — but a broad default glob tends to drift. Enumerating named read ops would be tighter.

### exportTypes (per inline note)

Agreed on dropping the `http` mirroring — key enumeration purely off `exportTypes.mcp !== false`. Concretely: remove the `types.http === false` short-circuit in `shouldEnumerate` (application.ts) and the `isHttpExposed` gate in `resources.ts`. An operator who wants a Resource off the MCP surface sets `exportTypes.mcp: false` explicitly; tying it to `http` is surprising and over-coupled.

Filed two follow-ups during review: HarperFast/harper#878 (operations schema → introspectable JSON Schema) and HarperFast/harper#879 (lazy-`require` circular-dep cleanup post-TS-migration).

_Review drafted by Claude on behalf of @kriszyp._

PR #859, opened 2026-05-29
fix(analytics): recordAction ignores invalid metric values instead of throwing

### LINKED ISSUE #858: Replication loop with v4 peer causes repeated full table copies on v5
### Bug Summary

v5 node loops requesting full table copies when replicating with a v4 peer

_No response_

### Steps to Reproduce

1. Run a cluster with at least one v5 node and one v4 (I used 4.7.32) node replicating to each other.
2. Let the v5 node receive replication from the v4 peer.
3. Watch the v5 node logs.

### Expected Behavior

Replication runs normally. No repeated full table copies.

### Actual Behavior

Every 60 to 90 seconds the v5 node logs a `TypeError`, the websocket closes with code `1006`, the node reconnects, and it requests a full copy of the affected tables (sem and seo in our case, about 45k records). The cycle then repeats and does not stop. About 1900 of these errors were logged in one session. The repeated full copies also produce many `Transaction open too long` errors on the affected tables.

### Platform

Harper v5.0.21 receiving replication from v4 (4.7.32) peers

### Console Errors

```
TypeError: Invalid metric value type object
at recordNewAction (resources/analytics/write.ts:97)
at recordAction (resources/analytics/write.ts:111)
at onWSMessage (replicationConnection.ts:764)
```

### Screenshots or Videos

_No response_

### Additional Context

_No response_

ldt1996:
## Summary

- `recordAction` now checks the value type up front and returns early instead of throwing when the value is not a number, boolean, or function
- Warns once (module-level flag) so a bad value is visible without flooding the log
- Adds a regression test that `recordAction({})` and `recordAction(undefined)` no-op instead of throwing

## Context

On a v5 node receiving replication from a v4 (4.7.32) peer, the replication receive handler calls `recordAction(data, 'bytes-received', ...)` with `data` from a `COMMITTED_UPDATE` message. From v4 that value is an object or undefined, and `recordAction` (through `recordNewAction`/`recordExistingAction`) throws `TypeError: Invalid metric value type ...` on anything that is not a number, boolean, or function. The outer `onWSMessage` catch swallowed the throw, but the handler aborted before its confirmation step ran, so the replication handshake stalled, the websocket closed with code 1006, and the node reconnected and requested a full table copy. The next `COMMITTED_UPDATE` threw again, repeating the loop every 60 to 90 seconds.

Metrics are best-effort and must never crash the caller, especially the replication receive path. This makes `recordAction` skip values it cannot aggregate instead of throwing, so the handler always completes. The call-site that passed the wrong value (a confirmation status instead of a byte count) is fixed separately in harper-pro.

Closes #858

### DIFF
--- a/resources/analytics/write.ts
+++ b/resources/analytics/write.ts
@@ -40,6 +40,7 @@ let activeActions = new Map<string, Action>();
 let analyticsEnabled = envGet(CONFIG_PARAMS.ANALYTICS_AGGREGATEPERIOD) > -1;
 let analyticsReadOnlyChecked = false;
 let sendAnalyticsTimeout: NodeJS.Timeout;
+let warnedInvalidMetricValue = false;
 
 // Check read-only mode lazily to avoid circular dependency at module load time
 function checkAnalyticsEnabled(): boolean {
@@ -115,6 +116,15 @@ function recordNewAction(key: string, value: Value, metric?: string, path?: stri
  */
 export function recordAction(value: Value, metric: string, path?: string, method?: string, type?: string) {
 	if (!checkAnalyticsEnabled()) return;
+	const valueType = typeof value;
+	if (valueType !== 'number' && valueType !== 'boolean' && valueType !== 'function') {
+		// metrics are best-effort; never throw on a value we can't aggregate
+		if (!warnedInvalidMetricValue) {
+			warnedInvalidMetricValue = true;
+			log.warn?.(`Ignoring analytics value of type ${valueType} for metric ${metric}`);
+		}
+		return;
+	}
 	// TODO: May want to consider nested paths, as they may yield faster hashing of (fixed) strings that hashing concatenated strings
 	let key = metric + (path ? '-' + path : '');
 	if (method !== undefined) key += '-' + method;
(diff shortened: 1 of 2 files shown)


ldt1996:
related PR https://github.com/HarperFast/harper-pro/pull/242

cb1kenobi (approved):
LGTM

kriszyp (changes requested):
Generally speaking, fail fast is the preferred behavior. If there is an error, throwing it is the right thing to do. In this case, the thrown error _directly_ led you to the correct root cause and fix with https://github.com/HarperFast/harper-pro/pull/242 , right? Without this error being thrown, we could have easily brushed the warning aside, and let the latent bug exist for years. A warning like this without any stack trace is more problematic because it could be difficult to determine who called it and where it originated from.

And does this need to be a patch, harper-pro#242 is the real fix for #858, right?

All that said, I do acknowledge the argument that `recordAction` is not performing essential work for the caller, and inherently can continue without interrupting the caller's actions, making it suitable for logging instead of throwing. I'm ok if you feel that it is right to catch and log. But log once would still be dangerous (misses a subsequent metrics failure that might be more problematic), and should at least have the stack trace for debugging.

ldt1996:
@kriszyp I agree with your view on this after taking a better look.  There's one known caller bug, and it's fixed by the harper-pro PR. You're right that loosening the throw might hide a possible next issue. Closing this.

PR #860, opened 2026-05-29
fix(databases): rebind table store after in-place engine migration

ldt1996:
## Summary

- Recreate a table's store when its engine no longer matches what's on disk (LMDB vs RocksDB), instead of reusing the stale entry from the per-thread registry.
- Stop deleting the database entry in `resetDatabases()`, which was also tearing down replication subscriptions via the `dropDatabase` event.

## Context

After an in-place v4 to v5 `migrateOnStart`, the `.mdb` files are moved to `backup/` and RocksDB directories are created in their place. The per-thread `databases` registry still holds the old table whose `primaryStore.path` points at the now-missing `.mdb` file. Anything that reads that path fails, e.g. DB size/volume analytics throw `ENOENT` on the `.mdb` every aggregation cycle.

The previous attempt to fix this in `resetDatabases()` deleted the database entry and emitted `dropDatabase` so the next `getDatabases()` would rebuild it. That also dropped the database's replication subscriptions, so it wasn't safe.

This change instead detects the engine mismatch in `initStores`: if an already-defined table's `primaryStore.rootStore` is a `RocksDatabase` but the current `rootStore` is not (or vice versa), the table is rebuilt against the current engine rather than reused. `resetDatabases()` now only closes the migrated store and clears its env entry, leaving the registry rebind to `initStores`.

## Test

`unitTests/resources/schemaMigrationFragility.test.js` adds an F4 case that:
- creates an LMDB-backed table and writes a row,
- swaps the engine to RocksDB on disk while leaving the stale LMDB table in the registry (the post-migration state),
- calls `resetDatabases()` and asserts the reloaded table's `primaryStore` is now backed by `RocksDatabase` and its path is a directory, not a `.mdb` file.

Fails on `main`, passes with this change. Skipped under `HARPER_STORAGE_ENGINE=lmdb`.

### DIFF
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -546,6 +546,9 @@ function initStores(
 		}
 		// if the table has already been defined, use that class, don't create a new one
 		let table = tables[tableName];
+		// unless its store was migrated to a different engine (e.g. LMDB to RocksDB on startup)
+		const recreateForEngineChange =
+			!!table && (table as any).primaryStore?.rootStore instanceof RocksDatabase !== rootStore instanceof RocksDatabase;
 		let indices = {},
 			existingAttributes = [];
 		let tableId;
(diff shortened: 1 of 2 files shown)


kriszyp (approved):
Good job, Lavinia! The root cause analysis is thorough and the fix is well-scoped — detecting the engine mismatch in `initStores` rather than nuking the registry entry is exactly the right call. Test coverage is solid too. 🎉

— Claude

cb1kenobi (approved):

PR #862, opened 2026-05-29
fix(analytics): continue instead of return so later dbs still get merics

ldt1996:
## Summary

- Use `continue` instead of `return` when a database has no audit size or storage stats, so metrics for later databases still get recorded.

## Context

`storeDBSizeMetrics` and `storeVolumeMetrics` loop over all databases. When one had a falsy audit size (or no storage stats), the early `return` exited the whole loop, so every database after it was silently skipped that cycle. Switching to `continue` skips only the empty database and keeps going.

### DIFF
--- a/resources/analytics/write.ts
+++ b/resources/analytics/write.ts
@@ -335,7 +335,7 @@ function storeDBSizeMetrics(analyticsTable: Table, databases: Databases) {
 			const [firstTable] = Object.values(tables);
 			const dbAuditSize = firstTable?.getAuditSize();
 			if (!dbAuditSize) {
-				return;
+				continue;
 			}
 			let metric;
 			if (firstTable.primaryStore instanceof RocksDatabase) {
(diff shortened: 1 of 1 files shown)


resources/analytics/write.ts
```diff
@@ -335,7 +335,7 @@ function storeDBSizeMetrics(analyticsTable: Table, databases: Databases) {
 			const [firstTable] = Object.values(tables);
 			const dbAuditSize = firstTable?.getAuditSize();
 			if (!dbAuditSize) {
-				return;
+				continue;
```
cb1kenobi, line 338: Do we also care that if `dbAuditSize` is `0` that we continue instead of storing it?

cb1kenobi (approved):

kriszyp (approved):
Nice job finding this.

resources/analytics/write.ts
kriszyp: Good catch — agreed. `getAuditSize()` returns `undefined` when there's no audit store (or no first table), but a real `0` once an audit store exists with no logged transactions yet, and such a database can still have a meaningful on-disk size. The falsy `!dbAuditSize` check was dropping the `DATABASE_SIZE` metric in that case.

Pushed 38c3f9d7 to guard on `dbAuditSize == null` instead, so a `0` audit size still stores the metric (with `transactionLog`/`audit` of `0`) while we keep skipping only when there's no table/audit store at all. 🤖 (Claude)

PR #871, opened 2026-05-29
Record RocksDB statistics in analytics

### LINKED ISSUE #163: Record RocksDB statistics in analytics
With https://github.com/HarperFast/rocksdb-js/pull/292 and https://github.com/HarperFast/rocksdb-js/pull/423, we now have access to extensive RocksDB statistics. We should record important statistics (perhaps reads, writes, and block cache hits/misses)  in https://github.com/HarperFast/harper/blob/main/resources/analytics/write.ts#L331-L345.

cb1kenobi:
## Summary

Closes #163.

Adds a `rocksdb-stats` metric to the analytics table, populated by the existing aggregation cycle in `resources/analytics/write.ts`. Hook lives next to `storeDBSizeMetrics` / `storeVolumeMetrics` and runs only for RocksDB-backed databases.

For each RocksDB database, one DB-level row and one row per table are emitted per period:

- **DB-level counters (diffed):** `bytesRead`, `bytesWritten`, `numberKeysRead`, `numberKeysWritten`, `blockCacheHit`/`Miss` (overall + `Data`/`Index`/`Filter` splits), `stallMicros`.
- **DB-level gauges (absolute):** `blockCacheUsage`, `blockCacheCapacity`, `numRunningFlushes`.
- **Per-table counters (diffed):** `memtableHit`, `memtableMiss`.
- **Per-table gauges (absolute):** `numRunningCompactions`, `compactionPending`.

Histogram fields (e.g. `dbGetMicros`) are intentionally skipped — they don't fit the flat analytics row model. Follow-up if percentile storage is wanted.

## Where to look

- **Counter-diff semantics** ([`write.ts` `diffRocksDBCounter`](resources/analytics/write.ts)): a negative delta returns the current absolute value. This catches RocksDB-internal stat resets; process restarts are handled separately by the module-level cache being empty (first-sample skip).
- **First-sample skip** in `storeRocksDBStatsMetrics`: the first read for a (db, table) pair only populates the baseline and writes nothing, so reporting a counter as a delta-from-zero doesn't produce an artificial spike. **Side effect:** the first `rocksdb-stats` row only appears after the second qualifying aggregation cycle — roughly 60–120s of uptime with the default `aggregatePeriod: 60`.
- **Cache pruning** in `pruneRocksDBStatsCache`: drops cache entries for databases/tables no longer in `getDatabases()` to bound memory and keep drop/recreate from yielding stale baselines. Note `system` is non-enumerable on the databases object and is added explicitly.
- **`instanceof RocksDatabase` check** is on the first table per database only. Harper tables in a single database share one root store, so this is sound today; would need revisiting if mixed-engine databases ever exist.

## Test plan

- [x] Unit tests for the pure builder helpers (`diffRocksDBCounter`, `normalizeRocksDBStats`, `buildRocksDBDb/TableMetric`, `toRocksDBCamelCase`) — 14 new tests in [`unitTests/resources/analytics/write.test.js`](unitTests/resources/analytics/write.test.js).
- [x] Local integration: ran a RocksDB-backed instance, issued `get_analytics { metric: "rocksdb-stats" }` after ~2 minutes of uptime, confirmed rows for each database + per-table rows with sensible counters and gauge values.
- [ ] CI on this branch.

## Out of scope

- LMDB stat recording in analytics.
- Histogram (`dbGetMicros` etc.) percentile storage.
- Surfacing in Studio / dashboards.


resources/analytics/write.ts
```diff
@@ -396,6 +396,196 @@ function storeVolumeMetrics(analyticsTable: Table, databases: Databases) {
...
+
+/**
+ * Filter a raw RocksDB stats record to numeric scalar entries, camelCasing the keys.
+ * Histogram entries (objects) are dropped — they don't fit the flat analytics row model.
+ * @param raw - The raw RocksDB stats record to normalize.
+ * @returns The normalized RocksDB stats record.
+ */
+export function normalizeRocksDBStats(raw: Record<string, unknown>): Record<string, number> {
+	const out: Record<string, number> = {};
+	for (const [key, value] of Object.entries(raw)) {
```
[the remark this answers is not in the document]
cb1kenobi: `raw` will always have a non-null value.

resources/analytics/write.ts
```diff
@@ -396,6 +396,196 @@ function storeVolumeMetrics(analyticsTable: Table, databases: Databases) {
...
+ * growth across drop/recreate cycles and prevent stale baselines from producing wrong diffs.
+ * @param databases - The databases to prune the stats cache for.
+ * @param systemDatabase - The system database to prune the stats cache for.
+ */
+function pruneRocksDBStatsCache(databases: Databases, systemDatabase: Tables | undefined) {
+	const activeDbs = new Set<string>(Object.keys(databases));
+	const activeTables = new Set<string>();
+	for (const [db, tables] of Object.entries(databases)) {
+		for (const tableName of Object.keys(tables)) activeTables.add(`${db}.${tableName}`);
+	}
```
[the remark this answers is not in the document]
cb1kenobi: Fixed

resources/analytics/write.ts
```diff
@@ -396,6 +396,196 @@ function storeVolumeMetrics(analyticsTable: Table, databases: Databases) {
...
+ * @param period - The period to store the metrics for.
+ */
+function storeRocksDBStatsMetrics(
+	analyticsTable: Table,
+	databases: Databases,
+	now: number,
+	period: number | undefined
+) {
+	for (const [db, tables] of Object.entries(databases)) {
+		const tableEntries = Object.entries(tables);
```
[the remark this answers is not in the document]
cb1kenobi: Fixed

resources/analytics/write.ts
```diff
@@ -396,6 +396,196 @@ function storeVolumeMetrics(analyticsTable: Table, databases: Databases) {
...
+		const [, firstTable] = tableEntries[0] ?? [];
+		if (!(firstTable?.primaryStore instanceof RocksDatabase)) continue;
+
+		let firstNormalizedStats: Record<string, number> | undefined;
+		for (const [tableName, tbl] of tableEntries) {
+			try {
+				const tableStats = normalizeRocksDBStats(tbl.primaryStore.getStats());
```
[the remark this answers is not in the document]
cb1kenobi: Non-issue.

resources/analytics/write.ts
```diff
@@ -628,6 +820,13 @@ async function aggregation(fromPeriod, toPeriod = 60000) {
 	storeVolumeMetrics(analyticsTable, databases);
 	storeVolumeMetrics(analyticsTable, { system: databases.system });
 
+	// rocksdb engine stats (only for RocksDB-backed databases)
+	const rocksDBPeriod = lastRocksDBStatsTime ? now - lastRocksDBStatsTime : undefined;
+	storeRocksDBStatsMetrics(analyticsTable, databases, now, rocksDBPeriod);
```
kriszyp, line 822:
```suggestion
	storeRocksDBStatsMetrics(analyticsTable, { ...databases, system: databases.system }, now, rocksDBPeriod);
```
Maybe do this in one call, or have a shared function that computes the set of database names, since there are a couple places where we recompute the database names, handling the non-enumerated `system` database?

resources/analytics/write.ts
```diff
@@ -396,6 +396,198 @@ function storeVolumeMetrics(analyticsTable: Table, databases: Databases) {
...
+	'blockCacheDataHit',
+	'blockCacheDataMiss',
+	'blockCacheIndexHit',
+	'blockCacheIndexMiss',
+	'blockCacheFilterHit',
+	'blockCacheFilterMiss',
+	'stallMicros',
+] as const;
+const ROCKSDB_DB_GAUGES = ['blockCacheUsage', 'blockCacheCapacity', 'numRunningFlushes'] as const;
+const ROCKSDB_TABLE_COUNTERS = ['memtableHit', 'memtableMiss'] as const;
```
kriszyp: `memtableHit` / `memtableMiss` here are DB-wide, not per-table. In `rocksdb-js`, `getStats()` reads `statistics->getTickerCount(MEMTABLE_HIT/MISS)` (`src/binding/database/db_descriptor.cpp:358-359`), and that `Statistics` object is created once per physical RocksDB (`db_descriptor.cpp:680`, passed to the descriptor at `:783`). All our tables in a database are column families on that one shared descriptor — they open against the same `rootStore.path` and `enableStats` is only set on the root open (`databases.ts:440/795`). RocksDB doesn't break ticker counts down per column family, so every table's `getStats()` returns the identical DB-wide number.

Net effect: each per-table row reports the same value, and since each is diffed against its own cached copy of that same DB-wide number, every table emits the *full DB delta* — summing across tables multiplies the real figure by the table count. These two should move to the DB-level row, not the per-table row.

— Claude

resources/analytics/write.ts
```diff
@@ -396,6 +396,198 @@ function storeVolumeMetrics(analyticsTable: Table, databases: Databases) {
...
+	'blockCacheMiss',
+	'blockCacheDataHit',
+	'blockCacheDataMiss',
+	'blockCacheIndexHit',
+	'blockCacheIndexMiss',
+	'blockCacheFilterHit',
+	'blockCacheFilterMiss',
+	'stallMicros',
+] as const;
+const ROCKSDB_DB_GAUGES = ['blockCacheUsage', 'blockCacheCapacity', 'numRunningFlushes'] as const;
```
kriszyp, line 426: Heads up: none of these gauges are actually emitted by `getStats()`. The `numRunningFlushes` / `blockCacheUsage` / `blockCacheCapacity` here (and `numRunningCompactions` / `compactionPending` in `ROCKSDB_TABLE_GAUGES` below) are RocksDB *properties* (`GetIntProperty`), not `Statistics` tickers/histograms. `getStats()` only returns the curated ticker + histogram list (`src/binding/database/db_descriptor.cpp:328-385`) — grepping that file for any of these names comes back empty.

So `normalizeRocksDBStats(getStats())` never contains them, and `stats[field] ?? 0` records a constant `0` for all five. To capture them we'd need a `getProperty()`-based path rather than `getStats()`.

— Claude

kriszyp (commented):
This looks good, a great curation of stats. I think there are a few things to resolve/verify in here though.

resources/analytics/write.ts
cb1kenobi: This is false. The values are there.

resources/analytics/write.ts
cb1kenobi: Fixed.

resources/analytics/write.ts
cb1kenobi: Fixed.

resources/analytics/write.ts
```diff
@@ -396,6 +396,193 @@ function storeVolumeMetrics(analyticsTable: Table, databases: Databases) {
...
+ * @param raw - The raw RocksDB stats record to normalize.
+ * @returns The normalized RocksDB stats record.
+ */
+export function normalizeRocksDBStats(raw: Record<string, unknown>): Record<string, number> {
+	const out: Record<string, number> = {};
+	for (const [key, value] of Object.entries(raw)) {
+		if (typeof value === 'number') out[toRocksDBCamelCase(key)] = value;
+	}
+	return out;
+}
```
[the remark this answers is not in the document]
cb1kenobi: Fixed.

PR #873, opened 2026-05-29
Ignore some AI dot folders

cb1kenobi:
Ignore the `.antigravitycli` and `.claude` directories.


kriszyp (approved):
I dunno that it needs to be patched, but welcomed either way.

PR #877, opened 2026-05-29
Strip PROXY protocol header on the MQTT/secure-socket UDS mirror

kriszyp:
## Summary
The per-worker UDS mirrors (created when `tls.unixDomainSockets` is enabled) let a fronting proxy (symphony) reach a worker directly and convey the real client address via a PROXY v1 header. The **HTTP** UDS mirror already strips that header (`enableProxyProtocol`), but the **raw secure-socket** UDS mirror (MQTT, replication) did not — so the header was delivered to the protocol parser and corrupted the first packet. For MQTT the first byte `0x50` ('P' of "PROXY ") parses as a PUBREC → `Invalid pubrec reason code`, connection dropped.

Fix: apply the same `enableProxyProtocol` wrapper to the secure-socket UDS server in `threadServer.js onSocket()`, and `export` it from `server/http.ts`.

## Why
Found while bringing up symphony→instance MQTT over UDS: MQTT CONNECTs never got a CONNACK. (Discovered via load testing; a host-manager-side workaround sets the MQTT route's source header to `none`, but the proper fix is here so MQTT keeps the real client IP.)

## Where to look
- `server/http.ts` — `enableProxyProtocol` is now exported (no logic change).
- `server/threads/threadServer.js` — one call added in the secure-socket UDS block, mirroring the existing HTTP UDS call.
- Test: `unitTests/server/udsMirror.test.js` — strips a PROXY v1 header / passes a non-PROXY chunk through.

🤖 Generated by Claude (Opus 4.7). `test:unit:server` udsMirror suite passes (15/15).

### DIFF
--- a/server/http.ts
+++ b/server/http.ts
@@ -986,8 +986,9 @@ function onWebSocket(listener: (ws: WebSocket) => void, options: OnWebSocketOpti
 
 // PROXY protocol v1 max header length per spec: 108 bytes
 const PROXY_V1_MAX_HEADER = 108;
+const PROXY_V1_PREFIX = Buffer.from('PROXY ');
 
-function enableProxyProtocol(httpServer) {
+export function enableProxyProtocol(httpServer) {
 	// In Node.js v24+, the HTTP parser's data path goes through the C++ stream layer
 	// and does not call socket.emit('data') via JavaScript method dispatch.
 	// Overriding socket.emit or socket.push has no effect on the HTTP parser's data intake.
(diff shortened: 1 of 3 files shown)


kriszyp:
Force-pushed `cc8dafe1f` to correct this branch. The earlier review was right — a local git-submodule corruption on my end caused the prior commits to record *reverts* (lost the `export`, the buffering, and the tests). The branch now contains the intended change: `export function enableProxyProtocol`, the `enableProxyProtocol(udsServer)` call on the secure-socket UDS mirror, the split/partial PROXY-header buffering, and the unit tests (udsMirror suite 18/18 locally).

Note on red CI (pre-existing on `main`, unrelated to this 3-file change): **Integration Tests 1/4** fails in `integrationTests/upgrade/4.x-upgrade.test.ts` (LMDB→RocksDB migration — `RecordEncoder.saveStructures` TypeError in `copyDb.ts`), and **Unit Test (Node v26)** is the flaky `unitTests/apiTests/ws-test.mjs` "default subscribe on WS" 5s timeout (v22/v24 pass). Neither touches `server/http.ts` / `threadServer.js`. Keeping this draft until those clear.

PR #250, opened 2026-03-18
Don't trigger source refresh for TTL-expired entries when doing `search()`

cb1kenobi:
When `search()` iterates over results and encounters a TTL-expired cache entry, `transformEntryForSelect()` would call `ensureLoadedFromSource()` and await the result before returning the record.

The fix separates the three cases that were conflated in a single condition:

- `onlyIfCached` + expired → return the "expired" indicator (unchanged)
- `INVALIDATED` or `EVICTED` → block and reload from source (unchanged)
- TTL-expired during `search()` → return stale cached data without touching the source; `get()` is the correct place to enforce TTL freshness

Note: I have no idea what I'm doing, but this seems to fix the flakey caching tests.

PR #880, opened 2026-05-30
chore(deps): remove @datadog/pprof from core

kriszyp:
## Summary
- `@datadog/pprof` is only consumed by `analytics/profile.ts`, which lives in **harper-pro**, not in core. Core has no direct `import`/`require` of it (only stale comments in test utils that guard against its native-build errors).
- Removes `@datadog/pprof` from core's direct `dependencies` in `package.json` and the top-level lockfile entry.
- It remains in `package-lock.json` only as a transitive **dev/peer** dependency of the published `harper` package, which is expected.

## Why
- pprof is being managed as a harper-pro-only dependency. Keeping it in core caused the `core:sync` job in harper-pro to overwrite/downgrade harper-pro's pinned pprof version on every sync.

## Companion change
- Paired with a harper-pro PR (#250) that updates `build-tools/sync-core.sh` to preserve harper-pro's pinned `@datadog/pprof` version across the core dependency sync.

## Test plan
- [ ] CI green (core builds and tests without the direct pprof dependency).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,6 @@
 			"dependencies": {
 				"@aws-sdk/client-s3": "^3.1012.0",
 				"@aws-sdk/lib-storage": "3.1045.0",
-				"@datadog/pprof": "^5.11.1",
 				"@endo/static-module-record": "^1.1.2",
 				"@fastify/autoload": "^6.3.1",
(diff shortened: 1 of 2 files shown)


kriszyp:
Syncing core on pro is kinda broken due to conflicting datadog versions, so fixing that.

PR #881, opened 2026-05-30
Add YCSB-style CRUD load test

kriszyp:
## Summary

Adds a YCSB-style CRUD load test under `benchmarks/ycsb/` that drives CRUD over the **HTTP REST interface** against a real Harper instance (booted via `@harperfast/integration-testing`), runs the standard A–F workloads, and reports throughput + latency percentiles as JSON for nightly tracking.

- `workload.mts` — uniform/zipfian/latest distributions (ports of the reference YCSB generators), A–F workloads, an acknowledged-counter so inserted keys only become readable once their write completes, and a closed-loop concurrent runner. Unit-tested (`workload.test.mts`, 10 tests).
- `restClient.mts` — `node:http` keep-alive transport (no new deps), round-robins across base URLs (cluster-ready).
- `harness.mts` / `run-single-node.mts` — CLI, load + run phases, results assembly; `--profile` runs the HTTP server on the main thread + `--cpu-prof`.
- `analyze-profile.mts` — summarizes the V8 profile by module / hot function.
- `.github/workflows/ycsb-nightly.yml` — nightly cron + manual dispatch, uploads results JSON as an artifact.

## Purpose

We had no CRUD-style load test for Harper. This gives us a repeatable throughput/latency baseline to track over time and a profiling entry point for finding perf opportunities. Files live under `benchmarks/` and are **not** part of the tsc build (run directly via Node type-stripping), so this is additive tooling — no runtime code is touched.

## Where to focus

- **REST scan syntax** (`restClient.mts`): scans use `GET /usertable/?id>=<key>&limit(<n>)`. Verified against `resources/search.ts`, but worth a sanity check.
- **Framework-internal dependency** (`run-single-node.mts`): the benchmark relies on `HARPER_SET_CONFIG` taking precedence over the framework's hard-coded `--THREADS_COUNT=1` / `--LOGGING_LEVEL=debug` CLI args (so `threads.count=4` and quiet logging win). Verified + commented; flagging since it's load-bearing.
- **Deliberate fidelity tradeoffs** (documented in the README, not bugs): the `latest` distribution is approximate (scrambled-zipfian mod keyspace), latency percentiles are nearest-rank (not HdrHistogram), and throughput is successful-ops/sec. Fine for trend tracking; calling out in case you'd want stricter fidelity before comparing to reference YCSB.

A cross-model review (Gemini perf/maintainability lens + a Harper-domain pass) found no blockers and no data-integrity/concurrency/security issues; actionable findings were addressed in this commit.

### DIFF
--- a/.github/workflows/ycsb-nightly.yml
+++ b/.github/workflows/ycsb-nightly.yml
@@ -1,0 +1,69 @@
+name: YCSB Nightly Load Test
+
+# Runs the YCSB-style CRUD load test against a single local Harper instance and
+# uploads the JSON results as an artifact for tracking throughput/latency over
+# time. Default-runner perf is modest, so treat absolute numbers as a trend
+# signal on a fixed runner, not a hardware-representative benchmark.
+
+on:
+  schedule:
+    - cron: '0 7 * * *' # 07:00 UTC nightly
+  workflow_dispatch:
+    inputs:
+      scale:
+        description: 'Workload scale'
+        required: true
+        type: choice
+        default: 'standard'
+        options:
+          - 'quick'
+          - 'standard'
+          - 'heavy'
+      workloads:
+        description: 'Comma-separated workloads (e.g. C,B,A,F,D,E)'
+        required: false
+        type: string
+        default: 'C,B,A,F,D,E'
+
+permissions:
+  contents: read
+
+jobs:
+  ycsb:
+    name: YCSB single-node (Node.js v24)
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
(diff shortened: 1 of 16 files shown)


kriszyp:
Thanks for the review — all findings are addressed in `0213dbe6f`:

- **Shell injection on the `workloads` dispatch input** → inputs are now passed via `env:` rather than `${{ }}` interpolation into the shell.
- **Insert failure stalling the ack frontier / leaking `pendingAcks`** → `acknowledgeInsert` now runs in a `finally`.
- **No request timeout** → per-request timeout (default 30s) so a hung socket can't block a closed-loop worker.
- **Startup probe could hang past its deadline** → `AbortSignal.timeout` on the probe.
- **Zero-keyspace crash** → clear guard error in `existingKey()`.
- **Implicit engine default** → `HARPER_STORAGE_ENGINE` is now always set.

Resolving the threads. Appreciate the thorough pass!

— Claude (Opus 4.7)

Issue #641, opened 2026-05-20
Deployment tracking, replicated payload delivery, and rollback


kriszyp:
# Deployment tracking, replicated payload delivery, and rollback

This plan becomes the body of a new issue in `HarperFast/harper`. Once approved, we'll pause the four in-flight PRs (#530, #531, #536, harper-pro #146) and rebuild the deploy_component story around a replicated `hdb_deployment` system table that doubles as the payload-delivery channel.

---

## Context

The original deploy_component effort split into four PRs solving four problems:

| In-flight PR | What it does |
|---|---|
| #530 | Multipart-streaming upload from CLI → origin (no 2 GB Buffer cap) |
| #531 | SSE-based progress events from server back to CLI |
| #536 | Staging the upload to a temp file on origin so peers can also receive it |
| harper-pro #146 | Direct-HTTPS peer relay because the WS operation transport has a frame cap |

The user's two insights collapse this story into a single coherent design:

1. **The deploy_component handler should write a persistent record from the very first byte** — so Studio (or any client) can observe and audit deploys independent of the CLI, errors are queryable after the fact, and rollback has a target to roll back to.
2. **If we're already storing the payload as a blob attribute on that record, we shouldn't need a separate staging file or a separate peer-relay channel** — Harper's replication ([`replication/replicationConnection.ts`](replication/replicationConnection.ts)) already supports streamed, chunked, back-pressured blob transfer (the `BLOB_CHUNK = 146` message type, sender at lines 1840–1896, receiver at lines 788–858). The 100 MB cap is **per chunk**, not per blob, so multi-GB payloads work today. #146's direct-HTTPS relay was solving a problem Harper had already solved elsewhere.

So this issue isn't "tracking on top of the deploy stack" — it's the new spine, and the staging/relay PRs go away.

---

### 1. New replicated system table: `hdb_deployment`

| Attribute | Type | Notes |
|---|---|---|
| `deployment_id` | string (UUID v4) | Hash attribute. Returned synchronously to the caller. |
| `project` | string | Component name. Indexed. |
| `package_identifier` | string \| null | Result of `derivePackageIdentifier()` ([Application.ts:446](components/Application.ts:446)); null for raw-payload deploys. |
| `payload_hash` | string \| null | SHA-256 of the tarball, computed once on origin during ingest. |
| `payload_size` | number \| null | Uncompressed bytes (precomputed by `getPackagedDirectorySize`). |
[...]

PR #884, opened 2026-05-30
fix(RecordEncoder): retain blob files still referenced by the new record on update

kriszyp:
## Summary

- `RecordEncoder` historically called `deleteBlobsInObject(existingEntry.value)` unconditionally on every update of a row that previously carried blobs.
- `deleteBlob` schedules `unlink()` ~`deletionDelay` ms later (default 500ms). If the new record carries the *same* `FileBackedBlob` (same `fileId`), the file gets unlinked after the put and the new row points at a missing path.
- Fix: `deleteBlobsInObject` now accepts an optional `retainedFileIds: Set<string>`; the encoder collects the new record's blob fileIds and threads them through. Blobs in the retained set are skipped.

## Why this matters — staging repro

[#641 deployment tracking](https://github.com/HarperFast/harper/issues/641): `DeploymentRecorder` issues many sequential puts on the same `hdb_deployment` row while keeping `payload_blob` attached (initial create → `ingestPayload` → phase transitions → per-peer `recordPeer` → `finish`). For payloads above `FILE_STORAGE_THRESHOLD` (8 KiB), the very first follow-up put schedules an `unlink` on the just-written file. Replication's `sendBlobs` then opens the file, gets ENOENT, and forwards an error chunk to the peer. The peer surfaces `Error in blob: ENOENT ... for record <id> from <origin>` and its `deploy_component` fails.

Reproduced live on the symphony-https 2-node staging cluster with a 171 MiB component tarball — both nodes had empty `blobs/system/0/0/` directories despite the row claiming `payload_blob_present: true`.

The bug is not deploy-specific. Any flow that updates a row carrying a file-backed blob without rewriting the blob attribute hits this — `description` field change next to a stored attachment, periodic timestamp bumps on a record with an embedded image, etc.

## Test plan

Added `unitTests/resources/blob.test.js` → `'updating an unrelated attribute does not unlink a still-referenced blob'`. The test:

- [x] Creates a record with a `>FILE_STORAGE_THRESHOLD` file-backed blob.
- [x] Performs the multi-update pattern (phase = pending → extracting → installing → loading → replicating → success), asserting the file survives each one.
- [x] Round-trips via `get` and `put`s the freshly-decoded blob, asserting retention across decoder instances (not just same JS object).
- [x] Explicitly nulls the blob and asserts the file *is* cleaned up — proves the retention set doesn't leak files when the new record genuinely drops the blob.

Without the fix, the test fails at the first follow-up put with `blob file must survive update that retains it` — exactly the production failure mode.

All 23 tests in `unitTests/resources/blob.test.js` pass. Lint + format clean. Cross-model review (codex) found no blocking issues; only feedback was the round-trip-via-decoded-blob test which is included.

## Performance

The extra cost is one in-memory recursive scan of the new record's tree, gated on `existingEntry.metadataFlags & HAS_BLOBS` (so untouched for the typical no-blobs path). The pre-existing scan of the old record is unchanged. No new I/O.

🤖 Generated by Claude

### DIFF
--- a/resources/RecordEncoder.ts
+++ b/resources/RecordEncoder.ts
@@ -18,7 +18,14 @@ import {
 } from './auditStore.ts';
 import * as harperLogger from '../utility/logging/harper_logger.ts';
 import './blob.ts';
-import { blobsWereEncoded, decodeFromDatabase, deleteBlobsInObject, encodeBlobsWithFilePath } from './blob.ts';
+import {
+	blobsWereEncoded,
+	decodeFromDatabase,
+	deleteBlobsInObject,
+	encodeBlobsWithFilePath,
+	findBlobsInObject,
+	getFileId,
+} from './blob.ts';
 import { getThisNodeId } from './nodeIdMapping.ts';
 import { recordAction } from './analytics/write.ts';
 import { RocksDatabase } from '@harperfast/rocksdb-js';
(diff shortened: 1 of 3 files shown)


kriszyp:
Heads-up on CI: the 3 failing checks (Integration Tests 2/4 on v22/v24/v26) are all the same pre-existing `upgrade and migrate LMDB to RocksDB` failure that's been failing on `main` for at least the last several runs (`97504f9e`, `8062f062`, `04c388f4`, …) with `AssertionError: {"error":"Must login"}`. Not caused by this PR — it's the 4.x→5.x upgrade harness, unrelated to RecordEncoder.

🤖 Generated by Claude

kriszyp:
Confirmed the 3 shard-2 failures are pre-existing on `main` and unrelated to this PR:

- All 3 jobs (v22/v24/v26 shard 2) fail exactly **one** test: `upgrade and migrate LMDB to RocksDB` (`fail 1` per shard). Other shards' tests all pass.
- The failure is an auth regression in the 4.x→5.x migration path: the first `sendOperation` after the upgrade returns `{"error":"Must login"}` (401), not anything in the blob/encoder path.
- `main` has been failing this same test continuously since `97504f9e` (right after the deploy-tracking-b2 merge into main). Bisecting back, **the last green Integration Tests run on `main` was `11d19b89` on 2026-05-29 16:14 UTC**. Likely-suspect window between `11d19b89..67d953fc`.
- The in-progress main CI run on the latest commit `68b576a3` is failing the same way (v22/v24/v26 shard 2 fail, Bun + Windows shard 2 pass), confirming the failure isn't introduced by anything on this branch.

Ran the encoder PR's unit test suite (`unitTests/resources/blob.test.js`) — all 23 tests pass including the new regression test. Lint + format clean.

🤖 Generated by Claude

Issue #914, opened 2026-06-01
Evaluate uNetworking.js (uWebSockets) for HTTP/WebSocket layer


kriszyp:
Evaluate uWebSockets.js as a replacement or supplement to Fastify/Node's built-in HTTP server for improved network throughput, especially for WebSocket-heavy workloads.

🤖 Filed by Claude on behalf of Kris.

PR #1096, opened 2026-06-02
feat(#914): uWebSockets.js HTTP/WebSocket backend (default-off)

kriszyp:
## uWebSockets.js as a per-worker HTTP/WebSocket backend (#914)

Harper's per-worker HTTP layer runs on Node's `http`/`http2` servers. For WebSocket-heavy and high-request-rate workloads that layer is the bottleneck — per-request and event-loop overhead dominates — and Bun, which I benchmarked earlier, didn't improve on it. This adds uWebSockets.js as a third per-worker backend alongside the Node and Bun paths, driving Harper's existing `httpChain`/`websocketChains` unchanged.

This started as a spike (#914) to test whether uWS was worth it. The thesis held — ≈1.9× less CPU and 1.66× more throughput on the real pipeline — and the backend has since been filled out to pass the entire integration suite under a gating CI job. So this is now a proposal to land it behind default-off flags, not a spike.

Two independent entry points, both **default off → no behavior change** unless enabled:

- **`HARPER_UWS_UDS`** — serves the plaintext UDS-behind-symphony mirror with uWS. The production target.
- **`HARPER_UWS_HTTP`** — backs a non-secure HTTP TCP port directly (`app.listen`, SO_REUSEPORT). A test/benchmark vehicle that lets the whole suite run over uWS without symphony in front.

### Why the UDS-behind-symphony shape

uWS has hard limits that would otherwise be blockers: no HTTP/2, no peer-certificate read API, and file-only TLS. Most Harper deployments already run behind **symphony**, which terminates TLS/mTLS/HTTP-2 and forwards plaintext HTTP/1.1 over a Unix domain socket with `X-Forwarded-For`. Running uWS there as a non-SSL `App().listen_unix()` sidesteps every one of those limits, because symphony owns TLS and client identity. Harper already exposes a pluggable per-worker UDS seam — both the Node and Bun paths drive it — so uWS slots in as a third implementation rather than a rewrite.

The direct-TCP path (`HARPER_UWS_HTTP`) exists so the integration suite and benchmarks can exercise uWS end-to-end without standing up symphony. It reads the real client IP from the TCP peer and only trusts `X-Forwarded-For` on the UDS path, so a direct client can't spoof loopback to satisfy local auth.

### Capabilities (validated through the real `startHarper` integration harness)

- **HTTP request/response** — REST/SQL, all data types, `@sealed`, binary/Brotli blob responses byte-exact, `Content-Encoding` passthrough, caching/ETags.
- **GraphQL** — POST bodies deserialize through the adapter's body stream (the Node-only `_nodeRequest` assumption is gone), so `/graphql` works on uWS and Bun.
- **Streaming responses** — Node streams / async-iterables stream incrementally with real backpressure (`res.write` + `res.onWritable`), chunked. `text/event-stream` (SSE) flushes headers up front, since uWS only flushes on the first write.
- **WebSocket upgrades** — native `app.ws()` bridged to a `ws`-shaped adapter; HTTP and WS coexist on one port. (MQTT-over-WS: JWT auth, topic ACLs, pub/sub, `$SYS`.)
- **Static files** — the `send`-based static plugin works: its `SendStream` is driven through a `ServerResponse` shim that captures the headers it writes and buffers the file.
- **Legacy Fastify routes** — when the chain doesn't handle a request, it delegates to the registered Fastify instance via `inject()`, so custom-function routes run behind uWS (parity with the Bun path).
- **Multiple `Set-Cookie`** — emitted as separate headers via `getSetCookie()` rather than comma-joined. uWS handles this correctly; Bun skips it.

### Validation

Full `test:integration:all` passes on all 6 shards under `HARPER_UWS_HTTP`, wired as a gating CI job alongside the Node and Bun variants — so the uWS path is held to the same bar as the others going forward. The new correctness work (client IP, GraphQL body, Fastify delegation, static, `Set-Cookie`) has unit and integration coverage; the earlier request/streaming/WebSocket work was cross-model reviewed (Codex + Gemini + Harper-domain).

### Benchmark (real `httpChain`, admin-authed `/`)

| path | CPU µs/req | req/s |
|---|---|---|
| Node UDS | 14.47 | 76,512 |
| **uWS UDS** | **7.57** | **127,152** |

≈ **1.9× less CPU / 1.66× more throughput**.

### Rollout

- **Ship default-off.** Both flags stay off in this PR; enabling uWS is opt-in, so merging changes nothing for existing deployments.
- **Prove it behind symphony.** The intended production path is `HARPER_UWS_UDS`. The next step is enabling it on a staging cluster behind symphony and watching real traffic before any default flip — a default change, if we make one, belongs in a later release, not here.
- **Keep `HARPER_UWS_HTTP` as a test/bench vehicle**, now guarded by CI. It's usable as a plaintext backend behind any terminating proxy, but it isn't the graduation target.
- **No hard dependency.** uWS is an optional dependency; when it's absent the uWS tests skip and the flags are inert, so nothing depends on it being present yet.

### Tradeoffs and caveats

- **Distribution tax.** uWS is a raw V8 addon, ABI-locked per Node major, with no musl/Alpine build — CI has to build it, and shipping it means owning that matrix. This is the main cost of graduating, and worth weighing explicitly.
- **Custom upgrade-middleware** (`onUpgrade` listeners that reject or modify a WS upgrade) isn't mirrored yet; default upgrade plus post-connection auth is.
- **Static responses are buffered**, not streamed, on this path — the `send` shim collects the file before writing. Fine for typical assets; large static files would want a streaming shim later.
- HTTP/2, peer certificates, and file-based TLS remain uWS non-features, which is exactly why the production shape puts symphony in front.

uWS is a drop-in third backend behind the existing seam, off by default, and now held to the full integration suite. The open question isn't whether it works but whether the distribution cost is worth the throughput on WebSocket-heavy workloads — which is where I'd like reviewer input.

🤖 Iterated with Claude (Opus) for @kriszyp


kriszyp:
## Live boot+bench results (full `httpChain` pipeline)

Booted real Harper from this branch (`@harperfast/integration-testing` `startHarper`, `threads.count=1`, `tls.unixDomainSockets=true`) and benchmarked the per-worker HTTP UDS socket through the real auth + routing pipeline. CPU/req measured from `/proc/<harper>/stat` (utime+stime) over an 8s window, 50 conns, admin-authed requests to `/` (→ 404 after auth+route-miss).

| path | CPU µs/req | req/s | p99 |
|---|---|---|---|
| Node UDS | 14.47 | 76,512 | 1ms |
| **uWS UDS** | **7.57** | **127,152** | **0ms** |

**uWS ≈ 1.91× less CPU and ≈ 1.66× more throughput** on the real pipeline. (This path exercises auth + middleware + route resolution but not a 200 resource read/serialize — a real 200 adds transport-invariant cost to both, so expect the steady-state win in the ~1.5–1.9× range.)

The live boot also surfaced a latent bug (now fixed in `470ba432`): `unhandled()` set `request._nodeRequest.user` for authenticated-but-unmatched requests, which throws for any adapter with a null `_nodeRequest` — i.e. **Bun today**, not just uWS. Null-guarded.

— Claude (Opus), for @kriszyp

kriszyp:
## Validation round: integration sweep + cross-model review (streaming/WS)

### Integration sweep under `HARPER_UWS_HTTP` — 43/43 pass
Ran a broad slice (database, server, security, components, mcp, mqtt) through uWS. Surfaced one real bug: **three component-deploy tests (early-hints, redirector, risk-query) failed** where they pass on baseline — the uWS HTTP path never populated `request.ip`, so `security/auth.ts` local-auth (`request.ip.includes('127.0.0.')`) failed and anonymous localhost requests got `AccessViolation: Must login`. Fixed (capture the TCP peer address); all 43 pass now.

### Cross-model review (Codex + Gemini + Harper-domain), adjudicated
No surviving blockers. Fixed:
- **[Codex P1 + sweep] Client IP / local-auth** — captured `res.getRemoteAddressAsText()` for the TCP path, and flipped `UwsRequest.ip` to prefer the real socket address over `X-Forwarded-For` so a direct client can't spoof `XFF: 127.0.0.1` to satisfy local auth (XFF trusted only on the symphony-UDS path).
- **[Codex P2] HEAD body** — uWS has no ServerResponse HEAD guard; `writeResponse` now suppresses the body for HEAD (REST already nulls it; this covers other handlers).
- **[Codex P2] WebSocket maxPayload** — the onWebSocket uWS branch now forwards `options.maxPayload` (was defaulting to 100 MiB).

Dropped / deferred:
- **[Gemini blocker] `Buffer.from(new Uint8Array(message))` "aliases uWS memory"** — false positive; it *copies* (proven: survives source neutering; corroborated by MQTT-over-WS 11/11 with async frame processing).
- **[Codex P2] GraphQL POST `_nodeRequest`** — pre-existing parity gap: `_nodeRequest` is null on Bun too; needs a body-based deserialize for non-Node adapters (tracked separately, not a uWS regression).
- **[Codex P2] raw Fastify on a uWS-backed port** — possible `SERVERS` collision; low reachability (MCP Fastify passes). Noted as a `HARPER_UWS_HTTP` (test-vehicle) limitation.

Adapter unit suite: 16 tests (adds `request.ip` and HEAD-suppression).

kriszyp:
## Cleared the two deferred Codex P2s (GraphQL POST body + Fastify/uWS port collision)

Following up on the earlier review round, both deferred P2s are now fixed on the branch.

### 1. GraphQL POST body deserialize (`b9e94a11a`)
`graphqlQuerying.ts` read the POST body from `request._nodeRequest` (the raw Node `IncomingMessage`), which is `null` on both the Bun and uWS request adapters → GraphQL POST 500'd off the Node path (`Cannot read properties of null (reading 'on')` in `streamToBuffer`). `REST.ts` already deserializes through `request.body`; GraphQL was the lone outlier. Fixed by reading `request.body` — a Readable-compatible body stream on every adapter — and dropping the stale `@ts-expect-error`.

- **Verified:** `graphql.test.mjs` **24/24 on both the Node and `HARPER_UWS_HTTP` paths** (was 0/24 → all 500 under the flag before the fix, which confirmed the bug was real).

### 2. Raw-Fastify listener colliding with the uWS HTTP port (`7bbe88dd5`)
Under `HARPER_UWS_HTTP` the app port is backed by uWS and `getHTTPServer` early-returns a `{ uws: true }` marker *before* it would have called `registerServer(server, port)`, so `SERVERS[port]` stays empty. If a legacy Fastify-routes app is then deployed, `fastifyRoutes` registers its raw `http.Server` via `server.http(fastify.server)`; with the port looking unused, `registerServer` set `SERVERS[port] = fastifyServer` and threadServer bound a Node http server competing with uWS on the same TCP port.

Fixed by mirroring the Bun path: non-function listeners on a uWS-backed port are diverted into the fallback map instead of `registerServer()`, so nothing lands in `SERVERS` to double-bind. Renamed `bunFallbackServers → fallbackServers` since the map is now shared by both non-Node backends.

- **Verified:** `components.test.mjs` (deploys a Fastify-routes component) **25/25 on both paths**, with the Fastify registration diverting cleanly and no bind collision.
- **Known limitation (noted for a parity follow-up):** request-time delegation to the Fastify fallback is not yet wired on the uWS handler (the `bunDelegateToNodeServer` equivalent), so raw-Fastify routes are unreachable under this flag — a clean not-found rather than a competing bind. Low reachability (requires `HARPER_UWS_HTTP` + a legacy Fastify-routes app + HTTP traffic to that route), acceptable for the current bench/test vehicle.

Still draft. Net: uWS is now validated as a full HTTP + streaming/SSE + WebSocket + GraphQL backend, with the two remaining review P2s closed.

kriszyp:
## Fastify fallback now works behind uWS + full integration suite promoted to CI

Two follow-ups from the previous round.

### 1. Fastify fallback delegation on the uWS HTTP path (`b0ec0b0c4`)
The prior commit stopped a raw `server.http(fastify.server)` from binding a competing Node server on a uWS port, but left the route unreachable (clean 404). That's now closed: when the request chain doesn't handle a request (`status === -1`) and a Fastify instance is registered for the port, the uWS handler delegates via `fastify.inject()` — Fastify's internal router, no socket — mirroring the Bun path, including SSE streaming and the `AUTHORIZE_LOCAL` pre-auth user forward.

- Extracted the shared inject core into `injectToFastify()` and routed **both** the Bun and uWS delegation paths through it.
- `fastifyRoutes` now registers its **app instance** for the http port(s). It previously only registered the `http.Server`, so *neither* Bun nor uWS could delegate to legacy routes (a latent Bun gap too). Renamed `bunFastifyInstances → fastifyInstances` / `registerBunFastifyInstance → registerFastifyInstance` since the map is no longer Bun-only.
- `UwsRequest` exposes `rawBody` for the inject payload.

**Verified:** `fastifyRoutes-test.mjs` (`GET /testApp/ping → 'pong'` + a REST resource on the same component) passes on **both** the Node and `HARPER_UWS_HTTP` paths — under the flag the Fastify route is served purely via inject-delegation. `graphql` 24/24, `components` 25/25, `mcp/sse-listchanged` 4/4 under the flag; 16 uWS unit tests green.

### 2. Full integration suite under `HARPER_UWS_HTTP` in CI (`4f8b10b2c`)
Added a `run-integration-tests-uws` job to `integration-tests.yml`, mirroring the existing Bun variant: the same 6-shard `test:integration:all` on Node 24, but with `HARPER_UWS_HTTP=1`. Only non-secure/non-ops app ports run through uWS; secure/replication/ops paths stay on Node. This replaces the manual local flag with continuous full-suite coverage of the uWS request/streaming/WS/GraphQL/Fastify-fallback path.

Still draft. First full-suite-under-uWS CI run is in flight — will report the shard results.

kriszyp:
## Full integration suite now green under `HARPER_UWS_HTTP` — job promoted to a gating check

Following the Fastify-delegation work and the CI promotion, the first full-suite run under `HARPER_UWS_HTTP` surfaced two pre-existing uWS-path gaps (Bun and Node were green on the same tests). Both are now fixed and the full 6-shard `test:integration:all` passes under uWS ([CI run 28724670219](https://github.com/HarperFast/harper/actions/runs/28724670219)), so the job is no longer `continue-on-error` — it gates alongside the Node and Bun variants.

### Fix 1 — static-file serving via `send` (`a8c84881c`)
The static plugin returns a `send` `SendStream`, which only runs when piped to a Node `ServerResponse` and writes its own headers there. The uWS path treated it as a plain Readable (`.on('data')` never starts a `SendStream`), and uWS only flushes status/headers on the first body write — so static responses hung and the client saw a `HeadersTimeout`. That's why **every deploy+access test hung ~300s** (deployed apps serve a static site).

Fix: pipe the `SendStream` into a `Writable` shim that captures the headers it writes and buffers the file (`bufferSendStream`, mirroring the Bun `fetchHandler` shim incl. `finished: false`). Gated on `handlesHeaders` (only the static plugin sets it), so real streaming/SSE bodies keep streaming.

### Fix 2 — multiple Set-Cookie headers (`fee336f0d`)
A WHATWG `Headers` comma-joins `Set-Cookie` on iteration, merging multiple cookies into one and corrupting values with commas (`expires=` dates). `writeHeaders` iterated directly, and `writeResponse` round-tripped a WHATWG `Headers` through `new Headers()` (joining before serialization). Fix: emit `Set-Cookie` individually via `getSetCookie()`, stop the lossy round-trip, and keep the delegation path's set-cookie array multi-valued. uWS now handles multi-`Set-Cookie` **correctly** — the same case Harper-on-Bun documents and skips.

### Result
- uWS: **6/6 shards green** (was 2/6 — shards 1/3/4/6 were exactly these two gaps).
- Regression sweep under the flag: graphql 24, components 25, mcp/sse 4, deploy-from-source 4, fastifyRoutes 2 (Node+uWS), 16 uWS unit tests.

Net across this PR: uWS is now a full HTTP + streaming/SSE + WebSocket + GraphQL + Fastify-fallback + static backend that passes the entire integration suite as a gating CI check. Still draft, pending the graduate-or-park call.

kriszyp:
## Cross-model review (Codex + Gemini + Harper-domain) of the new uWS work

Ran a thorough cross-model pass on the work since the last review (`561c750f1..HEAD`: Fastify delegation, `send`-based static, Set-Cookie, GraphQL body). Summary for reviewers so nothing gets rediscovered.

### No blockers
Both outside-model legs led with the *same two* "blockers", both **refuted**:
- **`request.headers.asObject` undefined → Fastify auth bypass** — false. `UwsRequest.headers` is a `RequestHeaders` with a real `.asObject` (the accessor `REST.ts` uses on every request); all uWS REST tests pass, which is impossible if it were undefined.
- **Set-Cookie comma-coerced by `.set()`** — false. The uWS response uses Harper's Map-based `Headers`, which preserves a `Set-Cookie` array; `writeHeaders` emits each cookie separately.

Both stemmed from assuming WHATWG-`Headers` semantics without the type in view (Gemini is diff-only; the Codex CLI leg stalled and fell back to a degraded analysis — see caveat).

Security probe was clean: the `INTERNAL_USER_HEADER` pre-auth forward strips any client-supplied copy before re-adding the chain-resolved user; SSE-delegation teardown and the GraphQL body cap are handled.

### Addressed (`2ac0b5d6a`)
- **`bufferSendStream` no longer swallows `send`'s status** — captures `statusCode`/`writeHead` so a 304 (conditional GET) or 206/416 (Range) is honored, not flattened to 200. *Clarification:* end-to-end 304/Range is still gated upstream — `send` can't read `If-None-Match`/`Range` from Harper's `RequestHeaders`, so it never emits a non-200 status on **any** backend today (verified: Node baseline also returns 200). That's a pre-existing cross-path limitation, filed as a separate follow-up; this change just removes the shim's status-swallow so it'll work once the upstream wiring is fixed.
- Skip re-copying already-`Buffer` chunks when draining a delegated response.
- Document the lowercased-`authorization` contract in `injectToFastify`.
- Refresh the fallback-divert comment (delegation is now wired; the `{ uws: true }` marker is guaranteed set by the `getServer(port)` call in the same loop iteration).

### Follow-ups (not blocking)
- **Static bodies are buffered, not streamed** (the `send` shim collects the file). Fine for typical assets; large files want a streaming shim. Matches the Bun path; already noted in the PR description.
- **True 304/Range for static** needs `send` fed a plain-headers request — a cross-path improvement (helps Node too), out of scope here.
- **Bun's SendStream shim has the same status-swallow** as the one just fixed on uWS — worth mirroring the fix there.

### Caveats
- The **Codex CLI leg failed** (stdin stall) and degraded to a non-code-trace fallback; worth a clean Codex re-run before merge for code-trace depth on the `send`/delegation paths.

server/http.ts
```diff
@@ -603,6 +657,210 @@ function getHTTPServer(port: number, secure: boolean, options: ServerOptions) {
...
+		recordAction(
+			executionTime,
+			'duration',
+			request.handlerPath,
+			request.method,
+			response.wasCacheMiss == undefined ? undefined : response.wasCacheMiss ? 'cache-miss' : 'cache-hit'
+		);
+		recordActionBinary(status < 400, 'success', request.handlerPath, request.method);
+		recordActionBinary(1, 'response_' + status, request.handlerPath, request.method);
+		logBunRequest(request, status, requestId, executionTime);
```
kriszyp: Should we use a more generic name (not Bun) here?

server/serverHelpers/uwsServer.ts
```diff
@@ -0,0 +1,450 @@
...
+		// address is set). request.ip feeds local-auth (security/auth.ts AUTHORIZE_LOCAL), rate limiting,
+		// and logging.
+		const ip = port != null ? normalizeAddress(Buffer.from(res.getRemoteAddressAsText()).toString()) : undefined;
+
+		const ac = new AbortController();
+		res.onAborted(() => ac.abort());
+
+		const dispatch = (bodyBuffer?: Buffer) => {
+			const request = new UwsRequest({ method, url, headers, secure, bodyBuffer, signal: ac.signal, ip });
+			Promise.resolve(handler(request))
```
kriszyp: Can this use when() instead of Promise.resolve() to avoid an unnecessary promise on synchronous handler return? (or is that not possible?)

server/serverHelpers/uwsServer.ts
```diff
@@ -0,0 +1,450 @@
...
+				if (total > maxBodyBytes) {
+					if (!ac.signal.aborted) res.cork(() => res.writeStatus('413 Payload Too Large').end());
+					ac.abort();
+					return;
+				}
+				// uWS neuters/reuses the ArrayBuffer once this callback returns, and the body is read
+				// asynchronously in the handler — so copy the bytes out now. `Buffer.from(arrayBuffer)`
+				// would alias uWS's memory; wrapping in a Uint8Array first forces an owned copy.
+				const part = Buffer.from(new Uint8Array(chunk));
+				buf = buf ? Buffer.concat([buf, part]) : part;
```
kriszyp: Doesn't Buffer.concat copy the input? When we concat we are double copying the chunk then? (can we avoid that?)

server/serverHelpers/uwsServer.ts
```diff
@@ -0,0 +1,165 @@
...
+		app.listen_unix((listenSocket: unknown) => {
+			if (listenSocket) resolve();
+			else reject(new Error(`uWS failed to bind unix socket ${socketPath}`));
+		}, socketPath);
+	});
+
+	return {
+		app,
+		close: () => app.close(),
+	};
```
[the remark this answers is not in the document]
kriszyp: Verified at runtime: `typeof app.close === 'function'` in uWebSockets.js v20.68.0 — `TemplatedApp.close()` exists (it closes the app's listen sockets); it's simply absent from the shipped `index.d.ts`. So `close: () => app.close()` is correct as written. Resolving.

—KrAIs (Claude Opus 4.8)

server/serverHelpers/uwsServer.ts
```diff
@@ -0,0 +1,165 @@
...
+ * by symphony upstream, the real client IP arrives via X-Forwarded-For, so this server only
+ * speaks plaintext HTTP/1.1 and never touches certificates.
+ *
+ * NOTE: this is the adapter core only. Wiring it into getUwsHTTPServer()/threadServer.js
+ * (alongside the Node and Bun UDS paths) and adding `uWebSockets.js` as an optional
+ * dependency is the remaining productionization step. The request-construction and
+ * response-serialization paths here are exercised by the spike benchmark in ~/dev/tmp/uws-poc.
+ */
+import { UwsRequest } from './Request.ts';
+import { Headers } from './Headers.ts';
```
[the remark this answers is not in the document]
kriszyp: Done — `STATUS_CODES` from `node:http` is now imported and drives the status line (`statusText`), no hardcoded table. Resolving.

—KrAIs (Claude Opus 4.8)

server/serverHelpers/uwsServer.ts
```diff
@@ -0,0 +1,165 @@
...
+
+	const onRequest = (res: UwsResponse, req: UwsRequestRaw, hasBody: boolean) => {
+		// uWS HttpRequest is only valid synchronously inside this callback — copy everything now.
+		const method = req.getMethod().toUpperCase();
+		const query = req.getQuery();
+		const url = req.getUrl() + (query ? '?' + query : '');
+		const headers: Record<string, string | string[]> = {};
+		req.forEach((k: string, v: string) => {
+			headers[k] = v;
+		});
```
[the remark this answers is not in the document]
kriszyp: Already handled — the `req.forEach` accumulator collapses duplicate header lines into an array (the `existing`/`Array.isArray` branch), so repeats like Cookie/Forwarded are preserved. Resolving.

—KrAIs (Claude Opus 4.8)

server/http.ts
```diff
@@ -603,6 +657,197 @@ function getHTTPServer(port: number, secure: boolean, options: ServerOptions) {
...
+				});
+				const respHeaders = new Headers();
+				for (const [k, v] of Object.entries(injectResult.headers)) {
+					if (v == null) continue;
+					// Keep Set-Cookie multi-valued (Harper Headers + writeHeaders emit each separately);
+					// only comma-join other repeated headers.
+					if (Array.isArray(v)) respHeaders.set(k, k.toLowerCase() === 'set-cookie' ? v : v.join(', '));
+					else respHeaders.set(k, String(v));
+				}
+				logBunRequest(request, injectResult.statusCode, requestId, performance.now() - startTime);
```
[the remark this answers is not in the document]
kriszyp: This path builds Harper's array-preserving `Headers` (not WHATWG), and Set-Cookie is passed through as an array that `writeHeaders` emits one entry at a time — so no comma-join occurs here. Resolving.

—KrAIs (Claude Opus 4.8)

server/serverHelpers/uwsServer.ts
kriszyp: Done in 37c2cf418 — the dispatch now uses `when(handler(request), ...)` so a synchronous handler return stays synchronous (no extra promise/microtask).

—KrAIs (Claude Opus 4.8)

server/serverHelpers/uwsServer.ts
kriszyp: Good catch — resolved in 37c2cf418 by removing the accumulation entirely: the body is now streamed (each chunk pushed straight into a Readable) and streamToBuffer owns concat + the size limit, so there's no more per-chunk `Buffer.concat` double-copy.

—KrAIs (Claude Opus 4.8)

server/http.ts
kriszyp: Renamed to `logHttpRequest` in 37c2cf418 (it serves the shared Bun/uWS delegation path, not just Bun).

—KrAIs (Claude Opus 4.8)

server/http.ts
```diff
@@ -317,6 +322,31 @@ function getHTTPServer(port: number, secure: boolean, options: ServerOptions) {
 	if (!httpServers[port]) {
 		// TODO: These should all come from httpOptions or operationsApiOptions
 		const serverPrefix = isOperationsServer ? 'operationsApi_network' : (usageType ?? 'http');
+		// uWS plaintext-HTTP path (#914, HARPER_UWS_HTTP): back a non-secure TCP HTTP port with
+		// uWebSockets.js directly instead of a Node http server. This is the flag used to run the
+		// integration suite through uWS (no symphony/UDS needed). WebSocket upgrades are not yet
+		// wired on this path, so it's opt-in and separate from HARPER_UWS_UDS.
```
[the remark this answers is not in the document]
kriszyp: Corrected in 37c2cf418 — the comment now notes WS upgrades are wired on this path via onWebSocket's uWS branch.

—KrAIs (Claude Opus 4.8)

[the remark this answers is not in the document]
kriszyp: Corrected in 37c2cf418 (stale sentence removed/updated).

—KrAIs (Claude Opus 4.8)

server/http.ts
```diff
@@ -317,6 +333,31 @@ function getHTTPServer(port: number, secure: boolean, options: ServerOptions) {
...
```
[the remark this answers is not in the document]
kriszyp: Corrected in 37c2cf418 (stale sentence removed/updated).

—KrAIs (Claude Opus 4.8)

package.json
```diff
@@ -250,6 +250,7 @@
 	"optionalDependencies": {
 		"bufferutil": "^4.0.9",
 		"segfault-handler": "^1.3.0",
+		"uWebSockets.js": "github:uNetworking/uWebSockets.js#v20.68.0",
```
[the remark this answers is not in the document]
kriszyp: Added a `uWebSockets.js` entry to dependencies.md in 37c2cf418 (git dep, glibc-only, optional, needs git at install).

—KrAIs (Claude Opus 4.8)

kriszyp:
## YCSB benchmark: local main baseline vs this branch (uWS off/on)

Ran the nightly `benchmarks/ycsb/run-single-node.mts --scale=standard` suite (200k records, 500k ops/workload, concurrency 64, RocksDB engine, `threads.count=4`) three times, all on the **same local machine** (12th Gen Intel i7-12700H, 20 threads) so the comparison isolates code/config from hardware:

1. **main** @ `39b479219` — current main, Node HTTP backend
2. **this branch** @ `37c2cf418` — `HARPER_UWS_HTTP` unset (Node backend, control)
3. **this branch** @ `37c2cf418` — `HARPER_UWS_HTTP=1` (uWS backend)

I didn't compare directly against last night's nightly-CI numbers — this box loads ~3x faster than the shared GitHub runner (23.6k rec/s here vs 8.4k there), so that comparison would conflate hardware with code. The local main run above is the real control.

### Throughput (ops/sec, higher = better)

| phase | main | branch (uWS off) | branch (uWS on) | uWS vs main | uWS vs control |
|---|---:|---:|---:|---:|---:|
| load (insert) | 23,619 | 20,723 | **31,138** | **+31.8%** | **+50.3%** |
| C — read-only | 40,618 | 41,021 | 41,103 | +1.2% | +0.2% |
| B — read-mostly | 40,665 | 39,900 | 42,173 | +3.7% | +5.7% |
| A — update-heavy | 25,371 | 27,049 | **33,333** | **+31.4%** | **+23.2%** |
| F — read-modify-write | 18,517 | 20,468 | **23,816** | **+28.6%** | **+16.4%** |
| D — read-latest | 40,772 | 39,426 | 43,307 | +6.2% | +9.8% |
| E — short scans | 4,271 | 3,734 | 4,290 | +0.4% | +14.9% |

### p99 latency (ms, lower = better) — the notable movers

| op | main | branch (uWS off) | branch (uWS on) | uWS vs main |
|---|---:|---:|---:|---:|
| A read | 10.10 | 9.45 | **5.54** | **−45%** |
| A update | 10.94 | 10.90 | **7.76** | **−29%** |
| F read | 9.89 | 6.06 | **5.35** | **−46%** |
| F rmw | 19.96 | 15.05 | **10.82** | **−46%** |

### Takeaways

- **Control (uWS off) tracks main within ~5% on every workload** — the branch introduces no regression on the default Node backend, which is the right sanity check given the branch's base has drifted from current main.
- **Write/mutate-heavy workloads are where uWS earns its keep**: 23–50% more throughput and 29–46% lower p99 on load, A, and F. That's consistent with the PR description's own `httpChain` microbenchmark (≈1.9× less CPU / 1.66× more throughput) — muted here because these are full end-to-end REST ops that also spend time in the RocksDB write path, not pure HTTP-layer time.
- **Pure-read workloads (C, D) barely move** — RocksDB-read-bound, not HTTP-bound, so a faster HTTP layer doesn't have much to grab onto there.
- **E (scan) throughput is flat; its p99 was noisy on a single rep** (main 45ms → control 80ms → uWS 70ms) — range-scan latency has high run-to-run variance at reps=1, so I wouldn't read anything into that number without more reps.

Happy to rerun with more reps or a different scale if useful.

🤖 Generated by Claude Opus for @kriszyp

server/serverHelpers/uwsServer.ts
```diff
@@ -0,0 +1,477 @@
...
+				if (ended || body.destroyed) return;
+				ended = true;
+				body.destroy(error && body.listenerCount('error') > 0 ? error : undefined);
+			};
+			ac.signal.addEventListener('abort', () => tearDown(new Error('request aborted')), { once: true });
+			res.onData((chunk: ArrayBuffer, isLast: boolean) => {
+				if (ended || body.destroyed) return;
+				total += chunk.byteLength;
+				if (total > maxBodyBytes) {
+					if (!ac.signal.aborted) res.cork(() => res.writeStatus('413 Payload Too Large').end());
```
cb1kenobi: **Medium: 413 can be written to an already-completed uWS response**

The only guard on this 413 write is `!ac.signal.aborted`. But `ac` is only aborted on client disconnect or on this same over-limit branch — it is *not* set when the handler responds normally. If a handler returns before consuming the body (the throttle path returns 503 without reading the body; validation/405/early-cache responses do the same) while the client is still uploading, `writeResponse`/the error path will have already `res.end()`ed. uWS keeps firing `onData` to drain the socket, so once `total > maxBodyBytes` this calls `res.writeStatus(...).end()` on a finished response — an invalid uWS state that can crash the worker (a DoS on the intended `HARPER_UWS_UDS` production path). Track whether a response was already sent and gate the 413 on it:

Suggested fix:
```suggestion
				if (total > maxBodyBytes) {
					if (!ac.signal.aborted && !responded) res.cork(() => res.writeStatus('413 Payload Too Large').end());
```

(set a `responded` flag in `writeResponse` and the error path of `dispatch`; alternatively, `tearDown()` + short-circuit `onData` once the response is sent.)

—
_Generated by Barber AI_

unitTests/server/serverHelpers/uwsServer.test.js
```diff
@@ -0,0 +1,376 @@
...
+						res.on('data', (c) => chunks.push(c));
+						res.on('end', () => resolve(Buffer.concat(chunks).toString()));
+					})
+					.on('error', reject)
+					.end();
+			});
+			assert.strictEqual(body, 'http');
+		});
+
+		it('populates request.ip from the TCP peer address', async function () {
```
cb1kenobi, line 337: **Low (test-coverage): the XFF-spoofing defense is not tested**

This proves `request.ip` comes from the TCP peer, but not the security-critical invariant the PR relies on: that a direct-TCP client's `X-Forwarded-For` is *ignored* when a peer address is present (so it can't spoof `127.0.0.1` to satisfy local auth). Add a case that sends `X-Forwarded-For: 127.0.0.1` to `/ip` over TCP and asserts the response is still the real peer (`127.0.0.1` here, but the point is XFF is not consulted), plus a UDS case asserting `ip` *does* fall back to XFF when no peer address exists.

—
_Generated by Barber AI_

unitTests/server/serverHelpers/uwsServer.test.js
```diff
@@ -0,0 +1,376 @@
...
+				};
+			case '/bigstream': {
+				const chunk = Buffer.alloc(64 * 1024, 0x61); // 64 KiB of 'a'
+				return { status: 200, body: chunkStream(Array.from({ length: 64 }, () => chunk)) }; // 4 MiB total
+			}
+			case '/method':
+				return { status: 200, body: 'method=' + request.method };
+			case '/boom':
+				throw new Error('kaboom');
+			case '/teapot':
```
cb1kenobi, line 109: 🫖

server/serverHelpers/uwsServer.ts
kriszyp: Fixed in 94307344c. Added a \`responseCompleted\` flag in \`onRequest\` and now guard all three write sites (handler result, error, and the 413) on it in addition to \`ac.signal.aborted\`. The handler can respond (or start streaming) without draining the body, so a later over-limit chunk would otherwise write a 413 onto an already-completed response and abort the process. — KrAIs

unitTests/server/serverHelpers/uwsServer.test.js
kriszyp: Added in 94307344c. Two tests: (1) on the direct-TCP path a spoofed \`X-Forwarded-For: 1.2.3.4\` does NOT override the authoritative socket peer IP (\`request.ip\` stays \`127.0.0.1\`); (2) on the UDS path (no socket peer address) the trusted proxy's XFF is still honored. Both pass locally. — KrAIs

cb1kenobi (approved):
Human reviewed. Looks great!

PR #1098, opened 2026-06-02
feat(hnsw): dynamic/query-time ef, rerank, and no-reindex on search-only option change

kriszyp:
Follow-up to #894 (int8 quantization, merged). Makes int8 HNSW production-quality — recall compensation + exact distances + cheap ef retuning.

## Summary
1. **Dynamic + query-time search ef** — `search()` resolves ef as: per-query `ef` > explicitly-configured `efConstructionSearch` > an auto-scaled `ef(N)` fallback (capped `sqrt(N)`, base 100). A fixed ef makes recall decay as the graph grows; this compensates. The cap favors latency — apps wanting more set `efConstructionSearch` or a per-query `ef`.
2. **No reindex on search-only option changes** (`databases.ts`) — changing `efConstructionSearch` previously forced a full reindex (and briefly 503'd the index). The change-check is split into structure-affecting (drives reindex) vs metadata-only (drives persistence); a custom index declares its search-only keys via `static searchOnlyOptions`. In-progress backfill state is preserved on metadata-only updates. Validated live: an ef change now persists + applies to the live index with no rebuild.
3. **Rerank** (`search.ts`) — a quantized index navigates on approximate distances, so for a nearest-neighbor (sort) query the search layer recomputes each candidate's distance from the record's full-precision vector and re-sorts, restoring exact `$distance` and ordering while keeping the fast int8 traversal.

## Why
From a recall/latency-vs-N sweep, int8 recall ≈ float at every scale but a fixed ef makes recall decay with N (search ef is the lever, not build quality — an A/B showed higher efConstruction barely helped while halving ingest, and `optimizeRouting=0.5` *helps* recall, so both were left alone). Rerank removes the approximate-`$distance` downside.

## Where to look
- `databases.ts` structural-vs-metadata split + the in-progress-indexing-state preservation (subtle: a metadata-only change mid-backfill must not mark a partial index complete).
- `search.ts` rerank: int8 + sort gating, the materialize/re-sort, missing-vector handling.
- ef resolution precedence in `HierarchicalNavigableSmallWorld.search()`.

## Open items / follow-ups (deliberately not here)
- **Default int8 flip** — separate deliberate compat change (a test-flip passed all 15 tests, but understand why float-assuming `verifyIntegrity` didn't break + add explicit default-int8 coverage + docs first).
- **`lt`/`le` int8 threshold queries** still use the index's approximate distance — exact threshold filtering needs over-fetch (follow-up).
- **Query-time `ef`** is accepted by the index but not yet plumbed from a query's `sort.ef` through core search — auto-scale covers the dynamic default.
- **Auto-ef cap** is conservative; the 100K calibration point is missing (box contention crashed it) — tune later.
- **Float32Array input seam** lands with #747.

## Review notes
Cross-model review run: **Codex** found one P2 (a metadata-only change mid-backfill could persist a partial index as complete) — **fixed** here. The **Gemini (agy) leg timed out** (loaded box) so no second-model coverage this round, and the worktree-isolated `reviewer` subagent was unavailable — adjudicated in-session. A Gemini re-run on a quieter box would be worthwhile. 15 unit tests pass.

_Generated with the assistance of an LLM (Claude Opus 4.8)._

### DIFF
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -1124,17 +1124,35 @@ export function table<TableResourceType>(tableDefinition: TableDefinition): Tabl
 
 			// note that non-indexed attributes do not need a dbi
 			if (attributeDescriptor?.attribute && !attributeDescriptor.name) attributeDescriptor.indexed = true; // legacy descriptor
-			// Include `embed` so a source/model change refreshes the embed registry.
-			const changed =
+			// Some index options affect only search, not the stored structure (e.g. HNSW's
+			// efConstructionSearch). Changing those should persist the new metadata but NOT trigger a
+			// reindex. A custom index declares such keys via a static `searchOnlyOptions`.
+			const indexType = attribute.indexed && typeof attribute.indexed === 'object' ? attribute.indexed.type : undefined;
+			const searchOnlyOptions: string[] = (indexType && CUSTOM_INDEXES[indexType]?.searchOnlyOptions) || [];
+			const stripSearchOnly = (indexed: any): any => {
+				if (!indexed || typeof indexed !== 'object' || searchOnlyOptions.length === 0) return indexed;
+				const copy = { ...indexed };
+				for (const key of searchOnlyOptions) delete copy[key];
+				return copy;
+			};
+			const commonChanged =
 				!attributeDescriptor ||
 				attributeDescriptor.type !== attribute.type ||
-				JSON.stringify(attributeDescriptor.indexed) !== JSON.stringify(attribute.indexed) ||
 				attributeDescriptor.nullable !== attribute.nullable ||
 				attributeDescriptor.version !== attribute.version ||
 				attributeDescriptor.enumerable !== attribute.enumerable ||
 				JSON.stringify(attributeDescriptor.properties) !== JSON.stringify(attribute.properties) ||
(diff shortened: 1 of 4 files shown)


dawsontoth (approved):
Gemini code smells like legit to work through, otherwise seems ok to me.

kriszyp:
Addressed all four Gemini review nits in 476e5fc1e:

- **autoScaleEf** — optional-chain `getStats?.()?.entryCount` so a missing stats object can't throw.
- **exactDistance** — convert a typed-array `recordVector` to a plain `Array` (distance fns guard on `Array.isArray`).
- **databases.ts** — optional-chain `attributeDescriptor?.` in both index-change condition blocks (previously relied on `||` short-circuit).
- **search.ts rerank** — comparison-based sort so `Infinity` sentinels for missing vectors sort last without producing `NaN`.

CI is fully green. The earlier `Integration Tests 4/4 (Bun)` failure was a flake in unrelated `@embed` caching tests ("embed should fire exactly once" + a 27s fake-Ollama timeout with a "Database is closing!" teardown race) — Bun-only (Node v22/v24/v26 all passed that shard), and it passed clean on the re-run.

— Claude

PR #1099, opened 2026-06-02
fix: normalize cert.uses to array in createTLSSelector

kriszyp:
## Summary

- Normalizes `cert.uses` to an array in `createTLSSelector` before quality-scoring TLS certs
- Adds a regression test verifying no `TypeError` is logged when `cert.uses` is a non-array scalar

## Problem

`cert.uses` is documented as an array but can arrive as a non-array in legacy or manually-configured entries. The old code used `cert.uses?.includes(...)` and `cert.uses?.length` directly — if `cert.uses` is a scalar without `.includes` (e.g. a number), this throws `TypeError: cert.uses?.includes is not a function`, caught per-cert and logged as a warning, silently causing cert selection to skip that certificate.

A secondary issue: when `cert.uses` was a string like `"https"`, the quality deduction used `string.length` (5) instead of `1`, penalizing the cert 5× more than intended.

## Fix

One-liner normalization before the quality block (security/keys.ts:748):
```ts
const uses = Array.isArray(cert.uses) ? cert.uses : cert.uses ? [cert.uses] : [];
```

## Test

Regression test in `unitTests/security/keys.test.js`: puts a cert with `uses: 42` (a number, no `.includes`), stubs `logger.error`, calls `createTLSSelector('https').initialize(null)`, and asserts no TypeError was logged for that cert.

*Note to reviewer:* The `trace` method is intentionally omitted from the logger stub — `logger.trace?.` relies on the optional chain short-circuiting when `trace` is undefined to avoid evaluating `server.ports` when `server=null` (a test-only condition). This is not a bug in the fix; it's an existing minor issue in `keys.ts` that can be addressed separately.

Generated by Claude Sonnet 4.6 🤖

### DIFF
--- a/security/keys.ts
+++ b/security/keys.ts
@@ -744,11 +744,13 @@ export function createTLSSelector(type, mtlsOptions?): any {
 								continue;
 							}
 							let quality = cert.is_self_signed ? 1 : 3;
+							// normalize: stored as scalar in legacy/manual entries, expected array
+							const uses = Array.isArray(cert.uses) ? cert.uses : cert.uses ? [cert.uses] : [];
 							// prefer operations certificates for operations API
-							if (cert.uses?.includes(type)) quality += 3;
-							else if (cert.uses?.includes('https'))
+							if (uses.includes(type)) quality += 3;
+							else if (uses.includes('https'))
 								quality += 0.5; // this was a legacy generic general use type
-							else quality -= (cert.uses?.length ?? 0) / 5; // if there are designed uses for this that don't match, dock points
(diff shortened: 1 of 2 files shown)


kriszyp:
Good catch — replaced the `keys.__set__('logger', …)` rewire call with a plain integration test that uses the real database and just asserts `createTLSSelector` resolves without throwing. Pushed in commit b2bd290.

Note: the smoke test can't directly assert the inner TypeError is absent (that would require logger injection or refactoring the dependency), but it exercises the fixed path and covers the outer contract. The production-code change itself is unaffected. 🙏

— Claude Sonnet 4.6

PR #1109, opened 2026-06-03
feat: log application path and table init during startup

kriszyp:
## Summary

- Logs `Loading application from <path>` at info level when `RUN_HDB_APP` is set, so users following the getting-started guide can confirm their app directory was found
- Logs `Initialized table "<Name>" (primaryKey: <field>)` for each table defined via `@table` in a GraphQL schema
- Both logs are guarded by `getWorkerIndex() === 0` so they emit exactly once even with multiple worker threads
- Together these fill the confidence gap between `harper dev /path/to/app` and `All root applications loaded`, without requiring a `describe_all` round-trip

## Example output

```
[main/0] [info]: Loading application from /home/user/first-harper-app
[main/0] [info]: Initialized table "Dog", primaryKey: id
[main/0] [info]: All root applications loaded
```

## Notes

Cross-model review (Gemini) caught a missing thread guard in `componentLoader.ts` — the initial commit only had the guard in `graphql.ts`. Fixed in the second commit along with `harperLogger.info?.()` optional chaining.

## Test plan

- [ ] Run `harper dev /path/to/app` with a schema containing one or more `@table` types — confirm each message appears once
- [ ] Run with `schema: "mydb"` on a `@table` directive — confirm `, schema: mydb` appears in the table log
- [ ] Start without `RUN_HDB_APP` set — confirm no spurious log line
- [ ] Verify no duplicate lines when running with multiple worker threads

🤖 Generated with [Claude Code](https://claude.com/claude-code) — cross-model review via Gemini

### DIFF
--- a/components/componentLoader.ts
+++ b/components/componentLoader.ts
@@ -77,6 +77,7 @@ export function loadComponentDirectories(loadedPluginModules?: Map<any, any>, lo
 	}
 	const hdbAppFolder = process.env.RUN_HDB_APP;
 	if (hdbAppFolder) {
+		harperLogger.info('Loading application from ' + hdbAppFolder);
 		cfsLoaded.push(
 			loadComponent(hdbAppFolder, resources, hdbAppFolder, {
 				isRoot: false,
--- a/components/componentLoader.ts
+++ b/components/componentLoader.ts
@@ -77,7 +77,7 @@ export function loadComponentDirectories(loadedPluginModules?: Map<any, any>, lo
 	}
 	const hdbAppFolder = process.env.RUN_HDB_APP;
 	if (hdbAppFolder) {
-		harperLogger.info('Loading application from ' + hdbAppFolder);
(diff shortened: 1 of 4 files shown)


components/componentLoader.ts
```diff
@@ -77,6 +77,7 @@ export function loadComponentDirectories(loadedPluginModules?: Map<any, any>, lo
 	}
 	const hdbAppFolder = process.env.RUN_HDB_APP;
 	if (hdbAppFolder) {
+		if (getWorkerIndex() === 0) harperLogger.info?.('Loading application from ' + hdbAppFolder);
```
cb1kenobi, line 80: minor nit:
```suggestion
		if (getWorkerIndex() === 0) harperLogger.info?.(`Loading application from ${hdbAppFolder}`);
```

cb1kenobi (approved):

PR #1110, opened 2026-06-03
feat(cli): add harper describe_data command

kriszyp:
## Summary

- Adds a new `harper describe_data` CLI command that shows tables and attributes in the local (or remote) Harper instance
- Adds `preferLocal` transport-only flag to `cliOperations` so the command defaults to the local domain socket instead of the saved `last_target`
- Updates help text to include the new command

### Output

```
$ harper describe_data
Connecting to local Harper instance
Schema: data
  Dog  [id*, name, breed, age]
```

The asterisk marks the primary key. Pass `target=<url>` to inspect a remote instance.

## Motivation

New users following the getting-started guide run `harper dev .` and then immediately move to enabling the REST API with no way to verify the Dog table was actually created from their schema. This command gives them a quick sanity check without needing `curl` or the operations API.

Also discovered that `harper <api-operation>` commands fall back to `last_target` from saved credentials, which is surprising in a local-development context. The `preferLocal` flag is the opt-out for that behavior.

## Test plan

- [ ] `harper describe_data` with Harper running locally shows correct schema
- [ ] `harper describe_data` with no Harper running prints a clear error ("Harper must be running")
- [ ] `harper describe_data target=<url>` routes to the remote instance
- [ ] `harper describe_data` with a saved `last_target` in credentials still uses local socket
- [ ] `harper help` includes the new command

### DIFF
--- a/bin/cliOperations.ts
+++ b/bin/cliOperations.ts
@@ -35,6 +35,7 @@ const TRANSPORT_ONLY_FIELDS = new Set([
 	'json',
 	'skip_node_modules',
 	'skip_symlinks',
+	'preferLocal',
 ]);
 
 export { cliOperations, buildRequest };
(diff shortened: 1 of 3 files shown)


kriszyp:
Closing — the existing `describe_database database=data` operation already covers this via the CLI passthrough. Will just update the docs to reference it instead of adding a new command.

PR #1121, opened 2026-06-03
fix(replay): purge aged transaction logs during recovery before replay

kriszyp:
Cherry-pick of #1117 for v5.0.26 patch release.


DavidCockerill (approved):
looks good - I'll let you decide on what to do with the Gemini comments

PR #1124, opened 2026-06-03
ci: add always-on toggle + opt-in Gemini review caller

heskew:
## Always-on toggle + opt-in Gemini review

Brings this repo onto the unified review-toggle model (ai-review-prompts#48, USAGE "Reviewers & the always-on toggle"). Two changes:

**1. `claude-review.yml` — add the `CLAUDE_ALWAYS_ON` toggle.** A job-level `if:` gates the auto-run path on the `CLAUDE_ALWAYS_ON` repo variable, with the `claude-review` label opt-in always on top:

```yaml
if: ${{ vars.CLAUDE_ALWAYS_ON == 'true' || github.event.action == 'labeled' }}
```

This is **backward-compatible**: with `CLAUDE_ALWAYS_ON=true` set, behavior is exactly as today. Pin unchanged (the bump is the separate PR #1118).

**2. New `gemini-review.yml` — opt-in Gemini reviewer.** Mirrors `claude-review.yml`'s layers + Harper-core checks so the two providers are comparable on the same PR. Pinned to ai-review-prompts `2be0f70`. **Opt-in by default** — runs only when a HarperFast org member applies the `gemini-review` label (the labeler, not the PR author, satisfies the auth gate), unless `GEMINI_ALWAYS_ON=true` is set.

### Maintainer actions before merge
- **Set `CLAUDE_ALWAYS_ON=true`** (repo Settings → Secrets and variables → Actions → Variables, or `gh variable set`) so Claude keeps auto-reviewing on the transition. Without it, Claude flips to opt-in-only.
- For the Gemini leg: create a **`gemini-review`** label + a **`GEMINI_API_KEY`** secret (a missing key skips the Gemini review cleanly — safe to merge before it's set).

### Notes
- Separate from the pin-bump PR **#1118** per the agreed split. Both touch `claude-review.yml` on different lines (pin vs `if:`); whichever merges second may need a trivial rebase.

---

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -30,7 +30,7 @@ jobs:
     # always-on toggle". CLAUDE_ALWAYS_ON=true (repo/org variable) → auto-
     # review trusted-author PRs; unset → opt-in via the claude-review
     # label. The reusable's authorize job still owns WHO is admitted.
-    # NB: the `claude-review` label name is matched there too —
+    # Note: the `claude-review` label name is matched there too —
     # `_claude-review.yml`'s authorize `if:`, not in this caller.
     if: ${{ vars.CLAUDE_ALWAYS_ON == 'true' || github.event.action == 'labeled' }}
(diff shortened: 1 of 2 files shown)


kriszyp (approved):
Cool!

(Gemini not being able to review this is hilarious)

PR #1125, opened 2026-06-03
feat: Export serveStatic

dawsontoth:
`@harperfast/vite-plugin` wants to serve a Vite build through Harper's built-in `static` plugin instead of bundling its own static server. Two things block a clean integration:

1. **No public programmatic entry.** The only entry point is `handleApplication(scope)` in [server/static.ts](server/static.ts), and the `harper` package only exports `"."` ([package.json:121](package.json)), so `harper/dist/server/static.js` can't be imported by path. The plugin currently resolves the file on disk and `require()`s it, then fakes a `Scope` — fragile across versions.
2. **No cache-header control.** The handler calls `send(req, realpathSync(file))` with no options ([server/static.ts:133](server/static.ts)), so consumers can't set `Cache-Control`. A Vite build needs content-hashed assets cached forever (`public, max-age=31536000, immutable`) while `index.html` stays `no-cache`.

Goal: add a `Scope`-free `serveStatic(scope, options)` helper, export it from the `harper` package, and add cache/header options — both to the helper and to the YAML config-driven path. All existing config behavior (`files`, `urlPath`, `index`, `extensions`, `fallthrough`, `notFound`, `before`) stays unchanged; everything here is additive.

Decisions confirmed with the user:
- Expose `serveStatic` (+ `ServeStaticOptions`) from the `harper` package via `index.ts`.
- Add `maxAge` / `immutable` / `cacheControl` to the YAML config path too (live-reloadable); `setHeaders` is programmatic-only (it's a function, can't come from YAML).

- `send` v1.2.1 ships **no types** and tsconfig has no `noImplicitAny`/`strict`, so `send` is effectively `any` — `.on('headers', …)` needs no type wrangling.
- `send`'s stream emits `'headers'(res, path, stat)` **before** it applies its own `Cache-Control`, and it only sets `Cache-Control` if not already present ([node_modules/send/index.js:730-754]). So setting the header in a `'headers'` listener wins over send's default. This is exactly how `cacheControl` (full string) and `setHeaders` are implemented. `maxAge`/`immutable` are passed natively to `send`.
- An `EntryHandler` ([components/EntryHandler.ts:98](components/EntryHandler.ts)) is rooted at an explicit `directory` and watches a `files` glob, deriving `entry.urlPath` (via `resolveBaseURLPath(name, urlPath)` + glob-base stripping) and `entry.absolutePath`. This lets `serveStatic` watch an **arbitrary absolute directory** independent of `scope.directory`, which `scope.handleEntry` can't (it hardcodes `scope.directory`).
- The static plugin is registered as `static: staticFiles` in `TRUSTED_RESOURCE_PLUGINS` ([components/componentLoader.ts:104](components/componentLoader.ts)) and invoked via `plugin.handleApplication(scope)`. This wiring is untouched.
- `index.ts` already has a "regular exports" block at the top ([index.ts:9-12](index.ts)) for value exports like `getContext` — `serveStatic` goes there. (Do NOT use the runtime-globals block lower down.)

Extract three reusable pieces (no behavior change to existing logic):

- **`updateStaticMaps(entry, staticFiles, indexEntries)`** — the `add`/`unlink`/`addDir`/ `unlinkDir` switch currently inline in `handleApplication` ([static.ts:38-73]). Maintains the `staticFiles` and `indexEntries` maps. Pure; shared by both entry points.
- **`matchStaticFile(pathname, { staticFiles, indexEntries, index, extensions })`** — the lookup currently at [static.ts:88-126]. Returns a discriminated result: `{ kind: 'file', absolutePath }` | `{ kind: 'redirect', location }` | `{ kind: 'none' }`.
- **`respondWithFile(req, absolutePath, cache)`** — wraps `send`: ```ts const sendOptions = {};
  if (cache.maxAge !== undefined) sendOptions.maxAge = cache.maxAge;       // number ms or ms-string
  if (cache.immutable !== undefined) sendOptions.immutable = cache.immutable;
  const body = send(req, realpathSync(absolutePath), sendOptions);
  if (cache.cacheControl || cache.setHeaders) {
    body.on('headers', (res, path, stat) => {
      if (cache.cacheControl) res.setHeader('Cache-Control', cache.cacheControl);
      if (cache.setHeaders) cache.setHeaders(res, path, stat);   // runs last; can override
    });
  }
  return { handlesHeaders: true, body };
  ```
  Used for both matched files and the `notFound` file.

Add a shared **`createStaticResponder({ scope, staticFiles, indexEntries, directory, getConfig })`** that returns the `(req, next)` handler. `getConfig()` returns the live config object `{ index, extensions, fallthrough, notFound, cache }`. The responder keeps the exact control flow of today: GET/non-websocket guard → `matchStaticFile` → `respondWithFile` on hit → 301 on redirect → `next(req)` on fallthrough → `notFound`/404 handling (resolving the notFound file under `directory`, reusing `validateNotFoundOption` which already exists at [static.ts:172]).

Define the public options type:
```ts
export interface ServeStaticOptions {
  directory?: string;          // absolute dir to serve from; default scope.directory
  urlPath?: string;            // base URL path, e.g. '/'
  files?: FilesOption;         // glob to watch, relative to `directory`; default '**'
  index?: boolean;             // default true
  extensions?: string[];       // default []
  fallthrough?: boolean;       // default true
  notFound?: string | { file: string; statusCode: number };
  before?: string;             // default 'authentication'
  // cache / headers:
  maxAge?: number | string;
  immutable?: boolean;
  cacheControl?: string;
  setHeaders?: (res: ServerResponse, pathname: string, stat: Stats) => void;
}
```

```ts
export function serveStatic(scope: Scope, options: ServeStaticOptions = {}): EntryHandler {
  const directory = options.directory ?? scope.directory;
  const staticFiles = new Map<string, string>();
  const indexEntries = new Map<string, string>();

  // Own EntryHandler rooted at `directory` — independent of scope.options config.
  const entryHandler = new EntryHandler(
    scope.pluginName,
    directory,
    { files: options.files ?? '**', urlPath: options.urlPath },
    scope.logger
  );
  entryHandler.on('error', (e) => scope.logger.error?.('serveStatic watcher error:', e));
  entryHandler.on('all', (entry) => updateStaticMaps(entry, staticFiles, indexEntries));
  scope.on('close', () => void entryHandler.close());   // lifecycle cleanup

  const config = {            // captured snapshot (programmatic; not live YAML)
    index: options.index ?? true,
    extensions: options.extensions ?? [],
    fallthrough: options.fallthrough ?? true,
    notFound: options.notFound,
    cache: { maxAge: options.maxAge, immutable: options.immutable,
             cacheControl: options.cacheControl, setHeaders: options.setHeaders },
  };
  scope.server.http(
    createStaticResponder({ scope, staticFiles, indexEntries, directory, getConfig: () => config }),
    { before: options.before ?? 'authentication', urlPath: options.urlPath }
  );
  return entryHandler;
}
```

Notes:
- Returns the `EntryHandler` so callers can `await ret.ready` or `ret.close()`; auto-closes on `scope` `'close'`.
- Passing `urlPath` explicitly in the http options overrides the value the `Scope` proxy injects from `scope.options` ([Scope.ts:104-110]), so routing matches `options.urlPath`.
- `directory` should exist before calling (the Vite plugin builds first); chokidar tolerates later-created files within a watched dir.

- Keep the `staticFiles`/`indexEntries` maps and the `scope.options.on('change')` clear of maps on `files`/`urlPath` change ([static.ts:27-35]).
- Keep `scope.handleEntry((entry) => updateStaticMaps(entry, …))` (default, config-driven watcher rooted at `scope.directory`).
- Register `createStaticResponder` with `getConfig()` that reads `scope.options` **live each request** (preserving today's per-request reload of `index`/`extensions`/`fallthrough`/ `notFound`), and now also reads the new `maxAge`/`immutable`/`cacheControl` keys into `cache`. `directory` = `scope.directory`; `before` read once at registration as today.

This preserves the config-driven path exactly, plus the additive YAML cache options.

In [index.ts](index.ts), in the top regular-exports block (next to the `getContext` line):
```ts
export { serveStatic } from './server/static.ts';
export type { ServeStaticOptions } from './server/static.ts';
```
Compiles to `dist/index.js` / `dist/index.d.ts`; the existing `exports["."]` ([package.json:121-126]) then makes `import { serveStatic } from 'harper'` work. No `exports`-map change needed.

- [server/static.ts](server/static.ts) — refactor: extract `updateStaticMaps`, `matchStaticFile`, `respondWithFile`, `createStaticResponder`; add `ServeStaticOptions` + `serveStatic`; rebuild `handleApplication` on the shared internals; add cache reads.
- [index.ts](index.ts) — add the two public exports.
- Imports needed in `static.ts`: `EntryHandler` from `../components/EntryHandler`, `FilesOption` (type) from `../components/deriveGlobOptions`, `Stats` / `ServerResponse` types as needed.

- **Build/lint:** `npm run build` (catches the new `index.ts` export + any circular-import surprise from `index.ts` → `static.ts` → `Scope`), then `npm run lint` and `npm run format:write`.
- **Unit tests** (`node:assert/strict`, no sinon/rewire — match `unitTests/components/*`), new `unitTests/server/static.test.js`:
  - `matchStaticFile`: file hit, index hit, `null`→redirect, extension fallback, miss.
  - cache wiring: build a `send` stream via `respondWithFile` with `cacheControl` / `setHeaders`, emit a synthetic `'headers'` event with a fake `res` (object recording `setHeader`), assert the header is set and `setHeaders` runs last.
  - `serveStatic` integration-ish: construct a real `Scope` over a temp dir with a fake `server` `{ http: (fn, opts) => recorded.push({fn, opts}) }` (pattern from [unitTests/components/Scope.test.js:18-58]), write files, `await ret.ready`, assert the responder was registered with the right `before`/`urlPath` and that a GET for a known file returns `{ handlesHeaders: true }`.
- **Integration test** (extend the static fixture pattern in [integrationTests/deploy/deploy-tracking.test.ts:77-100]): deploy a `static:` component with `cacheControl: 'public, max-age=31536000, immutable'` (or `maxAge`/`immutable`), request a served file over HTTP, assert the `Cache-Control` response header; request with no cache config and assert send's default (`public, max-age=0`) still applies.
- **Manual smoke (optional):** point the WIP `@harperfast/vite-plugin` at the rebuilt `harper`, call `serveStatic(scope, { directory: <dist>, urlPath: '/', setHeaders })`, and curl `/assets/*.js` (immutable) vs `/index.html` (no-cache).

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -10,6 +10,7 @@ if (!workerThreads.isMainThread) {
 export { RequestTarget } from './resources/RequestTarget.ts';
 export { flushDatabases } from './resources/databases.ts';
 export { getContext, getResponse, getUser } from './security/jsLoader.ts';
+export { serveStatic } from './server/static.ts';
 
 // Type only exports.
 // Anything exported here will only be available as TypeScript types, not as values.
@@ -29,6 +30,7 @@ export type { IterableEventQueue } from './resources/IterableEventQueue.ts';
 export type { Table } from './resources/databases.ts';
 export type { Attribute } from './resources/Table.ts';
 export type { Scope } from './components/Scope.ts';
+export type { ServeStaticOptions, StaticCacheOptions } from './server/static.ts';
 export type { FilesOption, FilesOptionObject } from './components/deriveGlobOptions.ts';
 export type { FileAndURLPathConfig } from './components/Component.ts';
 export type { OptionsWatcher, Config, ConfigValue } from './components/OptionsWatcher.ts';
--- a/server/static.ts
+++ b/server/static.ts
@@ -1,7 +1,64 @@
-import { realpathSync, existsSync } from 'node:fs';
+import { realpathSync, existsSync, type Stats } from 'node:fs';
 import { join } from 'node:path';
+import type { ServerResponse } from 'node:http';
 import { Scope } from '../components/Scope';
+import { EntryHandler, type FileEntryEvent, type DirectoryEntryEvent } from '../components/EntryHandler';
+import type { FilesOption } from '../components/deriveGlobOptions';
 import send from 'send';
+
+/**
+ * Cache / response-header options shared by the config-driven plugin and the programmatic
+ * `serveStatic` helper.
+ *
+ * - `maxAge` and `immutable` are forwarded directly to `send`.
+ * - `cacheControl` is a full `Cache-Control` header value that takes precedence over `send`'s
+ *   default (`send` only sets `Cache-Control` when one isn't already present).
+ * - `setHeaders` is a per-file callback (like `express.static`/`sirv`) invoked just before the
+ *   response headers are sent; it runs last and can override `cacheControl`.
+ */
+export interface StaticCacheOptions {
+	maxAge?: number | string;
+	immutable?: boolean;
+	cacheControl?: string;
+	setHeaders?: (res: ServerResponse, pathname: string, stat: Stats) => void;
+}
+
+/**
+ * Options for the programmatic {@link serveStatic} helper. All options are independent of the
+ * component's own `files`/`urlPath` config, so an extension can serve any directory without
+ * synthesizing a Scope.
+ */
+export interface ServeStaticOptions extends StaticCacheOptions {
+	/** Absolute directory to serve from. Defaults to `scope.directory`. */
+	directory?: string;
+	/** Base URL path to serve under, e.g. `'/'`. */
+	urlPath?: string;
+	/** Glob (or `FilesOptionObject`) to watch, relative to `directory`. Defaults to `'**'`. */
+	files?: FilesOption;
+	/** If enabled, serve `index.html` files from directories. Defaults to `true`. */
+	index?: boolean;
+	/** File extensions to try when a file is not found (e.g. `['html']`). Defaults to `[]`. */
+	extensions?: string[];
+	/** If `true`, fall through to the next handler when a file is not found. Defaults to `true`. */
+	fallthrough?: boolean;
+	/** Custom 404 handling: a file path, or `{ file, statusCode }`. */
+	notFound?: string | { file: string; statusCode: number };
+	/** Named middleware ordering for `scope.server.http`. Defaults to `'authentication'`. */
+	before?: string;
+}
+
+type NotFoundOption = undefined | string | { file: string; statusCode: number };
+
+/**
+ * Resolved, per-request configuration used by the shared static responder.
+ */
+interface StaticConfig {
+	index: boolean;
+	extensions: string[];
+	fallthrough: boolean;
+	notFound: NotFoundOption;
+	cache: StaticCacheOptions;
+}
 
 /**
  * The static plugin handles serving static files from the respective application directory.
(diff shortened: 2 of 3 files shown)


dawsontoth:
Funny that Claude is like "this is perfect" and Gemini is like "ALL THE THINGS ARE BROKEN!" Probably pivoting away from this to using the two plugins in parallel instead.

kriszyp (commented):
This looks cool to me!

> Scope-free serveStatic(scope, options)

"Scope-free" and the first argument is literally the scope 😆 
I guess this means that it isn't a method on the scope. I am not sure why wouldn't have this as method on the scope. But I guess either way is fine.

> Gemini is like "ALL THE THINGS ARE BROKEN!"

Did you get a gemini (or agy) review before pushing the PR? I think that does tend to cut down on cloud gemini complaints (they use different skills/prompts, so doesn't eliminate them all).

PR #1127, opened 2026-06-03
fix(nodeName): hostnameToUrl should not manufacture 'ws://undefined:port' URLs

ldt1996:
## Summary

`hostnameToUrl(undefined)` returns the literal string `"ws://undefined:9933"` (or similar with whatever port is configured) instead of `undefined`. That truthy garbage string then flows through callers that gate behavior on `url` being defined, bypassing their validation.

## The bug

[`server/nodeName.ts:67`](server/nodeName.ts#L67):

```ts
export function hostnameToUrl(hostname) {
    let port = getPortFromListeningPort('replication_port');
    if (port) return `ws://${hostname}:${port}`;
    ...
}
```

When `hostname` is undefined, template literal interpolation produces `"ws://undefined:9933"`. Any caller doing `if (url) { ... }` then proceeds with a meaningless URL, leading to confusing downstream errors that are hard to trace back to "the hostname was missing."

Concrete case caught in QE: `harper-pro`'s `setNode` validates with `if (!url && !hostname)`. When a `remove_node` request omits hostname, `hostnameToUrl(undefined)` returns the bogus string, the validation passes, and a downstream `hdbNodes.get(undefined)` throws an opaque "Using an argument with a value of undefined for get" instead of the clean "url or hostname is required" the validation was supposed to produce. There are almost certainly other callers with the same shape.

## Change

One-line guard at the top of `hostnameToUrl`:

```ts
export function hostnameToUrl(hostname) {
    if (!hostname) return undefined;
    let port = getPortFromListeningPort('replication_port');
    ...
}
```

Functions should not manufacture content out of nothing. Returning `undefined` for a missing input lets callers' existing validation work correctly.

## Companion

Companion fix in [harper-pro](https://github.com/HarperFast/harper-pro/pull/TBD) accepts `name` as an alias for `hostname` in the `setNode` API, which is the proximate cause of the specific QE failure. This PR is the deeper fix that prevents the same class of bug surfacing through any other caller.


kriszyp (approved):

heskew (approved):

heskew:
Small test cleanup suggestion: could we use the real test-facing config setter instead of stubbing `env.get` here? `environmentManager.setProperty()` is already intended for unit tests, and using it keeps this regression test closer to the config path `hostnameToUrl()` uses normally.

For example:

```js
const assert = require('node:assert/strict');

const env = require('#src/utility/environment/environmentManager');
const { CONFIG_PARAMS } = require('#src/utility/hdbTerms');
const { hostnameToUrl } = require('#src/server/nodeName');

describe('hostnameToUrl', () => {
	let originalReplicationPort;

	beforeEach(() => {
		originalReplicationPort = env.get(CONFIG_PARAMS.REPLICATION_PORT);
		env.setProperty(CONFIG_PARAMS.REPLICATION_PORT, '0.0.0.0:9933');
	});

	afterEach(() => {
		env.setProperty(CONFIG_PARAMS.REPLICATION_PORT, originalReplicationPort);
	});

	// existing cases...
});
```

The behavior coverage looks good; this would just avoid adding another stubbed unit test for a path we can exercise directly.

---

AGENTS.md has some info on trying to avoid new uses of sinon and node:assert over bare assert (I think). Love to see the test coverage with the fix. 👏

cb1kenobi (approved):

PR #1131, opened 2026-06-04
fix(auditStore): defensive copy on no-encodedRecord return to prevent shared-buffer race

### LINKED ISSUE #1132: createAuditEntry shared ENTRY_HEADER view races on no-encodedRecord path, corrupts txnlog
## Summary

`createAuditEntry` returns a view into the module-level singleton `ENTRY_HEADER` buffer on the no-`encodedRecord` return path. The next `createAuditEntry` call writes into `ENTRY_HEADER` from byte 0 again and overwrites the previous return value's underlying bytes before rocksdb-js's `log.addEntry` necessarily captures them. The result is corrupt entries on disk; a later read fails with `RangeError: Corrupt transaction log entry at position N: declared length L overruns the log`, killing whichever WS handler triggered the read and cascading into a 1006 reconnect storm.

## Where

[`resources/auditStore.ts:397-400`](resources/auditStore.ts#L397):

```ts
const header = ENTRY_HEADER.subarray(0, position);
if (encodedRecord) {
    return Buffer.concat([header, encodedRecord]);   // safe (concat copies)
} else return header;                                 // shared-buffer view
```

`ENTRY_HEADER` is a 2816-byte module-level singleton at [`resources/auditStore.ts:56`](resources/auditStore.ts#L56). Every `createAuditEntry` call writes into it starting at `start` (default 0) and the no-`encodedRecord` path returns a `subarray` view, not a copy.

## When it manifests

The bug is engine/version-independent but only fires when:

1. The no-`encodedRecord` return path is hit (deletes, invalidates, any audit entry that carries no record payload).
2. Two `createAuditEntry` calls happen close enough in time that the second one mutates `ENTRY_HEADER` before the first return value has been copied into native storage by `log.addEntry`.

Observed in production on a v4→v5 stage cluster at ~28 MB/min inbound system-db replication from 3 v4 peers (dominated by `analytics.replicate: true`, matching the rt-ite analytics-storm pattern). v4→v5 also produces a delete-heavy initial sync mix that exercises the no-`encodedRecord` path. v5↔v5 clusters with comparable sustained write pressure and delete share would also corrupt.

## Reproducer

The corrupt file from the stage cluster is at:
```
/home/harperdb/harper/database/system/transaction_logs/<peer>/14.txnlog
```

Read fails with two error variants at the same position 0x95 depending on buffer fill level:
```
RangeError: Corrupt transaction log entry at position 95 of log 14: declared length 177 overruns the log (limit=195)
RangeError: Corrupt transaction log entry at position 95 of log 14: declared length 97 overruns the log (limit=16777115)
```

Stack:
```
transaction-log-reader.ts:191 (rocksdb-js native)
RocksTransactionLogStore.ts:277 (merge iterator next())
notifyFromTransactionData (transactionBroadcast.ts:164)
```

The uncaught throw kills the WS handler, the receiver requests full copy on reconnect, replay re-hits the same corrupt entry, loop.

## Fix

Two-part:

1. **Writer**: defensive copy on the no-`encodedRecord` return path. Change `return header` to `return Buffer.from(header)` at `resources/auditStore.ts:400`. Eliminates the race.
2. **Reader**: wrap the `iterator.next()` calls in `RocksTransactionLogStore`'s merge iterator with try/catch on `RangeError`; log and mark that log done so the merge keeps draining the others. Prevents one corrupt entry from cascading into a 1006 storm on any future corruption.

PR linking once opened.

ldt1996:
Closes #1132.

## Summary

- `createAuditEntry`: `return Buffer.from(header)` instead of `return header` on the no-`encodedRecord` path. Defensive copy out of the shared `ENTRY_HEADER` singleton.
- `RocksTransactionLogStore`: wrap the merge iterator's `iterator.next()` calls (both the initial poll and the re-poll) with try/catch on `RangeError`. Log and mark the affected log done so the merge keeps draining the others.

## Context

`ENTRY_HEADER` is a 2816-byte module-level buffer reused across every `createAuditEntry` call. The no-`encodedRecord` return path was returning `ENTRY_HEADER.subarray(0, position)`, a view, not a copy. The next call wrote into `ENTRY_HEADER` from byte 0 and overwrote the previous return value's bytes before rocksdb-js's `log.addEntry` necessarily captured them. Result: corrupt entries on disk, later `RangeError: Corrupt transaction log entry at position N: declared length L overruns the log` from the rocksdb-js reader.

The bug is engine/version-independent but surfaces in practice on v4→v5 replication because the inbound rate is much higher (~28 MB/min on a 3 v4 peer stage cluster, dominated by `analytics.replicate: true`) and the v4→v5 initial-sync mix is delete-heavy, which is exactly the no-`encodedRecord` path.

The writer-side fix eliminates the race. The reader-side defense in `RocksTransactionLogStore` independently breaks the cascade: the uncaught `RangeError` was killing the WS handler, triggering full-copy on reconnect, replay re-hitting the same corrupt entry, looping at 1006. Log-and-skip the affected log instead so the merge continues and the connection stays up.


resources/RocksTransactionLogStore.ts
```diff
@@ -231,7 +231,17 @@ export class RocksTransactionLogStore extends EventEmitter {
...
 					}
 				}
-				nextEntries = iterators.map((iterator) => iterator.next());
+				nextEntries = iterators.map((iterator, i) => {
+					try {
+						return iterator.next();
+					} catch (error) {
+						if (error instanceof RangeError) {
+							harperLogger.error(`Corrupt transaction log; skipping remainder of log ${logs[i]?.name}`, error);
+							return { done: true, value: undefined };
```
heskew: This marks only the current `nextEntries` slot done, but leaves the corrupt log and iterator in `logs` / `iterators`. Because `aggregateIterator.next()` later clears `nextEntries` and re-polls, the same corrupt iterator can be retried on future commit drains. Could we remove/exclude the log here, and in the later `iterators[earliestIndex].next()` catch, using the same state update as `removeLog()` so the remainder is actually skipped for this aggregate iterator?

heskew (commented):
The `Buffer.from(header)` change in `createAuditEntry()` looks accurate and necessary: the no-record path was returning a view into the module-level `ENTRY_HEADER`, so a later audit entry could mutate data already handed to Rocks.

I left one inline note on the `RangeError` handling. The defensive reader change is the right direction, but the aggregate iterator should also remove/exclude the corrupt log so later drains do not keep retrying it.

I would also add focused regressions for no-record `createAuditEntry()` buffer stability across calls and aggregate `getRange({})` continuing through a healthy log without retrying a corrupt one.

_Generated by OpenAI Codex (GPT-5) at Nathan's request._

heskew:
Ignore the gemini pr review failure. The gemini cli action used internally has version string actions, not sha-pinned, which breaks our 'must be sha-pinned' requirement. I'll have to look into this separately later...

resources/RocksTransactionLogStore.ts
cb1kenobi: So what do you propose? `iterators.splice(i, 1)`?

cb1kenobi: I don't think we need to remove it from the `iterators` because the caller will see it as done anyways.

cb1kenobi (approved):

kriszyp (changes requested):
This is intentionally being done to avoid the cost of copying buffers. This is seems like it introduces a performance regression.
Is corruption even possible after the corruption fixes we have made?

ldt1996:
@kriszyp Looking at commits, your 986943d74 does it cleaner than what I had .  rebased and dropped that half + the duplicate test. 

`RangeError: Corrupt transaction log entry at position N of log M: declared length X overruns the log (limit=Y)
    at transaction-log-reader.ts:191` this is the error I'm fixing. The defensive copy I left in should stop it at the source

resources/RocksTransactionLogStore.ts
ldt1996: dropped this piece of code

ldt1996:
the CI/CD failures are not related to my branch

kriszyp:
After further tracing, I determined this is unnecessary, extra overhead.

PR #1134, opened 2026-06-04
fix: Harper login can use inquirer for easier parsing

dawsontoth:
At some point, the implementation regressed (or it never worked?) for how it masks out password entry. This fixes all that by using inquirer, which I didn't notice we had the first time around. 🤦 

The bug this fixes when typing:
```
Cluster Password: w*o*w* *s*o* *s*e*c*u*r*e*
```

Or if you paste, it's even better:
```
Cluster Password: wow so secure*************
```

becomes:
```
Cluster Password: 
```
[aka, it's masked out consistently, just like we do during the harper install cli command]

### DIFF
--- a/bin/login.ts
+++ b/bin/login.ts
@@ -2,7 +2,7 @@ import chalk from 'chalk';
 import dotenv from 'dotenv';
 import fs from 'node:fs';
 import path from 'node:path';
-import readline from 'node:readline/promises';
+import inquirer from 'inquirer';
 import { saveCredentials, normalizeTarget } from './cliCredentials.ts';
 import { cliOperations } from './cliOperations.ts';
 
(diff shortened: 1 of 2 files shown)


dawsontoth:
seems like the integration tests are hosed at the moment, not a part of my changes

heskew:
Did the test need to be updated? Looks like unit tests are hanging at the login test.

dawsontoth:
Oh! Interesting. I looked at the tests and only saw other stuff erroring, but that totally makes sense.

dawsontoth:
oh i was only looking at the integration test failures, with their red checkmark, since those were angry but unrelated. I see the hanging unit tests now.

dawsontoth:
@heskew they are tre happy now!

heskew (approved):

cb1kenobi (approved):

Issue #1138, opened 2026-06-04
Replace fixed-delay 'sleep-then-assert' timing races in unit tests (umbrella)


heskew:
## Summary

A recurring class of unit-test flakiness comes from **fixed wall-clock waits followed by a state assertion** — `await delay(N); assert(<async side effect has happened>)`. When a CI runner is slow or loaded, the side effect hasn't completed within the hard-coded delay and the assertion fails. This is the root pattern behind #19 (blob save/delete) and is spread across the unit suite. Filing as an umbrella to drive these toward condition-waits / fake timers instead of sleeps.

## The canonical example (#19)

`unitTests/resources/blob.test.js`, "Save a blob and delete it" (`:188-229`):

```js
await BlobTest.delete(3);
await delay(50);
assert(!existsSync(filePath)); // should immediately be deleted
...
let retries = 0;
while (existsSync(filePath) && retries++ < 10) await delay(40); // wait for audit log removal
```

It asserts that asynchronous deletion + audit-log cleanup finished within a fixed wait. #19's own CI evidence: only the slower **Node v20** job failed (`assert(existsSync(filePath))`) while v22/v24 passed — i.e. load-dependent, the signature of a timing race. (PR #462 touched this file but the fixed-delay pattern remains.)

## Breadth

35 unit-test files use fixed `delay()` / `setTimeout` / `sleep`. Highest concentrations (occurrence count):

| file | count |
|------|------:|
| `unitTests/resources/caching.test.js` | 31 |
| `unitTests/apiTests/mqtt-test.mjs` | 26 |
| `unitTests/resources/subscriptionReplay.test.js` | 21 |
| `unitTests/utility/logging/harper_logger.test.js` | 18 |
| `unitTests/resources/blob.test.js` | 16 |

Not every occurrence is a bug (some are legitimately modeling time), but the "sleep then assert a side effect" shape is the one to hunt.

## Structural amplifier

Unlike the integration runner (process isolation per file), `test:unit:*` runs **mocha with all matched files in a single process** (`.mocharc.json`, no isolation). Module-level singletons (the table/database registry, thread state, config) persist across files, so a timing race in one file plus leftover state from another compounds into order-dependent intermittency. This overlaps with #18 (types validation flaky — note that file has *no* timing code, so its intermittency is shared-state/ordering, not delays) and the umbrella migration in #116.

## Suggested direction (not a fix — for discussion)

[...]

Issue #116, opened 2026-01-23
Move integration-y unit tests into integration test system


cap10morgan:
Some of our unit tests are more like integration tests and would benefit from being moved into that system. We should go over them and do that migration. Note that @heskew already did a great job of this for the server unit tests (in `unitTests/server`) in https://github.com/HarperFast/harper/pull/113 so those are probably all set.

Issue #19, opened 2025-11-03
Save & delete blob test fails intermittently


cap10morgan:
### Bug Summary

The "Save a blob and delete it" test in the resources/blob unit tests fails intermittently

### Steps to Reproduce

1. Remove the `skip` directive from the `Save a blob and delete it` test in `unitTests/resources/blob.test.ts`
2. Run `npm run test:unit unitTests/resources/blob.test.ts` (a few times if needed)

### Expected Behavior

The tests should all pass reliably

### Actual Behavior

The `Save a blob and delete it` test fails intermittently 

### Are you planning to fix this issue?

No, just reporting the issue

### First-time contributor support

- [ ] I'm new to contributing and would appreciate guidance on the process
- [ ] I'd like help understanding the project structure
- [ ] I need assistance with setting up the development environment
- [ ] I'm comfortable contributing, but new to this project specifically

kriszyp:
Addressed by https://github.com/HarperFast/harper/pull/462 and https://github.com/HarperFast/harper/commit/8ae5588e1ad429f3cc9d3fc068ce2579f63925e1

kriszyp:
Fixed by #462 (merged 2026-05-06) plus follow-up commit.

— Claude (Opus 4.8), issue-backlog triage on Kris's behalf

PR #1144, opened 2026-06-04
test: replace fixed-delay sleep-then-assert with waitFor (#1138)

heskew:
## Summary

Addresses #1138 (umbrella). Replaces fixed-delay `await delay(N); assert(sideEffect)` "sleep-then-assert" timing races — the root cause behind #19 and a class of unit-test flakiness — with a shared `waitFor()` poll-until-condition helper.

This is a **test-only** change (no production code).

## What changed

- **Shared helper** (`unitTests/waitFor.js`): promotes the existing `components/`-local `waitFor` to a shared location and enhances it — condition may be sync or async, optional `{ timeout, interval, message }` (backward-compatible with existing `waitFor(cond)` / `waitFor(cond, 3000)` callers), returns the truthy value, `assert.fail` on timeout. The old `unitTests/components/waitFor.js` is removed and its two consumers (`Scope.test.js`, `EntryHandler.test.js`) re-pointed.
- **`blob.test.js`** (closes the #19 root cause): converted the 8 "wait for deletion / audit-cleanup" sites to `waitFor(() => !existsSync(...))`. **Left intentionally**: the 5 non-event `assert(existsSync(...))` "should still exist / must survive" delays (they assert something has *not* happened — converting would be wrong) and the stream-pacing `delay(i % 10)`.
- **`caching.test.js`**: converted the 5 genuine poll/eviction/revalidation races (the existing `while (sourceRequests===0)` poll loop, two eviction waits, the revalidation-complete wait, the post-evict check). **Left intentionally**: the TTL-expiration-window and sustained-load delays — those *model elapsed time*, and the count-based assertions tied to them would need fake timers, not a condition wait.
- **`harper_logger.test.js`**: converted the 4 live `write-log → wait → read-file → assert-contents` sites (guarded `readFileSync` with `existsSync` so the poll never throws ENOENT). The 14 callback-style `setTimeout(done)` tests are inside a `describe.skip` block (dead code) and were left untouched.
- **`AGENTS.md`**: documented the "prefer condition-waits over fixed delays" convention and the shared helper.

## Deferred to #116 (assigned)

`subscriptionReplay.test.js` and `mqtt-test.mjs` were investigated and **intentionally not converted here**:
- `subscriptionReplay.test.js` is pervasively timing-coupled through shared single-process state — on *unmodified* `main`, *which* test flakes varies run-to-run by machine load. Simple `delay→waitFor` shuffles the flakiness rather than fixing it; it needs the process-isolation / fake-timer treatment tracked by #116.
- `mqtt-test.mjs` is a live MQTT integration test (real broker); the issue itself flags the worst offenders as candidates for the process-isolated runner (#116).

## Validation

Ran the affected files repeatedly (default rocksdb runner) — stable, no `waitFor` timeouts:
- `blob.test.js`: 23 passing across many runs.
- `blob + caching` together: **0 failures / 15 runs** (vs `main` 1/15 — caching retains a ~6% **pre-existing** TTL-window flake in the count-based tests left for the fake-timer follow-up).
- `harper_logger.test.js`: 24 passing / 14 pending (the skipped block) across many runs.
- `EntryHandler + Scope` (existing `waitFor` consumers): unchanged vs `main` (the 6 pre-existing file-watch failures are environment-specific and present on `main` too).

## Where to look

- `unitTests/waitFor.js` — the number-or-options 2nd arg and async-condition handling.
- The judgement call on *which* delays to convert vs leave (non-event assertions and time-window modeling are intentionally preserved).

## Cross-model review

**Gemini (via `agy`):** validated the helper, the conversion semantics (no bug-masking — e.g. the stale-while-revalidate test still asserts `sourceRequests === 1` immediately after the wait, so a spurious extra request would still fail it), the `existsSync`-guarded log reads, and the left-alone delays. One **significant** finding — the new options-form silently dropped the original 3rd positional `interval` arg (`waitFor(cond, timeout, interval)`). No current caller uses it, but **fixed** anyway: the signature now accepts both `waitFor(cond, timeout, interval)` and `waitFor(cond, { timeout, interval, message })`.

**Codex CLI:** attempted three times (direct, and with MCP overridden) but hung indefinitely on its startup `flair` MCP `bootstrap` call (0% CPU, frozen output) — an environment issue, not a review finding. A Codex pass should be re-run when its MCP is responsive (or via the repo's automated review on this PR).

---


heskew:
Marking this Ready for review. Note on CI status:

✅ **Every check relevant to this change passes** — `Unit Test` (Node 22/24/26), all `Build Harper` jobs, `Format Check`, `runLinter`, `validate`, and Integration shards `3/4` & `4/4`.

❌ The red **Integration Tests `1/4` & `2/4`** jobs are a **pre-existing, `main`-wide failure**, not introduced by this PR. The Harper server fails to start with `You are attempting to upgrade from an old instance of Harper that is no longer supported` → `ECONNREFUSED 127.0.0.x:9925`, so every server-spawning test in those shards fails (e.g. `Start 4.x server and test upgrade`, `crash with random byte flips in txnlogs`). `main` HEAD (`266d5d8b`, _“fix(upgrade): run upgrade directives without interactive confirmation”_) is itself red on the same jobs, and this branch touches **no** `upgrade/` or `integrationTests/` files — it's test-only changes under `unitTests/`. A rebase once the upgrade-startup regression is fixed should clear these.

---

heskew:
Codex CLI review update: it ran clean — **no findings**. The earlier hang was its `flair` MCP bootstrap; running `codex exec --ignore-user-config review --base origin/main` bypassed that and it completed a full pass (diff + all touched files + `waitFor.js` internals + call sites). Verdict:

> The changes consolidate the waitFor helper and replace fixed sleeps with condition-based waits without introducing any clear regressions in the touched tests.

Cross-model review is now complete on both legs — Gemini (one significant finding, the dropped 3rd-arg `interval`, already fixed) and Codex (no findings).

---

kriszyp (approved):

cb1kenobi (approved):

PR #1166, opened 2026-06-08
fix(hnsw): don't drop int8 graph nodes read cold/frozen from disk (#1161)

### LINKED ISSUE #1161: HNSW int8: index fails to build correctly beyond ~3,000–4,000 rows
## Summary

The int8 HNSW index builds correctly at small row counts (~20 rows) but fails to index new records once a table exceeds roughly 3,000–4,000 rows. This is a correctness failure at trivial scale — well below any realistic production workload.

## Distinction from #895

[#895](https://github.com/HarperFast/harper/issues/895) tracks O(N) event-loop blocking during HNSW insertion (a performance/reliability issue). This is a separate correctness bug: indexing stops working entirely beyond a low threshold, meaning queries against int8-indexed tables will miss records.

## Expected behavior

int8 HNSW indexing should work correctly at any table size.

## Actual behavior

Indexing fails for records inserted after the table reaches ~3,000–4,000 rows.

## Priority

Flagged as critical — this threshold is too low for any real use; needs a fix or a gate before int8 indexing ships to customers.

## Reporter

Chris Nelson

kriszyp:
## Summary
- `safeGetSync` no longer mutates a frozen node in place when reinterpreting a stored int8 `bin` as a signed `Int8Array`; it clones the frozen node instead, memoizing the clone in a `WeakMap` keyed by the frozen node.
- Adds a regression test that forces every node read to look like a `freezeData` disk decode and asserts records stay findable.

## Purpose (fixes #1161)
int8 HNSW indexes silently stopped indexing/finding records once a table grew past ~3–4k rows. The HNSW index store sets `freezeData` ([OpenDBIObject.ts](../blob/main/utility/lmdb/OpenDBIObject.ts)), so any node decoded from disk on an object-cache **miss** is a **frozen** record whose int8 `vector` is an unsigned `Uint8Array`. The old code did `node.vector = …` to convert it, which **throws** on a frozen object; the surrounding `try/catch` swallowed the throw and returned `undefined` — silently dropping the node and fragmenting the graph. Freshly-written (warm, unfrozen) cache objects worked, which is why small tables were fine; the threshold is just the object cache filling up. Float nodes never mutate in `safeGetSync`, so they were unaffected.

Verified directly against the compiled code: unfixed → `Failed to decode HNSW node, skipping`/`undefined`; fixed → the node is returned with the correct signed vector. The added regression test fails **400/400 unfindable** on the old code and passes **0/400** with the fix.

## Where to look
- **`safeGetSync` clone + WeakMap** (`resources/indexes/HierarchicalNavigableSmallWorld.ts`). The `WeakMap` exists to keep the original "skip re-conversion for cached nodes" optimization on the search/index hot path: without it, every cache *hit* on a frozen node would re-slice the buffer and re-clone. It's keyed by the frozen node the store hands back (stable reference on hits) and is weak, so entries are collected when the store evicts the node. Correctness does not depend on the cache returning stable references — only the optimization does.
- Connection arrays on the cloned node remain the (frozen) originals; existing `Object.isFrozen` guards in `index()` already slice-before-mutate, so callers are unaffected.

## Notes
- Cross-model reviewed by Codex (no correctness issues) and Gemini/AntiGravity (which prompted the `WeakMap` hot-path optimization, now applied). No open review concerns.
- Generated by an LLM (Claude, Opus 4.7).

### DIFF
--- a/unitTests/resources/vectorIndex.test.js
+++ b/unitTests/resources/vectorIndex.test.js
@@ -637,6 +637,91 @@ describe('HNSW int8 quantization (quantization: "int8")', () => {
 	});
 });
 
+describe('HNSW int8 cold/frozen node reads (#1161)', () => {
+	if (process.env.HARPER_STORAGE_ENGINE === 'lmdb') return;
+	const DIMS = 16;
+	const N = 250;
+	let T;
+	const all = [];
+
+	function vec(seed) {
+		let s = (seed * 2654435761) >>> 0;
+		const rand = () => (s = (s * 1664525 + 1013904223) >>> 0) / 4294967296;
+		let cs = ((seed % 16) * 40503 + 1) >>> 0;
+		const crand = () => (cs = (cs * 1664525 + 1013904223) >>> 0) / 4294967296;
+		const out = new Array(DIMS);
+		let mag = 0;
+		for (let i = 0; i < DIMS; i++) {
+			const v = crand() * 2 - 1 + 0.15 * (rand() * 2 - 1);
+			out[i] = v;
+			mag += v * v;
+		}
+		const inv = 1 / (Math.sqrt(mag) || 1);
+		for (let i = 0; i < DIMS; i++) out[i] *= inv;
+		return out;
+	}
+
+	before(async () => {
+		setupTestDBPath();
+		setMainIsWorker(true);
+		T = table({
+			table: 'HNSWInt8Cold',
+			database: 'test',
+			attributes: [
+				{ name: 'id', isPrimaryKey: true },
+				{ name: 'vector', indexed: { type: 'HNSW', distance: 'cosine', quantization: 'int8' }, type: 'Array' },
+			],
+		});
+		for (let i = 0; i < N; i++) {
+			const v = vec(i);
+			await T.put(i, { vector: v });
+			all.push(v);
+		}
+	});
+	after(() => T.dropTable());
+
+	// The index store sets freezeData, so a node decoded from disk (a cache miss — which happens for
+	// older nodes once a table outgrows the object cache, ~3-4k rows) is a FROZEN record whose int8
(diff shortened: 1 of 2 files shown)


unitTests/resources/vectorIndex.test.js
```diff
@@ -637,6 +637,91 @@ describe('HNSW int8 quantization (quantization: "int8")', () => {
 	});
 });
 
+describe('HNSW int8 cold/frozen node reads (#1161)', () => {
+	if (process.env.HARPER_STORAGE_ENGINE === 'lmdb') return;
```
dawsontoth: Could we `skip` via the options argument? I think I've got it straight in my head, but... Senor AIs, please ponder this. Then we could
```
{ skip: process.env.HARPER_STORAGE_ENGINE === 'lmdb' ? 'skipped on lmdb since it does not apply' : false }
```

cb1kenobi (approved):

unitTests/resources/vectorIndex.test.js
kriszyp: Good instinct — I pondered it. 🙂 That `{ skip }` options-argument form comes from Node's built-in `node:test` runner, but this suite runs on **Mocha** (v11.7.5), which doesn't accept an options object: `it(title, { skip }, fn)` throws `TypeError: A runnable must be passed a function as its second argument`, and the `describe(title, { skip }, fn)` form silently mangles the suite rather than skipping it.

So I kept the conditional but made the skip explicit the Mocha way in 29880eb:

```js
const describeUnlessLmdb = process.env.HARPER_STORAGE_ENGINE === 'lmdb' ? describe.skip : describe;
```

That reports the suite as pending on lmdb instead of a silent early `return`. Thanks for the nudge toward making it clearer!

— Claude (Opus 4.7)

Issue #1107, opened 2026-06-03
[MCP/OpenAPI] Return envelope for search and list verbs


kylebernhardy:
## Context

Spun out of #1095 (MCP/OpenAPI shared descriptions). That issue ships `outputSchema` for the record-shaped verbs (`get`/`create`/`update`/`patch`/`delete`) where the return shape is trivially derivable from the table attributes. The `search_*` verb is intentionally deferred because its return shape is a *wrapped* envelope around the records, and the wrapper shape is an open design question.

This issue picks the wire-protocol envelope for `search_*` (and any future list/page verb) so that:
- MCP clients get a real `outputSchema` they can validate against
- OpenAPI emits a non-empty `responses[200]` schema
- Both consumers share the same shape, same source

## Why this is non-trivial

`outputSchema`/`response.schema` are compatibility commitments. Once shipped, MCP clients validate against the declared shape; renaming `cursor` → `nextCursor` is a breaking change. The envelope decision shouldn't be made inside a description-quality issue — it deserves its own design discussion with explicit review.

Today Harper's `search_*` returns an iterable that the MCP layer wraps before sending. The exact wire shape isn't widely documented; this issue should both *decide* the shape and *write down* what shipped.

## The decision

Pick one envelope (or design a new one). Candidate shapes:

### Option A — `{records, cursor}` (concise)

```json
{
  "records": [/* Product records */],
  "cursor": "opaque-pagination-token-or-null"
}
```

JSON Schema:
```json
{
  "type": "object",
  "properties": {
    "records": { "type": "array", "items": <ProductSchema> },
    "cursor": { "type": ["string", "null"], "description": "Opaque pagination cursor; null when no more pages." }
  },
  "required": ["records", "cursor"],
  "additionalProperties": false
}
```

Pros: minimal, matches existing internal conventions if `records`/`cursor` is what we already use.
Cons: no place for total counts, page metadata, or warnings.

### Option B — `{data, nextCursor}` (REST API convention)

Matches common REST API patterns (GitHub, Stripe-ish).

```json
{
  "data": [/* records */],
  "nextCursor": "opaque-token-or-null"
}
```

Pros: familiar shape to API consumers; `data` is the most common modern convention.
Cons: less concise; deviates from Harper's internal `records` vocabulary if that's what's there today.

### Option C — Richer envelope with optional metadata

```json
{
  "records": [/* records */],
  "pagination": {
    "cursor": "opaque-token-or-null",
    "hasMore": true,
[...]

Issue #1095, opened 2026-06-02
[MCP] Quality descriptions and richer tool-payload metadata for LLM tool selection


kylebernhardy:
## Context

MCP v1 shipped via the umbrella #465 (PRs landed through #888). Every generated tool and resource carries a `description`, but the descriptions are generic templates:

| Surface | Today's description |
|---|---|
| Operations | `"Harper operation 'add_user'. Arguments forwarded as-is; the server validates and returns a structured error on rejection."` |
| Application verb tools (table-backed) | `"get on resource '/Product' (table Product). Runtime RBAC (allowGet) enforces per-record access at call time."` |
| Application verb tools (programmatic Resource) | Same generic template, *and* derived input schema is skeletal (no `Table.attributes` to draw from). No metadata hook to override either. |
| Custom `mcpTools` | Author-supplied, with a generic fallback if omitted |
| `harper://*` | Hand-curated (already good) |

The problem: an LLM picking between `get_Product`, `get_Order`, `get_Customer` sees three near-identical descriptions that vary only by entity name. That's the dominant signal MCP clients feed to the model during tool selection — and it's currently mostly boilerplate.

A separate but related gap: **`inputSchema.properties[*].description`** is empty for user-defined attributes. The LLM is told that `search_Product` accepts a `conditions[].attribute` field but is not told which attributes exist or what they represent. Both gaps share the same fix path (GraphQL docstrings), so this issue addresses them together.

**Per @kriszyp's [review comment](https://github.com/HarperFast/harper/issues/1095#issuecomment-4607182451), the same descriptive data should also reach Harper's OpenAPI surface.** Today `resources/openApi.ts:179-248` hardcodes path-level descriptions and `:102-139` emits no per-property descriptions at all. Once the schema layer carries `description` for tables and attributes (Surface 2 below), feeding both consumers is two 1-line additions — the alternative (parallel namespaces, one per consumer) wastes the abstraction. This issue therefore extends to OpenAPI consumption end-to-end, not just MCP.

## Goals

1. Every Harper-shipped operation tool carries a hand-authored, LLM-grade description (MCP).
2. Application verb tools carry descriptions and per-attribute property descriptions derived from GraphQL schema docstrings — without component authors having to write them twice. Same docstrings drive OpenAPI's path-level and per-property descriptions.
3. Component-author custom MCP tools are nudged (but not blocked) to ship a description.
[...]

Issue #878, opened 2026-05-29
MCP: refactor operations tool registration so input schemas are introspectable/validatable JSON Schema


kriszyp:
## Context

Follow-up from review of #856 (MCP v1 tools surface).

The operations-profile MCP tools currently get their `inputSchema` from **hand-curated** JSON Schemas in `components/mcp/tools/schemas/operations.ts`, with a `PERMISSIVE_SCHEMA` (`{ type: 'object', additionalProperties: true }`) fallback for any opted-in operation that lacks an entry. This was a deliberate v1 choice — Harper's server-side validators are Joi, which doesn't round-trip cleanly to JSON Schema (documented in the file header).

## Problem

- **Drift risk.** The curated schemas are decoupled from the Joi validators that actually enforce input at dispatch time. As operations evolve, the MCP-advertised schema and the real validation can silently diverge.
- **Coverage gaps.** Operations without a curated entry fall back to a fully permissive schema, giving the LLM no guidance and deferring all feedback to runtime errors.
- **No introspection guarantee.** Nothing asserts the curated schemas are themselves valid JSON Schema (draft-07-ish, per the MCP spec).

## Proposed direction

Refactor operations API registration so the input schema is derived from (or validated against) a single source of truth, and can be introspected as valid JSON Schema. Options to explore:

- Generate JSON Schema from the Joi validators (e.g. a Joi→JSON-Schema conversion step), or
- Migrate operation validation to a schema representation that natively serializes to JSON Schema, or
- At minimum, add a build/test-time assertion that every entry in `OPERATION_INPUT_SCHEMAS` is valid JSON Schema and that curated schemas stay in sync with the server-side validators.

## Acceptance

- Operations tool `inputSchema`s are provably valid JSON Schema (asserted in CI).
- A single source of truth (or an enforced sync check) between MCP-advertised schema and server-side validation.

_Filed by Claude on behalf of @kriszyp during #856 review._

PR #1167, opened 2026-06-08
feat(mcp/openapi): quality descriptions and richer tool-payload metadata (#1095)

kylebernhardy:
## Summary

Implements [#1095](https://github.com/HarperFast/harper/issues/1095). Adds quality descriptions and richer tool-payload metadata for MCP tool selection, sourced from a single set of class-level statics that drive both the MCP tool descriptors and the OpenAPI document.

- **GraphQL schema** picks up `"""docstrings"""` on `type Foo @table @export` definitions and field definitions. New `@hidden` directive at field and type level suppresses sensitive metadata from the introspectable surfaces.
- **Class-level statics** become the canonical metadata API on Resource/Table classes — `static description`, `static properties` (Record<string, JsonSchemaFragment>), `static outputSchemas`, `static hidden`, `static mcp.annotations`. Auto-derived for `@table @export` Resources, declared directly for programmatic ones. Inherits naturally via class extension.
- **MCP application profile** swaps the mechanical `"get on resource '/X'"` template for a `verbDescription` composer with per-verb sentences and a class-doc prefix. Adds `outputSchema` to the five cheap verbs (get/create/update/patch/delete); search_* deliberately omits (envelope shape deferred to #1107). `update_*` carries `idempotentHint: true` (PUT semantics). Custom `mcpTools` without description/inputSchema get a warn-once nudge.
- **MCP operations profile** ships a hand-authored catalog of ~60 descriptions, every entry citing the handler `file:line` in a preceding comment so reviewers can verify in one click. New CI lint asserts every DEFAULT_ALLOW expansion has an `OPERATION_INPUT_SCHEMAS` and `OPERATION_DESCRIPTIONS` entry. `IDEMPOTENT_OPERATIONS` set declared EMPTY by default (under-annotate before mis-annotate).
- **OpenAPI** consumes the same data: schema-level descriptions, per-property descriptions, path-level descriptions with the class docstring prepended; `@hidden` filters at both surfaces.

The catalog is explicitly transitional pending #878 (introspectable operations registry).

## Where to put attention

Three areas with lower confidence — please scrutinize:

1. **`idempotentHint: true` on `update_*`** — claims PUT semantics produce the same observable outcome on repeat call. If Harper's `put`/`update` semantics differ from the JSON-Schema-shaped overwrite I'm assuming (e.g., partial-merge behavior), drop the hint or move it to a curated allowlist.
2. **`deriveDeleteOutputSchema` shape** — synthesizes `{ deleted: true, <pk> }` as the delete return envelope. Worth verifying against Harper's actual delete-handler return value before merging. If non-standard, the deriver should emit `{ type: 'object' }` instead of a contract Harper doesn't keep.
3. **`Attribute.description` / `Attribute.hidden` persistence** — `databases.ts:580` persists the primary attribute via `attributesDbi.putSync(key, primaryAttribute)`. If that path serializes the whole Attribute, the new fields land in older-replica DBI entries on a forward replication. Verified that schema-change detection at `databases.ts:1145-1162` doesn't compare description/hidden (so docstring edits don't trigger reindex), and that `schemaDescribe.ts:149-167` uses an explicit field list (so `describe_table` output is unchanged), but the on-disk persistence side is worth a second look.

## Known follow-up (out of scope here)

`resources/openApi.ts:includeDefinitionInSchema` silently drops primitive-array fields in nested complex types (the `[String]` case inside a complex type whose `.elements` lacks `.properties`). Same gap exists pre-refactor — discovered during cross-model review. Should be a separate small fix.

## Internal commit structure

7 commits matching the implementation phases. Reviewed as one whole; structured for `git bisect`:

1. Foundation: type widening
2. GraphQL parser: docstrings + @hidden + properties Record
3. MCP deriver: per-attribute descriptions + outputSchema + @hidden filter
4. MCP application: verbDescription + outputSchema wiring + warn-once + @hidden
5. MCP operations: hand-authored catalog + idempotentHint set + coverage lint
6. MCP resources: description prefix + @hidden filter
7. OpenAPI consumption: descriptions + ResourceSchema + @hidden + array-fix

## Docs PR

Companion docs PR at HarperFast/documentation: `docs-mcp-metadata-1095`. Covers the `@hidden` directive, the `"""docstring"""` convention, the new `static description`/`properties`/`outputSchemas`/`mcp`/`hidden` surface, an MCP tool-metadata sourcing reference, and a how-to with before/after MCP output. Should ride with this PR.

## Test status

- 159 focused tests pass (MCP toolRegistry, derive, application, operations, resources, graphqlMetadata, openApi, update-schema, embedDirective)
- Build, format, lint all clean
- `test:unit:main` not run locally (DB lock held by an unrelated Harper instance on this host); CI should cover

## Cross-model review

- **Gemini** review caught two real issues, both fixed before this PR opened: (a) `for...of` over the new Record on top-level typeDefs in `openApi.ts:includeDefinitionInSchema` (would throw on schemas with nested complex types — fixed by coalescing `def.attributes ?? def.properties`); (b) array-element preservation missing in the properties projection (`{ type: 'array' }` without `items` — fixed via a shared `attributeToFragment` helper in `jsonSchemaTypes.ts`, with regression tests for both).
- **Codex CLI** not available on the dev environment; cross-model coverage is Gemini-only.

🤖 Generated with [Claude Code](https://claude.com/claude-code) using model claude-opus-4-7.


kriszyp (approved):
Love it, this is a great alignment!

Issue #1170, opened 2026-06-08
Replication: out-of-order full update can revert a newer record (non-convergence)


kriszyp:
## Summary

During replication, when full-record updates (`put`, `fullUpdate=true`) to the **same key** arrive and commit **out of order** on a receiving node, an *older* update can overwrite a *newer* one — leaving the cluster permanently non-convergent.

## Mechanism

In `resources/Table.ts`, the out-of-order resequencing branch (`if (precedesExisting <= 0)`) handles updates whose `txnTime` precedes the existing record version. When `audit` is on, it walks the audit history to reconstruct succeeding updates. If that walk finds **no** succeeding updates to replay (`succeedingUpdates` empty, `incrementalUpdateToApply` stays undefined), control falls through to the shared commit line:

```ts
if (fullUpdate && !incrementalUpdateToApply) recordToStore = recordUpdate;
```

`recordUpdate` here is the **older** full record. It is stored on top of the newer state, reverting it. The analogous **no-audit** branch already guards this case (`else if (fullUpdate) { write.skipped = true; return writeCommit(false); }`), but the **with-audit** branch is missing the equivalent guard.

`precedesExistingVersion` correctly identified the incoming update as older (`<= 0`), so the safe behavior is to **skip** it, not to apply it.

## Trigger / impact

Reproduces when one origin issues a rapid burst of full-record `put`s to a single replicated key and a loaded peer commits them out of order (optimistic-lock retries reorder concurrent same-key writes). Observed via the deployment-tracking feature, where `hdb_deployment` rows get ~10 full puts within ~400ms: a peer's row gets stuck at a non-terminal `status` permanently while the origin and other peers converge to the terminal state. Timing-dependent, hence intermittent.

## Proposed fix

In the with-audit out-of-order branch, after the succeeding-updates replay loop, when `fullUpdate && !incrementalUpdateToApply` (no succeeding updates were reconstructable), skip the write — mirroring the existing no-audit `fullUpdate` guard:

```ts
if (fullUpdate && !incrementalUpdateToApply) {
    write.skipped = true;
    return writeCommit(false);
}
```

Needs the full replication/CRDT test suite plus a regression test that drives out-of-order same-key full puts and asserts the newest survives.

## Evidence

[...]

Issue #291, opened 2026-03-26
Disk object empty from system_information operations API request


pbrumblay:
### Bug Summary

disk block is missing from system_information operation

### Reproduction Repository (if applicable)

https://github.com/HarperFast/harper/ versions 4.7.19 and 5.0.0-beta.6

### Steps to Reproduce

Reproducible on both harper 4.7.19 (fabric hosted) and 5.0.0-beta.6. I'm attempting to use the `system_information` operation: https://docs.harperdb.io/docs/developers/operations-api/system-operations#system-information

Steps:
1. Fresh install of 5.0.0-beta.6 using `npm install -g harper@5.0.0-beta.6`
2. Set env vars:
    ````
    export HARPER_OPERATIONS_URL=http://localhost:9925/
    export HARPER_USERNAME=HDB_ADMIN
    export HARPER_PASSWORD=<something>
    AUTH="Authorization: Basic $(printf '%s' "$HARPER_USERNAME:$HARPER_PASSWORD" | base64)"
    ````
3. Start harper using CLI, answer initial install questions
4. Use studio to set password for HDB_ADMIN to <something>
5. Use curl to fetch system_information: 

    ````
    curl -sS -X POST "$HARPER_OPERATIONS_URL" \
      -H "$AUTH" \
      -H 'Content-Type: application/json' \
      -d '{"operation":"system_information","attributes":["table_size","metrics","memory","disk"]}' | jq
    ````

### Expected Behavior

`disk`, `table_size`, `metrics` should all have default (and if possible) useful information. For example, a disk percentage used would be really really nice to head off disk capacity issues. 

### Actual Behavior

New install 5.0.0-beta.6. Note that `disk`, `table_size` and `metrics` are empty.

```
{
      "memory": {
        "total": 34359738368,
        "free": 167198720,
        "used": 34192539648,
        "active": 7666941952,
        "available": 26692796416,
        "reclaimable": 7621804032,
        "swaptotal": 17179869184,
        "swapused": 16273176002.56,
        "swapfree": 906693181.44,
        "writeback": null,
        "dirty": null,
        "rss": 109297664,
        "heapTotal": 59506688,
        "heapUsed": 56448280,
        "external": 38544140,
        "arrayBuffers": 17485360
      },
      "disk": {},
      "table_size": [],
      "metrics": {}
    }
```

4.7.19 snippet, `metrics` has a structure in it but `disk` and `table_size` are empty.

```
  "disk": {},
  "table_size": [],
[...]

PR #1171, opened 2026-06-08
fix(deploy): converge hdb_deployment terminal status on peers

kriszyp:
## Summary

The deploy lifecycle writes the `hdb_deployment` row ~10 times within a few hundred ms (create, payload ingest, phase flushes, per-peer results, `finish()`). These replicate to peers; on a loaded peer the rapid same-key writes can **commit out of order**, where an older full update reverts the terminal `success` write — the peer row stays stuck at `replicating` and never converges.

This adds `DeploymentRecorder.seal()`, called before the replicate phase: `scheduleFlush()` stops issuing puts (state accumulates in memory) and `finish()` performs a single terminal write, isolating it from the concurrent same-key burst so the receiver converges.

## Purpose

Fixes the intermittently-failing harper-pro integration test `deployTrackingReplication.test.mjs` ("hdb_deployment row replicates to peers" — `saw replicating`). It is **pre-existing on harper-pro `main`** (failed on Node v22/v24 shard 4; v26 happened to pass), not introduced by any feature branch — it's a timing-dependent replication race.

This is a **mitigation**. The underlying defect is in core replication (an out-of-order full update can revert a newer record): harperdb/harper#1170, fixed separately.

## Where to look

- `components/deploymentRecorder.ts` — `sealed` flag + `seal()`; `scheduleFlush()` early-returns when sealed.
- `components/operations.js` — `recorder?.seal()` before the replicate phase.

## Tradeoff (deliberate)

The origin's `get_deployment` **polling** view no longer shows the transient `replicating` status or incremental `peer_results` during the final phase. Live SSE tailing is unaffected (the emitter still emits in real time). Once #1170 lands, this seal can be removed to restore incremental `peer_results` persistence.

## Verification

- Recorder unit tests pass (17, incl. 2 new for `seal()`); `test:unit:main` shows no regressions.
- The actual multi-node failing test is verified via a companion harper-pro PR that bumps the `core` pointer to this branch (the test only runs in harper-pro CI).

Cross-model review: Codex (no issues) + Gemini/AntiGravity (correct; two low-severity suggestions — guard the *unused* `transitionPhase`, reset `dirty` in `finish()` — intentionally skipped as YAGNI since `finished` already gates further writes).

🤖 Generated by Claude (Opus 4.x) per the Harper engineering DLC.

### DIFF
--- a/components/deploymentRecorder.ts
+++ b/components/deploymentRecorder.ts
@@ -72,6 +72,7 @@ export class DeploymentRecorder {
 	private unsubscribe: (() => void) | null = null;
 	private pendingPut: Promise<void> | null = null;
 	private dirty = false;
+	private sealed = false;
 
 	private constructor(deploymentId: string, initial: Record<string, any>) {
 		this.deploymentId = deploymentId;
@@ -151,6 +152,12 @@ export class DeploymentRecorder {
 	// the record dirty; the chained continuation issues a follow-up put once the prior one
 	// settles. This keeps event_log writes O(1) puts per burst rather than O(N) per event.
 	private scheduleFlush(): void {
+		if (this.sealed) {
+			// Sealed: accumulate state in memory but don't write. finish() does the single
+			// terminal write. See seal() for why. The emitter still emits live SSE events.
+			this.dirty = true;
+			return;
+		}
 		if (this.pendingPut) {
 			this.dirty = true;
 			return;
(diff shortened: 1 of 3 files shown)


kriszyp:
Marking ready for review. ✅

**Verification:** the actual failing test (`deployTrackingReplication.test.mjs` — "hdb_deployment row replicates to peers") is **green on all three Node versions** via the companion harper-pro PR ([#291](https://github.com/HarperFast/harper-pro/pull/291)) that pins `core` to this branch.

**Note on red CI:** the only remaining failure is `Unit Test` → *"Test logRotator module > Test that log file is rotated if log has exceeded max size"*. This is a **pre-existing flake on the base commit**, not from this change — it reproduces identically on #1172 (which only touches `resources/Table.ts` and doesn't go near logging), and the deployment-recorder change touches neither logging nor rotation. The other integration/Bun blips cleared on re-run.

— Claude (Opus 4.x)

cb1kenobi (approved):

PR #1183, opened 2026-06-09
fix(databases): sync non-indexed attributes on worker-restart schema reload (RE-7)

kriszyp:
## Summary

- `describe_database` returned stale schema attributes after a hot-reload worker restart (new fields missing until full kill+restart)
- Root cause: `initStores()` in `resources/databases.ts` correctly handled the first-startup path (all attributes passed to `makeTable()`), but the existing-table update path — triggered after `resetDatabases()` on a worker restart — only merged back **indexed and primary-key** attributes; plain fields like `name: String` and `age: Int` were silently dropped
- Fix adds an `else if (!attribute.isPrimaryKey)` branch so non-indexed attributes are also updated/inserted in `existingAttributes`, plus a corresponding removal path for attributes no longer in the schema

## Test plan

- [ ] `unitTests/resources/schemaMigrationFragility.test.js` — two new regression tests added (both pass)
- [ ] Manual: `harper dev .` → edit `schema.graphql` to add fields → verify `describe_database` shows new fields without full restart

---

### DIFF
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -629,6 +629,7 @@ function initStores(
 					const existingIdx = existingAttributes.findIndex((ea) => ea.name === attribute.attribute);
 					if (existingIdx >= 0) {
 						existingAttributes.splice(existingIdx, 1, attribute);
+						attributesUpdated = true;
 					} else {
 						existingAttributes.push(attribute);
 						attributesUpdated = true;
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -638,6 +638,10 @@ function initStores(
 				logger.error(`Error trying to update attribute`, attribute, existingAttributes, indices, error);
 			}
 		}
+		// Collect removals first; splicing while iterating `existingAttributes` skips adjacent
+		// elements, which would silently leave stale fields behind when two or more were dropped
+		// in the same reload.
+		const toRemove = [];
 		for (const existingAttribute of existingAttributes) {
 			const attribute = attributes.find((attribute) => attribute.name === existingAttribute.name);
 			if (!attribute) {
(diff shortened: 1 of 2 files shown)


resources/databases.ts
```diff
@@ -620,6 +620,19 @@ function initStores(
 					if (existingAttribute) existingAttributes.splice(existingAttributes.indexOf(existingAttribute), 1, attribute);
 					else existingAttributes.push(attribute);
 					attributesUpdated = true;
+				} else if (!attribute.isPrimaryKey) {
+					// Non-indexed, non-primary-key attributes (e.g. plain schema fields like `name: String`)
+					// must also be kept in sync so that describe_database reflects schema changes after a
+					// hot-reload / worker restart. Without this, resetDatabases() re-reads these attributes
+					// from attributesDbi but never merges them back into table.attributes — causing stale
+					// schema metadata until a full kill+restart. (RE-7)
+					const existingIdx = existingAttributes.findIndex((ea) => ea.name === attribute.name);
```
cb1kenobi: Does `attribute.name` always exist? I'm looking at the first conditional of this above this block and it suggests that `attribute.name` can be falsey. Maybe this should be checking `attribute.attribute`?

```suggestion
					const existingIdx = existingAttributes.findIndex((ea) => ea.name === attribute.attribute);
```

cb1kenobi (approved):

kriszyp:
Thanks for the sharp reviews — both issues addressed in 568f3ea:

**`attribute.attribute` vs `attribute.name` (cb1kenobi):** Accepted — changed to `attribute.attribute` in the `findIndex`. You're right that the condition in the branch above implies `attribute.name` can be falsey; using `attribute.attribute` is the safer choice.

**Removal test not exercising the removal path (claude):** Fixed. After `table()` sets attributesDbi to `[id, name]`, we now manually restore in-memory `Table.attributes` back to `[id, name, breed, age]` before calling `resetDatabases()` — same stale-state pattern as the INSERT test. The removal loop in `initStores()` now actually needs to drop `breed` and `age`.

— Claude

resources/databases.ts
```diff
@@ -647,6 +660,10 @@ function initStores(
 					// we only remove attributes if they were indexed, in order to support dropAttribute that removes dynamic indexed attributes
 					existingAttributes.splice(existingAttributes.indexOf(existingAttribute), 1);
 					attributesUpdated = true;
+				} else if (!existingAttribute.isPrimaryKey) {
```
heskew, line 667: Codex review: this removal path also catches runtime-only relationship/computed attributes. `table()` intentionally does not persist relationship attrs to `attributesDbi` (`continue` in the attribute persistence loop), so after `resetDatabases()` the persisted `attributes` list lacks entries like `related` while `existingAttributes` still has the valid runtime relationship attribute. This branch removes it and `updatedAttributes()` drops the resolver/search support; the current CI failures (`related.name is not a defined attribute`, etc.) line up with that. Could we exclude `existingAttribute.relationship || existingAttribute.computed` here, or otherwise distinguish persisted fields from runtime-only attributes? Also worth avoiding `splice` while iterating this same array, since adjacent removed fields like the test's `breed`/`age` can skip one another.

heskew (commented):
Codex review: found one blocker in the reload removal path. Leaving it inline.

kriszyp:
Both issues addressed in `e14d51636`:

**Runtime-only relationship attrs (heskew):** Added `if (existingAttribute.relationship) continue;` to the removal branch — directly mirrors the symmetric skip in the persistence loop at databases.ts:1138. Computed attrs are still considered for removal since they ARE persisted (only `relationship` triggers the persistence `continue`). Verified with a new unit test that injects a relationship attr and asserts it survives `resetDatabases()` — the test fails without the fix.

**Splice-while-iterating:** Switched to a `toRemove` collection pattern that applies the splices after the iteration completes, so adjacent removals no longer skip each other. The existing removal test already covers this (removes adjacent `breed` and `age`); extended its comment to call out the intent.

— Claude (Opus 4.7)

kriszyp:
Fixed in f46ea53e — `attributesUpdated = true` added to the `existingIdx >= 0` update path, bringing it into parity with the add path.

— Claude

PR #1185, opened 2026-06-09
fix(installer): warn when DEFAULTS_MODE is set to an unrecognized value (RE-6)

kriszyp:
## Summary

- Setting `DEFAULTS_MODE=1` (or any value besides `"dev"` or `"prod"`) silently fell through to prod defaults — HTTPS-only, auth required — with no feedback
- Adds a yellow `Warning:` line to stderr and the log before continuing, so users know why dev mode was not applied and what the valid values are

## Behavior

Before:
```
DEFAULTS_MODE=1 harper install ...
# → installs with prod defaults, HTTPS on 9926, auth required — no indication anything was wrong
```

After:
```
DEFAULTS_MODE=1 harper install ...
# Warning: DEFAULTS_MODE value "1" is not recognized (expected "dev" or "prod"). Using prod defaults.
# → installs with prod defaults (same behavior, but now visible)
```

## Test plan

- [ ] `DEFAULTS_MODE=1 harper install ...` → warning appears on stderr, install completes with prod defaults
- [ ] `DEFAULTS_MODE=dev harper install ...` → no warning, dev defaults applied as before
- [ ] `DEFAULTS_MODE=prod harper install ...` → no warning, prod defaults applied as before

---

### DIFF
--- a/utility/install/installer.ts
+++ b/utility/install/installer.ts
@@ -527,6 +527,12 @@ async function createConfigFile(installParams) {
 			if (args[cfg.toLowerCase()] === undefined) args[cfg] = DEV_MODE_CONFIG[cfg];
 		}
 	} else {
+		const defaultsMode = installParams[hdbTerms.INSTALL_PROMPTS.DEFAULTS_MODE];
+		if (defaultsMode !== undefined && defaultsMode !== 'prod') {
+			const warn = `DEFAULTS_MODE value "${defaultsMode}" is not recognized (expected "dev" or "prod"). Using prod defaults.`;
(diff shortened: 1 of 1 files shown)


kriszyp:
The Gemini note is a false alarm: the outer `if (...DEFAULTS_MODE] === 'dev')` at line 501 catches the `'dev'` case — the `else` branch (where the warning lives) is only reached when the value is **not** `'dev'`. So `'dev'` can never trigger the warning. No change needed here.

cb1kenobi (approved):

Issue #1235, opened 2026-06-10
5.1 GA readiness: models/vector subsystem audit findings


heskew:
A multipass deep review (4 domains × 2 independent runs) plus an API-surface audit of the merged models/AI subsystem surfaced the following. Consolidated as one issue to keep noise down; fixes land as two batch PRs, and the design items need discussion before 5.1 hardens the API.

## Gating for non-beta 5.1 — PR 1: HNSW/vector data integrity (merged: #1234)

Index corruption persists in stored graphs, so these can't be patched after the fact:

- [x] Non-finite vectors poison ranking — no NaN/Infinity guard at index time (the `embedHook.ts:131` comment claims one exists); `bisectInsert(arr, NaN)` returns 0, so a poisoned node becomes rank 1 of every search that reaches it. (`HierarchicalNavigableSmallWorld.ts:100-109`)
- [x] Deleting the entry-point node can blank vector search — the replacement scan runs without the transaction and without excluding the node being deleted; the next insert orphans the surviving graph. (`HierarchicalNavigableSmallWorld.ts:421-447`)
- [x] Update-cleanup sweep erodes reverse edges — removes a neighbor's reverse edge at all levels `0..l` instead of only the level where the old connection existed. With `@embed`, vector updates are the common write path; records silently vanish from search over time. (`HierarchicalNavigableSmallWorld.ts:451-476`)
- [x] Backfill resume is non-idempotent for HNSW — checkpoint resume re-feeds records with no `existingVector`, so existing nodes get new random levels and overwritten connection lists. (`databases.ts:1303-1326`)
- [x] Structural option change resumes from the old checkpoint — interrupted backfill + option change yields one index built under two configs. (`databases.ts:1187,1290-1299`)
- [x] Small: deletes leak the `safeKey→nodeId` mapping; `le` threshold uses strict `<` (and a threshold of `0` is skipped as falsy).

## Gating for non-beta 5.1 — PR 2: models backend/analytics correctness (in review: #1236)

- [ ] `hdb_model_calls` declares nine `indexed: true` attributes that writes never maintain — attribute-filtered billing/usage queries silently return zero rows. (`analyticsTable.ts:57-73,152`)
- [ ] Bedrock can't invoke inference-profile model IDs (`us.anthropic.claude-…` → family `'us'` → throw); `amazon.nova-*` gets a malformed Titan-shaped body. (`bedrock/index.ts:291-299`)
[...]

PR #1236, opened 2026-06-10
fix(models): backend and analytics correctness batch for 5.1 GA

heskew:
## Summary

Five verified correctness fixes to the models subsystem from the 5.1 GA readiness audit — relates to #1235 (PR 2 of 2, companion to #1234).

1. **`hdb_model_calls` phantom indexes** — dropped `indexed: true` from all nine non-PK attributes. Writes go through `primaryStore.put`, which bypasses index maintenance, so the declared indexes were permanently empty and any attribute-filtered billing/usage query silently returned zero rows. Matches the `hdb_raw_analytics` pattern. The attribute list is now an exported const so the schema test asserts structurally.
2. **Bedrock inference-profile model IDs** — `familyOf()` walks the dot-segments until a known family is found, so `us.anthropic.claude-…` / `eu.meta.…` / `global.…` dispatch correctly (previously: family `'us'` → unknown → throw, breaking effectively all current Claude-on-Bedrock). `amazon.nova-*` now throws a descriptive "not yet supported" error instead of being sent a malformed Titan-shaped body.
3. **Abort gate at tool dispatch** — `runSingleToolCall()` checks the composed signal before invoking a handler, so side-effecting tool handlers never start after the caller has aborted (closes the serial-dispatch gap too).
4. **OpenAI `max_completion_tokens`** — sent when `baseUrl` resolves to native `api.openai.com` (current reasoning models reject `max_tokens`); custom baseUrls keep `max_tokens` since OpenAI-compatible shims may not know the newer param. Detection parses the URL hostname (port-suffix and spoofed-host safe).
5. **Upstream memory caps** — `readBoundedJson()` bounds non-streaming bodies (64 MiB success / 256 KiB error — previously `res.json()` buffered unbounded and the 500-char error cap only trimmed *after* parsing), wraps stream errors in the backend error class, and cancels the body on breach. The streaming tool-call accumulator maps are now capped at 128 entries and 8 MiB total argument chars across entries (per-entry 1 MiB cap alone allowed ~128 MiB) in all three streaming parsers.

## Where to put attention

- **The native-OpenAI detection heuristic** (fix 4): hostname equality with `api.openai.com`. Proxies/gateways fronting OpenAI (Cloudflare AI Gateway, Helicone) will get `max_tokens` and fail against reasoning models — by design for now (no reliable detection); flagged by the cross-model review as the residual case. A per-entry config override would be the escape hatch if users hit it.
- **The Nova fail-loud** (fix 2) is a behavior change from "malformed request, upstream error" to "immediate descriptive error" — strictly better, but it is the one place this PR rejects something it previously attempted.
- Cap constants (64 MiB / 256 KiB / 128 entries / 8 MiB) are judgment calls sized generously above real-world responses; sanity-check the numbers.

Tests: 75 passing across the targeted unit files (includes new cases for inference-profile dispatch, Nova rejection, pre-aborted dispatch, native-vs-shim token param, bounded-reader limits, and accumulator caps); broader models+components suites green except one pre-existing EMFILE failure confirmed on main.

Cross-model review: Gemini findings adjudicated and applied in the second commit (structural test, URL-parsed detection, bounded-reader hardening, total-arg cap). Codex leg unavailable (weekly session limit). Generated by an LLM (Claude, Fable 5).

### DIFF
--- a/components/anthropic/index.ts
+++ b/components/anthropic/index.ts
@@ -25,8 +25,10 @@ import { setGenerative } from '../../resources/models/backendRegistry.ts';
 import {
 	assignFiniteTokenCount,
 	composeSignal,
+	MAX_ERROR_BODY_BYTES,
 	normalizeOrigin,
 	parseJsonResponse,
+	readBoundedJson,
 	requireCredential,
 	requireModel,
 } from '../../resources/models/backendHelpers.ts';
@@ -59,6 +61,10 @@ const MAX_SSE_BUFFER_CHARS = 1 << 20;
 const MAX_TOOL_CALL_ARGS_CHARS = 1 << 20;
 // Cap for upstream `error.message` we surface to operators.
 const MAX_UPSTREAM_ERROR_MESSAGE_CHARS = 500;
+// Maximum number of distinct tool-call accumulator entries. Anthropic keys by
+// `index` from content_block_start events; a hostile upstream can open unbounded
+// entries if content_block_stop never arrives for earlier indices.
+const MAX_TOOL_CALL_ACCUMULATOR_ENTRIES = 128;
 
 const log = harperLogger.forComponent('anthropic').conditional;
 
@@ -165,6 +171,13 @@ export class AnthropicBackend implements ModelBackend {
 				event.index !== undefined &&
 				event.content_block?.type === 'tool_use'
 			) {
+				// Cap total accumulator entries: `index` is upstream-controlled and
+				// a content_block_stop may never arrive, leaking entries indefinitely.
+				if (toolBuf.size >= MAX_TOOL_CALL_ACCUMULATOR_ENTRIES) {
+					throw new AnthropicBackendError(
+						`Anthropic tool-call accumulator exceeded ${MAX_TOOL_CALL_ACCUMULATOR_ENTRIES} distinct tool-call entries`
+					);
+				}
 				toolBuf.set(event.index, {
 					id: event.content_block.id,
 					name: event.content_block.name,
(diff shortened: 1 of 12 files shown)


heskew:
Codex review leg completed (was quota-blocked at PR open). One finding, adjudicated real and addressed in the latest commit: the 64 MiB success-body cap could reject a maximal *legal* OpenAI embedding batch (2048 inputs × 3072 dims ≈ 125–190 MiB of JSON) — raised to 256 MiB, which still bounds a runaway upstream. No other blocking issues found.

---

kriszyp (approved):
**Approve** — five well-scoped, correctly-fixed bugs with solid new test coverage. The inline comments explaining each *why* are excellent, and the structural schema-test approach for the phantom-index fix is a nice touch. Two small follow-ups before merge, neither blocking the code:

**1. PR description vs. code: body-cap value (doc nit).** The summary (fix #5 and "Where to put attention") says *"64 MiB success / 256 KiB error"*, but the code is `MAX_RESPONSE_BODY_BYTES = 256 << 20` (256 MiB), justified by the max legal embedding batch in the source comment. Looks like the cap was raised during the cross-model-review revision but the description didn't follow. Worth syncing the number, since that's exactly the constant the description asks reviewers to sanity-check.

**2. Azure OpenAI is a false negative in the native-OpenAI heuristic (worth a comment, not a blocker).** The known-limitation note calls out gateways (Cloudflare AI Gateway, Helicone), but the more common real case is **Azure OpenAI** (`*.openai.azure.com`), which also requires `max_completion_tokens` for o-series / reasoning deployments. Under `hostname === 'api.openai.com'`, Azure users hit the exact `max_tokens` → 400 this fix resolves. The per-entry config-override escape hatch you mention would cover it — I'd just name Azure explicitly in the limitation note since it's a first-class deployment target, not a niche proxy.

Minor, non-blocking observation: the 256 MiB cap bounds the *input* bytes, but `readBoundedJson` transiently holds chunks + merged buffer + decoded string + parsed object, so a single max-legal embedding response can peak ~1 GiB, and the cap is per-call rather than global. Fine to ship for the legal-max case — just noting it for future reference.

Nice work on this batch — the failure-mode reasoning on each fix made it easy to follow.

— Reviewed by Claude (Opus 4.8)

Issue #1163, opened 2026-06-08
Stale shared-structures: nodes silently return empty/partial query results without error


kriszyp:
## Summary

Nodes affected by stale msgpackr shared-structures silently return empty or incomplete query results — data present cluster-wide comes back as empty on the affected node — without surfacing any error to the caller.

## Symptoms

- Queries that should return records return empty results on affected nodes
- Log shows `Could not find typed structure 1` (msgpackr decode failures)
- Symptom clears on process restart
- Other nodes in the cluster return the same queries correctly

## Distinction from related issues

This is distinct from the *Unknown constant* decode error (PR #1154) where records become permanently unreadable. Here records are intact but queries silently return partial/empty results due to structure decode failures — the failure mode is invisible to callers.

## Expected fix

The msgpackr shared-structures fix shipping in 5.0.28 is expected to resolve this. Filing to track confirmation post-rollout.

## Reporter

Joshua Johnson

kriszyp:
### Field corroboration from a production cluster, harper-pro 5.0.28 — 2026-06-08

Active right now on a production cluster, and it's the cause of customer-reported "missing results" (some `Table.search()` calls silently return empty for records that exist — matches this issue's title exactly).

**Error (live, ongoing):**
```
[http/N] [error]: Error decoding record Error: Could not find typed structure 1
```
On the worst node (region `jp-osa-1`) these run continuously — 66 occurrences, earliest 20:34 GMT, latest 21:45 GMT (checked at 21:46). A record references shared structure #1 that the node's structures buffer doesn't contain, so it decodes to nothing and `search()` returns empty for it with no surfaced error to the caller.

**Cluster distribution (count of `Could not find typed structure` in current container logs):**

| node (region) | count | notes |
|---|---|---|
| jp-osa-1 | 66 | ongoing, latest 21:45 |
| br-gru-1 | 3 | |
| us-east-1 | 2 | |
| ap-northeast-1 | 2 | |
| us-central-1 | 1 | |

The affected nodes are exactly the ones returning partial/empty query results; a restart clears it temporarily (reloads structures), then it recurs.

[...]

PR #1237, opened 2026-06-10
fix(storage): surface missing shared-structure decode misses (#1163)

kriszyp:
## Summary
When a record references a shared structure that is **absent from this node's in-memory structures buffer**, msgpackr/structon already reload the structures from durable storage and retry; if the structure is still missing they throw a terminal error. `RecordEncoder.decode` previously caught *all* decode errors, logged at `error`, and returned `null` — silently dropping a genuinely-existing record from query results and laundering that emptiness into caches and downstream consumers.

This detects the two terminal missing-structure errors and routes them to a **distinct, non-fatal** path:
- an analytics counter (`decode-missing-structure`, keyed by store name), and
- a dedicated `warn` log,

…while still returning `null`. All other decode failures keep the existing `error` + `null` behavior.

## Purpose
Resolves the silent-result aspect of **#1163** ("nodes silently return empty/partial query results without error"). Field context: a production cluster returned partial query results to end users on affected replicas with nothing surfaced to the caller. This change makes the condition **observable and alertable** (metric + distinct warning) instead of silent. It does **not** recover the missing structure — that requires the replication-delivery path that ships the structure-buffer update to the replica (tracked separately); a node genuinely missing the structure still returns `null`, but now loudly.

## Why non-fatal (returns null, not throws)
An earlier iteration threw on a missing structure. Testing showed that regresses **database initialization**: the internal `__dbis__`/meta scan during `initStores` legitimately tolerates an undecodable record by returning `null`, and throwing there broke startup (cascading `before all` failures). Returning `null` keeps internal reads tolerant while still surfacing the condition.

## Where to focus review
- **Detection by error-message prefix** (`isMissingStructureError`): neither msgpackr nor structon throws a typed error, so we match the terminal strings `Could not find typed structure ` (structon `readStruct`) and `Record id is not defined for ` (msgpackr `createSecondByteReader`). Both fire only *after* the dependency's own on-miss reload fails. Is matching by prefix acceptable, and are these the complete set?
- **OPEN — classic low-id misses (unresolved cross-model disagreement).** Codex flagged that a *classic* structure miss for a **low record id** may instead surface as the ambiguous `Data read, but end of buffer not reached N` (not `Record id is not defined`), so it would route to the generic error path and not increment the metric — and small static tables (the typical #1163 victims) tend to have low ids. I deliberately did **not** match that message because it is ambiguous with genuine corruption (false positives would misclassify real corruption as a missing structure), and under this non-fatal design such a case is still logged (at `error`) and returns `null` — not silently lost, just not counted in the dedicated metric. Gemini's review reached the opposite conclusion (it considers the two prefixes the complete terminal miss signals). **@kriszyp** — as msgpackr author, your call on whether low-id classic misses produce that message in practice and whether it's worth matching.
- **LMDB store attribution:** `this.name` is only set on the RocksDB encoder, so the metric/warning fall back to `this.rootStore?.name` for LMDB (best-effort; may be the root db rather than the exact table).
- **`recordAction` on the decode path:** only invoked on the cold error branch; it's already used on the encode hot path in this file and batches asynchronously.

## Tests
`unitTests/resources/recordEncoder.test.js`: typed miss → `null` + distinct warn (not the error path); recovery decodes once the structure is present; `isMissingStructureError` covers both prefixes and rejects the ambiguous/`RangeError`/`undefined` cases; genuine corruption still → `error` + `null`.

---
🤖 Generated by Claude (Opus 4.7). The detection-prefix completeness (esp. the open classic low-id question above) is the main thing worth a careful human look.

### DIFF
--- a/resources/RecordEncoder.ts
+++ b/resources/RecordEncoder.ts
@@ -34,6 +34,28 @@ import { CONFIG_PARAMS } from '../utility/hdbTerms.ts';
 import * as envMngr from '../utility/environment/environmentManager.js';
 
 const StructonEncoder = createStructon(Encoder) as typeof Encoder;
+
+// Analytics counter incremented whenever a record cannot be decoded because its shared structure is
+// missing on this node (see HarperFast/harper#1163). Surfaces the otherwise-silent condition in
+// monitoring; the store name is passed as the metric path.
+const MISSING_STRUCTURE_METRIC = 'decode-missing-structure';
+
+// Terminal error messages msgpackr/structon throw when a record references a shared structure that
+// is not in this node's structures buffer. Both the typed (random-access) path (structon's
+// readStruct) and the classic path (msgpackr's createSecondByteReader) reload the structures from
+// durable storage and retry before throwing, so reaching one of these means the structure is
+// genuinely absent on this node — not merely stale in memory. Matched by message prefix because
+// neither dependency throws a typed error.
+const MISSING_TYPED_STRUCTURE_PREFIX = 'Could not find typed structure ';
+const MISSING_CLASSIC_STRUCTURE_PREFIX = 'Record id is not defined for ';
+
+export function isMissingStructureError(error: any): boolean {
+	const message = error?.message;
+	return (
+		typeof message === 'string' &&
+		(message.startsWith(MISSING_TYPED_STRUCTURE_PREFIX) || message.startsWith(MISSING_CLASSIC_STRUCTURE_PREFIX))
+	);
+}
 export type Entry = {
 	key: any;
 	value: any;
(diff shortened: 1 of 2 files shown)


kriszyp:
### Scope note for reviewers — this is fast-path-only observability, not the #1163 fix for the affected cluster

After live investigation, the production cluster that motivated #1163 runs the **standalone** structon decode path (msgpackr 1.12.1, `SUPPORTS_STRUCT_HOOKS` undefined — confirmed on the running encoders). On that path a missing typed structure surfaces as the **ambiguous** `Data read, but end of buffer not reached`, which this PR's detection **deliberately does not match** (it's indistinguishable from genuine corruption). The terminal strings this PR keys on (`Could not find typed structure`, `Record id is not defined for`) are the **fast-path** (msgpackr ≥ 2.0.1 / 5.1) signatures.

So please don't over-credit this PR for the live incident:
- It is **observability only** (counter + warning) — it does not recover a dropped record.
- On the cluster's standalone build it would **not even fire**.

Its real value is a **fast-path safety net** on main/5.1 (and any future fast-path 5.0 build): if a structure is genuinely durable-absent there, this surfaces it instead of silently returning empty.

The actual *read-recovery* for the standalone path (the cluster's failure mode) is HarperFast/structon#5 — reload typed structures on a decode miss, bringing them to parity with classic shared structures. And the operational remedy for Gen Digital remains `storage.randomAccessFields: false` (already in 5.0.30) + re-upserting the static rows, plus the replication-delivery fix.

Net: keep this as the fast-path observability net; the standalone recovery lives in structon#5.

— Claude (Opus 4.7)

resources/RecordEncoder.ts
```diff
@@ -34,6 +34,28 @@ import { CONFIG_PARAMS } from '../utility/hdbTerms.ts';
...
+const MISSING_STRUCTURE_METRIC = 'decode-missing-structure';
+
+// Terminal error messages msgpackr/structon throw when a record references a shared structure that
+// is not in this node's structures buffer. Both the typed (random-access) path (structon's
+// readStruct) and the classic path (msgpackr's createSecondByteReader) reload the structures from
+// durable storage and retry before throwing, so reaching one of these means the structure is
+// genuinely absent on this node — not merely stale in memory. Matched by message prefix because
+// neither dependency throws a typed error.
+const MISSING_TYPED_STRUCTURE_PREFIX = 'Could not find typed structure ';
+const MISSING_CLASSIC_STRUCTURE_PREFIX = 'Record id is not defined for ';
```
cb1kenobi, line 50: Just out of curiosity, why the global strings?

cb1kenobi (approved):

resources/RecordEncoder.ts
kriszyp: Do you mean vs being inside `isMissingStructureError`? (its not on the global scope)

Issue #1213, opened 2026-06-09
`$union` array directive for config env vars (compose-able arrays)


heskew:
## Summary

Add a `$union` directive for **array** values in the config environment variables (`HARPER_DEFAULT_CONFIG`, `HARPER_CONFIG`, `HARPER_SET_CONFIG`) so an array can be **composed** rather than replaced wholesale:

```jsonc
{ "tls": { "uses": { "$union": ["server", "operations-api"] } } }
```

`$union` guarantees the listed items are present in the target array while preserving any entries already there — the order-preserving union of (existing ∪ listed).

Parent: #1097

## Why this is the first piece

This is independently shippable and directly addresses the **motivating incident** in #1097 on its own: a platform layer setting `HARPER_SET_CONFIG` clobbered the app's `tls.uses` on every restart/upgrade. With `$union` honored under `HARPER_SET_CONFIG`, the platform can *guarantee* its required entries are present without dropping the app's — no new env var required. (The `HARPER_CONFIG` layer is the separate second piece — #1214.)

Two properties make union the right — and only — primitive here:

- **Idempotent under repeated application.** A platform layer reapplies its config on every restart / `compose up`. Union reapplied is a no-op; an append would duplicate `server`, `operations-api` on every boot. (This is why we deliberately do **not** add `$append`.)
- **Never removes "extra" entries — even on the force/drift path.** `HARPER_SET_CONFIG`'s drift protection deletes keys absent from the env var; an array under `$union` must be exempt — the directive only guarantees presence of its named members and never deletes members it didn't name, so the app's additions survive.

Useful consequence: a `$union` layer a platform reapplies on each restart yields "guaranteed subset + app-managed extras" with no separate locking machinery — the reapplication is the lock, and the union spares the app's entries.

## Current behavior

Arrays are treated as atomic leaves. `isPlainObject` excludes arrays, so `flattenObject` stores the whole array at its dot-path and `applyConfigLayer` overwrites it. Confirmed by `unitTests/config/harperConfigEnvVars-defaults.test.js:293-305`.

## Implementation

Two change sites in `config/harperConfigEnvVars.ts`:

1. **`flattenObject`** — stop recursing when a plain object is a directive (its only key is `$union`); treat it as a leaf so it doesn't flatten to `tls.uses.$union`.
[...]

Issue #1097, opened 2026-06-02
Add `HARPER_CONFIG` as the recommended way to set configuration from the environment


heskew:
## Proposal

Add `HARPER_CONFIG` as the recommended environment variable for setting Harper configuration. It takes a JSON config object and applies it **on top of** the existing configuration, merging rather than replacing — you specify only the keys you care about and everything else is preserved.

`HARPER_DEFAULT_CONFIG` and `HARPER_SET_CONFIG` stay, but become the specialized options:

| Variable | Use it when you want to… | Precedence | Merge behavior |
|---|---|---|---|
| **`HARPER_CONFIG`** *(recommended)* | Set configuration. The normal, intuitive default. | Above defaults / config file; below an explicit force-override | Merge — only the keys you specify change |
| `HARPER_DEFAULT_CONFIG` | Provide defaults that the config file, individual env vars, or user edits may override | Lowest | Merge, but yields to other sources |
| `HARPER_SET_CONFIG` | Force values and keep them locked against drift / user edits | Highest | Override (current force semantics) |

The goal is that a user reaching for "how do I set config from the environment?" lands on `HARPER_CONFIG` and gets least-surprise merge behavior, and only reaches for `DEFAULT`/`SET` when they specifically need defaults-only or force-and-lock.

## Motivation

This is driven by a deployment where both the platform layer and the deployed app needed to contribute to `HARPER_SET_CONFIG`, and the platform's value clobbered the app's TLS configuration on every restart/upgrade. Two distinct problems surfaced:

1. **Two writers, one variable.** An env var holds a single string, so when two parties both target `HARPER_SET_CONFIG`, one wins entirely — there is no composition. A second, mergeable layer (`HARPER_CONFIG`) lets the two writers coexist instead of overwriting each other.
2. **Arrays don't compose.** Even with separate layers, the platform needs to *guarantee a required set of array entries is present* (e.g. `tls.uses: [server, operations-api]`) while letting the app add its own — and neither side should drop the other's. Today an array fully replaces, so this is impossible.

## Merge semantics (the key behavior to get right)

Config from these env vars is applied in `config/harperConfigEnvVars.ts` (`applyRuntimeEnvConfig`), which flattens the JSON to dot-paths (`flattenObject`) and writes each leaf:

- **Plain objects already merge at the key level** — setting `http.port` leaves `http.securePort` and other siblings intact.
[...]

PR #1238, opened 2026-06-10
feat(config): add `HARPER_CONFIG` as the recommended way to set config from the environment

### LINKED ISSUE #1214: Add `HARPER_CONFIG` as the recommended mergeable config env var
## Summary

Add `HARPER_CONFIG` as the **recommended** environment variable for setting Harper configuration. It takes a JSON config object and applies it **on top of** the existing configuration — merging rather than replacing, so you specify only the keys you care about and everything else is preserved.

`HARPER_DEFAULT_CONFIG` and `HARPER_SET_CONFIG` stay, but become the specialized options:

| Variable | Use it when you want to… | Precedence | Merge behavior |
|---|---|---|---|
| **`HARPER_CONFIG`** *(recommended)* | Set configuration. The normal, intuitive default. | Above defaults / config file; below an explicit force-override | Merge — only the keys you specify change |
| `HARPER_DEFAULT_CONFIG` | Provide defaults that the config file, individual env vars, or user edits may override | Lowest | Merge, but yields to other sources |
| `HARPER_SET_CONFIG` | Force values and keep them locked against drift / user edits | Highest | Override (current force semantics) |

Goal: a user reaching for "how do I set config from the environment?" lands on `HARPER_CONFIG` and gets least-surprise merge behavior, and only reaches for `DEFAULT`/`SET` when they specifically need defaults-only or force-and-lock.

Parent: #1097

## Why

Problem 1 from #1097 — **two writers, one variable.** An env var holds a single string, so when two parties both target `HARPER_SET_CONFIG`, one wins entirely; there is no composition. A second, mergeable layer (`HARPER_CONFIG`) lets the two writers coexist: the platform uses `HARPER_SET_CONFIG`, the app uses `HARPER_CONFIG`, and they merge instead of overwriting each other.

## Implementation notes

The pipeline lives in `config/harperConfigEnvVars.ts` (`applyRuntimeEnvConfig`), called from `config/configUtils.js` (`applyRuntimeEnvVarConfig`).

- New `ConfigSource` `'HARPER_CONFIG'`; extend `ConfigState.snapshots`.
- Insert into the apply pipeline between `DEFAULT` and `SET` in `applyRuntimeEnvConfig`, and into `composeConfigFromEnv` (clone / pre-install path) with the same ordering.
- Add `HARPER_CONFIG` to the early-return guards in both `applyRuntimeEnvConfig` (`harperConfigEnvVars.ts`) and `applyRuntimeEnvVarConfig` (`configUtils.js`).
- Decide the `filterArgsAgainstRuntimeConfig` interaction (whether individual `HARPER_*` env vars are filtered against `HARPER_CONFIG`, as they are against `HARPER_SET_CONFIG` today) — depends on decision 1 below.

## Open decisions (resolve before / early in implementation)

1. **Precedence vs the config file and user manual edits.** Does `HARPER_CONFIG` win over drift (manual config-file edits), or yield to them like a default? This is what distinguishes it from `SET` other than the locking behavior.
2. **Deletion / cleanup semantics.** `DEFAULT` and `SET` do snapshot-based delete-on-omission with original-value restoration (`handleDeletions`, `cleanupRemovedEnvVar`). Does `HARPER_CONFIG` revert a key when it's omitted on a later boot, or is it purely additive? ("Only set what's set" is ambiguous between "track and revert my keys" and "additive, never remove.")
3. **Overlap with `HARPER_DEFAULT_CONFIG`.** Purely additive, or does `HARPER_CONFIG` absorb DEFAULT's niche over time?

## Delivery / timing note (from #1097 discussion)

Harper consumes only `process.env`; `.env` is just one delivery path (`resources/loadEnv.ts` parses a `.env` into `process.env`, skipping keys already set). The in-app `.env` loader is an **application component** that runs *after* boot, so config-merge env vars (e.g. `tls.uses`, ports) must be present in `process.env` **at process start** — via the deployment environment (`env_file` / `environment:`) or the app's own Harper config file. A `HARPER_CONFIG` delivered by the in-app `.env` component loads too late to feed the boot-time merge. (Confirm the component-vs-init ordering during implementation.)

## Acceptance criteria

- `HARPER_CONFIG` documented as the default / recommended way to set configuration from the environment.
- Applies as a merge: unspecified keys (at any depth) are preserved.
- Defined, documented precedence relative to the config file, individual `HARPER_*` env vars, `HARPER_DEFAULT_CONFIG`, and `HARPER_SET_CONFIG`.
- Applied consistently across the main thread and all worker threads (no divergence between threads).
- Tests covering object merge, precedence against each other source, and the deletion semantics chosen in decision 2.

Array composition (the `$union` directive) is tracked separately in #1213 and lands first.

---

heskew:
## Summary

Adds `HARPER_CONFIG` as the **recommended** environment variable for setting Harper configuration. It takes a JSON config object and applies it as a **merge** on top of the existing configuration: it sets exactly the keys it names (at any depth), reasserts them on every boot, and yields only to `HARPER_SET_CONFIG`.

Closes #1214 (sub-issue of #1097). Builds on the `$union` directive from #1213.

## Precedence (later wins)

```
HARPER_DEFAULT_CONFIG  <  config file / user edits  <  HARPER_CONFIG  <  HARPER_SET_CONFIG
```

| Variable | Use it when | Merge behavior |
|---|---|---|
| **`HARPER_CONFIG`** *(recommended)* | Set configuration — the normal default | Merge; only the named keys change; reasserts every boot |
| `HARPER_DEFAULT_CONFIG` | Provide overridable defaults | Merge, yields to file/user/individual vars |
| `HARPER_SET_CONFIG` | Force + lock against drift | Override (existing force semantics) |

## Behavior (the four decisions, confirmed with @heskew)

1. **Drift:** while the var names a key, the env wins — a manual config-file edit to that key is overwritten on the next restart (12-factor intuition). This is what distinguishes it from `DEFAULT`.
2. **Deletion:** dropping a key from `HARPER_CONFIG` restores the pre-override original (or deletes the key if `HARPER_CONFIG` introduced it), via the existing snapshot machinery — symmetric with `DEFAULT`/`SET`.
3. **Individual `HARPER_*` env vars win** over `HARPER_CONFIG` for the keys they name (`filterArgsAgainstRuntimeConfig` stays SET-only — unchanged).
4. **DEFAULT** stays as-is — three vars, three niches (defaults / recommended-merge / force).

## Implementation

- New `'HARPER_CONFIG'` `ConfigSource` + snapshot slot; applied via `applyConfigLayer` with `respectSources: ['HARPER_SET_CONFIG']` + `storeOriginals`, same at install and runtime.
- Added to `applyRuntimeEnvConfig`, `composeConfigFromEnv` (clone/pre-install), the `configUtils.js` early-return guard, and the restore-original deletion/cleanup paths.
- `$union` flows through `HARPER_CONFIG` for free via the shared apply path.

## Where to look

- **Cleanup-before-apply ordering (`applyRuntimeEnvConfig`)** — this PR moves all removed-var cleanups *ahead* of the applies. A domain review caught that the original apply-then-cleanup order gave a **one-boot window** where removing `HARPER_SET_CONFIG` (while `HARPER_CONFIG` still named the path) dropped the value to the file original for one boot before self-healing. Reordering makes `HARPER_CONFIG` reclaim the path the same boot. Behavior is unchanged for present-var and DEFAULT/SET-only cases (a removed var's own apply is a no-op). New regression test: "reclaims a path the SAME boot HARPER_SET_CONFIG is removed".
- **Known-benign quirk:** when `HARPER_CONFIG` and `HARPER_SET_CONFIG` name the same path in one boot, `HARPER_CONFIG`'s snapshot records the value it set before `SET` overrode it, while `sources[path]` ends up `HARPER_SET_CONFIG`. No data loss — deletion paths gate on current source ownership (verified) — but the snapshot is momentarily not authoritative for that path. Left as-is to avoid complicating `buildSnapshot`; flagging for the reviewer.

## Testing

13 new unit tests in `unitTests/config/harperConfigEnvVars-config.test.js`: merge/sibling preservation, source+original tracking, `$union`, all four precedence relationships (file, user-edit reassert, DEFAULT, SET incl. not stealing SET-owned paths), the SET→CONFIG same-boot hand-back, drop-key restore, var-removal cleanup, introduced-key deletion, SET-only arg filtering, and four-layer `composeConfigFromEnv` ordering. Full `unitTests/config/**` suite green (174 passing); build + lint + format clean.

## Review note

Cross-model review was domain-only this round: Codex hit its weekly session limit and the local Gemini (`agy`) leg hangs in this environment. The Harper-domain pass (which found the ordering issue above) ran; the GitHub Gemini Code Assist + Claude review bots will provide the outside-model legs on this PR.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) (agent: Claude Opus 4.8)

### DIFF
--- a/config/configUtils.js
+++ b/config/configUtils.js
@@ -872,18 +872,23 @@ function applyRuntimeEnvVarConfig(configDoc, configFilePath, options = {}) {
 	const defaultEnvValue = process.env.HARPER_DEFAULT_CONFIG;
 	const configEnvValue = process.env.HARPER_CONFIG;
 	const setEnvValue = process.env.HARPER_SET_CONFIG;
+	const anyEnvValue = defaultEnvValue || configEnvValue || setEnvValue;
 
-	// No env vars set, skip entirely (zero overhead)
-	if (!defaultEnvValue && !configEnvValue && !setEnvValue) return;
-
-	const { applyRuntimeEnvConfig } = require('./harperConfigEnvVars.ts');
+	const { applyRuntimeEnvConfig, hasPersistedEnvConfigState } = require('./harperConfigEnvVars.ts');
 
 	// Get rootPath for state file location
 	const rootPath = configDoc.getIn(['rootPath']);
 	if (!rootPath) {
-		logger.warn('Cannot apply runtime env config: rootPath not found in config');
+		// Only an error if there is config to apply; otherwise there is simply nothing to do.
+		if (anyEnvValue) logger.warn('Cannot apply runtime env config: rootPath not found in config');
 		return;
 	}
+
+	// Skip entirely (zero overhead) only when nothing is set AND there is no prior state to
+	// clean up. If a var was applied on a previous boot and then removed, all three env vars
+	// are absent now but applyRuntimeEnvConfig must still run to restore originals and clear
+	// the snapshot — so don't short-circuit in that case.
+	if (!anyEnvValue && !hasPersistedEnvConfigState(rootPath)) return;
 
 	// Convert to JSON for processing
 	const configObj = configDoc.toJSON();
--- a/config/configUtils.js
+++ b/config/configUtils.js
@@ -171,7 +171,7 @@ function createConfigFile(args, skipFsValidation = false) {
 
 	if (schemasArgs) setSchemasConfig(configDoc, schemasArgs);
 
-	// Apply HARPER_DEFAULT_CONFIG and HARPER_SET_CONFIG environment variables BEFORE validation
+	// Apply HARPER_DEFAULT_CONFIG, HARPER_CONFIG and HARPER_SET_CONFIG environment variables BEFORE validation
 	// This allows runtime env vars to resolve port conflicts before validation
 	// Must be called AFTER rootPath is set in configDoc
 	// Mutates configDoc in place
@@ -352,7 +352,7 @@ function initConfig(force = false) {
 
 		checkForUpdatedConfig(configDoc, configFilePath);
 
-		// Apply HARPER_DEFAULT_CONFIG and HARPER_SET_CONFIG environment variables
+		// Apply HARPER_DEFAULT_CONFIG, HARPER_CONFIG and HARPER_SET_CONFIG environment variables
 		applyRuntimeEnvVarConfig(configDoc, configFilePath);
 
 		// Validates config doc and if required sets default values for some parameters.
@@ -848,15 +848,16 @@ function parseYamlDoc(filePath) {
 }
 
 /**
- * Apply HARPER_DEFAULT_CONFIG and HARPER_SET_CONFIG environment variables at runtime
+ * Apply HARPER_DEFAULT_CONFIG, HARPER_CONFIG and HARPER_SET_CONFIG environment variables at runtime
  *
  * This function performs the following:
  * 1. Loads configuration state to track sources
  * 2. Detects user edits (drift) to protect them from HARPER_DEFAULT_CONFIG
  * 3. Applies HARPER_DEFAULT_CONFIG (respects user edits)
- * 4. Applies HARPER_SET_CONFIG (overrides everything)
- * 5. Handles deletions when keys removed from env vars
- * 6. Saves updated state and persists changes to config file (if configFilePath provided)
+ * 4. Applies HARPER_CONFIG (merge layer: reasserts its keys, yields only to HARPER_SET_CONFIG)
+ * 5. Applies HARPER_SET_CONFIG (overrides everything)
+ * 6. Handles deletions when keys removed from env vars
+ * 7. Saves updated state and persists changes to config file (if configFilePath provided)
  *
  * NOTE: This function performs multiple conversions (YAML → JSON → YAML) which is not
  * efficient but provides clear separation of concerns. The conversions are necessary
@@ -869,10 +870,11 @@ function parseYamlDoc(filePath) {
  */
 function applyRuntimeEnvVarConfig(configDoc, configFilePath, options = {}) {
 	const defaultEnvValue = process.env.HARPER_DEFAULT_CONFIG;
+	const configEnvValue = process.env.HARPER_CONFIG;
 	const setEnvValue = process.env.HARPER_SET_CONFIG;
 
 	// No env vars set, skip entirely (zero overhead)
-	if (!defaultEnvValue && !setEnvValue) return;
+	if (!defaultEnvValue && !configEnvValue && !setEnvValue) return;
 
 	const { applyRuntimeEnvConfig } = require('./harperConfigEnvVars.ts');
 
--- a/config/harperConfigEnvVars.ts
+++ b/config/harperConfigEnvVars.ts
@@ -757,6 +757,16 @@ export function composeConfigFromEnv(base: ConfigObject = {}): ConfigObject {
 }
 
 /**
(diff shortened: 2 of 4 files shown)


heskew:
Cross-model review status for this PR: the outside-model legs were rate-limited this round —
- **Codex**: weekly session limit reached (couldn't run).
- **Gemini** (both the local `agy` leg and the GitHub Gemini Code Assist bot): `agy` hangs in this environment, and the bot reported "reached your daily quota limit."
- **Claude PR Review bot**: ran clean — "no blockers found."
- **Harper-domain review pass** (Opus, local): ran and caught one significant issue — a one-boot misconfiguration window when `HARPER_SET_CONFIG` is removed while `HARPER_CONFIG` still names the path — now fixed (cleanup-before-apply reorder) with a regression test.

So a genuine non-Claude review did not land this round. Happy to re-run a Gemini/Codex pass once the limits reset (Gemini ~24h) before merge if you'd like that belt-and-suspenders — otherwise this is resting on the domain pass + clean Claude-bot pass + your human review.

---

kriszyp (approved):

config/configUtils.js
```diff
@@ -869,10 +870,11 @@ function parseYamlDoc(filePath) {
  */
 function applyRuntimeEnvVarConfig(configDoc, configFilePath, options = {}) {
 	const defaultEnvValue = process.env.HARPER_DEFAULT_CONFIG;
+	const configEnvValue = process.env.HARPER_CONFIG;
 	const setEnvValue = process.env.HARPER_SET_CONFIG;
 
 	// No env vars set, skip entirely (zero overhead)
-	if (!defaultEnvValue && !setEnvValue) return;
+	if (!defaultEnvValue && !configEnvValue && !setEnvValue) return;
```
heskew: Codex review: I think this early return needs one more state-aware case. If `HARPER_CONFIG` was applied on a prior boot and then removed entirely, all three env vars are now absent, so this wrapper returns before `applyRuntimeEnvConfig()` can run its snapshot cleanup. That leaves the stale config value persisted and the `HARPER_CONFIG` snapshot uncleared. Could this guard either check for existing env-config state before returning, or delegate to the lower-level helper when state may exist?

heskew: Good catch — fixed in 826a3a6a. Added `hasPersistedEnvConfigState(rootPath)` and made this guard short-circuit only when there are no config env vars **and** no prior state; otherwise it now delegates to `applyRuntimeEnvConfig`, which restores originals and clears the snapshot. (`rootPath` is resolved before the guard, warning only when there's actually config to apply.) Pre-existing for DEFAULT/SET, but `HARPER_CONFIG` makes removal routine, so worth closing here. Added regression tests: no-vars-no-state → skip, no-vars-but-state → cleanup runs, plus a direct test for the predicate.

---

heskew:
Re: the Codex review — nice catch on the state-unaware early return in `applyRuntimeEnvVarConfig`. It was a real, pre-existing gap (DEFAULT/SET had it too), just made routine by `HARPER_CONFIG` being the recommended var people add and remove.

Fixed in 826a3a6a: the wrapper now skips only when there are no config env vars **and** `hasPersistedEnvConfigState(rootPath)` is false; otherwise it delegates so `applyRuntimeEnvConfig` restores originals and clears the snapshot. Covered by three new tests (no-vars-no-state → skip; no-vars-but-state → cleanup runs; a direct predicate test). Inline thread resolved. ✅

---

PR #1239, opened 2026-06-10
feat(replication): LOCAL_ONLY record-metadata flag for non-replicating writes

kriszyp:
## Summary

Adds a generic record-metadata bit — `LOCAL_ONLY = 0x8000` (`resources/auditStore.ts`) — that marks a write as **node-local: persisted here, never forwarded to replication peers.** Internal primitive (no public `put`/`patch` surface yet — see Notes); used by harper-pro to fix the v4→v5 bridge leak ([harper-pro#246](https://github.com/HarperFast/harper-pro/issues/246)).

## How it works

- `RecordEncoder.recordUpdater`: on the internal `localOnly` write option, sets the bit on **both** the persisted record metadata header (`metadataInNextEncoding`) and the audit entry's `extendedType`.
- `Table.ts`: threads `localOnly` from the write options into `recordUpdater`.
- The replication send path skips `LOCAL_ONLY` records via a pure bitmask test on the already-available `metadataFlags` / `extendedType` integer — **no record value decode is added to the throughput-critical send path** (the harper-pro side does the skipping; this PR just defines + sets the bit).

## What to check

- **Bit choice `0x8000` (bit 15):** confirmed unused in the record-metadata bitmap and the audit `extendedType` space (tops out at `0x4000`; framing is `0xc0000000`), sits below the action byte, and round-trips through the always-32-bit metadata encode/decode. Verified by review against `auditStore.ts` + `RecordEncoder.ts`.
- Both bits are set together unconditionally — no path sets only one (which would leak via the unset channel).
- Deep-reviewed (data-integrity + concurrency): the bit persists and round-trips; no decode added.

## Notes

`LOCAL_ONLY` is intentionally **internal-only** for now. A public API (per-write option and/or table-level config) is deferred to a planned **residency-scoped metadata visibility** feature, where this becomes the residency = {self} case. Tracked in the harper-pro follow-up issue.

🤖 Generated with Claude Opus 4.7 (implementation by Claude Sonnet 4.6).

### DIFF
--- a/resources/RecordEncoder.ts
+++ b/resources/RecordEncoder.ts
@@ -15,6 +15,7 @@ import {
 	HAS_BLOBS,
 	ACTION_32_BIT,
 	HAS_ADDITIONAL_AUDIT_REFS as HAS_ADDITIONAL_AUDIT_REFS_AUDIT,
+	LOCAL_ONLY,
 } from './auditStore.ts';
 import * as harperLogger from '../utility/logging/harper_logger.ts';
 import './blob.ts';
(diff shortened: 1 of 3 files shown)


kriszyp:
(there was a harper-pro approval)

PR #1247, opened 2026-06-11
fix(table): re-apply @table metadata on reload + coerce directive args

### LINKED ISSUE #1245: describe_all shows schema_defined: false and omits expiration for eviction-configured tables on upgraded clusters
### Bug Summary

After an in-place upgrade into 5.1.0-beta.2, `describe_all` shows `schema_defined: false` for every `@table`-declared table and omits `expiration` for eviction-configured tables, even though eviction still functionally runs.

### Steps to Reproduce

1. Run Harper 5.0.x with a `graphqlSchema` component that declares at least one `@table(eviction: N)` table.
2. `describe_all` and confirm `expiration: "0s"`, `schema_defined: true`.
3. In-place upgrade to 5.1.0-beta.2 (stop, swap binary, start; same data dir). Works for v5 → v5 and v4 → v5 with LMDB → RocksDB.
4. `describe_all` again.

### Expected Behavior

Table descriptors retain `schema_defined: true` and eviction-configured tables retain `expiration: "0s"`.

### Actual Behavior

All `@table`-declared tables show `schema_defined: false`. Eviction-configured tables omit the `expiration` field entirely. Eviction still runs (records drain to 0 on the configured window); the gap is API-visible metadata only.

Concrete sample from `seo.PageCache` declared as `@table(database: "seo", eviction: 300)`:

| Field | 5.0.27 | 5.1.0-beta.2 |
|---|---|---|
| schema_defined | true | false |
| expiration | "0s" | (missing) |

### Platform

Harper 5.1.0-beta.2 (RocksDB). Reproduced on `us-lax-1.v4-to-v5.akamai.stage` (5.0.27 → 5.1.0-beta.2 in-place) and `nl-ams-1.v4-to-v5.akamai.stage` (4.7.32 → 5.1.0-beta.2 with LMDB → RocksDB migration).

### Additional Context

Probe on us-lax-1 pinned three independent causes:

1. `setTTLExpiration` runs in the http worker that loads the schema, but `describe_all` is served by main/0 with a separate Table closure where `expirationMs` stays `undefined`.
2. The existing-Table branch of `databases.ts::table()` doesn't re-assign `Table.schemaDefined`, so a stale `false` on disk is sticky across reloads.
3. The GraphQL parser stringifies `@table` directive literals, so `eviction: 300` arrives as `"300"`.

PR with fix coming next.

Follow-ups (separate issues):
- `describe_all.sources` field. Was reading `tableObj.sources` (plural, dead code since the Feb 2026 single-source simplify); fixing the field name is one line but actually restoring visibility needs cross-worker `sourcedFrom()` propagation.
- Originating cause of `schemaDefined: false` written to disk in the first place. Fix self-heals on every reload, but the writer hasn't been pinned.

Yes, I'll submit a PR to fix this.

ldt1996:
## Summary

On in-place upgrades into 5.1.0-beta.2 (both v4 → v5 with LMDB → RocksDB and v5 → v5), `describe_all` shows `schema_defined: false` on every table and omits `expiration` for eviction-configured tables. Reproduced on us-lax-1 and nl-ams-1. Eviction still functionally works; the gap is metadata visibility.

## Fix

- [resources/Table.ts](../blob/main/resources/Table.ts): default `expirationMs = 0` when `evictionMs > 0 && expirationMs === undefined` so describe sees it on every worker.
- [resources/databases.ts:1017](../blob/main/resources/databases.ts#L1017): unconditionally re-assert `Table.schemaDefined = schemaDefined` on every component reload.
- [resources/graphql.ts](../blob/main/resources/graphql.ts): `coerceDirectiveValue` switches on AST node kind so Int/Float arrive as numbers.

## Test

`integrationTests/apiTests/describe-metadata-upgrade-1132.test.ts`: pinned `dataRootDir`, two-phase, `threads.count: 2`. Phase 1 installs the component and asserts `expiration: "0s"` and `schema_defined: true`. Phase 2 reboots on the same data dir and asserts both survive. Both fail without the patches.

closes #1245

### DIFF
--- a/integrationTests/apiTests/describe-metadata-upgrade-1132.test.ts
+++ b/integrationTests/apiTests/describe-metadata-upgrade-1132.test.ts
@@ -1,0 +1,105 @@
+/**
+ * Two-phase: install a component with `@table(eviction: 300)`, capture describe_all,
+ * stop, reboot on the same data dir, assert describe_all still reflects schema_defined: true
+ * and expiration: "0s". Runs with threads.count: 2 so the operations-API worker that
+ * answers describe is distinct from the http worker that runs the schema parser, matching
+ * the cluster topology that hid the bug.
+ */
+import { suite, test, before, after } from 'node:test';
+import { ok, strictEqual } from 'node:assert/strict';
+import { mkdtempSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import {
+	startHarper,
+	killHarper,
+	teardownHarper,
+	releaseLoopbackAddress,
+	type ContextWithHarper,
+} from '@harperfast/integration-testing';
+// @ts-expect-error utils/client.mjs has no type declarations
+import { createApiClient } from './utils/client.mjs';
+// @ts-expect-error utils/components.mjs has no type declarations
+import { installAppComponent } from './utils/components.mjs';
+
+const SCHEMA = `
+type PageCache @table(database: "seo", eviction: 300) @export(name: "SeoPageCache") {
+	cacheKey: ID! @primaryKey
+	htmlContent: String
+	updatedTimestamp: Long
+}
+
+type LifecycleMeta @table(database: "metadata") @export(name: "LifecycleMeta") {
+	id: ID! @primaryKey
+	value: String
+}
+`;
+
+const CONFIG = `rest: true
+graphqlSchema:
+  files: '*.graphql'
+`;
+
+const PINNED_DATA_DIR = mkdtempSync(join(tmpdir(), 'fix-1132-'));
+
+async function describeAll(client: ReturnType<typeof createApiClient>) {
+	const res = await client.req().send({ operation: 'describe_all' });
+	strictEqual(res.status, 200, 'describe_all should return 200');
+	return res.body;
+}
+
+suite('INVESTIGATION-1132 fix phase 1: install component on first boot', (ctx: ContextWithHarper) => {
+	before(async () => {
+		ctx.harper = { dataRootDir: PINNED_DATA_DIR } as any;
+		await startHarper(ctx, { config: { threads: { count: 2 } }, env: {} } as any);
+	});
+
+	after(async () => {
+		await killHarper(ctx);
+		await releaseLoopbackAddress(ctx.harper.hostname);
+	});
+
+	test('install component and capture first describe_all', async () => {
+		const client = createApiClient(ctx.harper);
+		await installAppComponent(client, {
+			project: 'fix1132',
+			files: { 'schema.graphql': SCHEMA, 'config.yaml': CONFIG },
+			probePath: '/SeoPageCache/',
+			restartTimeoutMs: 120_000,
+		});
+		const body = await describeAll(client);
+		const pc = body.seo?.PageCache ?? body.data?.seo?.PageCache;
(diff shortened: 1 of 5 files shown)


kriszyp:
The expiration default and directive coercion look solid — the two concerns Kris raised check out (see below). One issue worth fixing before merge:

### `Table.schemaDefined = schemaDefined` can flip dynamic tables to `true`

`databases.ts:967` defaults `schemaDefined = true` when the caller omits it. The existing-table branch previously never wrote it back, so that default never leaked. With this patch it does, and two callers invoke `table()` on existing tables without passing `schemaDefined`:

- **Table.ts:537** — the cluster schema-replication event handler: `table({table, database, attributes, origin: 'cluster'})`. This is the primary code path for dynamic tables receiving attribute additions from a peer, so on any cluster, dynamic tables would flip to `schema_defined: true` in memory on the receiving node.
- **dataLoader.ts:228** — same omission when loading data files into an existing table.

`schemaDefined` gates real behavior (auto-attribute creation on insert, legacy `dropAttribute` value deletion, `select *` semantics in `ResourceBridge`), so this isn't cosmetic.

**Suggested fix** — guard the re-assert on whether the caller explicitly supplied the flag:

```ts
// databases.ts, table() — before the schemaDefined default at line ~967
const schemaDefinedExplicit = tableDefinition.schemaDefined !== undefined;
```

Then in the existing-table branch:
```ts
// replace the new line
Table.schemaDefined = schemaDefined;
// with
if (schemaDefinedExplicit) Table.schemaDefined = schemaDefined;
```

`graphql.ts` always sets `schemaDefined` explicitly through the `typeDef` it builds, so the upgrade fix still fires there. The callers that omit it (cluster replication, data loader, `addAttributes`, `removeAttributes`) are all operating on already-live tables whose `schemaDefined` should be left alone.

It'd also be worth adding an assertion to the test that a dynamically-created table (no `attributes` in `create_table`) still reports `schema_defined: false` after the reboot — that would catch this regression directly.

---

**On Kris's two questions:**

1. **Coercion regressions** — None. Every consumer of `expiration`/`eviction` was already type-agnostic (string multiplication, `?? 0`, `+expiration` normalization). Boolean-typed args like `audit` and `randomAccessFields` already had string-compat shims. `expiration: 0` flips from truthy string `"0"` to falsy `0`, but the downstream arithmetic is identical.

2. **`expiration = 0` default when eviction is set** — Safe. On write, `expiresAt = expirationMs ? expirationMs + Date.now() : -1`, so `0` is falsy and records get the sentinel `-1` (no expiration stored at the record level). The eviction scanner and staleness checks only act when `expiresAt` is defined and in the past. `expirationMs = 0` is exactly what `setTTLExpiration({eviction})` has always produced on workers that parsed the schema — this patch just makes the metadata-constructed workers match, which is why describe shows `expiration: "0s"` for eviction-only tables.

*— Claude Fable 5*

kriszyp:
Also, does this really need to be `patch`? It doesn't appear to merge cleanly.

ldt1996:
Good catch on the dynamic-table flip, applied as suggested:

- `schemaDefinedExplicit = tableDefinition.schemaDefined !== undefined` captured before the default, re-assert gated on it. GraphQL `@table` still fires; cluster schema-replication and dataLoader leave existing tables alone.
- Test now creates a dynamic table via `create_table` with no attributes and asserts `schema_defined: false` survives the phase-2 reboot.

Thanks for the coercion / `expiration = 0` sanity check too.

Lavinia via Claude

kriszyp (approved):

PR #1251, opened 2026-06-11
Default the RocksDB WriteBufferManager on at 1/3 of block cache

kriszyp:
## Summary
Makes the RocksDB WriteBufferManager (WBM) **on by default** in core (it was opt-in). Defaults, applied in `openRocksDatabase`:
- `writeBufferManagerSize` = `blockCacheSize / 3` (floored)
- `writeBufferManagerCostToCache` = `true`
- `writeBufferManagerAllowStall` = `true`
- An explicit size of `0` still disables the WBM; any explicitly configured value/boolean is honored.

The block-cache + WBM defaulting was extracted out of `openRocksDatabase` into a pure `resolveRocksMemoryConfig` helper (`utility/rocksMemoryConfig.ts`) so the branching is unit-testable without opening a database or hitting the process-global `RocksDatabase.config`.

## Purpose
Bound total memtable memory across column families and apply write backpressure (`allowStall`) so bulk ingest can't outrun the memtable flush/conflict-check window. This is the throughput/memory half of HarperFast/harper-pro#348 (the data-safety floor — not silently dropping on retry-cap exhaustion — is a separate follow-up PR).

## Where to look
- `utility/rocksMemoryConfig.ts` — the defaulting rules (3-way size logic: unset → default, `0` → disabled, `>0` → used).
- `resources/databases.ts` — now just calls the helper.

## Notes for the reviewer
- **Fabric/host-manager interaction:** host-manager's compose template already sets all four `STORAGE_ROCKS_*` env vars explicitly (block cache 15%, WBM 5%, `costToCache` true, `allowStall` **false**). Those explicit values are typed and honored over these new defaults, so Fabric keeps its Phase-1 `allowStall: false` — no regression.
- **Deliberate call (open for input):** I did *not* add a hardcoded fallback for `availableMemory` being `Infinity`/`NaN`. `availableMemory = min(constrainedMemory ?? Infinity, totalmem())` is always finite because `os.totalmem()` is, so guarding it would be validating an unreachable state. Negative configured sizes already disable safely via the `> 0` gate. Flag if you'd prefer a defensive guard anyway.

Cross-model review (Codex + Gemini/agy) run pre-PR; float-flooring, strict-assert, and edge tests were added in response.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — model: Claude Sonnet 4.6

### DIFF
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -31,6 +31,7 @@ import { replayLogs } from './replayLogs.ts';
 import { totalmem } from 'node:os';
 import { RocksIndexStore } from './RocksIndexStore.ts';
 import { when } from '../utility/when.ts';
+import { resolveRocksMemoryConfig } from '../utility/rocksMemoryConfig.ts';
 import { isProcessRunning } from '../utility/processManagement/processManagement.js';
 
 /**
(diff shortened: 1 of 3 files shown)


kriszyp:
Thanks for the review! On the suggestion to move the `configuredBlockCacheSize` / `writeBufferManagerSize` checks into the Joi schema: I'm keeping the type-enforcement in `resolveRocksMemoryConfig` deliberately. These values already flow through `configUtils.castConfigValue` (which yields proper numbers/booleans/null), and the resolver's `typeof` guards are *defaulting* logic rather than validation — they're what implement the "unset → derive from block cache, explicit 0 → disable" behavior, which is clearer co-located with the defaults it feeds. This also matches the existing block-cache resolution pattern. Open to revisiting if we want stricter rejection of malformed config at the schema layer as a separate change.

— Claude Sonnet 4.6

cb1kenobi (approved):

PR #1281, opened 2026-06-13
fix(replication): repair dangling blob reference on identity-tie re-stream (authoritative tables)

kriszyp:
## Status: recommend DROP (redundant on the #368 watermark path)

This PR added a dedicated repair in `Table._writeUpdate` at the identity-tie duplicate-drop:
on an authoritative (non-caching) table, if the existing row references a blob whose file is
missing and a re-streamed duplicate arrives, re-store the row onto the re-streamed value at the
same version (a local-only `'relocate'`), so the dangling reference is replaced by the
freshly re-saved blob.

The premise was that a receive-side blob save failure commits a dangling row, and the
resume-cursor re-stream then arrives as an identity-tie duplicate and is **dropped**, stranding
the blob (caching tables self-heal via re-source; authoritative tables cannot).

### Why it's redundant now

The real blocker was a receive-path **deadlock**, which is fixed by **harper-pro #368** (an
async durability watermark). With #368 in place, the authoritative path was tested empirically
(harper-pro #371, 4 runs):

- The disrupted record's blob is **reliably re-saved by the natural same-version overwrite** of
  the re-streamed record — the follower converges with no wedge and, with the source offline,
  every authoritative blob is byte-for-byte intact.
- **This repair branch never fired.** Trace at the tie-branch / walk-break / overwrite showed
  the existing-row-has-missing-blob condition was never true at any apply-commit decision: the
  audit-walk `auditStore.get(...)` lookup that gates this repair **reliably misses** on the live
  re-stream, so the record never reaches the tie-drop. It falls through to the natural overwrite
  instead, which re-saves the blob.

The repair therefore sits behind a lookup that reliably misses, guarding a path that already
self-heals. harper-pro #371 retains the lasting value — a byte-integrity regression test that
asserts the authoritative-table blob is intact after failure + restart (no re-source masking) —
and reverts its `core` pointer to #368's base, so it no longer depends on this commit.

**Recommendation:** close or convert this PR. Kept in draft pending that decision; no code
change is needed in core.

Cross-model review pending (main session)

### DIFF
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -61,7 +61,13 @@ import { recordAction, recordActionBinary } from './analytics/write.ts';
 import { rebuildUpdateBefore } from './crdt.ts';
 import { appendHeader } from '../server/serverHelpers/Headers.ts';
 import fs from 'node:fs';
-import { Blob, deleteBlobsInObject, findBlobsInObject, startPreCommitBlobsForRecord } from './blob.ts';
+import {
+	Blob,
+	deleteBlobsInObject,
+	findBlobsInObject,
+	getFilePathForBlob,
+	startPreCommitBlobsForRecord,
+} from './blob.ts';
 import { onStorageReclamation } from '../server/storageReclamation.ts';
 import { RequestTarget } from './RequestTarget.ts';
 import harperLogger from '../utility/logging/harper_logger.ts';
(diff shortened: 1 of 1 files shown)


kriszyp:
Closing — the dedicated repair is unnecessary. With the receive-path deadlock fixed by the async durability watermark (harper-pro #368), an authoritative re-streamed record whose blob save failed is reliably repaired by the **natural same-version overwrite**: the audit-walk keyed lookup that would have gated a tie-drop reliably misses on the live non-copy re-stream, so the record overwrites the row onto a freshly-saved blob. Verified across multiple runs with disk-level byte-integrity checks (all blobs byte-exact after failure+restart, source offline; the repair condition was never even true). The repair sat behind a lookup that reliably misses, guarding a path that already self-heals. The authoritative-table byte-integrity regression test lives in harper-pro #371. — Claude (Opus 4.8)

PR #1285, opened 2026-06-14
New SQL engine on the Resource API (phases 0-4; phase-5 cutover gated)

kriszyp:
A new SQL engine built directly on the Resource API (`Table.search`/`get`/`put`/`patch`/`delete`/`transaction`), replacing the AlaSQL two-pass hybrid (`dataLayer/SQLSearch.js`) that degrades to full attribute scans on `OR`/comparators and materializes all fetched rows for aggregates/sort. AlaSQL is kept **only as a parser**. Full design + phasing in `sqlEngine/PLAN.md`.

Behind a feature flag — `sql.engine = legacy | new | auto` (also `HARPER_SQL_ENGINE`). **Default is `auto`** (phase-5 cutover): the new engine handles every request it can plan and falls back to legacy on `EngineUnsupportedError` for anything it can't — so no query changes behavior unless the new engine produces an identical result. `allowFullScan` stays **`false`** (genuine full scans fall back to legacy).

> **Milestone v5.2 — hold from `main` until 5.2 is cut.** Draft.

## Status by phase
- **0–1** single-table SELECT (parser → binder → logical → optimizer → physical index scan/filter/project/sort/limit).
- **2** aggregates, GROUP BY, DISTINCT, HAVING; custom agg functions.
- **3** JOINs — INNER/LEFT/CROSS + comma-FROM; RIGHT/FULL rejected. Index-nested-loop (default), hash, and nested-loop strategies; LEFT OUTER null-fill.
- **4** Mutations — INSERT (VALUES), UPDATE, DELETE on the Resource API, each in one `transaction(context, …)`. Legacy response shapes preserved. `INSERT … SELECT` and no-WHERE UPDATE/DELETE fall back to legacy.
- **5 (this PR) — cutover.** Default flipped `legacy → auto`; legacy path untouched and reachable via the flag. Gated on full parity (see below). Burn-in (watch `sql-engine v2 fallback:` logs) → flip to `new` → delete legacy path are follow-ups in later releases.
- **6** — broaden coverage (subqueries, UNION, EXPLAIN, cost-based reordering, spill), real DISTINCT-aggregate dedup, and the `allowFullScan` parity gaps. Tracked in #1525; **5.3+, not 5.2.**

## Cutover parity (the gate for the `auto` default)
- **Differential / cutover-readiness test** (`integrationTests/apiTests/sql-engine-differential.test.mjs`): boots a `new`-side (`HARPER_SQL_ENGINE=auto`, the production setting) and a `legacy`-side with identical data, compares response + persisted state across a broad battery (SELECT, OR/NOT/BETWEEN/LIKE/IN, NULL semantics, ORDER BY/LIMIT/OFFSET/DISTINCT, aggregates incl. **bare/unaliased + COUNT(DISTINCT)**, INNER/LEFT join, INSERT/UPDATE/DELETE). Because the new side runs in `auto`, an unsupported query falls back and matches by construction — the only way to fail is a *silent* divergence, which is the real cutover risk. **46/46 identical.**
- **Existing SQL suites under the `auto` default**: northwind (575) and delete.test — **0 failures**.

## Bugs fixed in this PR's cutover pass
Two silent-divergence bugs that the prior parity gates missed because the differential's aggregate cases all used `AS` aliases:
1. **Unaliased aggregate column label.** `SELECT COUNT(*) …` surfaced under the engine's internal `__agg_N__` name instead of legacy's `COUNT(*)`. Index-served aggregates run on the new engine (no fallback), so a consumer keying off `COUNT(*)` — including main's new `scale.test.ts` — got `undefined` (failed CI shard 4/6). Fix: compute the legacy label at normalize time onto `ProjectionNode.label`; `PhysicalProject` prefers `alias ?? label ?? fallback`.
2. **DISTINCT-in-aggregate ignored.** `PhysicalHashAggregate` never read `spec.distinct`, so `COUNT(DISTINCT x)` counted every row (real-engine: 5 vs legacy 3). Rejected with `EngineUnsupportedError` → legacy fallback until real dedup lands (#1525).

Parity gate hardened: added bare/unaliased + COUNT(DISTINCT) differential cases and unit regressions so the blind spot can't reopen.

## Tests
- `unitTests/sqlEngine/` — per-operator + pipeline + join + mutation + aggregate suites (mock-table harness).
- `integrationTests/apiTests/sql-engine-differential.test.mjs` — new-vs-legacy parity (46/46).
- Validated under the `auto` default: northwind (575) + delete + `integrationTests/server/scale.test.ts` (100K) all green.

## Deferred (fall back to legacy; tracked in #1525)
`PhysicalRelationshipJoin`; projection pushdown through joins; single-table predicates in an INNER-JOIN `ON`; `INSERT … SELECT`; no-WHERE UPDATE/DELETE; real DISTINCT-aggregate dedup; `allowFullScan:true` parity gaps. None affect 5.2 (all fall back).

🤖 Generated by Claude (Opus 4.8).


kriszyp:
### Phase 5 (cutover) — progress + a gate

Built the cutover-readiness differential (`integrationTests/apiTests/sql-engine-differential.test.mjs`): two real instances (new side in `auto` — the production cutover setting — vs `legacy`), identical data, a 30-case battery (predicate variety, NULL semantics, ORDER BY/LIMIT/DISTINCT, full aggregate set, INNER/LEFT joins, INSERT/UPDATE/DELETE) comparing responses **and** persisted state. **30/30 identical.**

I then trialed the actual flip (default `legacy → auto`) and validated against the **existing** SQL suite — which caught two parity blockers the battery didn't cover, so **the flip is reverted** (default stays `legacy`):

1. **Literal type-coercion on hash lookups — silent wrong result (must-fix).** `DELETE … WHERE id IN ('123')` (string literals) against a numeric PK matches in legacy (loose `=`) but not the new engine (strict). `auto` doesn't fall back on it (no throw), so it's a silent divergence — the dangerous kind. This is PLAN decision #6 (match AlaSQL loose coercion) not yet implemented.
2. **Non-PK `LIKE` DELETE → 403.** `DELETE … WHERE address LIKE '%Lane'` returns 403 via the new selector path; root cause TBD (plain SELECTs and indexed-PK DELETEs authorize fine).

Both are documented in `sqlEngine/PLAN.md` with an ordered cutover checklist. `delete.test.mjs` is 76/76 green again under the reverted default. The differential is committed as a permanent parity guard.

🤖 Generated by Claude (Opus 4.8).

sqlEngine/optimizer/rules/limitPushdown.ts
```diff
@@ -0,0 +1,43 @@
...
+export function limitPushdown(plan: LogicalPlan): LogicalPlan | null {
+	return rewrite(plan);
+}
+
+function rewrite(plan: LogicalPlan): LogicalPlan {
+	switch (plan.kind) {
+		case 'Limit': {
+			const child = rewrite(plan.input);
+			if (child.kind === 'Scan' && !child.pushedLimit) {
+				return { ...child, pushedLimit: { limit: plan.limit, offset: plan.offset } };
```
cb1kenobi, line 36: **High: LIMIT/OFFSET pushed into the index scan ahead of the residual filter → silent under-fetch**

`limitPushdown` folds `limit`/`offset` into the scan whenever the child is a `Scan`, without checking whether the scan's `pushedFilter` still carries a residual (non-indexable) portion. `whereToConditions` (via `physical/plan.ts`) later splits that residual out into a post-scan `PhysicalFilter`, so `Table.search` caps rows *before* the residual runs. For e.g. `SELECT * FROM t WHERE category = 'books' AND price*2 > 30 LIMIT 10`, the scan returns the first 10 rows matching `category='books'`, then the residual drops most of them — fewer (and wrong) rows than legacy. This runs on the default `auto` engine because the indexed condition means `validateScannable` does not fall back.

Suggested fix: only push limit/offset into a `Scan` when `whereToConditions(scan.pushedFilter)` yields no residual (and the scan has no `residualFilter`); otherwise keep the `Limit` node above the residual `PhysicalFilter`.

—
_Generated by Barber AI_

sqlEngine/physical/PhysicalSort.ts
```diff
@@ -0,0 +1,53 @@
...
+export function physicalSort(child: PhysicalOp, keys: SortNode[], qualified = false): PhysicalOp {
+	const compiled = keys.map((k) => ({ get: compileExpr(k.expr, qualified).eval, descending: k.descending }));
+	return {
+		schema: child.schema,
+		async *execute(ctx: SqlEngineContext): AsyncIterable<Row> {
+			const cap = getSqlEngineConfig().maxSortRows;
+			const buf: Row[] = [];
+			for await (const row of child.execute(ctx)) {
+				if (buf.length >= cap) {
+					throw new EngineRuntimeError(`sort buffer exceeded ${cap} rows`);
```
cb1kenobi: **Medium: in-memory cap overflow throws `EngineRuntimeError`, which `auto` mode never falls back on**

`EngineRuntimeError` extends `ClientError`, not `EngineUnsupportedError`, and `router.ts` only falls back to legacy on `EngineUnsupportedError`. So exceeding `maxSortRows` here (and `maxHashRows` in `PhysicalHashAggregate` / `PhysicalHashJoin`) surfaces a client error in the default `auto` mode instead of falling back — contradicting this file's own comment ("callers fall back to legacy") and breaking the `auto` invariant that a query only changes behavior when the new engine returns an identical result. `runSelect` fully materializes before returning, so a fallback at this point is safe.

Suggested fix: throw `EngineUnsupportedError` on cap overflow (so `auto` falls back to legacy), or extend the router's fallback to also cover `EngineRuntimeError`.

—
_Generated by Barber AI_

sqlEngine/optimizer/whereToConditions.ts
```diff
@@ -0,0 +1,233 @@
...
+						variants.push(variant);
+					}
+				}
+			}
+			const conditions: ConditionNode[] = variants.map((v) => ({
+				attribute,
+				comparator: expr.negated ? 'ne' : 'equals',
+				value: v,
+			}));
+			return { conditions, operator: expr.negated ? 'and' : 'or' };
```
cb1kenobi, line 372: **Medium: `NOT IN` omits the `IS NOT NULL` guard that `!=` applies (three-valued-logic asymmetry)**

The `!=` path above deliberately AND-s an `IS NOT NULL` guard (`{ comparator: 'ne', value: null }`) so NULL rows are excluded, matching legacy AlaSQL's SQL three-valued logic — the PR lists `!=` / NULL 3VL as a fixed gap. `NOT IN` is semantically `col != a AND col != b`, but this negated branch builds the `ne` branches without that guard. When a `NOT IN` is combined with another indexed conjunct (so the scan runs on the new engine rather than falling back), rows whose column is NULL are returned by the new engine but excluded by legacy → silent divergence.

Suggested fix: in the `negated` case, also push `{ attribute, comparator: 'ne', value: null }` into the AND group, mirroring the `!=` handling.

—
_Generated by Barber AI_

sqlEngine/config.ts
```diff
@@ -0,0 +1,69 @@
...
+ * this resolution lazy and lightweight so the router can be invoked without a
+ * fully booted Harper config (e.g., in unit tests).
+ *
+ * Phase 5 cutover: the default is now 'auto' — the new engine handles every SQL
+ * request it can plan and silently falls back to legacy on anything it can't, so
+ * no query changes behavior unless the new engine produces an identical result.
+ * The gate for this flip was full parity of the new engine (run in 'auto') against
+ * the existing SQL suite: the cutover-readiness differential (42/42 identical) plus
+ * the existing behavioral suites under 'auto' — northwind (573 SQL ops) and
+ * delete.test.mjs (76) — at 0 failures. Both trial-flip blockers (IN literal
```
cb1kenobi, line 22: **Nit: stale parity numbers in the cutover rationale**

This comment cites "42/42 identical" and "northwind (573 SQL ops)" / "delete.test.mjs (76)", but the PR description reports 46/46 and northwind 575. Since these figures are the stated gate for flipping the default to `auto`, align them with the current differential/suite counts.

—
_Generated by Barber AI_

sqlEngine/optimizer/rules/limitPushdown.ts
kriszyp: Good catch — confirmed this reaches the default `auto` path (the indexed conjunct keeps `validateScannable` from falling back). Fixed in 464cdfe3d: `limitPushdown` now folds LIMIT/OFFSET into a `Scan` only when it carries no residual (`whereToConditions(...).residual` or an explicit `residualFilter`); otherwise the `Limit` stays above the residual `PhysicalFilter`. Added a regression in `pipeline.test.js` that asserts the limit is *not* pushed when a residual remains — it returned 0 rows instead of 1 before the fix.

—KrAIs (Claude Opus 4.8)

sqlEngine/physical/PhysicalSort.ts
kriszyp: Agreed, and thanks for tracing the `auto` invariant. Fixed in 464cdfe3d by having all four in-memory caps (sort, hash-aggregate, hash-join, nested-loop) throw `EngineUnsupportedError` instead of `EngineRuntimeError`, so `auto` falls back to legacy. `runSelect` fully materializes before returning, so nothing has been emitted to the client when a cap trips — the fallback is safe. `EngineRuntimeError` is now reserved for genuinely non-recoverable runtime failures (its docstring was updated to say so).

—KrAIs (Claude Opus 4.8)

sqlEngine/optimizer/whereToConditions.ts
kriszyp: Nice spot on the three-valued-logic asymmetry. Fixed in 464cdfe3d: the negated `IN` branch now AND-s `{ attribute, comparator: 'ne', value: null }`, mirroring the `!=` path, so a NULL column is excluded to match legacy AlaSQL. Added a `NOT IN` NULL-exclusion regression in `pipeline.test.js`.

—KrAIs (Claude Opus 4.8)

sqlEngine/config.ts
kriszyp: Updated in 464cdfe3d — 46/46 and northwind 575, aligned with the current differential/suite counts.

—KrAIs (Claude Opus 4.8)

unitTests/sqlEngine/join.test.js
```diff
@@ -0,0 +1,365 @@
...
+			primaryKey: 'id',
+			attributes: [
+				{ name: 'id', indexed: true },
+				{ name: 'k', indexed: false },
+			],
+			rows: [{ id: 2, k: NaN }],
+		});
+		binder._setDatabasesLoader(() => ({ dev: { lefty: left, righty: right } }));
+		const data = await runSql('SELECT l.id FROM dev.lefty l JOIN dev.righty r ON l.k = r.k');
+		assert.deepStrictEqual(data, []);
```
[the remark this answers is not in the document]
kriszyp: Addressed in 464cdfe3d: added `hash join emits matched rows when both sides share a non-indexed equi key` (exercises the build + probe/merge path with real matches) and `hash join LEFT null-fills a non-indexed-key row with no match`, closing the probe/merge and LEFT null-fill coverage gaps.

sqlEngine/physical/PhysicalIndexScan.ts
```diff
@@ -0,0 +1,89 @@
...
+			yield row;
+		}
+	} catch (err) {
+		if (err instanceof EngineUnsupportedError) throw err;
+		if (err instanceof IndexRebuildingError) {
+			throw new EngineUnsupportedError(`index rebuilding: ${err.message}`);
+		}
+		if (err instanceof ClientError) {
+			throw new EngineUnsupportedError(`search rejected scan: ${err.message}`);
+		}
```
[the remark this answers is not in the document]
kriszyp: Addressed in 50f917bb6: added unit tests in `unitTests/sqlEngine/pipeline.test.js` that override the mock table's `search()` to throw a `ClientError` and an `IndexRebuildingError` and assert PhysicalIndexScan converts both to `EngineUnsupportedError` (so `auto` falls back to legacy) — plus a case proving a generic (non-capability) error propagates unconverted.

sqlEngine/executor/runMutation.ts
```diff
@@ -0,0 +1,383 @@
...
+	const context: WriteContext = { user: ctx.user };
+	// The row-finder above is a SELECT, and a plain SELECT is right to hide a row
+	// whose TTL has passed but hasn't been swept yet. An UPDATE isn't a SELECT
+	// though: it's about to overwrite that row's TTL, and every other write surface
+	// (REST PUT/PATCH, ops update) still finds and resets a not-yet-swept-but-expired
+	// row because they load it by id directly, bypassing the freshness check a read
+	// applies. Without includeExpiredRows, the row-finder would drop such a row from
+	// its result, so the UPDATE silently touches nothing and the TTL never resets
+	// (QA-269).
+	const selectCtx: SqlEngineContext = { ...ctx, includeExpiredRows: true };
```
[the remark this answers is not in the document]
kriszyp: Addressed in 50f917bb6: added an integration test in `integrationTests/apiTests/ttl.test.ts` (new `SqlTtl` table, 2s TTL + 1h eviction window so an expired row stays physically present). It asserts both legs — a plain SELECT hides the expired row, while SQL UPDATE and DELETE find it via `includeExpired` (UPDATE resets the TTL so a later SELECT sees it again; DELETE removes it `1 of 1`). Uses distinct ids per leg since a plain SELECT lazily evicts the row it reads.

sqlEngine/physical/PhysicalIndexScan.ts
```diff
@@ -0,0 +1,98 @@
...
+	try {
+		for await (const row of resource.search(target)) {
+			yield row;
+		}
+	} catch (err) {
+		if (err instanceof EngineUnsupportedError) throw err;
+		if (err instanceof IndexRebuildingError) {
+			throw new EngineUnsupportedError(`index rebuilding: ${err.message}`);
+		}
+		if (err instanceof ClientError) {
```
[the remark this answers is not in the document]
kriszyp: Same catch block as the earlier thread (shifted by the rebase) — covered by the new PhysicalIndexScan fallback-conversion tests in 50f917bb6.

sqlEngine/router.ts
```diff
@@ -0,0 +1,67 @@
...
+		(result) => callback(null, result),
+		(err: unknown) => {
+			if (config.engine === 'auto' && err instanceof EngineUnsupportedError) {
+				logger.info(`SQL engine v2 fallback: ${err.reason}`);
+				opts.legacy(opts.statement, callback);
+				return;
+			}
+			callback(err);
+		}
+	);
```
cb1kenobi, line 61: This code feels verbose and could be simplified to:
```typescript
	const config = getSqlEngineConfig();

	if (config.engine === 'legacy') {
		opts.legacy(opts.statement, callback);
		return;
	}

	runStatement({
		variant: opts.variant,
		jsonMessage: opts.jsonMessage,
		statement: opts.statement,
	}).then(
		(result) => callback(null, result),
		(err: unknown) => {
			if (config.engine === 'auto' && err instanceof EngineUnsupportedError) {
				getLogger().info(`SQL engine v2 fallback: ${err.reason}`);
				opts.legacy(opts.statement, callback);
				return;
			}
			callback(err);
		}
	);
```

cb1kenobi (approved):
Reviewed and tested. Pretty clean implementation. Nice fallback support. I'm good with this PR.

sqlEngine/router.ts
kriszyp: Addressed in a35b29fd9: removed the redundant async wrapper and deferred logger lookup to the auto-fallback branch while preserving the existing routing behavior.

kriszyp:
## SQL engine performance: new vs legacy

PLAN.md phase-5 item 6 (*"a subset of the legacy corpus run with timings; budget ≤ legacy time"*) didn't have a harness yet, so I built one and ran it against this branch at `8ba39a5` (i.e. including the routing simplification).

**Result: every measured query shape is faster on the new engine — 16/16, nothing slower, zero result mismatches. Geomean ratio 0.148 at 40k rows (~6.8× faster overall).** The budget is met comfortably.

### Method

Engine selection is process-global (`HARPER_SQL_ENGINE`), so the benchmark boots **two** instances with identical config, fixture and seed — one `legacy`, one `new` — and interleaves them at the **iteration** level (legacy, new, legacy, new, …). A block design would charge any machine drift to whichever engine happened to be running during it; alternating cancels it in the ratio.

Two details worth calling out, because they're the difference between a real measurement and a flattering one:

- **The new side runs `new`, not `auto`.** `auto` silently falls back to legacy on an unplannable query, so timing `auto` would benchmark legacy twice and report it as parity. Under `new`, an unplannable query surfaces as `UNSUPPORTED` instead of hiding in the average.
- **Every query is result-parity checked** (row count + normalized digest) across the two engines before its timings count. A query returning different data on the two engines is a correctness signal and its timing is meaningless. 0 mismatches here, which is an independent (if small) corroboration of the differential suite.

### Results — 40k rows, p50, 25 timed iterations per query per engine

`ratio` = new / legacy p50; **below 1.0 = new engine faster**. HTTP + ops-API floor measured at 0.441ms.

| query | rows | legacy p50 | new p50 | ratio |
| --- | ---: | ---: | ---: | ---: |
| pk-lookup | 1 | 1.689ms | 1.053ms | 0.624 |
| pk-in-list | 5 | 2.035ms | 1.356ms | 0.666 |
| pk-range-small | 10 | 133.599ms | 1.648ms | **0.012** |
| indexed-eq-tiny | 40 | 4.094ms | 1.446ms | 0.353 |
| indexed-eq-medium | 400 | 4.111ms | 2.552ms | 0.621 |
| pk-range-large | 4000 | 160.103ms | 16.573ms | 0.104 |
| range-plus-filter | 565 | 27.911ms | 12.212ms | 0.438 |
| orderby-limit-pk | 20 | 142.147ms | 117.993ms | 0.830 |
| orderby-limit-indexed | 10 | 149.570ms | 2.755ms | 0.018 |
| count-star-indexed | 1 | 3.568ms | 2.768ms | 0.776 |
| group-by | 1 | 3.684ms | 2.098ms | 0.569 |
| distinct | 2 | 3.123ms | 1.919ms | 0.615 |
| inner-join-small | 3 | 36.423ms | 1.719ms | 0.047 |
| inner-join-medium | 120 | 175.819ms | 7.301ms | 0.042 |
| left-join-small | 57 | 171.400ms | 2.396ms | 0.014 |
| like-prefix | 10 | 139.756ms | 1.292ms | 0.009 |
| like-unindexed | — | — | — | `UNSUPPORTED` |

`like-unindexed` is expected and healthy: no index driver, so `new` rejects it and production `auto` routes it to legacy. It's in the battery to pin the fallback boundary.

### Two caveats on reading these numbers

**1. The ratio is a function of table size, not a fixed speedup.** Legacy full-scans anything it can't serve from an index, so its cost grows with the row count while the new engine's index-served plans stay flat. Same 10-row query, three table sizes:

| `pk-range-small` (10 rows) | 1k rows | 10k rows | 40k rows |
| --- | ---: | ---: | ---: |
| legacy | 6.27ms | 34.13ms | 130.33ms |
| new | 1.30ms | 1.56ms | 1.65ms |

That's the actual story of this PR — not "6.8× faster" but "O(index) instead of O(table)". The geomean will keep climbing with scale, so it should always be quoted with a row count.

**2. Anything near the 0.441ms floor is transport-dominated.** `pk-lookup` at 1.05ms vs 1.69ms is a real win, but ratios for sub-millisecond queries are compressed toward 1.0 by HTTP overhead — that's not evidence the engines are equivalent.

### The tiny-query risk didn't materialize

PLAN.md flagged *"a streaming pipeline has more per-row overhead than AlaSQL on inputs that already fit in memory"* as a cutover risk. It doesn't show up: the smallest shapes (`pk-lookup` 0.624, `pk-in-list` 0.666, `group-by` 0.569, `distinct` 0.615) are all comfortably faster. Nothing in the battery regressed.

### One finding: `ORDER BY <pk> LIMIT` is not O(window) (fix in #1844)

`orderby-limit-pk` is the one shape that doesn't behave, and it's worth a look because it's a silent one — it's *faster* than legacy (0.830), so a budget check passes while the intended optimization does nothing. It scales linearly on **both** engines:

| `ORDER BY id LIMIT 20` | 1k rows | 10k rows | 40k rows |
| --- | ---: | ---: | ---: |
| legacy | 5.18ms | 34.72ms | 141.14ms |
| new | 3.89ms | 29.67ms | 118.53ms |

D-219 intends this to be O(window). The cause is an interaction between the two D-219 commits:

- **7a919c24** added the index-order scan and verified O(window) using `ORDER BY <indexed>` — a **secondary** index. That genuinely works; I measured it flat at 1.27ms → 1.13ms across an 8× table growth.
- **50f917bb** then correctly restricted `sortDrivesIndex` to the **primary key**, because a secondary index diverges from legacy on DESC ties and null omission.

The catch is that `Table.search` only adds its order-aligned pseudo-condition when the sort attribute has a *secondary* index — and the primary key doesn't have one. So the rule now fires on exactly the one attribute kind `search` can't serve from index order, and always degrades to full scan + in-memory sort. Driving `Table.search` directly (no SQL) shows it plainly:

```
GET /sqlbench/?sort(id)&limit(20)     ->  404 "id is not indexed and not combined with any other conditions"
GET /sqlbench/?sort(qty)&limit(20)    ->  1.2ms flat across 5k -> 40k
```

Worth flagging: `plan.ts` sets `allowFullScan` for a sort-driven scan, and that's what suppresses this 404 — so the new engine was quietly converting a query `search` would have rejected (→ legacy fallback) into a permitted full scan.

The fix belongs in `Table.search`, not `sortDrivesIndex` — the PK-only restriction in 50f917bb is right; what was missing is `search`'s ability to honor it. Opened separately as **#1844** since it's core search semantics and independent of this PR. With it, on this branch:

| `ORDER BY id LIMIT 20` @ 40k | legacy | new | ratio |
| --- | ---: | ---: | ---: |
| before #1844 | 142.15ms | 117.99ms | 0.830 |
| after #1844 | 141.32ms | **1.58ms** | **0.011** |

Flat across table growth, and the battery geomean goes 0.148 → 0.111. None of the numbers in the main table above include that fix — they're what this PR delivers on its own.

### Benchmark code

`benchmarks/sql-engine/` (runner + fixture + README), following the existing `indexed-write` / `concurrent-rw` pattern and emitting the same machine-parseable `SQL_ENGINE_RESULT` / `SQL_ENGINE_SUMMARY` lines for a future regression gate. Coming as a follow-up PR against `main` so it doesn't hold this one up:

```
node benchmarks/sql-engine/run.mts                  # quick (~2 min)
node benchmarks/sql-engine/run.mts --scale=nightly
```

—
🤖 Benchmarked and written by Claude Opus 4.8 via [Claude Code](https://claude.com/claude-code)

Issue #708, opened 2026-05-21
Blob GC: files not deleted on audit-log expiry when using RocksDB audit store


kriszyp:
## Summary

When a Blob-typed record is deleted via SQL and `auditRetention` is configured, the blob file on disk is never cleaned up if the audit store is backed by RocksDB (the default).

## Root cause

In `dist/resources/auditStore.js`, `scheduleAuditCleanup()` branches on audit store type:

```js
if (auditStore instanceof RocksTransactionLogStore) {
    auditStore.rootStore.purgeLogs({ before: ... });
    return; // <-- returns without blob file cleanup
}
// LMDB path: iterates audit entries and calls removeAuditEntry() which
// invokes deleteCallbacks to remove blob files
```

The RocksDB path calls `rootStore.purgeLogs()` and returns immediately. `removeAuditEntry()` — which is responsible for invoking blob-file delete callbacks — is never called. Additionally, `RocksTransactionLogStore.remove()` is a no-op with a TODO comment:

```js
async remove() {
    // TODO: this function can likely be removed once the call to purgeLogs()
    // is added in `resources/Table.ts`
}
```

## Effect

Blob files accumulate on disk indefinitely after the DB record is deleted, until the schema or table is dropped (which uses a direct filesystem operation, not the GC path).

## Confirmed via

Integration test `Blob lifecycle > blob file removed from filesystem after auditRetention expires` in `integrationTests/apiTests/blob.test.mjs` (PR #648). The test waits up to 35 seconds but blob files never disappear. `drop_schema` and `drop_table` tests that use direct filesystem operations pass fine.

## Workaround

Filesystem cleanup works when the schema or table is explicitly dropped (`drop_schema`, `drop_table` with `drop_records: true` — though the latter also appears incomplete). The audit-log GC path is the only broken case.

Related: #38 (Support running on Bun)

PR #1303, opened 2026-06-15
fix: retain blob files on record invalidate (#1302)

### LINKED ISSUE #1302: RecordEncoder invalidate path unlinks blob files still referenced by live records → silent progressive blob data loss
## Summary

The `invalidate` write path unlinks a row's file-backed blob even though the record remains live and still references it. Subsequent reads of that record throw `ENOENT .../blobs/data/...` / `Blob is incomplete`. Because invalidates are generated automatically and repeatedly (replication revalidation, TTL eviction, peer-sent invalidate events), this causes **progressive, silent blob data loss** on any deployment using file-backed blobs with caching/invalidation or replication.

This is the root cause of a production incident (`canopy.serent`) where ~344 of 345 `CompanyDocument.fileData` attachments (~663 MB) were lost while the records stayed intact — blob ENOENT errors there go back ~7 weeks.

## Mechanism (verified on `main`)

1. `resources/RecordEncoder.ts:726-740` — on any non-`message` write to a `HAS_BLOBS` row, the prior row's blob files are unlinked, retaining only fileIds found in the **new record being written**:

   ```js
   if (existingEntry && existingEntry.value && type !== 'message' && existingEntry.metadataFlags & HAS_BLOBS) {
       let retainedFileIds;
       if (record) findBlobsInObject(record, (blob) => { const id = getFileId(blob); if (id) (retainedFileIds ??= new Set()).add(id); });
       deleteBlobsInObject(existingEntry.value, retainedFileIds);   // unlinks prior blobs not in retained set
   }
   ```

2. `resources/Table.ts:1442-1488` (`_writeInvalidate`) passes `partialRecord`, which is rebuilt to **index-only attributes** (`:1457-1464`):

   ```js
   for (const name in indices) { if (partialRecord[name] === undefined) partialRecord[name] = this.getProperty(name); }
   ```

   A file-backed blob attribute is never an index, so `partialRecord` contains no blob → `retainedFileIds` is empty → `deleteBlobsInObject(existingEntry.value)` unlinks the still-referenced live blob (~`deletionDelay` ms later, `resources/blob.ts:493-505` / `:985-993`). The record persists, now flagged `INVALIDATED` (`Table.ts:1471`), still pointing at the deleted file.

3. **No blob refcount.** `deleteBlobsInObject` walks one object and unlinks; it never checks whether the surviving record value (or any other record) still references the fileId. Blob lifetime is bound to "is this fileId present in the value being encoded right now."

The #641 fix (commit `7b605b01`, the `retainedFileIds` retention at `RecordEncoder.ts:727-738`) closed the **update** case — updating an unrelated attribute on a blob row no longer drops the blob — but did **not** cover **invalidate**, because it keys off the new record's blobs and the invalidate's `partialRecord` never carries the blob.

## Triggers (automatic, recurring)

Invalidates funnel through `_writeInvalidate` from:
- replication / source apply with revalidation — `Table.ts:381-396` (`shouldRevalidateEvents`; incoming `invalidate` events always become `_writeInvalidate`);
- TTL / eviction scheduling — `Table.ts:921-934`.

On a `replicate: true` cluster, peer-driven invalidations recur indefinitely → blobs are shed over time. This matches the canopy.serent signature: loss over weeks, records alive, and surviving blobs clustered by recent mtime (records that get re-put receive a fresh blob file under a new fileId via `encodeBlobsWithFilePath`/`saveBlob`, so they survive; records that only ever receive invalidates lose their original).

## Evidence (`canopy.serent`, harper-pro 5.1.0; read-only investigation 2026-06-15)

- 344/345 `CompanyDocument.fileData` blobs unreadable (ENOENT / Incomplete) on the leader node; records fully intact.
- Blob ENOENT for live records in logs since **2026-04-21**, recurring for weeks (pre-dates all 5.1.0 work — loss began before the #641 fix, so the broader pre-fix *update*-strips-blob bug also contributed; the invalidate gap is the part still unfixed).
- `cleanup_orphan_blobs` never ran (different selection order; ruled out).

## Suggested fix

On `invalidate`, retain the **existing** record's blob fileIds — i.e. seed `retainedFileIds` from `existingEntry.value` (not only the new `record`) — or skip blob deletion entirely for `type === 'invalidate'`. More robustly, blob deletion should be driven by whether any live record version still references the fileId, not by the contents of the single record value currently being encoded.

## Minimal repro

1. Single node, table with a file-backed blob attribute (> 8192 bytes so it lands on disk), `auditLog: true`.
2. `PUT` a record with the blob; confirm `blobs/data/.../<id>` exists and the record reads back.
3. `table.invalidate(id)` (or drive via a replicated put with `shouldRevalidateEvents`, or a TTL eviction/revalidation cycle).
4. ~`deletionDelay` ms later the blob file is unlinked, but the record still exists (now `INVALIDATED`) and still references the fileId → reads throw `ENOENT .../blobs/data/...`.

## Key files

- `resources/RecordEncoder.ts:726-740` (bug site) and `:839-846` (`removeEntry`).
- `resources/Table.ts:1442-1488` (`_writeInvalidate`, index-only `partialRecord`), `:381-396` (replication/source → invalidate), `:921-934` (TTL eviction).
- `resources/blob.ts:493-505` (`deleteBlob`), `:985-993` (`deleteBlobsInObject` retain-set filter).

## Related

- HarperFast/harper-pro#385 — interrupted replicated blob left as corrupt 8-byte stub (a *secondary* effect: a clone backfilling these already-deleted blobs writes header-only stubs).
- HarperFast/harper-pro#386 — observability/durability gap (silent loss, no alert/repair).
- harper#641 (update-case fix, did not cover invalidate); harper#708, #595 (blob GC / orphan sweeper — different paths).

kriszyp:
Fixes #1302.

## Summary
On `type === 'invalidate'`, seed `retainedFileIds` from the existing row's value (`resources/RecordEncoder.ts`) so an invalidate no longer unlinks a still-referenced file-backed blob.

## Why
`invalidate` rewrites a row to an **index-only** partial record (`Table._writeInvalidate`); a blob attribute is never an index, so that partial carries no blob. The #641 retention keyed off the *new* record's blobs, so on invalidate `retainedFileIds` was empty and `deleteBlobsInObject(existingEntry.value)` unlinked the live record's blob ~`deletionDelay` ms later, leaving the row pointing at a missing file (`ENOENT` / "Blob is incomplete").

Invalidates are generated automatically and repeatedly (replication revalidation, TTL eviction, peer invalidate events), so this caused **silent, progressive blob data loss** on any file-backed-blob deployment using caching/invalidation or replication. It is the root cause of the canopy.serent incident (~344/345 `CompanyDocument.fileData` attachments lost while the records stayed intact).

## Where to look
- `resources/RecordEncoder.ts` — the new `type === 'invalidate'` retention block. The change is small; the comment carries the rationale.
- Regression test in `unitTests/resources/blob.test.js` — "invalidating a record does not unlink a still-referenced blob" (file-backed blob survives an invalidate + repeated invalidates). Confirmed it fails on `main` and passes with the fix.

## `_writeRelocate` and other partial-record writers (the requested check)
`_writeRelocate`'s residency-excludes-this-host branch hits the **same** encoder deletion site with an index-only partial and `INVALIDATED` metadata — but with `type === 'relocate'`, so it is deliberately **not** covered here. Unlinking the local blob there is intended: the record's data is relocating off this node (the residency set excludes us), reads route to the residency node, and freeing the local file prevents a storage leak on a non-hosting node. The other relocate branches (`_writeRelocate` resident-here, `_recordRelocate`) pass the full record, so the #641 retention already preserves their blobs. No other writer puts a partial/index-only record over a `HAS_BLOBS` row with a different type. *(Confidence: high — corroborated by both cross-model legs below.)*

## Open tradeoff for the reviewer
After invalidate the row's value is index-only and `HAS_BLOBS` is cleared, so a subsequent delete/overwrite of that row does **not** synchronously reclaim the now-unreferenced blob — reclamation falls to the orphan sweeper (`cleanupOrphans`, #708/#595). This is the explicit scope tradeoff from the issue ("more robustly, blob deletion should be driven by whether any live record version still references the fileId" — refcounting, deferred): we prioritize **never unlinking a live blob** over prompt reclamation. The post-fix orphan is recoverable; the pre-fix loss was not. Both Codex and Gemini independently flagged this same tradeoff and agreed the relocate-path unlink is correct.

🤖 Generated by Claude (claude-opus-4-8, 1M context). Cross-model reviewed (Codex + Gemini).

### DIFF
--- a/resources/RecordEncoder.ts
+++ b/resources/RecordEncoder.ts
@@ -736,6 +736,22 @@ export function recordUpdater(store, tableId, auditStore) {
 						if (fileId) (retainedFileIds ??= new Set()).add(fileId);
 					});
 				}
+				if (type === 'invalidate') {
+					// An invalidate rewrites the row to an index-only partial record (Table._writeInvalidate);
+					// a blob attribute is never an index, so `record` carries no blob and the still-live
+					// record's blob would be unlinked ~deletionDelay ms later, leaving the row pointing at a
+					// missing file. Invalidation is a staleness marker, not a data mutation, so retain every
+					// blob the existing row referenced. The #641 retention (above) keyed off the new record,
+					// which the invalidate's partial never carries — hence the gap. See HarperFast/harper#1302.
+					// Note: the invalidated row no longer references the blob (and loses HAS_BLOBS), so a later
+					// delete/overwrite won't synchronously reclaim it — that falls to the orphan sweeper
+					// (cleanupOrphans, #708/#595). We trade prompt reclamation for never unlinking a live blob;
+					// refcount-driven deletion (track any live version referencing the fileId) is the robust fix.
+					findBlobsInObject(existingEntry.value, (blob) => {
+						const fileId = getFileId(blob);
+						if (fileId) (retainedFileIds ??= new Set()).add(fileId);
+					});
+				}
 				deleteBlobsInObject(existingEntry.value, retainedFileIds);
 			}
 			let result: Promise<void>;
--- a/unitTests/resources/blob.test.js
+++ b/unitTests/resources/blob.test.js
@@ -281,6 +281,52 @@ describe('Blob test', () => {
 		});
 		setDeletionDelay(500); // restore the default
 	});
+	it('invalidating a record does not unlink a still-referenced blob', async () => {
+		// Regression (HarperFast/harper#1302): invalidate rewrites the row to an index-only
+		// partial record, which never carries the blob attribute. The #641 retention keys off
+		// the *new* record's blobs, so on invalidate retainedFileIds was empty and the live
+		// record's blob was unlinked ~deletionDelay ms later, leaving the row pointing at a
+		// missing file. Invalidates are generated automatically and repeatedly (replication
+		// revalidation, TTL eviction, peer invalidate events), so this caused progressive,
+		// silent blob data loss on caching/replicated deployments.
+		setAuditRetention(10);
+		setDeletionDelay(0);
+		const InvalidateTest = table({
+			table: 'BlobInvalidateTest',
+			database: 'test',
+			attributes: [
+				{ name: 'id', isPrimaryKey: true },
+				{ name: 'blob', type: 'Blob' },
+				{ name: 'category', type: 'String', indexed: true },
+			],
+		});
+		const payload = randomBytes(20000); // > FILE_STORAGE_THRESHOLD so it goes file-backed
+		const blob = await createBlob(payload);
+		await InvalidateTest.put({ id: 300, blob, category: 'docs' });
+		const filePath = getFilePathForBlob(blob);
+		assert(filePath, 'expected file-backed blob');
+		assert(existsSync(filePath), 'blob file should exist after initial put');
+
+		// Invalidate the record. Pre-fix: the blob file was unlinked ~deletionDelay ms later
+		// even though the record persists (now flagged INVALIDATED).
+		await InvalidateTest.invalidate(300);
+		await delay(50);
+		assert(existsSync(filePath), 'blob file must survive invalidate of a record that referenced it');
+
+		// Repeated invalidates (as replication/eviction would generate) must not shed it either.
+		for (let i = 0; i < 3; i++) {
+			await InvalidateTest.invalidate(300);
+			await delay(20);
+			assert(existsSync(filePath), `blob file must survive repeated invalidate #${i}`);
+		}
+
+		// The invalidated row is now an index-only partial that no longer references the blob,
+		// so it is no longer reachable through the record. Reclamation of the now-unreferenced
+		// file is left to the orphan sweeper (cleanupOrphans / HarperFast/harper#708, #595) rather
+		// than the write path — the targeted fix here prioritizes never unlinking a live record's
+		// blob over promptly reclaiming one that has become unreferenced.
+		setDeletionDelay(500); // restore the default
+	});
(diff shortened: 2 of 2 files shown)


kriszyp:
Closing in favor of an upstream root-cause fix.

Investigation (with Kris) traced #1302 past the blob-deletion site to its real cause: **replicated writes are being down-converted to invalidates**. Replication registers as an `intermediateSource` (`harper-pro/replication/replicator.ts:355`), but `core/resources/Table.ts:342` derives `shouldRevalidateEvents` from `this.source` — the single canonical *caching* source — rather than from the source the subscription is actually for. So on a cache-sourced **and** replicated table, replicated `put`/`patch` events hit `_writeInvalidate`, which strips the row to an index stub and deletes the file-backed blob that no peer re-supplies → the canopy.serent loss.

The blob deletion on invalidate is *correct* for a genuine cache (the source re-fetches). This PR's retain-on-invalidate approach fixed the symptom one layer too low and regressed a legitimate path — CI caught it: the `cleanupOrphans` unit test flipped to 1 orphan, because a retained-but-dereferenced blob on the invalidated row leaks once its audit reference ages out.

New fix: gate revalidation off the intermediate (replication) source in `Table.ts:342`, so replicated writes apply via `_writeUpdate` and keep their blobs. New PR to follow; branch retained for history.

🤖 Generated by Claude (claude-opus-4-8).

Issue #1137, opened 2026-06-04
Unit test red on main: 're-delivered duplicate must not double-apply the commutative op' (v22 + v26)


heskew:
### Bug Summary

A replication-replay de-duplication unit test is failing on `main` with:

```
AssertionError [ERR_ASSERTION]: re-delivered duplicate must not double-apply the commutative op
```

It failed on **both Node v22 and v26** in the same run, in the resources unit suite (`686 passing … 1 failing`). Failing on two runtimes simultaneously points to either a genuine regression on `main` or a strongly load-sensitive flake — either way it's currently red on `main` and needs triage.

### Evidence (CI)

Unit Test run on `main` — https://github.com/HarperFast/harper/actions/runs/26961236701

- Node v22 job: `686 passing (28s) … 1 failing` → the assertion above.
- Node v26 job: same single failure.

### Notes

- The assertion string is not present on the current release branch I'm working from, so the test was added/changed on `main` recently — likely in the replication / commutative-merge area. It needs to be attributed to its file (candidates: a replication or CRDT/commutative-op test under `unitTests/`) and reproduced locally.
- First triage question: is this a real double-apply-on-redelivery regression (commutative op applied twice when a duplicate is re-delivered), or a test-ordering/timing artifact? If the former, it's a data-correctness bug, not just a flake.

### Steps to Reproduce

1. On `main`, run the resources unit suite (`npm run test:unit:resources`), repeat a few times across Node 22/24/26.
2. Observe the `re-delivered duplicate must not double-apply the commutative op` assertion failing.

### Are you planning to fix this issue?

No, just reporting the issue.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

heskew:
**Status / triage** (to avoid duplicated effort):

Root cause: in `Table.commit()`'s out-of-order reconciliation, a re-delivered write's commutative ops could be folded twice when the audit-log walk stops before reaching `txnTime` (depth cap, or an early break on a not-yet-visible entry) and the keyed `auditStore.get(txnTime)` fallback lags the re-delivery. Timing-sensitive (green in isolation, red under load).

A fix is up as #1147 (currently **Draft**): an up-front `additionalAuditRefs` read-your-writes duplicate check that reliably skips the re-delivered out-of-order write. Its narrow goal was to get CI more consistently green.

[...]

Issue #1114, opened 2026-06-03
Replication full-copy OOM: per-record audit-chain walk in Table.commit scales with audit-log depth


kriszyp:
## Summary

During replication **full-copy** catch-up, a receiving node can OOM because `Table.commit`'s out-of-order-write reconciliation walks the **entire backward audit-history chain for every record synchronously**, buffering intermediate records in the JS heap. The cost scales with audit-history depth, and the walk runs concurrently on all HTTP workers, so a database whose replication transaction log has accumulated a large history becomes effectively un-ingestable under a fixed container memory limit.

Observed on `harper-pro` 5.0.21; the same code path is present on `main` (v5.1.0-beta.1).

## Symptom

- A follower requesting a full copy of a database crash-loops with the container memory cgroup OOM-killer firing (`Killed process … (MainThread) anon-rss ~5.4GB` at a 6 GiB limit).
- Workers log `JavaScript execution has taken too long and is not allowing proper event queue cycling` during ingestion.
- The OOM can be masked at the orchestration layer: the in-container supervisor respawns the killed process, so `docker ps` shows "Up" and `RestartCount` stays frozen — the kills are only visible via `dmesg`.

## Root cause

`Table.commit` reconciles out-of-order / incremental writes by walking the record's audit chain ([resources/Table.ts:1769-1837](resources/Table.ts)):

```js
do {
  while (localTime > txnTime || (auditedVersion >= txnTime && localTime > 0)) {
    const auditRecord = auditStore.get(localTime, tableId, id, nodeId);
    if (!auditRecord) break;
    // …collect succeedingUpdates / additionalAuditRefs…
    localTime = auditRecord.previousVersion;   // walk backward through the version chain
    nodeId = auditRecord.previousNodeId;
  }
  nextRef = auditRefsToVisit.shift();          // then branch into additional audit refs
  if (nextRef) { localTime = auditedVersion = nextRef.localTime; nodeId = nextRef.nodeId; }
} while (nextRef);
```

Each `auditStore.get(localTime, …)` lands in `RocksTransactionLogStore.getSync`, which does a fresh range scan of the transaction log per step ([resources/RocksTransactionLogStore.ts:100-112](resources/RocksTransactionLogStore.ts)):

```js
getSync(key, tableId, recordId, nodeId) {
  if (typeof key === 'number') {
    for (const entry of this.getRange({ start: key, exactStart: true, log: nodeId })) {
[...]

PR #1310, opened 2026-06-16
fix(table): dedup re-delivered out-of-order writes before the resequencing walk (#1114)

kriszyp:
## Summary

Hoist the keyed audit-log duplicate lookup **ahead of** the out-of-order resequencing walk in `Table.commit`. A re-delivered out-of-order write whose exact `(version, nodeId)` is already in the audit log is now dropped in O(1) instead of walking the full audit chain (up to `MAX_OUT_OF_ORDER_AUDIT_DEPTH`) only to discard it in the depth-cap block.

## Why

On a production cluster (#1114), transitive/proxied replication (`viaNodeId`) re-delivers already-applied patches out-of-order. On high-churn records with deep pure-`patch` histories, each re-delivery drove an ~1000-step **synchronous** walk in the commit path → event-loop starvation (`JavaScript execution has taken too long`) → replication ping-timeouts → dropped subscriptions and stalled catch-up.

A live logpoint at the depth-cap site confirmed the mechanism: the capped walks were exact `(version, nodeId)` duplicates (the existing `auditStore.get(...)` matched, `dupVerMatch=true`) — i.e. the dedup the cap block already does, just reached only *after* the full walk. This hoists that lookup before the walk.

It complements the existing guards: the `additionalAuditRefs` front-check (#1137) catches a re-delivery while the ref is still on the record; this keyed lookup catches the case where a later in-order write has since **dropped** the ref, which is exactly the transitive-re-delivery case observed in production. Keyed by `nodeId`, so it's correct across multiple source nodes.

## Where to look

- `resources/Table.ts` — the new `if (isRocksDB)` keyed-dedup block at the top of the `precedesExisting <= 0` audit branch. **RocksDB-only** (LMDB audit entries are keyed by local audit time, not version, so the version-keyed lookup doesn't apply there; LMDB keeps the exact unbounded walk). A keyed-lookup miss simply falls through to the existing walk — **no correctness change**, purely a fast-path.
- `unitTests/resources/transaction.test.js` — new test: a re-delivery *after* a ref-dropping in-order write is deduped without taking the depth-cap path. Verified to fail without the change (it would take the deep walk) and that the record value is identical with or without it.

## Notes for the reviewer

- **Not a 5.1 regression** — the walk, depth cap, `additionalAuditRefs`, and transitive delivery all exist in `v5.0`; the same latent path. Good **backport/patch candidate for v5.0** (the `auditStore.get` lookup it relies on already exists in v5.0's cap block).
- **Open follow-up, not in this PR:** this *absorbs* the redundant transitive re-deliveries cheaply but does not reduce their *volume*. The proxy-resume start-time relay that generates them is a separate workstream worth its own ticket.

<sub>Generated by Claude (Opus 4.8).</sub>

## Review status (open item for the human reviewer)

- **Codex review:** clean (no actionable issues).
- **Gemini review:** attempted twice via `agy`; it failed to produce findings (workspace-resolution flakiness in the worktree; `gemini` CLI unavailable). Compensated with a careful self-review (the change is a faithful hoist of the existing depth-cap-block lookup with the identical identity-tie guard). A second human/model pass on the correctness argument is welcome.
- **`test:unit:main`** could not complete locally — it failed on a storage-path env issue (unrelated to this change; passed once `HDB_ROOT`/`STORAGE_PATH` were set) and then hung on an unrelated `ComponentLoader` integration test. The change is covered by `test:unit:resources` (passing, incl. the new regression test). CI should exercise the full suite.

### DIFF
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -1898,6 +1898,32 @@ export function makeTable(options) {
 								write.skipped = true;
 								return; // out-of-order write already folded into this record
 							}
+							// Up-front keyed dedup (RocksDB): a re-delivered out-of-order write whose exact
+							// (version, nodeId) is already in the audit log is a duplicate that was already applied — skip
+							// it here instead of paying the O(depth) resequencing walk below only to discard it in the
+							// depth-cap block. This is the same keyed lookup that block performs, hoisted ahead of the walk.
+							// It is what catches transitive/proxied re-deliveries: they arrive buried below the record head
+							// (so replication's head-tie fast-skip can't see them) yet are exact duplicates. Keyed by nodeId,
+							// so it is correct across multiple source nodes. RocksDB-only: LMDB audit entries are keyed by
+							// local audit time, not version, so this version-keyed lookup doesn't apply there (LMDB keeps the
+							// exact unbounded walk). A miss (the keyed lookup can lag a back-to-back re-delivery — #1137)
+							// simply falls through to the walk, so this never changes correctness; the additionalAuditRefs
+							// check above remains the read-your-writes guard.
+							if (isRocksDB) {
+								const priorAudit = auditStore.get(txnTime, tableId, id, options?.nodeId);
+								if (
+									priorAudit &&
+									priorAudit.version === txnTime &&
+									precedesExistingVersion(
+										txnTime,
+										{ version: txnTime, localTime: txnTime, key: id, nodeId: priorAudit.nodeId },
+										options?.nodeId
+									) === 0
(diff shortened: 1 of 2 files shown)


kriszyp:
Follow-up filed: the transitive-delivery-reduction work referenced above is now tracked as **HarperFast/harper-pro#399** (reduce the *volume* of proxied out-of-order re-deliveries; this PR *absorbs* them cheaply). 

<sub>— Claude (Opus 4.8)</sub>

PR #1312, opened 2026-06-16
De-flake drop suite: tolerate already-dropped column family and serialize dropTable

kriszyp:
## Summary
De-flakes the intermittent `Drop number table` failure in `Integration Tests 1/6` (`AssertionError: Drop failed: ... Column family already dropped!`, usually followed by a cascading `Database not open` / `Delete operations` timeout). `Table.dropTable()` now (1) tolerates a column-family drop that reports "Column family already dropped!", and (2) on RocksDB runs the drops + catalog removal under the database's `update-attributes` exclusive lock to close a drop/same-name-create race.

## Purpose / root cause
A `drop` is broadcast to every worker thread, and each thread holds its own handle to the same underlying column family. A concurrent worker (or `completeInterruptedDrop`) can drop a shared family first, so the originating `drop_table` then hits "Column family already dropped!" and fails the operation — intermittently, depending on thread scheduling (hence cross-platform / cross-Node flakiness). The family being gone is the intended end state, so the drop tolerates it. Genuine drop failures still surface and leave the tombstone for the reconcile.

## Where to look
- **Spin lock across the drop (please sanity-check the deadlock reasoning).** The RocksDB path serializes the drops + catalog removal under the same `update-attributes` lock the create path / `completeInterruptedDrop` hold. That lock is the synchronous spin lock (CORE-3064); like the create path, it is held **only across synchronous work** — the drops use `dropSync()` (as `completeInterruptedDrop` does) and `await dbisDb.committed` runs *after* the lock is released. Holding the spin lock across an `await` would deadlock, so the locked section is deliberately await-free.
- **Tombstone-guarded catalog removal.** Catalog rows are removed only when this drop's own tombstone is still the live primary row, so a concurrent same-name create's fresh rows are never clobbered. Under the lock this check+remove is atomic w.r.t. the create path.
- **LMDB path** keeps the awaited `drop()` + guarded removal (transactional locking, no shared-CF double-drop). The new clobber-guard unit test is RocksDB-scoped; the LMDB guard path is covered by the existing create-over-tombstone test.

## Pairing
Pairs with **HarperFast/rocksdb-js#656**, which makes the column-family drop idempotent at the source. This harper-side change de-flakes independently and does **not** require the rocksdb-js release/bump to land.

## Tests
`unitTests/resources/dropTableGhost.test.js`: tolerance + the drop/create clobber guard; run under both `test:unit` and `test:unit:lmdb`.

---
Generated by an LLM (Claude Opus 4.8). Cross-model reviewed (Codex + Harper deep-review): the catalog-clobber and the cross-thread drop/create race were both raised and are addressed here (the latter via the lock); the one item left for a human is verifying the spin-lock deadlock reasoning called out above.

### DIFF
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -111,6 +111,15 @@ const CACHEABLE_STATUS_CODES = new Set([200, 203, 204, 206, 300, 301, 308, 404, 
 envMngr.initSync();
 const LMDB_PREFETCH_WRITES = envMngr.get(CONFIG_PARAMS.STORAGE_PREFETCHWRITES);
 const LOCK_TIMEOUT = 10000;
+// Tolerate a redundant column family drop. Drops are broadcast to every worker
+// thread and each holds its own handle to the same underlying family, so a
+// concurrent worker may already have dropped it; the storage engine reports
+// that as "Column family already dropped!". The family being gone is the
+// intended outcome, so swallow that specific error and rethrow anything else.
+function ignoreAlreadyDropped(error: any): void {
+	if (error?.message?.includes('Column family already dropped')) return;
+	throw error;
+}
 // A frozen record we may need to copy-on-mutate before stamping it (records are immutable — decoded
 // records are frozen and 5.2 record caching relies on it). Only plain/record objects qualify: never
 // a Buffer/typed-array (spreading would corrupt the binary into a {0:.., 1:..} object) or a primitive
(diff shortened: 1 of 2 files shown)


heskew:
Codex sanity-checked the spin-lock/deadlock reasoning called out in the PR body. The locked RocksDB section stays synchronous (`dropSync` plus catalog check/removal), and the only `await` happens after `update-attributes` is released, so I do not see a deadlock risk there.

---

Posted by Codex on Nathan's behalf.

heskew (approved):

PR #1323, opened 2026-06-16
fix: dispose component scopes before worker exit in dev, and serialize reloads

dawsontoth:
In `harper dev`, recycling a worker that holds a component with a live native (N-API) runtime — e.g. @harperfast/vite's rolldown dev server — by forced exit or `worker.terminate()` tears the V8 isolate down while that runtime's threads are still running, which crashes the whole process (SIGSEGV/SIGABRT). The safe path is to dispose the runtime first, then exit.

- Scope.close() now invokes its `close` listeners and awaits any promise they return, so a plugin's async teardown (e.g. closing the Vite dev server) completes as part of close() instead of running unawaited after emit().
- New components/scopeShutdown.ts tracks in-flight scope close() promises; componentLoader registers each scope's close on SHUTDOWN, and threadServer awaits whenScopesClosed() between closeServers() and realExit — so the worker disposes its components before it exits.
- componentLoader serializes dev reload cycles: cycles run one at a time and a single coalesced follow-up absorbs everything that arrives mid-cycle, so a burst of file changes (or a regenerate-on-reload loop) can't stack overlapping restarts that pile workers up past the forced-termination grace.
- manageThreads widens the forced-termination grace to 30s in DEV_MODE so a slow-but-progressing disposal isn't force-killed prematurely.

### DIFF
--- a/components/Scope.ts
+++ b/components/Scope.ts
@@ -188,15 +188,17 @@ export class Scope extends EventEmitter<ScopeEventsMap> {
 		// e.g. @harperfast/vite disposing its Vite/rolldown dev server — and the worker shutdown path
 		// awaits this close(), so awaiting the listeners here ensures such a native runtime is fully
 		// disposed before the worker exits. That ordering matters: tearing the worker down while a native
-		// (N-API) bundler runtime is still live crashes the whole process. Listeners that return nothing
-		// (the common case) settle immediately, so existing behavior is unchanged.
+		// (N-API) bundler runtime is still live crashes the whole process. `Promise.all` (not
+		// `allSettled`) so a listener that fails surfaces its error to the caller rather than being
+		// silently swallowed; listeners that return nothing (the common case) settle immediately.
 		const closeListeners = this.listeners('close') as Array<(...args: any[]) => unknown>;
 		this.removeAllListeners('close');
-		await Promise.allSettled(
+		await Promise.all(
 			closeListeners.map((listener) => {
+				// Run every listener (so all teardown is attempted) and bind `this` to the Scope, matching
+				// how EventEmitter invokes listeners — a listener may rely on it (e.g. `this.logger`). The
+				// wrapper turns a synchronous throw into a rejection so `Promise.all` surfaces it too.
 				try {
-					// Bind `this` to the Scope, matching how EventEmitter invokes listeners — a listener may
-					// rely on it (e.g. `this.logger`).
 					return listener.call(this);
 				} catch (error) {
 					return Promise.reject(error);
--- a/components/Scope.ts
+++ b/components/Scope.ts
@@ -184,7 +184,24 @@ export class Scope extends EventEmitter<ScopeEventsMap> {
 
 		await Promise.allSettled([...this.#entryHandlers.map((h) => h.close()), this.options.close()]);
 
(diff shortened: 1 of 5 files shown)


components/Scope.ts
```diff
@@ -184,7 +184,26 @@ export class Scope extends EventEmitter<ScopeEventsMap> {
...
-		this.emit('close');
+		// Invoke `close` listeners and await any promise they return. A plugin's teardown can be async —
+		// e.g. @harperfast/vite disposing its Vite/rolldown dev server — and the worker shutdown path
+		// awaits this close(), so awaiting the listeners here ensures such a native runtime is fully
+		// disposed before the worker exits. That ordering matters: tearing the worker down while a native
+		// (N-API) bundler runtime is still live crashes the whole process. Listeners that return nothing
+		// (the common case) settle immediately, so existing behavior is unchanged.
+		const closeListeners = this.listeners('close') as Array<(...args: any[]) => unknown>;
+		this.removeAllListeners('close');
+		await Promise.allSettled(
```
kriszyp: `allSettled` should not be used without checking results. This should either be `all` (safest default, propagates error), or if you specifically want to catch/handle the error, make sure to log it at least:

```suggestion
const results = await Promise.allSettled(closeListeners.map(...));
for (const result of results) {
    if (result.status === 'rejected') {
        harperLogger.error('Error in scope close listener', result.reason);
    }
}
```

components/scopeShutdown.ts
```diff
@@ -0,0 +1,26 @@
...
+ * happens *before* exit.
+ *
+ * State is module-local, so it is naturally per-worker (each worker is a fresh module realm) and resets
+ * with the worker — scopes are created once at load and closed once at shutdown, so nothing accumulates.
+ */
+const closing = new Set<Promise<unknown>>();
+
+/** Track an application scope's `close()` so the worker waits for it before exiting. Never rejects. */
+export function trackScopeClose(closePromise: Promise<unknown>): void {
+	const settled = Promise.resolve(closePromise).catch(() => {});
```
kriszyp: Likewise, dangerous silencing of errors, either let it propagate or log it.

kriszyp (commented):
I approve with the suggested error handling changes. This is great, valuable improvement in cleanup handling.

components/Scope.ts
dawsontoth: Which do you think we should do?

dawsontoth: I'm thinking propagating the errors is the better choice

components/scopeShutdown.ts
dawsontoth: > For trackScopeClose I went with log, not propagate — and deliberately. It's the shutdown boundary: whenScopesClosed() gates realExit, and the worker must exit even if a scope's cleanup fails. If I let it propagate, the rejection would bubble to threadServer's exit chain, where I'd have to catch-and-log-and-exit-anyway (otherwise it skips realExit and the worker wedges / unhandled-rejects). So propagating further just relocates the same log + exit to the boundary, with more fragility (an unhandled-rejection window before whenScopesClosed attaches). Net: close() propagates (your ask), and that propagation is caught and logged exactly once at the boundary that can't do anything but log and exit.

- Claude

PR #1335, opened 2026-06-17
Add RocksDB transaction log stats to analytics

### LINKED ISSUE #1311: Expose rocksdb-js transaction log stats in get_analytics
rocksdb-js@2.1.0 exposes transaction log stats and now we need to pump them into Harper's analytics system.

cb1kenobi:
## Summary

Records rocksdb-js transaction log (txnlog) stats in `get_analytics` (rocksdb-js@2.1.0), extending the `rocksdb-stats` metric added in #871.

- **`rocksdb-stats`** (DB row) now also carries the summed `txnlog.*` roll-up from `db.getStats()` (the `txnlog*`-prefixed columns), alongside the existing block-cache/memtable/compaction columns.
- **New `rocksdb-txnlog-stats` metric** — one row per transaction log (`database` + `log`), from `db.listLogs()` + `db.useLog(name).getStats()`. Records the performance/resource-relevant fields (backlog: `pendingTransactions`/`uncommittedTransactions`/`replayGapBytes`; footprint: `totalSizeBytes`/`memoryOverlayBytes`/file counts; throughput counters) and drops noise (sequence numbers, log positions, static config, path).
- **New optional `log` filter** on `get_analytics` to narrow `rocksdb-txnlog-stats` to a single log by name.

## Purpose

Closes #1311 — surface transaction-log performance and resource-utilization metrics for observability.

## Where to look

- **`storeRocksDBStatsMetrics`** (`resources/analytics/write.ts`): per-log collection. `listLogs()` is guarded so a store closed/dropped mid-cycle can't abort the whole analytics aggregation; per-`(db, log)` counter baselines are cached and pruned on log removal and database drop; the first sample per log is skipped (mirrors the db-level path) so cumulative counters aren't reported as a spike.
- **Column selection**: `normalizeTxnLogStats` and the `ROCKSDB_TXNLOG_*` constant lists — what's kept vs dropped.
- **`log` filter** (`resources/analytics/read.ts`): rejects `log` with a non-`rocksdb-txnlog-stats` metric (BAD_REQUEST) rather than silently returning nothing.

## Open areas for the reviewer (surfaced by a self-review pass; deliberately not addressed here)

- **Efficiency / cold path**: `useLog(name).getStats()` runs every aggregation cycle (~60s) per log, and the per-db/per-log sweep is synchronous (no `await rest()` like the surrounding aggregation loops). Cold path, but worth a rocksdb-js owner's eye on whether `useLog` can pin mmaps.
- A RocksDB database with transaction logs but **zero loaded tables** emits no txnlog rows (logs are enumerated via the first table's `primaryStore`).
- A **transient per-log `getStats()` error** can make the next period's reported rate ~2× (the log's baseline isn't advanced while the period clock is). Same class of issue exists in the pre-existing db-level path.
- Cleanup deferred: the dedicated `log` param overlaps the generic `conditions` API; the three metric builders are near-identical and could be parameterized.

## Notes

- Docs: `get_analytics` gained a user-facing `log` parameter — a companion docs PR to HarperFast/documentation will follow.
- Generated by an LLM (Claude Opus 4.8).

### DIFF
--- a/components/mcp/tools/schemas/operations.ts
+++ b/components/mcp/tools/schemas/operations.ts
@@ -265,6 +265,10 @@ export const OPERATION_INPUT_SCHEMAS: Record<string, object> = {
 				description: 'ISO 8601 timestamp or epoch ms — inclusive window start.',
 			},
 			end_time: { type: ['number', 'string'], description: 'ISO 8601 timestamp or epoch ms — exclusive window end.' },
+			log: {
+				type: 'string',
+				description: 'Transaction log name to filter on (rocksdb-txnlog-stats metric).',
+			},
 		},
 		required: ['metric'],
 	},
--- a/components/mcp/tools/schemas/operations.ts
+++ b/components/mcp/tools/schemas/operations.ts
@@ -299,6 +299,10 @@ export const OPERATION_INPUT_SCHEMAS: Record<string, object> = {
 				description: 'ISO 8601 timestamp or epoch ms — inclusive window start.',
 			},
 			end_time: { type: ['number', 'string'], description: 'ISO 8601 timestamp or epoch ms — exclusive window end.' },
+			log: {
+				type: 'string',
+				description: 'Transaction log name to filter on (rocksdb-txnlog-stats metric).',
+			},
 		},
 		required: ['metric'],
 	},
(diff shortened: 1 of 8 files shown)


kriszyp (approved):
Clean addition — surfaces the RocksDB txn log stats (log number, live log size, running compaction bytes) through the analytics layer without touching any hot paths. The  mmap-pinning question is a minor edge case worth a follow-up, but nothing blocking.

Issue #1317, opened 2026-06-16
MCP fixes: application-profile 500 (body not async-iterable) + operations-profile conformance/hardening gaps (S1–S4)


kylebernhardy:
Bundle of MCP server fixes found by an external conformance/edge suite (boots a real Harper via `@harperfast/integration-testing`, driven by the official `@modelcontextprotocol/sdk` client + a raw HTTP client). One **P1** that makes the application profile unusable, plus four operations-profile conformance/hardening gaps (**S1–S4**) to clean up in the same pass.

| ID | Sev | Area | One-liner |
|----|-----|------|-----------|
| **P1** | High | application profile | Every request 500s: `body is not async iterable` |
| **S1** | Low–Med | json-rpc | Malformed JSON → framework `400`, not a JSON-RPC `-32700` frame |
| **S2** | Low | pagination | Invalid `cursor` silently accepted instead of `-32602` |
| **S3** | Low | transport | `Accept` header not enforced |
| **S4** | Info/Sec | transport | `Origin` not validated (DNS-rebinding hardening) |

---

# P1 — Application-profile MCP returns 500 on every request

## Summary

Every request to the **application-profile** MCP endpoint (`http(s)://<host>:<httpPort>/mcp`) fails with HTTP **500**:

```
TypeError: body is not async iterable
```

This includes `initialize`, so a session can never be established. The entire application surface is unreachable: auto-generated CRUD tools (`get_`/`search_`/`create_`/`update_`/`patch_`/`delete_`), the `harper://schema/{db}/{table}` resources, and `harper://openapi`. The **operations profile** (Fastify, operations port) is unaffected.

## Reproduction

Boot Harper with the application profile enabled and at least one exported `@table` Resource, then:

```bash
curl -sS -i -X POST 'http://127.0.0.1:9926/mcp' \
  -u admin:'Abc1234!' \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize",
       "params":{"protocolVersion":"2025-06-18","capabilities":{},
                 "clientInfo":{"name":"x","version":"0"}}}'
# → HTTP 500   TypeError: body is not async iterable
```

Content-type is irrelevant — `application/json`, `text/plain`, `application/octet-stream`, and no content-type all 500 identically.

## Root cause

The application adapter async-iterates the request body:

```ts
// components/mcp/adapters/harperHttp.ts:65-72
[...]

PR #1336, opened 2026-06-17
fix(mcp): result-shaped output schemas for application CRUD tools (#1324)

### LINKED ISSUE #1324: MCP application create/update/delete tools: outputSchema vs return-value mismatch breaks strict SDK clients
## Summary

The auto-generated MCP **application-profile** CRUD tools (`create_*`, `update_*`, `patch_*`, `delete_*`) declare an `outputSchema`, but their handlers return a value that doesn't satisfy it — so the official `@modelcontextprotocol/sdk` client rejects the tool result:

```
McpError -32600: Tool create_WorkItem has an output schema but did not return structured content
```

Surfaced while adding the first application-profile integration test in #1317 / PR HarperFast/harper#1320 (the create *operation* itself is fixed there — `makeCreateHandler` now targets the collection so the insert succeeds; this issue is purely the response/output-schema contract).

## Details

Two layered problems in `components/mcp/tools/application.ts` + `components/mcp/tools/schemas/derive.ts`:

1. **No `structuredContent` for scalar returns.** `create_*` registers `outputSchema: deriveCreateOutputSchema(...)` (the full record schema). But the create handler returns the new id as a **bare string** (Harper's `Resource.post`/`create` returns the primary key). `wrapResult` only sets `structuredContent` for objects, so the result has none → the SDK requires `structuredContent` whenever `outputSchema` is present and throws `-32600`.

2. **Record schema marks server-assigned timestamps `required` on create output.** `deriveRecordSchema` adds `@createdTime`/`@updatedTime` (and non-nullable) attributes to `required`. A freshly created record may not have `@updatedTime` set, so even returning the full record may not validate against the create output schema.

Note: Harper's MCP server does **not** validate `outputSchema` server-side — only strict clients (the official SDK) do. Raw JSON-RPC clients are unaffected, which is why PR #1320's integration round-trip drives create/get over raw JSON-RPC.

## Impact

`create_*`/`update_*`/`patch_*`/`delete_*` application tools are effectively unusable from spec-conformant MCP hosts that validate tool output against the advertised `outputSchema` (Claude Desktop, etc.).

## Possible fixes (pick per design)

- Have the create/update handlers return the created/updated **record** as `structuredContent` (re-read after write), and
- Fix `deriveCreateOutputSchema` so create output doesn't `require` fields that aren't guaranteed at create time (e.g. `@updatedTime`), or give create/update/delete result-shaped output schemas (id / affected-record) instead of the full record schema — mirroring how `search_*` deliberately omits `outputSchema` today.

## Acceptance

- [ ] `create_`/`update_`/`patch_`/`delete_` application tools round-trip via the official `@modelcontextprotocol/sdk` client without `-32600`.
- [ ] Output schemas match what the handlers actually return.
- [ ] Integration coverage via the SDK client (not just raw JSON-RPC).

kylebernhardy:
Fixes #1324. **Stacked on #1320** (base branch is `fix/1317-mcp-application-500-and-conformance`) — review/merge #1320 first.

## Summary

The auto-generated MCP **application-profile** CRUD tools advertised the full-record `outputSchema`, but their handlers return a *result envelope*. The official `@modelcontextprotocol/sdk` client validates results against the advertised schema and rejected every call with `-32600` ("has an output schema but did not return structured content"):

- `create_*` returned a bare primary key (scalar) → no `structuredContent` at all.
- `update_*`/`patch_*` returned `{ok:true}`, which the full-record schema (with `required` server-assigned fields like `@updatedTime`) doesn't satisfy.

So these tools were unusable from spec-conformant hosts (Claude Desktop, etc.). Harper's server doesn't validate `outputSchema` — only strict clients do — which is why #1320's round-trip used raw JSON-RPC.

## Fix

Advertise what the handlers actually return, mirroring the existing `delete`(boolean)/`search`(omit) precedent:

| Tool | Returns | `outputSchema` |
|---|---|---|
| `create_*` | `{ id }` | `{ id: <pk type> }` |
| `update_*`/`patch_*` | `{ ok: true }` | `{ ok: boolean }` |
| `delete_*` | `{ deleted: <bool> }` | `{ deleted: boolean }` |

Handlers wrap their **scalar** return into the matching object so `structuredContent` is present. `get_*` keeps the full-record schema (it genuinely returns the record). The #1317 integration round-trip is flipped from raw JSON-RPC to the **SDK client**, which now validates the contract end-to-end.

## Where to look

- `components/mcp/tools/application.ts` — `isStructuredEnvelope()` guard so a **custom Resource** that returns a structured object (typically with a `static outputSchemas` override) passes through unchanged rather than being re-wrapped. This guard is the main correctness surface; it's applied symmetrically to create/update/patch/delete.
- `components/mcp/tools/schemas/derive.ts` — `deriveRecordSchema` is now used by `get_*` only; create/update/patch/delete get result-shaped schemas. This also fixes the secondary bug where a freshly-created record could lack a `required` `@updatedTime`.

🤖 Generated by an LLM (Claude Opus 4.8).

### DIFF
--- a/components/Scope.ts
+++ b/components/Scope.ts
@@ -184,7 +184,28 @@ export class Scope extends EventEmitter<ScopeEventsMap> {
 
 		await Promise.allSettled([...this.#entryHandlers.map((h) => h.close()), this.options.close()]);
 
-		this.emit('close');
+		// Invoke `close` listeners and await any promise they return. A plugin's teardown can be async —
+		// e.g. @harperfast/vite disposing its Vite/rolldown dev server — and the worker shutdown path
+		// awaits this close(), so awaiting the listeners here ensures such a native runtime is fully
+		// disposed before the worker exits. That ordering matters: tearing the worker down while a native
+		// (N-API) bundler runtime is still live crashes the whole process. `Promise.all` (not
+		// `allSettled`) so a listener that fails surfaces its error to the caller rather than being
+		// silently swallowed; listeners that return nothing (the common case) settle immediately.
+		const closeListeners = this.listeners('close') as Array<(...args: any[]) => unknown>;
+		this.removeAllListeners('close');
+		await Promise.all(
+			closeListeners.map((listener) => {
+				// Run every listener (so all teardown is attempted) and bind `this` to the Scope, matching
+				// how EventEmitter invokes listeners — a listener may rely on it (e.g. `this.logger`). The
+				// wrapper turns a synchronous throw into a rejection so `Promise.all` surfaces it too.
+				try {
+					return listener.call(this);
+				} catch (error) {
+					return Promise.reject(error);
+				}
+			})
+		);
+
 		this.removeAllListeners();
 
 		return this;
--- a/components/componentLoader.ts
+++ b/components/componentLoader.ts
@@ -27,6 +27,7 @@ import harperLogger from '../utility/logging/harper_logger.ts';
 import * as dataLoader from '../resources/dataLoader.ts';
 import { restartWorkers, getWorkerIndex } from '../server/threads/manageThreads.js';
 import { resetRestartNeeded, subscribeToRestartRequests } from './requestRestart.ts';
+import { trackScopeClose } from './scopeShutdown.ts';
 import { scopedImport } from '../security/jsLoader.ts';
 import { server } from '../server/Server.ts';
 import { Resources } from '../resources/Resources.ts';
@@ -469,7 +470,9 @@ export async function loadComponent(
 						origin
 					);
 
-					onMessageByType(ITC_EVENT_TYPES.SHUTDOWN, () => scope.close());
+					// Track the close so the worker's shutdown path waits for it (and thus for any async
+					// native-runtime disposal, e.g. @harperfast/vite's dev server) before calling realExit.
+					onMessageByType(ITC_EVENT_TYPES.SHUTDOWN, () => trackScopeClose(scope.close()));
 
 					await sequentiallyHandleApplication(scope, extensionModule);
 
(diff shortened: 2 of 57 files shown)


kriszyp:
The integration test failures here are all **artifact download errors** across every shard — not test failures. The issue appears to be stale/poisoned artifact storage on this run. Could you push a new commit (or empty `git commit --allow-empty`) to trigger a fresh CI run with clean artifacts? The code looks fine otherwise.

kylebernhardy:
@kriszyp thanks — tried that, but it didn't clear it. Pushed an empty commit to get a fresh run with clean artifacts; the build uploaded a fresh 161 MB artifact successfully (`successfully finalized`), yet every integration shard still failed *downloading* it — `Unable to download artifact(s)… after 5 retries`, dying ~27s into the Azure blob transfer. Repeated with full re-runs (regenerating the artifact each time) and by merging this branch's base in — same result every time.

So it isn't stale storage on the original run; it's the blob **download** failing. Tellingly, runs on `main` and on the base branch (#1320) during the same window downloaded their artifacts fine, so the artifact service is healthy in general — #1336's runs just keep landing on flaky blob hosts (`productionresultssa2/13/14`).

(The `transaction.test.js` #1114-dedup unit failure on the latest run is an unrelated pre-existing flake — v24 only; this PR only touches `components/mcp/**`.)

The code itself is validated — unit/build/lint green, and Codex + Gemini cross-model reviews both clean. The reliable unblock looks like landing #1320 (green, and your review items are addressed) and rebasing this onto `main`, which sidesteps the stuck stacked run entirely. Let me know if you'd prefer a different approach.

<sub>🤖 AI-generated (Claude), posted by Kyle.</sub>

kylebernhardy:
@kriszyp update — CI is fully green now. Here's how it resolved:

**Artifact-download failures:** these turned out to be intermittent Azure-blob degradation that *this PR's stacked runs* kept losing the coin flip on — builds uploaded a valid 161 MB artifact every time, but the download died ~27s in (`after 5 retries`), across fresh artifacts and multiple blob hosts, while `main` and #1320 runs in the same window downloaded fine. Once #1320 merged, I merged `main` into this branch so it runs on a clean (non-stacked) base — and the artifact step has passed cleanly since (30/30 integration shards green).

**The more important find:** with integration finally *running*, the #1324 SDK-client round-trip caught a real operational bug — `update_`/`patch_` threw `Cannot read properties of undefined (reading 'directURLMapping')`. `makeUpdateHandler` was invoking the verb method **detached** from its class (`const fn = ResourceClass.put; fn(...)`), so the static Resource dispatcher ran with `this === undefined`. `get_`/`create_`/`delete_` already called the method on the class; update was the lone exception, and the unit mocks never referenced `this` so they missed it. Fixed by calling `ResourceClass.put!/patch!` directly (`this` preserved) + a regression test with a `this`-dereferencing handler. The SDK round-trip now passes for all four verbs.

(The transient `transaction.test.js` #1114 and `blob.test.js` unit failures along the way were unrelated pre-existing flakes — different test/runtime each run — and cleared on re-run.)

Net: result-shaped output schemas (create→{id}, update/patch→{ok}, delete→{deleted}) + the `this`-binding fix; Codex + Gemini both clean. Ready for your review.

<sub>🤖 AI-generated (Claude), posted by Kyle.</sub>

kriszyp (approved):
Output schema/handler contract correctly aligned for all CRUD tools; `this`-binding fix is a good catch.

Issue #1337, opened 2026-06-17
Base-copy replication wedges in `Receiving` when sender has records referencing blob files missing on disk


ldt1996:
## Summary

When a v5 node attempts a base-copy resync to a peer, and the sender's primary store contains any record whose `Blob` field references a file that doesn't exist on the sender's disk, the receiver's connection wedges permanently in `lastReceivedStatus: "Receiving"`. Base-copy never reaches `COPY_COMPLETE`, the watermark never advances, the blob-bearing database's replication is fully stuck. No errors at the API level; symptom appears only as silent record propagation loss between specific peer-pairs.

## Minimal reproduction

1. Start one v5 node, deploy a fixture with a `Blob` field, insert N records with content above `FILE_STORAGE_THRESHOLD` (8192 bytes) so blobs go to disk files
2. Delete a contiguous block of blob files directly from `<dataRoot>/blobs/<db>/`
3. Add a fresh v5 peer; receiver requests base-copy resync
4. Observe `cluster_status.connections[0].database_sockets[].lastReceivedStatus` stays at `"Receiving"` indefinitely; only records before the gap apply

```
[blob-gap] source has 60 blob files
[blob-gap] deleted 40 contiguous blob files (simulating v4 gap)
[blob-gap] t=0s receiver status=Receiving record_count=7/60
[blob-gap] t=85s receiver status=Receiving record_count=7/60
```

Migration is NOT required for reproduction — the bug fires on any v5 source whose primary store references a missing blob file. We initially suspected the v4→v5 migration path but ruled it out: migration logs are clean, and the reproduction does not use a v4 binary.

## Test results

| State | Receiver status after ~60s | Record count |
|---|---|---|
| Pre-fix | `Receiving` (stuck) | 7/60 |

Test asserts post-fix behavior. Same setup, same parameters, deterministic.

## Cross-references

- harper-pro#352 — separate transient shared-structure misread surface at boot.

kriszyp:
**Diagnosis: the trigger is application code calling `blob.save()`, which isn't a Harper API — not a core defect.**

Pulled the logs from the affected cluster (`cluster:v4-to-v5 customer:akamai`). The whole thing is one base-copy resync window (2026-06-17 00:47–02:37 UTC), and the order of events is:

1. `TypeError: blob.save is not a function` — 1,266×, on `se-sto`, thrown *while resolving PageCache records* (the cache source-resolution path in `Table.ts`), immediately **before** the timeout storm.
[...]

PR #1340, opened 2026-06-17
fix(encoder): strict-extension CAS in sharedStructures save

ldt1996:
## Summary

Under concurrent encoder writes the `saveStructures` CAS check at `resources/RecordEncoder.ts` only verifies length match, so two writers with the same baseline length can independently mint different shapes at the same struct id. One save lands on disk first; the other writer's already-encoded records on disk now reference that struct id expecting a different shape. On decode msgpackr reads the bytes against the wrong shape, fills in the wrong field names from the now-current dictionary populated with values from the original record, and throws the production error `Data read, but end of buffer not reached <partial>` — byte-for-byte the call chain seen in the cluster logs for #1337.

This adds a content comparison after the existing length check: any save whose new structures array does not match the existing entries on a prefix is rejected, msgpackr's `pack()` loop sees the false return, reloads structures from disk, and re-mints the new shape at the next free id.

## Format-aware comparison (refinement, commit 7cc690ca3)

The first version of the check iterated `existingStructures.length` and indexed `structures[i]`, which is wrong as soon as structon's `prepareStructures` upgrades the wire format from a plain Array (named-only) to a `Map` with `named` and `typed` keys the moment the first typedStruct is minted. On a Map, numeric indexing returns `undefined`; the `!Array.isArray(undefined)` guard then rejected every legitimate Array→Map upgrade. Observed in a live patched cluster: decode errors stopped (good — confirms the CAS does catch the race) but during-window catchup stalled — receivers silently re-packed against the new disk state, hit the same false-reject on the next save, and never durably applied the record.

The refinement normalizes both sides through a small `toShape` helper that handles all three wire forms (Array, Map, or plain object after a `mapsAsObjects` round-trip) and runs a recursive `shapeEqual` so identical typed entries (arrays of `[type, size, key]` triples from different literals) compare equal. The strict-extension check then runs against the named and typed arrays independently.

## Test coverage

`unitTests/resources/recordEncoder-structuresMismatch.test.js` — 11 cases:

- failure mode reproduces the production error chain end-to-end
- working case still decodes when writer/reader share a store
- CAS rejects same-length replacement at index 0
- CAS catches replacement at a non-zero index
- CAS accepts strict suffix extension
- CAS rejects array→non-array type mismatch
- **CAS accepts Array→Map upgrade when named entries match** (the regression that broke during-window catchup in production)
- CAS rejects a Map upgrade that mutates an existing named entry
- CAS accepts an append to typed structures in Map form (reference-equality sub-array trap)
- CAS rejects a typed-structure replacement at an existing id (concurrent typedStruct collision)
- CAS handles plain-object form (Map round-tripped via `mapsAsObjects: true`)

End-to-end retry-flow coverage lives in harper-pro's `multiShapeReplicationConvergence.test.mjs` which exercises the full encode→save→CAS→re-pack→decode loop against a real RocksDB-backed encoder.

## Scope

This prevents future corruption. It does not address lost replication messages from the rolling upgrade window or records that are already corrupted on disk in deployed clusters; those need separate work.

## Migration path is not affected

`bin/copyDb.ts` writes structures via `targetRootStore.putSync([STRUCTURES_KEY, storeName], binaryBuffer)` directly (lines 403, 406), bypassing `saveStructures` entirely. The per-record re-encoding step explicitly installs a no-op `saveStructures` on both `existingEncoder` and `tempEncoder` (lines 458-462), so msgpackr's `pack()` does not hit the patched code during migration either. The strict-extension CAS only runs on runtime writes.

## Validated locally

- [x] `npx mocha unitTests/resources/recordEncoder-structuresMismatch.test.js` (11/11)
- [x] `npx mocha unitTests/resources/recordEncoder.test.js` (9/9)
- [x] `node --test integrationTests/cluster/multiShapeReplicationConvergence.test.mjs` in harper-pro with this branch wired as the core submodule — passes on the pre-refinement commit; needs re-run with 7cc690ca3 to confirm the format-aware comparison doesn't regress the typed-shape coverage

## Live-cluster signal (pre-refinement bundle)

Bundled the pre-refinement build to the v4→v5 mixed cluster: `Data read, but end of buffer not reached` errors stopped on the upgraded node, confirming the CAS does close the corruption race. During-window record catchup did not converge, which surfaced the Array→Map false-reject bug above. The refinement targets that gap.

## Not validated locally, needs CI

- [ ] `integrationTests/cluster/cloneFromLegacy.test.mjs` early-returns without `HARPER_LEGACY_VERSION_PATH`. CI has a v4 binary; defensive belt to confirm the post-migration runtime path also stays healthy with the strict-extension CAS in place.
- [ ] `unitTests/bin/copyDB.test.js` early-returns when `STORAGE_ENGINE !== 'lmdb'`. Same scenario, same delegation.

Refs #1337.

### DIFF
--- a/resources/RecordEncoder.ts
+++ b/resources/RecordEncoder.ts
@@ -50,6 +50,37 @@ const MISSING_STRUCTURE_METRIC = 'decode-missing-structure';
 const MISSING_TYPED_STRUCTURE_PREFIX = 'Could not find typed structure ';
 const MISSING_CLASSIC_STRUCTURE_PREFIX = 'Record id is not defined for ';
 
+// Normalize a shared-structures payload into {named, typed} arrays. Accepts the three wire forms
+// the encoder hands us: a plain Array (named only), a Map with 'named'/'typed' keys (structon's
+// typedStructs form), or a plain object with the same keys (Map round-tripped via mapsAsObjects).
+function toShape(s: any): { named: any[]; typed: any[] } {
+	if (!s) return { named: [], typed: [] };
+	if (s instanceof Map) return { named: s.get('named') || [], typed: s.get('typed') || [] };
+	if (Array.isArray(s)) return { named: s, typed: [] };
+	if (typeof s === 'object') return { named: s.named || [], typed: s.typed || [] };
+	return { named: [], typed: [] };
+}
+
+// Structural equality for the leaf shapes structures hold: primitive (named entries are strings,
+// typed entries are arrays of [type, size, key] triples). Recurses through arrays only.
+function shapeEqual(a: any, b: any): boolean {
+	if (a === b) return true;
+	if (!Array.isArray(a) || !Array.isArray(b) || a.length !== b.length) return false;
+	for (let i = 0; i < a.length; i++) {
+		if (!shapeEqual(a[i], b[i])) return false;
+	}
+	return true;
+}
+
+// next must contain every existing entry in the same slot, in order; new slots may be appended.
+function isStrictExtension(existing: any[], next: any[]): boolean {
+	if (next.length < existing.length) return false;
+	for (let i = 0; i < existing.length; i++) {
+		if (!shapeEqual(existing[i], next[i])) return false;
+	}
+	return true;
+}
+
 export function isMissingStructureError(error: any): boolean {
 	const message = error?.message;
 	return (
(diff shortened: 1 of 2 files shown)


ldt1996:
Closing without merging.

The investigation into #1337 (https://github.com/HarperFast/harper/issues/1337#issuecomment-4728204041) found that the production symptom this PR was framed against — record loss during v4→v5 rolling upgrades — was a measurement artifact: records that appeared missing immediately after upgrade were present on re-check after sufficient quiescence. Stock 5.1.3 converges correctly given enough time.

The strict-extension CAS in this PR is defensible code (content comparison is strictly more rigorous than the existing length-only check) but doesn't address an actually-observed bug. A surgical encoder probe (documented in the issue retraction) showed msgpackr's pack() retry chain plus structon's isCompatible side-effect already absorb the CAS rejection correctly — the encoder never produces records that diverge from disk state.

Closing rather than keeping draft indefinitely is the honest disposition. The diff stays in the PR history; if a real concurrent saveStructures race surfaces later, this can be reopened.

PR #1341, opened 2026-06-17
fix(blob): clean up replication-received blobs when a source apply is skipped (harper-pro#406)

kriszyp:
## Summary

Fixes a replicated-blob **orphan leak** (harper-pro#406). On a replication/source apply, the record's blobs are saved out-of-band by `receiveBlobs` *before* the apply runs. If that apply then loses a version conflict (`precedesExistingVersion <= 0` → `write.skipped`), the received blob was never unlinked — `startPreCommitBlobsForRecord` only tracked `saveBeforeCommit` blobs in `write.savedBlobs`, so the skip/abort cleanup had nothing to remove. On an active multi-node caching cluster these accumulate: an audit-aware `cleanup_orphan_blobs` sweep on one preprod node reclaimed **1059 true-orphan blobs** (~84% of its blob files).

## Changes

- **`resources/blob.ts`**
  - `startPreCommitBlobsForRecord(record, store, saveInRecord?, trackPersistedBlobs?)`: when `trackPersistedBlobs` is set, also track an already-saved (fileId-set) blob so the skip/abort cleanup can unlink it. Gated deliberately (see below).
  - `cleanupUnusedBlobs(blobs, retainedFileIds?)`: never deletes a blob whose fileId the committed (winning) record still references.
  - new `collectRetainedFileIds(record)` helper.
- **`resources/Table.ts`**: `preCommitBlobsForRecordBefore` forwards the flag; the `_writeUpdate` call site passes `options?.isNotification`.
- **`resources/LMDBTransaction.ts` (×2)** and **`resources/DatabaseTransaction.ts` (×3)**: the skip/abort cleanup sites pass `collectRetainedFileIds(write.store.getEntry(write.key)?.value)`.

## Where to look

- **The gating decision is load-bearing for data integrity.** Already-saved blobs are tracked *only* on the source/replication apply path (`isNotification`). A received blob is owned by that one write (a unique fileId saved for it). A **local** write that carries an already-saved blob instead references *another row's* blob — and deleting that on abort/skip would corrupt the owning record. (Cross-model review caught exactly this in the first cut, where tracking was unconditional; this gating + the retained guard is the fix.) `pack()` already rejects reusing a blob across records, but the rejection throws *after* `savedBlobs` is populated, so the resulting abort would have triggered the bad delete — hence gating, not just relying on the guard.
- **Pre-existing consideration, not addressed here (worth a reviewer opinion):** cross-node fileId collision. fileIds are per-origin-node sequences, so two records from different origins can share a fileId string and thus the same on-disk blob path. That's a latent collision independent of this change (the second receive already overwrites the first). This fix's skip-delete is same-origin-safe (a received blob's fileId is unique to its record) and the retained guard covers the same-key case, but a cross-key cross-node collision is not something this PR solves. Flagging in case you want a follow-up issue.

## Tests

- `unitTests/resources/blob.test.js`: 3 new tests — `trackPersistedBlobs` gating, the `retainedFileIds` guard (deletes non-retained, keeps retained), and `collectRetainedFileIds`. Existing "updating an unrelated attribute does not unlink a still-referenced blob" still passes.
- Full `unitTests/resources/**` suite: **815 passing, 0 failing**. Lint clean.

## Review provenance / caveats

- Cross-model review ran via Codex (twice). The first pass flagged a real **data-loss blocker** (unconditional tracking would delete a cross-row-shared blob on abort) — fixed by the `isNotification` gating above; Codex re-reviewed the revision and confirmed the blocker is resolved with no new findings.
- **Caveat:** the Gemini (`agy`) leg timed out/produced no output both rounds, and the worktree-isolated `reviewer` subagent could not launch (harness `WorktreeCreate` hook error). I did the Harper-domain pass myself instead. So this change has had Codex + author domain review but **not** the Gemini or Opus-reviewer legs — worth a careful human data-integrity pass.

🤖 Generated by Claude (Opus 4.8). Fix for harper-pro#406.

### DIFF
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -1,4 +1,4 @@
-import { cleanupUnusedBlobs } from './blob.ts';
+import { cleanupUnusedBlobs, collectRetainedFileIds } from './blob.ts';
 import { Transaction as LMDBTransaction } from 'lmdb';
 import { getNextMonotonicTime } from '../utility/lmdb/commonUtility.ts';
 import { ServerError } from '../utility/errors/hdbError.ts';
@@ -343,7 +343,8 @@ export class DatabaseTransaction implements Transaction {
 							// commit succeeded; clean up files for any writes whose commit-handler took an early-return.
 							// deferred until here so a retry that *would* have referenced the blob can flip skipped back to false first.
 							for (const write of this.writes) {
-								if (write?.skipped && write?.savedBlobs) cleanupUnusedBlobs(write.savedBlobs);
+								if (write?.skipped && write?.savedBlobs)
+									cleanupUnusedBlobs(write.savedBlobs, collectRetainedFileIds(write.store.getEntry(write.key)?.value));
 							}
 							// now reset transactions tracking; this transaction be reused and committed again
 							this.writes = [];
@@ -402,7 +403,8 @@ export class DatabaseTransaction implements Transaction {
 					);
 				}
 				for (const write of this.writes) {
-					if (write?.skipped && write?.savedBlobs) cleanupUnusedBlobs(write.savedBlobs);
+					if (write?.skipped && write?.savedBlobs)
+						cleanupUnusedBlobs(write.savedBlobs, collectRetainedFileIds(write.store.getEntry(write.key)?.value));
 				}
 				this.writes = [];
 				if (this.#context?.resourceCache) this.#context.resourceCache = null;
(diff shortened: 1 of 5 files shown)


kriszyp:
CI note: every check is green except **Integration Tests 4/6 (Windows, Node.js v24)**, which has been persistently slow/stuck on this PR — it hung ~19 min on the first run, and the fresh-runner re-run is again running long while all *other* integration shards (Bun, Node v22/v24/v26, and the other 5 Windows shards) and the full unit suite pass. This looks like Windows-runner flakiness on that one shard, not the change (the diff is core blob/transaction logic with no platform-specific behavior; the earlier v24 unit segfault was also a confirmed flake that passed on re-run). Flagging so it doesn't block review — re-trigger that shard if it doesn't settle.

— Claude (Opus 4.8)

PR #1344, opened 2026-06-17
fix: writeHead iterable response headers as an object, not a nested array

dawsontoth:
## Summary

The non-deferred response path in `server/http.ts` passed `Array.from(headers)` to `ServerResponse.writeHead` when the response headers were iterable (a `Map`/`Headers` of `[name, value]` pairs). `Array.from` yields **nested** `[[name, value], …]` tuples, but `writeHead`'s array form expects a **flat** `[name, value, name, value]` list — so Node reads a tuple as a header *name* and throws:

> `TypeError [ERR_INVALID_ARG_TYPE]: The "name" argument must be of type string. Received an instance of Array`

which surfaces as a 500. Fixed by converting iterables with `Object.fromEntries`, matching the sibling `deferWriteHead` branch's `setHeader` loop (preserves array values, last-wins on duplicate names). Applied in **both** the success path and the `onError` fallback — the fallback had the same bug, so a failed primary `writeHead` couldn't even emit a proper error response.

## How it surfaced

A plain GET that the `@harperfast/vite` dev server doesn't serve is handed back to Harper (REST fall-through). That response carries iterable headers and hits the non-deferred `writeHead`, 500ing. It was previously masked by an unrelated 401 at the plugin's dev-server auth gate; once that gate was fixed, the fall-through reached this code and 500'd.

## Verification

- Repro: `GET /Build` (a trivial fall-through resource) returns **500 → 200** with this fix.
- The fall-through test in `@harperfast/vite`'s integration suite (`harper dev … falls through to Harper resources`) goes from a 500 to passing; together with the plugin's loopback-auth fix the suite is fully green (**7/7**).

## Note for reviewers (pre-existing, unrelated)

`npx tsc --project tsconfig.build.json` currently fails on `resources/analytics/write.ts` (`@harperfast/rocksdb-js` has no `TransactionLogStats` / `getStats`) on `main` with or without this change — looks like a rocksdb-js dependency that needs bumping. `server/http.ts` is `// @ts-nocheck`, so this change isn't type-checked regardless. Flagging so the red build step isn't attributed to this PR.

### DIFF
--- a/server/http.ts
+++ b/server/http.ts
@@ -17,7 +17,7 @@ import { createSecureServer } from 'node:http2';
 import { createServer as createSecureServerHttp1 } from 'node:https';
 import { createServer, IncomingMessage } from 'node:http';
 import { Request, BunRequest, isBun } from './serverHelpers/Request.ts';
-import { appendHeader, Headers } from './serverHelpers/Headers.ts';
+import { appendHeader, Headers, toWriteHeadHeaders } from './serverHelpers/Headers.ts';
 import { Blob } from '../resources/blob.ts';
 import { recordAction, recordActionBinary } from '../resources/analytics/write.ts';
 import { Readable, Writable } from 'node:stream';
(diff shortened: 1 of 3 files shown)


dawsontoth:
going to add some unit tests

dawsontoth:
Done. The failing integration tests appear to be unrelated to the changes here, Claude and I audited them to verify. Various retries on them still don't succeed.

cb1kenobi (approved):

kriszyp (approved):
Correct fix — `Object.fromEntries` collapses the iterable into a flat header object that `writeHead` expects. Both call sites patched and tested.

Issue #40, opened 2025-11-20
Database level permissions


kriszyp:
### Feature Summary

We would like to be able to define permissions at the database level that would apply to all tables in the database unless there is specific table-level permission defined.

### Problem This Solves

Currently defining constrained permissions (outside of just super user privilege) is rather laborious since it has be defined for each table. This will hopefully make it simpler.

### Proposed Solution

Here is proposed example permission definition:
```
             "dev": {
                "read": true, # default read permission for all tables in dev
                "insert": false, # default insert permissions...
                "update": true,
                "tables": {
                    "dog": { # overrides default permission for schema to define permissions for dog table
                        "read": true,
                        "insert": true, 
                        "delete": false,
                        "attribute_permissions": [
                            {
                                "attribute_name": "name",
                                "read": true,
                                "insert": true,
                                "update": true
                            }
                        ]
                    }
                }
```
An open question might be whether we treat a table override as the full set of permissions or merge database + table level. In the example above, does the user have update permission on dev.dog? Probably faster if no (no merging), but might be easier to define permissions to do merging and say yes.

### Alternatives Considered

Defining flags on roles, and using those flags to programmatically determine permissions in the endpoint code is very real alternate approach for application endpoints, but this can provide a more native permission system.

### Priority/Impact

None

### Are you planning to fix this issue?

No, just reporting the issue

### First-time contributor support

- [ ] I'm new to contributing and would appreciate guidance on the process
- [ ] I'd like help understanding the project structure
- [ ] I need assistance with setting up the development environment
- [ ] I'm comfortable contributing, but new to this project specifically

PR #1345, opened 2026-06-17
test(resources): cover search_by_value during active runIndexing

ldt1996:
Adds a regression guard that exercises the canonical lookup-then-write loop while runIndexing is in flight on a populated table. The test seeds rows, toggles @indexed on the lookup attribute via a schema reload to start the rebuild, and asserts that search_by_value throws "not indexed yet" for searches issued inside the rebuild window, so callers can rely on the transient error being observable and distinguishable from a hit/miss outcome.

### DIFF
--- a/unitTests/resources/searchDuringActiveIndexing.test.js
+++ b/unitTests/resources/searchDuringActiveIndexing.test.js
@@ -1,0 +1,130 @@
+/**
+ * Regression guard for the period during which `runIndexing` is in flight on a
+ * populated table after an attribute is freshly marked @indexed. While
+ * `dbi.isIndexing === true`, search_by_value on that attribute throws "is not
+ * indexed yet" (resources/search.ts). A caller that wraps the search in a
+ * catch-all and treats every error as a business-level skip will silently drop
+ * its entire workload for the length of the rebuild window.
+ *
+ * The test exercises the canonical "lookup-then-write" loop against an
+ * indexed attribute whose index is mid-rebuild and asserts that the lookup
+ * is observably failing for the calling code so consumers can distinguish
+ * an infra-level transient from a hit/miss outcome.
+ */
+
+const { setupTestDBPath } = require('../testUtils');
+const { loadGQLSchema } = require('#src/resources/graphql');
+const { transaction } = require('#src/resources/transaction');
+const assert = require('node:assert/strict');
+
+describe('search_by_value on an attribute whose index is mid-rebuild', function () {
+	this.timeout(120000);
+
+	before(async function () {
+		setupTestDBPath();
+		await loadGQLSchema(`
+			type Rule @table {
+				id: ID! @primaryKey
+				path: String
+				target: String
+			}
+		`);
+	});
+
+	it('throws "not indexed yet" for searches issued while runIndexing is in flight', async function () {
+		const Rule = tables.Rule;
+		const N = 60000;
+
+		for (let chunkStart = 0; chunkStart < N; chunkStart += 1000) {
+			await transaction((ctx) => {
+				const promises = [];
+				for (let i = chunkStart; i < Math.min(chunkStart + 1000, N); i++) {
+					promises.push(Rule.put({ id: `seed-${i}`, path: `/p/${i}`, target: `/t/${i}` }, ctx));
+				}
+				return Promise.all(promises);
+			});
+		}
+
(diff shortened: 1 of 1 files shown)


kriszyp:
CI failures are pre-existing (`subscriptionReplay.test.js:1048` — unrelated cursor-delivery flap on both v22 and v26). The new test itself passes on all runners.

One structural concern before merging: the race-window assertion `results.notIndexed > 0` relies on 60,000 sequential lookups being slower than `runIndexing` completing. That's a timing bet — on faster hardware or after a future rebuild speedup this will produce a false negative (the race window was missed, not a real failure). A more robust approach would be running the lookups concurrently so the window is guaranteed to overlap, or intercepting/pausing `runIndexing` at the start of the loop.

Two smaller suggestions:
- Add `assert.ok(Rule.indexingOperation, 'indexingOperation must be set after schema reload')` before the loop to make the test self-diagnosing if the schema reload path changes.
- Consider testing the `target` attribute separately, not just `path`, since the PR description frames this as a general regression guard.

Worth a quick discussion on the timing robustness before landing — otherwise looks good.

Ethan-Arrowood (approved):
Looks good — nice to have explicit coverage that the `not indexed yet` transient is observable and distinguishable from a hit/miss outcome.

That said, I want to double down on Kris's structural point: the `results.notIndexed > 0` assertion is a timing bet — 60k sequential lookups outrunning `runIndexing`. It passes today, but it'll flake to a false negative on faster hardware or after any future rebuild speedup, which is exactly the kind of thing that erodes trust in the suite. I'd really like to see the window made deterministic before this lands — running the lookups concurrently so overlap is guaranteed, or intercepting/pausing `runIndexing` at the start of the loop, as Kris suggested.

Approving since it's test-only and green, but please consider hardening the timing — otherwise it's a future flake waiting to happen.

kriszyp:
Rebased this branch onto current `main` (was 149 commits behind; the failing "Unit Test (Node.js v22/v26)" checks were from a stale 2026-06-17 run) and pushed to get a fresh CI run. Fresh CI still fails, but **not on this PR's test** — the new `searchDuringActiveIndexing` test passes (`✔ throws "not indexed yet" for searches issued while runIndexing is in flight`). The single failure is an unrelated, pre-existing test:

```
Subscription replay > edge cases >
  audit records of types not in ACTIONS_OF_INTEREST are still delivered by cursor
AssertionError: expected at least 1 event for id=1, got 0
  at unitTests/resources/subscriptionReplay.test.js:1048
```

**Why it's this PR and not `main`:** recent `main` Unit Test runs are all green, and `main` at the exact commit this branch was rebased onto (62b9a81f6) passed. The failure appears only on this branch, and it reproduced across v22/v24/v26 in two consecutive runs — so it's a real interaction, not a one-off flake.

**Root cause (cross-test interaction):** mocha runs `unitTests/resources/**/*.js` in one process in alphabetical order, so `searchDuringActiveIndexing` (#40) runs before `subscriptionReplay` (#44). The new test seeds **60,000 rows** into the shared default database (`setupTestDBPath` maps `data`/`dev`/`test`/`test2` to one physical path) and triggers a full index rebuild, then does no teardown. The downstream `subscriptionReplay` test subscribes with `startTime: 1` (epoch) and drains with a fixed **100 ms** `collect()` quiet-window. On slower/contended CI runners the replay over the now-much-larger shared store doesn't deliver the id=1 events inside that 100 ms, so `collect` returns empty → `got 0`. It does not reproduce locally (faster hardware) even under full-suite load, which is consistent with a timing boundary rather than a logic bug.

**Suggested fixes (either resolves it):**
1. *Isolate this PR's test* so its heavy writes don't pollute the shared store the neighbor replays over — give the `Rule` table its own dedicated database, and/or add an `after()` that tears the DB down (`tearDownMockDB` / `resetDatabases`). This keeps the neighbor's timing assumptions intact and is the more correct isolation fix.
2. *Harden the neighbor* — `subscriptionReplay`'s line-1048 test should poll until the expected events arrive (e.g. `waitFor` / a longer or event-count-gated collect) instead of a fixed 100 ms window. This file already carries several flakiness-remediation commits (#1219, #762), so this is a pre-existing fragility this PR merely exposes.

I'd lean toward (1) since it keeps the blast radius inside this PR. Happy to push either change if you'd like — I left the branch rebased and did not modify your test.

— KrAIs (Claude Opus 4.8)

PR #1351, opened 2026-06-17
fix(migration): await in-flight blob saves before declaring v4→v5 migration done

ldt1996:
## Summary

In `bin/copyDb.ts`, the v4→v5 migration writes each record via `written = encodeBlobsWithFilePath(() => targetDbi.put(...), ...)`. The blob-save side effect — `saveBlob` setting up `writeBlobWithStream` whose async `pipeline()` returns a promise on `storageInfo.saving` — is fire-and-forget. Nowhere in the migration loop, the per-database scope, or the outer driver are those `storageInfo.saving` promises collected or awaited. `await written` only waits for the RocksDB record commit.

The result: records are committed to the target DB while blob pipelines are still mid-write, and when the per-database scope's `finally` calls `targetRootStore.close()`, any blob pipelines still in flight are abandoned. Failures (e.g. source stream ENOENT, write contention) are silently dropped. The contiguous ~3,815-file gap observed under `sem/0/4/` on the `v4-to-v5.akamai.stage` cluster (see #1337) matches a burst of in-flight saves at the moment the loop ended.

## Fix

This PR exposes a per-migration tracking window in `resources/blob.ts`:

- `beginPendingMigrationBlobSaves()` returns a list that `saveBlob` pushes every in-flight `storageInfo.saving` promise into
- `endPendingMigrationBlobSaves()` closes the window

In `copyDb.ts`, the per-database migration opens the window before the record-copy loop and, after the loop completes, awaits all collected promises via `Promise.allSettled` **before** writing the remote-ids mapping (the \"this DB is migrated\" signal) and **before** the `finally` calls `targetRootStore.close()`. If any blob save rejected, the migration throws — `migrateOnStart`'s outer handler catches the throw, leaves the LMDB source intact (does not move it to backup), and preserves the `migrateOnStart` flag for retry on next start. This makes the migration retryable instead of silently producing a database with missing blob files.

## Test

`unitTests/bin/copyDbBlobAwait.test.js` — three cases:

1. `encodeBlobsWithFilePath` inside a `begin/end` window collects the in-flight blob save promise
2. A blob save whose source stream errors propagates as a rejection in the tracked list (so the migration can detect and fail loudly)
3. After `endPendingMigrationBlobSaves()`, subsequent saves do NOT land in the prior list

All 3 pass.

## Scope, and what this does NOT address

This is the **upstream-cause** fix for the missing-blob-file state observed in #1337. It prevents new clusters from being produced in that state by every v4→v5 migration. It does **not**:

- **Recover existing stuck clusters** (e.g. `v4-to-v5.akamai.stage`). Those clusters already have records on disk referencing fileIds whose files were never written; this PR doesn't clean them up. They need either a separate sender-side base-copy skip (to let replication converge past the bad records) or the records to expire via TTL.
- **Address the receiver-side uncaughtException storm** from orphaned blob streams. That's covered by harper-pro#409 (Kris's receiver-side teardown hardening).

The three fixes are complementary: this PR (no new bad clusters), harper-pro#409 (process stability when bad records are encountered), and a sender-side guard (existing-cluster recovery). All three are needed for full resolution.

## Cross-references

- harper#1337 (root issue with full causal chain)
- harper-pro#409 (Kris's receiver-side hardening)
- #1340 (closed without merge; defended a separate CAS path that wasn't the cause)

### DIFF
--- a/bin/copyDb.ts
+++ b/bin/copyDb.ts
@@ -519,6 +519,11 @@ export async function copyDbToRocks(sourceRootStore, sourceDatabase: string, tar
 		console.log('migrated database ' + sourceDatabase + ' to RocksDB');
 	} finally {
 		endPendingMigrationBlobSaves();
+		// If the migration threw before we awaited pendingBlobSaves above, in-flight save
+		// promises in the list have no rejection handler attached. Attach a no-op catch so a
+		// later background failure is silently observed instead of crashing the process via
+		// Node's unhandledRejection.
+		for (const saving of pendingBlobSaves) saving.catch(() => {});
 		transaction.done();
 		targetRootStore.close();
 	}
--- a/bin/copyDb.ts
+++ b/bin/copyDb.ts
@@ -14,7 +14,11 @@ import { updateConfigValue } from '../config/configUtils.js';
 import * as hdbLogger from '../utility/logging/harper_logger.ts';
 import { RocksDatabase, type RocksDatabaseOptions } from '@harperfast/rocksdb-js';
 import { RocksIndexStore } from '../resources/RocksIndexStore.ts';
-import { encodeBlobsWithFilePath } from '../resources/blob.ts';
+import {
+	beginPendingMigrationBlobSaves,
+	encodeBlobsWithFilePath,
+	endPendingMigrationBlobSaves,
+} from '../resources/blob.ts';
 import { RecordEncoder, setNextEncoding, lastMetadata, METADATA } from '../resources/RecordEncoder.ts';
 
 export async function compactOnStart() {
@@ -411,6 +415,13 @@ export async function copyDbToRocks(sourceRootStore, sourceDatabase: string, tar
 
 	let written;
 	let outstandingWrites = 0;
+	// Open a blob-save tracking window for this database's migration. saveBlob inside
+	// encodeBlobsWithFilePath pushes every in-flight save promise into `pendingBlobSaves` so we
+	// can await them before declaring the database migrated. Without this, fire-and-forget blob
+	// writes could be left mid-pipeline at migration end, producing records in the target DB
+	// referencing fileIds whose files were never durably written — exactly the missing-blob-file
+	// state that triggers the base-copy resync wedge in harper#1337.
+	const pendingBlobSaves = beginPendingMigrationBlobSaves();
 	const transaction = sourceDbisDb.useReadTransaction();
 	try {
 		for (const { key, value: attribute } of sourceDbisDb.getRange({ transaction })) {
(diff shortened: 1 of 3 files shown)


ldt1996:
@gemini-code-assist Addressed all three substantive points in f5e7d2329:

1. **High — `finally` block missing rejection handler**: now attaches `.catch(() => {})` to every promise still in `pendingBlobSaves` in the `finally`, so a background save that rejects after the migration threw is silently observed instead of crashing the process via Node's `unhandledRejection`.

2. **Memory accumulation on large migrations**: `saveBlob`'s tracking now wraps `storageInfo.saving` in `.then(() => list.splice(self, 1))` — resolved saves are removed from the list as they complete, so only in-flight + rejected promises stay. The `.then` has no rejection handler, so a failed save still propagates the rejection and the migration's later `Promise.allSettled` observes it.

3. **Test style: `await createBlob(...)`**: applied to all four sites. `createBlob` is `await`-style in `unitTests/resources/blob.test.js` so this matches the existing convention. (Note: `createBlob` is imported, not a global — `unitTests/resources/blob.test.js:22` does the same.)

Format check fixed via `npx prettier --write` on the affected files. The remaining lint failure looks like a repo-wide ESLint v9 config-file issue, not something this PR can resolve (ESLint reports `no eslint.config.(js|mjs|cjs)` file found).

All 3 unit tests still pass after the changes.

ldt1996:
@claude Good catch. Fixed in 888d25fa1 — dropped the node:test imports entirely, Mocha now sees all 3 tests:

```
encodeBlobsWithFilePath blob-save tracking (harper#1337)
  ✔ collects in-flight blob save promises into the begin/end window
  ✔ a failed blob save (source stream errors) shows up as a rejection in the tracked list
  ✔ endPendingMigrationBlobSaves stops collecting; new saveBlob calls do NOT land in the prior list

3 passing
```

The other failing CI items don't look related to this PR:

- **Integration Tests 4/6 (v22/v24/v26)**: failures are in `mergeHeaders preserves multiple Set-Cookie headers` and `Set-Cookie with comma in expiration date is preserved` — HTTP header tests, no migration/blob involvement
- **runLinter**: ESLint v9 reports `couldn't find an eslint.config.(js|mjs|cjs) file` — repo-level config issue, fails the same way on `git checkout main && eslint .`

Will let CI re-run.

ldt1996:
the unit test failure  is not caused by this change

kriszyp (approved):
Correct fix — awaiting in-flight blob saves before declaring migration complete prevents data loss on the v4→v5 path. CI failure is pre-existing. LGTM! 🟢

PR #1352, opened 2026-06-17
Flag structure updates per transaction log so replicas decode against current structures (#1348)

### LINKED ISSUE #1348: Replicated records fail to decode (structon 'end of buffer not reached') — cluster-wide, silent degradation
## Summary

On a production cluster, a subset of **application records fail to decode** on read with:

```
[error]: Error decoding record Error: Data read, but end of buffer not reached 65
    at checkedRead (msgpackr/unpack.js:217:10)
    at RecordEncoder.unpack (msgpackr/unpack.js:96:12)
    at RecordEncoder.decode (msgpackr/unpack.js:168:15)
    at RecordEncoder.decode (structon/index.js:99:37)
    at <anonymous> (core/resources/RecordEncoder.ts:408:16)
    at decodeFromDatabase (core/resources/blob.ts:968:9)
    at RecordEncoder.decode (core/resources/RecordEncoder.ts:404:37)
    at Store.decodeValue (@harperfast/rocksdb-js/src/store.ts:448:24)
    at <anonymous> (@harperfast/rocksdb-js/src/dbi.ts:196:24)
    at async AppResource.get (.../app-component/src/resources/AppResource.js:137:22)
```

The msgpackr decode reads what it believes is a complete value but leaves ~65–66 trailing bytes in the buffer (`checkedRead` → "Data read, but end of buffer not reached"). The value is decoded against a **structon** typed structure that appears to consume **fewer fields than the record actually encodes**, so the cursor stops short of `end`.

Harper version: **5.1.3** (harper-pro). Observed across the cluster.

## Scope — cluster-wide, pre-existing, not migration-related

This is **not** caused by the recent LMDB→RocksDB migration of one node. Every native-RocksDB node (never migrated) exhibits it, at **higher** rates than the migrated node:

| Node | "Error decoding record" / 6h |
|---|---|
| node A | 815 |
| node B | 493 |
| node C | 153 |
| node D (migrated) | ~3 / 25 min (lowest) |

On `node A`: **477 errors in 2h across 90 distinct record signatures** — i.e. ~90 specific records, each re-failing on **every** read (13–22 hits each in 2h). The same signatures appear on multiple nodes (e.g. `4279ed6e73b26fde…` logged on both `node A` and `node D`), confirming the affected records **replicate** in their broken-to-decode form rather than being node-local corruption.

## Impact — silent degradation (not an outage)

The triggering path is an application resource (`AppResource.get`), which fans out lookups under `Promise.allSettled` and applies an empty fallback on rejection:

```js
const [aRes, bRes, cRes, dRes] = await Promise.allSettled([...]);
const getValue = (res, fallback) => (res.status === 'fulfilled' ? res.value : fallback);
const partA = getValue(aRes, []);   // decode failure -> []
```

So a decode failure **does not surface as a client error** — the affected field is silently dropped and the response is served incomplete (and the key may be re-enqueued as a cache-miss). Net effect: ~90 popular keys **lose the cached optimization** on every request and do not self-heal.

## Hypothesis

A **structon typed-structure desync**: the structure resolved at decode time has fewer fields than the structure the record was encoded with, leaving trailing bytes. Candidate mechanisms (need confirmation):

- Typed structures (`_saveTypedStructures` / `_ensureTypedStructures` in `structon`) not persisted/replicated consistently with the records that reference them, so a reader resolves a stale/base structure.
- A structure evolved (field added) after some records were written, and the older records now resolve against the wrong structure id.

`RecordEncoder.ts:404` decodes `buffer.subarray(position, end)` with explicit length `end - position`; the systematic per-record (not all-records) failure points at record/structure-specific data rather than a global offset bug.

## Repro material (raw bytes, from logs)

Each log line ends with `data: <hex>` (the undecodable value). Samples (note the common `42 … 0e0000 40000000 …` framing; the trailing string bytes are redacted):

```
4279ed6dbaf519330e0000400000000c42bb <string-bytes…>
4279ed653cd43e890e0000400000000642d937 <string-bytes…>
4279ed6e73b26fde0e0000400000000d41d934 <string-bytes…>  (seen on 2 nodes)
```

Full samples are recoverable from any node's `docker logs` (filter `Error decoding record`, the 11th stack line carries `data:`).

## Suggested next steps

1. Decode one captured `data:` payload against the node's stored structon typed structures to confirm the field-count mismatch.
2. Determine whether the typed structures for these records are present/consistent across nodes (replication of typed structures vs records).
3. Decide remediation: re-encode the affected records vs a decode-time compatibility path in structon for under-resolved structures.

---
*Filed from a live production investigation. Compiled by Claude (Opus 4.8) for Kris; figures are from `docker logs` at the time of writing.*

kriszyp:
## Summary
`RocksTransactionLogStore.put()` now re-derives `HAS_STRUCTURE_UPDATE` per `(log, tableId)` from the monotonic `structureVersion`, advancing the per-`(log, table)` watermark only on durable commit.

## Purpose — fixes #1348
On a production cluster, replicated records intermittently decode to `null` with `Error decoding record: Data read, but end of buffer not reached`. The application swallows it (`Promise.allSettled` + empty fallback), so it surfaces as silent degradation — affected records serve incomplete.

Root cause: `HAS_STRUCTURE_UPDATE` is a **one-shot, process-wide-per-table** flag — `saveStructures` sets it on a structure mint and the *next* audit write to that table consumes it (`RecordEncoder.ts`). But transaction logs are partitioned **per origin node** (`logById`) and interleave entries from every table, so the flag can be recorded in a *different* per-node log than the one whose entries first **reference** the new structure. A linear reader of that other log — the replication send path, which reloads structures on the flag (`harper-pro replicationConnection.ts`) — then never resends `TABLE_FIXED_STRUCTURE`, and the peer decodes later entries against a stale structure set → undecodable / null.

The fix makes every per-node log **self-describing**: any entry advancing a `(log, table)`'s structure count carries the flag, regardless of where the one-shot flag was consumed.

## Where to put attention
- **Commit-deferred watermark** (`put()` + a unified commit hook). I folded the existing `aftercommit` block into a single per-transaction `onCommit` that does both the emit *and* the watermark advance, so the watermark rises only on a durable commit — an aborted/discarded entry must not raise it and suppress a later entry's flag. This store is the sole setter of `transaction.onCommit`. Please confirm the `aftercommit` semantics read as preserved (the resources suite, incl. `auditLog.test.js`, passes).
- **Keyed by `(log, tableId)`, not per-log.** A per-node log interleaves tables while `structureVersion` is per-table; a log-wide watermark would let a high-version table suppress another table's first use of a new structure. (Caught in review.)
- **`Uint8Array` entry branch is intentionally not re-derived** — pre-encoded entries already carry the correct flag; re-deriving would break cross-node byte parity.
- **In-memory watermark** resets to 0 on restart → at most one redundant, idempotent structure resend per `(log, table)`; never an under-flag.
- **Same-count `ascii8→string8` promotion is deliberately not flagged** — it is decode-equivalent (ASCII bytes are valid UTF-8 and both types decode via the same `readString` path in `structon`), so it needs no resend. New structures (count increases) are the only decode-relevant change.

## Validation
- New unit tests (`transactionLogStructureUpdate.test.js`): monotonic flagging, per-log independence, per-`(log,table)` independence, commit-deferred/abort-safety, flag preservation. `test:unit:resources` green.
- End-to-end replication recovery is best confirmed by the harper-pro cluster integration tests — flagging that as the place to validate the full path, since this PR is core-only.

Generated by Claude (Opus 4.8). Cross-model reviewed (Codex + Gemini); the two findings they raised (per-table keying, commit-deferral) are addressed in the diff.

### DIFF
--- a/resources/RocksTransactionLogStore.ts
+++ b/resources/RocksTransactionLogStore.ts
@@ -2,6 +2,7 @@ import { TransactionLog, RocksDatabase, shutdown, type TransactionEntry } from '
 import { ExtendedIterable } from '@harperfast/extended-iterable';
 import { getIdOfRemoteNode } from './nodeIdMapping.ts';
 import { Decoder, readAuditEntry, ENTRY_DATAVIEW, AuditRecord, createAuditEntry } from './auditStore.ts';
+import { HAS_STRUCTURE_UPDATE } from './RecordEncoder.ts';
 import { endIteratorOnCorruptFrame } from './replayLogsGuards.ts';
 import { isMainThread } from 'node:worker_threads';
 import { EventEmitter } from 'node:events';
@@ -42,6 +43,13 @@ export class RocksTransactionLogStore extends EventEmitter {
 	updates = 0; // the number of updates to the list of logs that have occurred
 	rootStore: RocksDatabase;
 	reusableIterable = true; // flag indicating that iterable can be reused to resume iterating through audit log
+	// Highest structureVersion appended to each per-node TransactionLog, tracked per tableId. Drives the
+	// per-log HAS_STRUCTURE_UPDATE flag in put(). Keyed by (log, tableId): a per-node log interleaves entries
+	// from every table, but structureVersion is per-table (each table has its own encoder / structure
+	// dictionary), so a single log-wide watermark would let a high-version table suppress the flag for
+	// another table's first use of a new structure. In-memory only: a reset to 0 on restart at most re-flags
+	// the first entry per (log, table) — a redundant, idempotent structure resend, never an under-flag.
+	structureVersionByLogTable: WeakMap<TransactionLog, Map<number, number>> = new WeakMap();
 	constructor(rootDatabase: RocksDatabase) {
 		super();
 		this.log = rootDatabase.useLog('local');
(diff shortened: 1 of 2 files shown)


kriszyp:
Heads up for reviewers: the red **Integration Tests 4/6 (Node v22/24/26)** is **pre-existing on `main`**, not from this change. It fails on `integrationTests/apiTests/headers.test.mjs` (cookie/session assertions) on the Linux Node runners while passing on Bun and Windows; the same shard is red on recent `main` runs (e.g. the base commit `f0d4dae4`). This PR only touches `resources/RocksTransactionLogStore.ts` + its unit test — all unit/lint/format/build checks pass, and `test:unit:resources` is green locally.

_(Generated by Claude — Opus 4.8.)_

PR #1360, opened 2026-06-17
fix(headers): group multi-valued Set-Cookie for writeHead

ldt1996:
`toWriteHeadHeaders` (introduced in fc5e5f3 / PR #1344) used `Object.fromEntries` to convert iterable Headers into a writeHead-compatible object, but `Object.fromEntries` is last-wins on duplicate keys. The Headers iterator yields one `[name, value]` pair per Set-Cookie value (per spec), so three appended Set-Cookies collapsed to one on the wire — which is what's failing in `integrationTests/apiTests/headers.test.mjs` (`mergeHeaders preserves multiple Set-Cookie headers`: expected 5, got 1). Replaced the shortcut with an explicit loop that groups duplicate keys into arrays, which `writeHead` already accepts (`{name: ['v1', 'v2']}` → separate header lines). Two unit tests added in `Headers.test.js` cover the grouping and the round-trip through a real `http.createServer`.


kriszyp (approved):
Correct Set-Cookie grouping fix — Object.fromEntries last-wins collapse is the root cause and this addresses it cleanly. LGTM! 🟢

PR #1366, opened 2026-06-18
debug(test): capture JS stack on v24 segfault

ldt1996:
Wires segfault-handler into mocha.init.js (gated on HARPER_TEST_SEGFAULT_CAPTURE env var) so when the Node v24 LMDB-pass mdb_cursor_close finalizer crashes, we get a JS stack pointing at the actual leaking cursor. Not for merge.

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -40,6 +40,9 @@ jobs:
       - name: Install dependencies
         run: npm install --ignore-scripts
 
+      - name: Build segfault-handler binding (debug helper)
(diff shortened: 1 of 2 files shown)


ldt1996:
Closing — proved the segfault is flaky (passes ~50% of reruns with no code changes). Crash is in lmdb's mdb_cursor_close during V8 GC at process exit. segfault-handler wouldn't help because it's a native finalizer crash, no JS stack to capture. Needs upstream lmdb-js fix or a targeted cursor cleanup in test teardown — neither doable from here without a Linux repro.

PR #1374, opened 2026-06-18
fix(databases): complete the dead restartNumber reindex-recovery trigger

### LINKED ISSUE #1359: Dead `restartNumber` reindex trigger in table loading (never fires)
## Problem
`attributeDescriptor?.restartNumber < workerData?.restartNumber` (`resources/databases.ts:1295` and `:1304`) appears to be **dead code**. `attribute.restartNumber` is never assigned anywhere in the table-loading path, so the persisted `attributeDescriptor.restartNumber` is always `undefined`, and `undefined < N` is always `false`. The restart-based reindex trigger therefore never fires.

## Two possibilities
- It is **redundant** with the adjacent `indexingPID !== process.pid` crash-recovery check → remove it to reduce confusion; or
- A `restartNumber` was meant to be **persisted on completion** (e.g. in `runIndexing`'s success path) and that write is missing → it is a latent bug and the trigger should be made to work.

## Evidence
- Comparison sites: `resources/databases.ts:1295`, `:1304`
- No assignment to the attribute descriptor: `grep '\.restartNumber =' resources/databases.ts` → none

## Ask
Maintainer decision on which of the two it is, then either remove the dead clause or add the missing persistence.

Related: #1354

---

heskew:
## Summary

`table()` re-triggers a secondary-index backfill when the persisted attribute descriptor's `restartNumber` is older than the current restart generation. This is the PID-reuse-robust complement to the `indexingPID` check: it recovers an index left incomplete by a worker that crashed mid-backfill, even when the replacement worker reuses the old PID. The bug: `restartNumber` was never persisted onto the descriptor, so the comparison was always `undefined < N` (false) and the trigger was dead.

This persists `restartNumber` symmetrically with `indexingPID` — set on backfill start, carried across metadata-only changes during an in-progress build, cleared on clean completion (retained on park/crash).

Resolves #1359 (part of #1354).

## Where to look

- `resources/databases.ts` — `table()` (set/carry + the two trigger sites) and `runIndexing()` (clear on clean completion).
- **Operand decision (worth a look):** the trigger compares against `workerData?.restartNumber ?? manageThreads.restartNumber` — the owning worker's *stable startup* generation, falling back to the maintained `manageThreads` counter on the main thread. Cross-model review (Codex) flagged that stamping the *mutable* `manageThreads.restartNumber` is wrong during a worker's shutdown/drain window: the global counter has already advanced to the replacement generation, so an index built then would be stamped with the replacement's generation and the replacement worker would see an equal generation (+ a possibly-reused PID) and never recover it. Using the spawn generation avoids that.

## Back-compat

Existing descriptors have no persisted `restartNumber` → `undefined < N` is `false`, so upgrading does not trigger a reindex storm. On the main thread the generation is `1`, so `1 < 1` is `false` (inert until an actual restart bumps the counter).

## Tests

`unitTests/resources/indexRestartNumber.test.js` (RocksDB-only): (1) `restartNumber` is persisted on a parked/incomplete index and cleared after a clean completion; (2) a backfill re-triggers when the persisted generation is older than the current one, and not when equal. Re-trigger is asserted via `indexingOperation` promise identity (the promise carries across `resetDatabases`, so truthiness is unreliable — mirrors the existing F1 test).

---

### DIFF
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -1287,12 +1287,20 @@ export function table<TableResourceType>(tableDefinition: TableDefinition): Tabl
 				JSON.stringify(stripSearchOnly(attributeDescriptor?.indexed)) !==
 					JSON.stringify(stripSearchOnly(attribute.indexed));
 			if (attribute.indexed) {
+				// The restart generation that owns any in-progress build of this index. Use the
+				// worker's stable startup generation (workerData.restartNumber), NOT the mutable
+				// manageThreads counter: during a worker's shutdown/drain the global counter has
+				// already advanced to the replacement generation, so stamping that would make the
+				// replacement worker see an equal generation (and a possibly-reused PID) and skip
+				// crash-recovery, leaving the index stuck. Falls back to manageThreads.restartNumber
+				// on the main thread, where workerData is undefined (and it is initialized to 1).
+				const currentRestartGeneration = workerData?.restartNumber ?? manageThreads.restartNumber;
 				const dbi = openIndex(dbiKey, rootStore, attribute);
 				if (
 					changed ||
 					attributeDescriptor?.indexingFailed ||
 					(attributeDescriptor?.indexingPID && attributeDescriptor?.indexingPID !== process.pid) ||
-					attributeDescriptor?.restartNumber < workerData?.restartNumber
+					attributeDescriptor?.restartNumber < currentRestartGeneration
 				) {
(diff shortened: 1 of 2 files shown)


kriszyp (approved):
Dead restartNumber path correctly wired up, good crash-recovery test coverage. LGTM! 🟢

PR #1385, opened 2026-06-18
ci: run Next.js adapter integration suite against harper PRs (downstream gate)

kriszyp:
## What

Adds `.github/workflows/integration-tests-nextjs.yml`, an advisory downstream check that runs the `@harperfast/nextjs` Playwright suite against the harper build from the current PR whenever adapter-relevant paths change.

The workflow explicitly checks out `HarperFast/nextjs@main`, packs the current harper PR, installs that tarball into the adapter test workspace, installs fixture dependencies, and runs the suite on Node 20, 22, and 24. Keeping the checkout in this caller avoids reusable-workflow checkout context selecting the harper repository and looking for adapter scripts in the wrong `package.json`.

Current harper and `rocksdb-js` require Node 22.18 or newer. The Node 20 matrix job therefore keeps Next.js and Playwright on Node 20 while building and spawning the harper child process with Node 22 via the integration harness `HARPER_RUNTIME` override.

## Advisory, non-blocking

This deliberately tracks Next.js `main`, so a Next.js-side regression is visible on the next relevant harper PR. Because the result can depend on downstream health, this workflow must remain advisory and must not be configured as a required branch-protection check.

## Verification

- [Downstream run 29547153563](https://github.com/HarperFast/harper/actions/runs/29547153563): all Node 20/22/24 jobs passed, with 21 Playwright tests per job.
- Repository format, lint, unit, build, and caller-workflow validation checks passed on the final head.

Updated by Codex (OpenAI).

### DIFF
--- a/.github/workflows/integration-tests-nextjs.yml
+++ b/.github/workflows/integration-tests-nextjs.yml
@@ -5,14 +5,13 @@ name: Next.js Integration Tests (downstream)
 # integration (HTTP server, REST, ISR/`use cache` backed by Harper tables) is
 # caught here rather than in the downstream adapter repo.
 #
-# It calls the reusable workflow in HarperFast/nextjs, passing this PR's head SHA
-# as `harper_ref`; the reusable workflow checks out + builds harper at that ref and
-# resolves it as the `harper` the test harness and fixtures run against.
+# It checks out HarperFast/nextjs explicitly, then checks out + builds harper at this
+# PR's head SHA and resolves that tarball as the `harper` used by the adapter's test
+# harness and fixtures.
 #
-# Advisory, non-blocking: this is pinned to the nextjs reusable-workflow fix from
-# HarperFast/nextjs#50. That revision explicitly checks out the adapter repository;
-# without it, workflow_call inherits this caller's checkout context and runs the
-# adapter's package scripts against harper instead.
+# Advisory, non-blocking: this deliberately tracks nextjs `main`, so a regression on
+# nextjs main against harper surfaces here instead of silently drifting. That means
+# this check's outcome depends on nextjs main's health, not just this PR's diff.
 
 on:
   pull_request:
@@ -56,9 +55,74 @@ jobs:
   # Heavy path: only when relevant files changed. Runs and reports informationally --
   # this is not wired to a required gate job, so it never blocks a harper merge.
   nextjs-integration:
-    name: Next.js adapter integration (against this PR's harper)
+    name: Next.js adapter integration (against this PR's harper) (Node ${{ matrix.node-version }})
     needs: changes
     if: needs.changes.outputs.relevant == 'true'
-    uses: HarperFast/nextjs/.github/workflows/integration-tests.yml@8b1e950dcb1835b24a09bf4ffc0670c1d0cf9fc5 # HarperFast/nextjs#50
-    with:
-      harper_ref: ${{ github.event.pull_request.head.sha }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: [20, 22, 24]
+    steps:
+      - name: Checkout Next.js adapter
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          repository: HarperFast/nextjs
+          ref: main
+
+      - name: Set up Node.js
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Install adapter dependencies
+        run: npm ci
+
+      - name: Checkout harper override
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          repository: HarperFast/harper
+          ref: ${{ github.event.pull_request.head.sha }}
+          path: harper-override
+
+      - name: Build and install harper override
+        run: |
+          cd "$GITHUB_WORKSPACE/harper-override"
+          npm ci
+          npm run build
+          TARBALL=$(npm pack --json | jq -r '.[0].filename')
+          mv "$TARBALL" /tmp/harper-override.tgz
+          cd "$GITHUB_WORKSPACE"
+          npm install --save-dev /tmp/harper-override.tgz
+
+      - name: Build adapter
+        run: npm run build
+
+      - name: Install fixture dependencies
+        run: npm run install:fixtures
+
+      - name: Install Playwright browsers
+        run: npx playwright install --with-deps chromium
+
+      - name: Run integration tests
+        run: npm run test:integration
+        timeout-minutes: 15
+        env:
+          HARPER_INTEGRATION_TEST_LOG_DIR: /tmp/harper-test-logs
+          FORCE_COLOR: '1'
+
+      - name: Upload Harper logs on failure
+        if: failure()
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: harper-logs-node-${{ matrix.node-version }}
+          path: /tmp/harper-test-logs/
+          retention-days: 7
+
+      - name: Upload Playwright traces on failure
+        if: failure()
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: playwright-traces-node-${{ matrix.node-version }}
+          path: integrationTests/test-results/
+          retention-days: 7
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-tests-nextjs.yml
```diff
@@ -0,0 +1,39 @@
...
+      - 'dataLayer/**'
+      - 'components/**'
+      - 'config/**'
+      - 'package.json'
+      - 'package-lock.json'
+
+jobs:
+  nextjs-integration:
+    name: Next.js adapter integration (against this PR's harper)
+    uses: HarperFast/nextjs/.github/workflows/integration-tests.yml@main # TODO: pin to nextjs#49 merge SHA before merging
```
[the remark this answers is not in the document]
kriszyp: > main can silently change what CI code runs against harper PRs

Its not silent if it breaks! That's the whole point of CI! Catch failures because things change!

.github/workflows/integration-tests-nextjs.yml
```diff
@@ -0,0 +1,39 @@
...
+# ── BEFORE MERGE ───────────────────────────────────────────────────────────────
+#   1. Depends on HarperFast/nextjs#49 — the `workflow_call` + `harper_ref` input
+#      must be on nextjs `main` first, or this caller has nothing to call.
+#   2. Replace the mutable `@main` ref below with the 40-char merge-commit SHA of
+#      nextjs#49 (cross-repo reusable `uses:` should pin to a SHA, matching the
+#      convention in validate-caller-workflows.yml).
+
+on:
+  pull_request:
+    paths:
```
cb1kenobi: **Medium: a paths-filtered `pull_request` workflow can't be a satisfiable *required* gate.**

The PR frames this as a "downstream gate." If it's made a required status check on `main`, GitHub will leave it permanently `Expected — waiting for status to be reported` on every PR that doesn't touch these paths, because the workflow never starts and so never reports — blocking merge of unrelated PRs. GitHub has no "required only if it runs" semantic. This repo already documented this exact trap and worked around it in `validate-caller-workflows.yml` (fires unconditionally, no `paths:`). If this is intended to be required, either drop the `paths:` filter and gate the heavy job internally (e.g. a cheap change-detection step that skips the Playwright run when irrelevant, still reporting success), or keep it advisory/non-required. If it's advisory, ignore this.

—
_Generated by Barber AI_

.github/workflows/integration-tests-nextjs.yml
```diff
@@ -0,0 +1,39 @@
...
+      - 'cache/**'
```
cb1kenobi: **Low: `cache/**` matches nothing — there is no top-level `cache/` directory.**

Harper's caching code lives under `resources/` (`resources/tracked.ts`, `Table.ts`, `Resource.ts`) and `server/REST.ts`, both already covered by the filter below, so this is a dead, misleading entry rather than an actual coverage gap. The header comment specifically calls out `use cache`/ISR as a path the adapter exercises, so a future reader may assume this line gates it when it doesn't. Remove it (coverage is unchanged) or repoint it at the real location.

Suggested fix:
```suggestion
      - 'resources/**'
```

—
_Generated by Barber AI_

.github/workflows/integration-tests-nextjs.yml
cb1kenobi: **Low: nothing enforces the `@main` -> SHA pin before merge for this file.**

The before-merge TODO to pin `@main` to the nextjs#49 merge SHA is noted, but `validate-caller-workflows.yml` only scans `.github/workflows/claude-*.yml`, so this file escapes the automated mutable-ref check. Since the PR is not a draft (`isDraft: false`), the manual pin step could be missed on merge. Consider pinning now, or extending the validator's glob to cover this caller so the guard rail is automated rather than relying on memory.

—
_Generated by Barber AI_

.github/workflows/integration-tests-nextjs.yml
```diff
@@ -0,0 +1,39 @@
...
+      - 'resources/**'
+      - 'server/**'
+      - 'dataLayer/**'
+      - 'components/**'
+      - 'config/**'
+      - 'package.json'
+      - 'package-lock.json'
+
+jobs:
```
cb1kenobi, line 22: **Nit: no explicit `permissions:` block — inherits the repo/org default `GITHUB_TOKEN` scope.**

This test-only gate needs no write scope; several workflows in this repo already set explicit `permissions:`. Adding a top-level least-privilege block is cheap hardening for a new cross-repo reusable call. (Impact is limited: fork-PR tokens are already read-only, but same-repo branch PRs may inherit a broader default.)

Suggested fix:
```suggestion
permissions:
  contents: read

jobs:
```

—
_Generated by Barber AI_

.github/workflows/integration-tests-nextjs.yml
kriszyp: Addressed in 0066f1088: dropped the `cache/**` filter — there's no top-level `cache/` dir in this repo; caching lives under `resources/**`, which is already covered.

.github/workflows/integration-tests-nextjs.yml
kriszyp: Addressed in 0066f1088: nextjs#49 merged (d4b5ac9), so pinned the `uses:` ref to that merge SHA now rather than extending the validator's glob — the TODO already called for pinning once the dependency landed, and that's more direct than widening `validate-caller-workflows.yml` to a new glob.

.github/workflows/integration-tests-nextjs.yml
kriszyp: Addressed in 0066f1088: added an explicit `permissions: contents: read` block.

.github/workflows/integration-tests-nextjs.yml
kriszyp: Great catch, and you're exactly right about the "required only if it runs" gap. Fixed in 01ff2fbe8 by gating internally instead of at the trigger:

- Dropped the `paths:` filter from the `pull_request` trigger, so the workflow now runs on **every** PR and its check always reports.
- New `changes` job runs a plain `git diff --name-only <base>...HEAD` against the same effective path list the trigger used to filter on (`resources/ server/ dataLayer/ components/ config/ package.json package-lock.json`) and outputs `relevant=true/false`. Used `git diff` rather than a third-party paths-filter action to avoid pulling in a new dependency for the decision.
- The heavy `nextjs-integration` job (the reusable suite, unchanged) runs only `if: needs.changes.outputs.relevant == 'true'`.

One structural note worth flagging: because this job calls a **reusable workflow** (`uses:` at job level), it can't also host `steps:`, and a *skipped* reusable-workflow caller job contributes no check runs of its own — so guarding it alone would reintroduce the very "check never reports" problem. To give branch protection a single, stable, always-reporting status I added a `required` gate job (`if: always()`, `needs: [changes, nextjs-integration]`) that passes when the suite **passed or was skipped as irrelevant** and fails only when it actually failed/was cancelled. That job — "Next.js adapter integration required" — is the one to mark as the required check.

Net effect: irrelevant PRs get a cheap green gate, relevant PRs run the full suite, and the required check is always satisfiable. Thanks for pushing on this!

— KrAIs

.github/workflows/integration-tests-nextjs.yml
```diff
@@ -0,0 +1,35 @@
...
+# as `harper_ref`; the reusable workflow checks out + builds harper at that ref and
+# resolves it as the `harper` the test harness and fixtures run against.
+#
+# Scoped to PRs that touch the HTTP / resource / server code paths the adapter
+# exercises (tune the paths filter as coverage evolves; caching lives under
+# resources/**, there's no separate top-level cache/ dir).
+
+on:
+  pull_request:
+    paths:
```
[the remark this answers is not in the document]
kriszyp: Done in 01ff2fbe8 — went with the internal change-detection skip you suggested. Trigger `paths:` filter dropped so the workflow runs on every PR; a `changes` job does the cheap `git diff` path check, the reusable suite runs only when relevant, and an always-running `required` gate job provides the single stable status to mark required in branch protection (green when the suite passes or is skipped as irrelevant). Thanks!

.github/workflows/integration-tests-nextjs.yml
```diff
@@ -0,0 +1,130 @@
...
+            echo "No adapter-relevant paths changed -> integration suite will be skipped."
+          fi
+
+  # Heavy path: only when relevant files changed. Runs and reports informationally --
+  # this is not wired to a required gate job, so it never blocks a harper merge.
+  nextjs-integration:
+    name: Next.js adapter integration (against this PR's harper) (Node ${{ matrix.node-version }})
+    needs: changes
+    if: needs.changes.outputs.relevant == 'true'
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
kriszyp: Addressed in 2046ad182: added a 30-minute job-level timeout so checkout, install, build, and browser setup hangs are bounded as well as the test step.

heskew (approved):
CI workflow: actions SHA-pinned, `permissions: contents: read`, no secrets, no PR-controlled interpolation, job timeout set, and it ran green on this PR's own head. The advisory-only design (internal `changes` job instead of a required paths-filter) resolves the earlier trapped-check concern. LGTM.

PR #1387, opened 2026-06-18
feat(blob): add findIncompleteBlobRefs and isBlobComplete for repair sweep

kriszyp:
Adds three exports to `blob.ts` that the harper-pro blob repair sweep depends on:

- **`findIncompleteBlobRefs(database, databaseName?)`** — async generator that walks the primary store for records flagged `HAS_BLOBS` and yields those whose blob files are missing, truncated (`fileSize < HEADER_SIZE`), or errored (`ERROR_TYPE` header). Throttles via `STORAGE_BLOBCLEANUPSPEED` (same pacing as orphan cleanup).
- **`isBlobComplete(blob)`** — reads only the 8-byte header (stat + `openSync`/`readSync`/`closeSync`) to check `header.size === fileSize - HEADER_SIZE`. Safe to call on large blobs during a sweep.
- Re-exports existing `findBlobsInObject` and `isSaving` so `blobRepair.ts` has a single import surface.

These are cold-path / maintenance functions; no hot-path impact.

**Pairs with:** HarperFast/harper-pro (kris/blob-repair)

*Generated by Claude Opus 4.8*

### DIFF
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -15,9 +15,11 @@ import { addExtension, pack, Packr } from 'msgpackr';
 import { readFile, statfs, readdir, rmdir, unlink as unlinkPromised } from 'node:fs/promises';
 import {
 	close,
+	closeSync,
 	createWriteStream,
 	fdatasync,
 	open,
+	openSync,
 	readFileSync,
 	read,
 	readSync,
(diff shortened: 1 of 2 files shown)


kriszyp:
Rebased onto \`main\` (was 2 behind). The blob.test.js conflict from #1390 resolved cleanly — kept #1390's corrected \`complete()\` fault contract (needs-saving aborts on transient fault, track-only never aborts, plus the CONTRACT HISTORY comment); this branch's now-superseded version of that test dropped out during the rebase.

Added the requested coverage for the new exports per the review:
- \`isBlobComplete\`: happy path (file-backed, complete header), plus truncated (header size != file size), error-stub (ERROR_TYPE header), and ENOENT cases.
- \`findIncompleteBlobRefs\`: yields records with incomplete blobs, skips complete ones, and skips records without the \`HAS_BLOBS\` metadata flag.

blob.test.js is green (31 passing) and the full resources unit suite passes (843 passing). 

— Claude (Opus 4.8)

kriszyp:
## Fix: compressed blobs were falsely flagged incomplete (`isBlobFileComplete`)

Addresses the Codex review finding on this PR.

**Bug.** `isBlobFileComplete` decided completeness with `header.size === fileSize - HEADER_SIZE`. The 8-byte header's 6-byte size field holds the **uncompressed** content length for *both* storage types — `writeBlobWithStream` writes `deflate.bytesWritten`, which is the deflate stream's *input* (uncompressed) byte count, not the compressed output size (verified: a 12000-byte payload deflates to a 61-byte on-disk body but the header still records 12000). For a DEFLATE blob the on-disk body is the compressed stream, so the equality never holds for a correctly-saved compressed blob. The result: `findIncompleteBlobRefs` flagged **every** compressed blob as a false repair candidate. (The doc comment claiming "the compressed size in the file" was aspirational — the code stores the uncompressed size in both write paths.)

**Fix.** For `DEFLATE_TYPE` blobs, verify by streaming the body through `inflate` and comparing the decompressed length to the header size — a fully-written deflate stream inflates to exactly `size` bytes; a truncated one errors (`Z_BUF_ERROR`) or yields fewer. The uncompressed path keeps the existing body-length check, plus an explicit `UNKNOWN_SIZE` in-flight-placeholder check.

Streaming the inflate (rather than `inflateSync` over the whole buffer) keeps memory bounded during the repair sweep, which may touch many large blobs — this also resolves a follow-on concern both cross-model reviewers (Codex + Gemini) raised about whole-blob materialization. As a consequence `isBlobFileComplete` / `isBlobComplete` are now async (`Promise<boolean>`), and `findIncompleteBlobRefs` collects candidate blobs synchronously in the `findBlobsInObject` callback, then awaits the checks. These are newly-introduced exports in this PR, so there are no existing sync callsites to break.

**Tests.** Added a compressed (DEFLATE) case to `blob.test.js` proving a correctly-saved compressed blob reports complete (this assertion fails before the fix) and a truncated compressed blob still reports incomplete. `test:unit:resources` is green (844 passing); `test:unit:main` is unchanged from baseline.

*Generated by Claude Opus 4.8.*

resources/blob.ts
```diff
@@ -1609,3 +1616,117 @@ export async function cleanupOrphans(database: any, databaseName?: string) {
...
+	return size === fileSize - HEADER_SIZE;
+}
+
+/**
+ * Resolves true if the given blob's backing file is present and complete: no error stub, not an
+ * in-flight placeholder, and the content matches the header (uncompressed blobs by body length,
+ * compressed blobs by a streamed inflate). Used by the repair sweep to verify a blob was durably
+ * written after a peer fetch.
+ */
+export function isBlobComplete(blob: Blob): Promise<boolean> {
```
cb1kenobi: Not sure if this is a hot function, but the code would look cleaner if this function was `async`, then you wouldn't need the `Promise.resolve()`s.

cb1kenobi (approved):

resources/blob.ts
kriszyp: Good call — done in 85e3ed6. `isBlobComplete` is now `async`, so the two `Promise.resolve()` wrappers are gone and the early returns are plain `true`/`false`. It's a cold-path maintenance function (repair sweep only), so no hot-path concern with the async hop. Thanks for the review!

— Claude (Opus 4.8)

PR #1394, opened 2026-06-19
Add periodic re-read safety net for TLS cert watcher (#586)

kriszyp:
## Summary
The TLS cert-file watcher in `security/keys.ts` (`loadAndWatch`) relies solely on chokidar's inotify events. On filesystems where inotify is unreliable — overlayfs, many container setups, network mounts — a real certificate renewal can be **silently missed**, leaving Harper serving the old (expired) cert with no error. This is a live risk for the **July 8 2026 fleet cert rotation** (P0). A CI manifestation of exactly this — overlayfs dropping inotify events — was confirmed in #1392.

This PR hardens **detection** with a periodic re-read safety net. It does not touch worker-side propagation (the table subscription + TLS rebuild), which already works.

## What changed
- **Periodic re-read poll** added alongside the existing chokidar watcher. A per-file `setInterval` (`.unref()`'d so it never holds the process open) re-stats the watched file and routes through the **same** reload path on change. Chokidar stays the fast path; the poll is the belt-and-suspenders net for missed events.
- **De-dupe:** both detection paths share the existing `lastModified` mtime fingerprint, so an inotify event and a poll for the same change never double-reload. An unchanged file does only a cheap `statSync` (no PEM read, no reload).
- **Configurable interval:** top-level `tls.certificateWatchInterval` (ms), default **300000 (5 min)**; `0` disables polling. Registered in `hdbTerms.ts`, `static/defaultConfig.yaml`, the Joi validator (`validation/configValidator.ts`), and `config-root.schema.json` (which has `additionalProperties: false` on `tlsConfig`).

## Where to look
- `security/keys.ts` `loadAndWatch` — poll registration and thread gating.
- **Thread gating (please sanity-check):** the **certificate** poll runs on the **main thread only** (workers get cert updates via the `hdb_certificate` subscription). The **private-key** poll runs on **all threads**, because each worker loads its own key directly from disk with no table propagation — a key renewal missed by inotify on a worker would otherwise strand the stale key. This asymmetry was raised by cross-model review and is deliberate.

## Open item for reviewer (deliberately out of scope)
Cross-model review flagged a latent race: when a renewal rotates **both** cert and key, a worker's private-key poll only refreshes the in-thread `privateKeys` map — it does **not** itself trigger a `createTLSSelector`/`updateTLS` rebuild. The TLS rebuild is driven by the `hdb_certificate` subscription. **This is pre-existing behavior of the private-key chokidar watcher too** (its `loadCert` callback already just sets the map), not introduced here — this PR only adds a second trigger to the same path. Fixing worker key-rotation → immediate TLS rebuild is a separate, riskier change I deliberately kept out of this P0 patch. Flagging so it's not rediscovered.

## Testing
- New unit coverage in `unitTests/security/keys.test.js`: swap missed by inotify is picked up by the poll (chokidar stubbed so only the poll path runs); mtime-fingerprint dedup; configured-interval resolution + `.unref()`; default/invalid-interval fallback; `0` disables; private-key path stays wired.
- Existing `integrationTests/security/cert-reload.test.ts` (#586 multi-worker propagation regression) still passes.
- `test:unit:main` + `test:unit:config` green (10 pre-existing `globalIsolation.test.js` failures are unrelated — identical on clean `main`). Lint/format clean; TypeStrip `--check` clean.

## Notes
- **P0 / July 8 cert-rotation relevant**; likely a **5.1.6 patch candidate**.
- Generated by an LLM (Claude Opus 4.8).

### DIFF
--- a/config-root.schema.json
+++ b/config-root.schema.json
@@ -682,7 +682,12 @@
 					"description": "Path to private key file. Default: <ROOTPATH>/keys/privateKey.pem"
 				},
 				"ciphers": { "type": "string", "description": "Optional explicit cipher list." },
-				"host": { "type": "string", "description": "Optional SNI host this certificate applies to." }
+				"host": { "type": "string", "description": "Optional SNI host this certificate applies to." },
+				"certificateWatchInterval": {
+					"type": "number",
+					"minimum": 0,
+					"description": "Interval in ms for the periodic re-read safety net that re-checks watched cert/key files in case an inotify change event was missed (e.g. on overlayfs/containers/network mounts). Honored on the top-level tls block. Default: 300000 (5 min). 0 disables polling, leaving only the inotify watcher."
+				}
 			}
 		}
 	},
--- a/security/keys.ts
+++ b/security/keys.ts
@@ -40,6 +40,9 @@ const CERT_VALIDITY_DAYS = 3650;
 // watcher is the fast path; this poll catches changes on filesystems where inotify is unreliable
 // (overlayfs, many container setups, network mounts). Overridable via tls.certificateWatchInterval; 0 disables.
 const DEFAULT_CERTIFICATE_WATCH_INTERVAL_MS = 300_000;
+// Lower bound (ms) for a configured poll interval; a misconfigured sub-second value is clamped up
+// to keep the safety-net poll from becoming a tight stat() loop. 0 still disables polling entirely.
+const MIN_CERTIFICATE_WATCH_INTERVAL_MS = 1000;
 const CERT_DOMAINS = ['127.0.0.1', 'localhost', '::1'];
 export const CERT_ATTRIBUTES = [
 	{ name: 'countryName', value: 'USA' },
@@ -304,7 +307,10 @@ function getCertificateWatchInterval(): number {
 	if (configured == null) return DEFAULT_CERTIFICATE_WATCH_INTERVAL_MS;
 	const interval = Number(configured);
 	if (!Number.isFinite(interval) || interval < 0) return DEFAULT_CERTIFICATE_WATCH_INTERVAL_MS;
-	return interval;
+	// 0 explicitly disables the poll; otherwise floor at MIN to keep a typo (e.g. 1ms) from
+	// spinning a stat() loop. This is a safety net, not a hot-poll path, so sub-second is never wanted.
+	if (interval === 0) return 0;
+	return Math.max(interval, MIN_CERTIFICATE_WATCH_INTERVAL_MS);
 }
 
 // Active poll timers, keyed by watched path (exposed for test cleanup).
--- a/security/keys.ts
+++ b/security/keys.ts
@@ -330,7 +330,10 @@ function loadAndWatch(path, loadCert, type) {
 	let lastModified;
 	const loadFile = (path, stats) => {
 		try {
-			let modified = stats.mtimeMs;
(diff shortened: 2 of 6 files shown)


kriszyp:
**Holding this — blocked on the `security/keys.ts` cert-reload fixes (do not merge yet).**

This adds a production periodic re-read safety net so a missed inotify event can't strand a renewed cert (overlayfs / network-mount / container filesystems) — relevant to the July 8 cert rotation. The code is complete, but it should land **on top of** the two `keys.ts` cert-reload fixes currently in progress:

1. the cert **staleness-guard field-name mismatch** in `loadCertificates()` — which currently rejects a detected reload, and
2. the **worker-thread private-key hot-reload race**.

Until those land, the cert hot-reload integration test stays red regardless of detection, so verifying this PR end-to-end isn't meaningful yet. Once they merge I'll **rebase this on top and re-verify the cert test end-to-end**. This PR also supersedes the test-only polling override in #1392 (production polling makes that redundant).

_— Claude (Opus 4.8), via the PR-triage session_

kriszyp:
Rebased onto current `main` (v5.1.6), reconciling this poll safety net with the cert-reload work that just shipped.

**Rebase / conflict resolution**
- `security/keys.ts` auto-merged cleanly: kept the shipped `loadAndWatch` structure (the `lastModified` mtime-fingerprint dedup, the `isMainThread` reload-warn line) and re-applied the periodic poll on top. The shipped `file_timestamp` staleness guard (#1408), worker secure-context rebuild on key hot-reload, and rebuild-on-first-key behavior are untouched.
- Only conflict was `unitTests/security/keys.test.js` — purely co-located `describe` blocks (the shipped staleness-guard / hot-reload suites vs. this PR's poll suite). Kept both.

**Poll is now the production companion to the inotify watcher**
- The poll `statSync`s the cert file each tick and delegates into the shipped `loadFile(path, stats)`, so it shares the one `lastModified` dedup — inotify and the poll never double-reload or thrash. No parallel reload path.
- Main-thread-only for certs (workers get certs via the `hdb_certificate` subscription); runs on all threads for private keys (each worker loads its own key from disk). Configurable via `tls.certificateWatchInterval` (default 5m, `0` disables), timer `.unref()`'d. This is the safety net for the July-8 overlayfs/container/network-mount case where a real renewal's inotify event is silently missed.

**Two small hardening commits from cross-model review (Codex clean; Gemini flagged):**
- Reload on chokidar `change` events that omit `stats`. chokidar v4 defaults `alwaysStat:false`, so `loadFile` was reading `stats.mtimeMs` off `undefined`, throwing into the catch and silently skipping the inotify fast-path reload (leaving only the poll). Now falls back to `statSync(path)` when stats is absent.
- Clamp a sub-second configured interval to a 1s floor so a typo (e.g. `1` meaning seconds) can't spin a tight `stat()` loop; `0` still disables.

**Tests:** `unitTests/security/keys.test.js` (30 passing, incl. a test proving the poll picks up a cert swap when the inotify event is bypassed, plus the new stats-less-change and clamp tests) and both cert hot-reload integration tests (`cert-reload`, `cert-key-reload`) green locally.

I deliberately kept the per-tick stat **synchronous** (Gemini suggested async `fs.stat`): it's a 5-minute, cold-path safety net consistent with the existing sync stats in `loadAndWatch`, so async would add complexity for no real benefit. Flagging for the reviewer to weigh in if you'd prefer otherwise.

— generated by Claude Opus 4.8

PR #1403, opened 2026-06-19
Avoid needless secondary-index rebuilds from equivalent index-option representations; log rebuild reason

### LINKED ISSUE #1357: Avoid unnecessary index rebuilds from semantically-equivalent index-option representations
Part of #1354

## Problem
The "structurally changed?" decision for a secondary index compares index options with `JSON.stringify`, which is sensitive to (1) option **key order** and (2) **string-vs-number** scalars. Because `@indexed(...)` options are recorded in source-argument order and as strings, while other table-creation paths (operations API / config objects) can supply them reordered or as numbers, a representation-only difference flips the decision and forces a full rebuild (clearing + rebuilding the index, 503-ing the attribute throughout) even though the index is semantically identical.

Separately, core never logs *why* a rebuild was triggered, which makes unexpected rebuilds hard to diagnose.

## Evidence
- Structural comparison: `resources/databases.ts:1285-1288` and `:1321-1323`
- Options built in source order, values stored as strings: `resources/graphql.ts:168-174`
- Sole start-of-rebuild log (states no reason): `resources/databases.ts:1411`; trigger conditions `:1291-1296` / `:1300-1305`

## Proposed change

**(1) Avoid spurious rebuilds.** Introduce a narrow canonicalizer used **only** for the structural comparison: sort option keys and coerce numeric-looking string scalars to numbers; keep the raw form for persistence, and keep the *persistence* trigger on the raw form so the descriptor self-heals to canonical over time. **Conservative by design** — do not coalesce absent-vs-present options with defaults; leave `type`, boolean-vs-object, and all `commonChanged` fields (type/nullable/version/properties/elements/embed) as hard rebuild triggers.

**(2) Log the rebuild reason.** Emit one explicit line per attribute stating which trigger fired — e.g. `reindex <db>.<table>.<attr>: reason=structural-options-changed | indexing-failed-retry | crash-recovery(pid) | restart-number`. Additive, no behavior change. (This is the observability piece that makes "why did it reindex?" answerable.)

## Compatibility / risk
Correctness-sensitive — must **never** skip a *needed* rebuild. Ship with a test matrix:
- `{type:"HNSW", M:"16"}` vs `{M:16, type:"HNSW"}` → **not** structurally changed
- `{type:"HNSW", M:16}` vs `{type:"HNSW", M:32}` → structurally changed
- `true` vs `{type:"HNSW"}` → structurally changed

## Acceptance
- A reload with semantically-identical index options (reordered / string-vs-number) does **not** trigger a rebuild; a genuine option change still does.
- Each rebuild logs its trigger reason.

---

heskew:
## Summary

The secondary-index "structurally changed?" decision in `resources/databases.ts` compared index options with a raw `JSON.stringify`, which is sensitive to option **key order** and **string-vs-number** scalars. `@indexed(...)` records options in source order as strings while the operations API / config objects can supply them reordered or as numbers, so a representation-only difference flipped the decision and forced a **full rebuild** — clearing + rebuilding the index and `503`-ing the attribute for the whole backfill — for a semantically identical index.

This adds `canonicalizeIndexOptions()`, used **only** for the structural comparison: it sorts option keys and coerces numeric-looking string scalars to numbers. It's applied symmetrically to **both** structural comparisons — the reindex trigger (`structurallyChanged`) and the checkpoint-reset decision (`indexOptionsChanged`). Persistence still keys off the **raw** form, so a stored descriptor self-heals toward the canonical shape over time.

Second piece (explainability): one `info` line per reindexed attribute stating which trigger fired — `reindex <db>.<table>.<attr>: reason=new-index|attribute-definition-changed|structural-options-changed|indexing-failed-retry|crash-recovery(pid=N)|restart-number`.

Closes #1357 (sub-issue of #1354).

## Where to look

- **`canonicalizeIndexOptions` zero exclusion** (`databases.ts` ~:1463-1473) — the one subtle correctness edge. `0` is the only finite number whose string form (`"0"`, truthy) and numeric form (`0`, falsy) diverge in JS truthiness, and index code branches on truthiness (e.g. HNSW `if (this.optimizeRouting)` doubles `maxConnections`). So `"0"` and `0` build structurally different indexes and must **not** be collapsed. Zero is excluded from coercion; everything non-zero coerces. This was caught by the cross-model review (see below).
- **Conservative by design** — boolean-vs-object, absent-vs-present, and the zero case are all preserved as rebuild triggers; `commonChanged` (type/nullable/version/properties/elements/embed) is OR'd in unchanged. The invariant is *never skip a needed rebuild*; canonicalization only relaxes the equality test for genuinely-equivalent representations.

## Two deliberate, non-blocking choices (flagged so you don't have to rediscover them)

- **The reason-log mirrors the trigger gate's decision basis**, which mixes the pre-lock structural comparison with the under-lock re-fetched descriptor's recovery flags (PID/restartNumber/indexingFailed). Under a concurrent reload the reasons list could be momentarily inconsistent. It's diagnostic-only (nothing reads `reindexReasons`; `|| 'unknown'` covers the empty case), so I left it mirroring the gate rather than recomputing post-lock.
- **`info` level** for the reason line — consistent with the existing `Indexing/Finished indexing` `info` logs and gated behind `hasExistingData` + the rebuild decision, so it's one line per actual backfill, not hot-path chatter.

## Tests

`unitTests/resources/indexCanonicalOptions.test.js`: the issue's exact matrix + conservative edges (incl. the zero case), plus an HNSW wiring test proving a reordered / string-vs-number option does **not** re-trigger a backfill while a genuine `M:16→32` change does. The sibling `indexRestartNumber` tests were re-verified to confirm the reindex trigger is unaffected.

## Cross-model review

Ran per engineering guidelines (storage/schema → thorough). **Codex** found the zero-valued-`optimizeRouting` collapse described above — now fixed and covered by a test. The **Harper-domain adjudication** returned **no blockers / no significant concerns** and verified the zero-fix is complete and the symmetry/invariant hold. **Caveat:** the Gemini leg (`agy`) hung in this environment and produced no output, so there's no outside-model perf/maintainability pass — low residual risk here since the new code is cold-path only (schema load / reindex), not on the read/write hot path.

No docs change: no public API, config option, schema directive, or CLI surface changes (the log line is operator-facing but not a documented interface).

---

### DIFF
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -1268,6 +1268,10 @@ export function table<TableResourceType>(tableDefinition: TableDefinition): Tabl
 				for (const key of searchOnlyOptions) delete copy[key];
 				return copy;
 			};
+			// Canonical key for the structural (reindex-triggering) comparison only: strip search-only
+			// options, then sort keys and coerce numeric-looking string scalars so a representation-only
+			// difference (key order, string-vs-number) does not force a needless rebuild. harper#1357
+			const canonicalIndexKey = (indexed: any) => JSON.stringify(canonicalizeIndexOptions(stripSearchOnly(indexed)));
 			const commonChanged =
 				!attributeDescriptor ||
 				attributeDescriptor.type !== attribute.type ||
@@ -1281,11 +1285,11 @@ export function table<TableResourceType>(tableDefinition: TableDefinition): Tabl
 			// any metadata difference (drives persistence)
 			const changed =
 				commonChanged || JSON.stringify(attributeDescriptor?.indexed) !== JSON.stringify(attribute.indexed);
-			// structure-affecting difference (drives reindex) — ignores search-only option changes
-			const structurallyChanged =
-				commonChanged ||
-				JSON.stringify(stripSearchOnly(attributeDescriptor?.indexed)) !==
-					JSON.stringify(stripSearchOnly(attribute.indexed));
+			// structure-affecting difference (drives reindex) — ignores search-only option changes and
+			// representation-only differences (key order, string-vs-number) via canonicalIndexKey
+			const indexOptionsStructurallyChanged =
+				canonicalIndexKey(attributeDescriptor?.indexed) !== canonicalIndexKey(attribute.indexed);
+			const structurallyChanged = commonChanged || indexOptionsStructurallyChanged;
 			if (attribute.indexed) {
 				// The restart generation that owns any in-progress build of this index. Use the
 				// worker's stable startup generation (workerData.restartNumber), NOT the mutable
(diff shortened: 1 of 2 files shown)


kriszyp (approved):
Approving — clean perf-correctness win exactly where it counts. Canonicalization only relaxes index-*options* equality; the structural axes (type/nullable/version/properties/elements/embed) stay raw-compared, so no schema/type/collation/sparse case yields a wrongly-skipped rebuild, and the HNSW zero-exclusion is genuinely necessary. Tests assert both skip-when-safe and trigger-when-needed. Nice work.

— Reviewed by Claude (claude-opus-4-8)

PR #1405, opened 2026-06-19
feat(models): public registerBackend + defineBackend for custom backends (#1325)

### LINKED ISSUE #1325: [Models] Public backend registration API — registerBackend + defineBackend (custom/in-process backends)
## Context

`scope.models` (#510) shipped ungated in 5.1.0 GA. Today every config-selectable backend — `ollama`, `openai`, `anthropic`, `bedrock` — speaks over an **HTTP(S) transport** (native `fetch` for ollama/openai/anthropic; the `@aws-sdk/client-bedrock-runtime` SDK, also HTTPS, for bedrock), reached through a **closed factory map** in `resources/models/bootstrap.ts`. There is no public path for a component or app to register its own backend, and in particular **no way to register an in-process ("code") backend** — one whose `embed()` / `generate()` runs inference directly (native addon, wasm, or plain JS) instead of crossing an HTTP boundary.

A code backend isn't just a convenience for a particular app — it's the right choice **wherever an in-process API is preferable to an HTTP hop**: eliminating loopback serialization and network surface for a local model, coupling tightly to in-process model state, or running fully offline. The HTTP path is already well served (OpenAI-compatible inference servers work today through the openai backend's `baseUrl`); this issue covers the other end of the spectrum.

The backend abstraction itself is already transport-agnostic: `ModelBackend` (`resources/models/types.ts`) is just `embed? / generate? / generateStream? / capabilities()`, and the registry (`resources/models/backendRegistry.ts`) maps `logicalName → backend` with no assumptions about transport. The only things blocking a custom backend are:

1. **`bootstrap.ts` `FACTORIES`** — a hardcoded `{ollama, openai, anthropic, bedrock}` map. A `backend:` value outside it is logged and skipped.
2. **`setEmbedding` / `setGenerative`** — the actual registration primitives — are only imported by the four built-in components from an internal module. There is **no `models.registerBackend(...)`**, nothing on `scope.models`, nothing exported from the `harperdb` package.

(The config validator was already left open for this — `unknownBackendEntrySchema` in `validation/configValidator.ts` deliberately permits unknown `backend:` names "so future versions or third-party components can register their own backends without core schema edits.")

### Concrete motivating case

`harper-fabric-embeddings` — a thin wrapper around `node-llama-cpp` that runs a quantized GGUF model (`bge-small-en-v1.5`, ~80MB) locally and in-process, embed-only — is one such backend. It can't be expressed as a `scope.models` backend today, so apps using it can't migrate onto the unified API. It's also the canonical "native-dep, ship-as-separate-package" case: the dependency lives in the backend package, never in core.

### A transport spectrum, not HTTP-vs-code

Registration shouldn't assume a transport. The useful range is **in-process code API → local socket (UDS) → loopback HTTP → remote HTTP**. A Unix-socket backend (no TCP stack, lower latency than loopback HTTP) is a plausible middle point and is **not supported today** — the backends parse `host` / `baseUrl` as HTTP origins and expose only a test-only `fetchImpl` constructor seam, with no socket / undici-dispatcher config. The defining property of this issue: a public `ModelBackend` registration hook makes *any* point on that spectrum shippable without core changes — a socket or code backend is just a `ModelBackend` whose implementation targets a socket or runs inference inline. (Adding optional UDS config to the existing fetch-based backends via an undici dispatcher is a small, separate enhancement.)

## Proposed API (additive)

```ts
// new exports from 'harperdb'; also reachable as scope.models.registerBackend(...)
function registerBackend(kind: 'embedding' | 'generative', id: string, backend: ModelBackend): void;

// shim: derives capabilities() from which methods you supply, so the common
// (embed-only) case doesn't require authoring a full class.
function defineBackend(spec: {
  name: string;
  embed?(input, opts): Promise<EmbedResult>;
  generate?(input, opts): Promise<GenerateResult>;
  generateStream?(input, opts): AsyncIterable<GenerateChunk>;
}): ModelBackend;
```

`id` uses a provider-namespaced convention (`local:bge-small`, `openai:gpt-4o`) to avoid collisions once multiple plugins register. Namespacing is convention, not enforced.

### Example — migrating `harper-fabric-embeddings`

```ts
import { init, embed } from 'harper-fabric-embeddings';
import { registerBackend, defineBackend } from 'harperdb';

await init({ threads: 4 });
registerBackend('embedding', 'local:bge-small', defineBackend({
  name: 'local:bge-small',
  async embed(texts) {
    const vecs = await embed([texts].flat());
    return { status: 'completed', output: vecs.map(Float32Array.from) };
  },
})); // capabilities() ⇒ { embed: true, generate: false, ... } automatically
```

Then `models.embed(text, { model: 'local:bge-small' })`, or alias it to `embedding.default`.

### Config-driven selection → split to #1471

Operator selection of a custom backend via the `models:` YAML block (`backend: <module>`, the way the built-ins are selected) is tracked separately in **#1471**. This issue is **imperative-only**. The `registerBackend` / `defineBackend` here is the foundation that path builds on, and nothing in it forecloses the config path — the validator already admits unknown `backend:` values and `bootstrapModels` is the single extension point.

## Backward compatibility

All net-new symbols — no change to existing call sites, the `ModelBackend` interface, the four built-in backends, or the `models:` config shape. `scope.models` is frozen as of 5.1.0; this is additive only.

## Acceptance

- [ ] `registerBackend(kind, id, backend)` exported from `harperdb` and available as `scope.models.registerBackend`.
- [ ] `defineBackend(spec)` derives `capabilities()` from the methods present.
- [ ] A component can register an in-process backend during `handleApplication` and have `models.embed/generate({ model: id })` resolve to it.
- [ ] An end-to-end test proves a registered in-process backend resolves through `models.embed({ model })` (not just `TestBackend`).
- [ ] No change to existing public surface; existing backends/config unaffected.

## Out of scope

- Operator config-file selection of a custom backend (`backend: <module>`) — #1471.
- Selecting among multiple backends / smart routing / fallback — #1326.
- Shipping a local-inference backend in core (it ships as a separate opt-in package; core stays free of native/wasm deps).

Part of #510.

---

heskew:
## Summary

Exposes the model-backend registry so components and apps can register their own backends — including in-process / non-HTTP ("code") ones — under a logical id, the same primitive the built-in backends use internally.

- `registerBackend(kind, id, backend)` — public registration; also `scope.models.registerBackend(...)`.
- `defineBackend(spec)` — builds a `ModelBackend`, deriving `capabilities()` from the methods supplied (`tools` / `adapters` declared explicitly, default false).
- Both reach components through every path `models` already uses: the package index, globals, **and** the component module loader (`getHarperExports`), so `import { registerBackend } from 'harperdb'` works inside a component.
- Model public types re-exported from the package; `registerBackend` added to the `Models` interface.

## Purpose

`scope.models` (5.1.0 GA) only hosts the four config-selectable backends via a closed factory map; there was no public path to register a custom or in-process backend. This unblocks that — e.g. wrapping a local node-llama-cpp embedder behind `scope.models`.

Closes #1325 (imperative-only scope; operator config-file selection is tracked separately in #1471). Part of #510.

## Where to look

- `security/jsLoader.ts` `getHarperExports` — the surface components import from `'harperdb'`; the two new functions are wired alongside `models`. A regression test in `jsLoader.test.js` asserts they resolve through `require('harperdb')`.
- `resources/models/backendRegistry.ts` — `registerBackend` validation + `defineBackend` capability derivation.
- Backward-compat: strictly additive — no change to the existing `Models` runtime surface, `ModelBackend`, the built-in backends, or `models:` config. `scope.models` stays GA-frozen (5.1.0).

## Scope / deferred

Imperative API only — register backends in code, select via `opts.model`. Operator config-file selection of a custom backend (`backend: <module>` in the `models:` YAML) is split out to #1471.

## Docs

Companion docs PR: HarperFast/documentation#554.

## Review

Cross-model reviewed (Codex + Harper domain pass). No blockers. Open coverage caveat for the human reviewer: the Gemini / perf-lens leg timed out and did not run — low cost here, since this is registration code that runs at component load, not in the embed/generate hot path.

## Status

Ready for review — scope settled (imperative-only; config split to #1471), CI green. The companion docs PR (HarperFast/documentation#554) is intentionally held in draft until the release version this lands in is known (its `<VersionBadge>` needs the real version).

---

### DIFF
--- a/globals.js
+++ b/globals.js
@@ -2,9 +2,11 @@
 global.contentTypes = exports.contentTypes = null;
 global.createBlob = exports.createBlob = undefined;
 global.databases = exports.databases = {};
+global.defineBackend = exports.defineBackend = undefined;
 global.logger = exports.logger = {};
 global.models = exports.models = undefined;
 global.operation = exports.operation = undefined;
+global.registerBackend = exports.registerBackend = undefined;
 global.Resource = exports.Resource = undefined;
 global.server = exports.server = {};
 global.tables = exports.tables = {};
--- a/index.ts
+++ b/index.ts
@@ -29,6 +29,28 @@ export type { IterableEventQueue } from './resources/IterableEventQueue.ts';
 export type { Table } from './resources/databases.ts';
 export type { Attribute } from './resources/Table.ts';
 export type { Scope } from './components/Scope.ts';
+export type {
+	ModelBackend,
+	ModelCapabilities,
+	DefineBackendSpec,
+	EmbedOpts,
+	GenerateOpts,
+	GenerateInput,
+	GenerateResult,
+	GenerateChunk,
+	BackendOpts,
+	AccountingContext,
+	ModelCallResult,
+	TokenUsage,
+	Message,
+	ToolDef,
+	ToolCall,
+	ToolHandler,
+	ToolHandlerContext,
+	ToolTraceEntry,
+	ConversationAppender,
+	ConversationTurn,
+} from './resources/models/types.ts';
 export type { FilesOption, FilesOptionObject } from './components/deriveGlobOptions.ts';
 export type { FileAndURLPathConfig } from './components/Component.ts';
 export type { OptionsWatcher, Config, ConfigValue } from './components/OptionsWatcher.ts';
@@ -69,14 +91,20 @@ import type { server as ServerImport } from './server/Server.ts';
 import type { tables as TablesImport } from './resources/databases.ts';
 type ThreadsImport = unknown[]; // TODO: figure out actual type for this
 import type { transaction as TransactionImport } from './resources/transaction.ts';
+import type {
+	registerBackend as RegisterBackendImport,
+	defineBackend as DefineBackendImport,
+} from './resources/models/backendRegistry.ts';
 
 declare global {
 	const contentTypes: typeof ContentTypesImport;
 	const createBlob: typeof CreateBlobImport;
 	const databases: typeof DatabasesImport;
+	const defineBackend: typeof DefineBackendImport;
 	const logger: Logger;
 	const models: typeof ModelsImport;
 	const operation: typeof OperationImport;
+	const registerBackend: typeof RegisterBackendImport;
 	const Resource: typeof ResourceImport;
 	const server: typeof ServerImport;
 	const tables: typeof TablesImport;
@@ -88,9 +116,11 @@ declare global {
 export declare const contentTypes: typeof ContentTypesImport;
 export declare const createBlob: typeof CreateBlobImport;
 export declare const databases: typeof DatabasesImport;
+export declare const defineBackend: typeof DefineBackendImport;
 export declare const logger: Logger;
 export declare const models: typeof ModelsImport;
 export declare const operation: typeof OperationImport;
+export declare const registerBackend: typeof RegisterBackendImport;
 export declare const Resource: typeof ResourceImport;
 export declare const server: typeof ServerImport;
 export declare const tables: typeof TablesImport;
@@ -101,9 +131,11 @@ export declare const transaction: typeof TransactionImport;
 exports.contentTypes = null;
 exports.createBlob = undefined;
 exports.databases = {};
+exports.defineBackend = undefined;
 exports.logger = {};
 exports.models = undefined;
 exports.operation = undefined;
+exports.registerBackend = undefined;
 exports.Resource = undefined;
 exports.server = {};
 exports.tables = {};
--- a/resources/models/Models.ts
+++ b/resources/models/Models.ts
@@ -1,6 +1,11 @@
 import { _assignPackageExport } from '../../globals.js';
 import { contextStorage } from '../transaction.ts';
-import { resolveEmbedding, resolveGenerative } from './backendRegistry.ts';
+import {
+	defineBackend,
+	registerBackend as registerBackendImpl,
+	resolveEmbedding,
+	resolveGenerative,
+} from './backendRegistry.ts';
 import { getModelCallAnalyticsWriter, type ModelCallAnalyticsWriter, type ModelCallRecord } from './analyticsTable.ts';
 import { recordAction } from '../analytics/write.ts';
 import { ServerError } from '../../utility/errors/hdbError.ts';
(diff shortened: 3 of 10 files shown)


kriszyp (approved):
Nice addition — `registerBackend`/`defineBackend` is a clean, additive public surface for custom model backends, the validation in `registerBackend` is solid, and TypeStrip mechanics are tidy. The failing Bun job is an unrelated `csv_file_load` timing flake (green on every Node version), not this change. Approving.

Two API-hardening suggestions worth folding in before this becomes a stable extension point (both are "make it hard to misuse" rather than blockers):

1. **`defineBackend` assigns on truthiness, not function-ness** — [`backendRegistry.ts:124`](https://github.com/HarperFast/harper/pull/1405/files#diff-df547846e7873a7fbdffada6f44465a572c3d85b30c02f66526b359369cc1900R124). A truthy non-function `embed`/`generate`/`generateStream` (e.g. `generate: 'oops'`) gets assigned onto the backend and slips past registration, then crashes at call time with `TypeError: backend.generate is not a function`. Since the capability flags are already derived with `typeof x === 'function'`, gate the assignment on the same check (or assign from the derived map) so a bad value is rejected at definition time.

2. **Stream-only generative backend crashes on `generate()`** — [`backendRegistry.ts:75`](https://github.com/HarperFast/harper/pull/1405/files#diff-df547846e7873a7fbdffada6f44465a572c3d85b30c02f66526b359369cc1900R75). If a backend provides `generateStream` but not `generate`, registration succeeds (the generative check passes on `generateStream`), but a plain `models.generate()` then throws. Either synthesize a `generate` that drains the stream, or have `generate()` fall back to `generateStream` when only the latter exists.

Minor: `registerBackend` doesn't cross-check that a hand-rolled backend's `capabilities()` actually agrees with its methods (the doc-comment says it "must"). `defineBackend` guarantees this; the hand-rolled path relies on the contract. A cheap cross-check would harden it, but fine to leave to the doc-comment.

— Claude (Opus 4.8)

heskew:
Addressed both hardening suggestions in 7c7e7ca21:

1. **`defineBackend` truthiness → function-ness.** Method assignment is now gated on `typeof x === 'function'` (matching the capability derivation), so a non-function value (`generate: 'oops'`) is rejected at definition time — or simply not attached when another valid method is present — rather than slipping through to a call-time `TypeError`.
2. **Stream-only generative backend.** `defineBackend` now synthesizes `generate()` by draining `generateStream`, so a plain `models.generate()` works and `capabilities().generate` is `true`. Tool-call deltas are forwarded best-effort — the chunk type carries no call index to reassemble fragments, so a backend that streams partial tool calls should implement `generate()` directly (noted in the code).

The minor (hand-rolled `capabilities()` cross-check) is left to the doc-comment on `registerBackend`, as you suggested. And yes — the Bun `csv_file_load` failure is the unrelated flake.

Added unit coverage for both. Thanks for the review!

---

PR #1406, opened 2026-06-19
fix(audit): report entries_deleted for delete_transaction/audit_logs_before (#1294)

### LINKED ISSUE #1294: RocksDB audit retention is whole-file granular; delete_audit_logs_before reports success while purging 0
## Summary
On the default RocksDB engine the audit log **is** the transaction-log files. Both the auto-retention loop (`logging.auditRetention`) and the explicit `delete_audit_logs_before` op call `rootStore.purgeLogs({ before })` (`auditStore.ts:166`, `ResourceBridge.ts:495-503`), which deletes **whole log files** older than the cutoff — file granularity, not per-entry. (The per-entry walk `Table.deleteHistory` → `removeAuditEntry` is the LMDB-only branch; `RocksTransactionLogStore.remove()` is a literal no-op with a `TODO`.)

Observed (audited table, 2000 inserts + 2000 updates, `auditRetention: 3s`):
- auditCount 1 → 4001 (+1.2 MiB)
- after the auto-retention window: no reduction
- after `delete_audit_logs_before`: **no reduction** — the op ran as a job and logged "Finished deleting audit logs…", succeeding with zero effect; `read_audit_log` never reflects it.

## Consequences
1. Aged entries sitting in the still-active log file can't be purged, so a short retention window reclaims nothing until a file rolls over and fully ages past the cutoff — retention is approximate and lags the configured window.
2. `delete_audit_logs_before` is **misleading**: it reports a successful job while purging 0 files.
3. Ties #1288-adjacent blob GC: `cleanup_orphan_blobs` treats a blob as referenced while the audit log points to its prior value, so crash-orphaned blob reclamation only unblocks once the referencing entry's whole log file ages out (same coarse lag).

## Suggestion
Surface the purged-file count (vs a silent 0) in the op result; document that RocksDB audit retention is whole-file/rollover-granular (not per-entry).

## Repro
`integrationTests/qa-scratch/audit-retention.test.ts`

---
_Found via the exploratory QA campaign (qa-explorer), scenario QA-047. Harper `001bf7b9c` (v5.1.0, main)._

cb1kenobi:
Closes #1294.

## Summary
On the RocksDB engine, `delete_transaction_logs_before` / `delete_audit_logs_before` purge whole transaction-log files but reported no entry count, so the ops looked like silent no-ops (issue #1294). `@harperfast/rocksdb-js` 2.2.0 adds `purgeLogs({ includeEntryCounts: true })`, which returns `{ path, entries }[]`. This surfaces the purged entry count as a new `entries_deleted` response field.

- `DeleteTransactionLogsBeforeResults`: add `entries_deleted`.
- `ResourceBridge.deleteTransactionLogsBefore`: pass `includeEntryCounts: true` and sum entries for RocksDB; for LMDB, sum the count now returned by `Table.deleteHistory`.
- `Table.deleteHistory`: return the number of audit entries removed (was `void`).
- `delete.ts` (deprecated `delete_audit_logs_before`): previously mapped `results.transactions_deleted` — a field that never existed on the new bridge result, so it **always reported 0**. Now maps `entries_deleted`.

## Where to look
- **`Table.deleteHistory` signature change** (`void` → `Promise<number>`): only caller is `ResourceBridge`; return value was previously ignored. Additive/back-compat.
- **`DeleteTransactionLogsBeforeResults` constructor param order**: `entriesDeleted` inserted before `logFilesDeleted`. The class is only ever constructed with no args, so positional callers are unaffected.
- **Cross-engine symmetry**: `entries_deleted` is now populated for both RocksDB (entries in purged whole files) and LMDB (per-entry deletions). RocksDB remains whole-file/rollover-granular — a short retention window still reclaims nothing until a file fully ages past the cutoff (documented in issue #1294); this PR makes the *reported count* truthful, it does not change purge granularity.

## Testing
- `test:unit:main` (2755 passing), `test:unit:resources` (841 passing), `delete.test.js` updated to assert the deprecated-op mapping.
- Integration `transaction-logs.test.mjs` extended to assert `entries_deleted: 0` in the no-purge case.
- A positive-count integration assertion was intentionally not added: forcing a RocksDB log-file rollover in-test is fragile (purge only deletes files entirely before the flushed position).

## Notes
- New user-facing response field `entries_deleted` — a companion docs update (HarperFast/documentation) is likely warranted.
- Cross-model review: Codex (clean, no regressions). The Gemini/`agy` reviewer returned empty output across several attempts (tooling issue), so a second model review is still outstanding.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (model: Claude Opus 4.8)

### DIFF
--- a/dataLayer/delete.ts
+++ b/dataLayer/delete.ts
@@ -109,7 +109,8 @@ export async function deleteAuditLogsBefore(deleteObj: any) {
 	await pGlobalSchema(deleteObj.schema, deleteObj.table);
 	harperLogger.info(`Finished deleting audit logs before ${deleteObj.timestamp}`);
 
-	return new DeleteAuditLogsBeforeResults(results.start_timestamp, results.end_timestamp, results.transactions_deleted);
+	// `deleteTransactionLogsBefore` returns `entries_deleted`; the legacy result exposes it as `transactions_deleted`.
+	return new DeleteAuditLogsBeforeResults(results.start_timestamp, results.end_timestamp, results.entries_deleted);
 }
 
 /**
--- a/dataLayer/harperBridge/DeleteTransactionLogsBeforeResults.ts
+++ b/dataLayer/harperBridge/DeleteTransactionLogsBeforeResults.ts
@@ -4,16 +4,19 @@
 export class DeleteTransactionLogsBeforeResults {
 	start_timestamp?: number;
 	end_timestamp?: number;
+	entries_deleted: number;
 	log_files_deleted: number;
 
 	/**
 	 * @param {number} startTimestamp
 	 * @param {number} endTimestamp
+	 * @param {number} entriesDeleted
 	 * @param {number} logFilesDeleted
 	 */
-	constructor(startTimestamp?: number, endTimestamp?: number, logFilesDeleted = 0) {
+	constructor(startTimestamp?: number, endTimestamp?: number, entriesDeleted = 0, logFilesDeleted = 0) {
 		this.start_timestamp = startTimestamp;
 		this.end_timestamp = endTimestamp;
+		this.entries_deleted = entriesDeleted;
 		this.log_files_deleted = logFilesDeleted;
 	}
 }
--- a/dataLayer/harperBridge/ResourceBridge.ts
+++ b/dataLayer/harperBridge/ResourceBridge.ts
@@ -493,16 +493,19 @@ export class ResourceBridge extends BridgeMethods {
 			if (tables) {
 				for (const table of Object.values(tables)) {
 					if (table.primaryStore instanceof RocksDatabase) {
-						const deleted = table.primaryStore.purgeLogs({ before });
+						const deleted = table.primaryStore.purgeLogs({ before, includeEntryCounts: true });
 						totalResults.log_files_deleted += deleted.length;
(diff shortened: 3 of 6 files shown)


kriszyp (approved):
Solid fix — the root cause (reading `transactions_deleted` which was never populated, always returning 0) is clearly identified and correctly fixed, the LMDB path gets parity via `deleteHistory` returning the count, and the CI failures are the pre-existing cert-reload timing flake unrelated to this change.

Two minor suggestions before/after merge:

1. **NaN guard on `file.entries`** — `purgeLogs` returns strings under a version mismatch or mock, so `file.entries` can be `undefined` and propagate `NaN` into the API response. One-liner: `(file?.entries ?? 0)` in the `.reduce`. Low production risk given the pinned rocksdb-js version, but worth hardening.

2. **`deleteObj.schema` fallback in the no-table branch** — `getDatabases()[deleteObj.database || deleteObj.schema]` is applied in the `getTable` path but not in the RocksDB purge path (line ~49). Pre-existing, not introduced here, but a natural cleanup opportunity while this area is open.

Both are non-blocking for this fix. LGTM! 🟢

— *reviewed by Claude Opus 4.8 + Gemini via agy*

PR #1411, opened 2026-06-19
Abort over-time write transactions instead of force-committing (#1407)

### LINKED ISSUE #1407: Force-committing an over-time transaction leaves orphaned secondary-index entries (atomicity violation)
# Core bug: force-committing an over-time transaction leaves orphaned secondary-index entries

## Summary

The long-transaction monitor **force-`commit()`s a still-open, in-use transaction and then reuses it**, instead of aborting and surfacing the error to the caller. This (a) violates the transaction's atomicity contract by committing a partial write set the application never asked to commit, and (b) can desynchronize secondary-index maintenance from the on-disk record, leaving **orphaned index entries** that no record-level write can repair. The only recovery is a full index rebuild (drop/recreate the attribute, or rebuild the table).

Observed in production on **v4.7.19**; root-caused in the **v5** source (`main`/beta) — same force-commit-while-open design is present in both, so the analysis and fix target current Harper.

## Impact

- **Severity: high.** Silent, persistent index corruption on any `@indexed` attribute that is updated inside a transaction held open long enough to be force-committed (default 30s).
- A range search on the corrupted attribute returns rows **that do not match the predicate** (e.g. `WHERE t <= now` returns rows whose stored value is in the future). Consumers that select work by an indexed range (schedulers, queues, TTL scans) **re-select the same rows forever** → runaway re-processing / "stuck queue."
- Atomicity violation: a partial subset of a transaction's writes is committed without the caller's consent.

## Observable symptoms

- Log spam: `Transaction was open too long and has been committed, from table: <T>/` across worker threads.
- For affected keys, the secondary index and the record diverge:
  - `search_by_conditions` `nextRenderTime <= now` returns the key **and** reports its value as `> now`.
  - The same key appears in **both** the `<= now` and `> now` index searches (two index entries for one key).
  - Reading the record by primary key shows the **correct** value — only the index is wrong.
- `delete` + re-`put` of the row does **not** clear it (delete only removes the index entry matching the record's *current* value; the orphan sits at a stale value).
- It is cleared **only** by dropping the attribute's index DBI (`drop_attribute` → `dropDBI`) and repopulating, i.e. a true index rebuild.

## Root cause

Two interacting mechanisms in core:

### 1. The monitor force-commits and *reuses* an open transaction
`resources/LMDBTransaction.ts`, `startMonitoringTxns()` (~L341–358):
```ts
if (txn.timeout <= 0) {
    harperLogger.error(`Transaction was open too long and has been committed, from table: ...`);
    txn.commit();              // commits the PARTIAL write set
    txn.timeout = txnExpiration;
}
```
- `commit()` with no options leaves `this.open = OPEN` (L176) — not `LINGERING`/`CLOSED` — and on success resets `this.writes = []; this.timestamp = 0; this.next = null` (L268–270) so the **same transaction object keeps being used** by the still-running request.
- `commit()` is asynchronous (source "before" actions L144–167, optimistic-write resolution L240–286). During those `await`s the request's own code interleaves and calls `addWrite()` (L103) / `commit()`, mutating the shared `writes` / `validated` / `writeIndex` / `timestamp` state out from under the in-flight commit.

### 2. Secondary indexes are maintained as a diff against a *cached* old value
`resources/table.ts`, per-write commit (~L2058):
```ts
updateIndices(id, existingRecord, recordToStore, transaction && { transaction });
```
- `existingRecord = existingEntry?.value` (~L1755); `existingEntry` is the JS-cached `write.entry`, only re-read from disk on retry (`LMDBTransaction.ts` L189: `if (retries > 0 || !write.entry) write.entry = write.store.getEntry(write.key)`).
- `updateIndices` removes index entries for `existingRecord`'s indexed values and adds entries for `recordToStore`'s values. When a force-commit / reuse / reordering makes `existingRecord` stale relative to the actual on-disk state, the **remove targets the wrong old value** → the real old index entry is never removed → orphan.

Net: long txn → force-commit mid-flight → index diff computed against a stale `existingEntry` → orphaned secondary-index entry that survives all record-level writes.

## Reproduction (minimal)

> Status: the production symptoms below the line are **observed/verified** (v4.7.19). These minimal steps are **derived from the v5 source analysis** and have not yet been executed as a standalone test — a unit test confirming them (and serving as the fix's regression test) is the recommended next step.

1. Table with an `@indexed` numeric attribute `t`.
2. Open a transaction; `put` a row with `t = future`. Inside the same transaction, do enough work / hold it open past `txnExpiration` (lower it via `setTxnExpiration` to make this fast), while issuing repeated updates to the row's `t`.
3. Let the monitor force-commit the transaction one or more times while updates are in flight.
4. Query `search_by_conditions t <= now`.

**Expected:** rows where the stored `t <= now`.
**Actual:** rows whose stored `t` is in the future are returned (orphaned index entry), and the same key also appears in `t > now`. `delete`+`put` does not fix it; only an index rebuild does.

## Proposed fix

**Primary — abort and surface, do not force-commit (see design recommendation below).**
The monitor should `txn.abort()` (already implemented, L303: sets `CLOSED`, clears writes) so the next `addWrite` throws "transaction no longer open" (L87) and the request fails cleanly and rolls back. Update the log/error wording from "has been committed" to "aborted after exceeding the open-transaction limit."

**Durable guarantee — make index maintenance unable to orphan (do this regardless).**
Derive the index "remove" set from the value actually stored **under the LMDB write lock**, not from a cached `write.entry`. In the transactional commit path (`LMDBTransaction.ts` L221–228) re-read `write.entry = store.getEntry(write.key)` before `updateIndices`, and/or have `updateIndices` reconcile removals against the on-disk entry. Then a stale cached entry — from force-commit, retry, or CRDT reordering — can never leave an orphan.

**If auto-recovery for long transactions is kept (availability variant):**
- After force-commit, mark the transaction `LINGERING` (commit with `{ doneWriting: true }`, L176) rather than `OPEN`. Subsequent writes then hit the lingering branch (L90–101) and commit immediately via `ImmediateTransaction` — atomic, fresh-entry-read — instead of accumulating in a reset shared array that races the monitor.
- Add a per-transaction `committing` guard so the monitor never commits a transaction mid-async-operation, and `addWrite` during a commit routes to a fresh transaction. Removes the `writes`/`writeIndex` interleave.

## Design recommendation to maintainers

**Long-transaction slowness should be surfaced to the application, not silently resolved by committing on its behalf.**

The current behavior trades a *performance* problem (a transaction held open too long) for a far worse *correctness/durability* problem (atomicity violation + persistent index corruption). Silently committing a partial write set:

- **Breaks the contract the application is coding against.** The app issued a set of writes expecting all-or-nothing at *its* commit point. Committing a prefix at an arbitrary moment produces a state the app can neither predict nor reason about.
- **Is not actionable.** A background commit happens invisibly; the app gets no signal, so it cannot retry, chunk, or back off. An *error* is actionable — the app can split the work, commit explicitly in batches, or fix the hot loop.
- **Converts a localized slowness into cluster-wide instability** (here: runaway re-processing from a corrupted index).

The correct contract should be: **core enforces a transaction time limit by aborting and raising an error; the application owns how it handles long-running work** . Performance is the app's responsibility to manage; consistency is core's responsibility to guarantee. Core should never silently mutate committed state to paper over app-side slowness.

(The component that triggered this — a `claim` loop holding a 30s+ transaction with a per-row cross-table read — is being fixed app-side to keep transactions short. That removes the *trigger*, but core should not corrupt an index when a transaction does run long.)

## References (v5 source)

- `resources/LMDBTransaction.ts`: `startMonitoringTxns` L341–358 (force-commit), `commit()` L114–302, `OPEN` reuse L176 + reset L268–270, retry-only entry re-read L189, `LINGERING` path L90–101, `abort()` L303.
- `resources/table.ts`: `updateIndices(id, existingRecord, recordToStore, ...)` ~L2058, `existingRecord = existingEntry?.value` ~L1755.

kriszyp:
Fixes #1407.

## Summary

The long-transaction monitor force-`commit()`d a still-open, in-use transaction and reused it. That (a) committed a partial write set the application never asked to commit (atomicity violation) and (b) could desync secondary-index maintenance from the on-disk record, leaving **orphaned index entries** that only a full index rebuild repairs. Observed in production on v4.7.19; the same force-commit-and-reuse design is present in v5.

This makes the monitor **abort and poison** a write-bearing transaction that exceeds the open-transaction limit instead of force-committing it. Once poisoned (`timedOut`), any subsequent `addWrite`/`commit` throws an actionable `transactionOpenTooLongError` (503), so the request rolls back cleanly and the application owns how it splits long-running work. Core keeps the consistency guarantee; it no longer silently mutates committed state to paper over app-side slowness.

Per the issue's "design recommendation", this is the **abort-and-surface** option (confirmed with the team), applied symmetrically to **both** transaction engines: the RocksDB `DatabaseTransaction` monitor and the `LMDBTransaction` monitor.

## What changed

- `DatabaseTransaction` / `LMDBTransaction` monitors: a timed-out **write-bearing** txn is aborted via `abortDueToTimeout()` (sets `timedOut`, forces `open = CLOSED` before releasing the read snapshot so the discard path is taken, then `abort()`). Guards in `addWrite`/`commit` throw `transactionOpenTooLongError` once poisoned.
- **Read-only** long transactions keep their prior behavior (no atomicity/index risk — e.g. a large scan/export), so ongoing reads aren't poisoned.
- **Multi-store**: `hasPendingWrites()` walks the `next` chain, so a transaction that reads database A (head, no writes of its own) and writes database B (`next`) is correctly treated as write-bearing rather than misclassified read-only and force-committed. `abortDueToTimeout()` poisons + aborts the whole chain.
- **Canonical-source apply / replay excluded**: `sourceApply` and `isReplay` transactions keep the prior force-commit behavior — they have no resubscribe/resume path, so aborting a write would drop it while the resume cursor advances past it (a harper-pro#348-class permanent divergence).
- Tests (`unitTests/resources/txn-tracking.test.js`): write-bearing abort leaves no record/orphaned index entry; multi-store write on the `next` chain is not force-committed; source-apply txn is preserved. Pass on both RocksDB and LMDB.

## Where to look / open items

- **`transactionOpenTooLongError` status code — now 422** (was 503; cross-model review and the Gemini bot both flagged 503 as inviting auto-retry of the same doomed transaction). 422 signals the request itself must change. Swap to 408/400 if you'd prefer.
- **In-flight-commit race (low severity, not fixed here).** If the monitor aborts a write txn whose async `commit()` is already in flight (awaiting `before` hooks), the resumed continuation does a second `--readTxnsUsed` (→ -1) and a double `transaction.abort()` that the pre-existing try/catch swallows. Cross-model review confirmed the **data is still correctly rolled back and the request errors out** — the only artifact is an inert negative counter on a dead txn object. I left the hot commit-continuation untouched rather than add a guard under a race that's hard to unit-test; flagging it for a reviewer's call.

Reviewed cross-model (Gemini + Harper-domain deep-review); the two findings above are the only items not fully resolved in the diff. Generated with Claude Opus 4.8 (1M context).


resources/LMDBTransaction.ts
```diff
@@ -347,14 +350,32 @@ function startMonitoringTxns() {
...
+					} catch (error) {
+						harperLogger.debug?.(`Error aborting timed out transaction: ${error.message}`);
+					}
+				} else {
+					// Read-only long transaction (no atomicity/index risk — e.g. a large scan or export), or a
+					// canonical-source apply/replay that must never drop a write: preserve the prior behavior of
+					// committing to close out the snapshot without poisoning the transaction.
+					txn.commit();
+					txn.timeout = txnExpiration;
+				}
```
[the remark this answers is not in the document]
kriszyp: Good catch — wrapped the read-only/source-apply branch's `txn.commit()` in try/catch and added a `.catch()` on the async path, mirroring the `DatabaseTransaction` monitor so a source-apply async commit can't surface an unhandled rejection. Fixed in 21ed179c2.

— Claude (Opus 4.8)

resources/DatabaseTransaction.ts
```diff
@@ -441,6 +461,37 @@ export class DatabaseTransaction implements Transaction {
...
+	abortDueToTimeout(): void {
+		for (let txn: DatabaseTransaction = this; txn; txn = txn.next) {
+			txn.timedOut = true;
+			// Force the CLOSED path when releasing the read snapshot: doneReadTxn() flushes lingering writes via
+			// commit() while open === LINGERING, and commit() now throws transactionOpenTooLongError once poisoned.
+			// Closing first makes the release discard (abort) the uncommitted writes instead, which is the intent.
+			txn.open = TRANSACTION_STATE.CLOSED;
+			txn.abort();
+		}
+	}
```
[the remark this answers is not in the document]
kriszyp: Agreed. Restructured `abortDueToTimeout` to poison every link in the chain first (set `timedOut` + `open = CLOSED`), then abort each inside its own try/catch — so a throw from one link's `abort()` can't leave later links un-poisoned and eligible for a force-commit. Fixed in 21ed179c2.

— Claude (Opus 4.8)

resources/DatabaseTransaction.ts
```diff
@@ -34,6 +34,20 @@ let txnExpiration = envMngr.get(CONFIG_PARAMS.STORAGE_MAXTRANSACTIONOPENTIME) ??
...
+ * write set preserves atomicity and avoids the index corruption described in issue #1407: the
+ * application gets an actionable error and owns how it splits long-running work into smaller
+ * transactions, while core keeps the consistency guarantee.
+ */
+export function transactionOpenTooLongError(): ServerError {
+	return new ServerError(
+		'Transaction was aborted after exceeding the open-transaction limit; split long-running work into smaller transactions',
+		503
+	);
+}
```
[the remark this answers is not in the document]
kriszyp: Changed 503 → 422. The condition is deterministic for a given transaction shape, so a retryable status (503/408) would invite clients/gateways to auto-retry the same doomed long transaction; 422 signals the request itself must change (split the work). Fixed in 21ed179c2.

— Claude (Opus 4.8)

resources/DatabaseTransaction.ts
```diff
@@ -173,6 +194,7 @@ export class DatabaseTransaction implements Transaction {
 	}
 
 	addWrite(operation: TransactionWrite) {
+		if (this.timedOut) throw transactionOpenTooLongError();
```
cb1kenobi, line 201: This is weird ergonomics to throw a function. Instead I would recommend declaring a `TransactionOpenTooLongError` class that extends `ServerError`, then throw a new instance.

resources/DatabaseTransaction.ts
```diff
@@ -441,6 +464,45 @@ export class DatabaseTransaction implements Transaction {
...
+		// Poison every link first, then abort each, so a throw from one link's abort() can't leave later links
+		// in the chain un-poisoned (and thus eligible to be force-committed by a later commit cascade).
+		for (let txn: DatabaseTransaction = this; txn; txn = txn.next) {
+			txn.timedOut = true;
+			// Force the CLOSED path when releasing the read snapshot: doneReadTxn() flushes lingering writes via
+			// commit() while open === LINGERING, and commit() now throws transactionOpenTooLongError once poisoned.
+			// Closing first makes the release discard (abort) the uncommitted writes instead, which is the intent.
+			txn.open = TRANSACTION_STATE.CLOSED;
+		}
+		for (let txn: DatabaseTransaction = this; txn; txn = txn.next) {
```
cb1kenobi, line 506: Why the second loop? Could the first loop do the abort?

kriszyp:
Flagging a reproduction this PR should resolve. In an exploratory pass (qa-explorer, main `28db4fde4`) we hit the silent-partial-drop variant of #1407 from the **custom-resource handler** side:

- A custom resource handler whose transaction exceeds `storage.maxTransactionOpenTime` (default 30s) gets **force-committed mid-handler** by the long-transaction monitor. Writes made *before* the fire point are silently lost; writes *after* it land on a fresh implicit txn and survive; the request still returns **HTTP 200** ("count: 6" but only 1/6 rows durable, reproduced 2/2).
- Narrow trigger: `write(s) → idle async suspend with no intervening write (a bare `await`/slow external I/O > maxTransactionOpenTime) → write(s)`. Bulk writes do **not** trigger it (each `put()` resets `txn.timeout` at `DatabaseTransaction.ts:115`). Natural shape: a handler that writes a pending row, `await`s a slow external API (>30s), then writes the result — the pre-await row is the casualty.

This PR's **abort-instead-of-force-commit** is exactly the right direction (the client should get an error / the writes should roll back atomically, rather than a silent 200 + partial drop). One thing worth confirming before merge: that the abort path covers the **custom-resource handler** force-commit (not just the replication/apply path), and that the caller receives a non-2xx. Repro: `integrationTests/qa-scratch/qa314-overtime-atomicity.test.ts` (happy to port it into a regression test for this PR).

— filed by KrAIs (Claude) on Kris's behalf

resources/DatabaseTransaction.ts
kriszyp: Agreed — the factory-function pattern is unusual. Will refactor to a `TransactionOpenTooLongError` class extending `ServerError`, then `throw new TransactionOpenTooLongError()` at the call sites in a follow-up commit.

— KrAIs (claude-sonnet-4-6)

resources/DatabaseTransaction.ts
kriszyp: The separation is intentional: the first loop poisons every link in the chain (`timedOut=true`, `open=CLOSED`) before any `abort()` call. If `abort()` throws at link N in a combined loop, links N+1…M remain un-poisoned — their `timedOut` is still `false`, making them eligible to force-commit on the next monitor pass. Splitting the loops guarantees the poison-all-first invariant even if any `abort()` throws.

— KrAIs (claude-sonnet-4-6)

kriszyp:
Follow-up on the custom-resource-handler repro flagged above — the promised regression test has landed: `integrationTests/resources/txn-overtime-atomicity.test.ts` (+ fixture in the sibling `txn-overtime-atomicity/` dir), pushed to this branch.

**Confirmed: the abort-and-poison fix covers the custom-resource-handler path**, not just replication/apply. The test drives a custom Resource whose handler writes a row, awaits past `storage.maxTransactionOpenTime` with no intervening write (the exact narrow trigger described above), then writes a second row:

- On this branch: the request returns **422** and neither the pre-await nor the post-await row is durably committed — atomic rollback, no partial drop.
- On `main` (pre-#1407 baseline, checked against the same test): the request returns **200** and *both* rows land (the force-commit + fresh-implicit-txn split), confirming the test is a real regression guard and not a tautology.

Also ran the PR's own `unitTests/resources/txn-tracking.test.js` and the full `test:unit:resources` suite (852 passing) against this branch — no regressions.

One incidental note, unrelated to this PR's diff: `npm run build` on this branch currently fails `tsc` type-checking at `resources/DatabaseTransaction.ts:305` (`Promise<number | void>` not assignable to `MaybePromise<void>`) because the branch predates a `rocksdb-js@2.2.0` type-widening compat cast that's already on `main` (`transaction.commit() as Promise<void>`). Doesn't block `dist` emission (`noEmitOnError` is off) or this test, but worth a rebase before merge to pick up a clean `tsc` pass.

— filed by KrAIs (Claude) on Kris's behalf

PR #1442, opened 2026-06-22
fix(encoder): honor structures.isCompatible in CAS check for typed-only saves

### LINKED ISSUE #1441: Typed-struct CAS clobber loses concurrent structure updates
### Bug Summary

Concurrent typed-struct saves silently overwrite each other; records the first writer encoded against decode as null after the clobber.

### Steps to Reproduce

1. Run `npx mocha unitTests/resources/recordEncoder.test.js --grep "RecordEncoder typed-struct CAS regression"`
2. The first test (`rejects a conflicting typed-only save when isCompatible is on structures, not the parameter`) reproduces the clobber: a second writer saves a different typed struct at the same `lastTypedLen` and the CAS check fails to reject it.
3. Cross-process production trigger: two threads/connections on a v5 RocksDB node both call `_saveTypedStructures` against the same table around the same time, e.g. during a v4→v5 base-copy resync apply.

### Expected Behavior

`saveStructures` rejects a same-length typed-struct save with conflicting content (returns false), preserving the first writer's persisted struct. Records encoded against it continue to decode correctly.

### Actual Behavior

`saveStructures` returns true and overwrites the persisted struct. Records the first writer encoded against now reference a struct whose shape has been replaced. On read they surface as `Data read, but end of buffer not reached` or `Unexpected end of MessagePack data`. `RecordEncoder.decode` catch block at `resources/RecordEncoder.ts:457` logs and returns null, laundering the corruption as a missing record.

### Platform

### Console Errors

```
[http/2] [error]: Error decoding record Error: Data read, but end of buffer not reached
    at checkedRead (/home/harperdb/.npm-global/lib/node_modules/@harperfast/harper-pro/node_modules/msgpackr/unpack.js:217:10)
    at RecordEncoder.unpack (/home/harperdb/.npm-global/lib/node_modules/@harperfast/harper-pro/node_modules/msgpackr/unpack.js:96:12)
    at RecordEncoder.decode (/home/harperdb/.npm-global/lib/node_modules/@harperfast/harper-pro/node_modules/structon/index.js:99:37)
    at Store.decodeValue (/home/harperdb/.npm-global/lib/node_modules/@harperfast/harper-pro/node_modules/@harperfast/rocksdb-js/src/store.ts:473:24)
```

### Additional Context

Mechanism: `structon`'s `_saveTypedStructures` (`node_modules/structon/index.js:165`) calls `this.saveStructures(structures)` without forwarding `isCompatible` as the second argument. The structures Map carries `isCompatible` as an attached property which Harper's `saveStructures` override at `resources/RecordEncoder.ts:280` never consults; the fallback CAS check evaluates `existingStructures.length !== undefined` which on the typed-only path is `Map.length(undefined) !== undefined` = false, so the conflicting save commits.

Fix in `harperfast/harper-pro:5.1.6-fixes2`: extract `compat = isCompatible ?? structures?.isCompatible` and use that in both branches of the CAS check.

Regression tests added in `unitTests/resources/recordEncoder.test.js` under the `RecordEncoder typed-struct CAS regression` describe block (2 tests; without the fix the first reports "conflicting same-length typed-only save must be CAS-rejected; got true → silent clobber").

ldt1996:
Fixes #1441.

## Summary

structon's `_saveTypedStructures` (`node_modules/structon/index.js:165`) calls `this.saveStructures(structures)` without forwarding `isCompatible` as the second argument; the structures Map carries `.isCompatible` as an attached property. Harper's `saveStructures` override at `resources/RecordEncoder.ts` only inspected the parameter form, so the typed-only call path fell into the else branch where `existingStructures.length !== undefined` evaluates `Map.length(undefined) !== undefined` = false. A conflicting same-length typed-struct save would commit and silently clobber the previous one. Records the first writer encoded against then reference a struct whose shape has been replaced, surfacing as `Data read, but end of buffer not reached` or `Unexpected end of MessagePack data` on decode, returned as null via the existing catch path.

## Fix

Extract `compat = isCompatible ?? (structures as any)?.isCompatible` and use that in both branches of the CAS check. msgpackr's own pack call site (`node_modules/msgpackr/pack.js:190`), which already passes `structures.isCompatible` as the second arg, is unaffected.

## Test plan

- [x] `npx mocha unitTests/resources/recordEncoder.test.js` passes locally (11/11)
- [x] New describe block `RecordEncoder typed-struct CAS regression` (2 tests) reproduces the silent clobber without the fix and passes with it
- [x] First test asserts the typed-only call path is now CAS-rejected
- [x] Second test asserts the explicit-parameter (msgpackr.pack) path still works

## Production observation

Predicted symptom verified on Akamai v4→v5 stage cluster (5.1.6) during a rolling upgrade soak.

### DIFF
--- a/resources/RecordEncoder.ts
+++ b/resources/RecordEncoder.ts
@@ -277,16 +277,22 @@ export class RecordEncoder extends StructonEncoder {
 				// re-pack; paired with the msgpackr fix that marks structures uninitialized on
 				// save-failure, the re-pack reloads the durable structures, rebuilds the transition
 				// trie, re-mints, and re-saves — so the record always references a persisted structure.
+				// structon's _saveTypedStructures calls saveStructures(structures) without the second
+				// arg, but prepareStructures attaches .isCompatible on the Map itself. Fall back to it
+				// so typed-only saves keep CAS protection; otherwise the else-branch check is
+				// `Map.length(undefined) !== undefined` = false, and a conflicting same-length typed
+				// struct silently clobbers the previous one.
+				const compat = isCompatible ?? (structures as any)?.isCompatible;
 				const committed = this.rootStore.transactionSync(
 					(txn) => {
 						const sharedStructuresKey = [Symbol.for('structures'), this.name];
 						const existingStructuresBuffer = txn.getBinarySync(sharedStructuresKey);
 						const existingStructures = existingStructuresBuffer ? this.decode(existingStructuresBuffer) : undefined;
-						if (typeof isCompatible == 'function') {
-							if (!isCompatible(existingStructures)) {
+						if (typeof compat == 'function') {
+							if (!compat(existingStructures)) {
 								return false;
 							}
-						} else if (existingStructures && existingStructures.length !== isCompatible) {
+						} else if (existingStructures && existingStructures.length !== compat) {
 							return false;
 						}
 						txn.putSync(sharedStructuresKey, structures);
--- a/resources/RecordEncoder.ts
+++ b/resources/RecordEncoder.ts
@@ -265,6 +265,13 @@ export class RecordEncoder extends StructonEncoder {
 		const superSaveStructures = this.saveStructures;
 		const superGetStructures = this.getStructures;
 		this.saveStructures = function (structures, isCompatible): boolean | undefined {
+			// structon's _saveTypedStructures calls saveStructures(structures) without the second
+			// arg, but prepareStructures attaches .isCompatible on the Map itself. Fall back to it
+			// so typed-only saves keep CAS protection on both storage paths; otherwise the LMDB
+			// branch forwards `undefined` and the RocksDB else-branch check is
+			// `Map.length(undefined) !== undefined` = false, and a conflicting same-length typed
+			// struct silently clobbers the previous one.
+			const compat = isCompatible ?? (structures as any).isCompatible;
 			if (this.isRocksDB) {
 				// transactionSync returns the callback's value on commit, but returns `undefined`
 				// when the txn was aborted (it swallows ERR_ALREADY_ABORTED). The success path here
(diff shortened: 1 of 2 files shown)


cb1kenobi (approved):
I think this looks good

Ethan-Arrowood (approved):
lgtm!

kriszyp:
Really appreciate the careful writeup here, @ldt1996 — the `_saveTypedStructures` / `isCompatible`-as-an-attached-property detail is a genuinely subtle catch, and forwarding it is a correct robustness improvement for the standalone (non-hook) encoder path. 🙏

I went digging on whether this is reachable on Harper's actual path, and wanted to share what I found, because I think the production `Data read, but end of buffer not reached` symptom has a different origin:

**1. The path this guards isn't exercised by Harper's encode.** Harper wraps msgpackr v2, which advertises `SUPPORTS_STRUCT_HOOKS`, so structon runs the **fast path**. There, `_structonEncode` / `_saveTypedStructures` (`index.js:79/88/165`) are never invoked — every structure save goes through msgpackr's pack, which already passes `isCompatible` as the 2nd arg (`msgpackr .../node.cjs:1429`) using structon's `prepareStructures` (named + typed length CAS).

Instrumenting the real `RecordEncoder`, every `saveStructures` call is `argc=2` with `isCompatible` a function — including under a `maxOwnStructures` cap. A faithful stale-view concurrent race is CAS-**rejected** (the second writer reloads and re-mints at a new id; no clobber). The fast-path `_prepareStructures` wiring and the typed-length check both predate structon `v1.0.0`, so this holds for every `1.0.x`.

**2. No other callers.** There are no direct `saveStructures` / `_saveTypedStructures` callers anywhere in harper core outside the override definition itself, so the no-param route the else-branch guards against isn't reachable in practice.

The test reproduces the clobber by calling `enc.saveStructures(conflicting)` directly without the 2nd arg — which faithfully models structon's *standalone* call — but Harper's encode doesn't take that route, so I don't believe this change alters Harper's runtime behavior.

**Where the `end of buffer` likely comes from instead:** the structures **replay / replication-apply** path (`replayLogs.ts`, `case 'structures'`) writes the durable composite structures key via `txn.putSync` guarded only by a length (shrink) check — **no CAS** — while replay decodes each log's values (including peer logs) with the local decoder against that one shared key. A divergent same-or-longer structures buffer from another lineage can replace an existing id there, which is exactly what surfaces as `Data read, but end of buffer not reached` on later reads. That path is untouched by this PR.

**Suggestion:** keep the `isCompatible` forwarding as defense-in-depth — ideally fixed in **structon itself**, where the contract is actually broken (`_saveTypedStructures` should forward `structures.isCompatible`, matching msgpackr's pack), with a test that drives the standalone path. But I don't think it resolves the Akamai symptom; that looks like it needs a fix in the structures replay/apply path. Happy to pair on a repro to confirm.

— Claude (Opus 4.8), working with @kriszyp

ldt1996:
@kriszyp you're right, closing this.   recovery should be in #1446 + harper-pro#455, and prevention tracked in #1453

Issue #1423, opened 2026-06-20
Blob read on an incomplete/being-written/truncated file hangs the HTTP response (~20–60s) instead of returning an error


kriszyp:
## Bug Summary

A GET of a blob whose on-disk file is incomplete — being actively written, truncated, or deleted mid-read — hangs the HTTP connection for ~20–60s instead of returning a prompt error. The connection is held open until an internal timeout fires and the client gives up. A *cleanly* missing blob file (descriptor present, file fully absent at open time) correctly returns 404; only the "open succeeds but content is incomplete" path hangs.

Found via exploratory QA. Characterized on `resources/blob.ts`; please re-confirm against `main`.

## Steps to Reproduce

Any of these single-node states (all reproduce, both RocksDB and LMDB):
1. **Write-in-flight:** begin writing a large file-backed blob and delete/truncate its on-disk file (`{dataRootDir}/blobs/{db}/{p}/{p}/{fileId}`) mid-write, then GET the blob → hits the `start()` ENOENT retry loop (~1000 × 20ms ≈ 20s).
2. **Truncated after write:** fully write a blob, truncate its file on disk, then GET → hits the `pull()` "Blob is incomplete" path → `FILE_READ_TIMEOUT` (60s).
3. **Unlink after write:** fully write, unlink the file, GET → same as (2).

## Expected Behavior

A prompt, clean error — 404/410 (gone) or 503 (temporarily unavailable if a write/repair is in progress).

## Actual Behavior

The HTTP response never sends; the client times out (~20s for the ENOENT retry loop, up to 60s for the `pull()` timeout). `status=-1, timedOut=true, 0 bytes`.

## Root Cause

Two unbounded/over-long waits on the incomplete-blob path in `resources/blob.ts`:
- the `start()` ENOENT retry loop (~1000 iterations × 20ms),
- the `pull()` `FILE_READ_TIMEOUT` (60s) when `open()` succeeds but the file is incomplete.

## Impact

Medium. The hung request is **connection-isolated** — it does *not* occupy the worker event loop and concurrent requests stay healthy (verified: 4 concurrent hangs do not stall a 4-worker pool; no DoS, no uncaughtException single-node). The issue is held connections + bad observability (client gets a timeout, not an actionable error), not an availability threat.

## Suggested Fix

Bound/abort the ENOENT retry and the `pull()` read-timeout paths and return a clean error (404/410, or 503 if a write/repair is in flight) rather than hanging the response.

[...]

PR #1444, opened 2026-06-22
fix(blob): idle watchdog on writeBlobWithStream source to unwedge stalled replication receives

### LINKED ISSUE #1443: Blob replication wedges when source stream stalls without finishing
### Bug Summary

`writeBlobWithStream` waits forever when the source stream goes silent without ending or erroring; in replication this pins the per-database apply consumer at `lastReceivedStatus:"Receiving"` indefinitely.

### Steps to Reproduce

1. `npx mocha unitTests/resources/blob.test.js --grep "saveBlob with idle source stream"`
2. The 3 tests in that describe block exercise the prod-observed states:
   - PassThrough destroyed before save started
   - chunks arrived but no end/error frame ever does
   - mid-stream chunk, then destroy, no further chunks
3. Without the fix, each test's `saveBlob.saving` stays pending past the 2-3s wait; the assertion fails with "saveBlob.saving never settled".
4. Cross-process production trigger: a sender's BLOB_CHUNK frames stop arriving before the closing `finished:true`, or arrive with `finished:true,error:"..."` that the receiver records by destroying its blobsInFlight PassThrough; the subsequent audit record references the same fileId, `receiveBlobs` hands the (destroyed or idle) stream to `saveBlob`, and the pipeline sits waiting for data that never arrives.

### Expected Behavior

`saveBlob.saving` settles (resolves or rejects) in finite time even when the source stream is abandoned, so `outstandingBlobsToFinish` drains and the per-database apply consumer can advance.

### Actual Behavior

`saveBlob.saving` never settles. `outstandingBlobsToFinish` retains the stuck promise. The apply consumer's drain await blocks. `cluster_status` shows the affected channel `lastReceivedStatus:"Receiving"` with no progress, no further log signature, until the connection is force-recycled.

### Platform

### Console Errors

With the fix's idle watchdog enabled via `HARPER_BLOB_STREAM_IDLE_TIMEOUT_MS=120000`, the stalls surface as:

```
[http/2] [error] [replication]: Blob save failed for b9d from i5s-nl-ams-1 Error: Blob source stream idle for 120000ms (fileId=2951)
```

Common upstream trigger on the sender side is an ENOENT against the sender's own blob storage:

```
[http/1] [error]: Blob send error ENOENT, open '/home/harperdb/harper/blobs/seo/0/4/18c'
```

### Additional Context

Mechanism: `writeBlobWithStream` (`resources/blob.ts:727`) runs `pipeline(stream, [compressedStream,] writeStream, finished)`. `pipeline` waits for the source to emit `end` or `error`. A source that has not ended and is not errored sits in `stream.readable=true, writableEnded=false`; nothing wakes the pipeline up, `saveBlob.saving` never settles, and the consumer at `replicationConnection.ts` Promise.all over `outstandingBlobsToFinish` blocks indefinitely.

Fix in `harperfast/harper-pro:5.1.6-fixes2`: opt-in source-idle watchdog in `writeBlobWithStream`, armed via env var `HARPER_BLOB_STREAM_IDLE_TIMEOUT_MS` (default 0 = off, replication sets 120000ms). Armed on every `data` event; on expiry calls `stream.destroy(new Error(...))` so pipeline rejects cleanly. Cleared in the `finished` callback.

Production verification: nl-ams-1 fired the watchdog 420 times across the post-upgrade catch-up window, each catch unwedging an apply that would have stalled permanently on 5.1.6 stock.

Regression covered by `unitTests/resources/blob.test.js` "saveBlob with idle source stream (replication wedge regression)" (3 tests).

ldt1996:
Fixes #1443.

## Summary

`writeBlobWithStream` runs `pipeline(stream, [compressedStream,] writeStream, finished)`. `pipeline` waits for the source to emit `end` or `error`. A source that has not ended and is not errored sits forever; nothing wakes the pipeline up, `saveBlob.saving` never settles, `outstandingBlobsToFinish` retains the stuck promise, and the per-database replication apply consumer's drain await blocks indefinitely. In production this pins the affected (sender, receiver, database) tuple at `lastReceivedStatus:"Receiving"` with no log signature.

## Fix

Opt-in source-idle watchdog in `writeBlobWithStream`, armed via `HARPER_BLOB_STREAM_IDLE_TIMEOUT_MS` (default 0 = off so behavior is unchanged for callers that do not set it; replication's blob-receive path sets 120000ms). Watchdog is armed on every `data` event; on expiry it calls `(stream as Readable).destroy(new Error(...))` so the pipeline rejects cleanly with `Blob source stream idle for <ms>ms (fileId=<id>)`. Cleared in the existing `finished` callback so successful streams pay nothing past the last `data` event.

The replication layer's separate `blobsInFlight` timer only helps when the stuck stream is still in that map; this guard catches the case where a finishing chunk arrived but the source never ended.

## Test plan

- [x] `npx mocha unitTests/resources/blob.test.js --grep "saveBlob with idle source stream"` passes (3/3)
- [x] New describe block `saveBlob with idle source stream (replication wedge regression)` covers the three prod-observed states: destroyed-before-save, chunks-but-never-ends, mid-stream-then-destroy
- [x] Pre-existing `Blob test` suite failures (`#1423`) are unrelated to this change (verified by running on clean main)

## Production observation

Akamai v4→v5 stage cluster (5.1.6-fixes2): post-upgrade catch-up on nl-ams-1 fired the watchdog 420 times, each catch unwedging an apply that would have stalled permanently on 5.1.6 stock; 6077 corresponding ENOENT blob-send errors on `/home/harperdb/harper/blobs/resilience/...` confirm the upstream missing-blob source on the sender.

## Companion change

A complementary sender-side timeout lives in harper-pro (`replication/replicationConnection.ts` `sendBlobs` per-chunk `Promise.race` against `setTimeout`); both halves are bundled in `harperfast/harper-pro:5.1.6-fixes2`. That PR will be opened separately against harper-pro.

### DIFF
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -724,6 +724,18 @@ export function saveBlob(blob: FileBackedBlob, deleteOnFailure = false) {
 /**
  * Create a blob from a readable stream
  */
+// Optional source-stream idle timeout for writeBlobWithStream. When > 0, a stream that goes this
+// long without delivering data is force-destroyed so the save promise rejects in finite time.
+// Default 0 (off) preserves prior behavior. Replication's blob-receive path enables it: without
+// this, pipeline(source, writeStream, finished) sits forever on a sender that never sends the
+// closing `finished:true` BLOB_CHUNK, saveBlob.saving never settles, outstandingBlobsToFinish
+// keeps the stuck promise, and the per-database apply consumer's drain await wedges. Surfaced in
+// production as a (sender, receiver, database) tuple pinned at lastReceivedStatus="Receiving"
+// indefinitely. Read at call time so tests and operators can change it without a process restart.
+function getBlobStreamIdleTimeoutMs(): number {
+	return Number(process.env.HARPER_BLOB_STREAM_IDLE_TIMEOUT_MS) || 0;
+}
+
 function writeBlobWithStream(blob: Blob, stream: NodeJS.ReadableStream, storageInfo: StorageInfo): Blob {
 	const { filePath, fileId, store, compress, flush } = storageInfo;
 	storageInfo.saving = new Promise((resolve, reject) => {
@@ -738,6 +750,26 @@ function writeBlobWithStream(blob: Blob, stream: NodeJS.ReadableStream, storageI
 			// if we know the size, we can write the header immediately
 			writeStream.write(createHeader(blob.size)); // write the default header
 			wroteSize = true;
+		}
+		// Source-idle watchdog: armed on every 'data' event, destroys the source if it goes silent
+		// (no end, no error, no more data) for longer than the configured threshold so pipeline
+		// rejects cleanly. Off by default; replication sets HARPER_BLOB_STREAM_IDLE_TIMEOUT_MS so an
+		// abandoned BLOB_CHUNK stream cannot wedge the apply consumer. The replication layer's
+		// separate blobsInFlight timer only helps when the stuck stream is still in that map; this
+		// guard catches the case where a finishing chunk arrived but the source never ended.
+		let idleTimer: NodeJS.Timeout | undefined;
+		const idleTimeoutMs = getBlobStreamIdleTimeoutMs();
+		if (idleTimeoutMs > 0) {
+			const armIdleTimer = () => {
+				if (idleTimer) clearTimeout(idleTimer);
+				idleTimer = setTimeout(() => {
+					(stream as Readable).destroy(
+						new Error(`Blob source stream idle for ${idleTimeoutMs}ms (fileId=${fileId})`)
+					);
+				}, idleTimeoutMs).unref();
+			};
+			stream.on('data', armIdleTimer);
+			armIdleTimer();
 		}
 		let compressedStream: any;
 		if (compress) {
@@ -758,6 +790,10 @@ function writeBlobWithStream(blob: Blob, stream: NodeJS.ReadableStream, storageI
 		}
 		// when the stream is finished, we may need to flush, and then close the handle and resolve the promise
 		function finished(error?: Error) {
+			if (idleTimer) {
+				clearTimeout(idleTimer);
+				idleTimer = undefined;
+			}
 			const fd = (writeStream as any).fd;
 			if (error) {
 				store.unlock(lockKey);
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -724,16 +724,21 @@ export function saveBlob(blob: FileBackedBlob, deleteOnFailure = false) {
 /**
  * Create a blob from a readable stream
  */
-// Optional source-stream idle timeout for writeBlobWithStream. When > 0, a stream that goes this
-// long without delivering data is force-destroyed so the save promise rejects in finite time.
-// Default 0 (off) preserves prior behavior. Replication's blob-receive path enables it: without
-// this, pipeline(source, writeStream, finished) sits forever on a sender that never sends the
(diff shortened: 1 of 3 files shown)


kriszyp:
The receiver-side watchdog is the load-bearing half here (it rescues independent of the sender), and the three regression tests nailing the destroyed/never-ended/torn-down PassThrough cases are great. 🙏

Same blocker as the sender PR (harper-pro#451): **defaults to `0` (off)** and nothing sets `HARPER_BLOB_STREAM_IDLE_TIMEOUT_MS` — the comment says "replication's blob-receive path enables it," but that setter doesn't exist in harper or harper-pro. As written it ships inert.

Live stakes: a customer cluster **preprod 5.1.7** is wedged on exactly this (harper-pro#453) — a stalled source blob stream leaves `outstandingBlobsToFinish` pinned, the apply consumer's drain await never resolves, the follower sticks at `Receiving`/`ver=0`, and replicated deploys are blocked with no self-heal. This watchdog *enabled* is what unwedges it.

Suggestion: give it a non-zero default in the replication receive path (mirror harper-pro `REPLICATION_BLOBTIMEOUT ?? 120000`), so 5.1.8 actually fixes the wedge rather than shipping a dormant switch. Glad to help wire the enablement + add a rolling-restart-during-copy integration case.

*— Claude (Opus 4.8), from live a customer cluster investigation*

kriszyp:
Pushed an enablement commit (`eb6a309`) per @kriszyp's call: `getBlobStreamIdleTimeoutMs()` now defaults to **120000ms** instead of `0`; `HARPER_BLOB_STREAM_IDLE_TIMEOUT_MS` still overrides (set `0` to disable).

One thing to confirm since this lives in core `blob.ts` (not replication-only): it now applies to **every** `writeBlobWithStream` caller. I traced them — the only source that can sit idle this long is the replication-fed `storageInfo.source` PassThrough; the `Readable.from(buffer/stream)` callers deliver immediately and finish, and your re-arm-on-pause guard means legitimate write backpressure never trips it. So the practical blast radius is the replication receive path this targets. If you'd rather keep it strictly replication-scoped, the alternative is having harper-pro's replication init set the env rather than core defaulting it — happy to switch to that if you prefer. The three regression tests still pin the env explicitly, so they're unaffected.

Nice fix — the destroyed/never-ended/torn-down PassThrough coverage is exactly the a customer cluster (#453) shape. 🙏

*— Claude (Opus 4.8)*

resources/blob.ts
```diff
@@ -724,6 +724,23 @@ export function saveBlob(blob: FileBackedBlob, deleteOnFailure = false) {
...
+// overrides it (set 0 to disable). The timer re-arms while the source is paused, so legitimate write
+// backpressure never trips it — only a genuinely silent source is destroyed. Read at call time so
+// tests and operators can change it without a process restart.
+const DEFAULT_BLOB_STREAM_IDLE_TIMEOUT_MS = 120000;
+function getBlobStreamIdleTimeoutMs(): number {
+	const configured = process.env.HARPER_BLOB_STREAM_IDLE_TIMEOUT_MS;
+	return configured != null ? Number(configured) : DEFAULT_BLOB_STREAM_IDLE_TIMEOUT_MS;
+}
+
 function writeBlobWithStream(blob: Blob, stream: NodeJS.ReadableStream, storageInfo: StorageInfo): Blob {
```
Ethan-Arrowood: I think we can switch `NodeJS.ReadableStream` should become `Readable` instead.

Ethan-Arrowood (approved):

resources/blob.ts
ldt1996: agreed

PR #1446, opened 2026-06-22
fix(encoder): reload structures and retry once on dict-lag decode error

### LINKED ISSUE #1445: Reader-side classic shared-structures dict lag silently nulls records on decode
### Bug Summary

When the local in-memory shared-structures dict is shorter than the persistent buffer the writer used, `RecordEncoder.decode` throws "Data read, but end of buffer not reached" and returns the record as null. The stored bytes are intact.

### Steps to Reproduce

1. `npx mocha unitTests/resources/recordEncoder.test.js --grep "structure-version drift on read"`
2. The 3 tests in that describe block model the prod state: writer published a 3-field classic shared structure, reader's in-memory `structures` is at the 2-field prior version.
3. Without the fix, decode produces a null and logs `Error decoding record Error: Data read, but end of buffer not reached`.
4. Cross-process production trigger: v4→v5 in-place upgrade where the first-upgraded node's local singleton encoder ingests records via base-copy resync from peers before its in-memory structures buffer reloads to match what got persisted.

### Expected Behavior

Decode tolerates an in-memory dict lag: on `Data read, but end of buffer not reached`, reload the persistent structures via `getStructures()` and retry once. If reload doesn't help, return null and log to the generic error path (existing behavior).

### Actual Behavior

Decode logs the error and returns null on the first attempt. The record is silently treated as missing by every reader of that struct ID until the singleton encoder happens to refresh.

### Platform

### Console Errors

```
[http/2] [error]: Error decoding record Error: Data read, but end of buffer not reached
    at checkedRead (/home/harperdb/.npm-global/lib/node_modules/@harperfast/harper-pro/node_modules/msgpackr/unpack.js:217:10)
    at RecordEncoder.unpack (/home/harperdb/.npm-global/lib/node_modules/@harperfast/harper-pro/node_modules/msgpackr/unpack.js:96:12)
    at Store.decodeValue (/home/harperdb/.npm-global/lib/node_modules/@harperfast/harper-pro/node_modules/@harperfast/rocksdb-js/src/store.ts:473:24)
data: 4279eed6d58d4a030e0000400000000142c5...
```

### Additional Context

Mechanism: `RecordEncoder.decode` catch block at `resources/RecordEncoder.ts:449` returns null on any non-missing-structure decode error. Typed (random-access) structures auto-reload via structon's `_readStruct` on a miss (see harper#1163), but the classic-structures path has no equivalent recovery for tail-shape mismatch.

Fix in `harperfast/harper-pro:5.1.6-fixes3`: on a `Data read, but end of buffer not reached` error, call `getStructures()` and `_mergeStructures()` then retry decode once. `_reloadingStructures` instance flag guards a single retry to avoid recursion when reload doesn't advance the dict.

Production observation: 23 unique HttpCache records on nl-ams-1 (first upgraded node) silently lost over a 13-minute post-upgrade window. se-sto-1 (second upgrade, had a v5 peer) lost 0, isolating the trigger to the first-upgrade base-copy resync window.

Distinct from harper#1441: that one is a writer-side CAS clobber that destroys the persisted structures; this one is a reader-side in-memory lag while persisted structures are intact.

Regression tests added in `unitTests/resources/recordEncoder.test.js` under the `RecordEncoder structure-version drift on read` describe block (3 tests: recovery, recursion guard, pure symptom).

ldt1996:
Fixes #1445.

## Summary

`RecordEncoder.decode` catch block returns null on any non-missing-structure decode error. When the local in-memory `structures` array is shorter than the persistent buffer the writer used, msgpackr's `checkedRead` flags leftover bytes for the (N+k)-th field and throws "Data read, but end of buffer not reached". The stored bytes are intact: only the in-memory singleton encoder dict has fallen behind what `saveStructures` already persisted. Without recovery, every read of that struct ID returns null until the singleton happens to refresh, silently laundering data as missing.

Typed (random-access) structures auto-reload via structon's `_readStruct` on a miss (`harper#1163`). The classic-structures path has no equivalent recovery for the tail-shape mismatch.

## Fix

In the existing catch block, when the error message matches `end of buffer not reached`, call `getStructures()` (reads the persistent buffer), `_mergeStructures()` to update the in-memory dict, then retry `decode` once. A `_reloadingStructures` instance flag guards a single retry so a reload that doesn't actually advance the dict falls straight through to the original null + error log path rather than recursing.

## Test plan

- [x] `npx mocha unitTests/resources/recordEncoder.test.js` passes (12/12, includes the 3 new dict-lag tests)
- [x] `recovers via getStructures() reload when persistent dict has caught up to writer` asserts the fix succeeds when storage has the fresh dict
- [x] `falls through to null + error log when reload does not advance the dict` asserts the recursion guard prevents infinite retry and real corruption still surfaces
- [x] `reports "Data read, but end of buffer not reached" when reader's structures are behind writer's` asserts the symptom is still detected (no regression of the existing log path)

## Production observation

Akamai v4→v5 stage cluster (5.1.6-fixes2): 23 unique HttpCache records on the first upgraded node silently lost over a 13-minute post-upgrade window. Second upgrade (with a v5 peer to sync from) lost 0, isolating the trigger to the first-upgrade base-copy resync window when the local singleton encoder ingests records before its in-memory dict reloads.

## Distinct from #1441

That bug is a writer-side CAS clobber: the persisted structures buffer itself was overwritten. This bug is a reader-side in-memory lag: the persisted buffer is fine, only the local encoder hasn't reloaded.

### DIFF
--- a/resources/RecordEncoder.ts
+++ b/resources/RecordEncoder.ts
@@ -448,18 +448,20 @@ export class RecordEncoder extends StructonEncoder {
 				);
 				return null;
 			}
-			// Classic shared-structures dict lag: persisted structures advanced past the in-memory copy,
-			// so msgpackr decodes the known fields and leaves the tail. Reload from storage and retry
-			// once. structon does the equivalent on typed-struct misses (harper#1163).
-			if (!this._reloadingStructures && /end of buffer not reached/i.test(error?.message)) {
-				try {
-					const fresh = this.getStructures();
-					if (fresh) this._mergeStructures(fresh);
-				} catch (reloadError) {
-					harperLogger.warn?.('Failed to reload structures during decode-retry', reloadError);
-				}
+			// In-memory structures dict diverged from durable (harper#1453); reload and retry. Guard
+			// is set before getStructures() because on RocksDB it itself decodes the structures buffer.
+			if (
+				!this._reloadingStructures &&
+				/end of buffer not reached|unexpected end of messagepack data/i.test(error?.message)
+			) {
 				this._reloadingStructures = true;
 				try {
+					try {
+						const fresh = this.getStructures();
+						if (fresh) this._mergeStructures(fresh);
+					} catch (reloadError) {
+						harperLogger.warn?.('Failed to reload structures during decode-retry', reloadError);
+					}
 					return this.decode(buffer, options);
 				} finally {
 					this._reloadingStructures = false;
--- a/resources/RecordEncoder.ts
+++ b/resources/RecordEncoder.ts
@@ -464,10 +464,7 @@ export class RecordEncoder extends StructonEncoder {
 					const fresh = this.getStructures();
 					if (fresh) this._mergeStructures(fresh);
 				} catch (reloadError) {
-					harperLogger.warn?.(
-						'Failed to reload structures during decode-retry; continuing to error path',
-						reloadError
-					);
+					harperLogger.warn?.('Failed to reload structures during decode-retry; continuing to error path', reloadError);
 				}
 				this._reloadingStructures = true;
 				try {
--- a/resources/RecordEncoder.ts
+++ b/resources/RecordEncoder.ts
@@ -114,10 +114,12 @@ export class RecordEncoder extends StructonEncoder {
 	declare saveStructures: any;
 	declare getStructures: any;
 	declare _writeStruct: any;
+	declare _mergeStructures: any;
 	structureUpdate?: any;
 	isRocksDB: boolean;
 	name: string;
 	useVersions: boolean;
+	_reloadingStructures?: boolean;
 	constructor(options) {
 		options.useBigIntExtension = true;
 		// Bound the per-encoder typed-structure dictionary. It is append-only and pinned on the
@@ -449,11 +451,7 @@ export class RecordEncoder extends StructonEncoder {
 			// Classic shared-structures dict lag: persisted structures advanced past the in-memory copy,
 			// so msgpackr decodes the known fields and leaves the tail. Reload from storage and retry
 			// once. structon does the equivalent on typed-struct misses (harper#1163).
-			if (
-				!this._reloadingStructures &&
-				typeof this.getStructures === 'function' &&
-				/end of buffer not reached/i.test(error?.message)
-			) {
+			if (!this._reloadingStructures && /end of buffer not reached/i.test(error?.message)) {
 				try {
 					const fresh = this.getStructures();
 					if (fresh) this._mergeStructures(fresh);
--- a/resources/RecordEncoder.ts
+++ b/resources/RecordEncoder.ts
@@ -446,6 +446,36 @@ export class RecordEncoder extends StructonEncoder {
 				);
 				return null;
 			}
+			// Reader-side classic shared-structures dictionary lag: msgpackr decoded the N fields its
+			// local dict knew about, then checkedRead flagged leftover bytes for the (N+k)-th field.
+			// The persistent structures buffer is already at N+k (the writer published it via
+			// saveStructures); the local singleton encoder just hasn't reloaded since. Force a reload
+			// and retry once. Mirrors structon's auto-reload for the typed-struct missing case
+			// (harper#1163), but for classic structures the upstream path doesn't self-recover on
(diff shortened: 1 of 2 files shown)


resources/RecordEncoder.ts
```diff
@@ -446,6 +448,23 @@ export class RecordEncoder extends StructonEncoder {
...
+					if (fresh) this._mergeStructures(fresh);
+				} catch (reloadError) {
+					harperLogger.warn?.('Failed to reload structures during decode-retry', reloadError);
+				}
+				this._reloadingStructures = true;
+				try {
+					return this.decode(buffer, options);
+				} finally {
+					this._reloadingStructures = false;
+				}
```
Ethan-Arrowood: I think the recursion guard is set one step too late here. On the RocksDB path `getStructures()` calls `decode()` on the structures buffer itself, so if that decode ever throws the same "end of buffer not reached" error it re-enters this branch with `_reloadingStructures` still `false` and recurses unbounded. Setting the flag *before* `getStructures()` and wrapping the whole reload+retry in one `try/finally` closes the gap. Behavior is identical on the happy path and the "reload didn't advance" path (the inner `try/catch` still just warns and falls through to a clean null + error log):

```suggestion
				this._reloadingStructures = true;
				try {
					try {
						const fresh = this.getStructures();
						if (fresh) this._mergeStructures(fresh);
					} catch (reloadError) {
						harperLogger.warn?.('Failed to reload structures during decode-retry', reloadError);
					}
					// Retry once. getStructures() above already ran under the guard, so the nested
					// structures-buffer decode can't re-enter this reload branch and recurse.
					return this.decode(buffer, options);
				} finally {
					this._reloadingStructures = false;
				}
```

Ethan-Arrowood (approved):
Nice fix — the diagnosis is well-supported by the production observation, and the three tests cover the recovery, the guard-against-recursion, and the symptom-still-surfaces cases cleanly. This is the right reader-side complement to the writer-side CAS fix in #1442.

Approving. One non-blocking hardening I left inline: the `_reloadingStructures` guard is set just after `getStructures()`, but on the RocksDB path `getStructures()` itself decodes the structures buffer, so setting the flag *before* that call (and wrapping reload + retry in one try/finally) closes a narrow re-entrancy gap. Safe to take as a follow-up.

kriszyp:
Following up here now that the root cause is fully pinned (consolidated in #1453): I think this PR is **superseded by the prevention fix in #1455**, and I'd suggest closing it — but wanted to lay out the reasoning since your diagnosis here is what drove us to the root cause. 🙏

**What the byte-level trace on the live node showed:** the durable structures dictionary is canonical/correct; the failing reads come from a *worker whose in-memory dict forked* — it minted the v5 classic record-structures from scratch during the migration window (because the migration persisted no classic dictionary), assigned ids that diverge from durable, and being a read-heavy GET worker it never re-mints, so msgpackr's missing-id reload (`unpack.js:247`) never fires and it decodes every record against the stale fork → `Data read, but end of buffer not reached` / null.

**Why prevention supersedes the reader-reload:** msgpackr's contract is that **structures only grow, never mutate** — so a reader has exactly two cases: the id is present-and-correct, or absent-and-needs-loading (which line 247 already handles). A *present-but-wrong* id can only exist if that invariant has been violated. #1455 restores the invariant at the source: `copyDbToRocks` now builds and persists one canonical classic dictionary at the end of migration, so every worker **adopts** it instead of minting a divergent one. Once that holds, "present-but-wrong" never arises and the reload-and-retry added here would never execute.

Adding the present-but-wrong reload would mean *normalizing* an invariant violation — silently healing a state that shouldn't exist — which masks exactly the class of bug we just root-caused rather than preventing it. (To be fair: it isn't data-dangerous — since durable is canonical, the reload would actually heal that particular read — so this is a "wrong layer / papers over the violation" concern, not a correctness hazard.)

So my recommendation is to **close this in favor of #1455**, but I'll leave the actual call to you and @kriszyp since it's your PR. For the record, #1442 is a separate matter — I left a note there; it targets the typed-struct CAS path which isn't reachable via Harper's fast-path encode, so it's orthogonal to this, not superseded by #1455.

— Claude (Opus 4.8), working with @kriszyp

ldt1996:
cloosing this in favor of #1455

PR #1452, opened 2026-06-22
Clarify upgrade migration log: show the data→software transition and migration purpose

### LINKED ISSUE #1451: Upgrade log names the migration's version, not the release being installed (misleading)
## Problem

During an upgrade, Harper logs a line like:

```
Running upgrade for version 5.1.0
```

even when the installed/target version is something later (e.g. 5.1.7). The full sequence a user sees:

```
This version of Harper Pro is 5.1.7
Harper Pro is completing an update to version 5.1.7. ...
Starting upgrade process...
[notify]: Running upgrade for version 5.1.0
Running upgrade for version 5.1.0
Harper Pro was successfully upgraded to version 5.1.7
```

The `5.1.0` is the version of the **upgrade directive** (the data-migration step), not the software version. Directives run once, when the data's last-run version crosses the directive's version boundary (`current < directive_version <= installed`; see `upgrade/directives/directivesController.ts` → `getVersionsForUpgrade`). The only directive currently registered is the 5.1.0 one (`upgrade/directives/5-1-0.ts`, which creates `system.hdb_deployment`), so that line prints `5.1.0` for *any* upgrade that crosses it — regardless of the actual from/to versions.

This is confusing: it reads like Harper is reverting to 5.1.0, and it says nothing about what the migration actually does.

## Proposed fix

Make `upgrade/directivesManager.ts` log the actual transition and the migration's purpose instead of the bare directive version, e.g.:

```
Starting upgrade process: applying 1 data migration to bring data from 5.0.x up to software 5.1.7.
Applying migration 1 of 1 (introduced in 5.1.0) — create system.hdb_deployment table for deployment tracking
```

- Header line shows the real data → software transition and the migration count.
- Per-migration line frames the version as "introduced in X" and includes a human-readable description (new optional `description` field on the directive object).

Log/notify text only — no behavior change.

heskew:
## Summary

Reworks the upgrade-time log wording so each migration line says what's actually happening instead of printing a bare directive version.

Before (installing 5.1.7):

```
Starting upgrade process...
Running upgrade for version 5.1.0
```

After:

```
Starting upgrade process: applying 1 data migration to bring data from 5.0.22 up to software 5.1.7.
Applying migration 1 of 1 (introduced in 5.1.0): create system.hdb_deployment table for deployment tracking
```

## Why

`Running upgrade for version <X>` printed the version that *introduced* a migration directive, not the software version being installed. With only the 5.1.0 directive registered, it prints `5.1.0` for any upgrade that crosses it (e.g. while installing 5.1.7) — which reads like a downgrade and says nothing about what the migration does. It came up when the line surprised people watching a 5.1.7 upgrade. Fixes #1451.

## Changes

- `upgrade/directivesManager.ts` — header logs the real data → software transition and migration count; per-migration line is now `Applying migration N of M (introduced in <version>): <description>`.
- `upgrade/directives/5-1-0.ts` — added an optional `description` field (set on the 5.1.0 directive).
- Unit tests for the new wording and the empty-migrations branch.

No change to the upgrade/migration logic — `log`/`console` text only.

## Review notes

- **Where to look:** the new header/branch logic in `processDirectives` (`upgrade/directivesManager.ts`).
- **Deliberate tradeoff (flagging since the cross-model review raised it):** I did *not* add an `undefined` guard on the header's version fields. The sole caller (`bin/upgrade.js`) always supplies both, and a missing-version object already routes to `getVersionsForUpgrade → []` (the "no migrations" branch), so it can't crash — a guard would be defensive cruft.
- **Docs:** no documentation-repo change — these log strings aren't documented.

## Cross-model review

Harper-domain pass via the `cross-model-review` skill (Opus 4.8): **no blockers, no significant concerns.** Caveat: the Gemini/`agy` leg timed out this run, so there was no second-model (Gemini) coverage.

---

### DIFF
--- a/unitTests/upgrade/directivesController.test.js
+++ b/unitTests/upgrade/directivesController.test.js
@@ -34,6 +34,9 @@ describe('directivesController — hdb_deployment table-creation directive', () 
 		const directive = directivesController.getDirectiveByVersion('5.1.0');
 		expect(directive, 'expected a directive registered for version 5.1.0').to.exist;
 		expect(directive.version).to.equal('5.1.0');
+		expect(directive.description, 'directive should carry a human-readable description for the upgrade log').to.be.a(
+			'string'
+		).that.is.not.empty;
 		expect(directive.async_functions).to.be.an('array').that.is.not.empty;
 		expect(directive.async_functions[0].name).to.equal('createHdbDeploymentIfMissing');
 	});
--- a/unitTests/upgrade/directivesManager.test.js
+++ b/unitTests/upgrade/directivesManager.test.js
@@ -1,0 +1,78 @@
+'use strict';
+
+const sinon = require('sinon');
+const chai = require('chai');
+const { expect } = chai;
+
+const directivesManager = require('#src/upgrade/directivesManager');
+const directivesController = require('#src/upgrade/directives/directivesController');
+const { UpgradeObject } = require('#src/upgrade/UpgradeObjects');
+
+// Coverage for the upgrade log wording: the per-migration line must describe the actual
+// data -> software transition and what each migration does, rather than printing the bare
+// directive version (which is the release that *introduced* the migration, not the version
+// being installed — e.g. "Running upgrade for version 5.1.0" while installing 5.1.7).
+describe('directivesManager.processDirectives — upgrade log output', () => {
+	const sandbox = sinon.createSandbox();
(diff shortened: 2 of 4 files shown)


kriszyp (approved):
Excellent!

DavidCockerill (approved):

Issue #1424, opened 2026-06-20
Blob reads and replication-send don't validate bytes against the descriptor blob.size — truncated blobs served as valid and propagate to replicas


kriszyp:
## Bug Summary

Nothing cross-validates the bytes returned (or replicated) for a blob against the length recorded in the DB descriptor (`blob.size`). The read/send paths trust the on-disk file's self-describing header. As a result, certain truncated/corrupted blob files are served to clients as valid (shorter) content with no error, and — confirmed in a cluster — propagate to a new replica via full copy as if complete.

Found via exploratory QA. Characterized on `resources/blob.ts` (core); please re-confirm against `main`.

## Steps to Reproduce

Seed a record with a 64KB file-backed blob (descriptor `blob.size = 65536`), then corrupt the on-disk file and read it:

- **T1/T2 (natural partial write — on-disk header still claims 65536, content shorter): SAFE.** `bytes()` and the replication-send path detect `header_size > file_content_size` and throw "Incomplete blob". (Good — included as the control.)
- **T3 (file overwritten to < the 8-byte header): SILENT.** Header parse is skipped; if byte[1] == DEFLATE_TYPE (0x01), `deflate(empty)` yields ~8 garbage bytes returned `ok=true`.
- **T4 (on-disk header rewritten to a *self-consistent* smaller size, content truncated to match): SILENT.** `bytes()` returns the short content; the descriptor still says 65536 but is never consulted. In a 2-node cluster, a fresh node joined via full copy **stored and served the 256-byte corrupt blob as complete** (no error, no gap flag). Not propagated by incremental updates (a blob is only re-streamed if the blob field changed).

## Expected Behavior

A truncated/incomplete blob should be detected (length mismatch vs the stored `blob.size`) and surfaced as an error on read, and must not be sent to a replica as a complete blob.

## Actual Behavior

T3 returns garbage bytes `ok=true`; T4 returns short content silently and propagates to a full-copy replica as "complete".

## Root Cause

`pack()`/the send path only check the on-disk header's self-consistency (`header_size === file_content_size`), never the authoritative descriptor `blob.size` (which is stored and survives truncation). Plus a missing `rawBytes.length < HEADER_SIZE` guard (T3).

## Impact

[...]

PR #1456, opened 2026-06-23
fix(blob): bound the stream() incomplete-content wait so a truncated blob can't spin a worker at 100% CPU (#1454)

### LINKED ISSUE #1454: Blob stream() busy-spins a worker at 100% CPU on a present-but-truncated blob whose writer lock stays held (survives #1423/#1424)
## Summary

When `FileBackedBlob.stream()` reads a blob file that is **present but truncated** — the 8-byte header records a final `size = X`, but only `Y < X` content bytes are on disk — **and** the blob write-lock reads as held, the `pull()` → `readMore()` reader enters an **unbounded, backoff-free busy loop that pegs the worker thread at ~100% CPU**. Several streams across the worker pool peg all cores regardless of request rate.

This is **not** fixed by #1423/#1424 (shipped in 5.1.7); the spin branch is byte-for-byte identical from 5.0.28 through 5.1.8.

## Production context

Observed on `prod-dyn` / `prod-gar` (GDI, harper-pro 5.0.28): both boxes pegged ~14/16 cores for 12h+ post-cutover while serving ~0.5–2 origin req/s. A CPU profile (6 threads, ~140k samples each) put ~73% of CPU in `FileBackedBlob.stream()` → `readMore()` — `fs.read` (~48%) + `Buffer.allocUnsafe(256 KB)` (~21%), ~1.5% idle. parse5/cache-miss transform was a minor secondary cost. The trigger was a wedged GDI blob-replication backlog (~11–12h behind): bodies stored as external blobs with a 2h expiration evict before the backlog ships them, and partial/stalled replication writes leave truncated blob files with a self-consistent header and a held lock — exactly the spin trigger.

## Mechanism

In the `bytesRead === 0` branch of `readMore()` (`resources/blob.ts`):

1. Re-read header → `size = X`; `X > totalContentRead (Y)` is true.
2. `checkIfIsBeingWritten()` returns `true`. Note it **caches** `isBeingWritten` once set. The lock is in-process and released on `unlock()` or `DBHandle::close()` (which reaps even expired-owner locks), so a *dead* writer cannot leave it held — the pin is a **live** in-progress write whose source stream stalled and never reached `unlock()` (the `writeBlobWithStream` pipeline that #1444 / #1443 watchdogs). This reader-side bound is the complementary half: a read must never spin regardless of why the write is stuck.
3. `readSync(position)` returns 0 (no new bytes on disk).
4. `else if (!resumeIfWriterFinished())`: `resumeIfWriterFinished()` re-reads the header, sees the size is *known* (`!== UNKNOWN_SIZE`), calls `readMore()`, and **returns `true`**.
5. Because it returned `true`, the `getBlobReadTimeout()` timer in that branch **never arms** — and `readMore()` was already re-entered with no `setTimeout`/backoff.
6. → back to step 1, forever. Each iteration does a 256 KB `allocUnsafe` + an async `read` at EOF (0 bytes) with no delay → 100% CPU.

Because `blob.size` is written into the header up front when known (the replication path), "known header size + content still arriving" is a **legitimate** in-progress state — so the existing code actually busy-spins through every slow known-size read until the writer catches up, and spins *forever* when the write stalls with the lock held.

## Why #1423/#1424 miss it

- **#1423** (`getBlobReadTimeout()` deadline) lives in the `else if (!resumeIfWriterFinished())` branch — the branch the spin never enters, because `resumeIfWriterFinished()` returns `true`. Setting `storage.blobReadTimeout` does not bound the spin.
- **#1424** (descriptor cross-check) is guarded by `isFullRead` (`start === undefined && end === undefined`); HTTP serving streams a range via `body.stream()`, so it is skipped. Even on a full read it only fires when on-disk header ≠ record descriptor; the common "header claims full, content short" truncation has header == descriptor and is not caught here.

## Fix

Bound the `resumeIfWriterFinished()` → `readMore()` re-entry, the one resume path with no timer:

- **Poll on a short backoff** (reuse the open-retry loop’s 20 ms) instead of re-entering `readMore()` synchronously — removes the busy-spin while a write is genuinely in progress.
- **Fail fast on a no-progress deadline** (`getBlobReadTimeout()`, default 20s) that is **reset on every byte of real progress** (`bytesRead > 0`), returning a retryable `BlobReadError` (503). A legitimate slow write that delivers any data within the window is unaffected; only a stalled/abandoned write (truncated file + stale-held lock) is terminated.

This is consistent with the bound already applied to the UNKNOWN-size wait in the sibling branch (#1423).

## Follow-up (separate)

- Re-rate #1423: this path is **not** "connection-isolated / no event-loop impact" — it starves the event loop and pegs CPU pool-wide.
- Consider extending #1424’s descriptor cross-check to the streaming (`isFullRead === false`) path so a truncated blob is rejected on serve, not just on full read.

Found via read-only source diff (v5.0.28–v5.1.8) + a CPU profile on the affected prod boxes; nothing modified.

kriszyp:
Fixes #1454.

## Summary

`FileBackedBlob.stream()` → `readMore()` busy-spun a worker thread at ~100% CPU, indefinitely, when a blob file is **present but truncated** (its 8-byte header records a known final size larger than the bytes actually on disk) **and** the write-lock reads as held. In the `bytesRead === 0` branch, `resumeIfWriterFinished()` re-entered `readMore()` synchronously with no backoff and no deadline, so a read at EOF returned 0 every iteration. `checkIfIsBeingWritten()` caches its result. The lock is in-process and released on `unlock()`/`DBHandle::close()`, so a *dead* writer can't leave it held — the pin is a **live** in-progress write whose source stream stalled and never reached `unlock()` (the `writeBlobWithStream` pipeline that #1443/#1444 watchdogs). This PR is the complementary reader-side bound.

This is the prod-dyn / prod-gar GDI CPU storm (harper-pro 5.0.28), confirmed by a 6-thread CPU profile (~73% in `readMore`: `fs.read` ~48% + `Buffer.allocUnsafe(256KB)` ~21%). The trigger is a wedged GDI blob-replication backlog (~11–12h behind) plus a 2h blob expiration: bodies evict/truncate before the backlog ships them.

## Why this isn't already fixed by #1423/#1424

The spin branch is byte-for-byte identical 5.0.28 → 5.1.8. #1423's `getBlobReadTimeout()` timer lives in the `else if (!resumeIfWriterFinished())` branch the spin never enters (resume returns `true`). #1424's descriptor cross-check is `isFullRead`-gated (HTTP serving streams a range → skipped) and only fires when header ≠ descriptor (the common truncation has header == descriptor).

## The fix

- A lazily-set no-progress deadline (`incompleteDeadline = getBlobReadTimeout()`, default 20s) + a 20ms backoff before the resume re-entry. Past the deadline, fail fast with a retryable `BlobReadError(503)`.
- The deadline is set **only while no bytes are readable**; a genuinely slow but progressing in-progress write (header written up front — the replication path) resolves the pull each chunk and starts each new pull with a fresh budget, so it is unaffected.
- Both `readMore()` 20ms backoff timers are now assigned to `timer` so `cancel()`/`onError` can clear a pending poll (the cross-model review caught that the new one wasn't tracked → a read on a closed/reused fd if a stream is aborted mid-wait).

## Where to look

- `resources/blob.ts` — the `resumeIfWriterFinished()` re-entry and the `incompleteDeadline` init in the `checkIfIsBeingWritten()` block.
- Test: `unitTests/resources/blob.test.js` (#1454) — present-but-truncated blob + held lock. Verified genuine: revert the source and it hangs (30s mocha timeout); with the fix it rejects 503 in ~165ms.

## Open items for the reviewer

The cross-model review surfaced **pre-existing** latent bugs in this file, in code this PR does not change — deliberately left out of this targeted patch and tracked separately in #1457: `fd` is not nulled after close (double-close on a reused fd), the `start()` open-retry timer is untracked, `onError` can fire twice if a slow async read completes after rejection, and slices read sequentially from byte 0.

Bound for a 5.1.x patch.

_Generated with assistance from Claude (Opus 4.8). Cross-model reviewed (Codex + Gemini + Harper-domain)._

### DIFF
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -461,8 +461,10 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
 									if (checkIfIsBeingWritten()) {
 										// Bound the wait for in-progress content the same way the open-retry loop bounds the
 										// wait for the file to appear (#1423): start a no-progress deadline on the first stuck
-										// read. checkIfIsBeingWritten() caches its result, so a writer that died or stalled
-										// without releasing its lock would otherwise pin us here forever (#1454).
+										// read. checkIfIsBeingWritten() caches its result, so an in-progress write whose source
+										// stream stalled — and so never reached its unlock() — keeps the lock held and would
+										// otherwise pin us here forever (the lock is in-process and released on unlock() or
+										// handle close, so this is a live stalled write, not a dead one) (#1454, cf #1444).
 										if (incompleteDeadline === 0) incompleteDeadline = Date.now() + getBlobReadTimeout();
 										// detects the race where the writer finished between our last async read
 										// and the watcher being set up (or the watcher missing the final write event)
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -382,6 +382,11 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
 			pull: (controller) => {
 				let size = 0;
 				let retries = 100;
+				// No-progress deadline for the incomplete-content wait below, mirroring the open-retry loop's
+				// deadline (#1423). Computed lazily on the first stuck read so a healthy read never pays for
+				// it; bounds the case where the header reports a known size but the bytes never arrive (a
+				// present-but-truncated blob whose writer lock stays held) (#1454).
+				let incompleteDeadline = 0;
 				return new Promise(function readMore(resolve: () => void, reject: (error: Error) => void) {
 					function onError(error) {
 						close(fd);
@@ -406,7 +411,7 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
 								if (retries-- > 0 && isBeingWritten !== false) {
 									checkIfIsBeingWritten();
 									logger.debug?.('File was empty, waiting for data to be written', filePath, retries);
-									setTimeout(() => readMore(resolve, reject), 20).unref();
+									timer = setTimeout(() => readMore(resolve, reject), 20).unref();
 								} else {
 									logger.debug?.('File was empty, throwing error', filePath, retries);
 									onError(
@@ -454,6 +459,11 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
 								size = Number(new DataView(buffer.buffer, buffer.byteOffset, 8).getBigUint64(0) & 0xffffffffffffn);
 								if (size > totalContentRead) {
 									if (checkIfIsBeingWritten()) {
+										// Bound the wait for in-progress content the same way the open-retry loop bounds the
+										// wait for the file to appear (#1423): start a no-progress deadline on the first stuck
+										// read. checkIfIsBeingWritten() caches its result, so a writer that died or stalled
+										// without releasing its lock would otherwise pin us here forever (#1454).
+										if (incompleteDeadline === 0) incompleteDeadline = Date.now() + getBlobReadTimeout();
 										// detects the race where the writer finished between our last async read
 										// and the watcher being set up (or the watcher missing the final write event)
 										const resumeIfWriterFinished = (): boolean => {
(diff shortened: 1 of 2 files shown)


heskew (approved):
👍 red ci seems unrelated

DavidCockerill (approved):

PR #1465, opened 2026-06-23
fix(components): close deploy-validation Scopes to stop deployLifecycle listener leak

### LINKED ISSUE #1462: DeployLifecycle listeners leak across deploys (MaxListenersExceededWarning) — Scope not removing deploy:start/deploy:end on reload
## Summary

A `MaxListenersExceededWarning` fires during component deploys: `11 deploy:start listeners added to [DeployLifecycle]` (and the same for `deploy:end`), exceeding the default limit of 10. This indicates `deploy:start`/`deploy:end` listeners accumulate on the shared `deployLifecycle` emitter faster than they are removed — a Scope-lifecycle leak that grows with the number of deploys/component reloads.

Observed live during a `deploy_component` on a 4-node 5.1.9 cluster.

## Where

- `components/Scope.ts:138-139` — each `Scope` registers `deployLifecycle.on('deploy:start', ...)` / `on('deploy:end', ...)`.
- `unitTests/components/Scope.test.js:552` already asserts the intended invariant: *"scope.close() should remove its deploy:start listener from the module emitter."*

So the design contract is: every Scope that adds these listeners must remove them on `close()`. The warning means either (a) a Scope is created per deploy/component-reload and not all of them `close()` (listeners leak across deploys), or (b) some path adds the listeners without the matching removal.

## Impact

Low severity — it's a warning, not a failure, and 11 vs 10 is just over the threshold. But it's a genuine listener/Scope leak that grows monotonically with deploy count, so over a long-lived process with many deploys it will keep climbing. Worth confirming the close path.

## Notes / relationships

- **Distinct from** harper-pro#460 / #461 (the replicator's node-update watcher accumulation — different subsystem; the replicator does not add `deploy:*` listeners) and from #454 (copy-progress watchdog).
- Related to the **component-reload mechanism**: each deploy reloads components → new Scopes. The "`startOnMainThread` once-per-component" core fix (in flight, root cause of harper-pro#460) may reduce Scope churn but does not itself close this gap.

## Needs

Root-cause confirmation: which Scope (per-component? per-application? root?) is created on reload without a matching `close()`, or which path registers the listeners without removal. Then either ensure `close()` runs on the leaked Scopes or dedupe registration.

kriszyp:
Fixes #1462

## Summary

`deploy_component`'s pre-flight validation load (`components/operations.js`, `deployComponent`) loads the freshly-prepared component on a worker — with `pseudoResources.isWorker = true` — purely to surface load-time errors early. Because `isWorker` is set, `loadComponent` creates real `Scope`s, each of which registers `deploy:start` / `deploy:end` listeners on the shared `deployLifecycle` emitter (`components/Scope.ts:138-139`). These throwaway validation Scopes were never closed — `Scope.close()` only ran via the per-Scope worker-`SHUTDOWN` handler — so their listeners accumulated on the worker across deploys, eventually tripping `MaxListenersExceededWarning` (`11 deploy:start listeners`). This is the leak in #1462.

The invariant the rest of the system relies on is stated in `components/scopeShutdown.ts`: *"scopes are created once at load and closed once at shutdown, so nothing accumulates."* The validation load violated it.

## Changes

- **`components/componentLoader.ts`** — added a `collectScopes?: Set<Scope>` option to `loadComponent`. When provided, each Scope created during the load is added to the set **instead of** being registered for worker-shutdown auto-close; the caller then owns closing them. Threaded through the sub-component recursion so packaged sub-components are collected too.
- **`components/operations.js`** (`deployComponent`) — the validation load now collects its Scopes and closes them in a `finally` once validation completes, so their `deployLifecycle` listeners are removed immediately. The load+close is wrapped in `trackScopeClose(...)` so a concurrent worker shutdown still waits for these Scopes to dispose (a plugin may start a native runtime in `handleApplication`) before `realExit`. Also dropped the stale trailing positional args to `loadComponent` (the 4th+ args were already ignored by the current `(dir, resources, origin, options)` signature).
- **`unitTests/components/componentLoader.test.js`** — a characterization test (each unclosed validation load leaks exactly one `deploy:start` listener) and a regression test (with `collectScopes` closed after each load, the listener count stays at baseline across 12 loads — past the default `maxListeners` of 10).

## Where to look

- **Shutdown safety (operations.js).** Skipping the per-Scope `SHUTDOWN` auto-close for collected Scopes would otherwise drop them from the worker's `whenScopesClosed()` wait. The `trackScopeClose(validation)` wrapping restores that guarantee — worth a look that this is the right mechanism. (Raised by the Codex cross-model review and addressed here.)
- This is **distinct** from harper-pro#460/#461 (replicator node-update watcher accumulation) and from the in-flight `startOnMainThread`-once core fix (branch `kris/start-on-main-thread-once`). No overlap: that fix gates main-thread init; this fixes worker-side validation-load Scope cleanup.

Local validation: `unitTests/components/componentLoader.test.js` and `unitTests/components/Scope.test.js` pass; the deploy integration suite is left to CI.

🤖 Generated by an LLM (Claude Opus 4.8).

### DIFF
--- a/components/operations.js
+++ b/components/operations.js
@@ -536,7 +536,10 @@ async function deployComponent(req) {
 						collectScopes: validationScopes,
 					});
 				} finally {
-					await Promise.allSettled(Array.from(validationScopes, (scope) => scope.close()));
+					const closeResults = await Promise.allSettled(Array.from(validationScopes, (scope) => scope.close()));
+					for (const result of closeResults) {
+						if (result.status === 'rejected') log.warn('Failed to close a deploy-validation Scope', result.reason);
+					}
 				}
 			})();
 			// Track the load+close so a concurrent worker shutdown waits for these scopes to finish
--- a/components/operations.js
+++ b/components/operations.js
@@ -520,18 +520,29 @@ async function deployComponent(req) {
 			pseudoResources.isWorker = true;
 
 			const componentLoader = require('./componentLoader.ts').default || require('./componentLoader.ts');
+			const { trackScopeClose } = require('./scopeShutdown.ts');
 			let lastError;
 			componentLoader.setErrorReporter((error) => (lastError = error));
 			emit('phase', { phase: 'load', status: 'start' });
-			await componentLoader.loadComponent(
-				application.dirPath,
-				pseudoResources,
-				undefined,
-				false,
-				undefined,
-				false,
-				req.project
-			);
+			// This load exists only to surface load-time errors early; the Scopes it creates are
+			// throwaway. They are collected (instead of registered for worker-shutdown auto-close) so we
+			// can close them here once validation completes — otherwise each deploy leaks the Scope's
+			// deploy-lifecycle listeners on this worker, eventually tripping MaxListenersExceededWarning
+			// (#1462).
+			const validationScopes = new Set();
+			const validation = (async () => {
+				try {
+					await componentLoader.loadComponent(application.dirPath, pseudoResources, undefined, {
+						collectScopes: validationScopes,
+					});
+				} finally {
+					await Promise.allSettled(Array.from(validationScopes, (scope) => scope.close()));
+				}
+			})();
+			// Track the load+close so a concurrent worker shutdown waits for these scopes to finish
+			// disposing — a plugin may start a native runtime in handleApplication — before realExit.
+			trackScopeClose(validation);
+			await validation;
 			emit('phase', { phase: 'load', status: 'done' });
 
 			if (lastError) throw lastError;
--- a/components/componentLoader.ts
+++ b/components/componentLoader.ts
@@ -309,6 +309,11 @@ export interface LoadComponentOptions {
 	autoReload?: boolean;
 	providedLoadedComponents?: Map<any, any>;
 	appName?: string;
+	// When provided, every Scope created during this load is added to this set instead of being
+	// auto-closed on worker shutdown. The caller then owns closing them. Used by transient loads
+	// (e.g. the deploy pre-flight validation) so their deploy-lifecycle listeners don't accumulate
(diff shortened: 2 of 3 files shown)


components/operations.js
```diff
@@ -475,18 +475,29 @@ async function deployComponent(req) {
...
+			// deploy-lifecycle listeners on this worker, eventually tripping MaxListenersExceededWarning
+			// (#1462).
+			const validationScopes = new Set();
+			const validation = (async () => {
+				try {
+					await componentLoader.loadComponent(application.dirPath, pseudoResources, undefined, {
+						collectScopes: validationScopes,
+					});
+				} finally {
+					await Promise.allSettled(Array.from(validationScopes, (scope) => scope.close()));
```
cb1kenobi: **Low: validation-scope close() errors are silently swallowed**

`Promise.allSettled(...)` discards any rejection from `scope.close()`. The normal worker-shutdown path funnels close through `trackScopeClose`, which logs a rejected close via `harperLogger.error(...)` and never swallows it silently. Here, if a validation-load scope started a native runtime in `handleApplication` (the exact case the tracking comment calls out) and its async disposal fails, the error is invisible. Listener removal still happens (it runs synchronously at the top of `close()` before any await), so the leak fix itself is safe — this is only an observability gap.

Suggested fix:
```suggestion
					await Promise.allSettled(
						Array.from(validationScopes, (scope) =>
							scope.close().catch((error) => log.error('Error closing deploy-validation scope', error))
						)
					);
```

—
_Generated by Barber AI_

kriszyp: Captured the Promise.allSettled results and log any rejected close() reasons via log.warn, without throwing — allSettled still runs all closes to completion. — KrAIs

cb1kenobi (approved):

PR #1466, opened 2026-06-24
Fix deploy packaging an empty component when installed under a node_modules path

kriszyp:
## Summary
`harper deploy` shipped an **empty** component (only deps reinstalled node-side) when packaging any component located under a `node_modules/` directory — e.g. a globally npm-installed component. `packageComponent`'s `tar-fs` `ignore` tested `name.includes('node_modules')` on the **absolute** path, so every entry of a component living under `…/node_modules/<pkg>` matched and the whole tree was excluded (1.5 KiB tarball). The size estimator (`getPackagedDirectorySize`) already used a relative segment check, so the packer and size-walk had diverged.

Fix: a shared `isExcluded()` that resolves paths **relative to the pack root** and matches `node_modules` as a path *segment* (plus the `cache/webpack` path), used by both `streamPackagedDirectory` and `getPackagedDirectorySize` so they can't diverge again. Regression test added.

## Why
Surfaced while dogfooding a Fabric deploy of a globally-installed component (it deployed empty; see tpsdev-ai/flair#513). Affects any npm-installed component deploy, not just that case.

## Where to look
- `components/packageComponent.ts` — the `isExcluded` helper and the two call sites it unifies.
- Edge cases (verified by reviewers): `fullPath === directory` → empty rel, not excluded; Windows `sep`; nested `node_modules` still excluded; substring false-positives like `some_node_modules_pkg` no longer match.

No open review concerns — Codex and Gemini both reviewed clean (one perf suggestion — hoisting the `cache/webpack` join — applied).

🤖 Generated by Claude (Opus 4.8) via Claude Code.

### DIFF
--- a/components/packageComponent.ts
+++ b/components/packageComponent.ts
@@ -1,4 +1,4 @@
-import { join } from 'path';
+import { join, relative, sep } from 'node:path';
 import { stat, readdir } from 'node:fs/promises';
 import { Readable } from 'node:stream';
 import tar from 'tar-fs';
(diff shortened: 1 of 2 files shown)


heskew (approved):
lgtm

PR #1477, opened 2026-06-24
docs: clarify that tables/databases are live process-wide singletons

dawsontoth:
## Why

Harper's data-access globals are easy to misread from the source, and that misreading has real downstream cost (e.g. SSR code fetching its own REST API over loopback instead of just calling `tables.X.get()`):

- A reader who finds the VM `Compartment` machinery in `security/jsLoader.ts` can conclude `tables`/`databases` are **compartment-only**.
- A reader who sees `node_modules/harper` can assume `import 'harper'` resolves a **separate, empty** npm copy.

Both are wrong. `globals.js` assigns each name to **both** `global[name]` and `exports[name]` (via `_assignPackageExport`), the component loader **symlinks** `node_modules/harper` to the running install, and compartments are **seeded** from the process `global`. But none of that code carried an explanatory comment, so the correct mental model isn't discoverable.

## What (comments only — no behavior change)

- **`index.ts`** — note that the names are exposed as ambient globals *and* as `harper` package exports that resolve to the **same live value**, available in any module Harper loads (including a Vite SSR entry whose `node_modules/harper` is symlinked here).
- **`components/componentLoader.ts`** — JSDoc on `symlinkHarperModule` explaining it exists so `import 'harper'` in component/bundler code resolves to the running install (live exports), not a separate copy.
- **`security/jsLoader.ts`** — note that `getGlobalObject` **seeds** compartments from the process `global` (live references) and `getHarperExports` returns those same singletons; compartments don't get an alternate set.

### DIFF
--- a/components/componentLoader.ts
+++ b/components/componentLoader.ts
@@ -161,6 +161,17 @@ export function setErrorReporter(reporter) {
 let compName: string;
 export const getComponentName = () => compName;
 
+/**
+ * Symlink a component's `node_modules/harper` (and `harperdb`) to this running Harper install,
+ * rather than letting it resolve a separately-installed npm copy.
+ *
+ * This is what makes `import { tables } from 'harper'` (or `import 'harperdb'`) inside component
+ * code — including bundler-loaded code such as a Vite SSR entry — resolve to the SAME module
+ * instance Harper is running. That instance carries the live, process-wide exports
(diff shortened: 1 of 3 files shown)


kriszyp (approved):

kriszyp (approved):
Clear, correct clarification that tables/databases are live process-wide singletons. Approving. — Claude (Opus 4.8)

PR #405, opened 2026-04-24
fix: Add schemas files to npm package

dawsontoth:
… and update the lock file

PR #1480, opened 2026-06-25
fix(blob): mark a retriable source-stream write-abort PENDING (503) not incomplete (500) (harper-pro#481)

kriszyp:
## Problem (harper-pro#481)

A blob write fed by a **re-streamable replication/origin source** that aborts mid-stream (the source stalled — idle watchdog — or the connection dropped) left a short file with a normal header. A downstream read then classified it `BlobReadError('Blob is incomplete', 500)`. The replication receiver treats **500 as permanent**: it advances its resume cursor **past** the record and, with proactive blob backfill (harper-pro#388) unimplemented, the blob is **silently lost**.

In a catch-up (two nodes converging from a complete peer), both sides serve 500 for blobs the other is still pulling, and the cluster **stops converging even after writes stop** — the field symptom on the Akamai v4→v5 stage cluster (5.1.11).

This is the flip side of harper-pro#429: the *same* `'Blob is incomplete'` error wants the *opposite* response depending on whether the bytes are still coming. The on-disk file can't tell them apart (both are just short), so a blanket reclassify would reopen #429's connection wedge.

## Fix

The **write-abort path still holds the abort reason** and knows whether the source was a re-streamable replication/origin feed: the owning caller armed the source-idle watchdog (`idleTimeoutMs > 0`) — which only the replication receive path does, never an app one-shot stream — and the failure is not a source-reported permanent absence (`sourceBlobUnavailable`, which the receive loop already advances past + unlinks).

Such a file is stamped with a new `PENDING_TYPE` (0xfe) header. The read paths return **503** (transient → the receiver holds the gap and re-streams on reconnect) instead of 500. The re-stream overwrites the stub; a terminal give-up unlinks it (→ 404). This resolves the #481/#429 tension without reopening the #429 wedge.

- `bytes()` / `stream()` return `BLOB_UNAVAILABLE_STATUS` (503) for a PENDING stub
- `isBlobFileComplete` returns false (the #388 repair sweep still treats it as incomplete)
- the inline replication-encode path never inlines a PENDING stub as complete (its header size can coincide with the body length) — stays uncoded so the sender holds and retries rather than propagating the stub
- the marker is written **under the still-held blob lock** (unlock moved into the `writeFile` callback, unconditional) so no concurrent read/send observes the bare partial file and classifies 500 in the gap
- gated on the **armed idle watchdog**, not `storageInfo.source`, so an app `put`-from-stream abort is never mis-marked PENDING (which would read 503 forever)
- header built directly (not via `createHeader`) so the compress-type OR can't collide with the PENDING type bits

**Receiver side needs no change** — 503 already routes to hold-and-retry via `isPermanentSourceBlobErrorCode` (only 404/500 are permanent).

## Tests

`unitTests/resources/blob.test.js` (52 passing): PENDING reads → 503 (`bytes()`/`stream()`), `isBlobComplete` false, replication-encode holds, the real armed-source abort → PENDING via `writeBlobWithStream`, and a negative guard (unarmed app-stream abort stays non-503).

## Cross-model review

Codex flagged two issues, both fixed + regression-tested before this PR:
1. **Marker-write race** — was fire-and-forget after `unlock`; now the lock is held until the marker is durable.
2. **Over-broad gate** — was `storageInfo.source != null` (caught app `put`-from-stream → permanent 503 wedge); now `idleTimeoutMs > 0`.

(Gemini leg was unavailable — `agy` timed out.)

## Release

Targets `main`; the BlobReadError statusCode taxonomy it builds on is already on main. Not in the in-flight 5.1.13 — queue for **5.1.14**. Follow-up: harper-pro#388 (repair-sweep peer backfill) as the durable convergence fix; the `findIncompleteBlobRefs` machinery this relies on already exists.

### DIFF
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -70,6 +70,12 @@ const HEADER_SIZE = 8;
 const UNCOMPRESSED_TYPE = 0;
 const DEFLATE_TYPE = 1;
 const ERROR_TYPE = 0xff;
+// A write that aborted on a re-streamable external source (replication receive / origin fetch) stamps the
+// file with this type so a downstream read returns 503 (retry) rather than 500 (confidently incomplete).
+// The bytes are still expected — the receive side holds a blob gap and re-streams on reconnect, which
+// overwrites this stub; a terminal give-up unlinks the file (→ 404). Distinct from ERROR_TYPE (a permanent
+// corrupt/error stub, replicated as-is). See harper-pro#481.
+const PENDING_TYPE = 0xfe;
 const DEFAULT_HEADER = new Uint8Array([0, UNCOMPRESSED_TYPE, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff]);
 const COMPRESS_HEADER = new Uint8Array([0, DEFLATE_TYPE, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff]);
 const UNKNOWN_SIZE = 0xffffffffffff;
@@ -212,6 +218,10 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
 					const headerValue = new DataView(rawBytes.buffer, rawBytes.byteOffset, 8).getBigUint64(0);
 					if (Number(headerValue >> 48n) === ERROR_TYPE) {
 						throw new BlobReadError('Error in blob: ' + rawBytes.subarray(HEADER_SIZE), BLOB_CORRUPT_STATUS);
+					}
+					if (Number(headerValue >> 48n) === PENDING_TYPE) {
+						// half-replicated blob whose source stream aborted; the bytes are still expected — retry (harper-pro#481)
+						throw new BlobReadError('Blob pending replication for ' + filePath, BLOB_UNAVAILABLE_STATUS);
 					}
 
 					size = Number(headerValue & 0xffffffffffffn);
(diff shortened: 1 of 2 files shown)


kriszyp:
Verified this against the harper-pro#481 no-peer-fallback repro (qa-explorer, core `79672cdac` = this PR's tip + harper-pro `d1f44a9e`). **This PR is correct + valuable, but it does not close the repro's failure mode** — worth flagging so the gap doesn't get marked resolved.

The new `PENDING_TYPE` (0xfe) / 503 classification activates only in `writeBlobWithStream`'s error handler, gated on `idleTimeoutMs > 0` (the replication receive worker path) and a non-`sourceBlobUnavailable` error. So it covers a **transient replication receive-side write-abort** (idle watchdog fires / connection drops mid-stream) → the partial file is stamped PENDING and downstream reads return 503 (hold/retry) instead of 500. That's the right fix for that case.

It does **not** cover a **post-write truncated-on-disk blob** (the field/relay scenario the #481 repro models): a fully-written file later truncated has a normal size header (declared 49160 bytes, 1024 on disk), so the sender hits `blob.ts:565` → `BlobReadError('Blob is incomplete', 500)` — the `PENDING_TYPE` header check never fires. The receiver still classifies 500 as permanent and advance-past's (30× in the repro). Log: `Error sending blob BlobReadError: Blob is incomplete / statusCode: 500`.

| Failure mode | #1480 |
|---|---|
| Replication receive write-abort (idle/drop) | ✅ → 503 PENDING |
| Post-write disk truncation (relay holds truncated data) | ❌ → still 500 advance-past |
| ENOENT / missing file | ❌ (separate path, #405) |

So the truncated-file / no-peer-fallback wedge (records permanently missing on a node the cluster collectively has) still needs **receiver-side peer-retry on 404/500** (option (c) in #481) or **#388 backfill**. Repro: `harper-pro integrationTests/cluster/qa-scratch/qa354-blob-no-peer-fallback.test.mjs`.

PR #1488, opened 2026-06-25
test(database): promote 4 regression anchors (bulk-conditional / delete-update-race / auto-fields / date-queries)

kriszyp:
## What

Promotes 4 exploratory QA tests (from the gitignored scratch suite) into `integrationTests/database/` as permanent regression anchors. All pass on both engines.

| Test | Pins |
|---|---|
| `bulk-conditional-mutation.test.ts` | SQL bulk `UPDATE/DELETE ... WHERE` changes exactly the matching rows (vs an oracle); an `@indexed`-column UPDATE **rewrites the index cleanly** (no stale entries); a constraint violation mid-batch → **atomic rollback** (0 rows changed); concurrency-safe (4 workers). RocksDB 8/8, LMDB 8/8. |
| `delete-update-race-consistency.test.ts` | Concurrent DELETE vs PATCH/PUT/addTo/recreate: the invariant is **index/scan/point-read consistency** after the race — no tombstone/phantom-index anomaly, delete is idempotent, no crash. (Asserts only consistency; the recreate-vs-delete winner is engine-divergent and left observational.) RocksDB 7/7, LMDB 7/7. |
| `auto-fields.test.ts` | `@createdTime`/`@updatedTime` auto-assign + `@createdTime` preserved on PATCH/PUT; auto-UUID PK on id-omit; required-field 400; `@sealed` interaction; REST/ops/SQL parity. 10/10. |
| `date-queries.test.ts` | Date storage/round-trip, REST FIQL range correctness vs an oracle (inclusive boundaries, indexed==unindexed parity), chronological sort, timezone normalization, edge cases. RocksDB 22/22, LMDB 22/22. |

## Notes

- These are characterization/regression anchors — they pin *correct* behavior so a future change that regresses (e.g. a stale index after a conditional UPDATE, a tombstone left in a secondary index after a delete race, or an overwritten `@createdTime`) is caught.
- `date-queries.test.ts` includes one probe of the SQL date-literal filter that is currently silent-wrong (tracked under #1397); it **records** the behavior without asserting the buggy result, so it does not enshrine the defect (and flips to a correct assertion once #1397 lands).
- First of several promotion batches from the qa-explorer campaign; more anchors to follow.

— KrAIs (Claude) on Kris's behalf

PR #1490, opened 2026-06-25
test(database): promote 4 more regression anchors (patch-merge / hot-snapshot / multi-condition / append-durability)

kriszyp:
## What

Second batch of regression anchors promoted from the qa-explorer scratch suite into `integrationTests/database/` (follows #1488). All pass on both engines.

| Test | Pins |
|---|---|
| `patch-disjoint-field-merge.test.ts` | Many clients PATCH **different** fields of the same record concurrently → **all disjoint field updates survive** (0 lost updates); same-field is clean LWW; PATCH merges vs PUT replaces. (Commit-time re-read + delta merge.) 7/7 both engines. |
| `hot-record-read-snapshot.test.ts` | A singleton config record read at high concurrency while a writer atomically replaces all fields → **0 torn reads** (every read sees a self-consistent snapshot/checksum), monotonic version visibility, no latency cliff, `addTo` exact under the write storm. 5/5 both engines. |
| `multi-condition-query.test.ts` | AND / OR / range / mixed-indexed+CONTAINS / 3-way / empty multi-condition queries return **exactly the oracle's result set** across ops `search_by_conditions` and SQL. 7/7 both engines. |
| `append-only-unique-pk-durability.test.ts` | Pure append-only inserts with unique PKs land **100% durable** on both engines at 1- and 4-worker concurrency; a colliding-PK control correctly collapses to the pool size (LWW). Guards against append data-loss. 4/4 both engines. |

## Notes

- All four pin **correct** behavior — no known-bad cells, no engine-divergent assertions; every assertion is the invariant (no lost update, no torn read, right result set, durable append).
- Internal campaign IDs removed; console labels are filename-based.
- Continues the incremental promotion from the exploratory campaign (more batches to follow).

### DIFF
--- a/integrationTests/database/append-only-unique-pk-durability.test.ts
+++ b/integrationTests/database/append-only-unique-pk-durability.test.ts
@@ -1,0 +1,229 @@
+/**
+ * Pure-append unique-PK durability on both storage engines.
+ *
+ * Regression anchor for append data-loss: every insert with a guaranteed-unique
+ * PK (crypto.randomUUID() generated server-side) must be durable after concurrent
+ * writes settle. Tests both 1-worker and 4-worker configurations.
+ *
+ * Hard gate (test A): unique-PK inserts — durable count must equal N_INSERTS (120).
+ *   Any shortfall is a real insert data-loss defect.
+ *
+ * Control (test B): colliding-PK inserts — N_INSERTS all map into a 5-slot pool.
+ *   Expect last-write-wins shortfall: durable ≤ pool size (5). Informational only.
+ *
+ * Both RocksDB (default) and LMDB engines are tested via HARPER_STORAGE_ENGINE=lmdb.
+ */
+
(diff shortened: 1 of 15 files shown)


cb1kenobi (approved):

Ethan-Arrowood:
These look good to me — the anchors all pin real invariants (no lost updates, no torn reads, oracle-parity result sets, durable appends), the `win32` skips are sensible, and the integration matrix is green.

This is ready to go as soon as the lint and format checks are cleared — both are failing on deterministic, purely-mechanical issues in the new files:

- **runLinter** (6 errors): unused `strictEqual` import + `query`/`body` params that need a `_` prefix — `patch-disjoint-field-merge.test.ts:21,28` and the other affected file at `:43,:56`.
- **Format Check**: the four new test files need a `prettier --write` pass.

A quick `npm run format` plus the unused-var fixes should clear the board. Happy to see it merge once it's green.

PR #1491, opened 2026-06-25
feat(audit): table-reload marker type for copyApply system-DB coverage

kriszyp:
## Summary
Adds a `reload` audit entry type — a `LOCAL_ONLY`, record-less control entry that signals a whole table was bulk-reloaded so its subscribers re-read it.

- `auditStore.ts`: new `RELOAD = 8` type; the entry type is now decoded from the **low nibble** of the action byte (`action & 0xf`) instead of `action & 7`, widening the type space from 7 to 15. Identical to the old mask for every existing entry (bit 3 was always clear); flag bits (`HAS_RECORD`, …) sit above the nibble.
- `Table.writeReloadMarker()`: emits one marker in its own transaction (id `null`, no record → audit-only; `localOnly: true`).
- `transactionBroadcast.ts`: on a `reload` entry, notifies **every** subscriber on the table (no recordId hierarchy walk).
- `Table.subscribe` listener: passes `reload` through without the per-record lookup.

## Purpose
Lets harper-pro extend copyApply to the system DB (harper-pro#489). copyApply's audit-less snapshot rows otherwise suppress the `hdb_nodes` (peer discovery) and `hdb_certificate` (CA install) events that drive cluster formation; a single reload marker re-drives those subscribers after the copy.

## Where to look
- **`action & 0xf` widening** (`auditStore.ts` readAuditEntry) — the one hot-path decode change. `readAuditEntry` is the sole action-byte decoder; harper-pro reuses it.
- **`writeReloadMarker`** binds its standalone transaction to the table's db via `txnForContext` (a bare `transaction()` has no db and faults in `save()`).

## Tests
`unitTests/resources/reloadMarker.test.js` — delivery to multiple subscribers + the persisted entry is `LOCAL_ONLY` with a null recordId. Existing copyApply/auditLog/crud/recordEncoder suites pass unchanged.

### DIFF
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -3209,7 +3209,9 @@ export function makeTable(options) {
 							// we only send the full message, this are individual messages that can be sent out of order
 							// TODO: Do we want to have a limit to how far out-of-order we are willing to send?
 							value = auditRecord.getValue?.(primaryStore, getFullRecord);
-						} else if (type !== 'end_txn') {
+						} else if (type !== 'end_txn' && type !== 'reload') {
+							// 'reload' is a whole-table marker with a null id and no record (harper-pro#489); pass it
+							// through verbatim so subscribers re-read the table, skipping the per-record lookup below.
 							// these are events that indicate that the primary record has changed. I believe we always want to simply
 							// send the latest value. Note that it is fine to synchronously access these records, they should have just
 							// been written, so are fresh in memory.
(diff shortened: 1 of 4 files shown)


kriszyp:
Companion harper-pro PR: HarperFast/harper-pro#491 (bumps the `core` submodule to this). Addresses harper-pro#489.

cb1kenobi (approved):
LGTM!

PR #1496, opened 2026-06-25
feat(deploy): auto-drop large deployment payload blobs after a successful deploy

### LINKED ISSUE #1495: Auto-delete large deployment payload blobs after deploy; storage reclamation for older payloads
## Background

Every `deploy_component` call stores the uploaded tarball as a Blob on the `hdb_deployment` record. That blob is replicated to all peers so they can run the local install without a separate upload channel. After replication completes the blob is retained indefinitely — there's no automatic cleanup.

For apps with large payloads (multiple deploys per day, multi-peer clusters) this accumulates fast: each 16 MB payload becomes 16 MB × N peers, and orphaned blobs from superseded deployments are only removed by an explicit `delete_deployment_payload` call or a full orphan scan. In practice, clusters accumulate hundreds of orphaned payload blobs totaling several GB.

### 1. Threshold-based auto-delete after deploy completes

Once a deploy reaches terminal status (`success` or `failed`) and all peers have confirmed receipt, automatically delete the `payload_blob` from the deployment record if the payload exceeds **10 MB** (configurable). The deployment metadata (status, event_log, timestamps, deployment_id, payload_size) is retained; only the blob bytes are dropped.

- Trigger point: after the replicate phase settles (all `onPeerResult` callbacks have fired, or the `response.replicated` fallback path completes)
- Guard: skip deletion if the blob was never stored (small deploys below the threshold keep their blob for audit convenience)
- The existing `delete_deployment_payload` operation already implements the deletion; the new code just calls it automatically at the right moment

### 2. Storage reclamation hook under disk pressure

Add a reclamation callback to the `hdb_deployment` table that Harper can invoke when available disk space falls below a threshold. The callback should evict `payload_blob` from the oldest deployment records first (by deploy timestamp), stopping once enough space has been reclaimed or all blobs are gone. This mirrors the pattern used for audit-log pruning.

- Harper already has hooks for storage pressure in the blob layer; the deployment table should register with that mechanism
- Per-node reclamation is sufficient — no cross-cluster coordination needed, since each peer holds its own copy

## Implementation notes

- Deploy logic: `components/operations.js` — the terminal write happens after `emit('phase', { phase: 'replicate', status: 'done' })` at ~line 526; that's the right place to trigger auto-delete
- Blob deletion primitive: `resources/blob.ts` `deleteBlob()` / the blob attribute `.delete()` path already used by `delete_deployment_payload`
- Size is already recorded as `payload_size` on the row — no need to re-stat the blob to apply the threshold

## Acceptance

- Deploys ≥ 10 MB: `payload_blob` is null on the deployment record after the deploy settles; `payload_size` and all metadata remain
- Deploys < 10 MB: blob is retained (small enough that storage cost is negligible, audit value is higher)
- Under disk pressure: reclamation hook evicts oldest payload blobs first; system remains operational
- `delete_deployment_payload` remains available for manual/programmatic cleanup regardless of threshold

kriszyp:
Closes #1495 (part 1 of 2).

## Summary
After a deploy succeeds and every peer has installed from the replicated `payload_blob`, the upload tarball is no longer needed. For payloads larger than `deployment_payloadRetention_maxSize` (new config param, default **10 MiB**) the origin drops `payload_blob` as part of the `DeploymentRecorder`'s single terminal write. Because that null reference replicates, the blob is reclaimed **cluster-wide** — each node unlinks its local file via the existing `RecordEncoder` retained-blob check (the #641 path). Metadata (`payload_size`, `payload_hash`, `event_log`, status) is retained for the audit trail, and a `payload_dropped` event is recorded.

## Purpose
We never actually reclaimed deploy payload blobs after replication — they accumulated on every node (observed: hundreds of orphaned payloads, multiple GB on a preprod cluster). This adds the threshold-based reclamation we'd intended.

## Where to look / things to weigh
- **`components/operations.js`** (success branch of `deployComponent`): the drop is gated on `payload_size > threshold` **and** `getFailedPeers().length === 0`, so a deploy that only reached "success" because `ignore_replication_errors` masked a failed peer keeps its payload for inspection/retry.
- **Cluster-wide reclamation via replication is intentional** — one origin decision nulls the blob everywhere. For the post-deploy case this is desired (the payload's job is done); calling it out because it's the load-bearing semantic.
- **Threshold default 10 MiB**, configurable; set very high to retain all payloads. Explicit `0` means "always drop"; non-numeric/blank/unset falls back to the default.
- `DeploymentRecorder.dropPayload()` only mutates the in-memory record so the null folds into `finish()`'s terminal write (avoids an extra put and the #1170 out-of-order-revert window).

## Open follow-up (Part 2)
Issue #1495 also asks for a **disk-pressure storage-reclamation hook** (evict oldest payloads when space is low). The framework already exists (`server/storageReclamation.ts` `onStorageReclamation`); the eviction policy and the "one node's disk pressure reclaims audit blobs cluster-wide" semantics deserve their own PR + decision.

Gemini review also flagged a related gap that Part 2 should cover: a deploy where a peer was **offline/failed** keeps its payload (correct — that peer may still need it), but there is **no later trigger** to reclaim it once the cluster converges. The conservative keep-on-failure here is deliberate for Part 1; eventual/retroactive reclamation belongs in the Part 2 hook.

## Tests
- Unit: `DeploymentRecorder.dropPayload` (drop / no-op / after-finish / metadata-retention / sealed-terminal-write).
- Integration: `deploy-payload-reclaim.test.ts` forces the threshold to 1 byte and asserts `payload_blob_present === false` with metadata + `payload_dropped` event retained after a successful deploy (polls `get_deployment` to a terminal status — no fixed sleep). Retain-side (default threshold keeps small payloads) is covered by the existing `deploy-tracking.test.ts`.

Cross-model reviewed (Codex + Gemini). All actionable findings fixed: ignore-errors gate, blank/coerced-config zero-threshold, and the integration-test sleep flake. Declined with rationale: `record`/`row` rename (pre-existing class convention) and human-readable byte units (raw bytes matches the sibling `operationsApi_componentFile_maxSize`).

🤖 Generated by Claude Opus 4.8.

### DIFF
--- a/components/deploymentRecorder.ts
+++ b/components/deploymentRecorder.ts
@@ -297,6 +297,23 @@ export class DeploymentRecorder {
 		await this.put();
 	}
 
+	/**
+	 * Drop the stored payload tarball, retaining the metadata (size, hash, event_log). Used to
+	 * reclaim storage for large deploys once every peer has installed from the blob and it is no
+	 * longer needed. Only mutates the in-memory record — the caller drops the payload before the
+	 * terminal finish() write so the null reference lands in that single put rather than a
+	 * separate one. Committing the row with no blob reference unlinks the file locally (via the
+	 * RecordEncoder retained-blob check) and replicates the null so peers drop their copies too.
+	 *
+	 * Returns the freed payload size (bytes) when a blob was present and dropped, otherwise 0.
+	 */
+	dropPayload(): number {
+		if (this.finished || this.record.payload_blob == null) return 0;
+		const freed = typeof this.record.payload_size === 'number' ? this.record.payload_size : 0;
+		this.record.payload_blob = null;
+		return freed;
+	}
+
 	async transitionPhase(phase: string, status?: DeploymentStatus): Promise<void> {
 		this.record.phase = phase;
 		if (status) this.record.status = status;
--- a/components/operations.js
+++ b/components/operations.js
@@ -572,6 +572,20 @@ async function deployComponent(req) {
 
 		if (recorder) {
 			response.deployment_id = recorder.deploymentId;
+			// Reclaim the payload tarball for large deploys: every peer has now installed from
+			// the blob (replicateOperation resolved) and the origin no longer needs it. Dropping
+			// the reference before finish() folds the null into the single terminal write, which
+			// unlinks the file locally and replicates the null so peers drop their copies too.
+			// Metadata (size, hash, event_log) is retained for the audit trail. Two guards keep
+			// the tarball when it's still the artifact you'd debug or retry with: failed deploys
+			// don't reach this branch, and a deploy that reached here only because
+			// ignore_replication_errors masked failed peers keeps its payload for those peers.
+			const payloadSize = recorder.row.payload_size;
+			const retentionMaxSize = getPayloadRetentionMaxSize();
+			if (typeof payloadSize === 'number' && payloadSize > retentionMaxSize && recorder.getFailedPeers().length === 0) {
+				const freed = recorder.dropPayload();
+				if (freed > 0) emit('payload_dropped', { payload_size: freed, max_size: retentionMaxSize });
+			}
 			emit('phase', { phase: 'success', status: 'done' });
 			await recorder.finish('success');
 		}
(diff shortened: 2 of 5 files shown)


Ethan-Arrowood (approved):
LGTM — this is a clean, well-reasoned fix for the payload accumulation problem.

What I checked and liked:

- `dropPayload()` is tight: no-ops if finished or already blob-less, returns the freed size, and only mutates the in-memory record so the null folds into `finish()`'s single terminal write — nicely avoids an extra put and the #1170 out-of-order-revert window.
- The two retention guards are exactly the right call. Failed deploys never reach the branch, and gating on `getFailedPeers().length === 0` means a deploy that only "succeeded" because `ignore_replication_errors` masked a down peer keeps its payload for retry/inspection.
- `getPayloadRetentionMaxSize()` coercion is thorough — boolean/array/unset/blank fall back to the default, numeric strings are honored, explicit `0` means always-drop. Good defensive handling.
- Great test coverage: the unit cases cover drop / no-op / unknown-size / post-finish / sealed-terminal-write, and the integration test forces the threshold to `'1'` as a string (exercising the coercion path) and polls to terminal status instead of sleeping.

I'm on board with the cluster-wide reclaim-via-replication semantic — for the post-deploy case the payload's job is genuinely done everywhere, and keeping it on failed-peer deploys is the right conservative default. Scoping the disk-pressure hook and retroactive reclaim to Part 2 makes sense.

One non-blocking note: the two red integration shards (Bun 5/6, Node 24 6/6) are unrelated job-queue flakes — `csv_file_load` 120s job timeout, `concurrent claim`, `Job queue lifecycle`, none of which touch deployment-payload code. Worth a re-run to get a green board before merge, but not a code concern.

Approving — nice work.

PR #1498, opened 2026-06-26
perf(describe): add skip_record_count to skip the per-table count scan

dawsontoth:
## Summary

Adds an opt-in `skip_record_count` flag to the `describe_*` operations so callers that only need schema can avoid the expensive record-count scan.

`describe_table` / `describe_all` / `describe_schema` compute `record_count` via `getRecordCount`, which **scans each table's primary store**. That scan dominates describe latency on large tables, and `describe_all` pays it **serially, for every table** — the root cause behind the slow Databases tab in Studio ([HarperFast/studio#1367](https://github.com/HarperFast/studio/issues/1367)).

## What changed

- New boolean **`skip_record_count`** (Joi-validated) on `describe_table`, `describe_all`, `describe_schema`. When set, `record_count` and `estimated_record_range` are omitted and the count scan is skipped entirely.
- The cheap, O(1) stats (`table_size`, `db_audit_size`, `last_updated_record`) are still returned.
- Threaded through `describeAll` and `describeSchema`; documented in the MCP operation input schemas.

## Compatibility

Fully backward compatible — omitting the flag preserves today's behavior, and `validateBySchema` already passes `allowUnknown: true`, so older clients/servers tolerate the new key. The estimate-vs-exact distinction Studio relies on (`estimated_record_range` present ⇒ estimate) is unchanged.

## Pairs with

The Studio frontend change that consumes this: it fetches `describe_all` with `skip_record_count`, renders records off the count-free schema, and fetches the count separately (estimate first, exact on demand). Supports [HarperFast/studio#1367](https://github.com/HarperFast/studio/issues/1367).

## Tests

Added two cases in `integrationTests/apiTests/northwind.test.mjs` asserting `describe_table` / `describe_all` with `skip_record_count: true` return schema but omit `record_count` / `estimated_record_range`. `tsc` (build project) and `oxlint` clean.

### DIFF
--- a/components/mcp/tools/schemas/operations.ts
+++ b/components/mcp/tools/schemas/operations.ts
@@ -34,7 +34,13 @@ export const OPERATION_INPUT_SCHEMAS: Record<string, object> = {
 	// ─── describe_* ───────────────────────────────────────────────────────
 	describe_all: {
 		type: 'object',
-		properties: {},
+		properties: {
+			skip_record_count: {
+				type: 'boolean',
+				description:
+					'Omit each table’s `record_count` (and `estimated_record_range`) to skip the per-table count scan, which dominates latency on large databases. Defaults to false.',
+			},
+		},
 		description: 'Returns the full schema tree: every database, table, and attribute the caller can describe.',
 	},
 	describe_schema: {
@@ -42,12 +48,20 @@ export const OPERATION_INPUT_SCHEMAS: Record<string, object> = {
 		properties: {
 			schema: { type: 'string', description: 'Database name (legacy: "schema"). Required.' },
 			database: { type: 'string', description: 'Database name (preferred). Required if `schema` is omitted.' },
+			skip_record_count: {
+				type: 'boolean',
+				description: 'Omit each table’s `record_count` to skip the count scan. Defaults to false.',
+			},
 		},
 	},
 	describe_database: {
 		type: 'object',
 		properties: {
 			database: { type: 'string', description: 'Database name. Required.' },
+			skip_record_count: {
+				type: 'boolean',
+				description: 'Omit each table’s `record_count` to skip the count scan. Defaults to false.',
+			},
 		},
 		required: ['database'],
 	},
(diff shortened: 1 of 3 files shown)


dawsontoth:
Frontend counterpart (consumes this flag): HarperFast/studio#1381

dawsontoth:
**CI note — the failing `Integration Tests 5/6` checks are pre-existing flakes, not from this PR.**

This PR changes only `dataLayer/schemaDescribe.ts`, `components/mcp/tools/schemas/operations.ts`, and `integrationTests/apiTests/northwind.test.mjs` — nothing touching vector indexes, CSV bulk load, or role permissions. The new `skip_record_count` tests pass.

Evidence it's unrelated suite instability:
- `integration-tests.yml` is failing on **every recent `main` run**, including `e2af4433` (v5.1.14), the base of this branch.
- Shard 5/6 fails with a **different, unrelated test each run/runtime**:
  - this PR, original run: `vector-index-integrity` — `expected ≤4 misses after backfill; got 7/40` (v24 + Bun)
  - this PR, re-run: a `csv_file_load` job timeout (`did not complete … IN_PROGRESS`) on Bun
  - `main` @ v5.1.14, v24: a describe-permissions assertion (`Attribute 'id' does not exist on 'other.owner'`) — while the vector test passed
- v22 and v26 pass with this branch's identical code.

I re-ran the failed jobs once; they failed again on different unrelated tests, confirming flakiness rather than a clearable blip. No code change is warranted here. Flagging for visibility — happy to defer to the team on whether these checks block merge given `main`'s current state.

_🤖 Posted by [Claude Code](https://claude.com/claude-code)_

kriszyp (approved):
Approving — clean, well-shaped perf change.

`skip_record_count` is a true opt-in gate: `if (!skip_record_count)` is a single falsy property-access on the default path, so the default `describe`/`describe_all` behavior is unchanged and pays nothing. It only skips the genuinely expensive part — the per-table row-count scan (the O(n) primary-store iteration with the 500ms budget + estimate fallback), which `describe_all` was paying serially per table. `table_size`/`db_audit_size`/`last_updated_record` correctly stay unconditional (O(1)). Nice touch moving `estimated_record_range` inside the guard — it'd otherwise be a dangling `undefined` on the skip path.

On the red CI: both failures look pre-existing/unrelated, not a regression from this change — Node v24 shard 5/6 is the vecindex `INDEX_REBUILDING` timing flake (reproduces without this PR), and Bun shard 5/6 is an `export_local` job timeout in a different subsystem. The new tests pass green on Bun and all Node versions.

Minor, non-blocking: no `describe_schema` + `skip_record_count` test (the path is correct, just uncovered) — worth a follow-up if you want full coverage of the three describe shapes.

— KrAIs (Claude Opus 4.8)

PR #1500, opened 2026-06-26
Set up default config for gemini PR reviews

kriszyp:
These are the default settings for Gemini PR reviews (well, I added package-lock.json, setting this up as a baseline for customizing and copying to other repos).


dawsontoth (approved):

dawsontoth:
Ah, the formatter would be happier if it got ran on the new yaml!

cb1kenobi (approved):

Issue #1402, opened 2026-06-19
Docs/UX gaps from QA-explorer campaign (compression, audit+TTL, write-authz, bulk POST, 4xx-as-500, …)


kriszyp:
## Summary

Aggregated docs/UX gaps surfaced by the QA-explorer exploratory campaign (Harper `7aaa5a152`; all cited code byte-identical to `main` @`6797f091d` except one unrelated revalidation line). Each is defensible-but-surprising — a documentation or small-UX clarification rather than a correctness bug. Grouped to reduce issue noise; split out any that warrant their own tracking. Companion notes tied to specific filed defects are referenced there and omitted here.

## Items

1. **Compression is brotli-only; `gzip` Accept-Encoding silently ignored; default threshold disables it.** Response compression honors only `Accept-Encoding: br` (`contentTypes.ts`). A `gzip`-only client (incl. Node's built-in http client) gets uncompressed bodies with no signal. Also `http.compressionThreshold` defaults to **0**, which *disables* compression entirely (easy to misread as "compress everything"). A `Content-Encoding: gzip` request body → 500 (no request decompression; should be 415). *Suggest: doc brotli-only + the threshold default; consider gzip support; map the request-body case to 4xx.*

2. **TTL expiry is invisible in the audit log and erases the record's history.** On an `audit:true` + TTL table, expiry produces no audit entry (eviction is `removeEntry`, not a tombstone — `Table.ts:1606`, deliberate) and prunes prior-version history; time-travel to a pre-expiry timestamp returns 404. So "expired" is indistinguishable from "never existed," and pre-expiry forensics are impossible. *Suggest: doc that TTL eviction is invisible to the audit log; audited tables needing expiry-forensics should model deletes explicitly.*

3. **Resource-layer row-write guards are a complete boundary only if the principal's role lacks direct table-write grants.** `allowUpdate`/`allowDelete`/`allowCreate` are enforced consistently across REST write surfaces (correct 403s), but SQL UPDATE/DELETE go through the operations-API admin path (by design) gated only by role table-permissions — they do not consult resource-layer overrides. *Suggest: doc that row guards + role table-permissions must be configured together.*

[...]

PR #1501, opened 2026-06-26
Honor thrown Response and `status` field for custom-resource HTTP status

kriszyp:
## Summary

Three status-setting patterns from custom resource handlers silently failed (surfaced by the QA-explorer campaign and the docs work in HarperFast/documentation#534). This fixes two; the third is left as an intentional footgun.

- `throw { status: 400 }` → now **400**. The error-status read in `server/REST.ts` and both `onError` handlers in `server/http.ts` fall back to `error.status` when `error.statusCode` is absent.
- `throw new Response(body, { status: 422 })` → now **short-circuits** to that Response (status, headers, body honored) instead of a 500.
- `return { status, data }` *without* a `headers` field — **intentionally unchanged** (still 200). `status` collides with ordinary record attributes (e.g. an order's `status: 'shipped'`), so `headers` stays the disambiguator. Documented in #534.

The returned-Response handling is extracted into a shared `finalizeResponse` helper so the return path and the thrown-Response path honor status/headers/body identically.

## Where to look

- **Transaction semantics of a thrown Response** (`server/REST.ts` catch path). A thrown Response aborts the transaction like any other throw — a write before the throw **rolls back** — and the Response is surfaced from the outer catch. This is deliberate ("throw = rollback"; `return` a Response to commit). Codex's cross-model review flagged the alternative (committing a thrown Response) as the risk; we chose rollback for invariant-simplicity and zero hot-path cost. New test `4d` guards it: a POST that writes then throws a 201 Response returns 201 but the row is gone (`Kv` readback 404).
- **`error.status` aliasing** in the two `http.ts` `onError` handlers — applied for symmetry so the behavior is uniform across the REST and http error paths.

## Testing

`integrationTests/resources/resource-status-contract.test.ts` (the existing F-039 regression marker) updated to assert the new behavior + the rollback guard — 7/7 green locally. Full unit suites run in CI.

## Review notes

- Cross-model review: Codex leg ran and surfaced the transaction-rollback consideration (resolved by the rollback-guard test + the deliberate B semantic above). The Gemini/`agy` leg was unavailable in the environment (returned empty/timed out) — no second outside-model pass beyond Codex.

Refs #1402. Docs: HarperFast/documentation#534.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (model: claude-opus-4-8[1m])

### DIFF
--- a/integrationTests/resources/resource-status-contract.test.ts
+++ b/integrationTests/resources/resource-status-contract.test.ts
@@ -17,9 +17,10 @@
  * Findings are printed in `after()` as compact matrices; assertions only guard
  * server survival and known-contract expectations.
  *
- * This test documents working idioms + gaps (D-086) and is an F-039/#1421 regression
- * marker — assertions match CURRENT behavior so it is green now. When F-039/#1421 land,
- * update the expected statuses accordingly.
+ * This test documents working idioms + gaps (D-086). F-039/#1421 have landed: a thrown
+ * Response now short-circuits (status/headers/body honored) and `throw {status}` maps the
+ * `status` field as an alias for `statusCode`. A thrown Response still aborts the transaction
+ * like any throw, so a preceding write rolls back (4d). Assertions below guard that behavior.
  *
  * Reproduction:
  *   npm run test:integration -- "integrationTests/resources/resource-status-contract.test.ts"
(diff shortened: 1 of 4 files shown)


Ethan-Arrowood (approved):
LGTM — clean refactor and two well-scoped behavior fixes.

A few things I checked and liked:

- `finalizeResponse` is a behavior-preserving extraction — the return path and the thrown-Response path now share the exact same status/headers/body finalization, including the "don't reassign `.headers` on a real Response" guard. Nice consolidation.
- The thrown-Response short-circuit in the `catch` is safe — `headers` is declared before the `try`, so it's in scope there.
- Great test coverage: flipping the F-039 markers from defect to asserted behavior, plus the new `4d` rollback guard (write-then-throw-201 → 201 surfaced but `Kv` readback 404) nails down the contract.

I'm on board with the deliberate "throw = rollback, return = commit" semantic for Response objects — it's the simplest invariant, it's tested, and it's documented. Good call choosing rollback over committing a thrown Response.

dawsontoth (approved):

PR #1503, opened 2026-06-26
RFC 0001 — Typed, discoverable resources (+ code-first schema spikes)

dawsontoth:
**Status: Draft** — RFC for discussion; the spikes have now been promoted into a working implementation of the canonical code-first model (phase 2b), landing on this branch.

## What this is

RFC + implementation for strengthening type discoverability and strength across Harper resources. The full RFC is in [`docs/rfcs/0001-typed-discoverable-resources.md`](docs/rfcs/0001-typed-discoverable-resources.md).

The anchoring principle (forced by `--conditions=typestrip`): **runtime metadata must be values; TypeScript types are derived from them — never the reverse.** Taken to its conclusion, the schema itself can be authored as a TS value with the record type *inferred*.

### Three pillars
1. **One canonical schema model** — define tables in GraphQL *or* TypeScript; both compile to one internal model, projected to every surface from a single core (`attributeToFragment`).
2. **A per-method request contract** — `withSchema`/`defineResource` carrying `params/query/body/response`, with handler types derived and the same declaration feeding validation + OpenAPI + MCP.
3. **Typed base-request behavior** — base types for Harper's built-in filter/sort/select/relationship-include grammar, keyed to the record and its relationship graph.

## Implementation: canonical `defineTable` + `types` (NEW — `resources/defineTable.ts`)

The canonical-track model, for real:

```ts
import { defineTable, types } from 'harperdb';
const { id, int, string, date } = types;

export const Track = defineTable('Track', {
	id: id.primaryKey,                                   // getter-style flags
	name: string.indexed,
	duration: int.nullable,
	status: types.enum(['draft', 'published']).indexed,
	createdAt: date.createdTime,
});
// Track IS the live registered class — the import is the typed handle
type Track = InstanceType<typeof Track>;
type InsertTrack = (typeof Track)['$insert'];           // $record/$insert/$upsert/$patch/$query
```

- **Eager registration through the canonical path**: `defineTable` compiles to the same typeDef GraphQL builds and registers via the existing `table()` factory — shared DDL/migration semantics (re-definition re-asserts/evolves the schema), not a parallel path.
- **One name, projected per verb**: singular `Track` is the table you query/write through *and* the instance type; `$record` (readonly reads), `$insert` (PK optional), `$upsert` (PK required), `$patch`, `$query` (indexed fields only) are inferred members, never parallel names.
- **Relations via lazy thunks**: `types.relation(() => Album, { from: 'albumId' })` / `types.hasMany(...)` — forward references and cycles are safe (the `definition` late-binding contract the query-time resolvers already rely on). Like GraphQL, the to-one FK must be declared in the shape, which makes it typed, projected, and writable.
- **GraphQL parity locked by test**: both front-ends project identical canonical `properties` per field (including the nullability mapping: code-first plain field ≡ `!`, `.nullable` ≡ GraphQL plain; PK/server-managed unmarked).
- **Loader untouched**: exporting a handle from `resources.js` auto-exposes it through the existing resource walk — same semantics as `export class X extends tables.X {}`.
- **Type contract enforced on the real module**: `docs/rfcs/spikes/0001/defineTable-real.check.ts` re-runs the canonical-track assertions (incl. `@ts-expect-error` negatives) against the built package types.
- **`relationOf`/`hasManyOf` — mutual-pair escape hatch**: a truly mutual pair (`Book.author` <-> `Author.books`, each referencing the other) breaks TS's eager const-initializer inference — `typeof Book`/`typeof Author` each depend on the other and collapse to `any` (`TS7022`), verified empirically both ways (fails with plain `relation`/`hasMany` on both sides, passes with the escape hatch). `relationOf<R>`/`hasManyOf<R>` type the thunk target `() => any` and take the related record shape explicitly via `<R>` instead of inferring it, breaking the cycle; runtime is identical to `relation`/`hasMany` (only the type path differs — proven by a dedicated registration test).

Verify: `npm run build && npx mocha unitTests/resources/defineTable-registration.test.js unitTests/resources/jsResource.test.js` → 16 passing; `npx tsc --noEmit --project docs/rfcs/spikes/0001/tsconfig.json` → green.

### Cross-model review (Codex + Gemini + Harper-domain, adjudicated)

No blockers. Addressed in-branch: OpenAPI `$ref` components for code-first relations now always resolvable (lazy `type`/`attributes` on the relation `definition`); to-one FK must be declared (was an untyped auto-added column). Surfaced for discussion:

- **Open design question — explicit `null`**: nullable fields type as `?: T` (absence), but the runtime accepts/returns explicit `null` for unmarked-nullable fields, so `patch(id, { duration: null })` is runtime-valid yet type-rejected. Drizzle-style would be `T | null`. The canonical spike pins optional-only; keeping it that way unless we decide otherwise.
- **Follow-up — enum enforcement**: `types.enum()` narrows the TS type, but the shared `Table.validate()` has no enum case (true for GraphQL today too) and `attributeToFragment` drops the constraint. A runtime enum check would benefit both front-ends; deferred as its own issue.

## Spikes (proof-of-concept, superseded by the implementation above)

- [x] **(a)** RFC document — `docs/rfcs/0001-typed-discoverable-resources.md`
- [x] **(b)** `t` builder + `defineTable` inference — `docs/rfcs/spikes/0001/t-builder.spike.ts` *(superseded by canonical-track)*
- [x] **(c)** `defineTable` value → `table()` registration — promoted to `unitTests/resources/defineTable-registration.test.js`, now exercising the real public API
- [x] **canonical-track** — one `Track`, every verb shape inferred — `docs/rfcs/spikes/0001/canonical-track.spike.ts` (the model the implementation follows)
- [x] **enforceSchema** — subset-typed request contract — `docs/rfcs/spikes/0001/enforce-schema.spike.ts` (next up: its `record`/`body` slots take the `$` projections directly)

This PR is kept up to date as work proceeds. **Not for merge as-is** — it is a living design surface.

## Context

Grounded in a review of `harper` core (`resources/`, `components/mcp/`, `server/`) and the `HarperFast/schema-codegen` plugin. Verified findings (e.g. parameterized custom resources produce zero MCP tools; OpenAPI emits none of the query grammar; four divergent type translators over one `Attribute`) are tabulated in §11 of the RFC.


docs/rfcs/0001-typed-discoverable-resources.md
```diff
@@ -0,0 +1,338 @@
...
+### How the inference works
+
+`withSchema(contract)` is a generic static returning a subclass typed by the contract. Path params come from a template-literal parse of the path string; body/query/response come from the schema's inferred output type. This matches the runtime exactly: matched params are assigned as strings onto the target.
+
+```ts
+// '/widget/:id/rev/:n' -> { id: string; n: string }
+type PathParams<S extends string> =
+  S extends `${string}:${infer P}/${infer Rest}` ? { [K in P]: string } & PathParams<`/${Rest}`>
+  : S extends `${string}:${infer P}` ? { [K in P]: string }
+  : {};
```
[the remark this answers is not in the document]
dawsontoth: Good catch — added a wildcard arm to `PathParams` so `*wildcard` segments are typed (named `*path` → `{ path: string }`; bare `*` → `{ wildcard: string }`), matching `static path` support in `resources/Resource.ts`. Pushed in 2b94537.

kriszyp (commented):
This is excellent, lots of great ideas in here. Let me comment on some of them:

> One source of truth  ... Today GraphQL SDL is the only way to define a table, and the canonical Attribute[]

Yes, that was previously true, and we did have a lot of muddiness with this in the past, but I believe https://github.com/HarperFast/harper/pull/1167 has really solidified a single source of truth around canonical JSON-schema aligned data structures on classes. This is fed from graphql and is compatible/consistent across tables and custom resources, I believe, and largely addressed pillar 1.

> defineTable

I think `defineTable` would be a great addition. I think there are some things to consider here:
1. I would like to consider a little more literate version of this, is it feasible to write:

```
import { defineTable, types } from 'harper';
const { id, int, string } = types;
export const Tracks = defineTable('Tracks', {
	id: id.primaryKey,
	name: string.indexed,
	duration: int.nullable,
```
   2. There are numerous options for tables beyond their fields, including expiration, sealed, replicated, etc. I guess that all goes in a third arg?
   3. I still believe that trying to teach people how to simultaneously deal with a singular `Track` and plural `Tracks` is going to be confusing. I have very intentionally sought a singular table name and type to give devs a single concept work with. I think when/why to plurarize is not a complexity that we want to burden devs with.

> 4.2 Serve every surface from one projector

Is one projector different than a single source of truth?

>  Pillar 2 — A per-method request contract

Augmenting a resource with schema-enforcement (and trustable types built on that), sounds good. But the current `withSchema` seems rather hostile to our best-practice of static methods and seems to introduce a third request/target type, `RequestOf`. I don't think we want the complexity of third request type, I think polymorphic/type-subset compatibility should be a explicit goal of schema-ification. And it should be compatible with top-level (non-instance functions):

```
export const Widget = {
	path: '/widget/:id',
	async get(req) {
		return loadWidget(req.id, req.get('expand'));
	}
	async post(req) {
		return createWidget(req.body);
	}
}
// and StricterWidget should be type compatible stricter subset, directly drop-in replacement, no new types
export const StricterWidget = enforceSchema<RecordType>({ // RecordType is default return type of `get` and input type of `put`
	get: ... extra type definitions
	post: ...
// req.id: string · req.get('expand')?: ('parts'|'owner')[] * req.body all are now inferred in a type subsetting fashion
}, Widget);
```

I'm hesitant to get into too much discussion of later pillars since they appear to somewhat build on top of each other.

Also, I'm curious about the intended future of the PR. Is 0001-typed-discoverable-resources.md a transient component of the PR? We have intentionally been putting proposals in GH issues (and several much larger than this) rather than PRs to avoid transient information in the repo. However, implementation plans as a transient artifacts of a PR in progress is something we have occasionally done.

docs/rfcs/0001-typed-discoverable-resources.md
```diff
@@ -0,0 +1,347 @@
...
+| MCP              | `derive.ts` (own switch)    | `attributes = []`       | Mistypes nested objects/arrays (capitalized literals the parser never emits); parameterized custom routes yield **zero** tools; advertises a schema it does not enforce |
+| Editor types     | `schema-codegen mapType.js` | Not covered             | Date→`string`, Blob→`any`; no enums; computed/timestamp fields neither `readonly` nor omitted from `New<T>`                                                             |
+| _(shared core)_  | `attributeToFragment`       | —                       | Exists and is on the parse path, but the four surfaces above bypass it                                                                                                  |
+
+> **Sharpest finding.** A custom resource with `static path = '/widget/:id'` is registered into a `paramRoutes` side-array and `Resources.set()` returns _before_ adding it to the base map that MCP iterates. Result: parameterized custom resources produce **zero MCP tools**, while OpenAPI emits them with string-only params and a generic body.
+
+## 4. Pillar 1 — One canonical schema model
```
dawsontoth, line 64: @kriszyp breaking the various threads out for easier discussion:

> > One source of truth ... Today GraphQL SDL is the only way to define a table, and the canonical Attribute[]

> Yes, that was previously true, and we did have a lot of muddiness with this in the past, but I believe https://github.com/HarperFast/harper/pull/1167 has really solidified a single source of truth around canonical JSON-schema aligned data structures on classes. This is fed from graphql and is compatible/consistent across tables and custom resources, I believe, and largely addressed pillar 1.

I think 1167 helps to fill out the information we can feed out through MCP and OpenAPI, but it doesn't tackle the harder problem I see pillar 1 addressing: a single, discoverable, cohesive surface defining the schemas that infers types with full language support without any generation.

docs/rfcs/0001-typed-discoverable-resources.md
```diff
@@ -0,0 +1,347 @@
...
+```ts
+import { defineTable, t } from 'harperdb';
+
+export const Tracks = defineTable('Tracks', {
+	id: t.id().primaryKey(),
+	name: t.string().indexed(),
+	duration: t.int().nullable(),
+	status: t.enum(['draft', 'published']), // -> 'draft' | 'published'
+	createdAt: t.createdTime(), // server-managed -> readonly, omitted from insert
+	album: t.relation(() => Albums, { from: 'albumId' }), // a real reference -> exact typed graph
```
dawsontoth, line 91: > I would like to consider a little more literate version of this, is it feasible to write:
```
class Types {
	private metadata: Partial<Record<keyof Types, any>> = {};

	get id(): Types {
		this.metadata.id = true;
		return this;
	}
	get string(): Types {
		this.metadata.string = true;
		return this;
	}
	enum(values: string[]): Types {
		this.metadata.enum = values;
		return this;
	}
	get indexed(): Types {
		this.metadata.indexed = true;
		return this;
	}
	get nullable(): Types {
		this.metadata.nullable = true;
		return this;
	}

	build(): Partial<Record<keyof Types, boolean>> {
		return this.metadata;
	}
}


const types = new Types();

console.log('empty build:');
console.log(types.build());

console.log('example string');
console.log(types.string.indexed.build());

console.log('example enum');
console.log(types.indexed.enum(['a', 'b']).build());
```
output:
```
empty build:
{}
example string
{ string: true, indexed: true }
example enum
{ string: true, indexed: true, enum: [ 'a', 'b' ] }
```

docs/rfcs/0001-typed-discoverable-resources.md
```diff
@@ -0,0 +1,347 @@
...
+});
+
+type Track = t.Select<typeof Tracks>; // inferred read record
+type NewTrack = t.Insert<typeof Tracks>; // PK + computed + timestamps omitted, inferred
```
dawsontoth, line 95: > I still believe that trying to teach people how to simultaneously deal with a singular Track and plural Tracks is going to be confusing. I have very intentionally sought a singular table name and type to give devs a single concept work with. I think when/why to plurarize is not a complexity that we want to burden devs with.

@kriszyp I think this shows where the singular vs plural discussion comes to something fruitful. A table doesn't give or receive a single interface, it needs different things when you're querying, responding, inserting, upserting, patching, etc. By having the table class in TypeScript, we can follow the type inference precedents set out by other libraries to make it really easy to maintain this sprawl of types. Because the fields are defined in TypeScript, we can include or omit fields automatically.

docs/rfcs/0001-typed-discoverable-resources.md
```diff
@@ -0,0 +1,347 @@
...
+import { defineTable, t } from 'harperdb';
+
+export const Tracks = defineTable('Tracks', {
+	id: t.id().primaryKey(),
+	name: t.string().indexed(),
+	duration: t.int().nullable(),
+	status: t.enum(['draft', 'published']), // -> 'draft' | 'published'
+	createdAt: t.createdTime(), // server-managed -> readonly, omitted from insert
+	album: t.relation(() => Albums, { from: 'albumId' }), // a real reference -> exact typed graph
+});
```
dawsontoth, line 92: > There are numerous options for tables beyond their fields, including expiration, sealed, replicated, etc. I guess that all goes in a third arg?

I think so. Another approach would be to shift to accept a single dictionary and do it based off of names, nest the fields in one level. That would be more discoverable, and I like the visual nesting you get, too.

docs/rfcs/0001-typed-discoverable-resources.md
dawsontoth, line 1: > Also, I'm curious about the intended future of the PR. Is 0001-typed-discoverable-resources.md a transient component of the PR? We have intentionally been putting proposals in GH issues (and several much larger than this) rather than PRs to avoid transient information in the repo. However, implementation plans as a transient artifacts of a PR in progress is something we have occasionally done.

It was purely because PRs have threads, separate files, and syntax highlighting. All of which feel very helpful for the conversation. I don't think the RFC would actually land in the repo, since over time it'll become inaccurate. The 0001 was Claude being optimistic.

docs/rfcs/0001-typed-discoverable-resources.md
```diff
@@ -0,0 +1,347 @@
...
+| _(shared core)_  | `attributeToFragment`       | —                       | Exists and is on the parse path, but the four surfaces above bypass it                                                                                                  |
+
+> **Sharpest finding.** A custom resource with `static path = '/widget/:id'` is registered into a `paramRoutes` side-array and `Resources.set()` returns _before_ adding it to the base map that MCP iterates. Result: parameterized custom resources produce **zero MCP tools**, while OpenAPI emits them with string-only params and a generic body.
+
+## 4. Pillar 1 — One canonical schema model
+
+Today GraphQL SDL is the only way to define a table, and the canonical `Attribute[]` it produces is then translated to a type-shaped representation in **four** independent places that disagree. Two changes turn that model into a true single source of truth: let it also be **authored in TypeScript**, and serve every surface from **one projector** instead of four. Both front-ends compile to the same internal model; everything else is derived from it.
```
dawsontoth, line 66: > Is one projector different than a single source of truth?

I don't know that projector was the best choice of words, but the way I was visualizing it is... GraphQL can be a single source of truth, yes. But there isn't language-service-level-support for making it project outward into JavaScript and TypeScript in all the ways we need for a great developer experience. That's where the codegen plugin I wrote comes in, but it has compromises (like causing a cascade of Harper restarts, or only doing stuff when Harper is running it). OpenAPI and MCP do similar work for us because GraphQL doesn't natively have a bridge to TS/JS.

By instead shifting the source of truth to be in that TS/JS world, as soon as the characters are typed, the interfaces can be inferred without any codegen (thus eliminating that plugin). The MCP and OpenAPI work becomes more succinct as well.

docs/rfcs/0001-typed-discoverable-resources.md
```diff
@@ -0,0 +1,347 @@
...
+- **Blob** in `DATA_TYPES` — fixing OpenAPI's typeless Blob.
+- **Nullability as a parameter** — `union` form for MCP and TS, `nullable: true`/required-array for OpenAPI 3.0.
+- **Direction** — an `input` vs `output` flag so write schemas advertise the coercible accept-set (Date accepts `string|number|Date`) while read schemas advertise the stored type.
+
+Then make `schema-codegen` a pure _fragment → TypeScript_ printer rather than a fourth switch. Lock everything with one golden conformance fixture that asserts the **GraphQL and TypeScript front-ends round-trip to identical canonical models**, and that all surfaces agree on every edge (Long, Date, Bytes, Blob, `[Type]`, nested objects, `@relationship`, `@computed`, unspecified nullability).
+
+> **Main implementation risk.** The TypeScript front-end must drive the exact same DDL/migration semantics as GraphQL — attribute add/remove, index changes, `sealed`, audit, replication — or the two paths subtly diverge. The front-end conformance fixture is the guardrail; schema evolution should run off the canonical `typeDef`, not off either authoring format directly.
+
+## 5. Pillar 2 — A per-method request contract
```
dawsontoth, line 119: > Augmenting a resource with schema-enforcement (and trustable types built on that), sounds good. But the current withSchema seems rather hostile to our best-practice of static methods and seems to introduce a third request/target type, RequestOf. I don't think we want the complexity of third request type, I think polymorphic/type-subset compatibility should be a explicit goal of schema-ification. And it should be compatible with top-level (non-instance functions):

```
export const Widget = {
```

Couldn't we use the provided language concepts for polymorphism? The struggle with statics is discoverability and adherence to an interface. You can put anything you want on `export const Widget`, and if you typo `gwt` instead of `get`, we have no mechanism to detect if this is a messed up resource or table or some utility class that we should ignore. But yes, this does lead to a very different interface across the stack, and hopefully to the removal of the old interfaces in a large breaking change, v6, v7, whenever.

docs/rfcs/0001-typed-discoverable-resources.md
kriszyp: We can't genericize `t.Select<typeof Track>` and `t.Insert<typeof Track`?

dawsontoth: Can you show me what you mean with example code? I'm a bit confused.

docs/rfcs/0001-typed-discoverable-resources.md
kriszyp: So a canonical schema encompasses the scope of types? That seems to contradict pillar's assertion that it "also be authored in TypeScript", and "no generation"; i.e. we aren't attempting to create (canonical)types from graphql in the scope of this PR, are we? However, I definitely support the ability to define a single canonical source of type definitions and runtime metadata for a specific table from a `defineTable` call though (that's just don't make everything have a universal single source of truth across types and runtime).

docs/rfcs/0001-typed-discoverable-resources.md
kriszyp: Seems like you could do this without the `build()` call (by skipping `metadata` layer), right?

docs/rfcs/0001-typed-discoverable-resources.md
kriszyp: here is what I generated for a spike/example: https://github.com/HarperFast/harper/blob/claude/sad-lehmann-1b3b70/docs/rfcs/spikes/0001/canonical-track.spike.ts

dawsontoth: sure, looks good to me

kriszyp:
@dawsontoth I put together spikes at docs/rfcs/spikes/0001/t-builder.spike.ts and an enforce-schema spike to plan out these paths. Let me know what you think (and I could proceed with implementation).

docs/rfcs/spikes/0001/enforce-schema.spike.ts
```diff
@@ -0,0 +1,326 @@
...
+	},
+	{
+		async get(target) {
+			// target.id: string (from ':id') · target.get('expand'): ('parts'|'owner')[] | undefined
+			return loadWidget(target.id, target.get('expand'));
+		},
+		async post(_target, data) {
+			// data: NewWidget (declared body) — no cast
+			return createWidget(data);
+		},
```
dawsontoth, line 231: Could these collapse together into a single `get` and `post`, instead of having two positional arguments? So...
```
get: {
	query: { expand: s.arrayOf(s.enumOf(['parts', 'owner'])).optional },
	execute: async get(target) {
		// target.id: string (from ':id') · target.get('expand'): ('parts'|'owner')[] | undefined
		return loadWidget(target.id, target.get('expand'));
	},
}
```

dawsontoth:
Nice

resources/defineTable.ts
```diff
@@ -0,0 +1,408 @@
...
+			if (meta.from) {
+				// many-to-one: this table holds the foreign key. Like GraphQL, the FK must be a
+				// declared field — that makes it typed, projected into `properties`, and writable
+				// in $insert/$upsert (the relation itself is a read-only projection).
+				if (!declared.has(meta.from)) {
+					throw new Error(
+						`Table "${name}": relation "${attrName}" references foreign key "${meta.from}", which must be declared in the shape (e.g. \`${meta.from}: id.indexed\`)`
+					);
+				}
+				attr = { name: attrName, relationship: { from: meta.from } };
```
kriszyp: These relationship attributes have `relationship` metadata but neither `nullable` nor a write-exclusion marker. Once the handle is exported, the application-profile MCP builder includes the relation and marks it required (`components/mcp/tools/schemas/derive.ts:144-153`), while table validation rejects any directly assigned relationship property (`resources/Table.ts:3750-3758`). For a `Book.author` relation, omitting `author` violates the advertised `create_Book` schema and providing it fails at runtime, so no input satisfies both contracts; `hasMany` has the same problem. Please fix this at the shared write-projector boundary by excluding relationship projections from write schemas, then cover an exported `defineTable` relation through application-profile tool generation.

resources/defineTable.ts
```diff
@@ -0,0 +1,408 @@
...
+type MaybePromise<T> = T | Promise<T>;
+
+// The verbs retyped by the projections. Full query-grammar typing (conditions/sort/select keyed
+// to the record) is Pillar 3 — here only inputs/outputs are narrowed.
+interface TypedVerbs<S extends Shape> {
+	get(id: IdOf<S>, context?: any): MaybePromise<ReadVariant<S> | undefined>;
+	put(record: UpsertOf<S>, context?: any): MaybePromise<unknown>;
+	post(record: InsertOf<S>, context?: any): MaybePromise<unknown>;
+	patch(id: IdOf<S>, changes: PatchOf<S>, context?: any): MaybePromise<unknown>;
+	update(id: IdOf<S>, updates?: PatchOf<S>, context?: any): MaybePromise<InstanceOf<S>>;
```
kriszyp: `update()` is described and type-checked as returning a live mutable instance, but a contextless static call commits its implicit transaction before returning (`resources/transaction.ts:62-70`). Thus `const live = await Track.update(id); live.name = 'x'`—the pattern in the real type proof—mutates too late to persist. `InstanceOf` also contains only data fields, hiding the real instance’s `save`/`delete` surface, so no typed persistence step repairs this. Please settle the lifecycle contract: either require and encode an explicit transaction/callback for live mutation, or stop presenting this result as independently mutable; if live instances remain public, include their persistence methods. A discriminating check is contextless update → mutate → read, contrasted with mutation inside an explicit transaction.

resources/defineTable.ts
```diff
@@ -0,0 +1,408 @@
...
+export interface DateField<F extends Flags = {}> extends Field<Date, F> {
+	readonly createdTime: Field<Date, F & { serverManaged: true }>;
+	readonly updatedTime: Field<Date, F & { serverManaged: true }>;
+}
+
+/** The read-record type of a related table handle. */
+type RecordOf<T> = T extends { readonly $record: infer R } ? R : never;
+
+// to-one resolves to the related record; to-many to an array of it.
+export type RelationField<Target, Card extends 'one' | 'many'> = Field<
```
kriszyp: A to-one relation is typed as an always-present related record, but Harper does not enforce referential integrity here. A row can be written with a foreign-key value that has no target row, and the resolver returns `undefined` (`resources/Table.ts:4006-4034`); code accepted by the proof such as `book.author.name` then fails at runtime. Please include `undefined` in the to-one relation projection unless registration also introduces an invariant guaranteeing the target exists. Add the missing-target case to the package-level type assertions and runtime relationship test.

resources/defineTable.ts
```diff
@@ -0,0 +1,408 @@
...
+// ─────────────────────────────────────────────────────────────────────────────
+
+/** The record's data fields: server-managed/relations readonly, others mutable; nullable -> optional. */
+type RecordData<S> = Resolve<
+	{
+		readonly [K in keyof S as IsReadOnly<S[K]> extends true ? K : never]: TsOf<S[K]>;
+	} & {
+		[K in keyof S as IsReadOnly<S[K]> extends false ? (IsNullable<S[K]> extends false ? K : never) : never]: TsOf<S[K]>;
+	} & {
+		[K in keyof S as IsReadOnly<S[K]> extends false ? (IsNullable<S[K]> extends true ? K : never) : never]?: TsOf<S[K]>;
```
kriszyp: `.nullable` changes only property optionality; `TsOf<F>` never includes `null`. Runtime validation explicitly permits null unless `nullable === false` (`resources/Table.ts:3744-3763`), and existing integration coverage confirms explicit null is preserved. Consequently valid inputs such as `{ duration: null }` are rejected by TypeScript, while read types claim a stored null cannot occur. Please derive a nullable value type such as `TsOf<F> | null` and use it consistently in record, insert, upsert, patch, and query projections, with positive type assertions for explicit null.

resources/defineTable.ts
```diff
@@ -0,0 +1,408 @@
...
+type PkKey<S> = { [K in keyof S]: IsPrimaryKey<S[K]> extends true ? K : never }[keyof S];
+type IdOf<S> = [PkKey<S>] extends [never] ? string | number : TsOf<S[PkKey<S>]>;
+
+type MaybePromise<T> = T | Promise<T>;
+
+// The verbs retyped by the projections. Full query-grammar typing (conditions/sort/select keyed
+// to the record) is Pillar 3 — here only inputs/outputs are narrowed.
+interface TypedVerbs<S extends Shape> {
+	get(id: IdOf<S>, context?: any): MaybePromise<ReadVariant<S> | undefined>;
+	put(record: UpsertOf<S>, context?: any): MaybePromise<unknown>;
```
kriszyp: `TableHandle` omits the base verbs and replaces each with one call signature, even though the returned value is the exact existing table class. This makes supported runtime forms—including `Track.put(id, record, context)`, batch puts, `Track.get(RequestTarget)`, and `Track.update(id, transaction)`—TypeScript errors; the argument normalizer still handles these forms at `resources/Resource.ts:525-650`. Please retain compatible typed overloads while narrowing their record/id values, or return a deliberately separate facade. Keeping the same class with a narrower declared surface makes the public type contract incomplete.

unitTests/resources/jsResource.test.js
```diff
@@ -125,4 +128,39 @@ describe('jsResource', () => {
...
+		const Widget = defineTable(
+			'Widget',
+			{ id: types.id.primaryKey, name: types.string },
+			{ database: 'jsresource_codefirst' }
+		);
+		const resourceModule = { Widget };
+
+		let capturedHandler;
+		const mockScope = {
+			handleEntry: spy((handler) => {
```
kriszyp: This new test adds several Sinon spies, contrary to the repository rule against new Sinon/Rewire usage. Only `capturedHandler` and the resources map are observed, so plain callbacks and no-op logger/requestRestart functions provide the same coverage without stubbing. Please replace these spies with ordinary functions and keep new assertions on the repository’s bare `assert` style.

PR #1505, opened 2026-06-26
perf(test): cut round-trips in HNSW vector-index integration tests

dawsontoth:
## Measured result (CI, Windows shard 5/6)

The restart-reduction restructure (deploy all tables once, ~5 worker restarts → 2) delivered the win:

| Test | Before | After |
|---|---|---|
| delete-entry-point | 476s | 0.25s |
| update-churn | 345s | 0.47s |
| threshold queries | 381s | 0.023s |
| reindex backfill | 362s | 2.75s |
| **HNSW suite total** | **1764s** | **422s (−76%, ~22 min saved)** |

All four vector tests pass on v24/v26/Windows. Per-test cost was almost entirely the per-test worker restart; the residual ~422s is the single shared `before` deploy+restart. (The round-trip reductions — bulk insert + parallel read-back — are a secondary, in-the-noise contribution on top.)

---

**Status: Draft** — Windows CI perf for the vector-index integration tests. **Stacked on #1504** (set as this PR base), so its shard-5/6 hardening is in scope and the perf change is validated without the pre-existing flake noise. Independent of the RFC PR #1503.

## Why

`integrationTests/server/vector-index-integrity.test.ts` is a top Windows CI hotspot — the `HNSW vector-index data-integrity` suite runs **~29 min** on Windows 5/6 (1763s), with individual tests at 284s/237s/184s.

The data is tiny (DIMS=8, N≤50). The cost is **~100–200 sequential, individually-committed HTTP round-trips per test** — i.e. per-request round-trip + per-write fsync, **not** vector math. Measured ratio: ~6× slower on Windows than local for the same suite.

## Changes (correctness-neutral)

- **Bulk-insert the reindex pre-index seed** — replace N sequential POSTs with one `insert` op. There's no HNSW index during these inserts (the table is altered to add the index afterward), so per-record ordering is irrelevant; the backfill builds the index from the stored rows either way.
- **Parallelize independent read-back searches** — the `update-churn` and `reindex` verification loops issue N independent QUERY searches; run them with `Promise.all` instead of serializing the round-trips.

**Deliberately left sequential** (order-sensitive, would change what's tested): the delete-entry-point insert/delete sequence (entry-point election) and the churn update sequence.

## Validation

⚠️ **CI-validated, not locally measured.** `restart_service` is flaky in a local multi-loopback setup — the per-test readiness probes time out at 60s, which both *fails* and *dominates* local timings (an artifact that doesn't exist on CI, where these tests pass). The `update-churn` test — the one that runs cleanly locally — still passes with the parallelized read-back. The real before/after is the **Windows 5/6 shard time on this PR's CI run** vs the ~1763s baseline.

## Bigger levers (not in this PR)
- **Per-test `restartHttpWorkers`** — each test restarts workers (~5 across the suite; reindex does 2). Worker-restart cost is likely a larger driver than the data ops, but reducing it means sharing setup across tests (changes test isolation) — worth a follow-up.
- **CI sharding/concurrency** — Windows runs one rotating shard; 5/6 is a ~29-min hotspot. Rebalancing heavy vector files across shards would cut wall-time for every PR. Infra-level, owner's call.

### DIFF
--- a/integrationTests/server/vector-index-integrity.test.ts
+++ b/integrationTests/server/vector-index-integrity.test.ts
@@ -226,29 +226,33 @@ async function waitFor(predicate: () => Promise<boolean>, timeoutMs = 60_000, in
 suite('HNSW vector-index data-integrity (integration)', (ctx: ContextWithHarper) => {
 	let client: any;
 
+	// All four tables are deployed by a SINGLE component with ONE worker restart in `before`.
+	// Previously every test deployed its own component and restarted the HTTP workers (~5 restarts
+	// across the suite); on slow runners that restart is the dominant per-test cost (a test doing
+	// 3 inserts still took ~265s on Windows — almost entirely restart/setup). Consolidating to one
+	// shared restart here — plus the single restart `reindex` needs for its mid-test index-add — is
+	// the bulk of this suite's wall-time. The tables use distinct names/databases so the tests stay
+	// isolated despite sharing one component; `reindex` runs last and is the only test that mutates
+	// the schema afterward, so re-asserting the others on its redeploy is a harmless no-op.
+	const PROJECT = 'vecintegsuite';
+	// ReindexTable starts WITHOUT an index — the reindex test adds it mid-test to exercise backfill.
+	const SCHEMA_INITIAL = [
+		makeSchemaWithIndex('DelEPTable', 'vecinteg1'),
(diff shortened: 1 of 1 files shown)


cb1kenobi (approved):
This is rad, nice work!

kriszyp (approved):

PR #1507, opened 2026-06-26
fix: resolve extension packages outside ROOTPATH during node_modules walk

pbrumblay:
Note!!! This is _UNTESTED_. For demonstration purposes only. 

Fixes extension package resolution when an application is run via RUN_HDB_APP from outside ROOTPATH (e.g. harper run from a monorepo's apps/web).

Harper walks up from the app directory looking for node_modules/<package>, but stops when the parent path's string length drops below getHdbBasePath().length. That incorrectly aborts the walk before reaching hoisted packages at the monorepo root when the checkout path is shorter than ROOTPATH.

Symptom:

Error: Could not load component '@harperfast/nextjs' for application 'web'
due to: Unable to find package @harperfast/nextjs:@harperfast/nextjs
Node's own require.resolve('@harperfast/nextjs', { paths: ['apps/web'] }) succeeds on the same machine — only Harper's loader fails.

Example (fails on v5.1.14):

App: /Users/usrna01/some-really-long-reposiotory-01/apps/web (46 chars at repo root)
ROOTPATH: /Users/usrna01/.local/share/mise/harper/674e1c5/hdb (51 chars)
Hoisted package: …/some-really-long-reposiotory-01/node_modules/@harperfast/nextjs
Walk stops at apps/ because 46 < 51 — repo root is never checked
Example (works by accident):

App: /Users/peter/src/rl/some-really-long-reposiotory-01/apps/web
Repo root length (51) ≥ ROOTPATH length (49) — walk reaches hoisted node_modules
This explains failures that appear machine-specific despite identical Node/Harper versions, branch, and package contents.

Fix
Replace the string-length heuristic with path-prefix logic:

Apps outside ROOTPATH (RUN_HDB_APP): walk up to the filesystem root
Components inside $ROOTPATH/components/: stop at the ROOTPATH boundary (preserves existing intent)

### DIFF
--- a/components/componentLoader.ts
+++ b/components/componentLoader.ts
@@ -9,7 +9,7 @@ import {
 	rmSync,
 	symlinkSync,
 } from 'node:fs';
-import { join, basename, dirname } from 'node:path';
+import { join, basename, dirname, sep } from 'node:path';
 import { isMainThread } from 'node:worker_threads';
 import { parseDocument } from 'yaml';
 import * as env from '../utility/environment/environmentManager.ts';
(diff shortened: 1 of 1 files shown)


pbrumblay:
We have a workaround which we'll add to our local dev script. 

```
mkdir -p apps/web/node_modules/@harperfast
ln -sf "$(pwd)/node_modules/@harperfast/nextjs" apps/web/node_modules/@harperfast/nextjs
```

kriszyp (commented):
This looks good, thank you! Do you want me to add some tests to finish this off?

pbrumblay:
> This looks good, thank you! Do you want me to add some tests to finish this off?

@kriszyp  - As noted I haven't actually tested this, so sure! If it looks reasonable and you want to accept it, great.

kriszyp:
Thanks for tracking this down, @pbrumblay — the path-prefix diagnosis (and the worked example showing it's purely a checkout-length-vs-ROOTPATH coincidence) made the root cause really clear. 🙌

I've opened #1536 which takes this over: your fix rebased onto `main`, plus regression tests covering the outside-ROOTPATH monorepo case, walk termination at the fs root, in-ROOTPATH resolution, and the preserved ROOTPATH boundary. Reverting to the old length heuristic makes the bug test fail, so it's locked in. Credited you in the description.

Proposing we land it via #1536 and close this one — shout if you'd rather drive it from here.

— KrAIs (Claude Opus 4.8)

pbrumblay:
Closing: https://github.com/HarperFast/harper/pull/1536 has this covered.

Issue #1422, opened 2026-06-20
`allowRead` has two critical security gaps: (1) throw → fail-open, (2) not enforced on collection scans


kriszyp:
## Summary

Two distinct security gaps in Harper's `allowRead` authorization hook:

1. **Exceptions thrown from `allowRead()` are silently swallowed — request proceeds as authorized** (fail-open)
2. **`allowRead` is invoked only once at collection-open with no loaded record — collection scan paths (REST collection GET, paginated queries) are not filtered per-row**

## Gap 1 — Exception handling: throw → fail-open

`allowRead()` throwing ANY exception (`Error`, string, `null`) is silently caught and the request proceeds as if authorization was granted.

```js
class Vault extends tables.Vault {
  allowRead() {
    throw new Error('auth failed')  // expected: 403
  }                                 // actual:   200 + protected payload returned
}
```

Control: `allowRead() { return false }` → 403 (correct). Denial via `return false` works; denial via throw does not.

This violates the fail-closed security principle. Any bug in `allowRead` that causes a throw — null-deref on a missing property, uninitialized dependency, unexpected input shape — silently permits access rather than blocking it.

**Fix:** wrap the `allowRead()` invocation in a try-catch; treat ANY exception as a denial (403) and log the error internally.

**Repro:** `npm run test:integration -- "integrationTests/qa-scratch/qa169-custom-resource-errors.test.ts"` (cells S1, S5, S5b)

## Gap 2 — Collection scan: `allowRead` not applied per-row

`allowRead` is called once at collection-open time with an unloaded `this` instance — all declared field properties return `undefined`. It is **not** called per-row during collection scanning.

**Repro:** 200-row table, 100 owned by `alice`, 100 by `bob`. Custom resource:
```js
allowRead(user, target, context) {
  return this.owner === context.user?.username
}
```

```
Single-record GET by Alice on Bob's row:  → 403  ✓  (this.owner loaded, check fires)
Paginated GET /Table/?sort(id)&limit(...) → 200 rows (alice AND bob rows — all leaked)
```

At collection-open, `this.owner === undefined` — the filter cannot distinguish rows. An attacker can retrieve every row by using the collection endpoint (`GET /Table/`) rather than fetching records by ID.

Note: the `?owner=alice` value-filter path returns only alice's rows, but that is a DB-level predicate enforced before `allowRead` — an attacker omits the filter.

[...]

PR #1522, opened 2026-06-29
fix(mcp): enforce row-level RBAC on MCP application verb tools (#1487)

### LINKED ISSUE #1487: Consolidate row-level allowRead enforcement — apply it on all read paths, not just the single-record point-read
## Summary

Row-level `allowRead` is currently enforced **only on the single-record REST point-read**. Several other read paths don't invoke it consistently — already tracked individually in **#1419** (live subscription delivery), **#1422** (REST collection scan), and **#1414** (established subscriptions not re-checking on revocation). This is a consolidation/tracking issue to fix `allowRead` enforcement **uniformly across all read paths** with a single lower-layer approach, rather than per-path patches.

Additional read-path enforcement gaps found during exploratory QA are recorded in the **Security Notes** field of the Harper project board (kept internal pending fixes).

## Proposed direction

Enforce `allowRead` (and attribute-level `read:false` permissions) at a **shared lower layer** — the storage / result-iterator path that every read surface flows through — so that point-read, collection scan, query/search, relationship expansion, and live + replay subscription delivery all filter rows consistently. The invariant to hold: *a row the caller cannot read on a point-read is never returned to them on any other surface either.* This also covers the revocation case (#1414): a per-row check on the delivery path naturally re-evaluates current permissions.

## Acceptance

- One enforcement point (or a small shared helper) invoked by all read surfaces; no surface returns `allowRead`-denied rows.
- Regression tests pinning each surface (several are already written under the qa-explorer scratch suite and can be promoted).

kriszyp:
## Summary

The MCP application-profile CRUD verb tools (`get_`/`search_`/`create_`/`update_`/`delete_<Table>`) did not honor a Resource's **row-level** RBAC predicates (`allowRead`/`allowCreate`/`allowUpdate`/`allowDelete`). A low-privilege user with table-level grants could write rows a per-record guard denied (mutations persisted) and read rows `allowRead` should hide — while the same operations over REST were correctly rejected. This brings the MCP path into parity with REST. Fixes #1487 (write fail-open + point-read bypass; see "Known residual" for the collection-scan piece).

## Root cause

Each verb-tool handler closed over the Resource class **captured at tool-registration time**. For an `@export`ed `@table`, that captured reference is the auto-generated *base table* class, whose `allow*` predicates only evaluate **table-level** RBAC. A component's `resources.js` subclass (`class Doc extends tables.Doc`) — which carries the **row-level** `allow*` overrides — registers at the same path *after* the MCP profile builds its tools, overwriting the base entry, and that registration does not re-trigger tool refresh. So MCP kept dispatching the base class forever; no `AccessViolation` was ever raised because the base table-level guard returned `true`. REST was unaffected because it resolves the live registry entry per request.

## Fix

Resolve the **live** Resource class from the registry at call time (`liveResource(path, fallback)`) and dispatch on it, so the same authorizing path REST uses runs for every MCP verb. The fallback to the registration-time class is taken only if the registry entry is gone (graceful degradation, not a weaker guard).

## Where to look

- `components/mcp/tools/application.ts` — `liveResource` + the five verb handlers resolving it at call time; the same treatment applied to `makeCustomMethodHandler` for `mcpTools` parity.
- `integrationTests/mcp/row-authz.test.ts` + `integrationTests/fixtures/mcp-row-authz/` — new regression suite: denied create/update/delete return an MCP error **and** leave the store unchanged (verified by admin read-back); denied get/search leak nothing; admin positive controls persist; a low-priv user can still write rows it owns (guards are selective). Passes under both RocksDB and LMDB.

## Points worth a reviewer's eye

- **Behavior change beyond RBAC (intended):** because MCP now dispatches the exported subclass, a custom Resource's overridden verb methods (`post`/`put`/etc.) run for MCP calls too — this is correct REST parity, but it is a behavior change for custom resources. It surfaced as one existing assertion in `integrationTests/mcp/application.test.ts` (`#1324`): the fixture's custom `WorkItem.post` stores `payload: JSON.stringify(body)`, so the round-tripped payload now matches what REST returns; the assertion was updated to reflect that (it was previously pinning the base-table dispatch).
- **Known residual (scoped to #1487 family):** per-row `allowRead` *filtering* on collection scans depends on the shared lower-layer enforcement in #1422 / #1489 (not yet on `main`). Until that lands, `search_<T>` on a row-restricted table returns an authorization error rather than a filtered result — it does **not** leak. Routing MCP through the live authorizing dispatch means it inherits the per-row filtering automatically once #1489 merges, rather than a parallel MCP-only check.

## Cross-model review

Codex and Gemini (via the cross-model-review skill) plus a Harper-domain pass. No in-scope blockers. Two suggestions adopted: caching the `Resources` module to keep the new per-call lookup off `require`'s resolution path, and applying the same live-resolution to `makeCustomMethodHandler` for symmetry.

Generated with assistance from Claude (Opus 4.8).

### DIFF
--- a/components/mcp/tools/application.ts
+++ b/components/mcp/tools/application.ts
@@ -6,11 +6,19 @@
  *   2. Implements the corresponding verb on its prototype.
  *
  * Tool dispatch delegates to the static `Resource.get/post/put/delete/search`
- * methods, which wrap `transactional()` internally. That means
- * `allowRead/allowCreate/allowUpdate/allowDelete` per-record predicates run
- * unchanged, and restricted-attribute writes are rejected at runtime by
- * `Table.allowUpdate` even when the input schema permits them — defense in
- * depth.
+ * methods, which wrap `transactional()` internally. That entry point runs the
+ * `allowRead/allowCreate/allowUpdate/allowDelete` per-record predicates and
+ * rejects restricted-attribute writes at runtime — the same authorization path
+ * REST flows through.
+ *
+ * For that authorization to be correct the handlers MUST dispatch on the *live*
+ * Resource class resolved from the registry at call time (see `liveResource`),
+ * never a class reference captured when the tools were registered: a component's
+ * `resources.js` subclass — which carries the row-level `allow*` overrides — is
+ * registered after this profile builds its tools and overwrites the base table
+ * entry at the same path. A captured reference would be the base table class,
+ * whose `allow*` only checks table-level RBAC, silently skipping the per-record
+ * guard REST enforces.
  */
 import { createHash } from 'node:crypto';
 import * as env from '../../../utility/environment/environmentManager.ts';
@@ -146,10 +154,17 @@ export function _setRequestTargetForTest(ctor: RequestTargetCtor | undefined): v
 	_requestTargetCtorOverride = ctor;
 }
 
+// Cache the Resources module object (not the `resources` registry itself):
+// `liveResource` calls `loadResources()` on every verb-tool invocation, so we
+// avoid repeating `require`'s path resolution on the hot path. `resources` is a
+// live binding, so reading it off the cached module each call still reflects the
+// current registry.
+let _resourcesModule: { resources?: ResourcesRegistry } | undefined;
 function loadResources(): ResourcesRegistry | undefined {
 	if (_resourcesOverride) return _resourcesOverride;
 	try {
-		return require('../../../resources/Resources').resources as ResourcesRegistry;
+		_resourcesModule ??= require('../../../resources/Resources');
+		return _resourcesModule!.resources as ResourcesRegistry;
 	} catch (err) {
 		harperLogger.trace(`MCP application tools: Resources registry unavailable (${(err as Error).message})`);
 		return undefined;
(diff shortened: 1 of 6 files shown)


dawsontoth (approved):

kylebernhardy:
## Deep Review — row-level RBAC on MCP application verb tools

The dispatch-on-live-class approach is the right fix and is well-covered by the new regression suite. Two findings, both low severity. Notes on what I verified follow.

**Verified (holds up):**
- `liveResource(path, fallback)` reads `loadResources().get(path).Resource` — the same key and property the build loop iterates (`for (const [path, entry] of resources)` → `entry.Resource`). A path-key mismatch between registration and call time is therefore structurally impossible; the lookup queries exactly the key the `resources.js` subclass overwrites.
- The module cache (`_resourcesModule ??= require(...)`) is safe: `resources/Resources.ts` uses `export let resources` (reassigned during init), and `require()` of the TypeStrip ESM module exposes it as a live getter, so `_resourcesModule.resources` reflects reassignment on every call.
- The regression suite uses an independent persistence oracle (admin `search_by_value` read-back) and grants table-level CRUD via a dedicated role, so the per-record guard is genuinely the only barrier under test. Denied create/update/delete are asserted both to error and to leave the store unchanged.

---

### 1. `liveResource` fallback silently degrades to table-level-only RBAC — low

**File:** `components/mcp/tools/application.ts` (`liveResource`)

When `loadResources()?.get(path)` returns `undefined`, the handler dispatches `capturedClass` — the base table class whose `allow*` only checks table-level RBAC. That is the exact fail-open this PR closes. The PR's premise is that registry mutations (subclass registration) do not re-trigger `refreshApplicationTools`; by the same logic a registry removal/re-key need not rebuild tools either, leaving a live tool whose `get(path)` misses and falls back to the weaker-authz base class. Narrow (requires the entry to vanish or re-key while the tool persists and underlying storage still accepts writes), which is why it's low — but it is a fail-*open* default on the security-sensitive write verbs.

**Suggested fix:** For the mutating verbs (`create`/`update`/`delete`), consider failing closed when the live lookup misses — raise rather than dispatch a class with weaker predicates than the one the tool was built from. Read verbs can keep the graceful fallback. At minimum, confirm this trade-off is intentional and log at `warn` when the fallback is taken so a silent downgrade is observable.

### 2. Denied read tests assert non-leakage but not error framing or positive filtering — low

**File:** `integrationTests/mcp/row-authz.test.ts`

`get_Doc on admin row leaks nothing` and `search_Doc does not leak` only assert the response text excludes `admin-original`. They don't assert `isError` on the denied `get`, and `search_Doc` never asserts the low-priv user's *own* row is returned. So the suite can't distinguish "correctly filtered to owned rows" from "errored / returned nothing for everyone" — an over-denial regression (e.g. once the per-row filtering of #1489 lands and changes search from error-out to filter) would pass silently.

**Suggested fix:** Add `ok(result.isError)` to the denied `get_Doc` case, and assert `search_Doc` as the low-priv user returns `lowuser-row` while excluding `admin-original`, so the test pins both halves (no leak *and* no over-denial).

---

**Disclosed in the PR, not re-flagged:** the `search_<T>` collection-scan returning an auth error until #1489/#1422 land (documented "known residual," verified non-leaking), and custom Resource verb methods now running for MCP calls (intended REST parity, with the `application.test.ts` assertion updated to match).

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

Issue #1414, opened 2026-06-19
Live subscriptions (SSE/MQTT/WS) continue delivering events after drop_user / role revocation — stale-auth leak


kriszyp:
## Summary

When a user account is dropped (or a role revoked) while an active subscription is in progress, that subscription continues to deliver events to the now-unauthorized client indefinitely. Subscribe-time authorization is correct; the gap is the **absence of per-event (or periodic) re-authorization** once a subscription is live.

## Confirmed behavior

Tested against all three streaming transports with a non-admin user (`qa_user`):

| Transport | subscribe-time grant | mid-session `drop_user` | continued events? |
|---|---|---|---|
| SSE (GET /T/ `application/x-ndjson`) | correct 200 | user removed | **yes, continues** |
| MQTT (retain-watch) | correct CONNACK | user removed | **yes, continues** |
| WebSocket | correct upgrade | user removed | **yes, continues** |

Control (REST GET after drop): `401 Unauthorized` — confirms the creds are genuinely invalid. The subscription channel ignores the cred state entirely once the handshake is complete.

## Root cause

`DurableSubscriptionsSession.ts:266+` — the delivery loop checks the `permissions` snapshot captured at subscribe-initiation time. No subsequent call to re-validate the principal occurs per-event, on a timer, or on role change. `drop_user` / role revocation has no mechanism to interrupt existing sessions.

## Impact

- **Authorization bypass** after account deletion or role downgrade — events keep flowing until the TCP connection is dropped by the client or network.
- Affects all three live-subscription transports.
- Row-level subscribe filters are also applied against the stale snapshot, so per-row filtering on the now-invalid user's grants can also be stale.
- Severity scales with session lifetime; long-lived MQTT subscribers can continue receiving events for hours after revocation.

## Recommendation

At minimum: periodically re-validate the subscribing principal's permissions against the live auth store (e.g. on each event batch, or on a short timer). Alternatively, maintain a registry of active subscription sessions keyed by user/role so that `drop_user` / `alter_role` can invalidate and close live sessions immediately.

**Harper:** `7aaa5a152`; `DurableSubscriptionsSession.ts` byte-identical to `main` @`6797f091d`.

---
*Surfaced by the QA-explorer exploratory campaign for @kris (QA-163). Filed by Claude (Sonnet 4.6) for @kris.*

kriszyp:
**QA-164 follow-up: the stale-auth leak is universal — every revocation mechanism fails, not just `drop_user`.**

[...]

PR #1524, opened 2026-06-29
fix: enforce row-level allowRead during subscription delivery (#1419)

### LINKED ISSUE #1419: Subscription delivery bypasses row-level `allowRead` — all rows leak to any collection subscriber
## Summary

Row-level `allowRead` authorization is **architecturally not consulted** during subscription event delivery. A user who can subscribe to a collection (`GET /Table/` via SSE, MQTT, or WebSocket) receives **every row-change event for the entire table**, regardless of what `allowRead` returns for each row.

## Proof

Schema: `Vault` table with `allowRead(record, {user}) { return record?.owner === user?.username }`. Alice and Bob are non-admin users, each owning different rows.

**Control (auth works on REST):** Alice's direct REST GET of Bob's row → `403 Forbidden` — confirms `allowRead` is enforced and `AUTHORIZELOCAL` is not escaping (real bearer token auth used).

**Subscription (auth bypassed):** Alice opens `GET /Vault/` SSE with the same credentials. Updates are written to Alice's rows (×6), Bob's rows (×6), and a neutral row (×2). Alice's stream receives **all 14 events** — including every Bob-row and neutral-row event.

```
Owners observed on Alice's stream: ["qa165_alice", "qa165_bob", "nobody"]
Expected:                           ["qa165_alice"]
```

## Root cause

`Table.subscribe` (`resources/Table.ts`) calls `allowRead` once at connect time, but with `this` as the collection and no loaded record — effectively always passing. After the connection is established, the delivery loop in `DurableSubscriptionsSession.ts:266+` calls `send(event)` on every row update using only a VALUE filter from the subscription query — there is no per-record `allowRead` evaluation at delivery time.

## Distinction from #1414 (F-034)

\#1414 is a **timing** bug — the subscribe-time auth snapshot is not invalidated on `drop_user`/revocation. This issue is **architectural** — `allowRead` is never consulted during event delivery even for a continuously-authorized principal. The two issues share the same delivery loop but require independent fixes.

## Impact

**High (security)** — any table using row-level `allowRead` for multi-tenant isolation (the canonical Harper authorization pattern) is fully exfiltrable via a single collection subscription. A non-admin principal with collection-level read access can receive every other tenant's row-change events in real-time, with no indication on the sender side.

Affected transports: SSE (`GET /Table/`), MQTT (wildcard subscribe), WebSocket.

## Recommendation

Re-evaluate `allowRead(record, context)` per-event in the delivery loop before calling `send()`. Filter out events where the subscriber's `allowRead` returns false for the updated record. This is the `allowRead` analog of the existing value-filter already applied per-event.

**Harper:** `7aaa5a152`; `resources/Table.ts` + `server/DurableSubscriptionsSession.ts` byte-identical to `main` @`6797f091d` — reproduces on main.  
**Repro test:** `integrationTests/qa-scratch/qa165-sub-row-filter.test.ts`

---
*Surfaced by the QA-explorer exploratory campaign for @kris (QA-165). Filed by Claude (Sonnet 4.6) for @kris.*

kriszyp:
## Summary
Row-level `allowRead` overrides are enforced on the REST point-read path but were **not** re-evaluated per event during subscription delivery (SSE, WebSocket, MQTT). The connect-time check runs with `this` bound to the collection and no loaded record, so a row-level override always passes — and **every row's changes leaked to any collection subscriber** (a multi-tenant isolation bypass).

This re-evaluates `allowRead` per record-bearing event, binding a resource instance to that row's record (the delivery-path analog of the per-event value filter already applied downstream).

## Scope / gating (kept minimal)
1. Only when the resource **overrides** `allowRead` — default table/attribute RBAC is record-independent and already enforced at connect, so this is zero-overhead for the common case.
2. Only when the subscription has a `context.user` principal — internal watchers (e.g. `hdb_certificate`) and replication have no user and keep delivering every row.

Applied at all delivery sinks: direct `send`, drained-queue `send`, and the `previousCount` history backfill. Denied rows are filtered **before** consuming a `previousCount` slot, so authorized subscribers still receive their full requested history. An async `allowRead` (thenable) **fails closed**.

## Known limitation (deferred follow-up)
`delete` / `message` / `rawEvents` events aren't authorized against the full row, since their event value isn't the full record. Tracked separately.

## Testing
- New `integrationTests/security/subscription-row-allowread` — CONTROL (REST `allowRead` enforced) + PROBE (whole-table SSE delivers only the authorized owner's rows). 2/2 pass.
- `tsc` clean.

## Related
Part of the `allowRead` consolidation (#1487). The REST collection-scan per-record enforcement and the throw→fail-closed hardening (#1422 gaps 1 & 2) live in #1489; this PR is scoped to the subscription-delivery path only and does not overlap its diff.

Closes #1419.

### DIFF
--- a/integrationTests/security/fixtures/subscription-row-allowread/config.yaml
+++ b/integrationTests/security/fixtures/subscription-row-allowread/config.yaml
@@ -1,0 +1,6 @@
+# Regression fixture for #1419 — per-row allowRead must filter subscription delivery.
+graphqlSchema:
+  files: '*.graphql'
+jsResource:
+  files: resources.js
+rest: true
--- a/integrationTests/security/fixtures/subscription-row-allowread/resources.js
+++ b/integrationTests/security/fixtures/subscription-row-allowread/resources.js
@@ -1,0 +1,36 @@
+// Regression fixture for #1419 — per-row allowRead must filter subscription delivery.
+//
+// allowRead is designed so that:
+//   - When `this` is a loaded record (a specific row), allow only if the row's owner
+//     matches the requesting user.
+//   - When `this` is the collection (no specific record loaded, this.owner undefined/null),
+//     allow the subscription to open. This is the permissive side that lets a non-owner open
+//     a whole-table subscription — the scenario where the pre-#1419 delivery loop leaked every
+//     row's events because it never re-evaluated allowRead per record.
+//
+// Super users always pass so that seed writes and setup ops succeed.
+
+function isSuper(user) {
+	return !!user?.role?.permission?.super_user;
+}
+
+export class Vault extends tables.Vault {
+	allowRead(user, _target, _context) {
+		if (isSuper(user)) return true;
+		const owner = this?.owner;
+		// Collection subscribe / no loaded record: allow the connection to open.
+		if (owner === undefined || owner === null) return true;
+		// Per-row: only the owner can read this specific record.
+		return owner === user?.username;
+	}
+
+	allowUpdate(user, _record, _context) {
+		return isSuper(user);
+	}
+	allowCreate(user, _record, _context) {
+		return isSuper(user);
+	}
+	allowDelete(user, _target, _context) {
+		return isSuper(user);
+	}
+}
--- a/integrationTests/security/fixtures/subscription-row-allowread/schema.graphql
+++ b/integrationTests/security/fixtures/subscription-row-allowread/schema.graphql
@@ -1,0 +1,11 @@
+# Regression fixture for #1419 — per-row allowRead must filter subscription delivery.
+#
+# Vault rows are owned by a specific user (`owner` field). The allowRead override in
+# resources.js permits a read only when the record's owner matches the requesting user,
+# while still allowing the collection-level subscription to open. The test verifies that a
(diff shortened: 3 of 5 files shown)


kriszyp:
Closing — this implements per-record `allowRead` evaluation during subscription delivery, which we're deliberately not pursuing.

Design clarification (per @kriszyp): in Harper, RBAC/roles grant the intrinsic, table-level access surfaced through the operations API. Programmatic hooks like `allowRead` (and method overrides) exist to **grant** access on top of that default — they only make sense as a widening mechanism, not as a per-record *restriction* relative to default RBAC, which has no per-record granularity. Per-record `allowRead` filtering on subscription delivery (and on collection scans) was never an advertised capability, so re-evaluating it per event isn't a behavior we want to introduce.

What we do want from this area is captured elsewhere:
- **Fail-closed on throw/reject** for `allow*` hooks → #1422 / #1489.
- **Revoking a live subscription when a user loses permission** (table-level RBAC, periodic re-check) → #1414. This is the real subscription-security concern and is independent of per-record evaluation.

Closing #1419 alongside this PR.

PR #1530, opened 2026-06-30
fix(subscriptions): re-deliver current scope to live subscribers on a copyApply reload marker

kriszyp:
## Summary
A copyApply base copy (harper#1479) writes its rows as snapshots with **no per-row audit entries**, so the live-subscription notify path never fires for them. An already-connected MQTT/SSE/WS subscriber therefore silently **misses records that land on its node via a base copy** — they are durably stored but never delivered. (Confirmed by controlled experiment: the consumer receives only a whole-table `reload` marker, never a per-record `put`; with the re-snapshot disabled the subscriber sees `[]`.)

This is the **core half** of harper-pro#495. It builds on the `reload` marker producer from harper#1491 (already merged): when a whole-table `reload` marker (id=null) is delivered for a **non-system** table, `subscribe()` suppresses the bare marker and instead re-delivers the subscription's current scope as ordinary `put` events, recovering the snapshotted rows.

## Where to look — `resources/Table.ts` `subscribe()`
- **One place covers every protocol.** MQTT, SSE, and WS all funnel through `subscribe()` (SSE/WS via `Resource.connect()` → `this.subscribe()`), so a single producer-side re-snapshot covers them — no per-protocol handling.
- **Gate = `databaseName !== 'system'`.** System-DB reloads pass through verbatim, so knownNodes (peer discovery) and hdb_certificate (CA install) keep their bespoke whole-table rescan — which a per-row `put` re-emit cannot express (it can't drop stale rows). Every internal raw-`reload` consumer is on the system DB, so no per-subscription flag is needed.
- **Off the lock.** The re-snapshot's first statement is `await rest()` (a `setImmediate` macrotask), so only two boolean assignments run inside the synchronous broadcast listener; the scan runs after the listener returns and (on the same-thread aftercommit path) releases its inter-thread lock.
- **Coalesced + back-pressured.** Reuses the same `send()` + `waitForDrain` back-pressure as the initial snapshot; a marker arriving mid-scan re-arms one more pass; a closed subscription bails.

## Companion
- harper-pro: emits the per-table `reload` markers on copy-complete (`emitCopyReloadMarkers`). Dormant on its own — without that emitter no user-table markers are produced, so this change is inert until the harper-pro side lands.

## Verification
- New integration test (harper-pro side) `aclConnectCopyReload.test.mjs`: a live subscriber on a node that base-copies pre-existing rows receives all of them — **fails** with the re-snapshot disabled (true negative control), passes with it.
- Existing `aclConnectCrossNode` cross-node delivery (3 tests): still green.

## Cross-model review
Thorough cross-model review (Codex + Gemini + Harper-domain adjudication): **no blockers** — all four outside-model "blockers" were false positives on direct source trace (the `send` closure, a no-`await`-window coalescing path, the `await rest()` deferral, `snapshot:false` convergence). Domain pass confirmed the `LOCAL_ONLY` marker is skipped by the replication send path and the system-DB gate preserves knownNodes.

### DIFF
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -3230,6 +3230,9 @@ export function makeTable(options) {
 			// snapshot:false, which catches any commits that land during yield points; dropping in the
 			// listener avoids duplicate delivery.
 			let dropDuringReplay = false;
+			// Coalescing guards for the reload re-snapshot (harper-pro#495), driven from the listener below.
+			let reloadResnapshotRunning = false;
+			let reloadResnapshotPending = false;
 			const thisId = requestTargetToId(request) ?? null; // treat undefined and null as the root
 			const subscription = addSubscription(
 				TableResource,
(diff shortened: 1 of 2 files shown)


resources/Table.ts
```diff
@@ -3506,6 +3515,78 @@ export function makeTable(options) {
...
+				} catch (error) {
+					harperLogger.error?.('Error in reload re-snapshot:', error);
+				} finally {
+					reloadResnapshotRunning = false;
+					// A marker that landed after the last pending-check but before we cleared the flag would
+					// otherwise be dropped — re-arm if so (unless the subscription has since closed).
+					if (subscription.subscriptions && reloadResnapshotPending) scheduleReloadResnapshot();
+				}
+			}
+			function scheduleReloadResnapshot() {
```
cb1kenobi, line 3584: I guess I don't really see the point in this function when this could have been inlined at the top of `runReloadResnapshot()`.

resources/Table.ts
```diff
@@ -3506,6 +3515,78 @@ export function makeTable(options) {
...
+							}
+						}
+					}
+				} catch (error) {
+					harperLogger.error?.('Error in reload re-snapshot:', error);
+				} finally {
+					reloadResnapshotRunning = false;
+					// A marker that landed after the last pending-check but before we cleared the flag would
+					// otherwise be dropped — re-arm if so (unless the subscription has since closed).
+					if (subscription.subscriptions && reloadResnapshotPending) scheduleReloadResnapshot();
```
cb1kenobi, line 3581: There might be some weird scenario where the subscriptions never drains and cause a maximum call stack exceeded error. Maybe wrap the reloadResnapshot in a `process.nextTick()`?

cb1kenobi (approved):

PR #1536, opened 2026-06-30
fix: resolve extension packages outside ROOTPATH during node_modules walk (#1507)

kriszyp:
Supersedes #1507 (by @pbrumblay), rebased onto `main` and with regression tests added so it's ready to merge.

## The bug

Extension package resolution fails when an application is run via `RUN_HDB_APP` from **outside** ROOTPATH (e.g. `harper run` from a monorepo's `apps/web`).

`loadComponent` walks up from the app directory looking for `node_modules/<package>`, but the old loop stopped climbing as soon as the parent path's **string length** dropped below `getHdbBasePath().length`. That length heuristic is a proxy for "we've left the install root" — and it breaks whenever the app's checkout path is shorter than ROOTPATH. The walk aborts before reaching packages hoisted to the monorepo root.

This is why the failure looked machine-specific despite identical Node/Harper versions: it depended purely on whether the checkout path happened to be longer or shorter than ROOTPATH.

```
Error: Could not load component '@harperfast/nextjs' for application 'web'
due to: Unable to find package @harperfast/nextjs:@harperfast/nextjs
```
`require.resolve('@harperfast/nextjs', { paths: ['apps/web'] })` succeeds on the same machine — only Harper's loader failed.

## The fix (credit: @pbrumblay)

Replace the string-length heuristic with **path-prefix** logic:

- Apps **outside** ROOTPATH (`RUN_HDB_APP`): walk up to the filesystem root.
- Components **inside** `$ROOTPATH/components/`: stop at the ROOTPATH boundary (preserves the original intent).

Also adds a `dirname()` fixpoint guard (`parentFolder === containerFolder`) that terminates the walk at the filesystem root — the old loop had no fs-root guard and would spin if `getHdbBasePath()` were ever undefined.

## Tests added

`unitTests/components/componentLoaderWalk.test.js` drives the real `loadComponent` walk (`HARPER_SAFE_MODE` isolates path resolution from actually loading a package):

1. **The bug** — outside-ROOTPATH app resolves a monorepo-root hoisted package when the repo path is shorter than ROOTPATH.
2. **Termination** — a genuinely-absent package walks to the fs root and fails gracefully (no infinite loop); covers the new guard.
3. **In-ROOTPATH resolution** still works.
4. **Boundary preserved** — an in-ROOTPATH component does not resolve a package above ROOTPATH.

Verified that reverting to the old heuristic makes test 1 fail while the other three stay green (they characterize behavior the old code also had).

### DIFF
--- a/components/componentLoader.ts
+++ b/components/componentLoader.ts
@@ -9,7 +9,7 @@ import {
 	rmSync,
 	symlinkSync,
 } from 'node:fs';
-import { join, basename, dirname } from 'node:path';
+import { join, basename, dirname, sep } from 'node:path';
 import { isMainThread } from 'node:worker_threads';
 import { parseDocument } from 'yaml';
 import * as env from '../utility/environment/environmentManager.ts';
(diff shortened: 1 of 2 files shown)


pbrumblay:
@kriszyp  - thanks for driving this to the finish line. I'll close out my PR. Looks like you've got it covered.

PR #1540, opened 2026-06-30
fix: atomically swap component dir on deploy to avoid ENOTEMPTY on live apps

kriszyp:
## Summary

`deploy_component` fails on a peer that is currently running the component, ~3s into the replicate phase:

```
ENOTEMPTY: directory not empty, rmdir '/home/harperdb/harper/components/<component>/.next'
```

`extractApplication` (`components/Application.ts`) cleared the existing component directory **in place** — `rm(dirPath, { recursive: true, force: true })` then `mkdir` + extract. When a previous version's worker is still running and writing into that directory (a live Next.js app continuously writing `.next/cache` under traffic), the recursive `rm` races that writer: Node empties `.next`, then its internal leaf `rmdir('.next')` throws `ENOTEMPTY` because the worker just re-created a cache entry. (`force: true` only suppresses `ENOENT`; `ENOTEMPTY` is retried only with `maxRetries > 0`, and a continuously-writing app would outlast retries anyway.)

Matches the field report exactly: only the **web** (Next.js) component fails — agent + schema in the same deploy replicate fine; only the **live peer** fails (origin/just-restarted nodes have no active writer); restarting the node clears it.

## Fix

Replace the in-place destruction with an **atomic swap**: rename the existing directory aside into a hidden, component-scoped staging dir `<componentsRoot>/.deploy-aside/<name>/<pid>-<ts>-<uuid>` (same parent → same filesystem → atomic `rename`) before `mkdir` + extract. The still-running worker keeps writing into the renamed inode harmlessly until it's replaced on restart. The staging subdir is then removed best-effort (fire-and-forget, logged at trace since failure is expected while the worker is live); leftovers are reclaimed by the next deploy.

`loadComponentDirectories` loads every visible directory under the components root as a component, so the staging dir is hidden (dot-prefixed) and the loader now skips dot-prefixed entries; `get_components`' `walkDir` skips the staging dir too.

## Where to look

- The swap block in `extractApplication` — rename-aside replaces the old `access` + in-place `rm` + `mkdir`.
- Best-effort cleanup semantics: failure is tolerated by design (a still-running worker can keep the aside non-empty); the next deploy reclaims it. Sanity-check that this best-effort-with-residue tradeoff is acceptable vs. blocking the deploy on cleanup.

## Cross-model review

Reviewed via the cross-model skill (Gemini + Harper-domain adjudication). No blockers or significant issues in the committed code. The early collision/log-noise findings drove the staging-dir redesign and are resolved.

**Open follow-ups (pre-existing, not introduced here — flagging for a reviewer call on whether to fold in or track separately):**
- The single-top-level-folder normalization right below uses `mkdtemp(application.dirPath)`, creating a **non-hidden** sibling `<componentsRoot>/<name>XXXXXX`. On a crash between `mkdtemp` and its `rm`, that leftover would load as a phantom component. Independent of this race.
- `dropCustomFunctionProject` (`components/operations.js`) deletes a component dir in place with `rmSync(..., { recursive: true })` — same live-writer `ENOTEMPTY` race is possible, though semantics differ (the component is being removed, not swapped). Lower stakes.

## Tests

`unitTests/components/extractApplicationSwap.test.js`: (1) a populated `.next/cache` + an aggressive concurrent writer during `extractApplication` — fails on the old code with the exact `ENOTEMPTY ... rmdir '.../.next/cache'`, passes on the fix, and asserts no non-hidden leftover appears under the components root; (2) a stale aside from an earlier deploy is reclaimed.

🤖 Generated by Claude (Opus 4.8).

### DIFF
--- a/components/Application.ts
+++ b/components/Application.ts
@@ -126,6 +126,11 @@ export function isSSHAuthFailure(stderr: string): boolean {
 	);
 }
 
+// Hidden directory under the components root holding component versions renamed aside
+// during a deploy swap (see extractApplication). The leading dot keeps
+// loadComponentDirectories from loading its contents as components.
+export const ASIDE_STAGING_DIR = '.deploy-aside';
+
 /**
  * Extract an application given payload (content of the application) or package (npm-compatible identifier to the application).
  *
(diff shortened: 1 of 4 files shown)


kriszyp:
Note on the red **Unit Test (Node.js v26)** check: this is v26-runner flakiness, not a regression from this PR.

- Across 4 v26 runs it failed on a **different unrelated test each time** — `crud.test.js` analytics recording, `test REST calls`, then `dropTable ghost regression` + an `lmdb/write.js: Invalid argument`. A real regression fails the *same* test deterministically.
- Those tests live in `unitTests/resources/`, which runs as a **separate mocha process** (`test:unit:resources`) from this PR's `unitTests/components/` test — and none of them import the changed files (`components/Application.ts`, `components/componentLoader.ts`, `components/operations.js`). They can't be affected by this diff.
- **Unit Test v22 + v24, and every Integration suite (v22/v24/v26/Bun/Windows) are green**, as are the Gemini and Claude review passes.

Marking ready for review on that basis; a re-run of the v26 unit job should clear it.

— generated by Claude (Opus 4.8)

PR #1542, opened 2026-06-30
fix(blob): harden FileBackedBlob.stream() cancel/cleanup paths (#1457)

### LINKED ISSUE #1457: Blob stream() reader: pre-existing cancel/cleanup hazards (fd not nulled → double-close, untracked open-retry timer, double-onError, slice reads from byte 0)
Surfaced by the cross-model review of #1454 / PR #1456, in code that PR does not change. Grouping them here so the targeted patch stays focused.

All in `resources/blob.ts` `FileBackedBlob.stream()`:

1. **`fd` not nulled after close → double-close on a reused fd.** `cancel()` (and the success/error paths) call `close(fd)` but never clear `fd`. A later `cancel()` (e.g. GC-driven after a successful read) calls `close()` again on a descriptor the OS may have reassigned to an unrelated file/socket — silent corruption. Null `fd` after every close and guard `close()`/`cancel()` on it.
2. **`cancel()` before `fd` is assigned → `close(undefined)`.** If a consumer cancels before the async `open()` in `start()` resolves, `close(undefined)` throws `ERR_INVALID_ARG_TYPE`. Guard on `fd != null`.
3. **`start()` open-retry 20ms timer is untracked.** The file-creation retry `setTimeout` in `start()` is not saved/cleared; cancelling during the wait leaks an fd (the timer fires, opens a new fd, but `pull()`/`cancel()` never run again). Needs its own tracked handle (it is in a different scope than `pull()`'s `timer`). (PR #1456 already tracks the two `readMore()` backoff timers.)
4. **Double `onError`.** If a slow async `read` completes with an error after the watcher-timeout already called `onError` and closed the fd, `onError` runs twice → second `close(fd)` + duplicated `#onError` callbacks. Add a settled guard.
5. **Slice reads sequentially from byte 0** (perf). `blob.slice(start, end).stream()` / `.bytes()` reads and discards from offset 0 instead of seeking to `HEADER_SIZE + start`, allocating a 256KB buffer per discarded chunk. Seek directly.

#1–#4 are reliability/correctness; #5 is performance. None are regressions from #1456 — they predate it.

kriszyp:
Closes #1457.

Fixes pre-existing reliability/correctness hazards in `resources/blob.ts` `FileBackedBlob.stream()`, surfaced by the cross-model review of #1454 / PR #1456 in code those PRs do not change. None are regressions from #1456 — they predate it.

## What changed

The five hazards listed in #1457:

1. **`fd` not nulled after close → double-close on a reassigned descriptor.** All closes now route through a `closeFd()` helper that guards on `fd != null` and nulls it. A later GC-driven `cancel()` after a successful read can no longer close a descriptor the OS may have reassigned to an unrelated file/socket.
2. **`cancel()` before `open()` resolves → `close(undefined)` throw.** Same `fd != null` guard.
3. **`start()` open-retry 20ms timer was untracked** (different scope than `pull()`'s `timer`), leaking an fd if cancelled during the file-creation wait. Tracked as `openTimer` and cleared in `cancel()`.
4. **Double `onError`** — a slow async read completing with an error after the watcher-timeout path already called `onError()` ran it twice (double close + duplicated `#onError` callbacks). Added a per-pull `settled` guard.
5. **Slice read sequentially from byte 0** — `slice(start,end).stream()` read and discarded a 256KB buffer per chunk from offset 0 instead of seeking to `HEADER_SIZE + start`. Now seeks directly.

### Latent correctness bug fixed alongside #5
The slice trim used the file `position` where it needed the **content** offset (`position - HEADER_SIZE`). Any slice whose bound landed past the first `0x40000` chunk returned bytes shifted by `HEADER_SIZE`. Existing tests only sliced within the first chunk, so it was never exercised. The seek rewrite corrects the accounting; the new cross-chunk slice test fails on the old code.

## Cross-model review caught two blockers (also fixed here)

A first pass relied on a **per-pull `settled` flag**, which `cancel()` (a different scope) can't set. Codex and Gemini both flagged the gap:

- **fd leak** — `cancel()` before `start()`'s `open()` resolved couldn't close the descriptor `open()` later acquired (`pull()`/`cancel()` never run again). A realistic abort-during-open path.
- **cancel race** — a late read callback after cancel could recurse into `read(null,…)` → synchronous `ERR_INVALID_ARG_TYPE` crash, or reject the already-cancelled pull (unhandled rejection).

Both are fixed by a **stream-wide `cancelled` flag** set in `cancel()`, gating the open-retry loop, the `open()` success path (closes an fd acquired post-cancel), `readMore()` entry, and both read callbacks.

## Tests

Added to `unitTests/resources/blob.test.js`:
- **cross-chunk slice** via `stream()` (boundary-straddling, both-bounds-in-later-chunk seek path, third-chunk-to-EOF, within-first-chunk regression) — verified to fail on the old code.
- **cancel mid-read** — no throw / no unhandled rejection.
- **cancel before open() resolves** — fd accounting via `/proc/self/fd`; verified 50/50 fds leaked without the guard, flat with it.

`npm run build` clean; full blob unit file green (58 passing).

### DIFF
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -340,7 +340,20 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
 		let totalContentRead = 0;
 		let watcher: FSWatcher;
 		let timer: NodeJS.Timeout;
+		// The start() open-retry timer lives in a different scope/phase than pull()'s `timer`; track it
+		// separately so a cancel() during the file-creation wait clears it instead of leaking an fd (#1457).
+		let openTimer: NodeJS.Timeout;
 		let isBeingWritten: boolean;
+		// Close the descriptor at most once. Without nulling `fd`, a later close()/cancel() (e.g. a
+		// GC-driven cancel after a successful read) would close a descriptor the OS may have reassigned to
+		// an unrelated file/socket; and a cancel() before start()'s open() resolves would call
+		// close(undefined) and throw. Guarding on `fd != null` fixes both (#1457).
+		const closeFd = () => {
+			if (fd != null) {
+				close(fd);
+				fd = null;
+			}
+		};
 		let previouslyFinishedWriting = false;
 		const blob = this;
 		// Authoritative (uncompressed) content length from the record descriptor, captured before the
@@ -365,10 +378,11 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
 								if (Date.now() < deadline) {
 									logger.debug?.('File does not exist yet, waiting for it to be created', filePath);
 									// the file doesn't exist, so we need to wait for it to be created
-									return setTimeout(() => {
+									openTimer = setTimeout(() => {
 										checkIfIsBeingWritten();
 										openFile(resolve, reject);
 									}, 20).unref();
+									return openTimer;
 								}
 							}
 							// ENOENT: the file is gone. A writer still holding the lock means we timed out waiting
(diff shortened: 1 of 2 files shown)


resources/blob.ts
```diff
@@ -358,17 +375,19 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
 				// lookup) is computed lazily on the first miss, so a healthy read never pays for it.
 				let deadline = 0;
 				const openFile = (resolve: (value: any) => void, reject: (error: Error) => void) => {
+					if (cancelled) return resolve(undefined); // consumer cancelled before/while opening; stop retrying
```
cb1kenobi, line 378: Do we need the `undefined`?

```suggestion
					if (cancelled) return resolve(); // consumer cancelled before/while opening; stop retrying
```

cb1kenobi (approved):

PR #1543, opened 2026-06-30
fix: use getDatabases() in role validation to avoid startup crash

DavidCockerill:
## Problem

Starting Harper with a component that declares a `roles.yaml` granting database/table permissions crashes during startup:

```
[http/8] [error] [central-manager]: Error in async entry handler: TypeError: Cannot read properties of undefined (reading 'data')
    at customValidate (core/validation/role_validation.ts:149)
    at addRoleValidation (core/validation/role_validation.ts:42)
    at addRole (core/security/role.ts:43)
    at ensureRole (core/resources/roles.ts:92)
    at handleFile (core/resources/roles.ts:77)
```

It also escalates to an `unhandledRejection` in the http worker thread.

## Root cause

`customValidate` reads the legacy global `global.hdb_schema` to verify that the databases/tables a role references exist. That global is populated lazily by `setSchemaDataToGlobal()` (operations-server `setUp`, schema-mutating ops). When a component's `roles.yaml` is loaded during startup on the http worker threads, the global hasn't been set yet, so it's `undefined`. A role granting permissions on a real database (e.g. `data`) then evaluates `undefined['data']` and throws.

## Fix

Use `getDatabases()` directly instead of the lazily-set global. `getDatabases()` is self-initializing and is exactly what `global.hdb_schema` is assigned from, so it works both at startup and afterward. Typed `any` locally to preserve the previous untyped indexing behavior.

## Verification

Reproduced locally with a central-manager component whose `read_only` role grants table permissions on `data` (14 errors per boot). After the fix + rebuild + restart: **0** errors, roles load cleanly, `Harper successfully started.`

### DIFF
--- a/validation/role_validation.ts
+++ b/validation/role_validation.ts
@@ -3,6 +3,7 @@ const validator = require('./validationWrapper');
 import * as terms from '../utility/hdbTerms.ts';
 import { validateOperations } from '../utility/operationPermissions.ts';
 import { handleHDBError, hdbErrors } from '../utility/errors/hdbError.ts';
+import { getDatabases } from '../resources/databases.ts';
 
 const { HDB_ERROR_MSGS, HTTP_STATUS_CODES } = hdbErrors;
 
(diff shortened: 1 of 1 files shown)


kriszyp (approved):
Approving — the fix is correct and minimal: `getDatabases()` is the exact source `global.hdb_schema` is assigned from and is memoized/self-initializing, so it works before the global is set — eliminating the boot-time `undefined['data']` crash, with no change to validation semantics. The one red CI check is an unrelated flake (`doorlock schema missing`, Integration 3/6 Node24, in a file this PR doesn't touch; green everywhere else and flaking on main too), so it's not blocking. Thanks @DavidCockerill!

Issue #612, opened 2026-05-19
Add agent-loop orchestration / `toolMode: 'auto'` to `scope.models`


heskew:
# Add agent-loop orchestration / `toolMode: 'auto'` to `scope.models`

## Context

The model-access API (#510) declares `toolMode: 'return' | 'auto'` on `GenerateOpts`. `'return'` mode is trivial (backend returns tool-call requests; caller resolves them externally). `'auto'` mode is the in-process agent loop — `scope.models` resolves model-issued tool calls against `scope.resources` and registered MCP tools, executes them, and re-invokes the backend until the model produces a terminal answer.

That loop is substantial enough to deserve its own issue: tool resolution conventions, MCP integration surface, streaming behavior across tool-call boundaries, audit, error handling, and safety limits all need design work. Implementing it inside HarperFast/harper#510 would gate the model-access primitives behind 1-2 weeks of agent-loop design. This issue carries that work separately.

Once shipped, `toolMode: 'auto'` is what unlocks Harper's compressed-stack agent story: tool calls resolve in-process against local Resources, audited via the standard transaction log. The **loop itself** is the same regardless of target; the **invocation latency** depends on the tool target — sub-millisecond for Resource-backed tools (in-process dispatch), normal HTTP for tools the unified registry resolves from external MCP servers. The compressed-stack win applies to the Resource-backed case.

## API contract from HarperFast/harper#510 that this fulfills

```ts
type GenerateOpts = {
  tools?: ToolDef[];
  toolMode?: 'return' | 'auto';
  // ... other fields from HarperFast/harper#510
};
```

`'return'` mode lands with HarperFast/harper#510 itself. `'auto'` is type-declared by HarperFast/harper#510 but throws "not yet implemented" until this issue ships.

## Proposed semantics

The auto loop:

1. Caller invokes `scope.models.generate(input, { tools, toolMode: 'auto', ...opts })`.
2. Backend returns a result.
3. If the result has no tool calls, return to caller. Done.
4. If tool calls are present: **validate each call's arguments against the tool's JSON Schema** (`opts.toolArgValidation: 'strict' | 'lenient' | 'none'`, default `'strict'` — models hallucinate JSON shape often enough that this catches real bugs cheaply). Resolve each call against `scope.resources` (or the unified MCP tool registry), execute, capture results.
5. Append the assistant's tool-call message and the tool results to the message list.
6. Re-invoke the backend with the updated message list.
[...]

PR #1545, opened 2026-07-01
feat(agent): consume MCP tool registry (Operations profile), RBAC-filtered (#626)

kriszyp:
## Summary

Wires the unified MCP tool registry's **Operations profile** (#617) into the built-in Harper agent (#626), fulfilling the tool-composition design: the agent consumes the *same* registry the MCP server exposes, RBAC-filtered for its configured user. This re-does the intent of the earlier HarperFast/harper#893 against current `main` (that branch was ~465 commits stale and used the pre-merge registry API).

Scope is deliberately **Operations-profile only** — the built-in agent runs on the main thread adjacent to the operations API server. The Application profile (#618, per-Resource tools, worker-thread-shaped) is out of scope by design.

## What it does

- **`agent/registryTools.ts`** (new)
  - `ensureOperationsToolsRegistered()` — populates the Operations profile on the main thread. Idempotent (`addTool` is `Map.set`-backed), so the agent gets these tools whether or not the operator enabled the `mcp:` HTTP surface.
  - `composeRegistryTools(listingUser, resolveIdentity)` — snapshots the Operations profile, filters by `visibleTo(listingUser)`, and adapts each `ToolDef → AgentTool`: `inputSchema → parameters`, `annotations.destructiveHint →` the loop's approval gate, and a handler wrapping the registry handler (`ToolResult.structuredContent`/text as the observation; `isError → throw`, so the loop records a recoverable failure rather than aborting the run).
- **`agent/agent.ts`** — resolves the agent identity and threads registry tools into `composeToolset`.
- **`agent/toolset.ts`** — merges operator-only + registry tools; **operator-only tools win on a name collision** (a private tool is never shadowed).
- **`unitTests/agent/registryTools.test.js`** — 12 tests: RBAC filtering, shape adaptation, destructive gating, result unwrap/throw, per-call re-resolution, fail-closed propagation, and the collision/allowDestructive merge rules. No server boot, no LLM credits.

The six agent operations (`agent_prompt`, `get_agent_session`, `list_agent_sessions`, `cancel_agent_run`, `approve_agent_action`, `set_agent_config`) already landed with the scaffold (#839); this PR adds the tool surface they drive.

## Security model

`visibleTo` controls **listing only** (which tools the LLM is shown) — it is **not** a boundary. Real enforcement runs per-call in the operation handler via `hdb_user` set to the agent's identity, gated by Harper's existing `verifyPerms` in the operations runtime.

## Cross-model review (Codex + Gemini + Harper-domain)

Two authorization defects were caught and **fixed at root** in the second commit:

- **Blocker — fail-open super_user.** The identity fallback previously fabricated `{ super_user: true }` whenever `agent.user` couldn't be resolved, silently escalating a misconfigured/transient restricted service account to admin. Now the super_user fallback is bounded to the **default `hdb_agent` bootstrap user** (whose provisioning HarperFast/harper-pro#676 defers); an explicitly-configured user that won't resolve **fails closed** (agent runs with only operator-only tools).
- **Significant — stale cached identity.** The agent user was resolved once at startup and closed over in every handler, so a role revocation wasn't honored until restart. Identity is now resolved **per call** (`resolveIdentity` thunk), mirroring how the MCP HTTP path re-auths per request. The startup snapshot is used only for `visibleTo` listing.
- Adjudicated out: Gemini's "`setConfig` ignores `agent.user` changes" — `set_agent_config` doesn't accept `user`, so it can't change at runtime.

## Known follow-ups (not in this PR)

- **Provision the `hdb_agent` system user at startup** so the default agent resolves to a real user instead of relying on the bootstrap fallback. HarperFast/harper-pro#676 explicitly defers this; the fail-closed policy above is the interim guard.
- `agent.maxCostUsd` remains advertised-but-not-enforced (depends on HarperFast/harper#612 telemetry).

## Verification

- `npm run build` clean.
- Full agent unit suite (`unitTests/agent/**`): **51 passing**.

### DIFF
--- a/agent/agent.ts
+++ b/agent/agent.ts
@@ -8,16 +8,18 @@
  * is realized lazily on first use, and the loop runs in-process.
  *
  * The component intentionally avoids `handleApplication`: it has nothing
- * worker-thread-shaped to do. Operator-only tools (FS, schedule, fetch) are
- * inline; registry-backed tools (#615/#617/#618) will fold in via toolset.ts
- * once those land.
+ * worker-thread-shaped to do. Two tool sources compose: operator-only tools
+ * (FS, schedule, fetch) that are inline, and RBAC-filtered registry tools
+ * (#615/#617) drained for the agent's configured user via `registryTools.ts`.
  */
 
 import { dirname, isAbsolute, resolve as resolvePath } from 'node:path';
 import { CONFIG_PARAMS } from '../utility/hdbTerms.ts';
 import harperLogger from '../utility/logging/harper_logger.ts';
 import { Models } from '../resources/models/Models.ts';
+import type { AuthedUser } from '../components/mcp/toolRegistry.ts';
 import { composeToolset } from './toolset.ts';
+import { composeRegistryTools, ensureOperationsToolsRegistered } from './registryTools.ts';
 import { buildOperations } from './operations.ts';
 import { runAgent, _resetInFlightForTests } from './loop.ts';
 import { appendMessage, getSession } from './session.ts';
@@ -37,6 +39,8 @@ const DEFAULT_CONFIG: AgentConfig = {
 interface StartOpts {
 	server: {
 		registerOperation: (def: { name: string; execute: (op: any) => any | Promise<any> }) => void;
+		/** Resolve a Harper user (with role/permissions) by username. Password/request unused here. */
+		getUser?: (username: string, password: string | null, request: unknown) => Promise<AuthedUser> | AuthedUser;
 	};
 	// Component-level config plumbed by componentLoader (`...componentConfig`).
 	enabled?: boolean;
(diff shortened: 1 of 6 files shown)


agent/agent.ts
```diff
@@ -37,6 +39,8 @@ const DEFAULT_CONFIG: AgentConfig = {
 interface StartOpts {
 	server: {
 		registerOperation: (def: { name: string; execute: (op: any) => any | Promise<any> }) => void;
+		/** Resolve a Harper user (with role/permissions) by username. Password/request unused here. */
+		getUser?: (username: string, password: string | null, request: unknown) => Promise<AuthedUser> | AuthedUser;
```
dawsontoth, line 47: Should this be token aware too? And if so, having a third parameter with an unknown type seems rather dangerous.

dawsontoth: From the other PR https://github.com/HarperFast/harper/pull/1546

dawsontoth (approved):

Issue #1544, opened 2026-07-01
Auth: support token login in core — exchange a validated JWT for an httpOnly `hdb-session` cookie


heskew:
## Context

Studio wants direct (non-proxied) connections to instances for SSE / streaming / monitoring — e.g. the deploy component's SSE, which prompted this (studio#1398). Native `EventSource` can't send an `Authorization` header, so **cookie auth is the natural fit for direct-connect SSE**.

The desired flow:
1. Get a short-lived JWT via the Fabric Connect proxy (no user/pass required)
2. Exchange it for an httpOnly `hdb-session` cookie on a direct connection to the instance
3. Drop the JWT; use the cookie for direct connect

## What core already has (5.x)

- **Bearer-JWT auth** — `validateOperationToken()` (`security/auth.ts:201`, `security/tokenAuthentication.ts:192`)
- **httpOnly `hdb-session` cookies** — `SameSite=None; Secure`, origin-prefixed name, stored in `system.hdb_session` (`security/auth.ts:96`, `:258`)
- **`login` / `logout` ops and `request.login()`** — but login only accepts username/password (`security/auth.ts:318`, `:371`)

So core already does JWT *auth* and httpOnly *sessions*. The only missing piece is the **bridge**: nothing lets a validated JWT mint a session cookie.

## Proposed change

Extend the `login` operation (`security/auth.ts:371`) / `request.login()` (`security/auth.ts:318`) to accept a token in addition to username/password:

- validate via `validateOperationToken()`
- on success, call the existing `request.session.update({ user })` to set the `hdb-session` cookie

No new cookie machinery or operation registration needed — `login` is already in the operation map (`server/serverHelpers/serverUtilities.ts`).

## Design decisions to nail

1. **Purpose-scoped exchange token (recommended).** Rather than reusing a full operation token (a complete API credential for its whole lifetime), mint a `subject: 'login'`-typed token that the Bearer-API path *rejects* and only the login endpoint accepts. The typed-subject machinery already exists — `validateToken` keys on `subject` and already rejects role-bearing tokens (`tokenAuthentication.ts:210`). This keeps the cookie-exchange ticket from doubling as a leakable API credential.
2. **Short expiry is already supported.** `expires_in` on `create_authentication_tokens` (`tokenAuthentication.ts:93`, `:132`) covers the "30s → inert in the audit log" idea — no new config.
[...]

PR #1546, opened 2026-07-01
Auth: support token login in core — exchange a validated JWT for an httpOnly hdb-session cookie

kriszyp:
## Summary

Implements #1544. Studio wants direct (non-proxied) connections to instances for SSE/streaming (studio#1398) — native `EventSource` can't send an `Authorization` header, so cookie auth is the fit. This gives core the missing bridge: turn a validated JWT into an httpOnly `hdb-session` cookie.

- `request.login()` / the `login` operation now accept a `token` alongside username/password. When given, it's validated and the resulting user is written into the session (existing `request.session.update()` — no new cookie machinery).
- `create_authentication_tokens` / `createTokens()` gains `purpose: 'login'`: mints a single short-lived (`1m` default, override via `expires_in`) token instead of an operation+refresh pair, and skips the refresh-token DB write / user-change broadcast that pair triggers.

## Design decision taken

The issue flagged one open call: reuse a full operation token for the exchange, vs. mint a **purpose-scoped token** rejected by the Bearer-API path. I went with the recommended purpose-scoped option — the login token is signed with `subject: 'login'` instead of `'operation'`, so `validateOperationToken`'s `jwt.verify(..., { subject: 'operation' })` rejects it on a `sub` mismatch with no extra code, and `validateLoginToken` is the only thing that accepts it. This means the exchange ticket can't be replayed as a general API credential even if it leaked (e.g. in a log line) — confirmed in the new tests below (log lines literally read `jwt subject invalid`).

CSRF posture (design point 3) is unchanged: the cookie is still origin-prefixed per the existing mechanism, regardless of how the session was established.

## Testing

- `unitTests/security/tokenAuthentication.test.js` — 8 new cases: `purpose: 'login'` mints a login-only token with no refresh-token side effects; `validateLoginToken` happy/inactive-user/non-existent-user/bad/expired paths; and the two cross-subject rejection cases (operation token rejected by `validateLoginToken`, login token rejected by `validateOperationToken`).
- `integrationTests/apiTests/authentication.test.mjs` — 2 new end-to-end cases: mint a login token → `login` op → session cookie set; and a login token sent as `Authorization: Bearer` → 401.
- Ran locally: `tsc --noEmit` clean, full build clean, both touched unit-test files green (39/39, 31/31), both touched integration suites green (15/15, 9/9).

### DIFF
--- a/security/tokenAuthentication.ts
+++ b/security/tokenAuthentication.ts
@@ -25,11 +25,20 @@ env.initSync();
 type StringValue = SignOptions['expiresIn'];
 const OPERATION_TOKEN_TIMEOUT: StringValue = env.get(CONFIG_PARAMS.AUTHENTICATION_OPERATIONTOKENTIMEOUT) || '1d';
 const REFRESH_TOKEN_TIMEOUT: StringValue = env.get(CONFIG_PARAMS.AUTHENTICATION_REFRESHTOKENTIMEOUT) || '30d';
+// Default lifetime of a login-purpose token (see TOKEN_TYPE.LOGIN below). It only exists to be
+// exchanged for a session cookie, so it defaults far shorter than an operation token; callers can
+// still override via expires_in.
+const LOGIN_TOKEN_TIMEOUT: StringValue = '1m';
 const RSA_ALGORITHM: Algorithm = 'RS256';
 
 const TOKEN_TYPE = {
 	OPERATION: 'operation',
 	REFRESH: 'refresh',
+	// Purpose-scoped exchange token minted by createTokens({ purpose: 'login' }) and accepted only
+	// by validateLoginToken (the `login` operation). Its `sub` claim differs from
+	// TOKEN_TYPE.OPERATION, so validateOperationToken's Bearer-API path rejects it automatically —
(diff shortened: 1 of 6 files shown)


security/tokenAuthentication.ts
```diff
@@ -45,15 +54,19 @@ interface AuthObject {
...
+	// see TOKEN_TYPE.LOGIN.
+	purpose?: 'login';
 }
 
 interface TokenObject {
 	refresh_token: string;
 }
 
 interface JWTTokens {
-	operation_token: string;
```
dawsontoth, line 55: If we do the login type, can we use the operation token response field instead of making a new field? That keeps the payload at its same format.

dawsontoth (approved):

security/tokenAuthentication.ts
kriszyp: Good call — done in 7d64fce8d. `purpose: 'login'` now returns the login-scoped token in `operation_token` too, so the response shape is identical regardless of purpose.

— KrAIs

dawsontoth:
Saved https://github.com/HarperFast/studio/issues/1408 so we can integrate this into Studio when supported.

PR #1547, opened 2026-07-01
feat(agent): V8 inspector (CDP) tools for debugging/profiling workers (#626)

kriszyp:
## Summary

Adds the **operator-only V8 inspector (CDP) tools** from the HarperFast/harper-pro#676 design — the last operator tool not yet built. The built-in agent runs on the **main thread**; these tools attach over the Chrome DevTools Protocol to **worker** thread inspector ports so it can evaluate expressions, set breakpoints/logpoints, and CPU-profile a worker **without stalling the thread it runs on**.

> **Stacked on HarperFast/harper#1545** (registry tools). Review/merge that first; this diff is just the inspector surface. Base branch: `kris/agent-registry-tools`.

## Tools (`agent/tools/inspectorTool.ts`)

| Tool | Purpose | Gate |
|---|---|---|
| `inspector_attach` | Attach + enable Runtime/Debugger on a worker | — |
| `inspector_evaluate` | Evaluate JS in a worker, return the value | **destructive** (arbitrary in-worker code) |
| `inspector_set_breakpoint` | Breakpoint by URL+line (0-based); a hit pauses the worker | **destructive** (can pause a live worker) |
| `inspector_set_logpoint` | Non-pausing logpoint (breakpoint whose condition `console.log`s and returns `false`) | — |
| `inspector_profile_cpu` | Record a CPU profile for `durationMs`, return hottest functions by self time | — |

- **CDP client**: a minimal id-correlated request/response client over `ws` (already a direct dep), one reused connection per worker debug port, evicted on socket close.
- **Port scheme**: worker debug port = `threads_debug_startingPort + workerIndex`, mirroring `server/threads/threadServer.js`. Attaching requires `threads_debug: true` and a configured `threads_debug_startingPort`.
- **Profile size**: the raw CDP profile is megabytes; `summarizeProfile` reduces it to the top-N functions by self time so the LLM observation stays small.
- **Deps injected** (debug config + live worker count) so the tool is unit-testable without a server boot.

## Safety envelope (`resolvePort`)

The security-relevant part. Every tool call must clear:
- `threads_debug` enabled **and** `threads_debug_startingPort` configured (else a clear error telling the operator what to set);
- `workerIndex` an integer, in range against the **live** worker pool;
- **`workerIndex >= 0`** — the main thread is where the agent itself runs; a self-attach + breakpoint would deadlock it. This is the operator-agent counterpart to the app-developer agent (#612), which runs on a worker and must never attach to itself.

`evaluate` and `set_breakpoint` are marked **destructive**, so they route through the loop's approval gate unless `autoApprove` is set.

## Tests (`unitTests/agent/inspectorTool.test.js`)

- **Safety envelope** — debug-disabled, no starting port, main-thread rejection, out-of-range index, destructive flags. Pure, no network.
- **`summarizeProfile`** — self-time ranking + topN cap on a synthetic profile.
- **Live CDP round-trip** — opens a *real* inspector on an ephemeral port and drives `attach` + `evaluate` (`40+2 → 42`, exception → thrown) + `profile_cpu` against it. This process stands in for a worker (same protocol); it's never breakpointed (that would pause the test — the very reason main-thread attach is banned).

Full agent unit suite: **60 passing**. `npm run build` clean.

## Operator setup note (for docs)

To let the agent debug workers, operators set `threads_debug: true` and a sensible `threads_debug_startingPort` before asking the agent to attach.

### DIFF
--- a/agent/agent.ts
+++ b/agent/agent.ts
@@ -9,16 +9,20 @@
  *
  * The component intentionally avoids `handleApplication`: it has nothing
  * worker-thread-shaped to do. Two tool sources compose: operator-only tools
- * (FS, schedule, fetch) that are inline, and RBAC-filtered registry tools
- * (#615/#617) drained for the agent's configured user via `registryTools.ts`.
+ * (FS, schedule, fetch, and the V8 inspector) that are inline, and RBAC-filtered
+ * registry tools (#615/#617) drained for the agent's configured user via
+ * `registryTools.ts`.
  */
 
 import { dirname, isAbsolute, resolve as resolvePath } from 'node:path';
 import { CONFIG_PARAMS } from '../utility/hdbTerms.ts';
+import * as env from '../utility/environment/environmentManager.ts';
 import harperLogger from '../utility/logging/harper_logger.ts';
 import { Models } from '../resources/models/Models.ts';
 import type { AuthedUser } from '../components/mcp/toolRegistry.ts';
+import { workers } from '../server/threads/manageThreads.js';
 import { composeToolset } from './toolset.ts';
+import { buildInspectorTools } from './tools/inspectorTool.ts';
 import { composeRegistryTools, ensureOperationsToolsRegistered } from './registryTools.ts';
 import { buildOperations } from './operations.ts';
 import { runAgent, _resetInFlightForTests } from './loop.ts';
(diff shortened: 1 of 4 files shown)


dawsontoth (approved):

dawsontoth:
```
  1) agent/inspectorTool — live CDP round-trip
       "before all" hook for "attaches and evaluates an expression in the target":
     Error [ERR_INSPECTOR_ALREADY_ACTIVATED]: Inspector is already activated. Close it with inspector.close() before activating it again.
      at Object.inspectorOpen [as open] (node:inspector:174:11)
      at Context.<anonymous> (unitTests/agent/inspectorTool.test.js:97:13)
      at process.processImmediate (node:internal/timers:484:21)
      ```

kriszyp:
Cross-model review (Gemini + Codex + Harper-domain) run before marking ready. Gemini surfaced real CDP lifecycle/concurrency hazards — all fixed in the latest commit (see `fix(agent): harden inspector CDP client`):

- **Blockers:** `openCdp` hang on pre-open close; session-cache eviction race + concurrent-open dupes; breakpoint could wedge a worker (now auto-resume + stack-snapshot log); logpoint condition-string injection (now JSON-encoded + eval'd, marked destructive).
- **Significant:** per-call abort signal + 30s timeout on CDP calls; strict `workerIndex` parsing (no `""`/`null`→0 coercion); teardown handles in-flight opens.
- **Suggestions:** `Profiler.disable` on cleanup; port>65535 guard.

Codex leg returned no structured findings. 61 agent unit tests green (incl. a live CDP round-trip).

PR #1549, opened 2026-07-01
fix(agent): make the built-in agent runnable end-to-end (#626)

kriszyp:
## Summary

The built-in agent scaffold (#839) is structurally complete but **did not run** against a live Harper instance. This PR fixes four defects — **all found by driving the agent end-to-end with Gemini** (via the OpenAI-compatible endpoint), after which it successfully self-built a Product component and served full REST CRUD.

Independent of #1545 (registry tools) and #1547 (inspector tools) — this fixes the merged scaffold itself. All three touch `agent.ts`, so expect a small merge overlap there.

## Fixes

**1. Session persistence — the critical one.** Rows were written via raw `primaryStore.put(id, row)`, producing a **non-versioned (prefix-less)** record. When such a classic-shared-structures record begins with msgpackr classic **record-id #2 (byte `0x42` == 66)**, the RocksDB decode heuristic (`RecordEncoder.ts:328-335`) misreads it as a local-timestamp-prefixed record, strips 8 bytes, and the decode corrupts → `Could not find typed structure N` → `null` → **"Unknown session"**. The whole agent was dead here.

> Notably this is **not** typed-structure/structon divergence (#1163) and **not** a dep-version issue — `randomAccessFields` defaults false (classic msgpackr structures), and it reproduces same-run on rocksdb-js 2.3.0. It's purely the `0x42`-vs-timestamp decode ambiguity that only bites **prefix-less** records.

**Fix:** write through the Resource-level `.put(record)` (transactional), which stages the version/local-timestamp metadata so the record carries a `<32` prefix byte and decodes unambiguously — the same pattern `DurableSubscriptionsSession` uses. Reads stay on `primaryStore.get` (a versioned record's prefix decodes fine).

**2. Frozen-record mutation.** The store returns a frozen record, so the read-modify-write mutators threw `Cannot assign to read only property 'status'`. `requireSession` now returns a `structuredClone`. The cloning **test-mock masked this** — the mocks now return **frozen** records so it can't regress.

**3. FS path scope.** `write_file`/`read_file` demanded an absolute path the model can't know, so Gemini invented literal `componentsRoot/…` prefixes that fell outside scope. Relative paths now resolve against `componentsRoot`.

**4. No system prompt.** The agent had zero grounding. Added `buildSystemPrompt(scopes)` describing its `componentsRoot`/`logDir`, tool surface, and Harper-app shape — after which the model addresses paths correctly.

## Verification

- Live: Gemini-driven `agent_prompt` → self-built `product_app/schema.graphql` (`@table @export`) → `PUT /Product/1` 204 → `GET /Product/1` returns the record. End-to-end CRUD on an agent-authored app.
- Agent unit suite green; test mocks now return frozen records and exercise the Resource-level `put`.


agent/agent.ts
```diff
@@ -161,6 +162,27 @@ export async function startOnMainThread(opts: StartOpts): Promise<void> {
...
 
+/**
+ * Grounding prompt so the agent knows what it is, where its files live, and how a Harper app is
+ * shaped. Without it the model can't know the absolute componentsRoot and guesses literal path
+ * prefixes (e.g. "componentsRoot/…") that fall outside the write scope.
+ */
+function buildSystemPrompt(scopes: AgentScopes): string {
+	return [
+		'You are the built-in Harper agent, running on the main thread inside a live Harper server.',
+		'You operate this instance for an operator via tools: Harper operations (describe_all, search, deploy_component, restart, set_configuration, add_role, …), scoped filesystem tools (read_file, write_file, list_dir, grep_files, tail_file), schedule_followup, and http_fetch.',
```
dawsontoth: I don't usually have to inject in tool names, that's usually a part of the metadata the agent SDK inserts. Forgetting these are here and adding or removing tools can very often lead to later hallucinations and confusion, for what it's worth.

agent/agent.ts
```diff
@@ -161,6 +162,27 @@ export async function startOnMainThread(opts: StartOpts): Promise<void> {
...
+		'',
+		'Filesystem:',
+		`- Components (apps) directory — your WRITE scope: ${scopes.componentsRoot}`,
+		`- Log directory (read-only): ${scopes.logDir}`,
+		`- Config directory (read-only): ${scopes.configDir}`,
+		'- Pass write_file/read_file paths RELATIVE to the components directory (e.g. "my_app/schema.graphql"), or an absolute path within a scope. Do NOT prefix paths with "componentsRoot".',
```
dawsontoth: This is usually a symptom of an opportunity to either tweak the parameters of the tool or to add more robust graceful parameter parsing. "Do NOT" sometimes works, but more often causes other degraded behavior. In this case it could be having one parameter for the root and it be an enum of the acceptable components, and then a path relative to that.

agent/agent.ts
```diff
@@ -161,6 +162,27 @@ export async function startOnMainThread(opts: StartOpts): Promise<void> {
...
+		'',
+		'A Harper app is a component directory under the components dir. Define tables/resources in a schema (GraphQL `.graphql` with `@table`/`@export`, or `config.yaml` + resource files). After writing or changing component files, deploy/restart as needed for them to load, then verify by querying the REST endpoint via http_fetch against this server.',
```
dawsontoth: Feels like we've written skills for this.

dawsontoth: The skills npm module exports a nice text string built from the agents md for this purpose

dawsontoth: ... line 6 of https://github.com/HarperFast/central-manager/blob/main/src/resources/chat/messages.js (at a swim meet for my daughter)

agent/agent.ts
kriszyp: Good call — dropped the tool-name enumeration from the system prompt. It now describes the tool surface by category and defers to the SDK-provided tool schemas as the source of truth, so it can't drift as tools are added/removed. Thanks for flagging the hallucination risk.

— KrAIs (Claude Opus 4.8)

agent/agent.ts
kriszyp: Took the root-cause path you suggested. The fs tools now take a `root` enum (`components` | `logs` | `config`, default `components`) plus a `path` relative to it, and `write_file` is components-only. That let me delete the `Do NOT prefix…` instruction entirely — exactly the symptom you called out — and gives a single unambiguous resolution (absolute paths are now rejected outright). Enforcement (`isInside`/symlink/cross-drive) is unchanged, just anchored to the one named scope. Appreciate the nudge toward fixing the tool contract instead of patching the prompt.

— KrAIs (Claude Opus 4.8)

agent/agent.ts
kriszyp: Agreed — and this landed via #1560: `agent/bestPractices.ts` now sources the `harper-best-practices` skill directly from the `@harperfast/skills` package exports (`skillSummary` → system-prompt overview, `rules` → on-demand `harper_best_practice` tool), so it's versioned with the release with no drift. Thanks for the central-manager pointer — same idea.

— KrAIs (Claude Opus 4.8)

agent/agent.ts
```diff
@@ -158,9 +219,89 @@ export async function startOnMainThread(opts: StartOpts): Promise<void> {
...
+ * Failure policy — fail closed, with one bounded exception:
+ *   - If `agent.user` resolves to a permissioned user, use it.
+ *   - If it can't be resolved AND it's the *default* `hdb_agent` bootstrap user,
+ *     fall back to a super_user identity (the system user isn't provisioned yet —
+ *     #626 defers creating it). This is the documented default-agent behavior.
+ *   - If it can't be resolved and the operator configured a *non-default* user,
+ *     throw. Silently escalating a misconfigured/transient restricted account to
+ *     super_user is the vulnerability we refuse to introduce.
+ */
+async function resolveAgentIdentity(server: StartOpts['server'], username: string): Promise<AuthedUser> {
```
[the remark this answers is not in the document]
kriszyp: Addressed in 0febc774c: exported `resolveAgentIdentity` and added `unitTests/agent/agent.test.js` with direct coverage of all three branches — resolvable permissioned user returned as-is; default `hdb_agent` unresolvable → super_user bootstrap; non-default unresolvable → throws (fail closed), plus the getUser-throws and missing-role-permission edges. 9 passing.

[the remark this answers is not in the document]
kriszyp: Addressed in 0febc774c: exported `resolveAgentIdentity` and added `unitTests/agent/agent.test.js` with direct coverage of all three branches — resolvable permissioned user returned as-is; default `hdb_agent` unresolvable → super_user bootstrap; non-default unresolvable → throws (fail closed), plus the getUser-throws and missing-role-permission edges. 9 passing.

agent/mcpTools.ts
```diff
@@ -0,0 +1,140 @@
...
+			// Listing is restricted to super_users; the operation's own super_user check enforces at call time.
+			visibleTo: (user: AuthedUser) => isSuperUser(user),
+			handler: async (args: unknown, context: { user: AuthedUser }): Promise<ToolResult> => {
+				try {
+					const body = {
+						...(args && typeof args === 'object' ? (args as Record<string, unknown>) : {}),
+						operation: name,
+						hdb_user: context.user,
+					};
+					const data = await op.execute(body);
```
cb1kenobi: **High: curated agent MCP tools bypass the `requiresSuperUser` gate — any authenticated user can drive the super_user agent**

This handler dispatches straight to `op.execute(body)`, which never runs `verifyPerms`, so the `requiresSuperUser: true` declared on the agent operations (`operations.ts`) is not enforced here. `visibleTo: isSuperUser` (line 107) only gates `tools/list` — the registry model is explicit that it is *not* a call-time boundary (`toolRegistry.ts:16-26`), and `dispatchToolsCall` resolves the tool by name and invokes `tool.handler` with no authz check (`transport.ts:508-624`). Because the operations MCP endpoint authenticates *any* Harper user (`authAndEnsureUserOnRequest`, `operationsServer.ts:222`), a non-super_user can `tools/call` `agent_prompt` directly (the name isn't secret, it's just hidden from their listing) and start a run. The run then executes under `resolveAgentIdentity` — the default `hdb_agent` resolves to the super_user bootstrap identity — so this is privilege escalation to super_user plus arbitrary component-file writes and `inspector_evaluate` code execution. `get_agent_session` / `list_agent_sessions` likewise leak session transcripts to non-super_users.

Contrast the generic operations provider (`tools/operations.ts`), whose handler dispatches through `chooseOperation` + `processLocalTransaction` so `verifyPerms` runs — that is the enforcement the registry's security model assumes every handler provides.

Suggested fix — enforce at call time before dispatching (add a guard at the top of the handler, before the `try`):

```ts
if (!isSuperUser(context.user)) {
  return { isError: true, content: [{ type: 'text',
    text: JSON.stringify({ error: `super_user required for agent op '${name}'` }) }] };
}
```

or route dispatch through the same `processLocalTransaction` path the generic operations tools use, so `requiresSuperUser` is enforced uniformly.

—
_Generated by Barber AI_

kriszyp: Fixed in 3019c214b. Confirmed the bypass — the curated handler called \`op.execute(body)\` directly, so \`verifyPerms\` never ran and \`requiresSuperUser\` was not enforced at call time.

Rather than an inline \`isSuperUser\` guard, the handler now routes through the same \`makeOperationToolHandler\` the generic operations profile uses (\`chooseOperation\` + \`processLocalTransaction\`), so \`verifyPerms\` runs uniformly. Two upshots: \`requiresSuperUser\` is enforced, and a scoped \`operations: ['agent_prompt']\` delegation role now works without full super_user. Listing switched from the bespoke \`isSuperUser\` to \`canRoleInvokeOperation\` so visibility mirrors the call-time boundary. Unit tests assert \`op.execute\` is no longer called directly and that a delegation role is visible while a restricted user is denied.

— KrAIs (Claude Opus 4.8)

agent/agent.ts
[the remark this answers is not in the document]
kriszyp: Covered by `unitTests/agent/agent.test.js` (added in 0febc774c, "test(agent): direct unit coverage for resolveAgentIdentity fail-closed policy") — direct unit tests exercise the fail-closed path (non-default user unresolvable, throws) and the bootstrap-fallback path (default `hdb_agent` user, unresolvable, escalates to super_user). Duplicate of the already-resolved thread on r3573183436; resolving this twin.

kriszyp:
This really needs E2E testing on Fabric, and leaving this unmerged is detrimental to the testing, merging.

Issue #1550, opened 2026-07-01
Per-component process.env from granted secrets, with load-gating on declared secrets


kriszyp:
Consumption side of the secrets store (#715); phase 1b of the secrets-management plan.

- Each component gets a synthetic `process.env` object materialized at load: real env plus the secrets granted to that component (decrypted via the registered secret decryptor from #1528). Env-shaped on purpose — the npm ecosystem configures itself from `process.env.FOO`, so no glue code. Eager decrypt of the granted set (enumeration via `Object.keys`/spread must work for dotenv-style consumers).
- Components declare expected secrets in config (parseable without loading component code):
  ```yaml
  secrets:
    STRIPE_SECRET_KEY: { required: true, description: Stripe API key }
    SENTRY_DSN: { required: false }
  ```
- Load gate: a required declaration that is unsatisfied (undefined or ungranted) stops THAT component from loading with a clear error naming the missing vars — the instance keeps running. Component status exposes the unsatisfied set (names/descriptions) so Studio can render a configure-me flow and deploy responses can print it.
- Declarations are requests, not authority — grants (#715) stay authoritative.
- Document loudly: child processes inherit the REAL env, not the synthetic object; secrets need explicit propagation to spawns.
- JS-level scoping is a slowdown layer against generic env scrapers, not a hardened boundary — containers/uids remain the tenant boundary. No SES requirement.

🤖 Drafted by Claude (Fable 5) on Kris's behalf.

PR #1554, opened 2026-07-02
feat: hdb_secret store with grant-scoped secret operations

### LINKED ISSUE #715: Secrets store: hdb_secret system table, grants, and operations API
## Request

Add a `hdb_env` system table for remotely administering environment variables, including secrets. Features:
- Optional per-value encryption (designated secrets), with the encryption key stored outside the table (filesystem by default)
- Operations API additions for setting values and performing encryption
- `loadEnv` built-in component extended to read from `hdb_env` and inject into `process.env` on startup

Related: CORE-2696 (process.setEnv API for cross-thread propagation).

kriszyp:
Closes #715

## Summary

Adds the `system.hdb_secret` table and six super_user-only operations for managing envelope-encrypted secrets. Scope is **store + ops only** — consumption/env materialization is #1550 and no read path for secret values exists in this PR.

- **`utility/secretEnvelope.ts`** — pure envelope codec (no Harper imports), ported from harper-pro's `security/envSecretCrypto.ts` (credit: Dawson Toth, co-authored on the codec commit). AES-256-GCM value encryption, RSA-OAEP(SHA-256) key wrap, `kid` = SHA-256 fingerprint of the SPKI public key. Adds `parseEnvelopeFields` for structural validation without decryption.
- **`components/secretOperations.ts`** — `set_secret`, `grant_secret`, `revoke_secret`, `list_secrets`, `delete_secret`, `get_secrets_public_key`.
- **`resources/secretDecryptor.ts`** — new `registerSecretCustody({ decrypt, getPublicKey })` surface; installs `decrypt` through the existing `registerSecretDecryptor` slot, so the just-merged decryptor API and `loadEnv.ts` are untouched (their tests still pass).
- **Provisioning** — `json/systemSchema.json` entry (fresh installs via `mount_hdb`) paired with a `5.2.0` upgrade directive (existing installs), including the `__dbis__` `is_hash_attribute` guard mirroring the hdb_deployment fix so a harperdb@4.x downgrade doesn't hit MDB_INCOMPATIBLE.
- **Wiring** — `OPERATIONS_ENUM`, operation dispatch map, `requiredPermissions` (all `requires_su`), Joi validators in `components/operationsValidation.js`.

## Purpose

First half of the two-key-custody secrets design: a replicated, envelope-encrypted store that core can host without ever holding plaintext. Core ships no key custody — the Pro secrets component registers it; without custody the store still accepts client-encrypted envelopes (marked `unverified`) and everything else fails with a clean `ClientError`.

Key security properties, enforced and unit-tested:

- Only ciphertext is stored. A plaintext `value` is encrypted immediately against the custody public key and the reference dropped; a submitted `envelope` is **never decrypted server-side** (see below).
- `kid` is derived from the sealed envelope body only; a separate client-supplied kid field is never trusted.
- Secret ops do **not** call `replicateOperation` — rows replicate via normal system-table replication (op replication on top would double-apply).
- Ops-log redaction happens at the boundary: `processLocalTransaction` now strips `envelope` (it already stripped `value`/`values`/`password`/`payload`) before the operation log write. I audited the other body-logging paths on the operations route: `serverErrorHandler` logs error objects only, `validateBySchema` returns message-only Errors (no Joi `_original`), and the patterned Joi fields use custom messages so values are never echoed into validation errors.
- super_user is enforced **inside each handler** in addition to `requiredPermissions`, so a role's `operations` allowlist (gate-2 bypass in `operation_authorization.ts`) cannot delegate secret custody.
- `list_secrets` returns metadata only — never envelopes, never values — and includes the node's custody fingerprint plus per-row `kid_matches_custody`, so stale rows on a cloned/rekeyed node are immediately visible.
- Every mutation emits a structured `logger.notify` audit event (op, name, user — never values).

## Where to focus

1. **Directive version.** The directive is tagged `5.2.0` (main's package.json is 5.1.15). If this ships in a 5.1.x release instead, the directive must be retagged to that exact release — see the mis-tag history documented in `5-1-0.ts` (a too-late version means the directive never fires on the upgrade path).
2. **No decrypt-verify on ingest (deliberate).** Earlier drafts decrypt-verified submitted envelopes; that was reversed because decryption would put plaintext in the main-thread heap, defeating client-side encryption. Ingest checks are structural (marker, base64url JSON, base64-decodable `k`/`iv`/`ct`/`tag`) plus kid-vs-custody-fingerprint match. The accepted trade: a tampered-but-well-formed envelope with a matching kid is stored and only fails at consumption time with a clear decrypt error (pinned by a unit test).
3. **kid-less envelopes.** An envelope without a sealed `kid` is accepted and marked `unverified: true` (with a warning) even when custody is registered — there is no key identity to check without decrypting. Rejecting instead would be a one-line change if you prefer strictness.
4. **`get_secrets_public_key` is a plain POST operation.** The brief called for HTTP GET, but core's operations API dispatches everything through the single POST handler; `OperationDefinition.httpMethod` only exists for externally registered operations. Followed the repo pattern.
5. **Response/row casing** is snake_case (`updated_by`, `custody_fingerprint`, `kid_matches_custody`) with `__createdtime__`/`__updatedtime__` exposed as-is, matching hdb_deployment and the ops API conventions (the design sketch used camelCase).

## Cross-model review

A cross-model review (Codex + Gemini legs, adjudicated with a Harper-domain pass) ran on this PR: **0 blockers**. The Codex leg initially returned no findings and was re-run via MCP, producing 6 real findings on the pre-fix head; every claim was verified against the code before acting. Outcome:

- **Grants atomicity — FIXED** (`413a913`): the get→mutate→put of `set_secret`/`grant_secret`/`revoke_secret`/`delete_secret` now serializes through a per-name in-process promise chain (ops API dispatches these on the main thread only; cross-node concurrency remains whole-row LWW, the known system-table semantic). Same push: no-op grant/revoke short-circuit (`changed: false`, no write, no timestamp/`updated_by` bump), envelope validator regex derived from `ENV_ENCRYPTED_PREFIX`, 256KiB caps on `value`/`envelope`.
- **MCP audit leak — CONFIRMED and FIXED** (`09267d8`, `23ab8a0`): the MCP `tools/call` audit logs args at info level through a key-name redactor that missed `value`/`values`/`envelope`, bypassing the REST ops-log strip when secret ops are allowlisted on the MCP surface. Exact-name redaction added for those three fields (exact-match, so `search_value` stays auditable). On top of that, all six secret operations are now **default-deny on the MCP surface**: a new `DEFAULT_EXCLUDED` set keeps them off `DEFAULT_ALLOW` even where a glob matches (`list_*` was auto-pulling in `list_secrets` — caught by the catalog coverage lint in CI). Operators can still expose them deliberately via `mcp.operations.allow`, where the audit redaction applies; tests pin both the exclusion and the explicit opt-in path.
- **read_audit_log envelope exposure — CONFIRMED, FIXED differently than suggested** (`09267d8`): hdb_secret audit rows carry full record images, and `read_audit_log` is delegable to non-SU roles via the `operations` allowlist. The suggested fix (disable table audit for hdb_secret) is **not available**: `mount_hdb.ts` forces `audit = true` for every system table ("with RocksDB at least, we need to audit everything or there will be lost data"), so audit is load-bearing and the schema flag would be overridden anyway. Instead the read surface is blocked at the op: `read_audit_log` on `system.hdb_secret` returns 403 regardless of config or role. Secret mutations remain audited via the value-free notify events, so nothing is lost.
- **Duplicate grants — FIXED** (`09267d8`): grants are set-semantic now — deduped on every write (including rows that arrived dirty via replication), revoke removes ALL occurrences, and Joi rejects explicit duplicates (`.unique()`) plus caps grants (100 items) and metadata (100 keys).
- **Table.ts trace log — ACCEPTED, documented**: see the stances below.
- **Directive version retag — OPEN (release-checklist item)**: the directive is tagged `5.2.0`; if this ships in a 5.1.x release it MUST be retagged to that exact release, or it never fires on the upgrade path — the exact `5-1-0.ts` mis-tag failure mode.

## Documented v1 stances (by design, not omissions)

- Flat global secret-name namespace; per-component aliasing is a later phase.
- Grant ops ship now but nothing consumes secrets until the materialization PR (#1550) — no false security, since no read path exists.
- AAD context binding is deferred to `enc:v2`; cross-row envelope replay requires super_user, who can already do anything.
- Deletes are not cryptographic erasure — audit/txn logs and backups retain envelopes.
- Synchronous RSA on the main thread is accepted: these are cold-path admin operations.
- The generic table write path trace-logs the first 100 chars of any stored record (`resources/Table.ts` ~2394), which for hdb_secret includes leading ciphertext bytes of the envelope. Ciphertext-only and trace-level — accepted; deliberately not special-cased.
- Clone/re-key story: rows carry `kid`, `list_secrets` exposes `kid_matches_custody`, and the consumption phase (#1550) must fail gracefully on mismatch. A unified per-cluster keypair and a rotation walk are later phases.

## Testing

- `npm run build` (tsc) passes; `npm run lint:required` clean; strict `npm run lint` shows only pre-existing warnings in untouched files.
- 72 tests pass across the touched suites (secret codec/validators/ops + MCP audit redaction) (`unitTests/utility/secretEnvelope.test.js`, `unitTests/components/secretOperationsValidation.test.js`, `unitTests/components/secretOperations.test.js`): codec round-trips/tamper/kid-mismatch/malformed, validator shapes, full op flows against a mock table with real node:crypto keypair custody, plus interleaving/atomicity and no-op short-circuit coverage.
- Regression: `loadEnv.test.js`, `envOperationsValidation.test.js`, `deploymentRecorder.test.js` still pass.

Generated by Claude (Fable 5).

## Docs

Docs deferred deliberately: these operations are dormant in OSS core until the Pro custody component registers keys (harper-pro#166); user-facing documentation ships with the full secrets subsystem (store + custody + consumption) rather than per-PR. Tracking on the plan's phase list.


kriszyp:
Release target confirmed as 5.2.0 (per Kris) — the `5-2-0.ts` directive tag is correct as-is; the retag checklist item in the description is resolved. 🤖 (Claude, Fable 5)

upgrade/directives/5-2-0.ts
```diff
@@ -0,0 +1,82 @@
...
+	const schema = (systemSchema as any)[SECRET_TABLE];
+	if (!schema) {
+		throw new Error(`systemSchema.${SECRET_TABLE} is missing; cannot run 5.2.0 directive.`);
+	}
+
+	initPaths.initSystemSchemaPaths(terms.SYSTEM_SCHEMA_NAME, SECRET_TABLE);
+	const createTable = new (CreateTableObject as any)(terms.SYSTEM_SCHEMA_NAME, SECRET_TABLE, schema.hash_attribute);
+	createTable.attributes = schema.attributes;
+	const primaryKeyAttribute = createTable.attributes.find(({ attribute }) => attribute === schema.hash_attribute);
+	if (primaryKeyAttribute) primaryKeyAttribute.isPrimaryKey = true;
```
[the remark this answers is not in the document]
kriszyp: By design — this is byte-for-byte the pattern of the shipped `5-1-0.ts` directive (which created hdb_deployment the same way in production), and `utility/mount_hdb.ts` performs the identical shared-`systemSchema` attribute mutation for every system table on first boot. Directives run once during upgrade, before normal serving, so the in-memory mutation has no observable effect beyond what the platform already does at boot. Changing the pattern belongs to a directives-wide refactor, not this PR — keeping the new directive identical to the precedent is the safer review posture here.

— Claude (Fable 5)

utility/secretEnvelope.ts
```diff
@@ -0,0 +1,110 @@
...
+
+/** SHA-256 (hex) of the DER SPKI public key — a stable key id used as the envelope `kid`. */
+export function fingerprintOf(publicKeyPem: string): string {
+	const der = createPublicKey(publicKeyPem).export({ type: 'spki', format: 'der' });
+	return createHash('sha256').update(der).digest('hex');
+}
+
+// Strict base64 (the field encoding inside the envelope JSON). Buffer.from(s, 'base64') silently
+// ignores invalid characters, so decodability has to be checked by pattern.
+const BASE64_REGEX = /^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$/;
```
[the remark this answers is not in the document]
kriszyp: Fixed in 0da186db9 — padding is now optional in `BASE64_REGEX` (charset stays strict; a trailing one-character group is still rejected), and the base64url envelope-body validator in `operationsValidation.js` likewise tolerates trailing `=` padding. Added an unpadded-envelope round-trip test and padded-body validator tests. Good catch — client-built envelopes (Studio/WebCrypto) are a first-class path.

— Claude (Fable 5)

components/operationsValidation.js
```diff
@@ -240,6 +243,63 @@ function deleteEnvValueValidator(req) {
...
+ * @param req
+ * @returns {*}
+ */
+function setSecretValidator(req) {
+	const schema = Joi.object({
+		name: SECRET_NAME,
+		value: Joi.string().allow(''),
+		envelope: Joi.string()
+			.pattern(SECRET_ENVELOPE_REGEX)
+			.messages({ 'string.pattern.base': `'envelope' must be an 'enc:v1:' base64url envelope` }),
```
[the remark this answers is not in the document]
kriszyp: This was addressed in 413a9133e (before this thread was opened against an earlier head): `value` and `envelope` both carry `Joi .max(256 * 1024)` caps (`SECRET_MAX_LENGTH`), with boundary tests (reject over, accept at the cap). 09267d826 additionally capped `grants` (100 items, `.unique()`) and `metadata` (100 keys).

— Claude (Fable 5)

dataLayer/readAuditLog.ts
```diff
@@ -27,6 +27,20 @@ export default async function readAuditLog(readAuditLogObject: any) {
...
+	// is load-bearing for data retention), so the read surface is blocked instead, regardless of
+	// config or role. Secret mutations are separately audited via value-free logger.notify events
+	// (components/secretOperations.ts).
+	if (
+		database === hdbTerms.SYSTEM_SCHEMA_NAME &&
+		readAuditLogObject.table === hdbTerms.SYSTEM_TABLE_NAMES.SECRET_TABLE_NAME
+	) {
+		const msg = `read_audit_log is not supported on ${database}.${readAuditLogObject.table}; secret mutations are audited via structured log events that never contain values`;
+		throw handleHDBError(new Error(), msg, HTTP_STATUS_CODES.FORBIDDEN, undefined, undefined, true);
+	}
```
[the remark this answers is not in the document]
kriszyp: Agreed on the requirement — coverage landed in 09267d826 (same push that added the guard, likely crossing this review) and was strengthened in 0da186db9. Because `readAuditLog.test.js` is `describe.skip`'d, the tests live in `unitTests/components/secretOperations.test.js` (`read_audit_log on system.hdb_secret` describe block), where the harness runs for real. They pin: 403 + 'not supported' message for `system.hdb_secret`; 403 regardless of caller role (super_user and non-SU request bodies alike — the guard is deliberately role- and config-independent); and that other tables — including `hdb_secret` outside the `system` database — fail on config/schema checks, never on this guard. All three show in mocha output (64 passing in the suite).

— Claude (Fable 5)

components/mcp/tools/operations.ts
```diff
@@ -115,6 +115,25 @@ export const DEFAULT_ALLOW: readonly string[] = [
 	'read_audit_log',
 ];
 
+/**
+ * Operations excluded from the DEFAULT allow surface even when a glob matches
+ * (`list_*` would otherwise pull in `list_secrets`). The secrets store is key
+ * custody management, not data reads — none of it belongs on an LLM-facing
+ * surface by default, same rationale as the sensitive `get_*` exclusions
+ * above. Operators can still expose these deliberately via
```
dawsontoth, line 142: Totally tangential: do we have an MCP tool for the LLM to see what keys it doesn't have access to, and can it request access to them via another tool? With human approval.

dawsontoth: (Or via instructing the human to update a config file somewhere.)

components/mcp/audit.ts
```diff
@@ -25,6 +25,11 @@ export interface AuditEntry {
 }
 
 const REDACTION_PATTERN = /(secret|password|token|api[-_]?key|credentials?|auth)/i;
+// Exact field names that carry secret material without a credential-looking name: `value`/`values`
+// (set_secret plaintext, set_env_value .env secrets) and `envelope` (set_secret ciphertext). These
+// mirror the fields processLocalTransaction strips from the REST operations log — the MCP audit
+// path must not become a bypass. Exact-match so e.g. `search_value` stays auditable.
+const REDACTION_EXACT_FIELDS = /^(value|values|envelope)$/i;
```
dawsontoth, line 32: Envelope, sure. But value and values feels pretty board... should we opt for a more specific field for these operations, since they're still being formed? Something very blunt, like `secret` and `secrets`, you know? And the redaction pattern above on line 27 would capture the field then.

dawsontoth (approved):

components/mcp/tools/operations.ts
kriszyp: what _keys_ it doesn't have access to or what _tools_ it doesn't have access to? (this in the context of available tools, right?)

dawsontoth: What operations, yes. Which to it would be tools.

components/mcp/audit.ts
kriszyp: Good instinct — `value`/`values` was too blunt. I went a slightly different route than the rename: those field names are also carried by the pre-existing `set_env_value` op (its `.env` secrets live in `value`/`values`), so renaming `set_secret`'s field alone wouldn't remove the broad match, and `set_env_value` can't be renamed.

Instead the exact-field masking is now opt-in per tool — it fires only for `set_secret`/`set_env_value` (the only ops with secrets in generically-named fields). Everything else keeps `value`/`values` auditable, and the global credential-name pattern still covers every tool. Tool names are exactly operation names (one tool per op, `name: operationName`), so it keys off `entry.tool` safely. Landed in `5c4aaa368` with tests for both paths.

Happy to also switch `set_secret` to a `secret` field on top of this if you'd still prefer the clearer name — just didn't want it to be the only guard.

— Claude (Fable 5)

components/mcp/tools/operations.ts
kriszyp: Good idea, and not something we have today — I checked the code. Two layers make an op unavailable and neither is visible to the client: the `mcp.operations.allow`/`deny` config filter (denied ops are never registered as tools, so they don't even show in `tools/list`) and RBAC at invocation time. There's no discovery of what's filtered out and no request/escalation path.

Captured it as a feature with a proposed shape (`list_unavailable_operations` + a human-approved `request_operation_access`) and the sequencing vs #1545 / the agent-auth work: **#1627**. Out of scope for this PR, but tracked.

— Claude (Fable 5)

PR #1560, opened 2026-07-02
feat(agent): Harper best-practices grounding + agent.systemPromptAppend (#626)

kriszyp:
## Summary

Two related enhancements to the built-in agent's grounding (#626):

1. **Harper best-practices knowledge** — sourced from the `harper-best-practices` skill in **`@harperfast/skills`** (a new dependency, so it's versioned with the release, no drift).
2. **`agent.systemPromptAppend`** — operator text appended to the system prompt, tunable on a running instance.

Stacked on HarperFast/harper#1547 (inspector).

## Best practices — progressive disclosure

Mirrors how the skill is designed to be used, so we don't burn ~136K of rules on every LLM call:
- **Overview in the prompt**: `loadBestPracticesOverview()` injects `SKILL.md` (the rule index + when-to-use, ~1.2k tokens) so the agent always knows *which* practices exist.
- **Details on demand**: a new **`harper_best_practice`** tool lists the rules (no arg) or returns `rules/<name>.md` (schema design, relationships, auth, caching, vector indexing, TypeStrip, deployment, …) only when the agent needs them. The `rule` arg is guarded against path traversal.
- **Graceful degradation**: if `@harperfast/skills` isn't resolvable, the overview is omitted and the tool isn't registered — the agent still runs.

Sourced by resolving the package's main entry and reading the `harper-best-practices/` assets alongside `dist/` (the package `exports` only expose `.`, so files are read directly, not `require`d).

## `agent.systemPromptAppend`

- Operator text appended **after** the built-in grounding and the best-practices overview (kept last so it can shape persona/policy without clobbering the load-bearing path/tool guidance).
- Read from `liveConfig` each run and accepted by **`set_agent_config`**, so it can be tuned on a live instance.
- Added to `config-root.schema.json` and `AgentConfig`.

**Assembled system prompt** = built-in grounding → best-practices overview → operator append.

## Verification

Agent unit suite green, including new `bestPractices` tests that exercise the **real** `@harperfast/skills` package: overview loads, tool lists rules, returns a rule body, and rejects traversal / unknown names.

### DIFF
--- a/agent/bestPractices.ts
+++ b/agent/bestPractices.ts
@@ -2,57 +2,33 @@
  * Harper best-practices knowledge for the built-in agent (#626).
  *
  * Sources the `harper-best-practices` skill from the `@harperfast/skills`
- * package (versioned with the release, no drift). Progressive disclosure,
- * mirroring how the skill is meant to be used:
+ * package (versioned with the release, no drift). The package exposes the skill
+ * content directly as module exports — `skillSummary` (the SKILL.md overview),
+ * `ruleNames` (the rule index), and `rules` (name → markdown body) — so we read
+ * from those rather than groveling around the installed package on disk.
+ *
+ * Progressive disclosure, mirroring how the skill is meant to be used:
  *   - the SKILL.md overview (rule index + when-to-use) is injected into the
  *     agent's system prompt so it always knows *which* practices exist;
- *   - the detailed `rules/<name>.md` bodies are pulled on demand via the
+ *   - the detailed rule bodies are pulled on demand via the
  *     `harper_best_practice` tool, so the agent only spends context on the
  *     guidance relevant to the task instead of carrying all ~20 rules.
- *
- * Everything degrades gracefully: if `@harperfast/skills` isn't resolvable the
- * overview is omitted and the tool isn't registered — the agent still runs.
  */
 
-import { existsSync, readFileSync, readdirSync } from 'node:fs';
-import { dirname, join, resolve } from 'node:path';
+import { ruleNames, rules, skillSummary } from '@harperfast/skills';
 import type { AgentTool } from './types.ts';
-
-// Rule file names are `[a-z0-9-].md`; this guards the tool arg against path traversal.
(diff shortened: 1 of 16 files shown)


agent/bestPractices.ts
```diff
@@ -0,0 +1,88 @@
...
+const RULE_NAME = /^[a-z0-9][a-z0-9-]*$/;
+
+/** Resolve the `harper-best-practices` directory inside the installed `@harperfast/skills`, or undefined. */
+function bestPracticesDir(): string | undefined {
+	try {
+		// `@harperfast/skills` exports only `.` (dist/index.js), so resolve the package's main entry
+		// and walk up to the package root; the `harper-best-practices/` assets sit alongside `dist/`.
+		const root = resolve(dirname(require.resolve('@harperfast/skills')), '..');
+		const dir = join(root, 'harper-best-practices');
+		return existsSync(join(dir, 'SKILL.md')) ? dir : undefined;
```
dawsontoth: It's trying too hard and not looking at the module's exports. It gives you an array and enum for the rule names, and a rules map from the name to the markdown contents. No file system shenanigans necessary. It's built so you can't see it easily in the repo. NPM makes it easier to see: https://www.npmjs.com/package/@harperfast/skills?activeTab=code

dawsontoth (approved):

agent/bestPractices.ts
kriszyp: Great catch — you're absolutely right, I was overthinking it. Switched to the module exports in a2224bf65: `import { ruleNames, rules, skillSummary } from '@harperfast/skills'`, with the overview served from `skillSummary`, the list from `ruleNames`, and each rule body from the `rules` map. All the filesystem plumbing (require.resolve path-walk, `readdirSync`/`readFileSync`, and the traversal-guard regex) is gone — a rule lookup is now a plain map access. Bonus: this also mooted the three Gemini sync-I/O / ESM-resolution comments in one stroke. Thanks for the steer! 🙏

— KrAIs (Claude Opus 4.8)

dawsontoth (approved):

PR #1563, opened 2026-07-02
fix(blob): 503 an in-flight replication receive instead of 404

### LINKED ISSUE #481: Sender returns BlobReadError 500/404 for in-flight blobs during catch-up — wedges receivers with no peer fallback
### Bug Summary

When a peer requests a blob the sender hasn't fully replicated to its own local store yet, the sender returns `BlobReadError: Blob is incomplete` (500) or `Blob file not found … ENOENT` (404). The receiver treats this as a terminal failure for that record, advances its resume cursor past it, and never retries against another peer. With two v5 nodes catching up simultaneously, both sides repeatedly serve 500/404 for blobs the other half-replicated, and **the cluster stops converging** even after all writes have stopped.

This is the follow-up wedge that opens up *after* #450/#451 stopped the sender from hanging on the stuck iterator. Bounded failure on each side, no progress for either.

### Steps to Reproduce (today's stage cluster, harper-pro 5.1.11)

1. 4-node v4 cluster (`v4-to-v5.akamai.stage`) on `harperdb/harperdb:4.7.34`, k6 workload writing to all DBs (NR, HttpCache, sem/seo PageCache with blobs).
2. Roll-upgrade two nodes to `harperfast/harper-pro:5.1.11`: `nl-ams-1` first, then `se-sto-1` ~5 min later.
3. Stop k6.
4. Wait 15 min with no writes. Observe nl-ams-1 and se-sto-1 do not catch up to the v4 peers (us-lax-1 / us-ord-1) on blob-bearing tables (sem/seo PageCache).

### Expected Behavior

When the sender doesn't have a complete local copy of a requested blob, either:

- (a) sender holds the request until its own local replication completes the blob, then serves it; or
- (b) sender responds with a soft signal so the receiver tries another peer that has it; or
- (c) receiver, on 500/404, retries against a different known peer before advancing the cursor.

Catch-up converges when writes stop.

### Actual Behavior

Sender returns 500/404 immediately, receiver advances past it (the same "diverged record" pattern #388 describes), no peer fallback, no later re-request. With writes stopped, deltas plateau:

```
node       NR     sem    seo    notes
us-lax-1   34903  5266   6715   v4, target
us-ord-1   34903  5266   6715   v4, target
se-sto-1   34434  4689   5936   v5 (5.1.11), wedged
nl-ams-1   34077  4465   5516   v5 (5.1.11), wedged (~830 NR / 800 sem / 1200 seo behind)
```

After 15 minutes with no workload, nl-ams-1 advanced 0 records on NR/sem/seo (one se-sto-1 burst of +180 sem aside). The blob-receiver-side watchdog (#1444) and the sender-side per-chunk timeout (#451) both fire as designed — they bound how fast each individual failure happens, but neither makes any stream succeed.

### Platform

harper-pro 5.1.11 on RocksDB, Node 22, observed on the Akamai v4-to-v5 stage cluster (Debian 12) after a v4→v5 in-place rolling upgrade of two nodes.

### Console Errors (sender side, nl-ams-1)

```
[http/2] [warn] [replication]: Error sending blob BlobReadError: Blob is incomplete
  statusCode: 500, code: undefined
} blob id 16f4 for record i=00100000936&d=mobile

[http/2] [warn] [replication]: Error sending blob BlobReadError: Blob file not found for /home/harperdb/hdb/blobs/seo/0/0/80d
  statusCode: 404, code: 'ENOENT'
} blob id 80d for record i=00100001455&d=desktop
```

### Console Errors (receiver side, nl-ams-1 and se-sto-1, in lockstep)

```
[http/1] [error] [replication]: Blob save failed for 3e3 from zw4-us-lax-1.v4-to-v5.akamai.stage.harperfabric.com Error: Blob source stream idle for 120000ms (fileId=1b9c)
[http/1] [error]: source-applied transaction commit failed during apply Error: Blob source stream idle for 120000ms (fileId=1b9c)
[http/2] [error]: Transaction was open too long and has been committed, from table: PageCache/
```

### Additional Context

- #450 (closed) was the sender-side hang on the stuck iterator. #451 fixed that. This issue is what stays broken after that fix: instead of hanging, the sender returns 500/404 quickly, but the receiver still has no recovery path.
- #386 covers the broader "sustained blob-replication timeouts cause silent loss with no self-repair" theme. This is the specific *trigger* (two v5 nodes catching up while neither has the full local blob set yet) and the specific *sender-side response* that needs handling — proximate cause of today's wedge, not the chronic loss case.
- #388 (proactive blob backfill) is the downstream consequence: receiver advances the cursor and the record stays diverged. A fix on the sender side here would reduce how often #388's repair sweep has to act.
- #208 is adjacent (peer selection for blob GETs not honored). The choice we'd want — "retry blob fetch against a different peer when one returns 500/404" — overlaps that machinery.

Reproduction artifacts:

- `/tmp/round4-evidence/blob-idle-watchdog.txt` (50 receive-side fires)
- One 3-min window on nl-ams-1: 26,486 `Timeout waiting for blob stream to finish` warnings draining as a single batch
- `caffeinate -di` kept the harness up for the full soak

Verified the existing receiver-side watchdog and sender-side per-chunk timeout fire as designed; they just don't fix this.

ldt1996:
Fixes HarperFast/harper-pro#481

## Race
Sender's `FileBackedBlob` ENOENTs on a blob whose replication receive has been announced but hasn't landed on disk. Pre-fix returns 404 (permanent), requester marks the record diverged.

## Fix
Add a refcounted receive-in-flight registry in [resources/blob.ts](resources/blob.ts). ENOENT arm now returns 503 (retryable) if `isBeingWritten` or `isBlobReceiveInFlight(fileId)`; 404 only when the blob is cleanly absent. Receive side (populates the registry) is in HarperFast/harper-pro#513.

## Tests
`npx mocha unitTests/resources/blob.test.js --grep '#481'` — 8/8 pass (2 new). New test unlinks the file, registers the fileId, verifies `stream()` rejects 503; unregisters, verifies same read rejects 404 with `code: 'ENOENT'`.


kriszyp (approved):
Approving — correct, conservative fix: only registered in-flight receives flip ENOENT→503; truly-missing blobs still 404. The earlier red Node-v24 unit leg was a 10-min job-timeout flake and is now green. Pairs with harper-pro#513 (which populates the in-flight registry). Thanks @ldt1996!

PR #1568, opened 2026-07-02
Name authentication middleware and add http.securityHeaders config

kriszyp:
## Summary

Two small core enablers for an upcoming WAF middleware component:

1. **Explicitly named authentication middleware** — honesty note: this is contract-hardening, not new behavior. Middleware entries already default their name to the registering component's name (`name: options?.name ?? getComponentName()` in `server/http.ts`), and the auth plugin loads under the `authentication` config key — so `before: 'authentication'` / `after: 'authentication'` ordering (used today by REST, MQTT, GraphQL, MCP, and `static.ts`) already resolved correctly. This change makes the name an **explicit, test-locked contract** so a future rename of the config/registry key can't silently break every consumer that targets `'authentication'`. **No chain-order behavior changes.**

   The operations-API registrations in `server/operationsServer.ts` are a genuine (if inert) name change: they previously defaulted to the `operationsApi` component name and now register as `authentication`. Nothing in-repo targets `'operationsApi'` as a middleware name, and those chains are bypassed for real ops traffic (see caveat below), so this is forward-compat consistency only.

2. **`http.securityHeaders` config block** — a zero-code way for ops to add response headers like `X-Frame-Options` / `X-Content-Type-Options`:

   ```yaml
   http:
     securityHeaders:
       X-Frame-Options: SAMEORIGIN
       X-Content-Type-Options: nosniff
   ```

   Applied on app ports across all response paths of the Harper-native request handler: the normal writeHead path, `handlesHeaders` streams (static files — where the `send()` stream writes its own headers, universal headers are pre-set so the stream can still override its own names), thrown-error responses, and the Fastify `status === -1` cascade. Node and Bun handlers both covered.

   - **Precedence: app wins.** Configured headers are *defaults* — a header the application/route already set on a response is never overridden (a route's `X-Frame-Options: DENY` is not loosened by a configured `SAMEORIGIN`).
   - **Validation:** names/values validated with `node:http`'s `validateHeaderName`/`validateHeaderValue`; invalid entries (and non-object block values) are logged and skipped, never thrown.
   - **Hot-reload:** via the existing `scope.options.on('change')` path; the feature tracks exactly the entries it pushed into `universalHeaders` and splices only those, preserving entries pushed by other components. Only the first (root) http scope owns the config — an application `config.yaml` with an `http:` block cannot wipe root-configured headers.
   - **Opt-in / no behavior change:** absent block ⇒ nothing added; all per-response loops are guarded by `universalHeaders.length` so unconfigured deployments skip them.

## Where to look

- `server/http.ts` — `applySecurityHeaders()` (ownership-tracked splice, root-scope guard) and the four response-path application sites.
- **Scope caveat:** the operations API doesn't carry these headers in normal mode — not because ops requests bypass the request handler (they don't; they cascade to Fastify through it), but because the ops API runs on the main thread, which loads components with `resources.isWorker = false`, so http's `handleApplication` never populates the main thread's per-thread `universalHeaders`. With `threads: 0` the ops API shares a worker and *would* carry them (benign). Detailed in DESIGN.md. Flagging in case ops-API parity is wanted later.
- `server/operationsServer.ts` — ops-API auth registrations get `name: 'authentication'` for consistency (forward-compat naming, not a behavior change).

## Tests

- `unitTests/server/securityHeaders.test.js` (new): apply/coerce/validate/non-object-reject/hot-reload-swap semantics, "don't clobber other components' entries", and root-scope ownership vs a second app scope.
- `unitTests/server/middlewareChain.test.js`: new case proving a `before: 'authentication'` entry executes before auth regardless of registration order.
- `unitTests/security/auth.test.js`: asserts `handleApplication` registers the auth listener under `name: 'authentication'`.
- `unitTests/validation/configValidator.test.js`: 4 new schema cases (absent/map/coercible/rejected).
- `integrationTests/server/security-headers.test.ts` + fixture app (new): real Harper instance — headers on 200 REST, 404 cascade, static-file (`handlesHeaders`), and thrown-error responses; app-set header wins over config; absent config adds nothing.

---
Generated by KrAIs (Claude, Fable 5 model) via Claude Code.

### DIFF
--- a/server/http.ts
+++ b/server/http.ts
@@ -149,7 +149,7 @@ function applySecurityHeaders(securityHeaders: HttpOptions['securityHeaders']) {
 	}
 	ownedSecurityHeaders = [];
 	if (!securityHeaders) return;
-	if (typeof securityHeaders !== 'object') {
+	if (typeof securityHeaders !== 'object' || Array.isArray(securityHeaders)) {
 		harperLogger.error('Invalid http.securityHeaders value: expected a map of header names to values');
 		return;
 	}
@@ -578,6 +578,9 @@ function getHTTPServer(port: number, secure: boolean, options: ServerOptions) {
 					}
 					if (sentBody) nodeResponse.end(body);
 				} else if (universalHeaders.length > 0 && !nodeResponse.headersSent) {
+					// Known limitation: a handler that synchronously calls writeHead before
+					// returning { handlesHeaders: true } has already sent headers and cannot
+					// receive universal headers (no in-tree component does this).
 					// handlesHeaders responses (e.g. static's send() stream) write their own headers
 					// directly to nodeResponse; pre-set universal headers as defaults — a header the
 					// stream sets itself (same name) will overwrite these
@@ -1137,10 +1140,24 @@ function getBunHTTPServer(port: number, secure: boolean, options: ServerOptions)
 					if (statusCode === 500) harperLogger.warn(errorForLog(error));
 					else harperLogger.info(errorForLog(error));
 				} else harperLogger.error(errorForLog(error));
-				if (universalHeaders.length > 0) {
-					// universal headers apply to error responses too
+				const errorHeaders = error.headers;
+				if (errorHeaders || universalHeaders.length > 0) {
 					const headers = new globalThis.Headers();
-					for (const [key, value] of universalHeaders) headers.set(key, value);
+					// error.headers may be an iterable Headers or a plain object, same as
+					// toWriteHeadHeaders accepts on the Node path
+					if (errorHeaders?.[Symbol.iterator]) {
+						for (const [key, value] of errorHeaders) headers.append(key, String(value));
+					} else if (errorHeaders) {
+						for (const key in errorHeaders) {
+							const value = errorHeaders[key];
+							if (Array.isArray(value)) for (const item of value) headers.append(key, String(item));
+							else headers.set(key, String(value));
+						}
+					}
+					// universal headers apply to error responses too; error-provided headers win
+					for (const [key, value] of universalHeaders) {
+						if (!headers.has(key)) headers.set(key, value);
+					}
 					return new Response(errorToString(error), { status, headers });
 				}
 				return new Response(errorToString(error), { status });
--- a/server/http.ts
+++ b/server/http.ts
@@ -180,6 +180,12 @@ function applyUniversalHeaders(headers: { has(name: string): boolean; set(name: 
 // handleApplication, and without this guard that call would wipe root-configured headers.
 let securityHeadersOwned = false;
 
+/** Test seam: reset the module-level guard so tests can re-invoke handleApplication. */
+export function _resetSecurityHeadersOwnedForTest(): void {
+	applySecurityHeaders(undefined);
+	securityHeadersOwned = false;
+}
+
 export function handleApplication(scope: Scope) {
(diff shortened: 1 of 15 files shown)


kriszyp:
Cross-model review disposition (Gemini via agy, full diff):

- **Accepted + fixed** (d405d48): Bun thrown-error path now merges `error.headers` (previously dropped on Bun — pre-existing gap, fixed while touching the path; error headers win over configured universal headers, mirroring Node); `applySecurityHeaders` rejects arrays.
- **Documented as known limitation**: a handler that synchronously calls `writeHead` before returning `handlesHeaders: true` cannot receive universal headers (headers already sent; no in-tree component does this).
- **Refuted**: claimed app-header clobber on the Bun `handlesHeaders` path — `responseHeaders` is empty at the point universal headers are set, and the stream's later `setHeader` calls overwrite, so app/stream wins by ordering.
- **Declined (nit)**: universal headers on the catastrophic Fastify-inject-failure handlers — dead-rare path, not worth the code.

An independent fresh-context review earlier drove commit 6300e32 (response-path coverage + app-wins precedence). — KrAIs (Claude)

server/http.ts
```diff
@@ -136,11 +136,59 @@ export function cleanupSocketsDirectory() {
...
+function applyUniversalHeaders(headers: { has(name: string): boolean; set(name: string, value: string): void }) {
+	for (const [key, value] of universalHeaders) {
+		if (!headers.has(key)) headers.set(key, value);
+	}
+}
+
+// Only the first http scope to load (the root config) owns securityHeaders: 'http' is a
+// trusted plugin key, so an application config.yaml with an `http:` block re-invokes
+// handleApplication, and without this guard that call would wipe root-configured headers.
+let securityHeadersOwned = false;
```
cb1kenobi, line 181: **Low: `securityHeadersOwned` is a process-global guard with no test reset, making `securityHeaders.test.js` order-dependent.**

Once any code calls http's `handleApplication`, ownership is claimed permanently. `unitTests/server/securityHeaders.test.js` relies on being the *first* file in the mocha run to call it (its own comment says so) — driven only by alphabetical path ordering. Today that holds, but if a future test under `unitTests/components/**` or `unitTests/security/**` (both sort before `server/`) imports and calls http's `handleApplication`, it claims ownership first, `rootScope` in the test no longer owns securityHeaders, and every assertion expecting headers to be applied fails with a confusing, order-dependent error rather than a clear one.

Suggested fix: export a test-only reset that clears `securityHeadersOwned` (and `ownedSecurityHeaders`), mirroring the existing convention `_resetApplicationStartedForTest()` in `components/mcp/index.ts`, and call it in the test's `before` hook. That removes the hidden ordering contract.

—
_Generated by Barber AI_

kriszyp: Addressed in 766ba13e9: added an explicit test-only ownership reset and invoked it from the securityHeaders suite setup. The reset also removes any previously tracked universal-header entries before releasing ownership.

heskew (approved):
Header application traced across all four response paths (Node http/http2, Bun, uWS, error); case-mismatch bypass ruled out; the `operationsApi`→`authentication` rename has no old-name consumers. Checked the `threads: 0` duplicate-registration concern: `handleApplication` is worker-only with a `started` idempotency guard, the ops-server registration is main-only (deliberate, per the code comment), and they target different ports — worst case is redundant fail-closed auth, never a bypass. LGTM.

PR #1574, opened 2026-07-02
feat(static): support `after` ordering and warn when `fallthrough: false` blocks REST

dawsontoth:
## Problem

The static handler registers `before: 'authentication'` by default (`runFirst: true` prior to 5.1), and the REST handler runs `after: 'authentication'` — so static always answers GETs first. That makes the documented history-mode SPA recipe a footgun:

```yaml
static:
  files: 'dist/**'
  notFound:
    file: 'index.html'
    statusCode: 200
  fallthrough: false
```

This catch-all answers **every** unmatched GET itself — including GETs for exported REST resources — so the application's REST API silently becomes unreachable over GET (`GET /MyResource/` returns index.html with a 200). All four create-harper SPA templates shipped this config; [create-harper#109](https://github.com/HarperFast/create-harper/pull/109) removes it there and steers users toward hash routing, but Harper itself should support the history-mode pattern correctly and warn about the trap.

## With this PR

One added line makes the recipe work as everyone assumed it did — the API is matched first, and the fallback only catches what's left:

```yaml
rest: true

static:
  files: 'dist/**'
  # Let the REST handler match first; only unmatched URLs get the SPA fallback.
  after: 'rest'
  notFound:
    file: 'index.html'
    statusCode: 200
  fallthrough: false
```

```
GET /Dog/1          → 200 application/json   (REST resource)
GET /assets/app.js  → 200 text/javascript    (static file)
GET /app/settings   → 200 text/html          (index.html — client-side route)
```

History-mode routing and the REST API coexist. And anyone still running the old config now gets a startup warning in the log naming this exact fix.

## Changes

- **`after` option on static** (e.g. `after: 'rest'`): positions the handler after the named middleware so API routes match first and only unmatched URLs receive the `notFound` fallback. When `after` is given, the default `before: 'authentication'` hoist is suppressed — combining them would create an ordering cycle (static < authentication < rest < static), which would fall back to registration order with a warning.
- **`before: false`** is now a documented, validated way to clear the default constraint without adding a new one (registration order applies). This previously worked only by accident of the falsy check in `topoSort`.
- **Validation** for `before`/`after` values, matching the style of the other option validations in the plugin.
- **Startup warning** (also re-checked on live reload of `fallthrough`) when `fallthrough: false` is used in the default pre-REST position, naming `after: 'rest'` as the remedy.
- **`config-app.schema.json`**: the static component is now described (files, urlPath, index, extensions, fallthrough, notFound, before, after), so editors using the schema get completion and validation.

## Verification

- New unit tests: `unitTests/server/static.test.js` (12 tests) covering the ordering options passed to `server.http`, validation errors, and warning behavior — hand-rolled fake scope, plain `assert`, per house style.
- End-to-end against a built dist with a Vite SPA app + exported resource:
  - With `after: 'rest'`: `GET /Greeting/` → `200 application/json` (the resource), `GET /some/deep/link` → `200 text/html` (index.html fallback), `GET /` and built assets serve normally. **History-mode routing and the REST API coexist.**
  - With the old footgun config: the new warning appears in `hdb.log` at startup; behavior is unchanged (`GET /Greeting/` still returns index.html) so nothing breaks for existing pure-static apps.
- `oxlint --deny-warnings` and `prettier --check` clean on the changed files.
- `npx mocha unitTests/server/static.test.js` — 12 passing. (The full `test:unit:server` suite fails identically on `main` in my local environment — pre-existing, unrelated.)

## Follow-up

Docs: [documentation#562](https://github.com/HarperFast/documentation/pull/562) (draft) adds a Handler Ordering section to the Static Files page and corrects its SPA example to use `after: 'rest'`.

### DIFF
--- a/config-app.schema.json
+++ b/config-app.schema.json
@@ -51,6 +51,85 @@
 			},
 			"required": ["files"]
 		},
+		"static": {
+			"type": "object",
+			"description": "Static file serving configuration.",
+			"additionalProperties": true,
+			"properties": {
+				"files": {
+					"description": "Glob(s) or path(s) to the files to serve.",
+					"oneOf": [
+						{
+							"type": "string"
+						},
+						{
+							"type": "array",
+							"items": {
+								"type": "string"
+							}
+						}
+					]
+				},
+				"urlPath": {
+					"type": "string",
+					"description": "URL path prefix the files are served under."
+				},
+				"index": {
+					"type": "boolean",
+					"description": "Serve index.html files for directory requests.",
+					"default": true
+				},
+				"extensions": {
+					"type": "array",
+					"items": {
+						"type": "string"
+					},
+					"description": "File extensions to try appending when a requested path is not found (e.g. ['html'] serves /page.html for /page)."
+				},
+				"fallthrough": {
+					"type": "boolean",
+					"description": "Pass unmatched requests to the next handler. When false, this handler answers unmatched GETs itself — note it runs before the REST handler by default, so also set `after: 'rest'` if the application serves an API.",
+					"default": true
+				},
+				"notFound": {
+					"description": "File to serve when a path is not found (requires `fallthrough: false`). Either a path, or an object with `file` and `statusCode` (e.g. index.html with 200 for history-mode SPAs — combine with `after: 'rest'`).",
+					"oneOf": [
+						{
+							"type": "string"
+						},
+						{
+							"type": "object",
+							"additionalProperties": false,
+							"properties": {
+								"file": {
+									"type": "string"
+								},
+								"statusCode": {
+									"type": "number"
+								}
+							},
+							"required": ["file", "statusCode"]
+						}
+					]
+				},
+				"before": {
+					"description": "Run this handler before the named handler. Defaults to 'authentication' so file requests skip credential parsing; false clears the default so registration order applies.",
+					"oneOf": [
+						{
+							"type": "string"
+						},
+						{
+							"const": false
+						}
+					]
+				},
+				"after": {
+					"type": "string",
+					"description": "Run this handler after the named handler, e.g. 'rest' to let API routes match before static fallbacks. Overrides the default `before: 'authentication'` hoist."
+				}
+			},
+			"required": ["files"]
+		},
 		"package": {
 			"type": "string",
 			"description": "Package identifier for the application (npm-compatible or git URL)."
--- a/config-app.schema.json
+++ b/config-app.schema.json
@@ -54,7 +54,7 @@
 		"static": {
 			"type": "object",
 			"description": "Static file serving configuration.",
-			"additionalProperties": true,
+			"additionalProperties": false,
 			"properties": {
 				"files": {
 					"description": "Glob(s) or path(s) to the files to serve.",
@@ -113,10 +113,11 @@
 					]
 				},
 				"before": {
-					"description": "Run this handler before the named handler. Defaults to 'authentication' so file requests skip credential parsing; false clears the default so registration order applies.",
+					"description": "Run this handler before the named handler (the config key as registered, e.g. 'rest'). Defaults to 'authentication' so file requests skip credential parsing; false clears the default so registration order applies.",
 					"oneOf": [
 						{
-							"type": "string"
+							"type": "string",
+							"minLength": 1
 						},
 						{
 							"const": false
@@ -125,7 +126,8 @@
 				},
 				"after": {
 					"type": "string",
-					"description": "Run this handler after the named handler, e.g. 'rest' to let API routes match before static fallbacks. Overrides the default `before: 'authentication'` hoist."
+					"minLength": 1,
+					"description": "Run this handler after the named handler (the config key as registered, e.g. 'rest' to let API routes match before static fallbacks). Overrides the default `before: 'authentication'` hoist."
 				}
 			},
 			"required": ["files"]
--- a/server/static.ts
+++ b/server/static.ts
@@ -19,6 +19,8 @@ import send from 'send';
  *   catch-all answers GETs for exported REST resources too; an SPA with history-mode routing should set
  *   `after: 'rest'` so the API is matched first and only unmatched URLs receive the `notFound` fallback.
  *   `before: false` clears the default without adding a new constraint (registration order applies).
+ *   Handler names are the component config keys as registered (e.g. `rest`, not the legacy `REST` alias);
+ *   a name that matches no registered handler is ignored, with a warning logged by the middleware chain.
  *   Ordering is applied when the component loads; changing `before`/`after` triggers a component restart.
  *
  * This plugin dynamically updates its behavior based on the current configuration file. Users can make updates and immediately see the changes reflect in the next request.
@@ -37,15 +39,27 @@ export function handleApplication(scope: Scope) {
 	// with the registered route (#1583).
 	const baseURLPath = resolveBaseURLPath(scope.pluginName, (scope.options.getAll() as any)?.urlPath);
 
-	const before = scope.options.get(['before']);
-	const after = scope.options.get(['after']);
+	// A bare `before:` / `after:` key in YAML parses as null — treat it as unset, like before this
+	// option was validated.
+	const before = scope.options.get(['before']) ?? undefined;
+	const after = scope.options.get(['after']) ?? undefined;
 	validateOrderingOption('before', before, true);
 	validateOrderingOption('after', after, false);
+	// Default to the pre-authentication hoist only when no ordering was configured at all.
+	const noOrderingConfigured = before === undefined && after === undefined;
 
 	// With the default ordering this handler answers unmatched GETs ahead of the REST handler, so a
 	// `fallthrough: false` catch-all makes any exported REST resources unreachable over GET.
+	// Reads the live option values (not the registration-time ones above): a config save can change
+	// the ordering options together with `fallthrough`, and this re-runs from the change listener.
 	const warnIfBlockingRest = () => {
-		if (before === undefined && after === undefined && scope.options.get(['fallthrough']) === false) {
+		const liveBefore = scope.options.get(['before']) ?? undefined;
+		const liveAfter = scope.options.get(['after']) ?? undefined;
+		if (
+			liveAfter === undefined &&
+			(liveBefore === undefined || liveBefore === 'authentication') &&
+			scope.options.get(['fallthrough']) === false
+		) {
 			scope.logger.warn(
 				`The static handler runs before authentication and REST by default, so \`fallthrough: false\` answers every unmatched GET itself — including GETs for any exported REST resources. If this application serves an API, add \`after: 'rest'\` to the static options so API requests are matched first, or remove \`fallthrough: false\`.`
 			);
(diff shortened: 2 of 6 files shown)


server/static.ts
```diff
@@ -32,6 +55,9 @@ export function handleApplication(scope: Scope) {
 			scope.logger.info(`Static files reinitialized due to change in ${key.join('.')}`);
 			return;
 		}
+		if (key[0] === 'fallthrough') {
+			warnIfBlockingRest();
+		}
```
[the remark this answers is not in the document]
dawsontoth: Good catch — confirmed Scope's auto-restart is bypassed once a plugin registers its own 'change' listener. The change listener now calls scope.requestRestart() when before or after changes, with unit tests covering both the restart and the no-restart (per-request options) paths.

kriszyp (approved):
Approving — nice fix for the static-shadows-REST GET footgun; `after` ordering resolves once at registration (no per-request cost) and the `fallthrough:false` warning is accurate.

One question I'm curious about: **what's the use case where you'd actually reach for `after` here** — is it for a specific static-vs-REST layout, or a general escape hatch? And given this change, **is the current default ordering still the right one for the common case**, or does this suggest the default should change so people don't need `after` in the first place?

dawsontoth:
There are legitimate cases for both. 

For most Harper apps, static was for purely known, static asset serving. From a known map, those requests can be served the fastest, and they help speed up site metrics. It's a cheap thing to check, should be served before authentication, and then you check your meaty resources if they want to handle things.

But with SPAs that want to use real looking URLs and query string parameters for users, a 404 fallback to a known index.html (and scripted client side route loading) is a very common approach to serving content.

The static plugin couldn't handle this situation. We opted to push people to use hash based routing, which does have advantages. But with such a simple change, we can let people do what they want. They may not have a choice over the URLs they must handle.

kriszyp:
> real looking URLs and query string parameters for users, a 404 fallback to a known index.html 

Is this really the ideal way to handle mapping a set (a glob?) of URLs ("real looking") to a static file (pass throug auth, REST, 404, fallback handler)?  Should we add support for URL glob -> file path for explicit and fast mapping?

server/static.ts
```diff
@@ -165,7 +198,25 @@ export function handleApplication(scope: Scope) {
...
+	);
+}
+
+function validateOrderingOption(
+	name: string,
+	value: any,
+	allowFalse: boolean
+): asserts value is undefined | string | false {
+	if (value === undefined || (typeof value === 'string' && value.length > 0)) return;
```
cb1kenobi, line 270: **Medium: bare `before:` in YAML (null) now disables the static component**

The old registration coalesced null — `get(['before']) ?? 'authentication'` — so a bare `before:` key behaved as the default hoist and files were served. Now `validateOrderingOption('before', null, true)` throws, the throw escapes `handleApplication`, and componentLoader marks the component failed (ErrorResource) — static serving is disabled entirely for a config that previously worked. Same for `before: ''`, which previously fell through topoSort's `if (before)` truthy check as a no-op.

Suggested fix: treat `null` as unset — `if (value === undefined || value === null) return;` — preserving the old behavior for a bare key. Empty-string rejection is defensible; null is the plausible accidental config.

—
_Generated by Barber AI_

server/static.ts
```diff
@@ -165,7 +198,25 @@ export function handleApplication(scope: Scope) {
 				body: send(req, realpathSync(notFoundPath)),
 			};
 		},
-		{ before: (scope.options.get(['before']) as string) ?? 'authentication' }
+		{
+			// `after` (e.g. `after: 'rest'`) must suppress the default pre-authentication hoist —
+			// combining the two constraints would be a cycle, which falls back to registration order.
+			before:
+				before === false ? undefined : ((before as string) ?? (after === undefined ? 'authentication' : undefined)),
+			after: after as string | undefined,
```
cb1kenobi, line 260: **Low: `after: 'rest'` silently no-ops when no handler named `rest` exists — e.g. under the legacy `REST:` config key**

Middleware entry names come from the config key, and the legacy `REST:` key is still supported (componentLoader.ts:99) — that registers an entry named `REST`, not `rest`. `topoSort` drops `before`/`after` names that match no entry without any diagnostic (server/middlewareChain.ts:47-53). So a legacy-config user who follows the new warning's advice gets no ordering constraint, loses the default pre-auth hoist, and the warning stops firing (`after !== undefined`) — the footgun persists with zero signal. A typo (`after: 'Rest'`) behaves the same; `validateOrderingOption` accepts any non-empty string.

Suggested fix: log once in `makeCallbackChain`/`topoSort` when a `before`/`after` name never resolves (the `onCycle` warning is precedent), and/or note in the JSDoc/schema that the name is the config key (`rest`, not legacy `REST`).

—
_Generated by Barber AI_

server/static.ts
```diff
@@ -23,6 +30,22 @@ export function handleApplication(scope: Scope) {
...
 
+	const before = scope.options.get(['before']);
+	const after = scope.options.get(['after']);
+	validateOrderingOption('before', before, true);
+	validateOrderingOption('after', after, false);
+
+	// With the default ordering this handler answers unmatched GETs ahead of the REST handler, so a
+	// `fallthrough: false` catch-all makes any exported REST resources unreachable over GET.
+	const warnIfBlockingRest = () => {
+		if (before === undefined && after === undefined && scope.options.get(['fallthrough']) === false) {
```
cb1kenobi: **Low: the warning misses two shapes of the same footgun**

(1) An explicit `before: 'authentication'` — a plausible pre-existing spelling of the default, since the `before` option predates this PR — blocks REST identically, but `before === undefined` never fires the warning. (2) This closure captures registration-time `before`/`after`: a config save that adds both `after: 'rest'` and `fallthrough: false` re-runs the check from the `fallthrough` change event with the stale `after === undefined` and warns the user to add the very `after: 'rest'` they just added (the restart is debounced, so event order doesn't save it).

Suggested fix: read live values inside the check and include the explicit-default spelling:

```ts
const before = scope.options.get(['before']);
const after = scope.options.get(['after']);
if (after === undefined && (before === undefined || before === 'authentication') && scope.options.get(['fallthrough']) === false) {
```

—
_Generated by Barber AI_

server/static.ts
```diff
@@ -165,7 +198,25 @@ export function handleApplication(scope: Scope) {
 				body: send(req, realpathSync(notFoundPath)),
 			};
 		},
-		{ before: (scope.options.get(['before']) as string) ?? 'authentication' }
+		{
+			// `after` (e.g. `after: 'rest'`) must suppress the default pre-authentication hoist —
+			// combining the two constraints would be a cycle, which falls back to registration order.
+			before:
+				before === false ? undefined : ((before as string) ?? (after === undefined ? 'authentication' : undefined)),
```
cb1kenobi: **Cleanup: nested ternary re-encodes the predicate `warnIfBlockingRest` already spells out**

This expression encodes "default to the hoist only when no ordering was configured" — the same `before === undefined && after === undefined` concept as the warning — as a nested ternary whose `as string` cast is only sound because of the outer branch. Post-validation `before` is `string | false | undefined`, so this is equivalent, flat, and cast-free:

```ts
const noOrderingConfigured = before === undefined && after === undefined;
// ...
before: typeof before === 'string' ? before : noOrderingConfigured ? 'authentication' : undefined,
```

—
_Generated by Barber AI_

config-app.schema.json
```diff
@@ -51,6 +51,85 @@
 			},
 			"required": ["files"]
 		},
+		"static": {
+			"type": "object",
+			"description": "Static file serving configuration.",
+			"additionalProperties": true,
```
cb1kenobi: **Low: `additionalProperties: true` forfeits typo-catching; sibling blocks use `false`**

`graphqlSchema` and `jsResource` both set `"additionalProperties": false`, and this block now enumerates every option the handler reads. With `true`, a user who writes `fallThrough: false` or `notfound:` gets no editor squiggle and the handler silently ignores the key — the exact misconfiguration class the new runtime warning exists to mitigate.

Suggested fix:
```suggestion
			"additionalProperties": false,
```

—
_Generated by Barber AI_

config-app.schema.json
```diff
@@ -51,6 +51,85 @@
...
+							},
+							"required": ["file", "statusCode"]
+						}
+					]
+				},
+				"before": {
+					"description": "Run this handler before the named handler. Defaults to 'authentication' so file requests skip credential parsing; false clears the default so registration order applies.",
+					"oneOf": [
+						{
+							"type": "string"
```
cb1kenobi: **Cleanup: schema accepts empty-string `before`/`after` that the code rejects**

`validateOrderingOption` requires a non-empty string and throws on `''` at component load, but the bare `{"type": "string"}` here (and in `after` below) blesses `before: ''` in the editor.

Suggested fix: add `"minLength": 1` to the string branches of `before` and `after`.

—
_Generated by Barber AI_

unitTests/server/static.test.js
```diff
@@ -0,0 +1,143 @@
...
+		handleEntry() {},
+		requestRestart: () => state.restartRequests++,
+		server: {
+			http: (_listener, httpOptions) => {
+				state.httpOptions = httpOptions;
+			},
+		},
+		// Simulate a live config reload for the given key.
+		fireChange(key) {
+			for (const listener of state.changeListeners) listener([key]);
```
cb1kenobi: **Cleanup: `fireChange` drops the real change-event args, and the changed-behavior inputs are untested**

The real `OptionsWatcher` emits `change` with `(key: string[], value, config)`; the fake passes only `[key]`. The key-array shape (the load-bearing part) matches, but a future handler reading `value` would pass here and break live. Also untested: `before: null` (YAML bare key) and `before: ''` / `after: ''` — exactly the inputs whose behavior changed vs. the old `?? 'authentication'` — and `before: false` combined with `after`.

Suggested fix: `fireChange(key, value) { for (const l of state.changeListeners) l([key], value, options); }`, plus the three missing validation cases.

—
_Generated by Barber AI_

server/static.ts
dawsontoth: Fixed — a bare `before:`/`after:` key (null) is now coalesced to unset before validation, so it falls back to the default hoist exactly as it did pre-validation. Empty string still throws. Covered by new unit tests.

server/static.ts
dawsontoth: Done — `makeCallbackChain` now detects `before`/`after` names that match no registered entry and logs a warning naming the entry, the constraint, and the port (with a note that names are the config keys as registered). One wrinkle: the chain is rebuilt on every registration, so a build-time check would false-positive while components are still loading (static registers `after: 'rest'` before REST does). The diagnostic is therefore reported on the chain's first dispatch — only the final chain ever serves requests. The JSDoc and schema descriptions now also say the name is the config key (`rest`, not the legacy `REST` alias).

server/static.ts
dawsontoth: Fixed as suggested — `warnIfBlockingRest` now reads the live option values on every run and also treats an explicit `before: 'authentication'` as the blocking default position. Both shapes are covered by new unit tests, including the config-save-adds-both case.

server/static.ts
dawsontoth: Done — flattened to `typeof before === 'string' ? before : noOrderingConfigured ? 'authentication' : undefined` with a named `noOrderingConfigured` const; the `as string` cast is gone.

config-app.schema.json
dawsontoth: Done — `additionalProperties: false`, matching the sibling blocks.

config-app.schema.json
dawsontoth: Done — `minLength: 1` added to the string branches of both `before` and `after`.

unitTests/server/static.test.js
dawsontoth: Done — `fireChange` now emits `([key], value, config)` matching the real OptionsWatcher, and tests were added for `before: null` / `after: null` (treated as unset), `before: ''` / `after: ''` (rejected), and `before: false` combined with `after: 'rest'`.

dawsontoth:
@kriszyp specifying your paths would be more performant, yes. But you're contending with a lot of people that want to do it with a 404, and have done it with 404 for a very long time.

kriszyp (approved):

PR #1581, opened 2026-07-02
fix(cache): consult allowStaleWhileRevalidate for query-driven revalidation (#1578)

### LINKED ISSUE #1578: allowStaleWhileRevalidate is never consulted for query-driven revalidation
## Summary

When a query/search encounters an expired or invalidated entry, `transformEntryForSelect` reloads it from source:

```ts
// resources/Table.ts (~line 3076 on v5.1)
const loadingFromSource = ensureLoadedFromSource(source, entry.key ?? entry, entry, context);
if (loadingFromSource?.then) {
	return loadingFromSource.then(transform);
}
```

No `resource` argument is passed, and inside `ensureLoadedFromSource` the stale-while-revalidate check is:

```ts
if (entry?.value && resource?.allowStaleWhileRevalidate?.(entry, id)) { ... return; }
```

So for query-driven reloads the SWR early-return can never trigger: a table that implements `allowStaleWhileRevalidate()` gets stale-served single-record `get`s, but queries touching the same expired records always block on the upstream source fetch. For a query sweeping many expired rows this means serial upstream fetch latency inside query iteration, which is exactly the case SWR exists to avoid.

## Considerations for a fix

- There is no loaded resource instance in the transform; the reviewer-suggested `this.prototype` as the `resource` would work for prototype-defined `allowStaleWhileRevalidate` implementations but needs design review (the method is documented as an instance method receiving `(entry, id)`).
- Serving stale rows from queries while revalidating in the background is a user-visible behavior change and should come with tests covering query iteration over expired entries.

## Context

Surfaced by review on #1577 (which threads a `resource` argument through these functions for `wasLoadedFromSource()` but deliberately did not change this call site). Related: #1571, #1575, #1576.

Behavior predates #1575/#1577; observed on v5.1 (5.1.15) and the same code exists on main.

_Filed by an LLM (Claude Fable 5) at the user's request._

kriszyp:
## Summary

Fixes #1578. When a query/search hits an expired or invalidated entry, `transformEntryForSelect` reloads it via `ensureLoadedFromSource` — but it passed **no resource**, so the stale-while-revalidate early-return could never fire. A table implementing `allowStaleWhileRevalidate()` got stale-served single-record `get`s, yet queries touching the same expired rows always **blocked on the upstream source fetch** — serial upstream latency inside query iteration, exactly what SWR exists to avoid.

## Fix

The single-record `get` path passes its loaded resource instance as the `resource`; queries have none. Since `allowStaleWhileRevalidate` is an instance method, the query path now **constructs a resource instance per expired/invalidated row** (via the same `new constructor(id, context)` the framework's `getResource` uses) and threads it through.

Key design points:
- **Real instance, not the class prototype.** Every resource prototype chain ends in a tracked-property `Proxy`; reading an absent property on a *non-instance* (probing for an undefined hook, or a user hook touching `this.x`) invokes `getChanges()` with no backing state and throws. A real instance degrades gracefully, exactly as the single-record path does for non-SWR tables.
- **Per row, not memoized per query.** Constructing per row keeps `this.getId()` / instance state correct for row-dependent policies (matching single-record semantics). It only runs for rows already headed to source, so the allocation is negligible against the source fetch.
- **Non-SWR tables unchanged** — an absent hook resolves to `undefined` through the instance's proxy and the query still blocks on source.

## Tests

Adds query-iteration tests over expired rows (in `unitTests/resources/caching.test.js`):
- **SWR**: serves the stale row while revalidating in the background — uses a deterministic source gate (no timing race).
- **Non-SWR**: still blocks on source (guards against silent behavior change).
- **Per-row identity**: a `this.getId()`-based policy sees each row's own id across a multi-row query (verified to fail under a memoized single instance).

Local: 160 passing across the caching / query / query-tier1 / resources unit suites.

## Cross-model review

Reviewed with Codex + Gemini (thorough) before opening. Both flagged the initial approach (which passed the bare prototype guarded by a `hasResourceMethod` probe): prototype-as-`this` breaks hooks that read instance state, and the probe could invoke a shadowing accessor. Reworked to the per-row real-instance approach above, which resolves both. A Codex re-review then caught a residual **row-identity** bug in an interim memoized variant (first row's id reused for all rows) — fixed by constructing per row, now covered by the per-row-identity regression test. Gemini's remaining notes (constructor arg, test flakiness) are addressed (real id passed; deterministic gate).

Behavior predates #1575/#1577; observed on v5.1 (5.1.15) and main.

_Authored with Claude Opus 4.8 (1M context)._

### DIFF
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -3039,6 +3039,7 @@ export function makeTable(options) {
 			}
 			let transformCache;
 			const source = this.source;
+			const resourceClass = this;
 			// Transform an entry to a record. Note that *this* instance is intended to be the iterator.
 			const transform = function (entry: Entry) {
 				let record;
@@ -3076,7 +3077,17 @@ export function makeTable(options) {
 								message: 'This entry has expired',
 							};
 						}
-						const loadingFromSource = ensureLoadedFromSource(source, entry.key ?? entry, entry, context);
+						// Stale-while-revalidate is an instance method, but a query has no per-row resource
+						// instance to consult (the single-record `get` path passes `this`). Construct one for this
+						// row — via the same `new constructor(id, context)` the framework's getResource uses — so
+						// the hook sees the current row's identity (this.getId()), matching the single-record path.
+						// It must be a real instance, not the bare class prototype: every resource prototype chain
+						// ends in a tracked-property Proxy, so reading an absent property (probing for an undefined
+						// `allowStaleWhileRevalidate`, or a hook touching `this.x`) on a non-instance invokes
+						// getChanges() with no backing state and throws (harper#1578). This runs only for the
+						// expired/invalidated rows already headed to source, so the allocation is negligible.
+						const swrResource = new resourceClass(entry.key ?? entry, context);
+						const loadingFromSource = ensureLoadedFromSource(source, entry.key ?? entry, entry, context, swrResource);
 						if (loadingFromSource?.then) {
 							return loadingFromSource.then(transform);
 						}
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -3080,13 +3080,19 @@ export function makeTable(options) {
 						// Stale-while-revalidate is an instance method, but a query has no per-row resource
 						// instance to consult (the single-record `get` path passes `this`). Construct one for this
 						// row — via the same `new constructor(id, context)` the framework's getResource uses — so
-						// the hook sees the current row's identity (this.getId()), matching the single-record path.
-						// It must be a real instance, not the bare class prototype: every resource prototype chain
-						// ends in a tracked-property Proxy, so reading an absent property (probing for an undefined
-						// `allowStaleWhileRevalidate`, or a hook touching `this.x`) on a non-instance invokes
-						// getChanges() with no backing state and throws (harper#1578). This runs only for the
-						// expired/invalidated rows already headed to source, so the allocation is negligible.
+						// the hook sees the current row's identity (this.getId()) and record state, matching the
+						// single-record path. It must be a real instance, not the bare class prototype: every
+						// resource prototype chain ends in a tracked-property Proxy, so reading an absent property
+						// (probing for an undefined `allowStaleWhileRevalidate`, or a hook touching `this.x`) on a
+						// non-instance invokes getChanges() with no backing state and throws (harper#1578). We also
+						// load the stale entry into it (as the single-record path does via _updateResource) so a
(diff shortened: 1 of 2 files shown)


resources/Table.ts
```diff
@@ -3076,7 +3077,17 @@ export function makeTable(options) {
...
+						// Stale-while-revalidate is an instance method, but a query has no per-row resource
+						// instance to consult (the single-record `get` path passes `this`). Construct one for this
+						// row — via the same `new constructor(id, context)` the framework's getResource uses — so
+						// the hook sees the current row's identity (this.getId()), matching the single-record path.
+						// It must be a real instance, not the bare class prototype: every resource prototype chain
+						// ends in a tracked-property Proxy, so reading an absent property (probing for an undefined
+						// `allowStaleWhileRevalidate`, or a hook touching `this.x`) on a non-instance invokes
+						// getChanges() with no backing state and throws (harper#1578). This runs only for the
+						// expired/invalidated rows already headed to source, so the allocation is negligible.
+						const swrResource = new resourceClass(entry.key ?? entry, context);
```
cb1kenobi, line 3093: **Low: query SWR instance is unloaded, so parity with the single-record path is only partial**

`new resourceClass(id, context)` builds an instance with id and context but no loaded record: unlike the single-record `get` path, which passes `this` *after* `_loadRecord`/`_updateResource` has populated the record (Table.ts:723,737), nothing loads the stale entry into `swrResource` here. So an `allowStaleWhileRevalidate` hook that consults instance record state (e.g. `this.getRecord()` or `this.someField`) sees data on `get` but nothing on a query, even though the comment states this matches the single-record path. `getId()` matches (and is tested); record state does not.

The `(entry, id)` arguments are the safe contract (`entry.value` is the stale record). Either document that the hook must rely on its `(entry, id)` args rather than `this.*` beyond `getId()`, or load the entry for full parity, e.g. `resourceClass._updateResource(swrResource, entry)` before the call.

—
_Generated by Barber AI_

kriszyp: Good catch — went with full parity (option 2). `3a1538d` now calls `resourceClass._updateResource(swrResource, entry)` for entry/version metadata, then `swrResource.setRecord(record)` using the already-dereferenced `record` (a lazy query entry's `.value` can be a GC'd deref, so we don't rely on `entry.value`). A hook now sees the same `this.getRecord()`/`this.<field>` state on a query as on `get`. The test hook asserts the loaded record is visible and its id matches the row. Thanks for the precise line-level note!

cb1kenobi (approved):

PR #1582, opened 2026-07-02
Two-tier component secret delivery + env declarations (#1550)

### LINKED ISSUE #1550: Per-component process.env from granted secrets, with load-gating on declared secrets

kriszyp:
Closes #1550. **Stacked on #1554** (`kris/secrets-store-715`) — the diff shows only the consumption side. Retarget to `main` after #1554 merges; do **not** merge #1554 via squash while this is stacked.

## Summary

Consumption of the `hdb_secret` store, delivered in two tiers:

- **Global tier** (rows with empty `grants`): decrypted at the start of each component-load cycle and materialized into the **real** `process.env`, before applications load. Precedence: a pre-existing real env var always wins (logged). Child processes inherit these — same semantics as `.env` today. No isolation promise.
- **Scoped tier** (rows with non-empty `grants`): **never** in `process.env` — not inherited by child processes, invisible to env dumps. Exposed only through the per-component accessor: `import { secrets } from 'harper'` and `scope.secrets`. The accessor is a superset view: scoped rows granted to the component **plus** its declared global-tier names, so app code uses `secrets.FOO` uniformly and ops can later tighten a secret from global to granted without breaking the app.

Component configs declare env expectations in an `env:` block:

```yaml
env:
  NODE_ENV: production                                        # string → inline literal into process.env (.env semantics, supports enc:v1:)
  STRIPE_SECRET_KEY: { required: true, description: Stripe }  # object → declaration, satisfied from the store
  SENTRY_DSN: { required: false }
```

An unsatisfied `required: true` declaration gates that component's load (reason: `missing` | `ungranted` | `custody-unavailable`); the instance keeps running and nothing of the gated component is registered. Literals are applied only after the whole block validates and the gate passes. `get_components` exposes the declared-but-unsatisfied set per component (`unsatisfiedEnv`: name/description/required/reason/tier — metadata only, never values) for Studio configure-me rendering.

## Accessor binding (the investigated part)

- **vm, compartment, and the default `vm-current-context` loader**: `import { secrets } from 'harper'` (and CJS `require('harper')` inside the app loader) resolves to the per-scope synthetic `harper` module (`security/jsLoader.ts` `getHarperExports`) — **exact** binding. This includes the default mode: the `harper` specifier is intercepted by `resolveModule` before Node resolution, so the symlinked package is not consulted.
- **native loader / bundler-loaded code / natively-loaded deps**: those reach the process-wide singleton package, whose `secrets` export resolves the loading component via an AsyncLocalStorage context that `scopedImport` binds for the duration of each component load. Access from an ambiguous context (e.g. lazily at request time) **fails loudly** with an error naming the component-resolution problem — it never guesses.
- Recommended idiom (documented in the module header): destructure at module top level — `const { STRIPE_SECRET_KEY } = secrets;` — where binding is exact in every mode.

JS-level scoping is documented as a **slowdown layer, not a security boundary** (shared process; containers/uids are the boundary; no SES hardening claimed).

## Degraded modes

- No custody registered (OSS core / material missing): required declarations report `custody-unavailable`; global rows log and stay out of env; node boots.
- `hdb_secret` table missing (pre-upgrade): store treated as empty, `missing` reasons note the store is unavailable; node boots.
- Late custody / changed secrets: heal on restart or reload cycle (each cycle re-reads the store; env-declaring component loads also refresh the snapshot so deploy validation in a long-lived worker gates against post-boot `set_secret`/`grant_secret`). No live re-materialization in v1.

## Where to focus review

- **The `secrets` export is also an ambient global** (via `_assignPackageExport`). Its loud-failure traps mean `util.inspect`/`console.log` of a container holding it (e.g. `console.log(global)`) outside a load context throws. Deliberate trade-off for "never silently hand back another component's secrets"; symbols and `then` are exempted so `await`/inspect probes don't explode.
- **Main-thread reload semantics**: `loadedPaths` is never cleared in production, so already-loaded components early-return on main-thread reload; the declaration registries are therefore overwrite-on-reprocess (a cycle-level wipe would permanently empty `get_components.unsatisfiedEnv` — caught by the cross-model review leg).
- **Per-load store scan**: each env-declaring component load re-scans `hdb_secret` (small system table, load-time only) to keep out-of-cycle gates fresh.
- **Gate containment**: gate failure = status `failed` + log + early return; it deliberately does **not** install an `ErrorResource` at `/` (the pre-existing per-plugin catch does, which can clobber the root URL space — not repeated here).

## Explicitly out of scope (follow-ups)

- `sensitive: false` plaintext rows in `hdb_secret` (ops-managed non-secret vars with the same grants/declaration/UI machinery) — touches #1554's operations while in review.
- Live re-materialization on secret change (v1 = restart/reload heals).
- Truly process-level vars (`NODE_OPTIONS`, `UV_THREADPOOL_SIZE`) — need container-level injection (the P3 surface), not the store.
- `env:` in the **root** config is warn-and-ignore for now (root loads before custody registration/materialization; needs its own ordering story).
- Docs: user-facing (`env:` block + accessor); companion documentation PR to follow when this retargets to main.

## Tests

`unitTests/components/componentSecrets.test.js` (42: materialization/precedence/retraction, literal + enc:v1: semantics, declaration shapes, all three gate reasons, accessor visibility/superset/enumeration/collision, ALS binding incl. concurrent loads, loud ambiguous failure, proxy invariants) and `unitTests/components/componentSecretsGate.test.js` (2: loader-level gate containment — component fails, instance survives, literals land on the passing path). Existing `loadEnv` behavior untouched.

Cross-model review (adjudicated; all fixes in follow-up commits):

- **Codex leg** — 6 findings, all confirmed and fixed: root-resource poisoning on gate failure, literal-before-gate env mutation, stale declaration registries, `scope.secrets` identity mismatch on RUN_HDB_APP paths, stale store snapshot on deploy-path loads, missing `preventExtensions` trap. Plus 1 valid residual on the first fix round (a cycle-level registry wipe vs. the never-cleared `loadedPaths`) — resolved with overwrite-on-reprocess semantics.
- **Gemini leg** — 7 findings: 4 confirmed and fixed (concurrent-load scan/decrypt herd + out-of-order snapshot overwrite → single-flight `materializeGlobalSecrets`; gate rejecting an env fallback the accessor would serve for granted-but-undecryptable rows → gate now matches; `Object.prototype` names masquerading as secrets under dynamic access → null-prototype views). 1 documented-as-designed (env literals share `process.env` cross-component with load-order visibility — inherent to `.env` semantics, now stated in the module docs). 2 dismissed with evidence: the "Blocker" claim that AsyncLocalStorage context is lost during dynamic-import ESM top-level evaluation is empirically false on our supported Node range (guarded by a permanent regression test covering top-level read, destructure, and post-top-level-await); the `catch (error)` primitive-throw concern doesn't apply (the guarded block only throws `Error` instances and matches the loader's existing pattern).

Generated by Claude (Fable) with cross-model review by Codex and Gemini.


components/componentSecrets.ts
```diff
@@ -0,0 +1,415 @@
...
+		ownKeys() {
+			return Reflect.ownKeys(resolveBoundSecrets());
+		},
+		getOwnPropertyDescriptor(_target, property) {
+			if (typeof property === 'symbol') return undefined;
+			const descriptor = Object.getOwnPropertyDescriptor(resolveBoundSecrets(), property);
+			// The views are frozen; report configurable so the proxy invariant against its (extensible,
+			// empty) target holds.
+			return descriptor && { ...descriptor, configurable: true };
+		},
```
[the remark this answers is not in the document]
kriszyp: Fixed in b8e78bc54 — the enumeration traps (`has`/`ownKeys`/`getOwnPropertyDescriptor`) now report an empty object when no component-load binding is active, so inspectors/serializers (`util.inspect`, spread, `in`) can never crash the process; direct property reads (`secrets.MY_KEY`) stay loud. This was a genuine gap: the docstring already promised inspect-safety but only the `get` trap delivered it. Tests updated to pin both halves (loud direct read, safe enumeration incl. `util.inspect` and spread). Good catch!

— Claude (Fable 5)

index.ts
```diff
@@ -90,6 +90,7 @@ import type { Logger } from './utility/logging/logger.ts';
 import type { models as ModelsImport } from './resources/models/Models.ts';
 import type { operation as OperationImport } from './server/serverHelpers/serverUtilities.ts';
 import type { Resource as ResourceImport } from './resources/Resource.ts';
+type SecretsImport = Readonly<Record<string, string>>; // per-component secrets view (#1550)
```
dawsontoth: Could this type live in componentSecrets.ts, or closer to it? And be... imported...

unitTests/components/componentSecrets.test.js
```diff
@@ -0,0 +1,528 @@
...
+		table.restore();
+		for (const key of ENV_KEYS) {
+			const prior = savedEnv.get(key);
+			if (prior === undefined) delete process.env[key];
+			else process.env[key] = prior;
+		}
+	});
+
+	describe('global-tier materialization', () => {
+		it('materializes empty-grants rows into the real process.env', async () => {
```
dawsontoth: 🤔 I like the capability of surfacing the secrets in the real process.env, but I'm not inline with the triggering of it being when there are empty-grants. The grants map back to component names? If this is the best way to do it all, that's ok. But if we could be clearer in the contract when something will surface in the process.env (mentioning it by name?) that feels more transparent and expected to me.

kriszyp: Yeah, this is a pretty critical aspect of the design, and I don't know that the PR description does it justice. There are several important considerations here:
* There are many 3rd party packages out there that rely on (at least by default) `process.env` to read variables/secrets.
* The application scoping's default configuration, vm-current-context, does not isolate the global scope, so we can't make `process.env` be specific each application.
* The majority of Harper deployments only have a single substantive application (that would actually be consuming secrets).
* Our application scopes are not truly hardened, they don't create a real hermetic boundary between components anyway.

With this in mind, I decided to go with two distinct paths for secrets:
* The generic unscoped (no application scoping) secrets, a default path that is easily satisfies 3rd party modules, and in the majority case, doesn't worry about hiding secrets from other components that probably aren't there.
* The scoped secrets that really are scoped to each component. These require specifically accessing them through `import { secrets } from 'harper'`. Requires explicit code, but most secure mechanism if an app is willing to code for it.

I do wonder if the "grant" vs "scope" terminology is little confusing here ("unscoped" kinda implies broadly available, "not granted" kinda implies the opposite)? It does sound like clarity is perhaps what you are most interested in here.

dawsontoth: Na, I get that, what I'm really imagining slightly different is how we describe where a secret should show up.

Like... look at these two examples. First, env scoped:
```
{
    "name": "API_KEY",
    "value": "encrypted secret stuff goes here",
    "exposeAs": "process.env"
}
```
That can be accessed anywhere as `process.env.API_KEY`, presumably. But if you want more protections:
```
{
    "name": "OTHER_KEY",
    "value": "other encrypted secret stuff goes here",
    "exposeAs": "harper.secret",
    "components": ["status-check"]
}
```

Does that make sense now, what I'm thinking about? It's more about how we let developers configure their secrets. I like that it communicates the trade-off a bit too. And if you try to do `"process.env"` with `"components"`? The types would fail!
```
type SecretDefinition = ProcessEnvSecret | HarperSecret;

interface ProcessEnvSecret {
    name: string;
    value: string;
    exposeAs: "process.env";
}

interface HarperSecret {
    name: string;
    value: string;
    exposeAs: "status-check"
    components?: string[];
}
```

kriszyp: Yeah, that makes sense. I guess it seems little on the verbose side (and `harper.secret` doesn't actually represent a parallel valid syntax). I submit as evidence your typo (I think) of `exposeAs: "status-check"` :). And using "components" instead of "grants" feels like it shifts the expectation to omission meaning that it is not constrained, broadly available. Also worth noting that without a scoping to a component/app, the secrets *are* still available on `import { secrets} from 'harper'`.
Perhaps another type to consider (middle-ish verbosity?):
```
type SecretDefinition = ProcessEnvSecret | ScopedSecret;

interface ProcessEnvSecret {
    name: string;
    value: string;
    processEnv: true;
}

interface ScopedSecret {
    name: string;
    value: string;
    grants?: string[];
}
```

dawsontoth: sure, that looks good to me

index.ts
kriszyp: Done in 3c6c3666b — extracted \`SecretsView\` (exported from \`components/componentSecrets.ts\`, right next to the accessor it describes) and index.ts now imports it as a type. Same alias (\`SecretsImport\`) so the ambient/export wiring is untouched. Thanks!

— KrAIs (Opus 4.8)

unitTests/components/componentSecrets.test.js
kriszyp: Implemented the explicit-tier shape we landed on. A secret now declares its tier directly instead of it being inferred from empty grants:

- `processEnv: true` → materialized into the real `process.env` (global, `.env` semantics).
- `grants: [...]` → scoped, accessor-only, restricted to the listed components.
- The two are mutually exclusive (a processEnv secret is already global, so scoping it is rejected in `set_secret`/`grant_secret`). A row with neither is inert until granted — omission stays restrictive, as you wanted.

Split across the stack: store + ops (schema `processEnv` attribute, `set_secret`/`grant_secret`/`list_secrets`) went to #1554; consumption (materialization + gate + accessor now key on the flag) stayed here. Full secret suite green (116 tests).

— KrAIs (Opus 4.8)

dawsontoth (approved):

Issue #1571, opened 2026-07-02
loadedFromSource is unobservable with string-id Table.get() — set on an internal RequestTarget; Context declares it but it is never assigned


kylebernhardy:
## Summary

For a caching table (`sourcedFrom()`), the cache-disposition flag `loadedFromSource` is set on the **RequestTarget of the `get` call** — and when the caller passes a plain string id (the most common calling convention), the static-get dispatch mints an *internal* `RequestTarget` the caller never sees. The flag is faithfully set on that internal object and discarded. There is no way to observe hit/miss from resource code with a string id.

Compounding it: the `Context` interface *declares* the field (`resources/ResourceInterface.ts:98` — `loadedFromSource?: boolean`), which implies it should be readable from the request context, but **no code path ever assigns it to a context**. Reading `context.loadedFromSource` after `Cache.get(stringId)` silently yields `undefined`.

## Impact

Any app computing hit/miss this way records **every lookup as a cache hit** — silently wrong data, no error. In StormScout this was a latent day-one bug: every `Server-Timing … desc=` value and stored cache-status row said `hit`, including 2–3 s upstream source fetches.

## Repro

```ts
// resource method, table wired with sourcedFrom()
await Cache.get(key);                                        // string id
console.log((this.getContext() as any).loadedFromSource);    // undefined — always

import { RequestTarget } from 'harper';
const t = new RequestTarget();
t.id = key;
await Cache.get(t);
console.log(t.loadedFromSource);   // true on source fetch, false on cache hit — works
```

Verbatim probe output (5.1.6 and 5.1.15, identical):

```
"f1_context_flag_after_string_get": null,
"f1_target_flag_after_target_get": false
```

## Code locations (5.1.15)

- `resources/Resource.ts:637` — string-id branch of static `get` dispatch: `query = new RequestTarget(); query.id = id;` (internal target, never returned to the caller)
- `resources/Table.ts:4804` — `target.loadedFromSource = true` (this request acquired the resolution lock and fetched from source)
- `resources/Table.ts:4789` — `= false` after waiting on another request's in-flight fetch (dedupe join — so "hit" with a multi-second duration is expected behavior, worth documenting)
- `resources/Table.ts:730` / `:746` — `= false` on plain cache hits
- `resources/ResourceInterface.ts:98` — declared on `Context`, never assigned

## Suggested fix

[...]

PR #1590, opened 2026-07-03
Make target.loadedFromSource the sole cache-disposition signal; drop the Context mirror (#1576)

### LINKED ISSUE #1576: wasLoadedFromSource() is documented as reporting cache disposition but is an unoverridden stub that always returns undefined
## Summary

The docs ([Resource API reference, `wasLoadedFromSource(): boolean`](https://docs.harperdb.io/docs/reference/resources/resource-api#wasloadedfromsource-boolean)) state:

> For caching tables, indicates that this request was a cache miss and the data was loaded from the source resource.

But the only implementation is the base-class stub, which no subclass (including the caching `TableResource` produced by `makeTable`) ever overrides:

```ts
// resources/Resource.ts:71
wasLoadedFromSource(): boolean | void {
	// Subclasses should override if needed
}
```

`grep -rn "wasLoadedFromSource" resources/` shows only the interface declaration (`resources/ResourceInterface.ts:52`) and this stub. Calling it on a caching-table resource instance always returns `undefined`, regardless of disposition.

Additionally, the default `get()` on a table returns a frozen `RecordObject`, not a resource instance, so the method is not even reachable in the place most developers would look for hit/miss information.

## Suggested resolution (either)

1. Implement the override on caching tables — e.g. return the same value that #1575 now records via `setLoadedFromSource()` (target/context mirroring), or derive it from the entry's disposition; or
2. Remove the method from the interface and the docs section, since #1575 makes `context.loadedFromSource` (and `target.loadedFromSource`) the supported way to observe cache disposition.

## Related

- #1571 — `loadedFromSource` was unobservable with string-id gets (root observability gap)
- #1575 — fix making `context.loadedFromSource` work; its companion docs PR HarperFast/documentation#563 originally cross-referenced `wasLoadedFromSource()` and had to remove the reference because of this stub (found via cross-model review on that PR).

Found on branch v5.1 (5.1.15); the stub is identical on main.

_Filed by an LLM (Claude Fable 5) at the user's request._

kriszyp:
Follow-up to #1575 (already merged to `v5.1`). Fixes #1576.

## Why

#1575 mirrored cache disposition onto the request **Context** so plain-id gets could observe it (#1571). But `loadedFromSource` is fundamentally a **per-`get()` result**: each `get()` has its own `RequestTarget`, whereas the `Context` is shared for the whole request lifecycle. Mirroring onto the shared Context makes it **last-write-wins** — an internal caching `get()` inside a resource's own `get()` overwrites the entry get's disposition before the caller reads it.

The REST handler already reads the precise per-get value from `target.loadedFromSource` ([`server/REST.ts:231`](https://github.com/HarperFast/harper/blob/v5.1/server/REST.ts)). And #1571's "unobservable on string-id gets" is really a usage question best answered by docs: **observe disposition on the `RequestTarget` you pass to the get.** So the Context mirror is code papering over a documentation gap, and it introduces a semantic that doesn't hold up.

## What

Remove the Context mirror and standardize on `target.loadedFromSource`:
- `setLoadedFromSource()` now writes **only** the target (keeping the primitive-id guard #1575 added); Context param and all context writes dropped.
- Removed the context-only marking in `ensureLoaded()` (it passed `undefined` target — pure context write).
- Removed the now-unused `Context.loadedFromSource` declaration (it predated #1575 as declared-but-never-assigned).

## Retained from #1575 (the parts that were genuinely right)

- **Per-get `false`-marking on cache HITS** in the `loadAsInstance=false` value path and `ensureLoaded` (`Table.ts` ~1227/~1240). Before #1575 those paths set the flag on source fetch but left it `undefined` on cache hits, so `REST.ts` (`if (loadedFromSource !== undefined)`) **skipped the `Age` header on hits**. This is a real `target`-side fix, independent of the mirror.
- The **primitive-target guard** in the helper.

## Testing

Caching unit suite converted to observe `target.loadedFromSource`; **15/15 passing** against `dist` (incl. the `loadAsInstance=false` hit/miss case and stale-while-revalidate). The plain-id "observable on context" test was removed (its premise — the mirror — is gone; target hit/miss is covered by the existing `Can load cached data` test).

## Docs

HarperFast/documentation#563 reframed to make `target.loadedFromSource` canonical and drop `context.loadedFromSource`.

## Note for downstream

The OCSP/CRL cert-verification code (`security/certificateVerification/*`) does plain-id gets and previously read the (always-`undefined`) `wasLoadedFromSource()`. With the mirror gone, disposition there must be observed by passing a `RequestTarget`; tracked separately.

_Authored by KrAIs (Claude Opus 4.8), at Kris's direction._

### DIFF
--- a/resources/RequestTarget.ts
+++ b/resources/RequestTarget.ts
@@ -56,7 +56,8 @@ export class RequestTarget extends URLSearchParams {
 	declare previousResidency?: string[];
 
 	// Action tracking
-	/** Cache disposition of a get on a caching table; also mirrored onto the request Context. */
+	/** Cache disposition of this get on a caching table: true if loaded from the source, false if
+	 * served from cache. Set per-get; read it on the RequestTarget you passed to the get. */
 	declare loadedFromSource?: boolean;
 	declare createdNewId?: string;
 
--- a/resources/ResourceInterface.ts
+++ b/resources/ResourceInterface.ts
@@ -94,12 +94,6 @@ export interface Context {
 	sourceApply?: boolean;
 	originatingOperation?: OperationFunctionName;
 	previousResidency?: string[];
-	/** Cache disposition of the most recent get on a caching table in this context: true if the get
-	 * fetched from source — including when a source error fell back to a stale record (staleIfError);
-	 * false if served from cache — including stale-while-revalidate responses (the source fetch
-	 * continues in the background) and waits on another request's in-flight source fetch.
-	 * Subsequent gets in the same context overwrite it. */
-	loadedFromSource?: boolean;
 	nodeName?: string;
 	resourceCache?: Map<Id, any>;
 	_freezeRecords?: boolean; // until v5, we conditionally freeze records for back-compat
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -727,7 +727,7 @@ export function makeTable(options) {
 							// return 504 (rather than 404) if there is no content and the cache-control header
 							// dictates not to go to source
 							if (!this.doesExist()) throw new ServerError('Entry is not cached', 504);
-							if (hasSourceGet) setLoadedFromSource(target, request, false); // mark it as cached
+							if (hasSourceGet) setLoadedFromSource(target, false); // mark it as cached
 						} else if (resourceOptions?.ensureLoaded) {
 							const loadingFromSource = ensureLoadedFromSource(
 								(this.constructor as any).source,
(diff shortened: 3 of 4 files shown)


kriszyp:
Closing — v5.1's #1575 (the base this PR reworks) was reverted after we caught it riding into an unintended v5.1.16 patch (it should not have shipped as a patch). The Context-mirror bug and the fix this PR describes both only ever existed on `v5.1`, which no longer carries any of it.

The net end-state this PR was going for (target-only `loadedFromSource`, Age-header-on-hits fix, primitive guard, no Context mirror) has been ported to `main`/5.2 as one clean PR instead: #1626.

— KrAIs

Issue #1595, opened 2026-07-03
Model backends have no retry/backoff — a single transient provider error fails the whole write for @embed


kylebernhardy:
## Summary

Split out from #1593 (contributing factor 2). The model backends (`components/openai/index.ts#post` and siblings) throw immediately on any non-OK status — no `429`/`Retry-After` handling, no backoff on `5xx`. Because `@embed` computes embeddings per record at write time, a single transient provider error fails the whole write, and for a bulk ingest, the batch. The only resilience today is `requestTimeoutMs` and multi-candidate failover (which requires several backends configured under one logical name).

## Suggested direction

Retry-with-backoff on retriable statuses (`429`, `502`, `503`, `529`), honoring `Retry-After` when present, in the backend layer (so all callers benefit) with sane caps (e.g. 2-3 attempts, jittered exponential backoff, overall deadline respecting `requestTimeoutMs` and the caller's abort signal). Design questions: which layer (backend post vs Models dispatch), interaction with multi-candidate failover ordering, and whether @embed bulk writes should batch/queue instead.

## Context

- #1593 fixed the wire-model resolution bug and improved error surfacing; retry was deliberately left out of that PR as a behavior/design change.
- Found while dogfooding the docs-site replatform (bulk embedding ingest).

_Filed by an LLM (Claude Fable 5) while working #1593._

kriszyp:
Appears to be a duplicate of / consolidate under #1594 — model-backend retry/backoff — #1594 covers both resilience gaps.

— Claude (Opus 4.8), issue-backlog triage

heskew:
Duplicate of #1594 — consolidated there (its body now carries this issue's retriable-status list, backoff/deadline caps, and the layer decision under "Consolidated from #1595"). Closing per the duplicate triage.

PR #1596, opened 2026-07-03
@embed / models.embed no longer forwards the logical model name as the provider wire model id

### LINKED ISSUE #1593: @embed / models.embed: logical model name is forwarded as the wire model id (404 when they differ), and the error is unhelpful
### Bug Summary

When a model's logical name differs from the provider's model id, `@embed` (and `models.embed({ model })`) sends the **logical name** to the provider as the wire model, producing a 404. The failure is compounded by a sanitized error that hides the cause and by no retry/backoff in the backend.

### Environment

- Harper **5.1.15**, Node v24.16.0, macOS
- Embedding via the `openai` backend pointed at Google's Gemini OpenAI-compatible endpoint

### Configuration

```yaml
models:
  embedding:
    default:                 # logical name
      backend: openai
      baseUrl: https://generativelanguage.googleapis.com/v1beta/openai/
      apiKey: ${GEMINI_API_KEY}
      model: gemini-embedding-001   # provider/wire model id
```

Schema:

```graphql
type SearchChunk @table {
  # ...
  embedText: String
  embedding: [Float] @embed(source: "embedText", model: "default")
}
```

### Steps to Reproduce

1. Configure the embedding model above (logical name `default`, wire model `gemini-embedding-001`).
2. Write any record to a table with the `@embed` attribute (e.g. via an ingest resource).
3. The write fails; the app-facing error is `Failed to compute embedding for attribute "embedding"` (500).
4. The server log shows the real error:
   `OpenAIBackendError: OpenAI /embeddings returned HTTP 404: models/default is not found ... not supported for embedContent`

Note the provider received `models/**default**` — the *logical* name — not `gemini-embedding-001`.

### Expected Behavior

`@embed(model: "default")` / `models.embed(text, { model: "default" })` should resolve `default` to the configured entry and call the provider with that entry's `model` (`gemini-embedding-001`).

### Actual Behavior

The logical name is used verbatim as the wire model, so any config where `logical name !== provider model id` 404s. It works only when the logical name happens to equal the provider id, or when no `model` option is passed (falls through to the backend's configured default).

### Root Cause (from the installed source)

- `resources/models/embedHook.ts` → `createDefaultEmbedder` calls `models.embed(text, { model, inputType: 'document' })` with the directive's `model` (the logical name).
- `resources/models/Models.ts` → `resolveCandidates` uses `model` as the logical name to select the backend (correct), **but the same `opts.model` is then passed through to the backend call**.
- `components/openai/index.ts` → `embed()` does `const model = opts.model ?? this.#defaultModel`, so a provided `opts.model` **overrides** the configured `model` and becomes the wire model id.

So `opts.model` is doing double duty: logical selector *and* wire override. `models.embed(text)` (no option) works precisely because `opts.model` is undefined and the backend uses its configured `#defaultModel`.

### Contributing Factors

1. **Unhelpful surfaced error.** `embedHook.ts` rethrows a generic `Failed to compute embedding for attribute "embedding"`; the actual 404 with the offending model name is only at `error` level in the server log. The sanitized message sent me down a rate-limit rabbit hole before I found the log line.
2. **No retry/backoff.** `components/openai/index.ts#post` throws immediately on any non-OK status — no `429`/`Retry-After` handling, no backoff. Combined with `@embed`'s per-record write-time embedding, a single transient provider error fails the whole write (and, for a bulk ingest, the batch). The only resilience is `requestTimeoutMs` and multi-candidate failover (only if several backends are configured under one logical name).

### Suggested Fixes

- Don't forward the logical name as the wire model. After `resolveCandidates` selects the backend, call it **without** overriding `opts.model` (let the backend use its configured `model`), or pass the resolved entry's `model` rather than the logical name.
- Surface the backend error's status + message through the `@embed` failure (or at least reference the log line) so the wire-model mismatch is diagnosable without reading server logs.
- Consider retry-with-backoff on retriable statuses (`429`, `5xx`, honoring `Retry-After`) in the backend or embed hook — `@embed`'s per-record model makes bulk embedding fragile without it.

### Workaround

Name the logical model to equal the provider model id:

```yaml
models:
  embedding:
    gemini-embedding-001:
      backend: openai
      baseUrl: https://generativelanguage.googleapis.com/v1beta/openai/
      apiKey: ${GEMINI_API_KEY}
      model: gemini-embedding-001
```

and `@embed(source: "embedText", model: "gemini-embedding-001")`. The docs' own example uses `model: "default"`, which hits this bug whenever the provider id differs.

### Context

Found while dogfooding the docs-site replatform (semantic search over Harper docs, embeddings via `@embed`).

kylebernhardy:
Fixes #1593.

## Summary

`opts.model` on the models facade did double duty: it selects the configured logical entry (via `resolveCandidates`) **and** was spread into the backend call options, where every backend does `opts.model ?? configuredModel`. So `@embed(model: "default")` sent the literal string `default` to the provider as the wire model — a 404 whenever the logical name differs from the provider model id (the docs' own example config hits this).

A `toBackendOpts()` helper in `resources/models/Models.ts` strips the logical name at all three dispatch sites (`embed`, `generate`, `generateStream`); backends now always use their configured wire model. Callers could never legitimately pass a wire model directly — `resolveCandidates` throws `ModelBackendNotFoundError` for any name that isn't a configured logical entry — so this closes the accidental override path without removing a real capability. Analytics still records the logical name (the helper clones before deleting).

Second fix: the `@embed` sanitized error (deliberately terse — raw backend errors can carry URLs/key tails) now includes safe identifiers: the error class name and the provider's HTTP status, via a new structured `upstreamStatus` field on all four backend error classes. `upstreamStatus` is distinct from `ServerError.statusCode` (Harper's own response status, which defaults to 500 and would have been misleading here — caught by cross-model review).

## Where to look

- **One correctness-enforcing break for invalid configs:** an entry that omitted its backend `model` setting previously worked by accident when the caller's logical name doubled as the wire model; it now fails with the backend's "model required" error. Valid configs (including the issue's workaround where names are equal) behave identically.
- `upstreamStatus` is set at the three HTTP throw sites (openai/anthropic/ollama); bedrock's error class has the field for shape symmetry but its throws are SDK/setup errors with no upstream HTTP status.

## Testing

3 regression tests (one per dispatch method) assert the backend receives no `model` option — all verified failing without the fix; a sanitized-error test asserts class name + upstream 404 appear while key tails, hostnames, and Harper's own 500 do not. `test:unit:resources` passes (997).

Retry/backoff (the issue's third suggestion) is deliberately split out as #1595. Reported on 5.1.15 — v5.1 patch candidate once baked. No docs change needed: this makes the docs' existing `model: "default"` example work as documented.

_Generated by an LLM (Claude Fable 5), cross-reviewed by Codex (statusCode-vs-upstreamStatus finding addressed) and Gemini (approved, no blockers)._


kriszyp (approved):
Reviewed via Claude review-queue. Correct, well-scoped fix — `toBackendOpts()` stripping the logical model name at all three dispatch sites resolves a real wire-protocol bug (`opts.model` was silently winning over the provider's configured model id). Tests assert the fix directly.

One gap worth a fast follow-up rather than blocking: `BedrockBackendError` gains the `upstreamStatus` param, but bedrock's HTTP dispatch (`client.send(command)`) has no try/catch, so SDK `ServiceException`s propagate raw and never populate it. This is 3-of-4 providers covered for the stated #1593 goal ("provider's HTTP status... via upstreamStatus on all four backend error classes") — degrades safely (no crash), just an incomplete diagnostic for bedrock specifically. Worth wrapping `client.send` and mapping `err.$metadata?.httpStatusCode` → `upstreamStatus` in a follow-up.

Approving — solid fix, the bedrock gap doesn't need to block this.

dawsontoth (approved):

kriszyp:
Addressed the bedrock gap in-branch (commit bbc92fc9d) rather than deferring: `#sendWithAbort` — the single choke point all three dispatch methods route through — now catches SDK `ServiceException`s, re-throws as `BedrockBackendError` with `$metadata.httpStatusCode` → `upstreamStatus`, and preserves the raw SDK error as `cause` for the unsanitized server log. Aborts/timeouts propagate untouched. All four backends now populate `upstreamStatus`. 3 regression tests added (status mapped / no-status → undefined / abort passthrough); bedrock suite 34 passing.

— KrAIs 🤖

PR #1599, opened 2026-07-04
feat(auth): let registered operations declare permissions for scoped delegation

### LINKED ISSUE #1598: Registered operations can't declare permissions — forces hardcoded inline super_user checks
## Problem

Operations added via `server.registerOperation` have no way to declare their authorization requirements to the central `verifyPerms` / `requiredPermissions` system. As a result:

- They can't participate in the role `operations` allowlist (the two-gate check + SU-bypass in `utility/operation_authorization.ts`), so an operator can't grant a specific registered op to a scoped, non-super_user role.
- Components work around this with **hardcoded inline `requireSuperUser` checks** inside each handler (e.g. the built-in agent ops in `agent/operations.ts`), which is an all-or-nothing gate and duplicates auth logic outside the standard path.
- A registered op keyed by its handler's `.name` also collides when handlers are anonymous arrows (`execute: async (op) => ...` → all named `"execute"`), so even manual `requiredPermissions` registration isn't viable for them.

The invariant we want: **an operation declares its permissions through the standard authorization system**, rather than hand-rolling an inline check that the role model can't see or override.

## Motivation

The built-in agent (#626) exposes six super_user-only ops (`agent_prompt`, `get_agent_session`, `list_agent_sessions`, `cancel_agent_run`, `approve_agent_action`, `set_agent_config`). We want a **least-privilege delegation role** — a scoped identity that can drive the agent (`operations: ['agent']`) without full super_user — which is the two-principal direction from the WAF/agent security discussions. That's impossible while the ops are gated by inline `requireSuperUser`.

## Proposed direction

Add a general seam so `registerOperation` can declare a permission (e.g. `requiresSuperUser`), registering it into `requiredPermissions` keyed by the snake_case op name and making the handler's lookup key stable — so registered ops flow through `verifyPerms` and the existing role `operations` allowlist like built-ins do. Then convert the agent ops to declare their permission instead of the inline gate, and add an `agent` operation-permission group.

Backward compatible: only ops that opt in are affected; undeclared registered ops and super_user behavior are unchanged.

## Scope
- Core: `server.registerOperation` + `OperationDefinition` + a `requiredPermissions` registration helper in `operation_authorization.ts`.
- Agent: drop inline `requireSuperUser`, declare the ops SU-by-default (`agent/operations.ts`).
- Groups: add `agent` to `operationPermissions.ts`.

Targeting 5.2.

kriszyp:
Closes HarperFast/harper#1598. Supports the built-in agent (#626) delegation-role UX.

## Problem

Operations added via `server.registerOperation` had no way to declare their authorization to the central `verifyPerms` / `requiredPermissions` system, so they couldn't participate in the role `operations` allowlist and instead hand-rolled inline `super_user` checks (the built-in agent ops did exactly this). Anonymous arrow handlers also all share the name `"execute"`, so they can't be keyed by function name. Net effect: a registered op is all-or-nothing super_user, never grantable to a scoped least-privilege role.

## Change (general seam)

- **`OperationDefinition.requiresSuperUser`** (opt-in). When set, `registerOperation` registers the op with the central authz system via the new **`registerOperationPermission`** helper (`operation_authorization.ts`) and makes it grantable in a role's `operations` allowlist. Undeclared ops keep today's behavior — only opt-in ops are affected.
- The op is registered under its snake_case name (a fresh per-registration wrapper carries that name for the `verifyPerms` lookup), so it flows through `verifyPerms` and the existing two-gate `operations` allowlist + SU-bypass exactly like a built-in.
- **Agent ops (#626):** dropped the inline `requireSuperUser`, now declare `requiresSuperUser: true`. Same SU-only default, enforced centrally (a non-SU without a grant now gets a clean `403 OP_IS_SU_ONLY` instead of `400 OP_NOT_FOUND`).
- **`agent` operation-permission group** (`operationPermissions.ts`): `agent_prompt`, `get_agent_session`, `list_agent_sessions`, `cancel_agent_run`, `approve_agent_action`. An operator grants a least-privilege delegation role with `operations: ['agent']`. **Excludes `set_agent_config`** — changing the agent's model/provider/approval policy stays an operator (super_user) action.

## Authorization semantics

- Default: super_user only (`requiresSuperUser: true` → `requires_su` entry). super_user callers unchanged.
- A scoped, non-super_user role with `operations: ['agent']` (or individual ops) reaches the ops via the existing SU-bypass gate; RBAC is still enforced per call.
- Undeclared registered ops and all built-in ops are untouched.

## Cross-model review (thorough: Codex + Gemini + Harper-domain adjudication)

Two real issues surfaced and **fixed** (second commit):
1. **Shared-handler corruption (blocker, both models):** the original in-place `.name` mutation would corrupt the lookup if one handler were registered under two op names → now a fresh per-registration wrapper; the caller's function is never mutated.
2. **Dynamic ops not grantable (significant, Codex):** `validateOperations` (add_role/alter_role + impersonation) only accepted enum/group names → `registerOperationPermission` now also marks the name grantable, completing the general seam.

Adjudicated **not** actionable (documented for reviewers):
- **`set_agent_config` grantable via an explicit `operations: ['set_agent_config']` grant** (Gemini "blocker"): by-design SU-bypass parity — the same is true of `restart` / `set_configuration` today, and it's deliberately excluded from the `agent` group so `operations: ['agent']` never confers it. Granting a dangerous SU op explicitly is an operator's deliberate choice.
- **Omitted-`requiresSuperUser` "fail-open/unreachable"** (Gemini): false positive — that path is unchanged from today's behavior (super_user-reachable; non-SU fails closed with `400 OP_NOT_FOUND`).
- Minification breaking `.name`, hot-path deopt, `.ts` import extension: N/A for Harper (no minification; the *existing* authz already keys on `.name`, and the new code sets the name explicitly; `.ts` extensions are the TypeStrip convention; registration is startup-time, not hot).

## Tests

`unitTests/server/serverHelpers/serverUtilities.test.js` (seam: name registration without mutating the caller, shared-handler independence, non-enum op becomes grantable, opt-in gating, SU allowed / non-SU denied / non-SU+grant allowed), plus `operationPermissions` (`agent` group) and existing `operation_authorization` SU-bypass coverage. 201 passing across the affected suites + agent suite; build/prettier/oxlint clean.

## Notes
- Targets 5.2 (with the rest of the HarperFast/harper-pro#676 stack).
- Residual (out of scope): a direct in-code call to a handler bypasses `verifyPerms` — acceptable, the handlers are ops-API-coupled.


dawsontoth (approved):

server/serverHelpers/serverUtilities.ts
```diff
@@ -131,14 +131,31 @@ export type OperationDefinition = {
 	httpMethod?: 'DELETE' | 'GET' | 'HEAD' | 'OPTIONS' | 'PATCH' | 'POST' | 'PUT' | 'TRACE'; // method to use for REST
 	isJob?: boolean;
 	parametersSchema?: any[];
+	// When set, the operation declares its authorization requirement to the central verifyPerms
+	// system so it participates in the role `operations` allowlist (grantable to a scoped role)
+	// instead of a hand-rolled inline check. `true` = super_user by default. Omit to keep the
+	// pre-existing behavior (no central entry; the caller enforces its own auth, if any).
```
cb1kenobi: **Nit: document the `requiresSuperUser: false` case — it opens the op to any authenticated user**

The comment covers `true` and omit, but not `false`. Passing `false` still takes the `requiresSuperUser !== undefined` branch and registers a central entry with `perms: []`, so `verifyPerms` hits the `perms.length === 0 → return null` path and allows the op for *any* authenticated user with a valid role, not just super_user. In an auth seam that tri-state is a footgun; worth spelling out that `false` = any authenticated user.

—
_Generated by Barber AI_

unitTests/server/serverHelpers/serverUtilities.test.js
```diff
@@ -564,4 +564,96 @@ describe('Test serverUtilities.js module ', () => {
...
+			hdb_user: { active: true, role: { permission: { super_user: true }, role: 'admin' }, username: 'su' },
+		});
+
+		before(function () {
+			server.registerOperation({ name: SU_OP, execute: async () => ({ ok: true }), requiresSuperUser: true });
+		});
+
+		after(function () {
+			// Keep the global op map clean — these test-only ops shouldn't leak into other suites.
+			for (const op of [SU_OP, OPEN_OP, 'test_name_pinning_op', 'shared_op_a', 'shared_op_b', 'dyn_grantable_op']) {
```
cb1kenobi, line 597: **Low: teardown leaves entries in the global authz registries**

`after()` clears `OPERATION_FUNCTION_MAP`, but each `requiresSuperUser` registration also mutates two other process-global sets that are never cleaned: `requiredPermissions` (operation_authorization) and `dynamicallyRegisteredOps` (operationPermissions, via `registerGrantableOperation`). So `dyn_grantable_op`, `test_registered_su_op`, etc. stay registered for the rest of the mocha process. No cross-suite collision today (names are unique), but it undercuts the exact cross-file-leak hygiene this same PR adds in `operation_authorization.test.js`. Consider exposing a reset/delete helper on those modules and clearing them here too.

—
_Generated by Barber AI_

kriszyp:
Addressed the review feedback in `5822f3953`:

- **Rest-param forwarding** (Gemini): the renaming wrapper now forwards all args (`(...args) => execute(...args)`) so it's a transparent pass-through and never narrows the handler's call contract. (In practice ops are only ever called with the single `operation` arg, but the wrapper shouldn't be the thing that assumes that.)
- **`requiresSuperUser: false` documented** (@cb1kenobi): the doc comment now spells out the tri-state — `false` registers an empty-perms entry, which `verifyPerms` treats as "no specific permission required" and opens the op to any authenticated user with a valid role. Flagged explicitly as *not* a "still locked down" default.
- **Teardown registry hygiene** (@cb1kenobi): added symmetric `unregisterOperationPermission` / `unregisterGrantableOperation` helpers (mirroring the register side) and the test `after()` now clears all three process-global registries — `OPERATION_FUNCTION_MAP`, `requiredPermissions`, and the grantable-ops set — instead of just the op map.

Affected unit suites green (serverUtilities 62, operation_authorization + operationPermissions 106); build + lint clean.

PR #1601, opened 2026-07-04
fix(serialization): surface @computed scalars on default reads + guard cyclic @enumerable serialization (#1484)

### LINKED ISSUE #1484: REST GET silently omits @computed fields from responses (computed getters are non-enumerable)
REST GET responses (collection scan **and** single-record point-read) silently **omit `@computed` fields**, while ops `search_by_value`(get_attributes) and SQL `SELECT *`/`SELECT computedField`/aggregates all return them with correct values.

**Root:** `@computed` getters are installed as **non-enumerable** properties on `structPrototype` — `resources/graphql.ts:229` only sets `enumerable: true` when the separate `@enumerable` directive is applied, so a plain `@computed` field defaults to non-enumerable. REST serialization uses `JSON.stringify`, which walks only own-enumerable properties → the computed getter is invisible. The ops path iterates the `attributes[]` array explicitly, so it always includes computed fields.

**Workaround (predicted):** annotate the computed field with `@enumerable`.

**Fix:** make `@computed` fields enumerable by default (or have the REST serializer include computed getters explicitly) so REST GET matches ops/SQL.

**Other `@computed` semantics are correct:** computed-on-read (no stale value), index materialized at write with no staleness under source-field churn, direct writes to a computed field rejected 400, both engines identical.

**Severity:** medium — silent omission on the most common read surface (REST clients never see the computed value).

**Repro:** `npm run test:integration -- "integrationTests/qa-scratch/qa363-computed.test.ts"`

— filed by KrAIs (Claude) on Kris's behalf (qa-explorer campaign, main @ 28db4fde4 v5.1.12)

kriszyp:
Closes #1484.

Default reads now surface non-table `@computed` attributes, and cyclic `@enumerable` relationships serialize repeated ancestors as primary-key reference stubs instead of overflowing. The bounded per-table serializer keeps acyclic enumerable tables on a guard-free fast path and uses a path-scoped `(table, primary key)` guard only when runtime cycles are possible.

The guarded path preserves transport-native values: it identifies both promoted classic msgpackr records and structon random-access records through a shared `RecordObject` runtime base, recursively materializes only Harper records and ordinary containers, and leaves `Date`, `Buffer`/typed arrays, and `ArrayBuffer` intact for CBOR and MessagePack. Per-encoder record subclasses keep table-specific computed getters isolated.

Reviewer attention: `resources/Table.ts` owns the cycle guard and guarded materialization; `resources/RecordEncoder.ts` owns the shared runtime base. A computed resolver's declared scalar type cannot constrain its JS return value, so every surfaced non-table computed takes the guarded path.

Coverage includes mutual/self/runtime-computed cycles, classic msgpackr and structon records, CBOR and MessagePack fidelity, and RocksDB + LMDB. No documentation change is needed: this restores declared computed-field behavior and preserves the existing binary transport contract.

Generated with Claude Opus on the original implementation and Codex on the native-serialization follow-up. Final follow-up review: Gemini 3.6 Flash via direct API found no unresolved issues; the local Claude CLI review route was unavailable (execution timeout).

### DIFF
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -62,14 +62,7 @@ import { getWorkerIndex, getWorkerCount } from '../server/threads/manageThreads.
 import { HAS_BLOBS, auditRetention, removeAuditEntry } from './auditStore.ts';
 import { buildEmbedBefore, createDefaultEmbedder, type EmbedAttribute, type Embedder } from './models/embedHook.ts';
 import { autoCast, autoCastBooleanStrict } from '../utility/common_utils.ts';
-import {
-	recordUpdater,
-	removeEntry,
-	PENDING_LOCAL_TIME,
-	type RecordObject,
-	type Entry,
-	entryMap,
-} from './RecordEncoder.ts';
+import { recordUpdater, removeEntry, PENDING_LOCAL_TIME, RecordObject, type Entry, entryMap } from './RecordEncoder.ts';
 import { recordAction, recordActionBinary } from './analytics/write.ts';
 import { rebuildUpdateBefore } from './crdt.ts';
 import { appendHeader } from '../server/serverHelpers/Headers.ts';
@@ -272,6 +265,68 @@ function cloneConditions(conditions: any[]): any[] {
 		return copy;
 	});
 }
+// Ambient, path-scoped cycle guard for the enumerable-struct `toJSON` serialization path. A record on
+// a cyclically-enumerable table can (transitively) reference itself, which would recurse forever through
+// JSON.stringify. The struct getters resolve related records by id — and without a shared cache each
+// resolution decodes a fresh instance — so object-identity tracking is defeated; we key on (tableId, id).
+// Only cyclic tables allocate/consult this map; acyclic-enumerable tables keep a guard-free fast path.
+// Keyed by table class → set of primary keys currently on the serialization path (avoids any tableId
+// stringification/collision concern).
+let structSerializationVisited: Map<any, Set<any>> | null = null;
+// Resolve Harper records within a guarded serialization, so none escape back to an encoder that would call
+// their toJSON after our path-scoped unwind and miss the cycle. Native values (Date, Buffer/typed arrays,
+// ArrayBuffer, etc.) stay intact for CBOR/MessagePack.
+function resolveStructForJSON(value: any): any {
+	if (value == null || typeof value !== 'object') return value;
+	const isRecordObject = value instanceof RecordObject;
+	if (isRecordObject) {
+		const toJSON = (value as any).toJSON;
+		// The table-installed toJSON resolves every surfaced value through this function before returning.
+		if (typeof toJSON === 'function') return toJSON.call(value);
+	}
+	if (Array.isArray(value)) {
+		let resolvedArray: any[] | undefined;
+		for (let index = 0; index < value.length; index++) {
+			if (!(index in value)) continue;
+			const original = value[index];
+			const resolved = resolveStructForJSON(original);
+			if (resolved !== original) {
+				resolvedArray ??= value.slice();
+				resolvedArray[index] = resolved;
+			}
+		}
+		return resolvedArray ?? value;
+	}
+	const prototype = Object.getPrototypeOf(value);
+	if (!isRecordObject && prototype !== Object.prototype && prototype !== null) return value;
+	// Materialize eligible containers in one pass so accessors run exactly once. A RecordObject without a
+	// table toJSON becomes plain response data rather than a detached record-shaped object with no entryMap metadata.
+	const resolvedObject = Object.create(isRecordObject ? Object.prototype : prototype);
+	for (const key of Object.keys(value)) {
+		const original = value[key];
+		resolvedObject[key] = resolveStructForJSON(original);
+	}
+	return resolvedObject;
+}
+// Is `start` reachable from itself through @enumerable table-typed edges? (Includes self-loops, e.g. a
+// tree table with an enumerable parent/children relationship.) Walks the per-table `enumerableRelationDefs`
+// graph, resolving each definition's `.tableClass` here — this runs lazily on first serialization, by which
+// point every table class is assigned (so self/forward refs whose tableClass was unset at collection time
+// resolve correctly).
+function detectCyclicEnumerable(start: any): boolean {
+	const queue = start.enumerableRelationDefs ? [...start.enumerableRelationDefs] : [];
+	const seen = new Set();
+	while (queue.length) {
+		const target = queue.pop()?.tableClass;
+		if (!target) continue;
+		if (target === start) return true;
+		if (seen.has(target)) continue;
+		seen.add(target);
+		if (target.enumerableRelationDefs) for (const def of target.enumerableRelationDefs) queue.push(def);
+	}
+	return false;
+}
+
 // #section: setup-and-factory
 export function makeTable(options) {
 	const {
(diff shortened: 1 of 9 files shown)


resources/Table.ts
```diff
@@ -283,6 +336,72 @@ export function makeTable(options) {
...
+				let added = false;
+				try {
+					if (isTop) structSerializationVisited = new Map();
+					const id = this[primaryKey];
+					// Composite/array primary keys decode to a fresh array each read, so identity-based membership
+					// would miss the same logical record; normalize to a stable string key. The replacer keeps a
+					// BigInt component from throwing (JSON.stringify can't serialize BigInt natively).
+					idKey = Array.isArray(id)
+						? JSON.stringify(id, (_, v) => (typeof v === 'bigint' ? v.toString() : v))
+						: id;
```
cb1kenobi: **Low: object-typed single primary key defeats the cycle guard**

The `idKey` normalization only stringifies array PKs; a single-column object-typed key (a `Bytes`/`Uint8Array` or `Date` PK) falls to the `: id` branch and is tracked by object identity. As the guard's own comment notes, related records decode to a *fresh* instance each resolution, so `ids.has(idKey)` never matches for such a key — the cycle is never collapsed to a stub and serialization recurses until the stack overflows, the exact failure this guard exists to prevent. Narrow trigger (cyclic-`@enumerable` table + object-typed single PK), hence Low. Normalizing any non-primitive id also subsumes the array case.

Suggested fix:
```suggestion
					idKey =
						id !== null && typeof id === 'object'
							? JSON.stringify(id, (_, v) => (typeof v === 'bigint' ? v.toString() : v))
							: id;
```

—
_Generated by Barber AI_

resources/Table.ts
```diff
@@ -4240,21 +4365,25 @@ export function makeTable(options) {
...
+					// The type definition (set early) marks a table-typed attribute; its `.tableClass` may not be
+					// assigned yet for self/forward refs, so key table-typedness off the definition, not tableClass.
+					const relationDef = attribute.definition || attribute.elements?.definition;
+					// Surface on the default read when @enumerable, or when a @computed attribute is NOT
+					// table-typed (harper#1484 — computed scalars). Table-typed computeds stay lazy.
+					if (attribute.enumerable || (attribute.computed && !relationDef)) enumerableAttributeNames.push(name);
+					// Only @enumerable *table-typed* attributes create a serialization edge that can cycle.
+					if (attribute.enumerable && relationDef) enumerableRelationDefs.add(relationDef);
+					// A surfaced computed with an unconstrained type could return a live entity at runtime.
+					if (attribute.computed && !relationDef && (attribute.type == null || attribute.type === 'Any'))
```
cb1kenobi: **Low: `hasUnconstrainedComputed` only catches `Any`/untyped computeds**

This flag routes a table to the guarded path when a surfaced `@computed` could return a live entity the static edge graph can't see — but it only fires for `type == null || type === 'Any'`. A `@computed` declared with a concrete *non-table* scalar type (so `relationDef` is falsy) whose resolver actually returns a live struct still takes the raw fast path (~line 366) with no guard Map, and can overflow if that entity cycles. The resolver's real return type isn't knowable from the declared type, so consider routing any surfaced non-table `@computed` through the guarded path rather than only `Any`/untyped ones.

—
_Generated by Barber AI_

resources/Table.ts
```diff
@@ -4240,21 +4365,25 @@ export function makeTable(options) {
...
-								return json;
-							},
-						});
-					}
+					// The type definition (set early) marks a table-typed attribute; its `.tableClass` may not be
+					// assigned yet for self/forward refs, so key table-typedness off the definition, not tableClass.
+					const relationDef = attribute.definition || attribute.elements?.definition;
+					// Surface on the default read when @enumerable, or when a @computed attribute is NOT
+					// table-typed (harper#1484 — computed scalars). Table-typed computeds stay lazy.
+					if (attribute.enumerable || (attribute.computed && !relationDef)) enumerableAttributeNames.push(name);
```
cb1kenobi, line 4803: **Low: computed scalars now surface to non-REST consumers too — confirm intended**

Installing this `toJSON` for a table whose only surfaced attribute is a `@computed` scalar (such a table installed no `toJSON` on `main`) means the computed value now appears in *every* struct serialization — MQTT/WebSocket/SSE subscription payloads and the change feed — not just the REST GET this PR targets. The PR body and tests cover only REST. Confirm surfacing computed scalars to subscription consumers is intended; if not, gate it to the REST read path. (Replication/audit are unaffected — they persist via the storage encoder path.)

—
_Generated by Barber AI_

resources/Table.ts
kriszyp: Good catch — fixed in 367d8507e. `idKey` now normalizes any non-primitive PK (`typeof id === 'object'`), covering object/`Date`/`Bytes` single-column keys and subsuming the array case, so an object-PK cycle collapses to a `{ [primaryKey]: id }` stub instead of overflowing.

— KrAIs (Claude Opus 4.8)

resources/Table.ts
kriszyp: Agreed — the declared scalar type can't be trusted to bound what the resolver actually returns. Broadened in 367d8507e: any surfaced non-table `@computed` now routes through the guarded path (`hasUnconstrainedComputed` → `hasSurfacedComputed`), not just `Any`/untyped ones. These tables already install `toJSON` and build a plain object, so the guarded-vs-fast delta is small relative to closing a stack-overflow. Added regression test **B4** (a `String`-typed `@computed` returning a live entity cycle) which overflowed on the fast path before this change and now serializes as a reference stub.

— KrAIs (Claude Opus 4.8)

resources/Table.ts
kriszyp: Intended. `@enumerable` attributes already surface to MQTT/WS/SSE via this same struct `toJSON` on `main`, so computed scalars joining them is consistent rather than new-in-kind; and `toJSON` has no transport context, so REST-only gating isn't feasible without a projection layer. Replication/audit are unaffected (storage-encoder path), as you noted.

— KrAIs (Claude Opus 4.8)

cb1kenobi (approved):

Ethan-Arrowood:
Code looks right, but the unit failures are this PR's: the basicREST computed-index assertion asserts the pre-fix behavior and needs updating, cache-test's 400 needs a look, and the new test imports `node:assert/strict` which lint now rejects (#1753). Happy to approve once CI is green.

kriszyp:
CI was red on 3 things after the rebase — pushed fixes for the two that were real, left the third alone since it's yours to weigh in on:

1. **Its own `node:assert/strict` lint violation** in the new `computed-and-cyclic-serialization.test.ts` — fixed (house rule is `node:assert`).
2. **Two pre-existing tests broke against Fix A's own stated intent**, not a code bug:
   - `basicREST-test.mjs` asserted `ageInMonths` (a non-enumerable `@computed @indexed` scalar) stays `undefined` on a default read — that's exactly the pre-fix behavior this PR's Fix A deliberately changes ("computed scalars... their getters are non-enumerable, so JSON.stringify used to silently drop them... must include"). Updated the assertion to `288` to match.
   - `cache-test.mjs`'s "change source and get" test round-trips a GET response straight into a PUT, stripping only `nameTitle` before sending it back. Now that `ageInMonths` also surfaces by default, it rode along and got rejected as a write to a computed attribute (400). Added `delete data.ageInMonths` alongside the existing strip.

Didn't touch anything else — just wanted CI green so you can review the actual diff without noise. Let me know if you want the `ageInMonths` default-surfacing behavior itself reconsidered instead of the test updated.

<sub>— KrAIs (Claude Sonnet 5)</sub>

PR #1602, opened 2026-07-04
fix(mcp): surface parameterized custom resources as MCP tools

kriszyp:
## Summary

Custom resources on a parameterised path (`static path = '/widget/:id'`) are stored by `Resources.setParamRoute` in a `paramRoutes` side-array and never inserted into the base resources `Map`. MCP's `buildApplicationTools` iterated only the Map, so these resources produced **zero MCP tools** — even though OpenAPI (which already iterates `paramRoutes`) exposed them. This enumerates `paramRoutes` for MCP tool building so the tool surface matches the REST surface.

## What to look at

The enumeration is deliberately restricted to what the **generated verb handlers can actually bind**, so we never advertise a tool that would dispatch to the wrong target:

- **Single-`:id` routes** → `get`/`update`/`patch`/`delete` tools. These handlers bind `target.id = args.id` correctly. `create`/`search` are suppressed: `makeCreateHandler` forces `target.isCollection = true` and never binds `id`, and `makeSearchHandler` is collection-scoped — neither matches "the record named by `:id`".
- **Multi-segment / named-wildcard routes** (`/files/*rest`) → no generated verb tools yet (nothing binds the extra segments), but author-defined `mcpTools`/`mcpPrompts` still register, since they carry their own schemas and handler methods.
- `liveResource()` (live-registry dispatch, added in #1487) now also consults `paramRoutes`, so param-route tools don't silently fall back to the registration-time class.

Regression coverage in `unitTests/components/mcp/tools/application-paramroutes.test.js` (get present; create/search absent on `:id`; custom tools present + generated verbs absent on wildcard). Full MCP tools suite green (94 passing).

## Provenance

The core enumeration fix was authored by @dawsontoth as part of the discussion on #1503 (RFC 0001) — extracted here so it can land independently of the RFC. The binding-mode restrictions (single-`:id` scoping, `create`/`search` suppression, custom-tool preservation on wildcards) were added during extraction in response to a Codex cross-model review that converged over four rounds.

Cross-model review: Codex — clean (no actionable correctness issues) after the changes above. Gemini/`agy` timed out (twice) and did not produce a review.

_Prepared by an AI agent (Claude, Opus 4.8). Not the original commit author — see per-commit `Co-Authored-By` trailers._


dawsontoth (approved):

dawsontoth:
try rebasing

Issue #1417, opened 2026-06-19
Multi-worker HTTP rolling restart produces ~0.6–1.2s whole-pool connection-refused gap


kriszyp:
## Summary

`restartWorkers('http')` (called during rolling HTTP restart) computes `maxWorkersDown=1` to maintain connectivity, but in practice the entire worker pool goes connection-refused for approximately **0.6–1.2 s** before any new worker is ready to accept.

## Observed behavior

With `threads: { count: 4 }` and a continuous request loop at ~50 req/s:

1. `restartWorkers('http')` is called (e.g. on config hot-reload or deploy).
2. Within ~20ms, ALL four workers stop accepting connections simultaneously — `ECONNREFUSED` on every request.
3. After ~0.6–1.2s, the first new worker finishes init and begins accepting — connectivity resumes.
4. The remaining workers stagger online over the next ~1s.
5. Net gap: **0.6–1.2s of total connection-refused** visible to any external client.

Expected: at most `maxWorkersDown=1` workers down at any instant; the other 3 remain available throughout.

## Root cause (hypothesis)

`restartWorkers` likely signals all workers to restart in a single round before waiting for replacements to become ready. The "rolling" invariant (`maxWorkersDown`) appears to govern WHICH workers restart, not WHEN they restart — if all workers receive the restart signal before any has fully started its replacement, the pool is momentarily empty.

## Impact

- Any live HTTP client (REST, GraphQL, WebSocket handshake) sees ~1s of `ECONNREFUSED` during any rolling restart.
- Deploy/config-reload triggers this — both are routine operational events.
- Severity scales with worker count (more workers → more overlap in the restart window).

## Repro

```
npm run test:integration -- "integrationTests/qa-scratch/qa156-rolling-restart.test.ts"
```
(4 workers, continuous HTTP load, `restartWorkers('http')` mid-test; asserts no gap > 100ms — fails.)

**Harper:** `7aaa5a152`; byte-identical to `main` @`6797f091d`.

---
*Surfaced by the QA-explorer exploratory campaign for @kris (QA-156). Filed by Claude (Sonnet 4.6) for @kris.*

kriszyp:
Addressed by #1432 (overlap HTTP worker restarts to eliminate the connection-refused gap), merged 2026-06-22.

— Claude (Opus 4.8), issue-backlog triage on Kris's behalf

PR #1605, opened 2026-07-05
feat(server): surface external port conflicts on all platforms; single-worker default without SO_REUSEPORT

kriszyp:
## Summary

Stacked on #1603 (MQTT listeners share their port via `SO_REUSEPORT` on non-macOS). This makes external port conflicts **detected and loudly logged on every platform**, and aligns two related behaviors on platforms without working `SO_REUSEPORT` (macOS/Windows).

`listenOnPorts()` used to swallow every `EADDRINUSE` — a workaround for a Node <20.11.1 reusePort bug now outside Harper's supported range (`^22.18 || >=24`). So an unrelated process squatting a Harper port (the original symptom: a second Harper instance holding 8883) silently received Harper's MQTT traffic with no error logged anywhere.

## Design

Every `EADDRINUSE` with a possible in-process explanation is structurally ruled out, so what remains is unambiguously external and gets logged (port + owning component + error):

| Case | Why EADDRINUSE must be external |
|---|---|
| `reusePort` listener (Linux) | siblings share the port, never collide — even across overlapping restarts |
| main thread (HTTP/ops ports) | binds every port (awaited) before any worker starts; never restarts |
| dedicated listener's owner worker (MQTT via `server.socket()`) | never bound by the main thread; now bound by a **single owner worker** (lowest eligible index) instead of every worker racing; restarts are non-overlapping without reusePort |

The one remaining benign case — a worker's exclusive HTTP bind losing to the main thread on macOS/Windows — stays silently swallowed. Everything still resolves, so a squatted port never stalls boot.

Supporting changes:
- **`restartWorkers()` no longer pre-starts replacements on macOS** (`canPreStartReplacement` now excludes darwin, like Windows/Bun). This also fixes a latent bug: the pre-started replacement could never bind the exclusive ports the old worker still held, its `EADDRINUSE` was swallowed, and worker-owned listeners (MQTT) were left **permanently unbound after every component-reload restart on macOS**. The main thread keeps serving HTTP throughout, so only worker-owned listeners see the brief shutdown-first gap.
- **`threads.count` defaults to 1 on macOS/Windows** (`setDefaultThreads`): without `SO_REUSEPORT`, extra HTTP workers can never share the server ports — the CPU-based default just spawned workers that serve no direct TCP traffic. Explicit `threads.count` still overrides.

## Where to look

- **The ownership claims are the crux.** Verified in code + empirically: main thread binds HTTP/ops before workers (`socketRouter.ts` awaits `listenOnPorts()` pre-spawn); component plugins don't run `handleApplication` on the main thread, so MQTT listeners exist only in workers.
- **`canPreStartReplacement` on darwin**: #1417 (rolling-restart conn-refused) was a Linux `SO_REUSEPORT`-pool fix and Linux is unchanged. `rolling-restart.test.ts` fails identically on macOS on the base branch (pre-existing), so CI/Linux remains the gating signal.
- **`threadRange`**: the dedicated-listener owner is the lowest *eligible* index, not hardcoded 0.

## Test

`integrationTests/server/external-port-conflict.test.ts` squats the MQTT secure port before boot and asserts the conflict is logged **on every platform** (skip removed) and Harper still starts. Plus platform-stubbed unit coverage for `setDefaultThreads` and the `dedicatedListener` marker.

---
🤖 Generated by Claude (Fable 5 / Opus 4.8 across revisions). Cross-model reviewed (Codex + Gemini) at each design iteration; earlier reviews drove the pivot from worker-gating → reusePort-only → this universal design.

### DIFF
--- a/integrationTests/server/external-port-conflict.test.ts
+++ b/integrationTests/server/external-port-conflict.test.ts
@@ -15,7 +15,7 @@
  * still starts (the conflict is surfaced, not fatal).
  */
 import { suite, test, before, after } from 'node:test';
-import { ok } from 'node:assert/strict';
+import { ok } from 'node:assert';
 import { createServer, type Server } from 'node:net';
 import { mkdtemp, readFile } from 'node:fs/promises';
 import { join } from 'node:path';
--- a/integrationTests/server/external-port-conflict.test.ts
+++ b/integrationTests/server/external-port-conflict.test.ts
@@ -1,0 +1,111 @@
+/**
+ * Regression test for external MQTT port-conflict detection.
+ *
+ * With reusePort (Linux), sibling workers share a port and never raise EADDRINUSE; without it
+ * (macOS/Windows), the main thread binds every port before any worker starts and never
+ * restarts. Either way an EADDRINUSE on the reusePort path or the main thread is never a
+ * benign in-process collision — it means an unrelated process already holds the port and will
+ * silently receive this listener's traffic. Harper used to swallow that EADDRINUSE
+ * unconditionally, so a squatter on the MQTT port went completely unreported (the original
+ * symptom: a developer's second Harper instance holding 8883, MQTT connections silently
+ * routed to it).
+ *
+ * This test pre-occupies the MQTT secure port with an unrelated plain socket server, boots
+ * Harper on the same address, and asserts (a) Harper logs the external conflict and (b) Harper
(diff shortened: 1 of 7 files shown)


Ethan-Arrowood (approved):
LGTM — the ownership analysis is convincing. Branch is conflicting with main but the hunks look mechanical (drain-hook context drift); approving on that assumption — ping me if the rebase turns out to touch anything real.

PR #1620, opened 2026-07-06
ci(docker): add smoke test that builds and boots the image

kriszyp:
## Why

The Docker publish workflow only runs on `release: published` and pushes the built image straight to Docker Hub — it never **starts** the container. So runtime-only breakage in the image sails through CI unnoticed.

That's exactly how #1619 (entrypoint written as a 0-byte file → `exec format error` on every `5.1.x` container) reached published images.

## What

Add a `pull_request` / `push`-to-`main` smoke test, path-filtered to the `Dockerfile` (and `.dockerignore` / the workflow itself) so it only runs when the image build actually changes. It builds the image (`linux/amd64`, loaded locally) and then:

1. **Entrypoint sanity** — asserts `/usr/local/bin/docker-entrypoint.sh` is non-empty, executable, and parses as valid shell (`sh -n`). This alone would have caught #1619.
2. **Boot check** — starts the container and waits (up to 60s) for the Operations API on `:9925` to answer with any HTTP status; fails fast if the container exits early, and dumps logs on failure.

## Verification

Ran both steps locally against the image built from the #1619 fix branch:

- Entrypoint check: `entrypoint OK (117 bytes, valid shell)`
- Boot check: `Operations API responded HTTP 200 after 4s`

For contrast, on the current (broken) `main` Dockerfile the entrypoint check fails with a 0-byte file — i.e. the test correctly red/greens on the bug.

## Note

Stacked on #1619 (this branch includes that fix so its own CI is green). Once #1619 merges, rebasing reduces this PR to just the workflow file. Merge #1619 first.

### DIFF
--- a/.github/workflows/docker-smoke.yml
+++ b/.github/workflows/docker-smoke.yml
@@ -22,20 +22,31 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+# Building and booting the image only needs read access to the checkout.
+permissions:
+  contents: read
+
 jobs:
   smoke:
-    runs-on: ubuntu-latest
+    # arm64 is exercised only on push to main (the publish workflow ships both
+    # arches); PRs stay amd64-only to keep the extra build off every PR run.
+    strategy:
+      fail-fast: false
+      matrix:
+        include: ${{ github.event_name == 'push' && fromJSON('[{"platform":"linux/amd64","runner":"ubuntu-latest"},{"platform":"linux/arm64","runner":"ubuntu-24.04-arm"}]') || fromJSON('[{"platform":"linux/amd64","runner":"ubuntu-latest"}]') }}
+    runs-on: ${{ matrix.runner }}
+    timeout-minutes: 20
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
       - name: Setup Docker Buildx
         uses: docker/setup-buildx-action@d7f5e7f509e45cec5c76c4d5afdd7de93d0b3df5 # v4.1.0
 
-      - name: Build image (linux/amd64, load locally)
+      - name: Build image (${{ matrix.platform }}, load locally)
         uses: docker/build-push-action@f9f3042f7e2789586610d6e8b85c8f03e5195baf # v7.2.0
         with:
           context: .
-          platforms: linux/amd64
+          platforms: ${{ matrix.platform }}
           load: true
           tags: harperfast/harper:smoke-test
 
--- a/.github/workflows/docker-smoke.yml
+++ b/.github/workflows/docker-smoke.yml
@@ -1,0 +1,87 @@
+name: Docker Image Smoke Test
+
+# Builds the container image and boots it to catch breakage that only shows up
+# at runtime (e.g. an empty/unexecutable entrypoint producing "exec format
+# error"). The publish workflow only runs on release and never starts the
+# container, so this guards the gap for changes to the image build.
+
+on:
+  pull_request:
+    paths:
+      - Dockerfile
+      - .dockerignore
+      - .github/workflows/docker-smoke.yml
+  push:
+    branches: [main]
+    paths:
+      - Dockerfile
+      - .dockerignore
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-smoke.yml
```diff
@@ -0,0 +1,87 @@
...
+      - Dockerfile
+      - .dockerignore
+      - .github/workflows/docker-smoke.yml
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  smoke:
```
cb1kenobi, line 30: **Low (hardening): no explicit `permissions` block — the job runs with the repo's default token scope**

With no `permissions:` key, this workflow inherits the repository's default `GITHUB_TOKEN` permissions, which may be read/write. Building and booting the image only needs read access. Set least privilege explicitly (OpenSSF Scorecard flags exactly this).

Suggested fix:
```suggestion
  smoke:
    permissions:
      contents: read
```

—
_Generated by Barber AI_

.github/workflows/docker-smoke.yml
```diff
@@ -0,0 +1,87 @@
+name: Docker Image Smoke Test
+
+# Builds the container image and boots it to catch breakage that only shows up
+# at runtime (e.g. an empty/unexecutable entrypoint producing "exec format
+# error"). The publish workflow only runs on release and never starts the
+# container, so this guards the gap for changes to the image build.
+
+on:
+  pull_request:
+    paths:
```
cb1kenobi, line 10: **Low: a `paths`-filtered workflow cannot safely be a required status check**

Because of the `paths:` filter this workflow only runs when `Dockerfile` / `.dockerignore` / the workflow itself change. If it is ever marked as a required check in branch protection, PRs that don't touch those paths will leave the check permanently pending and block merges (a well-known GitHub behavior). If you intend it as a merge gate, either drop the path filter or add a companion always-running job that reports success as a stand-in. If it isn't meant to be required, no change needed — just be aware.

—
_Generated by Barber AI_

.github/workflows/docker-smoke.yml
kriszyp: Good call — added an explicit \`permissions: contents: read\` in 05f86d0. I put it at the workflow level (the job only reads the checkout to build/boot the image), so both the amd64 and the new arm64 leg inherit least privilege. Thanks! (AI-assisted response.)

.github/workflows/docker-smoke.yml
kriszyp: Thanks — noted. This is not intended as a required status check, precisely because of the path filter (the smoke test is a guard for image-build changes, not a universal merge gate), so no change here. Appreciate flagging it for the record. (AI-assisted response.)

.github/workflows/docker-smoke.yml
```diff
@@ -0,0 +1,87 @@
...
+      - .dockerignore
+      - .github/workflows/docker-smoke.yml
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  smoke:
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
kriszyp: Added \`timeout-minutes: 20\` on the smoke job in 05f86d0. (AI-assisted response.)

.github/workflows/docker-smoke.yml
```diff
@@ -0,0 +1,87 @@
...
+# error"). The publish workflow only runs on release and never starts the
+# container, so this guards the gap for changes to the image build.
+
+on:
+  pull_request:
+    paths:
+      - Dockerfile
+      - .dockerignore
+      - .github/workflows/docker-smoke.yml
+  push:
```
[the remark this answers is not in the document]
kriszyp: Added an arm64 leg in 05f86d0 via a dynamic matrix, restricted to \`push\` to main (PRs stay amd64-only), mirroring the publish workflow's \`ubuntu-24.04-arm\` runner. (AI-assisted response.)

Ethan-Arrowood (approved):
Nice guard — this would have caught #1619 before it shipped.

cb1kenobi (approved):

PR #1626, opened 2026-07-06
Make target.loadedFromSource the sole cache-disposition signal (net of v5.1 #1575 + #1590)

kriszyp:
Brings the **net end-state** of the `loadedFromSource` cache-disposition work to `main`/5.2. That work landed only on `v5.1` as two PRs — #1575 (initial fix) and #1590 (rework) — and was never on `main`. This is the single clean end-state, not a replay of #1575-then-#1590.

## Why

Cache disposition (`loadedFromSource`: served from source vs. cache) is fundamentally a **per-`get()` result** — each `get()` has its own `RequestTarget`. #1575 additionally mirrored disposition onto the request **Context**, which is shared for the whole request lifecycle, making it **last-write-wins**: an internal caching `get()` inside a resource's own `get()` overwrites the entry get's disposition before the caller reads it. #1590 dropped that mirror and standardized on `target.loadedFromSource`, which `server/REST.ts` already reads.

## What changed

- **`target.loadedFromSource` is the sole signal.** New `setLoadedFromSource(target, bool)` helper in `Table.ts` writes only the `RequestTarget`, with a **primitive-id guard** (`typeof target === 'object'`) — instance-API calls can pass a primitive id that can't hold the flag. No Context writes anywhere.
- **`Age` header now emitted on cache hits.** The `loadAsInstance=false` value path and the `ensureLoaded` branch (`Table.ts` ~1229/~1242) previously set the flag only on source fetch, leaving it `undefined` on hits — so `REST.ts` (`if (loadedFromSource !== undefined)`) skipped the `Age` header on hits. They now mark `false` on hits.
- **Removed dead code:** the `wasLoadedFromSource()` method (`Resource.ts` / `ResourceInterface.ts`) and the `Context.loadedFromSource` declaration (`ResourceInterface.ts`), both now unused.

## Where to look

- `resources/Table.ts` is the substance — 7 small hunks routing all disposition writes through the guarded helper plus the two new cache-hit markings. The get() path is hot; the helper is a trivial branch, no added async.
- **Downstream note (open, tracked separately):** `security/certificateVerification/{ocsp,crl}Verification.ts` still call `cacheEntry.wasLoadedFromSource?.()`. Removing the method is behavior-preserving there — it always returned `undefined`/void, and the optional-chaining call now yields `undefined`, i.e. the same "cache hit" branch as before. To actually observe disposition there, a `RequestTarget` must be passed to the get; that's out of scope here (matches #1590's downstream note). This is the one spot worth a reviewer's eye.

## Testing

Ported the v5.1 caching unit coverage to `unitTests/resources/caching.test.js`: a `loadAsInstance=false` hit/miss case and a stale-while-revalidate `loadedFromSource=false` assertion. **15/15 caching tests pass** against `dist`.

_Generated by KrAIs (Claude Opus 4.8), at Kris's direction._


kriszyp:
CI note: the red **Unit Test** job (v22/v24/v26) is the pre-existing `subscriptionReplay.test.js:1048` flake introduced by #1345 ("expected at least 1 event for id=1, got 0") — already red on `main` across all Node versions, unrelated to this change and tracked/fixed by #1625. The **resources** unit suite that actually exercises this change is fully green locally (1009 passing, incl. the caching suite 15/15). Everything else on this PR (builds, all integration shards, format, lint) is green.

_Posted by KrAIs (Claude Opus 4.8)._

resources/Table.ts
```diff
@@ -4650,6 +4651,11 @@ export function makeTable(options) {
 		}
 	}
 
+	function setLoadedFromSource(target: RequestTarget | undefined, loadedFromSource: boolean) {
+		// cache disposition is a per-get result, recorded on the RequestTarget of the get (#1576)
+		// target may be a primitive id on instance-API calls, which can't hold the flag
+		if (target && typeof target === 'object') target.loadedFromSource = loadedFromSource;
```
cb1kenobi, line 4737: **Low: guard and one new marking lack test coverage**

The `typeof target === 'object'` guard is effectively a bug fix — a primitive-id instance-API `get()` on a caching table would previously attempt `(id).loadedFromSource = false`, which throws in strict mode. The added tests cover the `loadAsInstance=false` hit/miss and the SWR case, but not:

- the primitive-id path — a regression test (`await CachingTable.get(32)` with an `onlyIfCached`/`ensureLoaded` context, no `RequestTarget`) would lock in the guard and prove the original crash is gone.
- the new `onlyIfCached` cache-hit marking at `Table.ts:1232` (the `loadAsInstance=false` branch) — currently untested; only its `loadAsInstance=true` twin at line 730 has equivalent behavior.

Both are low-risk, but the guard fix in particular deserves a regression test.

—
_Generated by Barber AI_

kriszyp: Thanks — good catch, added both. 🙏 (commit 25a1f8c70)

- **Guard regression:** `primitive-id instance-API get does not throw setting cache disposition` — drives the real primitive-`target` path via `getResource(41, {}, { ensureLoaded: true })` (how source-apply/replication loads records, `Table.getResource(id, …)`). I confirmed it fails without the guard with `Cannot create property 'loadedFromSource' on number '41'`. I used `getResource(..., { ensureLoaded: true })` rather than a `get(32)`-with-context call because the static `get()` wraps a primitive id into a fresh `RequestTarget` (transactional arg-normalization), so it never actually reaches `setLoadedFromSource` with a primitive — only the low-level instance API does.
- **onlyIfCached hit marking:** `onlyIfCached cache hit marks loadedFromSource=false with loadAsInstance=false` — covers the value-path branch (the `loadAsInstance=true` twin's counterpart).

Both green across 35+ local runs of the caching suite.

_Posted by KrAIs (Claude Opus 4.8), at Kris's direction._

Ethan-Arrowood:
seems like a rebase is needed to know if this can land cleanly. Rerequest review once rebased. Thanks!

cb1kenobi (approved):

Issue #631, opened 2026-05-20
[Models] Phase 4 — OpenAI-compatible /v1/* gateway as built-in Resources


heskew:
## Scope

Expose `POST /v1/embeddings` and `POST /v1/chat/completions` on the REST port so unmodified LangChain.js, Vercel AI SDK, OpenAI SDK, MCP-sampling clients, etc. can hit Harper for inference. The endpoints are thin protocol-translation wrappers over `scope.models`.

What ships:

- Two built-in Resources registered via `scope.resources.set(path, ResourceClass)`, dispatched through the REST port (`HTTP_PORT`, 9926) like any other Resource.
- Request-body translation (OpenAI shape ↔ internal `EmbedOpts` / `GenerateOpts` shape).
- SSE streaming on `/v1/chat/completions` for `stream: true`, using `openaiStream()` from **#514** to format the OpenAI chunk-delta envelope (including the terminal `[DONE]` sentinel).
- OpenAI-shape error responses (`{ error: { message, type, code, param } }`) for genuine drop-in compatibility.
- `GET /v1/models` — minimal endpoint listing registered backends' advertised model names so clients that probe for available models work.
- Tenant scoping resolved from `request.user` via the same accounting path Phase 1 set up.

## Why this is in core

The issue's Acceptance section asserts "Unmodified LangChain.js / OpenAI SDK client successfully completes a chat against Harper." That requires `/v1/*` to actually exist; this is where it lives.

This is the single highest-leverage external-adoption piece in #510 — LangChain, Vercel AI SDK, OpenAI SDK, MCP sampling, and many internal tools speak the OpenAI shape natively.

## Registration pattern

Built-in Resource registration matches the existing precedent at `resources/login.ts:4`:

```typescript
// resources/models/v1/index.ts
export function handleApplication(scope: Scope) {
  scope.resources.set('v1/embeddings', V1Embeddings);
  scope.resources.set('v1/chat/completions', V1ChatCompletions);
  scope.resources.set('v1/models', V1Models);
}

class V1Embeddings extends Resource {
  static async post(_target, body, request) {
    // Translate OpenAI shape → scope.models.embed()
    // Return OpenAI-shape response
  }
}

class V1ChatCompletions extends Resource {
  static async post(_target, body, request) {
    // Translate OpenAI shape → scope.models.generate() / generateStream()
    // For stream: true, yield via openaiStream() from #514
  }
}

class V1Models extends Resource {
  static async get(_target, _request) {
[...]

PR #1632, opened 2026-07-06
Fix SSE hang + uncaughtException on finite generator streamed to completion (#1628)

### LINKED ISSUE #1628: SSE: streaming a finite async generator to completion hangs the response + throws uncaughtException (terminal-step serialize)
## Summary
Any custom Resource that streams a **plain async generator to natural completion over SSE** (`Accept: text/event-stream`) **hangs the HTTP response and throws a process `uncaughtException`** the instant the generator returns. Data events + `[DONE]` are delivered, but the response never closes. This sits directly on the LLM/`openaiStream` completion pattern #631 is building toward, and on any bounded/export-over-SSE stream.

## Root cause (code-confirmed)
`server/serverHelpers/contentTypes.ts` — `transformIterable`'s `next()` unconditionally calls `transform(step.value)` even on the generator's terminal step `{ value: undefined, done: true }`. SSE `serialize()` (contentTypes.ts:135) opens with `if (message.acknowledge) message.acknowledge()`, but `message` is `undefined` on that step → `undefined.acknowledge` throws:
```
TypeError: Cannot read properties of undefined (reading 'acknowledge')
    at serialize (server/serverHelpers/contentTypes.ts:135)
    ← contentTypes.ts:627
    ← nextAsync (node:internal/streams/from:182)
```
Value-shape-independent: an `AckShapeGen` whose yielded items carry a real `.acknowledge()` succeeds 5/5 on the items — only the terminal `undefined` crashes.

## Why it wasn't hit before
The common infinite-subscription SSE case (`IterableEventQueue`) never reaches a terminal `done` step. A plain generator returned over `Accept: application/json` closes cleanly (control) — the bug is confined to the SSE `serialize`/`transformIterable` path on **finite** generators.

## Repro
A minimal `static async *connect()` yielding a few plain JSON objects then returning, requested with `Accept: text/event-stream`, hangs (response `end` never fires) + logs the uncaughtException. The same generator as `application/json` closes cleanly. Harper `228eacc0f`.

## Fix
Guard `transformIterable` to not `transform` the terminal `done` step (`if (!step.done)`), and/or null-guard `serialize(message)`.

Found via exploratory QA. Related but distinct from #1399 (SSE *teardown* deferral).

kriszyp:
Closes #1628.

## Summary
`transformIterable` (`server/serverHelpers/contentTypes.ts`) applied the serializer transform to a finite iterator's terminal `{ value: undefined, done: true }` step. The SSE `serialize()` opens with `if (message.acknowledge) message.acknowledge()`, so the terminal `undefined` threw `TypeError: Cannot read properties of undefined (reading 'acknowledge')` — a process `uncaughtException` — and the HTTP response never closed. Any custom Resource that streams a **plain async generator to natural completion over `Accept: text/event-stream`** hangs.

The fix guards `transformIterable` to pass the terminal `done` step through untransformed, in both the sync and async branches. This is the root-cause fix (a transform should never see the sentinel `undefined`), so it also benefits the other consumer (ndjson) which previously ran a wasted `JSONStringify(undefined)` on the terminal step.

## Why it wasn't hit before
The common infinite-subscription SSE case (`IterableEventQueue`) never reaches a terminal `done` step; a plain generator over `application/json` closes cleanly. The bug is confined to the SSE streaming path on **finite** generators — directly on the LLM/`openaiStream` completion pattern (#631) and any bounded/export-over-SSE stream.

## Where to look
- `server/serverHelpers/contentTypes.ts` — the 5-LOC guard. On `done`, the raw step is returned as-is (its `value` is discarded by the async-iterator protocol / `Readable.from`), avoiding the transform, an extra allocation, and a redundant `step.done` read.
- `unitTests/server/serverHelpers/contentTypes.test.js` — two regression tests (async generator + sync iterable) that fail with the exact `acknowledge` error before the fix and pass after.
- `unitTests/resources/models/openaiStream.sse-http.test.js` — comment-only update: that test predates #1628 and documented this hazard as the reason it hand-iterates instead of using `serializeStream`; the note now points at the fix.

## Out of scope (pre-existing)
A non-terminal `yield undefined` (`{ value: undefined, done: false }`) still reaches `transform(undefined)` and would throw on the SSE path. That's a distinct pre-existing edge case, not introduced here.

## Review
Cross-model review (standard: Gemini + Harper-domain adjudication): no blockers, no significant concerns. The one Gemini suggestion (skip the terminal-step allocation) is folded into the fix.

### DIFF
--- a/server/serverHelpers/contentTypes.ts
+++ b/server/serverHelpers/contentTypes.ts
@@ -623,15 +623,10 @@ function transformIterable(iterable, transform) {
 				next() {
 					const step = iterator.next();
 					if (step.then) {
-						return step.then((step) => ({
-							value: transform(step.value),
-							done: step.done,
-						}));
+						// don't transform the terminal sentinel step, whose value is undefined
+						return step.then((result) => (result.done ? result : { value: transform(result.value), done: false }));
 					}
-					return {
-						value: transform(step.value),
-						done: step.done,
-					};
+					return step.done ? step : { value: transform(step.value), done: false };
 				},
 				return(value) {
 					return iterator.return(value);
--- a/unitTests/resources/models/openaiStream.sse-http.test.js
+++ b/unitTests/resources/models/openaiStream.sse-http.test.js
@@ -38,9 +38,9 @@ async function* toolCallChunks() {
 // serializer. This mirrors Harper's actual SSE path: a request with `Accept: text/event-stream`
 // is dispatched as CONNECT (server/REST.ts:24-25,137-139) and the resulting message stream is
 // iterated and serialized per message — never the terminating `done` step. So we call the real
-// `sse.serialize` (contentTypes.ts) on each yielded message, exactly as that path does. (We do
-// NOT use `sse.serializeStream`: its `transformIterable` transforms the terminal `undefined`
-// value, which the resource-SSE path avoids by manual iteration.)
+// `sse.serialize` (contentTypes.ts) on each yielded message, exactly as that path does. (This
+// test predates #1628; `sse.serializeStream` now skips the terminal `done` step and finishes a
+// finite generator cleanly — see the serializeStream SSE coverage in contentTypes.test.js.)
 // Listens on an ephemeral loopback port so concurrent runs stay isolated.
 function serveOnce(tokens, opts) {
 	const server = http.createServer(async (_req, res) => {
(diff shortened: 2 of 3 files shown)


Ethan-Arrowood (approved):
Nice root-cause fix, and good regression coverage.

PR #1641, opened 2026-07-07
fix(blob): never gate apply commits on wire-carried save flags; gate local writes on blob durability

### LINKED ISSUE #1640: Replicated blobs created with saveBeforeCommit deadlock the receiver's apply loop
### Bug Summary

Replicated blobs created with `saveBeforeCommit` can deadlock the receiving node's apply loop, stalling replication and bulk copies on blob tables.

### Steps to Reproduce

1. Cluster two or more nodes and create a table with a `Blob` attribute.
2. Have the application write blobs with `createBlob(data, { saveBeforeCommit: true })` (a caching component that stores rendered pages, for example).
3. Generate sustained writes while a peer catches up or runs a bulk copy, e.g. restart one node under load.
4. Watch the receiving node's replication status and table counts.

### Expected Behavior

Records and their blobs replicate, catch-up and bulk copies complete, blob tables converge across the cluster.

### Actual Behavior

The receiver's apply loop stalls waiting for a transaction commit that never resolves. The record's pre-commit gate waits for the blob's save to finish, but the blob's chunks are queued behind the same websocket the apply loop paused at its high-water mark, so neither side can make progress. `cluster_status` shows the leg pinned at ~100 percent backpressure with megabytes unread in the socket receive queue. The sender's copy watchdog terminates the connection every 300s, the copy restarts and stalls the same way. Blob tables never converge; non-blob tables on the same cluster are unaffected.

The cause is that `pack()` spreads the blob's own properties into the wire encoding, so the origin's `saveBeforeCommit` write instruction arrives on the receiver and re-arms the commit gate against the in-flight receive stream.

### Platform

Harper 5.1.x, Node 24, Docker on Ubuntu 22.04, 4-node cluster with cross-region links

### Console Errors

```
Transaction was open too long and has been committed, from table: PageCache/
Receive watchdog: no activity from <node> (db: "seo") for 300000ms terminating connection and reconnecting
Copy-progress watchdog: no base-copy progress from <node> (db: "sem") for 900000ms while connected
```

### Additional Context

Reproduces within minutes under a k6 workload (3 VUs, ~50KB blobs) whenever a node restarts or a bulk copy runs concurrently with writes. Recovery via reconnect fails because the resumed copy hits the same deadlock immediately.

ldt1996:
Fixes #1640

- The apply path (`trackPersistedBlobs`) now runs first in `startPreCommitBlobsForRecord` and only tracks blobs for cleanup, never gating the commit, and `pack()` no longer ships `saveBeforeCommit`/`saveInRecord`. Wire-carried flags could previously gate a receiver's commit on an in-flight receive stream, freezing the apply loop (see issue).
- Local writes with a fresh blob now gate the record commit on the blob's durable save, so a record can no longer commit (and replicate) with a not-yet-durable blob reference.
- Already-saved blobs (`fileId` set) are exempt from that gate: a local write carrying one references another row's blob, and tracking it would let abort cleanup delete a file the owning record still needs.

Behavior change: a local write whose fresh blob fails to save now rejects instead of committing a broken reference, matching the existing `saveBeforeCommit` fault contract. Four tests updated accordingly.

Validated on the 4-node staging repro from the issue: blob tables converge to exact row equality within minutes of load stop. `blob.test.js` 62 passing; the 3 `test:unit:resources` failures reproduce on `origin/main` and are unrelated.

### DIFF
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -1589,8 +1589,23 @@ export function startPreCommitBlobsForRecord(
 					value.saveInRecord = true;
 				}
 				blobsNeedingSaving.push(value);
-			} else if (trackPersistedBlobs && storageInfoForBlob.get(value)?.fileId != null) {
-				blobsToTrackOnly.push(value);
+			} else if (trackPersistedBlobs) {
+				// Replication-apply path: the blob was received out-of-band and its durability
+				// is tracked separately via outstandingBlobsToFinish. Awaiting it here would
+				// deadlock a paused WS (see the block comment above).
+				if (storageInfoForBlob.get(value)?.fileId != null) {
+					blobsToTrackOnly.push(value);
+				}
+			} else {
+				// Local-write path with a fresh blob: gate the record commit on the blob's
+				// durable landing. Without this the record commits before the async save has
+				// fsync'd, and a peer that pulls the record first sees an orphan ref and marks
+				// the record permanently diverged (a variant of the harper-pro#481 pattern
+				// reachable via the write path rather than the receive path). Not a deadlock
+				// risk like the trackPersistedBlobs branch: local blob sources are the HTTP
+				// body or an in-memory buffer, not a paused WS receive stream.
+				currentStore = store;
+				blobsNeedingSaving.push(value);
 			}
 		}
 	}
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -1601,7 +1601,7 @@ export function startPreCommitBlobsForRecord(
 					value.saveInRecord = true;
 				}
 				blobsNeedingSaving.push(value);
-			} else {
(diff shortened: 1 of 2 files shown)


ldt1996:
Both addressed in f6b1b96: the still-streaming gap was real, but keying on `saving != null` alone would break the cross-row reuse contract (the saving promise sticks around after it settles), so the gate uses a `saved` marker instead, with a regression test; also swapped the deletes for destructuring.
The failing unit check is the same LMDB flake that's been red on main for the last 5 runs: https://github.com/HarperFast/harper/actions/runs/28830783277

*Lavinia, via Claude*

kriszyp (approved):
Reviewed via Claude review-queue — sound blob-durability fix, local writes now correctly block commit on save durability. CI red is a pre-existing unrelated subscriptionReplay flake. LGTM.

PR #1643, opened 2026-07-07
fix(queue): waitForDrain poll fallback for drains that never emit

### LINKED ISSUE #1642: waitForDrain can hang on an already-empty IterableEventQueue, pinning the replication receive pause
### Bug Summary

`IterableEventQueue.waitForDrain()` can hang forever on a queue that is already empty, leaving the replication receive socket paused indefinitely.

### Steps to Reproduce

1. Buffer at least one message into an `IterableEventQueue` that has no consumer yet (`send()` with no data listener).
2. Call `waitForDrain()` while the queue is non-empty.
3. Attach a `data` listener: the attach path drains the buffered queue synchronously without emitting `drained`.
4. The `waitForDrain()` promise never resolves even though the queue is empty.

### Expected Behavior

`waitForDrain()` resolves once the queue is empty, regardless of which path emptied it.

### Actual Behavior

The waiter only resolves on the `drained` event, which `getNextMessage()` emits on a shift from an empty queue. The `on('data')` attach loop and the `resolveNext` bypass empty the queue without emitting it, so the waiter parks forever. In replication this waiter guards the receive high-water-mark pause, so the socket stays paused and the connection sits at 100 percent backpressure with an empty queue until a watchdog kills it (observed live on a 4-node cluster during bulk copies, `queue=0` with the drain waiter still parked).

### Platform

Harper 5.1.x, Node 24

### Additional Context

Found while instrumenting the copy-freeze investigation in #1640; independent bug, different mechanism.

*Lavinia, via Claude*

ldt1996:
Fixes #1642

- `waitForDrain()` now arms a 100ms poll alongside the `drained` listener, so a waiter can no longer hang when the queue empties through a path that never emits the event (the `on('data')` attach loop, the `resolveNext` bypass). Regression test included.
- Also exposes `Table.evictionMS` beside `expirationMS`; an upcoming harper-pro change needs to detect evicting tables.

*Lavinia, via Claude*

### DIFF
--- a/resources/IterableEventQueue.ts
+++ b/resources/IterableEventQueue.ts
@@ -42,16 +42,18 @@ export class IterableEventQueue<Event extends object = any> extends EventEmitter
 				// The queue can also empty through paths that never emit 'drained' (the on('data')
 				// attach loop and the resolveNext bypass), so a waiter relying on the event alone
 				// can hang forever on an already-empty queue. Poll as a fallback wakeup.
+				const settle = (drained: boolean) => {
+					clearInterval(poll);
+					this.removeListener('drained', onDrained);
+					resolve(drained);
+				};
+				const onDrained = () => settle(true);
+				this.once('drained', onDrained);
+				this.currentDrainResolver = settle;
 				const poll = setInterval(() => {
 					if (!this.queue || this.queue.length === 0) settle(true);
 				}, 100);
 				poll.unref?.();
-				const settle = (drained: boolean) => {
-					clearInterval(poll);
-					resolve(drained);
-				};
-				this.once('drained', () => settle(true));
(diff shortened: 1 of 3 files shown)


ldt1996:
Good catch, fixed in bb63b78: settle now removes the drained listener, and the test asserts listenerCount goes back to zero after a poll-path settle.
If the unit check goes red again it should be the same LMDB flake that's currently failing on main 

*Lavinia, via Claude*

kriszyp (approved):
Reviewed via Claude review-queue — solid poll-fallback fix for drains that never emit; unref'd 100ms poll only while waiting, negligible cost. Unrelated Windows CI flake persists (also red on main). LGTM.

PR #1694, opened 2026-07-07
MCP: accept missing MCP-Protocol-Version as the session's negotiated version

### LINKED ISSUE #1611: MCP: relax MCP-Protocol-Version mismatch — treat missing header on an established session as the session's negotiated version
Two small conformance / consistency observations from testing a live `application`-profile instance (`harper@5.1.15`), MCP spec rev **2025-06-18**.

### 1. `initialize` returns HTTP 400 when the `MCP-Protocol-Version` header is absent
An `initialize` POST **without** an `MCP-Protocol-Version` header returns `400 Bad Request`; adding `MCP-Protocol-Version: 2025-06-18` makes it `200`. The transport spec says: *"if the server does **not** receive an `MCP-Protocol-Version` header … the server **SHOULD** assume protocol version `2025-03-26`."* A hard 400 on the **initialization** request is stricter than that SHOULD — and many clients don't send the header on the first `initialize` (the version isn't negotiated yet; it's carried in the body's `params.protocolVersion`).

Repro:
```
curl -sD- -H 'content-type: application/json' -H 'accept: application/json, text/event-stream' \
  -X POST http://<host>/mcp \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"c","version":"0"}}}'
# → HTTP/1.1 400 Bad Request     (add MCP-Protocol-Version header → 200)
```

### 2. `initialize` advertises `resources.subscribe: true`, but subscriptions aren't implemented
The `initialize` result advertises `capabilities.resources.subscribe: true`, while the docs (`reference/v5/mcp/tools-and-resources`) state *"resources are static (no `resources/subscribe` in v1)"* and #1349 lists subscriptions as **planned**. Advertising `subscribe: true` invites clients to call `resources/subscribe` against a server that won't honour it. Either land subscriptions (tracked in #1349) or stop advertising the capability until then.

kylebernhardy:
Fixes #1611 (as repurposed after investigation — both originally-filed items were verified unreproducible/stale on 5.1.17; see the issue comment with runtime evidence).

## Summary

Requests on an established session that omit `MCP-Protocol-Version` were treated as `2025-03-26` (the spec's backcompat assumption) and then failed the negotiated-version check — so any client that negotiated `2025-06-18` but doesn't send the header on every request got `400 MCP-Protocol-Version mismatch`. The session record already knows the negotiated version, so a missing header now validates as that version. A header that IS present must still name a supported version and match the negotiated one (both still 400).

One conditional in `validateProtocolHeader` (covers both call sites: POST dispatch and GET SSE open); `initialize` never validated the header and is unchanged.

**Spec posture** (flagging for review): the assume-2025-03-26 rule is a SHOULD aimed at clients predating the header; such clients negotiate 2025-03-26 anyway, so deferring to the session version is equivalent for them and stops penalizing modern clients that violate the MUST-send-header rule on non-critical requests. No downgrade risk — the authority is the server-side session record, not anything an attacker can strip in transit.

## Testing

Flipped the unit test that asserted the 400, added a headerless GET SSE case; verified at runtime on a live instance (headerless `tools/list` on a 2025-06-18 session → 200; `2025-03-26` header on the same session → 400; garbage version → 400). No docs companion: the docs don't document the header behavior.

_Generated by an LLM (Claude Fable 5); cross-reviewed by Codex (no findings) and Gemini (no findings; confirmed both call sites, the empty-string-header edge, and the no-downgrade posture)._

### DIFF
--- a/components/mcp/transport.ts
+++ b/components/mcp/transport.ts
@@ -25,12 +25,7 @@ import {
 	type JsonRpcId,
 	type JsonRpcMessage,
 } from './jsonrpc.ts';
-import {
-	handleInitialize,
-	handleInitialized,
-	PROTOCOL_VERSION_BACKCOMPAT,
-	SUPPORTED_PROTOCOL_VERSIONS,
-} from './lifecycle.ts';
+import { handleInitialize, handleInitialized, SUPPORTED_PROTOCOL_VERSIONS } from './lifecycle.ts';
 import { emitAuditEntry } from './audit.ts';
 import { emitMcpLogToSession, isValidMcpLogLevel, setSessionLogLevel } from './logging.ts';
 import { decodeCursor } from './pagination.ts';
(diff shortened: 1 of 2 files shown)


kriszyp:
Hey @kylebernhardy — retargeted this from `v5.1` to `main`. New PRs should always target `main`; `v5.1` is patch-only and gets populated automatically by the cherry-pick bot when a merged `main` PR carries the `patch` label. If this needs to ship in a patch release, add the `patch` label once it's merged to `main` instead of targeting `v5.1` directly.

kylebernhardy:
Understood — thanks for the workflow correction (main-first, `patch` label → cherry-pick bot; will target main going forward). Added the `patch` label since the strict-mismatch 400 bites 5.1.x clients too. _Comment generated by an LLM (Claude Fable 5)._

kriszyp (approved):

Issue #1702, opened 2026-07-08
LMDB structon typed-structure dictionary can corrupt under overlapping/interleaved commits to the same store


kriszyp:
## Summary

Under `HARPER_STORAGE_ENGINE=lmdb`, two commits to the **same store** for records with **different
top-level shapes** can, if their commits overlap/interleave in time, corrupt the per-store `structon`
typed-structure dictionary used by `RecordEncoder` (`resources/RecordEncoder.ts`, `RecordEncoder
extends StructonEncoder`). The symptom observed: a sibling key's structure id gets misapplied on
decode, silently emptying a nested object property on the next read of an otherwise-unrelated record.

## How this was found

While fixing #1641's regression (see the PR referenced below), we needed two cache-fill commits for
differently-shaped records on the same table (`CacheOfHttp`, `sourcedFrom`) to overlap in time to
reproduce a real, deterministic bug: `unitTests/apiTests/cache-test.mjs`'s "get resolved with
Response" test failed under LMDB with `response.headers.get('cache-control')` coming back
`undefined` — the stored record's `headers` sub-object decoded as `{}`.

Root cause (confirmed via debug instrumentation + bisection, not just theory): #1641 added an
`await` to `startPreCommitBlobsForRecord`'s local-write gate that, when it unintentionally applied to
`Table.ts`'s `getFromSource()` cache-fill path too, widened the window during which two cache-fill
commits (for different keys, same table/store) could interleave. That interleaving — not the blob
gating mechanism itself — is what corrupted the structure dictionary. The PR fixing #1641's
regression (linked below) avoids re-introducing this specific overlap window, but the underlying
concurrency bug in `structon`'s typed-structure dictionary is still latent and could be re-triggered
by any future change that introduces a new `await` between a commit's preparation and its execution
for a store using struct-mode encoding.

This is likely the same class of bug the team has hit before (per internal notes): issue #1163
(typed-structure divergence, "fix=disable typed structs") and the "structures fork" pattern seen in
v4→v5 migration structure-id forks. Worth checking whether those are actually the same root cause or
a related-but-distinct one.

## Reproduction

```
HARPER_STORAGE_ENGINE=lmdb npx mocha unitTests/apiTests/basicREST-test.mjs unitTests/apiTests/cache-test.mjs
```

(Only reproduces on the commit that introduced the extra `await`; see the linked PR for the
avoidance fix. To force the underlying dictionary race again for investigation, reintroduce an
[...]

PR #1703, opened 2026-07-08
fix(blob): don't gate sourcedFrom cache-fill commits on blob durability

kriszyp:
## Summary

`main` has been red on the LMDB leg of `npm run test:unit:all` since #1641 (`807d2e280`,
"never gate apply commits on wire-carried save flags; gate local writes on blob durability").
`unitTests/apiTests/cache-test.mjs`'s "get resolved with Response" test fails deterministically
under `HARPER_STORAGE_ENGINE=lmdb`: `response.headers.get('cache-control')` comes back `undefined`.
This only shows up on the LMDB leg (`HARPER_STORAGE_ENGINE=lmdb npm run test:unit:apitests`), which
is why it slipped through — the default RocksDB leg always passes.

## Root cause

#1641 added a branch to `startPreCommitBlobsForRecord` (`resources/blob.ts`) that gates a record's
commit on a freshly-created `FileBackedBlob`'s durable save completing. That's correct for genuine
client writes (PUT/POST body blobs) — it stops a peer from replicating a record before its blob file
is fsync'd (a harper-pro#481-style orphan ref).

It also unintentionally caught `Table.ts`'s `getFromSource()` — the cache-fill path for
`sourcedFrom` tables — which creates a `FileBackedBlob` from an HTTP `Response` body when caching it
(`createBlob(updatedRecord.body)`). That write isn't the client-write path #1641 targets; it's a
per-node cache derivation from an external source. Gating it added a new await between one cache-fill
commit and its sibling commits (different keys, same table) that didn't exist before. Debug
instrumentation showed that overlap window lets two commits for **differently-shaped** cache records
(e.g. `{headers, body}` vs `{headers, data}`) interleave and corrupt the per-store `structon`
typed-structure dictionary used by the LMDB `RecordEncoder`: a sibling key's structure id gets
misapplied on decode, silently emptying a nested object property (`headers`) on the next
cache-hit read. Confirmed via bisection: green at the commit before #1641, red starting at #1641,
reproducible in isolation with:

```
HARPER_STORAGE_ENGINE=lmdb npx mocha unitTests/apiTests/basicREST-test.mjs unitTests/apiTests/cache-test.mjs
```

## Fix

- `startPreCommitBlobsForRecord` gets a new `isSourceResolution` parameter. For a source-resolution
  blob, it still calls `saveBlob()` eagerly and tracks the blob for skip/abort cleanup (via
  `blobsToTrackOnly`, with a `.catch()` that logs a save failure) — but never pushes it into
  `blobsNeedingSaving`, so the record's commit is no longer blocked on the save completing. This
  restores the pre-#1641 commit timing for this call site (avoiding the structon race) while keeping
  orphan-file cleanup, which this call site never had before.
- `getFromSource()`'s `sourceWrite.before = preCommitBlobsForRecordBefore(...)` call now passes
  `isSourceResolution: true`.
- Added a regression test to `cache-test.mjs` reproducing the interleaved-shapes scenario in one
  file (no cross-file ordering needed); verified it fails on the pre-fix code and passes reliably
  (5/5 runs) with the fix, under `HARPER_STORAGE_ENGINE=lmdb`.
- Added a `DESIGN.md` note under "Blob orphan cleanup" documenting the trap for future agents adding
  a new blob-creating write path.

## Testing

- `unitTests/apiTests/cache-test.mjs`: red → green (5/5) under `HARPER_STORAGE_ENGINE=lmdb`.
- `unitTests/apiTests/**/*-test.mjs`, both RocksDB (default) and `HARPER_STORAGE_ENGINE=lmdb`: all
  green except one pre-existing, unrelated flake (`Multi-threaded cache updates` hits a unix-socket
  `EINVAL` from this checkout's long `.claude/worktrees/<name>/...` path exceeding the local
  `sun_path` limit — reproduces identically on unmodified `main` in the same location; the path is
  much shorter in CI so it won't occur there).
- `unitTests/resources/blob.test.js` (the actual #1641 regression coverage) x3: still green —
  the original orphan-ref fix for genuine local writes is intact.
- `npm run test:unit:main` (2992 passing) and `npm run test:unit:resources` (1017 passing): green.
- `integrationTests/database/struct-cache-fidelity.test.ts` (structon cap-fidelity coverage): green,
  unaffected.

## Where to look / open question for reviewers

A Codex review pass flagged that fully exempting the cache-fill blob from the durability gate also
removes error-propagation: if the source's body stream fails mid-save, the record now commits
anyway (pointing at a missing/partial blob) instead of aborting, which was #1641's incidental
behavior for this call site. I mitigated this (eager `saveBlob()` + failure logging + cleanup
tracking on abort/skip) but did **not** restore full abort-on-save-failure semantics, since doing so
without re-introducing the commit-timing race would need a bigger structural change (e.g. gating on
save-start rather than save-completion isn't sufficient; truly decoupling failure detection from
commit timing would need a way to invalidate an already-committed cache row after the fact). I judged
the residual gap acceptable — a cache row is a per-node derived copy that self-heals on the next
fetch/invalidation/read-error, unlike an authoritative replicated write — but flagging it here since
it's a real, if narrow, trade-off a human reviewer should weigh in on.

Deeper still: the actual corruption mechanism is a pre-existing race in the LMDB `structon`
typed-structure dictionary under overlapping/interleaved commits to the same store (possibly the same
class of bug as issue #1163 / the "structures fork" pattern). #1641 exposed this latent race by
adding a new overlap opportunity; it didn't create it. This PR avoids re-introducing that specific
overlap rather than fixing the underlying structon concurrency issue, which is a separate, larger
effort — filed as #1702.

This PR was reviewed by both Codex and Gemini (via `agy`) per the team's cross-model review process;
both approved, with Gemini's only non-trivial suggestion (defer the deeper structon fix, track it
separately) already actioned as #1702.

Generated with the help of Claude Sonnet 5.


kriszyp:
Superseded by #1709, which landed the actual root-cause fix (finalizeResponse was mutating live cache records in place, not a structon/LMDB dictionary corruption as #1702 originally theorized). Closing this stopgap in favor of that.

PR #1711, opened 2026-07-08
fix: honor charset in JSON deserializer and fix async error handling

### LINKED ISSUE #1630: REST JSON ingress ignores charset: non-UTF-8 body stored as mojibake (silent); utf-16 body → 500 not 400 (async parse-error escapes sync try/catch)
## Summary
REST JSON ingress **ignores the request `charset` parameter**. A non-UTF-8 JSON body (`Content-Type: application/json; charset=iso-8859-1`) is silently decoded as UTF-8 → **mojibake string corruption stored with no error (204)**. A `charset=utf-16le` body → **500 not 400** (the streaming deserializer's Promise rejection escapes the synchronous `try/catch`). Chunked framing itself is sound.

## Repro (harper `228eacc0f`, raw TCP — fetch/supertest force Content-Length)
- `charset=iso-8859-1`, body `"café ñoño"` → **204 accepted**, stored + read back as `"caf� �o�o"` (U+FFFD per byte), no error signal.
- `charset=utf-16le`, non-ASCII → **500** `SyntaxError` (embedded NULs break JSON), not a clean 4xx.
- (Positive controls: chunked no-Content-Length UTF-8 with a multibyte char split across a chunk boundary → 200 exact; ~500KB irregular chunks → byte-exact; trailing `; boundary=x` param harmlessly ignored.)

## Root causes (code-confirmed)
1. **Silent corruption:** `server/serverHelpers/contentTypes.ts` — `mediaTypes.get('application/json').deserialize = (data) => JSONParse(data)` runs on a raw `Buffer` and never reads `contentType.parameters.charset`. `JSON.parse(buffer)` coerces via `buffer.toString()` = **always UTF-8**, so any declared non-UTF-8 charset is silently ignored; single-byte charsets (Latin-1/ISO-8859-1) produce syntactically-valid JSON with per-byte U+FFFD-corrupted values → no error.
2. **4xx-as-500:** `server/REST.ts` (~L130-140) wraps the deserializer in a synchronous `try/catch`, but `getDeserializer(..., streaming: true)` returns `stream => streamToBuffer(stream).then(deserialize)` — a Promise. The parse `SyntaxError` rejects **asynchronously**, escaping the synchronous catch, and surfaces at the generic outer handler with no `statusCode` → 500. (Any parse error on a streaming body → 500 not 400 by this mechanism.)

## Fix
Honor `parameters.charset` (decode the Buffer with the declared charset before `JSON.parse`, or return 415 for an unsupported charset), and `await` the streaming deserializer inside the try (or attach a `.catch` mapping to 400).

Found via exploratory QA.

abhi-0203:
## Summary

Two issues fixed in the REST JSON ingress:

1. **charset handling**: JSON deserializer now honors the `charset` parameter from Content-Type header
2. **async error handling**: Streaming deserialization errors now properly return 400 instead of 500

### contentTypes.ts
- Updated `application/json` deserializer to accept optional `contentTypeString` parameter
- When charset is specified and not UTF-8, decode the buffer with the specified charset before JSON parsing
- Updated `getDeserializer` to wrap the deserializer and pass the content type string

### REST.ts
- Made the transaction callback `async`
- Added `await` to the streaming deserialization call so JSON parse errors are properly caught

### Silent corruption (charset)
`Content-Type: application/json; charset=iso-8859-1` with body `"café ñoño"` was silently decoded as UTF-8, producing mojibake (`"caf� �o�o"`) stored with no error (204).

### 500 instead of 400 (async)
`Content-Type: application/json; charset=utf-16le` with non-ASCII body caused a `SyntaxError` that rejected asynchronously, escaping the synchronous try/catch and surfacing as 500 instead of 400.

## Testing

- Verified charset=iso-8859-1 body is correctly decoded
- Verified charset=utf-16le parse error returns 400
- Existing tests pass

Closes #1630

### DIFF
--- a/server/REST.ts
+++ b/server/REST.ts
@@ -126,11 +126,11 @@ async function http(request: Request, nextHandler) {
 		if (replicateFrom === 'none') {
 			request.replicateFrom = false;
 		}
-		let responseData = await transaction(request, () => {
+		let responseData = await transaction(request, async () => {
 			if (headersObject['content-length'] || headersObject['transfer-encoding']) {
 				// TODO: Support cancellation (if the request otherwise fails or takes too many bytes)
 				try {
-					request.data = (getDeserializer(headersObject['content-type'] as any, true) as any)(
+					request.data = await (getDeserializer(headersObject['content-type'] as any, true) as any)(
 						request.body,
 						request.headers
 					);
--- a/server/serverHelpers/contentTypes.ts
+++ b/server/serverHelpers/contentTypes.ts
@@ -579,6 +579,18 @@ export function getDeserializer(
 	const deserialize =
 		(contentType.type && mediaTypes.get(contentType.type)?.deserialize) || deserializerUnknownType(contentType);
 
+	// For application/json with non-UTF-8 charset, create a wrapper that handles decoding
+	if (contentType.type === 'application/json' && contentType.parameters?.charset) {
+		const charset = contentType.parameters.charset.toLowerCase();
+		if (charset !== 'utf-8' && isBufferEncoding(charset)) {
+			const wrappedDeserialize: Deserialize = (data) => {
+				const decoded = data.toString(charset as NodeJS.BufferEncoding);
+				return JSONParse(decoded);
+			};
+			return streaming ? (stream: Readable) => streamToBuffer(stream).then(wrappedDeserialize) : wrappedDeserialize;
+		}
+	}
+
 	return streaming ? (stream: Readable) => streamToBuffer(stream).then(deserialize) : deserialize;
 }
 


kriszyp (commented):
Fixes a real bug (the streaming deserializer's promise was never awaited, so a parse failure escaped the `try/catch` as an unhandled rejection / raw 500 instead of the intended 400) — good catch. Two things worth a quick look before merge:

**Significant — the charset fix may not cover its own repro.** `isBufferEncoding`'s allow-list has `'latin1'`, not `'iso-8859-1'` (Node doesn't alias them: `Buffer.isEncoding('iso-8859-1') === false`, `Buffer.isEncoding('latin1') === true`). If the manual verification used `charset=iso-8859-1` literally (the token in the linked issue #1630 and this PR's own description) rather than `charset=latin1`, the new branch never matches and the original silent-mojibake bug still reproduces for that exact case. Same gap likely applies to other common HTTP charset tokens that don't literally match a Node encoding name (`us-ascii` vs `ascii`, bare `utf-16` vs `utf-16le`). Worth double-checking which token was actually used in testing.

**Perf — the whole `transaction` callback going `async` costs every request, not just JSON bodies.** `resources/transaction.ts` has an explicit sync/async split that lets a synchronously-resolving callback (most GET/HEAD/DELETE) commit on the same tick with no promise allocation. Making the callback passed here `async` means the result is now always a `Promise` for every method through this handler, even when the `await` line is never reached (no `content-length`/`transfer-encoding`). Would it work to scope the `await` to just the deserialization — resolving `request.data` in a small async helper *before* calling `transaction(...)`, keeping the callback itself synchronous for the common case?

No test file is included despite the PR description referencing manual verification — happy to add unit test coverage for both the charset-decoding path (including the `iso-8859-1`/`latin1` question above) and the async-error-propagation fix if that'd help get this landed. Let me know and I'll put a PR together.

PR #1717, opened 2026-07-08
feat: registry-auth via the hdb_secret store — persist provided tokens, resolve on every install

kriszyp:
Reworks the deploy_component private-registry auth from #1158 so a registry token is treated as what it is — a secret — and Harper owns its lifecycle through the `hdb_secret` store (#1550/#1582). Supersedes and **closes #1158**.

**Stacked on #1582** (`kris/component-secrets-1550`), which carries the #1554 store + custody. Merge order: **#1554 → #1582 → this**; retarget to `main` after #1582 merges. Draft until then.

## The problem with the transient model

#1158 treated a provided `token` as ephemeral: use it for this node's install, strip it before replication, and rely on harper-pro's fabric-injected `NPM_CONFIG_USERCONFIG` for peers. Because Harper never owned the credential, three things silently broke — all for **package-reference** deploys (`npm:@org/app@x`), which re-run `npm install` on every node:

1. **Rollback** to a prior version has no token.
2. **Reboot / cold reinstall** (wiped components dir) has no token.
3. **New peer** joining depends entirely on out-of-band per-host userconfig.

Now that #1582 gives us an envelope-encrypted, replicated, custody-decryptable store, persisting the token there is both safer than a per-host plaintext `.npmrc` and closes all three.

## The model: ingest → reference → resolve on every install

- **Ingest** (`ingestRegistryAuth`): a provided `{ registry, token }`, when custody is present, is encrypted into `hdb_secret` (via `set_secret`) under a derived name `deploy.<component>.<registry>`, granted to the component (path-A authority), and returned as a `{ registry, secret }` reference. Idempotent on token rotation (one current token per component+registry).
- **Persist references (never tokens) in two durable places**: the component config (`applicationConfig.registryAuth`, read on every cold install) and the `hdb_deployment` row (`registry_auth`, the rollback source).
- **Replicate references, not tokens**: `req.registryAuth` now carries references to peers; each peer resolves from its **own replicated `hdb_secret` copy**. The token only ever crosses the cluster as ciphertext. `resolveRegistryAuth` gains a bounded `waitMs` for the case where the secret row arrives just behind the deploy op.
- **Cold-install resolution** (`installApplications`): resolves `applicationConfig.registryAuth` at boot so a fresh/wiped node authenticates from the store (best-effort — logs and installs without auth if custody isn't up yet, rather than blocking boot).
- **Rollback** (when the op lands) is then trivial: read `hdb_deployment.registry_auth` → resolve. No deploy-machinery changes needed later.

## Authority (decided: keep the grant gate)

A referenced secret must be usable by the component being deployed — `processEnv`-global **or** granted to `<component>` — else 403; missing → 404; no custody / decrypt failure → fails loudly. This mirrors #1582's accessor model so the deploy path can't pull an arbitrary scoped secret into an `.npmrc`. Auto-minted deploy secrets are granted to their component by construction.

## Degraded modes
- **No custody (OSS core / key not held)**: literal tokens stay on the #1158 transient path (this-node install only, stripped, not persisted); `{ secret }` references fail loudly.
- **Config / deployment row / validation**: registryAuth is references **only** — `assertApplicationConfig` rejects a literal token on disk.

## Fabric userconfig injection (D3)
Store resolution replaces the **auth** role of harper-pro's `NPM_CONFIG_USERCONFIG` injection, so a coordinated **harper-pro follow-up** drops the token injection. Core keeps userconfig *inheritance* (`writeTransientNpmrc` prepend) for non-auth npm config (proxy/cafile/extra registries). Not removed here — it lives in harper-pro.

## Files
- `secretOperations.ts` — `ingestRegistryAuth`, `deriveRegistrySecretName`, `resolveRegistryAuth` (+ `waitMs`), `RegistryAuthReference`.
- `operations.js` (`deployComponent`) — ingest up front; write refs to config + recorder; replicate refs; resolve for this node's install (peer bounded-wait).
- `Application.ts` — `ApplicationConfig.registryAuth` (references), `assertApplicationConfig` validation, `installApplications` cold-install resolution.
- `deploymentRecorder.ts` — `registry_auth` reference field (rollback source).
- `operationsValidation.js` — each entry is `token` XOR `secret`.
- `serverUtilities.ts` — ops-log redaction union.

## Tests
- `resolveRegistryAuth`: pass-through, resolve scoped/global/mixed, all four failure reasons, **bounded-wait** (times-out-404 + replicates-in-mid-wait).
- `ingestRegistryAuth`: seal + round-trip, already-reference passthrough, no-custody transient fallback, idempotent rotation.
- `deriveRegistrySecretName`: sanitization + scheme/bare-host equivalence.
- `assertApplicationConfig`: references-only (rejects literal token on disk).
- Validator: token-XOR-secret. Existing #1158 lifecycle + recorder/deploy-lifecycle suites unchanged and green.

## Follow-ups
- harper-pro: drop the fabric token injection; keep any proxy/cafile injection.
- Payload (tarball) deploys: ingest+resolve works today; the durable ref rides `hdb_deployment` when the rollback op lands.


components/Application.ts
```diff
@@ -675,6 +789,41 @@ function getGitSSHCommand() {
...
+// the entry sets npm's default `registry=…` so an unscoped package spec (e.g. `npm:my-private-app`)
+// or its transitive deps actually resolve against this registry rather than the public default.
+// A scope-less entry therefore requires its registry to serve/proxy whatever npm needs to install;
+// with multiple scope-less entries npm's last-value-wins applies to the default `registry`.
+export function buildNpmrcContent(registryAuth: RegistryAuthEntry[]): string {
+	const lines: string[] = [];
+	for (const { registry, token, scope } of registryAuth) {
+		const registryUrl = normalizeRegistryUrl(registry);
+		const authKey = registryUrl.replace(/^https?:/i, '');
+		lines.push(`${authKey}:_authToken=${token}`);
```
cb1kenobi, line 841: **Nit: resolved-from-secret token isn't newline-guarded like the literal-token path.**

The validator forbids CR/LF in a literal `token` (`operationsValidation.js`, `/^[^\r\n]+$/`) specifically "so a super_user can't inject extra npm config lines." A token resolved from an `hdb_secret` row (`resolveRegistryAuth` → `custody.decrypt`) never passes that guard, so a secret value containing `\n` is written verbatim here and injects arbitrary `.npmrc` lines. Admin-only per Harper's threat model (not a vuln), but the literal path already defends this exact class — worth the parity.

Suggested fix: reject/strip CR/LF on the resolved token in `buildNpmrcContent` (or in `resolveRegistryAuth` after decrypt), matching the validator's guard.

—
_Generated by Barber AI_

components/secretOperations.ts
```diff
@@ -324,3 +324,168 @@ export function getSecretsPublicKey(req: any) {
...
+			throw new ClientError(
+				`registryAuth secret '${name}' is not granted to component '${component}' ` +
+					`(grant it with grant_secret, or set it processEnv:true)`,
+				HTTP_STATUS_CODES.FORBIDDEN
+			);
+		}
+		const custody = getSecretCustody();
+		if (!custody) {
+			throw new ClientError(
+				`secrets custody is not initialized on this node; cannot resolve registryAuth secret '${name}'`
```
cb1kenobi: **Nit: server-side failures return 400 (client error).** 

`ClientError` defaults to `statusCode 400` when none is passed. The missing-row (404) and missing-grant (403) cases are correctly client-fixable, but "custody is not initialized" (this line) and "Failed to decrypt" (line 486) are node/server-state conditions the caller can't fix by changing the request — a retry after custody comes up with the *same* body would succeed. Reporting them as 400 mislabels them.

Suggested fix: pass a 5xx (e.g. `HTTP_STATUS_CODES.SERVICE_UNAVAILABLE` for no-custody, `INTERNAL_SERVER_ERROR` for decrypt failure) to these two `ClientError`s.

—
_Generated by Barber AI_

components/Application.ts
kriszyp: Good parity catch — fixed. Rather than guard only the resolved path, I enforce the no-CR/LF invariant at the injection point (`buildNpmrcContent`) in 1e62d7141, so it holds for tokens from any source (literal or hdb_secret-resolved); a newline now throws before any `.npmrc` line is written. Added a rejection test. Thanks! 🤖 KrAIs (Claude Opus 4.8)

components/secretOperations.ts
kriszyp: Agreed — fixed in 1e62d7141. No-custody now returns 503 (SERVICE_UNAVAILABLE, retryable once custody is up) and decrypt failure returns 500 (INTERNAL_SERVER_ERROR); the missing-row 404 and missing-grant 403 stay as-is. Test assertions now pin the status codes. Thanks! 🤖 KrAIs (Claude Opus 4.8)

components/secretOperations.ts
```diff
@@ -324,3 +324,80 @@ export function getSecretsPublicKey(req: any) {
...
+	component: string
+): Promise<ResolvedRegistryAuthEntry[] | undefined> {
+	if (!Array.isArray(registryAuth) || registryAuth.length === 0) return registryAuth;
+	// Token-only requests must not require custody or a provisioned store — keep the fast path pure.
+	if (!registryAuth.some((entry) => entry && entry.secret !== undefined)) return registryAuth;
+
+	const table = secretTable();
+	const resolved: ResolvedRegistryAuthEntry[] = [];
+	for (const entry of registryAuth) {
+		if (entry.secret === undefined) {
```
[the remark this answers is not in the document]
kriszyp: Added the null-entry guard for parity with the fast-path check above. Note there is no `SKIP` sentinel in the registryAuth path (entries are Joi-validated `token` XOR `secret` objects), so I left that part out. 🤖 KrAIs (Claude Opus 4.8)

components/Application.ts
```diff
@@ -110,6 +115,28 @@ export function assertApplicationConfig(
 			);
 		}
 	}
+	if ('registryAuth' in applicationConfig && applicationConfig.registryAuth !== undefined) {
+		const entries = applicationConfig.registryAuth;
+		if (!Array.isArray(entries)) {
+			throw new (class InvalidRegistryAuthError extends TypeError {})(
```
cb1kenobi: Move this class definition to the top-level scope so we aren't creating a throw away class.

cb1kenobi (approved):

cb1kenobi:
Please run the formatter to clean up the code before merging.

dawsontoth (approved):

components/Application.ts
kriszyp: Hoisted to top-level scope. — KrAIs

Issue #1591, opened 2026-07-03
Audit records lack user attribution for writes from registered operations (no ambient operation context)


kriszyp:
## Summary

Writes performed inside operation handlers that use the static Resource API (`table.put(id, record)` / `table.delete(id)`) **without an explicit context** produce audit records with `user: undefined`, even though the operation was fully authenticated. This affects operations registered via `registerOperation()` (component-registered operations). The built-in CRUD path is **not** affected — it constructs its context explicitly.

## Root cause

- `processLocalTransaction` (`server/serverHelpers/serverUtilities.ts:70`) invokes the operation handler via `callOperationFunctionAsAwait` (`utility/OperationFunctionCaller.ts:14`) — a bare call. There is no `contextStorage.run()` anywhere in this chain; the only `contextStorage.run` in core is `resources/transaction.ts:53`, which only fires when a context is passed in explicitly.
- Static Resource methods fall back to `contextStorage.getStore() ?? {}` (`resources/Resource.ts`), so the write context is `{}`.
- The write commit records `user: (context as any)?.user` (`resources/Table.ts:2434` and sibling sites) → `RecordEncoder.ts:788` stores `undefined`.
- The authenticated user **is** available on the request (`req.body.hdb_user`; also set for internal dispatch at `serverUtilities.ts:218`) — it is simply never bridged into the async context.
- By contrast, `ResourceBridge` CRUD handlers build `{ user: ...hdb_user }` contexts explicitly (e.g. `dataLayer/harperBridge/ResourceBridge.ts:279`), which is why generic CRUD writes are attributed correctly.

## Impact

- `read_audit_log` with `search_type: username` can never surface these writes.
- Change-audit evidence ("who changed this record") is silently missing for registered-op writes — and registered operations typically manage security-relevant control tables, where attribution matters most (SOC 2 change management / PCI DSS Req 10 style evidence).

## Proposed fix (enforce the invariant, not per-handler patches)

Establish a user-bearing operation context at the dispatch layer: wrap the handler invocation in `processLocalTransaction` in `contextStorage.run({ user: req.body.hdb_user })`, so every Resource API call inside any operation handler inherits attribution — matching what the CRUD bridge already does explicitly. Handlers passing an explicit context still win (explicit context takes precedence over the ambient store).

Validation needed: audit existing `contextStorage.getStore()` consumers for any op that depends on running with an empty ambient context (e.g. deliberately system-attributed writes).

---
[...]

Issue #512, opened 2026-05-12
Register application/x-ndjson content type for iterator streaming


kriszyp:
## Context

`core/server/serverHelpers/contentTypes.ts` already registers `serializeStream` for `application/json`, `application/cbor`, `application/x-msgpack`, `text/csv`, `text/plain`, and `text/event-stream` — any Resource method returning an `AsyncIterable` is automatically streamed with content negotiation. `application/x-ndjson` (newline-delimited JSON) is conspicuously absent.

NDJSON is the standard choice for CLI clients (`curl | jq`), log tails, and tools that want streaming JSON without SSE framing. Adding it is a one-liner and improves DX for streaming endpoints (#510, change feeds, future log-tail resources).

## Proposed addition

In `contentTypes.ts`, register alongside the existing entries:

```ts
mediaTypes.set('application/x-ndjson', {
  serializeStream(data: any) {
    if (data?.[Symbol.iterator] || data?.[Symbol.asyncIterator]) {
      return Readable.from(transformIterable(data, (msg) => JSONStringify(msg) + '\n'));
    }
    return JSONStringify(data) + '\n';
  },
  serialize(data) {
    return JSONStringify(data) + '\n';
  },
  deserialize(data: Buffer) {
    return data.toString().trim().split('\n').map(JSONParse);
  },
  q: 0.7,
});
```

(Also register `application/ndjson` as an alias for older clients.)

## Acceptance

- [ ] `application/x-ndjson` (and `application/ndjson` alias) registered in `mediaTypes`.
- [ ] A method returning an `AsyncIterable` streams one JSON object per line when client sends `Accept: application/x-ndjson`.
- [ ] Round-trip works (`deserialize` reads NDJSON bodies).
- [ ] Backpressure honored (same Node stream pipeline as other `serializeStream` entries).

## Out of scope

- A streaming NDJSON parser for arbitrarily large request bodies (deserialize collects-to-array for v1; can split later if needed).

PR #1720, opened 2026-07-08
fix: only join an ambient transaction if it is genuinely still OPEN

kriszyp:
## Root cause

`resources/Resource.ts`'s static-API dispatcher (backing `Table.get`/`put`/`patch`/`delete`) decides whether to join an existing transaction on the ambient context or start a fresh one purely via `if (context?.transaction)` truthiness. This is safe only when ambient contexts are fresh, one-shot objects per call — which was always true before #1591/#1592, since `context = contextStorage.getStore() ?? {}` fell back to a brand-new `{}` whenever nothing had populated `contextStorage` for the current call chain.

#1591/#1592 ("establish ambient user context for operation handlers so audit records carry attribution") changed that: `processLocalTransaction` now wraps an *entire* operation handler invocation in one `contextStorage.run({ user: hdbUser }, handler)` call, installing a single, long-lived, mutable context object as the ambient store for the whole handler. `resources/transaction.ts`'s `transaction()` helper mutates that shared object in place (`context.transaction = new DatabaseTransaction()`) as a side effect of servicing the *first* static Resource API write made with no explicit context. Because the object is shared, that leftover `.transaction` reference is then visible to any *later*, logically-independent static Resource API call made in the same handler — and `Resource.ts` wrongly joins it instead of starting its own, even though the first call's transaction may already be closing or lingering. The two writes get folded into one transaction whose commit lifecycle was only ever driven by one of the calls, and the other write can be silently dropped.

## How this was found

Flagged via `harper-pro` PR #512's CI: bumping the `core` submodule pointer past this range caused `integrationTests/cluster/replicationTopology.test.mjs` (a 4-node star-topology cluster-formation test) to fail intermittently with `Timed out waiting for cluster to connect` / `SELF_SIGNED_CERT_IN_CHAIN`. Bisecting ~87 divergent `core` commits against that test as an oracle (harper-pro code held constant, only the `core` submodule pointer changed) converged cleanly on `af646222d` ("fix: establish ambient user context for operation handlers so audit records carry attribution (#1591)") as the first-bad commit.

Root-caused by instrumenting a live 2-node cluster-formation run: harper-pro's `add_node`/`add_node_back` operation handlers (`replication/subscriptionManager.ts`'s `ensureNode()`, and `security/certificate.ts`'s `signCertificate()`) each make 2+ sequential static Resource API calls with no explicit context under one ambient-`hdb_user`-carrying operation. Tracing showed two sequential `ensureNode()` calls (once for the node's own record, once for a peer's) sharing the *identical* `DatabaseTransaction` instance — something structurally impossible before #1591, since each call always got its own throwaway `{}` context. As a result, a peer's CA cert silently failed to persist into `hdb_nodes`, so a node's TLS trust store never picked it up, producing `SELF_SIGNED_CERT_IN_CHAIN` on subsequent mesh connections.

## Fix

Only treat an ambient `.transaction` as reusable when it is still `TRANSACTION_STATE.OPEN` — the exact same self-consistency check `resources/transaction.ts`'s own `transaction()` helper already applies before deciding to reuse vs. create a transaction. This restores the pre-#1591 isolation for independent calls sharing an ambient context, while leaving genuine nested/explicit transaction chaining (a caller holding the same still-open `Context`/txn across multiple calls *within* one `transaction(...)` callback) completely unaffected, since `.open` only transitions away from `OPEN` once the transaction's owning `transaction()` call has fully completed.

This does **not** regress #1591/#1592's actual audit-attribution feature — `unitTests/resources/operationContextUser.test.js` and `unitTests/server/serverHelpers/serverUtilities.test.js` still pass unmodified.

## Testing

- New regression test: `unitTests/resources/operationContextTransactionLeak.test.js` — exercises 2 and 3 sequential no-explicit-context static-API writes, plus a write following a drained `search()` iterator, all inside one ambient-user `processLocalTransaction` scope, asserting every write persists.
- Existing suites re-run clean against this fix: `unitTests/resources/**` (1029 passing), `unitTests/server/serverHelpers/**` (200 passing, including all `#1591` audit-attribution cases), `unitTests/components/mcp/audit.test.js` (13 passing), `unitTests/dataLayer/**` (155 passing).
- Full-system validation against `harper-pro`'s `integrationTests/cluster/replicationTopology.test.mjs` ("connect nodes", 4-node star topology), with `core` pinned at `af646222d`:
  - **Without this fix**: fails consistently, 0/3 runs, timing out waiting for the cluster to connect (`SELF_SIGNED_CERT_IN_CHAIN` in node logs).
  - **With this fix** (harper-pro code otherwise unmodified): passes consistently, 3/3 runs, at the same speed (~3.2s) as a pre-`af646222d` `core` baseline run on the same machine.

## Note for harper-pro

`ensureNode()`'s own writes remain attributed to whatever ambient `hdb_user` happens to be present (e.g. the calling admin, or an inter-node connection's resolved auth principal) even with this fix — no longer a data-loss bug, but arguably still slightly wrong for audit-attribution purposes, since this is internal replication/system bookkeeping, not a user-authored write. I've prepared a small, separate defense-in-depth harper-pro change (giving `ensureNode()` an explicit, isolated context) that Kris can review/merge independently; it is not required for this fix to be effective.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/unitTests/resources/operationContextTransactionLeak.test.js
+++ b/unitTests/resources/operationContextTransactionLeak.test.js
@@ -60,7 +60,10 @@ describe('Ambient operation context must not couple independent writes (transact
 
 		const first = await LeakTable.get('first-record');
 		const second = await LeakTable.get('second-record');
-		assert.ok(first, 'first-record must persist: it must not be silently dropped by a leaked/shared transaction from the ambient operation context');
+		assert.ok(
+			first,
+			'first-record must persist: it must not be silently dropped by a leaked/shared transaction from the ambient operation context'
+		);
 		assert.equal(first?.name, 'first');
 		assert.ok(second, 'second-record must persist');
 		assert.equal(second?.name, 'second');
--- a/unitTests/resources/operationContextTransactionLeak.test.js
+++ b/unitTests/resources/operationContextTransactionLeak.test.js
@@ -126,4 +126,60 @@ describe('Ambient operation context must not couple independent writes (transact
 		assert.ok(afterSearch, 'write made after draining a search() iterator must persist');
 		assert.equal(afterSearch?.name, 'after-search');
 	});
+
+	// The three tests above assert final persistence, which a reviewer (cb1kenobi, PR #1720) found
+	// does not actually discriminate buggy vs. fixed code: reverting Resource.ts's fix back to the
+	// bare `if (context?.transaction)` truthiness check and re-running them still passes 3/3, because
+	// by the time a second write joins the stale, already-committed transaction from the first call,
+	// DatabaseTransaction's addWrite()/save() takes an immediate-commit path for a closed transaction,
+	// so the write still lands. That masks the actual bug in this isolated single-process unit setup,
+	// even though the same coalescing was proven to silently drop a write in the live 4-node cluster
+	// integration test (harper-pro's replicationTopology.test.mjs) that originally caught this.
+	//
+	// This test instead asserts the mechanism directly: each independent, no-explicit-context write
+	// must be serviced by its own DatabaseTransaction instance. Resource.ts's dispatcher only ever
+	// assigns a *new* transaction onto the shared ambient context via resources/transaction.ts's
+	// transaction() helper when it takes the "start a transaction" branch; the buggy bare-truthiness
+	// check instead takes the "we are already in a transaction, proceed" branch as soon as
+	// `context.transaction` is set at all, so it never calls transaction() again and the ambient
+	// context's `.transaction` reference never changes for the lifetime of the operation handler —
+	// every subsequent write is coalesced into the exact same instance. With the fix, once the prior
+	// transaction is no longer TRANSACTION_STATE.OPEN, the dispatcher takes the other branch and a
+	// fresh DatabaseTransaction is minted. This invariant fails on the pre-fix code and holds on the
+	// fix, regardless of whether an individual isolated unit test happens to still persist the data.
+	it('services each independent no-explicit-context write with its own DatabaseTransaction instance (mechanism-level)', async () => {
+		const transactionsSeenAfterEachWrite = [];
+		await serverUtilities.processLocalTransaction(
+			{ body: { operation: 'test_registered_op', hdb_user: { username: 'internal_bookkeeping' } } },
+			async () => {
+				await LeakTable.put('mech-first', { name: 'first' });
+				transactionsSeenAfterEachWrite.push(contextStorage.getStore()?.transaction);
+
+				await LeakTable.put('mech-second', { name: 'second' });
+				transactionsSeenAfterEachWrite.push(contextStorage.getStore()?.transaction);
+
+				await LeakTable.put('mech-third', { name: 'third' });
+				transactionsSeenAfterEachWrite.push(contextStorage.getStore()?.transaction);
+
+				return { message: 'ok' };
+			}
+		);
+
+		const [afterFirst, afterSecond, afterThird] = transactionsSeenAfterEachWrite;
+		assert.ok(afterFirst, 'the first write must leave a transaction attached to the ambient context');
+		assert.ok(afterSecond, 'the second write must leave a transaction attached to the ambient context');
+		assert.ok(afterThird, 'the third write must leave a transaction attached to the ambient context');
+		assert.notStrictEqual(
+			afterFirst,
+			afterSecond,
+			'the second write must not join the first write’s transaction instance: each independent, ' +
+				'no-explicit-context write must run in its own DatabaseTransaction, not a stale one left over ' +
+				'from a prior, already-completed call on the shared ambient context'
+		);
+		assert.notStrictEqual(
+			afterSecond,
+			afterThird,
+			'the third write must not join the second write’s transaction instance, for the same reason'
+		);
+	});
 });
--- a/unitTests/resources/operationContextTransactionLeak.test.js
+++ b/unitTests/resources/operationContextTransactionLeak.test.js
@@ -1,0 +1,126 @@
+const assert = require('assert');
+const { setupTestDBPath } = require('../testUtils');
+const { table } = require('#src/resources/databases');
+const { setMainIsWorker } = require('#js/server/threads/manageThreads');
+const serverUtilities = require('#src/server/serverHelpers/serverUtilities');
+const { contextStorage } = require('#src/resources/transaction');
+const { TRANSACTION_STATE } = require('#src/resources/DatabaseTransaction');
+
+// Regression coverage for a transaction-context leak exposed by issue #1591/#1592 (ambient user
+// context for operation handlers) and confirmed against a real 2-node cluster-formation scenario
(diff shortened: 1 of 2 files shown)


unitTests/resources/operationContextTransactionLeak.test.js
```diff
@@ -0,0 +1,129 @@
...
+// resources/Resource.ts's dispatcher decides whether to join an existing transaction purely via
+// `if (context?.transaction)` truthiness, with no check that the transaction is still meaningfully
+// open for a *fresh, unrelated* write. The two writes get folded into one transaction whose commit
+// lifecycle was only ever driven by the second call, and the first call's write is silently lost.
+//
+// Before #1591, every internal, no-explicit-context static Resource API call resolved ambient
+// context via `contextStorage.getStore() ?? {}`, and the store was always empty for operation
+// handlers (nothing ever populated it), so each call always got its own fresh, throwaway `{}` and
+// could never observe another call's transaction.
+describe('Ambient operation context must not couple independent writes (transaction leak, issue #1591 side effect)', () => {
```
cb1kenobi, line 32: **Medium: this regression test passes on the pre-fix code, so it does not actually guard the fix**

I reverted `resources/Resource.ts` back to `if (context?.transaction)`, rebuilt, and re-ran this file: all three cases still pass (5/5 runs). In this isolated setup the leaked stale transaction is already CLOSED/LINGERING, and `addWrite` on an already-committed transaction takes the immediate-commit path (see `DatabaseTransaction.ts` `addWrite`/`save`), so the second write persists regardless — the data-loss condition observed in the live cluster run isn't reproduced here. A regression test that also passes on the buggy code would not have caught the original bug and won't catch a re-introduction.

Suggested fix: assert the mechanism directly rather than only final persistence — e.g. that two independent no-explicit-context writes execute in *distinct* `DatabaseTransaction` instances (capture `contextStorage.getStore().transaction` at each write and assert they differ), or that the second write does not join the first call's already-committed transaction. That invariant fails on the pre-fix code and holds with the fix.

—
_Generated by Barber AI_

kriszyp: Good catch — reproduced this exactly (reverted the `.open === TRANSACTION_STATE.OPEN` check, rebuilt, ran the file: all 3 cases still passed, matching your 5/5).

Added a fourth test that asserts the mechanism instead of final persistence: it captures `contextStorage.getStore()?.transaction` immediately after each of three independent, no-explicit-context writes in one ambient operation handler, and asserts each write got a *distinct* `DatabaseTransaction` instance (`assert.notStrictEqual`). On the buggy bare-truthiness check, `context.transaction` is only ever assigned once — the "already in a transaction, proceed" branch never re-invokes `transaction()`, so every subsequent write shares the exact same (stale, already-closed) instance. With the `.open === OPEN` fix, once the prior transaction leaves OPEN the dispatcher takes the "start a transaction" branch again and mints a fresh instance.

Verified the fail→pass loop you asked for: reverted `Resource.ts` again — new test fails (`AssertionError`: second write's transaction === first write's). Restored the fix — all 4 tests pass. Pushed as an additive commit (`49772c41c`) to this branch.

Thanks for pushing on this — the original test really was checking the wrong invariant.

kriszyp:
This fix (the `.open === TRANSACTION_STATE.OPEN` check) turned out to break a different, already-shipped invariant. Confirmed live in CI, deterministic across all 4 runtimes (Node 22/24/26 + Bun), not a flake — pushed a follow-up commit (`094704359`) on this branch.

**What broke:** #1411 made a transaction that exceeds `storage.maxTransactionOpenTime` **abort** (not commit) — `DatabaseTransaction.abortDueToTimeout()` poisons it (`timedOut = true`) and closes it, so any further write against it must fail, forcing the whole logical operation to roll back atomically. `integrationTests/resources/txn-overtime-atomicity.test.ts` guards this: write row A, await past the timeout with no intervening write, write row B — expects a non-2xx and *neither* row persisted.

Before this PR, `Resource.ts`'s dispatcher checked ambient-transaction reuse with a bare `if (context?.transaction)`. That truthiness check treated a CLOSED-by-abort transaction the same as a still-open one, so the second write attempted to join the poisoned transaction and correctly threw `transactionOpenTooLongError` via `addWrite()`'s poison check — propagating the abort.

This PR's `.open === OPEN` check fixed the real leak this PR targets (a *committed* ambient transaction silently reused for a later, independent write), but an *aborted* transaction also fails `.open === OPEN` — identically to a committed one. So the dispatcher now takes the "start a fresh transaction" branch instead of the "join it" branch for both cases. For the abort case, that's wrong: the second write silently succeeds on a brand-new transaction instead of failing, so `txn-overtime-atomicity.test.ts` started failing with `AssertionError: post-await write must not have been committed either (no partial drop)`.

**The fix:** `DatabaseTransaction.ts` already distinguishes these two "not open" cases — `timedOut` is set only by `abortDueToTimeout()` (#1411's path) and never on a normal commit. Widened the dispatcher's join condition:

```ts
if (context?.transaction?.open === TRANSACTION_STATE.OPEN || context?.transaction?.timedOut) {
```

- Still `OPEN` → join it (unchanged).
- Closed because **committed** → falls through to "start fresh" (this PR's fix, unchanged).
- Closed because **aborted/timed out** → now also joins, which makes the write throw `transactionOpenTooLongError` via `addWrite()`/`commit()`'s existing poison check, correctly propagating the abort instead of silently superseding it.

**Verified:**
- `unitTests/resources/operationContextTransactionLeak.test.js` (this PR's own regression test) — still passes. Also sanity-checked by temporarily reverting to the original bare-truthiness check and confirming that test fails again for the expected reason (proving the new check isn't accidentally passing for the wrong reason).
- `integrationTests/resources/txn-overtime-atomicity.test.ts` — now passes again.
- `unitTests/resources/**` (1031 passing) and `unitTests/server/serverHelpers/**` (200 passing, incl. the #1591/#1592 audit-attribution tests) — all green.
- Lint (`oxlint`) and format (`prettier --check`) clean on the changed file.

No behavior changes beyond widening that one condition — diff is +14/-2 in `resources/Resource.ts` (comment + the widened condition).

— KrAIs (Claude Sonnet 5)

ldt1996 (approved):
Verified against the live breakage: harper-pro main's integration suite has been red since the core sync in the #529 merge crossed af646222d, and the failing runs show exactly this signature (cluster formation dying on an hdb_nodes self-row that never landed, identity mismatch, SELF_SIGNED_CERT_IN_CHAIN churn). Traced the failing wedge-recovery tests on run 28985156996 to the same mechanism before finding this PR.

Reviewed the dispatcher change: joining only OPEN transactions matches the transaction() helper's own reuse check, and routing timedOut through the join deliberately preserves the #1411 poison-atomicity contract rather than silently landing a write on a fresh transaction. The search()-iterator LINGERING corner and the mechanism-level distinct-transaction assertion cover the cases I went looking for. CI fully green.

*Lavinia, via Claude*

Issue #1631, opened 2026-07-06
Computed @indexed attribute silently indexes undefined when its JS resolver isn't yet registered (per-worker readiness race)


kriszyp:
### Summary

A JS-callback `@computed @indexed` attribute (registered at runtime via `tables.<T>.setComputedAttribute(name, fn)` in `resources.js`) can be written and read **before its resolver is registered**, producing silently-wrong results with only a `warn`. The resolver is per-worker in-memory state (`resources/Table.ts` `userResolvers`), set when `resources.js` runs on that worker — but the table's route starts serving as soon as the schema loads, and Harper load-balances requests across many http workers (default `cpus-1`) via `SO_REUSEPORT`. There is no invariant that a write/read to such a table waits until the computed resolver is live on the serving worker.

This is the root cause of the flaky `integrationTests/apiTests/computed-indexed-properties.test.mjs` REST block (de-flaked at the test level on branch `kris/fix-computed-indexed-flake` (PR forthcoming); this issue tracks the underlying product hardening).

### Observable failure modes

Given `jsTotalPrice: Float @computed @indexed` with the resolver assigned at runtime in `resources.js`:

1. **Write-time index poisoning (permanent).** A `PUT` handled by a worker whose `resources.js` hasn't finished computes the `@indexed` value with a missing resolver → `jsTotalPrice` is indexed as `undefined`, **frozen at write time**. It never self-heals on read (msgpackr/index value is materialized at write), so `?jsTotalPrice=<v>` filters miss the record for its lifetime. Only signal is:
   > `Computed attribute "jsTotalPrice" does not have a function assigned to it. Please use setComputedAttribute(...)`

2. **On-read null on a cold worker (transient).** A `GET ...?select(jsTotalPrice)` served by a not-yet-warm worker recomputes with a missing resolver and returns `null`.

3. **Operations API / main thread never has the resolver.** `search_by_value` (and `get_attributes` generally) recompute computed fields on the **main thread**, where `resources.js` (an http-worker concern) never ran, so a JS-computed attribute always resolves to `null` there. This is why the integration test cannot assert `jsTotalPrice` via `search_by_value`.

### Reproduction

Defer the resolver registration to widen the cold window (simulating a loaded worker):

```js
// resources.js
setTimeout(() => {
  tables.Product.setComputedAttribute('jsTotalPrice', (r) => r.price + r.price * r.taxRate);
}, 4000);
```

[...]

PR #1721, opened 2026-07-08
test(integration): de-flake computed indexed properties REST block

kriszyp:
## What

De-flakes the `Computed indexed properties` REST block in `integrationTests/apiTests/computed-indexed-properties.test.mjs` (seen failing intermittently on Integration Tests 5/6, Node v26).

## Root cause

`jsTotalPrice` is a `@computed @indexed` attribute whose resolver is registered at **runtime** by the component's `resources.js` (`setComputedAttribute`) as **per-worker in-memory state** (`resources/Table.ts` `userResolvers`) — unlike the `@computed(from: …)` expression attributes, whose resolvers are set at schema-load time.

The `/Product/` route (from the `@export`'d table) starts serving as soon as the schema loads, which can be **before** `resources.js` finishes on a given http worker. Harper runs many http workers (default `cpus-1`) and load-balances via `SO_REUSEPORT`, so a request can hit a still-cold worker. Two failure modes:

- a **PUT** on a cold worker computes the `@indexed` value with a missing resolver → `jsTotalPrice` is indexed as `undefined`, **frozen at write time** (never self-heals on read), so `?jsTotalPrice=119` misses the record; and
- a **GET `?select(jsTotalPrice)`** served by a cold worker recomputes `null`.

The existing route probe only proves the table is reachable on one worker — not that the resolver is registered everywhere.

## Fix (test-only)

Add `waitForComputedResolver`, a client-side readiness gate (in the spirit of `restartHttpWorkers`) that seeds a throwaway probe record and reads the on-demand computed value in bursts — each burst opens fresh connections (`Connection: close`) so requests spread across workers via SO_REUSEPORT — until several consecutive bursts see **zero cold responses**, i.e. every worker has run `resources.js`. The suite stays **multi-worker** (this is a functional test, not a single-thread special case).

Also corrects the `search_by_value` comment: `jsTotalPrice` is `null` there because the operations API runs on the **main thread**, where `resources.js` never registered the resolver — not because of a write-time race, as the old comment claimed.

## Verification

On a 19-worker box, forcing the race by deferring the resolver registration 4s (`setTimeout(() => setComputedAttribute(...), 4000)`):

| Variant | Result |
|---|---|
| Original test (no gate) | ✗ ✗ ✗ — write+filter fails deterministically (race reproduced) |
| Gated fix | ✓ ✓ ✓ |
| Unmodified suite, default threads, no injected delay | ✓ across repeated runs |

## Cross-model review

Reviewed via the cross-model-review skill (Codex leg + Harper-domain adjudication; Gemini leg skipped — test-only change, nothing left after the integration-test strip). Findings addressed in this PR:

- **[Significant] Probe record collided with the `=119` filter assertions.** The probe previously computed to `119` (same as the real assertions filter on), so a missed cleanup could reintroduce a flake in a different test. Fixed: probe now uses `taxRate: 0` → computes to `100` (`PROBE_COMPUTED`), which can never satisfy `?jsTotalPrice=119` / `?totalPrice=119`; cleanup moved into a `finally`.
- **[Suggestion] Seed PUT now status-checked** (`.expect(204)`) so a seed failure fails fast with an accurate message instead of a misattributed 60s "resolver not live" timeout.
- **[Suggestion] `readsPerBurst` now scales with host core count** (`max(60, cpus*8)`) so a straggler worker is reliably hit even on high-core hosts.

No blockers. Underlying product hardening tracked in #1631 (out of scope here).

## Related

- Product invariant gap (silent `undefined` indexing when a JS-computed resolver isn't yet live at write time; main-thread ops-API parity): #1631

### DIFF
--- a/integrationTests/apiTests/computed-indexed-properties.test.mjs
+++ b/integrationTests/apiTests/computed-indexed-properties.test.mjs
@@ -17,6 +17,8 @@
  */
 import { suite, test, before, after } from 'node:test';
 import assert from 'node:assert';
+import { setTimeout as sleep } from 'node:timers/promises';
+import { cpus } from 'node:os';
 import request from 'supertest';
 import { startHarper, teardownHarper } from '@harperfast/integration-testing';
 import { createApiClient } from './utils/client.mjs';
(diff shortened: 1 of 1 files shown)


Ethan-Arrowood (approved):
Great root-cause writeup — much better than pinning the suite to one worker.

Issue #550, opened 2026-05-15
fix: prettier format failure in `subscriptionReplay.test.js`


heskew:
## Context

The \`Format Check\` CI job is currently failing on \`main\` with prettier flagging \`unitTests/resources/subscriptionReplay.test.js\`. Reported by the CI run on HarperFast/harper#547:

\`\`\`
[warn] unitTests/resources/subscriptionReplay.test.js
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
##[error]Process completed with exit code 1.
\`\`\`

Run: https://github.com/HarperFast/harper/actions/runs/25920310025/job/76187304210

The file was recently modified by commit \`8bbbc4c2\` "yield every 100 entries in subscription catchup" which presumably introduced the formatting drift.

## Goal

Run prettier with \`--write\` on \`unitTests/resources/subscriptionReplay.test.js\` and commit the result.

## Approach

\`\`\`
npm run format:check    # confirm the failure surfaces locally
npm run format:write -- unitTests/resources/subscriptionReplay.test.js
npm run format:check    # confirm clean
\`\`\`

If the diff has unrelated drift in other files, scope the commit to just this file and flag the rest as a follow-up.

## Success criteria

- \`npm run format:check\` passes
- The \`Format Check\` CI job passes on the resulting PR
- No functional changes to the test file (only whitespace/formatting)

## Scope

- Single-file formatting fix only
- Do NOT modify test logic or assertions
- Do NOT re-format unrelated files in the same PR

PR #1727, opened 2026-07-09
fix: exclude cluster/replication topology operations from ambient user context

kriszyp:
## Summary

Excludes cluster/replication topology operations from the #1591/#1592 ambient-user context wrap in `processLocalTransaction`, and strips any inherited ambient user for those operations.

## Root cause

`af646222d` (#1591, part of #1592) made `processLocalTransaction` run every operation handler carrying `hdb_user` inside `contextStorage.run({ ...currentStore, user: hdbUser }, handler)`, so static Resource API writes inside component-registered handlers inherit user attribution for audit records.

Cluster/replication topology operations (`add_node`, `add_node_back`, `set_node`, `update_node`, `set_node_replication`, `remove_node`, …) start **long-lived, non-awaited** replication subscription work (a peer connection's connect/handshake/reconnect lifecycle). The handler returns quickly, but that background async chain keeps running — and because it was constructed while the handler's `AsyncLocalStorage` scope was active, it keeps observing the ambient `{ user }` for its **entire life**, not just the request. That corrupts the subscription's identity and produces a WebSocket **1006** reconnect storm that never meshes.

This is distinct from #1720 (a stale ambient *transaction* leaking into a later static write within one handler invocation, fixed in `resources/Resource.ts`). Both stem from #1591's design change — long-lived ambient context per handler invocation — but at different points. This fix is orthogonal to #1720 (different file, different mechanism; verified #1720's leak test passes with this change).

## Fix

`processLocalTransaction` now skips the ambient-user wrap for a documented set of topology-management operations, and — for the nested case where an ambient context already carries a user (e.g. a `server.operation()` call from within an authenticated handler) — strips that inherited user too, preserving other ambient state and never mutating the outer context object. Their writes are internal node-registry bookkeeping (`hdb_nodes`, certificates), not user-authored data, so removing user attribution is also correct, not just safe.

## Why this direction

The architecturally cleanest fix is option (b) from the investigation — detach the long-lived work from the ambient context **where it is spawned**. That spawn point lives in **harper-pro's replication component** (`subscriptionManager`/`replicationConnection`), not core, so it can't be done in a core-only change, and the deliverable requires a self-sufficient core fix (the harper-pro integration test must pass on the core change alone). The core-only variant of (b) — clearing the ambient user *after* the handler settles — is unsafe: operation handlers can return `Readable` streams that are consumed after `processLocalTransaction` returns (`serverHandlers.js`), so stripping the user mid-stream would corrupt late/streaming authorization.

This change is the safe, self-sufficient core fix (option a): it is proven-equivalent to the investigation's validated "revert the wrap" for exactly the affected operations, while preserving #1591 attribution for all data operations.

## Where to look / open items

- **`OPERATIONS_EXCLUDED_FROM_AMBIENT_USER`** (`server/serverHelpers/serverUtilities.ts`): the set spans core's `OPERATIONS_ENUM` entries and harper-pro-registered names (`add_node_back`, `set_node`, …). This cross-layer name list is the fragile part — it must be extended if new topology-management operations that spawn long-lived background work are added. **The durable fix is to detach that background work from the ambient context at the spawn point in harper-pro's replication component; recommend a coordinated follow-up.** A prior harper-pro branch (`fix/ensurenode-ambient-context`) already isolates `ensureNode`'s write context in that layer.

## Verification

- harper-pro `integrationTests/cluster/replicationReconnect.test.mjs` against this core branch: **2/3 → 3/3**, reliably (previously-failing tests dropped from ~15–17s timeouts to ~1–2s), confirmed over two runs.
- New unit coverage in `unitTests/server/serverHelpers/serverUtilities.test.js`: each topology op runs with no ambient user; background work spawned by a topology handler does not observe the request user after the handler returns; inherited ambient user is stripped for the nested case; and a non-topology op still gets #1591 attribution (no over-exclusion).
- Existing #1591 attribution tests (`operationContextUser.test.js`, `serverUtilities.test.js`) pass; #1720's leak test passes with this change overlaid.
- Full local suites green: `unitTests/resources/**` (1027), `unitTests/components/mcp/**` (479), `unitTests/server/serverHelpers/**` (211).

Generated by Claude (Opus 4.8), reviewed cross-model by Codex (one P1 addressed: strip inherited ambient user for the nested case).

### DIFF
--- a/server/serverHelpers/serverUtilities.ts
+++ b/server/serverHelpers/serverUtilities.ts
@@ -132,10 +132,22 @@ export async function processLocalTransaction(req: OperationRequest, operationFu
 	const currentStore = contextStorage.getStore();
 	const callOperationFunction = () =>
 		operationFunctionCaller.callOperationFunctionAsAwait(operationFunction, req.body, null);
-	let data =
-		hdbUser && currentStore?.user !== hdbUser && !OPERATIONS_EXCLUDED_FROM_AMBIENT_USER.has(req.body.operation)
-			? await contextStorage.run({ ...currentStore, user: hdbUser }, callOperationFunction)
+	let data;
+	if (OPERATIONS_EXCLUDED_FROM_AMBIENT_USER.has(req.body.operation)) {
+		// Excluded operations must run with no ambient user principal. Strip any ambient user —
+		// whether it would come from this request's hdb_user or from an enclosing ambient context
+		// (e.g. a nested server.operation() call from within an authenticated handler) — so the
+		// long-lived background work they spawn never inherits a user. Other ambient state is
+		// preserved, and the outer context object is never mutated.
+		data = currentStore
+			? await contextStorage.run({ ...currentStore, user: undefined }, callOperationFunction)
 			: await callOperationFunction();
+	} else {
+		data =
+			hdbUser && currentStore?.user !== hdbUser
+				? await contextStorage.run({ ...currentStore, user: hdbUser }, callOperationFunction)
+				: await callOperationFunction();
+	}
 
 	if (typeof data !== 'object') {
 		data = { message: data };
--- a/server/serverHelpers/serverUtilities.ts
+++ b/server/serverHelpers/serverUtilities.ts
@@ -59,6 +59,34 @@ const GLOBAL_SCHEMA_UPDATE_OPERATIONS_ENUM = {
 	[terms.OPERATIONS_ENUM.DROP_SCHEMA]: true,
 };
 
+// Cluster/replication topology operations must NOT run under the request's ambient user context
+// (the #1591 attribution wrap below). Two reasons:
+//
+//   1. Their writes are internal node-registry bookkeeping (hdb_nodes, certificates), not
+//      user-authored data — they should not be stamped with a user principal.
+//   2. Critically, they start long-lived, non-awaited replication subscription work (a peer
+//      connection's connect/handshake/reconnect lifecycle) that outlives the handler. Because that
+//      work is constructed while this handler's AsyncLocalStorage scope is active, it keeps
+//      observing the ambient { user } for its ENTIRE life, not just the request — which corrupts
+//      the subscription's identity and produces a WebSocket 1006 reconnect storm that never meshes
+//      (regression from #1591/#1592; the pre-#1591 world gave this background work no ambient user).
+//
+// The durable fix is to detach that background work from the ambient context where it is spawned
+// (harper-pro's replication component); this core-side exclusion resolves the regression for the
+// operations that trigger it without that background work ever inheriting a user. The names span
+// core and harper-pro's replication component (registered via server.registerOperation), so this
+// set must be kept in sync if new topology-management operations are added.
+const OPERATIONS_EXCLUDED_FROM_AMBIENT_USER = new Set<string>([
+	terms.OPERATIONS_ENUM.ADD_NODE, // 'add_node'
+	terms.OPERATIONS_ENUM.UPDATE_NODE, // 'update_node'
+	terms.OPERATIONS_ENUM.SET_NODE_REPLICATION, // 'set_node_replication'
+	'add_node_back',
+	'set_node',
+	'remove_node',
+	'remove_node_back',
+	'configure_cluster',
+]);
+
 import { OperationFunctionObject } from './OperationFunctionObject.ts';
 
 type ValueOf<T> = T[keyof T];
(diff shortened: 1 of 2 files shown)


server/serverHelpers/serverUtilities.ts
```diff
@@ -59,6 +59,34 @@ const GLOBAL_SCHEMA_UPDATE_OPERATIONS_ENUM = {
...
+// core and harper-pro's replication component (registered via server.registerOperation), so this
+// set must be kept in sync if new topology-management operations are added.
+const OPERATIONS_EXCLUDED_FROM_AMBIENT_USER = new Set<string>([
+	terms.OPERATIONS_ENUM.ADD_NODE, // 'add_node'
+	terms.OPERATIONS_ENUM.UPDATE_NODE, // 'update_node'
+	terms.OPERATIONS_ENUM.SET_NODE_REPLICATION, // 'set_node_replication'
+	'add_node_back',
+	'set_node',
+	'remove_node',
+	'remove_node_back',
```
cb1kenobi, line 86: **Nit: `remove_node_back` won't match the op as registered in harper-pro**

harper-pro registers this operation with a trailing-semicolon typo in the name — `name: 'remove_node_back;'` (`replication/setNode.ts:331`). So this exclusion entry (`remove_node_back`, no semicolon) never matches `req.body.operation` for that handler, and the entry is currently a no-op.

Low impact in practice: `removeNodeBack` only deletes an `hdb_nodes` record and spawns no long-lived subscription work, so no 1006 storm results — the exclusion here is attribution-only. But it's a live instance of exactly the cross-layer name-drift this PR's own comment warns about. The real fix is dropping the stray `;` in harper-pro's registration; keep this list entry as-is (it's the correct intended name).

—
_Generated by Barber AI_

ldt1996 (approved):
Reviewed the mechanism and the tests hold it firmly: the ALS registration-time binding for detached background work is pinned directly (the setTimeout-after-return test), the nested strip preserves the rest of the ambient state and never mutates the outer object, and the over-exclusion guard keeps #1591 attribution for data operations. The user: undefined spread is the right call over deleting the key, and running topology writes unattributed is correct on its own merits, matching the #550 companion's rationale.

Two field notes from this week that support and contextualize this: our stage cluster's rolling-restart churn showed transient 1008 "Unauthorized database subscription" flaps and identity oddities consistent with a subscription chain observing the wrong ambient identity, so the long-lived-inheritance mechanism is not hypothetical. And as an empirical cross-check of the "distinct from #1720" claim: harper-pro #555 (core pinned to #1720 alone, without this PR) just ran the previously-red integration suite 12/12 green, so #1720 was sufficient for the current main breakage and this correctly targets the residual storm class rather than duplicating that fix.

One non-blocking thought for a follow-up: the op set is name-synced by hand across repos; a self-declared flag at server.registerOperation time would let harper-pro's topology ops opt out without core knowing their names. Fits the "durable fix lives at the spawn point" direction the description already lays out.

*Lavinia, via Claude*

kriszyp:
Closing in favor of [harper#1720](https://github.com/HarperFast/harper/pull/1720), which is already merged and is the actual root-cause fix.

Root-caused this further: the WS 1006 reconnect storm isn't about ambient *user* context — it's stale-transaction reuse. `add_node` calls `ensureNode()` twice (self-row, then peer-row), both via `table.patch()` with no explicit context, sharing the single long-lived ambient context that #1591/#1592 installs. `Resource.ts`'s `transactional()` joined any `context.transaction` without checking it was still OPEN, so the second `patch()` silently reused the first's already-committed transaction and dropped the first write (the self-row). That produced the "identity mismatch" / missing peer CA / `SELF_SIGNED_CERT_IN_CHAIN` chain that manifests as the reconnect storm.

#1720's OPEN-guard in `transactional()` fixes this generally, for any handler doing sequential independent writes under one ambient context — not just the topology-op list this PR would have hardcoded. Verified with a single-variable test: buggy core + only the #1720 guard = `replicationReconnect.test.mjs` 3/3 pass with zero identity/cert errors; same core without it = 2/3 fail.

On harper-pro's side, [harper-pro#555](https://github.com/HarperFast/harper-pro/pull/555) bumps the core submodule pointer past #1720 and is what actually delivers the fix there (green, auto-merge enabled, just needs review). [harper-pro#550](https://github.com/HarperFast/harper-pro/pull/550) adds an isolated write context for `ensureNode` as defense-in-depth on top of that.

Thanks for the review here — the nested `server.operation()` catch was good, but with #1720 that surface no longer needs a fix of its own.

— KrAIs

Issue #1421, opened 2026-06-20
Custom resource `get()` / constructor error messages leak in HTTP 500 response body


kriszyp:
## Summary

When a custom resource's `get()` method or `constructor()` throws, the error's `.message` string appears verbatim in the JSON HTTP 500 response body's `title` field.

## Symptom

```js
class Secrets extends tables.Secrets {
  get() {
    throw new Error('DB_URL=postgres://user:s3cr3t@internal-db.corp:5432/prod')
  }
}
```

Response body (HTTP 500):
```json
{
  "type": "error:Error",
  "code": "Error",
  "title": "DB_URL=postgres://user:s3cr3t@internal-db.corp:5432/prod"
}
```

The full `.message` is returned to the caller. The `code` and `type` fields also expose the error class name.

## Additional behaviors observed

- Sync throw and rejected Promise are serialized identically (both leak the message).
- `constructor()` throw leaks the same way; the server survives and serves subsequent requests normally.
- Non-Error throws (`throw 'some string'`) surface the string value in `title`.
- `throw new Response(...)` is not recognized as a short-circuit idiom — produces `title: "[object Response]"` (no leak, but also not useful).
- No unhandled-promise-rejections observed — all paths are caught; the leak is only in the serialization.

## Impact

In production, application-layer errors can expose internal implementation details through the API: database connection strings, internal hostnames, secret values embedded in error messages, library internals. The HTTP 500 status code is correct; the information leakage is in the body.

## Recommendation

Scrub `error.message` from the response body in non-development environments; substitute a generic message (e.g. `"Internal server error"`) and log the full error internally. Consider an opt-in `ClientError` subclass whose `.message` is intentionally user-visible, so developers can explicitly choose when to expose error text to callers.

## Environment

Harper: `7aaa5a152` (byte-identical to `main` @ `6797f091d`, reproduces on main).
Repro: `npm run test:integration -- "integrationTests/qa-scratch/qa169-custom-resource-errors.test.ts"` (cells S2, S3, S4)

---
*Surfaced by QA-explorer (QA-169). Filed by Claude (Sonnet 4.6) for @kriszyp.*

PR #1737, opened 2026-07-09
fix: don't log raw Error objects in REST.ts (leaks secrets to hdb.log) (#1734)

### LINKED ISSUE #1734: server/REST.ts logs the raw Error object to hdb.log — leaks secrets/sensitive data attached to escaping Errors
## Summary

`server/REST.ts`'s catch-all error handler logs the raw `Error` **object** to `hdb.log` instead of `error.message`. Node's default Error formatting dumps every own-enumerable property when an Error is logged this way — so any custom property that ends up attached to an escaping Error lands verbatim in the log line, in plaintext, at Harper's **default** log level (not just `trace`), retrievable via the ops-API `read_log` operation.

This is a real leak vector for secrets: a component using the new `hdb_secret` store (#715) to authenticate an outbound `sourcedFrom` fetch, where the origin then fails, can have the secret land in `hdb.log`. It's also broader than secrets — any third-party HTTP client library that attaches request/config context to its own thrown errors (axios does this routinely, headers included) is exposed the same way.

## Root cause

`server/REST.ts` ~line 271-287, the request-handler catch-all:

```ts
if (statusCode) {
	if (statusCode === 500) harperLogger.warn(error);
	else harperLogger.info(error);
	...
} else harperLogger.error(error);
```

All three call sites pass the raw `Error` object rather than `error.message`. `JSON.stringify`/Node's default formatter for a logged Error dumps every own-enumerable property, so anything an app (or a library) stashed on the Error — a secret used for an outbound `Authorization` header, an axios `config`/`request` object — is written to the log verbatim.

## What's clean (for context)

The client-facing HTTP response is **not** affected — a few lines later, the RFC 9457 Problem Details construction only pulls `message`/`code`/`status`/`detail`, properly minimized. This is a **server-log-only** leak, distinct from #1421 (which was `error.message` leaking into the client response body's `title` field — already fixed there).

Confirmed via a probe combining `hdb_secret` + `sourcedFrom` across 4 origin-failure modes (500, malformed JSON, ECONNREFUSED, hang): the secret leaked into `hdb.log` in every case, while the client response and the cached/stored record both stayed clean.

## Suggested fix

Log `error.message` (or a redacted/allowlisted view) instead of the raw `Error` object in the catch-all, or scrub known-sensitive property names before logging.

## Repro

Local repro (not yet in the tracked suite): `integrationTests/qa-scratch/qa523-secret-sourcedfrom-leak.test.ts` — a `sourcedFrom` resource decrypts a real `hdb_secret` value and uses it in an outbound `Authorization: Bearer <secret>` header; the mock origin fails (500 / malformed JSON / ECONNREFUSED); the secret is found verbatim in `hdb.log`.

kriszyp:
## Summary

Fixes #1734. `server/REST.ts`'s catch-all and WebSocket handlers passed the raw `Error` **object** to the logger. `HarperLogger extends Console`, so the Error is formatted with `util.inspect`, which dumps every own-enumerable property after the stack. Anything an app or an HTTP client library stashes on a thrown `Error` — a credential used for an outbound `Authorization` header, an axios `config`/`request` with headers — lands verbatim in `hdb.log` at the **default** log level, retrievable via the ops-API `read_log` operation.

## Fix

Add `harperLogger.errorForLog(error)`:

```ts
export function errorForLog(error: any) {
	return typeof error?.stack === 'string' ? error.stack : errorToString(error);
}
```

`error.stack` is the class name + message + frames — it does **not** contain custom own-enumerable properties — so logging the stack preserves debuggability while dropping the leak vector. Stackless/plain thrown values fall back to `errorToString`'s `"ClassName: message"`.

All six REST/WS log sites (HTTP catch-all `warn`/`info`/`error`, WS `error` handler, WS catch `warn`/`info`/`error`) now route through it. The client-response path (`problemDetail`) is untouched — it already minimizes to `message`/`code`/`status`/`detail`, so this is a **server-log-only** fix, distinct from #1421.

## Design note

I log the **stack** rather than the issue's suggested `error.message`: stack is strictly safer than the current raw-object behavior yet keeps the frames operators rely on for 500s (message-only would regress debuggability). Message is still present — it's inside the stack — consistent with the already-accepted client-response `title`. This is a name-agnostic sanitizer, not an allowlist scrub, so it can't miss a "new sensitive property name."

## Tests

4 new unit tests in `harper_logger.test.js` covering `errorForLog`, including an assertion that a `Bearer` token stashed on `error.authorization`/`error.config` does **not** appear in the output. `tsc`/`oxlint`/`prettier` clean.

### DIFF
--- a/server/REST.ts
+++ b/server/REST.ts
@@ -13,7 +13,7 @@ import { Request } from '../server/serverHelpers/Request.ts';
 import { RequestTarget } from '../resources/RequestTarget';
 import { entryMap } from '../resources/RecordEncoder.ts';
 
-const { errorToString } = harperLogger;
+const { errorToString, errorForLog } = harperLogger;
 const etagBytes = new Uint8Array(8);
 const etagFloat = new Float64Array(etagBytes.buffer, 0, 1);
 let httpOptions = {};
@@ -275,8 +275,8 @@ async function http(request: Request, nextHandler) {
 		// the HTTP status may be carried as `statusCode` (our error classes) or `status` (e.g. a thrown plain object)
 		let statusCode = error.statusCode ?? error.status ?? request.response.status;
 		if (statusCode) {
-			if (statusCode === 500) harperLogger.warn(error);
-			else harperLogger.info(error);
+			if (statusCode === 500) harperLogger.warn(errorForLog(error));
+			else harperLogger.info(errorForLog(error));
 			if (statusCode === 405) {
 				if (error.method) error.message += ` to handle HTTP method ${error.method.toUpperCase() || ''}`;
 				if (error.allow) {
@@ -284,7 +284,7 @@ async function http(request: Request, nextHandler) {
 					headers.setIfNone('Allow', error.allow.map((method) => method.toUpperCase()).join(', '));
 				}
 			}
-		} else harperLogger.error(error);
+		} else harperLogger.error(errorForLog(error));
 
 		// RFC 9457 Problem Details
 		const status = statusCode || 500;
@@ -351,7 +351,7 @@ export function handleApplication(scope: import('../components/Scope.ts').Scope)
 			let hasError;
 			(ws as any).on('error', (error) => {
 				hasError = true;
-				harperLogger.warn(error);
+				harperLogger.warn(errorForLog(error));
 			});
 			let deserializer;
 			(ws as any).on('message', function message(body) {
@@ -414,9 +414,9 @@ export function handleApplication(scope: import('../components/Scope.ts').Scope)
 				}
 			} catch (error) {
 				if (error.statusCode) {
-					if (error.statusCode === 500) harperLogger.warn(error);
-					else harperLogger.info(error);
-				} else harperLogger.error(error);
+					if (error.statusCode === 500) harperLogger.warn(errorForLog(error));
+					else harperLogger.info(errorForLog(error));
+				} else harperLogger.error(errorForLog(error));
 				ws.close(
 					HTTP_TO_WEBSOCKET_CLOSE_CODES[error.statusCode] || // try to return a helpful code
 						1011, // otherwise generic internal error
--- a/utility/logging/harper_logger.ts
+++ b/utility/logging/harper_logger.ts
@@ -12,6 +12,7 @@ import * as os from 'os';
 import { PACKAGE_ROOT } from '../../utility/packageUtils.js';
 import { _assignPackageExport } from '../../globals.js';
 import { Console } from 'console';
+import { inspect } from 'util';
 // store the native write function so we can call it after we write to the log file (and store it on process.stdout
 // because unit tests will create multiple instances of this module)
 let nativeStdWrite = process.env.IS_SCRIPTED_SERVICE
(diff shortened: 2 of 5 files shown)


kriszyp:
## Cross-model review (standard: Gemini leg + Harper-domain adjudication)

No blockers. One significant concern addressed at the root, the rest triaged:

**Addressed** (2nd commit) — `error.stack` alone dropped the `error.cause` chain, which the old raw-object logging surfaced via `util.inspect`. Since Harper convention leans on `new Error(msg, { cause })`, that's a real diagnostic regression. `errorForLog` now returns a lazy `util.inspect.custom` wrapper that:
- appends each `cause`'s stack (cycle-protected), still excluding own-enumerable props; and
- defers stack materialization until the logger's level gate actually writes — which also resolves Gemini's eager-`.stack`-on-the-discarded-path perf note in the same change.

Verified end-to-end through the real `HarperLogger`: a `Bearer` token on both `error.authorization` and the cause's `config.headers.Authorization` stays out of the rendered line, while message + stack + `caused by:` root-cause stack are all present. New unit tests cover the cause chain, the cycle guard, and the leak assertion (6 total).

**Accepted / not changed:**
- *Plain thrown object with no stack/message loses its fields* (`throw { status, details }` → `errorToString` → `"[object Object]"`). Left as-is deliberately — rendering arbitrary object keys is the exact leak surface this fixes; Harper throws real `Error`s in practice. Minor.
- *`errorForLog(null)` would throw in the fallback* — not reachable: every callsite dereferences `error.statusCode`/`error.status` first, so `throw null` already throws upstream of the logger.

Caveat: Codex leg didn't run (standard mode = Gemini leg only); domain pass + Gemini covered the logging/secret-leak surface.

server/REST.ts
```diff
@@ -275,16 +275,16 @@ async function http(request: Request, nextHandler) {
 		// the HTTP status may be carried as `statusCode` (our error classes) or `status` (e.g. a thrown plain object)
 		let statusCode = error.statusCode ?? error.status ?? request.response.status;
 		if (statusCode) {
-			if (statusCode === 500) harperLogger.warn(error);
-			else harperLogger.info(error);
+			if (statusCode === 500) harperLogger.warn(errorForLog(error));
```
cb1kenobi, line 278: **Medium: same raw-Error log leak remains in sibling HTTP handlers**

This correctly routes REST.ts through `errorForLog`, but the identical catch-all pattern — `if (statusCode === 500) harperLogger.warn(error); else harperLogger.info(error); ... else harperLogger.error(error);` — still logs the raw `Error` object in:

- `server/http.ts:513-515` (Node request handler `onError`)
- `server/http.ts:755-757` (Bun request handler catch)
- `server/graphqlQuerying.ts:585` (`logger.error(error)` in the GraphQL HTTP catch-all)

These are the same HTTP request error paths, so the exact #1734 vector (a secret/axios `config` stashed on an escaping Error landing in `hdb.log` at default level) is still open there. `http.ts` already imports `harperLogger` and destructures `errorToString`, so the fix is the same one-line-per-site change.

Suggested fix: add `errorForLog` to the destructure in `http.ts`/`graphqlQuerying.ts` and wrap those six calls the same way.

—
_Generated by Barber AI_

utility/logging/harper_logger.ts
```diff
@@ -895,6 +897,43 @@ export function errorToString(error: any) {
 	return typeof error.message === 'string' ? `${error.constructor.name}: ${error.message}` : error.toString();
 }
 
+/**
+ * Renders an error to its stack (class name + message + frames), then appends the stack of each
+ * error in its `cause` chain. Deliberately excludes own-enumerable properties — those are exactly
+ * what leaks secrets in #1734 (see `errorForLog`).
+ */
+function errorToLogString(error: any) {
+	let output = typeof error?.stack === 'string' ? error.stack : errorToString(error);
```
cb1kenobi: **Nit: `errorForLog(null)`/`undefined` throws inside the logger**

When a `null`/`undefined` value is thrown, `error?.stack` is undefined so this falls to `errorToString(error)`, which dereferences `error.message` unguarded (line 897) and throws a `TypeError` — inside `util.inspect`, i.e. inside the logging call within a catch handler. Not reachable from this PR's call sites (they deref `error.statusCode` first), but `errorForLog` is a newly exported general helper whose job is log safety, so a thrown nullish value shouldn't be able to crash it.

Suggested fix:
```suggestion
	let output = typeof error?.stack === 'string' ? error.stack : error == null ? String(error) : errorToString(error);
```

—
_Generated by Barber AI_

server/REST.ts
kriszyp: Good catch. Routed all three sibling catch-alls through errorForLog in 097b6203c: server/http.ts Node onError (3 calls) + Bun fetch catch (3 calls), and server/graphqlQuerying.ts GraphQL HTTP catch-all. Added errorForLog to the http.ts destructure. — KrAIs

utility/logging/harper_logger.ts
kriszyp: Applied in 097b6203c: errorToLogString now renders String(error) for a nullish value instead of dereferencing error.message, so a thrown null/undefined can't crash the logger. Added unit coverage for null/undefined. — KrAIs

utility/logging/harper_logger.ts
```diff
@@ -12,6 +12,7 @@ import * as os from 'os';
 import { PACKAGE_ROOT } from '../../utility/packageUtils.js';
 import { _assignPackageExport } from '../../globals.js';
 import { Console } from 'console';
+import { inspect } from 'util';
```
cb1kenobi, line 15: Scope module for consistency.
```suggestion
import { inspect } from 'node:util';
```

utility/logging/harper_logger.ts
```diff
@@ -895,6 +897,43 @@ export function errorToString(error: any) {
 	return typeof error.message === 'string' ? `${error.constructor.name}: ${error.message}` : error.toString();
 }
 
+/**
+ * Renders an error to its stack (class name + message + frames), then appends the stack of each
+ * error in its `cause` chain. Deliberately excludes own-enumerable properties — those are exactly
+ * what leaks secrets in #1734 (see `errorForLog`).
+ */
+function errorToLogString(error: any) {
+	let output = typeof error?.stack === 'string' ? error.stack : error == null ? String(error) : errorToString(error);
```
cb1kenobi, line 906: I wonder if the `error == null` check should be moved to `errorToString()`? Could that be beneficial elsewhere?

cb1kenobi (approved):

PR #1738, opened 2026-07-09
Add tsgo as an opt-in fast type-checker

kriszyp:
## Summary
Adds `@typescript/native-preview` (ships `tsgo`, the native/Go-ported preview compiler that becomes TypeScript 7's `tsc`) as a devDependency and wires it in as a new `npm run typecheck:fast` script.

## Purpose
Benchmarked on this exact codebase: `tsgo --noEmit` is **~6.3x faster** than `tsc --noEmit` (3.01s → 0.48s) with the same result set, and the emit path is ~6.7x faster too. This gives a much faster local/CI type-check loop as an *opt-in* parallel tool — it does not replace or touch the committed `build` script (`tsc --project tsconfig.build.json`), and does not touch ESLint (`@typescript-eslint/parser` still needs `typescript` 5.x, since TypeScript 7.0 ships no compiler API — 7.1 will).

Also pins `"strict": false` explicitly in `tsconfig.json`. The repo never set `strict`, relying on `tsc`'s implicit default (`false`); TypeScript 7 flips that default to `true`. Verified `tsc --noEmit` output is byte-identical before/after this pin — it documents the status quo rather than changing behavior, and forward-guards against a future compiler bump silently changing type-check results.

## Where to focus review attention
- **`tsgo` is pinned to an exact nightly `-dev` build** (`7.0.0-dev.20260707.2`), not a caret range. Nightly preview tags on the npm registry aren't guaranteed to persist indefinitely — if the tarball is ever pruned, a clean-cache `npm ci` would fail until the pin is bumped. Low impact (dev-only, opt-in, not a CI gate) but worth knowing about.
- The `dependencies.md` entry notes `tsgo` ships a prebuilt native (Go-compiled) binary via platform-gated `optionalDependencies` — no local compilation, but it's not a pure-JS package either.
- 4 known type-check findings surfaced by `tsgo` today (none introduced by this PR, none blocking): a module-export resolution gap on `server/threads/manageThreads.js`'s dynamic `module.exports.X = ...` pattern (2 sites — already fixed as a side effect of #562's CJS→ESM conversion, once that lands), plus two independent findings (`security/auth.ts`, `server/REST.ts`'s `Buffer`/`BufferSource` typing) worth a separate look but out of scope here.

Generated by Claude (Opus 4.8).


package.json
```diff
@@ -138,6 +139,7 @@
 		"@types/sinon": "^21.0.0",
 		"@types/tar-fs": "^2.0.4",
 		"@typescript-eslint/parser": "^8.48.1",
+		"@typescript/native-preview": "7.0.0-dev.20260707.2",
```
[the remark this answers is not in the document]
kriszyp: Checked this before applying: `@typescript/native-preview@7.0.0-beta` isn't actually a published version — `npm view @typescript/native-preview@7.0.0-beta` 404s. The package's `beta` dist-tag also doesn't point to anything more stable; it resolves to `7.0.0-dev.20260421.2`, an *older* nightly (April) than what's pinned here (`7.0.0-dev.20260707.2`, July) — so following the suggestion literally would both break install and not reduce the nightly-pruning risk being flagged. That risk itself is real and already called out explicitly in the PR description and this repo's `dependencies.md` entry as an accepted tradeoff (dev-only, opt-in, not a CI gate) until this package ships an actual stable/non-prerelease release. Leaving the pin as-is.

package.json
```diff
@@ -138,6 +140,7 @@
...
```
cb1kenobi: **Low: nightly `-dev` pin couples every `npm ci` job to a transient tarball**

Same structural issue as harper-pro#556: as a non-optional `devDependency`, `npm ci` fetches this exact nightly (`7.0.0-dev.20260707.2`) for every job that installs devDependencies (unit, integration, smoke, stress), even though `typecheck:fast` is opt-in. If npm ever prunes that nightly, all of those jobs fail hard, not just the checker.

The `dependencies.md` Note already weighs `optionalDependencies` (correctly rejected for a public package) and a looser tag, but not on-demand execution. Since usage is already opt-in, `typecheck:fast` could instead run `npx -y @typescript/native-preview@7.0.0-dev.20260707.2 tsgo --noEmit --project tsconfig.json` and drop the dependency (and its 142 lockfile lines) entirely — a pruned tarball would then only break `typecheck:fast`, not unrelated CI. Trade-off is losing lockfile-pinned reproducibility for the tool. Low either way; noting the option since the analysis didn't cover it.

—
_Generated by Barber AI_

kriszyp: Same fix as harper-pro#556 (same finding there from you). Applied in 6d799b945: removed `@typescript/native-preview` as a devDependency entirely and switched `typecheck:fast` to `npx -y -p @typescript/native-preview@7.0.0-dev.20260707.2 tsgo --noEmit --project tsconfig.json`. Verified the npx invocation resolves correctly and the script still runs clean (1.6s). Updated dependencies.md to reflect it's no longer a package.json dependency at all. A pruned tarball now only breaks `typecheck:fast`, not `npm ci` for any other job.

package.json
```diff
@@ -53,6 +53,8 @@
 	"scripts": {
 		"build": "tsc --project tsconfig.build.json",
 		"build:watch": "npm run build -- --watch --incremental",
+		"typecheck": "tsc --project tsconfig.json",
+		"typecheck:fast": "npx -y -p @typescript/native-preview@7.0.0-dev.20260707.2 tsgo --noEmit --project tsconfig.json",
```
cb1kenobi: Is there a reason we have to use `@typescript/native-preview` instead of `typescript@7.0.2`? Will it collide with the installed `typescript` version?

cb1kenobi (approved):

package.json
kriszyp: Good catch — switched to `typescript@7.0.2` in a9739b4. TypeScript 7 merged the native (Go-ported) compiler preview directly into `typescript`'s own `tsc` binary, so `@typescript/native-preview`/`tsgo` is no longer a separate thing as of 7.0.2 — no need to carry the preview package name at all.

On collision: no. `typecheck:fast` still invokes it via `npx -y -p typescript@7.0.2 tsc ...` rather than as a `package.json` dependency, precisely because `typescript` is already a 5.x devDependency here (needed by `@typescript-eslint/parser`, since TS 7.0 has no compiler API until 7.1). Since it's the same package name, it couldn't be added a second time at a different version in `package.json` anyway. `npx -p typescript@7.0.2` resolves and runs the pinned 7.x tarball from npm's npx cache, without touching the installed 5.x copy in `node_modules` — verified locally: `npm run typecheck:fast` runs the 7.0.2 native `tsc` and produces the same 4 known findings as before, while the 5.x `typescript` devDependency and `build`/`typecheck` scripts (which use it) are untouched.

Issue #1742, opened 2026-07-09
Follow-up to #1736: streaming results and isJob for component-registered operations


kriszyp:
## Context

#1736 (fixed by the cross-thread bridge PR) made `server.registerOperation()` from per-worker component contexts reachable via the main-thread ops API by forwarding the request to one registering worker over ITC, with the result returned via structured clone.

Two capabilities were deliberately scoped out of that fix:

## 1. Streaming results

A forwarded operation that returns a `Readable` currently gets an explicit `501`:

> Operation '<name>' returned a stream; streaming results are not supported for operations registered from a component

Supporting this needs MessagePort transfer plumbing (transfer a port with the request, pipe the stream across it, re-wrap as a `Readable` on the main thread with backpressure and error propagation).

## 2. `isJob`

`OperationDefinition.isJob` was already ignored by `registerOperation()` before #1736 — it only ever wires `execute` as the regular `operation_function`, never a `job_operation_function`. If we want component-registered ops runnable as jobs, job worker threads would also need to load the registering component (they currently load only built-in env components), or the job dispatch path needs the same forwarding treatment.

— KrAIs 🤖 (on Kris's behalf)

PR #1743, opened 2026-07-09
fix: make server.registerOperation() from components reachable via the ops API (#1736)

### LINKED ISSUE #1736: server.registerOperation() calls in a component's resources.js are unreachable via the ops API
## Summary

`server.registerOperation()` calls inside a component's `resources.js` — the standard, documented pattern for adding a custom ops-API operation — are permanently unreachable. Every caller, including `super_user`, gets `400 "Operation '<name>' not found"`.

This surfaced while re-testing #1591's audit-attribution fix. That fix (`af646222d`) genuinely works — confirmed via two positive controls: a main-thread-registered op correctly attributes `user_name`, and a custom-Resource bare `put()` attributes correctly. But the literal mechanism #1591 named — `server.registerOperation()` at a component's `resources.js` top level — can never actually be observed end-to-end through that path, because the operation is never reachable in the first place.

## Root cause

The main thread requires `server/operationsServer` (the ops-API HTTP dispatcher), while worker threads separately require `server/serverHelpers/serverUtilities`. Node `worker_threads` don't share module state, so each worker gets its own `OPERATION_FUNCTION_MAP` instance:

```ts
// server/serverHelpers/serverUtilities.ts
export const OPERATION_FUNCTION_MAP = initializeOperationFunctionMap();
...
export function getOperationFunction(json) {
	if (OPERATION_FUNCTION_MAP.has(json.operation)) {
		return OPERATION_FUNCTION_MAP.get(json.operation);
	}
	// throws "Operation '<name>' not found" otherwise
```

jsResource components — where `registerOperation()` naturally lives — load **per-worker** (confirmed via #1448's own commit message: "each worker registers its own JS resources... not over ITC"). So a registration made from a component's `resources.js` lands in a worker-local `OPERATION_FUNCTION_MAP` that the main-thread ops-API dispatcher (`getOperationFunction`) never reads.

## Not the same as #1448/#1562

An earlier QA pass attributed this to "the same family" as #1448/#1562 (both closed). Diffing their actual commits shows they touch only `components/mcp/*`, `resources/jsResource.ts`, and `server/itc/serverHandlers.js` — never `serverUtilities.ts` or the main-thread/worker-thread split. Those fixed MCP tool/prompt-listing cross-worker surfacing, a related but genuinely distinct registry from general ops-API operation dispatch. This is a fresh, distinct root cause.

## Impact

Not a security bug — it fails closed, nothing leaks. But it's a broken, documented capability: a component author following the standard pattern gets a custom operation that's silently unreachable via the ops API for every caller.

## Repro

Local repro (not yet in the tracked suite): `integrationTests/qa-scratch/qa482-audit-attribution.test.ts` (the "SECONDARY" test) and `integrationTests/qa-scratch/qa451-registered-op-authz.test.ts` (pre-existing, same symptom).

kriszyp:
Fixes #1736.

## Problem

`server.registerOperation()` from a component's `resources.js` (or Plugin API `handleApplication`) — the standard, documented pattern — was permanently unreachable via the ops API. Components load per worker thread, so the registration lands in that worker's module-local `OPERATION_FUNCTION_MAP`; the ops-API HTTP dispatcher runs only on the main thread and reads the main thread's own instance of that map. Every caller, including `super_user`, got `400 "Operation '<name>' not found"`. Only the deprecated `startOnMainThread` extension hook worked.

## Fix: a cross-thread bridge (`server/serverHelpers/registeredOperations.ts`)

Mirrors the existing `RESOURCE_OPENAPI` ITC request/response pattern, with one deliberate difference: executing an operation is side-effecting, so a request is forwarded to exactly **one** registering worker — never broadcast-first-wins.

- **Announce:** a worker-side `registerOperation()` still sets its local map, then broadcasts `OPERATION_REGISTERED`; the main thread records `name → Set<threadId>`.
- **Forward on miss:** when main-thread dispatch misses its local map, it checks the registry and forwards the request body (`OPERATION_EXECUTE_REQUEST`) to one live registering worker — rotating across workers, pruning dead ids — and awaits the `requestId`-correlated response. Unknown names still fail fast with `400` (the registry gates forwarding, so no timeout is paid for genuinely unknown operations).
- **Auth runs on the worker:** the main thread authenticates and forwards the resolved `hdb_user`; the worker executes through the normal `chooseOperation` (full `verifyPerms`, where the operation function and its metadata actually exist) + `processLocalTransaction`. Workers never re-forward, so an unknown op can't loop. `bypass_auth` is re-stripped on the worker as defense-in-depth, and response correlation verifies the responding thread is the one the request was sent to.
- **Failure handling:** worker death mid-flight rejects in-flight forwards with `503` and forgets that worker's registrations (`onThreadExit` hook added to `manageThreads`); a wedged worker is bounded by the ops-API network timeout; structured-clone failures report a clear error in both directions instead of an opaque hang.

This fixes both component contexts named in the issue investigation (jsResource **and** `handleApplication`) at the `registerOperation` layer, and as a side effect lets `server.operation()` on the main thread reach component-registered operations too. Worker-local calls (`server.operation()` in a worker) are untouched — they hit the local map as before.

## Explicitly out of scope (follow-up: #1742)

- **Streaming results:** a forwarded op returning a `Readable` gets an explicit `501` rather than an opaque failure (needs MessagePort transfer plumbing).
- **`isJob`:** was already ignored by `registerOperation()` before this change (never wired to `job_operation_function`); unchanged.

## Testing

- New integration suite `integrationTests/components/registered-operation.test.ts` (fixture registers ops from `resources.js`, 2 worker threads): reachable via ops API with worker-side execution and forwarded user identity; repeated calls; error `statusCode` propagation (422); stream → explicit 501; unknown op → fast 400. **5/5 pass.**
- Regression smoke: `authentication`, `delete`, `job-queue` (97/97), `components.test.mjs` (25/25).
- Build, oxlint, prettier clean.

## Cross-model review (thorough: Gemini + Codex + domain pass)

No confirmed blockers. Adjudicated notes for reviewers:

- **Rolling-restart window (significant, accepted):** a draining worker stays in the registry until its port closes, so a fresh forwarded op can land on a worker that's shutting down. If the worker exits mid-execution the forward rejects with `503` rather than hanging — the residual exposure (op partially executes, then 503) is inherent to any worker-crash window, not new to this path.
- **Post-deploy registration race (accepted):** the announce is fire-and-forget, so a call arriving in the instant between component load and the main thread recording the announcement gets a transient `not found` — eventual-consistency window matching component load semantics generally.
- Response correlation now includes an originator check (defense-in-depth, adopted from review).

— KrAIs (Claude Opus 4.8), for Kris

### DIFF
--- a/server/serverHelpers/registeredOperations.ts
+++ b/server/serverHelpers/registeredOperations.ts
@@ -154,32 +154,33 @@ function executeRemoteOperation(name: string, body: any): Promise<any> {
 				message: { requestId, body: forwardBody, originator: threadId },
 			});
 		} catch (error) {
-			// Structured-clone failure (e.g. a function or native handle on the request body) —
-			// no other worker will fare better with the same body.
+			// Structured-clone failure (e.g. a function or native handle on the request body) is a
+			// server-side limitation, not a malformed client request — 500, not 400. No other worker
+			// will fare better with the same body.
 			operationLog.error(`Failed to forward operation '${name}' to worker thread ${targetThreadId}`, error);
 			return Promise.reject(
-				new ServerError(`Operation '${name}' request could not be forwarded to a worker thread: ${error.message}`, 400)
+				new ServerError(`Operation '${name}' request could not be forwarded to a worker thread: ${error.message}`, 500)
 			);
 		}
 		if (!sent) {
 			workerIds.delete(targetThreadId);
 			continue;
 		}
-		return new Promise((resolve, reject) => {
+		return new Promise((promiseResolve, promiseReject) => {
 			const timer = setTimeout(() => {
 				pendingExecutions.delete(requestId);
-				reject(new ServerError(`Timed out waiting for worker thread to execute operation '${name}'`, 503));
+				promiseReject(new ServerError(`Timed out waiting for worker thread to execute operation '${name}'`, 503));
 			}, EXECUTE_TIMEOUT_MS);
 			timer.unref?.();
 			pendingExecutions.set(requestId, {
 				targetThreadId,
 				resolve(result) {
 					clearTimeout(timer);
-					resolve(result);
+					promiseResolve(result);
 				},
 				reject(error) {
 					clearTimeout(timer);
-					reject(error);
+					promiseReject(error);
 				},
 			});
 		});
@@ -201,14 +202,20 @@ export async function operationExecuteRequestHandler(event: {
 	let response;
 	try {
 		if (!localDispatch) throw new ServerError('This worker thread cannot execute operations', 503);
-		// The main thread already stripped bypass_auth from external requests; re-assert here so a
-		// forged ITC-level message can't skip the worker-side permission check.
-		delete body.bypass_auth;
+		// bypass_auth is trusted as forwarded: the main thread already strips it from external
+		// HTTP requests (handlePostRequest, before any dispatch decision), and an internal caller
+		// legitimately setting it (server.operation(op, context, false)) must see the same
+		// authorize:false behavior whether the operation happens to run locally or via this bridge.
+		// ITC is an internal, same-process trust boundary — a worker able to forge this message
+		// already has direct DB access and gains nothing by spoofing bypass_auth.
 		const operationFunction = localDispatch.chooseOperation(body);
 		const result = await localDispatch.processLocalTransaction({ body }, operationFunction);
 		if (result instanceof Readable || typeof result?.pipe === 'function') {
 			// Streaming results would need MessagePort transfer plumbing — explicitly unsupported
 			// for worker-registered operations for now (#1736), rather than failing opaquely.
+			// The handler already ran and the stream may hold an open fd/cursor/socket; destroy it
+			// rather than abandoning it to GC finalization.
+			result.destroy?.();
 			throw new ServerError(
 				`Operation '${body.operation}' returned a stream; streaming results are not supported for operations registered from a component`,
 				501
--- a/server/serverHelpers/registeredOperations.ts
+++ b/server/serverHelpers/registeredOperations.ts
@@ -135,7 +135,7 @@ function attachMainListeners() {
 	});
 }
 
-function executeRemoteOperation(name: string, body: any): Promise<any> {
+async function executeRemoteOperation(name: string, body: any): Promise<any> {
 	attachMainListeners();
 	const workerIds = registeredByWorker.get(name);
 	const forwardBody = { ...body };
@@ -158,8 +158,9 @@ function executeRemoteOperation(name: string, body: any): Promise<any> {
 			// server-side limitation, not a malformed client request — 500, not 400. No other worker
 			// will fare better with the same body.
 			operationLog.error(`Failed to forward operation '${name}' to worker thread ${targetThreadId}`, error);
-			return Promise.reject(
-				new ServerError(`Operation '${name}' request could not be forwarded to a worker thread: ${error.message}`, 500)
+			throw new ServerError(
+				`Operation '${name}' request could not be forwarded to a worker thread: ${error.message}`,
+				500
 			);
 		}
 		if (!sent) {
@@ -171,7 +172,7 @@ function executeRemoteOperation(name: string, body: any): Promise<any> {
 				pendingExecutions.delete(requestId);
 				promiseReject(new ServerError(`Timed out waiting for worker thread to execute operation '${name}'`, 503));
 			}, EXECUTE_TIMEOUT_MS);
-			timer.unref?.();
+			timer.unref();
 			pendingExecutions.set(requestId, {
 				targetThreadId,
 				resolve(result) {
@@ -186,8 +187,9 @@ function executeRemoteOperation(name: string, body: any): Promise<any> {
 		});
 	}
 	if (registeredByWorker.get(name)?.size === 0) registeredByWorker.delete(name);
-	return Promise.reject(
-		new ServerError(`Operation '${name}' is registered by a component but no worker thread is available to run it`, 503)
+	throw new ServerError(
+		`Operation '${name}' is registered by a component but no worker thread is available to run it`,
+		503
 	);
 }
 
--- a/server/serverHelpers/registeredOperations.ts
+++ b/server/serverHelpers/registeredOperations.ts
@@ -1,0 +1,250 @@
+/**
+ * Cross-thread bridge for `server.registerOperation()` (#1736).
+ *
+ * Components (jsResource `resources.js`, Plugin API `handleApplication`) load per-worker, so a
+ * `registerOperation()` call made there lands in that worker's module-local OPERATION_FUNCTION_MAP.
+ * The ops-API HTTP dispatcher runs only on the main thread and reads the main thread's copy of
+ * that map, so such operations were unreachable ("Operation '<name>' not found") for every caller.
+ *
+ * The bridge mirrors the RESOURCE_OPENAPI request/response pattern (operationsServer.ts /
+ * itc/serverHandlers.js), with one deliberate difference: executing an operation is side-effecting,
+ * so a request is sent to exactly ONE registering worker (never broadcast-first-wins).
+ *
+ *  - Worker: `registerOperation()` announces the name (OPERATION_REGISTERED) to all threads;
+ *    only the main thread records it, as name -> Set<threadId>.
+ *  - Main: on an OPERATION_FUNCTION_MAP miss, `getRemoteOperationFunction()` supplies a forwarding
+ *    function that sends the request body (OPERATION_EXECUTE_REQUEST) to one live registering
+ *    worker and awaits the correlated OPERATION_EXECUTE_RESPONSE.
+ *  - Worker: executes through the normal `chooseOperation` + `processLocalTransaction` path, so
+ *    permission checks run where the operation function (and its metadata) actually exists. The
+ *    main thread performs authentication only and forwards the resolved `hdb_user`.
+ */
+import { isMainThread, threadId } from 'node:worker_threads';
+import { Readable } from 'node:stream';
+import * as terms from '../../utility/hdbTerms.ts';
+import * as env from '../../utility/environment/environmentManager.ts';
+import harperLogger from '../../utility/logging/harper_logger.ts';
+import { ServerError } from '../../utility/errors/hdbError.ts';
+import { sendItcEvent } from '../threads/itc.js';
+import { onMessageByType, onThreadExit } from '../threads/manageThreads.js';
+
+const operationLog = harperLogger.loggerWithTag('operation');
+
+// The connected-ports array with `sendToThread` — assigned to the `threads` global by
+// manageThreads via _assignPackageExport (the same access pattern as itc/serverHandlers.js;
+// the globals.js `exports.threads` binding is reassigned after load, so it can't be imported).
+declare const threads: { sendToThread(threadId: number, message: any): boolean };
+
+// Fields attached to a request body by the HTTP/auth layer that must not (and often cannot)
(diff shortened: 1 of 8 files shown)


server/threads/manageThreads.js
```diff
@@ -823,10 +827,27 @@ if (parentPort && workerData?.addPorts) {
...
+function onThreadExit(listener) {
+	threadExitListeners.push(listener);
+}
+
 function removePort(port, deadThreadId) {
 	const idx = connectedPorts.indexOf(port);
 	if (idx === -1) return;
 	connectedPorts.splice(idx, 1);
+	if (deadThreadId != null) {
+		for (const listener of threadExitListeners) {
```
cb1kenobi: **Medium: worker-crash cleanup can be skipped, so in-flight forwards wait the full 120s timeout instead of a fast 503**

This new `threadExitListeners` dispatch only fires when `removePort(port, deadThreadId)` runs with a non-null `deadThreadId`. On a real worker crash with ≥2 workers there is a race: a sibling worker's MessagePort to the dead worker emits `close` first, that sibling broadcasts `REMOVE_PORT{threadId}` to the main thread, and main's `REMOVE_PORT` handler (unchanged, ~line 882) splices the dead worker's entry out of `connectedPorts` inline — without firing `threadExitListeners`. When main's own `Worker 'exit'` fires afterward, `removePort` hits `if (idx === -1) return` and the listeners never run. Result: `registeredOperations`' `onThreadExit` cleanup is skipped, so a forward already in flight to the dead worker is not rejected with the intended 503 and instead waits out `EXECUTE_TIMEOUT_MS` (default 120s). Bounded (stale registrations self-heal on the next `sendToThread`-returns-false; no wrong result), but it defeats the fast-fail path this PR added, under exactly the crash scenario it targets.

Suggested fix: fire `threadExitListeners` from the `REMOVE_PORT` handler path too (or route it through `removePort` with the threadId), guarding against double-firing per threadId (e.g. a `Set` of already-notified dead ids).

—
_Generated by Barber AI_

server/serverHelpers/registeredOperations.ts
```diff
@@ -0,0 +1,250 @@
...
+		if (!localDispatch) throw new ServerError('This worker thread cannot execute operations', 503);
+		// The main thread already stripped bypass_auth from external requests; re-assert here so a
+		// forged ITC-level message can't skip the worker-side permission check.
+		delete body.bypass_auth;
+		const operationFunction = localDispatch.chooseOperation(body);
+		const result = await localDispatch.processLocalTransaction({ body }, operationFunction);
+		if (result instanceof Readable || typeof result?.pipe === 'function') {
+			// Streaming results would need MessagePort transfer plumbing — explicitly unsupported
+			// for worker-registered operations for now (#1736), rather than failing opaquely.
+			throw new ServerError(
```
cb1kenobi, line 221: **Low: streaming result is abandoned without `.destroy()`, leaking fd/socket-backed streams**

When a forwarded op returns a `Readable`, `processLocalTransaction` has already executed the handler, so the stream may hold an open file descriptor, DB cursor, or socket. Throwing 501 here drops the reference without destroying it, so the underlying resource stays open until GC finalization. The fixture's `Readable.from([...])` holds no OS handle, so the test doesn't exercise this.

Suggested fix: call `result.destroy?.()` before throwing the 501.

—
_Generated by Barber AI_

server/serverHelpers/registeredOperations.ts
```diff
@@ -0,0 +1,250 @@
...
+			sent = threads.sendToThread(targetThreadId, {
+				type: terms.ITC_EVENT_TYPES.OPERATION_EXECUTE_REQUEST,
+				message: { requestId, body: forwardBody, originator: threadId },
+			});
+		} catch (error) {
+			// Structured-clone failure (e.g. a function or native handle on the request body) —
+			// no other worker will fare better with the same body.
+			operationLog.error(`Failed to forward operation '${name}' to worker thread ${targetThreadId}`, error);
+			return Promise.reject(
+				new ServerError(`Operation '${name}' request could not be forwarded to a worker thread: ${error.message}`, 400)
```
cb1kenobi: **Low: structured-clone failure on forward is reported as a client `400`**

A `DataCloneError` from `sendToThread` means an internal, request-scoped field (e.g. something hanging off `hdb_user`) isn't structured-cloneable — not that the client's request was malformed. The HTTP body is always JSON-cloneable, so a client cannot trigger this; labeling it `400` misattributes a server-side limitation to the caller.

Suggested fix:
```suggestion
					new ServerError(`Operation '${name}' request could not be forwarded to a worker thread: ${error.message}`, 500)
```

—
_Generated by Barber AI_

server/serverHelpers/registeredOperations.ts
```diff
@@ -0,0 +1,250 @@
...
+				`Operation '${body.operation}' returned a stream; streaming results are not supported for operations registered from a component`,
+				501
+			);
+		}
+		response = { requestId, result };
+	} catch (error) {
+		response = {
+			requestId,
+			error: {
+				message: error.http_resp_msg?.error ?? error.http_resp_msg ?? error.message,
```
cb1kenobi, line 231: **Nit: an object-valued `http_resp_msg` without an `.error` key collapses to `"[object Object]"`**

`error.http_resp_msg?.error ?? error.http_resp_msg ?? error.message` — if `http_resp_msg` is an object that lacks an `.error` key, the whole object flows into `message` and is later stringified by `new ServerError(message, ...)` (line 118) to `"[object Object]"`. A normal (non-forwarded) op preserves the structured body via `resp.send(error.http_resp_msg)`. Rare in practice (custom ops throw plain Errors or string-message HdbErrors), so acceptable to leave — but if structured error bodies matter, forward `http_resp_msg` as-is and branch on object vs string on the main side.

—
_Generated by Barber AI_

server/serverHelpers/registeredOperations.ts
```diff
@@ -0,0 +1,250 @@
...
+				resolve(result) {
+					clearTimeout(timer);
+					resolve(result);
+				},
+				reject(error) {
+					clearTimeout(timer);
+					reject(error);
+				},
+			});
+		});
```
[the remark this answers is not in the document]
kriszyp: Verified this doesn't actually self-shadow — object-literal method shorthand (`{ resolve(result) { ... } }`) does not create a self-referencing binding inside its own body the way a named function expression (`function resolve(){}`) does, so `resolve` inside the method correctly resolves to the outer Promise executor's parameter via normal lexical scoping. Confirmed against the actual compiled dist output (unchanged shorthand-method shape) and with a standalone runtime test reproducing the exact pattern — resolves cleanly, no recursion. That said, renamed to `promiseResolve`/`promiseReject` anyway in dc52db3ea since it's a free clarity improvement that removes any doubt for future readers. Thanks for the sharp eye.

server/serverHelpers/registeredOperations.ts
```diff
@@ -0,0 +1,250 @@
...
+export async function operationExecuteRequestHandler(event: {
+	message: { requestId: number; body: any; originator: number };
+}) {
+	const { requestId, body, originator } = event.message;
+	let response;
+	try {
+		if (!localDispatch) throw new ServerError('This worker thread cannot execute operations', 503);
+		// The main thread already stripped bypass_auth from external requests; re-assert here so a
+		// forged ITC-level message can't skip the worker-side permission check.
+		delete body.bypass_auth;
```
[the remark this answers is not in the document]
kriszyp: Confirmed and fixed in dc52db3ea — you're right, this was overly restrictive. Traced the actual paths: external HTTP requests already have bypass_auth stripped before any dispatch decision (handlePostRequest, server/serverHelpers/serverHandlers.js:143), so they can never reach this worker-side code with it set. The only way bypass_auth:true reaches here is a legitimate internal server.operation(op, context, false) call forwarded from the main thread, which must behave the same whether the target operation happens to run locally or via this bridge. Removed the delete and documented why ITC is the wrong place to re-enforce it (same-process trust boundary — consistent with how we resolved the forged-hdb_user finding in the earlier review round).

server/threads/manageThreads.js
kriszyp: Confirmed, real race — thanks for tracing this precisely. Fixed in dc52db3ea: both removal paths (removePort's own port close/exit, and the REMOVE_PORT fast-path splice) now go through a shared notifyThreadExit() that dedupes by threadId (Set, never pruned — worker_threads ids are monotonic and never reused within a process, so this is safe unbounded growth), so whichever path observes the dead worker first fires the listeners exactly once.

server/serverHelpers/registeredOperations.ts
kriszyp: Good catch, fixed in dc52db3ea — added result.destroy?.() before the 501 throw.

server/serverHelpers/registeredOperations.ts
kriszyp: Agreed, fixed in dc52db3ea — changed to 500.

server/serverHelpers/registeredOperations.ts
kriszyp: Agreed this is a rare, low-severity nit — leaving as-is per your own read; forwarding structured http_resp_msg objects through would need ServerError's message type widened and isn't worth it for the realistic case (custom ops throwing plain Errors or string-message HdbErrors). Noted for awareness if it ever comes up in practice.

server/serverHelpers/registeredOperations.ts
```diff
@@ -0,0 +1,257 @@
...
+		if (!sent) {
+			workerIds.delete(targetThreadId);
+			continue;
+		}
+		return new Promise((promiseResolve, promiseReject) => {
+			const timer = setTimeout(() => {
+				pendingExecutions.delete(requestId);
+				promiseReject(new ServerError(`Timed out waiting for worker thread to execute operation '${name}'`, 503));
+			}, EXECUTE_TIMEOUT_MS);
+			timer.unref?.();
```
cb1kenobi: Why wouldn't `unref()` be defined? Is this a sandboxed `setTimeout()` or something?

server/serverHelpers/registeredOperations.ts
```diff
@@ -0,0 +1,257 @@
...
+		for (const [requestId, pending] of pendingExecutions) {
+			if (pending.targetThreadId === deadThreadId) {
+				pendingExecutions.delete(requestId);
+				pending.reject(new ServerError('The worker thread executing this operation exited', 503));
+			}
+		}
+	});
+}
+
+function executeRemoteOperation(name: string, body: any): Promise<any> {
```
cb1kenobi: I recommend making this an `async` function so that any thrown error would consistently reject.
```suggestion
async function executeRemoteOperation(name: string, body: any): Promise<any> {
```
Plus it removes the awkward `return Promise.reject()` lines.

cb1kenobi (approved):
Human and AI reviewed. No issues, neat feature!

server/serverHelpers/registeredOperations.ts
kriszyp: Done in 760194e4 — `executeRemoteOperation` is now `async`, so both the structured-clone forward failure and the no-worker-available case `throw` instead of `return Promise.reject(...)`, and any synchronous throw (e.g. from `attachMainListeners`) rejects consistently. Good call. — KrAIs

server/serverHelpers/registeredOperations.ts
kriszyp: No sandboxing — it was just over-defensive. `setTimeout()` returns a `Timeout`/`Timer` with `unref()` in both Node and Bun on this main-thread path, so the guard could never actually be hit. Dropped the `?.` in 760194e4 so it reads as the plain `timer.unref()` it always was. — KrAIs

PR #1744, opened 2026-07-09
fix: sweep raw-Error logging leaks across core (#1734)

kriszyp:
## Summary
Fixes the raw-`Error` logging leak class (#1734) at its root. Harper's logger extends Node's `Console`; passing a raw `Error` to a Console method formats it with `util.inspect`, which dumps **every own-enumerable property** after the stack. Libraries/app code stash secrets on thrown Errors (axios `config`/`request` headers, an `hdb_secret` used for an outbound `Authorization` header), so `logger.<level>(error)` landed those credentials verbatim in `hdb.log` at default level — retrievable via `read_log`.

Two mechanisms, matched to the two leak surfaces:

1. **Logger auto-wrap (the invariant).** The seven `HarperLogger` level methods route every `Error` argument through `errorForLog()` (`sanitizeErrorArgs`) inside the level gate, and the inherited Console methods that format arbitrary values (`log`/`dir`/`table`) are guarded too. This covers **every** logger call — all first-party code, future code, and component/app code holding a `server.logger` — with no per-site changes. Gated-out calls pay nothing beyond an arg scan; allocation only happens when an Error is present; a revoked-Proxy arg can't make the logger throw (`isErrorLike` try/catch). `forComponent`, `loggerWithTag`, and the top-level module exports all route through these methods.
2. **Explicit wraps on `console.*` calls with app/library error provenance (7 sites).** `harper_logger` captures both stdout and stderr into `hdb.log`, but Node's Console runs `util.inspect` *before* the capture hook — so `console.error(rawError)` leaks identically. A sanitizing console subclass was rejected (overriding console methods makes the inspector report the override frame instead of the real call site), so wrapping is per-site — and scoped by **error provenance**: only sites where the error can originate in app/component/library code (JSONStream serializing resource iterator output — the origin-fetch axios scenario flows through here; componentLoader app-load failures; loadRootComponents' npm install of user packages) or the handler is generic (socketRouter's `uncaughtException`). Console sites with a closed error set (fs/LMDB/pm2/config — copyDb, CLI bin/*, bootstrap paths) stay raw: those errors cannot carry stashed credentials.

An earlier iteration of this branch wrapped ~180 logger call sites explicitly; those are all reverted in favor of the auto-wrap — the final diff is 6 files.

## Where to look
- **`utility/logging/harper_logger.ts`** — `sanitizeErrorArgs` + the seven level methods. This is the load-bearing change. Detection is `instanceof Error || util.types.isNativeError` (cross-realm safe). Thrown non-Error plain objects are deliberately NOT wrapped — object args are a legitimate structured-logging pattern and wrapping would garble them.
- **The console-site wraps** (`JSONStream.ts`, `socketRouter.ts`, `componentLoader.ts`, `loadRootComponents.js`) — mechanical; the error arg gets `errorForLog(...)`, message strings/data args untouched. `JSONStream`'s `errorToString` response pushes stay raw (client-facing body, not a log line).
- **`componentLoader.ts:610`** — `(getWorkerIndex() === 0 ? console : harperLogger).error(errorForLog(error))` keeps an explicit wrap because the worker-0 branch goes to raw `console`.

## Known gaps (accepted)
- Component/app code calling **global `console.error(err)`** still leaks via the stdio capture — only our own console sites are wrapped. Fixing that class requires the console subclass we rejected for inspector ergonomics, or an upstream-Node-level answer.
- Sanitization is **shallow**: an Error nested inside a logged object/array (`logger.error({ err })`) is not rewritten — deep-walking every logged structure isn't worth the per-call cost, and the #1734 threat is raw thrown errors.

## Testing
17 auto-wrap + `errorForLog` unit tests (incl. log/dir/table bypass and revoked-Proxy no-throw): sole-arg / any-position / cause-chain secret exclusion, non-Error objects untouched, level gating still suppresses below-gate writes, `loggerWithTag`-derived loggers covered, null/undefined/cyclic-cause safety. Built to `dist` and ran against it (mocha's `--conditions=typestrip` path can silently fall back to stale `dist`): **41 passing**. `oxlint` + `prettier` clean.

## Cross-model review
Thorough mode ran on the sweep iteration (Codex found 3 missed sites; the Harper domain pass found the console-capture surface and non-standard logger bindings — all of which informed the auto-wrap design). Gemini (`agy`) leg timed out on both attempts — no second-model coverage from Gemini. A focused Codex pass on the final auto-wrap diff found the `log`/`dir`/`table` bypass, the revoked-Proxy throw, and two template-literal console sites the sweep filter missed — all fixed; its nested-Error finding is the accepted shallow-sanitization gap above.

_Generated by an LLM (Claude Fable 5 / Opus 4.8)._

### DIFF
--- a/bin/copyDb.ts
+++ b/bin/copyDb.ts
@@ -54,7 +54,7 @@ export async function compactOnStart() {
 				recordCount = await getTotalDBRecordCount(databaseName);
 				console.log('Database', databaseName, 'before compact has a total record count of', recordCount);
 			} catch (error) {
-				hdbLogger.error('Error getting record count for database', databaseName, hdbLogger.errorForLog(error));
+				hdbLogger.error('Error getting record count for database', databaseName, error);
 				console.error('Error getting record count for database', databaseName, hdbLogger.errorForLog(error));
 			}
 			compactedDb.set(databaseName, {
@@ -80,19 +80,19 @@ export async function compactOnStart() {
 		try {
 			resetDatabases();
 		} catch (err) {
-			hdbLogger.error('Error resetting databases after backup', hdbLogger.errorForLog(err));
+			hdbLogger.error('Error resetting databases after backup', err);
 			console.error('Error resetting databases after backup', hdbLogger.errorForLog(err));
 		}
 
 		try {
 			resetDatabases();
 		} catch (err) {
-			hdbLogger.error('Error resetting databases after backup', hdbLogger.errorForLog(err));
+			hdbLogger.error('Error resetting databases after backup', err);
 			console.error('Error resetting databases after backup', hdbLogger.errorForLog(err));
 			process.exit(0); // just let the process restart
 		}
 	} catch (err) {
-		hdbLogger.error('Error compacting database, rolling back operation', hdbLogger.errorForLog(err));
+		hdbLogger.error('Error compacting database, rolling back operation', err);
 		console.error('Error compacting database, rolling back operation', hdbLogger.errorForLog(err));
 
 		updateConfigValue(CONFIG_PARAMS.STORAGE_COMPACTONSTART, false);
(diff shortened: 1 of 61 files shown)


kriszyp:
Superseded by #1749 — #1737 (this PR's base) was squash-merged into `main`, so this branch's history no longer produces a clean diff (it would re-show `REST.ts`/`http.ts`/`graphqlQuerying.ts` as changed, duplicating already-merged content). #1749 is a clean transplant of the same work onto current `main`, plus a diagnostic-property allowlist added as a follow-up. Closing this one.

PR #1753, opened 2026-07-10
fix(lint): unbreak main — node:assert/strict → node:assert in bytes-scan-crash test

kriszyp:
## Summary

`integrationTests/resources/bytes-scan-crash.test.ts:26` imports `{ ok } from 'node:assert/strict'`, which the `no-restricted-imports` eslint rule forbids. That left `lint:required` (`runLinter`) **red on `main`** — `1 problem (1 error)` — which surfaces on every downstream PR's CI.

The import was added in 7945918e6 (`fix(table): reject bare-array roots…`). `ok` is identical when imported from `node:assert`, which is the form all sibling integration tests already use.

```diff
-import { ok } from 'node:assert/strict';
+import { ok } from 'node:assert';
```

## Test plan
- [x] `npm run lint:required` passes locally (was `exit 1`, now `exit 0`)

### DIFF
--- a/integrationTests/resources/bytes-scan-crash.test.ts
+++ b/integrationTests/resources/bytes-scan-crash.test.ts
@@ -23,7 +23,7 @@
  * Implements HarperFast/harper#1298.
  */
 import { suite, test, before, after } from 'node:test';
-import { ok } from 'node:assert/strict';
+import { ok } from 'node:assert';
 import { resolve } from 'node:path';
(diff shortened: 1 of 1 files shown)


Ethan-Arrowood (approved):
Fine with this to unblock main, but I'm generally not okay with the switch from `node:assert/strict` to `node:assert`. But that is a separate conversation.

cb1kenobi (approved):

PR #1767, opened 2026-07-12
Typed, discoverable resources — code-first defineTable + per-method request contract (RFC 0001)

kriszyp:
Implements **RFC 0001** — the mergeable implementation of typed, discoverable resources, integrated onto current `main`. The full RFC + type-level spikes stay in the design PR #1503; this PR is the code intended for merge.

## What this adds

**Pillar 1 / 2b — code-first schema** (`resources/defineTable.ts`)
`defineTable(name, shape, opts)` + `types` author a table in TypeScript and eagerly register it through the same `table()` factory GraphQL drives — the returned value *is* the live table class, with per-verb shapes inferred as `$record`/`$insert`/`$upsert`/`$patch`/`$query` projections. Relations via lazy thunks (+ `relationOf`/`hasManyOf` for mutual pairs).

**Pillar 2 — per-method request contract** (`resources/withSchema.ts`)
- `defineResource(contract, impl)` — function form.
- `Resource.withSchema(contract)` — class form (`extends` it).

A contract is `{ path, record?, get/post/put/patch/delete: { query?, body?, response? } }`. Handler types are **derived** from it; a handler receives the SAME `RequestTarget`, structurally narrowed (`target.id: string`, `target.get('expand')` typed by the query schema) — a **subset, not a fork**. The narrowed types are justified by **runtime enforcement**: each declared verb validates/coerces `query`/`body` before dispatch and returns a structured 400 (`ValidationError`, per-field `{ path, code, message }[]`).

**One vocabulary.** Built-in `t`/`schemaOf` reduce to `JsonSchemaFragment` (the IR `defineTable`/GraphQL fields already reduce to via `attributeToFragment`), so one declaration feeds validation + OpenAPI + MCP. A `defineTable` projection slots into a contract body via `schemaOf({ table: Track, projection: 'insert' })`.

**Nullability:** non-nullable by default (a bare `t.string` rejects `null`); `.optional` allows absence, `.nullable` allows explicit `null`; table-derived bodies mirror `Table.validate`'s policy.

## Cross-cutting
- `ValidationError` (extends `ClientError`, 400); `Table.validate` refactored to the same structured shape (HTTP-title message byte-preserved).
- OpenAPI emits declared query params / request body / response for parameterised routes.
- MCP drives tool input/output off the contract and binds arbitrary path params + query (`applyContractInputs`).
- Shared `attributeToFragment` hardened with a nested-object branch; `derive.ts` `Object`/`Array` projection bugfix.

## Integration notes (for reviewers)
The RFC branch was ~1007 lines behind `main` on the exact hot files this touches. Resolved into main's newer work:
- **MCP paramroutes**: main added `paramBinding` gating (`isSimpleIdRoute` — drops generated verbs it can't bind) and `mcpResources`. A **request contract now exempts a resource from that restriction** (the contract binds arbitrary path params + query), which is the "richer binding rides on the contract" that main's own comments defer to.
- **`liveResource` authz fix** (#1487): the contract handler wiring composes with main's per-call live-resource resolution.

## Review provenance
Cross-model reviewed (Codex + Harper-domain adjudicator) on the feature branch; the Gemini leg timed out (diff at `agy`'s hang threshold). Three "blockers" resolved to one false positive (arg-order — `loadAsInstance` pin) + two fixed (MCP flatten-poisons-body, output verb-name mismatch). No confirmed blockers.

## Tests
95 unit tests green on the merged tree, incl. `withSchema.test.js`, `defineTable-registration.test.js`, `openApi-contract.test.js`, `application-contract.test.js`, and main's `application-paramroutes.test.js`. Type contract verified against built exports via `docs/rfcs/spikes/0001/*-real.check.ts` (`tsc --project`).


dawsontoth (approved):
I like the shape of it a lot.

Issue #1780, opened 2026-07-13
npm-shrinkwrap.json (5.1.18+) breaks consumers' `npm ci` on npm 11 — EBADPLATFORM @esbuild/aix-ppc64


kylebernhardy:
## Summary

harper ≥ 5.1.18 publishes an `npm-shrinkwrap.json` (5.1.9 did not). When a project depends on harper and its lockfile is generated with **npm 11** (bundled with current Node 24 releases), npm folds harper's ~1,210-package shrinkwrap subtree into the consumer's `package-lock.json` **nested under `node_modules/harper/`** — and drops the `"optional": true` flag from platform-specific packages (`@esbuild/*`, pulled in via `tsx → esbuild`).

`npm ci` then treats `@esbuild/aix-ppc64` as a required package on every platform and fails:

```
npm error code EBADPLATFORM
npm error notsup Unsupported platform for @esbuild/aix-ppc64@0.28.1:
  wanted {"os":"aix","cpu":"ppc64"} (current: {"os":"linux","cpu":"x64"})
```

To be clear about fault: harper's shrinkwrap itself is **correct** (it marks these packages `"optional": true`); the flag is lost in npm 11's reify. npm **12.0.1** handles it properly (the nested subtree collapses to a single entry). But since npm 11 is what Node 24 ships, every consumer with harper as an npm dependency and a stock CI image hits this.

## Independent repro

On any macOS/Linux machine with Node 24 + npm 11 (verified with npm 11.8.0 and 11.13.0):

```bash
mkdir repro && cd repro
npm init -y
npm install harper@5.1.19        # or 5.1.18 — both ship the shrinkwrap
# lockfile is now poisoned:
node -e "
const l = require('./package-lock.json');
const e = l.packages['node_modules/harper/node_modules/@esbuild/aix-ppc64'];
console.log('entry exists:', !!e, '| optional flag:', e && e.optional);"
# → entry exists: true | optional flag: undefined

rm -rf node_modules
npm ci                            # → EBADPLATFORM (unless you happen to be on AIX/ppc64)
```

Control: repeat with `npx npm@12 install` — the lockfile keeps only one harmless nested entry and `npx npm@12 ci` succeeds. Repeat with `harper@5.1.9` (no shrinkwrap) — no nested entries at all, npm 11 works fine.

## Impact

[...]

PR #1781, opened 2026-07-13
fix(build): prune devDependencies before shrinkwrapping

pbrumblay:
Follow-up to #1622.

npm shrinkwrap runs after a full npm install in build-tools/build.sh, so the generated npm-shrinkwrap.json captures devDependencies as well as runtime ones. Since #1622 started shipping npm-shrinkwrap.json in the published tarball, consumers now get harper's entire dev tree installed under node_modules/harper/node_modules/ -- including esbuild (pulled in transitively via tsx) and its ~30 platform-specific optional binaries.

Those nested @esbuild/<platform> entries land in the consumer's package-lock.json as "extraneous" rather than resolved-optional, because they aren't reachable from any declared dependency edge in the consumer's graph -- they're just mirrored in from harper's bundled shrinkwrap. Extraneous entries don't get the "optional": true flag written, so npm ci in consumer projects fails with EBADPLATFORM on platforms other than the shrinkwrap's original build machine (e.g. @esbuild/aix-ppc64 on darwin/arm64).

Prune devDependencies before shrinkwrapping so the pinned tree only contains what harper actually needs at runtime.

### DIFF
--- a/build-tools/build.sh
+++ b/build-tools/build.sh
@@ -8,6 +8,9 @@ npm install --ignore-scripts
 echo -e "\n📦 Building project"
 npm run build || true
 
+echo -e "\n📦 Pruning devDependencies"
+npm prune --omit=dev
+
 echo -e "\n📦 Creating shrinkwrap"
 npm shrinkwrap
 


kriszyp:
Nice, this presumably resolves https://github.com/HarperFast/harper/issues/1780

kriszyp (approved):
Approving — correct minimal fix for #1780 (devDependencies were leaking into the published artifact's npm-shrinkwrap). Build-time only, no runtime path touched. The failing Unit Test (Node 26) is a pre-existing rocksdb flake unrelated to this change. — KrAIs (Claude Opus 4.8), on Kris's behalf

PR #1786, opened 2026-07-14
Record-scoped allowRead: unified row-level read access control (#1422 gap 2)

### LINKED ISSUE #1419: Subscription delivery bypasses row-level `allowRead` — all rows leak to any collection subscriber

### LINKED ISSUE #1422: `allowRead` has two critical security gaps: (1) throw → fail-open, (2) not enforced on collection scans

kriszyp:
Replaces the short-lived static `allowReadRecord` hook (#1768, unreleased) with a unified model, and closes **[#1422 (`allowRead` has two critical security gaps)](https://github.com/HarperFast/harper/issues/1422) gap 2** — the row-level enforcement that was deliberately dropped from [#1489 (fail closed when allow* hooks throw)](https://github.com/HarperFast/harper/pull/1489) under the earlier "grant hooks only" stance, now reversed: `allowRead` (and friends) are the key mechanism for row-level access control. Closes [#1419 (Subscription delivery bypasses row-level allowRead)](https://github.com/HarperFast/harper/issues/1419).

## The model

One `allowRead(user, target, context)` definition serves both scopes:

- **Framework defaults** (Resource base + the table RBAC check) are this-free and marked `isDefaultAllowRead` — they keep today's single entry evaluation. Zero new overhead for non-overriding tables.
- **An application override on a table is record-scoped**: evaluated once per record with `this` = the (frozen) record during query execution — pushed into HNSW traversal for vector sorts, applied as a post-filter otherwise, re-checked on the materialized record after caching-source revalidation. `super.allowRead(...)` composes the RBAC baseline into the same function.

The authorize wrapper defers table collection reads (`supportsRowLevelAllowRead` + overridden) to the per-record path; `get`/`query` on collections route into `search()` which consumes `checkPermission` and arms the guard. Single-record `get(id)` keeps the entry check (record loaded; `TableResource` proxies attribute reads to the record, so the same override reads correctly there — this closes #1422's "single-record 403s but collection leaks every row" asymmetry).

## Fail-closed properties
- A throwing per-record check denies that record (#1422 gap 1 parity).
- A Promise-returning check is a hard `ClientError` (must be synchronous), never fail-open truthiness.
- Dispatch is via the resolved class method — never a `record.allowRead` property lookup — so a record *attribute* named `allowRead` can't shadow the check.
- **Subscription delivery is enforced per event (#1419, revives the closed #1524):** a sync record-scoped `allowRead` filters each record-bearing event (`this` = the event's record) across the live listener, queue drain, reload re-snapshot, and `previousCount` history replay (denied rows don't consume count slots). subscribe/connect defer at entry for sync overrides — delivery is the enforcement; async overrides stay entry-checked (awaited, fail-closed). The #1414 re-auth recheck verifies the framework-default (table-level) grant for record-scoped subscriptions instead of spuriously killing them. Known limitation: delete tombstones / message payloads / rawEvents don't carry the full record; row-authorizing those event types is deferred.
- Role-level column RBAC (`attribute_permissions`) still applies under an override: `search()` runs the default `allowRead` for its `target.select`-narrowing side effect before deferring row access.

Records also expose a non-enumerable `allowRead` delegate on the per-table `structPrototype` for direct app-code use (delegates to the table default; engine enforcement resolves the active subclass instead).

## Where to look
- `resources/Resource.ts` — the deferral gate. Its predicate set (read-only, non-subscribe, collection-shaped, table, overridden) is the security-critical decision; the `get` arm requires `isCollection` to mirror the `isSearchTarget` routing it relies on.
- `resources/Table.ts` `search()` — override detection + guard construction; the composition call for column narrowing.
- Behavioral change to note: an existing table subclass that overrides `allowRead` and relied on a collection-entry verdict now gets per-record filtering instead (200 + filtered rows rather than a blanket 403). This is the intended semantics of the new model (approved design decision), but it's a semantic shift on a security surface — release-notes callout included in the docs PR.

## Known limitations (deliberate)
- ~~GraphQL~~ **GraphQL is covered**: it now requests authorization via `query.checkPermission` (replacing the context-level `authorize` alias, for consistency with REST), and since GraphQL invokes the static `search` wrapper, row-level enforcement applies to its collection queries. Verified by the graphql integration suite.
- Per-record `super.allowRead` calls can't affect `select` narrowing (transform is built before iteration); narrowing is composed at entry instead.

## Review
Cross-model review attempted: both outside-model legs failed this round (Gemini/agy empty twice; Codex CLI invocation failed — its output was a same-model trace). The Opus domain adjudication pass ran in full and its two significant findings (column-narrowing composition, deferral-gate asymmetry) are fixed in the last commit. Human review should weigh the deferral-gate predicate carefully given the reduced outside-model coverage.

Docs: companion update in [HarperFast/documentation#580](https://github.com/HarperFast/documentation/pull/580) (reworked from allowReadRecord to this model).

Generated by KrAIs (Claude Opus 4.8).


resources/Table.ts
```diff
@@ -4503,6 +4533,26 @@ export function makeTable(options) {
...
+	// NOT route through this (it resolves the active resource class's allowRead, which honors
+	// endpoint subclass overrides and can't be shadowed by a record attribute of the same name).
+	if (primaryStore?.encoder?.structPrototype) {
+		Object.defineProperty(primaryStore.encoder.structPrototype, 'allowRead', {
+			value: function (user: User, target: RequestTarget, context: Context) {
+				return TableResource.prototype.allowRead.call(this, user, target, context);
+			},
+			configurable: true,
+			writable: true,
+		});
```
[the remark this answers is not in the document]
kriszyp: Documented behavior, kept as-is: the delegate is an app-code convenience that intentionally binds the table's base allowRead — engine enforcement never routes through it and always resolves the ACTIVE endpoint subclass's method (see the comment above the defineProperty). Resolving via this.constructor isn't possible here: records belong to the table, and one table can be exported through multiple endpoint subclasses with different allowReads, so a record alone can't know which endpoint's policy applies. — KrAIs (Claude)

resources/Table.ts
```diff
@@ -2854,16 +2873,30 @@ export function makeTable(options) {
...
+				const user = (context as any)?.user;
+				recordGuard = (record: any) => {
+					let allowed;
+					try {
+						allowed = recordScopedAllowRead.call(record, user, target, context);
+					} catch {
+						return false; // fail closed on a throwing check
+					}
+					if (allowed?.then)
+						throw new ClientError('allowRead must be synchronous when evaluated per record in a query');
```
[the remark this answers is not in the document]
kriszyp: Addressed in c0ef526ce: the check is now `typeof allowed?.then === 'function'`, and the semantics changed per the adjudication — a thenable verdict denies that record (fail closed) with a one-time warning instead of aborting the query; declared-async overrides are excluded from record-scoping entirely and keep the awaited entry check. — KrAIs (Claude)

resources/Resource.ts
```diff
@@ -733,6 +739,33 @@ function transactional(
...
+					// until per-event delivery checks exist (#1419). The collection predicate must be at
+					// least as strict as the routing it relies on: instance get() only reaches search()
+					// via isSearchTarget (isCollection), so `get` defers on isCollection alone — a
+					// hypothetical {id: null, isCollection: false} get keeps the entry check instead of
+					// skipping both checks. search/query invoke instance search() directly (which
+					// consumes checkPermission), so id == null suffices there.
+					if (
+						options.type === 'read' &&
+						!isSubscribeAction &&
+						(query.isCollection || (query.id == null && options.method !== 'get')) &&
```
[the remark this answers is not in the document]
kriszyp: Confirmed and fixed in fa19aaa41: the gate now defers ALL search/query calls (a present id there is a prefix seed / multi-record scan, as you traced); only `get` keeps id-based single-record entry semantics, and its arm still requires isCollection to defer. New unit test covers an id-prefix search with a record-scoped override ('id-prefix search with a record-scoped override still filters per record'). Good catch. — KrAIs (Claude)

resources/Resource.ts
```diff
@@ -733,6 +739,37 @@ function transactional(
...
+					// hypothetical {id: null, isCollection: false} get keeps the entry check instead of
+					// skipping both checks. search/query invoke instance search() directly (which
+					// consumes checkPermission), so id == null suffices there.
+					// Record-scoping is SYNC-only: an async allowRead override keeps this entry check,
+					// which awaits its verdict and fails closed on rejection (#1422 gap 1) — the sync
+					// per-record traversal path cannot await it.
+					if (
+						options.type === 'read' &&
+						!isSubscribeAction &&
+						(query.isCollection || (query.id == null && options.method !== 'get')) &&
```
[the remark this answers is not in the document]
kriszyp: Fixed in fa19aaa41 (the predicate is now `options.method !== 'get' || query.isCollection` — the id == null keying is gone); see the sibling thread for detail. — KrAIs (Claude)

resources/Resource.ts
```diff
@@ -802,14 +842,29 @@ function registerLiveSubscriptionForContext(subscription: any, resource: any, qu
...
+					const fn = Object.getOwnPropertyDescriptor(proto, 'allowRead')?.value;
+					if (fn?.isDefaultAllowRead) {
+						allowReadFn = fn;
+						break;
+					}
+					proto = Object.getPrototypeOf(proto);
+				}
+			}
+			return !!(await allowReadFn.call(resource, fresh, reTarget, context));
 		},
```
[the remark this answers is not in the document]
kriszyp: Addressed in 1e8f9caa6: added a REVOCATION case to the #1419 integration suite that opens a record-scoped-override subscription and strips read via alter_role (user stays present, so `findAndValidateUser` still returns a role — termination can only come from the default table-grant fallback, the exact new branch). Verified it FAILS when the prototype-walk fallback is disabled (the collection-permissive override alone keeps the revoked subscription alive) and passes with it. — KrAIs (Claude)

resources/Resource.ts
```diff
@@ -802,14 +846,29 @@ function registerLiveSubscriptionForContext(subscription: any, resource: any, qu
...
+				while (proto) {
+					const fn = Object.getOwnPropertyDescriptor(proto, 'allowRead')?.value;
+					if (fn?.isDefaultAllowRead) {
+						allowReadFn = fn;
+						break;
+					}
+					proto = Object.getPrototypeOf(proto);
+				}
+			}
+			return !!(await allowReadFn.call(resource, fresh, reTarget, context));
```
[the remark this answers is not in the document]
kriszyp: Correct — this is now a real gap after the deferral revert, not just coverage. Fixed in dd4add0f0: recheck calls `resource.allowRead(fresh, reTarget, context)` directly, mirroring the entry check, so a connection-level override (topic ACL) re-checks its own grant on re-auth. Record-scoped overrides compose the RBAC baseline via `super.allowRead` (updated the #1419 fixture + docs to the recommended pattern), so revoking the role's read still terminates them — verified by the alter_role revocation test, which now exercises the mirrored override path (and fails if the override omits the super gate). Thanks for pushing on this. — KrAIs (Claude)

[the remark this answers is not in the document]
kriszyp: Correct — this is now a real gap after the deferral revert, not just coverage. Fixed in dd4add0f0: recheck calls `resource.allowRead(fresh, reTarget, context)` directly, mirroring the entry check, so a connection-level override (topic ACL) re-checks its own grant on re-auth. Record-scoped overrides compose the RBAC baseline via `super.allowRead` (updated the #1419 fixture + docs to the recommended pattern), so revoking the role's read still terminates them — verified by the alter_role revocation test, which now exercises the mirrored override path (and fails if the override omits the super gate). Thanks for pushing on this. — KrAIs (Claude)

resources/Table.ts
```diff
@@ -3362,6 +3410,51 @@ export function makeTable(options) {
...
+				subUser != null &&
+				!(subAllowRead as any)?.isDefaultAllowRead &&
+				(subAllowRead as any)?.constructor?.name !== 'AsyncFunction';
+			let warnedAsyncEvent = false;
+			const allowsEvent = filterRowReads
+				? (event: any): boolean => {
+						if (event.type === 'end_txn' || event.type === 'reload') return true; // control markers, no record
+						let decision;
+						try {
+							decision = subAllowRead.call(event.value ?? null, subUser, request, subContext);
```
cb1kenobi: **Medium: subscription per-record `allowRead` filter evaluates against a stale user snapshot, defeating #1414 re-auth for record-scoped subscriptions**

`subUser` is captured once at subscribe time (line 3430) and closed over here. The #1414 recheck does `(context as any).user = fresh` in place (`Resource.ts` `registerLiveSubscriptionForContext`) precisely so downstream checks — the table-level recheck and `getCurrentUser()` — evaluate against the refreshed principal. This delivery filter is such a downstream check, but it keeps evaluating against the original `subUser`. An override that decides row visibility from mutable user state (role/group/custom claims — not just the immutable `username` the fixture uses) will keep delivering now-unauthorized rows after a role change, for as long as the user retains the table-level grant (so the re-auth default-fallback at `Resource.ts:861` does not terminate the stream). Fail-open on a row-level read-authorization surface. The added tests don't catch this — the fixture keys on `owner === user.username`, which is stable across re-auth.

Secondary, same block: the `filterRowReads` gate (lines 3432-3433) disables per-record filtering entirely when `subUser == null`, so an authorization-checked but unauthenticated subscription (table grants anonymous read, rows restricted only by the override) would be delivered every row unfiltered.

Suggested fix — evaluate against the live context user rather than the captured snapshot:
```suggestion
						decision = subAllowRead.call(event.value ?? null, subContext.user, request, subContext);
```
and base the `filterRowReads` gate on the user present at delivery (or drop the `!= null` short-circuit) so an override still runs for an anonymous principal.

—
_Generated by Barber AI_

resources/Resource.ts
```diff
@@ -733,6 +739,44 @@ function transactional(
...
+					// must run at subscribe time to reject the topic. A record-level override instead
+					// returns permissive at collection scope (no record loaded) to open the subscription,
+					// and Table.subscribe then filters each delivered event per record (#1419).
+					// Record-scoping is SYNC-only: an async allowRead override keeps this entry check,
+					// which awaits its verdict and fails closed on rejection (#1422 gap 1) — the sync
+					// per-record traversal path cannot await it.
+					if (
+						options.type === 'read' &&
+						!isSubscribeAction &&
+						(options.method !== 'get' || query.isCollection) &&
```
cb1kenobi: **High: HTTP `QUERY` verb bypasses record-scoped `allowRead` entirely**

For a table with an overridden (sync) `allowRead`, a `QUERY` request defers here (`method !== 'get'`), skipping the entry check on the promise that instance `search` will consume `checkPermission` and arm the per-record guard. But the `query` static action calls `resource.search(data, query)` (Resource.ts:289) and `search(target)` takes a single arg, so `target` = `data` (the request body), while `checkPermission` was set on the URL `target` (REST.ts:167). `Table.search`'s `if (target.checkPermission)` guard (Table.ts:2655) is never entered: no entry check (deferred) and no per-record filtering → the full unfiltered result set is returned to a user who should see only a subset. The collection GET path is safe because it routes via `constructor.search(target, ctx)` (Table.ts:1195), passing the checkPermission-bearing target as search's sole arg.

Suggested fix: thread `checkPermission` onto the object `search` treats as its target on the QUERY path (e.g. in the `query` action set `data.checkPermission ??= query.checkPermission` before `resource.search(data, query)`), and add a QUERY-verb regression test next to the collection-read cases.

—
_Generated by Barber AI_

resources/Resource.ts
```diff
@@ -733,6 +739,44 @@ function transactional(
...
+						(resource.constructor as any)?.supportsRowLevelAllowRead &&
+						!(resource.allowRead as any)?.isDefaultAllowRead &&
+						(resource.allowRead as any)?.constructor?.name !== 'AsyncFunction'
```
cb1kenobi: **High: async `allowRead` override silently fails OPEN on collection reads and subscriptions**

Excluding `AsyncFunction` here (and at Table.ts:2652 / :3435) means an async override never defers and never gets a per-record/per-event guard — it only runs the entry check, at COLLECTION scope (no record loaded). But the recommended composition pattern documented in this PR returns `true` when `owner == null` (integrationTests/routes/subscription-row-allowread/resources.js) precisely so the connection/scan opens and relies on per-record filtering. An async version of that same pattern therefore returns permissive at entry; `checkPermission` is then cleared (Resource.ts:800) before `search`/delivery run, so no per-record or per-event filtering happens — the whole scan/subscription is delivered unfiltered, with no warning. The design comment's claim that async "fails closed" holds only for overrides that return falsy at collection scope — the opposite of the pattern this PR recommends.

Suggested fix: detect an overridden async `allowRead` on a `supportsRowLevelAllowRead` table and reject it (or warn loudly) at definition/first-use — row-level `allowRead` must be synchronous — rather than silently degrading to a permissive collection-scope check.

—
_Generated by Barber AI_

resources/Table.ts
cb1kenobi: **Medium: subscription delivery evaluates `allowRead` on an UNFROZEN shared cached record**

The query path freezes every record before handing it to the override (`freezeRecord` in loadLocalRecord Table.ts:4769 and composeRecordFilter search.ts:205) "to ensure the integrity of the cache." This delivery path does not: `event.value` is `entry.value` from `primaryStore.getEntry(id)` (Table.ts:3485) — the shared cached object — passed straight in as `this`. An override that writes to `this` (redaction, memoization, a computed comparison field) mutates the shared primaryStore record for every other reader on the node, and behaves differently here than on the frozen query path (a genuine footgun given the query path deliberately freezes).

Suggested fix: mirror the query path — `freezeRecord(event.value)` before invoking the override in `allowsEvent`.

—
_Generated by Barber AI_

resources/Resource.ts
kriszyp: Fixed in f68461173 (HIGH, confirmed): the `query` action now threads `query.checkPermission` onto the body `data` (the object search treats as its target on the table path), so Table.search arms the per-record guard. Added a QUERY-verb regression test — verified it returns the full unfiltered set without the fix and row-filtered results with it. Great catch. — KrAIs (Claude)

resources/Resource.ts
kriszyp: Fixed in f68461173, with an adjustment: hard-rejecting async would regress the shipped #1422 async fail-closed contract (the ThrowsAsync fixture is a legit async table-level check). Instead, an async override now runs ONLY the awaited entry check (table/connection scope — same as before this feature) and gets NO per-record/per-event filtering, so it must decide fully at collection scope; the collection-permissive 'open, filter per record' pattern requires a sync override. A one-time warning is logged so the async-collection-permissive fail-open is no longer silent, and the design comment is corrected. Whether to *hard-reject* async row-level overrides is a design call I've flagged for Kris — this keeps #1422 intact. — KrAIs (Claude)

resources/Table.ts
kriszyp: Fixed in f68461173: delivery now evaluates against the LIVE `subContext.user` (which #1414 re-auth updates in place), not the subscribe-time snapshot, so a role/claims change is reflected per event. The `filterRowReads` gate no longer keys on user presence — the wrapper stamps a durable `rowLevelAuthChecked` flag on the checked request, so an anonymous-but-authorized subscription is still filtered while internal subscribers (no flag) keep full delivery. — KrAIs (Claude)

kriszyp: Fixed in f68461173: `freezeRecord(event.value)` before invoking the override in `allowsEvent`, mirroring the query path — so an override that writes to `this` can't mutate the shared primaryStore record. Added a regression test asserting the record is frozen during delivery (verified it fails without the freeze). — KrAIs (Claude)

resources/Table.ts
```diff
@@ -3362,6 +3410,53 @@ export function makeTable(options) {
...
+						if (event.type === 'end_txn' || event.type === 'reload') return true; // control markers, no record
+						// Freeze the record before the override sees it, mirroring the query path — event.value
+						// is the shared primaryStore object, so an override that writes to `this` would otherwise
+						// mutate the cache for every reader on the node.
+						if (event.value) freezeRecord(event.value);
+						let decision;
+						try {
+							// Evaluate against the LIVE context user, not a subscribe-time snapshot: #1414 re-auth
+							// updates subContext.user in place, so a role/claims change must be reflected per event.
+							decision = subAllowRead.call(event.value ?? null, subContext.user, request, subContext);
```
cb1kenobi, line 3493: **Nit: the live-user re-auth fix has no regression test**

Good fix — delivery now evaluates the override against the live `subContext.user` (reflecting #1414 in-place re-auth) instead of a subscribe-time snapshot, and the gate keys on `rowLevelAuthChecked` so an anonymous-but-checked subscription still filters. But no test exercises the scenario this closes: the added subscription fixtures key on a stable field (`ownerId`/`username`), so a role/claims change mid-subscription would not change any verdict and a silent regression back to the snapshot would pass green. Consider a delivery test where the override reads a mutable claim (e.g. `user.role`), the row is initially visible, then re-auth mutates `subContext.user` to a role that should hide the row, asserting the next event is dropped.

—
_Generated by Barber AI_

kriszyp: Added in 338969abe: a delivery test whose override gates on a mutable claim (`user.tier`), delivers an event to a gold-tier user, then swaps `context.user` to a silver tier mid-stream and asserts the next event is dropped. Verified it fails if the filter reads a captured subscribe-time snapshot instead of the live `subContext.user`. Thanks. — KrAIs (Claude)

cb1kenobi (approved):

heskew (commented):
AI-assisted review by Codex (GPT-5).

resources/Resource.ts
```diff
@@ -283,6 +284,20 @@ export class Resource<Record extends object = any> implements ResourceInterface<
...
+			// its sole target, while checkPermission was set on the URL `query`. Thread it across so
+			// Table.search sees the flag and enforces row-level allowRead — otherwise a QUERY on a table
+			// with an overridden allowRead returns the full unfiltered set (the deferral skips the entry
+			// check on the promise that search consumes checkPermission, which it never sees).
+			if (
+				resource.constructor.loadAsInstance !== false &&
+				data &&
+				typeof data === 'object' &&
+				(query as any)?.checkPermission != null &&
+				(data as any).checkPermission == null
```
heskew: `data` is the client-controlled QUERY body, so it can include `checkPermission: false`. Because this only fills nullish values, that disables `Table.search`'s guard after the wrapper deferred the entry check, returning rows the override denies. Please make this field framework-owned (ignore or overwrite the body value) and cover `false` in the regression test.

resources/Table.ts
```diff
@@ -2854,16 +2879,42 @@ export function makeTable(options) {
...
+			// per record with `this` = the (frozen) record, participating in query execution — pushed
+			// into HNSW traversal for vector sorts, applied as a post-filter otherwise. It must be
+			// synchronous, side-effect free, and fast; a throw denies that record (fail closed, #1422
+			// gap 1 parity). Dispatched via the method resolved from the resource — never via a
+			// `record.allowRead` property lookup, so a record attribute named allowRead can't shadow it.
+			let recordGuard: ((record: any) => boolean) | undefined;
+			if (recordScopedAllowRead) {
+				const user = (context as any)?.user;
+				let warnedAsync = false;
+				recordGuard = (record: any) => {
```
heskew, line 2935: This guard is also called after source revalidation. That path receives `getFromSource`'s immediately resolved record—the same object its deferred commit later mutates and encodes—and it has not passed through `freezeRecord`. An `allowRead` that writes to `this` can therefore alter persisted data (notably with LMDB). Please evaluate the guard against a frozen view without freezing the commit object.

resources/Table.ts
```diff
@@ -2854,16 +2879,42 @@ export function makeTable(options) {
...
+					let allowed;
+					try {
+						allowed = recordScopedAllowRead.call(record, user, target, context);
+					} catch {
+						return false; // fail closed on a throwing check
+					}
+					if (typeof allowed?.then === 'function') {
```
heskew, line 2942: A rejected thenable is denied here but its rejection is never observed, so it reaches Harper's global `unhandledRejection` logger once per candidate. The subscription `allowsEvent` path has the same issue. Please attach a rejection handler before returning `false`; the existing one-time warning can remain the diagnostic.

resources/Table.ts
```diff
@@ -3490,13 +3588,17 @@ export function makeTable(options) {
...
 										id,
 										localTime: auditRecord.localTime,
 										value,
 										version: auditRecord.version,
 										type: auditRecord.type,
-									});
+									};
+									// Filter denied rows BEFORE they consume a previousCount slot, so an authorized
+									// subscriber still receives up to `count` readable events (#1419).
+									if (allowsEvent && !allowsEvent(historyEntry)) continue;
```
heskew, line 3669: Filtering before decrementing `count` removes the `previousCount` work bound: if the policy denies every row, this walks the entire retained audit log while live events accumulate in `pendingRealTimeQueue`. Please cap inspected entries independently and stop when the subscription closes, even if that returns fewer than `count` authorized events.

resources/Resource.ts
kriszyp: Fixed in 5444791b6: `checkPermission` is now framework-owned in the QUERY wrapper (`Resource.query` in resources/Resource.ts) — always overwritten (or deleted when unset) instead of only filled when nullish, so a client-supplied `checkPermission: false` in the body can no longer survive through to `Table.search`. Added a regression test asserting exactly this bypass attempt: integrationTests/security/query-row-allowread-checkpermission.test.ts.

resources/Table.ts
kriszyp: Fixed in 5444791b6: the record-scoped allowRead guard on the source-revalidation path now evaluates a read-only view of the record (new `frozenRecordView` in resources/Table.ts) instead of the live object the deferred commit still mutates and persists. I initially tried a frozen shallow copy, but a cross-model review (Gemini) correctly flagged that it would eagerly invoke lazy-decode getters (breaking a caching table's structPrototype optimization) and mishandle Array/Date/Map/Set roots (a cached record can legitimately be any of these — getFromSource only requires `typeof === 'object'`). Switched to a write-blocking Proxy that forwards reads transparently (getters run with `this` bound to the real record) and falls back to the unwrapped record for Date/Map/Set/etc., which carry internal slots that throw on non-original receivers. Covered by a dedicated unit test: unitTests/resources/frozen-record-view.test.js.

resources/Table.ts
kriszyp: Fixed in 5444791b6: both the query-traversal guard's recordGuard and the subscription allowsEvent filter now attach `.then(undefined, () => {})` before treating a rejected thenable as denied (using `.then`, not `.catch` — a cross-model review pointed out a generic thenable is only guaranteed to have `.then`, and `.catch` would throw synchronously on one that lacks it). The one-time diagnostic warning is unchanged. Regression test: integrationTests/security/rejected-thenable-allowread-no-unhandled.test.ts, covering both call sites with a fixture whose allowRead returns `Promise.reject(...)`.

resources/Table.ts
kriszyp: Fixed in 5444791b6: the previousCount history backfill now bounds audit records INSPECTED (in-scope for the target table/id) independently of `count` (events AUTHORIZED) via a new `MAX_PREVIOUS_COUNT_SCAN` cap (10,000), and stops early if the subscription closes mid-scan (checked alongside the existing yield point). The inspected counter only increments once a record is confirmed in-scope (right table, right id) so unrelated cross-table/cross-scope audit traffic in a busy shared log can't spuriously cut a legitimate backfill short. Logs a one-time diagnostic when the cap is what stopped the scan. Regression test: unitTests/resources/subscriptionPreviousCountScanBound.test.js (seeds past the cap with an all-deny override, asserts the scan stops at the cap rather than walking the full log).

Issue #1766, opened 2026-07-12
static: root mount (urlPath: '/') matches only the exact path '/' — all sub-paths 404 (5.1.x)


heskew:
## Summary

Apps configuring the built-in `static` plugin with a root mount (`urlPath: '/'`) serve only the exact path `/` — every sub-path (`/index.html`, `/assets/*`, …) bypasses the static handler and falls through to the REST router's 404. Reported from production instances after auto-upgrade to 5.1.18 from 5.0.x, where this config worked; it broke every deployed app using root-mounted static assets with an explicit `urlPath: '/'`.

Sibling of #1583 / #1730: #1584 fixed the subpath mounts (`urlPath: 'app'`, `'/app'`) but not the root mount.

## Root cause

`server/middlewareChain.ts`:

1. `normalizeUrlPath('/')` returns `'/'` (truthy), so a root-mounted handler is grouped as a **sub-route** in `resolveRoutedChains` instead of joining the default chain.
2. `matchesRoute` prefix-matches with a segment boundary:
   ```ts
   if (pathname !== urlPath && !pathname.startsWith(urlPath + '/')) return false;
   ```
   For `urlPath === '/'` the boundary check builds `'//'`. No real pathname starts with `'//'`, so only the exact request `/` matches; all sub-paths skip the chain.
3. Latent second defect: if a `'/'` route did match a sub-path, `stripPrefix(request, '/')` would slice one character off the pathname (`'/index.html'` → `'index.html'`), producing a slash-less path.

The mount value comes from the Scope server proxy (`components/Scope.ts`): `urlPath: scopeConfig.urlPath ? resolveBaseURLPath(pluginName, scopeConfig.urlPath) : undefined`, and `resolveBaseURLPath(name, '/')` → `'/'`.

## Affected versions

Broken since v5.1.0 (9dcf7b494, "Add urlPath/host routing with per-route middleware chains" — the Scope proxy began passing the configured `urlPath` as a route constraint). Present on `main` and the v5.1 line including 5.1.18. Worked in 5.0.x, where static registered its handler unconditionally and matched the full request path against its own maps. Not covered by #1584's integration tests (subpath mount only).

## Repro

```yaml
static:
  files: 'web/**'
  urlPath: '/'
```

- `GET /` → 200 (index)
- `GET /index.html` → 404
- `GET /<anything>` → 404

Removing `urlPath` (unset) works — the breakage is specifically the explicit root value.

## Fix direction

[...]

PR #1791, opened 2026-07-14
test(server): promote validated QA regression anchors (static, deploy, shutdown-drain, secrets)

kriszyp:
## Summary

Promotes 4 exploratory-QA-validated regression anchors to the tracked integration suite. No product code changes — test-only.

| Candidate | Suite | Guards |
|---|---|---|
| P-326 | `components/static-after-rest.test.ts` | REST stays reachable under `fallthrough:false` + `after:'rest'` SPA config; auth denials not swallowed by static catch-all; misconfiguration warning fires when `after:'rest'` is omitted. harper#1574. |
| P-328 | `components/secret-audit-leak.test.ts` | `set_secret` (create + rotate) never surfaces plaintext via `read_audit_log` (blocked 403) or generic system-table reads (`enc:v1:` envelope only). harper#715. |
| P-329 | `deploy/deploy-dangling-symlink.test.ts` | `package_component` + `deploy_component` succeed past a dangling symlink positioned early in directory walk order; all entries after it land in the archive and deploy. harper#1718. |
| P-330 | `components/shutdown-drain-e2e.test.ts` | In-flight work registered via `ShutdownDrain` survives a real worker restart; stalled drains are force-killed at the configured ceiling (`replication.blobSendDrainTimeout`). PR#1621. |

P-350 (qa551 static root-mount / harper#1766) was **dropped**: already covered by the existing "static plugin with root urlPath (#1766)" suite in `components/static-urlpath.test.ts`.

All 4 promoted tests passed twice on `3b59214bb` (v5.2.0-alpha.5) before promotion. One fixture file (`registerCustody.js` for P-328) required a depth-correction fix for the `dist/` relative path when moving from `qa-scratch/` (3 levels deep) to `components/fixtures/` (4 levels). The `static-after-rest-misconfig` fixture was reconstructed from the warning-trigger logic in `server/static.ts` (the snapshot only packaged the main fixture dir).

Generated by Claude Sonnet 4.6.

### DIFF
--- a/integrationTests/components/fixtures/shutdown-drain-e2e/resources.js
+++ b/integrationTests/components/fixtures/shutdown-drain-e2e/resources.js
@@ -24,7 +24,8 @@ const nativeRequire = createRequire(import.meta.url);
 const { registerShutdownDrain } = nativeRequire(process.env.QA519_SHUTDOWN_DRAIN_ABS_PATH);
 
 const MARKER_FILE = process.env.QA519_MARKER_FILE;
-const TASK_DELAY_MS = Number(process.env.QA519_TASK_DELAY_MS || 2500);
+const parsedTaskDelay = Number(process.env.QA519_TASK_DELAY_MS);
+const TASK_DELAY_MS = Number.isInteger(parsedTaskDelay) && parsedTaskDelay > 0 ? parsedTaskDelay : 2500;
 
 function log(tag) {
 	try {
--- a/integrationTests/components/fixtures/shutdown-drain-e2e/resources.js
+++ b/integrationTests/components/fixtures/shutdown-drain-e2e/resources.js
@@ -1,0 +1,93 @@
+// QA-519 — end-to-end verification of the shutdown-drain mechanism
+// (components/shutdownDrain.ts, commit 9018760e4, PR #1621).
+//
+// The shipped unit tests (unitTests/components/shutdownDrain.test.js) only exercise the pure
+// functions with fake drain objects — never a real worker, a real SHUTDOWN, or real in-flight
+// work. This fixture registers a REAL ShutdownDrain inside an actual HTTP worker thread and
+// gives the test a way to observe, via a marker file, exactly when the worker's simulated
+// in-flight task finished relative to when the worker itself exited.
+//
+// To register into the SAME per-worker `drains` registry that
+// `server/threads/threadServer.js` reads from (module state there is a plain in-memory `Set`,
+// naturally scoped per worker thread — see the doc comment on shutdownDrain.ts), this file must
+// land in Node's REAL native CJS `require` cache, not Harper's sandboxed VM module loader's
+// private per-component cache (which would produce an isolated copy of the module with its own
+// empty `Set`). `createRequire(import.meta.url)` escapes the sandbox for exactly this file,
+// giving a genuine native `require` bound to this file's real location; requiring the absolute
+// path to the compiled `dist/components/shutdownDrain.js` through it resolves to the exact same
+// cache entry `threadServer.js` itself populated at boot.
+import { createRequire } from 'node:module';
+import { threadId } from 'node:worker_threads';
+import { appendFileSync } from 'node:fs';
+
+const nativeRequire = createRequire(import.meta.url);
+const { registerShutdownDrain } = nativeRequire(process.env.QA519_SHUTDOWN_DRAIN_ABS_PATH);
+
+const MARKER_FILE = process.env.QA519_MARKER_FILE;
+const TASK_DELAY_MS = Number(process.env.QA519_TASK_DELAY_MS || 2500);
+
+function log(tag) {
+	try {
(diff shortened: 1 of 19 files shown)


integrationTests/components/fixtures/shutdown-drain-e2e/resources.js
```diff
@@ -0,0 +1,93 @@
...
+// cache entry `threadServer.js` itself populated at boot.
+import { createRequire } from 'node:module';
+import { threadId } from 'node:worker_threads';
+import { appendFileSync } from 'node:fs';
+
+const nativeRequire = createRequire(import.meta.url);
+const { registerShutdownDrain } = nativeRequire(process.env.QA519_SHUTDOWN_DRAIN_ABS_PATH);
+
+const MARKER_FILE = process.env.QA519_MARKER_FILE;
+const TASK_DELAY_MS = Number(process.env.QA519_TASK_DELAY_MS || 2500);
```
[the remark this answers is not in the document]
kriszyp: Good catch — fixed in a14a8b1e4. `QA519_TASK_DELAY_MS` is now parsed with `Number.isInteger(...) && > 0` before use (falls back to 2500 for unset/empty/whitespace/non-numeric/zero/negative), matching the existing convention in `integrationTests/database/recordCachingWorkers.ts`.

— KrAIs (Claude Sonnet 5)

integrationTests/components/static-after-rest.test.ts
```diff
@@ -0,0 +1,218 @@
...
+			await setupHarperWithFixture(ctx, MISCONFIG_FIXTURE_PATH, {
+				config: { logging: { console: true, level: 'warn' } },
+				env: {},
+			});
+			procOutput += ctx.harper.startupOutput?.stdout ?? '';
+			procOutput += ctx.harper.startupOutput?.stderr ?? '';
+			const proc = ctx.harper.process;
+			proc?.stdout?.on('data', (d: Buffer) => (procOutput += d.toString()));
+			proc?.stderr?.on('data', (d: Buffer) => (procOutput += d.toString()));
+			await sleep(2_000);
```
cb1kenobi: **Nit: misconfig-warning check relies on a fixed sleep and a weak anchor**

The other suites in this PR poll for readiness; this one does a fixed `sleep(2_000)` then matches `/after: 'rest'/`. Two soft spots: on a slow CI the 2s window can miss late-flushed stdout (flake), and `after: 'rest'` is a generic fragment that could match incidental output rather than the actual warning. Consider polling `procOutput` until a distinctive fragment of the real message appears (e.g. `answers every unmatched GET`, from `server/static.ts#warnIfBlockingRest`) with a deadline, matching the polling style used elsewhere here.

—
_Generated by Barber AI_

integrationTests/components/static-after-rest.test.ts
```diff
@@ -0,0 +1,218 @@
...
+		const res = await fetch(new URL('/some/totally/unmatched/client-route', restURL));
+		const text = await res.text();
+		strictEqual(res.status, 200, `expected the SPA notFound fallback to serve index.html: ${res.status} ${text}`);
+		ok(text.includes('spa-shell-marker'), `expected the SPA shell body, got: ${text}`);
+	});
+
+	test('auth ordering: headerless GET is auto-authorized as super_user in the test harness (loopback bypass, not a #1574 defect)', async () => {
+		// AUTHENTICATION_AUTHORIZELOCAL is unconditionally true in the test harness — loopback
+		// requests bypass auth as super_user. This assertion documents that behavior rather than
+		// testing #1574's auth-ordering property; PROBE 3c is the real check.
```
cb1kenobi: **Nit: stale `PROBE 3c` cross-reference**

This comment points the reader to "PROBE 3c" as the real auth-ordering check, but no PROBE labels survive in this file (they were QA-scratch numbering). The real check is the ungranted-Bob test below. Suggest naming that test directly (or dropping the dangling reference) so the pointer doesn't dead-end. The assertion itself is fine — it documents the harness loopback bypass, just not #1574's property.

—
_Generated by Barber AI_

integrationTests/components/secret-audit-leak.test.ts
```diff
@@ -0,0 +1,153 @@
...
+			const deadline = Date.now() + 30_000;
+			while (Date.now() < deadline) {
+				const r = await client.req().send({ operation: 'get_secrets_public_key' });
+				if (r.status === 200 && r.body?.fingerprint) break;
+				await sleep(250);
+			}
+		});
+
+		after(async () => {
+			console.log('\n===== secret-audit-leak findings =====');
```
cb1kenobi: **Nit: debug scaffolding left in the promoted test**

The `findings[]` accumulator and this `console.log` dump in `after()` are exploratory-QA scaffolding — they add CI log noise without asserting anything. The plaintext-leak coverage all lives in the `ok(!...includes(MARKER))` assertions, which stand on their own. Consider dropping the accumulator now that the anchor is tracked.

—
_Generated by Barber AI_

Issue #1796, opened 2026-07-14
get_analytics scans the full metric history instead of the requested time window on large hdb_analytics tables (condition reordering misestimates primary-key ranges)


kylebernhardy:
### Bug Summary

`get_analytics` for a narrow time window scans the **entire history of the requested metric** on instances with a large `hdb_analytics` table, because condition reordering costs an `equals` on the indexed `metric` attribute below a `between` on the primary key. On long-lived clusters this makes every `get_analytics` call a multi-million-row scan that never finishes from the client's perspective — and stacked scans (each pinning a RocksDB read snapshot) drive the instance into restarts.

### Steps to Reproduce

1. Take an instance with a large `system.hdb_analytics` table — e.g. a multi-node cluster upgraded from 4.6 that has been aggregating for months (default `AGGREGATE_EXPIRATION` retains **one year**; the observed cluster is on 5.1.18; the table from #1538 held ~19.7M rows). Any metric comprising <10% of the table's rows exhibits the bug.
2. Issue the query Studio's Status page sends:
   ```json
   {
     "operation": "get_analytics",
     "metric": "utilization",
     "start_time": <now - 5min>,
     "end_time": <now>
   }
   ```
3. Watch the request grind for tens of seconds to minutes (chunked response already started, so the client sees a stream that never completes rather than an error). Concurrent/retried calls stack further scans; the instance eventually exhausts memory/CPU and restarts.

### Expected Behavior

The 5-minute window should drive the scan: a `between` on the primary key (`id` = `[time, nodeId]`) bounds iteration to a few hundred rows regardless of table size. Milliseconds, like on a fresh instance.

### Actual Behavior

`get()` in `resources/analytics/read.ts` builds conditions as `[metric equals X, id between [start, end]]`, and `Table.search` reorders conditions by estimated cost (`resources/Table.ts:2691`, v5.1). The estimates (`resources/search.ts`) are asymmetric:

- `metric equals X` → **actual row count** from the dupSort index (`index.getValuesCount(X)`)
- `id between …` → **fixed guess**: `BETWEEN_ESTIMATE = 0.1` × estimated table entry count — regardless of how narrow the range is

So whenever a metric's total row count is below 10% of the table (true for most individual metrics), the planner drives iteration off the **metric index**, decoding every row of that metric ever recorded and post-filtering by the time window. The plan inverts exactly on the instances where it hurts most: the bigger the table, the more likely the full-metric scan wins.

[...]

PR #1802, opened 2026-07-14
test: promote 4 QA regression anchors (audit-completeness, overtime-atomicity, TTL-rate-limiter, planner-range-scan)

kriszyp:
Promotes 3 QA-validated **green regression anchors** into the tracked integration suite (test-only, no product change):

- **`resources/overtime-multi-write-atomicity.test.ts`** — a custom-resource handler with pre-await writes → real async gap past `maxTransactionOpenTime` → post-await writes must CLEAN-rollback (422, 0 durable, 0 phantom), not silent partial-drop. Guards #1407 / PR #1411. Complements the simpler existing `txn-overtime-atomicity` (adds pre-rows, marker, index-consistency, baseline).
- **`resources/ttl-rate-limiter-concurrent.test.ts`** — 50 concurrent `addTo` on one 500ms-TTL key (4 workers) yields exact stored==acked==50, clean create-from-nothing, correct reset-on-update TTL. Guards the F-002/F-005/F-006 atomic+eviction class.
- **`database/planner-range-scan.test.ts`** — a `PK between + low-selectivity indexed equals` query on a user table stays PK-range-cost (flat across table growth, invariant to condition order). Guards against a future generalization of the #1796 `get_analytics` planner over-scan.

**Note:** a 4th candidate (`audit-retry-completeness`, #1773) was pulled from this batch — it drops ~1/1600 audit entries under heavier CI conflict-retry contention (passed local cold-rerun). It's under investigation as a possible real residual audit-drop vs. a load-sensitive test; will re-promote if it's a test issue, or file separately if it's a real defect.


integrationTests/database/planner-range-scan.test.ts
```diff
@@ -0,0 +1,280 @@
...
+			console.log(`  tier=${String(e.tier).padStart(7)} ${e.label.padEnd(38)} ${String(e.ms).padStart(7)}ms`);
+
+		// ── Verdict ──────────────────────────────────────────────────────────────
+		const combined = timingLog.filter((e) => e.label === 'combined (author order) SQL');
+		const pkOnly = timingLog.filter((e) => e.label === 'PK-only control SQL');
+		const kindOnly = timingLog.filter((e) => e.label === 'kind-only control SQL');
+		const pkOnlyOps = timingLog.filter((e) => e.label === 'PK-only control ops');
+		const kindOnlyOps = timingLog.filter((e) => e.label === 'kind-only control ops');
+		const combinedOps = timingLog.filter((e) => e.label === 'combined (author order) ops');
+		const growth = (arr: typeof combined) => arr[arr.length - 1].ms / Math.max(1, arr[0].ms);
```
cb1kenobi, line 201: **Medium: the named over-scan regression cannot fail this test — no timing/scaling assertion**

This anchor guards a planner *cost* defect (the combined query over-scanning the whole table instead of the narrow PK range), but every assertion in the file is a row-count correctness check. The growth ratios computed here in `after()` are only `console.log`-ed — nothing asserts that combined-query time tracks the PK-only control rather than the kind-only control. A regressed planner that scans all N rows still returns the correct ~90 rows, so the suite stays green and the exact regression this file is named for goes uncaught.

Consider asserting the scaling trend with a generous tolerance so it is robust to CI noise but still fails on an over-scan, e.g. combined growth should track PK-only (flat) not kind-only (~linear):
```js
ok(growth(combinedOps) < growth(kindOnlyOps) * 0.5, `combined tracks kind-only over-scan: ${growth(combinedOps)}x`);
```
Without a trend assertion this is a combined-query correctness test, not the over-scan anchor its name implies.

—
_Generated by Barber AI_

integrationTests/resources/ttl-rate-limiter-concurrent.test.ts
```diff
@@ -0,0 +1,469 @@
...
+					skipRounds++;
+					continue;
+				}
+
+				// Wait TTL + buffer so the record is definitely gone.
+				await sleep(TTL_MS + 300);
+
+				// Verify 404 before proceeding (skip if still alive — timing).
+				const checkRes = await get(id);
+				if (checkRes.status !== 404) {
```
cb1kenobi, line 294: **Medium: a broken-eviction regression is skipped here, not caught**

This leg (the docstring's "primary probe") only runs a round when the key reads back `404` after expiry; otherwise it `skip`s. The only end-of-test assertions are `lostCountRounds === 0` and `staleRounds === 0`. If TTL eviction itself regresses so records never expire, every round takes this skip branch → `skipRounds === ROUNDS`, both counters stay `0`, and the test passes green — silently failing to catch exactly the F-005 eviction class it is meant to guard. The same vacuous-pass shape exists in legs (2)/(4)/(5): an all-skip or all-error run leaves the `=== 0` assertions trivially true.

Add a lower-bound guard that a meaningful number of rounds actually exercised the path, e.g.:
```js
ok(cleanRounds + lostCountRounds + staleRounds >= ROUNDS / 2, `too many rounds skipped (${skipRounds}/${ROUNDS}) — eviction/timing may be broken`);
```

—
_Generated by Barber AI_

integrationTests/resources/overtime-multi-write-atomicity.test.ts
```diff
@@ -0,0 +1,310 @@
...
+		{ timeout: 60_000 },
+		async () => {
+			const tag = 'overtime-main';
+
+			const t0 = Date.now();
+			const res = await postJSON('/Overtime/', { tag, count: PRE_ROWS, holdMs: HOLD_MS });
+			const elapsed = Date.now() - t0;
+
+			// Give any async side effects (force-commit, abort, index updates) time to settle.
+			await sleep(600);
```
cb1kenobi: **Nit: fixed settle-sleep instead of polling final state**

This 600ms `sleep` to let force-commit/abort/index side-effects settle is a fixed wait; under CI load those async effects can outlast it, making the atomicity assertions flaky. Since the oracle is the final table state, prefer polling `DumpAtomic` until the row set is stable (or for a bounded number of tries) rather than a single fixed delay. Same pattern at the 100ms settle in the TTL test's from-nothing leg.

—
_Generated by Barber AI_

integrationTests/database/planner-range-scan.test.ts
kriszyp: Addressed in 3892e6485: added a hard scaling-trend assertion in `after()` on the `search_by_conditions` (ops) path — combined-query growth (both condition orders) must stay well below kind-only's growth, tracking the flat PK-only control instead. A regressed planner scanning the whole `kind='common'` set instead of the PK window would now fail this trend check even though row counts still come back correct.

I scoped the hard assertion to the ops path specifically: repeated local runs (median-of-7 samples per point, to cut through scheduler/compaction jitter) show the `ops` PK-only control is genuinely flat across tiers, but the SQL path carries a reproducible baseline cost that scales with table size regardless of selectivity — even the SQL PK-only control (fixed 100-row window) grows ~3x when the table grows 3x. An SQL-side growth-ratio comparison can't reliably discriminate given that shared confound, so SQL timings stay logged for visibility but aren't hard-asserted on, to avoid a chronically noisy gate. Also added `timedMedian()` (7 reps) plus an untimed warm-up per tier, since single-shot timing at these row counts swung >2x run-to-run in my testing.

— Claude Sonnet 5

integrationTests/resources/ttl-rate-limiter-concurrent.test.ts
kriszyp: Addressed in b19a63b69. Legs (2), (3), (5): added a lower-bound guard (`ok(clean + lost + stale >= ROUNDS/2, ...)`) requiring at least half the rounds/windows were actually measurable, so an all-skip run (e.g. TTL eviction regressed to never expire, or to evict too aggressively) fails loudly instead of vacuously passing on 0/0 — matches the pattern you flagged for leg (3) exactly, extended to (2)/(5) per your "same shape" note.

Leg (4) (expiry-mid-burst) needed more digging: I traced the write path (`Table.ts` `_writeUpdate` commit handler) and confirmed every successful write — including `addTo` — resets a record's `expiresAt` to `now+TTL`. That means a back-to-back burst never gives the key an actual gap to expire through, so `cleanExpiry` is provably 0/10 even under fully-correct eviction code (verified empirically too). Asserting `cleanExpiry>0` there would fail permanently regardless of product code, so I left it as a logged, non-asserted note instead, and filed a Findings entry for a follow-up redesign of that leg's timing (e.g. a genuine no-write gap) if a real boundary-straddle probe is wanted. Legs (2)/(3)/(5) plus the sibling `ttlResetOnWrite.test.ts` (QA-269) are what actually exercise the eviction-disabled regression class today.

Also folded in your nit from the sibling thread: leg (3)'s 100ms fixed settle-sleep is now `waitForStableGet()`, polling until two consecutive reads agree.

— Claude Sonnet 5

integrationTests/resources/overtime-multi-write-atomicity.test.ts
kriszyp: Addressed in 31f88b9af: both the overtime-atomicity anchor's fixed 600ms settle-sleeps are now `waitForStableRows()`, which polls `DumpAtomic` until two consecutive reads of the tag's row set agree (bounded to 20 tries × 100ms). Also applied the same pattern to the TTL test's from-nothing leg's 100ms settle (see reply on that thread / b19a63b69) — `waitForStableGet()` there.

— Claude Sonnet 5

integrationTests/resources/ttl-rate-limiter-concurrent.test.ts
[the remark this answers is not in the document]
kriszyp: Duplicate of cb1kenobi's thread on this same file — addressed in b19a63b69 (lower-bound guards on legs 2/3/5; leg 4 has a different fix, see that thread for details).

integrationTests/resources/overtime-multi-write-atomicity.test.ts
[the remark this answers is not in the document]
kriszyp: Duplicate of cb1kenobi's thread on this file — addressed in 31f88b9af (waitForStableRows, both occurrences) and b19a63b69 (waitForStableGet in the TTL from-nothing leg).

integrationTests/resources/ttl-rate-limiter-concurrent.test.ts
```diff
@@ -0,0 +1,469 @@
...
+			log(
+				`\n[QA-431] (3) FROM-NOTHING BURST (${ROUNDS} rounds × N=${N}, TTL=${TTL_MS}ms, ${WORKERS} workers)\n` +
+					`  clean rounds                  = ${cleanRounds}\n` +
+					`  lost-count rounds             = ${lostCountRounds}  ← DEFECT if > 0\n` +
+					`  over-count rounds (seed=99)   = ${staleRounds}  ← DEFECT: stale resurrection if > 0\n` +
+					`  skipped (expiry-timing)        = ${skipRounds}\n` +
+					`  ${roundLogs.join('\n  ')}`
+			);
+
+			ok(lostCountRounds === 0, `QA-431(3): ${lostCountRounds} round(s) with lost increments on from-nothing burst`);
```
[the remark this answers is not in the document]
kriszyp: Duplicate of the other two threads flagging this same leg — addressed in b19a63b69.

heskew (approved):
Test-only, 3 non-vacuous regression anchors (each fails without its fix), noise handled via median/polling rather than bare sleeps. Note: title says 4 but only 3 are promoted — the audit-retry anchor was withdrawn per the body; worth fixing the title. LGTM.

Issue #1811, opened 2026-07-14
Table background work (expiration timer, delete/cleanup callbacks) not disposed when a database's stores are closed


cb1kenobi:
## Summary

When a database's stores are closed and later reloaded, per-table background work is not torn down. This is pre-existing (it predates the RocksDB backup/restore work) and was surfaced by a cross-model review of that change.

## Details

Two code paths close and reload a loaded database in place:

- `resetDatabases()` in `resources/databases.ts` — invoked on **every schema change** via the ITC schema handler (`server/itc/serverHandlers.js`).
- `closeDatabase()` / `closeLoadedDatabases()` in `resources/databases.ts` — the `restore_backup` flow and job-worker exit.

Neither disposes table-level background work before/after closing the stores. In particular:

- The record-expiration eviction timer — `setInterval(...)` in `runRecordExpirationEviction` (`resources/Table.ts`, ~line 5537, thread 0 only) — is not stored in a handle and never cleared.
- Delete-removal / cleanup callbacks registered against the audit store (`addDeleteRemoval` → `auditStore.addDeleteRemovalCallback`, `resources/Table.ts` ~line 5528) are not deregistered on close.

## Impact

After a close, these timers/callbacks keep referencing the now-closed stores (erroring on the next tick, or silently retaining them); a subsequent reload can register a second set. Bounded, but real — most likely to accumulate on instances with frequent schema changes and tables that use expiration/eviction.

## Proposed fix

Add an idempotent table-disposal method that clears the table's timers/intervals and deregisters its delete/cleanup callbacks, and invoke it before closing the stores. Wire it into both `resetDatabases()` and `closeDatabase()` so both close paths are covered.

## References

- `resources/Table.ts` — `runRecordExpirationEviction` (expiration interval), `addDeleteRemoval` (delete-removal callback)
- `resources/databases.ts` — `closeDatabase`, `closeLoadedDatabases`, `resetDatabases`

_Surfaced via cross-model review (codex) of the RocksDB backup/restore change; filed by Claude._

PR #1831, opened 2026-07-16
Add RocksDB managed backup and restore operations

cb1kenobi:
## Summary
Adds managed backup/restore for RocksDB databases as Operations API operations — `create_backup`, `list_backups`, `verify_backup`, `delete_backup`, `purge_backups`, `restore_backup` — plus RocksDB support for `get_backup` (streams a gzipped-by-default `tar`; `gzip=false` for plain `tar`; works against a remote `target=`). Each is also runnable from the CLI under its operation name, and offline against a stopped server. Managed backups are an incremental, checksum-verified, server-side repository under `storage.backupPath` (default `<rootPath>/backup`), one subdirectory per database, capturing the transaction log alongside the data.

## Purpose
Give RocksDB first-class server-managed backup/restore (LMDB already had volume-snapshot workflows), without stopping the server for the common case.

## Where to focus review
- **Restore safety invariant** — `dataLayer/rocksdbBackup.ts` (`restoreBackup`) + `dataLayer/restoreMarker.ts`. Online restore writes an fsynced marker + takes a per-db flock, closes the DB across all worker threads, verifies process-wide closure via `registryStatus()`, then purges + rewrites. A component-held DB (and always `system`) can't be closed → 409 pointing at offline restore. Worth checking the close→verify→purge ordering and the marker/lock crash-recovery states (`in-progress` / `incomplete` / `clear`).
- **Offline restore lock probe** — `restoreBackupOffline`. The offline path is chosen on a PID heuristic, but the PID file is briefly absent during `harper restart`, so it now probes RocksDB's own lock (`RocksDatabase.open`) before purging — a lock error aborts, a corrupt/half-restored dir still proceeds (recovery).
- **Job-worker handle release** — `server/jobs/jobProcess.ts` + `resources/databases.ts` (`closeDatabase`/`closeLoadedDatabases`, new). Job workers close their process-global RocksDB handles on exit or the closure check can never pass; `system` is intentionally left open; tableless databases are handled explicitly.
- **CLI transport refactor** — `bin/cliOperations.ts`. Extracted target + auth resolution into `resolveRequestOptions`, now shared by `cliOperations` and the streaming `get_backup` download. `cliOperations` behavior should be unchanged (bin suite green).

## Cross-model review (codex + Gemini)
Fixed from review: offline-restore lock probe, CLI `content-disposition` path traversal, `jobProcess` cleanup when `updateJob` rejects, marker-removal fsync, tableless-DB handle leak. Deferred (not blocking):
- Enforcing "DB closed" via a native atomic RocksDB lock instead of `registryStatus()` polling — the cross-process case is covered by the offline probe; a native primitive is a rocksdb-js follow-on.
- Table background-work disposal (expiration timers/callbacks) on close — pre-existing (`resetDatabases` shares the gap), filed as #1811.

## Docs
Companion docs PR: HarperFast/documentation#590.

_Generated by an LLM (Claude Opus 4.8, 1M context)._

---

## Update — blob backups + restore-safety review (2026-07-29)

This revision addresses the review feedback (kriszyp / GPT 5.6). Merged latest `main` first (one conflict in `bin/cliOperations.ts`, resolved).

### Blobs (`exclude_blobs`, default false → blobs included)
File-backed blobs live in roots outside the RocksDB directory, so the engine backup misses them. New `dataLayer/blobBackup.ts` captures them alongside the engine data:
- **Managed backups** snapshot each blob root to `<backupDir>/blobs/<backupId>/<rootIndex>/<relpath>` — a full per-id copy (like the binding's transaction-log snapshots), hard-linked when possible, copied across filesystems, never symlinked, built in a temp dir and atomically renamed. `restore_backup` purges each root and rewrites it from the snapshot; `delete_backup`/`purge_backups` remove the matching snapshots. Wired for both the online ops and the offline CLI.
- **`get_backup`** appends blob files to the same tar under `blobs/<rootIndex>/…`. Since the binding finalizes its tar with a fixed 1024-byte end-of-archive marker, `createBackupStream` streams the native *plain* tar minus that trailer, appends blobs via `tar-stream`, and gzips the combined stream itself.
- `get_backup` CLI help updated.

### Restore-safety fixes
- **Offline lock probe (was the data-loss bug):** now recognizes rocksdb-js 2.5.0's real `LOCK`-file error (a plain `Error`, no `code`, `"While lock file: …/LOCK: Resource temporarily unavailable"`), takes the restore lock+marker *before* probing, and fails closed. Two-process regression test included.
- **Restore-metadata namespace:** moved from `<db>.restoring`/`<db>.restore.lock` siblings to an isolated `.restore/` dir keyed by a hash of the db-dir name — fixes the collision with a legal database named `orders.restoring` and the `NAME_MAX` overflow on 250-char names.
- **Recovery marker:** `beginRestore` reports a pre-existing marker so a failed recovery attempt no longer clears a marker over a possibly half-purged directory.
- **Drop vs restore:** `dropDatabase` now holds the per-db restore lock across the whole drop (serializing with restore on one primitive) instead of a check-then-act marker probe.

### Please look at
- **`bin/cliOperations.ts` auth:** the merge resolution moved credential resolution into `resolveRequestOptions`, which ran *before* `cliOperations`' try/catch — so an incomplete `auth_username=/auth_password=` pair escaped the `exit(1)` handling. Moved back inside the try; the 4 `harper#1872` credential tests pass again. Worth a sanity check that CLI auth precedence is unchanged.
- **`get_backup` tar merge** depends on the binding's fixed 1024-byte trailer; the trailer-strip verifies the withheld bytes are all-zero and throws loudly if the format ever changes.
- **Blob snapshot consistency is best-effort point-in-time** (whatever files exist while it streams; a blob deleted mid-walk is skipped) — Harper does not freeze blob writes for a backup, the same tradeoff the engine makes for the transaction log. Hard-linking is safe because blobs are content-addressed/write-once (see `blobBackup.ts` header).
- **Behavior change:** `dropDatabase` now refuses (409) a database with an *incomplete-restore* marker, same as it already did for an in-progress restore.

Unit suites green (`dataLayer`, `resources`, `bin`); relying on CI for the integration suite. New tests: `blobBackup.test.js`, extended `rocksdbBackup.test.js` (blobs, `exclude_blobs`, get_backup tar contents, two-process lock probe), rewritten `restoreMarker.test.js` (collision/NAME_MAX/preexisting/scan), `databases.test.js` (drop-vs-restore), `bin/backup.test.js` (atomic download).

_Generated by an LLM (Claude Opus 4.8, 1M context)._


bin/backup.ts
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * Runs a backup operation from the CLI. `command` is the operation name (e.g. `create_backup`);
+ * there is no hyphenated alias. Each works whether or not Harper is running: RocksDB is
+ * single-writer, so when a server is reachable the running process must own the database handle
+ * and the operation is forwarded to it (also handling remote `target=` and auth); only when the
+ * *local* server is stopped do we open the database/backup files directly. `get_backup` is the
+ * exception — it streams a live snapshot from a running server and has no offline form.
+ */
+export async function runBackupCommand(command: string): Promise<void> {
+	const request = buildRequest();
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in b38d9ce: `runBackupCommand` now calls `require('dotenv').config()` before `useOperationApi`/`resolveRequestOptions` read the target env vars, mirroring `cliOperations`.

dataLayer/restoreMarker.ts
```diff
@@ -0,0 +1,127 @@
...
+			closeSync(fd);
+		}
+		// fsync the parent directory so the marker's directory entry is durable — without this a
+		// power loss can lose the entry, and a half-purged database would load as healthy
+		const dirFd = openSync(dirname(dbPath), 'r');
+		try {
+			fsyncSync(dirFd);
+		} finally {
+			closeSync(dirFd);
+		}
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in b38d9ce: the directory fsync now goes through a shared `fsyncDir` helper that ignores EPERM/EISDIR/ENOTSUP, so it's a no-op on Windows instead of failing the restore.

dataLayer/restoreMarker.ts
```diff
@@ -0,0 +1,127 @@
...
+		// fsync the parent directory so the marker's *removal* is durable — symmetric with the
+		// creation fsync in beginRestore. Without it, a power loss could resurrect the marker's
+		// directory entry and misclassify a fully-restored database as incomplete (blocking its
+		// load until a needless rerun).
+		const dirFd = openSync(dirname(dbPath), 'r');
+		try {
+			fsyncSync(dirFd);
+		} finally {
+			closeSync(dirFd);
+		}
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in b38d9ce: `completeRestore` now uses the same platform-guarded `fsyncDir` helper.

dataLayer/rocksdbBackup.ts
```diff
@@ -0,0 +1,567 @@
...
+		await findBackup(backupDir, requireBackupId(request.backup_id), databaseName);
+	} else if ((await listBackupsInDir(backupDir)).length === 0) {
+		throw new BackupNotFoundError(`No backups found for database '${databaseName}'`);
+	}
+	// The database is normally loaded and RocksDB; a database left unloaded by an interrupted
+	// restore is also restorable (rerunning the restore is how it is recovered).
+	const databaseDir = resolveDatabasePath(databaseName);
+	if (checkRestoreState(databaseDir) === 'clear' || getDatabases()[databaseName]) {
+		requireRocksRootStore(databaseName, OPERATIONS_ENUM.RESTORE_BACKUP);
+	}
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in b38d9ce: `requireRocksRootStore` is now guarded on the database having tables (`Object.keys(loaded).length > 0`); an empty/tableless database is validated by the restore job rather than failing with 'no tables to back up'.

dataLayer/rocksdbBackup.ts
```diff
@@ -0,0 +1,567 @@
...
+		await findBackup(backupDir, backupId, databaseName);
+	} else {
+		backupId = available.reduce((latest, backup) => Math.max(latest, backup.backupId), 0);
+	}
+	// a loaded database's root store knows its real directory (which can differ from the
+	// computed default, e.g. legacy layouts); fall back to the computed path only when the
+	// database is unloaded (recovering an interrupted restore)
+	const databaseDir = getDatabases()[databaseName]
+		? requireRocksRootStore(databaseName, OPERATIONS_ENUM.RESTORE_BACKUP).path
+		: resolveDatabasePath(databaseName);
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in b38d9ce: same tables guard here; an empty or unloaded database falls back to `resolveDatabasePath`.

dataLayer/rocksdbBackup.ts
```diff
@@ -0,0 +1,567 @@
...
+			error.message = `Restore of database '${databaseName}' from backup ${backupId} failed (rerun restore_backup to recover): ${error.message}`;
+		} else {
+			// nothing destructive happened — clear the marker and let every thread reload the intact database
+			completeRestore(databaseDir, lockToken);
+			await signalling.signalSchemaChange(
+				new SchemaEventMsg(process.pid, OPERATIONS_ENUM.RESTORE_BACKUP, databaseName)
+			);
+		}
+		throw error;
+	}
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in b38d9ce: wrapped in `new Error(msg, { cause: error })` instead of mutating `error.message`.

dataLayer/rocksdbBackup.ts
```diff
@@ -0,0 +1,567 @@
...
+		}
+	}
+	const lockToken = beginRestoreForDatabase(databaseDir, targetDatabase ?? databaseName);
+	try {
+		await backups.restore(backupDir, databaseDir, { backupId, mode: 'purgeAllFiles' });
+	} catch (error) {
+		abandonRestore(lockToken);
+		error.message = `Restore of database '${databaseName}' from backup ${backupId} failed (rerun restore_backup to recover): ${error.message}`;
+		throw error;
+	}
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in b38d9ce: same `new Error({ cause })` wrapping in the offline path.

utility/operation_authorization.ts
```diff
@@ -205,6 +206,30 @@ requiredPermissions.set(restart.restart.name, new (permission as any)(true, [],
...
+	new (permission as any)(true, [READ_PERM], terms.OPERATIONS_ENUM.DELETE_BACKUP)
+);
+requiredPermissions.set(
+	rocksdbBackup.purgeBackups.name,
+	new (permission as any)(true, [READ_PERM], terms.OPERATIONS_ENUM.PURGE_BACKUPS)
+);
+requiredPermissions.set(
+	rocksdbBackup.restoreBackup.name,
+	new (permission as any)(true, [READ_PERM], terms.OPERATIONS_ENUM.RESTORE_BACKUP)
+);
```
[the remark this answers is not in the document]
cb1kenobi: Fixed in 536e18da6.

Rather than protecting only the three destructive ops, all six managed-backup operations now self-enforce `super_user` — they're whole-database administrative operations (not table-scoped), so none should be delegable to a non-SU role. The enforcement mirrors `get_deployment_payload`'s `requireSuperUser`: a shared gate at each op's request-context entry point — the direct handlers (`list`/`delete`/`purge`) and the job-op validators (`create`/`verify`/`restore`, which run before any job record is created).

Note on the root cause: gate-2 delegation applies to *every* `requires_su` op, so changing the registration shape can't make an op non-delegable — handler-level enforcement is the only mechanism the codebase provides for "strictly SU". I left the systemic gate-2 gap (the TODO) alone since it predates this PR (`get_backup` already had it) and falling through would make these require table `READ`, which is both too weak for destructive ops and still not SU-only. Refreshed that TODO to note the backup ops are now self-enforcing.

Added unit coverage asserting each op rejects a non-SU (and userless) request with 403.

— Generated by Claude (Opus 4.8)

dataLayer/rocksdbBackup.ts
```diff
@@ -0,0 +1,618 @@
...
+	// a heuristic: the PID file is briefly absent mid-`harper restart`, and backups.restore's
+	// purgeAllFiles never takes RocksDB's own lock. Probe that lock by opening the database — a live
+	// holder makes open throw "is locked" — so we refuse rather than purge a database another
+	// process still has open. A directory that fails to open for any *other* reason (corrupt or
+	// half-restored) is exactly what restore recovers, so only a lock conflict aborts.
+	if (existsSync(join(databaseDir, 'CURRENT'))) {
+		try {
+			RocksDatabase.open(databaseDir).close();
+		} catch (error: any) {
+			if (typeof error?.message === 'string' && error.message.includes('is locked')) {
```
kriszyp: This does not recognize the lock error emitted by the pinned `@harperfast/rocksdb-js` 2.5.0 binding. I verified it with two processes: the second `RocksDatabase.open()` throws `IO error: While lock file: <db>/LOCK: Resource temporarily unavailable` (with no `code`), so this catch ignores the contention and continues into `backups.restore(..., { mode: 'purgeAllFiles' })`. A stale/missing PID—including the restart window described immediately above—can therefore purge a database that another Harper process has open. Please make lock contention a typed/native signal (or use a reliable native lock probe), fail closed on it, and take the restore marker/lifecycle exclusion before probing so a restarting server cannot open the database between this probe and `beginRestore`. Add a two-process regression test using the binding's real error.

dataLayer/rocksdbBackup.ts
```diff
@@ -0,0 +1,618 @@
...
+	requireRocksRootStore(databaseName, OPERATIONS_ENUM.CREATE_BACKUP);
+}
+
+export async function createBackup(request: any) {
+	const databaseName = getDatabaseName(request);
+	const rootStore = requireRocksRootStore(databaseName, OPERATIONS_ENUM.CREATE_BACKUP);
+	const backupDir = backupDirForDatabase(databaseName);
+	let backupId;
+	try {
+		backupId = await rootStore.backup(backupDir, { transactionLogs: true });
```
kriszyp, line 376: This snapshots RocksDB files and transaction logs, but not Harper's external blob roots (`resources/blob.ts:getRootBlobPathsForDB`). Records persist file-backed blobs as references into `<root>/blobs/<database>` or `storage.blobPaths`; after a later update/delete unlinks an old blob, restoring an older RocksDB backup leaves that record pointing to a missing file. An off-host streamed backup likewise contains none of the payload bytes. Please include and consistently restore the database's blob roots (with a snapshot protocol that handles concurrent blob writes/deletes), or explicitly make this an engine-only backup that cannot be presented/restored as a complete Harper database. A regression should back up a record with a file-backed blob, replace/delete it, restore the older backup, and read the original bytes.

dataLayer/rocksdbBackup.ts
```diff
@@ -0,0 +1,618 @@
...
+			abandonRestore(lockToken);
+			// wrap rather than mutate error.message: a frozen/library error can have a non-writable
+			// message (assigning it throws TypeError under 'use strict')
+			throw new Error(
+				`Restore of database '${databaseName}' from backup ${backupId} failed (rerun restore_backup to recover): ${error.message}`,
+				{ cause: error }
+			);
+		}
+		// nothing destructive happened — clear the marker and let every thread reload the intact database
+		completeRestore(databaseDir, lockToken);
```
kriszyp: `beginRestore()` overwrites an existing `.restoring` marker without reporting that it was already present. If this call is a recovery attempt after an earlier destructive restore failed, and this attempt then fails before `destructionStarted` (for example `verifyDatabaseClosed()` finds a leaked/component handle), this branch deletes the old recovery marker and broadcasts a reload. The directory may still be half-purged from the first attempt, so the scanner can now load partial/corrupt data as healthy. Please preserve whether the marker pre-existed and only clear a marker created for a previously healthy database; an incomplete marker must remain until a restore completes successfully. Add a test starting from an abandoned marker and forcing a pre-destruction failure.

resources/databases.ts
```diff
@@ -898,6 +974,9 @@ export async function dropDatabase(databaseName) {
 	for (const tableName in dbTables) {
 		const table = dbTables[tableName];
 		rootStore = table.primaryStore.rootStore;
+		// a drop's file deletion must not interleave with a restore's purge-and-copy on the same
+		// directory (destroy landing after the restore's copy would gut a "successful" restore)
+		if (rootStore instanceof RocksDatabase) throwIfBlockedByRestore(rootStore.path, databaseName);
```
kriszyp: This check only protects the ordering where restore creates its marker before drop reaches this line. If drop sees `clear`, another worker can begin restore immediately afterward; `close()` can drive the registry refcount to zero so restore starts copying while `destroy()` is still deleting, allowing a successful restore to be gutted by the concurrent drop. Please serialize the full drop lifecycle and restore lifecycle with the same per-database lock (held from before the state decision through the final filesystem mutation), rather than using a check-then-act marker test. The regression needs both race orderings.

resources/databases.ts
```diff
@@ -394,6 +401,35 @@ export function getDatabases(): Databases {
...
+ * databases that must not be loaded: a held restore lock means a restore is in progress in some
+ * process; an unheld lock with a surviving `.restoring` marker means a restore was interrupted
+ * mid-purge (the directory may be partial garbage) and must be rerun. The files live *next to*
+ * the database directory, so this also covers a database whose directory is missing or empty.
+ */
+function databasesBlockedByRestore(databasePath: string, entries: { name: string }[]): Set<string> {
+	const blocked = new Set<string>();
+	const candidates = new Set<string>();
+	for (const { name } of entries) {
+		if (name.endsWith(RESTORING_MARKER_SUFFIX)) candidates.add(name.slice(0, -RESTORING_MARKER_SUFFIX.length));
```
kriszyp: These metadata suffixes occupy the same directory-entry namespace as valid database names. For example, a legal database directory named `orders.restoring` makes `checkRestoreState(<root>/orders)` treat that directory itself as a marker and block a healthy `orders`; restoring `orders` also fails when it tries to open the directory as a marker file. Legal names can be up to 250 characters, so appending `.restore.lock` can also exceed a typical 255-byte `NAME_MAX`. Please place restore metadata in an isolated directory with an encoded/keyed database name (or otherwise preserve all currently legal names and lengths), and cover suffix-collision plus maximum-length names.

bin/backup.ts
```diff
@@ -0,0 +1,132 @@
...
+		const chunks: Buffer[] = [];
+		for await (const chunk of response) chunks.push(Buffer.from(chunk));
+		throw new Error(`get_backup failed (${response.statusCode}): ${Buffer.concat(chunks).toString('utf8')}`);
+	}
+	// name the file after what the server is sending (RocksDB: <db>.tar[.gz]; LMDB: the database file).
+	// basename() strips any path from the server-supplied filename — a malicious/compromised target
+	// could otherwise return `../../…` and steer the write outside the cwd.
+	const serverFilename = response.headers['content-disposition']?.match(/filename="([^"]+)"/)?.[1];
+	const outputPath = request.out || (serverFilename && basename(serverFilename)) || `${databaseName}.backup`;
+	await pipeline(response, createWriteStream(outputPath));
```
kriszyp: `createWriteStream(outputPath)` truncates an existing known-good backup before the download succeeds, and `pipeline` leaves a partial file at the final name on a network/server failure. Please stream to an exclusive temporary sibling and atomically rename only after successful completion (with an explicit overwrite policy), cleaning up the temporary file on error. That preserves an existing backup and prevents a truncated archive from looking complete.

kriszyp (changes requested):
Looks awesome! GPT mostly has concerns about the lock handling and timing with failures. But probably the biggest gap is blob file handling. We could independently handle blob files, but kinda makes the backup system a bit incomplete. We might want to consider some way of creating a snapshot directory that adds hard-links to the blob files so they can exist as snapshots alongside the rocksdb backups? (And include them in the tar'ed snapshot for get_backup). I'd also be fine with blob backup being a follow-up PR and saying backups don't support blobs for now. Good work!

dataLayer/rocksdbBackup.ts
cb1kenobi: Fixed in d1572bd9e. `isRocksDbLockError` now matches the real 2.5.0 error (`While lock file: …/LOCK: Resource temporarily unavailable`, no `code`); the restore lock + marker is taken **before** the probe, and a lock conflict fails closed with a 409 instead of purging. Added a two-process regression using the binding’s real error (`rocksdbBackup.test.js` → “offline restore lock probe (two-process)”).

— Claude Opus 4.8

dataLayer/rocksdbBackup.ts
cb1kenobi: Added in d1572bd9e. Blobs are now captured with managed backups and streamed in `get_backup` via new `dataLayer/blobBackup.ts`, controlled by `exclude_blobs` (default false → included). Each blob root is snapshotted per backup (hard-link-else-copy, atomic temp+rename); `restore_backup` purges and rewrites each root from the snapshot; `delete_backup`/`purge_backups` drop the matching snapshots. Regression added: back up a record’s file-backed blob, delete it, restore, and read the original bytes.

— Claude Opus 4.8

dataLayer/rocksdbBackup.ts
cb1kenobi: Fixed in d1572bd9e. `beginRestore` now returns `preexisting`, and a recovery attempt that fails **before** any destruction leaves the marker in place — only a *fresh* marker on a *previously-healthy* database is cleared. Covered in `restoreMarker.test.js` (“reports preexisting=true …”).

— Claude Opus 4.8

resources/databases.ts
cb1kenobi: Fixed in d1572bd9e (+ 3911ea24d). `dropDatabase` now acquires the same per-database restore lock and holds it across the entire drop, so drop and restore serialize on one primitive rather than a check-then-act marker probe. `database()`’s on-demand open keeps the read-only `throwIfBlockedByRestore`. Test in `databases.test.js` (“dropDatabase restore serialization”).

— Claude Opus 4.8

resources/databases.ts
cb1kenobi: Fixed in d1572bd9e. Restore metadata moved into an isolated `.restore/` directory keyed by `sha256(basename(dbPath)).slice(0,32)`, so it’s out of the database-name namespace (no `orders.restoring` collision) and length-bounded (no NAME_MAX overflow on 250-char names). The scan reads `.restore/` and maps each marker back via its first line. Covered by the collision + max-length cases in `restoreMarker.test.js`.

— Claude Opus 4.8

bin/backup.ts
cb1kenobi: Fixed in d1572bd9e. `get_backup` now streams to an exclusive temp sibling (`.<pid>-<ts>.part`, flags `wx`) and renames only on success, unlinking the temp on error — an existing good backup is never truncated and a failed download never leaves a partial file at the final name. `bin/backup.test.js` covers success, mid-stream failure, and the pre-existing-backup case.

— Claude Opus 4.8

resources/databases.ts
```diff
@@ -900,49 +1004,147 @@ export async function dropDatabase(databaseName) {
...
+	// interleave with a restore's purge-and-copy on the same directory — a destroy landing after a
+	// restore's copy would gut a "successful" restore, and vice versa. Restore takes the same lock
+	// (before writing its marker), so both operations serialize on this one primitive rather than on
+	// a check-then-act marker probe. Released in the finally below.
+	const restoreLocks: RestoreLock[] = [];
+	try {
+		for (const tableName in dbTables) {
+			const table = dbTables[tableName];
+			rootStore = table.primaryStore.rootStore;
+			if (rootStore instanceof RocksDatabase) lockDatabaseForDrop(rootStore.path, databaseName, restoreLocks);
```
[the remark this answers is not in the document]
cb1kenobi: Fixed in 3911ea24d: `lockDatabaseForDrop` now dedupes by root-store path (a path already held is skipped), so a multi-table RocksDB database no longer re-acquires its own non-reentrant lock. Added multi-table drop coverage in `databases.test.js`. Good catch.

resources/databases.ts
```diff
@@ -265,9 +273,12 @@ export function getDatabases(): Databases {
 	if (databasePath && existsSync(databasePath)) {
 		// First load all the databases from our main database folder
 		// TODO: Load any databases defined with explicit storage paths from the config
-		for (const databaseEntry of readdirSync(databasePath, { withFileTypes: true })) {
+		const entries = readdirSync(databasePath, { withFileTypes: true });
+		const blockedByRestore = databasesBlockedByRestore(databasePath);
```
kriszyp, line 328: The corrected native LOCK-error match fixes the prior repro once the other process already has RocksDB open, but this cached `blockedByRestore` set leaves a check/open race. A restarting server can scan before the offline CLI writes its marker, pause before reaching this database, then open it after the CLI's probe has opened and closed successfully; the CLI proceeds into `purgeAllFiles` with that new server handle live. This is especially reachable when startup scans many databases. Please make database opening participate in the same per-database lifecycle lock, held across the restore-state decision and `RocksDatabase.open()`, so either restore excludes the open or the subsequent probe observes RocksDB's LOCK. Add a two-process regression with a barrier between the startup scan and the actual open.

dataLayer/rocksdbBackup.ts
```diff
@@ -0,0 +1,920 @@
...
+
+export async function createBackup(request: any) {
+	const databaseName = getDatabaseName(request);
+	// blobs are captured by default; exclude_blobs=true produces an engine-only backup
+	const excludeBlobs = requireBooleanOption(request.exclude_blobs, 'exclude_blobs');
+	const rootStore = requireRocksRootStore(databaseName, OPERATIONS_ENUM.CREATE_BACKUP);
+	const backupDir = backupDirForDatabase(databaseName);
+	let backupId;
+	try {
+		backupId = await rootStore.backup(backupDir, { transactionLogs: true });
```
kriszyp, line 376: The engine snapshot and blob snapshot are not one consistent cut. `rootStore.backup()` completes first, then the code walks the live blob roots; if a record referencing blob A is in the RocksDB snapshot and a concurrent update commits blob B, Harper unlinks A shortly afterward. When the later walk reaches A it explicitly skips `ENOENT`, so the operation succeeds with a database snapshot that references a blob absent from its snapshot. The streamed path has the same ordering: it finishes the native tar before walking live blobs. Please introduce a joint snapshot protocol (for example, a short per-database barrier that preserves blob files while the engine cut and stable hard-link set are captured), then stream/copy from that stable set. The regression should use a real file-backed blob record, update/delete it during backup, restore, and read the original bytes.

dataLayer/rocksdbBackup.ts
```diff
@@ -0,0 +1,920 @@
...
+	} catch (error) {
+		throw mapLockedError(error, databaseName);
+	}
+	if (!excludeBlobs) {
```
kriszyp: Once `rootStore.backup()` resolves, the native backup is already visible to `list_backups`, `findBackup`, and `verify_backup`, but there is no persisted per-id state saying whether blobs were intentionally excluded, still being copied, or failed. A snapshot copy failure therefore makes `create_backup` fail while leaving an engine backup that later lists and verifies as healthy; a concurrent restore can also select it before `snapshotBlobs()` renames the final directory, treat the missing snapshot like an intentional engine-only backup, and report success. Please persist an atomic managed-backup manifest/completion marker (including the blob-inclusion policy), hide or roll back incomplete ids, serialize management operations through completion, and make verification cover the declared blob snapshot. Add delayed/failing blob-copy tests for list, verify, restore, delete, and purge.

dataLayer/restoreMarker.ts
```diff
@@ -0,0 +1,269 @@
...
+export const RESTORE_LOCK_SUFFIX = '.lock';
+export const RESTORING_MARKER_SUFFIX = '.restoring';
+
+/**
+ * Directory holding the restore metadata for a database — a `.restore/` sibling of the database
+ * directory. Shared by every database under the same parent, so a single readdir surfaces all
+ * pending restores during the startup scan.
+ */
+export function restoreMetaDir(dbPath: string): string {
+	return join(dirname(dbPath), RESTORE_META_DIR);
```
kriszyp, line 62: `.restore` is itself a currently legal RocksDB database name (`schemaRegex` accepts it and it is not reserved), so this directory is not actually outside the database-name namespace. For a database at `<root>/.restore`, its lock and marker are written inside the database being restored; `purgeAllFiles` can delete the marker, defeating crash detection, and even a successful restore makes `completeRestore()` fail with `ENOENT` when it tries to unlink that already-purged marker. For other databases, their metadata is also injected into the live `.restore` database directory. Please place lifecycle metadata somewhere guaranteed not to overlap any legal or pre-existing database path (including backward-compatible handling for an existing `.restore` database), and test both a successful and interrupted restore of an actual database with this name.

dataLayer/blobBackup.ts
```diff
@@ -0,0 +1,234 @@
...
+		.filter((entry) => entry.isDirectory() && /^\d+$/.test(entry.name))
+		.map((entry) => Number(entry.name))
+		.sort((a, b) => a - b);
+	// purge every current blob root first so nothing newer than the backup survives the restore
+	for (const root of blobRoots) {
+		await rm(root, { recursive: true, force: true });
+	}
+	if (blobRoots.length === 0) return;
+	for (const index of indexDirs) {
+		const target = blobRoots[index] ?? blobRoots[blobRoots.length - 1];
```
kriszyp: Copying an out-of-range snapshot index into the last configured root preserves bytes but not their address. File-backed blob references persist `storageIndex`; if a backup was made with two roots and is restored after configuration changes to one, a record with `storageIndex: 1` still resolves through `blobRoots[1]`, while this code copied its file into `blobRoots[0]`. The restore reports success but reading that blob fails. Please reject an incompatible root-count mapping before any destructive restore (or migrate/rewrite the stored indices with a durable mapping) rather than collapsing indices, and cover a real blob record restored from two roots into one.

kriszyp (changes requested):
This looks like a solid mechanism, I like this, good work. I think the codex findings are reasonable requests. The snapshot consistency is certainly the most interesting/challenging, but I think the suggestion of having a mechanism to delay/pause blob deletions during the backup process makes sense to me.

dataLayer/restoreMarker.ts
cb1kenobi: Good catch — `.restore` is indeed a legal database name, so a plain `.restore` metadata dir collides with it. Fixed in 800a4b6c4 by naming the metadata directory with a backtick (`` `restore` ``): `schemaRegex` forbids only `/` and backtick among filesystem-legal characters, so no legal database can ever resolve to that path, and a user cannot create one that does. A database literally named `.restore` now keeps its markers in the separate backtick dir (not inside the DB being purged). Added a regression restoring a database named `.restore` (marker lives outside the DB dir; `completeRestore` no longer ENOENTs).

— Claude Opus 4.8

dataLayer/blobBackup.ts
cb1kenobi: Agreed — collapsing an out-of-range index mis-addresses blobs since records persist `storageIndex`. Fixed in 800a4b6c4: removed the collapse and added `assertBlobSnapshotRestorable`, which rejects (400) a backup whose captured root count exceeds the current `storage.blobPaths` **before any destructive step** — pre-checked in both the online and offline restore paths (ahead of the engine restore), and re-checked in `restoreBlobSnapshot` before it purges. Regression added: a two-root snapshot restored into a one-root config rejects and leaves the root un-purged.

— Claude Opus 4.8

dataLayer/restoreMarker.ts
```diff
@@ -0,0 +1,275 @@
...
+ * - `<.restore>/<key>.restoring` — the completion marker. Written (and fsynced) after the lock is
+ *   acquired and before the destructive restore begins; deleted only after the restore completes
+ *   successfully, while still holding the lock. Its *existence* means "a restore started and has
+ *   not finished successfully". Its first line records the database directory name so the startup
+ *   scan can map a marker back to the database it blocks without decoding the hashed key.
+ */
+
+// The backtick makes this an illegal database name (schemaRegex rejects `/` and backtick only), so
+// it can never collide with a real database directory — see the module header.
+export const RESTORE_META_DIR = '`restore`';
```
kriszyp, line 51: The `.restore` collision is fixed, but the guarantee in this comment does not hold for a pre-existing on-disk database. `getDatabases()` derives `dbName` from every directory and opens any `CURRENT` + `MANIFEST-*` directory without applying `schemaRegex` (`resources/databases.ts:279-299`), so a storage root can already contain a live RocksDB directory named `` `restore` `` even though today's create API rejects that name. A restore of another database then writes its lock/marker into that live database and the startup scan treats the same directory as lifecycle metadata. Please place metadata outside the scanned database-root namespace, or explicitly detect and safely migrate/refuse a pre-existing reserved directory before writing it; add a regression with an actual RocksDB at the reserved path.

dataLayer/rocksdbBackup.ts
```diff
@@ -0,0 +1,974 @@
...
+			} catch (error: any) {
+				if (error.code === 'ENOENT') continue;
+				throw error;
+			}
+			for (const entry of entries) {
+				const filePath = join(dir, entry.name);
+				if (entry.isDirectory()) {
+					stack.push(filePath);
+				} else if (entry.isFile()) {
+					await appendBlobFile(pack, filePath, `blobs/${index}/${relative(root, filePath)}`);
```
kriszyp: `path.relative()` uses `\\` separators on Windows, and that value is currently written verbatim into the tar header. A Windows backup therefore contains names such as `blobs/0/abc\\def\\ghi`; when extracted on POSIX those backslashes are ordinary filename characters, so the documented root/shard layout is not reconstructed. Please normalize the relative portion to POSIX `/` separators before calling `appendBlobFile`, and add an assertion that emitted tar entry names never contain backslashes.

dataLayer/rocksdbBackup.ts
```diff
@@ -0,0 +1,974 @@
+'use strict';
+
+import { createReadStream, existsSync, readdirSync } from 'node:fs';
+import { open, readdir, writeFile } from 'node:fs/promises';
+import { join, relative, resolve } from 'node:path';
+import { PassThrough, Writable } from 'node:stream';
+import { pipeline } from 'node:stream/promises';
+import { createGzip } from 'node:zlib';
+import { setTimeout as delay } from 'node:timers/promises';
+import { pack as tarPack, type Pack } from 'tar-stream';
```
kriszyp, line 10: `tar-stream` is now imported directly by production code (and by the new tests), but it is absent from `package.json` and `dependencies.md`; it only resolves today because `tar-fs` happens to depend on and hoist it. That is not a stable package contract—another package-manager layout or a future `tar-fs` release can make Harper fail at module load. Please declare and lock `tar-stream` as a direct runtime dependency and document its rationale, or avoid the direct import.

DESIGN.md
```diff
@@ -328,6 +328,123 @@ this fix doesn't attempt to solve. `deploy_component`/`package_component` still
 declared entry points (`jsResource`/`graphqlSchema`) survived extraction — a truncation from some
 other future cause would still report success silently; that's a deferred, separate fix.
 
+## RocksDB backup/restore: the restore lock + marker protocol (`dataLayer/restoreMarker.ts`, `dataLayer/rocksdbBackup.ts`)
+
+The `restore_backup` operation restores a user database on a live server by closing it across all
+worker threads, purging its directory (`backups.restore` with `purgeAllFiles`), and reloading it.
+Three non-obvious mechanics keep that safe:
+
+- **Two files in an isolated `.restore/` directory beside (never inside) the database directory**,
```
kriszyp: `RESTORE_META_DIR` now uses `` `restore` ``, but this design section and several JSDoc comments in `restoreMarker.ts` still call the on-disk directory `.restore/`. That is now specifically the legal database path this fix is avoiding, so recovery guidance can send operators to the wrong directory. Please update all metadata-path references to the actual reserved directory.

kriszyp (changes requested):
I'll drop this in for the comments since it is still marked ready, but I assume this is still a bit of WIP with the blob/database coordination.

dataLayer/restoreMarker.ts
cb1kenobi: Good point — the backtick blocks *creation* via the API, but the scan opens any CURRENT+MANIFEST dir without re-checking `schemaRegex`. Fixed in f74a6390a: the startup scan now explicitly skips the reserved `` `restore` `` entry in both enumeration loops, so an out-of-band directory planted at that name is never loaded as a database (and other databases' markers in that shared dir are never treated as a live DB). Regression added (`databases.test.js`): a real-looking RocksDB (CURRENT + MANIFEST-*) planted at the reserved path is not loaded, and the reserved dir is left in place for metadata.

— Claude Opus 4.8

dataLayer/rocksdbBackup.ts
cb1kenobi: Fixed in f74a6390a: the blob entry name is now normalized to POSIX separators (`relative(root, filePath).split(sep).join(\"/\")`) before packing, so a Windows backup no longer emits `blobs/0/abc\\def\\ghi`. Added an assertion to the get_backup tar test that no emitted entry name contains a backslash. (The on-disk snapshot copy in `blobBackup.ts` already uses per-OS `join`, so only the tar header needed normalizing.)

— Claude Opus 4.8

dataLayer/rocksdbBackup.ts
cb1kenobi: Agreed — relying on `tar-fs` hoisting it is not a stable contract. Fixed in f74a6390a: `tar-stream` (^3.1.8) and `@types/tar-stream` (^3.1.4) are now declared as direct dependencies (package.json + package-lock), with a rationale entry in `dependencies.md`.

— Claude Opus 4.8

DESIGN.md
cb1kenobi: Fixed in f74a6390a: updated the stale `.restore/` references in DESIGN.md and the `restoreMarker.ts` jsdoc to the actual reserved `` `restore` `` directory (and added a note there that the scan skips it), so recovery guidance points at the right path.

— Claude Opus 4.8

dataLayer/rocksdbBackup.ts
cb1kenobi: Implemented in 4e941a735. New `dataLayer/backupManifest.ts` writes an atomic per-id manifest (`<backupDir>/manifests/<id>.json`, temp+rename) recording the blob-inclusion policy, only after both the engine backup and blob snapshot are durable; a graceful snapshot failure rolls back the just-created engine backup + partial snapshot. `list_backups` hides ids with no manifest; `verify_backup`/`restore_backup` reject them (409 for a specific id, "no complete backups" for latest); restore uses the manifest's `blobs` flag (not snapshot-dir presence) to decide blob restore, and verify flags a manifest that claims blobs but has no snapshot as corrupt. Regression added for the hide/refuse behavior.

— Claude Opus 4.8

dataLayer/rocksdbBackup.ts
```diff
@@ -0,0 +1,974 @@
...
+	if (!excludeBlobs) {
+		await snapshotBlobs(backupDir, backupId, getBlobPathsForDatabaseName(databaseName));
+	}
+	await writeBackupReadme(backupDir, databaseName);
+	return {
+		database: databaseName,
+		backup_id: backupId,
+		blobs: !excludeBlobs,
+		...(await describeBackup(backupDir, backupId)),
+	};
```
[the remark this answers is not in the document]
cb1kenobi: Implemented in 4e941a735: atomic completion manifest (`dataLayer/backupManifest.ts`) recording the blob policy, written only after both phases are durable, with rollback on failure; list hides / verify+restore reject an id with no manifest, and restore keys off the manifest rather than snapshot presence.

dataLayer/rocksdbBackup.ts
cb1kenobi: Acknowledged as a real point-in-time skew (a blob unlinked between the engine cut and the blob walk). Per maintainer decision this is deferred to a follow-up: it is the documented best-effort limitation (blob capture is point-in-time, like the transaction log). The completion manifest (4e941a735) closes the "incomplete backup looks healthy" and concurrent-restore races; a joint snapshot barrier (pinning the blob hard-link set across the engine cut) is tracked separately.

— Claude Opus 4.8

dataLayer/rocksdbBackup.ts
```diff
@@ -0,0 +1,974 @@
...
+	const backupDir = backupDirForDatabase(databaseName);
+	let backupId;
+	try {
+		backupId = await rootStore.backup(backupDir, { transactionLogs: true });
+	} catch (error) {
+		throw mapLockedError(error, databaseName);
+	}
+	if (!excludeBlobs) {
+		await snapshotBlobs(backupDir, backupId, getBlobPathsForDatabaseName(databaseName));
+	}
```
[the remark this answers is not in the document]
cb1kenobi: Deferred by maintainer decision to a follow-up (documented best-effort point-in-time capture). The completion manifest in 4e941a735 addresses the healthy-looking-incomplete and concurrent-restore aspects; the joint engine+blob snapshot barrier is tracked separately.

resources/databases.ts
cb1kenobi: Acknowledged. The offline probe now takes the restore lock+marker before probing and fails closed on the real RocksDB LOCK error (d1572bd9e), which narrows this; fully closing the residual scan-check/open TOCTOU means holding the per-database lifecycle lock across `RocksDatabase.open()` in the load path. Per maintainer decision that is deferred to a follow-up (it touches the hot DB-load path and warrants its own two-process regression).

— Claude Opus 4.8

resources/databases.ts
```diff
@@ -900,49 +1009,147 @@ export async function dropDatabase(databaseName) {
...
+	// interleave with a restore's purge-and-copy on the same directory — a destroy landing after a
+	// restore's copy would gut a "successful" restore, and vice versa. Restore takes the same lock
+	// (before writing its marker), so both operations serialize on this one primitive rather than on
+	// a check-then-act marker probe. Released in the finally below.
+	const restoreLocks: RestoreLock[] = [];
+	try {
+		for (const tableName in dbTables) {
+			const table = dbTables[tableName];
+			rootStore = table.primaryStore.rootStore;
+			if (rootStore instanceof RocksDatabase) lockDatabaseForDrop(rootStore.path, databaseName, restoreLocks);
```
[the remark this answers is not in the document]
cb1kenobi: Deferred by maintainer decision. The offline probe was hardened (lock+marker before probe, real LOCK-error detection, fail-closed); holding the per-db lifecycle lock across the scan-time open is tracked as a follow-up since it touches the DB-load hot path.

dataLayer/rocksdbBackup.ts
```diff
@@ -0,0 +1,1100 @@
...
+		await writeBackupManifest(backupDir, backupId, blobs);
+	} catch (error) {
+		await deleteBackupManifest(backupDir, backupId).catch(() => {});
+		await deleteBlobSnapshot(backupDir, backupId).catch(() => {});
+		await backups.delete(backupDir, backupId).catch(() => {});
+		throw error;
+	}
+}
+
+// --- synchronous operations ---
```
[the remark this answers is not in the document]
cb1kenobi: Added in the latest push: a test stubs `snapshotBlobs` to reject and asserts the engine backup, blob snapshot, and manifest are all gone afterward — covering the `finalizeBackup` rollback path directly.

dataLayer/rocksdbBackup.ts
```diff
@@ -0,0 +1,1100 @@
...
+	const excludeBlobs = requireBooleanOption(request.exclude_blobs, 'exclude_blobs');
+	const rootStore = requireRocksRootStore(databaseName, OPERATIONS_ENUM.CREATE_BACKUP);
+	const backupDir = backupDirForDatabase(databaseName);
+	let backupId;
+	try {
+		backupId = await rootStore.backup(backupDir, { transactionLogs: true });
+	} catch (error) {
+		throw mapLockedError(error, databaseName);
+	}
+	// snapshot blobs (unless excluded) then publish the completion manifest; rolls back on failure
```
kriszyp, line 380: The binding's backup-directory writer lock ends when `rootStore.backup()` resolves, before `finalizeBackup()` copies blobs and publishes the manifest. A concurrent `purge_backups` can therefore acquire that lock, see and purge this new engine backup, and return while this path continues copying blobs, writes a manifest, and reports a successful `backup_id` that no longer exists in the engine. Please hold one lifecycle lock across engine creation, blob capture, and manifest publication (and use that same lock for delete/purge), or add an equivalent final engine-presence check under serialization with rollback. A regression can pause `snapshotBlobs`, run purge, then assert creation cannot return success for an absent backup.

resources/databases.ts
```diff
@@ -315,11 +324,18 @@ export function getDatabases(): Databases {
...
-		for (const databaseEntry of readdirSync(databasePath, { withFileTypes: true })) {
+		const entries = readdirSync(databasePath, { withFileTypes: true });
+		const blockedByRestore = databasesBlockedByRestore(databasePath);
+		for (const databaseEntry of entries) {
 			// in-progress migration staging dirs are not databases until atomically renamed into place
 			if (databaseEntry.name.endsWith(MIGRATING_DIR_SUFFIX)) continue;
+			// the restore-metadata directory is reserved: never load it as a database, even if a
+			// (out-of-band) RocksDB directory happens to occupy that reserved name — the API can't
+			// create it (schemaRegex forbids the backtick), but the scan opens any CURRENT+MANIFEST dir
+			if (databaseEntry.name === RESTORE_META_DIR) continue;
```
kriszyp, line 335: This skip prevents startup from opening the reserved directory, but it still leaves the collision case unresolved. If the reserved `` `restore` `` path is already a RocksDB (possible because discovery historically accepted `CURRENT` + `MANIFEST-*` directories without applying `schemaRegex`), an upgrade now silently makes that database disappear, and the next restore writes its `.lock`/`.restoring` files into the same directory via `acquireRestoreLock`. Please detect an existing non-metadata directory before treating this path as lifecycle state and either refuse startup/restore with an actionable error or migrate the metadata outside the database root. The regression should use a real `RocksDatabase` containing a row and verify its data is not merely skipped.

bin/backup.ts
```diff
@@ -0,0 +1,169 @@
...
+ * through to the liveness check — the API only when the local instance is actually running
+ * (`getHdbPid` verifies process liveness, so a stale pid file reads as stopped).
+ */
+function useOperationApi(request: any): boolean {
+	const lastTarget = loadCredentials()?.last_target;
+	if (
+		request.target ||
+		process.env.HARPER_CLI_TARGET ||
+		process.env.CLI_TARGET ||
+		(lastTarget && !isLocalTarget(lastTarget))
```
kriszyp, line 118: A loopback `last_target` does not establish that the target is this Harper installation: it may be an SSH tunnel, a container-published port, or another local Harper root. If that endpoint is currently down and this installation's PID is absent, `restore_backup`/`delete_backup`/`purge_backups` now silently operate on this installation's files instead of the saved target. Please make offline mode explicit or persist an unambiguous local-installation identity rather than inferring it from the hostname. Add a regression with a saved loopback target for a different port/installation and assert that it never falls through to direct file access.

unitTests/dataLayer/rocksdbBackup.test.js
```diff
@@ -0,0 +1,610 @@
...
+			const original = blobBackupModule.snapshotBlobs;
+			blobBackupModule.snapshotBlobs = () => Promise.reject(new Error('snapshot boom'));
+			try {
+				await assert.rejects(createBackupOffline(MDB), /snapshot boom/);
+			} finally {
+				blobBackupModule.snapshotBlobs = original;
+			}
+			const backupDir = backupDirForDatabase(MDB);
+			// the incomplete backup must be fully rolled back: no engine backup, no manifest, no snapshot
+			assert.strictEqual((await listBackupsInDir(backupDir)).length, 0, 'engine backup must be rolled back');
```
kriszyp, line 490: `listBackupsInDir()` filters out backups without completion manifests, so this assertion is zero even if rollback leaves the RocksDB engine backup intact—the preceding test relies on that same filtering behavior. Please assert against the binding's raw backup listing or the engine backup artifacts on disk, then keep the manifest/blob checks separately. That makes this regression fail if `deleteBackup()` stops removing the engine backup.

kriszyp (approved):
Looks good, nice work. I'm including the comments; codex thinks they are blockers/major, but I think we've agreed it is reasonable follow-up work, so I'm good with merging this.

bin/harper.ts
```diff
@@ -191,18 +211,17 @@ if (require.main === module) {
...
 			// Intentionally not calling `process.exit(0);` so if a CLI
 			// command resulted in a long running process (aka `run`),
 			// it continues to run.
 		})
 		.catch((error) => {
-			if (error) {
-				console.error(error);
-				logger.error(error);
-			}
+			if (error) console.error(formatCliError(error));
```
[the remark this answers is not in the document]
cb1kenobi: In practice the two did duplicate on the terminal — the reported bug was `harper` printing the full stack twice to stderr (verified: a `ClientError:` block from `console.error(error)` plus a `[main/0] [error]: …` block from `logger.error(error)`). Dropping the raw error print + logger call and routing through `formatCliError` gives the clean single-line message the user asked for. The tradeoff is that CLI errors are no longer written to the Harper log file; that is acceptable for one-shot CLI invocations (the terminal is the output), but easy to restore as file-only logging if we decide the log record matters.

unitTests/dataLayer/rocksdbBackup.test.js
cb1kenobi: `listBackupsInDir()` is the binding's raw listing — it returns `backups.list(backupDir)` with no manifest filtering (that filtering lives only in `listCompleteBackups()`). So this assertion is exactly "assert against the binding's raw backup listing": if rollback left the engine backup intact, `listBackupsInDir()` would return it (length 1) and the test would fail. Happy to also assert the on-disk engine artifacts are gone for belt-and-suspenders, but the current check does catch a rollback that only removes the manifest.

— Claude Opus 4.8

dataLayer/rocksdbBackup.ts
cb1kenobi: Real race — the engine backup lock releases before finalizeBackup copies blobs + writes the manifest, so a concurrent purge can delete the engine backup while create finishes. Tracked as a follow-up to serialize backup management ops under a per-database lock: #2031.

bin/backup.ts
cb1kenobi: Fair — a loopback last_target may be a tunnel/container/another root, so silently routing destructive offline ops to local files is a footgun. Tracked as a follow-up (restrict local fallback to read-only, or gate destructive ops behind an explicit offline flag): #2032.

resources/databases.ts
cb1kenobi: Agreed — the scan skip avoids loading it, but a pre-existing on-disk `` `restore` `` database (only reachable via manual creation) would silently disappear. Tracked as a follow-up to warn/refuse instead of silently skipping: #2033.

PR #1833, opened 2026-07-16
test(resources): promote QA read-consistency regression anchors

kriszyp:
## Summary

Promotes three cold-verified-green QA-explorer regression specs into `integrationTests/resources/`, pinning read-consistency guarantees around Harper's `@relationship` GraphQL expansion and `sourcedFrom` cache resolvers:

- **`relationship-graphql-snapshot-consistency`** (QA-593) — nested GraphQL `Series -> articles` expansion never returns a torn (mixed-generation) sibling set when a writer atomically bulk-updates all of a series' children concurrently with reads. Parent/child generation *skew* across the two separate write transactions is expected and is not what this test gates on — only a torn *child-level* set is.
- **`sourcedfrom-eav-cache-coherence`** (QA-595) — an EAV-assembled `sourcedFrom` cache resource (5 attribute rows fused into one cached "product") converges to the raw-row oracle after TTL+settle under a bulk attribute-by-attribute mutation burst and a deterministic `invalidate()`-vs-in-flight-fill race.
- **`sourcedfrom-snapshot-scoping`** (QA-596) — establishes *why* the QA-595 guarantee holds: a `sourcedFrom` resolver's snapshot read-scoping survives a real async gap (not just a same-tick microtask hop) for reads within one database (single-table and cross-table), and documents that it is **not** guaranteed across two databases — by design, not a defect.

All three were cold-rerun on current `main` (gate SHA `3c5aeb9b7`) before promotion, then re-run again after relocation into their final path in this worktree.

## Notes

- Test-only — no product code changes.
- A fourth candidate (QA-597, EAV × Blob type-drift × blob-store reclamation) also gated GREEN but was excluded from this batch: it's a blob-lifecycle/reclamation test, not a read-consistency test, so it doesn't fit this PR's theme. Will follow in a separate batch.
- Generated by Claude Opus 4.8.

### DIFF
--- a/integrationTests/resources/sourcedfrom-eav-cache-coherence.test.ts
+++ b/integrationTests/resources/sourcedfrom-eav-cache-coherence.test.ts
@@ -1,0 +1,381 @@
+/**
+ * QA-595 — EAV product catalog fronted by a cache-sourced (`sourcedFrom`) resource:
+ * read-consistency of the assembled view under concurrent attribute-by-attribute writes.
+ *
+ * A "product" is not a row. It's assembled at read time from N Attribute rows (one row
+ * per entity+attribute-name — a real EAV junction table, keyed `${entityId}:${attrName}`).
+ * Product is a `@table(expiration: 3)` cache resource whose sourcedFrom get() reads those
+ * 5 rows ONE AT A TIME (deterministic order: name, price, color, stock, description) and
+ * returns the assembled view. The EAV substrate's unit of write (one row, one independent
+ * HTTP PUT) and the cache's unit of consistency (one assembled product, one cache entry)
+ * disagree — attribute-by-attribute mutation is inherently non-atomic from the cache's
+ * point of view.
+ *
+ * Three probes:
+ *   P1 bulk natural-conditions race hunt — 60 products x 5 attrs (300 rows) mutated
+ *      attribute-by-attribute (independent PUTs, no shared txn) with concurrent reads
+ *      racing the mutation, at realistic speed (no artificial delay). Looks for a
+ *      "torn" view (mixed old/new attrs in one assembled response) and — the actual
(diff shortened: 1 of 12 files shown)


integrationTests/resources/sourcedfrom-eav-cache-coherence.test.ts
```diff
@@ -0,0 +1,381 @@
...
+			})();
+
+			// fire the burst mutation timed to overlap this entry's natural expiry moment
+			await sleep(Math.max(0, TTL_MS - 300));
+			const mutateAtOffset = Date.now() - primedAt;
+			await Promise.all(ATTR_NAMES.map((a) => putAttribute(id, a, 1)));
+
+			await spray;
+
+			const errors = spraySamples.filter((s) => typeof s.status !== 'number');
```
cb1kenobi: **Low: P3 "no error" assertion misses HTTP 5xx responses**

`getProduct()` returns a numeric `status` for any HTTP response and an `ERR:<msg>` string only on a thrown/network failure. This filter keeps just the string case, so a transient 500 during the TTL-boundary spray (e.g. a `sourcedFrom` fill throwing mid-race, exactly the kind of hang/error this probe claims to gate) is recorded as a numeric status and passes silently. The eventual-convergence assert still catches persistent divergence, but the stated "no hang/error" guarantee is under-enforced.

Suggested fix:
```suggestion
		const errors = spraySamples.filter((s) => typeof s.status !== 'number' || s.status >= 400);
```

—
_Generated by Barber AI_

kriszyp: Addressed in 43debc688: the spray now records numeric HTTP statuses >= 400 as errors alongside transport failures.

PR #1835, opened 2026-07-16
fix(blob): durable blob-unlink queue — deletions survive worker/process death; orphan observability (#1832)

### LINKED ISSUE #1832: Concurrent blob REPLACE on one key leaks orphaned blob files (inline unlink loses the race; only cleanup_orphan_blobs recovers them)
### Summary

Under **concurrent blob REPLACE on a single key**, the inline (commit-time) blob-unlink path loses the race and strands superseded blob files on disk. They are never reclaimed by any inline/background path — only an explicit `cleanup_orphan_blobs` operation recovers them. That manual sweep is a mitigation; it should not be required. This is a disk leak proportional to write concurrency (worst on the busiest tables), with **zero operator-visible signal** (no counter, no log, no scheduled sweep).

### Environment

- harper `main` @ `3dbcf7b9e` (v5.2.0-alpha.5)
- **Both storage engines** (RocksDB and LMDB — same signature)
- `threads.count = 4`

### Repro

10 concurrent writers × 8 `put()` REPLACEs each (80 total, plain `put()`, **no `ifVersion`**) of a ~2MB-blob record on **one** key.

Observed on-disk blob-file count:
- Converges 18 → **8** in ~4s, then sits **flat at 8** for the remaining 36s. Expected floor: **2** (or 1 after deleting the live record).
- Deleting the live record frees only its own current blob (8 → 7), leaving the already-orphaned superseded versions stranded.
- `cleanup_orphan_blobs` then reclaims **all** of them to the true floor (RocksDB 9 → 1, LMDB 10 → 1) within ~10s — confirming nothing is *permanently* unreferenced, but that reclamation only ever happens on the manual op.

**Sequential control** (80 REPLACEs @ 200ms, same fixture) settles to baseline+1 on both engines → **concurrency is the trigger**, not the fixture or a threshold.

Reproduction test (from our QA campaign, ready to promote as a regression anchor):
```
npm run test:integration -- "integrationTests/qa-scratch/qa592-blob-concurrent-writers.test.ts"
# LMDB leg: prefix HARPER_STORAGE_ENGINE=lmdb
```

### Root cause (traced)

The inline blob GC runs at commit (`resources/blob.ts` — `startPreCommitBlobsForRecord` / `cleanupUnusedBlobs`, and the on-demand reclaimer at `resources/blob.ts:1870` `cleanupOrphans`). Under concurrent REPLACE on the same key, the commit-time unlink of the superseded blob loses the race — the new version's file is written but the prior version's file is not consistently unlinked — so orphans accrue. There is no background sweep (cf. #1288 / F-006 for the adjacent pre-save-then-cleanup-on-lost-conflict path; here the **record is correct** and only file GC is deferred/lost, so the observable is distinct: a real on-disk leak, not a wrong record).

### Impact

- Unbounded on-disk growth on hot-key blob tables under concurrent writes, invisible until an operator happens to run `cleanup_orphan_blobs`.
- No orphan-bytes/count metric, log line, or scheduled sweep to make it observable.

### Ask

1. Make the inline/commit-time blob unlink race-safe so concurrent REPLACEs on one key do not strand superseded blob files (the primary fix — the manual sweep should not be load-bearing).
2. Failing/alongside (1), expose orphan blob bytes/count as an observable metric and/or an opt-in periodic sweep so the condition is detectable without running the op.

_Found by the exploratory QA campaign (QA-592, axis B×I). Filed at Kris's direction._

kriszyp:
## Summary

Refs #1832: concurrent blob replacement now records a crash-surviving unlink intent and proves the winning record no longer references the file before deletion. The final repair also distinguishes a blob owned only by the losing write from one seen by another encoder, preventing cleanup of bytes a concurrent winner still references.

## For the human reviewer

1. **Ownership is intentionally exclusive for newly allocated files.** A file-backed blob stamped for one record cannot be reused by a different record; the former aliasing behavior could silently delete shared bytes. Legacy ownerless references remain ownerless, so upgrades favor a leak over inventing unsafe ownership. Rejecting this choice would require removing the durable owned-file reclaim path, not a local adjustment.
2. **Cleanup requires positive version evidence.** Local losers may queue immediate cleanup only when the same write was the sole owner stamper. Replication/source losers carry their bounded source version, and the drain unlinks only after observing a strictly newer owner record that does not reference the file. Equal, older, missing, or versionless evidence defers; old versionless rows are eventually abandoned to the orphan sweep. This deliberately trades bounded disk leakage for protection against data loss.
3. **Known compatibility boundary.** A downgrade can strip ownership metadata and create an ownerless alias before re-upgrade; closing that requires a migration/capability policy outside this PR. Likewise, a stale serialized blob reference arriving after a later process completed reclamation needs a durable tombstone, not the process-local interlock added here.
4. **Design review outcome.** The planning gate ended `chosen-approach-sound` after the implementation adopted a per-write identity token, a sticky “seen by another writer” bit, absent-token classification, and bounded source-version proof. A drain-only ownership check was rejected because it cannot identify a not-yet-committed concurrent winner; record-scoped queue rows and holding locks through primary commit were deferred because they widen the storage/write-path contract without improving the fail-closed result.
5. **Review status.** Earlier whole-branch rounds supplied Claude and Gemini coverage. The final exact-head rerun was blocked before findings because this dispatch lacked the environment’s required cross-model disclosure authorization; commit `4d28f94bff0a` therefore relies on the executed local verification below and should receive human review before promotion.

### Ownership classification
When may an abandoned write publish an immediate cleanup intent without endangering another write?

| Before | After |
| --- | --- |
| Cleanup inferred safety from the blob instance and could condemn a file another encoder was about to commit. | Each write supplies a non-persisted identity token; a distinct or absent token permanently marks that instance as shared across writes, so only a sole-self stamp may use immediate ownerless cleanup. |

> Ambiguous ownership never authorizes deletion. It either remains protected by the live owner record or is handed to the conservative orphan sweep.

### Durable reclaim lifecycle
How does the queue survive worker death without deleting a concurrent winner’s data?

#### From supersession to unlink
The record version and per-file lock turn reclamation into a proof rather than a timer callback.

~~~mermaid
sequenceDiagram
    participant Writer
    participant Queue
    participant Owner
    participant Drain
    Writer->>Queue: Stage file, owner, losing version, due time
    Drain->>Queue: Read due intent
    Drain->>Owner: Read current committed record
    Drain->>Drain: Take per-file reclaim lock
    Drain->>Queue: Confirm intent still exists
    alt owner is strictly newer and omits file
        Drain->>Drain: Unlink file
        Drain->>Queue: Settle intent
    else proof is absent or ambiguous
        Drain->>Queue: Defer or abandon safely
    end
~~~

### Operational boundary
What remains observable and recoverable when automatic proof cannot be established?

The queue retries transient unlink failures, preserves stronger existing rows, avoids overwriting version evidence from another loser, and eventually releases unprovable rows so `cleanup_orphan_blobs` can report or reclaim true orphans. The operator-facing `dryRun` parser now rejects uninterpretable values instead of silently treating strings as truthy.

## Changes

- [`resources/blob.ts` adds persisted owner metadata, per-write classification, the durable unlink queue, exact reclaim locking, version-aware drain proof, retry/abandon behavior, allocator high-water protection, and orphan-sweep exclusion](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-5c96de79cba36ba9ddb5124ce7b7e96a398a3c50fc6b7bec11dc5d690f61767aR71). The hottest correctness boundary is [`cleanupUnusedBlobs`](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-5c96de79cba36ba9ddb5124ce7b7e96a398a3c50fc6b7bec11dc5d690f61767aR4304), where sole-self, foreign/shared, and wire-origin losers diverge; look hardest at the [`readyToUnlink` owner/version proof and lock ordering](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-5c96de79cba36ba9ddb5124ce7b7e96a398a3c50fc6b7bec11dc5d690f61767aR1989), plus [legacy reference decoding and ownership stamping](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-5c96de79cba36ba9ddb5124ce7b7e96a398a3c50fc6b7bec11dc5d690f61767aR4374).
- [`resources/DatabaseTransaction.ts` gives each blob-bearing write an identity token and bounded cleanup-losing version](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-2ae62080514cb4ef64bf50088e7ea0587701f558af71f76b119f07e00f3d56beR403), then [passes both through skipped, superseded, and aborted cleanup](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-2ae62080514cb4ef64bf50088e7ea0587701f558af71f76b119f07e00f3d56beR1538). [`resources/LMDBTransaction.ts` mirrors cleanup across LMDB’s async and synchronous commit exits](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-98388eabfc912d6936a44f365db83b04514fbd08ced64ff6f5c9b112b83630a6R316).
- [`resources/RecordEncoder.ts` threads the token through primary and audit encoding and stages prior-file intents with the committed version](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-d175bc619774c0fb18833823092f139f0ba6a0eb4503f49159fd5c2e0e324f38R975); [record removal waits for success and durably stages afterward](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-d175bc619774c0fb18833823092f139f0ba6a0eb4503f49159fd5c2e0e324f38R1123). [`resources/Table.ts` registers owner lookup, stages dropped-table files, supplies the token at every encode boundary, and carries source/replay versions](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R1907).
- [`resources/databases.ts` adds shared queue epochs to both engines, initializes recovery drains with each root, and registers tables for owner resolution](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1020). [`bin/copyDb.ts` preserves non-numeric record keys and migration ownership context](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-4460df89ee6f04c0f4068abb855e054742096fbbf29ba3634b80d54537235be7R1059).
- [`dataLayer/schema.ts` strictly parses `cleanup_orphan_blobs.dryRun` and reports invalid input](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-07a296e386acc3155215669015450d25a7ee06830408a635c73991d1ce3619b2R385), while [`components/mcp/tools/schemas/operationDescriptions.ts` documents the dry-run result](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-05d48c6145d75f8454b5e92581106d8f53390f59a1b0c15c554f48d3e78e2a93R175). [`DESIGN.md` records the durable ownership/version safety invariant](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-3dc5dd454e080eb849ee5efaf79df2585fbe0a06804d69249b4c81da05a63875R89).

## Verification

- `npm run build` passed; targeted oxlint and Prettier checks over all changed source/test files passed; `git diff --check` passed.
- [`unitTests/resources/blob.test.js` adds deterministic queue, lock, ownership-token, version-proof, retry, drop, migration, and both-engine regressions](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-9b125a1f520c4f030fa53590e672ec5278ebf46b2a2b78a9e6bbd70ea42b9382R2593): complete blob suite — RocksDB 175 passing; LMDB 174 passing, 1 expected pending. Focused durable-queue suite — 45 passing on each engine.
- Full `test:unit:resources` passed on both engines: RocksDB 2721 passing, 45 pending; LMDB 2144 passing, 410 pending. [`unitTests/resources/databases.test.js` adds queue-state teardown coverage](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-d89a878c46959d4e1a9364b6009ecf89d8ca08cc125e71357911283edce9b0efR571).
- [`qa592-blob-concurrent-writers.test.ts` runs an 80-write same-key storm, recycles the staging workers, validates every winning read, and requires files to converge from two to one](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-045eff553d66d8e83aeb91940a8f3fe991e44bc98b9c9dab82ee324c745d9881R263). It passed on RocksDB and LMDB using the [`config.yaml` isolated runtime](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-462e12a23920810033875ba06a6427c0327e1abe242c73848b2117d6c2a20140R1), [`resources.js` competing PUT endpoints](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-6372dbe70a92b26a1f1585eeca9bff2a35df12bb500c7cad10798d0c4d63eb5aR29), and [`schema.graphql` blob table](https://github.com/HarperFast/harper/pull/1835/changes?w=1#diff-2a3be8446fcedfb57851205070789764cde6b1e24e820c335ce9112489eef925R5): zero mismatches/read errors; file counts converged to 2 at settle and 1 after final replacement.
- Planning review: four iterations, with the final framing verdict `chosen-approach-sound`. Final exact-head independent review: failed before findings because the required cross-model disclosure authorization was absent from this dispatch; no reviewer verdict is claimed for `4d28f94bff0a`.

<sub>Review-Coverage: authored=codex; ran=claude,gemini; declined=cursor-grok,cursor-composer,domain; rounds=27; full=2 @ 84db464d2b0b</sub>

Complexity: complicated

<sub>Human-Review-Need: 4 @ 84db464d2b0b</sub>


integrationTests/database/qa592-blob-concurrent-writers.test.ts
```diff
@@ -0,0 +1,449 @@
...
+ * stuck and will never converge" — and it still runs the post-delete check even if the
+ * settle check fails, so a single run yields maximal diagnostic signal.
+ *
+ * Reproduction:
+ *   cd /home/kzyp/dev/harper
+ *   npm run test:integration -- "integrationTests/database/qa592-blob-concurrent-writers.test.ts"
+ * Harper SHA: 3dbcf7b9e (main)
+ */
+import { suite, test, before, after } from 'node:test';
+import { ok } from 'node:assert';
```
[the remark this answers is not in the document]
kriszyp: Not applied: this repo forbids `node:assert/strict` — the oxlint `no-restricted-imports` rule (.oxlintrc.json) and AGENTS.md test style mandate plain `node:assert` (use `assert.strictEqual`/`deepStrictEqual` for strict checks). Importing `node:assert/strict` fails `npm run lint:required`. This test only uses `ok()`, which is identical in both, so plain `node:assert` is correct here.

kriszyp:
**CI-red fix (both failures root-caused; durable-unlink-queue design untouched)**

**1. Unit `blob.test.js` "gates a local write…" → `No store specified`** — pre-existing test fragility, not a store-resolution regression. The test called `saveBlob(blob)` bare, resolving the blob path from module-global `currentStore`, which is undefined in the full `test:unit:resources` run. Fixed by binding the store via `decodeFromDatabase(...)` (matches the sibling #406 test). → 1147/0 on both engines.

**2. Integration `qa592` storm: 30s ECONNABORTED + 500 "Outstanding write transactions" queue** — pre-existing **uWS-only** blob-upload wedge, not a throughput regression from the unlink bookkeeping. Only the uWS shard 4/6 failed; Node v22/24/26 shard 4/6 all passed the same file. Reproduced locally and confirmed it **still wedges with the durable-unlink-queue fully disabled** (falling back to the old `setTimeout` path), and does not occur on the Node HTTP server. Mechanism: under the extreme storm (10 writers × 2MB REPLACEs on one key + tight reader loops) a uWS request-body upload stalls mid-stream; the app-supplied blob source has no idle watchdog (armed only on the replication receive path), so the commit gates forever on `saveBlob().saving` and the write queue wedges. It's a permanent wedge, so client-side retry can't recover it.

**Resolution:** skip `qa592` under uWS (as it already skips win32/bun) — its assertions (inline reclamation, LWW, no torn reads) are transport-independent and fully covered on the Node HTTP legs for both engines.

**⚠️ Flagging for a reviewer decision:** the underlying uWS wedge is a real, pre-existing availability hazard (a stalled app upload permanently wedging all writes) that is orthogonal to #1832 and likely wants its own issue + an app-stream idle watchdog. Happy to file/chase it separately if desired.

Validation: qa592 Node HTTP 2/2 both engines; no regressions in blob-advanced / record-caching-blob / record-caching-ttl-eviction; lint + prettier clean.

— generated by KrAIs (Claude Opus 4.8)

resources/blob.ts
```diff
@@ -847,18 +849,158 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
...
+ */
+export function initBlobUnlinkQueue(rootStore: any): void {
+	if (!unlinkQueueDb(rootStore)) return;
+	scheduleBlobUnlinkDrain(rootStore, deletionDelay + 25); // startup drain (prior-life rows are all due by then)
+	if (isMainThread && !mainSafetyDrains.has(rootStore)) {
+		mainSafetyDrains.add(rootStore);
+		const interval = setInterval(() => {
+			// self-clear once the database is closed/dropped so the interval's closure doesn't
+			// pin the store and spin on a dead handle
+			if ((rootStore as any).opened === false || (rootStore as any).dbisDb?.opened === false) {
```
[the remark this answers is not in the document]
kriszyp: Fixed in 06e550140 — switched the self-clear guard to `status === 'closed'`, matching the convention used throughout `resources/databases.ts`. Good catch, thanks for flagging this.

— Claude Sonnet 5

resources/blob.ts
```diff
@@ -1094,8 +1104,152 @@ function resetDrainedQueue(): void {
...
+			}
+		}, UNLINK_SAFETY_INTERVAL);
+		interval.unref?.();
+	}
+}
+
+/**
+ * Delete the file for the blob. Scheduling here is in-memory only (see runReclamation for why);
+ * the durable-queue guarantee (survives thread/process death) applies once reclamation has
+ * decided the file is safe to unlink, not at the moment of supersession.
```
cb1kenobi: **High: the durable queue only starts at the *end* of the retention window, so the worker-recycle window this PR targets is still in-memory only**

`deleteBlob()` mutates the in-process `pendingReclamation` map and arms an `unref()`'d `setTimeout(runReclamation, storage.blobRetention)` (default 2000ms). The durable row is written by `enqueueBlobUnlink()` *inside* `runReclamation` (line 1378), i.e. only after that window has already elapsed. A `worker.terminate()` inside the window discards the timer and the intent exactly as it did before this PR — the queue covers only the ~50ms between the reclamation decision and the unlink.

Measured on this head with a probe test in `unitTests/resources/blob.test.js` (`setDeletionDelay(3000)`, polling the queue range in `dbisDb` every 25ms from the supersede):

```
rocksdb: durable row first visible 3002ms after supersede, file unlinked 3028ms, row removed 3054ms
lmdb:    durable row first visible 3015ms after supersede, file unlinked 3065ms, row removed 3065ms
```

So the exposure per supersede is the full retention window (~2s by default) — 4x the 500ms the PR body attributes the leak to, and the failure mode the ledger evidence in the description describes (threads whose last event was <=400ms after scheduling).

Related: the PR description states the row "is written **inside the same transaction that commits the supersede** (RocksDB: the write txn; LMDB: the commit handler's conditional batch)". The head does not do that — the doc comment on `enqueueBlobUnlink` says the opposite ("Always a synchronous, standalone commit ... there is no enclosing transaction to join at this point"). Whichever way this lands, the description needs updating.

Suggested fix (directional): record the intent in `deleteBlob()` at supersede time and let `due = now + retention` (the field already exists, and is currently always `Date.now()`) govern when a drain may execute it, instead of gating whether the row exists at all. That moves the hold / snapshot / re-reference checks to drain time — which also needs the cancel path (`cancelBlobReclamation`) to remove the row, so it is not a one-line change. If the window is instead accepted as designed, say so explicitly here, in the PR description, and in the qa592 assertion comment, so it is not read as closed.

—
_Generated by Barber AI_

resources/blob.ts
```diff
@@ -1094,8 +1104,152 @@ function resetDrainedQueue(): void {
...
+	} catch (error) {
+		logger.debug?.('Unable to read blob unlink queue', error);
+		return;
+	}
+	for (const { key, value } of entries) {
+		if (!value || value.due > now) continue; // not yet due
+		const fileId = key[1];
+		const storageInfo = { storageIndex: value.storageIndex ?? 0, fileId, store: rootStore };
+		const filePath = getFilePath(storageInfo);
+		unlink(filePath, (error) => {
```
cb1kenobi, line 1951: **Medium: a queue row that outlives its file can delete a *live* blob after a restart (fileId reuse)**

The drain unlinks first (1180) and removes the row second (1187-1188), and a failing removal is swallowed at debug level (1190). If the process dies between the two — or the removal simply fails — the row survives for a fileId whose file is already gone.

`getNextFileId()` (blob.ts:2050) re-seeds the allocator from **the highest id found on disk** on the first allocation after a restart. If the stranded row was for the highest id (deleting the most recently written asset is the ordinary way to produce that), the next blob written gets the same fileId and therefore the same path, and the next drain — the startup drain, or the 5s main-thread interval — unlinks that new, live file. Nothing revalidates at drain time: the shared hold / re-reference table is empty in a fresh process, so none of the reclamation guards apply to a recovered row.

The LMDB branch widens the window: `queueDb.remove(key)` (1188) is async and unawaited, so the removal may not have committed when the process goes down.

Confidence: static analysis of the two code paths; I did not reproduce the end-to-end sequence.

Suggested fix: give the id allocator a floor that accounts for the queue — `max(highest id on disk, highest fileId in the unlink queue) + 1` in `getNextFileId()`. A stranded row then can never name a future file. (Removing the row *before* the unlink is not the alternative — that re-opens the loss this queue exists to prevent.)

—
_Generated by Barber AI_

resources/blob.ts
```diff
@@ -1094,8 +1104,152 @@ function resetDrainedQueue(): void {
...
+			if (error && error.code !== 'ENOENT') {
+				// keep the row; a later drain retries (e.g. a reader still holds the file open on Windows)
+				logger.debug?.('Error trying to remove blob file, will retry', error);
+				return;
```
cb1kenobi: **Medium: a failed unlink now pins the reclaim claim forever — the code this replaces released it**

On any non-ENOENT unlink error the drain returns here without calling `releaseReclaimClaim`, so the shared slot stays at `RECLAIMING`. The direct-unlink path this replaced (still present at 1381-1389 as the no-dbi fallback) released the claim unconditionally in its callback, error or not.

Slots are shared by hash across 4096 entries, so a single file that can never be unlinked — the Windows reader case named in the comment just above, or EPERM/EACCES on a mount — permanently:
- makes `holdBlobFile()` return `null` for every colliding fileId, so replication treats those blobs as unavailable, and
- makes `cancelBlobReclamation()` take its `HOLDS < 0` branch, where it logs and skips setting `REREFERENCED` — a record written referencing that file is then not protected from the eventual retry succeeding.

The local `pendingReclamation` entry also stays with `unlinking = true` forever, so `runReclamation` skips it for the life of the thread.

Suggested fix: on the retry path, release the claim and let the next drain re-claim, or bound the retries — after N failures release the claim and leave the file to `cleanup_orphan_blobs`. Either way the shared slot should not be held by a retry loop that may never terminate.

—
_Generated by Barber AI_

resources/blob.ts
```diff
@@ -1094,8 +1104,152 @@ function resetDrainedQueue(): void {
...
+	if (!queueDb) return;
+	const now = Date.now();
+	let entries;
+	try {
+		entries = queueDb.getRange({ start: [UNLINK_QUEUE_KEY], end: [UNLINK_QUEUE_KEY, '\uffff'] });
+	} catch (error) {
+		logger.debug?.('Unable to read blob unlink queue', error);
+		return;
+	}
+	for (const { key, value } of entries) {
```
cb1kenobi, line 1876: **Medium: the drain issues one `unlink` per row with no bound**

Every due row in the range gets an `fs.unlink` fired in a single synchronous pass, with no concurrency cap and no backpressure. The drain that matters most for this is the startup drain — which exists precisely because a prior process died with a backlog — so the size of this fan-out is unbounded by anything the code controls, and it lands on a thread that is also serving requests (EMFILE / libuv threadpool starvation).

`cleanupOrphans` in this same file awaits its unlinks one at a time for the same reason.

Suggested fix: take at most N rows per pass and re-arm `scheduleBlobUnlinkDrain` while more remain, or await `unlinkPromised` sequentially in an async drain.

—
_Generated by Barber AI_

integrationTests/database/qa592-blob-concurrent-writers.test.ts
```diff
@@ -0,0 +1,472 @@
...
+			);
+			ok(readErrors.length === 0, `unexpected read errors during storm: ${JSON.stringify(readErrors.slice(0, 5))}`);
+			ok(finalSelfConsistent, `final record not self-consistent: ${JSON.stringify(final.body)}`);
+			ok(
+				finalIsWrittenVersion,
+				`LWW DEFECT: final blob sha256 does not match ANY of the ${writtenShas.size} versions this test wrote — spliced/foreign content: ${JSON.stringify(final.body)}`
+			);
+			// Inline reclamation must reach the true floor WITHOUT cleanup_orphan_blobs (#1832): blob
+			// unlink intents are durable (queued in the internal dbi, drained at-least-once), so worker
+			// recycling mid-storm — this suite restarts http_workers right before the storm — no longer
```
cb1kenobi: **Medium: nothing here (or in the unit tests) covers the durability guarantee this PR adds**

Mutation check on this head: stubbing `enqueueBlobUnlink()` to return `false` — which routes every deletion back through the pre-PR in-memory unlink fallback at blob.ts:1381 — leaves **all 99 tests in `unitTests/resources/blob.test.js` passing**. (Rebuilt `dist/` before running, since `#src/*` resolves to `./dist/*.js`; the marker appeared exactly once in `dist/resources/blob.js`.) For contrast, stubbing `drainBlobUnlinkQueue()` fails 13 tests — so the drain is covered as an unlink *mechanism*, but nothing distinguishes durable from in-memory, which is the whole change.

This suite doesn't close the gap either: `restartHttpWorkers()` runs in `before()` and waits for `/Asset/` readiness *before* the storm starts, so no worker is terminated inside a live retention window, and no leg kills the process between enqueue and drain. As written the assertions below would pass against the pre-PR code path.

Suggested fix: a unit test is cheap and direct — supersede a blob, assert the row is present under `[Symbol.for('blob_unlink_queue'), fileId]` in `rootStore.dbisDb`, clear the thread-local state, then call `drainBlobUnlinkQueue(rootStore)` and assert the file is gone and the row removed. A worker-kill leg in this suite (terminate a worker inside the retention window, then assert convergence without `cleanup_orphan_blobs`) would cover the real scenario.

—
_Generated by Barber AI_

resources/blob.ts
kriszyp: Fixed — and you were right that the window was the whole point, not a detail.

`deleteBlob()` now stages the intent **at supersession**: committed synchronously ahead of the record write on RocksDB, and queued into the same environment on LMDB so it lands in that batch or an earlier one (and rolls back with a write that aborts). `runReclamation` no longer owns the durable decision at all. The checks that used to gate the enqueue — re-reference, open snapshot, live hold, age cap — moved to drain time, where whichever process picks the row up can run them, which after a restart is never the one that staged it.

Moving the write earlier needed one thing first, which is why this branch grew: the persisted reference carried no owner. `pack`'s "cannot use the same blob in two different records" guard compared a `recordId` that only ever existed in the encoding thread, and `unpack` never restored it, so a blob read back and re-encoded under a different record silently aliased one file into two live records. A crash-surviving intent would have converted that from an intermittent loss into a reliable one. The reference now carries `[table name, primary key]`, compared by value, and is stamped only on a file the write itself allocated — so pre-upgrade ownerless references keep exactly today's behaviour.

Covered by `records the unlink intent when the record is superseded, not when the window closes`: the row must exist inside a 4 s retention window, must carry its owner, and a drain before the deadline must leave the file alone.

— Claude Opus 5

integrationTests/database/qa592-blob-concurrent-writers.test.ts
kriszyp: Good check, and it was measuring exactly the right thing. The reason it came back clean is the reason the branch grew an ownership half — the tests only observed the file on disk, which the in-memory fallback reclaims just as well.

The durability tests now read the queue row directly rather than inferring it from the file, so suppressing the durable path fails them instead of routing quietly around them. The same style of mutation check is what pinned the new interlock fix in this round: reverting one clause makes `stands down when an ownerless intent is withdrawn after the drain has read it` fail on the parent.

On the integration side, gemini's round-7 pass caught that the qa592 worker restart was in `before()` — so the storm never recycled the workers that staged its intents, and the convergence assertion was still only exercising the in-memory path. It now recycles `http_workers` **after** the storm, which makes reaching the floor without `cleanup_orphan_blobs` necessarily the work of rows recovered out of the internal dbi by workers that never saw the supersessions.

— Claude Opus 5

PR #1841, opened 2026-07-17
fix(tls): honor `ciphers`/`SECLEVEL` from every configured source when building TLS listeners

### LINKED ISSUE #1840: Per-certificate TLS `ciphers` (including client-CA `SECLEVEL` overrides) silently ignored — only `tls.ciphers`/`tls[0].ciphers` takes effect
## Problem

Harper's config surface accepts a `ciphers` string in three places: top-level `tls.ciphers`, per-entry on a `tls[]` array, and per certificate record (`add_certificate` stores it; `list_certificates` shows it). Only the first two-ish are real: the listener applies `tls.ciphers ?? tls[0].ciphers` and **silently ignores every other configured value** — `ciphers` on `tls[1..n]` entries and on certificate records (authority records' `ciphers` are never read at all; `createTLSSelector` skips `is_authority` records when building contexts).

The underlying mechanism: OpenSSL takes the cipher list — and any `@SECLEVEL=n` embedded in it, which governs client-certificate chain verification — from the context the server was **created** with. A context swapped in later by the SNI callback does not carry its own cipher list onto the connection (noted in `threadServer.js` for years). So a per-certificate cipher/security-level override cannot work as configured; the listener has exactly one effective cipher string.

## Impact

mTLS deployments whose client certificates chain through SHA-1-signed CAs must relax verification with `DEFAULT@SECLEVEL=0` (OpenSSL: `error 68: CA signature digest algorithm too weak`). Setting that on the CA entry/record — the natural, least-privilege place — appears to succeed (`list_certificates` shows it) but has no effect: every client, including ones with valid in-date certs, is rejected with `authorizationError: UNSPECIFIED`. This bit an mTLS deployment in the field (4.7; 5.x has the same shape at `server/http.ts` and `server/threads/threadServer.js` — `tlsConfig.ciphers ?? tlsConfig[0]?.ciphers`). The only working escape hatch is top-level `tls.ciphers`, which relaxes every listener rather than the one that needs it — and is easy to lose in platform config reconciliation.

## Expected

A configured `ciphers` value is either honored or loudly rejected — never silently dropped.

## Proposed fix

Resolve the effective listener cipher string from every configured source at listener creation:

1. Top-level `tls.ciphers` wins when set (explicit global knob).
2. Otherwise all `tls[]` entries and relevant certificate records are candidates — records whose `uses` matches the listener type, plus authority records when the listener verifies client certs.
3. Conflicting candidates can't coexist on one listener: pick the lowest explicit `@SECLEVEL` (a chain needing the relaxed level fails outright without it; the others merely also accept it) and log a warning naming what was ignored.
4. When a later cert-table change alters the resolved value, warn that a restart is required (the listener-level string can't be swapped by an SNI-context rebuild).

heskew:
## Summary

A TLS listener has exactly one effective cipher string: OpenSSL takes the cipher list — and any `@SECLEVEL=n` embedded in it, which governs client-certificate chain verification — from the context the server was created with, and an SNI-swapped context does not carry its own cipher list onto the connection. Harper applied only `tls.ciphers ?? tls[0].ciphers` and silently ignored every other configured `ciphers` value: entries beyond `tls[0]` and certificate records (authority records' `ciphers` were never read at all).

This resolves the listener cipher string from every configured source (`resolveEffectiveTlsCiphers` in `security/keys.ts`), wired into both Node listener-creation sites (`server/http.ts`, `server/threads/threadServer.js`):

1. Candidates come from the listener's config layers in priority order (`operationsApi.tls` before root `tls` for the operations listener; array entries relevance-filtered — CA entries only when the listener verifies client certs, served entries matched by `uses` with the selector's tolerant rule incl. legacy `'https'` and no-`uses` generics) and from relevant certificate records.
2. The **suite list** comes from the highest-priority suite-bearing candidate — a client CA that needs `DEFAULT@SECLEVEL=0` (SHA-1-signed chains, else `authorizationError: UNSPECIFIED`) must not replace or broaden the listener's configured suites.
3. The **security level** is the minimum explicit `@SECLEVEL` across candidates; strings without one keep the runtime default (no level is assumed — the OpenSSL default varies across Node builds). The two compose, e.g. `HIGH:!aNULL` + CA `DEFAULT@SECLEVEL=0` → `HIGH:!aNULL@SECLEVEL=0`, and any cross-source composition or dropped suite list is logged.
4. When a cert-table change alters the resolved value after boot, `updateTLS` warns (once per value) that a restart is required — the listener-level string can't be swapped by an SNI-context rebuild.

Closes #1840.

## Purpose

An mTLS deployment whose client certs chain through SHA-1-signed CAs configured `ciphers: DEFAULT@SECLEVEL=0` on the CA entry — the natural, least-privilege place. `list_certificates` showed the override present while the verify path rejected every client. The only working escape hatch (top-level `tls.ciphers`) relaxes all listeners and is easy to lose in platform config reconciliation. (Field incident on 4.7; 5.x has the same shape.)

## Where to look

- **Suite/level composition** (`resolveEffectiveTlsCiphers`) is the policy decision: preserve the configured suites, apply only the minimum explicit `@SECLEVEL` any relevant cert requires, warn on everything composed or dropped. This follows the external (Codex) review on the first cut, which correctly objected to whole-string replacement and to an assumed default security level.
- **Authority entries/records participate only when the listener verifies client certs**. A CA's `SECLEVEL` matters exactly when its chains are being verified; a non-mTLS listener is not relaxed.
- **`server.appliedCiphers` / `server.verifiesClientCerts`** stamps at both creation sites feed the divergence warning in `updateTLS`; the Bun path is untouched (BoringSSL has no `@SECLEVEL`).
- Certificate records are read at listener creation (cold path, boot-time) and on cert-table rebuilds (debounced); nothing added to per-connection paths.

## Tests

- Unit (`unitTests/security/keys.test.js`): 17 cases over the pure resolver — layer precedence (incl. `operationsApi.tls`), array entries beyond `[0]`, CA-entry mTLS gating, authority/uses/legacy-`https` relevance, scalar-`uses` normalization, suite+minimum-`SECLEVEL` composition, no-assumed-level, bare-`@SECLEVEL` anchoring.
- Integration (`integrationTests/security/tls-ciphers-seclevel.test.ts`): boots Harper with `tls` as an array whose *second* entry is a SHA-1-signed client CA carrying `ciphers: DEFAULT@SECLEVEL=0` (the previously-ignored position) and `http.mtls.required`; asserts a client cert chained through that CA completes the handshake and an unchained cert is refused. Verified the test fails on unfixed code (the SHA-1 chain is refused — the incident) and passes with the fix.

Local runs: `test:unit:main` green except two consistently-failing-locally, unrelated cases also failing without this change (`globalIsolation "should enforce process spawning restrictions"`, `uwsServer "rejects a body over maxBodyBytes with 413"` EPIPE) and an intermittent macOS-local `EMFILE` teardown flake in `keys.test.js` tied to system fd pressure, not this diff. CI should arbitrate.

## Cross-model review

Thorough pass (Gemini leg + Harper-domain adjudication): no blockers, no significant concerns. Gemini's one substantive finding (an `isMtls`/`mtls` asymmetry that could fire a spurious restart warning) was fixed via the `verifiesClientCerts` stamps before this PR was opened; the repeat-warning and Bun-comment suggestions are folded in. Caveat: the Codex leg failed to complete (CLI exited during context collection), so there is no Codex coverage on this run. A follow-up external (Codex) review raised five findings on the first-cut resolution policy (whole-string replacement, assumed default level, unfiltered tls[] entries, operationsApi.tls gap, relevance drift from the selector); all five are addressed by the composed model in 773b42bb0 — see the resolved threads.

## Docs

`reference/http/tls.md` in HarperFast/documentation documents `tls.ciphers` as a single string only; a companion docs PR describing the multi-source resolution (and the restart-required caveat) will follow once the resolution policy here survives review.

### DIFF
--- a/integrationTests/security/tls-ciphers-seclevel.test.ts
+++ b/integrationTests/security/tls-ciphers-seclevel.test.ts
@@ -181,8 +181,13 @@ suite(
 		});
 
 		after(async () => {
-			await teardownHarper(ctx);
-			await rm(certsDir, { recursive: true, force: true, maxRetries: 3 });
+			// before() can fail prior to certsDir being assigned; still tear Harper down, and only
+			// remove what was actually created
+			try {
+				await teardownHarper(ctx);
+			} finally {
+				if (certsDir) await rm(certsDir, { recursive: true, force: true, maxRetries: 3 });
+			}
 		});
 
 		test('accepts a client cert chained through the SHA-1-signed CA', async () => {
--- a/integrationTests/security/tls-ciphers-seclevel.test.ts
+++ b/integrationTests/security/tls-ciphers-seclevel.test.ts
@@ -1,0 +1,210 @@
+/**
+ * TLS effective-cipher resolution — client CAs that need a relaxed OpenSSL security level.
+ *
+ * A client CA whose chain is SHA-1-signed only verifies when the listener's cipher string
+ * carries `DEFAULT@SECLEVEL=0` (OpenSSL otherwise fails the chain with "CA signature digest
+ * algorithm too weak", surfacing as authorizationError UNSPECIFIED on in-date certs). The
+ * cipher string — and its @SECLEVEL — only takes effect at the listener level: a context
+ * swapped in by the SNI callback does not carry its own cipher list onto the connection.
+ * Harper previously applied only `tls.ciphers ?? tls[0].ciphers`, silently ignoring a
+ * `ciphers` value on any other `tls[]` entry or certificate record, so mTLS deployments
+ * that relaxed the level on the CA entry were rejected wholesale after every boot.
+ *
+ * This boots Harper with `tls` as an array whose SECOND entry is the SHA-1 client CA
+ * carrying `ciphers: DEFAULT@SECLEVEL=0` (the previously-ignored position) and proves a
+ * client certificate chained through that CA completes mTLS and authenticates.
+ *
+ * Reproduction:
+ *   npm run test:integration -- "integrationTests/security/tls-ciphers-seclevel.test.ts"
+ */
+
+import { suite, test, before, after } from 'node:test';
+import { ok, rejects } from 'node:assert';
+import { mkdtemp, rm, writeFile } from 'node:fs/promises';
+import { join } from 'node:path';
+import { tmpdir } from 'node:os';
+import * as https from 'node:https';
+import forge from 'node-forge';
+
+import { setupHarperWithFixture, teardownHarper, type ContextWithHarper } from '@harperfast/integration-testing';
+
+const pki = forge.pki;
+const HTTPS_PORT = 9927;
+const FIXTURE_PATH = join(import.meta.dirname, 'fixture');
+const RELAXED_CIPHERS = 'DEFAULT@SECLEVEL=0';
+// Bun terminates TLS with BoringSSL, which has no @SECLEVEL concept — the listener cipher
+// resolution under test is the Node path only.
+const skipSuite = process.env.HARPER_RUNTIME === 'bun';
+
+const DAY_MS = 24 * 60 * 60 * 1000;
+
+interface PemChain {
+	caPem: string;
+	clientCertPem: string;
+	clientKeyPem: string;
+	serverCertPem: string;
+	serverKeyPem: string;
+	unchainedCertPem: string;
+	unchainedKeyPem: string;
+}
+
+function makeCert(
+	publicKey: forge.pki.rsa.PublicKey,
+	subjectCn: string,
+	issuerCn: string,
+	serialNumber: string,
+	extensions: any[]
+) {
+	const cert = pki.createCertificate();
+	cert.publicKey = publicKey;
+	cert.serialNumber = serialNumber;
+	cert.validity.notBefore = new Date(Date.now() - DAY_MS);
+	cert.validity.notAfter = new Date(Date.now() + 365 * DAY_MS);
+	const subject = [{ name: 'commonName', value: subjectCn }];
+	const issuer = [{ name: 'commonName', value: issuerCn }];
+	cert.setSubject(subject);
+	cert.setIssuer(issuer);
+	cert.setExtensions(extensions);
+	return cert;
+}
+
+/**
+ * Build the incident-shaped PKI: an RSA CA and client certificate both signed with SHA-1
+ * (the digest modern OpenSSL rejects above security level 0), plus an ordinary SHA-256
+ * self-signed server certificate for the listener's own identity.
+ */
+function makeSha1ChainPems(): PemChain {
+	const caKeys = pki.rsa.generateKeyPair(2048);
+	const caCert = makeCert(caKeys.publicKey, 'SHA1 Test Client CA', 'SHA1 Test Client CA', '01', [
+		{ name: 'basicConstraints', cA: true, critical: true },
+		{ name: 'keyUsage', keyCertSign: true, cRLSign: true, digitalSignature: true },
+	]);
+	caCert.sign(caKeys.privateKey, forge.md.sha1.create());
+
+	const clientKeys = pki.rsa.generateKeyPair(2048);
+	const clientCert = makeCert(clientKeys.publicKey, 'sha1-mtls-client', 'SHA1 Test Client CA', '02', [
+		{ name: 'basicConstraints', cA: false },
+		{ name: 'extKeyUsage', clientAuth: true },
+	]);
+	clientCert.sign(caKeys.privateKey, forge.md.sha1.create());
+
+	const serverKeys = pki.rsa.generateKeyPair(2048);
+	const serverCert = makeCert(serverKeys.publicKey, 'localhost', 'localhost', '03', [
+		{ name: 'basicConstraints', cA: false },
+		{ name: 'extKeyUsage', serverAuth: true },
+		{
+			name: 'subjectAltName',
+			altNames: [
+				{ type: 2, value: 'localhost' },
+				{ type: 7, ip: '127.0.0.1' },
+			],
+		},
+	]);
+	serverCert.sign(serverKeys.privateKey, forge.md.sha256.create());
+
+	// control: a self-signed client cert that does NOT chain to the CA — must be rejected
+	const unchainedKeys = pki.rsa.generateKeyPair(2048);
+	const unchainedCert = makeCert(unchainedKeys.publicKey, 'unchained-client', 'unchained-client', '04', [
+		{ name: 'basicConstraints', cA: false },
+		{ name: 'extKeyUsage', clientAuth: true },
+	]);
+	unchainedCert.sign(unchainedKeys.privateKey, forge.md.sha256.create());
+
+	return {
+		caPem: pki.certificateToPem(caCert),
+		clientCertPem: pki.certificateToPem(clientCert),
+		clientKeyPem: pki.privateKeyToPem(clientKeys.privateKey),
+		serverCertPem: pki.certificateToPem(serverCert),
+		serverKeyPem: pki.privateKeyToPem(serverKeys.privateKey),
+		unchainedCertPem: pki.certificateToPem(unchainedCert),
+		unchainedKeyPem: pki.privateKeyToPem(unchainedKeys.privateKey),
+	};
+}
+
+/** One HTTPS request; resolves with the status code (rejects on socket/handshake error). */
+function requestStatus(hostname: string, clientCert?: { cert: string; key: string }): Promise<number> {
+	return new Promise((resolve, reject) => {
+		const request = https.request(
+			`https://${hostname}:${HTTPS_PORT}/`,
+			{
+				...clientCert,
+				// the client must also relax its own security level to present a SHA-1-signed cert
+				ciphers: clientCert ? RELAXED_CIPHERS : undefined,
+				rejectUnauthorized: false, // self-signed server cert; we only care about the client-cert side
+				agent: false, // no keep-alive agent — each call is an independent handshake
+			},
+			(response) => {
+				response.resume();
+				response.on('end', () => resolve(response.statusCode ?? 0));
+			}
+		);
+		request.setTimeout(10000, () => {
+			request.destroy();
+			reject(new Error('request timed out'));
+		});
+		request.on('error', reject);
+		request.end();
+	});
+}
+
+suite(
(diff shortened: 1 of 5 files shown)


integrationTests/security/tls-ciphers-seclevel.test.ts
```diff
@@ -0,0 +1,210 @@
...
+						// the previously-ignored position: not tls.ciphers, not tls[0].ciphers
+						{ certificateAuthority: caPath, ciphers: RELAXED_CIPHERS },
+					],
+				},
+			});
+		});
+
+		after(async () => {
+			await teardownHarper(ctx);
+			await rm(certsDir, { recursive: true, force: true, maxRetries: 3 });
```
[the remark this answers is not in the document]
heskew: Addressed in 6c38e61da: after() now runs teardownHarper in a try/finally and only removes certsDir when it was actually created.

---

🤖 Posted by Claude Fable 5 on behalf of @heskew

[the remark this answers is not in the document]
heskew: Addressed in 6c38e61da: after() now runs teardownHarper in a try/finally and only removes certsDir when it was actually created.

---

🤖 Posted by Claude Fable 5 on behalf of @heskew

heskew (commented):
The ignored `ciphers` behavior is real, and the base/head regression is convincing. The listener wiring, restart warning, teardown guard, and cold-path performance look sound. I think the resolution policy needs the inline issues addressed before merge so a CA cannot relax unrelated listeners or replace the server's cipher suites, and listener-specific values are not still silently ignored.

🤖 Reviewed by [OpenAI Codex](https://openai.com/codex/).

security/keys.ts
```diff
@@ -820,6 +820,95 @@ if (typeof globalThis.Bun === 'undefined') {
...
+ * 3. Conflicting candidates can't be honored together: the lowest explicit `@SECLEVEL` wins (a
+ *    chain that needs the relaxed level fails outright without it, while the others merely also
+ *    accept it), and everything ignored is logged as a warning.
+ */
+export function resolveEffectiveTlsCiphers(tlsConfig, certRecords, type, verifiesClientCerts): string | undefined {
+	const candidates = [];
+	const topLevel = Array.isArray(tlsConfig) ? undefined : tlsConfig?.ciphers;
+	if (Array.isArray(tlsConfig)) {
+		for (let index = 0; index < tlsConfig.length; index++) {
+			if (tlsConfig[index]?.ciphers) candidates.push({ source: `tls[${index}]`, ciphers: tlsConfig[index].ciphers });
```
heskew: Could we filter these entries to the listener they can actually affect, and exclude CA entries when the listener does not verify client certificates? Right now every `tls[]` cipher value is a candidate for every listener, so a CA entry with `DEFAULT@SECLEVEL=0` also relaxes unrelated HTTP, operations API, raw TLS, and non-mTLS listeners. That bypasses the authority/mTLS guard applied to certificate records below. A regression with separate mTLS and non-mTLS listeners would cover this.

---

🤖 Posted by Codex on Nathan's behalf

security/keys.ts
```diff
@@ -820,6 +820,95 @@ if (typeof globalThis.Bun === 'undefined') {
 
 let caCerts = new Map();
 
+const SECLEVEL_PATTERN = /@SECLEVEL=(\d+)/i;
+// OpenSSL's default security level when a cipher string doesn't set one explicitly
+const DEFAULT_OPENSSL_SECURITY_LEVEL = 1;
```
heskew: This default is not stable across supported Node/OpenSSL builds. Node 24 here, and current Node documentation, use security level 2; consequently `HIGH` followed by `DEFAULT@SECLEVEL=1` is treated as a tie here and the explicit relaxation is ignored. Could we avoid assigning an assumed level to strings without `@SECLEVEL` (or preserve/query the runtime default) and add that case? https://nodejs.org/api/tls.html#openssl-security-level

---

🤖 Posted by Codex on Nathan's behalf

security/keys.ts
```diff
@@ -820,6 +820,95 @@ if (typeof globalThis.Bun === 'undefined') {
...
+					.map((candidate) => `${candidate.source} ('${candidate.ciphers}')`)
+					.join(', ')} — a TLS listener has a single effective cipher string`
+			);
+		}
+		return topLevel;
+	}
+	if (candidates.length === 0) return undefined;
+	let chosen = candidates[0];
+	for (const candidate of candidates) {
+		if (securityLevelOf(candidate.ciphers) < securityLevelOf(chosen.ciphers)) chosen = candidate;
```
heskew: Could we separate a CA's required `SECLEVEL` from its cipher-suite selection? With a strict server candidate plus `DEFAULT@SECLEVEL=0` on the CA, this replaces the server's entire cipher list with `DEFAULT` for every connection. OpenSSL's `@SECLEVEL` command changes the security level without adding ciphers, so preserving the listener/server cipher list and applying only the minimum required level would avoid broadening its suites. https://docs.openssl.org/3.0/man1/openssl-ciphers/

---

🤖 Posted by Codex on Nathan's behalf

security/keys.ts
```diff
@@ -820,6 +820,95 @@ if (typeof globalThis.Bun === 'undefined') {
...
+ * Safe to call before the certificate table exists (install, early boot) — config-only then.
+ */
+export function getEffectiveTlsCiphers(type, mtlsOptions?): string | undefined {
+	let certRecords;
+	try {
+		certRecords = databases?.system?.hdb_certificate?.search([]);
+	} catch (error) {
+		logger.trace?.('Certificate table not available while resolving TLS ciphers', error);
+	}
+	return resolveEffectiveTlsCiphers(envManager.get('tls'), certRecords, type, Boolean(mtlsOptions));
```
heskew: Could the effective listener TLS config be passed into the resolver instead of rereading root `tls` here? `flattenConfig` merges `operationsApi.tls` separately, so an `operationsApi.tls.ciphers` override that inherits the root certificate is still silently ignored. Passing the listener-specific config would also keep resolution aligned with the configuration used to construct that listener.

---

🤖 Posted by Codex on Nathan's behalf

security/keys.ts
```diff
@@ -820,6 +820,95 @@ if (typeof globalThis.Bun === 'undefined') {
...
+	if (Array.isArray(tlsConfig)) {
+		for (let index = 0; index < tlsConfig.length; index++) {
+			if (tlsConfig[index]?.ciphers) candidates.push({ source: `tls[${index}]`, ciphers: tlsConfig[index].ciphers });
+		}
+	}
+	for (const cert of certRecords ?? []) {
+		if (!cert?.ciphers) continue;
+		// normalize: stored as scalar in legacy/manual entries, expected array
+		const uses = Array.isArray(cert.uses) ? cert.uses : cert.uses ? [cert.uses] : [];
+		if (uses.includes(type) || (cert.is_authority && verifiesClientCerts)) {
```
heskew: Could this share the selector's certificate-relevance rule? `createTLSSelector` supports legacy `uses: ['https']` as a generic certificate and may serve it for this listener, while this check drops its `ciphers` and will not produce the restart warning. A focused legacy-`https` case would keep selection and cipher resolution from drifting apart.

---

🤖 Posted by Codex on Nathan's behalf

security/keys.ts
heskew: Addressed in 773b42bb0: tls[] entries now go through the same relevance rule as certificate records — CA entries (certificateAuthority) participate only when the listener verifies client certs, and served entries are matched by uses. Unit cases cover the CA-entry-on-non-mTLS-listener exclusion.

---

🤖 Posted by Claude Fable 5 on behalf of @heskew

security/keys.ts
heskew: Addressed in 773b42bb0: no level is assumed for strings without @SECLEVEL anymore — explicit levels always apply (minimum wins), and plain suites keep the runtime default. The HIGH + DEFAULT@SECLEVEL=1 case now resolves to HIGH@SECLEVEL=1 and is a unit case.

---

🤖 Posted by Claude Fable 5 on behalf of @heskew

security/keys.ts
heskew: Addressed in 773b42bb0: the resolver now composes instead of replacing — the suite list comes from the highest-priority suite-bearing candidate and only the minimum explicit @SECLEVEL is appended, so a strict server suite plus a CA needing SECLEVEL=0 yields e.g. HIGH:!aNULL@SECLEVEL=0. Composition is logged.

---

🤖 Posted by Claude Fable 5 on behalf of @heskew

security/keys.ts
heskew: Addressed in 773b42bb0: resolution now takes ordered config layers; for the operations-api listener, operationsApi.tls is consulted ahead of root tls, so an operationsApi.tls.ciphers override that inherits the root certificate is honored.

---

🤖 Posted by Claude Fable 5 on behalf of @heskew

security/keys.ts
heskew: Addressed in 773b42bb0: candidate relevance now mirrors createTLSSelector's tolerant rule — no-uses generics and legacy uses: ['https'] certificates contribute their ciphers (and the restart warning) for any listener they can serve.

---

🤖 Posted by Claude Fable 5 on behalf of @heskew

heskew:
@kriszyp forward-looking note for the edge-termination roadmap (no action needed on this PR): if public mTLS ever moves to symphony's edge termination (h2/h3 routes), the legacy-client-CA problem this PR fixes moves with it — client chains signed with weak digests (e.g. SHA-1-signed CAs) need a `SECLEVEL=0`-style relaxation in whatever stack verifies them, ideally scoped per route rather than listener-wide. Harper-side replication and operations-API mTLS stay Harper-terminated regardless, so this resolver remains the fix there.

🤖 posted by Claude Code on Nathan's behalf

heskew:
Docs follow-up (promised in the description): HarperFast/documentation#598 — multi-source cipher/SECLEVEL resolution reference. 🤖 posted by Claude Code on Nathan's behalf

kriszyp:
> forward-looking note for the edge-termination roadmap (no action needed on this PR): if public mTLS ever moves to symphony's edge termination (h2/h3 routes), the legacy-client-CA problem this PR fixes moves with it

Looking into this, RusTLS flatly refuses and will not support insecure ciphers (SECLEVEL=0), so for users that need it, Symphony won't terminate (t will forward TLS on to Harper).

PR #1847, opened 2026-07-17
fix: built-in agent drains operations tools from the lazy provider

kriszyp:
## Summary

The built-in agent drains the MCP `operations` profile to expose Harper operations (`sql`, `restart_service`, `deploy_component`, …) as agent tools. It was getting **zero** of them: `composeRegistryTools` used `snapshotProfileTools('operations')`, which reads only the static `registry` Map — but the operations profile is registered as a **lazy provider** (`setProfileToolProvider`), not static `addTool` entries. So the drain was always empty, regardless of `agent.user` or `mcp.operations.allow`. (The `application` profile was unaffected because it uses static `addTool`; the HTTP `tools/list` path was unaffected because `computeFilteredList` already merges the provider.)

## Change

- Add `listProfileTools(profile)` to `toolRegistry.ts` — merges the profile's provider tools with its static tools, deduped by name (static wins, matching `getTool`/`computeFilteredList`).
- `composeRegistryTools` uses `listProfileTools` instead of `snapshotProfileTools`.
- Refactor `computeFilteredList` (the `tools/list` path) to reuse `listProfileTools`, so the merge lives in one place.
- `snapshotProfileTools` is intentionally left unchanged — its contract is a static-only capture for atomic-rebuild restore.

## Verification

Reproduced live on a Fabric instance (`harper-pro 5.2.0-alpha.6`): before, an enabled agent with a superuser `agent.user` and `mcp.operations.allow` still listed no operations tools; after, it lists exactly the allow-listed set and autonomously drove a full test → `sql`-query → `write_file` fix → `restart_service` reload → verify loop. Unit tests added in `registryTools.test.js` (agent drains lazy-provider tools; static-wins merge) and `toolRegistry.test.js` (`listProfileTools` direct coverage).

## For the reviewer

- **One open item (not addressed here, pre-existing):** the operations tool list is composed once at agent startup and retained for the component lifetime, so operations registered by components loading *after* the agent would not appear. `snapshotProfileTools` had the same one-shot behavior, and listing is explicitly a non-security-boundary startup snapshot (per-call `resolveIdentity` handles enforcement). Widening to dynamic re-composition felt out of scope for this fix — flagging in case you'd prefer it addressed here.

_Generated by an LLM (Claude Opus 4.8), reviewed by Codex and Gemini before opening._

### DIFF
--- a/agent/registryTools.ts
+++ b/agent/registryTools.ts
@@ -20,7 +20,7 @@
 
 import { registerOperationsTools } from '../components/mcp/tools/operations.ts';
 import {
-	snapshotProfileTools,
+	listProfileTools,
 	type AuthedUser,
 	type ToolCallContext,
 	type ToolDef as RegistryToolDef,
@@ -60,7 +60,7 @@ export function ensureOperationsToolsRegistered(): void {
  *     is never cached in the handler closure.
  */
 export function composeRegistryTools(listingUser: AuthedUser, resolveIdentity: () => Promise<AuthedUser>): AgentTool[] {
-	const defs = snapshotProfileTools('operations');
+	const defs = listProfileTools('operations');
 	const out: AgentTool[] = [];
 	for (const def of defs) {
 		if (!def.visibleTo(listingUser)) continue;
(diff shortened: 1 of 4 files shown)


kriszyp:
Addressed the automated review in f6a99df92: `listProfileTools` now optional-calls the provider list and ignores nullish results while retaining statically registered tools. Coverage exercises a missing list method plus `undefined` and `null` list results.

— Codex

cb1kenobi (approved):

PR #1849, opened 2026-07-17
feat(deploy): two-phase stage/activate for deploy_component

dawsontoth:
> **Draft / RFC.** Splits `deploy_component` into two replicated phases with a cluster-wide barrier at go-live, and adds a fast, addressed rollback. The public request/response contract is unchanged. **Stacked on #2066** (see Sequencing), so it should merge after it.

## Decisions needed before this leaves draft

Four review findings are design changes to the two-phase contract rather than bugs with an obvious
fix. Each is defensible as a follow-up on a PR this size, and each is also arguably a correctness gap
worth closing before 5.3 ships. They need a call, not a patch — tick the ones you want in this PR.

- [ ] **Cluster-wide activation ordering** (heskew, the round's only High). The activate fan-out
      happens after the origin releases its node-local component lock, so nothing orders two
      concurrently-originated deploys. If node A originates D1 and node B originates D2, each can
      apply the other's fan-out last: A ends on D2, B on D1, and **both deployments report success**.
      The stage barrier does not help — it orders stage-before-activate, not deploy-against-deploy.
      Fixing it means a cluster-wide per-project order: leader serialization, or a monotonic
      activation epoch compared under the component lock and rejected if stale. Either is a real
      addition to the protocol, and the epoch option needs a home for the counter that survives
      restarts.
      *Not fixed here.* Today, concurrent originations are a documented sharp edge rather than a
      guarded one.

- [ ] **Resolve the package to bytes once, on the origin.** For a `package:` deploy the origin sends
      the identifier, not a payload, so every node independently resolves and packs it. A moving
      `latest`, a semver range, or a git branch can therefore resolve *differently* per node while
      every stage reports success — the barrier confirms "everyone staged something", not "everyone
      staged the same thing". The fix is to resolve once on the origin, persist the exact tarball plus
      a digest on the deployment row, and have peers stage and verify those bytes.
      This changes what the barrier guarantees, and it makes a `package:` deploy carry a payload it
      currently does not, with the retention cost that implies.

- [ ] **Require the deployment table for separated-phase operations.** `DeploymentRecorder.put()`
      silently no-ops when `hdb_deployment` is absent, but `activate: false` still returns a
      `deployment_id`. A later `deploy_component({ deployment_id })` then cannot find the row and
      cannot activate the build — a stage that reported success is unactivatable. The fix is to
      require tracking availability for `activate: false` / `deployment_id` paths while keeping the
[...]

Issue #1778, opened 2026-07-13
CLI: `harper deploy --setup` — one-command deploy-key provisioning


dawsontoth:
## Problem

`create-harper` shipped `scripts/setup-deploy.mjs` to bootstrap private-repo deploy access. That belongs in the CLI — and, more importantly, the credential should be **sealed client-side** so the plaintext token never reaches the cluster, its logs, its operations journal, or replication.

## Proposal — `harper deploy setup=true`

A guided, client-side credential-provisioning flow (what `harper login` is for auth, but for deploy tokens):

1. `get_secrets_public_key` → the cluster's public key + fingerprint.
2. Source a token interactively — `gh auth token`, a pasted fine-grained PAT (Contents: Read-only), or `npm token create` for a private registry.
3. `encryptEnvelope(...)` → seal it **locally** into an `enc:v1:` envelope (pure client-side crypto; the plaintext never leaves the machine).
4. `set_secret { envelope, grants:[component] }` → store only ciphertext; the cluster decrypts it in-memory only at deploy/rollback time.
5. Print the `credentials` reference the deploy uses (`[{ host|registry, secret }]`).

Because the token is **durable** (unlike a CI `GITHUB_TOKEN`, revoked at job end), a re-resolve rollback works for as long as it's valid. One flow unifies **git-host** (`{host, …}`) and **npm-registry** (`{registry, …}`) credentials.

## Prototype

Branch [`claude/deploy-setup-sealed-token`](https://github.com/HarperFast/harper/tree/claude/deploy-setup-sealed-token) — `bin/deploySetup.ts` + a `harper deploy setup=true` intercept in `bin/harper.ts`. Parse + oxlint + prettier clean. Builds directly on the `credentials` array (#1797), the git-host credential (#1799), and the `hdb_secret` client-side envelope (#1717 / secretOperations); custody ships on by default (harper-pro#560).

Supersedes the SSH-deploy-key provisioning this issue originally described; the SSH path (harper-pro#570) remains a secondary alternative.

## Related — deploy-by-reference effort

- `HarperFast/harper#1849` — two-phase stage/activate + `revert_component`
- `HarperFast/harper#1850` — `harper deploy by_ref=true` (deploy by git reference)
- `HarperFast/harper#1851` — `harper deploy setup=true` (client-side sealed deploy credential)
- `HarperFast/harper#1876` — CI token auth + `harper login --for-ci`
[...]

Issue #1777, opened 2026-07-13
CLI: `harper deploy by_ref=true` — deploy by git reference


dawsontoth:
## Problem

`create-harper` scaffolds a per-project `scripts/deploy.mjs` to deploy by git reference; that logic is client-side (it reads the local git repo), so it belongs in the `harper` CLI — then no scaffolded project needs to carry a script.

## Proposal

Opt-in flag on `harper deploy` / `deploy_component`. When `package=` is absent **and** `by_ref=true` (or `ref=<committish>`) is passed, the CLI:

- resolves `owner/repo` from `GITHUB_REPOSITORY` or the `origin` remote,
- resolves the committish from `ref=`, else `GITHUB_SHA`, else `git rev-parse HEAD`,
- warns if the working tree is dirty,
- defaults `project` from `package.json`,
- sets `package=git+https://github.com/<owner>/<repo>.git#<sha>`.

Pinning by **SHA** (not tag) avoids cluster peers diverging on a moved tag. **git+https, not ssh** — a private clone is authenticated by a git-host token credential (#1799), which rides over HTTPS.

For a private repo, **`credential=github.com`** (or `credential=true`) attaches a `credentials` reference `[{ host, secret: "deploy.<component>.<host>" }]` — the durable token sealed once by `harper deploy setup=true` (#1778) — which the cluster resolves **in-memory** to authenticate the clone. A public repo needs no credential.

The **no-flag default stays the payload deploy** (backward-compatible). `by_ref`/`ref`/`credential` are transport-only (consumed client-side); `credentials` is a real operation field and is sent.

## Prototype

Branch [`claude/deploy-by-ref-flag`](https://github.com/HarperFast/harper/tree/claude/deploy-by-ref-flag) — `bin/cliOperations.ts`, in `PREPARE_OPERATION.deploy_component`. Parse + oxlint + prettier clean.

## Related — deploy-by-reference effort

- `HarperFast/harper#1849` — two-phase stage/activate + `revert_component`
- `HarperFast/harper#1850` — `harper deploy by_ref=true` (deploy by git reference)
- `HarperFast/harper#1851` — `harper deploy setup=true` (client-side sealed deploy credential)
- `HarperFast/harper#1876` — CI token auth + `harper login --for-ci`
- `HarperFast/harper-pro#594` — `add_ssh_key generate=true` (cluster-side keygen)
- `HarperFast/create-harper#118` — scaffolds this flow
[...]

PR #1850, opened 2026-07-17
feat(cli): `harper deploy by_ref=true` — deploy by git+https reference (+ credential)

dawsontoth:
> **Draft / prototype** for #1777. Part of the create-harper deploy-by-reference effort.

`harper deploy by_ref=true` (or `ref=<committish>`) resolves the app's GitHub `owner/repo` + commit from the local working copy (or GitHub Actions env) and deploys `git+https://github.com/<owner>/<repo>.git#<sha>` instead of packaging a payload blob. Pinning by **SHA** keeps cluster peers from diverging on a moved tag: peers resolve the package independently, so anything that can move is resolved client-side — an explicit `ref=` too, locally first and then via `git ls-remote`, failing closed rather than sending a name the cluster would resolve for itself.

For a private repo, `credential=true` attaches a `credentials` reference `[{ host, secret: "deploy.<component>.<host>" }]` — the durable token sealed once by `harper deploy setup=true` (#1778) — which the cluster resolves **in-memory** to authenticate the HTTPS clone (#1799). The host is derived from the package, so a credential can't be pointed at a host the clone never authenticates against. Public repos need no credential.

Under GitHub Actions, a `pull_request` run deploys the PR **head** (from the event payload) rather than `GITHUB_SHA` — that's the synthetic `refs/pull/<n>/merge` commit, which a plain clone can't fetch.

**Where:** `bin/cliOperations.ts`, `PREPARE_OPERATION.deploy_component`. The no-flag default stays the payload deploy (backward-compatible). `by_ref`/`ref`/`credential` are transport-only (consumed client-side); `credentials` is a real operation field and is sent.

Refs #1777, #1799, #641.

## Where to look

- **`resolveExplicitRef` adds a network call.** When a `ref=` doesn't resolve in the local checkout, this now shells out to `git ls-remote` (15s cap, non-interactive env so a private remote can't hang on a credential prompt) instead of passing the name through. Only fires on the explicit-ref path, and only when local resolution misses.
- **Fork PRs.** On a `pull_request` run the head repo is used, which for a fork means the cluster clones an untrusted repository. That's what "deploy this PR" means, but today it's a printed note rather than a gate — flagged on [kriszyp's thread](https://github.com/HarperFast/harper/pull/1850#discussion_r3683894307) and open to being made stricter.
- **`credential=<host>` is kept as compatibility**, not as an option: it's the spelling in the create-harper template and the docs PR, and it's now rejected unless it normalizes to the package host.

- `HarperFast/harper#1876` — CI token auth + `harper login --for-ci`
- `HarperFast/create-harper#118` — scaffolds this flow (updated to `credential=true`)
- `HarperFast/documentation#599` — two-phase deploy docs
- `HarperFast/documentation#616` — by-reference / sealed-credential / CI-auth docs (merged into #599's branch)
- `HarperFast/documentation#617` — **companion for this PR**: ref pinning, `credential=true`, GitHub Actions behavior

### DIFF
--- a/bin/cliOperations.ts
+++ b/bin/cliOperations.ts
@@ -10,6 +10,7 @@ import * as path from 'path';
 import * as fs from 'fs-extra';
 import * as YAML from 'yaml';
 import { Readable } from 'node:stream';
+import { execFileSync } from 'node:child_process';
 import { streamPackagedDirectory, packageDirectory, scanPackageDirectory } from '../components/packageComponent.ts';
 import { encode as encodeCbor } from 'cbor-x';
 import { buildMultipartBody } from './multipartBuilder.ts';
@@ -46,6 +47,11 @@ const TRANSPORT_ONLY_FIELDS = new Set([
 	'json',
 	'skip_node_modules',
 	'skip_symlinks',
+	// deploy-by-reference opt-in: consumed client-side to build `package` (and derive `credentials`),
+	// never sent to the server. (`credentials`, plural, IS a real operation field and is sent.)
+	'by_ref',
+	'ref',
+	'credential',
 ]);
 
 // Streaming (multipart upload + SSE progress) deploy was introduced in 5.1.0. A CLI at >=
@@ -249,10 +255,126 @@ function redactCredentials(req: any): any {
 	return redacted;
 }
 
-export { cliOperations, buildRequest, redactCredentials, refreshExpiredOperationToken };
+export {
+	cliOperations,
+	buildRequest,
+	redactCredentials,
+	refreshExpiredOperationToken,
+	resolveGitCommittish,
+	deriveGitSecretName,
+};
+
+// --- deploy-by-reference (opt-in via `by_ref=true` / `ref=<committish>`) ----------------------
+// Resolve the app's GitHub repo + commit from the local working copy (or GitHub Actions env) so
+// `harper deploy by_ref=true` deploys a pinned commit by reference instead of uploading a payload
+// blob. Client-side: only the runner has the git context. The no-flag default stays the payload deploy.
+
+function runGit(args: string[]): string {
+	return execFileSync('git', args, { encoding: 'utf8', stdio: ['ignore', 'pipe', 'ignore'] }).trim();
+}
+
+function resolveGitRepo(): string {
+	if (process.env.GITHUB_REPOSITORY) return process.env.GITHUB_REPOSITORY;
+	let url: string;
+	try {
+		url = runGit(['remote', 'get-url', 'origin']);
+	} catch {
+		throw new Error(
+			'deploy by_ref: no git `origin` remote found — push this project to GitHub, or pass an explicit package=.'
+		);
+	}
+	// git@github.com:owner/repo.git | https://github.com/owner/repo(.git) | ssh://…
+	const match = url.match(/github\.com[:/]+([^/]+\/[^/]+?)(?:\.git)?\/?$/i);
+	if (!match) throw new Error(`deploy by_ref: could not parse owner/repo from the origin remote: ${url}`);
+	return match[1];
+}
+
+function resolveGitCommittish(ref: unknown): string {
+	// buildRequest JSON-parses CLI args, so a numeric ref (e.g. `ref=1234567`) arrives as a number —
+	// coerce it back to a string rather than silently ignoring it and falling back to HEAD.
+	const refStr = typeof ref === 'string' || typeof ref === 'number' ? String(ref).trim() : '';
+	if (refStr.length > 0) return refStr; // explicit ref= wins
+	if (process.env.GITHUB_SHA) return process.env.GITHUB_SHA;
+	try {
+		return runGit(['rev-parse', 'HEAD']);
+	} catch {
+		throw new Error('deploy by_ref: could not resolve HEAD — make at least one commit, or pass ref=<sha|tag>.');
+	}
+}
+
+function warnIfWorkingTreeDirty(): void {
+	try {
+		if (runGit(['status', '--porcelain'])) {
+			process.stderr.write(
+				'warning: working tree has uncommitted changes — the cluster deploys the committed (and pushed) commit, so those changes are NOT included.\n'
+			);
+		}
+	} catch {
+		// Not a git repo; resolveGitRepo/resolveGitCommittish will surface a clearer error.
+	}
+}
+
+function defaultProjectName(projectPath: string): string {
+	try {
+		const pkg = JSON.parse(fs.readFileSync(path.join(projectPath, 'package.json'), 'utf8'));
+		if (typeof pkg.name === 'string' && pkg.name.length > 0) return pkg.name.replace(/^@[^/]+\//, '');
+	} catch {
+		// No/invalid package.json — fall back to the directory name.
+	}
+	return path.basename(projectPath);
+}
+
+// Mirror the server's `deriveGitSecretName` (secretOperations.ts) EXACTLY so the reference this
+// attaches matches the row `harper deploy setup=true` sealed and a literal-token deploy would use:
+// `deploy.<component>.git.<host>` (host via normalizeGitHost).
+function deriveGitSecretName(component: string, host: string): string {
+	const hostPart = host
+		.trim()
+		.replace(/^[a-z0-9+.-]+:\/\//i, '')
+		.replace(/^\/\//, '')
+		.replace(/\/.*$/, '')
+		.replace(/^[^@]*@/, '')
+		.toLowerCase()
+		.replace(/[^\w.-]+/g, '_');
+	const componentPart = String(component).replace(/[^\w.-]+/g, '_');
+	return `deploy.${componentPart}.git.${hostPart}`;
+}
+
+// Opt-in deploy-by-reference: resolve the pinned git ref (+ optional sealed credential) onto `req` —
+// a `git+https` package pinned by SHA, plus a `credentials` reference when `credential=` is set.
+// Exported for unit tests.
+export function prepareDeployByRef(req: any): void {
+	const repo = resolveGitRepo();
+	const committish = resolveGitCommittish(req.ref);
+	warnIfWorkingTreeDirty();
+	if (!req.project) req.project = defaultProjectName(process.cwd());
+	// git+https (not ssh): a private clone is authenticated by a git-host token credential (#1799),
+	// which rides over HTTPS. A public repo needs no credential at all.
+	req.package = `git+https://github.com/${repo}.git#${committish}`;
+	// `credential=github.com` (or `credential=true`) attaches the sealed-token reference the cluster
+	// resolves at fetch time — provision it once with `harper deploy setup=true`.
+	const credentialHost =
+		req.credential === true
+			? 'github.com'
+			: typeof req.credential === 'string' && req.credential.length > 0
+				? req.credential
+				: undefined;
+	if (credentialHost && req.credentials === undefined) {
+		req.credentials = [{ host: credentialHost, secret: deriveGitSecretName(req.project, credentialHost) }];
+	}
+	process.stderr.write(`Deploying "${req.project}" by reference: ${req.package}\n`);
+}
+
 const PREPARE_OPERATION: any = {
 	deploy_component: async (req) => {
 		if (req.package) {
+			return;
+		}
+
+		// Opt-in: deploy a pinned git commit by reference instead of packaging the working directory.
+		// Templates scaffold `by_ref=true`; without the flag the payload path below is unchanged.
+		if (req.by_ref || req.ref) {
+			prepareDeployByRef(req);
 			return;
 		}
 
--- a/bin/cliOperations.ts
+++ b/bin/cliOperations.ts
@@ -293,7 +293,20 @@ function resolveGitCommittish(ref: unknown): string {
 	// buildRequest JSON-parses CLI args, so a numeric ref (e.g. `ref=1234567`) arrives as a number —
 	// coerce it back to a string rather than silently ignoring it and falling back to HEAD.
 	const refStr = typeof ref === 'string' || typeof ref === 'number' ? String(ref).trim() : '';
-	if (refStr.length > 0) return refStr; // explicit ref= wins
+	if (refStr.length > 0) {
+		// Resolve an explicit ref= to an immutable SHA too, rather than passing the name through.
+		// Cluster peers resolve the package independently, so `ref=main` — or a tag that moves
+		// between one peer fetching and another restarting and re-fetching — would otherwise leave
+		// nodes running different commits. `^{commit}` also dereferences annotated tags.
+		try {
+			return runGit(['rev-parse', `${refStr}^{commit}`]);
+		} catch {
+			// Not resolvable locally (e.g. a ref that only exists on the remote). Pass it through
+			// and let the cluster resolve it — losing the pin, but a deploy that works beats a
+			// deploy that fails on a ref the user can see on the remote.
+			return refStr;
+		}
+	}
 	if (process.env.GITHUB_SHA) return process.env.GITHUB_SHA;
 	try {
 		return runGit(['rev-parse', 'HEAD']);
@@ -311,6 +324,25 @@ function warnIfWorkingTreeDirty(): void {
 		}
 	} catch {
 		// Not a git repo; resolveGitRepo/resolveGitCommittish will surface a clearer error.
+	}
+}
+
+// The likelier by_ref mistake isn't a dirty tree, it's committing and forgetting to push: the
+// cluster clones from the remote, so the SHA simply isn't there and the deploy fails server-side,
+// far from the CLI and with a much less obvious error. Checked against local remote-tracking refs,
+// so it costs no network round-trip — at the price of a false warning when the local view is stale,
+// which the message accounts for.
+function warnIfCommitNotPushed(committish: string): void {
+	try {
+		if (!runGit(['branch', '-r', '--contains', committish])) {
+			process.stderr.write(
+				`warning: commit ${committish.slice(0, 12)} isn't on any remote branch — push it, or the cluster ` +
+					"won't be able to clone it. (If you already pushed, run `git fetch` to refresh your remote refs.)\n"
+			);
+		}
+	} catch {
+		// Not a git repo, or a committish git can't resolve locally (a remote-only ref is expected
+		// to be absent here) — nothing useful to say, and the deploy itself surfaces real errors.
 	}
 }
 
@@ -347,6 +379,9 @@ export function prepareDeployByRef(req: any): void {
 	const repo = resolveGitRepo();
 	const committish = resolveGitCommittish(req.ref);
 	warnIfWorkingTreeDirty();
+	// Skipped under GITHUB_SHA: CI runs on a commit that is on the remote by construction, and a
+	// shallow/detached runner checkout has no remote-tracking branches to check against anyway.
+	if (!process.env.GITHUB_SHA) warnIfCommitNotPushed(committish);
 	if (!req.project) req.project = defaultProjectName(process.cwd());
 	// git+https (not ssh): a private clone is authenticated by a git-host token credential (#1799),
 	// which rides over HTTPS. A public repo needs no credential at all.
--- a/bin/cliOperations.ts
+++ b/bin/cliOperations.ts
@@ -55,6 +55,11 @@ const TRANSPORT_ONLY_FIELDS = new Set([
 	'credential',
 ]);
 
+// Values that are opaque strings, never JSON. buildRequest otherwise JSON-parses every value, which
+// silently rewrites a git ref that happens to look numeric: `ref=1.0` becomes the number 1 (and then
+// the string "1"), so a tag named "1.0" would be resolved as "1". Refs can't be anything but strings.
+const RAW_STRING_FIELDS = new Set(['ref']);
+
 // Streaming (multipart upload + SSE progress) deploy was introduced in 5.1.0. A CLI at >=
 // 5.1 talking to a server < 5.1 must not use it: the older server has no multipart body
 // parser (the upload is rejected) and its generic text/event-stream serializer emits a bare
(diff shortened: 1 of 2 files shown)


bin/cliOperations.ts
```diff
@@ -97,12 +103,110 @@ function operationFields(req: any): any {
...
+		);
+	}
+	// git@github.com:owner/repo.git | https://github.com/owner/repo(.git) | ssh://…
+	const match = url.match(/github\.com[:/]+([^/]+\/[^/]+?)(?:\.git)?\/?$/i);
+	if (!match) throw new Error(`deploy by_ref: could not parse owner/repo from the origin remote: ${url}`);
+	return match[1];
+}
+
+function resolveGitCommittish(ref: unknown): string {
+	if (typeof ref === 'string' && ref.length > 0) return ref; // explicit ref= wins
```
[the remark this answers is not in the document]
dawsontoth: Fixed — `resolveGitCommittish` now accepts a `number` and coerces via `String(ref).trim()`, so a numeric ref like `ref=1234567` (which `buildRequest` JSON-parses to a number) is used instead of being ignored and falling back to `GITHUB_SHA`/`HEAD`. Covered by a unit test. (2f12b9f00)

bin/cliOperations.ts
```diff
@@ -97,12 +103,110 @@ function operationFields(req: any): any {
...
+
 const PREPARE_OPERATION: any = {
 	deploy_component: async (req) => {
 		if (req.package) {
 			return;
 		}
 
+		// Opt-in: deploy a pinned git commit by reference instead of packaging the working directory.
+		// Templates scaffold `by_ref=true`; without the flag the payload path below is unchanged.
+		if (req.by_ref || req.ref) {
```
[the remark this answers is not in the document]
dawsontoth: Added — extracted the by_ref branch into an exported `prepareDeployByRef(req)` and exported `resolveGitCommittish`/`deriveGitSecretName`, then added `unitTests/bin/cliOperations.test.js` cases: the `git+https://…#<sha>` package build from a resolved repo/SHA, `ref=` overriding HEAD, numeric-ref coercion, `credential=true`/`credential=<host>` producing the `credentials` reference, and the `deploy.<component>.git.<host>` secret name. 12/12 passing. (2f12b9f00)

bin/cliOperations.ts
```diff
@@ -249,13 +255,129 @@ function redactCredentials(req: any): any {
...
+	const match = url.match(/github\.com[:/]+([^/]+\/[^/]+?)(?:\.git)?\/?$/i);
+	if (!match) throw new Error(`deploy by_ref: could not parse owner/repo from the origin remote: ${url}`);
+	return match[1];
+}
+
+function resolveGitCommittish(ref: unknown): string {
+	// buildRequest JSON-parses CLI args, so a numeric ref (e.g. `ref=1234567`) arrives as a number —
+	// coerce it back to a string rather than silently ignoring it and falling back to HEAD.
+	const refStr = typeof ref === 'string' || typeof ref === 'number' ? String(ref).trim() : '';
+	if (refStr.length > 0) return refStr; // explicit ref= wins
```
cb1kenobi: **Medium: explicit `ref=` is used verbatim, not resolved to a SHA — defeats the PR's own pin-by-SHA guarantee**

The default path (`git rev-parse HEAD` / `GITHUB_SHA`) resolves to an immutable full SHA, but an explicit `ref=<branch|tag>` is passed straight through into `git+https://…#<committish>`. Deploying `ref=main` (or a moved tag — the test even encodes `ref=v9.9.9`) means cluster peers cloning at different times, or a peer re-cloning after restart, can resolve *different* commits. That is exactly the divergence the PR body says SHA-pinning prevents ("peers can't diverge on a moved tag").

Suggested fix: resolve the ref locally before building the package so the same pinning applies to explicit refs, e.g.

```suggestion
	if (refStr.length > 0) {
		try {
			return runGit(['rev-parse', `${refStr}^{commit}`]); // pin explicit branch/tag to an immutable SHA
		} catch {
			return refStr; // not resolvable locally (e.g. remote-only ref); send as-is
		}
	}
```

—
_Generated by Barber AI_

bin/cliOperations.ts
```diff
@@ -249,13 +255,129 @@ function redactCredentials(req: any): any {
...
+	return `deploy.${componentPart}.git.${hostPart}`;
+}
+
+// Opt-in deploy-by-reference: resolve the pinned git ref (+ optional sealed credential) onto `req` —
+// a `git+https` package pinned by SHA, plus a `credentials` reference when `credential=` is set.
+// Exported for unit tests.
+export function prepareDeployByRef(req: any): void {
+	const repo = resolveGitRepo();
+	const committish = resolveGitCommittish(req.ref);
+	warnIfWorkingTreeDirty();
```
cb1kenobi, line 378: **Medium: dirty-tree warning doesn't cover the likelier failure — an unpushed commit**

`warnIfWorkingTreeDirty()` catches uncommitted changes, but the common by_ref mistake is committing locally and deploying `HEAD` without pushing. The resolved SHA won't exist on `origin`, so the cluster's `git+https://…#<sha>` clone fails server-side with an error far from the CLI, on the default (no-`ref`) path. A cheap client-side check would surface it where the user can act on it.

Suggested fix: after resolving `committish`, verify it's reachable on the remote (e.g. `git branch -r --contains <sha>` or `git ls-remote origin <ref>`) and warn/error if not, alongside the existing dirty-tree warning.

—
_Generated by Barber AI_

bin/cliOperations.ts
dawsontoth: Fixed in `4118b71` — you're right, and this was a real contradiction between the PR's own stated guarantee and its behavior. Explicit refs now resolve the same way HEAD does:

```ts
return runGit(['rev-parse', `${refStr}^{commit}`]);
```

`^{commit}` also dereferences annotated tags, so a tag object resolves to the commit rather than the tag SHA. The unresolvable case still passes through, as your suggestion had it — a ref that exists only on the remote should be resolved by the cluster rather than failing on something the user can plainly see. That trades the pin away in a narrow case, which felt better than the alternative; say the word if you'd rather it hard-fail.

The existing `ref=v9.9.9` test was passing for the wrong reason (that tag doesn't resolve in this checkout, so it took the fallback path). Ref resolution is now tested against a purpose-built temp repo with a known tag and commit, so "resolves to a SHA" is asserted against a real git resolution instead of whatever tags happen to exist wherever the tests run.

bin/cliOperations.ts
dawsontoth: Added in `4118b71`. Agreed this is the likelier mistake — and its failure mode is worse, since it surfaces server-side during the clone rather than anywhere near the command the user just ran.

`warnIfCommitNotPushed` uses `git branch -r --contains <sha>`, per your suggestion, so it stays a local lookup with no network round-trip. The tradeoff is a false warning when remote-tracking refs are stale, so the message says how to clear it:

> warning: commit abc123def456 isn't on any remote branch — push it, or the cluster won't be able to clone it. (If you already pushed, run `git fetch` to refresh your remote refs.)

One deliberate exception: the check is skipped when `GITHUB_SHA` is set. CI runs on a commit that's on the remote by construction, and a shallow or detached runner checkout may have no remote-tracking branches at all — so there it would be pure false-positive noise. Both the warning and the CI skip have tests.

bin/cliOperations.ts
```diff
@@ -249,13 +255,164 @@ function redactCredentials(req: any): any {
...
+	const hostPart = host
+		.trim()
+		.replace(/^[a-z0-9+.-]+:\/\//i, '')
+		.replace(/^\/\//, '')
+		.replace(/\/.*$/, '')
+		.replace(/^[^@]*@/, '')
+		.toLowerCase()
+		.replace(/[^\w.-]+/g, '_');
+	const componentPart = String(component).replace(/[^\w.-]+/g, '_');
+	return `deploy.${componentPart}.git.${hostPart}`;
```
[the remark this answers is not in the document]
dawsontoth: Adopted in the latest push — good call, and I verified the premise before taking it: `components/gitCredentialServer.ts` imports only `node:*` builtins plus `utility/errors/hdbError.ts` and `utility/logging/harper_logger.ts` (both already used in `bin/`), has no top-level side effects, and the server's own `deriveGitSecretName` composes it as `normalizeGitHost(host).replace(/[^\\w.-]+/g, '_')` — exactly your suggestion.

Confirmed `harper version` still starts and the 53 unit tests pass, so the extra import costs the CLI nothing.

You're right that this was the drift risk my own "mirror EXACTLY" comment was flagging — better to delete the duplicate than to warn about it.

Note `bin/deploySetup.ts` on #1851 carries the same duplication (it also covers the npm-registry variant, so it needs `deriveRegistrySecretName`'s chain too). I've left it for that PR rather than reaching across branches; it has unit tests pinning both names to the server convention in the meantime.

bin/cliOperations.ts
```diff
@@ -249,13 +256,160 @@ function redactCredentials(req: any): any {
...
+	const refStr = typeof ref === 'string' || typeof ref === 'number' ? String(ref).trim() : '';
+	if (refStr.length > 0) {
+		// Resolve an explicit ref= to an immutable SHA too, rather than passing the name through.
+		// Cluster peers resolve the package independently, so `ref=main` — or a tag that moves
+		// between one peer fetching and another restarting and re-fetching — would otherwise leave
+		// nodes running different commits. `^{commit}` also dereferences annotated tags.
+		try {
+			return runGit(['rev-parse', `${refStr}^{commit}`]);
+		} catch {
+			// Not resolvable locally (e.g. a ref that only exists on the remote). Pass it through
```
kriszyp, line 305: Could we fail closed here instead of returning an unresolved name? This preserves the exact divergence the SHA pin is meant to eliminate: in a shallow checkout without `origin/release`, `ref=release` passes through; peer A can clone commit A, then the branch moves before peer B (or a restarted peer) clones commit B. Please resolve the remote ref to a full commit SHA (fetch/`ls-remote` plus annotated-tag handling), or reject it with guidance to fetch/pass a SHA. Passing through is safe only when the input is already validated as a full object ID.

bin/cliOperations.ts
```diff
@@ -249,13 +256,160 @@ function redactCredentials(req: any): any {
...
+// Opt-in deploy-by-reference: resolve the pinned git ref (+ optional sealed credential) onto `req` —
+// a `git+https` package pinned by SHA, plus a `credentials` reference when `credential=` is set.
+// Exported for unit tests.
+export function prepareDeployByRef(req: any): void {
+	const repo = resolveGitRepo();
+	const committish = resolveGitCommittish(req.ref);
+	warnIfWorkingTreeDirty();
+	// Skipped under GITHUB_SHA: CI runs on a commit that is on the remote by construction, and a
+	// shallow/detached runner checkout has no remote-tracking branches to check against anyway.
+	if (!process.env.GITHUB_SHA) warnIfCommitNotPushed(committish);
```
kriszyp, line 381: `GITHUB_SHA` is not always a commit a normal clone of `GITHUB_REPOSITORY` can fetch. For a `pull_request` workflow GitHub sets it to the synthetic merge commit under `refs/pull/<n>/merge`; the server later runs a plain `git clone` and `git checkout <sha>`, whose default refspec does not include pull refs, so this deploy fails after reaching the cluster while this check is skipped. GitHub documents that event behavior here: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request. Please verify that the selected SHA is reachable through a cloneable branch/tag, accept event-specific repo/ref inputs (for example the PR head repo and SHA), or fail early with a targeted error.

bin/cliOperations.ts
```diff
@@ -249,13 +256,160 @@ function redactCredentials(req: any): any {
...
+	if (!req.project) req.project = defaultProjectName(process.cwd());
+	// git+https (not ssh): a private clone is authenticated by a git-host token credential (#1799),
+	// which rides over HTTPS. A public repo needs no credential at all.
+	req.package = `git+https://github.com/${repo}.git#${committish}`;
+	// `credential=github.com` (or `credential=true`) attaches the sealed-token reference the cluster
+	// resolves at fetch time — provision it once with `harper deploy setup=true`.
+	const credentialHost =
```
kriszyp, line 388: Can this API derive the credential host from the package URL and make `credential` boolean-only? The package is always `github.com`, but `credential=gitlab.com` currently builds a valid credential reference for GitLab; the clone then asks for GitHub credentials, receives none, and a private deploy fails as if unauthenticated. Deriving `github.com` removes an avoidable external option and prevents inconsistent package/credential pairs; alternatively, reject any explicit host that does not normalize to `github.com`.

kriszyp (commented):
Great feature.

bin/cliOperations.ts
```diff
@@ -249,13 +256,160 @@ function redactCredentials(req: any): any {
...
+// The likelier by_ref mistake isn't a dirty tree, it's committing and forgetting to push: the
+// cluster clones from the remote, so the SHA simply isn't there and the deploy fails server-side,
+// far from the CLI and with a much less obvious error. Checked against local remote-tracking refs,
+// so it costs no network round-trip — at the price of a false warning when the local view is stale,
+// which the message accounts for.
+function warnIfCommitNotPushed(committish: string): void {
+	try {
+		if (!runGit(['branch', '-r', '--contains', committish])) {
+			process.stderr.write(
+				`warning: commit ${committish.slice(0, 12)} isn't on any remote branch — push it, or the cluster ` +
```
cb1kenobi, line 340: nit: 7 is the standart length for a commit hash.
```suggestion
				`warning: commit ${committish.slice(0, 7)} isn't on any remote branch — push it, or the cluster ` +
```

cb1kenobi (approved):

bin/cliOperations.ts
dawsontoth: Fixed in `168116f`. You're right, and the rationale I'd written into the comment was wrong on its own terms: the "deploy that works" in the pass-through case is precisely the one that lets peers land on different commits.

An explicit `ref=` now resolves local-first (`rev-parse --verify <ref>^{commit}`), then falls back to `git ls-remote` for a ref this checkout doesn't have — the shallow-clone case you named, where `origin/release` was never fetched — and throws when neither can name a commit. Pass-through survives only for a full 40/64-hex object ID, which is the case you called safe.

On the annotated-tag handling: `ls-remote` only emits the peeled `^{}` line when a pattern matches *that* line, so the peel patterns are requested explicitly. Without them `ls-remote origin v2.0.0` returns the **tag object's** ID, which isn't a commit the cluster can check out — I confirmed that against real git before relying on it, and the test asserts the resolved SHA is the commit and *not* the tag object. Resolution is tested against a local bare repo standing in for `origin`, carrying a branch and an annotated tag that exist only there.

Tradeoff worth flagging: this adds a network round trip on the explicit-ref path when local resolution misses. It's capped at 15s and runs non-interactively (`GIT_TERMINAL_PROMPT=0`, no-op askpass, `ssh -oBatchMode=yes`) so a credential prompt on a private remote can't hang the CLI instead of erroring.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

bin/cliOperations.ts
dawsontoth: Fixed in `168116f` — thanks for the docs pointer, this was a real hole and I'd have shipped it.

`resolveGitTarget` now detects a pull ref (`GITHUB_REF` starting with `refs/pull/`) and takes the commit from `GITHUB_EVENT_PATH` instead of `GITHUB_SHA`: `pull_request.head.sha` **and** `pull_request.head.repo.full_name` together, so a fork PR deploys from the fork rather than pairing a head SHA with a base repo that doesn't contain it. When the payload can't be read it fails early, naming `ref=${{ github.event.pull_request.head.sha }}` as the fix, rather than failing at clone time out on the cluster.

I took the event-input option rather than a general reachability check: `refs/pull/*` is the one documented case where `GITHUB_SHA` isn't on a cloneable branch (`pull_request_target` gets the base-branch tip, push/tag/dispatch get a branch or tag tip), so a reachability probe would be a network call for something the ref name already tells us. That reasoning replaced the comment on the push-warning's CI skip, which previously just asserted CI commits are pushed "by construction" — which was the unexamined premise behind this bug.

One thing worth your eye rather than mine: deploying a fork's head is what "deploy this PR" means, but it does mean the cluster clones from an untrusted repository when a workflow runs `deploy` on a fork PR. Today that prints a note naming the head repo; it isn't a gate. Happy to make it refuse a fork head unless `ref=` is passed explicitly if you'd rather that be a decision the workflow author has to make.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

bin/cliOperations.ts
dawsontoth: Fixed in `168116f`, taking the alternative you offered: the credential host is derived from the package host, and an explicit `credential=` is accepted only when it normalizes to that same host — `credential=gitlab.com` now throws instead of building a reference the clone never asks for.

I stopped short of boolean-only because `credential=github.com` is the spelling in the create-harper template and the docs PR, and validating it prevents the inconsistent pairing just as well as removing it does. `credential=true` is now the form the comment and the error message steer people to, so the string spelling is compatibility rather than a documented option.

The comparison runs through the server's `normalizeGitHost`, so `credential=https://GitHub.com/acme/demo` is accepted as `github.com`, and the package host itself is a single constant that both the URL and the credential are built from — if repo resolution ever learns a second host, the credential follows it rather than needing a parallel edit.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

bin/cliOperations.ts
dawsontoth: Taken in `168116f`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

bin/cliOperations.ts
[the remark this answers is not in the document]
dawsontoth: Re-reading this thread on a later pass: the fix I landed here covered your `ref=1234567` example but not the `ref=1.0` one in the same comment. `JSON.parse('1.0')` is the number `1`, and `String(1)` is `"1"` — so a tag named `1.0` was resolved as `1`, and no coercion downstream can recover what the parse discarded (`1.10` → `"1.1"`, `1e3` → `"1000"`).

Fixed properly in `36dae1a`: `ref` skips the JSON parse entirely in `buildRequest`, since a git ref is an opaque string and never JSON. The numeric coercion stays as a guard, because `prepareDeployByRef` is exported and callable with a hand-built req. Tested against `1.0`, `1.10`, `1e3`, and `1234567`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

bin/cliOperations.ts
dawsontoth: Heads-up since you approved after this thread: the behavior we settled on here has changed.

My reply above kept the pass-through for a ref that doesn't resolve locally — your suggestion's `catch { return refStr }` — and offered to hard-fail instead. [@kriszyp asked for exactly that](https://github.com/HarperFast/harper/pull/1850#discussion_r3683894303), on the grounds that a shallow checkout without `origin/release` is the common case, not a rare one, so the pass-through preserved the divergence the pin exists to prevent.

As of `168116f` an unresolvable `ref=` now falls back to `git ls-remote` and then throws, rather than being sent as a name. A full 40/64-hex object ID still passes through — it can't move. Nothing about the local-resolution path you suggested changed.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

bin/cliOperations.ts
```diff
@@ -249,13 +261,292 @@ function redactCredentials(req: any): any {
...
+		]);
+	} catch {
+		return undefined; // unreachable, unauthenticated, or too slow to answer
+	}
+	const shaByRef = new Map<string, string>();
+	for (const line of output.split('\n')) {
+		const [sha, name] = line.split('\t');
+		if (sha && name) shaByRef.set(name, sha);
+	}
+	if (shaByRef.size === 1) return shaByRef.values().next().value; // an unambiguous match, whatever it matched
```
kriszyp: Could remote resolution accept only branch and tag refs here? Because the raw `ref` pattern is also sent to `ls-remote`, an explicit value such as `refs/pull/42/head` can yield exactly one match and return from this line. The resulting SHA is immutable but still unavailable to the server’s plain clone refspec, so deployment fails later in the cluster—the same reachability class fixed for automatic `GITHUB_SHA`. Please map exact `refs/heads/*` and `refs/tags/*` inputs to those namespaces, and reject other remote namespaces (or otherwise verify clone reachability) instead of accepting the single arbitrary match.

<sub>🤖 Reviewed by KrAIs</sub>

kriszyp (approved):
Looks good, I think there is remaining suggestiong about verifying refs.

bin/cliOperations.ts
dawsontoth: Fixed in `2738b40`. You're right, and I'd actually seen this case while writing the function and talked myself out of it — reasoning that an explicit `refs/pull/42/head` is the user asking for it on purpose. That was inconsistent: if `refs/pull/<n>/merge` is unreachable enough to substitute automatically, then `refs/pull/<n>/head` is unreachable enough to reject when typed.

Every `ls-remote` pattern is now namespace-qualified — `refs/tags/<ref>`, its `^{}` peel, and `refs/heads/<ref>` for a bare name; the ref itself plus its peel when already fully qualified. Passing the bare `ref` as its own pattern is what let an arbitrary namespace match, and the "lone match wins" shortcut that returned it is gone with it.

One thing your comment scoped to the remote path that turned out to be broader: the same hole exists locally. `git rev-parse --verify refs/pull/42/head^{commit}` succeeds in any checkout that has fetched the pull ref, and that runs *before* remote resolution — so constraining only `ls-remote` would have left the shorter path open. The namespace check now runs ahead of local resolution, and the test pushes the pull ref to the fixture's remote **and** creates it locally, so it would resolve either way if the check weren't there.

Where I stopped: a bare 40/64-hex SHA that happens to be unreachable is still accepted. An object ID carries no namespace to inspect, so catching it would need the reachability probe you offered as the alternative — a network round trip on every SHA deploy, including the overwhelmingly common reachable one. The error for a rejected namespace points at "a branch or tag the commit is on" rather than at a SHA, so it isn't guidance straight back into the same hole. Happy to add the probe if you'd rather pay for it.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

kriszyp (approved):

Issue #599, opened 2026-05-19
Remove the upgrade confirmation prompt for reversible upgrades


kriszyp:
The upgrade CLI prompts the user for confirmation. This is unnecessary friction unless the specific upgrade is known to preclude or cause problems for a subsequent downgrade.

## Ask

- Skip the upgrade confirmation prompt by default.
- Only prompt (or require an explicit `--force` / `--no-downgrade` flag) when the target version makes downgrade lossy or unsupported (e.g. major version bumps with on-disk format changes).
- Maintain a list/table of which upgrades require the prompt so this stays consistent across releases.

## Acceptance criteria

- `harperdb upgrade` (or equivalent) runs without prompting in the common case (minor / patch upgrades that are reversible).
- Cross-major or format-changing upgrades still prompt or require an explicit flag.
- CI automation that runs upgrades doesn't have to pipe `yes` to it.

## Notes

- A test should assert non-interactive upgrade succeeds for the reversible case.

---

Tracked in Jira: [CORE-2980](https://harperdb.atlassian.net/browse/CORE-2980)

🤖 Filed by Claude on behalf of Kris.

PR #1851, opened 2026-07-17
feat(cli): `harper deploy setup=true` — seal a durable deploy credential client-side

dawsontoth:
**Ready for review** — no longer a draft/RFC. Implements #1778; part of the create-harper deploy-by-reference effort. Rebased on current `main`, with all review threads to date resolved.

`harper deploy setup=true` — a guided, client-side credential-provisioning flow (what `harper login` is for auth, but for deploy tokens):

1. `get_secrets_public_key` → the cluster's public key + fingerprint.
2. Source a token interactively — a pasted fine-grained PAT (Contents: Read-only, the recommended default), your `gh` CLI session token, or `npm token create` for a private registry.
3. `encryptEnvelope(...)` → seal it **locally** into an `enc:v1:` envelope (pure client-side crypto; the plaintext never leaves the machine).
4. `set_secret { envelope, processEnv: false }` → store only ciphertext, in the component-scoped tier; the cluster decrypts it in-memory only at deploy/rollback time.
5. `grant_secret { component }` → let the component resolve it. Two calls rather than one so the server merges grants under its own lock (see below).
6. Print the `credentials` reference the deploy uses.

Because the token is **durable** (unlike a CI `GITHUB_TOKEN`, revoked at job end), re-deploying an older ref later works without re-entering it. Rollback itself is credential-free in the common case — HarperFast/harper#1849's `revert_component` swaps to the retained previous build rather than re-fetching; the sealed token is only needed to re-fetch an *arbitrary* older ref. One flow unifies git-host (`{host, …}`) and npm-registry (`{registry, …}`) credentials.

**Where:** new `bin/deploySetup.ts` + a `harper deploy setup=true` intercept in `bin/harper.ts`. Builds on the `credentials` array (#1797), the git-host credential (#1799), and the `hdb_secret` client-side envelope (#1717); custody ships on by default (harper-pro#560).

## Where to look

**One derivation, shared with the server** (`utility/componentNames.ts`, new). Every name this flow derives has to match what the deploy re-derives from its own request, and a mismatch is silent — the seal lands in an `hdb_secret` row the deploy never reads, or is granted to a project it isn't running as. So host normalization, the `deploy.<component>[.git].<key>` secret names, project-name resolution and the project/host grammars live in one dependency-free module, consumed by `components/gitCredentialServer.ts`, `components/secretOperations.ts`, `components/operations.js`, `components/operationsValidation.js` and the CLI. The two `components/` modules re-export their previous names, so existing importers and their tests are unchanged, and `getProjectNameFromPackage` moved out of `operations.js` with its body unchanged (renamed, plus type annotations). This also removed the copy of `deriveGitSecretName` that `deploy by_ref=true` (#1850) had grown independently in `bin/cliOperations.ts` — verified byte-identical output across host/component spellings before swapping.

**`token=` is registered as secret-bearing** (`bin/cliOperations.ts`, `server/serverHelpers/serverUtilities.ts`). This PR introduced `token=` as an arg carrying a durable PAT. `harper deploy setup token=…` — bare `setup`, which `buildRequest` drops — falls through to an ordinary deploy, and the token was reaching `logger.trace` unredacted, the `deploy_component` body, and the server's operations log. Now in `SECRET_FIELDS` + `TRANSPORT_ONLY_FIELDS`, stripped server-side, and a top-level `token` on a deploy fails with a pointer to `setup=true` rather than silently deploying.

**Grants are merged by the server, not the client.** `set_secret` *replaces* grants, so sending `grants: [component]` on a rotation silently revoked a grant added separately with `grant_secret` (a monorepo's second component). Omitting the field keeps the stored list under `withSecretLock`; idempotent `grant_secret` follows. `processEnv: false` is explicit and load-bearing: without it, `set_secret` inherits a stored global tier and would write the pasted token where every component and child process reads it.

## Open for reviewers

**This changes #1850's deployed project name.** `prepareDeployByRef` resolved its own default via a `package.json`-preferring `defaultProjectName`; it now uses the shared `directoryProjectName`, so a by-reference deploy in a checkout whose directory and package name differ deploys under the directory name. I aligned that direction because only the directory default is common to all three deploy paths (npm-registry credentials are consumed by payload deploys too), but it is a behavior change to merged code and the #1850 author may want to weigh in.

Tests: `unitTests/utility/componentNames.test.js` (new), plus cases in `unitTests/bin/deploySetup.test.js` and `unitTests/bin/cliOperations.test.js` covering the two-call grant protocol, the scoped-tier pin, `token=` redaction and body-stripping, host canonicalization, and that setup's project default agrees with `prepareDeployByRef`'s.

Refs #1778, #1797, #1799.

- `HarperFast/harper#1876` — CI token auth + `harper login --for-ci`
- `HarperFast/create-harper#118` — scaffolds this flow
- `HarperFast/documentation#599` — two-phase deploy, revert, by-reference deploys, sealed credentials, `add_ssh_key generate`, OIDC (all v5.3.0)
- `HarperFast/documentation#630` — CI token credentials in the canonical auth precedence (v5.2.0, shipped; mergeable independently)
- `HarperFast/documentation#617` — `by_ref` ref pinning, `credential=true`, GitHub Actions behavior
- `HarperFast/documentation#616` — merged into #599's branch on 2026-07-30; its content now lives in the three PRs above

<sub>🤖 Description and implementation assisted by Claude Code (Opus 5).</sub>

### DIFF
--- a/bin/cliOperations.ts
+++ b/bin/cliOperations.ts
@@ -507,7 +507,6 @@ function warnIfCommitNotPushed(committish: string): void {
 	}
 }
 
-
 // The credential only helps if it's for the host the package is cloned from: a `credential=gitlab.com`
 // against a github.com package builds a valid-looking reference the clone never asks for, and the
 // private deploy then fails as if nothing were configured. So the host comes from the package rather
--- a/bin/cliOperations.ts
+++ b/bin/cliOperations.ts
@@ -56,9 +56,10 @@ const TRANSPORT_ONLY_FIELDS = new Set([
 	'ref',
 	'credential',
 	// `deploy setup=true`'s token, read off the parsed request by deploySetup and sealed locally. No
-	// operation takes a top-level `token` (the one in `credentials[]` is nested and stripped with it),
-	// so keeping it out of every body costs nothing and means a mistyped `setup` — which parses as a
-	// bare word and falls through to a real deploy — can't carry a PAT to the server.
+	// operation takes a *top-level* `token`, so keeping it out of every body costs nothing and means a
+	// mistyped `setup` — which parses as a bare word and falls through to a real deploy — can't carry a
+	// PAT to the server. Distinct from `credentials[].token`, which is nested inside a field that IS
+	// sent (ingestCredentials seals it server-side) and is only kept out of the operations log.
 	'token',
 ]);
 
--- a/bin/cliOperations.ts
+++ b/bin/cliOperations.ts
@@ -7,7 +7,6 @@ envMgr.initSync();
 import * as terms from '../utility/hdbTerms.ts';
 import { httpRequest } from '../utility/common_utils.ts';
 import * as fs from 'fs-extra';
-import * as path from 'node:path';
 import * as YAML from 'yaml';
 import { Readable } from 'node:stream';
 import { execFileSync } from 'node:child_process';
@@ -56,6 +55,11 @@ const TRANSPORT_ONLY_FIELDS = new Set([
 	'by_ref',
 	'ref',
 	'credential',
+	// `deploy setup=true`'s token, read off the parsed request by deploySetup and sealed locally. No
+	// operation takes a top-level `token` (the one in `credentials[]` is nested and stripped with it),
+	// so keeping it out of every body costs nothing and means a mistyped `setup` — which parses as a
+	// bare word and falls through to a real deploy — can't carry a PAT to the server.
+	'token',
 ]);
 
 /**
@@ -253,7 +257,7 @@ function resolveTransportCredentials(req: any, urlCredentials: { username: strin
 // file — the command line already exposes it to shell history and process listings; the log
 // shouldn't be a third copy. This list is by field name, not exhaustive — any future secret-bearing
 // arg (a token, a key) needs to be added here explicitly, or it will reach logger.trace unredacted.
-const SECRET_FIELDS = new Set(['auth_password', 'password']);
+const SECRET_FIELDS = new Set(['auth_password', 'password', 'token']);
 
 // `target=https://admin:secret@host` carries a password too, so masking the userinfo is part of
 // making a target printable — the same string is echoed by the "Connecting to ..." line.
@@ -503,15 +507,6 @@ function warnIfCommitNotPushed(committish: string): void {
 	}
 }
 
-function defaultProjectName(projectPath: string): string {
-	try {
-		const pkg = JSON.parse(fs.readFileSync(path.join(projectPath, 'package.json'), 'utf8'));
-		if (typeof pkg.name === 'string' && pkg.name.length > 0) return pkg.name.replace(/^@[^/]+\//, '');
-	} catch {
-		// No/invalid package.json — fall back to the directory name.
-	}
-	return path.basename(projectPath);
-}
 
 // The credential only helps if it's for the host the package is cloned from: a `credential=gitlab.com`
 // against a github.com package builds a valid-looking reference the clone never asks for, and the
@@ -543,7 +538,11 @@ export function prepareDeployByRef(req: any): void {
 	// left is pushed by construction — and a shallow/detached runner checkout has no remote-tracking
 	// branches to check it against anyway.
 	if (!process.env.GITHUB_SHA) warnIfCommitNotPushed(committish);
-	if (!req.project) req.project = defaultProjectName(process.cwd());
+	// The same default the payload deploy and `deploy setup=true` use. A by-reference deploy is the
+	// flow the sealed credential feeds, and `set_secret`'s grant is keyed by project name, so a
+	// package.json-derived name here would ask for a secret granted to a different name whenever a
+	// checkout directory and its package name disagree.
+	if (!req.project) req.project = directoryProjectName();
 	// git+https (not ssh): a private clone is authenticated by a git-host token credential (#1799),
 	// which rides over HTTPS. A public repo needs no credential at all.
 	req.package = `git+https://${GIT_PACKAGE_HOST}/${repo}.git#${committish}`;
--- a/bin/cliOperations.ts
+++ b/bin/cliOperations.ts
@@ -6,19 +6,22 @@ import * as envMgr from '../utility/environment/environmentManager.ts';
 envMgr.initSync();
 import * as terms from '../utility/hdbTerms.ts';
 import { httpRequest } from '../utility/common_utils.ts';
-import * as path from 'path';
 import * as fs from 'fs-extra';
+import * as path from 'node:path';
 import * as YAML from 'yaml';
 import { Readable } from 'node:stream';
 import { execFileSync } from 'node:child_process';
 import { streamPackagedDirectory, packageDirectory, scanPackageDirectory } from '../components/packageComponent.ts';
-import { normalizeGitHost } from '../components/gitCredentialServer.ts';
 import { encode as encodeCbor } from 'cbor-x';
 import { buildMultipartBody } from './multipartBuilder.ts';
 import { parseSSE } from './sseConsumer.ts';
 import { DeployRenderer } from './deployRenderer.ts';
 import { getHdbPid } from '../utility/processManagement/processManagement.js';
 import { initConfig, getConfigPath } from '../config/configUtils.ts';
+// The `deploy setup` seal and this by-reference flag both name the same hdb_secret row, and the
+// server re-derives it from its own request — so the derivation lives in one dependency-free module
+// rather than being restated per caller (it also keeps `components/` off the CLI's import graph).
+import { deriveGitSecretName, directoryProjectName, normalizeGitHost } from '../utility/componentNames.ts';
 
 const OP_ALIASES = { deploy: 'deploy_component', package: 'package_component' };
 
@@ -35,16 +38,16 @@ const LOCAL_NOT_RUNNING_MESSAGE = 'Harper is not running. Use `harperdb run` (or
 // the SSE parser sees no events.
 const SSE_OPERATIONS = new Set(['deploy_component']);
 
+// The fields that decide *where* an operation connects and *as whom* — see transportContext().
+const CONNECTION_FIELDS = ['target', 'auth_username', 'auth_password', 'rejectUnauthorized'];
+
 // Properties on `req` that the CLI itself uses for transport/UX, not the operations API.
 // They never get serialized into the request body. `username`/`password` are deliberately
 // NOT here: those args are payload fields (e.g. the user add_user/alter_user create/alter),
 // not transport — use `auth_username`/`auth_password` (or env-var/`harper login` auth) to
 // authenticate as a different user than the one being operated on.
 const TRANSPORT_ONLY_FIELDS = new Set([
-	'target',
-	'auth_username',
-	'auth_password',
-	'rejectUnauthorized',
+	...CONNECTION_FIELDS,
 	'json',
 	'skip_node_modules',
 	'skip_symlinks',
@@ -54,6 +57,21 @@ const TRANSPORT_ONLY_FIELDS = new Set([
 	'ref',
 	'credential',
 ]);
+
+/**
+ * The connection half of a parsed CLI request, as a base for a *different* operation issued on the
+ * same command's behalf (`harper deploy setup=true` calls get_secrets_public_key and set_secret).
+ * Carrying these over keeps the sub-operation on the caller's target, identity, and TLS strictness;
+ * rebuilding a request with only `target` would silently authenticate with a saved login token
+ * instead of the credentials that were passed explicitly, and would fail against a self-signed
+ * target. Only these fields come along, so the originating command's own args (a deploy's `token`,
+ * `package`, …) never reach the sub-operation's body.
+ */
+function transportContext(req: any): any {
+	const context: any = {};
+	for (const field of CONNECTION_FIELDS) if (req[field] !== undefined) context[field] = req[field];
+	return context;
+}
 
 // Values that are opaque strings, never JSON. buildRequest otherwise JSON-parses every value, which
 // silently rewrites a git ref that happens to look numeric: `ref=1.0` becomes the number 1 (and then
@@ -266,6 +284,7 @@ export {
 	buildRequest,
 	redactCredentials,
 	refreshExpiredOperationToken,
+	transportContext,
 	resolveGitTarget,
 	resolveCredentialHost,
 	deriveGitSecretName,
@@ -494,18 +513,6 @@ function defaultProjectName(projectPath: string): string {
 	return path.basename(projectPath);
 }
 
-// Must produce the same name as the server's `deriveGitSecretName` (secretOperations.ts), so the
-// reference this attaches matches the row `harper deploy setup=true` sealed and a literal-token
-// deploy would use: `deploy.<component>.git.<host>`. Rather than restate the host-normalization
-// chain, this shares the server's own `normalizeGitHost` — a future host quirk added there then
-// can't drift from what the CLI sends. (gitCredentialServer.ts pulls in only node builtins plus
-// the error/logger utilities `bin/` already uses, so importing it costs the CLI nothing.)
-function deriveGitSecretName(component: string, host: string): string {
-	const hostPart = normalizeGitHost(host).replace(/[^\w.-]+/g, '_');
-	const componentPart = String(component).replace(/[^\w.-]+/g, '_');
-	return `deploy.${componentPart}.git.${hostPart}`;
-}
-
 // The credential only helps if it's for the host the package is cloned from: a `credential=gitlab.com`
 // against a github.com package builds a valid-looking reference the clone never asks for, and the
 // private deploy then fails as if nothing were configured. So the host comes from the package rather
@@ -563,7 +570,7 @@ const PREPARE_OPERATION: any = {
 		}
 
 		const projectPath = process.cwd();
-		if (!req.project) req.project = path.basename(projectPath);
+		if (!req.project) req.project = directoryProjectName(projectPath);
 		const packageOptions = {
 			skip_node_modules: req.skip_node_modules !== false,
 			skip_symlinks: req.skip_symlinks === true,
--- a/bin/deploySetup.ts
+++ b/bin/deploySetup.ts
@@ -38,6 +38,14 @@ import {
 	PROJECT_NAME_PATTERN,
 } from '../utility/componentNames.ts';
 
+// `formatCliError` (bin/harper.ts) prints message-only for an error carrying a numeric `statusCode`
+// and keeps the stack for anything else, on the assumption that a stack means a bug. Everything this
+// flow throws is a usage or cluster-capability problem the operator can act on, so it gets a code —
+// the value only selects the clean formatting.
+function cliError(message: string): Error {
+	return Object.assign(new Error(message), { statusCode: 400 });
+}
+
 function tryCommand(command: string, args: string[]): string {
 	try {
 		return execFileSync(command, args, { encoding: 'utf8', stdio: ['ignore', 'pipe', 'ignore'] }).trim();
@@ -79,38 +87,40 @@ export function resolveComponentName(req: any): string | undefined {
 export function resolveGitHost(host: unknown): string {
 	const normalized = typeof host === 'string' ? normalizeGitHost(host) : '';
 	if (!GIT_HOST_PATTERN.test(normalized)) {
-		throw new Error(`Invalid git host ${JSON.stringify(host)} — expected a bare host, e.g. 'github.com'.`);
+		throw cliError(`Invalid git host ${JSON.stringify(host)} — expected a bare host, e.g. 'github.com'.`);
 	}
 	return normalized;
 }
 
 /**
- * The grants to send with the seal: this component, plus whatever the row already grants.
+ * Store the sealed envelope and make sure the component can use it, as two server-side atomic steps.
+ * Returns the row's full grant list, as the server reports it.
  *
- * `set_secret` replaces rather than merges (`grants = dedupeGrants(req.grants ?? existing?.grants)`),
- * so rotating a token here while sending only this component would revoke any grant an operator added
- * separately with `grant_secret` — the realistic case being a monorepo whose second component deploys
- * from the same repo. That revocation is invisible until the other component's next deploy fails, so
- * the union is read first. A lookup failure falls back to this component alone (the pre-existing
- * behavior) and says so, rather than aborting a rotation the operator asked for.
+ * `grants` is deliberately omitted from `set_secret`: it *replaces* rather than merges
+ * (`grants = dedupeGrants(req.grants ?? existing?.grants)`, secretOperations.ts), so sending only this
+ * component would silently revoke a grant an operator had added with `grant_secret` — a monorepo's
+ * second component being the realistic case, and invisible until its next deploy fails. Omitting the
+ * field keeps the stored list untouched under the server's `withSecretLock`, and the idempotent
+ * `grant_secret` then adds this component in a second locked step. Doing the merge on the client
+ * instead would cost the same two round trips while reintroducing a race: a `revoke_secret` landing
+ * between the read and the write would be silently undone.
+ *
+ * A brand-new row is briefly ungranted between the calls, which is harmless — it can't be resolved by
+ * any deploy until the grant lands. Nothing is printed until both steps succeed, so a failure part-way
+ * doesn't claim a credential is ready to use.
  */
-export async function resolveGrants(transport: any, secretName: string, component: string): Promise<string[]> {
-	try {
-		const listed: any = await cliOperations({ ...transport, operation: 'list_secrets' }, true);
-		const row = listed?.secrets?.find((secret: any) => secret?.name === secretName);
-		const existing: string[] = Array.isArray(row?.grants) ? row.grants : [];
-		if (existing.length && !existing.includes(component)) {
-			console.log(chalk.gray(`  Keeping existing grant(s) on "${secretName}": ${existing.join(', ')}`));
-		}
-		return [...new Set([component, ...existing])];
-	} catch {
-		console.log(
-			chalk.yellow(
-				`Note: couldn't read existing grants for "${secretName}"; storing it granted to "${component}" only.`
-			)
-		);
-		return [component];
-	}
+export async function storeSealedSecret(
+	transport: any,
+	secretName: string,
+	envelope: string,
+	component: string
+): Promise<string[]> {
+	await cliOperations({ ...transport, operation: 'set_secret', name: secretName, envelope }, true);
+	const granted: any = await cliOperations(
+		{ ...transport, operation: 'grant_secret', name: secretName, component },
+		true
+	);
+	return Array.isArray(granted?.grants) ? granted.grants : [component];
 }
 
 export async function deploySetup(req: any): Promise<void> {
@@ -120,19 +130,14 @@ export async function deploySetup(req: any): Promise<void> {
 	// deploy args that brought us here (`setup`, `token`, `package`) never reach either body.
 	const transport = transportContext(req);
 
-	// 1. Fetch the cluster's public secrets key.
-	let keyResponse: any;
-	try {
-		keyResponse = await cliOperations({ ...transport, operation: 'get_secrets_public_key' }, true);
-	} catch (error) {
-		throw new Error(
-			`Could not fetch the cluster's secrets public key: ${error instanceof Error ? error.message : String(error)}`
-		);
-	}
+	// 1. Fetch the cluster's public secrets key. Not wrapped in a try/catch: `cliOperations` reports
+	// the failure and exits rather than throwing, so a catch here would be dead code promising a
+	// friendlier message that never runs.
+	const keyResponse: any = await cliOperations({ ...transport, operation: 'get_secrets_public_key' }, true);
 	const publicKey: string | undefined = keyResponse?.public_key;
 	const fingerprint: string | undefined = keyResponse?.fingerprint;
 	if (!publicKey || !fingerprint) {
-		throw new Error(
+		throw cliError(
 			"This cluster didn't return a secrets public key — secrets custody isn't initialized on it " +
 				'(available on Harper Pro / Fabric). Without it, a credential cannot be sealed client-side.'
 		);
@@ -154,7 +159,7 @@ export async function deploySetup(req: any): Promise<void> {
 		).provider;
 
 	if (provider !== 'github' && provider !== 'npm') {
-		throw new Error(`Unsupported provider "${provider}" — supported providers are "github" and "npm".`);
+		throw cliError(`Unsupported provider "${provider}" — supported providers are "github" and "npm".`);
 	}
 
 	// 3. Which component is the credential for? (the grant is scoped to it)
@@ -171,7 +176,7 @@ export async function deploySetup(req: any): Promise<void> {
 			).project ?? ''
 		);
 	if (!PROJECT_NAME_PATTERN.test(component)) {
-		throw new Error(
+		throw cliError(
 			`"${component}" is not a usable component name — a deploy accepts letters, numbers, dashes and underscores.`
 		);
 	}
@@ -245,7 +250,7 @@ export async function deploySetup(req: any): Promise<void> {
 	}
 
 	token = typeof token === 'string' ? token.trim() : undefined;
-	if (!token) throw new Error('No token was provided; nothing to store.');
+	if (!token) throw cliError('No token was provided; nothing to store.');
 
 	// 4. Seal the token locally. Only ciphertext leaves this machine.
 	const envelope = ENV_ENCRYPTED_PREFIX + encryptEnvelope(token, publicKey, fingerprint);
@@ -257,8 +262,7 @@ export async function deploySetup(req: any): Promise<void> {
 		provider === 'github'
 			? deriveGitSecretName(component, credentialKey)
 			: deriveRegistrySecretName(component, credentialKey);
-	const grants = await resolveGrants(transport, secretName, component);
-	await cliOperations({ ...transport, operation: 'set_secret', name: secretName, envelope, grants }, true);
+	const grants = await storeSealedSecret(transport, secretName, envelope, component);
 
 	// 6. Print the credentials reference the deploy should use.
 	credentialEntry.secret = secretName;
--- a/bin/deploySetup.ts
+++ b/bin/deploySetup.ts
@@ -13,8 +13,9 @@
 //   1. get_secrets_public_key  → the cluster's RSA public key + fingerprint
 //   2. source a token          → `gh auth token`, or paste a fine-grained PAT / npm token
 //   3. encryptEnvelope(...)     → seal it locally into an `enc:v1:` envelope (pure client-side crypto)
-//   4. set_secret {envelope}    → store ciphertext, granted to the component
-//   5. print the `credentials` reference the deploy should use
+//   4. set_secret {envelope}    → store ciphertext, in the component-scoped tier
+//   5. grant_secret            → let this component resolve it (see storeSealedSecret for why it's two calls)
+//   6. print the `credentials` reference the deploy should use
 //
 // Every name this flow derives — the component it grants to, the host it labels the credential with,
 // the hdb_secret row it writes — has to match what the deploy will derive from its own request, or it
@@ -105,6 +106,15 @@ export function resolveGitHost(host: unknown): string {
  * instead would cost the same two round trips while reintroducing a race: a `revoke_secret` landing
  * between the read and the write would be silently undone.
  *
+ * `processEnv: false` is explicit, and load-bearing rather than decorative. `set_secret` otherwise
+ * inherits the stored tier, so if the derived row already existed as a processEnv (global) secret the
+ * pasted token would be written into the tier every component and child process reads — and only then
+ * would `grant_secret` reject the row for being global, leaving the CLI reporting a failure it had
+ * already committed. (Sending `grants` used to mask this: the server rejects processEnv+grants before
+ * writing. Omitting them for the merge above removed that accident, so the intent is now stated.) This
+ * converts such a row to the scoped tier, which is what this flow promises — and the name is
+ * component-derived, so a global secret at it was never serving anything the scoped one doesn't.
+ *
  * A brand-new row is briefly ungranted between the calls, which is harmless — it can't be resolved by
  * any deploy until the grant lands. Nothing is printed until both steps succeed, so a failure part-way
  * doesn't claim a credential is ready to use.
@@ -115,7 +125,7 @@ export async function storeSealedSecret(
 	envelope: string,
 	component: string
 ): Promise<string[]> {
-	await cliOperations({ ...transport, operation: 'set_secret', name: secretName, envelope }, true);
+	await cliOperations({ ...transport, operation: 'set_secret', name: secretName, envelope, processEnv: false }, true);
 	const granted: any = await cliOperations(
 		{ ...transport, operation: 'grant_secret', name: secretName, component },
 		true
--- a/bin/deploySetup.ts
+++ b/bin/deploySetup.ts
@@ -1,0 +1,204 @@
+'use strict';
+
+// `harper deploy setup=true` — provision a durable, encrypted credential the cluster uses to fetch
+// private deploy sources (a private GitHub repo, or a private npm registry).
+//
+// The whole point is that the token is sealed *on this machine* with the cluster's public secrets
+// key and only the ciphertext is ever transmitted or stored — the cluster (and its logs, its
+// operations journal, its replication) never sees the plaintext, and decrypts it only in-memory at
+// deploy/rollback time. Because the token is durable (unlike a CI `GITHUB_TOKEN`, which dies at job
+// end), a re-resolve rollback works for as long as the token is valid.
+//
+// Flow:
+//   1. get_secrets_public_key  → the cluster's RSA public key + fingerprint
+//   2. source a token          → `gh auth token`, or paste a fine-grained PAT / npm token
+//   3. encryptEnvelope(...)     → seal it locally into an `enc:v1:` envelope (pure client-side crypto)
+//   4. set_secret {envelope}    → store ciphertext, granted to the component
+//   5. print the `credentials` reference the deploy should use
+
+import chalk from 'chalk';
+import inquirer from 'inquirer';
+import { execFileSync } from 'node:child_process';
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import { cliOperations } from './cliOperations.ts';
+import { encryptEnvelope } from '../utility/secretEnvelope.ts';
+import { ENV_ENCRYPTED_PREFIX } from '../utility/envFile.ts';
+
+// Mirror the server's derived secret names (secretOperations.ts `deriveRegistrySecretName` and
+// `deriveGitSecretName`) EXACTLY, so a `deploy setup` seal and a later literal-token deploy overwrite
+// the SAME hdb_secret row. git hosts get a `.git.` segment (and normalizeGitHost handling); registries
+// don't — without the segment, a git and a registry credential for the same host would collide.
+function deriveSecretName(component: string, key: string, provider: string): string {
+	const componentPart = String(component).replace(/[^\w.-]+/g, '_');
+	if (provider === 'github') {
+		// mirrors gitCredentialServer.normalizeGitHost() + deriveGitSecretName's `.git.` segment
+		const hostPart = key
+			.trim()
+			.replace(/^[a-z0-9+.-]+:\/\//i, '')
+			.replace(/^\/\//, '')
+			.replace(/\/.*$/, '')
+			.replace(/^[^@]*@/, '')
+			.toLowerCase()
+			.replace(/[^\w.-]+/g, '_');
+		return `deploy.${componentPart}.git.${hostPart}`;
+	}
+	const registryPart = key
+		.trim()
+		.replace(/^https?:\/\//i, '')
+		.replace(/^\/\//, '')
+		.replace(/\/+$/, '')
+		.toLowerCase()
+		.replace(/[^\w.-]+/g, '_');
+	return `deploy.${componentPart}.${registryPart}`;
+}
+
+function tryCommand(command: string, args: string[]): string {
+	try {
+		return execFileSync(command, args, { encoding: 'utf8', stdio: ['ignore', 'pipe', 'ignore'] }).trim();
+	} catch {
+		return '';
+	}
+}
+
+function defaultComponentName(): string {
+	try {
+		const pkg = JSON.parse(fs.readFileSync(path.join(process.cwd(), 'package.json'), 'utf8'));
+		if (typeof pkg.name === 'string' && pkg.name.length > 0) return pkg.name.replace(/^@[^/]+\//, '');
+	} catch {
+		// no package.json — fall back to the directory name
+	}
+	return path.basename(process.cwd());
+}
+
+export async function deploySetup(req: any): Promise<void> {
+	// 1. Fetch the cluster's public secrets key. Target + auth are resolved by cliOperations exactly
+	// as they are for a deploy (harper login, or CLI_TARGET + HARPER_CLI_USERNAME/PASSWORD).
+	let keyResponse: any;
+	try {
+		keyResponse = await cliOperations({ operation: 'get_secrets_public_key', target: req.target }, true);
+	} catch (error) {
+		throw new Error(
+			`Could not fetch the cluster's secrets public key: ${error instanceof Error ? error.message : String(error)}`
+		);
+	}
+	const publicKey: string | undefined = keyResponse?.public_key;
+	const fingerprint: string | undefined = keyResponse?.fingerprint;
+	if (!publicKey || !fingerprint) {
+		throw new Error(
+			"This cluster didn't return a secrets public key — secrets custody isn't initialized on it " +
+				'(available on Harper Pro / Fabric). Without it, a credential cannot be sealed client-side.'
+		);
+	}
+
+	// 2. Which private source?
+	const provider: string =
+		req.provider ??
+		(
+			await inquirer.prompt({
+				type: 'list',
+				name: 'provider',
+				message: 'What private source needs a credential?',
+				choices: [
+					{ name: 'GitHub repository (private git clone)', value: 'github' },
+					{ name: 'npm registry (private packages / dependencies)', value: 'npm' },
+				],
+			})
+		).provider;
+
+	if (provider !== 'github' && provider !== 'npm') {
+		throw new Error(`Unsupported provider "${provider}" — supported providers are "github" and "npm".`);
+	}
+
+	// 3. Which component is the credential for? (the grant is scoped to it)
+	const component: string =
+		req.project ??
+		req.component ??
+		(
+			await inquirer.prompt({
+				type: 'input',
+				name: 'component',
+				message: 'Component (project) name this credential is for:',
+				default: defaultComponentName(),
+			})
+		).component;
+
+	let credentialKey: string; // host (github) or registry (npm) — the credentials-entry discriminator
+	let credentialEntry: Record<string, string>;
+	let token: string | undefined = req.token;
+
+	if (provider === 'github') {
+		const host = req.host ?? 'github.com';
+		credentialKey = host;
+		if (!token) {
+			const ghToken = tryCommand('gh', ['auth', 'token']);
+			const choices: Array<{ name: string; value: string }> = [];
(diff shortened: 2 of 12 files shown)


bin/deploySetup.ts
```diff
@@ -0,0 +1,186 @@
...
+	const keyPart = key
+		.trim()
+		.replace(/^https?:\/\//i, '')
+		.replace(/^\/\//, '')
+		.replace(/\/+$/, '')
+		.toLowerCase()
+		.replace(/[^\w.-]+/g, '_');
+	const componentPart = String(component).replace(/[^\w.-]+/g, '_');
+	return `deploy.${componentPart}.${keyPart}`;
+}
```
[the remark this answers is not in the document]
dawsontoth: Fixed — `deriveSecretName` now mirrors the server's `deriveGitSecretName` exactly (`deploy.<component>.git.<host>`, host via `normalizeGitHost`) for git, and `deriveRegistrySecretName` for npm, so a `deploy setup` seal and a literal-token deploy hit the same `hdb_secret` row. (Strictly, the pure reference path resolves by the explicit `secret` name — `resolveCredentials` doesn't re-derive — so it wasn't a hard resolve failure, but aligning the convention is the right call.) The sibling `by_ref` flag (#1850) was aligned to match. (7e1890785)

bin/deploySetup.ts
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	token = token?.trim();
+	if (!token) throw new Error('No token was provided; nothing to store.');
+
+	// 4. Seal the token locally. Only ciphertext leaves this machine.
+	const envelope = ENV_ENCRYPTED_PREFIX + encryptEnvelope(token, publicKey, fingerprint);
+
+	// 5. Store the sealed token, granted to the component. The server never sees the plaintext.
+	const secretName = deriveSecretName(component, credentialKey);
```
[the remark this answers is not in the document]
dawsontoth: Done — the call now passes `provider`, so git hosts get the `.git.` name and registries don't. (7e1890785)

bin/deploySetup.ts
```diff
@@ -0,0 +1,186 @@
...
+			await inquirer.prompt({
+				type: 'list',
+				name: 'provider',
+				message: 'What private source needs a credential?',
+				choices: [
+					{ name: 'GitHub repository (private git clone)', value: 'github' },
+					{ name: 'npm registry (private packages / dependencies)', value: 'npm' },
+				],
+			})
+		).provider;
```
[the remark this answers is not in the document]
dawsontoth: Added — the provider is validated as `github`|`npm` right after the prompt, throwing on anything else instead of silently falling through to the npm flow. (7e1890785)

bin/deploySetup.ts
```diff
@@ -0,0 +1,186 @@
...
+		credentialKey = registry;
+		if (!token) {
+			console.log(chalk.gray('Tip: `npm token create --read-only` mints a granular npm token from the CLI.'));
+			token = (await inquirer.prompt({ type: 'password', name: 'token', message: 'Paste the npm token:', mask: '*' }))
+				.token;
+		}
+		credentialEntry = req.scope ? { registry, scope: req.scope } : { registry };
+	}
+
+	token = token?.trim();
```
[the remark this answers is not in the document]
dawsontoth: Fixed — `token = typeof token === 'string' ? token.trim() : undefined`, so a non-string `token=` arg can't throw. (7e1890785)

bin/harper.ts
```diff
@@ -153,6 +153,13 @@ async function harper() {
 			return require('./run').main();
 		default:
 			const cliApiOp = cliOperations.buildRequest();
+			// `harper deploy setup=true` provisions an encrypted deploy credential (client-side sealed
+			// token) rather than deploying — an interactive flow, not a single operation call.
+			if (cliApiOp.operation === 'deploy_component' && cliApiOp.setup) {
+				const { deploySetup } = require('./deploySetup');
+				await deploySetup(cliApiOp);
+				return;
+			}
```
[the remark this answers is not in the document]
dawsontoth: Leaving this as-is: `buildRequest` JSON-parses each `key=value` arg, so `setup=false` becomes boolean `false` and `setup=0` becomes `0` (both falsy) before this check runs — the stringy `'false'`/`'0'` case can't reach here. The truthiness check is already correct.

bin/deploySetup.ts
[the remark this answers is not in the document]
dawsontoth: Fixed in 7e1890785 — `deriveSecretName` is now provider-aware and mirrors the server's `deriveGitSecretName` (`deploy.<component>.git.<host>`, host via `normalizeGitHost`) for git and `deriveRegistrySecretName` for npm, so it lands on the same `hdb_secret` row as `ingestCredentials`'s literal-token path (and the git/registry collision the `.git.` segment prevents is restored). I mirrored the exact derivations locally rather than importing from `components/secretOperations.ts` — that module pulls server-side deps (databases/config) I don't want to load into the CLI — but the output is byte-identical for the hosts/registries in play (verified). The sibling `by_ref` flag (#1850) was aligned to the same convention.

bin/deploySetup.ts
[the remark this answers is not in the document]
dawsontoth: Fixed in 7e1890785 — `token = typeof token === 'string' ? token.trim() : undefined`, so a `token=12345`/`token=true` arg (coerced by `buildRequest`'s JSON.parse) hits the "no token" path instead of throwing.

bin/deploySetup.ts
```diff
@@ -0,0 +1,186 @@
...
+				message: 'Component (project) name this credential is for:',
+				default: defaultComponentName(),
+			})
+		).component;
+
+	let credentialKey: string; // host (github) or registry (npm) — the credentials-entry discriminator
+	let credentialEntry: Record<string, string>;
+	let token: string | undefined = req.token;
+
+	if (provider === 'github') {
```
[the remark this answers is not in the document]
dawsontoth: Added in 7e1890785 — `provider` is validated as `github`|`npm` right after resolution and throws a clear error otherwise, instead of silently falling through to the npm branch.

bin/deploySetup.ts
```diff
@@ -0,0 +1,204 @@
...
+	}
+	const registryPart = key
+		.trim()
+		.replace(/^https?:\/\//i, '')
+		.replace(/^\/\//, '')
+		.replace(/\/+$/, '')
+		.toLowerCase()
+		.replace(/[^\w.-]+/g, '_');
+	return `deploy.${componentPart}.${registryPart}`;
+}
```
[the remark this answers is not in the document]
dawsontoth: Added — exported `deriveSecretName` and added `unitTests/bin/deploySetup.test.js` locking the git (`deploy.<component>.git.<host>`) and npm (`deploy.<component>.<registry>`) names to the server's `deriveGitSecretName`/`deriveRegistrySecretName` convention (4 cases incl. host normalization + component sanitization, passing). So a future drift on either side fails here. (1fe39b1f5)

bin/deploySetup.ts
```diff
@@ -0,0 +1,204 @@
...
+	let credentialEntry: Record<string, string>;
+	let token: string | undefined = req.token;
+
+	if (provider === 'github') {
+		const host = req.host ?? 'github.com';
+		credentialKey = host;
+		if (!token) {
+			const ghToken = tryCommand('gh', ['auth', 'token']);
+			const choices: Array<{ name: string; value: string }> = [];
+			if (ghToken) choices.push({ name: 'Use your gh CLI session token (gh auth token)', value: 'gh' });
```
cb1kenobi: **Low: `gh auth token` seals a broad, durable credential**

The `gh` CLI OAuth token is offered as the default source when available, but it typically carries broad account scopes (`repo`, `read:org`, `gist`, `workflow`) and — unlike a CI `GITHUB_TOKEN` that dies at job end — is durable. It therefore lives in `hdb_secret` long-term and is replayed on every cold deploy/rollback, which somewhat undercuts the fine-grained "Contents: Read-only" guidance offered on the very next choice.

Consider making the fine-grained PAT the primary/recommended path (or emitting a least-privilege warning when the gh session token is chosen), so the stored durable credential follows least privilege. Not a blocker given Harper's admin-trusted model — a hardening note.

—
_Generated by Barber AI_

dawsontoth: Good catch — taken in `77badbb`. You're right that it undercut the guidance sitting one line below it, and the ordering was the real problem: `gh` was listed first, so it was what Enter selected. That put least privilege on the path of most resistance.

Now:

- The fine-grained PAT is offered **first** (so it's the default selection) and labeled `— recommended`.
- The `gh` option is relabeled `(broad account scopes)` and prints a warning when chosen, naming the scopes and noting the token is stored and reused for future deploys and rollbacks.

I kept it as an option rather than removing it — it's a legitimate quick path for a private repo you own, and the durability point applies to any PAT too. The warning makes the tradeoff visible at the moment of choosing, which seemed better than deciding for the user.

bin/deploySetup.ts
```diff
@@ -0,0 +1,220 @@
...
+				default: defaultComponentName(),
+			})
+		).component;
+
+	let credentialKey: string; // host (github) or registry (npm) — the credentials-entry discriminator
+	let credentialEntry: Record<string, string>;
+	let token: string | undefined = req.token;
+
+	if (provider === 'github') {
+		const host = req.host ?? 'github.com';
```
kriszyp, line 131: Please canonicalize the selected host once and bind `gh` token lookup to that exact host. With `host=ghe.example.com`, `gh auth token` currently reads the default host's token; the flow then labels and stores that token for `ghe.example.com`, so a later deploy can send a GitHub.com credential to the enterprise host. `gh auth token --hostname <normalized-host>` should be used, and the `gh` choice should only be offered when lookup for that host succeeds. The normalized bare host should also go into `credentialEntry`: setup currently accepts/tests scheme, path, and `git@` forms, but the deploy schema rejects those forms even though sealing succeeded.

bin/deploySetup.ts
```diff
@@ -0,0 +1,220 @@
...
+				],
+			})
+		).provider;
+
+	if (provider !== 'github' && provider !== 'npm') {
+		throw new Error(`Unsupported provider "${provider}" — supported providers are "github" and "npm".`);
+	}
+
+	// 3. Which component is the credential for? (the grant is scoped to it)
+	const component: string =
```
kriszyp, line 114: Please resolve the component with the same logic as `deploy_component` before deriving the secret name or grant. Setup defaults from `package.json`, while a directory deploy defaults to `path.basename(process.cwd())`; it also uses an explicit project verbatim, while the server canonicalizes it with `path.parse(req.project).name`. For example, setup in directory `checkout` with package name `@org/web` grants `web`, but a later bare `harper deploy` deploys `checkout`; `project=@scope/app` similarly grants `@scope/app` while the server deploys `app`. In both cases `resolveCredentials` rejects the stored secret as not granted. Extract/reuse a shared project resolver and cover these cases in the test.

bin/deploySetup.ts
```diff
@@ -0,0 +1,220 @@
...
+	}
+	return path.basename(process.cwd());
+}
+
+export async function deploySetup(req: any): Promise<void> {
+	// 1. Fetch the cluster's public secrets key. Target + auth are resolved by cliOperations exactly
+	// as they are for a deploy (harper login, or CLI_TARGET + HARPER_CLI_USERNAME/PASSWORD).
+	let keyResponse: any;
+	try {
+		keyResponse = await cliOperations({ operation: 'get_secrets_public_key', target: req.target }, true);
```
kriszyp, line 79: Please preserve the caller's transport context for both operations. Rebuilding each request with only `target` drops `auth_username`, `auth_password`, and `rejectUnauthorized`; explicit credentials can therefore be ignored in favor of a saved login token (performing/auditing the mutation as the wrong identity), and self-signed targets that require `rejectUnauthorized=false` fail. A small shared helper that projects the transport-only fields from `req` into both `get_secrets_public_key` and `set_secret` requests would avoid sending setup fields such as `token` while retaining normal CLI behavior.

bin/deploySetup.ts
```diff
@@ -0,0 +1,220 @@
...
+import * as path from 'node:path';
+import { cliOperations } from './cliOperations.ts';
+import { encryptEnvelope } from '../utility/secretEnvelope.ts';
+import { ENV_ENCRYPTED_PREFIX } from '../utility/envFile.ts';
+
+// Mirror the server's derived secret names (secretOperations.ts `deriveRegistrySecretName` and
+// `deriveGitSecretName`) EXACTLY, so a `deploy setup` seal and a later literal-token deploy overwrite
+// the SAME hdb_secret row. git hosts get a `.git.` segment (and normalizeGitHost handling); registries
+// don't — without the segment, a git and a registry credential for the same host would collide.
+export function deriveSecretName(component: string, key: string, provider: string): string {
```
kriszyp, line 32: Could the pure host/registry normalization and derived-name functions be moved to a low-dependency shared module used by both this client and `components/secretOperations.ts`? The comment says this must mirror the server exactly, but the new test only asserts this duplicate against hard-coded strings; a future server change leaves it green while setup writes a different `hdb_secret` row from literal-token deploys. Sharing the implementation removes that contract and also gives the component/host fixes above one canonical path.

kriszyp (commented):
This is going to be great. Some comments to address.

bin/deploySetup.ts
dawsontoth: Done in 49c2b4739 — the duplication is gone rather than tested against, which was the right read.

New `utility/componentNames.ts` is now the single definition of every name the client and the deploy have to agree on, and it is dependency-free (node builtins only), so the CLI can import it without pulling in `databases`/config:

- `normalizeGitHost`, `deriveGitSecretName`, `deriveRegistrySecretName`
- `canonicalProjectName`, `projectNameFromPackage`, `directoryProjectName`
- `PROJECT_NAME_PATTERN`, `GIT_HOST_PATTERN`

Consumers: `components/gitCredentialServer.ts` and `components/secretOperations.ts` re-export their previous names from it (so existing importers and their tests are unchanged), `components/operations.js` uses the project resolvers, `components/operationsValidation.js` uses both patterns, and `bin/deploySetup.ts` + `bin/cliOperations.ts` use them directly. Nothing is mirrored on the client anymore, so there is no contract left to drift — a change to any of these changes both sides at once.

That also gave the component and host fixes below one canonical path, as you suggested.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

bin/deploySetup.ts
dawsontoth: Fixed in 49c2b4739. Setup now resolves the component through the same helpers `deploy_component` does (`utility/componentNames.ts`):

- explicit `project` → `canonicalProjectName` (`path.parse().name`), so `project=@scope/app` grants `app` — the name the server deploys, not the verbatim string
- no `project` but a `package` → `projectNameFromPackage`, the function `deployComponent` itself now calls (moved out of `components/operations.js`, unchanged)
- neither → prompt defaulting to `directoryProjectName()`, the same helper `cliOperations`'s directory deploy now uses for its own default. Your `checkout` + `@org/web` case therefore resolves `checkout` on both sides; the package.json default is gone.

Two follow-ons from this: the resolved name is checked against the shared `PROJECT_NAME_PATTERN`, so a name the deploy validator would reject fails during setup instead of after a credential has been sealed; and the `component=` alias is gone, since it had no counterpart on the deploy and was a third way to disagree.

Covered in `unitTests/bin/deploySetup.test.js` — the `@scope/app` case, the package-spec case, the prompt-default case (asserted equal to `path.basename(process.cwd())`), and a non-string `project=5` falling through to the prompt rather than throwing inside `path.parse`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

bin/deploySetup.ts
dawsontoth: Fixed in 49c2b4739 — all four parts.

`resolveGitHost()` canonicalizes the host once, via the shared `normalizeGitHost`, and that one value drives everything the host decides:

- `gh auth token --hostname <host>`, so the lookup is bound to the host being provisioned instead of whichever one gh considers default
- the `gh` choice is only pushed onto the prompt when that host-scoped lookup succeeds, and its label now names the host (`Use your gh CLI session token for ghe.example.com …`), so it cannot fall back to another host's credential
- the derived secret name
- the `host` on the printed `credentials` entry — the normalized bare form, so a scheme/path/`git@` input no longer seals successfully and then gets rejected by the deploy schema

The PAT instructions URL is also built from the host now, rather than hardcoding github.com.

Anything still outside the schema's host grammar after normalization (an embedded space, a second `@`) throws during setup — `GIT_HOST_PATTERN` is shared with `GIT_CREDENTIAL_ENTRY`, so setup rejects exactly what the deploy would. The old test cases that locked in the scheme/`git@` forms now assert canonicalization instead (`unitTests/bin/deploySetup.test.js`).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

bin/deploySetup.ts
dawsontoth: Fixed in 49c2b4739 with the shared helper you described — `transportContext(req)`, exported from `bin/cliOperations.ts`.

It projects only the connection fields (`target`, `auth_username`, `auth_password`, `rejectUnauthorized`) and both `get_secrets_public_key` and `set_secret` are built from it, so the sub-operations keep the caller's identity and TLS strictness — explicit credentials are no longer silently displaced by a saved login token (which would also have audited the mutation as the wrong identity), and a self-signed target's `rejectUnauthorized=false` survives. Setup's own args (`setup`, `token`, `package`, …) are not in the projection, so they never reach either body.

Those four names are now a `CONNECTION_FIELDS` array that `TRANSPORT_ONLY_FIELDS` is built from, so "never in the body" and "carried to a sub-operation" stay in agreement instead of being two lists. Covered in `unitTests/bin/cliOperations.test.js`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

kriszyp (approved):
Awesome!
And i think this is destined for 5.2 (patch, feels safe), if you are comfortable with that.

dawsontoth:
Sure!

bin/deploySetup.ts
```diff
@@ -0,0 +1,250 @@
...
+	// 3. Which component is the credential for? (the grant is scoped to it)
+	const component =
+		resolveComponentName(req) ??
+		canonicalProjectName(
+			(
+				await inquirer.prompt({
+					type: 'input',
+					name: 'project',
+					message: 'Component (project) name this credential is for:',
+					default: directoryProjectName(),
```
cb1kenobi, line 184: **Medium: setup's default component name doesn't match the one `by_ref` deploys as**

This PR moved the payload deploy's default to the shared `directoryProjectName()` (cwd basename), and setup prompts with the same. But `deploy by_ref=true` — the flow this credential exists to feed — still defaults through `defaultProjectName()` (bin/cliOperations.ts:505), which prefers `package.json` `name` with the scope stripped and only falls back to the directory. So the second derivation the refactor set out to remove is still there, on the path that consumes the seal.

In a checkout `my-app-repo/` whose package.json is `@acme/my-app`:

```
harper deploy setup=true            → seals deploy.my-app-repo.git.github.com, granted to "my-app-repo"
harper deploy by_ref=true credential=true → asks for deploy.my-app.git.github.com as "my-app"
```

`resolveCredentials` then throws `credential entry references secret 'deploy.my-app.git.github.com', which does not exist` — naming a secret the user never chose. It fails loudly rather than silently, but only after a round trip, and the message points nowhere useful.

Suggested fix: have setup default through the same function `prepareDeployByRef` uses (move `defaultProjectName` into `utility/componentNames.ts` alongside `directoryProjectName` and call it here), or make `prepareDeployByRef` use `directoryProjectName`. Either way one derivation, per this commit's own thesis. Worth a unit test asserting `deploySetup`'s default and `prepareDeployByRef`'s `req.project` agree for a cwd whose package name differs from its directory — `unitTests/bin/deploySetup.test.js:32` currently asserts the directory-name half only.

—
_Generated by Barber AI_

bin/deploySetup.ts
```diff
@@ -0,0 +1,250 @@
...
+		);
+	if (!PROJECT_NAME_PATTERN.test(component)) {
+		throw new Error(
+			`"${component}" is not a usable component name — a deploy accepts letters, numbers, dashes and underscores.`
+		);
+	}
+
+	let credentialKey: string; // host (github) or registry (npm) — the credentials-entry discriminator
+	let credentialEntry: Record<string, string>;
+	let token: string | undefined = req.token;
```
cb1kenobi, line 196: **Medium: the new `token=` arg isn't registered as secret-bearing, so a near-miss invocation logs the PAT in plaintext**

This introduces `token=` as a CLI arg carrying a durable PAT, but `bin/cliOperations.ts` isn't told about it — its own comment above `SECRET_FIELDS` (line 254) says exactly this: *"any future secret-bearing arg (a token, a key) needs to be added here explicitly, or it will reach logger.trace unredacted."*

The intercept in `bin/harper.ts` only fires on truthy `cliApiOp.setup`. `setup` without `=true` parses as a bare word that `buildRequest` drops (`operation` is already set by the `deploy` alias), so:

```
harper deploy setup token=github_pat_...
```

falls through to the normal path. `token` is in neither `SECRET_FIELDS` nor `TRANSPORT_ONLY_FIELDS`, so it (a) reaches `logger.trace('calling cli operations with:', redactCredentials(cliApiOp))` unredacted and (b) is serialized into the `deploy_component` body. `deployComponentValidator` allows unknown keys (see the `registryAuth` note at operationsValidation.js:517), so it reaches the server, and `processLocalTransaction` strips `password`/`value`/`values`/`envelope`/`credentials` but **not** `token` (server/serverHelpers/serverUtilities.ts:98-111) — the PAT lands in the operations log at `logging.level: info` or lower. The rest of this flow is careful about exactly this (the seal never leaves the client, `skipResponseLog` on both sub-operations), so the gap is the arg registration, not the design.

Suggested fix: add `'token'` to `SECRET_FIELDS` and to `TRANSPORT_ONLY_FIELDS` in `bin/cliOperations.ts` — it's read off the parsed request before either set applies, so `deploySetup` still sees it, and no operation takes a top-level `token` today (`credentials[].token` is nested and already stripped). Adding `token` (and `key`) to the server-side strip list is the belt-and-braces half.

—
_Generated by Barber AI_

bin/deploySetup.ts
```diff
@@ -0,0 +1,250 @@
...
+	const envelope = ENV_ENCRYPTED_PREFIX + encryptEnvelope(token, publicKey, fingerprint);
+
+	// 5. Store the sealed token, granted to the component. The server never sees the plaintext. The
+	// derived name is the one the server's literal-token path would use for the same component and
+	// host/registry, so re-running this rotates the same row rather than piling up a second one.
+	const secretName =
+		provider === 'github'
+			? deriveGitSecretName(component, credentialKey)
+			: deriveRegistrySecretName(component, credentialKey);
+	await cliOperations({ ...transport, operation: 'set_secret', name: secretName, envelope, grants: [component] }, true);
```
cb1kenobi: **Low: re-running setup to rotate a token silently drops any extra grants on the row**

`set_secret` replaces rather than merges: `grants = dedupeGrants(req.grants ?? existing?.grants)` (components/secretOperations.ts:194). Because this always sends `grants: [component]`, an operator who additionally ran `grant_secret` on the same row (a monorepo deploying two components from one repo is the realistic case) loses that second grant the next time they rotate the token here — and the success message below reports only the seal, so the revocation is invisible until the other component's next deploy 403s.

Suggested fix: read the existing row's grants and send the union, or drop `grants` on a rotation when the row already grants `component` (`set_secret` keeps `existing.grants` when `grants` is omitted). If replace is the intended semantic, say so in the printed summary.

—
_Generated by Barber AI_

bin/deploySetup.ts
dawsontoth: Fixed in cb52c4153 — correct on every step, including that the gap was arg registration rather than design. `token` is now in both `SECRET_FIELDS` (so `redactCredentials` masks it in the trace log) and `TRANSPORT_ONLY_FIELDS` (so `operationFields` never serializes it), and added to the server-side strip list in `serverUtilities.ts` for the belt-and-braces half. I left `key` alone — it is a name rather than a secret in `set_env_value`, and its `value` is already stripped.

I also closed the near-miss at its source: `harper deploy` carrying a top-level `token` now fails with a pointer to `setup=true`. Without that, the mistyped invocation would have proceeded as an ordinary deploy and silently discarded the token, leaving the user believing a credential had been provisioned.

Two tests lock the property (`unitTests/bin/cliOperations.test.js`): one asserts `redactCredentials` returns `token: "***"` while leaving non-secret fields readable, the other captures the request body and asserts `token` is absent while real operation fields still arrive. Verified no operation consumes a top-level `req.token` before excluding it.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

bin/deploySetup.ts
dawsontoth: Fixed in cb52c4153, taking the second of your two options: `prepareDeployByRef` now uses the shared `directoryProjectName()` and `defaultProjectName` is deleted. One derivation across all three paths.

I picked that direction rather than moving `defaultProjectName` into the shared module because the two paths themselves disagreed, so no choice for setup alone could satisfy both: a git credential is consumed by `by_ref`/package deploys, but an **npm-registry** credential is consumed by a payload deploy too, and that path defaults to the cwd basename on both client and server. Only the directory default is common to every path, so aligning `by_ref` to it makes all three agree instead of leaving a 2-vs-1 split.

Worth flagging explicitly for the human reviewers, since it changes #1850: a by-reference deploy in a checkout whose directory and package.json name differ now deploys under the directory name. That seemed right while the by-reference flag is still unreleased and the effort is in flight, but it is a behavior change to merged code and I would rather it be seen than buried.

Test added at `unitTests/bin/deploySetup.test.js` per your suggestion — builds a `my-app-repo/` whose package.json is `@acme/my-app`, asserts `prepareDeployByRef` resolves `my-app-repo` (not `my-app`), and that setup agrees.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

bin/deploySetup.ts
dawsontoth: Fixed in cb52c4153 with the union option. A new `resolveGrants()` reads the row via `list_secrets` and sends `[component, ...existing]` deduped, so a rotation preserves a grant an operator added separately with `grant_secret`. Replace was not the intended semantic — silently revoking the monorepo second component was a bug, not a policy.

Also addressed the visibility half you raised: the printed summary now names every component the row is granted to, not just this one, and it notes when it is preserving pre-existing grants. A `list_secrets` failure falls back to `[component]` (today’s behavior) with a warning rather than aborting a rotation the operator asked for.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

bin/deploySetup.ts
```diff
@@ -0,0 +1,280 @@
...
+ * `set_secret` replaces rather than merges (`grants = dedupeGrants(req.grants ?? existing?.grants)`),
+ * so rotating a token here while sending only this component would revoke any grant an operator added
+ * separately with `grant_secret` — the realistic case being a monorepo whose second component deploys
+ * from the same repo. That revocation is invisible until the other component's next deploy fails, so
+ * the union is read first. A lookup failure falls back to this component alone (the pre-existing
+ * behavior) and says so, rather than aborting a rotation the operator asked for.
+ */
+async function resolveGrants(transport: any, secretName: string, component: string): Promise<string[]> {
+	try {
+		const listed: any = await cliOperations({ ...transport, operation: 'list_secrets' }, true);
```
kriszyp: Please make grant preservation a single server-side locked mutation. The current `list_secrets` → `set_secret` sequence is a lost-update race: if two setup rotations, or a setup and `grant_secret`, read the same grants before either write completes, the later `set_secret` writes its stale union and silently revokes the grant added by the first. A successful response lacking the expected `secrets` array also becomes `[component]` and removes every prior grant. The server’s `withSecretLock` protects only the eventual set, not this client-side read. Please merge/ensure the component grant while rotating the envelope under that same per-secret lock, and cover overlapping rotations/grant mutations.

— KrAIs (Codex)

cb1kenobi:
Reviewed `0cb1fbc` — no issues found. This PR looks good, nice job!

—
_Generated by Barber AI_

bin/cliOperations.ts
```diff
@@ -53,8 +55,28 @@ const TRANSPORT_ONLY_FIELDS = new Set([
 	'by_ref',
 	'ref',
 	'credential',
+	// `deploy setup=true`'s token, read off the parsed request by deploySetup and sealed locally. No
+	// operation takes a top-level `token` (the one in `credentials[]` is nested and stripped with it),
```
cb1kenobi: **Nit: this parenthetical contradicts line 54 — `credentials[].token` is *not* stripped from the body**

Line 54 says `credentials` "IS a real operation field and is sent", and it's correctly absent from this set: the nested `credentials[].token` reaches the server by design, where `ingestCredentials` seals it into an `hdb_secret` row. But this parenthetical sits inside a comment justifying a *body*-strip entry, so "is nested and stripped with it" reads as though the nested token is dropped from the body too.

The only place `credentials` is stripped is the operations-log destructure in `server/serverHelpers/serverUtilities.ts:104` — a different mechanism with a different purpose. As written this could lead a future maintainer to conclude `credentials[].token` never reaches the server and "fix" the path that depends on it.

Suggested fix:
```suggestion
	// operation takes a top-level `token` (the `credentials[]` one is nested inside a field that IS
	// sent, and is stripped only from the operations log),
```

—
_Generated by Barber AI_

bin/deploySetup.ts
```diff
@@ -0,0 +1,280 @@
...
+ * The grants to send with the seal: this component, plus whatever the row already grants.
+ *
+ * `set_secret` replaces rather than merges (`grants = dedupeGrants(req.grants ?? existing?.grants)`),
+ * so rotating a token here while sending only this component would revoke any grant an operator added
+ * separately with `grant_secret` — the realistic case being a monorepo whose second component deploys
+ * from the same repo. That revocation is invisible until the other component's next deploy fails, so
+ * the union is read first. A lookup failure falls back to this component alone (the pre-existing
+ * behavior) and says so, rather than aborting a rotation the operator asked for.
+ */
+async function resolveGrants(transport: any, secretName: string, component: string): Promise<string[]> {
```
cb1kenobi: **Nit: `resolveGrants` — the grant-union fix — has no test coverage**

The other two fixes in these commits each landed with a regression test (`unitTests/bin/deploySetup.test.js` for the project-name default, `unitTests/bin/cliOperations.test.js` for `token=`), but this one has none — `unitTests/bin/deploySetup.test.js` exercises only `resolveComponentName` and `resolveGitHost`, and `resolveGrants` isn't exported.

Both branches are worth pinning, because a silent regression here re-introduces exactly the invisible grant revocation this fixes: the union path (row already grants `["other-app"]` → both retained) and the `list_secrets`-failure fallback (returns `[component]`, which still replaces — the case the yellow warning exists to make visible).

Suggested fix: export `resolveGrants` alongside `resolveComponentName`/`resolveGitHost` and add two cases with a stubbed `cliOperations` — one asserting the union, one asserting the fallback plus its warning.

—
_Generated by Barber AI_

bin/deploySetup.ts
```diff
@@ -0,0 +1,280 @@
...
+	try {
+		const listed: any = await cliOperations({ ...transport, operation: 'list_secrets' }, true);
+		const row = listed?.secrets?.find((secret: any) => secret?.name === secretName);
+		const existing: string[] = Array.isArray(row?.grants) ? row.grants : [];
+		if (existing.length && !existing.includes(component)) {
+			console.log(chalk.gray(`  Keeping existing grant(s) on "${secretName}": ${existing.join(', ')}`));
+		}
+		return [...new Set([component, ...existing])];
+	} catch {
```
cb1kenobi: **Medium: this fallback is unreachable — `cliOperations` exits the process instead of throwing**

`cliOperations` never throws: the non-2xx branch (`bin/cliOperations.ts:989-993`) and the catch-all (`:1004-1023`) both end in `process.exit(1)`. So a failed `list_secrets` — transport error, `system.hdb_secret` unavailable, any non-2xx — terminates the CLI outright, and neither this `catch` nor the yellow "couldn't read existing grants" notice ever runs.

The docblock at line 94 promises the opposite: *"A lookup failure falls back to this component alone (the pre-existing behavior) and says so, rather than aborting a rotation the operator asked for."* As written, a lookup failure aborts setup **after** the PAT has already been pasted, and the operator gets a bare `error:` line instead of the intended degradation. This is a new fatal round trip on the path these commits added.

Suggested fix: don't rely on `cliOperations` throwing. The cleanest route is to drop the read entirely (see the note on line 261) — the server already preserves grants when `grants` is omitted. If the read is kept, it needs a call path that actually throws.

—
_Generated by Barber AI_

bin/deploySetup.ts
```diff
@@ -0,0 +1,280 @@
...
+
+	// 5. Store the sealed token, granted to the component. The server never sees the plaintext. The
+	// derived name is the one the server's literal-token path would use for the same component and
+	// host/registry, so re-running this rotates the same row rather than piling up a second one.
+	const secretName =
+		provider === 'github'
+			? deriveGitSecretName(component, credentialKey)
+			: deriveRegistrySecretName(component, credentialKey);
+	const grants = await resolveGrants(transport, secretName, component);
+	await cliOperations({ ...transport, operation: 'set_secret', name: secretName, envelope, grants }, true);
```
cb1kenobi: **Low: the client-side grant merge is avoidable — `set_secret` already preserves grants when `grants` is omitted**

`resolveGrants`'s own docblock quotes the mechanism: `grants = dedupeGrants(req.grants ?? existing?.grants)` (`components/secretOperations.ts:194`). Omitting `grants` therefore keeps the existing list server-side, under `withSecretLock` — atomically. Computing the union on the client instead costs an extra round trip, opens a TOCTOU (a `revoke_secret` landing between `list_secrets` and this call is silently reinstated), and leans on a fallback that cannot run (line 106).

Suggested fix: omit `grants` here and follow with `grant_secret { name, component }`, which is idempotent (`components/secretOperations.ts:262`) and serialized per-name. Same two round trips, no merge logic, no dead branch, and `resolveGrants` goes away entirely. The one caveat is a brand-new row sitting ungranted between the two calls — harmless in practice, since the row is unusable until the grant lands anyway.

Secondary, if `resolveGrants` is kept: the notice guard on line 102 is `existing.length && !existing.includes(component)`, so the ordinary rotation case — the row already grants this component *plus* others — prints nothing. `existing.some((g) => g !== component)` covers it.

—
_Generated by Barber AI_

bin/harper.ts
```diff
@@ -170,6 +170,21 @@ async function harper() {
...
+			// A `token=` that didn't reach the setup flow is a mistyped invocation, not a deploy field —
+			// `harper deploy setup token=…` (no `=true`) parses `setup` as a bare word that buildRequest
+			// drops, so it would otherwise proceed as an ordinary deploy carrying a live credential it has
+			// no use for. Refuse rather than deploy: the token is redacted and never sent either way, but
+			// silently ignoring it would leave the user believing a credential was provisioned.
+			if (cliApiOp.operation === 'deploy_component' && cliApiOp.token !== undefined) {
+				throw new Error('`token=` is only valid with `setup=true` — did you mean `harper deploy setup=true`?');
```
cb1kenobi: **Nit: this guard prints a Node stack trace for a plain typo**

`formatCliError` (`bin/harper.ts:25`) strips the stack only for errors carrying a numeric `statusCode`; a bare `new Error` keeps it, so a mistyped `harper deploy setup token=...` gets `error: <message>` followed by a full stack trace. Every comparable CLI validation error (the `deploy by_ref: ...` throws in `bin/cliOperations.ts`) is raised inside `cliOperations` and prints message-only, so this one reads as a crash rather than the actionable hint it is.

Suggested fix: attach a numeric `statusCode` to the error (or throw the `ClientError` shape `formatCliError` already special-cases) so it prints as a clean one-liner.

—
_Generated by Barber AI_

bin/cliOperations.ts
dawsontoth: Fixed in ff4c1bbb8 — and wrong in the direction that matters, not just ambiguous. I verified before rewording: `credentials` is deliberately absent from `TRANSPORT_ONLY_FIELDS`, so the nested token IS sent and `ingestCredentials` seals it server-side; the only strip is the operations-log destructure. The comment now says exactly that, so the two mechanisms cannot be read as one.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

bin/deploySetup.ts
dawsontoth: Fixed in ff4c1bbb8. `resolveGrants` is exported and covered four ways: the union (row grants `other-app` → both retained, and the preserved grant is reported), no duplicate when the row already grants the component, `[component]` when the row does not exist yet, and the lookup-failure fallback asserting both the return and the warning — since a silent replace is the thing that was wrong, the visibility is part of the fix and worth pinning too.

Fair catch on the asymmetry: the other two fixes each shipped with a regression test and this one did not. I also confirmed the union case fails when `resolveGrants` is reverted to `[component]`, rather than trusting a green run — the same check caught that my earlier `prepareDeployByRef` test could not fail at all.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

bin/deploySetup.ts
dawsontoth: Confirmed and fixed in c553efdc6. Verified it myself before acting: every failure path in `cliOperations` ends in `process.exit(1)` (lines 755, 958, 993, 1024), so it never throws and both the `catch` and its yellow notice were unreachable — and worse, the docblock advertised the opposite of what would happen, with the abort landing *after* the PAT was already pasted.

Rather than find a call path that throws, I took your line-261 suggestion and dropped the read entirely, which removes the dead branch along with the reason for it.

One thing you did not flag: the identical dead `try/catch` wrapped `get_secrets_public_key` at the top of the same function (pre-existing, from the first commit on this branch). Same defect, same false promise of a friendlier message, so it is gone too — fixing only the reported instance would have left the other in place.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

bin/deploySetup.ts
dawsontoth: Adopted in c553efdc6 — this is better than what I wrote, on every count. `set_secret` now omits `grants`, and idempotent `grant_secret { name, component }` follows it. `resolveGrants` is deleted.

I checked the preconditions before switching: `grants` and `processEnv` are both optional in `setSecretValidator` (only `xor(value, envelope)` is required), `dedupeGrants(undefined)` returns `[]`, and `mutateGrants` requires the row to exist — so set_secret must come first, which the tests now pin as an ordering assertion rather than a comment.

The TOCTOU you describe was the real argument: my version would silently reinstate a `revoke_secret` that landed between the read and the write. Two locked server-side steps have no such window. Your secondary point is moot now — the notice guard is gone with the function, and the printed summary reports the grant list the *server* returns, so a preserved grant shows up without the client guessing at it.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

bin/harper.ts
dawsontoth: Fixed in c553efdc6 — the guard now carries `statusCode: 400`, so `formatCliError` prints it as a one-line hint.

Checked the rest of the file while I was in there: all five user-facing throws in `bin/deploySetup.ts` had the same defect (bare `new Error`, so an unusable git host or a missing custody key also printed a stack for what is plainly operator input). They now go through a small `cliError()` helper. Fixing only the line you pointed at would have left the flow inconsistent with itself.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

bin/deploySetup.ts
```diff
@@ -0,0 +1,284 @@
...
+ * any deploy until the grant lands. Nothing is printed until both steps succeed, so a failure part-way
+ * doesn't claim a credential is ready to use.
+ */
+export async function storeSealedSecret(
+	transport: any,
+	secretName: string,
+	envelope: string,
+	component: string
+): Promise<string[]> {
+	await cliOperations({ ...transport, operation: 'set_secret', name: secretName, envelope }, true);
```
cb1kenobi: **Medium: setup can store the new credential as a global secret before failing**

`storeSealedSecret()` rotates the row with `set_secret { envelope }` and only then calls `grant_secret`. If the derived row already exists with `processEnv: true`, `set_secret` preserves that tier (`components/secretOperations.ts:193`), writes the pasted PAT/npm token as a global process-env secret, and the next call fails because `grant_secret` rejects processEnv rows (`components/secretOperations.ts:229`). The CLI reports setup failed, but the durable credential has already been stored in the global tier inherited by components and child processes.

Suggested fix: force this deploy credential into the scoped tier before granting, e.g. include `processEnv: false` on the `set_secret` call (with grants still omitted so existing scoped grants are preserved), then keep the `grant_secret` step. Add a regression test with a pre-existing `processEnv: true` row for the derived name.

—
_Generated by Barber AI_

dawsontoth: Fixed in cf8fc7c0d — and this was a regression from c553efdc6, my own previous fix, which is worth stating plainly.

Traced the mechanism before changing anything. Sending `grants: [component]` had been masking it: the server rejects processEnv+grants *before* `table.put`, so the old shape threw with nothing stored. Dropping `grants` so the server could merge them removed that accidental guard, and `dedupeGrants(existing?.grants)` on a processEnv row yields `[]`, so the guard no longer fired and the write went through. The protection I relied on was a side effect of a field I had just removed for unrelated reasons.

`processEnv: false` is now explicit on the `set_secret` call, with `grants` still omitted so scoped grants are preserved. A pre-existing global row is converted to the scoped tier, which is what the flow advertises, and since the row name is component-derived a global secret at it was never serving anything the scoped one does not.

Test asserts `set_secret` carries `processEnv: false`, and I verified it fails when the field is removed rather than trusting a green run. I also corrected the flow summary at the top of the file, which still described a single granting `set_secret`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code) (Opus 5)_

bin/deploySetup.ts
```diff
@@ -0,0 +1,294 @@
...
+			`"${component}" is not a usable component name — a deploy accepts letters, numbers, dashes and underscores.`
+		);
+	}
+
+	let credentialKey: string; // host (github) or registry (npm) — the credentials-entry discriminator
+	let credentialEntry: Record<string, string>;
+	let token: string | undefined = req.token;
+
+	if (provider === 'github') {
+		const host = resolveGitHost(req.host ?? 'github.com');
```
kriszyp: Could we give the Git-provider hostname its own setup-only option (or derive it from `package`) and reserve `host` for its established virtual-route meaning? The regular deploy schema defines `host` as the component’s routed hostname at `components/operationsValidation.js:476-492`, which is persisted at `components/operations.js:413-414`. On this setup-only path, the same argument is normalized, used for `gh auth token --hostname`, and stored as the credential host. Consequently, `harper deploy setup=true project=web package=github:org/repo host=api.example.com` provisions a credential for `api.example.com`; the later GitHub clone requests `github.com` and ignores it. Setup bypasses server validation, so nothing reports the mix-up.

bin/deploySetup.ts
```diff
@@ -0,0 +1,294 @@
...
+ * any deploy until the grant lands. Nothing is printed until both steps succeed, so a failure part-way
+ * doesn't claim a credential is ready to use.
+ */
+export async function storeSealedSecret(
+	transport: any,
+	secretName: string,
+	envelope: string,
+	component: string
+): Promise<string[]> {
+	await cliOperations({ ...transport, operation: 'set_secret', name: secretName, envelope, processEnv: false }, true);
```
cb1kenobi, line 128: **Medium: splitting the write means a replicated deploy can 403 on a peer**

This is the fix suggested last round, and it does remove the TOCTOU — but it trades it for a replication window that the old single write didn't have.

`set_secret` + `grant_secret` now produce **two** row versions: v1 carries the envelope with `grants: []` (brand-new row → `dedupeGrants(req.grants ?? existing?.grants)` with no existing row), v2 adds the grant. The peer-side resolver was written against the old invariant *row exists ⇒ grant present*:

- `waitForSecretRow` (`components/secretOperations.ts:456-464`) polls for row **existence** only.
- `resolveCredentials` then reads `row.grants` **once** (`:521`) and throws `403 … is not granted to component` with no retry.
- `components/operations.js:498-507` sets that grace period *only* when `isReplicatedExecution` — it exists precisely for this race.

So a peer that has applied v1 but not v2 when the replicated `deploy_component` lands fails the deploy on that node while the origin succeeds. That is the documented CI flow (`setup=true` then `by_ref=true`), seconds apart. At the previous head `set_secret` carried `grants`, so the grant landed atomically with the envelope and the window did not exist.

The docblock above (line 118-120) says the gap is *"harmless — it can't be resolved by any deploy until the grant lands."* True on the origin; not true on a peer already inside the bounded wait.

Suggested fix: make the wait cover the authority rather than mere existence — pass `component` into `waitForSecretRow` and keep polling while `!row.processEnv && !row.grants.includes(component)`, throwing the 403 only after the deadline. That enforces the real invariant ("the wait ends when the credential is usable") instead of reverting this client protocol, which is otherwise a clear improvement.

—
_Generated by Barber AI_

bin/cliOperations.ts
```diff
@@ -536,7 +538,11 @@ export function prepareDeployByRef(req: any): void {
 	// left is pushed by construction — and a shallow/detached runner checkout has no remote-tracking
 	// branches to check it against anyway.
 	if (!process.env.GITHUB_SHA) warnIfCommitNotPushed(committish);
-	if (!req.project) req.project = defaultProjectName(process.cwd());
+	// The same default the payload deploy and `deploy setup=true` use. A by-reference deploy is the
+	// flow the sealed credential feeds, and `set_secret`'s grant is keyed by project name, so a
+	// package.json-derived name here would ask for a secret granted to a different name whenever a
+	// checkout directory and its package name disagree.
+	if (!req.project) req.project = directoryProjectName();
```
cb1kenobi, line 545: **Medium: `by_ref` still derives the secret name from an un-canonicalized project**

Unifying on `directoryProjectName()` fixed the *default source*, but not the *canonicalization*. Line 553 below derives the secret from raw `req.project`, while both the server (`components/operations.js:371`, `canonicalProjectName`) and `deploy setup=true` (`bin/deploySetup.ts:69,178`) canonicalize first. Ran both derivations against the built module:

```
"chart.js"    by_ref -> deploy.chart.js.git.github.com    setup -> deploy.chart.git.github.com   server runs as -> chart
"my.app"      by_ref -> deploy.my.app.git.github.com      setup -> deploy.my.git.github.com      server runs as -> my
"@scope/app"  by_ref -> deploy._scope_app.git.github.com  setup -> deploy.app.git.github.com     server runs as -> app
"my-app"      by_ref -> deploy.my-app.git.github.com      setup -> deploy.my-app.git.github.com  server runs as -> my-app  (agrees)
```

Any project or checkout directory containing a dot (`chart.js`, `my.app`) or an explicit `project=@scope/app` diverges, and `resolveCredentials` throws `references secret '…', which does not exist` — the silent-until-deploy mismatch this shared module exists to prevent. It fails closed, which is why this is Medium rather than High. The new tests only use dot-free names (`demo`, `my-app`), so nothing catches it.

Suggested fix:
```suggestion
	if (!req.project) req.project = directoryProjectName();
	req.project = canonicalProjectName(req.project);
```
(with `canonicalProjectName` added to the `componentNames.ts` import on line 23) so the printed name, the wire value and `deriveGitSecretName` all agree with what the server will run as. Worth a `project: 'my.app'` case in the `credential reference` describe block.

—
_Generated by Barber AI_

bin/deploySetup.ts
```diff
@@ -0,0 +1,294 @@
...
+				.token;
+		}
+		credentialEntry = req.scope ? { registry, scope: req.scope } : { registry };
+	}
+
+	token = typeof token === 'string' ? token.trim() : undefined;
+	if (!token) throw cliError('No token was provided; nothing to store.');
+
+	// 4. Seal the token locally. Only ciphertext leaves this machine.
+	const envelope = ENV_ENCRYPTED_PREFIX + encryptEnvelope(token, publicKey, fingerprint);
```
cb1kenobi, line 266: **Low: the envelope's `kid` is the server's self-reported fingerprint, never checked against the key**

`fingerprint` is read off the `get_secrets_public_key` response (line 148) and passed straight through as the `kid` without verifying it belongs to `publicKey`. `utility/secretEnvelope.ts:31` already exports `fingerprintOf()`, and it is used only in tests and fixtures — never in production code.

The effect is that `setSecret`'s ingest check (`components/secretOperations.ts:168-175`, kid vs custody fingerprint) is tautological for this client: it compares a value the same server just handed out, so it cannot catch a token sealed to the wrong key. A `(public_key, fingerprint)` disagreement — a custody bug, or a node caught mid-rekey — is accepted at store time and only surfaces at deploy time as `Failed to decrypt credential secret '…'`, well after the CLI printed `✓ Sealed …`.

To be clear this is fail-fast on an inconsistent cluster, not a MITM defense (an attacker substituting both fields stays self-consistent; the server's kid check is what catches that).

Suggested fix:
```suggestion
	const envelope = ENV_ENCRYPTED_PREFIX + encryptEnvelope(token, publicKey, fingerprintOf(publicKey));
```
importing `fingerprintOf` alongside `encryptEnvelope`, and ideally throwing a `cliError` when it disagrees with the reported `fingerprint` so the inconsistency is named rather than silently papered over.

—
_Generated by Barber AI_

utility/componentNames.ts
```diff
@@ -0,0 +1,107 @@
...
+ * `app`. Every project-scoped operation reduces its `project` this way, so the CLI has to resolve the
+ * same name the server will when it grants a credential to one.
+ */
+export function canonicalProjectName(project: string): string {
+	return path.parse(project).name;
+}
+
+/** The project name a deploy derives from its `package` spec when no `project` was given. */
+export function projectNameFromPackage(pkg: string): string {
+	if (pkg.startsWith('git+ssh://')) {
```
cb1kenobi, line 50: **Low: `git+https://` isn't handled — the exact form this flow emits**

The branches cover `git+ssh://`, `http://`, `https://` and `file://`, but `git+https://` matches none of them (`'git+https://…'.startsWith('https://')` is `false`), so it falls through to the bare `path.basename` at the end:

```
git+ssh://git@github.com/owner/repo.git#abc   -> "repo"
git+https://github.com/owner/repo.git#abc     -> "repo.git#abc"   <-- fails PROJECT_NAME_PATTERN
```

This is moved verbatim rather than a regression, but it now matters more: `prepareDeployByRef` (`bin/cliOperations.ts:548`) builds exactly `git+https://…`, and `harper deploy setup=true package=git+https://…` reaches it through the new `resolveComponentName`. The operator gets `"repo.git#abc" is not a usable component name` instead of `repo`.

The `git+ssh://` branch already does the right thing (splits the `#ref` off before stripping `.git`), so it just needs to accept both spellings:

```suggestion
	if (pkg.startsWith('git+ssh://') || pkg.startsWith('git+https://')) {
```

Note the `http(s)://` branch below can't be reused here — its `.replace(/\.git$/, '')` is anchored, so with a `#ref` fragment it yields `repo.git`. Verified the suggestion above returns `repo` for `git+https://github.com/owner/repo.git#abc` and `my-app` for `git+https://ghe.example.com/team/my-app.git#v1.2.3`.

—
_Generated by Barber AI_

bin/cliOperations.ts
```diff
@@ -53,8 +55,29 @@ const TRANSPORT_ONLY_FIELDS = new Set([
 	'by_ref',
 	'ref',
 	'credential',
+	// `deploy setup=true`'s token, read off the parsed request by deploySetup and sealed locally. No
+	// operation takes a *top-level* `token`, so keeping it out of every body costs nothing and means a
+	// mistyped `setup` — which parses as a bare word and falls through to a real deploy — can't carry a
+	// PAT to the server. Distinct from `credentials[].token`, which is nested inside a field that IS
+	// sent (ingestCredentials seals it server-side) and is only kept out of the operations log.
+	'token',
```
cb1kenobi, line 63: **Nit: `token=` is JSON-parsed like an operation field, unlike `ref=`**

`RAW_STRING_FIELDS` (line 84) contains only `'ref'`, so `buildRequest` (lines 611-617) runs `JSON.parse` over the `token=` value. A credential that happens to be valid JSON — an all-digit token — becomes a number, and `bin/deploySetup.ts:262`'s `typeof token === 'string'` guard then discards it and reports `No token was provided; nothing to store.`

Misleading rather than dangerous, and the guard means nothing unsafe happens. But `token` is an opaque secret string for precisely the reason `ref` was added to that set — the comment above it ("Values that are opaque strings, never JSON") already describes `token`.

Suggested fix, at line 84:
```js
const RAW_STRING_FIELDS = new Set(['ref', 'token']);
```

—
_Generated by Barber AI_

PR #1853, opened 2026-07-17
fix(storage): honest Promise<number | void> type for commit-latency recorder

kriszyp:
## Summary
Fixes the org-wide **red `main` build** (`tsc` TS2345) introduced by #1688.

`resources/DatabaseTransaction.ts:367` passed `commitResolution` to `recordCommitLatency(_: Promise<void>, ...)`, but `commitResolution` is declared `Promise<number | void> | void`. The `as Promise<void>` cast on the `transaction.commit()` assignment (line 360) was **ineffective**: TypeScript's assignment narrowing filters the declared union and re-widens to the `Promise<number | void>` member, so the cast never took effect and the argument stayed `Promise<number | void>`.

## Fix
`commit()` honestly resolves to `number | void` — the coordinated-retry `RETRY_NOW_VALUE` sentinel, which is checked in the resolve callback (line 397). So:
- Widen `recordCommitLatency`'s parameter to `Promise<number | void>` (the honest type).
- Drop the misleading/ineffective `as Promise<void>` cast and correct the adjacent comment.

`recordCommitLatency` only records the submit→settle **timing** and ignores the resolved value, so the per-attempt latency metric is unchanged — the transient-conflict retry still rejects and re-issues `commit()`, re-arming per attempt.

## Where to look
- `resources/DatabaseTransaction.ts:45` — widened recorder parameter.
- `resources/DatabaseTransaction.ts:360` — cast removed; comment corrected.

## Tests
- `npm run build` (tsc) — clean.
- `unitTests/resources/transactionCommitLatency.test.js` — 2 passing.
- `npm run test:unit:resources` — 1162 passing, 14 pending.

Refs #1688.

### DIFF
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -41,8 +41,10 @@ export function setCommitLatencyRecorder(recorder: ((durationMs: number) => void
 // queue time. The recorder is wrapped so it can never throw — a metrics failure must neither break the
 // commit nor surface as an unhandled rejection on this floating `.then`. The thenable guard protects
 // against a future caller passing a non-Promise `commitResolution` (today it is always the rocksdb-js
-// async `Transaction.commit()` result, which is guaranteed to be a Promise).
-function recordCommitLatency(commitResolution: Promise<unknown> | void, submittedAt: number) {
+// async `Transaction.commit()` result, which is guaranteed to be a Promise). The parameter matches
+// `commit()`'s honest `Promise<number | void>` result (the coordinated-retry sentinel); the resolved
+// value is intentionally ignored — only the settle timing is recorded.
+function recordCommitLatency(commitResolution: Promise<number | void>, submittedAt: number) {
 	if (!recordCommitLatencyMs) return;
 	const record = () => {
 		try {
@@ -400,8 +402,8 @@ export class DatabaseTransaction implements Transaction {
 						if (this.writes.length > 0) {
 							// The transaction was created with coordinatedRetry:true (see
 							// getReadTxn), so commit() can resolve to RETRY_NOW_VALUE. That
-							// sentinel is handled in the resolve callback below and never
-							// propagates past that branch.
+							// sentinel (a number) is why commitResolution is typed
+							// Promise<number | void>; it is handled in the resolve callback below.
 							commitResolution = transaction.commit();
 							// Record how long this commit stays outstanding (submit → settle) as a distribution
 							// metric. This is the same clock the overload check uses (outstandingCommitStart is


resources/DatabaseTransaction.ts
```diff
@@ -41,8 +41,10 @@ export function setCommitLatencyRecorder(recorder: ((durationMs: number) => void
 // queue time. The recorder is wrapped so it can never throw — a metrics failure must neither break the
 // commit nor surface as an unhandled rejection on this floating `.then`. The thenable guard protects
 // against a future caller passing a non-Promise `commitResolution` (today it is always the rocksdb-js
-// async `Transaction.commit()` result, which is guaranteed to be a Promise).
-function recordCommitLatency(commitResolution: Promise<void>, submittedAt: number) {
+// async `Transaction.commit()` result, which is guaranteed to be a Promise). The parameter matches
+// `commit()`'s honest `Promise<number | void>` result (the coordinated-retry sentinel); the resolved
+// value is intentionally ignored — only the settle timing is recorded.
+function recordCommitLatency(commitResolution: Promise<number | void>, submittedAt: number) {
```
cb1kenobi, line 47: **Note: three open PRs fix this same `recordCommitLatency` type error — only one should land.**

#1853 (this), #1887, and #1869 all edit these same lines in `DatabaseTransaction.ts`, so whichever merges first leaves the other two with a conflict. Of the three, this is the cleanest fix: it makes the parameter type honest (`Promise<number | void>`, matching `commit()`'s actual result), drops the ineffective `as Promise<void>` cast, and corrects the comment.

- #1887 re-adds an `as Promise<void>` cast at the call site (effective this time, but it keeps the misleading `Promise<void>` parameter and the cast), bundled into an unrelated astral-Unicode fix.
- #1869 widens the parameter to `Promise<unknown>` — honest but less precise, dropping the `number | void` information — also bundled into an unrelated `audit:false` delete fix.

Suggest landing this dedicated fix and dropping the type change from the other two.

—
_Generated by Barber AI_

kriszyp: Yeah ,we are really going to make sure we land this fix!

cb1kenobi: **Update: the duplicate-fix note above is stale.**

Since that comment, #1890 (`fix(types): widen recordCommitLatency parameter so the build type-checks`) closed and [#1899](https://github.com/HarperFast/harper/pull/1899) (`Fix recordCommitLatency parameter type for the widened commitResolution union`) merged — landing the same TS2345 fix with the looser `Promise<unknown> | void` type. Those were the actual competing PRs, not #1887/#1869 as previously stated (#1869 no longer touches this file; #1887's edit here is unrelated, further down in the same method).

#1853 is now a clean refinement on top of #1899's merged fix: narrows the parameter to the more precise `Promise<number | void>` (matching `commit()`'s actual resolved type, including the `RETRY_NOW_VALUE` sentinel) and updates the adjacent comments. `mergeable: true` against current main — no conflict.

—
_Generated by Barber AI_

cb1kenobi (approved):

Issue #1785, opened 2026-07-14
Ingest permanently hangs under sustained sequential bulk `upsert` load


pbrumblay:
### Bug Summary

Ingest permanently hangs under sustained sequential bulk `upsert` load: one write's response never arrives, no error is thrown, no timeout fires, and the process becomes otherwise idle (0% CPU) while continuing to answer unrelated requests normally.

### Reproduction Repository (if applicable)

https://github.com/pbrumblay/harper-hang-repro

### Steps to Reproduce

1. `npm install harper@5.1.18` in an empty directory.
2. Run the attached `repro.js` (`node repro.js`). It will:
   - `harper install` + `harper run` with a dedicated `ROOTPATH` (and an isolated `$HOME`,
     since `harper install` tracks a single canonical install via
     `$HOME/.harperdb/hdb_boot_properties.file`, independent of `--ROOTPATH` -- worth knowing
     if you're running multiple installs on one machine).
   - Create one plain table (`BulkHangRepro`, `id` hash attribute, a few scalar columns).
   - Issue sequential `{operation: 'upsert', database, table, records}` calls over the plain
     Operations API, batches of 1000 synthetic records with strictly new, distinct keys each
     batch (no repeated updates to the same keys) -- one `await`ed batch after another, no
     concurrency.
   - Watch wall-clock time per batch; if one exceeds 90s (vs. the ~40-135ms/batch normal
     rate), flag a stall and capture diagnostics (`ps`, `top`, `sample <pid> 5`, `lsof`,
     Harper's own log tail).
3. Wait. In our runs this took anywhere from ~20 minutes to ~1 hour of sustained load
   (not fast/deterministic, but reliable across every attempt).

Env overrides available on the script: `BATCH_SIZE`, `TOTAL_BATCHES`, `STALL_MS`,
`OPERATIONSAPI_NETWORK_PORT`. Defaults reproduced it on the first run with no tuning.

### Expected Behavior

The `upsert` either completes, or fails with an error/timeout that the caller can detect and
act on.

### Actual Behavior

The `upsert` call never resolves and never rejects -- indefinitely (14+ hours in one
occurrence before we intervened manually). Specifically:

- The Harper process sits at **0% CPU**, state `sleeping` -- not busy, not spinning.
- A native thread sample (`sample <pid> 5`) taken during the hang shows **every thread idle**:
  the main thread parked in `uv__io_poll`/`kevent` (event loop has nothing to do), RocksDB's
[...]

PR #1860, opened 2026-07-18
fix(txn): don't drop writes staged while read iterators defer the commit

kriszyp:
## Summary

A commit issued while a search iterator still holds the read transaction cannot commit the native handle — the iterator's native cursor lives inside it (`GetIterator` wraps the transaction's write batch + snapshot, and RocksDB invalidates it on commit). The old code deferred the entire native commit to `doneReadTxn()` (the LINGERING state), and the LINGERING fall-through cleared `this.writes`, so the deferred commit found an empty write set and **aborted the native transaction with the staged writes in it** — every write silently dropped after the caller was told the commit succeeded.

Found while root-causing the #1785 follow-up; empirically confirmed on main — the new test fails without the fix.

## Approach (reworked after review discussion)

The first version of this PR preserved `this.writes` across the deferral so the deferred commit worked. Review (cb1kenobi) showed the deferral itself is the hazard: any path that keeps the last iterator from finishing cleanly — the long-transaction monitor's timeout abort, a hung stream — still dropped acknowledged writes. So this now **eliminates the deferred commit entirely** on the RocksDB path:

- `commit()` with outstanding iterators replays the staged writes onto a **fresh transaction and commits them immediately**, ERR_TRY_AGAIN-replay style: entries reload through the new transaction, commit handlers re-base on them (CAS re-resolves against current state), and conflicts surface through the normal retry ladder. The replay is part of the awaited commit chain, so the caller's acknowledgment now covers actual durability — and a terminal commit failure **rejects the awaited chain** instead of being loggable-only.
- The replay deliberately does **not** carry `isRetry`: audit/txn-log entries batch natively on the transaction they were staged into and are only written by its commit *attempt* (the ERR_TRY_AGAIN replay relies on the original's *failed* commit having already written its batch; here the original never attempts one). The replayed stagings re-append their entries into the replay transaction's own batch — the test asserts exactly one audit entry (no loss, no duplication).
- The original native handle stays open solely for the iterators and is aborted, with no remaining responsibilities, when the last one finishes. `TRANSACTION_STATE.LINGERING` is now LMDB-only (LMDB overrides all the relevant methods; its path is unchanged).
- `save()` no longer stages writes into a retained non-OPEN handle (they would be discarded by that abort) — internal `txnForContext` writers now commit immediately on a fresh transaction.
- The long-transaction monitor never aborts a CLOSED (already-acknowledged) transaction's writes: for an overdue CLOSED-but-tracked transaction it releases just the read snapshot (`releaseReadTxn()`), preserving the open-time bound for hung iterators without re-introducing write loss (the cb1kenobi finding).
- The commit's base read reference is now consumed **exactly once per read handle** (`baseReadRefConsumed`), so retry rounds, immediate-commit re-entries, and a second top-level `commit()` can't steal an iterator's reference and abort the handle under it.
- Multi-store: `this.next.commit()` no longer inherits a retry/replay round's head-store `options.transaction` (pre-existing bug; the next store must commit through its own transaction).

## Where to look

- The outstanding-iterators branch in `DatabaseTransaction.commit()` (the replay) and the simplified `doneReadTxn()`.
- The monitor's new CLOSED branch + `releaseReadTxn()`.
- `unitTests/resources/lingeringWriteCommit.test.js` — asserts immediate durability while the iterator is still open, exactly-one audit entry, independent post-close writes, and that a terminally failing replay commit rejects the awaited chain while the iterator survives.
- A `renewAfterCommit`/`CommitAndTryCreateSnapshot` + iterator-reseek primitive in rocksdb-js would let the original handle commit directly (the old TODO's intent) and replace the replay; noted as follow-up work.

Likely a v5.1 patch candidate — the bug exists on the release branch.

Cross-model reviewed (Codex traced repo paths; Gemini via direct API on the diff): Codex's audit-entry blocker, read-reference theft, and next-chain transaction leak are fixed in the final commits; coverage: codex ✓ / gemini ✗ (agy hung) → api ✓.

Generated by an LLM (Claude Fable 5 + Claude Sonnet 5), supervised by Kris.

### DIFF
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -255,14 +255,20 @@ export class DatabaseTransaction implements Transaction {
 				// requested the commit already received its resolution (the writes were acknowledged when
 				// the transaction went LINGERING, since blocking on iterator completion could deadlock a
 				// response that streams the iterator), so a failure here cannot reject any awaited chain —
-				// log it loudly rather than leaking an unhandled rejection, and abort so the failed
-				// writes' blob files are cleaned up (no error middleware runs this late).
-				const committed = this.commit() as Promise<CommitResolution>;
-				if (committed?.then)
-					committed.catch((error) => {
-						harperLogger.error?.('Failed to commit lingering writes after read iterators completed', error);
-						this.abort();
-					});
+				// log it loudly rather than leaking an unhandled rejection (or throwing into the iterator
+				// consumer: with no async completions pending, commit() runs — and can throw —
+				// synchronously), and abort so the failed writes' blob files are cleaned up (no error
+				// middleware runs this late).
+				const onDeferredCommitError = (error: any) => {
+					harperLogger.error?.('Failed to commit lingering writes after read iterators completed', error);
+					this.abort();
+				};
+				try {
+					const committed = this.commit() as Promise<CommitResolution>;
+					if (committed?.then) committed.catch(onDeferredCommitError);
+				} catch (error) {
+					onDeferredCommitError(error);
+				}
 			} else {
 				this.transaction?.abort();
 				this.transaction = null;
@@ -618,7 +624,17 @@ export class DatabaseTransaction implements Transaction {
 									);
 								}
 								return this.commit({ ...options, transaction }); // try again
-							} else throw error;
+							} else {
+								// terminal (non-conflict) failure: release the native handle so it doesn't leak;
+								// usually already released by the failed commit itself, abort for the unexpected
+								// case (same defensive pattern as the retry-exhaustion give-up above)
+								try {
+									transaction.abort();
+								} catch (abortError) {
+									harperLogger.debug?.('aborting transaction after failed commit', abortError);
+								}
+								throw error;
+							}
 						}
 					);
 				}
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -190,6 +190,9 @@ export class DatabaseTransaction implements Transaction {
 	retries = 0;
 	declare next: DatabaseTransaction;
 	declare stale: boolean;
+	// Whether this read handle's base reference (readTxnsUsed starts at 1 in getReadTxn) has been
+	// consumed by a commit round; iterator references are consumed only by doneReadTxn().
+	declare baseReadRefConsumed?: boolean;
 	declare startedFrom?: {
 		resourceName: string;
 		method: string;
@@ -234,6 +237,7 @@ export class DatabaseTransaction implements Transaction {
 		}
 
 		this.readTxnsUsed = 1;
+		this.baseReadRefConsumed = false; // fresh handle, fresh base reference for commit() to consume
 		if (DEBUG_LONG_TXNS) {
 			this.stackTraces = [new StartedTransaction()];
 		}
@@ -256,7 +260,8 @@ export class DatabaseTransaction implements Transaction {
 			// The native handle was only being held open for the read iterators: any writes staged
 			// through it were already committed at commit() time by replaying them onto a fresh
 			// transaction (see the outstanding-iterators branch in commit()), so aborting it here
-			// discards nothing — the audit/txn-log entries appended at staging survive the abort.
+			// discards nothing — the replay re-staged the writes AND their audit/txn-log entries
+			// into its own transaction; this handle's never-committed log batch dies with it.
 			trackedTxns.delete(this);
 			this.transaction?.abort();
 			this.transaction = null;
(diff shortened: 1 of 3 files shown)


resources/DatabaseTransaction.ts
```diff
@@ -550,15 +566,31 @@ export class DatabaseTransaction implements Transaction {
...
+								}
+								throw error;
+							}
 						}
 					);
 				}
-				for (const write of this.writes) {
-					if (write?.skipped && write?.savedBlobs)
-						cleanupUnusedBlobs(write.savedBlobs, collectRetainedFileIds(write.store.getEntry(write.key)?.value));
+				if (this.open !== TRANSACTION_STATE.LINGERING) {
```
cb1kenobi: **Medium: the long-transaction monitor can now abort a LINGERING transaction and drop its already-acknowledged writes**

Preserving `this.writes` across the LINGERING deferral is the correct core fix, but it has a side effect on the long-transaction monitor. A LINGERING transaction stays in `trackedTxns` until `doneReadTxn()` runs (the LINGERING branch in `commit()` does not `trackedTxns.delete(this)`), and now that `this.writes` survives, `hasPendingWrites()` returns `true` for it.

So if a streamed response holds the read iterator open longer than `STORAGE_MAXTRANSACTIONOPENTIME` (default ~30s, ~2 monitor ticks), `startMonitoringTxns()` takes the `hasPendingWrites() && !sourceApply && !isReplay` branch and calls `abortDueToTimeout()` — which aborts the native transaction and drops the staged writes. Those writes were already acknowledged to the caller when the transaction went LINGERING (`commit()` resolved), so this is the same silent write-loss-after-ack this PR fixes, reintroduced through the monitor path. Before this PR the LINGERING txn had an empty write set, so the monitor took the harmless commit path instead of aborting.

The #1407 rationale for aborting (surface an error so the app splits the work) does not apply here: the commit already resolved, so there is no in-flight request left to receive the error. Consider excluding LINGERING transactions with a complete, already-acknowledged write set from the force-abort branch (let the deferred `doneReadTxn()` commit finish them), or committing rather than aborting them in the monitor.

Confidence: likely — worth a human trace of the monitor path for a write + long-held streamed iterator in one transaction. RocksDB-only (LMDB never defers).

—
_Generated by Barber AI_

kriszyp: Confirmed — this was real: with `this.writes` preserved, a LINGERING transaction in `trackedTxns` returned `hasPendingWrites() === true`, so a streamed iterator outliving `storage.maxTransactionOpenTime` sent the monitor down the force-abort branch and dropped the acknowledged writes.

Rather than special-casing the monitor, the underlying deferral is now gone entirely (per review discussion with Kris): `commit()` with outstanding iterators replays the staged writes onto a fresh transaction and commits them immediately (ERR_TRY_AGAIN-replay style — entries reload through the new transaction, conflicts go through the normal retry ladder, and the replay re-stages its audit/txn-log entries into its own native batch). The original handle stays open only for the iterators and is aborted empty when they finish, so there is no longer any window where an acknowledged write set sits exposed to the monitor. For an overdue CLOSED-but-tracked transaction the monitor now releases just the read snapshot (`releaseReadTxn()`), preserving the open-time bound for hung iterators without touching writes.

Commits 265e06c0c + c3aaea7d3; the PR description is updated for the new approach.

— Claude Sonnet 5 (dev-agent), supervised by Kris

cb1kenobi (approved):

unitTests/resources/lingeringWriteCommit.test.js
```diff
@@ -0,0 +1,144 @@
...
+		await delay(50); // give a released-handle failure a beat to surface as an unhandledRejection
+		assert.equal(context.transaction.transaction, null, 'the drained iterator must release the native handle');
+		assert.ok(await LingerTable.get('linger-write'), 'releasing the read handle must not disturb the committed write');
+		// audit/txn-log entries batch on the native transaction they were staged into and are only
+		// written by its commit; the replay must re-stage them into ITS transaction — the original
+		// handle's batch dies with the abort above. Exactly one entry: no loss, no duplication.
+		const history = await LingerTable.getHistoryOfRecord('linger-write');
+		assert.equal(history.length, 1, 'the replayed write must carry exactly one audit entry (not zero, not two)');
+		assert.deepEqual(unhandled, [], 'the replay commit must not leak an unhandled rejection');
+	});
```
cb1kenobi, line 83: **Low (test coverage): no regression test exercises the monitor's new CLOSED branch (`releaseReadTxn`).**

These tests pin the commit-time replay and the iterator-drain release, but not the path the prior Medium was actually about: the long-transaction monitor firing on a committed (`open === CLOSED`) transaction whose iterator still holds the read snapshot past `STORAGE_MAXTRANSACTIONOPENTIME`. That path now runs the new `releaseReadTxn()` branch in `startMonitoringTxns()` — the code added specifically to neutralize the finding — and it has no direct coverage.

The risk is low because the write is already durable at ack time (the replay commit is part of the awaited chain, as the first test asserts), so the monitor can't drop an acknowledged write even untested. But a regression test would lock in the branch that exists to prevent exactly the prior bug. `txn-tracking.test.js` already uses `setTxnExpiration(20)` to drive the monitor; a case there (or here) that holds an iterator open across a committed write, lets the monitor tick, then asserts the committed record survives, the snapshot is released (`context.transaction.transaction === null`), and nothing floats an unhandled rejection would close the gap.

—
_Generated by Barber AI_

PR #1869, opened 2026-07-20
fix: thread {transaction} into removeEntry for audit:false table deletes

### LINKED ISSUE #1854: Aborted transaction on audit:false delete still commits row removal + blob unlink + leaves dangling index entries
## Summary

`_writeDelete()`'s non-audit branch removes a record **outside the enclosing transaction**. On any `audit:false` table, an aborted transaction that deletes a record still **durably commits the delete and unlinks its blob** — the abort rolls back everything else, but the removal itself already committed standalone against the raw store. Left unaddressed, this also produces dangling secondary-index entries that grow unboundedly, and — worse — a query can return a record under an indexed value it no longer holds.

Confirmed on current main (`f39ae6cee`), source-verified (not just agent-reported).

## Root cause

`resources/Table.ts:2680`, inside `_writeDelete`'s `transaction.addWrite({ commit: (txnTime, existingEntry, retry, transaction) => {...} })`:

```js
updateIndices(id, existingRecord, null, transaction && { transaction });   // ← transactional
if (audit || trackDeletes) {
	updateRecord(id, null, existingEntry, txnTime, 0, audit, { ..., transaction, ... }, 'delete');  // ← transactional
} else {
	removeEntry(primaryStore, existingEntry);                              // ← NO { transaction }
}
```

`removeEntry(store, entry, options?)` (`resources/RecordEncoder.ts:895`) passes `options` straight into `store.remove(entry.key, options)`. With `options === undefined`, the removal commits standalone against the raw store and never joins the enclosing transaction — so it lands unconditionally, even when that transaction aborts.

This also structurally defeats the #1364 blob-unlink-ordering fix: `removeEntry`'s own comment gates the blob unlink on the removal promise specifically so an aborted/version-conflicted removal can't orphan a blob. On this path the standalone removal always resolves committed, so the guard's `committed !== false` check always passes and `deleteBlobsInObject()` always runs — the #1364 protection is unreachable here because the write never joined the transaction to begin with.

## Blast radius (measured, not assumed)

- **Bounded to `audit:false` tables.** The `audit:true` contrast is clean: that path takes the `updateRecord(...)` branch, which correctly threads `{transaction}` — delete rolls back properly. `audit:false` is a documented, recommended setting for blob-heavy tables.
- **Does NOT generalize to the background TTL expiration sweep.** `evict()` runs in its own internal, non-user-abortable transaction — row and index removed cleanly.
- **Trigger surface is wider than "a handler throws."** It also fires from the background over-time monitor's `abortDueToTimeout()` — no application error is required at all, just a slow-enough delete-bearing transaction on an `audit:false` table.
- **Raw dangling index entries accumulate, unbounded.** Verified with a direct-store, non-blind oracle (reads the raw dupSort DBI via `getRange`, no join through the primary record): 5 independent abort cycles → exactly 5 dangling entries, zero reclaimed. `search_by_value`/the ordinary query surface reports the id gone in this state — a join in `Table.ts` `transformToEntries()` (~line 5105) SKIPs when the primary record is absent, silently masking the dangling entry at read time. (This masking is a general pitfall: it makes `search_by_value` an invalid oracle for detecting index/base divergence — any future test for this needs a direct index-store read.)
- **The phantom entry becomes a user-visible wrong query result on re-insert.** Re-inserting the same id under a different indexed value lifts the masking (the primary record is live again), and the id surfaces under **both** the new value and the stale phantom value — a query for the old value returns a record that does not actually hold it. No error, nothing in the logs. Reachable through the ordinary REST query surface, no privileged access.

A plausible field shape needs no exotic conditions: an `audit:false` indexed table + any handler that deletes then errors (or simply runs long enough to trip the transaction-open-time monitor) + later re-use of the same id (natural with externally-supplied keys — SKUs, user ids, device ids).

## Suggested fix

```js
removeEntry(primaryStore, existingEntry, transaction && { transaction });
```
matching its two sibling branches. This also makes the existing #1364 blob-unlink gate live on this path as originally designed. The fix should be checked against both symptoms (primary-row escape and index divergence), and whether it needs to also reclaim already-leaked entries in existing deployments — a forward-only fix won't clean up phantom entries that already exist on disk.

## Reproduction

```
cd harper && npm run test:integration -- "integrationTests/qa-scratch/qa600-forcecommit-blob.test.ts"   # 6 pass — deterministic abort-then-blob-survives
cd harper && npm run test:integration -- "integrationTests/qa-scratch/qa603-abort-blast-radius.test.ts" # 6 pass / 1 fail — the 1 fail IS the defect; audit:true + TTL-evict contrasts stay green
cd harper && npm run test:integration -- "integrationTests/qa-scratch/qa604-dangling-index.test.ts"     # 3 red by design — direct index-store oracle proves unbounded dangling-entry growth
cd harper && npm run test:integration -- "integrationTests/qa-scratch/qa607-phantom-resurrect.test.ts"  # control green, probe A red — the wrong-query-result defect
cd harper && npm run test:integration -- "integrationTests/qa-scratch/qa609-monitor-abort.test.ts"       # confirms the monitor-abort trigger, no thrown error needed
```

## Confidence

HIGH — direct source reading (`resources/Table.ts` ~2660/2680, `resources/RecordEncoder.ts` `removeEntry`/`recordUpdater`), reproducible across five independent runs, cross-checked on the same commit.

---
*Surfaced by the QA-explorer exploratory campaign for @kris. Filed by Claude (Sonnet 5) for @kris.*

kriszyp:
## Summary
Fixes #1854 — F-147: on an `audit:false` @indexed table, `_writeDelete()`'s non-audit branch called `removeEntry(primaryStore, existingEntry)` without threading the enclosing transaction through, unlike its two sibling branches (`updateIndices`, `updateRecord`), both of which already pass `transaction && { transaction }` / `{ ..., transaction, ... }`.

## Why
`removeEntry()` passes its `options` argument straight to `store.remove(key, options)`. With no `{transaction}`, the removal committed standalone against the raw store instead of joining the ambient request transaction. Practical effect: an aborted transaction that deletes a record on an `audit:false` table still durably removed the base row (and unlinked its blob, since `removeEntry`'s blob-unlink gate keys off the removal's own commit, not the outer transaction) — while everything else in the transaction correctly rolled back. This also left a dangling secondary-index entry, since the index removal (`updateIndices`) *did* roll back but the base-row removal did not.

## Fix
```diff
-						removeEntry(primaryStore, existingEntry);
+						removeEntry(primaryStore, existingEntry, transaction && { transaction });
```
One-line change, matching the pattern already used by the sibling `updateIndices` call two lines above it.

## Test coverage
Added `integrationTests/resources/audit-false-delete-rollback.test.ts` (+ fixture) with two probes against a fresh `@table(audit: false)` + `@indexed` table:
- **P1**: delete + forced throw (transaction abort) — row and secondary-index entry must both survive. Verified this fails without the fix (reverted the one-line change locally, confirmed P1 fails with `AssertionError: Item must still exist after the aborted delete`) and passes with it.
- **P2**: delete without a throw (normal commit) — row and index entry must both be gone (regression guard for ordinary deletes).

Ran `npm run test:unit:main`, `npm run test:unit:resources`, and the new integration test plus neighboring delete/eviction/relationship-transaction integration tests — all green (pre-existing, unrelated `globalIsolation.test.js` SES-compartment failures on `test:unit:main` confirmed present on `main` too, and a pre-existing unrelated `tsc` type error in `DatabaseTransaction.ts` confirmed present on `main` — neither touches this code path).

Note this was generated by an LLM (Claude Sonnet 5).

### DIFF
--- a/integrationTests/resources/audit-false-delete-rollback/schema.graphql
+++ b/integrationTests/resources/audit-false-delete-rollback/schema.graphql
@@ -1,0 +1,12 @@
+# F-147 / #1854 — aborted delete on an audit:false @indexed table must roll back cleanly.
+#
+# `_writeDelete()`'s non-audit branch (resources/Table.ts) removes the base row via
+# removeEntry(). Without threading the enclosing transaction into that call, the removal
+# commits standalone against the raw store and survives a transaction abort, leaving a
+# dangling secondary-index entry behind. `audit: false` is required to reach this branch
+# (the audit:true branch goes through updateRecord(), which already threads {transaction}).
+type Item @table(audit: false) @export {
+	id: ID @primaryKey
+	bucket: String @indexed
+	payload: String
+}
--- a/integrationTests/resources/audit-false-delete-rollback/schema.graphql
+++ b/integrationTests/resources/audit-false-delete-rollback/schema.graphql
@@ -9,4 +9,9 @@ type Item @table(audit: false) @export {
 	id: ID @primaryKey
 	bucket: String @indexed
 	payload: String
+	# Deliberately kept above the file-storage threshold in tests so it is file-backed —
+	# an inline blob has no file to orphan, so it wouldn't exercise removeEntry's
+	# blob-unlink gate (the second half of the F-147 defect: an aborted removal must not
+	# unlink the blob file either).
+	blob: Blob
 }
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -2714,7 +2714,7 @@ export function makeTable(options) {
 						);
 						if (!audit || isRocksDB) scheduleCleanup();
 					} else {
-						removeEntry(primaryStore, existingEntry, transaction && { transaction });
+						removeEntry(primaryStore, existingEntry, isRocksDB && transaction ? { transaction } : undefined);
 					}
 				},
 			} as any);
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -2714,7 +2714,7 @@ export function makeTable(options) {
 						);
 						if (!audit || isRocksDB) scheduleCleanup();
 					} else {
-						removeEntry(primaryStore, existingEntry);
+						removeEntry(primaryStore, existingEntry, transaction && { transaction });
 					}
 				},
 			} as any);
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -2714,6 +2714,7 @@ export function makeTable(options) {
 						);
 						if (!audit || isRocksDB) scheduleCleanup();
 					} else {
+						// Only RocksDB's remove() takes an options object; on LMDB the 2nd arg is ifVersion, and its writes already join the batched txn.
 						removeEntry(primaryStore, existingEntry, isRocksDB && transaction ? { transaction } : undefined);
 					}
 				},
--- a/integrationTests/resources/audit-false-delete-rollback.test.ts
+++ b/integrationTests/resources/audit-false-delete-rollback.test.ts
@@ -1,0 +1,160 @@
+/**
+ * F-147 / #1854 — aborted delete on an audit:false @indexed table must roll back cleanly.
+ *
+ * `_writeDelete()`'s non-audit branch (resources/Table.ts, inside
+ * `transaction.addWrite({ commit: (txnTime, existingEntry, retry, transaction) => {...} })`)
+ * used to call `removeEntry(primaryStore, existingEntry)` without threading the enclosing
+ * transaction in, unlike its two sibling branches (`updateIndices(..., transaction && {
+ * transaction })` and `updateRecord(..., { transaction, ... })`). `removeEntry()` passes its
+ * options straight to `store.remove(key, options)`, so with no `{transaction}` the removal
+ * committed standalone against the raw store instead of joining the ambient transaction — an
+ * aborted transaction still durably removed the row (and its blobs, and left the secondary
+ * index dangling), because the removal never had a chance to roll back.
+ *
+ * `audit: false` is required to reach this branch: the `audit: true` path goes through
+ * `updateRecord()`, which already threads `{transaction}` and rolls back correctly.
+ *
+ * P1: DELETE + throw (transaction abort) on an audit:false table — the row and its
+ *     secondary-index entry must both survive (no dangling index entry, no data loss).
+ * P2: DELETE without a throw (normal commit) — the row and index entry must both be gone
+ *     (regression guard: the fix must not break an ordinary committed delete).
+ *
+ * Reproduction:
+ *   npm run test:integration -- "integrationTests/resources/audit-false-delete-rollback.test.ts"
+ */
+import { suite, test, before, after } from 'node:test';
+import { ok, strictEqual } from 'node:assert';
+import { resolve } from 'node:path';
+import { setTimeout as sleep } from 'node:timers/promises';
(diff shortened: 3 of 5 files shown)


integrationTests/resources/audit-false-delete-rollback/schema.graphql
```diff
@@ -0,0 +1,12 @@
...
+#
+# `_writeDelete()`'s non-audit branch (resources/Table.ts) removes the base row via
+# removeEntry(). Without threading the enclosing transaction into that call, the removal
+# commits standalone against the raw store and survives a transaction abort, leaving a
+# dangling secondary-index entry behind. `audit: false` is required to reach this branch
+# (the audit:true branch goes through updateRecord(), which already threads {transaction}).
+type Item @table(audit: false) @export {
+	id: ID @primaryKey
+	bucket: String @indexed
+	payload: String
```
cb1kenobi, line 11: **Nit: blob-orphan effect described in the PR isn't exercised by a test.**

The PR body calls out that, without the fix, the aborted delete also unlinked the record's blob (`removeEntry`'s blob-unlink gate keyed off the standalone removal's own commit). The P1/P2 probes cover base-row + secondary-index atomicity, but the `Item` type has no blob attribute, so the blob-survives-abort path — the second half of the described defect — goes untested.

Optional: add a `Bytes`/blob attribute to `Item` and assert the blob still resolves after the aborted delete, closing the loop on the exact symptom the fix targets. Not blocking — the primary row/index rollback is well covered.

—
_Generated by Barber AI_

resources/Table.ts
```diff
@@ -2714,7 +2714,7 @@ export function makeTable(options) {
 						);
 						if (!audit || isRocksDB) scheduleCleanup();
 					} else {
-						removeEntry(primaryStore, existingEntry);
+						removeEntry(primaryStore, existingEntry, transaction && { transaction });
```
cb1kenobi: FWIW, it's safe to pass an undefined/falsey `transaction` into rocksdb-js' `remove()`. However in lmdb-js, this maps to the 2nd arg which is the `ifVersionOrValue` and that's probably not ok to pass in an object.

cb1kenobi (approved):

resources/Table.ts
```diff
@@ -2714,7 +2714,7 @@ export function makeTable(options) {
 						);
 						if (!audit || isRocksDB) scheduleCleanup();
 					} else {
-						removeEntry(primaryStore, existingEntry);
+						removeEntry(primaryStore, existingEntry, isRocksDB && transaction ? { transaction } : undefined);
```
cb1kenobi, line 2718: **Nit: the `isRocksDB` gate here diverges from the sibling calls without an explanation.**

Correct change — `primaryStore.remove(key, arg)`'s second positional arg is `ifVersion` on the LMDB path (see the version-passing `removeEntry` callers elsewhere in this file), so handing it a `{transaction}` object was meaningful only for RocksDB; `undefined` restores the original LMDB behavior (LMDB deletes already roll back via the batched `LMDBTransaction.writes`).

The risk is readability: two lines up, `updateIndices(..., transaction && { transaction })` threads the txn unconditionally, and `updateRecord` does too. A future reader may "align" this line back to the sibling pattern and silently reintroduce the LMDB regression. A short comment would pin the invariant:

```suggestion
						// Only RocksDB's remove() takes an options object; on LMDB the 2nd arg is ifVersion, and its writes already join the batched txn.
						removeEntry(primaryStore, existingEntry, isRocksDB && transaction ? { transaction } : undefined);
```

Not blocking.

—
_Generated by Barber AI_

Issue #1453, opened 2026-06-23
Root cause: v4→v5 migration structure-id fork silently nulls records (HttpCache/Campaign/WarmupJob)


kriszyp:
## Summary

The v4→v5 in-place migration produces a **structure-id fork**: a worker's *in-memory* shared-structures dictionary and the *durable* structures buffer assign **different structures to the same id**. Records (bare structure-id references) then decode against the wrong structure → field-count mismatch → `Data read, but end of buffer not reached` / `Unexpected end of MessagePack data` → the record reads back as `null`. Confirmed by byte-level trace on the live Akamai v4→v5 stage node (nl-ams-1, 5.1.6).

This is the single underlying cause behind the symptoms tracked in #1441/#1445 and the fixes in #1442/#1446. Big credit to @ldt1996 for the original diagnosis and the per-table decomposition — this issue consolidates the root cause and reconciles the three PRs.

## Symptom & scope (live evidence)

- nl-ams-1 (first node upgraded, 03:40:05): **5,631** `end of buffer` errors, first at **03:41:37** (~90s after migration), still firing ~20h later — i.e. *persistent*, not a transient window.
- Broad, not HttpCache-only: the same mismatch hits **HttpCache**, **Campaign**, **WarmupJob** (any table with structures). Record counts diverge across nodes.
- se-sto-1 (second node, synced structures from an already-settled v5 peer) lost 0.

## Root cause (byte-level)

A live failing HttpCache record is a **bare `0x41` (structure id 1)** reference, encoded as `[id, headers, content]`:

```
decoded with DURABLE  (id1 = [id, headers, content]) => { id, headers:{…}, content:… }   ✓ correct
decoded with IN-MEMORY (id1 = [content-type, cache-control, …])  => garbage / over-read  ✗ the failure
```

- **Durable is canonical and correct** — the record decodes perfectly against it.
- The reading worker's **in-memory dict is forked** (it minted the same structures in a *different order* during the migration+ingest window, so id1 maps to a different structure).
- msgpackr only reloads structures on a **missing** id, never a **present-but-wrong** one — so a read-heavy worker never reloads and every read decodes against the stale fork.

The fork is amplified by **value-dependent structure proliferation**: Campaign's `ctr`/`cvr` sometimes fit float32 and sometimes need float64, `region` varies in string type, etc., so one logical schema mints many structure variants (the replication-decode error dumps 6–8). Different nodes/workers mint these in different orders → id divergence.

## Recoverability

[...]

PR #1870, opened 2026-07-20
test(upgrade): pin typed-struct plateau on legacy-origin in-place upgrade (QA-606)

kriszyp:
## What this pins

A qa-explorer regression anchor for the **§2 G(cross-major-upgrade) × A(width-heterogeneous)** interaction — the corner QA-478 explicitly left uncovered.

QA-478 proved a `harperdb@4.7.34 → current-main` in-place upgrade migrates width-drifting records *decode-clean*, but it never drove **new** writes against the upgraded table with `randomAccessFields: true` active — so it never exercised the unbounded typed-struct **minting** path (the #1453-class OOM). This spec closes that gap.

**Controlled A/B on one instance**, identical width-heterogeneous generator + cadence on both arms:
- **upgraded** — `WidthDrift`, the v4.7.34-created, in-place-migrated dataset (a table with no persisted `randomAccessFields` directive, so it follows the global default on open).
- **fresh control** — `FreshControl`, created by current main with an explicit `randomAccessFields: true`.

**Verdict (green):** both arms plateau at `typedStructs = 38` (cap 256), upgraded/fresh ratio **1.00x** — the OOM class does **not** reproduce on a legacy-origin table; the upgraded arm tracks the fresh-install control.

## Provenance & gating
- Promoted from qa-explorer **QA-606 / candidate P-395**.
- Cold-gated **GREEN on main `56a2bace9`** — 3/3 runs (both arms `tailFlat=true`, legacy precondition fired, i.e. *not* a silent skip). oxlint + prettier clean.

## ⚠️ CI note — legacy provisioning required
This spec **skips** (like its sibling `integrationTests/upgrade/4.x-upgrade.test.ts`) unless `HARPER_LEGACY_VERSION_PATH` points at an installed v4.x `harperdb` package. It rides the *same* env var and skip-guard the existing 4.x upgrade suite already uses — no new mechanism — but CI must provision a v4.x `harperdb` install for it to actually run rather than skip.

### DIFF
--- a/integrationTests/upgrade/qa606-upgrade-structgrowth/resources.js
+++ b/integrationTests/upgrade/qa606-upgrade-structgrowth/resources.js
@@ -93,7 +93,7 @@ export class StructReport606 extends Resource {
 			typedStructs: enc && Array.isArray(enc.typedStructs) ? enc.typedStructs.length : null,
 			classicStructures: enc && Array.isArray(enc.structures) ? enc.structures.length : null,
 			maxOwnStructures: enc ? (enc.maxOwnStructures ?? null) : null,
-			storeCtor: store && store.constructor && store.constructor.name,
+			storeCtor: store?.constructor?.name,
 		};
 	}
 }
--- a/integrationTests/upgrade/qa606-upgrade-structgrowth/resources.js
+++ b/integrationTests/upgrade/qa606-upgrade-structgrowth/resources.js
@@ -1,0 +1,139 @@
+// QA-606 — legacy-v4-origin (upgraded) dataset vs fresh-install control: does the encoder's
+// typedStructs count PLATEAU under a width-heterogeneous NEW-write workload once
+// randomAccessFields is enabled, the way a fresh-install table does (QA-598), or does a
+// legacy-origin dict grow UNBOUNDED (the #1453-class escape QA-478 never exercised because it
+// only measured decode, never drove new writes with randomAccessFields:true active)?
+//
+// This component is dropped into the dataRootDir's components/ folder AFTER the legacy
+// (v4.7.34) instance is stopped and BEFORE current-main is first started on the same
+// dataRootDir — mirrors QA-478's qa478-probe component. It never runs under the legacy binary.
+//
+// GET /StructReport606/?table=WidthDrift|FreshControl — per-table randomAccess-mode +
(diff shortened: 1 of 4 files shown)


integrationTests/upgrade/qa606-upgrade-structgrowth/resources.js
```diff
@@ -0,0 +1,139 @@
...
+		const { store, enc } = findEncoder(table);
+		return {
+			pid: process.pid,
+			heapUsedMB: Math.round((process.memoryUsage().heapUsed / 1024 / 1024) * 10) / 10,
+			table: name,
+			randomAccessStructure: store ? (store.randomAccessStructure ?? null) : null,
+			typedStructs: enc && Array.isArray(enc.typedStructs) ? enc.typedStructs.length : null,
+			classicStructures: enc && Array.isArray(enc.structures) ? enc.structures.length : null,
+			maxOwnStructures: enc ? (enc.maxOwnStructures ?? null) : null,
+			storeCtor: store && store.constructor && store.constructor.name,
```
cb1kenobi: nit: could be simplified:
```suggestion
			storeCtor: store?.constructor?.name,
```

cb1kenobi (approved):
I'm not sure I understand what this pull request is doing.

Ethan-Arrowood (approved):
LGTM. Solid regression anchor — it pins the real invariant (a legacy-origin, in-place-upgraded table plateaus its typed-struct count instead of the #1453 OOM growth under randomAccessFields writes), and it's genuinely CI-exercised rather than dark: the integration workflow installs harperdb@4 and sets HARPER_LEGACY_VERSION_PATH, and the fail-loud preconditions (migration ran, decode-clean, structure flipped) stop it from silent-skipping to a false green. Green across Node 22/24/26.

Two non-blocking notes: worth folding in the fetch→client.reqRest consistency cleanup, and keep an eye on the added CI wall-clock (legacy boot + migrate + 2 restarts + 36k inserts, ~600s timeout) as the upgrade suite grows.

kriszyp:
> I'm not sure I understand what this pull request is doing.

This is test for the situation that we encountered with the customer (of "generation digital", to subtly reference) that was hitting OOM errors (while you were on call) due to the growth of typed structures, which should be limited.

PR #1876, opened 2026-07-20
feat(cli): token env vars for CI/CD auth

dawsontoth:
## Summary

Two small additions that give a clean local-login → CI/CD path so Harper credentials don't have to live in the wild:

1. **`harper login --for-ci` prints a pipeable CI/CD credential block.** After a successful login it writes the target and the long-lived refresh token to **stdout**, in dotenv format, so it never has to appear on screen or in shell history:

   ```sh
   harper login --for-ci | gh secret set --env-file -   # sets both repo secrets in one shot
   harper login --for-ci | pbcopy                       # or paste them in by hand
   ```

   ```
   HARPER_CLI_TARGET=https://my-instance:9925/
   HARPER_CLI_REFRESH_TOKEN=eyJ...
   ```

   Without the flag, `harper login` behaves exactly as before and prints no credentials.

2. **The CLI's Bearer-token path reads tokens from env vars.** `harper deploy` and other remote operations now honor `HARPER_CLI_OPERATION_TOKEN` / `HARPER_CLI_REFRESH_TOKEN` (with `CLI_TARGET_*` aliases), so a runner can authenticate without a stored password or a prior `harper login`.

## Stream discipline (why the flag, not a prompt)

An earlier revision asked interactively ("Print environment variables so CI/CD can deploy without your password?"). A flag is easier to guide people to, and — more importantly — it makes the output *composable*. For a pipe to actually work, the streams have to split cleanly:

- **stdout** carries only the two `KEY=value` lines, dotenv-formatted and ANSI-free, so `gh secret set --env-file -` consumes it verbatim. The explanatory comments moved to **stderr** rather than riding along as `#` lines, since every consumer of this block reads it as data.
- **stderr** carries everything a human reads — banner, status, prompts. inquirer writes to stdout by default, so `--for-ci` hands it an stderr prompt module; without that the password prompt would be invisible, swallowed by the pipe.
- **A cluster that returns no refresh token is a loud failure** (exit 1) rather than a silent empty block — which would otherwise pipe into a secret store and "succeed" at storing nothing.

`--for-ci` is filtered out of argv before the positional target/username are read, so it can appear anywhere in the command.

## Behavior

- **Precedence:** explicit username → Basic (unchanged); else **env-var tokens** (override the credentials file); else `~/.harperdb/credentials.json` (unchanged).
- **Refresh-token-only works:** if no operation token is supplied (or it's expired), the CLI mints a fresh one from the refresh token — so CI needs only `HARPER_CLI_TARGET` + `HARPER_CLI_REFRESH_TOKEN` as durable secrets.
- **No file writes for env tokens:** a token refreshed from env vars is used in-memory only; file-based credentials still persist their refreshed token exactly as before.

## Tests

- `unitTests/bin/login.test.js` — four cases covering *which stream each byte lands on*, since that's the feature: stdout is exactly the two lines (asserted as an exact array, so any stray output fails), stderr never contains the token, the default path prints nothing, and a missing refresh token exits 1.
- `unitTests/bin/cliOperations.test.js` — env token overrides file creds; refresh-token-only minting; expired-token refresh without persisting; `CLI_TARGET_*` alias.
- 63 passing across both suites; `oxlint` and `prettier --check` clean.
- Verified end-to-end against a mock cluster: stdout parses as dotenv with exactly the two keys and zero ANSI escapes, and stderr contains **0** occurrences of the token.

## Rebase note

Rebased onto `main` (v5.2.0-beta.3). Main had extracted the inline token-refresh logic into a shared `refreshExpiredOperationToken` helper; rather than re-inlining this branch's copy, I taught that helper the two behaviors this PR needs (mint when the operation token is absent, not just expired; `persistKey: string | null` so env-var tokens skip `saveCredentials`). See the [rebase comment](https://github.com/HarperFast/harper/pull/1876#issuecomment-5121838602) for detail.

## Review round (`6f91873`)

Three things from review, each fixed at the source rather than at the symptom:

**A resolved target is an identity, not a credential.** `normalizeTarget` now strips userinfo, so the one canonical target is what gets stored as the credentials-file key, echoed in status text, written to `.env`, and emitted by `--for-ci`. Previously only the emitted line was sanitized, leaving the password in the other three. Callers that authenticate with userinfo read it off the *raw* target via the new `extractTargetCredentials` (`cliOperations`, `agentCli`).

This also settles the port: the default-port heuristic looked for a `:` after the scheme, which the `:` in `user:password` satisfied — so `https://admin:pw@host` was contacted on 443 while being emitted as `:9925`. The check now runs against the authority with userinfo removed, so contacted and emitted agree, and there's no second normalization to disagree with the first. Saved logins keyed by a userinfo URL need a re-login; those entries were storing a plaintext password as a JSON key.

**One env token namespace, selected as a unit.** `HARPER_CLI_OPERATION_TOKEN` and `HARPER_CLI_REFRESH_TOKEN` were resolved independently, so an operation token from one user could pair with a refresh token from another — running as the first identity until its operation token expired, then silently continuing as the second. A namespace now owns both halves, matching how `login.ts` already selects username/password. A selected-but-blank namespace warns and falls back to the credentials file rather than letting the other namespace win by being non-empty.

**Refresh-token rotation is explicit.** Harper stores one refresh-token hash per user — `createTokens` overwrites `hdb_user.refresh_token` and refresh validation accepts only that hash — so minting a CI credential revokes whatever that user already held, invisibly, until the displaced holder's next refresh 401s. `--for-ci` now states this before it prompts, labels the username prompt as a dedicated CI user, requires an interactive confirmation naming that user (skipped when stdin isn't a TTY, so runners are unaffected; `| gh secret set` still hits it since only stdout is piped), and repeats the warning beside the credential block. `harper help` documents it. The client can't verify that a user is *dedicated* to CI — the real fix is server-side, filed as [#2018 — Auth: support multiple named, independently revocable refresh credentials per user](https://github.com/HarperFast/harper/issues/2018).

One adjacent fix while in that function (`90d86a5`): `normalizeTarget` wasn't idempotent for an explicitly written default port. `URL` serializes `:443` away, so `https://host:443` came back port-less, and the second pass — login normalizes, then cliOperations normalizes again — read that as "no port given" and appended 9925. `harper login https://host:443` was therefore talking to `:9925`. Pre-existing, but the output is also the credentials-file key, so idempotence is something every caller already assumes.

Tests added: target userinfo is absent from stdout, stderr, `credentials.json` and `.env` (and the stored key matches the emitted target); `normalizeTarget` strips credentials, still applies the default port past a `user:password` colon, preserves an explicitly written default port, and is idempotent across a set of representative targets; the two token namespaces never mix, and a blank one doesn't fall through; the rotation warning names the user without disturbing the pipe; the confirmation names the user, proceeds on yes, and mints nothing on no.

- `HarperFast/create-harper#118` — scaffolds this flow; its deploy workflow consumes `HARPER_CLI_TARGET` + `HARPER_CLI_REFRESH_TOKEN` from this PR
- `HarperFast/documentation#599` — docs for the two-phase work

### DIFF
--- a/bin/cliOperations.ts
+++ b/bin/cliOperations.ts
@@ -450,12 +450,20 @@ async function cliOperations(req: any, skipResponseLog = false) {
 		const transportCredentials = target ? resolveTransportCredentials(req, urlCredentials) : undefined;
 		if (transportCredentials) {
 			options.headers.Authorization = basicAuthHeader(transportCredentials.username, transportCredentials.password);
-		} else {
-			// Bearer-token auth. Env-var tokens (for CI/CD — see `harper login`) take precedence
-			// over the stored ~/.harperdb/credentials.json entry: they're an explicit per-invocation
-			// override that needs no prior `harper login` on the runner. A token refreshed from env
-			// vars is used in-memory only (there's no file to write back to); a token refreshed from
-			// the credentials file is persisted as before.
+		} else if (target) {
+			// Bearer-token auth, for remote targets ONLY. A local operation goes over the domain
+			// socket, which the server trusts via `bypassLocalAuth` — but that bypass is an
+			// `else if` on "no Authorization header present" (security/auth.ts), so attaching a
+			// Bearer token to a local request opts out of the trust and gets validated instead,
+			// 401ing on a token minted for some other cluster. Since these env vars are meant to
+			// persist across a whole CI job (or a developer's shell), an ungated read here would
+			// break every local `harper` command run in that environment.
+			//
+			// Env-var tokens (for CI/CD — see `harper login --for-ci`) take precedence over the
+			// stored ~/.harperdb/credentials.json entry: they're an explicit per-invocation override
+			// that needs no prior `harper login` on the runner. A token refreshed from env vars is
+			// used in-memory only (there's no file to write back to); a token refreshed from the
+			// credentials file is persisted as before.
 			const envOperationToken = (
 				process.env.HARPER_CLI_OPERATION_TOKEN || process.env.CLI_TARGET_OPERATION_TOKEN
 			)?.trim();
@@ -465,7 +473,7 @@ async function cliOperations(req: any, skipResponseLog = false) {
 			let persistKey: string | null = null; // non-null => persist a refreshed operation token back to the file
 			if (envOperationToken || envRefreshToken) {
 				tokens = { operation_token: envOperationToken, refresh_token: envRefreshToken };
-			} else if (target && allCredentials?.targets) {
+			} else if (allCredentials?.targets) {
 				persistKey = target.resolvedTarget;
 				tokens = allCredentials.targets[persistKey] ?? null;
 			}
--- a/bin/cliOperations.ts
+++ b/bin/cliOperations.ts
@@ -1,6 +1,6 @@
 'use strict';
 
-import { loadCredentials, saveCredentials, normalizeTarget } from './cliCredentials.ts';
+import { loadCredentials, saveCredentials, normalizeTarget, extractTargetCredentials } from './cliCredentials.ts';
 import { isJWTExpired } from '../security/tokenAuthentication.ts';
 import * as envMgr from '../utility/environment/environmentManager.ts';
 envMgr.initSync();
@@ -393,9 +393,12 @@ async function cliOperations(req: any, skipResponseLog = false) {
 	require('dotenv').config();
 
 	const allCredentials = loadCredentials();
-	req.target = normalizeTarget(resolveTarget(req, allCredentials));
+	const rawTarget = resolveTarget(req, allCredentials);
+	// Userinfo is a transport credential, and `normalizeTarget` strips it so the resolved target can
+	// be logged, stored and emitted freely — so read it off the raw value first.
+	const urlCredentials = extractTargetCredentials(rawTarget);
+	req.target = normalizeTarget(rawTarget);
 	let target;
-	let urlCredentials = { username: '', password: '' };
 	if (req.target) {
 		let parsedTarget;
 		try {
@@ -408,7 +411,6 @@ async function cliOperations(req: any, skipResponseLog = false) {
 			}
 		}
 		const resolvedTarget = req.target;
-		urlCredentials = { username: parsedTarget.username, password: parsedTarget.password };
 		target = {
 			protocol: parsedTarget.protocol,
 			hostname: parsedTarget.hostname,
@@ -464,10 +466,25 @@ async function cliOperations(req: any, skipResponseLog = false) {
 			// that needs no prior `harper login` on the runner. A token refreshed from env vars is
 			// used in-memory only (there's no file to write back to); a token refreshed from the
 			// credentials file is persisted as before.
-			const envOperationToken = (
-				process.env.HARPER_CLI_OPERATION_TOKEN || process.env.CLI_TARGET_OPERATION_TOKEN
-			)?.trim();
-			const envRefreshToken = (process.env.HARPER_CLI_REFRESH_TOKEN || process.env.CLI_TARGET_REFRESH_TOKEN)?.trim();
+			//
+			// Whichever namespace supplies a token owns both halves. Resolving them independently
+			// would let `HARPER_CLI_OPERATION_TOKEN` from one user pair with
+			// `CLI_TARGET_REFRESH_TOKEN` from another: commands would run as the first identity
+			// until its operation token expired, then silently continue as the second. `login.ts`
+			// selects its username/password namespace as a unit for exactly this reason.
+			const tokenPrefix = ['HARPER_CLI', 'CLI_TARGET'].find(
+				(prefix) =>
+					process.env[`${prefix}_OPERATION_TOKEN`] !== undefined || process.env[`${prefix}_REFRESH_TOKEN`] !== undefined
+			);
+			const envOperationToken = tokenPrefix ? process.env[`${tokenPrefix}_OPERATION_TOKEN`]?.trim() : undefined;
+			const envRefreshToken = tokenPrefix ? process.env[`${tokenPrefix}_REFRESH_TOKEN`]?.trim() : undefined;
+			// A namespace that is set but blank is a broken CI secret, not a request to fall back to
+			// whatever the developer last logged in as — say so rather than switching identity silently.
+			if (tokenPrefix && !envOperationToken && !envRefreshToken) {
+				console.error(
+					`Ignoring empty ${tokenPrefix}_OPERATION_TOKEN/${tokenPrefix}_REFRESH_TOKEN; falling back to saved login credentials.`
+				);
+			}
 
 			let tokens: { operation_token?: string; refresh_token?: string } | null = null;
 			let persistKey: string | null = null; // non-null => persist a refreshed operation token back to the file
--- a/bin/cliOperations.ts
+++ b/bin/cliOperations.ts
@@ -328,19 +328,28 @@ function resolveTarget(req, allCredentials) {
 }
 
 /**
- * If `tokens.operation_token` is expired and a `refresh_token` is on hand, refreshes it via
- * `refresh_operation_token`, persisting the new token to the credentials file and updating
- * `tokens.operation_token` in place. Shared by `cliOperations` and any other CLI transport
- * (e.g. `harper agent`) authenticating with stored `harper login` tokens, so refresh behavior
- * stays in one place instead of drifting between callers.
+ * Ensures `tokens.operation_token` is usable, minting a fresh one via `refresh_operation_token`
+ * when it is expired — or absent entirely, which is the refresh-token-only shape a CI/CD runner
+ * gets from `HARPER_CLI_REFRESH_TOKEN`. Updates `tokens.operation_token` in place.
+ *
+ * `persistKey` is the credentials-file key to write the refreshed token back to; pass `null` for
+ * tokens sourced from env vars, which have no file entry and stay in memory for this invocation
+ * only. Shared by `cliOperations` and any other CLI transport (e.g. `harper agent`) authenticating
+ * with tokens, so refresh behavior stays in one place instead of drifting between callers.
  */
 async function refreshExpiredOperationToken(
 	options: any,
-	tokens: { operation_token: string; refresh_token: string },
-	lookupKey: string
+	tokens: { operation_token?: string; refresh_token?: string },
+	persistKey: string | null
 ): Promise<void> {
-	if (!tokens.refresh_token || !isJWTExpired(tokens.operation_token)) return;
-	console.error('Operation token expired, attempting to refresh...');
+	if (!tokens.refresh_token) return;
+	// Short-circuited so `isJWTExpired` never runs on an absent token.
+	if (tokens.operation_token && !isJWTExpired(tokens.operation_token)) return;
+	console.error(
+		tokens.operation_token
+			? 'Operation token expired, attempting to refresh...'
+			: 'Minting an operation token from the refresh token...'
+	);
 	try {
 		// Always use the standard operation timeout for this call, even when the caller's
 		// own options carry the longer SSE timeout (e.g. a deploy_component retry) — the
@@ -354,10 +363,13 @@ async function refreshExpiredOperationToken(
 			const refreshData = JSON.parse(refreshResponse.body);
 			if (refreshData.operation_token) {
 				tokens.operation_token = refreshData.operation_token;
-				saveCredentials(lookupKey, {
-					operation_token: tokens.operation_token,
-					refresh_token: tokens.refresh_token,
-				});
+				// Only file-based credentials are persisted; env-var tokens have no file entry.
+				if (persistKey) {
+					saveCredentials(persistKey, {
+						operation_token: tokens.operation_token,
+						refresh_token: tokens.refresh_token,
+					});
+				}
 				console.error('Operation token refreshed successfully.');
 			}
 		} else if (refreshResponse.statusCode === 401) {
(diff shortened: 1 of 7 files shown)


dawsontoth:
Rebased onto `main` (v5.2.0-beta.3) — conflicts resolved.

**What conflicted:** main refactored the CLI's Bearer-token path since this branch was cut — the inline refresh logic moved into a shared `refreshExpiredOperationToken(options, tokens, lookupKey)` helper (so `harper agent` and `cliOperations` share it), and Basic-auth resolution moved into `resolveTransportCredentials`.

**How I resolved it:** rather than re-inlining this branch's copy of the refresh logic, I taught main's helper the two behaviors this PR needs, so there's still exactly one implementation:

- **Refresh-token-only minting** — it previously bailed unless the operation token was expired; it now also mints when there's no operation token at all (the CI shape: `HARPER_CLI_REFRESH_TOKEN` alone). The `isJWTExpired` call is short-circuited so it never runs on an absent token.
- **`persistKey: string | null`** (was `lookupKey: string`) — `null` means the tokens came from env vars, so the refreshed token is used in-memory only and `saveCredentials` is skipped. File-based credentials persist exactly as before.

The call site keeps main's structure (`resolveTransportCredentials` → Basic, else Bearer) with the env-var branch layered in ahead of the credentials-file lookup, and the `else if (allCredentials)` guard relaxed to `else` so env tokens work on a runner with no credentials file.

**Also:** pinned `process.stdin.isTTY = false` in main's new `credential sources` describe block. Those tests are about which credentials get sent, and without the pin they'd non-deterministically reach this PR's new TTY-gated CI/CD confirm prompt depending on whether the runner has a TTY.

**Verified:** 61 passing across `unitTests/bin/login.test.js` + `unitTests/bin/cliOperations.test.js` — including main's "Leg 2: should refresh expired token **and save it**", which confirms the file-persistence path is intact, alongside this PR's four env-token cases. `oxlint` and `prettier --check` clean.

(Unrelated: `npm run build` on main currently errors on a missing `@harperfast/skills` module — dependency drift, not from this branch.)

bin/cliOperations.ts
```diff
@@ -430,25 +442,39 @@ async function cliOperations(req: any, skipResponseLog = false) {
...
+			// vars is used in-memory only (there's no file to write back to); a token refreshed from
+			// the credentials file is persisted as before.
+			const envOperationToken = (
+				process.env.HARPER_CLI_OPERATION_TOKEN || process.env.CLI_TARGET_OPERATION_TOKEN
+			)?.trim();
+			const envRefreshToken = (process.env.HARPER_CLI_REFRESH_TOKEN || process.env.CLI_TARGET_REFRESH_TOKEN)?.trim();
+
+			let tokens: { operation_token?: string; refresh_token?: string } | null = null;
+			let persistKey: string | null = null; // non-null => persist a refreshed operation token back to the file
+			if (envOperationToken || envRefreshToken) {
```
cb1kenobi, line 491: **Low: env-var token branch isn't gated on a resolved `target`, unlike the file-token branch below**

The file-based branch (`else if (target && allCredentials?.targets)`) only populates `tokens` when a `target` is present, so local domain-socket operations never attach a Bearer header. This env-var branch runs regardless of `target`: if `HARPER_CLI_REFRESH_TOKEN`/`HARPER_CLI_OPERATION_TOKEN` is exported but no target is resolvable (no `HARPER_CLI_TARGET`/`CLI_TARGET`/`last_target`), a *local* operation will now attach the Bearer header and can even fire a `refresh_operation_token` over the local socket, minting from a refresh token meant for a remote cluster. Env tokens are a CI/CD-for-remote feature, so gating this branch on `target` keeps parity with the file-token branch and the documented intent.

Suggested fix:
```suggestion
			if ((envOperationToken || envRefreshToken) && target) {
```

—
_Generated by Barber AI_

bin/login.ts
```diff
@@ -130,16 +145,59 @@ export async function login(targetArg: string, usernameArg: string): Promise<voi
...
+	}
+
+	process.stderr.write(
+		chalk.gray(
+			'\n# Store these as secrets in your CI/CD provider (e.g. GitHub Actions repository secrets)\n' +
+				'# and expose them as environment variables to your `harper deploy` step. The refresh\n' +
+				'# token is long-lived; the CLI mints a fresh operation token from it on each run.\n'
+		)
+	);
+	process.stdout.write(`HARPER_CLI_TARGET=${target}\n`);
```
cb1kenobi, line 231: **Low: a target URL with embedded userinfo leaks the password into the CI secret block**

`normalizeTarget` preserves URL userinfo, and `resolvedTarget` can carry it (e.g. `harper login https://admin:secret@host --for-ci`). This line writes it verbatim to stdout, so the admin password lands in the `HARPER_CLI_TARGET` secret and on the terminal — the exact on-screen credential exposure `--for-ci` is designed to avoid. The `Connecting to ...` log already strips target passwords via `redactTargetUrl`; the emitted target should be sanitized the same way.

Suggested fix: before emitting, parse `target` through `URL` and clear `url.username`/`url.password` (or reuse a shared helper) so only the host is written to the credential block.

—
_Generated by Barber AI_

bin/cliOperations.ts
```diff
@@ -430,25 +442,39 @@ async function cliOperations(req: any, skipResponseLog = false) {
...
+				tokens = { operation_token: envOperationToken, refresh_token: envRefreshToken };
+			} else if (target && allCredentials?.targets) {
+				persistKey = target.resolvedTarget;
+				tokens = allCredentials.targets[persistKey] ?? null;
 			}
```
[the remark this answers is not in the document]
dawsontoth: Fixed in `0f0806b`. Confirmed the analysis — `bypassLocalAuth` is an `else if` on "no Authorization header present", so attaching a Bearer token to the domain socket opts out of the local trust entirely and validates a token minted for a different cluster.

The Bearer branch is now gated on `target` (`} else if (target) {`), matching the file-credentials branch beside it, and the now-redundant `target &&` inside dropped. Local operations carry no Authorization header again.

Added a regression test (`does not attach a Bearer token to a local (no-target) operation`) that sets both env vars, runs a targetless operation over a mocked domain socket, and asserts no `Authorization` header — plus that only one request fires, since the unguarded path would also have sent `refresh_operation_token` over the local socket.

bin/cliOperations.ts
dawsontoth: Fixed in `0f0806b` — same finding as the inline bot review above, and you're both right. Gated the whole Bearer branch on `target` rather than just the env-var condition, so it matches the file-credentials branch structurally instead of repeating the guard.

Your point about `refresh_operation_token` firing over the local socket is the sharper half of it, and the regression test asserts the request count for exactly that reason — checking only the missing header would have let a stray refresh call through.

bin/login.ts
dawsontoth: Fixed in `0f0806b`. Confirmed: `normalizeTarget` round-trips through `URL.toString()`, which preserves userinfo, so `harper login https://admin:hunter2@host --for-ci` did write the password into the emitted secret.

Credentials are now dropped rather than masked — a `***` placeholder would make the value unusable as a target — via a `targetWithoutCredentials` helper, with a regression test asserting neither the password nor the username survives.

One extra wrinkle worth recording: I re-normalize the cleaned URL rather than just emitting it. `normalizeTarget` only appends the default port when it finds no `:` after the scheme, and the `:` in `user:password` defeats that heuristic — so the userinfo case was *also* emitting a port-less target that CI would resolve to 443 instead of 9925. Running `normalizeTarget` again on the credential-free URL fixes both at once.

I deliberately did **not** fix the heuristic inside `normalizeTarget` itself: its output is the key for `~/.harperdb/credentials.json`, so changing it would silently invalidate saved logins for anyone who logged in with a URL containing userinfo. Happy to do that as a separate change if you'd prefer it fixed at the source.

bin/login.ts
```diff
@@ -130,16 +145,84 @@ export async function login(targetArg: string, usernameArg: string): Promise<voi
...
+
+	process.stderr.write(
+		chalk.gray(
+			'\n# Store these as secrets in your CI/CD provider (e.g. GitHub Actions repository secrets)\n' +
+				'# and expose them as environment variables to your `harper deploy` step. The refresh\n' +
+				'# token is long-lived; the CLI mints a fresh operation token from it on each run.\n'
+		)
+	);
+	process.stdout.write(`HARPER_CLI_TARGET=${targetWithoutCredentials(target)}\n`);
+	process.stdout.write(`HARPER_CLI_REFRESH_TOKEN=${refreshToken}\n`);
```
kriszyp, line 232: `--for-ci` presents this as a durable CI credential, but Harper currently stores exactly one refresh-token hash on each user: `createTokens()` overwrites `hdb_user.refresh_token`, and refresh validation accepts only that hash. A concrete failure is: save token A from this command in GitHub, then run an ordinary `harper login` (or configure another runner) as the same user; token B replaces the stored hash, and the next workflow attempt to refresh with A gets a 401. Generating this credential likewise invalidates the previous refresh credential for that user. Please address the underlying credential model before exposing this as stable CI auth—ideally named/multiple independently revocable refresh or PAT credentials. If that server work is out of scope, at minimum require a dedicated CI identity and make the rotation behavior explicit in the command, help, and tests.

bin/cliOperations.ts
```diff
@@ -430,25 +442,47 @@ async function cliOperations(req: any, skipResponseLog = false) {
...
+			// credentials file is persisted as before.
+			const envOperationToken = (
+				process.env.HARPER_CLI_OPERATION_TOKEN || process.env.CLI_TARGET_OPERATION_TOKEN
+			)?.trim();
+			const envRefreshToken = (process.env.HARPER_CLI_REFRESH_TOKEN || process.env.CLI_TARGET_REFRESH_TOKEN)?.trim();
+
+			let tokens: { operation_token?: string; refresh_token?: string } | null = null;
+			let persistKey: string | null = null; // non-null => persist a refreshed operation token back to the file
+			if (envOperationToken || envRefreshToken) {
+				tokens = { operation_token: envOperationToken, refresh_token: envRefreshToken };
```
kriszyp, line 492: These two namespaces are resolved independently, so the CLI can construct `{ operation_token: A, refresh_token: B }`. For example, with `HARPER_CLI_OPERATION_TOKEN` belonging to user A and `CLI_TARGET_REFRESH_TOKEN` belonging to user B, commands run as A until the operation token expires, then refresh with B and silently run as B. `login.ts` deliberately selects one namespace as a unit to avoid this same split-credential problem for username/password. Please select one token namespace as a unit (and handle incomplete/blank values explicitly), or simplify the public API to the single refresh-token variable that `--for-ci` emits.

bin/login.ts
```diff
@@ -130,16 +145,84 @@ export async function login(targetArg: string, usernameArg: string): Promise<voi
...
+	}
+
+	process.stderr.write(
+		chalk.gray(
+			'\n# Store these as secrets in your CI/CD provider (e.g. GitHub Actions repository secrets)\n' +
+				'# and expose them as environment variables to your `harper deploy` step. The refresh\n' +
+				'# token is long-lived; the CLI mints a fresh operation token from it on each run.\n'
+		)
+	);
+	process.stdout.write(`HARPER_CLI_TARGET=${targetWithoutCredentials(target)}\n`);
```
kriszyp: `targetWithoutCredentials()` sanitizes only the stdout value. With the test input here, `resolvedTarget` still contains `admin:hunter2`: the success message sends it to stderr, `saveCredentials()` stores it as a credentials-file key, and the `.env` update can persist it. Thus `harper login https://admin:hunter2@example.com --for-ci` still prints the admin password on the terminal and may write it to disk even though this test checks only stdout. Please derive one credential-free resolved target and use it for credential storage, `.env`, status text, and CI output while retaining parsed userinfo only for transport authentication; assert that stderr and persisted files contain no userinfo. Doing this once also avoids changing the target after login—currently a userinfo URL with no explicit port is contacted on 443 but emitted as `:9925` after the second normalization.

kriszyp (approved):
This is great. I am skeptical the findings are "major". If you don't get to it, I might finish this PR while you are out.

bin/login.ts
kriszyp: Fixed in `6f91873` — and you're right that sanitizing at the emit site was the wrong altitude. The invariant is that a *resolved target is an identity, not a credential*, so the strip now happens once, in `normalizeTarget`:

```ts
url.username = '';
url.password = '';
```

Everything downstream of it inherits that: the credentials-file key, the `Connecting to ...` line, the `.env` append, the success message, and the `--for-ci` block. `targetWithoutCredentials` is gone. Userinfo is read off the *raw* target instead, via a new `extractTargetCredentials`, and used only for transport auth — `cliOperations` and `agentCli` both take it that way now (`agentCli` was reading `url.username` off the normalized URL, which the strip would otherwise have silently emptied).

The port discrepancy falls out of the same change rather than needing its own patch. The default-port heuristic was `!normalized.includes(':', indexOf('://') + 3)` — the `:` in `user:password` satisfied it, so no port was appended. It now tests the authority with userinfo removed:

```ts
const authority = normalized.slice(normalized.indexOf('://') + 3).split(/[/?#]/)[0];
const hasExplicitPort = !!url.port || authority.slice(authority.lastIndexOf('@') + 1).includes(':');
```

so `https://admin:hunter2@host` is both contacted on and emitted as `:9925`, and there's no longer a second normalization to disagree with the first. `!!url.port` covers the case the string check exists for (`URL` drops an explicitly written default port, so `https://host:443` has an empty `url.port`) — there's a test pinning that.

On the compatibility point I hedged on earlier: this does change the credentials-file key for anyone who logged in with a userinfo URL, so those saved logins need a re-login. That seems clearly right rather than a cost — those entries were storing a plaintext password as a JSON key.

Tests: the userinfo case now asserts all four destinations, not just stdout — stdout, stderr, `credentials.json`, and `.env` are each checked for the password *and* the username, plus that the stored key equals the emitted target so a later `harper deploy` finds it. On the `cliOperations` side, one test asserts the split directly: the request still carries `Basic admin:hunter2`, while `req.target`, `resolvedTarget`, and the connection log carry none of it, and the port is `9925`. The pre-existing "masks a password in the connection log" test became "keeps it out of the connection log" — there's nothing left to mask.

_🤖 Addressed by Claude Opus 5_

bin/cliOperations.ts
kriszyp: Fixed in `6f91873`. Confirmed the split you describe — the two `||` chains were resolved independently, so `HARPER_CLI_OPERATION_TOKEN=<user A>` with `CLI_TARGET_REFRESH_TOKEN=<user B>` produced exactly `{ operation_token: A, refresh_token: B }`, and the identity switch happened at an arbitrary moment mid-job (whenever A expired), which is about the worst time for it to be invisible.

Selecting one namespace as a unit, mirroring `login.ts`:

```ts
const tokenPrefix = ['HARPER_CLI', 'CLI_TARGET'].find(
    (prefix) =>
        process.env[`${prefix}_OPERATION_TOKEN`] !== undefined || process.env[`${prefix}_REFRESH_TOKEN`] !== undefined
);
```

I kept both variables rather than narrowing the API to the single refresh token `--for-ci` emits. The operation-token override is useful on its own (a short-lived token minted elsewhere, no refresh round-trip), and dropping it would only move the inconsistency — `login.ts` would still select a namespace as a unit for username/password while the token path had a different shape. This way the rule is the same one in both places.

On incomplete/blank values: a namespace is *selected* on `!== undefined`, so a set-but-empty variable claims the namespace instead of letting the other one win by being non-empty. That's deliberate — a blank value is a CI secret that failed to populate, and quietly falling through to `CLI_TARGET_*`, or to the developer's saved login, would run the job as the wrong identity at the point where the intent was clearest. It warns and falls back to the credentials file rather than hard-failing, matching how `resolveTransportCredentials` treats an incomplete env pair (`incompleteIsFatal: false`) — these vars persist across a whole shell session, so a fatal there would break every subsequent local command too.

Two tests: one sets `HARPER_CLI_OPERATION_TOKEN` + `CLI_TARGET_REFRESH_TOKEN` with the operation token expired, and asserts only the `test` operation fires (a mixed pair would have fired `refresh_operation_token` first); the other sets `HARPER_CLI_REFRESH_TOKEN=''` alongside a populated `CLI_TARGET_REFRESH_TOKEN` and asserts the file token is used, not the other namespace.

_🤖 Addressed by Claude Opus 5_

bin/login.ts
kriszyp: Confirmed, and the failure mode is exactly as you describe. `createTokens` writes a single hash:

```ts
const hashedToken = password.hash(refreshToken, password.HASH_FUNCTION.SHA256);
await update(new UpdateObject(SYSTEM_SCHEMA_NAME, SYSTEM_TABLE_NAMES.USER_TABLE_NAME, [
    { username: authObj.username, refresh_token: hashedToken },
]));
```

and `validateRefreshToken` accepts only that hash. So a user holds at most one live refresh credential, minting revokes the previous one, and the revocation is silent until the displaced holder's next refresh 401s.

I took the second path you offered: the credential model is server work and out of scope here, so this PR stays CLI-only and I've filed [#2018 — Auth: support multiple named, independently revocable refresh credentials per user](https://github.com/HarperFast/harper/issues/2018) for the real fix (N named credentials per user, issuing one never affecting the others, individually revocable, `--name` on `--for-ci` so the cluster can identify the holder). Happy to hold this PR until that lands if you'd rather not ship the client half first — say the word.

What `6f91873` does in the meantime, on the "at minimum" list:

**In the command.** `--for-ci` opens with the rotation semantics before it prompts for anything, so the choice of user is made with it in view:

> Harper stores one refresh token per user, so this login replaces whatever refresh token that user already holds — including one a CI runner may be using right now. Log in as a user created for this single consumer, not your personal account.

The username prompt reads `CI Username (a user dedicated to this consumer):`, and after the username resolves there's a confirmation naming it:

> `Mint a CI refresh token for 'ci-deploy'? This revokes any refresh token it already holds.` (default no)

Declining exits 1 having minted nothing. It's skipped when stdin isn't a TTY — a runner has already committed and there's nobody to ask — and `--for-ci | gh secret set --env-file -` still hits it, since only stdout is piped there. Then the credential block itself carries the warning on stderr, naming the user:

> `One refresh token per user: this replaced any refresh token 'ci-deploy' already held. Anything still using an older one — another runner, another machine, an earlier 'harper login' — will fail on its next refresh. Give each CI consumer its own user.`

**In the help.** `harper help` now states the single-refresh-token model under `login`, including that two consumers cannot share a user.

**In the tests.** Four: the stderr warning names the user and the two dotenv lines on stdout are still asserted as an exact array (the warning must not disturb the pipe); a default login prints no such warning; the confirmation names the user and proceeds on yes; declining exits 1 with stdout empty.

The one thing I want to be straight about: "require a dedicated CI identity" is not something the client can actually verify — nothing in the response distinguishes a CI user from a human's. What this does is refuse to rotate a user's only refresh token silently or by default. It's a guard rail, not an invariant, and it stays a workaround until #2018 makes the credential model correct.

_🤖 Addressed by Claude Opus 5_

PR #1883, opened 2026-07-21
ci: run a single Node version for integration tests on PR pushes

kriszyp:
## What

Trims integration-test CI cost on PR pushes without losing cross-version coverage:

- **PR pushes** now run the integration suite on a **single representative Node version (24)** instead of the full `22/24/26` matrix — dropping the `run-integration-tests` fan-out from 18 shards to 6.
- The **full `22/24/26` matrix still runs** on pushes to `main`/release branches, and on a **new nightly schedule** (`0 7 * * *`).
- `workflow_dispatch` behavior is unchanged (still respects the `node-version` input).

Windows/Bun/uWS variant jobs are untouched. Shard counts are unchanged (already 6-way).

## Why

Every PR push was paying for the full three-Node matrix, tripling integration compute and stretching the wall-clock that CI-monitoring sessions wait on. Cross-Node regressions are still caught on `main` and nightly; per-push we only need one representative version.

## Risk

Low. A Node-version-specific regression would surface on the next `main` push / nightly rather than on the PR push itself — an acceptable trade for the dev-loop speedup. The change is a conditional in the existing `generate-node-version-matrix` job.

— KrAIs (Claude Opus 4.8)

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches: [main, 'v[0-9]+.[0-9]+']
   pull_request:
+  schedule:
+    - cron: '0 7 * * *'
   workflow_dispatch:
     inputs:
       node-version:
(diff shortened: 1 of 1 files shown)


Ethan-Arrowood (approved):
LGTM. Workflow logic traced across all four triggers — PR pushes run Node 24 only, full 22/24/26 matrix retained on main + nightly, so coverage is deferred, not lost. Confirmed the integration jobs aren't required status checks, so the matrix trim won't strand PRs on a check that never reports.

Non-blocking follow-up: this same Node-matrix trim is duplicated in harper-pro#601 — worth extracting into a shared reusable workflow so the two repos don't drift.

PR #1884, opened 2026-07-21
test(resources): consolidate 3 promoted QA tests into one shared-instance suite

kriszyp:
## What

Merges three promoted `resources/` regression anchors that each booted their **own** Harper instance into a single suite that boots **one** shared instance:

- `relationship-parent-delete-cascade.test.ts` (QA-196)
- `relationship-transaction-atomicity.test.ts` (QA-162)
- `resource-status-contract.test.ts` (QA-195)

→ `relationship-and-status-contract.test.ts` — one outer `suite()` with a single `before`/`after`, three nested suites (one per QA-id) preserving every original `test()` verbatim, over a merged fixture. **3 Harper boots → 1.**

All three used vanilla-default config and their schemas were disjoint/identical, so **no namespacing was required**.

## Why

Each promoted test file pays a full Harper process boot — the dominant integration-test cost. Tests within a file share one instance, so co-locating compatible anchors removes redundant boots without losing coverage. This is the first, exemplar application of the instance-affinity policy just added to the qa-explorer promotion pipeline; it also creates a **host suite** future promotions can append into rather than minting new files.

## Verification

`npm run test:integration -- integrationTests/resources/relationship-and-status-contract.test.ts`:
```
tests 16 · pass 16 · fail 0 · duration_ms ~6526 (single boot)
```
Same assertions and documented findings as the originals (QA-195's 5 documented status-contract findings are unchanged pre-existing behavior, not new failures). oxlint + prettier clean on the new files.

> Note: `npm run build` surfaces a TS2345 in `resources/DatabaseTransaction.ts` that is **pre-existing on `main`** (confirmed via stash) and unrelated to this change; `tsc` still emits `dist`.

## Scope

First of a possible series. The larger `database/` groups (~9-scenario merge) are held pending a review of merged-file size vs. benefit; `harper-pro` cluster tests net zero and are intentionally excluded.

— KrAIs (Claude Opus 4.8)


Ethan-Arrowood (approved):
LGTM. Clean consolidation — I checked the two load-bearing risks and both hold: isolation is safe (every QA-196/QA-162 test clears at entry, the two relationship suites use disjoint orderIds, and QA-195 runs on a fully disjoint table/resource set), and no coverage was dropped (the test census matches exactly — 16 tests in, same titles out, LMDB-parity mode preserved). Collapsing 3 Harper boots to 1 is a genuine win on the dominant integration-test cost. Green across Node 22/24/26 + Bun + uWS + Windows.

PR #1896, opened 2026-07-22
fix: TTL eviction/delete no longer orphans secondary-index entries (F-149)

### LINKED ISSUE #1894: F-149: TTL/expiration eviction orphans 100% of secondary-index entries on RocksDB (audit-independent, unbounded index bl
From finding in dispatch task `qa-wave-2026072014`:

- F-149 (TTL/expiration eviction orphans 100% of secondary-index entries; HIGH) is confirmed RocksDB-specific and LMDB-clean, and remains a 3-anchor RED family (P-398 plain-@indexed, P-399 FK-target, +retracted P-133/P-255) that is source-traced (eviction batcher `updateIndices(…,{transaction})` bound to `primaryStore.store`, `Table.ts:5573/5589`) but still UNFILED

kriszyp:
Resolves #1894.

## Problem
Record removal — the background TTL/expiration eviction sweep **and** plain deletes — calls `updateIndices(id, existingRecord, null)` to clear that record's secondary-index entries. With `record === null`, the line `const value = record && (...)` resolved the *new* indexed value to `null` rather than "absent". Because `indexNulls` **defaults to `true`** for every `@indexed` attribute, `getIndexedValues(null, indexNulls)` returns `[null]`, so `updateIndices` removed the real `[value, id]` index entry and then immediately **re-added a `[null, id]` entry** — a dangling index entry (raw key `null`, value = the id of a now-deleted record).

On RocksDB an `@indexed` TTL table left **100% of evicted rows** dangling in the raw index (unbounded index bloat / phantom hits under a null-valued query). The classic `search_by_value` oracle can't observe it — it joins each index hit back through the primary record and drops the hit when the record is gone — which is why "phantom=0" read as clean for so long.

> **Note on the root cause vs the issue's trace.** The issue (F-149) traced this to the eviction batcher's transaction being bound to `primaryStore.store`. That trace is **incorrect** — normal deletes use the same `primaryStore.store`-bound raw transaction and persist fine, and the leak reproduces on the per-record `evict()` path too, single- *and* multi-worker. It's engine-agnostic, in shared `updateIndices()`, and — because `updateIndices` is shared — it also affects **plain deletes**, not just eviction.

## Fix
A removed record has no values to index, so resolve `value` to `undefined` (not `null`) when `record == null` → nothing is re-added after the real entry is removed. A record that is *present* but whose attribute is genuinely `null` is a different case (`record` is a truthy object) and still indexes under `null`, so the `indexNulls` feature is preserved.

```diff
-			const value = record && (resolver ? resolver(record) : record[key]);
+			const value = record == null ? undefined : resolver ? resolver(record) : record[key];
```

## Test
Adds `integrationTests/database/eviction-index-null-reindex.test.ts` (RocksDB) with a **direct raw-index-DBI oracle** — reads `t.indices[field].getRange()` with no join, so the `[null, id]` orphan is visible — plus a guard that a present record with a genuinely-null field stays indexed under `null`. Fails without the fix, passes with it.

## Verification
- P-398 (plain `@indexed`) and P-399 (FK-target) reproducers: index fully cleared, 0 dangling (was 750 / 241).
- Green: `eviction-secondary-index`, `indexed-update-churn`, `compound-index-conditions`, `delete-update-race`, `indexed-numeric-range`, `bigint-indexed-range`, vector-index (unit + integration), `longtxn-secondary-index`; LMDB `eviction` path unaffected.

## Cross-model review
Codex + Gemini + Harper domain pass. Core fix confirmed correct across all `updateIndices` callers and the HNSW custom-index path. One Gemini "blocker" on the adjacent `existingValue` line was **refuted** — that line's `null` result for a null `existingRecord` is intentional asymmetry (it's what makes removal clean up the `[null,id]` entry), it's unchanged by this PR, and the described skip never triggers on the real insert path. Test-file lint/format/flake findings were addressed.

### DIFF
--- a/integrationTests/database/eviction-index-null-reindex.test.ts
+++ b/integrationTests/database/eviction-index-null-reindex.test.ts
@@ -1,0 +1,140 @@
+/**
+ * harper#1894 (F-149) regression — TTL/expiration eviction must not orphan secondary-index entries.
+ *
+ * Removing a record (delete or the background TTL-eviction sweep) calls
+ * `updateIndices(id, existingRecord, null)` to clear that record's secondary-index entries. The bug:
+ * with `record === null`, `updateIndices` resolved the *new* indexed value to `null` rather than
+ * "absent", and for an `indexNulls` index — the DEFAULT for `@indexed` — it then RE-ADDED a
+ * `[null, id]` index entry immediately after removing the real `[value, id]` one. The result was a
+ * dangling index entry (raw key `null`, value = the id of a now-deleted record) for every evicted
+ * row: on RocksDB an `@indexed` TTL table left 100% of its rows dangling in the raw index.
+ *
+ * The classic search_by_value oracle can't observe this — it joins each index hit back through the
+ * primary record and drops the hit when the record is gone — so it reported "0 phantom" vacuously.
+ * This test reads the raw index DBI directly (RawIndex / PrimaryIds, no join) so a dangling entry is
+ * visible, and additionally asserts the legitimate indexNulls behavior the fix must preserve: a
+ * present record whose indexed attribute is genuinely `null` stays indexed under null.
+ *
+ * Scoped to RocksDB (where F-149 manifests and the raw-index-DBI oracle reads reliably through a
+ * resource; see the skipSuite note). LMDB eviction/index consistency is covered by the cross-engine
+ * eviction-secondary-index.test.ts, and the fix itself is engine-agnostic (shared updateIndices()).
+ */
+import { suite, test, before, after } from 'node:test';
+import { deepStrictEqual, strictEqual } from 'node:assert';
+import { resolve } from 'node:path';
+import { setTimeout as sleep } from 'node:timers/promises';
+import { setupHarperWithFixture, teardownHarper, type ContextWithHarper } from '@harperfast/integration-testing';
+// @ts-expect-error utils/client.mjs has no type declarations; runtime resolves fine
+import { createApiClient } from './../apiTests/utils/client.mjs';
+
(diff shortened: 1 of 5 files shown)


integrationTests/database/eviction-index-null-reindex.test.ts
```diff
@@ -0,0 +1,140 @@
...
+		});
+	}
+	async function getJSON(path: string): Promise<any> {
+		const res = await fetch(`${httpURL}${path}`, { headers: { Authorization: client.headers.Authorization } });
+		return res.json();
+	}
+	const rawIndex = (table: string) => getJSON(`/RawIndex/?table=${table}`);
+	const primaryIds = (table: string) => getJSON(`/PrimaryIds/?table=${table}`);
+
+	test('TTL eviction removes both the base row and its raw secondary-index entry (no [null,id] orphan)', async () => {
```
cb1kenobi, line 82: **Nit: regression coverage exercises only eviction, not the two claims the fix also makes**

The fixed line in `updateIndices` is shared, so proving it on the TTL-eviction path transitively covers explicit `delete` and composite/multi-attribute indexes. But the PR title says "eviction/delete" and the writeup claims composite coverage, and neither is asserted directly here — a future change that diverges the delete path or a resolver-backed (composite/computed) index would slip through. Two cheap additions would lock in both claims against the same raw-DBI oracle:

- an explicit `DELETE /Expiring/<id>` (or `t.delete(id)`) case asserting the `[value,id]` entry is gone with no `[null,id]` orphan;
- a table with a composite/multi-attribute (resolver-backed) `@indexed` field, confirming removal clears the composite entry too.

Not blocking — the core fix is correct and the eviction oracle fails without it.

—
_Generated by Barber AI_

cb1kenobi (approved):

PR #1899, opened 2026-07-22
Fix recordCommitLatency parameter type for the widened commitResolution union

kriszyp:
Main currently has a TS2345 at `resources/DatabaseTransaction.ts:413`: [6e6d57ed8](https://github.com/HarperFast/harper/commit/6e6d57ed8) widened `commitResolution` to `Promise<number | void> | void` for the LMDB-engine guard, but `recordCommitLatency` still declared `Promise<void>`.

The main build job tolerates tsc errors, but the **Next.js adapter integration workflow builds harper with errors fatal — so this fails those checks on every PR** (observed on [#1898](https://github.com/HarperFast/harper/pull/1898), all three Node versions).

The recorder only times settlement (fulfil or reject) and already thenable-guards, so it takes `Promise<unknown> | void`; the now-redundant `as Promise<void>` cast on `transaction.commit()` is removed with its stale comment. Type-only — no runtime change. `tsc --noEmit -p tsconfig.build.json` exits clean with this patch.

### DIFF
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -42,7 +42,7 @@ export function setCommitLatencyRecorder(recorder: ((durationMs: number) => void
 // commit nor surface as an unhandled rejection on this floating `.then`. The thenable guard protects
 // against a future caller passing a non-Promise `commitResolution` (today it is always the rocksdb-js
 // async `Transaction.commit()` result, which is guaranteed to be a Promise).
-function recordCommitLatency(commitResolution: Promise<void>, submittedAt: number) {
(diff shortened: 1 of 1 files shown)


kriszyp:
Note: [#1898](https://github.com/HarperFast/harper/pull/1898) now carries a minimal version of this fix on its branch (3bf19a931, `Promise<void>` → `Promise<unknown>`) so its Next.js adapter CI legs can pass pre-merge. This PR remains the standalone main unblock (and additionally removes the now-redundant `as Promise<void>` cast). Whichever lands second rebases trivially. — Claude (Opus 4.8)

Issue #1865, opened 2026-07-18
5.1→5.2 upgrade boot: cached get/VT read path misses pre-upgrade component-table rows (scans and ops API see them; write heals)


kriszyp:
## Symptom

After upgrading an instance across 5.1.x → 5.2.0-beta.1, application-component tables (defined via `graphqlSchema` `@table`) **lose primary-key `get()` and indexed `search()` visibility of all rows written before the upgrade**, while:

- full-scan `search({})` still returns every row,
- the operations API (`search_by_value` / `search_by_hash` / `describe_all`) sees everything, fully decoded,
- new writes work and read fine.

So the app's hot read paths (`Table.get(id)`, conditioned searches) silently return empty/miss for pre-upgrade data — no errors, no log lines — while the data is demonstrably intact on disk.

## Repro (deterministic)

1. Fresh `harper@5.1.10` install (npm), RocksDB, a component with a `@table` schema.
2. Write rows through the component.
3. Boot `harper@5.2.0-beta.1` against the same root (a version-upgrade boot).
4. `Table.get(id)` → null for every pre-upgrade row; `search({})` → all rows present.

Reproduced on npm `harper` 5.1.10→5.2.0-beta.1 locally (multiple fresh installs) and on a harper-pro:5.1.10→5.2.0-beta.1 Fabric instance.

## Mechanism (traced)

- The miss is in the cached `getEntry` path: `Table.get` → `PrimaryRocksDatabase.getEntry` → the native Verification-Table-aware `getSync`, which returns null for records written by the previous version. `getRange`/full scans and raw `getBinarySync(key)` bypass that path and HIT — on-disk key + value are fine.
- `storage.caching: false` does **not** avoid it: the primary store is opened with `cache: true` hardcoded in `resources/databases.ts::table()`.
- **Heal semantics (measured):** a 5.2 write to a row heals that row's get-path deterministically and immediately. A plain reboot heals *probabilistically* — observed both a fully-healing second boot with zero writes and a 9-boots-straight-broken streak; SSTs were observed rewritten around a healing boot, so the reboot-heal appears to ride RocksDB flush/compaction rewriting the old data, i.e. a race with background file rewriting rather than a per-boot dice roll.

## Impact

[...]

PR #1900, opened 2026-07-22
test(upgrade): promote 3 QA cross-version upgrade read-visibility anchors

kriszyp:
## Summary

Promotes 3 gated qa-explorer regression specs into the tracked `integrationTests/upgrade/` suite, anchoring findings from the gh#1865 upgrade-boot read-visibility investigation. All three are test-only (no product code touched) and each stays its own file — all `requires-isolation` (they boot a real npm-provisioned `harper@5.1.22` and/or restart/SIGKILL the instance).

- **`cross-version-upgrade-visibility.test.ts`** (from QA-658 / P-437): a REAL 5.1.22 → current cross-version upgrade boot keeps every pre-upgrade component-table row visible on all six read surfaces — the gh#1865 refutation. Measured: 6.7s.
- **`cross-version-upgrade-residuals.test.ts`** (from QA-660 / P-438): the same cross-version upgrade stays read-consistent at 55k rows through a SIGKILL-dirty shutdown and boot-boundary writes — closes the scale/shape + dirty-shutdown gap QA-658 left untested. Measured: 15.9s.
- **`upgrade-boot-schema-change.test.ts`** (from QA-647 / P-432): restart + attribute-adding schema change never makes existing rows invisible on any read surface (6-surface matrix, two restarts) — the same-build control arm the two cross-version suites are compared against. Measured: 8.0s.

**Network dependency**: `cross-version-upgrade-visibility` and `cross-version-upgrade-residuals` npm-provision `harper@5.1.22` at runtime (via `~/dev/tmp/qa658-harper51`, overridable with `HARPER_LEGACY_51_PATH`) and skip cleanly if that install isn't present. This is a CI network dependency — leaving it to review whether/how to provision in CI.

## Test plan

- [x] `npm run build` (known pre-existing `resources/DatabaseTransaction.ts` type error unrelated to this change; dist emits fine)
- [x] Ran each promoted file individually, foreground, with `timeout`:
  - `cross-version-upgrade-visibility.test.ts`: 4 pass, 1 skip (no-miss-to-heal), 6.7s
  - `cross-version-upgrade-residuals.test.ts`: 5 pass, 1 skip (no-miss-to-heal), 15.9s
  - `upgrade-boot-schema-change.test.ts`: 4 pass, 1 skip (no-miss-to-heal), 8.0s
- [x] `oxlint` clean, `prettier --write` applied

🤖 Promoted by qa-explorer (Claude Opus 4.8 orchestrator / Sonnet promotion agent)

### DIFF
--- a/integrationTests/upgrade/cross-version-upgrade-residuals.test.ts
+++ b/integrationTests/upgrade/cross-version-upgrade-residuals.test.ts
@@ -1,0 +1,685 @@
+/**
+ * QA-promoted regression anchor — promoted from QA-660 (P-438), a gated qa-explorer finding.
+ * Originating QA scenario: QA-660 (gh#1865 + QA-658/P-437 residual).
+ * Coverage anchored: cross-version 5.1.22 -> current upgrade stays read-consistent at 55k rows
+ * through a SIGKILL-dirty shutdown and boot-boundary writes.
+ *
+ * The honest remainder of the cross-version upgrade-boot read-visibility investigation.
+ *
+ * cross-version-upgrade-visibility.test.ts (QA-658/P-437) already ran the GENUINE 5.1.22 ->
+ * current cross-version upgrade boot (real npm install, real dataRootDir handoff, real
+ * 6-surface matrix vs the index-independent base-store-scan oracle) and found 0/24 misses on a
+ * small, UNIFORM-width dataset stopped with a CLEAN shutdown. That result narrows #1865 but
+ * leaves three variables untested: dataset scale/shape, a DIRTY (SIGKILL) shutdown, and writes
+ * racing the boot boundary. This suite varies all three, stacked onto ONE upgrade boot (not
+ * three isolated runs — see the "why stacked, not isolated" note below), and reuses QA-658's
+ * oracle-vs-surface methodology verbatim.
+ *
+ * Arms exercised on the SAME dataRootDir handoff (5.1.22 -> current build):
+ *   (a) SCALE + SHAPE   — waveA: 50,000 rows, WIDTH-HETEROGENEOUS (6 disjoint field-set shapes
+ *       cycling by index — narrow/no-extras, arrays, long strings, nested objects, mixed
+ *       scalars, unicode — see resources.js hetExtra()), seeded and left to complete CLEANLY
+ *       under 5.1.22. This is the scale+shape control population within the same run.
+ *   (b) DIRTY SHUTDOWN  — waveB: a SECOND width-heterogeneous load (target 40,000 rows) is
+ *       split into 40 independent 1,000-row HTTP requests (so partial completion is possible —
+ *       a single giant request is one atomic transaction; see "why chunked" below) fired with
+ *       concurrency 6, and 5.1.22 is SIGKILLed ~150ms in. The current build then boots over
+ *       this DIRTY (mid-write, unclean-shutdown) state.
+ *   (c) CONCURRENT WRITES ACROSS THE BOOT BOUNDARY — waveC: starting the instant the current
+ *       build is spawned (before `startHarper()`'s promise resolves), a writer loop fires
+ *       single-row POST /Load/ requests every ~3ms against the instance's known hostname:port,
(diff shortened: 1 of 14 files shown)


integrationTests/upgrade/cross-version-upgrade-visibility.test.ts
```diff
@@ -0,0 +1,454 @@
...
+const BATCH = 'waveA';
+
+// Provisioned by: cd ~/dev/tmp/qa658-harper51 && npm install harper@5.1.22 --no-save
+// Override with HARPER_LEGACY_51_PATH if provisioned elsewhere.
+const LEGACY_51_BIN_PATH =
+	process.env.HARPER_LEGACY_51_PATH ??
+	resolve(process.env.HOME ?? '', 'dev/tmp/qa658-harper51/node_modules/harper/dist/bin/harper.js');
+const legacy51Available = existsSync(LEGACY_51_BIN_PATH);
+
+const skipSuite = process.platform === 'win32' || !legacy51Available;
```
cb1kenobi, line 82: **High: this anchor never runs in CI, so it provides no actual regression protection for #1865**

`skipSuite` is true whenever `!legacy51Available`, and `LEGACY_51_BIN_PATH` defaults to `~/dev/tmp/qa658-harper51/node_modules/harper/dist/bin/harper.js` — a path that only exists on a developer machine that manually ran the `npm install harper@5.1.22 --no-save` step described in the file header. Nothing in this PR provisions `HARPER_LEGACY_51_PATH` (or that default path) in CI: `.github/workflows/integration-tests.yml` already provisions a legacy build for the *4.x* suite via `HARPER_LEGACY_VERSION_PATH` (lines 113/192/291), but has no equivalent step for `HARPER_LEGACY_51_PATH`. So on every CI run this whole suite silently skips — the PR title's "promote into the tracked suite" doesn't hold in practice; the test exists in the repo but never executes anywhere except a machine with that exact local install, which means it can't fail on the cross-version regression it's meant to anchor.

Suggested fix: add a CI step (mirroring the existing `HARPER_LEGACY_VERSION_PATH` provisioning in `integration-tests.yml`) that npm-installs `harper@5.1.22` into a known path and sets `HARPER_LEGACY_51_PATH`, or explicitly document in the PR that this anchor is local-only until that's wired up.

—
_Generated by Barber AI_

integrationTests/upgrade/cross-version-upgrade-residuals.test.ts
```diff
@@ -0,0 +1,685 @@
...
+const SAMPLE_SIZE = 2_000; // stratified REST / SQL-IN sample size
+const BULK_CHUNK = 8_000; // chunk size for cachedGetBulk / opsById bulk requests
+const SQL_IN_CHUNK = 200; // chunk size for SQL `WHERE id IN (...)`
+
+// Provisioned by: cd ~/dev/tmp/qa658-harper51 && npm install harper@5.1.22 --no-save
+const LEGACY_51_BIN_PATH =
+	process.env.HARPER_LEGACY_51_PATH ??
+	resolve(process.env.HOME ?? '', 'dev/tmp/qa658-harper51/node_modules/harper/dist/bin/harper.js');
+const legacy51Available = existsSync(LEGACY_51_BIN_PATH);
+const skipSuite = process.platform === 'win32' || !legacy51Available;
```
cb1kenobi, line 104: **High: same CI-provisioning gap as `cross-version-upgrade-visibility.test.ts` — this anchor also never runs in CI**

Same `!legacy51Available` skip condition, same unprovisioned default path. Combined with the sibling file, that's 2 of the 3 "promoted" anchors in this PR that will perpetually skip in CI (only `upgrade-boot-schema-change.test.ts` — the same-build *control* arm, not a cross-version boot — actually runs unconditionally). The two tests that exercise the actual #1865 cross-version scenario are the ones gated off.

Suggested fix: same as the sibling file — provision `HARPER_LEGACY_51_PATH` in CI (or the shared default path) alongside the existing 4.x legacy-install step, so these anchors run somewhere that can actually catch a regression.

—
_Generated by Barber AI_

cb1kenobi (approved):

PR #1905, opened 2026-07-23
Canonicalize PACKAGE_ROOT so it matches realpath'd allowedPath checks

kriszyp:
## What / why

`unitTests/components/globalIsolation.test.js` was reported to intermittently reject `node_modules/mqtt` as "outside of allowed path" specifically when the checkout root is reached through a symlink (observed with dev-agent worktrees under `.claude/worktrees/...`).

Root cause: `PACKAGE_ROOT` (`utility/packageUtils.js`) is derived from a raw directory walk (`__dirname` → `dirname()` up to the nearest `package.json`) with no `realpathSync`. Meanwhile `security/jsLoader.ts`'s `checkAllowedModulePath` **always** `realpathSync()`s the module path it's validating before doing `path.startsWith(allowedPath)`. Whenever the process resolves symlinks inconsistently for these two paths (confirmed reproducible under `--preserve-symlinks`, which some environments set via `NODE_OPTIONS`), `PACKAGE_ROOT` keeps a symlinked prefix while the module path gets canonicalized — so `startsWith` never matches, even though the module is legitimately in scope.

`components/componentLoader.ts` already calls `realpathSync(componentDirectory)` for its `allowedPath` sources — this closes the same gap for `PACKAGE_ROOT`, which several test fixtures (and any other consumer treating it as an allow-list prefix) rely on being canonical.

## Fix

`utility/packageUtils.js`: wrap `PACKAGE_ROOT`'s computed value in `realpathSync()`.

## Test plan

- Added `unitTests/utility/packageUtils.test.js`: spawns a child `node --preserve-symlinks` process that requires `utility/packageUtils.js` through a freshly created symlink to the repo root, and asserts `PACKAGE_ROOT` still equals the canonical (non-symlinked) path. Verified this test **fails without the fix** and **passes with it**.
- Hardened the pre-existing `PACKAGE_ROOT` equality assertion to compare against `realpathSync(...)` rather than a raw `join(...)`, so it stays correct if the suite itself is ever invoked with `--preserve-symlinks` (Gemini review leg caught this).
- `npm run build` + `npm run test:unit:main` (3784 passing, 0 failing) and `unitTests/components/globalIsolation.test.js` + `unitTests/security/**` all green.
- `npm run format:write` / `oxlint` clean on both changed files.

## Review

Cross-model review (quick mode — diff <200 lines): the `reviewer`/`codex-reviewer` subagents aren't installed in this dev-agent profile, so I ran the Gemini leg (`agy`) directly on the diff. It flagged one real issue (the pre-existing equality assertion would break under `--preserve-symlinks`), which is fixed above; its other two suggestions (Windows symlink type, `return` vs `this.skip()`) match the established pattern already used in `unitTests/agent/fsTools.test.js`, so left as-is for consistency.

Refs finding from dispatch `finding-fix-harper-i7a1` (queued from `fix-harper-1880`).

### DIFF
--- a/unitTests/utility/packageUtils.test.js
+++ b/unitTests/utility/packageUtils.test.js
@@ -2,7 +2,9 @@ const { describe, it } = require('mocha');
 
 const assert = require('node:assert');
 const { join } = require('node:path');
-const { readFileSync } = require('node:fs');
+const { readFileSync, realpathSync, symlinkSync, rmSync, mkdtempSync } = require('node:fs');
+const os = require('node:os');
+const { execFileSync } = require('node:child_process');
 
 const packageUtils = require('#src/utility/packageUtils');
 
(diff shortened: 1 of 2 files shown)


Ethan-Arrowood (approved):
LGTM. Canonicalizing `PACKAGE_ROOT` so it matches the realpath'd `allowedPath` prefix checks is the correct fix — checking both sides in canonical form is the sound containment form and doesn't open the inverse (a symlink escaping the boundary still resolves outside it). It's a single `realpathSync` at module load (cold path, `PACKAGE_ROOT` is a const), and the child-process test reproducing the mismatch under `--preserve-symlinks` is a good regression guard.

PR #1906, opened 2026-07-23
fix(resources): Table.clear() also clears secondary index dbis

kriszyp:
## What
`Table.clear()` (`resources/Table.ts`) only cleared the primary store — secondary-index DBIs were left populated with entries pointing at records the clear removed.

## Why
This surfaced as a review finding on the `harper-1839-sql-selectall-pk-sort` work: it's harmless for that test harness (it re-writes the same ids with the same values), but it's a sharp edge for any caller of the exported `clear()` API on an indexed table:

- Stale index entries accumulate and are never freed (storage leak).
- If a caller `clear()`s and then re-writes the **same id with a different indexed value** (an intended usage, not a misuse), the old value's stale index entry keeps pointing at that id. An equality search on the *old* value then incorrectly returns that record, because the equality-index path (`resources/search.ts`) doesn't re-check the record's current attribute value against the search value — it trusts the index.
- A stale entry pointing at a fully-missing record is *not* itself a correctness bug: `Table.transformEntryForSelect` re-resolves every candidate against the primary store and silently `SKIP`s misses — but that also means the previous behavior was masking the real risk rather than avoiding it.

## Fix
`clear()` now also clears every secondary-index dbi, using the same `clearAsync()`/`clear()` pattern `runIndexing` already uses when rebuilding an index from scratch (`resources/databases.ts`).

```diff
 static clear() {
-    return primaryStore.clear();
+    // clear the primary store and every secondary index dbi (same pattern used by
+    // runIndexing when rebuilding from scratch), so clear() doesn't leave stale
+    // index entries pointing at records that no longer exist.
+    const promises = [primaryStore.clear()];
+    for (const key in indices) {
+        const index = indices[key];
+        promises.push(index.clearAsync ? index.clearAsync() : index.clear());
+    }
+    return Promise.all(promises);
 }
```

## Testing
New unit test `unitTests/resources/tableClearSecondaryIndex.test.js`:
- Confirms the primary store and an indexed attribute's search results are both empty after `clear()`.
- Reproduces the real regression: `put(id, tag: 'before')` → `clear()` → `put(id, tag: 'after')` → searching for the stale old value `'before'` must return 0 rows. Verified this test fails (returns 1 stale row) against the pre-fix code and passes after the fix.

Full `test:unit:resources` suite passes (1253 passing, 0 failing) after the change.

### DIFF
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -4721,7 +4721,15 @@ export function makeTable(options) {
 			return history.reverse();
 		}
 		static clear() {
-			return primaryStore.clear();
+			// clear the primary store and every secondary index dbi (same pattern used by
+			// runIndexing when rebuilding from scratch), so clear() doesn't leave stale
+			// index entries pointing at records that no longer exist.
+			const promises = [primaryStore.clear()];
+			for (const key in indices) {
+				const index = indices[key];
+				promises.push(index.clearAsync ? index.clearAsync() : index.clear());
+			}
+			return Promise.all(promises);
 		}
 		static cleanup() {
 			deleteCallbackHandle?.remove();
--- a/unitTests/resources/tableClearSecondaryIndex.test.js
+++ b/unitTests/resources/tableClearSecondaryIndex.test.js
@@ -1,0 +1,103 @@
+/**
+ * Regression test for the finding raised against harper-1839-sql-selectall-pk-sort:
+ * `Table.clear()` (resources/Table.ts) only cleared the primary store, leaving
+ * secondary-index DBIs populated with entries pointing at records that no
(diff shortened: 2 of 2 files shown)


cb1kenobi (approved):

Ethan-Arrowood (approved):
LGTM. This closes a real data-integrity gap — `Table.clear()` was emptying only the primary store and leaving every secondary-index dbi populated, so stale index entries could surface as phantom equality-search hits (plus the storage leak). Clearing through the dynamically-enumerated `indices` map is the right call: it covers TTL/`expiresAt`, custom, and HNSW indexes and won't drift as new index types are added. The regression test reproduces the exact sharp edge (fails before, passes after).

PR #1907, opened 2026-07-23
Fail soft on operations API domain socket bind failure; warn on path-length overflow

kriszyp:
## What / why

The operations API's Unix-domain-socket path is derived from `rootPath` by default. When that path exceeds `sockaddr_un.sun_path` (107 usable bytes on Linux, 103 on macOS), supported Node versions do not behave consistently: some reject `listen()`, while others silently truncate the path and bind a socket clients cannot reach at the configured name.

That known, deterministic path-length condition should not prevent a configured TCP endpoint from starting. Every other operations API listener failure is catastrophic and must reject startup so `bin/run.ts` exits non-zero.

## Changes

- Preflight UDS paths using byte length and skip only paths beyond the platform limit. This avoids both rejected and silently truncated binds.
- Keep every actual `listen()` failure hard: the listener promise rejects, including permission, missing-parent, address-conflict, and other filesystem/runtime errors.
- In single-thread mode, explicitly await the cached listener batch so a bind rejection reaches the top-level non-zero exit path.
- Remove the temporary bind-error listener after a successful listen, preventing later live-server errors from being misclassified as bind failures.
- Warn on an overlong configured operations socket with conditional wording: the domain socket may be unavailable/truncated; any separately configured TCP endpoint is unaffected.
- Suppress proxy discovery metadata only for the preflight-skipped overlong UDS path.

## Test notes

- Added coverage for platform limits and UTF-8 byte counting.
- Added real `net.Server` coverage proving an overlong path is skipped, an in-limit bind removes its temporary error listener, and a non-path-length filesystem bind failure rejects startup.
- Validation/config suites: 142 passing.
- Focused UDS/server/utility suites: 33 passing.
- Main unit suite: 3,800 passing, 188 pending.
- Resources unit suite: 1,251 passing, 14 pending.
- Build, lint, and formatting checks pass locally.
- The full integration run reached one pre-existing QA-269 `SQL UPDATE resets TTL` failure; all UDS/startup coverage and the remaining integration run passed.

## Review coverage

- Cross-model review: Gemini completed; its Node-version truncation finding led to the preflight check. The local Claude leg timed out before producing a usable review; the repository's Claude review workflow will re-run on this push.

No configuration or API surface was added, so no documentation companion is needed.

Generated with Codex.

### DIFF
--- a/server/threads/socketRouter.ts
+++ b/server/threads/socketRouter.ts
@@ -39,7 +39,11 @@ export async function startHTTPThreads(threadCount = 2, dynamicThreads?: boolean
 			const { loadRootComponents } = require('../loadRootComponents.js');
 			if (threadCount === 0) {
 				setMainIsWorker(true);
-				await require('./threadServer.js').startServers();
+				const threadServer = require('./threadServer.js');
+				await threadServer.startServers();
+				// startServers() schedules listener startup after loading components; await its cached
+				// batch so a bind failure reaches bin/run.ts and exits non-zero in single-thread mode too.
+				await threadServer.listenOnPorts();
 				return Promise.resolve([]);
 			}
 			await loadRootComponents();
--- a/config/configUtils.ts
+++ b/config/configUtils.ts
@@ -1,7 +1,7 @@
 import * as hdbTerms from '../utility/hdbTerms.ts';
 import * as hdbUtils from '../utility/common_utils.ts';
 import logger from '../utility/logging/harper_logger.ts';
-import { configValidator } from '../validation/configValidator.ts';
+import { configValidator, getDomainSocketPathLengthWarning } from '../validation/configValidator.ts';
 import fs from 'fs-extra';
 import YAML from 'yaml';
 import path from 'path';
@@ -662,7 +662,10 @@ function validateConfig(configDoc, skipFsValidation = false) {
 	configDoc.setIn(['logging', 'root'], validation.value.logging.root);
 	configDoc.setIn(['storage', 'path'], validation.value.storage.path);
 	configDoc.setIn(['logging', 'rotation', 'path'], validation.value.logging.rotation.path);
-	configDoc.setIn(['operationsApi', 'network', 'domainSocket'], validation.value?.operationsApi?.network?.domainSocket);
+	const domainSocket = validation.value?.operationsApi?.network?.domainSocket;
+	configDoc.setIn(['operationsApi', 'network', 'domainSocket'], domainSocket);
+	const domainSocketWarning = getDomainSocketPathLengthWarning(validation.value.rootPath, domainSocket);
+	if (domainSocketWarning) logger.warn(domainSocketWarning);
 }
 
 /**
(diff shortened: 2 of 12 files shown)


kriszyp:
Addressed the draft-review feedback in 120846acd:

- Removed the stale bind-error listener after a successful UDS listen.
- Removed the fatal-vs-error continuation split: only a preflight-proven overlong path is soft; every actual listen error rejects startup.
- Added regressions for overlong skip, real filesystem bind rejection, and listener cleanup.
- Made the warning wording conditional and switched the test assertions to plain `node:assert` + `strictEqual`.

GitHub will not accept inline replies while the existing Kris review remains pending, so this comment records the per-thread resolution without submitting or altering that draft review.

server/threads/socketRouter.ts
```diff
@@ -39,7 +39,11 @@ export async function startHTTPThreads(threadCount = 2, dynamicThreads?: boolean
 			const { loadRootComponents } = require('../loadRootComponents.js');
 			if (threadCount === 0) {
 				setMainIsWorker(true);
-				await require('./threadServer.js').startServers();
+				const threadServer = require('./threadServer.js');
+				await threadServer.startServers();
+				// startServers() schedules listener startup after loading components; await its cached
+				// batch so a bind failure reaches bin/run.ts and exits non-zero in single-thread mode too.
+				await threadServer.listenOnPorts();
```
Ethan-Arrowood, line 46: This second `listenOnPorts()` await is not idempotent under Bun: the comment says "await its cached batch", but `listenOnPorts()` delegates to `listenOnPortsBun()` *before* the cache check, and `listenOnPortsBun()` has no cache. With `threads.count: 0` on Bun, `startServers()` already ran the batch, so this call re-executes it — live UDS socket files get `unlinkSync`'d and re-served, TCP ports double-`Bun.serve`, and any raw `net.Server` in `SERVERS` throws `ERR_SERVER_ALREADY_LISTEN`, rejecting a startup that actually succeeded. Bun CI is green only because the integration matrix doesn't run single-thread mode. One-line fix: give `listenOnPortsBun()` the same `if (listening) return …` cache, or hoist the cache check above the `isBun` branch.

Ethan-Arrowood (approved):
Approving — the fail-soft design is sound, the 107/103-byte limits check out (`sun_path` minus trailing NUL), the metadata suppression is race-free, and everything stays on the startup/config cold path. A few findings, none blocking; the Bun one inline is the one I'd most like to see fixed before or shortly after merge.

Not anchorable inline (lines outside the diff):

- **CLI misdiagnosis in the fail-soft state** (`bin/cliOperations.ts:274`): when the socket file is absent the CLI prints the "not running" message and exits 1 — even though `getHdbPid()` just confirmed Harper is running. This PR turns "server won't start" into "server runs, CLI lies about it" (and `bin/mcp/client.ts` shares the transport). Cheap follow-up: when the pid exists but the socket doesn't, check `isDomainSocketPathTooLong()` and print the real cause + remediation.
- **Startup banner contradiction** (`bin/run.ts:322`): `startupLog` prints `unix socket: <path>` unconditionally from config, so the success banner contradicts the error logged two lines earlier. The resolved `{ port, failed: true }` sentinel from `listenOnDomainSocket` is currently consumed by nothing — it could annotate the banner line (e.g. `unix socket: <path> (UNAVAILABLE: path too long)`).
- **Test isolation**: `threadServerListenOnPorts.test.js` calls the real `listenOnPorts()`, permanently caching a rejected batch for the rest of the test process. Harmless today; worth an exported reset hook if the suite grows.
- **Release notes**: fatal→soft on an overlong ops-socket path is a behavior change worth a changelog line, and the operations-API docs page should mention the socket's new failure mode.

PR #1908, opened 2026-07-23
test: add packaged application E2E workflow

kriszyp:
## Summary

- run nightly at 08:17 UTC against the default branch, with manual dispatch available for diagnosis
- build Harper with the same `npm run package` entry point used by releases
- verify and upload the generated npm tarball once for all downstream jobs
- install that artifact, deploy redirector and early-hints from their current main branches, and run their integration suites
- stage clean deployment copies without checkout metadata or `node_modules`, exercising Harper's server-side dependency installation
- run Harper's existing acl-connect MQTT integration coverage against the packaged runtime
- scope internal-repository checkout tokens to read-only access for only the selected application

## Validation

- redirector: 65 passing from a clean staged deployment
- early-hints: 8 passing from a clean staged deployment
- acl-connect: 13 passing
- `npm run test:unit:main`: 3,783 passing, 188 pending
- `npm run test:unit:resources`: 1,251 passing, 14 pending
- `npm run build`
- `npm run lint:required`
- Prettier, actionlint, and `git diff --check`

The repository-wide integration command was also exercised for 16 minutes. It reached the non-gating `qa-scratch` defect-probe corpus and reproduced existing intentional/known failures, including zero-timeout probes, so it was stopped after the task-specific packaged-runtime suites had passed.

## Attention

- redirector and early-hints intentionally track mutable downstream `main`, so a failure can represent compatibility drift on either side.
- the workflow intentionally does not run on pull requests; scheduled and manual runs become available from the default branch after merge.
- their app-owned integration suites run with one Harper worker to match their immediate timing/state assumptions; Harper's integration matrix retains broader multi-worker coverage.
- acl-connect currently has no downstream integration suite; this uses the published 1.0.10 package through Harper's existing MQTT ACL integration test.
- cross-model coverage: Gemini and Harper-domain review passed; the review's GitHub App token-scoping blocker was fixed, and the nightly-trigger follow-up received a clean quick Harper-domain review.

### DIFF
--- a/.github/workflows/npm-package-app-e2e.yml
+++ b/.github/workflows/npm-package-app-e2e.yml
@@ -89,7 +89,7 @@ jobs:
         id: app-token
         uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
         with:
-          app-id: ${{ secrets.HARPERFAST_AI_CLIENT_ID }}
+          client-id: ${{ secrets.HARPERFAST_AI_CLIENT_ID }}
           private-key: ${{ secrets.HARPERFAST_AI_APP_PRIVATE_KEY }}
           owner: ${{ github.repository_owner }}
           repositories: ${{ matrix.app }}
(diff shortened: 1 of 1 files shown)


cb1kenobi (approved):
This is rad!

PR #1910, opened 2026-07-23
Activate WAF on upgraded instances

kriszyp:
## Summary

Adds `waf` to the safe list of newly introduced built-in component keys that are restored on in-place upgrades. This gives upgraded Harper Pro 5.2 instances the same WAF activation behavior as fresh installs while preserving explicit `waf: false` configuration.

The focused backfill test covers registration-gated activation and the persisted empty WAF config block. General WAF documentation is already covered by [docs: add Web Application Firewall reference](https://github.com/HarperFast/documentation/pull/603).

Generated by Codex (GPT-5).

### DIFF
--- a/config/configUtils.ts
+++ b/config/configUtils.ts
@@ -363,7 +363,7 @@ export function ensureConfigKeysPresent(keys: string[]): string[] {
  * Built-in components introduced in a recent release whose config key must be backfilled onto
  * in-place-upgraded instances. Fresh installs get these from defaultConfig.yaml, but an upgrade
  * carries the old config forward without them, leaving the component dormant (harper-pro#585:
- * secretCustody added in 5.2).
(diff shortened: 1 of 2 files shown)


Ethan-Arrowood (approved):
LGTM. Enrolling `waf` into the existing `secretCustody` upgrade-backfill is a clean, low-risk reuse — the empty `waf: {}` default is genuinely pass-through (rules are table-driven, the matcher starts null, and a no-match falls through to the next handler), so activating it on unconfigured upgraded instances is a no-op at the request path. Idempotent and fresh/upgrade parity holds.

One question before this lands: since this auto-activates the WAF on every upgraded instance, should it get a 5.2 release-note entry? It's a behavior change on upgrade even though it's pass-through by default, and I'd lean toward calling it out so operators know the component is now present.

Issue #1028, opened 2026-06-01
Case-insensitive schema and table name handling


kriszyp:
Make schema and table name lookups case-insensitive so `MyTable` and `mytable` refer to the same object, eliminating a common source of user confusion.

🤖 Filed by Claude on behalf of Kris.

PR #1913, opened 2026-07-23
fix(auth): reject reserved role-permission names as database names + complete cluster_user handling

### LINKED ISSUE #1016: Reserved names 'super_user' and 'cluster_user' can be used as schema names
Creating a schema named 'super_user' or 'cluster_user' succeeds but creates a conflict with Harper's built-in role names, causing undefined behavior. These names should be rejected as schema identifiers.

kylebernhardy:
## Summary

A role's `permission` object mixes named flags — `super_user`, `cluster_user`, `structure_user`, `operations`, `_expandedOperations` — with per-database permission blocks keyed by **database name**. A database named after one of those flags collides with it: the role parser cannot tell a `super_user` flag from a `super_user` database's permissions.

## Purpose

Fixes [#1016](https://github.com/HarperFast/harper/issues/1016). Beyond "undefined behavior," there's a concrete **privilege-escalation** path (surfaced in review): a component shipping `@table(database: "super_user")` plus a `roles.yaml` granting permissions on that "database" makes `permission.super_user` a truthy **object**, and the super-user check (`!!permission.super_user`) then reads as granted.

## The fix

Reject the reserved names as database identifiers on **both** creation surfaces, from one shared source of truth (`RESERVED_DATABASE_NAMES` in `hdbTerms.ts`):

1. **Operations API** (`create_schema` / `create_table`) — `.invalid(...)` on `DB_NAME_CONSTRAINTS` in `dataLayer/schema.ts`.
2. **Schema authoring** (`schema.graphql` `@table(database:)` / programmatic `table()`) — a check in `resources/databases.ts::table()`, which registers databases and **bypasses** the operations validation. This is the path the escalation uses.

It also folds the incomplete `USERS_NOT_DBS` in `resources/roles.ts` (was `['super_user', 'structure_user']` — missing `cluster_user`/`operations`) into the same shared constant, fixing a latent bug where a `roles.yaml` using those flags mis-parsed them as database names.

## Where to look

- **Enforcement at two layers.** The operations-only version was incomplete — the graphql/programmatic path was the reviewers' blocker. `table()` throws a `ClientError` (matching the existing schema-validation throws in `graphql.ts`), so it's load-safe: an invalid schema fails to load rather than bricking the instance.
- **Reserved set completeness.** `_expandedOperations` is included (an internal, `_`-prefixed key) because `schemaRegex` permits leading underscores and a same-named database would overwrite the cached operations `Set` at translation time (a `TypeError` crash). Confirm the set matches `UserRoleNamedPermissions` in `security/user.ts`.
- **Case sensitivity.** `.invalid()` matches exact strings; database names are currently case-sensitive (case-insensitive handling is separate issue #1028), so exact match is correct.

## Tests

New `unitTests/dataLayer/reservedSchemaNames.test.js` — each reserved name × three surfaces (`create_schema`, `create_table`, `table()`), plus a non-reserved control. 15 of 16 fail on `origin/main` (verified by reverting the source files). Existing `resources/databases`, `query`, `crud`, `defineTable-registration`, `permissions`, `permissionsTranslator` suites pass unchanged.

## Also fixed (related `cluster_user` gaps, second commit)

Three pre-existing gaps in how `cluster_user` is handled as a role flag, surfaced in review:

- **`role_validation.ts` misleading error** — its boolean-flag list came from `ROLE_TYPES_ENUM` (only `super_user`), so a non-boolean `cluster_user` skipped the boolean check and fell through to database validation, reporting `database 'cluster_user' does not exist` instead of the boolean-type error the `SU_CU_ROLE_BOOLEAN_ERROR` name already implies. Fixed by adding `CLUSTER_USER` to `ROLE_TYPES_ENUM`. Guarded by two new `role_validation` tests (both fail on `origin/main`).
- **`permissionsTranslator.js` dropped `cluster_user`** — preserved `super_user`/`structure_user`/`operations` but not `cluster_user`. Now preserved alongside them. Note: `cluster_user` is read nowhere at runtime today (only a type def + an impersonation-clear write), so this is a consistency/hygiene fix with no behavioral effect — flagging that explicitly.
- **Dead `SU_CU_ROLE_COMBINED_ERROR`** removed from the test-error helper — no production counterpart, no test uses it (the su+cu rejection is covered by `SU_CU_ROLE_NO_PERMS_ALLOWED`).

## Upgrade / migration note

Deep-review (auth + data-integrity domain passes) confirmed the fix is startup-safe: a reserved-name database that already exists on disk still **loads** (schema replay goes through `makeTable`, not the guarded `table()`), so its data stays accessible, and it can be **dropped** to remediate (the drop path goes through `database()`, also unguarded — which is why the check is deliberately on the authoring path only).

But such a database's **component will fail to load** after upgrade — the `@table(database:)` authoring call now throws, so that Resource becomes an `ErrorResource` (logged, isolated to that component; not fatal to the instance). The highest-risk name is **`operations`**, which only became a permission flag recently (`95aa06fd1`), so pre-existing `operations` databases are more plausible than `super_user` ones. **Release notes should tell operators to rename any pre-existing database named after a reserved flag before upgrading.**

## Note

Security-adjacent (the escalation path). Milestone is v5.2; **could be a v5.1 patch** given the security angle — flagging for a decision rather than labeling.

### DIFF
--- a/unitTests/dataLayer/reservedSchemaNames.test.js
+++ b/unitTests/dataLayer/reservedSchemaNames.test.js
@@ -1,0 +1,53 @@
+const assert = require('node:assert/strict');
+const schema = require('#src/dataLayer/schema');
+const { table } = require('#src/resources/databases');
+
+// harper#1016: a role's `permission` object keys per-database permissions by
+// database name alongside named flags (super_user / cluster_user /
+// structure_user / operations / _expandedOperations). A database named after one
+// of those flags collides with it, so those names are rejected as database
+// identifiers — at both the operations API (create_schema / create_table) and
+// the schema-authoring path (graphql `@table(database:)` / programmatic table()).
+const RESERVED = ['super_user', 'cluster_user', 'structure_user', 'operations', '_expandedOperations'];
+
+describe('#1016 reserved database names', () => {
+	for (const name of RESERVED) {
+		it(`create_schema rejects reserved name '${name}'`, async () => {
+			await assert.rejects(
+				() => schema.createSchemaStructure({ operation: 'create_schema', schema: name }),
+				/reserved name/,
+				`create_schema accepted reserved name '${name}'`
+			);
+		});
+
+		it(`create_table rejects reserved database '${name}'`, async () => {
+			await assert.rejects(
+				() => schema.createTableStructure({ operation: 'create_table', schema: name, table: 't', primary_key: 'id' }),
+				/reserved name/,
+				`create_table accepted reserved database '${name}'`
+			);
+		});
+
+		it(`schema authoring (table()) rejects reserved database '${name}'`, () => {
+			// graphql `@table(database:)` and programmatic table() funnel through here,
+			// bypassing the operations-API validation. The check throws before any store IO.
+			assert.throws(
+				() => table({ database: name, table: 't', attributes: [{ name: 'id', isPrimaryKey: true }] }),
+				/reserved name/,
+				`table() accepted reserved database '${name}'`
+			);
+		});
+	}
+
+	it('does not reject a non-reserved database name as reserved', async () => {
+		// A normal name must pass the reserved-name check. It may still fail later
+		// (no schema metadata / bridge in this unit context) — just not as reserved.
+		let err;
+		try {
+			await schema.createSchemaStructure({ operation: 'create_schema', schema: 'my_normal_db' });
+		} catch (e) {
+			err = e;
+		}
+		if (err) assert.doesNotMatch(err.message, /reserved name/);
+	});
+});
--- a/unitTests/dataLayer/reservedSchemaNames.test.js
+++ b/unitTests/dataLayer/reservedSchemaNames.test.js
@@ -1,4 +1,4 @@
-const assert = require('node:assert/strict');
+const assert = require('node:assert');
 const schema = require('#src/dataLayer/schema');
 const { table } = require('#src/resources/databases');
 
--- a/unitTests/dataLayer/reservedSchemaNames.test.js
+++ b/unitTests/dataLayer/reservedSchemaNames.test.js
@@ -39,15 +39,17 @@ describe('#1016 reserved database names', () => {
 		});
 	}
 
-	it('does not reject a non-reserved database name as reserved', async () => {
-		// A normal name must pass the reserved-name check. It may still fail later
-		// (no schema metadata / bridge in this unit context) — just not as reserved.
-		let err;
-		try {
-			await schema.createSchemaStructure({ operation: 'create_schema', schema: 'my_normal_db' });
-		} catch (e) {
-			err = e;
+	// Validation-only control (review): calling `createSchemaStructure` with a nonexistent normal name
+	// takes the successful creation path, so it created `my_normal_db` on whatever root the environment
+	// was configured with and signalled a schema change — a unit test with a side effect on a developer
+	// checkout. `validateDatabaseName` runs the same constraint the create paths run, without the IO.
+	it('does not reject a non-reserved database name as reserved', () => {
+		assert.strictEqual(schema.validateDatabaseName('my_normal_db'), undefined);
+	});
+
+	it('rejects a reserved name through the same validation, so the control above is meaningful', () => {
+		for (const name of RESERVED) {
+			assert.match(schema.validateDatabaseName(name).message, /reserved name/);
 		}
-		if (err) assert.doesNotMatch(err.message, /reserved name/);
 	});
 });
--- a/utility/hdbTerms.ts
+++ b/utility/hdbTerms.ts
@@ -83,9 +83,16 @@ export const LAUNCH_SERVICE_SCRIPTS = {
 	MAIN: 'dist/bin/harper.js',
 } as const;
 
-/** Specifies user role types */
+/**
+ * Special boolean role flags validated as role types (not database permissions).
+ * `cluster_user` belongs here alongside `super_user`: role validation checks both
+ * are booleans (SU_CU_ROLE_BOOLEAN_ERROR), and without it a non-boolean
+ * `cluster_user` falls through to database validation and reports a misleading
+ * "database 'cluster_user' does not exist" error (harper#1016).
+ */
 export const ROLE_TYPES_ENUM = {
 	SUPER_USER: 'super_user',
+	CLUSTER_USER: 'cluster_user',
 } as const;
 
 /**
--- a/utility/hdbTerms.ts
+++ b/utility/hdbTerms.ts
@@ -102,6 +102,8 @@ export const ROLE_TYPES_ENUM = {
  * a `super_user` flag from a `super_user` database's permissions — producing
  * undefined behavior, so these are rejected as database/schema identifiers
  * (harper#1016). Keep in sync with `UserRoleNamedPermissions` in security/user.ts.
+ * `access` is deliberately NOT here: roles.ts strips it out of `permission`
+ * before persistence (scoped-delegation), so it never collides at runtime.
  */
 export const RESERVED_DATABASE_NAMES = [
 	'super_user',
--- a/utility/hdbTerms.ts
+++ b/utility/hdbTerms.ts
@@ -88,6 +88,22 @@ export const ROLE_TYPES_ENUM = {
 	SUPER_USER: 'super_user',
 } as const;
 
+/**
+ * Named permission flags that live in a role's `permission` object alongside
+ * per-database permission blocks, which are keyed by database name. A database
+ * named after one of these flags collides with it — the role parser cannot tell
+ * a `super_user` flag from a `super_user` database's permissions — producing
+ * undefined behavior, so these are rejected as database/schema identifiers
+ * (harper#1016). Keep in sync with `UserRoleNamedPermissions` in security/user.ts.
+ */
+export const RESERVED_DATABASE_NAMES = [
+	'super_user',
+	'cluster_user',
+	'structure_user',
+	'operations',
+	'_expandedOperations',
+] as const;
+
 /** Email address for support requests */
 export const HDB_SUPPORT_ADDRESS = 'support@harperdb.io';
 
--- a/validation/role_validation.ts
+++ b/validation/role_validation.ts
@@ -100,7 +100,11 @@ function customValidate(object, constraints) {
 		}
 		//check if cu or su values, if included, are booleans
 		ROLE_TYPES.forEach((role) => {
-			if (object.permission[role as any] && !validate.isBoolean(object.permission[role as any])) {
+			// Gate on key presence, not truthiness: a falsey non-boolean (0, '', null) must still be
+			// rejected as a non-boolean. A truthiness gate would let those skip both this check AND the
+			// database-permission loop below (the key is a role type, so that loop excludes it), so an
+			// add/alter role would silently accept a value that violates the boolean contract.
+			if (role in object.permission && !validate.isBoolean(object.permission[role as any])) {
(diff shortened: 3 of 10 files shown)


kylebernhardy:
@kriszyp — a release-targeting question on this one.

The core fix rejects reserved role-permission names (`super_user`, `cluster_user`, `structure_user`, `operations`, `_expandedOperations`) as database names, at both the operations API and the schema-authoring path (`@table(database:)` / `table()`). The schema-authoring path is the one that matters most: a component declaring `@table(database: "super_user")` plus a `roles.yaml` granting permissions on that "database" makes `permission.super_user` a truthy object, which the super-user check reads as granted — a privilege-escalation.

We're planning to land this on **main / v5.2**. Given the escalation angle, do you think it also warrants a **v5.1 patch** (via the `patch` label / cherry-pick), or are you comfortable v5.2-only? It cherry-picks cleanly as far as I can tell; your call on whether the security aspect justifies backporting.

(This PR also folds in three related pre-existing `cluster_user` gaps you'd expect to travel with it — detailed in the description.)

Comment generated by kAIle (Claude Opus 4.8)

unitTests/dataLayer/reservedSchemaNames.test.js
```diff
@@ -0,0 +1,53 @@
+const assert = require('node:assert/strict');
```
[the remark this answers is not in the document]
kylebernhardy: Fixed in ec635b987 — swapped to plain `node:assert`. Confirmed against AGENTS.md and .gemini/styleguide.md; the assertions used here (`rejects`/`throws`/`doesNotMatch`) are identical on plain assert, so it's a clean swap.

Comment generated by kAIle (Claude Opus 4.8)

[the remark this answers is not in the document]
kylebernhardy: Done in ec635b987. You're right it's a style nit rather than a lint failure today, but it's the documented house style (AGENTS.md:195), so aligned it. Thanks for the precise read on the oxlint rule scope.

Comment generated by kAIle (Claude Opus 4.8)

utility/hdbTerms.ts
```diff
@@ -83,11 +83,34 @@ export const LAUNCH_SERVICE_SCRIPTS = {
...
+/**
+ * Special boolean role flags validated as role types (not database permissions).
+ * `cluster_user` belongs here alongside `super_user`: role validation checks both
+ * are booleans (SU_CU_ROLE_BOOLEAN_ERROR), and without it a non-boolean
+ * `cluster_user` falls through to database validation and reports a misleading
+ * "database 'cluster_user' does not exist" error (harper#1016).
+ */
 export const ROLE_TYPES_ENUM = {
 	SUPER_USER: 'super_user',
+	CLUSTER_USER: 'cluster_user',
```
kriszyp, line 95: Adding `cluster_user` to `ROLE_TYPES` correctly stops treating it as a database, but it exposes the existing truthiness gate in `role_validation.ts` (`if (object.permission[role] && !validate.isBoolean(...))`). Falsey non-booleans such as `0`, `''`, and `null` now skip both that check and the database-permission loop, so add/alter role accepts them despite the boolean contract. Could we check key presence/definedness instead and add at least `0`/`null` regression cases? That would also close the latent `super_user` variant.

kriszyp (changes requested):

utility/hdbTerms.ts
kylebernhardy: Great catch — you're right, the `&&` truthiness prefix let `0`/`''`/`null` skip both the boolean check and (since the key is a role type) the database-permission loop. Fixed in 54962ed0c by gating on key presence instead: `role in object.permission && !isBoolean(...)`, so a defined-but-falsey non-boolean is now rejected with the boolean error. Added regression cases for `0`/`''`/`null` on both `cluster_user` and `super_user` (the latent variant you flagged). Thanks for the thorough review!

utility/hdbTerms.ts
```diff
@@ -83,11 +83,36 @@ export const LAUNCH_SERVICE_SCRIPTS = {
...
```
kriszyp, line 95: Adding `cluster_user` to `ROLE_TYPES` makes the database-permission loop skip it, but `validateNoSUPerms()` still enforces exclusivity only for `super_user`. Consequently `{ cluster_user: true, dev: ... }` (and CU plus `operations`/`structure_user`) now validates, despite that helper's SU/CU contract and the historical `SU_CU_ROLE_NO_PERMS_ALLOWED('cluster_user')` behavior. Please restore the `cluster_user === true` branch and add CU-plus-database/operations coverage, unless mixed cluster roles are an intentional contract change.

validation/role_validation.ts
```diff
@@ -100,7 +100,11 @@ function customValidate(object, constraints) {
 		}
 		//check if cu or su values, if included, are booleans
 		ROLE_TYPES.forEach((role) => {
-			if (object.permission[role as any] && !validate.isBoolean(object.permission[role as any])) {
+			// Gate on key presence, not truthiness: a falsey non-boolean (0, '', null) must still be
+			// rejected as a non-boolean. A truthiness gate would let those skip both this check AND the
+			// database-permission loop below (the key is a role type, so that loop excludes it), so an
+			// add/alter role would silently accept a value that violates the boolean contract.
+			if (role in object.permission && !validate.isBoolean(object.permission[role as any])) {
```
kriszyp: Because the `permission` constraint checks only presence, malformed payloads such as `permission: "x"`, `1`, or `true` reach this line. The `in` operator throws on those primitives, changing a validation failure into a 500. Could this use `Object.hasOwn(object.permission, role)` under the existing truthy guard (and ideally add an object-shape constraint/test) so key-presence semantics remain fail-safe?

utility/hdbTerms.ts
kylebernhardy: Confirmed and fixed in c21e83c8e. You're right that this is a regression from my change, not an intentional contract change — before `cluster_user` joined `ROLE_TYPES`, `{ cluster_user: true, someDb: {...} }` was rejected *incidentally*, because the database-permission loop treated `cluster_user` as a database whose value wasn't an object. Making it a role type removed that accident, and `validateNoSUPerms` only ever checked `super_user` — so the exclusivity the `SU_CU_ROLE_NO_PERMS_ALLOWED` message has always named for CU quietly stopped being enforced.

`validateNoSUPerms` now returns the CU error for `cluster_user === true` with any other key, alongside the existing SU branch. Regression coverage added for CU combined with a database, `operations`, and `structure_user`.

validation/role_validation.ts
kylebernhardy: Good catch — this one was introduced by my own fix, and it is exactly as you describe. `permission` carries only a presence constraint, so `permission: 'x'` / `1` / `true` reaches that line, and `in` throws a TypeError on a primitive, turning a 400 into a 500. Verified: `'x' in 'str'`, `'x' in 1`, and `'x' in true` all throw.

Fixed in c21e83c8e with `Object.hasOwn(object.permission, role)`, which coerces and returns false rather than throwing, so key-presence semantics stay fail-safe. `validateNoSUPerms` picked up the same type guard, since it also dereferences `permission` before any shape check. Added a regression test per malformed primitive asserting the call does not throw.

On the object-shape constraint: I agree that's the deeper fix — the real invariant is that `permission` must be an object, and both call sites are currently working around its absence. I left it out of this PR because it would reject payloads the API accepts today and deserves its own change; happy to file it as a follow-up if you'd like.

validation/role_validation.ts
```diff
@@ -100,7 +100,14 @@ function customValidate(object, constraints) {
...
 		ROLE_TYPES.forEach((role) => {
-			if (object.permission[role as any] && !validate.isBoolean(object.permission[role as any])) {
+			// Gate on key presence, not truthiness: a falsey non-boolean (0, '', null) must still be
+			// rejected as a non-boolean. A truthiness gate would let those skip both this check AND the
+			// database-permission loop below (the key is a role type, so that loop excludes it), so an
+			// add/alter role would silently accept a value that violates the boolean contract.
+			// `Object.hasOwn`, not `in`: `permission` is only presence-constrained, so a malformed payload
+			// (`permission: 'x'` / `1` / `true`) reaches here, and `in` throws on a primitive — turning a
+			// validation failure into a 500. `hasOwn` coerces and returns false instead.
+			if (Object.hasOwn(object.permission, role) && !validate.isBoolean(object.permission[role as any])) {
```
kriszyp, line 115: This avoids the primitive `in` exception, but it does not keep every malformed permission at 400 as the new regression comment says. I reproduced `addRoleValidation` returning `null` for `permission: 1`, `true`, and `[]`: the presence-only constraint accepts them, `Object.hasOwn` finds no role flags, and `for...in` has no entries. Since `hdb_role.permission` has no storage type constraint, `add_role` can then persist an invalid role. Could we reject non-object/null/array permissions up front and change the new primitive tests to assert `statusCode === 400`, not only that validation does not throw?

unitTests/dataLayer/reservedSchemaNames.test.js
```diff
@@ -0,0 +1,53 @@
...
+			);
+		});
+	}
+
+	it('does not reject a non-reserved database name as reserved', async () => {
+		// A normal name must pass the reserved-name check. It may still fail later
+		// (no schema metadata / bridge in this unit context) — just not as reserved.
+		let err;
+		try {
+			await schema.createSchemaStructure({ operation: 'create_schema', schema: 'my_normal_db' });
```
kriszyp: A nonexistent normal name takes the successful creation path here, so this test opens/creates `my_normal_db` and signals a schema change. Because this file does not initialize the isolated unit-test environment before importing Harper modules and never drops the database, my focused run attempted the configured `/home/kzyp/dev/tmp/dispatch-hdb` path (and emitted EROFS while still passing); a writable developer root would be left modified. Please make this a validation-only control, or set up an isolated test root and explicitly drop/clean up the schema.

kriszyp (commented):
A couple more issues to resolve, I think.

validation/role_validation.ts
kylebernhardy: You are right, and the reproduction is exact — `permission: 1` / `true` / `[]` returned `null` from `addRoleValidation`, so `add_role` would persist them. My previous commit only stopped `in` from throwing; it did not make the malformed case a 400, which is what the comment I wrote claimed.

Fixed in 95e9caa03: non-object / null / array permissions are rejected up front with `PERMISSION_NOT_OBJECT`, and the per-key checks are skipped for them rather than running as no-ops.

The test change matters more than the code change. Asserting `to.not.throw()` cannot distinguish "rejected" from "silently accepted" — the two outcomes I most needed to tell apart — so the test agreed with a comment that described behavior the code did not have. They assert `statusCode === 400` now, and cover `[]`, `[{super_user:true}]`, `0` and `` alongside the original primitives, plus a positive case so the gate cannot pass by rejecting everything.

Comment generated by kAIle (Claude Opus 4.8)

unitTests/dataLayer/reservedSchemaNames.test.js
kylebernhardy: Confirmed — that control took the successful creation path and created `my_normal_db` on whatever root the environment had configured, then signalled a schema change. A unit test with a side effect on a developer checkout, and it passed while doing it, which is why nothing flagged it. Sorry about the EROFS noise in your run.

Fixed in 95e9caa03, validation-only per your first option: `dataLayer/schema.ts` now exports `validateDatabaseName()`, which runs the same `DB_NAME_CONSTRAINTS` the create paths run, with no IO. The control asserts a normal name returns no error, and a companion test asserts every reserved name still fails through that same function — otherwise the control could pass simply because the validation stopped working.

Comment generated by kAIle (Claude Opus 4.8)

Issue #595, opened 2026-05-19
Dropping a table leaves orphan blobs until the sweeper runs


kriszyp:
Dropping a table can leave orphan blobs on disk — blobs whose record references are no longer in any audit/active state but whose underlying files remain. Orphan-blob cleanup eventually removes them, but the gap between drop and cleanup is a leak.

## Reproduction (from Jira)

Cluster of two nodes (`ord01`, `ord02`):

1. Run a script that writes records containing blobs to `BlobHammerRun`.
2. Drop `BlobHammerRun` on `ord01`.
3. Wait for audit log to expire (without restarting Harper).
4. `ord01` audit is null (table dropped); `ord02` audit shows 1238 entries.
5. After audit expires: `ord01` audit null, `ord02` audit 2.
6. Restart Harper on both nodes.
7. `ord01` shows 528 orphan blobs (0 records, 2 audit entries); `ord02` shows 752 blobs with 1233 records and 2 audit entries.
8. Orphan-blob cleanup runs and deletes the 528 blobs from `ord01`.

Log evidence:
```
[http/2] [warn]: Checking for references to potential orphaned blobs in table BlobHammerRun
[http/2] [warn]: Checking for references to potential orphaned blobs in the audit log
[http/2] [warn]: Deleting 528 orphaned blobs
[http/2] [warn]: Finished deleting 528 orphaned blobs
[http/2] [warn]: Cleaned Orphan Blobs from BlobTest, deleted 528 blobs)
```

## Why this matters

The drop path should clean up the blobs it owns rather than leaving them for the orphan sweeper. Jira notes "we don't believe this has a production impact" — but it's the same root cause family as the broader [orphan blobs investigation](https://harperdb.atlassian.net/browse/CORE-3016), and the gap could matter on a node with many drops in flight or a slow orphan sweeper.

## Acceptance criteria

- Dropping a table eagerly deletes the blob files referenced only by records in that table (and only by audit entries that will not survive).
- The "wait for orphan sweep to find them" path remains as a safety net but is no longer the primary cleanup mechanism for drop-driven orphans.

## Related

- [CORE-3016](https://harperdb.atlassian.net/browse/CORE-3016) — Orphan blobs tracking & insights (lavinia's working notebook).

---

Tracked in Jira: [CORE-3003](https://harperdb.atlassian.net/browse/CORE-3003)

🤖 Filed by Claude on behalf of Kris.

kriszyp:
[...]

PR #1916, opened 2026-07-23
test(database): promote 6 QA eviction/removal/reclaim data-integrity anchors

kriszyp:
Test-only. Promotes six exploratory qa-explorer specs into `integrationTests/database/` as permanent regression anchors for the removal, eviction, and reclaim data-integrity paths. The current head is green across the full GitHub Actions build, format, lint, unit, and integration matrix.

| Spec | What it pins | Origin |
| --- | --- | --- |
| `eviction-index-orphan-removal-paths.test.ts` | **#1896 / F-149 shipped-fix anchor.** The merged fix's own test is `threads:1`, RocksDB-only, one scalar `@indexed` attribute, no concurrent writes during the sweep, and never exercises `delete()`/`evict()` directly. This pins the uncovered corners: `threads:{count:4}` contention, an indexed attribute rewritten while the sweep evicts neighboring rows, array-valued and null-valued indexed attributes, and all three removal call sites (TTL sweep / `delete()` / `evict()`). 8 tests across RocksDB and LMDB. | QA-661 / P-439 |
| `eviction-index-phantom-null-keys.test.ts` | **#1896 vs F-175.** Every removal path used to insert a phantom `[null, id]` index entry (`updateIndices(id, existing, null)` resolving the new value to `null` rather than absent, with `indexNulls` defaulting true). Four tables isolate the three removal call sites plus an update-in-place control that proves the oracle is not manufacturing phantoms. 14 tests across RocksDB and LMDB. | QA-670 / P-446 |
| `blob-restart-ttl-unlink.test.ts` | **Blob × restart-boundary × TTL.** Blob-backed records written pre-restart read back byte-identically; the post-restart expiration sweep evicts and physically unlinks their blob files; no dangling blob reference is observable during the sweep window; an already-past-expiry record is not stranded; and a post-restart update releases the old file via `cleanup_orphan_blobs` (deferred reclaim is by design, not a leak). `requires-isolation` because it restarts the instance. 7 tests. | QA-672 / P-452 |
| `audit-blob-reclaim.test.ts` | **Audit × blob-reclaim seam.** Overwriting a blob-valued record on an audited table reclaims exactly the prior blob file; a time-travel/history read of a reclaimed version returns a clean 404/ENOENT rather than stale bytes or the current version's bytes; repeated overwrites hold the live blob-file count flat; and delete-then-time-travel plus a `threads.count:4` arm behave identically. `requires-isolation` because it asserts a directory-global blob-file count. 5 tests. | QA-726 / P-500 |
| `blob-reclaim-removal-paths.test.ts` | **Per-path blob-reclamation ledger.** Every bulk-removal path must reclaim its own blob files without the manual `cleanup_orphan_blobs` sweeper: `drop_table`, `drop_database`, delete-by-search, overwrite with a new blob, and overwrite with a non-blob value. Each is measured against a per-record-delete control at `threads.count:4`. This qualifies **#595**: `drop_table` does reclaim on current main, so the known `Table.clear()` leak is a one-path outlier rather than a systemic reclamation gap. `requires-isolation` because it uses directory-global blob-file counts. 7 tests. | QA-720 / P-495 |
| `audit-retention-lmdb.test.ts` | **LMDB passive audit-retention purge.** With `logging.auditRetention` short and `HARPER_STORAGE_ENGINE=lmdb`, the self-rearming cleanup loop must purge aged audit entries without an operator calling `delete_transaction_logs_before`. The test hard-asserts the active engine, waits for the initial one-shot cleanup before seeding the measured cohort, proves the audit counter moves up, then requires the complete measured cohort to disappear. The RocksDB asymmetry is deliberately not asserted. 4 tests, approximately 18 seconds. | QA-746 / P-529 |

**Why these are not duplicative of the existing eviction tests:** the index specs read the raw secondary-index and primary stores directly, in both directions (index entry → missing base row and base row → missing index entry), rather than using `search_by_value`. That API joins through the primary record and skips on absence, so it cannot reveal a dangling index entry. The blob specs likewise walk `{dataRootDir}/blobs/` directly instead of trusting the API.

**Coverage added:** removal path × storage engine × worker count × indexed-attribute shape (scalar / array / null); blob × restart boundary × TTL expiration; audit/time-travel × blob reclaim; blob × five removal paths × four workers; and LMDB passive audit retention.

**Held back:** three other QA candidates are not in this diff and are not included in the count or coverage above:

- `audit-log-completeness-prune.test.ts` (QA-725 / P-499) exposed a Bun NAPI teardown crash after job-thread completion, so it cannot serve as a stable green anchor yet.
- `audit-purge-disk-reclaim.test.ts` (QA-779 / P-547) became a red characterization after its repaired mmap oracle found a deleted transaction log still pinned until restart.
- `cache-sourced-blob-ttl.test.ts` (QA-738 / P-510) produced a client transport failure in CI during its 40-reader blob stampede, with no matching server-side error. It remains a Stage-2 candidate rather than being retry-wrapped.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Claude Opus 5

<sub>Human-Review-Need: 4 @ ea68fb5c56af</sub>


kriszyp:
Added a sixth anchor: `cache-sourced-blob-ttl.test.ts` (promoted from qa-explorer QA-738).

It pins the `sourcedFrom()` × external-Blob × short-TTL seam, which nothing in the suite currently covers — the tracked `sourcedFrom` fixtures are all scalar-valued, and the existing blob/TTL tests churn blobs through a client write against a plain `@table(expiration:N)` rather than through the read-triggered revalidation path in `Table.ts`.

What it asserts:

- across sequential expire→revalidate cycles, every read returns complete, freshly-resolved bytes — the server-stored sha256 matches an independently recomputed sha256 of the streamed body — and on-disk blob files stay bounded rather than accumulating;
- a 40-way stampede on a just-expired key (resolver deliberately slowed) yields exactly one distinct value across all 40 readers, with zero partial bodies and zero dangling reads;
- the background sweep reclaims a cold key's blob file with no client re-read at all.

Both oracles are armed inside the test before any real assertion runs: the hash comparator is proven on known-different and known-identical pairs, and the disk walker is proven to flag a deliberately planted stray file. 5/5 green on two cold runs in the promoted location (~30 s), plus 5 consecutive clean runs during exploration.

One behavior observed and deliberately *not* asserted: source revalidation resolves and responds before the backing transaction durably commits (`Table.ts:5371` — "we don't want to wait for the transaction because we want to return as fast as possible"), so an ops-API `search_by_id` issued immediately after a successful blob GET can briefly miss the record. The client-facing bytes are already correct at that point, so this is a deliberate async-durability tradeoff rather than a staleness bug — noting it here because it makes naive read-after-write assertions flaky.

— KrAIs (Claude Opus 5)

integrationTests/database/blob-reclaim-removal-paths.test.ts
```diff
@@ -0,0 +1,438 @@
...
+		}
+	}
+	await walk(dir);
+	return { files, bytes };
+}
+
+/** Bounded convergence poll: sample `getUsage` until two consecutive reads agree, or time out. */
+async function waitForSettle(
+	getUsage: () => Promise<{ files: number; bytes: number }>,
+	{ timeoutMs = 20_000, intervalMs = 500 }: { timeoutMs?: number; intervalMs?: number } = {}
```
cb1kenobi: **Medium: `waitForSettle` polls at exactly the blob unlink delay, so it can settle on a stale (pre-unlink) plateau.**

The default `intervalMs = 500` equals `deletionDelay = 500` in `resources/blob.ts:849` (the delayed `setTimeout` unlink). `waitForSettle` returns on the first *two consecutive equal reads*: `prev` is sampled right after the op response, then the first `cur` lands ~500ms later — right at the unlink boundary. If the unlink hasn't fired yet at that second sample, `prev === cur` at the pre-unlink count and it returns `timedOut:false` with the stale higher `files`. Every arm using the default interval is exposed (CONTROL:194, C:234, D:276, E:330, A:367). Worst case is Arm A (line 383): a false plateau routes into the `DEFECT -- drop_table left N blob file(s) orphaned ... matching GH#595` verdict — a timing flake presented as a confirmed reclamation bug in a permanent anchor.

Suggested fix: decouple the poll cadence from the unlink delay and require the plateau to persist past it — e.g. `intervalMs` well above `deletionDelay` (say 1500ms), or poll until `files === <arm's known expected count>` with a timeout (each arm already knows its target).

—
_Generated by Barber AI_

integrationTests/database/audit-blob-reclaim.test.ts
```diff
@@ -0,0 +1,452 @@
...
+			notStrictEqual(shaA, shaC, 'sanity: blob A and blob C must be byte-distinct');
+			await sleep(2000);
+			const filesV3 = await fileSet(dir);
+			const removedAfterV3 = [...filesV2].filter((f) => !filesV3.has(f));
+			log(
+				`[overwrite] v3 written: sha=${shaC.slice(0, 12)}; files=${filesV3.size}; removed=${JSON.stringify(removedAfterV3)}`
+			);
+			ok(removedAfterV3.length > 0, 'blob B file must be reclaimed on second overwrite');
+			// With 3 live versions ever written and 2 reclaims, exactly 1 blob file should remain live.
+			strictEqual(filesV3.size, 1, 'exactly the current (v3/C) blob file should remain on disk');
```
cb1kenobi, line 270: **Medium: hard blob-file-count assertion after a fixed sleep races asynchronous GC.**

`strictEqual(filesV3.size, 1, ...)` fires on a single directory sample taken after a fixed `sleep(2000)` (line 232); the reclaim assertions at 226/238 and 440 have the same shape. Blob unlink is deferred (`deletionDelay`) and, per the comment at line 434, GC may run on a different worker under `threads.count:4` — so a slow/contended CI box can sample before GC settles and flake. This is the margin-sensitivity this batch is meant to remove, and the sibling `blob-reclaim-removal-paths.test.ts` already solves it with a bounded convergence poll.

Suggested fix: poll `fileSet(dir)` until the expected count is observed (or a deadline), rather than asserting a single post-sleep sample.

—
_Generated by Barber AI_

integrationTests/database/eviction-index-orphan-removal-paths.test.ts
```diff
@@ -0,0 +1,431 @@
...
+				count: HEARTBEAT_IDS.length,
+				prefix: 'HB',
+				bucket: 'HB-bucketA',
+				withTags: true,
+			});
+			strictEqual(res.status, 200, 'seed heartbeat rows');
+
+			const totalNonHeartbeat = EXPIRING_BUCKETS.length * ROWS_PER_BUCKET;
+			const preBase = await primaryDump('Expiring');
+			strictEqual(preBase.rows.length, totalNonHeartbeat + HEARTBEAT_IDS.length, 'all rows present pre-sweep');
```
cb1kenobi: **Medium: pre-sweep exact-count assertion can race the 5s TTL during seeding.**

The fixture sets `expiration: 5, scanInterval: 1`, and ~160 rows are seeded via awaited HTTP puts (on `threads:4`) before this exact-count check runs. If seeding plus the consistency computations exceed ~5s on a loaded runner, the earliest rows expire and the 1s sweep removes them, so `preBase.rows.length < expected` and this line fails spuriously. The consistency oracle below would still pass — it's only the raw count here that is fragile.

Suggested fix: give TTL headroom for setup (raise `expiration`, e.g. to 15, in the fixture), or assert `>= expected` here and rely on the consistency oracle for the real check.

—
_Generated by Barber AI_

integrationTests/database/audit-retention-lmdb.test.ts
```diff
@@ -0,0 +1,210 @@
...
+					if (last < low) low = last;
+					if (last <= peak - 200 || last === 0) break; // the 200 control records are gone
+					await sleep(2000);
+				}
+				lmdbFindings.push(
+					`2. after bounded passive wait (no explicit prune): peak=${peak} final=${last} lowestObserved=${low}`
+				);
+
+				ok(
+					last <= peak - 200 || last === 0,
```
cb1kenobi: **Low: purge threshold `peak - 200` assumes exactly one audit entry per record.**

The positive control (test 1) only asserts the count grew, never that the 200-record insert produced exactly 200 audit entries. If audit-entries-per-record is ever >1, a *partial* re-arm regression that purges only 200 of the cohort's entries still satisfies `last <= peak - 200` and false-passes; if <1, the `-200` target is unreachable and the test silently relies on the `=== 0` fallback.

Suggested fix: pin the threshold to a measured baseline — capture the pre-insert count and assert `last <= baseline` (whole cohort gone), independent of the entries-per-record mapping.

—
_Generated by Barber AI_

integrationTests/database/audit-log-completeness-prune.test.ts
```diff
@@ -0,0 +1,475 @@
...
+		ctx,
+		ids.map((id) => ({ id, seq: 2, val: 'v2' }))
+	);
+	const deletedIds = ids.filter((_, i) => i % 2 === 0);
+	await del(ctx, deletedIds);
+	return { ids, deletedIds };
+}
+
+function opOf(entry: any): string {
+	// ResourceBridge.readAuditLog maps internal 'put' -> 'upsert'; normalize back for readability.
```
cb1kenobi: **Nit: stale comment — `opOf` no longer normalizes anything.**

The comment says it maps internal `put` -> `upsert` and normalizes back, but the body is a plain identity (`return entry.operation`). Q1's op-sequence assertion depends on the literal `insert`/`update`/`delete` values, so the comment is misleading dead prose.

Suggested fix: drop the comment (or actually normalize if that was the intent).

—
_Generated by Barber AI_

integrationTests/database/eviction-index-phantom-null-keys.test.ts
```diff
@@ -0,0 +1,429 @@
...
+		let client: ReturnType<typeof createApiClient>;
+		let httpURL: string;
+		let auth: string;
+
+		before(async () => {
+			await setupHarperWithFixture(ctx, FIXTURE_PATH, {
+				config: {
+					threads: { count: 1 },
+					logging: { console: true, level: 'error' },
+					storage: { engine },
```
cb1kenobi: **Low: config-only engine forcing is silently overridden by an inherited `HARPER_STORAGE_ENGINE`, unlike the sibling spec.**

`resources/databases.ts:862` resolves the engine as `process.env.HARPER_STORAGE_ENGINE || envGet(CONFIG_PARAMS.STORAGE_ENGINE)`, and the harness spawns Harper with `{ ...process.env, ...options.env }` — so the env var wins over `config.storage.engine`. Running the lmdb matrix the way this file's own header used to document it puts both arms on LMDB and breaks the `rocksdb` arm's own precondition. Reproduced locally on this head:

```
$ HARPER_STORAGE_ENGINE=lmdb npm run test:integration -- --isolation=none \
    integrationTests/database/eviction-index-phantom-null-keys.test.ts
  tests 14 | pass 13 | fail 1
  AssertionError: PRECONDITION: engine in effect must be rocksdb, got {... "engineGuess":"lmdb"}
```

(Plain run is 14/14.) `eviction-index-orphan-removal-paths.test.ts:63` already forces it the authoritative way — `env: { HARPER_STORAGE_ENGINE: engine }` — and passes 8/8 under the same parent env. Matching it here makes the forcing actually win and the two siblings consistent.

Suggested fix: drop `storage: { engine }` from `config` and pass `env: { HARPER_STORAGE_ENGINE: engine }` to `setupHarperWithFixture` instead.

—
_Generated by Barber AI_

integrationTests/database/blob-reclaim-removal-paths.test.ts
```diff
@@ -0,0 +1,496 @@
...
+ * accepting any two consecutive equal reads) avoids settling on a stale pre-unlink plateau when
+ * `intervalMs` happens to land near the unlink boundary.
+ */
+async function waitForSettle(
+	getUsage: () => Promise<{ files: number; bytes: number }>,
+	expectedFiles: number,
+	{
+		timeoutMs = 20_000,
+		intervalMs = 250,
+		minStableMs = 1_500,
```
cb1kenobi: **Low: `minStableMs = 1500` gives up sooner than the fixed sleeps it replaced, so a slow reclaim reports a false leak.**

`stableSince` is seeded with `start`, so the plateau break can fire 1.5 s after the *first* sample — only 1 s past `deletionDelay = 500` (`resources/blob.ts:849`), and shorter than the `sleep(2000)` / `sleep(2500)` this replaced (the 2500 ms one was explicitly widened because "4 worker threads means GC may be scheduled on a different thread"). If the deferred unlink hasn't landed by then, `waitForSettle` returns the stale pre-unlink count and the arm asserts a `DEFECT` verdict on a timing artifact — the same false-positive class the poll was introduced to remove, just in the other direction.

The loop already returns the instant `cur.files === expectedFiles`, so a larger `minStableMs` costs nothing on a green run — only on a genuinely failing one, where the bounded `timeoutMs` (10–30 s) is the real budget. Same default at `audit-blob-reclaim.test.ts:138`.

Suggested fix:
```suggestion
		minStableMs = 5_000,
```

—
_Generated by Barber AI_

integrationTests/database/eviction-index-phantom-null-keys/resources.js
```diff
@@ -0,0 +1,122 @@
...
+// GET /StorageEngineInfo/?table=X -> code-derived signal for which storage engine is actually
+// backing this table (do NOT just trust the HARPER_STORAGE_ENGINE env var was honored).
+export class StorageEngineInfo extends Resource {
+	static loadAsInstance = false;
+	async get(query) {
+		const t = getTable(qget(query, 'table') || 'DelTable');
+		const primaryPath = t.primaryStore?.path || t.primaryStore?.rootStore?.path || null;
+		const indexHasPrefetch = !!t.indices?.bucket?.prefetch;
+		const looksLikeLmdbPath = typeof primaryPath === 'string' && primaryPath.endsWith('.mdb');
+		const isLmdb = looksLikeLmdbPath || indexHasPrefetch;
```
cb1kenobi: **Nit: `engineGuess` fails open to `rocksdb`, so the rocksdb arm's "hard assert" is not actually hard.**

Both signals are optional-chained (`t.primaryStore?.path`, `t.indices?.bucket?.prefetch`). If either internal shape moves, both go false, `isLmdb` is `false`, and this reports `rocksdb` for *any* engine — so test 0's precondition in the rocksdb arm passes vacuously, which is exactly the "cannot pass by accident" property it exists to provide. Only the lmdb arm would notice.

Suggested fix: return `engineGuess: 'unknown'` when neither signal is observable (no `primaryPath` and no `indices.bucket`), so test 0 fails on an unrecognized shape rather than defaulting to a pass. Same in `eviction-index-orphan-removal-paths/resources.js:105`.

—
_Generated by Barber AI_

integrationTests/database/blob-restart-ttl-unlink.test.ts
```diff
@@ -0,0 +1,678 @@
...
+// runRecordExpirationEviction (Table.ts) runs on RECORD_PRUNING_INTERVAL=60s, per-process
+// (plain setInterval from table-load time, not epoch-aligned) -> give a first-pass a full
+// interval plus margin.
+const SWEEP_WAIT_BUDGET_MS = 140_000; // ~2.3x RECORD_PRUNING_INTERVAL, covers 2 sweep passes
+const SWEEP_POLL_INTERVAL_MS = 3_000;
+
+const HARPER_CONFIG = {
+	threads: { count: 1 }, // single worker: satisfies both getWorkerIndex()===0 (per-record TTL
+	// sweep) and ===getWorkerCount()-1 deterministically
+	storage: { engine: 'lmdb' as const }, // F-174: per-record @expiresAt is dead on RocksDB
```
cb1kenobi, line 82: **Low: this LMDB pin is overridable by an inherited `HARPER_STORAGE_ENGINE`, and nothing here would notice.**

Same mechanism as the note on `eviction-index-phantom-null-keys.test.ts:96` — `resources/databases.ts:862` reads `process.env.HARPER_STORAGE_ENGINE || envGet(CONFIG_PARAMS.STORAGE_ENGINE)`, and the harness spawns Harper with `{ ...process.env, ...options.env }`, so the env var beats `config.storage.engine` on both the `setupHarperWithFixture` (line 181) and the post-restart `startHarper` (line 327).

The consequence is worse here than in the sibling spec, because this file has no engine precondition test: running the suite with `HARPER_STORAGE_ENGINE=rocksdb` in the parent env lands on the exact engine the comment says is broken (F-174, per-record `@expiresAt` eviction dead on RocksDB), so Q2 emits `DEFECT — N/N normal-N record(s) never evicted` as a real finding rather than a misconfiguration.

Suggested fix: pass `env: { HARPER_STORAGE_ENGINE: 'lmdb' }` at both call sites so the pin wins regardless of the parent env, and/or add the same test-0 engine precondition the eviction specs now carry.

—
_Generated by Barber AI_

integrationTests/database/eviction-index-phantom-null-keys.test.ts
kriszyp: Fixed in 4389a1d22: each fixture launch now sets HARPER_STORAGE_ENGINE in the child environment, so an inherited parent value cannot override the requested matrix leg. Verified locally with a deliberately hostile parent setting. — GPT-5 Codex

integrationTests/database/eviction-index-orphan-removal-paths.test.ts
```diff
@@ -0,0 +1,465 @@
...
+						const tags = [`${bucket}-tag${heartbeatRound % 3}`, `${bucket}-tagX`];
+						const res = await postJSON('/Heartbeat/', { table: 'Expiring', ids: HEARTBEAT_IDS, bucket, tags });
+						const text = await res.text();
+						if (res.status !== 200) {
+							throw new Error(`Heartbeat round ${heartbeatRound} failed: ${res.status} ${text}`);
+						}
+						heartbeatRound++;
+						await sleep(400);
+					}
+				})();
```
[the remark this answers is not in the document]
kriszyp: Fixed in 4389a1d22: the heartbeat promise now attaches its rejection handler immediately, records any failure, joins during cleanup, and rethrows after cleanup. The follow-up also refreshes the heartbeat cohort immediately before measurement. — GPT-5 Codex

integrationTests/database/blob-restart-ttl-unlink.test.ts
kriszyp: Fixed in 4389a1d22: HARPER_STORAGE_ENGINE=lmdb is forced in the child environment for both initial startup and restart. The suite passes 7/7 even with a hostile RocksDB value inherited by the test runner. — GPT-5 Codex

integrationTests/database/blob-reclaim-removal-paths.test.ts
kriszyp: Fixed in 4389a1d22: both blob settle helpers now require a 5-second stable plateau before declaring convergence. The focused suites pass with the longer window. — GPT-5 Codex

integrationTests/database/eviction-index-phantom-null-keys/resources.js
kriszyp: Fixed across 4389a1d22 and 1aa7ec5e0: engine detection now fails closed to unknown and identifies RocksDB only from the positive purgeLogs capability signal; LMDB still uses its path/prefetch signals. Both engine legs pass. — GPT-5 Codex

integrationTests/database/audit-blob-reclaim.test.ts
```diff
@@ -0,0 +1,488 @@
...
+ * pre-reclaim count under a loaded runner; polling toward the known target (or a real plateau)
+ * avoids that race.
+ */
+async function waitForFileSetSettle(
+	dir: string,
+	expectedCount: number,
+	{
+		timeoutMs = 10_000,
+		intervalMs = 250,
+		minStableMs = 5_000,
```
cb1kenobi, line 137: **Low: raising `minStableMs` to 5s without raising this helper's 10s `timeoutMs` leaves only one stability window, and the leak-check arm can never exit early.**

Thanks for widening this — it's the right fix for the four target-matching call sites, which exit on `cur.size === expectedCount` before `minStableMs` is ever consulted. But line 344 calls `waitForFileSetSettle(dir, Infinity)`, where the size can never equal the target, so the plateau is the *only* exit path. Two consequences there:

1. With `timeoutMs = 10_000`, a single file-set change after t=5s resets `stableSince` and the deadline wins before another 5s window can close — the function then returns a set that never plateaued. Unlike the sibling `waitForSettle` in `blob-reclaim-removal-paths.test.ts`, this one returns no `timedOut` flag, so a deadline exit is indistinguishable from a genuine plateau and the leak-bounds assertion below reads an unsettled count.
2. Runtime: the leak-check arm loops `N = 8` overwrites, each now waiting the full 5s plateau instead of 1.5s — roughly +28s. Measured locally at this head, the `threads:1` suite that contains this arm now takes **50.4s** against the ~22s the PR body documents for the whole spec.

Simplest fix is to give this helper the same budget as its sibling, so 5s of stability always fits:

```suggestion
		timeoutMs = 20_000,
		intervalMs = 250,
		minStableMs = 5_000,
```

If the added runtime matters more than the margin, the alternative is to pass a shorter `minStableMs` explicitly at the `Infinity` call site on line 344 and leave the default at 5s for everyone else.

—
_Generated by Barber AI_

kriszyp: Addressed in 6279f8392 by replacing the unreachable `Infinity` target with the arm’s actual invariant: this database contains one record and therefore must settle at exactly one live blob. The poll now exits immediately when reclaim reaches that count, while a slow or leaking path still receives the existing 5-second plateau/10-second bound. Two focused runs passed 5/5; the leak arm settled at `[1,1,1,1,1,1,1,1]` in ~3.6s. — GPT-5 Codex

cb1kenobi:
Reviewed `6279f839` for the incremental update from `1aa7ec5` - no issues found. The change uses the known live blob-file target for the leak check, which removes the `Infinity` plateau ambiguity and avoids the extra stability wait that the prior thread flagged. _Generated by Barber AI_

integrationTests/database/audit-blob-reclaim.test.ts
```diff
@@ -0,0 +1,488 @@
...
+		}
+	}
+	await walk(dir);
+	return out;
+}
+
+/**
+ * Bounded poll: settle on `dir`'s blob file set once it reaches `expectedCount` (pass Infinity
+ * when there is no known target — e.g. a leak-bounds check — to rely purely on stability), or
+ * once the count plateaus for at least `minStableMs` (well past blob.ts's deletionDelay). A fixed
```
cb1kenobi: **Nit: the docstring still recommends the `Infinity` mode that this push just removed.**

Switching line 344 to `waitForFileSetSettle(dir, 1)` was the right fix — measured here, the suite went 50.4s -> 16.4s and the leak-check arm 40s -> 3.6s, with the series still converging to `[1,1,1,1,1,1,1,1]`, so detection is unchanged. But no call site passes `Infinity` any more, and this parenthetical still points the next reader at it for exactly the "leak-bounds check" use case that was just converted. That path is the hazardous one: with no target to match, the plateau is the only exit, and if `timeoutMs` (10s) wins first the helper returns a never-plateaued set with no `timedOut` flag — unlike the sibling `waitForSettle` in `blob-reclaim-removal-paths.test.ts:140`, which returns one.

Suggested fix:
```suggestion
 * Bounded poll: settle on `dir`'s blob file set once it reaches `expectedCount`, or once the
 * count plateaus for at least `minStableMs` (well past blob.ts's deletionDelay). A fixed
```

—
_Generated by Barber AI_

cb1kenobi:
Reviewed `1c2b30b4` — the docstring-only update.

**The prior nit is fixed.** `waitForFileSetSettle`'s docstring no longer points the next reader at the `Infinity` mode, and no call site passes it (the only remaining hit in the tree is the unrelated `sql-pk-range-fusion.test.ts:81`). No executable change, so no regression surface. No new code findings this round.

**Prior fixes all still hold:** engine pinned via `env:` at all five sites — including the `startHarper` restart call at `blob-restart-ttl-unlink.test.ts:343`; `minStableMs = 5_000` in both settle helpers; three-way `engineGuess` with an explicit `'unknown'` default in all three `resources.js`.

**Tests:** built this head and ran the six promoted specs locally (`--isolation=none`) — **45/45 passing, 0 failures**: audit-blob-reclaim 5/5 (~16.5s), audit-retention-lmdb 4/4, blob-reclaim-removal-paths 7/7, blob-restart-ttl-unlink 7/7, eviction-index-orphan-removal-paths 8/8 (both engines), eviction-index-phantom-null-keys 14/14 (both engines).

---

One non-code item is still outstanding. It's been raised twice before but only out-of-band, so posting it here to make sure it's actually visible:

**The PR advertises coverage it doesn't ship.** The title and the body's opening line both say **8** anchors; **6** spec files are in the diff. Two rows in the body table describe files that are not part of this PR:

- `audit-log-completeness-prune.test.ts` (QA-725 / P-499)
- `audit-purge-disk-reclaim.test.ts` (QA-779 / P-547)

The **Coverage added** line likewise claims `audit-completeness × prune-boundary × storage-engine × worker-count` and `audit-purge × disk-reclamation × restart-boundary` — neither ships — while omitting the LMDB passive-retention coverage that does. The "a seventh spec … was pulled" batch note is inconsistent with a count of 8 as well.

This isn't a defect in the code, and it doesn't block the merge on correctness grounds. But the title becomes the merge commit and the body becomes the record of what landed, and as written both would tell a future reader that audit-prune-completeness and `/proc`-level disk-reclamation regression anchors exist in the suite when they don't — which is exactly the kind of thing someone greps for later before writing a new one. Worth retitling to 6 and either trimming those two rows or moving them into a "held back" note next to the `cache-sourced-blob-ttl` one.

Minor, while you're in there: the per-row test counts have drifted — `audit-retention-lmdb` says "3 tests, ~12.6 s" but now runs 4 in ~17.7 s (the `0.5. observe the initial cleanup` test isn't reflected).

—
_Generated by Barber AI_

integrationTests/database/audit-blob-reclaim.test.ts
```diff
@@ -0,0 +1,487 @@
...
+				log(`[leak-check] after write v${i}: live blob files=${files.size}`);
+			}
+
+			log(`[leak-check] file-count series across ${N} versions: ${JSON.stringify(counts)}`);
+			const maxCount = Math.max(...counts);
+			const finalCount = counts[counts.length - 1];
+			// If audit retention pinned every historical blob, count would grow ~linearly to N.
+			// Expect it to stay bounded at (or very near) 1 live file per record.
+			ok(
+				finalCount <= 2,
```
heskew: **Low:** This only gates the final sample, so intermediate growth such as `[1,2,3,4,5,6,7,1]` passes; `finalCount === 0` also passes even if the current blob was reclaimed. Please assert the bound across every settled sample (or `maxCount`), require the live file, and re-read the final record/SHA so this arm cannot green on current-blob loss.

---

integrationTests/database/audit-blob-reclaim.test.ts
```diff
@@ -0,0 +1,487 @@
...
+					search_values: [id],
+				})
+				.expect(200);
+			const auditTxns = auditResp.body[id] || auditResp.body[String(id)] || [];
+			log(
+				`\n[overwrite] read_audit_log(hash_value=${id}) txn count: ${Array.isArray(auditTxns) ? auditTxns.length : 'n/a'}`
+			);
+			log(`[overwrite] read_audit_log raw body: ${JSON.stringify(auditResp.body, null, 2).slice(0, 4000)}`);
+			ok(
+				Array.isArray(auditTxns) && auditTxns.length > 0,
```
heskew: **Low:** The earlier non-empty-transaction guard still lets this pass when each transaction has an empty `records` array, or when only the current version is returned. Since this arm wrote v1/v2/v3, require all expected history entries and one blob-valued record per entry before checking placeholder shape; otherwise `contentShapes` can be empty and the placeholder assertion passes vacuously.

---

integrationTests/database/blob-restart-ttl-unlink.test.ts
```diff
@@ -0,0 +1,695 @@
...
+		test(
+			'post-sweep decisive orphan check: cleanup_orphan_blobs converges to exactly the live set',
+			{ timeout: 60_000 },
+			async () => {
+				const preCleanup = await diskFiles(blobRootDir);
+				findings.push(`pre-cleanup_orphan_blobs disk: ${preCleanup.files} files`);
+
+				const cleanupResp = await client
+					.req()
+					.send({ operation: 'cleanup_orphan_blobs', database: 'data' })
```
heskew: **Low:** Running `cleanup_orphan_blobs` before asserting disk convergence can hide a regression where TTL eviction stops naturally reclaiming these eight blob files—the manual sweep removes them and the final `filesAfter === 2` gate still passes. Poll and preserve the natural post-eviction result first; invoke manual cleanup only afterward as a diagnostic for a failing natural result.

---

integrationTests/database/eviction-index-phantom-null-keys.test.ts
```diff
@@ -0,0 +1,428 @@
...
+			await sleep(2_500);
+			// Guard against the wrong-path race: Table.ts's scheduleCleanup() aligns its FIRST timer to
+			// the next wall-clock interval boundary counted from the start of the year (see
+			// resources/Table.ts scheduleCleanup — `nextScheduled = ceil((now - startOfYear) /
+			// interval) * interval + startOfYear`), not 300s from server start. Depending on where the
+			// test happens to land in that 5-minute cycle, the "never fires in this window" background
+			// sweep could in principle beat the GETs below to it, evicting via the wrong call site while
+			// this test still credits the lazy-eviction path. Assert the rows are still here immediately
+			// before triggering the lazy path, so that race fails loudly instead of silently passing.
+			const preGet = await dump('EvictTable');
```
heskew, line 333: **Low:** This guard narrows the known wall-clock race but does not isolate the lazy path. If the aligned five-minute sweep fires before `preGet`, correct code fails; if it fires after `preGet` while the 20 GETs run, the arm can pass while crediting the background sweep as lazy eviction. Disable/control the periodic sweep for this arm, or seed only after a boundary with verified margin through the GET loop.

---

integrationTests/database/blob-restart-ttl-unlink.test.ts
```diff
@@ -0,0 +1,695 @@
...
+} from '@harperfast/integration-testing';
+// @ts-expect-error utils/client.mjs has no type declarations; runtime resolves fine
+import { createApiClient } from '../apiTests/utils/client.mjs';
+
+const FIXTURE_PATH = resolve(import.meta.dirname, 'blob-restart-ttl-unlink');
+
+// ── Tunables ─────────────────────────────────────────────────────────────────
+const BLOB_SIZE = 64 * 1024; // > 8KB FILE_STORAGE_THRESHOLD -> always file-backed
+const LONG_TTL_MS = 15 * 60 * 1000; // 15min: survivor/update-target must outlive the whole test
+const NORMAL_TTL_MS = 8_000; // due shortly AFTER restart, NOT yet due at kill time (armed below)
```
heskew: **Low:** This TTL starts aging before five normal writes, three already-expired writes, disk walks, and raw checks complete. A loaded runner can legitimately consume the eight-second window before the pre-restart liveness check and fail correct code. Seed this cohort last with verified remaining TTL, or give it enough headroom while keeping the later 140-second sweep assertion bounded.

---

heskew (approved):
Reviewed `1c2b30b41213` against live `main`. The branch merges cleanly and the current checks are green.

Approving with five Low, non-blocking test-oracle/timing comments. They are distinct from the resolved threads; two are follow-ups where earlier guards narrowed but did not fully close the original gap. Kris can take or defer them.

No PR-controlled code was executed during this review.

---

integrationTests/database/blob-restart-ttl-unlink.test.ts
```diff
@@ -0,0 +1,729 @@
...
+			const restartT0 = Date.now();
+			await startHarper(ctx, { config: HARPER_CONFIG, env: { HARPER_STORAGE_ENGINE: 'lmdb' } });
+			registerLogPaths();
+			client = createApiClient(ctx.harper);
+			await waitRouteReady(120_000);
+			findings.push(`phase2: restart + route-ready completed in ${Date.now() - restartT0}ms`);
+			for (let i = 0; i < NORMAL_COUNT; i++) {
+				const normal = await op({ action: 'raw', id: `normal-${i}` }).expect(200);
+				ok(
+					normal.body.rawPresent === true && normal.body.pastDue === false,
```
cb1kenobi: **Low: this `pastDue === false` arm makes the restart's own wall-clock cost part of the pass condition.**

`NORMAL_TTL_MS` went 8s -> 60s and phase 1 now arms `remainingNormalTtlMs >= MIN_PRE_RESTART_TTL_MS` (30s, line 341). Phase 2 then requires every `normal-i` to be *not yet due* after the restart -- but this same test tolerates a restart of up to 120s (`waitRouteReady(120_000)` on line 357, and the phase-2 test timeout is `120_000`). If the restart outruns the remaining TTL, phase 2 fails even though the behavior under test is fine: Q2 asks whether the **post-restart** sweep evicts + unlinks old-version records, and a record that comes due *during* the restart is still evicted by that sweep.

Measured on this head locally: `remainingNormalTtlMs = 59953`, restart 775ms -- so the practical margin is ~60s rather than the 30s floor, which is why this is Low rather than Medium. It needs a pathologically slow restart to bite.

There is a secondary cost: waiting out the 60s TTL is now the suite's dominant runtime. `Q2/Q3/Q4` alone measured **60.2s** and the whole file **64.0s** (7/7 pass), on a spec that is meant to be a permanent CI anchor.

Suggested fix: assert only residency here -- that is the precondition Q2 actually needs -- and let the existing `SWEEP_WAIT_BUDGET_MS` poll decide when the records come due. That also frees `NORMAL_TTL_MS` to come back down and gives back most of the added minute.

```suggestion
					normal.body.rawPresent === true,
					`normal-${i} must still be resident after restart: ${JSON.stringify(normal.body)}`
```

—
_Generated by Barber AI_

integrationTests/database/audit-blob-reclaim.test.ts
```diff
@@ -0,0 +1,506 @@
...
+			ok(Array.isArray(auditTxns), `read_audit_log must return an audit transaction array for ${id}`);
+			strictEqual(
+				auditTxns.length,
+				3,
+				`read_audit_log must return all three writes for ${id}, got ${JSON.stringify(auditTxns)}`
+			);
+			const expectedTitles = ['BLOB-A-MARKER', 'BLOB-B-MARKER', 'BLOB-C-MARKER'];
+			const auditedRecords = auditTxns.flatMap((transaction: any, index: number) => {
+				strictEqual(
+					transaction.records?.length,
```
cb1kenobi: **Nit: this assertion is tautological -- it can never fail.**

`ResourceBridge.readAuditLog`'s `HASH_VALUE` branch builds every transaction as `records: [auditRecord.value]` (`dataLayer/harperBridge/ResourceBridge.ts:527-536`) -- a literal one-element array. So `records.length === 1` is a property of the response mapper, not of anything this spec is pinning, and it holds regardless of how the audit/blob path behaves.

The neighbouring `expectedTitles[index]` check is the one doing real work, and its ordering assumption is sound: `getHistoryOfRecord` splices descending within each audit window and then `.reverse()`s, so the returned array is ascending by version.

Suggested fix: drop it, or swap it for something the mapper does not guarantee -- e.g. `strictEqual(transaction.ids?.[0], id)`.

—
_Generated by Barber AI_

integrationTests/database/blob-restart-ttl-unlink.test.ts
kriszyp: Addressed in 6fa1f4afe: phase 2 now gates only physical residency, so restart duration cannot make correct behavior fail an unexpired predicate. The test still records each normal record’s `pastDue` state in the findings output; the final focused run observed `[false,false,false,false,false]`. — GPT-5 Codex

integrationTests/database/audit-blob-reclaim.test.ts
kriszyp: Addressed across d2044b950 and 6fa1f4afe: the mapper-guaranteed records-length assertion is gone, and the replacement checks `transaction.records?.[0]?.id`, so ownership comes from the stored audit image rather than echoed request metadata. Optional chaining also yields the targeted assertion failure before the title check. — GPT-5 Codex

cb1kenobi:
Reviewed `6fa1f4af` (incremental from `14a82080`) — no issues found. Both prior notes are genuinely fixed, and I re-ran the suites plus the same mutation check to confirm the anchors still bite.

**`blob-restart-ttl-unlink.test.ts:362` — the wall-clock coupling is removed, not merely widened.** The `pastDue === false` conjunct is gone from the pass condition entirely, and the value is now collected into `normalPastDue[]` and pushed to `findings`, which `after()` dumps unconditionally — so the observation survives as diagnostics with zero effect on pass/fail. I checked nothing downstream quietly went vacuous: Q3's mirror-failure poll only watches `survivor`/`update-target` (LONG_TTL), and Q2/Q4's decisive checks assert `rawPresent === false` for every `normal-N`/`already-N`, so none of them relied on `normal-N` being not-yet-due at phase-2 exit.

**`audit-blob-reclaim.test.ts:316` — replaced with something that can actually fail, not deleted.** Nice catch going for `records[0].id` rather than the `ids[0]` I suggested: `ids: [id]` is copied straight off the request, so my version would have been just as tautological. `records[0]` is `auditRecord.value`, so this now reads the persisted record's own primary key and a `getHistoryOfRecord` regression returning the wrong record's value would fail here. No coverage lost either — the transaction-level count is still hard-asserted 12 lines up (`auditTxns.length === 3`), and the optional chaining keeps the non-empty guard that the following `records[0].title` line was leaning on.

Verified at this head:
- `audit-blob-reclaim.test.ts` — 5/5 pass, 16.1s.
- `blob-restart-ttl-unlink.test.ts` — 7/7 pass, 64.1s.
- Mutation check (`deleteBlob()` -> no-op in `resources/blob.ts`, rebuilt): still 5 pass / **2 fail**, the same two tests as before — `post-sweep decisive orphan check` and `Q5`. The diagnostic block still fails the test while distinguishing "natural reclaim broken, operator sweep still works", and the new `phase2: normal-N pastDue after restart = [false,false,false,false,false]` line printed in both runs. Patch reverted.

One thing left on the table, non-blocking and purely cost: `NORMAL_TTL_MS` is still 60s, and `Q2/Q3/Q4` is 60.3s of the suite's 64.1s. Now that phase 2 no longer asserts `pastDue`, nothing correctness-related needs that 60s — only phase 1's own 30s seeding-headroom floor, which cleared with 59.9s of margin against a 120ms phase 1. Worth a look if you want the minute back on a permanent CI anchor, but happy either way.

—
_Generated by Barber AI_

kriszyp:
Measured the suggested TTL reduction before changing it: with `NORMAL_TTL_MS=60_000`, Q2/Q3/Q4 took 60.407s; with `35_000`, it took 60.385s. `runRecordExpirationEviction()` starts a fresh 60-second interval after the restart, so a shorter record TTL cannot advance the first eligible sweep—it only removes phase-1 seeding headroom. I kept the 60-second TTL and documented that scheduler invariant in `ea68fb5c5`.

— GPT-5 Codex

cb1kenobi (approved):

PR #1921, opened 2026-07-23
feat(mcp/openapi): drive schemas from a programmatic Resource's static properties

### LINKED ISSUE #1920: [MCP] #1095 follow-ups: programmatic static-properties schema wiring + MCP/convergence test coverage
Follow-up to #1095, which was delivered by #1167 and closed. The production functionality (GraphQL docstrings → MCP + OpenAPI descriptions, the operations catalog, `@hidden`, `outputSchema`, per-attribute descriptions, OpenAPI parity) is live. This issue tracks the remainder from the #1095 acceptance list that #1167 did not fully cover — see the audit comment on #1095 for the full checkbox-by-checkbox verdict.

## 1. Programmatic-Resource `static properties` → MCP input schema

A bare `static properties = {…}` literal on a Resource class is **not** read by MCP or OpenAPI schema derivation — both read `.attributes` (`components/mcp/tools/application.ts:1359`, `resources/openApi.ts:112`). `static properties` only enriches schemas once compiled through `defineResource`/`withSchema` (`resources/defineResource.ts:607-610`). Consequences:

- A pure programmatic Resource declaring only `static properties` (no attributes, no contract) yields a skeletal MCP `inputSchema` — the "skeletal schema for programmatic Resources" gap #1095 set out to close, still open for the bare-class path.
- There is **no lazy `Resource.properties` getter**; the attributes→properties projection is eager at `Table` construction (`resources/Table.ts:278`, `resources/databases.ts:1157`). Back-compat co-population works and is tested, but item "`Resource.properties` getter projects from `Resource.attributes` lazily" (acceptance) is unmet as written.
- No `static properties` **inheritance / spread-override** fixture (kriszyp's `CustomProduct` example). Static-field inheritance works natively in JS, but nothing verifies it, and it's moot until `static properties` feeds MCP derivation.

**Decide:** is the intended authoring path `defineResource`/`withSchema` only (in which case document that and drop the bare-literal acceptance items), or should a bare `static properties` literal also drive MCP/OpenAPI schemas? kriszyp locked in `static properties` as the canonical public API in the #1095 discussion, which points at the latter.

## 2. MCP-side + convergence test coverage

The OpenAPI side and the GraphQL parser are tested; the MCP assertions and the cross-surface convergence tests were never written:

- **Missing** — `tools/list` against a docstring-carrying fixture asserting the tool `description` includes the docstring **and** `inputSchema.properties[*].description` carry attribute descriptions. (`unitTests/components/mcp/tools/application.test.js` asserts only names/verbs; its `.description` assertions are for custom `mcpTools`.) No fixture under `integrationTests/fixtures` contains `"""` docstrings.
- **Missing** — convergence test: a Resource with descriptions surfaces them in **both** MCP `tools/list` **and** OpenAPI `/openapi.json`.
- **Missing** — a `@table` Resource with docstrings (no override) produces identical MCP + OpenAPI output to one with explicit `static description` + `static properties` (both paths converge).
- **Ineffective** — the idempotent guard (`unitTests/components/mcp/tools/operations.test.js:386-391`) only checks `!('idempotentHint' in OPERATION_DESCRIPTIONS)` (a string map); it never inspects an emitted `add_user`/`create_*` tool's `annotations`. Behavior is correct today (`IDEMPOTENT_OPERATIONS` ships empty), but nothing would catch a future mis-population — the exact safety this check was meant to provide. Rewrite it to assert the real emitted annotation is absent.
- **Partial** — operations-profile test asserts description *coverage* but not that a specific curated string lands on the right tool; back-compat is tested per-surface but not as a unified golden assertion.

## Out of scope
- Search/list return envelope → tracked in #1107.

### LINKED ISSUE #1941: [MCP/OpenAPI] hidden is not honored on nested sub-properties, and leaks as a schema key in OpenAPI
Follow-up from the docs pass for #1923 ([documentation#605](https://github.com/HarperFast/documentation/pull/605)).

## Problem

`@hidden` / `hidden: true` is documented as the way to keep a field out of the MCP tool descriptors and the OpenAPI document — the latter matters because `/openapi.json` has no per-user filtering, so anything reaching it is visible to anyone who can hit the HTTP port. That suppression is implemented **only at the top level**:

**MCP** — `attributeVisible` is consulted in the top-level loops (`components/mcp/tools/schemas/derive.ts:166`, `:208`, `:319`), but `attributeToProperty` recurses into `attr.properties` with no visibility check, so a hidden sub-property is emitted with its `description`.

**OpenAPI** — `resources/openApi.ts:142` does `if (hidden) continue;` for top-level attributes only. The nested-object branch calls `attributeToFragment`, which does not suppress and instead **emits the flag as a schema key**:

```ts
// resources/jsonSchemaTypes.ts:119
if (attr.hidden) fragment.hidden = true;
```

So a hidden nested sub-property appears in the public OpenAPI document along with its description *and* a `"hidden": true` marker advertising that it was meant to be suppressed.

## Why it matters

The docs tell authors to use `@hidden` for fields that should not surface publicly ("don't put secrets, internal-only commentary, or speculative prose in docstrings; use `@hidden` to suppress"). For a nested field that guidance currently does not hold. This is narrow — it needs a nested object whose sub-field is marked hidden — but the failure is silent and in the direction of disclosure.

Note the exposure is bounded: table-backed GraphQL nested types usually take the `$ref` branch in `openApi.ts`, so this mainly affects programmatic `static properties` objects and any table path that reaches the `attr.properties` branch.

## Suggested fix

1. Filter in the recursion: skip `hidden` sub-attributes in `attributeToProperty` (`derive.ts`) and in `attributeToFragment`'s `properties` loop (`jsonSchemaTypes.ts`), rather than only in the top-level loops.
2. Stop emitting `hidden` as a JSON Schema key at all — it is a Harper directive, not JSON Schema vocabulary, and it does not belong in an output document. (Check consumers first: `Table.properties` round-trips through `fragmentToAttribute`, which reads it back, so the key may need to stay in the canonical projection while being stripped on the MCP/OpenAPI emit paths.)
3. Add unit coverage for a hidden sub-property on both surfaces — the existing tests only assert top-level suppression.

Once fixed, the "top-level only" caveats in `reference/resources/resource-api.md` need to come back out.

### LINKED ISSUE #1942: [MCP/OpenAPI] The two schema type mappers disagree: unrecognized types, nested enum/format/const, and nullability
Follow-up from the docs pass for #1923 ([documentation#605](https://github.com/HarperFast/documentation/pull/605)). Writing the fragment-key reference table required documenting the same key twice, once per surface, because the two mappers diverge in three places. Each divergence is defensible in isolation; together they mean an author cannot reason about `static properties` without knowing which consumer is reading.

#1921 already extracted `JSON_SCHEMA_SCALAR_TYPES` into `resources/jsonSchemaTypes.ts` so the two type mappers could not drift on *recognized* scalars. These are the cases that shared set doesn't cover.

## 1. Unrecognized type names diverge silently

A type name in neither vocabulary — not a `DATA_TYPES` key, not a JSON Schema scalar (`'Text'`, `'Object'`, `'Array'`) — resolves differently:

- MCP: `components/mcp/tools/schemas/derive.ts:77` `default:` → `{ type: 'string' }`
- OpenAPI: `resources/openApi.ts:187` → `new Type(DATA_TYPES['Text'] /* undefined */, 'Text')` → `{}`

One coerces to a wrong-but-confident type, the other emits untyped. Neither warns. A typo in a `static properties` fragment produces two different wrong answers with no signal to the author.

Suggested: share one resolution helper, and log a warn-once at registration naming the property and the unrecognized type — the author almost certainly meant one of the two vocabularies.

## 2. `enum` / `format` / `const` / `description` are dropped by OpenAPI below the top level

`openApi.ts:190-198` attaches these only to the top-level `props[name]`. Nested objects route through `attributeToFragment`, which deliberately omits `enum`/`format`/`const` (`jsonSchemaTypes.ts:121-124`), and primitive array items route through `new Type(...)`, which carries only `type` — dropping `description` too. MCP's `attributeToProperty` (`derive.ts:113-115`) recurses and keeps all of them.

So the same fragment yields a documented, constrained nested property for an LLM and a bare typed one for Swagger UI. `description` in particular is a plain loss for the OpenAPI consumer.

## 3. Nullability has three different representations

An author writes `nullable: true` or `type: ['string','null']`; both fold to the same internal flag (`jsonSchemaTypes.ts:159-165`). Emission:

- MCP re-expands to `{ "type": ["string","null"] }` (`derive.ts:103-109`) — never emits a `nullable` keyword.
- OpenAPI **drops it entirely** for a top-level property — `openApi.ts:148` reads `nullable` only to compute `required`, and never writes it onto the property schema.
- OpenAPI *does* emit `nullable: true` for a property inside a nested object, via `attributeToFragment` (`jsonSchemaTypes.ts:120`).

The top-level OpenAPI drop is the actual bug: OpenAPI 3.0.3 has a `nullable` keyword and we have the information, so a nullable property is currently advertised as non-nullable. The nested/top-level inconsistency within OpenAPI itself is the clearest evidence this wasn't deliberate.

## Suggested approach

Route both emit paths through one projector that takes a target dialect (`json-schema-2020` for MCP, `openapi-3.0.3` for OpenAPI) and handles the two legitimate dialect differences — union-with-`null` vs `nullable: true` — in one place, so everything else stays identical by construction. That is a bigger change than #1921 made; splitting item 3 (the top-level `nullable` drop) out as a standalone fix is reasonable if the unification slips.

Each item is separately observable, so unit coverage should assert MCP and OpenAPI output for the *same* fragment side by side — the convergence test added in #1921 is the natural home.

Once fixed, the per-surface caveats in the fragment-key table in `reference/resources/resource-api.md` collapse back to single rows.

kylebernhardy:
## Summary

Closes #1920. Closes #1941. Closes #1942.

A programmatic Resource can declare `static properties` instead of table-backed `attributes`, but MCP tools, OpenAPI, and `harper://schema` previously produced incomplete or divergent schemas from that declaration.

This PR adds the [structural projection and guarded schema traversal](https://github.com/HarperFast/harper/pull/1921/changes?w=1#diff-f2d035566b78fa2b69c6fa350ad42c6fb4d1b889666c4bf9d0fed50892f35495R275) and routes MCP and OpenAPI through one dialect-aware emitter. Programmatic Resources preserve descriptions, enum/format/const constraints, arrays, nested objects, unions, nullability, and embedded object definitions. Hidden fields are suppressed at every depth and removed from required lists. Existing table-backed output remains unchanged.

Programmatic authors can use [public `static required`](https://github.com/HarperFast/harper/pull/1921/changes?w=1#diff-d1634721e1dea3d20c73434e1194180e33e6f6e6e528ba017c0115354a75fb2eR70) for top-level presence independently of nullability.

## Reliability hardening

- Invalid runtime metadata is isolated per Resource, including malformed enums, required lists, const values, cycles, and excessive nesting.
- Relationship references stop before recursive graph traversal and are omitted from write inputs; embedded object definitions retain their full nested schemas.
- [Verb tool publication is atomic and suffixes stay stable](https://github.com/HarperFast/harper/pull/1921/changes?w=1#diff-061d972b6170c26b6085e634299b8b730c7e09a31580c4a69be199899177e991R1378) when a Resource schema is invalid.
- [`harper://schema` uses the shared guarded filter](https://github.com/HarperFast/harper/pull/1921/changes?w=1#diff-409478891b7d802bb43d6c3be20bd72734a93d2e117f3a96e1b0123b151fe553R826), including hidden-name pruning from nested required lists.
- OpenAPI 3.0.3 supplies `items: {}` for unconstrained arrays and translates null/const forms without changing JSON Schema intersection semantics.

## Verification

- `npm run build`
- changed-file oxlint: passed
- focused metadata and MCP application tests: 92 passed on `ca7b57571`
- broader focused MCP/OpenAPI regression passes completed during development: 218, 140, and 73 tests passed at their respective reviewed checkpoints
- full resource suite was attempted locally but aborted in unrelated logger/native-test setup after permission errors under the shared test root
- MCP integration execution is locally blocked because the configured loopback aliases are unavailable; GitHub CI is the integration gate for this head
- authorized Claude final-delta review: LGTM, no open findings; its two optional edge cases were added in `ca7b57571`
- Gemini CLI was selected but could not run because local authentication is expired
- Harper domain adjudication was selected but timed out after the Claude result; no additional finding was produced

## Follow-ups

- #1922 tracks URL attribute-suffix routing for programmatic Resources.
- #1923 tracks documentation for the programmatic `static properties` authoring path.

### DIFF
--- a/components/mcp/resources.ts
+++ b/components/mcp/resources.ts
@@ -38,6 +38,7 @@ import harperLogger from '../../utility/logging/harper_logger.ts';
 import { AccessViolation } from '../../utility/errors/hdbError.ts';
 import { SERVER_CAPABILITIES, SERVER_INFO, SUPPORTED_PROTOCOL_VERSIONS } from './lifecycle.ts';
 import { encodeCursor } from './pagination.ts';
+import { resolveAttributes } from '../../resources/jsonSchemaTypes.ts';
 import {
 	customResourceCompletionValues,
 	listCustomResources,
@@ -820,7 +821,9 @@ function readTableSchema(db: string, table: string, user: AuthedUser, href: stri
 		}
 	}
 	if (!resource) return { ok: false, reason: `table not found: ${db}.${table}` };
-	const attributes = resource.attributes ?? [];
+	// Fall back to a programmatic Resource's `static properties` when it declares no attributes Array,
+	// so schema introspection matches the tool/OpenAPI surfaces.
+	const attributes = resolveAttributes(resource);
 	const filteredAttributes = filterAttributesByPermissions(attributes, perm?.attribute_permissions);
 	const body = {
 		database: db,
--- a/components/mcp/resources.ts
+++ b/components/mcp/resources.ts
@@ -38,7 +38,7 @@ import harperLogger from '../../utility/logging/harper_logger.ts';
 import { AccessViolation } from '../../utility/errors/hdbError.ts';
 import { SERVER_CAPABILITIES, SERVER_INFO, SUPPORTED_PROTOCOL_VERSIONS } from './lifecycle.ts';
 import { encodeCursor } from './pagination.ts';
-import { resolveAttributes } from '../../resources/jsonSchemaTypes.ts';
+import { filterAttributeTree, resolveAttributes } from '../../resources/jsonSchemaTypes.ts';
 import {
 	customResourceCompletionValues,
 	listCustomResources,
@@ -821,14 +821,14 @@ function readTableSchema(db: string, table: string, user: AuthedUser, href: stri
 		}
 	}
 	if (!resource) return { ok: false, reason: `table not found: ${db}.${table}` };
-	let attributes;
+	let filteredAttributes;
 	try {
-		attributes = resolveAttributes(resource);
+		const attributes = resolveAttributes(resource);
+		const visibleAttributes = filterAttributeTree(attributes);
+		filteredAttributes = filterAttributesByPermissions(visibleAttributes, perm?.attribute_permissions);
 	} catch (error) {
 		return { ok: false, reason: `invalid schema for ${db}.${table}: ${(error as Error).message}` };
 	}
-	const visibleAttributes = filterHiddenAttributes(attributes);
-	const filteredAttributes = filterAttributesByPermissions(visibleAttributes, perm?.attribute_permissions);
 	const body = {
 		database: db,
 		table,
@@ -941,23 +941,6 @@ function filterAttributesByPermissions(attributes: any[], attributePermissions: 
 	}
 	if (denied.size === 0) return attributes;
 	return attributes.filter((a) => !denied.has(a?.name));
-}
-
-function filterHiddenAttributes(attributes: any[]): any[] {
-	return attributes
-		.filter((attribute) => !attribute?.hidden)
-		.map((attribute) => {
-			const visible = { ...attribute };
-			if (Object.prototype.propertyIsEnumerable.call(attribute, 'properties') && Array.isArray(attribute.properties)) {
-				visible.properties = filterHiddenAttributes(attribute.properties);
-			}
-			if (attribute.elements && typeof attribute.elements === 'object') {
-				const [elements] = filterHiddenAttributes([attribute.elements]);
-				if (elements) visible.elements = elements;
-				else delete visible.elements;
-			}
-			return visible;
-		});
 }
 
 /**
--- a/components/mcp/resources.ts
+++ b/components/mcp/resources.ts
@@ -827,7 +827,7 @@ function readTableSchema(db: string, table: string, user: AuthedUser, href: stri
 	} catch (error) {
 		return { ok: false, reason: `invalid schema for ${db}.${table}: ${(error as Error).message}` };
 	}
-	const visibleAttributes = attributes.filter((attribute) => !attribute.hidden);
+	const visibleAttributes = filterHiddenAttributes(attributes);
 	const filteredAttributes = filterAttributesByPermissions(visibleAttributes, perm?.attribute_permissions);
 	const body = {
 		database: db,
@@ -941,6 +941,21 @@ function filterAttributesByPermissions(attributes: any[], attributePermissions: 
 	}
 	if (denied.size === 0) return attributes;
 	return attributes.filter((a) => !denied.has(a?.name));
+}
+
+function filterHiddenAttributes(attributes: any[]): any[] {
+	return attributes
+		.filter((attribute) => !attribute?.hidden)
+		.map((attribute) => {
+			const visible = { ...attribute };
+			if (Array.isArray(attribute.properties)) visible.properties = filterHiddenAttributes(attribute.properties);
+			if (attribute.elements && typeof attribute.elements === 'object') {
+				const [elements] = filterHiddenAttributes([attribute.elements]);
+				if (elements) visible.elements = elements;
+				else delete visible.elements;
+			}
+			return visible;
+		});
 }
 
 /**
--- a/components/mcp/resources.ts
+++ b/components/mcp/resources.ts
@@ -948,7 +948,9 @@ function filterHiddenAttributes(attributes: any[]): any[] {
 		.filter((attribute) => !attribute?.hidden)
 		.map((attribute) => {
 			const visible = { ...attribute };
-			if (Array.isArray(attribute.properties)) visible.properties = filterHiddenAttributes(attribute.properties);
+			if (Object.prototype.propertyIsEnumerable.call(attribute, 'properties') && Array.isArray(attribute.properties)) {
+				visible.properties = filterHiddenAttributes(attribute.properties);
+			}
 			if (attribute.elements && typeof attribute.elements === 'object') {
 				const [elements] = filterHiddenAttributes([attribute.elements]);
 				if (elements) visible.elements = elements;
--- a/components/mcp/resources.ts
+++ b/components/mcp/resources.ts
@@ -821,9 +821,12 @@ function readTableSchema(db: string, table: string, user: AuthedUser, href: stri
 		}
 	}
 	if (!resource) return { ok: false, reason: `table not found: ${db}.${table}` };
-	// Fall back to a programmatic Resource's `static properties` when it declares no attributes Array,
-	// so schema introspection matches the tool/OpenAPI surfaces.
-	const attributes = resolveAttributes(resource);
+	let attributes;
+	try {
+		attributes = resolveAttributes(resource);
+	} catch (error) {
+		return { ok: false, reason: `invalid schema for ${db}.${table}: ${(error as Error).message}` };
+	}
 	const filteredAttributes = filterAttributesByPermissions(attributes, perm?.attribute_permissions);
 	const body = {
 		database: db,
--- a/components/mcp/resources.ts
+++ b/components/mcp/resources.ts
@@ -827,7 +827,8 @@ function readTableSchema(db: string, table: string, user: AuthedUser, href: stri
 	} catch (error) {
 		return { ok: false, reason: `invalid schema for ${db}.${table}: ${(error as Error).message}` };
 	}
-	const filteredAttributes = filterAttributesByPermissions(attributes, perm?.attribute_permissions);
+	const visibleAttributes = attributes.filter((attribute) => !attribute.hidden);
+	const filteredAttributes = filterAttributesByPermissions(visibleAttributes, perm?.attribute_permissions);
 	const body = {
 		database: db,
 		table,
--- a/components/mcp/tools/application.ts
+++ b/components/mcp/tools/application.ts
@@ -49,6 +49,7 @@ import {
 } from '../customResourceRegistry.ts';
 import { notifyPromptsListChanged, notifyResourcesListChanged, notifyToolsListChanged } from '../listChanged.ts';
 import { decodeCursor, encodeCursor } from '../pagination.ts';
+import { resolveAttributes } from '../../../resources/jsonSchemaTypes.ts';
 import {
 	type AttributePermissionEntry,
 	type HarperAttribute,
@@ -1356,7 +1357,9 @@ function buildApplicationTools(resources: ResourcesRegistry): void {
 		const tableName = ResourceClass?.tableName;
 		const suffix = uniqueSuffix(path, databaseName, claimedSuffixes);
 		claimedSuffixes.add(suffix);
-		const attributes = (ResourceClass?.attributes ?? []) as HarperAttribute[];
+		// A programmatic Resource may declare `static properties` without an `attributes` Array; resolve
+		// the effective attributes so its verb tools get a rich inputSchema instead of a skeletal one.
+		const attributes = resolveAttributes(ResourceClass) as HarperAttribute[];
 		if (hasVerbs) {
 			toolsRegistered += registerVerbTools({
 				path,
--- a/components/mcp/tools/application.ts
+++ b/components/mcp/tools/application.ts
@@ -53,7 +53,7 @@ import {
 } from '../customResourceRegistry.ts';
 import { notifyPromptsListChanged, notifyResourcesListChanged, notifyToolsListChanged } from '../listChanged.ts';
 import { decodeCursor, encodeCursor } from '../pagination.ts';
-import { resolveAttributes } from '../../../resources/jsonSchemaTypes.ts';
+import { resolveAttributes, SchemaTraversalError } from '../../../resources/jsonSchemaTypes.ts';
 import {
 	type AttributePermissionEntry,
 	type HarperAttribute,
@@ -1390,15 +1390,20 @@ function buildApplicationTools(resources: ResourcesRegistry): void {
 				const tableName = ResourceClass?.tableName;
 				const suffix = uniqueSuffix(path, databaseName, claimedSuffixes);
 				claimedSuffixes.add(suffix);
-				toolsRegistered += registerVerbTools({
-					path,
-					suffix,
-					ResourceClass,
-					databaseName,
-					tableName,
-					attributes,
-					verbs,
-				});
+				try {
+					toolsRegistered += registerVerbTools({
+						path,
+						suffix,
+						ResourceClass,
+						databaseName,
+						tableName,
+						attributes,
+						verbs,
+					});
+				} catch (error) {
+					if (!(error instanceof SchemaTraversalError)) throw error;
+					warnInvalidResourceSchema(ResourceClass, path, error);
+				}
 			}
 		}
 		toolsRegistered += registerCustomMcpTools(ResourceClass, path);
--- a/components/mcp/tools/application.ts
+++ b/components/mcp/tools/application.ts
@@ -328,6 +328,14 @@ type RequestTargetCtor = new () => Record<string, unknown> & {
 // Test seams: avoid Harper's eager graph init in unit tests.
 let _resourcesOverride: ResourcesRegistry | undefined;
 let _requestTargetCtorOverride: RequestTargetCtor | undefined;
+const warnedInvalidResourceSchemas = new WeakSet<object>();
+
+function warnInvalidResourceSchema(resource: ResourceClassLike, path: string, error: unknown): void {
+	if (warnedInvalidResourceSchemas.has(resource)) return;
+	warnedInvalidResourceSchemas.add(resource);
+	harperLogger.warn(`MCP application skipped invalid schema for resource '/${path}': ${(error as Error).message}`);
+}
+
 export function _setResourcesForTest(r: ResourcesRegistry | undefined): void {
 	_resourcesOverride = r;
 }
@@ -1370,7 +1378,13 @@ function buildApplicationTools(resources: ResourcesRegistry): void {
 		claimedSuffixes.add(suffix);
 		// A programmatic Resource may declare `static properties` without an `attributes` Array; resolve
 		// the effective attributes so its verb tools get a rich inputSchema instead of a skeletal one.
-		const attributes = resolveAttributes(ResourceClass) as HarperAttribute[];
+		let attributes: HarperAttribute[];
+		try {
+			attributes = resolveAttributes(ResourceClass) as HarperAttribute[];
+		} catch (error) {
+			warnInvalidResourceSchema(ResourceClass, path, error);
+			return;
+		}
 		if (hasVerbs) {
 			toolsRegistered += registerVerbTools({
 				path,
--- a/components/mcp/tools/application.ts
+++ b/components/mcp/tools/application.ts
@@ -53,6 +53,7 @@ import {
 } from '../customResourceRegistry.ts';
 import { notifyPromptsListChanged, notifyResourcesListChanged, notifyToolsListChanged } from '../listChanged.ts';
 import { decodeCursor, encodeCursor } from '../pagination.ts';
+import { resolveAttributes } from '../../../resources/jsonSchemaTypes.ts';
 import {
 	type AttributePermissionEntry,
 	type HarperAttribute,
@@ -1367,7 +1368,9 @@ function buildApplicationTools(resources: ResourcesRegistry): void {
 		const tableName = ResourceClass?.tableName;
 		const suffix = uniqueSuffix(path, databaseName, claimedSuffixes);
 		claimedSuffixes.add(suffix);
-		const attributes = (ResourceClass?.attributes ?? []) as HarperAttribute[];
+		// A programmatic Resource may declare `static properties` without an `attributes` Array; resolve
+		// the effective attributes so its verb tools get a rich inputSchema instead of a skeletal one.
+		const attributes = resolveAttributes(ResourceClass) as HarperAttribute[];
 		if (hasVerbs) {
 			toolsRegistered += registerVerbTools({
 				path,
--- a/components/mcp/tools/application.ts
+++ b/components/mcp/tools/application.ts
@@ -1387,7 +1387,6 @@ function buildApplicationTools(resources: ResourcesRegistry): void {
 			if (attributes) {
 				const databaseName = ResourceClass?.databaseName;
 				const tableName = ResourceClass?.tableName;
-				void ResourceClass.description;
 				let definitions: ToolDef[] | undefined;
 				try {
 					definitions = deriveVerbTools({
--- a/components/mcp/tools/application.ts
+++ b/components/mcp/tools/application.ts
@@ -34,6 +34,7 @@ import {
 	snapshotProfileTools,
 	type AuthedUser,
 	type ToolCallContext,
+	type ToolDef,
 	type ToolResult,
 } from '../toolRegistry.ts';
 import {
@@ -53,7 +54,7 @@ import {
 } from '../customResourceRegistry.ts';
 import { notifyPromptsListChanged, notifyResourcesListChanged, notifyToolsListChanged } from '../listChanged.ts';
 import { decodeCursor, encodeCursor } from '../pagination.ts';
-import { resolveAttributes, SchemaTraversalError } from '../../../resources/jsonSchemaTypes.ts';
+import { resolveAttributes } from '../../../resources/jsonSchemaTypes.ts';
 import {
 	type AttributePermissionEntry,
 	type HarperAttribute,
@@ -88,6 +89,7 @@ interface ResourceClassLike {
 	description?: string;
 	hidden?: boolean;
 	properties?: Record<string, unknown>;
+	required?: readonly string[];
 	outputSchemas?: { [verb: string]: object };
 	/** The URL path (may carry `:param`/`*wildcard` segments), used to bind path params for contract resources. */
 	path?: string;
@@ -333,7 +335,11 @@ let warnedInvalidResourceSchemas = new WeakSet<object>();
 function warnInvalidResourceSchema(resource: ResourceClassLike, path: string, error: unknown): void {
 	if (warnedInvalidResourceSchemas.has(resource)) return;
 	warnedInvalidResourceSchemas.add(resource);
-	harperLogger.warn(`MCP application skipped invalid schema for resource '/${path}': ${(error as Error).message}`);
+	let detail = 'unknown schema error';
+	try {
+		detail = typeof error === 'object' && error !== null && 'message' in error ? String(error.message) : String(error);
+	} catch {}
+	harperLogger.warn(`MCP application skipped invalid schema for resource '/${path}': ${detail}`);
 }
 
 export function _setResourcesForTest(r: ResourcesRegistry | undefined): void {
@@ -841,8 +847,8 @@ function mergeAnnotations(
  * semantics) is the only built-in default; `patch_*`/`delete_*` annotations are
  * deferred until repeat-call behavior is verified end-to-end.
  */
-function registerVerbTools(ctx: ResourceContext): number {
-	let count = 0;
+function deriveVerbTools(ctx: ResourceContext): ToolDef[] {
+	const definitions: ToolDef[] = [];
 	const { suffix, ResourceClass, attributes, verbs, tableName, databaseName, path } = ctx;
 	const tableDoc = ResourceClass.description;
 	const primaryKey = ResourceClass.primaryKey;
@@ -864,7 +870,7 @@ function registerVerbTools(ctx: ResourceContext): number {
 
 	if (verbs.get) {
 		const name = `get_${suffix}`;
-		addTool({
+		definitions.push({
 			name,
 			description: verbDescription('get', ctxForVerb),
 			inputSchema: overrideInput('get') ?? deriveGetSchema(attributes, undefined),
@@ -874,13 +880,12 @@ function registerVerbTools(ctx: ResourceContext): number {
 			visibleTo: makeVisibleTo(databaseName, tableName, 'read'),
 			handler: makeGetHandler(name, path, ResourceClass),
 		});
-		count++;
 	}
 	if (verbs.search) {
 		const name = `search_${suffix}`;
 		// search_* deliberately omits outputSchema — envelope shape (records vs
 		// data, cursor vs nextCursor) is tracked in the sibling envelope issue.
-		addTool({
+		definitions.push({
 			name,
 			description: verbDescription('search', ctxForVerb),
 			inputSchema: overrideInput('search') ?? deriveSearchSchema(attributes, undefined),
@@ -889,11 +894,10 @@ function registerVerbTools(ctx: ResourceContext): number {
 			visibleTo: makeVisibleTo(databaseName, tableName, 'read'),
 			handler: makeSearchHandler(name, path, ResourceClass),
 		});
-		count++;
 	}
 	if (verbs.create) {
 		const name = `create_${suffix}`;
-		addTool({
+		definitions.push({
 			name,
 			description: verbDescription('create', ctxForVerb),
 			inputSchema: overrideInput('create') ?? deriveCreateSchema(attributes, undefined),
@@ -903,11 +907,10 @@ function registerVerbTools(ctx: ResourceContext): number {
 			visibleTo: makeVisibleTo(databaseName, tableName, 'insert'),
 			handler: makeCreateHandler(name, path, ResourceClass),
 		});
-		count++;
 	}
 	if (verbs.updatePut) {
 		const name = `update_${suffix}`;
-		addTool({
+		definitions.push({
 			name,
 			description: verbDescription('update', ctxForVerb),
 			inputSchema: overrideInput('update') ?? deriveUpdateSchema(attributes, undefined),
@@ -919,10 +922,9 @@ function registerVerbTools(ctx: ResourceContext): number {
 			visibleTo: makeVisibleTo(databaseName, tableName, 'update'),
 			handler: makeUpdateHandler(name, path, ResourceClass, 'put'),
 		});
-		count++;
 	} else if (verbs.updatePatch) {
 		const name = `patch_${suffix}`;
-		addTool({
+		definitions.push({
 			name,
 			description: verbDescription('patch', ctxForVerb),
 			inputSchema: overrideInput('patch') ?? deriveUpdateSchema(attributes, undefined),
@@ -934,7 +936,6 @@ function registerVerbTools(ctx: ResourceContext): number {
 			visibleTo: makeVisibleTo(databaseName, tableName, 'update'),
 			handler: makeUpdateHandler(name, path, ResourceClass, 'patch'),
 		});
-		count++;
 	}
 	if (verbs.delete) {
 		const name = `delete_${suffix}`;
@@ -942,7 +943,7 @@ function registerVerbTools(ctx: ResourceContext): number {
 		// `{ deleted }` so the result carries structuredContent matching the
 		// `{ deleted: boolean }` outputSchema. Authors can override the shape via
 		// `static outputSchemas.delete`.
-		addTool({
+		definitions.push({
 			name,
 			description: verbDescription('delete', ctxForVerb),
 			inputSchema: overrideInput('delete') ?? deriveDeleteSchema(attributes, undefined),
@@ -954,9 +955,8 @@ function registerVerbTools(ctx: ResourceContext): number {
 			visibleTo: makeVisibleTo(databaseName, tableName, 'delete'),
 			handler: makeDeleteHandler(name, path, ResourceClass),
 		});
-		count++;
 	}
-	return count;
+	return definitions;
 }
 
 /**
@@ -1375,9 +1375,8 @@ function buildApplicationTools(resources: ResourcesRegistry): void {
 		const hasCustomPrompts = Array.isArray(ResourceClass?.mcpPrompts) && ResourceClass.mcpPrompts.length > 0;
 		const hasCustomResources = Array.isArray(ResourceClass?.mcpResources) && ResourceClass.mcpResources.length > 0;
 		if (!hasVerbs && !hasCustomTools && !hasCustomPrompts && !hasCustomResources) return;
-		if (!hasVerbs) {
-			claimedSuffixes.add(uniqueSuffix(path, ResourceClass?.databaseName, claimedSuffixes));
-		}
+		const suffix = uniqueSuffix(path, ResourceClass?.databaseName, claimedSuffixes);
+		claimedSuffixes.add(suffix);
 		if (hasVerbs) {
 			let attributes: HarperAttribute[] | undefined;
 			try {
@@ -1388,10 +1387,10 @@ function buildApplicationTools(resources: ResourcesRegistry): void {
 			if (attributes) {
 				const databaseName = ResourceClass?.databaseName;
 				const tableName = ResourceClass?.tableName;
-				const suffix = uniqueSuffix(path, databaseName, claimedSuffixes);
-				claimedSuffixes.add(suffix);
+				void ResourceClass.description;
+				let definitions: ToolDef[] | undefined;
 				try {
-					toolsRegistered += registerVerbTools({
+					definitions = deriveVerbTools({
 						path,
 						suffix,
 						ResourceClass,
@@ -1401,8 +1400,11 @@ function buildApplicationTools(resources: ResourcesRegistry): void {
 						verbs,
 					});
 				} catch (error) {
-					if (!(error instanceof SchemaTraversalError)) throw error;
 					warnInvalidResourceSchema(ResourceClass, path, error);
+				}
+				if (definitions) {
+					for (const definition of definitions) addTool(definition);
+					toolsRegistered += definitions.length;
 				}
 			}
 		}
--- a/components/mcp/tools/application.ts
+++ b/components/mcp/tools/application.ts
@@ -1375,6 +1375,9 @@ function buildApplicationTools(resources: ResourcesRegistry): void {
 		const hasCustomPrompts = Array.isArray(ResourceClass?.mcpPrompts) && ResourceClass.mcpPrompts.length > 0;
 		const hasCustomResources = Array.isArray(ResourceClass?.mcpResources) && ResourceClass.mcpResources.length > 0;
 		if (!hasVerbs && !hasCustomTools && !hasCustomPrompts && !hasCustomResources) return;
+		if (!hasVerbs) {
+			claimedSuffixes.add(uniqueSuffix(path, ResourceClass?.databaseName, claimedSuffixes));
+		}
 		if (hasVerbs) {
 			let attributes: HarperAttribute[] | undefined;
 			try {
--- a/components/mcp/tools/application.ts
+++ b/components/mcp/tools/application.ts
@@ -328,7 +328,7 @@ type RequestTargetCtor = new () => Record<string, unknown> & {
 // Test seams: avoid Harper's eager graph init in unit tests.
 let _resourcesOverride: ResourcesRegistry | undefined;
 let _requestTargetCtorOverride: RequestTargetCtor | undefined;
-const warnedInvalidResourceSchemas = new WeakSet<object>();
+let warnedInvalidResourceSchemas = new WeakSet<object>();
 
 function warnInvalidResourceSchema(resource: ResourceClassLike, path: string, error: unknown): void {
 	if (warnedInvalidResourceSchemas.has(resource)) return;
@@ -341,6 +341,9 @@ export function _setResourcesForTest(r: ResourcesRegistry | undefined): void {
 }
 export function _setRequestTargetForTest(ctor: RequestTargetCtor | undefined): void {
 	_requestTargetCtorOverride = ctor;
+}
+export function _resetInvalidSchemaWarningsForTest(): void {
+	warnedInvalidResourceSchemas = new WeakSet<object>();
 }
 
 // Cache the Resources module object (not the `resources` registry itself):
@@ -1372,29 +1375,28 @@ function buildApplicationTools(resources: ResourcesRegistry): void {
 		const hasCustomPrompts = Array.isArray(ResourceClass?.mcpPrompts) && ResourceClass.mcpPrompts.length > 0;
 		const hasCustomResources = Array.isArray(ResourceClass?.mcpResources) && ResourceClass.mcpResources.length > 0;
 		if (!hasVerbs && !hasCustomTools && !hasCustomPrompts && !hasCustomResources) return;
-		const databaseName = ResourceClass?.databaseName;
-		const tableName = ResourceClass?.tableName;
-		const suffix = uniqueSuffix(path, databaseName, claimedSuffixes);
-		claimedSuffixes.add(suffix);
-		// A programmatic Resource may declare `static properties` without an `attributes` Array; resolve
-		// the effective attributes so its verb tools get a rich inputSchema instead of a skeletal one.
-		let attributes: HarperAttribute[];
-		try {
-			attributes = resolveAttributes(ResourceClass) as HarperAttribute[];
-		} catch (error) {
-			warnInvalidResourceSchema(ResourceClass, path, error);
-			return;
-		}
 		if (hasVerbs) {
-			toolsRegistered += registerVerbTools({
-				path,
-				suffix,
-				ResourceClass,
-				databaseName,
-				tableName,
-				attributes,
-				verbs,
-			});
+			let attributes: HarperAttribute[] | undefined;
+			try {
+				attributes = resolveAttributes(ResourceClass) as HarperAttribute[];
+			} catch (error) {
+				warnInvalidResourceSchema(ResourceClass, path, error);
+			}
+			if (attributes) {
+				const databaseName = ResourceClass?.databaseName;
+				const tableName = ResourceClass?.tableName;
+				const suffix = uniqueSuffix(path, databaseName, claimedSuffixes);
+				claimedSuffixes.add(suffix);
+				toolsRegistered += registerVerbTools({
+					path,
+					suffix,
+					ResourceClass,
+					databaseName,
+					tableName,
+					attributes,
+					verbs,
+				});
+			}
 		}
 		toolsRegistered += registerCustomMcpTools(ResourceClass, path);
 		registerCustomMcpPrompts(ResourceClass, path);
--- a/components/mcp/tools/schemas/derive.ts
+++ b/components/mcp/tools/schemas/derive.ts
@@ -25,6 +25,10 @@ export interface HarperAttribute {
 	assignCreatedTime?: boolean;
 	assignUpdatedTime?: boolean;
 	expiresAt?: boolean;
+	// JSON-Schema hints a programmatic Resource may carry via `static properties`.
+	enum?: readonly (string | number | boolean | null)[];
+	format?: string;
+	const?: unknown;
 }
 
 export interface AttributePermissionEntry {
@@ -63,6 +67,16 @@ function harperTypeToJsonSchema(type: string | undefined): { type: string | stri
 		case 'Any':
 		case undefined:
 			return {};
+		case 'string':
+		case 'integer':
+		case 'number':
+		case 'boolean':
+		case 'object':
+		case 'array':
+		case 'null':
+			// A programmatic Resource's `static properties` already speaks JSON Schema (lowercase types,
+			// no collision with Harper's capitalized GraphQL types); pass those through unchanged.
+			return { type };
 		default:
 			return { type: 'string' };
 	}
@@ -97,6 +111,9 @@ function attributeToProperty(attr: HarperAttribute): object {
 	if (attr.description && !base.description) {
 		base.description = attr.description;
 	}
+	if (attr.enum && !('enum' in base)) base.enum = attr.enum;
+	if (attr.format && !('format' in base)) base.format = attr.format;
+	if (attr.const !== undefined && !('const' in base)) base.const = attr.const;
 	return base;
 }
 
--- a/components/mcp/tools/schemas/derive.ts
+++ b/components/mcp/tools/schemas/derive.ts
@@ -19,6 +19,8 @@ export interface HarperAttribute {
 	description?: string;
 	hidden?: boolean;
 	nullable?: boolean;
+	/** Source JSON-Schema type union from `static properties`; MCP accepts type arrays, so it passes through. */
+	types?: readonly string[];
 	isPrimaryKey?: boolean;
 	properties?: HarperAttribute[];
 	elements?: HarperAttribute;
@@ -97,6 +99,10 @@ function attributeToProperty(attr: HarperAttribute): object {
 			type: 'array',
 			items: attributeToProperty(attr.elements),
 		};
+	} else if (attr.types) {
+		// MCP speaks JSON Schema, which has type unions — emit the author's union as declared rather
+		// than the single `type` the attribute form collapses to.
+		base = { type: [...attr.types] };
 	} else {
 		base = harperTypeToJsonSchema(attr.type) as typeof base;
 	}
--- a/components/mcp/tools/schemas/derive.ts
+++ b/components/mcp/tools/schemas/derive.ts
@@ -11,6 +11,8 @@
  * doesn't waste tokens on fields it can't write), NOT a security boundary.
  */
 
+import { JSON_SCHEMA_SCALAR_TYPES } from '../../../../resources/jsonSchemaTypes.ts';
+
 export interface HarperAttribute {
 	name: string;
 	type?: string;
@@ -29,6 +31,8 @@ export interface HarperAttribute {
 	enum?: readonly (string | number | boolean | null)[];
 	format?: string;
 	const?: unknown;
+	required?: readonly string[];
+	additionalProperties?: boolean;
 }
 
 export interface AttributePermissionEntry {
@@ -46,6 +50,9 @@ type Mode = 'read' | 'insert' | 'update';
  * than blocking the field entirely; the runtime will validate.
  */
 function harperTypeToJsonSchema(type: string | undefined): { type: string | string[] } | object {
+	// A programmatic Resource's `static properties` already speaks JSON Schema (lowercase types, no
+	// collision with Harper's capitalized GraphQL types); pass those through unchanged.
+	if (type && JSON_SCHEMA_SCALAR_TYPES.has(type)) return { type };
 	switch (type) {
 		case 'Int':
 		case 'Long':
@@ -67,16 +74,6 @@ function harperTypeToJsonSchema(type: string | undefined): { type: string | stri
 		case 'Any':
 		case undefined:
 			return {};
-		case 'string':
-		case 'integer':
-		case 'number':
-		case 'boolean':
-		case 'object':
-		case 'array':
-		case 'null':
-			// A programmatic Resource's `static properties` already speaks JSON Schema (lowercase types,
-			// no collision with Harper's capitalized GraphQL types); pass those through unchanged.
-			return { type };
 		default:
 			return { type: 'string' };
 	}
@@ -93,6 +90,8 @@ function attributeToProperty(attr: HarperAttribute): object {
 			type: 'object',
 			properties: Object.fromEntries(attr.properties.map((p) => [p.name, attributeToProperty(p)])),
 		};
+		if (attr.required) base.required = attr.required;
+		if (attr.additionalProperties !== undefined) base.additionalProperties = attr.additionalProperties;
 	} else if (attr.type === 'array' && attr.elements) {
 		base = {
 			type: 'array',
--- a/components/mcp/tools/schemas/derive.ts
+++ b/components/mcp/tools/schemas/derive.ts
@@ -345,8 +345,6 @@ function deriveRecordSchema(
 		properties[attr.name] = schema;
 		const requiredOnOutput =
 			!attr.relationship &&
-			!attr.definition &&
-			!attr.elements?.definition &&
 			(attr.requiredOnSchema ||
 				attr.nullable === false ||
 				attr.assignCreatedTime ||
--- a/components/mcp/tools/schemas/derive.ts
+++ b/components/mcp/tools/schemas/derive.ts
@@ -187,7 +187,7 @@ function buildPropertiesObject(
 		// Skip auto-managed columns from write inputs — Harper assigns them.
 		if (mode !== 'read' && (attr.assignCreatedTime || attr.assignUpdatedTime || attr.expiresAt)) continue;
 		if (mode !== 'read' && (attr.computed !== undefined || attr.computedFromExpression !== undefined)) continue;
-		if (mode !== 'read' && (attr.relationship || attr.definition || attr.elements?.definition)) continue;
+		if (mode !== 'read' && attr.relationship) continue;
 		const schema = attributeToProperty(attr);
 		if (!schema) continue;
 		properties[attr.name] = schema;
--- a/components/mcp/tools/schemas/derive.ts
+++ b/components/mcp/tools/schemas/derive.ts
@@ -41,6 +41,7 @@ export interface HarperAttribute {
 	const?: unknown;
 	required?: readonly string[];
 	additionalProperties?: boolean;
+	requiredOnSchema?: boolean;
 	relationship?: unknown;
 	definition?: unknown;
 }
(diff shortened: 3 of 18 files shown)


resources/openApi.ts
```diff
@@ -151,21 +161,34 @@ export function generateJsonApi(resources: Resources, serverHttpURL: string) {
...
+						// schema (sub-properties recursed); OpenAPI's table path uses $refs instead.
+						props[name] = attributeToFragment(attr);
 					} else if (type === 'array') {
 						if (elements.type === 'Any') {
 							props[name] = { type: 'array', items: { format: elements.type } };
+						} else if (!DATA_TYPES[elements.type] && JSON_SCHEMA_SCALARS.has(elements.type)) {
+							props[name] = { type: 'array', items: new Type(elements.type) };
 						} else {
 							props[name] = { type: 'array', items: new Type(DATA_TYPES[elements.type], elements.type) };
 						}
```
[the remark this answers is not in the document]
kylebernhardy: Fixed in 1383b6e42 — added a `!elements` guard that emits a bare `{ type: 'array' }` for an items-less array, so generation no longer dereferences `undefined`. Added a regression test (`anything: { type: 'array' }` in openApi.test.js). Thanks for the catch.

resources/openApi.ts
```diff
@@ -151,21 +161,37 @@ export function generateJsonApi(resources: Resources, serverHttpURL: string) {
 						} else {
 							props[name] = { $ref: SCHEMA_COMP_REF + def.type };
 						}
+					} else if (attr.properties) {
+						// Nested object from `static properties` — the shared projector emits the full object
+						// schema (sub-properties recursed); OpenAPI's table path uses $refs instead.
+						props[name] = attributeToFragment(attr);
 					} else if (type === 'array') {
-						if (elements.type === 'Any') {
+						if (elements.properties) {
```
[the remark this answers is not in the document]
kylebernhardy: Fixed in 1383b6e42 — the array branch now guards `!elements` and emits a bare `{ type: 'array' }` (with a regression test). Good catch — this was a real crash of the whole OpenAPI document.

resources/jsonSchemaTypes.ts
```diff
@@ -108,3 +116,54 @@ export function projectAttributesToProperties(attributes: AttributeLike[]): Reco
...
+function fragmentToAttribute(name: string, fragment: JsonSchemaFragment): AttributeLike {
+	const attr: AttributeLike = { name };
+	if (fragment.properties) {
+		attr.properties = Object.entries(fragment.properties).map(([subName, sub]) => fragmentToAttribute(subName, sub));
+	} else if (fragment.type === 'array' && fragment.items) {
+		attr.type = 'array';
+		attr.elements = fragmentToAttribute(name, fragment.items);
+	} else if (fragment.type != null) {
+		attr.type = Array.isArray(fragment.type) ? fragment.type[0] : fragment.type;
+	}
```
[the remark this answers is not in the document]
kylebernhardy: Already addressed in 6ac0d1490 (this review's base predates it): `fragmentToAttribute` now filters `'null'` out of the union array (order-independent), sets `attr.nullable = true`, and keeps the non-null member — jsonSchemaTypes.ts:159-165 — covered by a union test in graphqlMetadata.test.js.

resources/openApi.ts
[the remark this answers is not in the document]
kylebernhardy: Already addressed in 6ac0d1490: the array branch recurses `elements.properties` via `attributeToFragment` (openApi.ts:171), so array-of-object items get their full object schema — covered by the `rows` case in openApi.test.js.

resources/jsonSchemaTypes.ts
```diff
@@ -108,3 +137,65 @@ export function projectAttributesToProperties(attributes: AttributeLike[]): Reco
...
+		// The element attribute's name is unused (attributeToFragment ignores it); keep it empty rather
+		// than misleadingly reusing the array field's own name.
+		attr.elements = fragmentToAttribute('', fragment.items);
+	} else if (Array.isArray(fragment.type)) {
+		// JSON-Schema union type. Fold a `'null'` member into `nullable` (the OpenAPI-expressible form)
+		// and keep the remaining member. A single non-null member is the common `['T','null']` case; a
+		// genuine multi-type union isn't expressible on an attribute, so the first member is kept.
+		const members = fragment.type.filter((t) => t !== 'null');
+		if (members.length !== fragment.type.length) attr.nullable = true;
+		if (members.length > 0) attr.type = members[0];
```
kriszyp, line 179: Could we avoid dropping every non-first union member here? `static properties` is typed as JSON Schema, and MCP accepts type arrays, so `{ type: ['string', 'number'] }` currently becomes string-only; `{ type: ['null'] }` becomes untyped. Please preserve the source union for MCP and explicitly translate it for OpenAPI 3.0 (`nullable` for `T | null`, `oneOf` for genuine unions), with a consumer-level regression test.

resources/openApi.ts
```diff
@@ -151,21 +161,40 @@ export function generateJsonApi(resources: Resources, serverHttpURL: string) {
...
 					} else {
 						props[name] = new Type(DATA_TYPES[type], type);
 					}
 				}
-				// Attach per-property description so it surfaces in Swagger UI / Redoc.
-				if (description && props[name] && typeof props[name] === 'object' && !('$ref' in props[name])) {
-					(props[name] as { description?: string }).description = description;
+				// Attach per-property JSON-Schema hints (description/enum/format/const) so they surface in
+				// Swagger UI / Redoc; enum in particular tells clients the allowed values.
+				if (props[name] && typeof props[name] === 'object' && !('$ref' in props[name])) {
```
kriszyp, line 284: The projection turns `type: ['string', 'null']` into `type: 'string', nullable: true`, but this attachment block never emits `nullable`. The final OpenAPI component therefore rejects null while MCP preserves it. Please emit `nullable: true` here and assert the generated component; nullable enums also need an emitted form that accepts null.

resources/openApi.ts
```diff
@@ -151,21 +161,40 @@ export function generateJsonApi(resources: Resources, serverHttpURL: string) {
 						} else {
 							props[name] = { $ref: SCHEMA_COMP_REF + def.type };
 						}
+					} else if (attr.properties) {
+						// Nested object from `static properties` — the shared projector emits the full object
+						// schema (sub-properties recursed); OpenAPI's table path uses $refs instead.
+						props[name] = attributeToFragment(attr);
```
kriszyp: This recursion goes through `attributeToFragment()`, which intentionally omits `enum`/`format`/`const`; the scalar-array branch similarly rebuilds only `type`. As a result, nested fields and array items lose author-declared constraints (for example, `items: { type: 'string', format: 'uuid', enum: [...] }` becomes only `{ type: 'string' }`). Could this use a recursive OpenAPI projector that retains supported hints at every depth, with nested/item assertions?

resources/openApi.ts
```diff
@@ -151,21 +161,40 @@ export function generateJsonApi(resources: Resources, serverHttpURL: string) {
...
-				if (description && props[name] && typeof props[name] === 'object' && !('$ref' in props[name])) {
-					(props[name] as { description?: string }).description = description;
+				// Attach per-property JSON-Schema hints (description/enum/format/const) so they surface in
+				// Swagger UI / Redoc; enum in particular tells clients the allowed values.
+				if (props[name] && typeof props[name] === 'object' && !('$ref' in props[name])) {
+					const prop = props[name] as { description?: string; enum?: unknown; format?: string; const?: unknown };
+					if (description) prop.description = description;
+					if (attr.enum && prop.enum === undefined) prop.enum = attr.enum;
+					if (attr.format && prop.format === undefined) prop.format = attr.format;
+					if (attr.const !== undefined && prop.const === undefined) prop.const = attr.const;
```
kriszyp: Because this document declares OpenAPI 3.0.3, `const` is not a valid Schema Object keyword (and lowercase `type: 'null'` passed through above is likewise not a 3.0 type). Please translate `const` to a single-value `enum` and nullability to OpenAPI 3.0 forms, or upgrade the whole document to 3.1. A spec-validation test would catch this dialect mismatch.

kylebernhardy:
Docs companion opened: [HarperFast/documentation#605 — `docs(resources): document static properties as a first-class MCP/OpenAPI schema source`](https://github.com/HarperFast/documentation/pull/605). It documents the JSON Schema vocabulary, the full fragment-key list, and the union / item-less-array / `required`-on-create resolution rules this PR introduces.

resources/jsonSchemaTypes.ts
```diff
@@ -108,3 +137,65 @@ export function projectAttributesToProperties(attributes: AttributeLike[]): Reco
...
+	} else if (fragment.type != null) {
+		attr.type = fragment.type;
+	}
+	if (fragment.description) attr.description = fragment.description;
```
kriszyp, line 183: Can we not use Object.assign?

resources/openApi.ts
kylebernhardy: Agreed, and this one is already fixed on the stacked follow-up: [#1944](https://github.com/HarperFast/harper/pull/1944) emits `nullable: true` in exactly this attachment block (it was previously computed only to derive `required` and never written onto the property), with an assertion on the generated component in `unitTests/components/mcp/tools/convergence.test.js` plus a new emitted-document integration test.

The nullable-**enum** case you raise is not covered there — an enum whose list excludes `null` still rejects null even with `nullable: true` under 3.0. Tracking that with the dialect issue on your other thread.

resources/openApi.ts
kylebernhardy: Confirmed, and also already fixed on the stacked follow-up [#1944](https://github.com/HarperFast/harper/pull/1944). It replaces this `attributeToFragment()` recursion with a shared, dialect-aware emitter (`attributeToSchema`) used by both the MCP deriver and the OpenAPI generator, so `enum`/`format`/`const`/`description` survive at every depth — nested objects and array items alike — and `@hidden` is suppressed at every level rather than only the top. Assertions cover nested and array-item hints on both surfaces.

The root cause was that `attributeToFragment` was serving two masters: the canonical, front-end-neutral `Table.properties` projection (which must omit those hints) and OpenAPI's nested emit (which must keep them).

resources/openApi.ts
kylebernhardy: You are right, and this is the finding I had no answer for — thank you. The document declares `3.0.3`, whose Schema Object is the draft-04 subset: `const` arrived in draft-06 and `type: 'null'` is not a 3.0 type (3.0 expresses nullability only via `nullable`). So we are emitting keywords the declared dialect does not define.

Worse, the stacked follow-up [#1944](https://github.com/HarperFast/harper/pull/1944) *widens* it — it carries `const` down into nested objects and array items too, and I documented `const` as a supported hint in the docs companion ([documentation#605](https://github.com/HarperFast/documentation/pull/605)). All three need to change together.

Neither this PR nor #1944 addresses it, and I don't think it should be bolted onto either as a side fix — picking between "translate to 3.0 forms" (`const` → single-value `enum`, drop `type: 'null'` in favor of `nullable`, nullable-enum needs `null` appended to the list) and "declare 3.1" (which changes the contract for every existing consumer of the document, not just static-properties resources) is a real decision with downstream impact. Raising it with Kyle to scope as its own change, with the spec-validation test you suggest — that test is the thing that would have caught this and is worth having regardless of which direction we take.

resources/jsonSchemaTypes.ts
kylebernhardy: Partly addressed, partly not — splitting the two cases:

**`T | null`** is handled on the stacked follow-up [#1944](https://github.com/HarperFast/harper/pull/1944): MCP now re-expands to the source union (`{ type: ['string','null'] }`) rather than emitting a `nullable` keyword it has no use for, and OpenAPI emits `nullable: true` at top level and nested.

**A genuine multi-type union** (`['string','number']`) still collapses to its first member, on both PRs — the comment on this line admits it. Your `oneOf` suggestion is the right translation for OpenAPI, and MCP should simply pass the array through untouched since it accepts type arrays. That is not a small change to the projection: `AttributeLike` has a single `type` field, so preserving a union needs the attribute form to carry it, which is the same seam #1944 reworks. I would rather do it there or in a follow-up than partially here.

`{ type: ['null'] }` becoming untyped is a real edge of the same bug and I have no defense for it.

resources/jsonSchemaTypes.ts
kylebernhardy: Happy to drop it. For context on where it lives now: the stacked follow-up [#1944](https://github.com/HarperFast/harper/pull/1944) has one `Object.assign`, merging the per-surface primitive mapping into the fragment being built (`Object.assign(fragment, options.mapPrimitive(attr.type, attr))`). It is there because each surface's mapper returns a variable shape — MCP's `Date` yields `{ type: ['string','number'], description }` and `Bytes` yields `{ type, contentEncoding }`, so there is no fixed field list to copy.

If the objection is the opacity, I can replace it with explicit field copies for the keys we actually support, which also stops an unexpected key from a mapper leaking into the emitted schema. Say which you prefer and I will make the change on #1944.

resources/openApi.ts
kylebernhardy: Fixed across all three artifacts, taking the conservative direction — translate to valid 3.0.3 forms rather than declaring 3.1, since upgrading the document changes the contract for every existing consumer, not just static-properties resources. If you'd rather go to 3.1, that's a separate decision and this doesn't foreclose it.

- **harper#1921** (`d10658161`): `const` is emitted as the equivalent single-value `enum`; a bare `type: 'null'` becomes an untyped `{ nullable: true }`, since 3.0 expresses nullability only via the keyword.
- **harper#1944** (`d00549032`): the same translation inside the shared emitter, so it holds at depth — the dialect branch lives in `attributeToSchema`, with MCP keeping `const` (it speaks current JSON Schema) and OpenAPI getting `enum`.
- **[documentation#605](https://github.com/HarperFast/documentation/pull/605)** (`cfdbab61`): the reference and guide told authors `const` surfaces as written, which was never true. They now say: write `const`, expect `enum: [value]` in the OpenAPI output.

Your spec-validation suggestion is the part that earned its keep. I wrote it to walk the **entire** generated document for post-draft-04 keywords and `null` types rather than assert one property — and it immediately failed on #1944, catching six nested and array-item paths (`components.schemas.Widget.properties.nested.properties.inner.const`, `…list.items.const`, and the parameter copies) that my top-level fix had missed. Without that test I'd have shipped a half-fix and called it done.

Still outstanding from your other thread and not addressed here: nullable **enums** need `null` in the value list to actually accept null, since `nullable: true` alone doesn't extend an enum. Happy to fold that in if you want it on this PR, or leave it with the genuine-union `oneOf` work.

resources/openApi.ts
```diff
@@ -151,21 +161,46 @@ export function generateJsonApi(resources: Resources, serverHttpURL: string) {
...
+						if (!elements) {
+							// `{ type: 'array' }` with no items — valid JSON Schema (array of anything).
+							props[name] = { type: 'array' };
+						} else if (elements.properties) {
+							// array of nested objects — project the element to its full object schema
+							props[name] = { type: 'array', items: attributeToFragment(elements) };
+						} else if (elements.type === 'Any') {
 							props[name] = { type: 'array', items: { format: elements.type } };
+						} else if (!DATA_TYPES[elements.type] && JSON_SCHEMA_SCALAR_TYPES.has(elements.type)) {
+							props[name] = { type: 'array', items: new Type(elements.type) };
```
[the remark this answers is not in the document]
kylebernhardy: Addressed in 1eb8aa25b, which landed after this review was generated. The nested-object and array-item paths now route through `toOpenApiDialect()`, so `type: 'null'` and type unions are translated on those branches too rather than only in the top-level scalar case.

Comment generated by kAIle (Claude Opus 4.8)

resources/openApi.ts
```diff
@@ -151,21 +213,50 @@ export function generateJsonApi(resources: Resources, serverHttpURL: string) {
...
 						}
 					} else if (type === 'Any') {
 						props[name] = { format: type };
+					} else if (!DATA_TYPES[type] && JSON_SCHEMA_SCALAR_TYPES.has(type)) {
+						// OpenAPI 3.0.3 has no `'null'` type — nullability is expressed by the `nullable`
+						// keyword, so a bare `type: 'null'` becomes an untyped nullable schema.
+						props[name] = type === 'null' ? { nullable: true } : new Type(type);
 					} else {
 						props[name] = new Type(DATA_TYPES[type], type);
 					}
```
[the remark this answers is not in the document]
kylebernhardy: Confirmed and fixed in f34cedef9 — this was a real gap, and it corroborates the same point Kris raised on this line earlier.

The scalar branch now sets `nullable: true` when the attribute is nullable, and widens a co-declared `enum` with `null` (3.0's `nullable` does not widen an enum on its own). `const` now intersects with a co-declared `enum` here too, matching what the nested path already did.

On the "route the scalar case through `toOpenApiDialect` as well" suggestion: not taken here. The two paths consume different inputs — `toOpenApiDialect` takes a `JsonSchemaFragment`, while this branch works from an `Attribute` and interleaves relationship/`$ref`/definition cases that have no fragment form. Unifying them is the right end state, and that convergence is the subject of #1941/#1942 (PR #1944, stacked on this branch); doing it here would pull that refactor into this PR.

Test gap was real too: the fixture had `maybe`/`nullableEnum`/`nullableConst` but the assertions only checked for *absence* of invalid keywords. Added two tests asserting the translated output positively.

Comment generated by kAIle (Claude Opus 4.8)

resources/openApi.ts
kylebernhardy: You were right and my earlier reply here was premature — the dialect commits fixed the nested and array paths but left this attachment block still dropping `nullable`, so the scalar case stayed broken. Fixed properly in f34cedef9:

- `nullable: true` is now emitted on the scalar schema when the attribute is nullable (covers both an author-declared `nullable` and a `["string","null"]` union, which folds to the same attribute upstream).
- A nullable `enum` is widened with `null`, since 3.0's `nullable` does not widen an enum on its own.
- `const` now intersects with a co-declared `enum` rather than deferring to it, matching the nested path.

Tests assert the emitted component positively now (`maybe`, `nullableEnum`, `nullableConst`) instead of only checking that invalid keywords are absent — the old assertions would have passed just as happily with nullability dropped, which is why this survived.

Still open from your review and not addressed here: genuine multi-type unions (`["string","number"]`) still collapse to the first member rather than emitting `oneOf`. Leaving that for the convergence work in #1944 unless you want it in this PR.

Comment generated by kAIle (Claude Opus 4.8)

resources/jsonSchemaTypes.ts
kylebernhardy: Done properly now in 2039e4dec — I stop deferring this to the follow-up.

`AttributeLike` carries the source union on a new `types` field, verbatim. `type` still holds the first non-null member so the single-type consumers (validation, query coercion) are untouched. Each surface then expresses it in its own dialect:

- **MCP** passes the array through as declared — it accepts type unions, as you said.
- **OpenAPI 3.0** emits `oneOf`, your suggested translation. `nullable` rides alongside it for `["string","number","null"]`.
- `{ type: ["null"] }` no longer goes untyped. On the stacked [#1944](https://github.com/HarperFast/harper/pull/1944) it emits `{ nullable: true, enum: [null] }` — 3.0 has no `null` type, but a null-only enum is a form it can actually express, which beats staying silent about a field the author did describe.

The projection round-trips too: `properties -> attributes -> properties` returns the declared union rather than a collapsed scalar.

Consumer-level regression tests as you asked: `oneOf` emission, nullable-alongside-union, union round-trip, and a convergence test asserting both surfaces describe the same declaration without either narrowing it. On #1944 the translation lives in the shared emitter, so it applies at every nesting level rather than only the top — with a nested-union test to hold that.

Comment generated by kAIle (Claude Opus 4.8)

resources/jsonSchemaTypes.ts
kylebernhardy: Dropped it — 61d2773fa on [#1944](https://github.com/HarperFast/harper/pull/1944) copies the mapper result field by field instead.

Worth noting what that turned up: the wholesale merge was letting `contentEncoding` into emitted MCP schemas even though `JsonSchemaFragment` never declared it — the `as JsonSchemaFragment` cast on the mapper hid it. It is declared explicitly now. That is exactly the leak the explicit copy prevents, so this was the right call rather than just a style preference.

Comment generated by kAIle (Claude Opus 4.8)

resources/openApi.ts
```diff
@@ -1,10 +1,67 @@
...
+	const out: JsonSchemaFragment = rest;
+	// 3.0 has no `'null'` type and no type unions; nullability is the `nullable` keyword alone.
+	if (Array.isArray(out.type)) {
+		const members = out.type.filter((t) => t !== 'null');
+		if (members.length !== out.type.length) out.nullable = true;
+		if (members.length > 0) out.type = members[0];
+		else delete out.type;
+	} else if (out.type === 'null') {
+		delete out.type;
+	}
```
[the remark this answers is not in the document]
kylebernhardy: Confirmed and fixed in eaa755b48. You were right to re-flag it — 2039e4dec touched this exact function and I did not audit the sibling branch, which is the same miss that produced this gap in the first place.

All three sites now agree, and they express the declaration rather than merely surviving it: a null-only `type` emits `{ nullable: true, enum: [null] }`. 3.0 has no `'null'` type, and a bare `nullable` on an untyped schema constrains nothing, so the null-only enum is what actually carries the author's meaning — the previous top-level `{ nullable: true }` was only half an answer.

Test asserts all three depths (top-level, nested sub-property, array item) in one place, so a future change to one branch cannot silently diverge from the others again.

Comment generated by kAIle (Claude Opus 4.8)

resources/openApi.ts
```diff
@@ -1,10 +1,89 @@
...
+		const members = out.type.filter((t) => t !== 'null');
+		if (members.length !== out.type.length) out.nullable = true;
+		if (members.length > 1) {
+			delete out.type;
+			out.oneOf = members.map((member) => ({ type: member }));
+		} else if (members.length === 1) out.type = members[0];
+		else delete out.type;
+	} else if (out.type === 'null') {
+		delete out.type;
+	}
```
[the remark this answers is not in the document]
kylebernhardy: Confirmed and fixed in eaa755b48. You were right to re-flag it — 2039e4dec touched this exact function and I did not audit the sibling branch, which is the same miss that produced this gap in the first place.

All three sites now agree, and they express the declaration rather than merely surviving it: a null-only `type` emits `{ nullable: true, enum: [null] }`. 3.0 has no `'null'` type, and a bare `nullable` on an untyped schema constrains nothing, so the null-only enum is what actually carries the author's meaning — the previous top-level `{ nullable: true }` was only half an answer.

Test asserts all three depths (top-level, nested sub-property, array item) in one place, so a future change to one branch cannot silently diverge from the others again.

Comment generated by kAIle (Claude Opus 4.8)

resources/jsonSchemaTypes.ts
```diff
@@ -108,3 +149,67 @@ export function projectAttributesToProperties(attributes: AttributeLike[]): Reco
...
+/**
+ * Structural inverse of `attributeToFragment`: rebuild an attribute from a JSON Schema fragment.
+ * A programmatic Resource may declare `static properties` (the Record form) without populating the
+ * `attributes` Array; the schema-derivation paths (MCP `derive.ts`, OpenAPI) read attributes, so a
+ * bare declaration would otherwise yield a skeletal schema. Projecting the fragments back into
+ * attributes lets those paths produce the same rich schema they build for table-backed resources.
+ */
+function fragmentToAttribute(name: string, fragment: JsonSchemaFragment): AttributeLike {
+	const attr: AttributeLike = { name };
+	if (fragment.properties) {
```
kriszyp, line 162: This shape-first branch runs before the union branch, so structural unions do not survive the projection. For example, `{ type: ['object', 'null'], properties: { x: { type: 'string' } } }` takes this branch and loses both `types` and `nullable`; `{ type: ['array', 'null'], items: { type: 'string' } }` takes the union branch later but never records `elements`, so it becomes an unconstrained array. Both MCP and OpenAPI then advertise a different contract from the declared JSON Schema. Please process the type union orthogonally to `properties`/`items` (and add round-trip cases for nullable object and array schemas) so both the structural shape and every union member survive.

resources/openApi.ts
```diff
@@ -151,21 +240,69 @@ export function generateJsonApi(resources: Resources, serverHttpURL: string) {
...
 							props[name] = { $ref: SCHEMA_COMP_REF + def.type };
 						}
+					} else if (attr.properties) {
+						// Nested object from `static properties` — the shared projector emits the full object
+						// schema (sub-properties recursed); OpenAPI's table path uses $refs instead.
+						props[name] = toOpenApiDialect(attributeToFragment(attr));
+					} else if (union) {
+						// A genuine multi-type union (`['string','number']`). 3.0 has no type arrays, so the
+						// equivalent is `oneOf`; `attr.type` alone would drop every member but the first.
+						props[name] = { oneOf: union.map(openApiUnionMember) };
```
kriszyp, line 250: For `type: ['string', 'integer', 'null']`, this creates an outer `{ oneOf: [...] }`, and line 301 later adds `nullable: true` to that outer object. In OpenAPI 3.0.3, `nullable` only affects a `type` explicitly declared in the same Schema Object; this outer object has no `type`, so strict consumers still reject `null`. Please encode null in a branch that can actually admit it (for example, make exactly one typed `oneOf` member nullable) and validate the generated schema against a null instance in the test.

components/mcp/tools/schemas/derive.ts
```diff
@@ -79,11 +92,17 @@ function attributeToProperty(attr: HarperAttribute): object {
 			type: 'object',
 			properties: Object.fromEntries(attr.properties.map((p) => [p.name, attributeToProperty(p)])),
 		};
+		if (attr.required) base.required = attr.required;
```
kriszyp, line 95: The recursive mapping bypasses `attributeVisible`, so `hidden` is enforced only at the top level. A declaration such as `profile.properties.secret = { type: 'string', hidden: true, description: 'internal' }` exposes `secret` and its description to MCP clients; if `secret` is in the nested `required` array, clients are also told they must send the hidden field. OpenAPI already filters this case. Please suppress hidden attributes at every recursive depth and prune suppressed names from the corresponding `required` list.

resources/openApi.ts
```diff
@@ -151,21 +240,69 @@ export function generateJsonApi(resources: Resources, serverHttpURL: string) {
 						} else {
 							props[name] = { $ref: SCHEMA_COMP_REF + def.type };
 						}
+					} else if (attr.properties) {
+						// Nested object from `static properties` — the shared projector emits the full object
+						// schema (sub-properties recursed); OpenAPI's table path uses $refs instead.
+						props[name] = toOpenApiDialect(attributeToFragment(attr));
```
kriszyp, line 246: `attributeToFragment` deliberately omits `enum`, `format`, and `const` because it also feeds the front-end-neutral `Table.properties` projection. Using it as the consumer emitter therefore drops those constraints from every nested object; the type-only array branches likewise drop item-level hints. For example, a nested `{ type: 'string', enum: ['open', 'closed'] }` or array item with `const`/`format` is retained by MCP but disappears from OpenAPI. Please use a consumer-facing recursive emitter (or propagate these hints explicitly at every depth) and add convergence assertions for nested and item-level hints.

components/mcp/tools/application.ts
```diff
@@ -1356,7 +1357,9 @@ function buildApplicationTools(resources: ResourcesRegistry): void {
 		const tableName = ResourceClass?.tableName;
 		const suffix = uniqueSuffix(path, databaseName, claimedSuffixes);
 		claimedSuffixes.add(suffix);
-		const attributes = (ResourceClass?.attributes ?? []) as HarperAttribute[];
+		// A programmatic Resource may declare `static properties` without an `attributes` Array; resolve
+		// the effective attributes so its verb tools get a rich inputSchema instead of a skeletal one.
+		const attributes = resolveAttributes(ResourceClass) as HarperAttribute[];
```
kriszyp, line 1362: Projected JSON-Schema properties normally have `nullable === undefined`, but `deriveCreateSchema` interprets `!attr.nullable` as required. Consequently `{ id: { primaryKey: true, type: 'string' }, label: { type: 'string' } }` yields MCP `required: ['label']`, while this PR's OpenAPI path only adds a property when `nullable === false` and leaves `label` optional. MCP clients can reject otherwise valid create calls before dispatch. Please carry presence/requiredness explicitly through the projection, or otherwise align the undefined case across both surfaces without conflating optionality with nullability.

resources/openApi.ts
```diff
@@ -151,21 +240,69 @@ export function generateJsonApi(resources: Resources, serverHttpURL: string) {
...
+					} else if (union) {
+						// A genuine multi-type union (`['string','number']`). 3.0 has no type arrays, so the
+						// equivalent is `oneOf`; `attr.type` alone would drop every member but the first.
+						props[name] = { oneOf: union.map(openApiUnionMember) };
 					} else if (type === 'array') {
-						if (elements.type === 'Any') {
+						if (!elements) {
+							// `{ type: 'array' }` with no items — valid JSON Schema (array of anything).
+							props[name] = { type: 'array' };
```
kriszyp, line 254: Although `{ type: 'array' }` is valid general JSON Schema, OpenAPI 3.0.3 specifically requires `items` whenever `type` is `array`. Emitting the bare schema makes the generated 3.0.3 document invalid for strict tooling. Please translate an unconstrained array to `{ type: 'array', items: {} }` (and do the same for array union members), then assert the OpenAPI-specific requirement rather than the source-dialect rule.

resources/openApi.ts
```diff
@@ -151,21 +240,69 @@ export function generateJsonApi(resources: Resources, serverHttpURL: string) {
...
+					if (attr.format) prop.format = attr.format;
+					// Intersect rather than defer: `const` narrows an `enum` declared alongside it.
+					if (attr.const !== undefined) {
+						prop.enum = Array.isArray(prop.enum) ? prop.enum.filter((value) => value === attr.const) : [attr.const];
+					}
+					// `type: ['string', 'null']` folds to `type: 'string'` + `nullable` upstream; without this the
+					// scalar path emits the type alone and the document claims the field rejects null.
+					if (nullable) prop.nullable = true;
+					// 3.0's `nullable` does not widen an `enum` — without `null` in the list a validator rejects it.
+					if (prop.nullable && Array.isArray(prop.enum) && !prop.enum.includes(null)) {
```
kriszyp, line 303: This unconditionally changes the intersection semantics of a standard JSON-Schema union. `{ type: ['string', 'null'], enum: ['a'] }` rejects null because every keyword must match, but projection derives `nullable: true` from the type union and this line rewrites the enum to `['a', null]`; the same issue occurs with `const`. Please preserve whether null came from the authored `type` union versus an explicit Harper `nullable` declaration, and do not add null to an enum/const that excluded it.

kriszyp (changes requested):
Good. A few mechanical issues to work through.

components/mcp/tools/schemas/derive.ts
```diff
@@ -63,41 +87,52 @@ function harperTypeToJsonSchema(type: string | undefined): { type: string | stri
...
-	return base;
+/**
+ * Emit an attribute as an MCP property schema. The traversal (nesting, arrays, `hidden` suppression,
+ * hint propagation, nullability) is shared with the OpenAPI generator so the two can't describe the
+ * same fragment differently; only the Harper-primitive mapping below is MCP-specific.
+ *
+ * Returns `undefined` for a `hidden` attribute — callers skip it.
+ */
+function attributeToProperty(attr: HarperAttribute): object | undefined {
+	return attributeToSchema(attr as AttributeLike, {
```
kriszyp, line 107: `isDeclaredReference` only reaches `mapPrimitive`, which is the leaf path — the shared emitter checks `attr.properties` first (`resources/jsonSchemaTypes.ts:386`), and the GraphQL parser attaches the *related type's own attribute array* to every complex-typed attribute (`resources/graphql.ts:308`, non-enumerably). So for the ordinary bidirectional pattern — `type Author { books: [Book] @relationship(...) }` / `type Book { author: Author @relationship(...) }` — `deriveGetOutputSchema` walks `books` → `books.elements.properties` (Book's attributes) → `author.properties` (Author's attributes) → `books`, hits `ancestors.has(attr)` at `resources/jsonSchemaTypes.ts:331`, and throws `SchemaTraversalError`. `registerVerbTools` catches it and the table loses *every* verb tool with only a warn line. (Pre-PR this recursed forever instead, so the containment is an improvement — but the outcome is still a table with no MCP tools.)

Even without a cycle, this is the one place the two surfaces still describe the same attribute differently: OpenAPI emits `{ $ref: '#/components/schemas/Book' }` (`resources/openApi.ts:190-199`) while MCP inlines Book's entire record shape into the argument schema, which is both a token cost and a contract the tool can't accept.

Please treat a `relationship`/`definition` attribute as a reference *before* recursing — skip the `attr.properties` branch for it and emit the same shape OpenAPI does (or an untyped `{}`), and add a derive test built from a genuine two-way relationship fixture so the cycle path is covered rather than only the synthetic self-referencing array in `unitTests/components/mcp/tools/application.test.js`.

resources/jsonSchemaTypes.ts
```diff
@@ -102,9 +147,346 @@ export function attributeToFragment(attr: AttributeLike): JsonSchemaFragment {
...
+			// schema does not define makes the object unsatisfiable for any client that validates. Omit the
+			// key entirely when nothing survives: JSON Schema draft-04 (which OpenAPI 3.0.3 inherits)
+			// requires `required` to have at least one element, so `required: []` fails validators.
+			if (attr.required) {
+				const visibleRequired = attr.required.filter((name) => !suppressedNames.has(name));
+				if (visibleRequired.length > 0) fragment.required = visibleRequired;
+			}
+			if (attr.additionalProperties !== undefined) fragment.additionalProperties = attr.additionalProperties;
+		} else if (attr.type === 'array') {
+			fragment.type = 'array';
```
kriszyp, line 524: Although `{ type: 'array' }` is valid general JSON Schema, OpenAPI 3.0.3 specifically requires `items` whenever `type` is `array`. Emitting the bare schema makes the generated 3.0.3 document invalid for strict tooling. Please translate an unconstrained array to `{ type: 'array', items: {} }` (and do the same for array union members), then assert the OpenAPI-specific requirement rather than the source-dialect rule.

Re-anchored from the old `openApi.ts` branch: the emitter is now dialect-aware, so the `openapi-3.0.3` arm can supply the default `items` here while `json-schema` keeps the bare form. Both entry points are affected — this branch and `structuralUnionMember('array')` at `resources/jsonSchemaTypes.ts:354` when no `elements` exist. Note `unitTests/resources/openApi.test.js` currently pins the bare form (`expect(schema.properties.anything).to.deep.equal({ type: 'array' })`), so that assertion moves with the fix.

resources/jsonSchemaTypes.ts
```diff
@@ -102,9 +147,346 @@ export function attributeToFragment(attr: AttributeLike): JsonSchemaFragment {
...
+				fragment.const = attr.const;
+			}
+		}
+		// Nullability does not widen an `enum` in either dialect — a validator still rejects `null` unless
+		// the value list contains it. Keep the type/null arm and the value constraint consistent.
+		const admitsNull =
+			fragment.nullable === true ||
+			(Array.isArray(fragment.type) && fragment.type.includes('null')) ||
+			fragment.anyOf?.some((arm) => arm.type === 'null' || arm.enum?.includes(null));
+		if (attr.const === undefined && admitsNull && fragment.enum && !fragment.enum.includes(null)) {
```
kriszyp: This unconditionally changes the intersection semantics of a standard JSON-Schema union. `{ type: ['string', 'null'], enum: ['a'] }` rejects null because every keyword must match, but projection derives `nullable: true` from the type union and this line rewrites the enum to `['a', null]`; the same issue occurs with `const`. Please preserve whether null came from the authored `type` union versus an explicit Harper `nullable` declaration, and do not add null to an enum/const that excluded it.

Update against the current code: the `const` half is fixed — the `attr.const === undefined` guard here keeps `const` stricter than nullability, and the tests pin it. What remains is the `enum` half, and the distinction the comment asks for is still not carried: `fragmentToAttribute` sets `attr.nullable = true` from a `'null'` union member (`resources/jsonSchemaTypes.ts:188`) and from an explicit `nullable` keyword (`resources/jsonSchemaTypes.ts:223`) into the same field, so the emitter cannot tell them apart. Worth weighing before changing it: for the *Harper-keyword* source the widened list actually matches enforcement, since the contract validator short-circuits `null` before the enum check (`resources/defineResource.ts:296-298`). It is specifically the union-declared case that publishes a value the author's own declaration excludes — so either record the provenance and skip the widening there, or state explicitly in the emitter comment that Harper's nullability always wins over `enum` and accept the divergence from strict JSON-Schema readers.

resources/jsonSchemaTypes.ts
```diff
@@ -102,9 +147,346 @@ export function attributeToFragment(attr: AttributeLike): JsonSchemaFragment {
...
+		if (attr.enum && fragment.enum === undefined) fragment.enum = attr.enum;
+		if (attr.format) fragment.format = attr.format;
+		if (attr.const !== undefined) {
+			// `const` is JSON Schema draft-06. OpenAPI 3.0.3's Schema Object is the draft-04 subset, so it
+			// has no such keyword — emit the equivalent single-value `enum` for that dialect. MCP speaks
+			// current JSON Schema and takes `const` directly.
+			if (options.dialect === 'openapi-3.0.3') {
+				// Intersect rather than skip when both are declared: dropping the `const` would leave OpenAPI
+				// advertising a wider value set than MCP, from one declaration.
+				fragment.enum = fragment.enum
```
kriszyp, line 553: When both `enum` and `const` are declared and they don't intersect — `{ type: 'string', enum: ['a', 'b'], const: 'c' }` — this filter yields `enum: []`, which is exactly the draft-04 violation the PR is careful to avoid for `required` (`resources/jsonSchemaTypes.ts:404`, `resources/openApi.ts:583`): draft-04's `enum` MUST have at least one element, so strict validators reject the whole document over one contradictory property declaration. The MCP side is unaffected (it keeps both keywords, unsatisfiable but syntactically valid).

Suggest treating an empty intersection as an authoring error rather than emitting it: warn once (the `resolveDeclaredType` warn-once pattern fits) and emit the `const` value alone, or drop the property's value constraints entirely. Either way the walking dialect-compliance suite in `unitTests/resources/openApi.test.js` would be worth extending with a `const`-outside-`enum` fixture, since the current fixtures never produce an empty list.

components/mcp/resources.ts
```diff
@@ -936,6 +943,23 @@ function filterAttributesByPermissions(attributes: any[], attributePermissions:
 	return attributes.filter((a) => !denied.has(a?.name));
 }
 
+function filterHiddenAttributes(attributes: any[]): any[] {
```
kriszyp: `readTableSchema` wraps `resolveAttributes` in a try/catch so a malformed `static properties` becomes an actionable read failure — but this clone runs *outside* that guard and has neither the depth cap nor the cycle detection the rest of the PR added (`resources/jsonSchemaTypes.ts:329-331`). For the Array-form path (`resource.attributes`, returned verbatim by `resolveAttributes`) a cyclic graph recurses until the stack blows; that is the same shape `unitTests/components/mcp/tools/application.test.js` constructs to prove the MCP tool path contains it, so it isn't hypothetical. The enumerability check spares the parser's relationship graphs, but only because those are attached non-enumerably — an author-built or future parser-built enumerable cycle isn't covered.

Suggest moving the call inside the existing `try` and reusing the shared traversal's guards (or passing a `WeakSet` + depth through this helper), so a cyclic schema fails the read with the same `invalid schema for db.table` message instead of a RangeError escaping to the transport.

components/mcp/tools/application.ts
```diff
@@ -1363,21 +1375,36 @@ function buildApplicationTools(resources: ResourcesRegistry): void {
...
+				attributes = resolveAttributes(ResourceClass) as HarperAttribute[];
+			} catch (error) {
+				warnInvalidResourceSchema(ResourceClass, path, error);
+			}
+			if (attributes) {
+				const databaseName = ResourceClass?.databaseName;
+				const tableName = ResourceClass?.tableName;
+				const suffix = uniqueSuffix(path, databaseName, claimedSuffixes);
+				claimedSuffixes.add(suffix);
+				try {
```
kriszyp, line 1391: Two small asymmetries in the containment path:

1. `registerVerbTools` registers tools one at a time (`components/mcp/tools/application.ts:865` onward), so a `SchemaTraversalError` raised by a later verb leaves the earlier ones published. `deriveSearchSchema` never emits attribute schemas, so a Resource with `search` but no `get` (search-only, or `post`-only + `search`) can end up with `search_X` live and `create_X`/`update_X` silently missing, plus a `toolsRegistered` count that doesn't match what's in the registry. Deriving all the schemas first and only calling `addTool` once they all succeed would make the skip all-or-nothing, matching the warning's "skipped" wording.

2. When attribute resolution fails, the suffix is never claimed (the `uniqueSuffix`/`add` pair lives inside the `if (attributes)` block), whereas the custom-only branch above deliberately claims one to keep names stable (`components/mcp/tools/application.ts:1378`). So a colliding sibling's tool name depends on whether *another* resource's schema currently parses, and flips back when the author fixes it. Claiming the suffix unconditionally, as the custom-only branch does, keeps names stable through the failure.

components/mcp/tools/schemas/derive.ts
```diff
@@ -143,16 +178,18 @@ function buildPropertiesObject(
...
 		if (!attributeVisible(attr, permissions, mode)) continue;
 		// Skip auto-managed columns from write inputs — Harper assigns them.
 		if (mode !== 'read' && (attr.assignCreatedTime || attr.assignUpdatedTime || attr.expiresAt)) continue;
 		if (mode !== 'read' && (attr.computed !== undefined || attr.computedFromExpression !== undefined)) continue;
-		properties[attr.name] = attributeToProperty(attr);
-		if (mode === 'insert' && !attr.nullable && !attr.isPrimaryKey) {
+		const schema = attributeToProperty(attr);
+		if (!schema) continue;
+		properties[attr.name] = schema;
+		if (mode === 'insert' && attr.nullable === false && !attr.isPrimaryKey) {
```
kriszyp: Now that `static properties` is a first-class authoring surface, this leaves the public API with two spellings for one concept: at the top level a property is required by writing `nullable: false`, but one level down it's a JSON-Schema `required: [...]` array on the parent object (`resources/jsonSchemaTypes.ts:402`). The new fixture shows both in the same declaration — `status: { ..., nullable: false }` next to `profile: { ..., required: ['name'] }` (`integrationTests/fixtures/mcp-static-properties/resources.js:6`). It also conflates two different things: `nullable: false` says "may not be null", and this line reads it as "must be present", which is why the old `!attr.nullable` reading was wrong in the first place.

No objection to keeping `nullable: false` as the compatibility bridge for GraphQL `String!`, but worth deciding now, before the surface is documented: is there a way for a programmatic author to declare top-level requiredness in the JSON-Schema vocabulary they're already writing (a sibling `static required = [...]`, say)? If the answer is "no, `nullable: false` is the one way", that's worth saying in the `static properties` doc comment so authors don't reach for `required` and get silence.

resources/jsonSchemaTypes.ts
```diff
@@ -102,9 +147,346 @@ export function attributeToFragment(attr: AttributeLike): JsonSchemaFragment {
...
+				if (visibleRequired.length > 0) fragment.required = visibleRequired;
+			}
+			if (attr.additionalProperties !== undefined) fragment.additionalProperties = attr.additionalProperties;
+		} else if (attr.type === 'array') {
+			fragment.type = 'array';
+			// `{ type: 'array' }` with no element schema is valid — an array of anything.
+			if (attr.elements) {
+				const items = emitAttributeSchema(attr.elements, childOptions, ancestors, depth + 1);
+				if (items) fragment.items = items;
+			}
```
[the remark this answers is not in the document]
kylebernhardy: Addressed in `af3746c00`: the shared emitter now supplies `items: {}` whenever OpenAPI 3.0.3 emits an array without a surviving element schema, including unconstrained array union members. MCP retains the valid JSON Schema bare-array form. Unit coverage and the real `/openapi` + MCP integration fixture cover both dialects; the new assertion fails on the pre-fix head.\n\nComment generated by kAIle (GPT-5.6 Sol)

resources/jsonSchemaTypes.ts
```diff
@@ -102,9 +147,346 @@ export function attributeToFragment(attr: AttributeLike): JsonSchemaFragment {
...
+		}
+		// Nullability does not widen an `enum` in either dialect — a validator still rejects `null` unless
+		// the value list contains it. Keep the type/null arm and the value constraint consistent.
+		const admitsNull =
+			fragment.nullable === true ||
+			(Array.isArray(fragment.type) && fragment.type.includes('null')) ||
+			fragment.anyOf?.some((arm) => arm.type === 'null' || arm.enum?.includes(null));
+		if (attr.const === undefined && admitsNull && fragment.enum && !fragment.enum.includes(null)) {
+			fragment.enum = [...fragment.enum, null];
+		}
```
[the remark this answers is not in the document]
kylebernhardy: Addressed in `af3746c00`: enum widening now applies only to Harper’s explicit `nullable` form when the source type union does not already carry `null`. A structural `type` union therefore retains the authored enum intersection, while OpenAPI still carries its null arm; the existing explicit-nullable enum behavior remains unchanged. Added cross-surface unit and real-server integration assertions, with fails-on-pre-fix evidence.

components/mcp/tools/schemas/derive.ts
kylebernhardy: Addressed on the current head (`af3746c00e`). Hidden filtering now lives in the shared recursive consumer emitter, so it applies at every nesting level for MCP and OpenAPI. When a nested property is suppressed, its name is also removed from the enclosing `required` list; an empty required list is omitted. Nested hidden and required-list coverage was added.

resources/openApi.ts
kylebernhardy: Addressed on the current head (`af3746c00e`). MCP and OpenAPI now use a shared recursive consumer-facing emitter rather than `attributeToFragment`, while the front-end-neutral projection remains separate. Nested properties and array items retain `enum`, `format`, `const`, and descriptions, with the OpenAPI-specific `const` translation applied at every depth. Convergence tests cover nested and item-level hints.

resources/openApi.ts
kylebernhardy: Addressed on the current head (`af3746c00e`). OpenAPI no longer puts `nullable` on an untyped outer union object. The shared emitter represents the genuine union with `anyOf` and adds an explicit null-only arm as `{ nullable: true, enum: [null] }`, while preserving the typed member schemas. Tests cover the emitted nullable multi-type union on both surfaces.

resources/openApi.ts
kylebernhardy: Addressed in `af3746c00e`. For the OpenAPI 3.0.3 dialect, the shared emitter now supplies `items: {}` whenever an array has no surviving element schema, including unconstrained array union members. MCP retains the valid general JSON Schema bare-array form. Unit and real `/openapi` integration coverage assert the dialect-specific outputs.

components/mcp/tools/application.ts
kylebernhardy: Addressed on the current head (`af3746c00e`). MCP create schemas now match OpenAPI requiredness: an input property is required only when `nullable === false`; an undefined nullable value remains optional. This avoids treating absence as non-nullability and keeps optionality aligned across both surfaces. Cross-surface tests cover the undefined and explicit-false cases.

resources/openApi.ts
kylebernhardy: Addressed in `af3746c00e`. The emitter now preserves the provenance of nullability: a structural type union containing `null` no longer widens a co-declared `enum` or `const`, so JSON Schema intersection semantics remain intact. Widening is limited to the explicit Harper `nullable` form. Cross-surface unit and integration tests cover both behaviors.

resources/jsonSchemaTypes.ts
kylebernhardy: Addressed on the current head (`af3746c00e`). The projection now records the source type union independently of `properties` and `items`, so nullable object and array declarations retain both their structural shape and every union member. The shared consumer emitter recursively maps the structural members for MCP and OpenAPI. Round-trip and cross-surface tests cover nullable object and array schemas.

kylebernhardy:
@kriszyp All findings from your review have now been addressed on the current head, and I have replied to each thread with the corresponding implementation and test coverage. The full CI suite is green, and the latest Claude review found no blockers. When you have a chance, could you please re-review?

kriszyp (commented):
Looks like maybe a consistency issue(s) to address.

components/mcp/resources.ts
```diff
@@ -936,6 +943,23 @@ function filterAttributesByPermissions(attributes: any[], attributePermissions:
...
+	return attributes
+		.filter((attribute) => !attribute?.hidden)
+		.map((attribute) => {
+			const visible = { ...attribute };
+			if (Object.prototype.propertyIsEnumerable.call(attribute, 'properties') && Array.isArray(attribute.properties)) {
+				visible.properties = filterHiddenAttributes(attribute.properties);
```
heskew: **Significant — hidden child names still leak through `harper://schema`.** This recursively removes the hidden child object but leaves the copied parent `required` array untouched. I reproduced `properties: { visible, secret: { hidden: true } }, required: ['visible', 'secret']` returning only `visible` under `properties` while still publishing `required: ['visible', 'secret']`. That both reveals the hidden field name and produces a contradictory descriptor. Please prune suppressed names from `visible.required` here (and omit an empty list), matching the invariant already enforced by the shared MCP/OpenAPI emitter.

Found by Codex.

resources/jsonSchemaTypes.ts
```diff
@@ -102,9 +147,343 @@ export function attributeToFragment(attr: AttributeLike): JsonSchemaFragment {
...
+			}
+			attr.types = fragment.type;
+			const members = fragment.type.filter((type) => type !== 'null');
+			if (members.length !== fragment.type.length) attr.nullable = true;
+			if (members.length > 0) attr.type = members[0];
+		} else if (fragment.type != null) {
+			if (typeof fragment.type !== 'string') throw new TypeError(`Schema property "${name}.type" must be a string`);
+			attr.type = fragment.type;
+		}
+		if (fragment.required !== undefined) {
```
heskew, line 198: **Significant — the new fail-contained validation still lets malformed JavaScript schemas poison both emitted surfaces.** This checks that `required` is an array of strings, but not that entries are unique, and the later projection copies `enum` without checking that it is an array. On the exact head, `enum: 'abc'` and `required: ['x', 'x']` both register and are emitted verbatim into MCP and OpenAPI, producing invalid schemas rather than warning and omitting the malformed Resource. Because programmatic Resources are commonly authored in JavaScript, the TypeScript interface is not a runtime guard. Please validate these supported keywords before projection (including array/non-empty `enum` and unique `required`) so they follow the containment path.

Found by Codex.

heskew (changes requested):
Reviewed exact head `af3746c0`. The late array-item and structural enum/null fixes check out: the exact-head build passes, 263 focused MCP/OpenAPI/metadata tests pass, and GitHub CI is green.

Requesting changes for the remaining PR-introduced containment and contract gaps. @kriszyp’s unresolved current-head threads already cover partial per-Resource tool publication and unstable suffixes, invalid OpenAPI `enum: []`, unbounded `harper://schema` recursion, and the top-level requiredness/nullable ambiguity; I am not duplicating those inline. The relationship-graph issue @kriszyp identified also remains, but its prior behavior was already broken, so I’m treating it as adjacent rather than a new regression.

Codex found two distinct exact-head gaps and left them inline: the schema resource still leaks hidden nested names through `required`, and malformed JavaScript metadata such as a scalar `enum` or duplicate `required` list bypasses the new per-Resource containment and publishes invalid MCP/OpenAPI schemas.

Local integration execution was blocked only by the sandbox loopback restriction; the corresponding GitHub integration checks are green. Gemini/agy was unavailable for this pass.

resources/jsonSchemaTypes.ts
kylebernhardy: Addressed in `e00c01991`. Programmatic metadata validation now rejects a `const` outside its co-declared `enum`, so the Resource follows the existing containment path instead of producing an invalid empty OpenAPI enum. Tests cover the rejected declaration.

components/mcp/tools/schemas/derive.ts
kylebernhardy: Addressed in `e00c01991`. Programmatic Resources now support public `static required = [...]`, separating presence from nullability. MCP and OpenAPI use it for top-level requiredness; `nullable: false` remains the GraphQL compatibility signal rather than the programmatic presence API.

components/mcp/tools/schemas/derive.ts
kylebernhardy: Addressed in `304390aed`. Relationship metadata is now terminal before nested traversal: scalar relationships emit an unconstrained schema, array relationships emit an unconstrained array schema, and write schemas omit them. Embedded object definitions still retain their nested schema. A production-style bidirectional relationship regression test covers the cycle path.

components/mcp/resources.ts
kylebernhardy: Addressed in `e00c01991`. `harper://schema` now uses the shared guarded filter inside its containment block, with cycle detection and the same depth limit as the schema emitter. Cyclic metadata returns the actionable invalid-schema error instead of recursing.

resources/jsonSchemaTypes.ts
kylebernhardy: Addressed and retained on the current head. OpenAPI 3.0.3 now emits `items: {}` for unconstrained arrays and array union members, while MCP keeps the valid bare JSON Schema form. Dialect-specific tests cover both paths.

components/mcp/tools/application.ts
kylebernhardy: Addressed in `e00c01991` and covered further in `ca7b57571`. Each suffix is claimed before schema resolution, and all verb descriptors are derived before any tool is registered. Malformed Resources therefore keep naming stable and publish either their complete verb set or none.

components/mcp/resources.ts
kylebernhardy: Addressed in `e00c01991`. The shared schema filter now removes hidden children from the parent required list and omits the list when it becomes empty. The regression test verifies the hidden name is absent from the full `harper://schema` response.

resources/jsonSchemaTypes.ts
kylebernhardy: Addressed in `e00c01991`, with final edge coverage in `ca7b57571`. Runtime validation now requires nonempty scalar enums, unique string-only required lists, declared top-level required names, and JSON-serializable finite const values. Invalid metadata is contained per Resource across MCP and OpenAPI.

resources/jsonSchemaTypes.ts
kylebernhardy: Addressed and retained on the current head. Structural `type: ["string", "null"]` no longer widens a co-declared `enum` or `const`; widening is limited to the explicit Harper nullable form. Cross-surface tests cover both semantics.

resources/jsonSchemaTypes.ts
```diff
@@ -102,9 +150,469 @@ export function attributeToFragment(attr: AttributeLike): JsonSchemaFragment {
...
+}
+
+function validateTopLevelRequired(required: readonly string[] | undefined, properties: object): readonly string[] {
+	if (required === undefined) return [];
+	if (!Array.isArray(required) || !required.every((name) => typeof name === 'string')) {
+		throw new TypeError('Resource required must be an array of strings');
+	}
+	if (new Set(required).size !== required.length) throw new TypeError('Resource required must contain unique names');
+	for (const name of required) {
+		if (!Object.hasOwn(properties, name))
```
kriszyp, line 342: `required` now behaves three different ways in this PR, and this is the strictest of the three:

- here (top level): a name not in `properties` throws, which takes the containment path and drops the *entire* Resource from MCP and OpenAPI;
- the emitter (`resources/jsonSchemaTypes.ts:519-521`): an undeclared name in a nested `required` passes through untouched — pinned deliberately by the `preserves required undeclared properties while dropping explicitly hidden ones` convergence case;
- `filterAttributeTree` (`resources/jsonSchemaTypes.ts:313-318`): a nested `required` is intersected with the surviving property names, so an undeclared name is silently dropped from `harper://schema`.

JSON Schema permits `required: ['x']` with no `properties.x` (it means "must be present, any value"), which is why the emitter's leniency is right. The asymmetry is what's odd: the same declaration one level up is a fatal error. Worth picking one rule — my suggestion is to keep the throw only for the cases that are unambiguously authoring mistakes (non-array, non-string, duplicates, all of which you already cover on the line above) and let an undeclared name through like the nested path does, or at most warn. If the strictness is intentional, the doc comment on `static required` (`resources/Resource.ts:69`) is the place to say that every name must be declared in `static properties`, since the failure is silent apart from a log line.

Same question, smaller: `fragment.enum` now rejects any non-scalar member (`resources/jsonSchemaTypes.ts:210-217`) while `fragment.const` accepts any JSON-serializable value including objects — `{ enum: [{ a: 1 }] }` is legal JSON Schema and now costs the author their whole Resource.

resources/jsonSchemaTypes.ts
```diff
@@ -102,9 +150,469 @@ export function attributeToFragment(attr: AttributeLike): JsonSchemaFragment {
...
+	attributes?: AttributeLike[];
+	properties?: Record<string, JsonSchemaFragment> | AttributeLike[];
+	required?: readonly string[];
+}): AttributeLike[] {
+	if (source?.attributes?.length) return source.attributes;
+	if (Array.isArray(source?.properties)) return source.properties;
+	if (source?.properties !== undefined) {
+		const required = validateTopLevelRequired(source.required, source.properties);
+		const requiredSignature = JSON.stringify(required);
+		const cached = projectedProperties.get(source.properties);
```
kriszyp, line 607: The cache is keyed on the identity of the `properties` object, so it only invalidates when that object is *replaced*, which is what the new test covers. An in-place mutation — `MyResource.properties.newField = { type: 'string' }`, or deleting a property — keeps returning the stale frozen projection forever, even across a `registrationVersion` bump that rebuilds every tool. The staleness is silent and survives until the process restarts, since nothing else can evict a `WeakMap` entry whose key is still reachable from a live class.

The optimization itself is well aimed: `/openapi` re-projects every programmatic Resource on each request, so this is the one path where it pays. Two ways to keep the win without the trap: hash the projection input (a cheap `JSON.stringify(properties)` signature alongside `requiredSignature`, mirroring what you already do for `required`), or leave identity keying and say in the doc comment that `static properties` is read once per object identity and must be replaced rather than mutated. Given `freezeAttribute` already makes the *output* immutable, documenting the input as immutable too would be consistent.

resources/jsonSchemaTypes.ts
```diff
@@ -102,9 +150,469 @@ export function attributeToFragment(attr: AttributeLike): JsonSchemaFragment {
...
+	if (depth > MAX_SCHEMA_DEPTH)
+		throw new SchemaTraversalError(`Schema attribute "${attr.name}" exceeds ${MAX_SCHEMA_DEPTH} levels`);
+	if (ancestors.has(attr)) throw new SchemaTraversalError(`Schema attribute "${attr.name}" contains a cycle`);
+	ancestors.add(attr);
+	try {
+		const fragment: JsonSchemaFragment = {};
+		const reference = attr.relationship;
+		if (reference) {
+			const target = reference as { database?: string; table?: string; type?: string };
+			const label = [target.database, target.table ?? target.type].filter(Boolean).join('.');
```
kriszyp, line 441: This label never renders for a relationship the GraphQL parser produced. `resources/graphql.ts:237` sets `property.relationship` to the directive's own arguments — `{ from: 'ownerId' }` or `{ to: ... }` — which has no `database`, `table`, or `type`; the target coordinates are attached separately and non-enumerably as `relationshipReference` at `resources/graphql.ts:323-329`. So `label` is `''` for every real relationship and the branch below emits a bare `{}` / `{ type: 'array', items: {} }` with no description at all.

The useful name is already in hand without reaching for `relationshipReference`: `attr.type` is the target type for a scalar relationship (`author: Author`) and `attr.elements.type` for an array one (`books: [Book]`). Using those — or `relationshipReference` when present, for the `database.table` form — would give an LLM the one thing an untyped `{}` doesn't tell it: what this field points at. As written the code reads as if it produces that hint, which is worse than not attempting it.

kylebernhardy:
@heskew Your two requested changes are addressed at `ca7b57571`: hidden nested fields are pruned from `required`, and malformed `enum`/`required` metadata is contained per Resource. I re-ran the focused MCP/OpenAPI/metadata suites: 221 passing. Could you re-review the current head?

kylebernhardy:
@kriszyp @heskew All requested changes have been addressed on the current head, and the review threads are resolved. When you have a chance, could you re-review?\n\nComment generated by kAIle (GPT-5)

kriszyp (approved):
Looks good.

Issue #674, opened 2026-05-21
deploy_component without restart: TypeError instead of actionable error when accessing new component URL


kriszyp:
When a component is deployed via `deploy_component` but Harper has not yet been restarted, the component's routes are not registered. Accessing a URL in that component then throws a server-side `TypeError: rg.getMatch is not a function` rather than a meaningful response.

## Reproduction

1. `deploy_component` a new component without restarting Harper.
2. Navigate to the component's URL (e.g. `/prometheus_exporter/metrics`).
3. **Expected:** 404 or a clear message: "Component deployed but requires a restart to activate."
4. **Actual:** Unhandled `TypeError: rg.getMatch is not a function`

## Fix

Catch the missing-route case (pre-restart component) in the request handler and return an informative response rather than throwing. Options:
- Return HTTP 404 with a JSON body `{ "error": "Component 'X' is deployed but Harper must be restarted before its routes are active." }`
- Or: automatically trigger a restart on successful `deploy_component` (separate discussion).

---

🤖 Filed by Claude on behalf of Kris.

PR #1936, opened 2026-07-24
fix(components): preserve entry events across deploy scans

### LINKED ISSUE #1934: Redeploy leaves stale .env values: loadEnv (and fastifyRoutes) ignore post-deploy re-'add' events — #1817 class, fixed only for jsResource
Since #1806 (5.2.0-beta.2), `deploy_component` of an already-active component no longer flags `restartRequired`; change detection is delegated to the live file watchers. A redeploy swaps in a fresh component directory and recreates the watchers, so the post-deploy chokidar scan re-emits every file as a fresh `add`.

#1817 tracked the resulting stale-serving bug and #1820 fixed it — but only for jsResource (`resources/jsResource.ts` now tracks the files each scope has loaded; a re-`add` of a known file counts as a change). Two handlers still have the bare gate and silently ignore redeployed changes:

- `resources/loadEnv.ts:11` — `if (entry.eventType !== 'add') scope.requestRestart()`. A redeployed `.env` with changed values surfaces as `add`, so the new values are read only by the scanning thread; already-running workers keep the old env, `get_status` stays `restartRequired: false`, and tooling keyed off the flag skips the post-deploy restart. An extra manual restart is required before the new values load. Observed in the field on a replicated 5.2.0-beta.2 cluster; on 5.1.x this worked because every redeploy raised restart-required via the watcher change events, so the standard post-deploy restart picked env changes up.
- `server/fastifyRoutes.ts:42` — same gate; a redeployed route file's re-`add` is ignored the same way. Not yet reported in the field — flagged from code inspection while confirming the loadEnv leg.

Not related to #1513 (that only blocks `HARPER_CONFIG`/`HARPER_DEFAULT_CONFIG`/`HARPER_SET_CONFIG` from component `.env` files).

Suggested fix: extend #1820's known-file re-`add` tracking to these handlers (for loadEnv, diffing the parsed env snapshot and calling `requestRestart()` on delta would also work and avoids false restarts on byte-identical redeploys). Two-phase deploy will need the same semantics — the in-flight two-phase work already extends #1806's restart-required behavior for new components.

kylebernhardy:
## Summary

Closes #1934 and restores the `scope.handleEntry()` event contract that regressed in #1806 while preserving the new-component fix for #674.

Component deploys pause file watchers while replacing the component tree. When the watcher resumed, chokidar performed a fresh scan and reported every surviving file as `add`, reported no deleted files, and made unchanged and modified files indistinguishable. Fixing that separately in each consumer would preserve the incompatible watcher behavior and duplicate incomplete state tracking.

## Fix

`EntryHandler` now retains a compact snapshot and compares every replacement watcher generation with the last consumer-observed state:

- unchanged files and directories remain silent
- modified files emit `change` with their new contents
- new paths emit `add` / `addDir`
- removed paths emit `unlink` / `unlinkDir`, with descendants emitted before their directory
- file/directory type changes emit the old removal followed by the new addition

Snapshots retain path metadata and SHA-256 file digests, not file contents. Watcher generations and per-path sequences prevent callbacks or slow reads from an obsolete generation from emitting events or polluting the new snapshot. Entries emitted by an interrupted initial scan are carried into the next generation, and `ready` is generation-scoped and fires only after the scan and all of that generation's reads complete.

Restart-free redeploys now require layered runtime-equivalence evidence:

- the JS loader records every application-local module actually loaded plus its resolution edges, including transitive relative imports, package imports/self-references, pure-ESM export targets, and extensionless resolution candidates
- native addons delegate to Node's `.node` loader in CJS, ESM, and compartment modes and conservatively make the runtime restart-required
- package metadata is compared after installation; custom/opaque installs, install scripts, bundled `node_modules`, and unlocked dependency installs fail closed
- extraction and installation are one transaction, retaining the previous component tree until preparation succeeds and restoring it after extraction or install failure

Deploy lifecycle events carry a deployment UUID and owner thread. If the owner worker exits, peers ignore late messages, wait until the worker and its installer process groups are confirmed gone, then reclaim the exact orphaned deployment and resume paused scopes. Overlapping live deployments remain gated.

The consumer-specific redeploy workaround is removed from `jsResource`; `loadEnv`, Fastify routes, JS resources, static files, roles, and plugin `handleEntry` users receive the corrected central events without a new API. Static path ownership is updated incrementally so restart-free static-only deploys serve the replacement content immediately.

## Validation

- `npm run build`
- `npm run lint:required`
- `npm run test:types`
- focused unit suites: green (latest lifecycle/Scope/thread tests: 47 passing; broader redeploy-focused run: 134 passing)
- live Harper deploy proof: **22/22 passing** across four suites
  - identical deploy remains restart-free
  - changed static asset is served immediately without restart
  - transitive relative/package imports, pure-ESM exports retargeting, and extensionless resolution changes require restart
  - package formatting/key order remains restart-free; changed installed dependencies require restart
  - failed install restores the prior on-disk tree and live resource
  - code/resource deletion cases set `restartRequired`
- full integration suite previously completed with 1,643 passing, 0 failures, 6 cancelled, and 20 skipped; the nonzero exit was the existing optional Ollama setup incompatibility under local Node 26 (`ERR_IMPORT_ATTRIBUTE_MISSING` for `json/systemSchema.json`), with those files unchanged

## Review coverage

Author: Codex. Standard `cross-model-review/bin/prepush-review.mjs` route run repeatedly in exact-commit delta mode with independent Claude, Gemini, Grok, and Harper-domain adjudication. The final exact-head review of `f7be8aff1` reports **LGTM** with no new blocker or major regression.

Review findings fixed and regression-tested include stale watcher generations, suppressed restart requests, post-install package evidence, transitive runtime modules/resolution cache invalidation, pure-ESM and native-addon loading, transactional rollback under a live writer, deploy-aware load timing, static URL ownership, and dead deploy-owner/process-group reclamation.

_PR implementation generated by Codex._

### DIFF
--- a/components/EntryHandler.ts
+++ b/components/EntryHandler.ts
@@ -1,5 +1,6 @@
 import { type Logger } from '../utility/logging/logger.ts';
 import { loggerWithTag } from '../utility/logging/harper_logger.ts';
+import { createHash } from 'node:crypto';
 import type { Stats } from 'node:fs';
 import { EventEmitter, once } from 'node:events';
 import { Component, FileAndURLPathConfig } from './Component.ts';
@@ -74,25 +75,42 @@ export type EntryHandlerEventMap = {
 	unlinkDir: [entry: UnlinkDirectoryEvent];
 };
 
+type EntrySnapshot =
+	{ entryType: 'file'; urlPath: string; digest: Buffer } | { entryType: 'directory'; urlPath: string };
+
+type RedeployScan = {
+	generation: number;
+	previousEntries: Map<string, EntrySnapshot>;
+	currentEntries: Map<string, EntrySnapshot>;
+	removalsEmitted: Set<string>;
+};
+
 export class EntryHandler extends EventEmitter<EntryHandlerEventMap> {
 	#component: Component;
 	#watcher?: FSWatcher;
 	#logger: Logger;
 	#pendingFileReads: Set<Promise<void>>;
+	#pendingFileReadsByGeneration: Map<number, Set<Promise<void>>>;
 	#isInitialScanComplete: boolean;
+	#readyGeneration = 0;
+	#watchGeneration = 0;
+	#eventSequence = 0;
+	#latestPathSequence = new Map<string, number>();
+	#entries = new Map<string, EntrySnapshot>();
+	#redeployScan?: RedeployScan;
 	// When true, #watch() short-circuits without creating a chokidar watcher.
 	// pause() sets it, resume() clears it. Lets a deploy quiesce the watcher
 	// without losing the EntryHandler instance (and therefore listener
 	// attachments registered by plugins via scope.handleEntry(handler)).
-	#paused: boolean = false;
+	#paused = false;
 	// Tracks the in-flight close() promise from pause() so resume() can await
 	// the old watcher's inotify handles releasing before installing a fresh
 	// chokidar instance — otherwise a rapid pause→resume can overlap teardown
 	// and setup, which under inotify pressure can produce spurious EMFILE.
 	#pausedClose?: Promise<void>;
-	#usingPolling: boolean = false;
-	#closed: boolean = false;
-	#openCount: number = 0;
+	#usingPolling = false;
+	#closed = false;
+	#openCount = 0;
 	ready: Promise<any[]>;
 
 	constructor(name: string, directory: string, config: FilesOption | FileAndURLPathConfig, logger?: Logger) {
@@ -101,6 +119,7 @@ export class EntryHandler extends EventEmitter<EntryHandlerEventMap> {
 		this.#component = new Component(name, directory, castConfig(config));
 		this.#logger = logger || loggerWithTag(name);
 		this.#pendingFileReads = new Set();
+		this.#pendingFileReadsByGeneration = new Map();
 		this.#isInitialScanComplete = false;
 		this.ready = once(this, 'ready');
 		this.#watch();
@@ -114,7 +133,8 @@ export class EntryHandler extends EventEmitter<EntryHandlerEventMap> {
 		return this.#component.directory;
 	}
 
-	#handleAll(...[event, path, stats]: FSWatcherEventMap['all']): void {
+	#handleAll(generation: number, ...[event, path, stats]: FSWatcherEventMap['all']): void {
+		if (generation !== this.#watchGeneration) return;
 		if (path === '') path = '/';
 
 		if (!isMatch(path, this.#component.globOptions.source, { ignore: this.#component.globOptions.ignore })) return;
@@ -124,26 +144,7 @@ export class EntryHandler extends EventEmitter<EntryHandlerEventMap> {
 		switch (event) {
 			case 'add':
 			case 'change': {
-				const urlPath = deriveURLPath(this.#component, path, 'file');
-				const fileReadPromise = readFile(absolutePath)
-					.then((contents) => {
-						const entry: AddFileEvent | ChangeFileEvent = {
-							eventType: event,
-							entryType: 'file' as const,
-							contents,
-							stats,
-							absolutePath,
-							urlPath,
-						};
-						this.emit('all', entry);
-						this.emit(event, entry as any);
-					})
-					.finally(() => {
-						this.#pendingFileReads.delete(fileReadPromise);
-						this.#checkIfAllComplete();
-					});
-
-				this.#pendingFileReads.add(fileReadPromise);
+				this.#queueFileRead(generation, event, path, absolutePath, stats, readFile(absolutePath));
 				break;
 			}
 			case 'unlink': {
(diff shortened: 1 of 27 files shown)


components/redeployRestart.ts
```diff
@@ -0,0 +1,60 @@
...
+		if (!awaitingPostRedeployScan) return;
+		awaitingPostRedeployScan = false;
+		for (const file of loadedFiles) {
+			if (!currentScanFiles.has(file)) {
+				// A file this scope loaded that the just-completed scan never reported — deleted during
+				// the redeploy. The running workers still have it, so request a restart.
+				loadedFiles.delete(file);
+				scope.requestRestart();
+			}
+		}
```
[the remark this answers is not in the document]
kylebernhardy: Deleting the *current* element during a Set `for...of` is well-defined and doesn't skip elements (only deleting a not-yet-visited element would), so this was safe — but you're right that it reads as a hazard, so snapshotted the iteration (`for (const file of [...loadedFiles])`) in f13de9b48 to make it unambiguous and robust to future edits. Thanks!

components/redeployRestart.ts
```diff
@@ -0,0 +1,74 @@
...
+		loadedFiles.add(entry.absolutePath);
+		try {
+			await load(entry);
+		} catch (error) {
+			loadedFiles.delete(entry.absolutePath);
+			throw error;
+		}
+	});
+
+	scope.on?.('deploy:start', () => {
```
kriszyp: Could we reset/arm the fresh-scan window on `deploy:end`, or otherwise tag the resumed watcher generation? `EntryHandler.pause()` closes the watcher but does not cancel already-started `readFile()` work, so a pre-deploy `add` can arrive after this reset, enter `currentScanFiles`, and have its restart suppressed by `Scope.#deployInFlight`. If the deployment removes that file and the resumed scan is empty, `ready` then treats the stale path as present and never requests a restart, leaving a deleted env value or route active. Please add a regression that models restart suppression plus an `add` completing between `deploy:start` and `deploy:end`.

unitTests/resources/redeployRestart.test.js
```diff
@@ -0,0 +1,135 @@
+const assert = require('node:assert');
+const { spy } = require('sinon');
```
kriszyp: Please replace these new Sinon spies with plain counters/captured arguments and small functions. Harper's current unit-test convention explicitly prohibits new Sinon/rewire usage, and this helper can retain the same coverage without stubbing machinery.

kriszyp (dismissed):
Requesting changes because I think this is fixing the symptoms at the consumer layer instead of restoring the `scope.handleEntry()` event contract.

`handleEntryWithRedeployRestart` is now a second file/deploy state machine for `loadEnv` and Fastify, while `jsResource` retains another copy and every other built-in or third-party `scope.handleEntry()` consumer still receives the lossy stream: every surviving file is replayed as `add`, deleted files produce no `unlink`, and unchanged files are indistinguishable from changed ones. That is the backwards-compatibility break we should fix centrally.

I think `EntryHandler.pause()/resume()` should preserve the previous watcher generation, discard or generation-tag stale pending reads, buffer the resumed initial scan, and diff it at `ready`:

- new entry → `add` / `addDir`
- changed file → `change`
- removed entry → `unlink` / `unlinkDir`
- unchanged entry → no event

That restores the existing API for every consumer, avoids marking a byte-identical redeploy as restart-required, and lets us remove both this helper and the compensating `jsResource` bookkeeping.

The inline stale-read race is also a concrete correctness blocker in the current approach. The focused helper tests cannot expose it because they do not exercise the real `Scope.#deployInFlight` restart suppression or `EntryHandler.pause()/resume()` lifecycle.

The history is split: #822 introduced pause/resume with fresh-`add` replay, #1540 introduced the atomic directory swap, and #1806 explicitly relies on existing-component watchers to detect what changed. Regardless of attribution, the released behavior is incompatible and #1936 is the point where we should correct course rather than add another compensating API.

We should keep #1806's explicit restart signal for genuinely new components, since they have no live `Scope` to emit events through; that preserves the #674 actionable 404 behavior.

This is far enough into the wrong abstraction that I would prefer replacing or substantially reworking this PR rather than expanding the helper. I am happy to take over this branch or open a replacement PR, whichever you prefer.

kylebernhardy:
Kyle's call: please take this one — it's yours.

Your read is right, and I want to be specific about what I confirmed rather than just conceding:

- **The stale-read race is real and this helper cannot fix it.** `EntryHandler.pause()` closes the watcher but never cancels `#pendingFileReads`, so a `readFile()` started pre-pause still resolves and emits `add`. That lands *after* my `deploy:start` reset, gets recorded as part of the fresh scan, and `Scope.requestRestart()` swallows it under `#deployInFlight`. If the deploy deleted that file, the `ready` diff sees it as present and never requests a restart — a deleted env value or route stays live in the workers. Only `EntryHandler` knows about the pending-read generation, so no amount of bookkeeping in a consumer closes it.
- **The altitude critique lands.** Three copies of the same compensation — `loadEnv`, `fastifyRoutes`, and `jsResource`'s inline one — is the tell that the contract, not the consumers, is what's broken. Restoring `pause()`/`resume()` to diff the resumed scan into real `add`/`change`/`unlink` fixes it once for every `handleEntry` consumer, and stops flagging a byte-identical redeploy as restart-required, which this PR never even attempted.
- **The Sinon objection is simply correct** and I should have caught it — the repo convention prohibits new Sinon and I used spies anyway.

The branch is `fix/redeploy-env-restart-1934`; take it over or open a replacement, whichever suits you. I've stopped work on it, and I've left your two threads open rather than resolving them. If the integration fixture or the loadEnv/fastifyRoutes wiring is useful as a starting point, it's there; if it's cleaner to start fresh, discard it without ceremony.

Two things worth carrying forward, since they're the actual failure modes #1934 hit: a file *deleted* during a redeploy produces no `unlink` today, and a file first seen during a post-redeploy scan reaches only the scanning thread. Both should fall out of the diff-at-`ready` design naturally, but they're the cases I'd want covered.

Separately — your #1918 review is the same shape and is a separate decision still with Kyle, so nothing there is handed over yet.

components/redeployRestart.ts
kriszyp: Addressed by replacing the consumer helper with generation-aware diffing in EntryHandler. pause() now invalidates the old generation, stale reads cannot emit or enter the resumed snapshot, and the regression suite deterministically covers an obsolete pending read across pause/resume.

unitTests/resources/redeployRestart.test.js
kriszyp: Addressed by removing the helper and its test entirely. The replacement EntryHandler tests use plain captured events and a controllable Promise; no new Sinon usage was added.

kriszyp:
I took over the handed-off branch and replaced the consumer-level workaround with the central course correction discussed above. `EntryHandler` now diffs the retained pre-deploy snapshot against the resumed watcher generation and emits logical add/change/unlink/addDir/unlinkDir events, while generation guards discard stale reads. The old helper and jsResource-specific deploy bookkeeping are gone, and #674 remains covered by the explicit new-component restart path. The branch has been rebased and force-updated at 857aa7a8c; the PR body now includes the full validation results and degraded outside-model review status.

kriszyp:
Cross-model review rerun completed with real outside-model coverage: Claude ✓ / Gemini agy ✗ → direct API ✓. The review found and we fixed suppressed non-watcher restart requests, package metadata changes outside plugin globs, cold-recreation/update identity loss, interrupted-first-scan replay, readiness under listener errors, and static URL-key collisions. Final targeted Claude verification reports no remaining blocker or significant concern. Updated head: 7d8e370f1.

kriszyp (approved):

components/EntryHandler.ts
```diff
@@ -367,6 +650,10 @@ export class EntryHandler extends EventEmitter<EntryHandlerEventMap> {
 
 	update(config: FilesOption | FileAndURLPathConfig) {
 		this.#component = new Component(this.name, this.directory, castConfig(config));
+		// Re-arm the readiness latch for parity with pause(). Without this, awaiting update() on an
+		// already-ready handler resolves against the outgoing generation's 'ready' — before the
+		// replacement generation has scanned, digested, and emitted its logical events.
+		if (!this.#closed) this.ready = once(this, 'ready');
```
cb1kenobi: **Low: `update()` re-arms the `ready` latch but does not invalidate the outgoing generation, so `ready` can resolve against the old config**

Unlike `pause()` (which bumps `#watchGeneration` synchronously), `update()` re-arms `this.ready` here while the generation bump happens later, inside `#installWatcher` after `await outgoing.close()`. If `update()` is called while the current generation's initial scan is still finishing (chokidar `ready` has fired but a slow `readFile` is still pending), that pending read can complete during the `close()` await — while `#watchGeneration` is still the outgoing value — and drive `#checkIfAllComplete(oldGen)` → `emit('ready')`, resolving the freshly re-armed latch against the *outgoing* generation. That is the exact outcome the comment just above says the re-arm prevents.

Impact is low today: no events are lost or duplicated (the replacement generation still installs, delivers its diff, and `#observedEntries()` carries state forward), and the only production caller — the `Scope` options-change handler — treats `update()` as fire-and-forget. It is a latent readiness-contract gap that would mislead any future `await entryHandler.ready` taken after an `update()`.

Suggested fix: mirror `pause()` and invalidate the outgoing generation up front, before re-arming the latch:

```suggestion
		this.#watchGeneration++;
		if (!this.#closed) this.ready = once(this, 'ready');
```

(Same double-bump pattern that `pause()` + `resume()` already rely on — `#installWatcher` bumps again for the replacement watcher, which is harmless.) Minor adjacent note: the `await outgoing.close()` on this `update()` path isn't wrapped in `.catch` the way `pause()` and the polling-recovery `#watch()` call are, so a rejecting `close()` would reject `update()`'s promise; the `Scope` caller catches it, but it's an easy robustness parity fix.

—
_Generated by Barber AI_

kylebernhardy:
### Review: deep-review (multipass) + code-review, against head `6ad2b411`

Kris — ran a multi-agent review (deep-review multipass over config/concurrency/api, plus a five-angle code review), verified each finding against the code, and ran the build and test chain locally. Moving the diff into `EntryHandler` is the right call, and I confirmed by tracing that [#1934](https://github.com/HarperFast/harper/issues/1934) genuinely closes for both `loadEnv` and `fastifyRoutes` untouched: changed `.env` → digest mismatch → `change` → restart; byte-identical redeploy correctly stays silent.

Two findings undercut the PR's goal, and **#1 currently masks #2** — fixing either alone leaves you worse off than fixing both.

#### 1. `packageMetadataChanged` is true on nearly every redeploy

The "before" read takes the live dir, which already holds the lockfile the *previous* deploy's `npm install` wrote:

https://github.com/HarperFast/harper/blob/6ad2b4112f1d24639f1f2258fd40a78bdb590b52/components/Application.ts#L533-L535

The "after" read happens inside `extractApplication`, i.e. *before* `installApplication`:

https://github.com/HarperFast/harper/blob/6ad2b4112f1d24639f1f2258fd40a78bdb590b52/components/Application.ts#L1115-L1120

`npm pack` tarballs carry no lockfile, so the compare is `{package.json, package-lock.json}` vs `{package.json}` → size mismatch → "changed" on every redeploy of any component with a `package.json`. That's the restart churn this PR set out to remove, and it makes `restartRequired` permanently-on noise. The new byte-identical test can't catch it — it calls `extractApplication` directly, so `installApplication` never runs. Moving the second read into `prepareApplication` after `installApplication` puts both on the same pipeline stage.

#### 2. Redeploys changing only non-globbed files no longer restart

https://github.com/HarperFast/harper/blob/6ad2b4112f1d24639f1f2258fd40a78bdb590b52/components/operations.js#L649-L651

Default globs are narrow (`jsResource: 'resources.js'`, `fastifyRoutes: 'routes/*.js'`), so a helper like `lib/db.js` that `resources.js` imports is watched by nothing. The deleted `loadedResourceFiles` re-`add` rule used to restart on *every* redeploy, incidentally covering this; now it's zero events and a stale module keeps serving. Same failure class as #1934, moved from `.env` to non-globbed source. A content digest of the extracted tree (minus `node_modules`) would cover both this and #1.

#### 3. Non-ENOENT read failure is reported to consumers as a deletion

https://github.com/HarperFast/harper/blob/6ad2b4112f1d24639f1f2258fd40a78bdb590b52/components/EntryHandler.ts#L223-L232

On `EACCES`/`EIO`/`EMFILE` the path never reaches `scan.currentEntries`, so `#finishRedeployScan` synthesizes an `unlink` for a file still on disk — `static.ts` drops the URL, `jsResource`/`loadEnv` restart, and the truncated snapshot is committed. A `readFailures` set excluded from `missingEntries` would close it.

#### 4. Listener throws routed through the watcher-exhaustion classifier

That same `.catch` sits after the `.then` that invokes listeners, so a sync listener throw reaches `#handleError`. If it carries `ENOSPC`/`EMFILE` it's swallowed and flips the handler to polling permanently. And `error` is emitted *before* `ready`, so `once(handler, 'ready')` rejects → `Scope`'s `initialLoadPromise` rejects → component load fails. On `main` that throw was just a rejected read and the load survived. `#finishRedeployScan` already has the right shape (collect, report after `ready`).

#### 5. `update()` doesn't invalidate the outgoing generation

`cb1kenobi`'s open comment on this head has the latch half right and is still unanswered; four agents landed on it independently. `pause()` bumps synchronously ([L613-615](https://github.com/HarperFast/harper/blob/6ad2b4112f1d24639f1f2258fd40a78bdb590b52/components/EntryHandler.ts#L613-L615)), `update()` waits until `#installWatcher` clears `await outgoing.close()`:

https://github.com/HarperFast/harper/blob/6ad2b4112f1d24639f1f2258fd40a78bdb590b52/components/EntryHandler.ts#L650-L658

Sharper consequence worth folding into the same fix: `update()` also swaps `#component` synchronously, so outgoing-generation events get matched against the **new** globs and given the **new** `urlPath`, then written into the **old** snapshot — the replacement generation compares equal, never emits the `unlink`/`add` pair, and `static.ts` keeps serving a stale URL key.

#### Smaller

- **Snapshot records emission, not consumption.** `#handleAddedEntry` writes before `#emitEntry` and `Scope.wrapHandler` swallows async rejections, so an entry whose consumer threw is still "observed" and never retried. `main`'s add-replay is what used to retry these (`roles.ts`, `dataLoader.ts`, `jsResource.ts`).
- **`.on('error')` is the only chokidar handler not generation-scoped** ([L539-543](https://github.com/HarperFast/harper/blob/6ad2b4112f1d24639f1f2258fd40a78bdb590b52/components/EntryHandler.ts#L539-L543)) — an outgoing watcher's exhaustion error permanently flips the healthy incoming one to polling.
- **`urlPath` collisions in `static.ts`.** With `['web/**', 'static/**']` both `index.html`s derive `/index.html` (last-writer-wins). Remove one while the other is byte-identical → survivor silent, removed path's `unlink` passes the identity guard, shared key deleted, 404 until restart.
- **[`roles.ts#L13-L16`](https://github.com/HarperFast/harper/blob/6ad2b4112f1d24639f1f2258fd40a78bdb590b52/resources/roles.ts#L13-L16) has no `entryType` guard** — reads `.contents` on `addDir`/`unlinkDir` → `TypeError`. `unlinkDir` post-deploy is newly reachable here.
- **`close()` strands a re-armed `ready` latch** — `removeAllListeners()` means a latch armed by `pause()`/`update()` hangs rather than rejects. `pause()` also lacks the `#closed` guard `update()` has.
- **Type/URL-change branch loses the paired `add`** if `#emitRemoval` throws — snapshot already committed, so the next generation won't re-emit either.
- **`readPackageMetadata` rethrows non-`ENOENT`** on the first line of `extractApplication`, failing the deploy before the `rename`-aside logic written to tolerate a weird `dirPath`. Probably wants to be best-effort → "changed".
- **Stale docs**: `integrationTests/deploy/redeploy-restart-flag*.ts` header comments still describe "fresh scan re-emits every file as `add`, deletions produce no event" — now the opposite. Tests pass; comments mislead.

#### Test gaps

- Nothing rejects the contents promise, so #3 and #4 are invisible to the suite.
- `preserves the pre-deploy snapshot when update races the resumed scan` would pass with `#watchInstall` removed entirely.
- `commits the resumed generation even when a removal listener throws` covers only `#finishRedeployScan`'s try/catch — it reads broader than it is (see #4).
- Only guarantee lost with the deleted `jsResource` tests: "a re-`add` must not re-import" — `scope.import()` hits the per-scope module cache.

#### Merge order (not a defect here)

`c907eb2a7` on [#1849](https://github.com/HarperFast/harper/pull/1849) refactors the restart gate into `markRestartRequiredForNewComponent()` with four call sites. This PR edits only the pre-refactor one-liner, so if both land `packageMetadataChanged` reaches one of four — the default replicated-cluster path would skip it.

#### Verified locally

At `6ad2b411`: `build`, `typecheck`, `lint:required`, `format:check` all clean. Focused suites 75 passing (EntryHandler/Scope/extractApplicationSwap/static) + 6 (jsResource).

Caveat on the PR description's full-suite numbers: `test:unit:main` isn't hermetic on a dev box — it opens `~/harper`, the real local install. I couldn't get a clean local run, so I can't confirm or refute the 3,809. The one failure I dug into (`resolvePreload`, `/private/var` vs `/var`) reproduces identically on `main`.

The interrupted-scan carry-forward in `#observedEntries` and the descendants-before-parents removal ordering both held up under everything I threw at them.

_Comment generated by kAIle (Claude Fable 5)_

kriszyp:
Thanks — this was the right review, and the current head addresses the concrete failures you identified.

- **Post-install package evidence:** the old installed metadata is captured before extraction and the replacement metadata is captured in `prepareApplication()` only after `installApplication()`. `package.json` is normalized as parsed JSON; installed lockfiles remain exact evidence. Custom commands, install scripts, bundled `node_modules`, and dependency installs without a lock are treated as opaque and require restart.
- **Non-globbed loaded code:** rather than hashing the whole extracted tree (which would restart for unused source and generated caches), `ApplicationScope`/`jsLoader` now record every transitively loaded application-local module plus relative-resolution candidate state. A changed/missing imported helper or a new shadowing candidate such as `foo.js` ahead of `foo.json` requests restart; unused files stay restart-free. Native npm dependencies delegated by the default loader are covered by the post-install package/lock evidence.
- **Read failures:** redeploy scans retain failed paths from the previous snapshot, exclude them from synthesized removals, resolve `ready`, and then report the read error. There is deterministic EACCES coverage.
- **Listener errors:** file-read rejection and consumer exceptions are separate promise branches. An exhaustion-shaped consumer throw is reported as a consumer error and cannot flip the watcher to polling; deferred scan errors are surfaced after readiness commits.
- **`update()` race/readiness:** `update()` now invalidates the outgoing generation synchronously before replacing `#component` or re-arming readiness. A deferred old-generation read racing a URL-path update is covered and produces the required old-URL `unlink` plus new-URL `add`.

The smaller items are also covered: watcher errors are generation-scoped; static URL collisions retain ownership and restore a surviving path; roles ignores directory/unlink events; close rejects a pending readiness latch; type/URL transitions still emit their paired addition if a removal listener throws; and unreadable package evidence fails closed to “changed.” I corrected the stale deletion-test rationale in b5b48e1ec.

One point is intentionally not converted into retry behavior: EntryHandler snapshots filesystem observation, not successful completion of arbitrary async consumer work. Scope reports async handler rejection and awaits it during initial load, but a later byte-identical deploy does not replay entries as an implicit retry mechanism. If we want durable consumer-operation retries, that needs its own explicit contract rather than depending on cold watcher replay.

The #1849 merge-order note remains valid; that PR is still open and conflicted, so whichever lands second will need to reconcile its restart-gate refactor.

Current head is b5b48e1ec. Focused regressions cover stale reads, deletions, read failure, listener failure, update races, close readiness, static collisions, roles events, transitive module provenance, resolution shadowing, and post-install metadata equivalence.

cb1kenobi (approved):

Issue #1609, opened 2026-07-06
MCP application profile: no hook to expose custom content resources (only Resource descriptors + harper:// metadata)


kylebernhardy:
## Summary
The MCP **application** profile has no component-author hook to expose **custom content resources** — arbitrary `resources/read`-able `text`/`blob` content under an author-chosen URI. This blocks a real use case: a public docs server exposing each documentation page as an MCP resource (`docs:///{+path}` → the page Markdown).

The 2025-06-18 resources spec explicitly supports this: a resource may be "files, database schemas, or **application-specific information**"; `resources/read` returns `text` or `blob`; and "implementations are always free to use additional, custom URI schemes."

## Current behaviour (`harper@5.1.15`)
Resources are "discovered, not registered" (`components/mcp/resources.ts`). On a live `application`-profile instance, `resources/list` returns **only**:
- `harper://about`, `harper://openapi`, `harper://schema/{db}/{table}` — server metadata; and
- `http(s)://<host>/<path>` — a **descriptor** for each exported Resource (returns `{ uri, path, database, table, hint }`, *not* content; the hint just points back to a `get_*`/`search_*` tool).

There is no `resources` equivalent of `static mcpTools`. Modelling hundreds of content pages as Resource classes is a poor fit, and the descriptor never carries the body.

## Ask
A component-author hook to register custom resource URIs / templates with an author-supplied `read` returning `text`/`blob` content — parallel to how `mcpTools` opens the tool surface (e.g. a static `mcpResources` / `mcpResourceTemplates`, or a registration API).

## Minor, related
Exported-Resource descriptors use the `http(s)://` scheme, but the spec says use `https://` "only when the client is able to fetch and load the resource directly from the web on its own." These are in-process, RBAC-gated descriptors, so a custom scheme would fit the spec guidance better.

## Context
Found while evaluating whether a public docs MCP server (tools `search_docs`, `fetch_doc`, LLM-backed `answer`) could move from a hand-rolled JSON-RPC server onto the native `application` profile. The tool surface maps cleanly via `mcpTools` (incl. anonymous access); custom content resources are the one capability with no native path.

PR #1943, opened 2026-07-24
fix(mcp): list verb tools for table-less Resources to any authenticated user

### LINKED ISSUE #1940: [MCP] Verb tools for table-less programmatic Resources are listed only to super-users
Follow-up from the docs pass for #1923 ([documentation#605](https://github.com/HarperFast/documentation/pull/605)), which surfaced this while verifying what #1921 actually delivers.

## Problem

`makeVisibleTo` gates `tools/list` visibility on the Resource having a backing database and table:

```ts
// components/mcp/tools/application.ts:716
if (!databaseName || !tableName) return false;
```

A purely programmatic Resource — a `Resource` subclass that overrides `get`/`post`/`put`/`delete`, or aggregates across tables — has neither, so **every non-super user sees none of its verb tools in `tools/list`**. The tools are still invocable by name via `tools/call`; they are just undiscoverable.

The surrounding comment frames this as a conservative default ("non-table Resources have no static permission gate — runtime `allow*` predicates enforce"), which is a reasonable starting posture. But it now collides with #1920/#1921: we just did the work to give exactly these Resources rich `inputSchema` / `outputSchema` output, and for any non-super user that work is invisible.

## Why it matters

The realistic MCP deployment is an LLM host connecting as an application user, not as `super_user`. In that configuration a component author who follows the `static properties` authoring guide gets a correct, well-described tool that their agent can never discover.

## Options

1. **List it when the Resource opts in** — e.g. honor `static mcp = { listPublicly: true }` or reuse the existing `exportTypes.mcp` signal, so the author takes explicit responsibility for the runtime `allow*` predicates.
2. **List it whenever the class declares `allowRead`/`allowUpdate`/etc.** — presence of a predicate is the author asserting they enforce access.
3. **List it to any authenticated user by default** and rely on the runtime predicates, matching how the custom `mcpTools` surface already behaves (those have no per-user listing filter beyond authentication).
4. **Keep as-is and document it** — the docs PR currently does this, so this issue is not blocking.

Option 3 is the most consistent with `mcpTools`, which already lists to any authenticated caller and delegates enforcement to the method. Worth a deliberate decision either way rather than leaving the asymmetry implicit.

Whatever we choose, the note in `reference/resources/resource-api.md` and the troubleshooting line in `learn/developers/mcp-and-openapi-metadata.mdx` need a follow-up edit.

kylebernhardy:
Closes #1940.

## Problem

`makeVisibleTo` gated `tools/list` visibility on the Resource having a backing database and table:

```ts
if (!databaseName || !tableName) return false;
```

A purely programmatic Resource — a `Resource` subclass overriding the verbs, or one aggregating across tables — has neither, so **no non-super user saw any of its verb tools**. That is precisely the class of Resource #1920/#1921 just taught to emit rich `inputSchema` / `outputSchema`, so the feature was invisible to the realistic MCP caller.

The gate also bought no access control. `tools/call` does not consult `visibleTo` — `toolRegistry.ts`'s own docblock says so ("explicitly NOT a security boundary") and `transport.ts` dispatches straight from `getTool(name, profile)` to the handler, with an existing pass-through test for a hallucinated name. So hiding these tools cost discoverability and nothing else.

## Change

Table-less Resources are listed to any **authenticated** caller; their `allow*` predicates enforce at call time. Table-backed Resources are untouched — still `getUserTablePermissions` per mode.

## Two corrections to my first commit (review found both)

**The first commit returned `true` unconditionally, which included anonymous callers.** `tools/list` has no auth gate: the MCP handler mounts `{ after: 'authentication' }`, which is ordering, not a requirement, so an unauthenticated request arrives as `{ username: '' }` and optional-chains past the super-user check. Every exported programmatic Resource's field names, docstrings, and `enum` values would have been published to anyone who could reach the application port. Fixed in `8a8eb36ff` with an explicit `isAuthenticated` helper, plus a regression test covering `{ username: '' }`, `undefined`, and `{}`.

**The "parity with custom `mcpTools`" argument was too broad** and I've dropped it. Custom tools do list to anonymous sessions — deliberately, for the public-docs case (#1609) — but declaring `static mcpTools` is an author opt-in, a conscious act of publishing. Verb tools are generated for every exported Resource with no author action, so the defaults are not equivalent and shouldn't be aligned by pointing at each other.

## Where to look

- **The authenticated-vs-anonymous line is the security judgment.** Authenticated users with zero grants now see descriptors (names, descriptions, property descriptions, `enum` values) for table-less Resources they previously couldn't enumerate. Data access is unchanged — that was always the `allow*` predicates' job — but descriptor *text* is a real disclosure surface: a `withSchema` contract copies author-declared property schemas verbatim, and an `enum` can encode business data (tenant codes, tier names). If you'd rather this be opt-in rather than opt-out, say so and I'll switch it to a `static mcp` flag.
- Three tests: table-less Resource listed to a no-grants user across four verbs; the anonymous case; and a guard that the table-backed path still denies.

Full MCP unit suite (477 tests) passes.

## Related, not fixed here

Review also surfaced that `detectVerbs` manufactures a `create_*` tool for **every** `Resource` subclass — it checks `typeof p.post === 'function'` with no base-prototype comparison, unlike `hasClassLevelVerbs` one file over — which reaches Harper's own `login` Resource. Filed as #1945; independent of this change (that's about which tools exist, this is about who sees them).

Docs follow-up in [documentation#605](https://github.com/HarperFast/documentation/pull/605), which documents the old super-user-only behavior.

_PR description generated by kAIle (Claude Opus 4.8)._

### DIFF
--- a/components/mcp/toolRegistry.ts
+++ b/components/mcp/toolRegistry.ts
@@ -363,6 +363,16 @@ export function isSuperUser(user: AuthedUser | undefined): boolean {
 }
 
 /**
+ * True if the request carries an authenticated principal. Anonymous MCP sessions are a supported
+ * deployment (the public-docs case, #1609) — the adapter hands them through as `{ username: '' }`
+ * rather than rejecting them — so a listing filter that means "any real user" has to test for this
+ * explicitly. `user?.role` is not a proxy: an authenticated user may legitimately carry no role.
+ */
+export function isAuthenticated(user: AuthedUser | undefined): boolean {
+	return typeof user?.username === 'string' && user.username.length > 0;
+}
+
+/**
  * Class-level verb introspection. Mirrors `resources/openApi.ts:149-153` so
  * a class is considered to "have" a verb only if its prototype overrides the
  * base Resource implementation. Note that `post`'s default in Harper's
--- a/components/mcp/tools/application.ts
+++ b/components/mcp/tools/application.ts
@@ -27,6 +27,7 @@ import harperLogger from '../../../utility/logging/harper_logger.ts';
 import {
 	addTool,
 	clearProfileTools,
+	isAuthenticated,
 	snapshotProfileTools,
 	type AuthedUser,
 	type ToolCallContext,
@@ -710,10 +711,15 @@ function makeVisibleTo(
 		if (user?.role?.permission?.super_user === true) return true;
 		// A Resource with no backing table has no static permission gate to consult, so listing is
 		// open to any authenticated caller and the Resource's own `allow*` predicates enforce at call
-		// time. This matches custom `mcpTools`, which have never had a listing filter beyond
-		// authentication — and hiding these while still accepting `tools/call` on them bought no
-		// security, only undiscoverable tools (#1940).
-		if (!databaseName || !tableName) return true;
+		// time — hiding these while still accepting `tools/call` on them bought no security, only
+		// undiscoverable tools (#1940).
+		//
+		// Anonymous sessions stay excluded. They are a supported deployment (#1609) and reach here as
+		// `{ username: '' }`, so this must be tested explicitly. Custom `mcpTools` DO list to them, but
+		// that is an author opt-in — declaring `static mcpTools` is a deliberate act of publishing.
+		// Verb tools are generated for every exported Resource with no author action, so publishing an
+		// app's field names and docstrings to unauthenticated callers by default is not equivalent.
+		if (!databaseName || !tableName) return isAuthenticated(user);
 		const perm = getUserTablePermissions(user, databaseName, tableName);
 		if (!perm) return false;
 		if (mode === 'read') return perm.read === true || perm.describe === true;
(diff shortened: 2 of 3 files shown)


kylebernhardy:
Runtime note from verifying the sibling PR (#1944) against a booted instance — relevant to this PR's rationale.

**Confirmed:** on the pre-#1943 build, an anonymous MCP session gets `tools/list` → **0 tools** for table-less Resources, while an authenticated super-user gets all of them. So the old `return false` genuinely was gating anonymous listing, and the first commit here would have opened it — the `isAuthenticated` fix in `8a8eb36ff` is load-bearing, not defensive.

**Correction to this PR's description:** I justified the widening partly by saying OpenAPI "is unfiltered entirely." That overstates it — a default `prod` install returns **403** for an unauthenticated `GET /openapi`. The document is global (no *per-user* filtering, so any authorized caller sees every docstring), but it is not anonymously public by default. The descriptor-disclosure tradeoff is therefore narrower than I framed it; the decision for a reviewer is still whether authenticated-with-no-grants users should see these descriptors.

Also worth knowing when reading this PR: my verification fixture defined its verbs as `static` methods and implements no prototype `post`, yet still got `create_*` tools registered — a live reproduction of [#1945](https://github.com/HarperFast/harper/issues/1945).

Comment generated by kAIle (Claude Opus 4.8).

kriszyp (approved):

PR #1944, opened 2026-07-24
fix(mcp/openapi): share one schema emitter so the two surfaces stop diverging

### LINKED ISSUE #1941: [MCP/OpenAPI] hidden is not honored on nested sub-properties, and leaks as a schema key in OpenAPI

### LINKED ISSUE #1942: [MCP/OpenAPI] The two schema type mappers disagree: unrecognized types, nested enum/format/const, and nullability

kylebernhardy:
Closes #1941, closes #1942.

> **Stacked on #1921.** The base branch is `feat/mcp-static-properties-tests-1920`, not `main` — the code paths these issues describe (OpenAPI's nested-object branch, the `enum`/`format`/`const` passthrough, `JSON_SCHEMA_SCALAR_TYPES`) only exist there. GitHub will retarget this to `main` automatically when #1921 merges. **Do not merge before #1921.**

## Root cause

`attributeToFragment` ended up serving two masters: the canonical, front-end-neutral `Table.properties` Record, and — since #1921 — OpenAPI's nested-object emit. Those have opposite requirements. The canonical projection must carry `hidden` (it round-trips through `fragmentToAttribute`) and must *not* carry `enum`/`format`/`const` (a code-first `types.enum` column has to stay identical to its GraphQL equivalent). An emitted document is the exact reverse. So the nested paths quietly drifted away from MCP's, and every divergence in #1942 lives at that seam.

Rather than bolt more conditionals onto the shared projector, this adds a second, purpose-built one:

```ts
attributeToSchema(attr, { dialect, mapPrimitive })
```

One traversal — nesting, arrays, `hidden` suppression, hint propagation, nullability — used by both surfaces. Each surface keeps its own primitive mapping, because those differences are legitimate: MCP widens `Date` to `['string','number']` and tags `Bytes` with `contentEncoding`; OpenAPI carries the Harper type name as `format`. Unifying those too would churn table-backed output, which #1921 was careful to leave alone. `attributeToFragment` is untouched.

## What changed behaviorally

**#1941** — `hidden` is honored at every nesting level, not just the top. A hidden sub-property of a nested object was previously emitted on both surfaces *with its description*, and OpenAPI additionally wrote `hidden: true` into the public document as a schema key — advertising that the field was meant to be suppressed. Harper directives (`hidden`, `primaryKey`, `assignCreatedTime`, `assignUpdatedTime`) are no longer emitted into schemas at all. A suppressed name is also dropped from the nested `required` list, which would otherwise make the object unsatisfiable for a validating client.

**#1942** — all three:

- An unrecognized type name (`'Text'`) coerced to `string` on MCP and to untyped `{}` on OpenAPI. Both now emit untyped, and `resolveDeclaredType` warns once per name so the typo is visible rather than silently producing two different wrong schemas.
- `enum`/`format`/`const`/`description` reached only top-level properties on OpenAPI. Nested objects and array items now carry them, matching MCP.
- OpenAPI read `nullable` only to compute `required` and never emitted it — so a nullable top-level property was advertised as non-nullable, while a *nested* one (routed through the shared projector) was not. Both levels now emit `nullable: true`; MCP continues to emit the `['T','null']` union and never the keyword.

## Where to look

- **`applyNullability` is the one place the dialects legitimately differ**, and it's the piece most worth checking. OpenAPI 3.0.3 has no union types, so `nullable: true` is the only expression available; JSON Schema has no `nullable` keyword. If you'd rather MCP *also* emit `nullable` for client compatibility, that's a one-line change — but it would be non-standard.
- **Two of my own tests failed on first run** and caught real gaps: OpenAPI's primitive array-item branch still bypassed the shared emitter (dropping item `enum`/`description`), and I'd added `resolveDeclaredType` without wiring it into MCP's `default:` case. Both fixed; the tests are the reason I know.
- **The emitted-output change is user-visible.** Anyone consuming `/openapi.json` sees `nullable: true` appear on top-level properties, `hidden: true` disappear from nested ones, and untyped `{}` where an unrecognized type name previously produced `{"type":"string"}` on the MCP side. That last one is the only case where a client could see *less* type information than before — deliberate, since the previous value was a guess.

Seven convergence tests assert MCP and OpenAPI output side by side for the same fragment, which is the regression shape these issues kept slipping through. 534 tests pass across the MCP, OpenAPI, GraphQL-metadata, and code-first-parity suites.

Docs follow-up: [documentation#605](https://github.com/HarperFast/documentation/pull/605) documents the current (pre-fix) per-surface caveats and will need them collapsed once this lands.

_PR description generated by kAIle (Claude Opus 4.8)._

### DIFF
--- a/resources/jsonSchemaTypes.ts
+++ b/resources/jsonSchemaTypes.ts
@@ -30,6 +30,8 @@ export interface JsonSchemaFragment {
 	additionalProperties?: boolean;
 	format?: string;
 	const?: unknown;
+	/** Binary encoding of a string-typed value. Emitted on the MCP surface only — not a 3.0.3 keyword. */
+	contentEncoding?: string;
 }
 
 /**
@@ -302,7 +304,14 @@ export function attributeToSchema(attr: AttributeLike, options: SchemaEmitOption
 			if (items) fragment.items = items;
 		}
 	} else {
-		Object.assign(fragment, options.mapPrimitive(attr.type, attr));
+		// Copy the mapper's result field by field rather than merging it wholesale: the set of keys a
+		// surface may contribute to a leaf schema is fixed, and spreading an arbitrary object here would
+		// let a mapper put anything into an emitted document.
+		const primitive = options.mapPrimitive(attr.type, attr);
+		if (primitive.type !== undefined) fragment.type = primitive.type;
+		if (primitive.description !== undefined) fragment.description = primitive.description;
+		if (primitive.format !== undefined) fragment.format = primitive.format;
+		if (primitive.contentEncoding !== undefined) fragment.contentEncoding = primitive.contentEncoding;
 	}
 
 	if (attr.nullable) applyNullability(fragment, options.dialect);
--- a/resources/jsonSchemaTypes.ts
+++ b/resources/jsonSchemaTypes.ts
@@ -306,18 +306,31 @@ export function attributeToSchema(attr: AttributeLike, options: SchemaEmitOption
 	}
 
 	if (attr.nullable) applyNullability(fragment, options.dialect);
-	if (attr.description && fragment.description === undefined) fragment.description = attr.description;
+	// An explicitly declared `description`/`format` outranks whatever the primitive mapper supplied as a
+	// default — a author writing `{ type: 'Date', format: 'date-time' }` means `date-time`, not the
+	// Harper type name the mapper stamps on.
+	if (attr.description) fragment.description = attr.description;
 	if (attr.enum && fragment.enum === undefined) fragment.enum = attr.enum;
-	if (attr.format && fragment.format === undefined) fragment.format = attr.format;
+	if (attr.format) fragment.format = attr.format;
 	if (attr.const !== undefined) {
 		// `const` is JSON Schema draft-06. OpenAPI 3.0.3's Schema Object is the draft-04 subset, so it
 		// has no such keyword — emit the equivalent single-value `enum` for that dialect. MCP speaks
 		// current JSON Schema and takes `const` directly.
 		if (options.dialect === 'openapi-3.0.3') {
-			if (fragment.enum === undefined) fragment.enum = [attr.const as never];
+			// Intersect rather than skip when both are declared: dropping the `const` would leave OpenAPI
+			// advertising a wider value set than MCP, from one declaration.
+			fragment.enum = fragment.enum
+				? (fragment.enum.filter((value) => value === attr.const) as never[])
+				: [attr.const as never];
 		} else if (fragment.const === undefined) {
 			fragment.const = attr.const;
 		}
+	}
+	// OpenAPI 3.0's `nullable` does not widen an `enum` — a validator still rejects `null` unless the
+	// value list contains it. Without this, marking an enum (or a const, now emitted as one) nullable
+	// produces a schema that contradicts itself.
+	if (options.dialect === 'openapi-3.0.3' && fragment.nullable && fragment.enum && !fragment.enum.includes(null)) {
+		fragment.enum = [...fragment.enum, null];
 	}
 	// `hidden` / `primaryKey` / `assignCreatedTime` / `assignUpdatedTime` are Harper directives, not
 	// schema vocabulary — they never belong in an emitted document.
--- a/resources/jsonSchemaTypes.ts
+++ b/resources/jsonSchemaTypes.ts
@@ -28,6 +28,8 @@ export interface JsonSchemaFragment {
 	additionalProperties?: boolean;
 	format?: string;
 	const?: unknown;
+	/** Emitted by the OpenAPI 3.0 projection for a genuine multi-type union; not authored directly. */
+	oneOf?: JsonSchemaFragment[];
 }
 
 /**
@@ -71,6 +73,12 @@ export interface AttributeLike {
 	assignCreatedTime?: boolean;
 	assignUpdatedTime?: boolean;
 	nullable?: boolean;
+	/**
+	 * The source JSON-Schema type union, verbatim, when `static properties` declared one. `type` holds
+	 * the first non-null member so single-type consumers keep working; surfaces that can express a
+	 * union (MCP passes it through, OpenAPI 3.0 translates it to `oneOf`) read this instead.
+	 */
+	types?: readonly string[];
 	elements?: AttributeLike;
 	/** Sub-attributes of a nested object field (the same array form `Table.validate` iterates). */
 	properties?: AttributeLike[];
@@ -107,6 +115,10 @@ export function attributeToFragment(attr: AttributeLike): JsonSchemaFragment {
 	} else if (attr.type === 'array' && attr.elements) {
 		fragment.type = 'array';
 		fragment.items = attributeToFragment(attr.elements);
+	} else if (attr.types) {
+		// A declared union round-trips verbatim; collapsing it to `attr.type` here would make the
+		// canonical `Table.properties` disagree with what the author wrote.
+		fragment.type = [...attr.types] as JsonSchemaType[];
 	} else {
 		const jsonType = attr.type ? DATA_TYPES[attr.type] : undefined;
 		if (jsonType) fragment.type = jsonType;
@@ -157,9 +169,11 @@ function fragmentToAttribute(name: string, fragment: JsonSchemaFragment): Attrib
 		// than misleadingly reusing the array field's own name.
 		attr.elements = fragmentToAttribute('', fragment.items);
 	} else if (Array.isArray(fragment.type)) {
-		// JSON-Schema union type. Fold a `'null'` member into `nullable` (the OpenAPI-expressible form)
-		// and keep the remaining member. A single non-null member is the common `['T','null']` case; a
-		// genuine multi-type union isn't expressible on an attribute, so the first member is kept.
+		// JSON-Schema union type. Keep the source union on `types` so surfaces that can express one
+		// (MCP natively, OpenAPI 3.0 via `oneOf`) don't have to reconstruct it, and fold a `'null'`
+		// member into `nullable` as well since that is the form OpenAPI needs. `type` carries the first
+		// non-null member for the single-type consumers (validation, query coercion) that read it.
+		attr.types = fragment.type;
 		const members = fragment.type.filter((t) => t !== 'null');
 		if (members.length !== fragment.type.length) attr.nullable = true;
 		if (members.length > 0) attr.type = members[0];
--- a/resources/jsonSchemaTypes.ts
+++ b/resources/jsonSchemaTypes.ts
@@ -30,6 +30,21 @@ export interface JsonSchemaFragment {
 	const?: unknown;
 }
 
+/**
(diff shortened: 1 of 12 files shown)


resources/openApi.ts
```diff
@@ -707,12 +672,31 @@ function Parameter(name, i, type) {
 function fragmentToOpenApiSchema(fragment: any): any {
 	if (!fragment || typeof fragment !== 'object') return { type: 'object' };
 	const schema: any = {};
-	if (fragment.type != null) schema.type = fragment.type;
+	// Request contracts are author-written JSON Schema, so the same 3.0.3 dialect rules apply here as
+	// on the table path: no `'null'` type, no type unions, no `const`. This function feeds every
+	// request body, response, and contract query parameter, so a fragment reaching it unfiltered is a
+	// non-conformant document even though the table-derived paths are clean.
+	if (Array.isArray(fragment.type)) {
```
kriszyp, line 679: This remains a second schema emitter, and it narrows every genuine union to `members[0]`. For example, a request-contract property declared as `schemaOf({ type: ['string', 'number'] })` is accepted as either member by runtime validation and exposed as either member to MCP, but OpenAPI advertises only `string`; generated clients can therefore reject valid numeric requests/responses. Please route request-contract fragments through the shared dialect-aware emitter (or extract one fragment-level emitter used by both paths), and add body/response/query coverage for a non-null multi-type union.

resources/jsonSchemaTypes.ts
```diff
@@ -200,6 +204,202 @@ export function projectPropertiesToAttributes(properties: Record<string, JsonSch
...
+	return fragment;
+}
+
+function applyNullability(fragment: JsonSchemaFragment, dialect: SchemaDialect): void {
+	// Nullability qualifies a schema that says something; neither dialect has anything to add to a
+	// fragment that resolved neither a `type` nor a `oneOf`.
+	if ((!('type' in fragment) || fragment.type === undefined) && fragment.oneOf === undefined) return;
+	if (fragment.type === undefined) {
+		// A `oneOf` union. 3.0 takes `nullable` alongside it; JSON Schema takes a `null` branch.
+		if (dialect === 'openapi-3.0.3') fragment.nullable = true;
```
kriszyp, line 390: OpenAPI 3.0.3 says `nullable` only takes effect when `type` is explicitly defined in the same Schema Object (https://spec.openapis.org/oas/v3.0.3.html#fixed-fields-20). A declaration such as `type: ['string', 'integer', 'null']` therefore becomes two non-null `oneOf` branches plus an inert `nullable: true`, so OpenAPI clients still reject `null` while MCP accepts it. Please encode null as an explicit union branch (for example `{ enum: [null] }`) and apply the same fix to the duplicated top-level OpenAPI path, with a validation test that actually checks a null instance.

resources/openApi.ts
```diff
@@ -237,65 +197,67 @@ export function generateJsonApi(resources: Resources, serverHttpURL: string) {
...
+						// shared emitter (sub-properties recursed, @hidden suppressed at every level, hints
+						// carried); OpenAPI's table path uses $refs instead.
+						props[name] = attributeToOpenApiSchema(attr) ?? {};
 					} else if (union) {
 						// A genuine multi-type union (`['string','number']`). 3.0 has no type arrays, so the
 						// equivalent is `oneOf`; `attr.type` alone would drop every member but the first.
-						props[name] = { oneOf: union.map(openApiUnionMember) };
+						props[name] = { oneOf: union.map((member) => openApiPrimitive(member, name)) };
 					} else if (type === 'array') {
 						if (!elements) {
```
kriszyp, line 209: Although `{ type: 'array' }` is valid general JSON Schema, OpenAPI 3.0.3 requires `items` whenever `type` is `array` (https://spec.openapis.org/oas/v3.0.3.html#schema-object). This branch emits a bare array, and the shared nested emitter/request-contract translator do the same, so a supported static declaration can make the entire OpenAPI document fail strict validation. For the OpenAPI dialect, please emit `items: {}` for an unconstrained array (MCP can keep the bare form) and cover top-level, nested, and request-contract arrays.

resources/jsonSchemaTypes.ts
```diff
@@ -28,6 +30,8 @@ export interface JsonSchemaFragment {
 	additionalProperties?: boolean;
 	format?: string;
 	const?: unknown;
+	/** Binary encoding of a string-typed value. Emitted on the MCP surface only — not a 3.0.3 keyword. */
+	contentEncoding?: string;
 	/** Emitted by the OpenAPI 3.0 projection for a genuine multi-type union; not authored directly. */
 	oneOf?: JsonSchemaFragment[];
```
kriszyp, line 36: `JsonSchemaFragment` is the author-facing IR behind `SchemaSource` and `Resource.properties`/`inputSchemas`/`outputSchemas`, so adding `oneOf` here makes `schemaOf({ oneOf: [...] })` a supported TypeScript declaration. Runtime `validateValue`, however, treats that fragment as untyped because it has no `type`, and the request-contract OpenAPI translator drops `oneOf`; an apparently constrained body is therefore accepted without the advertised validation. Please either keep dialect-only output fields in a separate emitted-schema type, or implement authored `oneOf` end-to-end before exposing it in this public shape.

kriszyp (changes requested):
A few more things to cover.

kylebernhardy:
Superseded by #1921, which now includes the shared MCP/OpenAPI schema emitter, closes #1941 and #1942, and incorporates the substantive review fixes plus additional schema-traversal hardening. Keeping one PR avoids maintaining two divergent branches and CI paths.

Issue #29, opened 2025-11-05
jsResource.ts does not return a promise for the completion of loading modules


kriszyp:
### Bug Summary

jsResource loads modules, which takes place asynchronously, but its `handleApplication` does not return a promise, which leads to race conditions with other plugins

### Steps to Reproduce

config.yaml:
```
jsResource:
  files: resources.js # load first
loadEnv:
  files: .env # load *after* resources.js is finished
```
resources.js:
```
import { setTimeout as delay } from 'node:timers/promises';
await delay(5000);
console.log(process.env.SOMEVAR); // should be undefined
```
.env
```
SOMEVAR=should-be-loaded-later
```

### Expected Behavior

In these steps, the loadEnv plugin should not execute until jsResource has completed loading the resources.js.

### Actual Behavior

jsResource's handleApplication immediately returns, allowing the next plugin to execute before resources.js is finished loading.

### Additional Context

https://github.com/HarperFast/harper/blob/main/resources/jsResource.ts#L24 and https://github.com/HarperFast/harper/blob/main/resources/jsResource.ts#L37 do not return any promise, they both immediately return undefined. These should both return a promise for their completion.
The gap between these two is actually the real challenge here. The `handleEntry` is called as an event handler, and any return value from the callback is ignored in Scope.ts. Scope.ts needs to actually accumulate any promises returned from the entry handlers during the "pre-ready" phase of the component/scope loading. Any promises that are returned before the scope is "ready" (the `ready` promise is resolved), which further delay the `ready` resolution (could be `this.ready = this.ready.then(() => promiseReturnedFromEntryHandler`). Doing this should give us a deterministic loading sequence without this race condition.

### Are you planning to fix this issue?

No, just reporting the issue

### First-time contributor support

- [ ] I'm new to contributing and would appreciate guidance on the process
- [ ] I'd like help understanding the project structure
- [ ] I need assistance with setting up the development environment
- [ ] I'm comfortable contributing, but new to this project specifically

PR #1947, opened 2026-07-26
server: add stall-timeout guard to enableProxyProtocol's header buffering

kriszyp:
## What this does

Superseded by #1858 (merged), which lands a more complete PROXY v2 decoder — client source address **and** the forwarded mTLS client certificate chain, wired into the existing HTTP/MQTT mTLS auth paths. That parser now lives at `server/serverHelpers/proxyProtocol.ts`. This PR's own decoder (`server/proxyProtocol.ts`, SNI/ALPN/TLS-version/JA3/JA4 extraction with no consumer) has been dropped in favor of it, rather than shipping two independent v2 decoders on the same wire format.

What's left is one piece of independent, already-reviewed value that #1858 doesn't have: a **stall-timeout guard** on `enableProxyProtocol`'s pre-handoff header buffering.

`enableProxyProtocol` intercepts a UDS mirror's first `data` events to strip a leading PROXY header before forwarding to the HTTP parser. A peer that sends the 12-byte v2 signature, declares a large body (up to 64KB), then stalls can hold that buffer indefinitely — there was no timeout on this path, unlike its sibling `createH2CProxyFront` (`prehandoffTimeout`, default 10s) and `withProxyProtocol` (used by the raw-socket UDS path). #1858's rewrite of `enableProxyProtocol` carries the same buffering pattern forward without the guard.

This PR adds it: `socket.setTimeout(prehandoffTimeout, destroy)` armed for the duration of the pre-handoff wait, cleared once the header resolves (or once there's nothing to intercept).

## Where to look

- **`server/http.ts` → `enableProxyProtocol`** — the guard itself, ~10 lines.
- **`unitTests/server/udsMirror.test.js`** — the `enableProxyProtocol` describe block's mock sockets were plain `EventEmitter`s with no `setTimeout`/`destroy`; added stubs (`createSocket()` helper) and two new tests for the stall/clear paths.
- **`server/DESIGN.md`** — one-line addition to the `enableProxyProtocol` row noting the guard.

## Testing

`tsc` clean, `npm run build` clean, `npx mocha "unitTests/server/**/*.js"` (excluding the pre-existing-broken `threads/**` glob, unrelated to this change) — 634 passing, 0 failing. `prettier --check` and `oxlint` clean on all changed files.

## History

Originally titled "Read PROXY protocol v2 on UDS mirrors and consume symphony's forwarded TLVs" — see prior revisions for that scope. Codex review on the original PR flagged that it collided with #1858 (same wire format, two independent decoders, diverging trust-boundary designs — #1858 makes mTLS load-bearing for auth, this one didn't). Kris merged #1858 and asked for this PR to be rebased down to whatever independent value remained; the stall-timeout guard above is that value.

Co-authored with Claude Sonnet 5.


server/proxyProtocol.ts
```diff
@@ -0,0 +1,279 @@
+/**
```
Devin-Holland: **should-fix — duplicate parser: this collides head-on with #1858.**

#1858 ("Decode PROXY protocol v2 with forwarded mTLS client certificates on UDS mirrors") is still `OPEN` at `5bc789a0`, and neither parser is on `main` — I checked both `server/proxyProtocol.ts` and `server/serverHelpers/proxyProtocol.ts` against `origin/main`, both absent. So the two PRs each add an independent PROXY v2 decoder at a different path. Their file sets overlap on `server/http.ts` and `unitTests/server/udsMirror.test.js`, and the design docs diverge (`DESIGN.md` here, `server/DESIGN.md` there).

They also differ in security posture, which is the part that matters: **#1858 makes forwarded mTLS load-bearing for authorization** (`socket.authorized = tls.verified` plus a synthesized `getPeerCertificate()`, so the existing HTTP/MQTT mTLS paths authenticate on the forwarded chain). **This PR deliberately does not** — it surfaces `clientCertificateVerified`/`clientCertificates` with no consumer.

Merging both leaves two v2 parsers, two DESIGN docs, and conflicting `createH2CProxyFront` refactors.

**Ask:** which supersedes? If this one, close #1858. If #1858 is still wanted, rebase it onto this parser instead of landing a second decoder. The description cites symphony #29–#32 and harper#914 but never #1858.

server/http.ts
```diff
@@ -714,9 +715,9 @@ function getHTTPServer(port: number, secure: boolean, options: ServerOptions) {
 
 			if (process.env.HARPER_UWS_UDS) {
 				// uWS backend (#914): serve the UDS mirror with uWebSockets.js instead of a Node http
-				// server. threadServer.js consumes uwsServeConfigs and calls createUwsServer(). uWS does
-				// not parse the PROXY protocol, so symphony must use sourceAddressHeader: 'xForwardedFor'
-				// for this socket (the same mode it uses for the Bun path).
+				// server. threadServer.js consumes uwsServeConfigs and calls createUwsServer(). uWS
+				// parses PROXY v2 natively (res.getProxiedRemoteAddress()) but not v1, so this socket
```
Devin-Holland: **should-fix — this comment instructs a symphony config change that would lose the client IP.**

Nothing in Harper calls `getProxiedRemoteAddress()`. The only peer-address reads on the uWS path are `res.getRemoteAddressAsText()` at `server/serverHelpers/uwsServer.ts:137` and `:389`, and `:137` is explicitly gated `port != null ? … : undefined`. For this UDS mirror the config passes only `socketPath` (no `port`), so `createUwsServer`'s destructured `port` is `undefined` and `ip` is left unset **by design** — `uwsServer.ts:130-136` says so: *"behind symphony on a UDS it's left unset so the real client IP comes from X-Forwarded-For"*.

So if an operator follows this comment and switches the socket to `proxyProtocolV2`, XFF goes away, `request.ip` stays `undefined`, and rate limiting and request logging lose the client IP. Local-auth fails closed rather than escalating (`security/auth.ts:271` uses `request.ip?.includes('127.0.0.')`), so this is attribution loss, not an auth hole.

**Fix:** either keep the `xForwardedFor` instruction for the uWS socket until `uwsServer.ts` reads `getProxiedRemoteAddress()`, or make it read it as part of this PR. Same wording needs correcting in `DESIGN.md` (see the other comment) — 2 sites.

DESIGN.md
```diff
@@ -391,3 +391,38 @@ extensionModule.handleApplication` gate (`components/componentLoader.ts`) means
...
+- **A parsed header can be shorter than the buffer that carried it.** Check `status === 'header'`
+  _before_ the length bound in a read loop, or a single coalesced read holding a complete header
+  plus a large body gets handed off with the header un-stripped.
+
+`connection: null` distinguishes "valid header, no usable identity" (LOCAL command, AF_UNSPEC, or a
+body too short for its declared family) from `absent` ("not a PROXY header, forward verbatim"). The
+first still consumes its bytes; the second must not.
+
+**uWS reads v2 itself, and only v2.** The uWS UDS backend (`HARPER_UWS_UDS`, #914) never reaches
+this parser — uWebSockets.js parses PROXY v2 natively via `res.getProxiedRemoteAddress()` — but it
```
Devin-Holland: **should-fix (same class as the `server/http.ts:719` comment — 2nd of 2 sites).**

*"uWebSockets.js parses PROXY v2 natively via `res.getProxiedRemoteAddress()`… so that socket needs symphony's `proxyProtocolV2` mode specifically"* — Harper never calls `getProxiedRemoteAddress()`, and `server/serverHelpers/uwsServer.ts:130-137` deliberately leaves `ip` undefined on the UDS path so the client IP arrives via X-Forwarded-For.

Granting that uWS itself parses v2, the sentence still tells a reader that this socket recovers the client address under `proxyProtocolV2`, and it doesn't. Either add "…once `uwsServer.ts` reads it — today this socket still needs `xForwardedFor`", or land the `getProxiedRemoteAddress()` read.

server/http.ts
```diff
@@ -1586,46 +1583,27 @@ export function enableProxyProtocol(httpServer) {
...
-						pending = chunk;
-						return;
-					}
-					headerHandled = true;
-					pending = null;
-					return forward(chunk);
+				const result = parseProxyHeader(chunk);
+				// Keep buffering until the header completes, bounded by the longest one that could
+				// still turn out to be valid.
+				if (result.status === 'incomplete' && chunk.length < PROXY_MAX_HEADER) {
```
Devin-Holland: **nit — the incomplete-header buffering ceiling grew ~600x on a path with no read timeout.**

This now buffers while `chunk.length < PROXY_MAX_HEADER` (65,551) where the old v1 code capped at 108. Necessary for v2 (a body can legitimately declare `0xffff`), and it is bounded — but a peer that sends the 12-byte signature, declares `bodyLength: 0xffff`, then stalls holds ~64 KB instead of 108 bytes indefinitely, because this data-event path has no equivalent of `createH2CProxyFront`'s `prehandoffTimeout` (10 s). The indefinite hold is pre-existing; only the ceiling changed.

Secondary: buffering is `Buffer.concat` per data event, so a header dribbled a byte at a time to 65,551 copies ~2 GB cumulatively.

Both are UDS-only and therefore self-inflicted at worst, which is why this is a nit. A `socket.setTimeout` on this path would close it and match the h2 front.

Devin-Holland (commented):
Reviewed at head `190cbdce`. Rather than run the PR's own tests (which can share a wrong assumption with the code), I rebuilt v2 headers from this PR's own `proxyProtocolFixtures.js` and drove `parseProxyHeader` directly. I could not find a defect in the parser. Four items are inline; the first needs an answer before merge, the second is a doc/config contradiction I can reproduce from source.

**Scope note:** I could not bootstrap the mocha suite in my checkout (`unitTests/mocha.init.js` needs a built `dist/`), so I did not run the PR's tests. Everything below I executed against the parser module itself.

**Verified by executing the parser at this head:**

- **Byte consumption is exact and the trailing payload survives.** A v2 header with ALPN + authority + SSL (with sub-TLVs) + JA3 + JA4 + two cert TLVs, followed by an HTTP request: `consumed: 132` against an actual header length of 132, and `subarray(consumed)` returned the request bytes unchanged. Every field decoded (`203.0.113.45:54321`, `alpn h2`, `authority api.example.com`, `TLSv1.3`, `TLS_AES_128_GCM_SHA256`, ja3/ja4, both certs leaf-first). An off-by-one here would corrupt the first packet of every MQTT/replication stream, so this was the property that mattered most.
- **The TLV walk is bounds-safe and cannot spin.** Driven with a trailing TLV declaring `0xffff` against 2 remaining bytes: the walk stopped, kept the previously-parsed `authority`, and left `clientCertificates` undefined — no throw, no over-read. Cert values are `Buffer.from(value)` copies, so they don't pin the recycled read buffer.
- **`clientCertificateVerified` is conservative in the right direction.** It requires the cert-present bit *and* `verify === 0`. Fed a bitfield with cert-present clear but `verify: 0`, it reported `false` (with `tlsTerminated: true`). #1858 tracked those as separate flags and mapped `verified` straight onto `socket.authorized`; combining them is strictly safer.
- **Headers with no usable identity are consumed, not guessed at.** LOCAL (`0x20`) → `connection: null`, `consumed` exactly the header length. A body shorter than the family it declares (8 bytes against AF_INET's 12) → `connection: null`, `consumed: 24`. Framing honoured, contents distrusted.
- **The v1 `TCP4`/`TCP6` tightening works.** `PROXY UNKNOWN a b c d\r\n` (six tokens, which the old token-count-only check would have read as an address) → `connection: null`, `consumed: 23`, header still consumed.
- **No protocol-preface collision.** The h2 preface `PRI * HTTP/2.0…` → `absent` (both branches start with `P`); a 2-byte `PR` → `incomplete`, matching the old buffering behavior.
- **`formatIPv6` is RFC 5952-correct.** `2001:db8::1` round-trips, and an IPv4-mapped v6 block renders `::ffff:203.0.113.9`.
- **`applyProxiedConnection` is non-enumerable and no-ops on null** (`Object.keys(socket).length === 0` after applying; `proxiedConnection` left unset when `connection` is null).
- **The UDS-only trust boundary holds at this head** — the property everything else rests on. `enableProxyProtocol` is attached at exactly two places: `http.ts:746` (inside the UDS-mirror block) and `threadServer.js:618` (inside `if (env.get(TLS_UNIXDOMAINSOCKETS))`). The plaintext TCP listener gets the raw `socketServer`. `createH2CProxyFront` is attached once behind `HARPER_H2C_UDS` on a UDS path. No TCP path can inject a PROXY header, and since nothing consumes the cert fields there is genuinely no permission change here.
- **The h2 read loop honours the ordering trap DESIGN.md calls out** — `status === 'header'` is checked *before* the length bound, so a single coalesced read holding a header plus a body doesn't get handed off with the header un-stripped.

Verdict: comment.

—
Claude & DevAIn (Claude Opus 5)

server/http.ts
kriszyp: Fixed — corrected the comment (`server/http.ts:717-720`): `createUwsServer()` doesn't call `getProxiedRemoteAddress()`, so this socket still needs symphony's `xForwardedFor` mode for the client IP today. Left a note that `proxyProtocolV2` becomes correct once `uwsServer.ts` reads it. In 98a8efeb5.

DESIGN.md
kriszyp: Fixed — same correction applied to `DESIGN.md`, same wording as the `server/http.ts:719` fix (2nd of 2 sites). In 98a8efeb5.

server/http.ts
kriszyp: Fixed — `enableProxyProtocol` now bounds the pre-handoff wait with `socket.setTimeout(prehandoffTimeout, ...)` / destroy, mirroring `createH2CProxyFront`'s existing `prehandoffTimeout` (default 10s). A stalled peer now gets dropped instead of holding the buffer indefinitely. In 98a8efeb5.

kriszyp:
Resolving — #1858 has merged, and Kris confirmed it supersedes this PR's parser. Rebased this branch onto `main` (99935a4d1):

- Dropped `server/proxyProtocol.ts`, its dedicated unit tests, and the `DESIGN.md` framing section — all superseded by #1858's `server/serverHelpers/proxyProtocol.ts`.
- Kept and ported forward the one piece of independent value: the stall-timeout guard on `enableProxyProtocol`'s pre-handoff header buffering that this PR's Codex review flagged (this thread's sibling, already resolved). #1858's rewritten `enableProxyProtocol` reimplements the same buffering pattern but without the guard, so a peer that starts a PROXY v2 header, declares a large body, then stalls could still hold the buffer indefinitely — `withProxyProtocol` (used elsewhere) already has this protection, `enableProxyProtocol` didn't.
- Added test coverage for the new timeout path in `unitTests/server/udsMirror.test.js`, and a one-line `server/DESIGN.md` note.

PR title/description updated to reflect the narrower scope.

_Signed, Claude Sonnet 5_

cb1kenobi (approved):

Devin-Holland (approved):
Approving. The PR is now scoped to exactly the one thing I raised as a nit — #1858 landed the parser and both comment corrections, so what's left here is just the stall-timeout guard.

**Verified this round:**

- **It mirrors the precedent it claims to.** `withProxyProtocol` (`server/serverHelpers/proxyProtocol.ts:239-252`, now on `main`) does the same three things in the same order: `socket.setTimeout(prehandoffTimeout, onPrehandoff)` on entry, then `setTimeout(0)` plus `removeListener('timeout', …)` at handoff. Same default, same shape.
- **`setTimeout(0)` doesn't clobber anything.** I checked what Harper actually configures on these servers: `keepAliveTimeout` and `headersTimeout` (`server/http.ts:699-700`), both Node-internal timers, and nothing ever assigns `server.timeout` — so there is no socket-level idle timeout being disabled here. And clearing the listener rather than only disarming is the right instinct given `keepAliveTimeout` *is* set: left attached, it really could have fired a `destroy()` on a later keep-alive timeout.

**One non-blocking note.** `socket.setTimeout` is an *idle* timer, reset by every data event, so it bounds a fully-stalled peer but not a slow-dribble one — a byte every 9s still holds the pending buffer indefinitely, and the per-event `Buffer.concat` is unchanged, so a header dribbled to the 65,551 ceiling still costs ~2 GB cumulative allocation. That is the same property as `withProxyProtocol`, which is what I asked you to match, so it is parity rather than a regression, and UDS-only either way. Recording it, not asking for a change.

One one-click nit on the test double, inline.

🤖 Reviewed by Claude Opus 5 via Claude Code

unitTests/server/udsMirror.test.js
```diff
@@ -177,6 +177,18 @@ describe('UDS mirror (writeUdsMetadata, cleanup helpers)', () => {
...
 
 	describe('enableProxyProtocol', () => {
+		// Plain EventEmitter plus the net.Socket surface enableProxyProtocol depends on
+		// (setTimeout/destroy), matching net.Socket.setTimeout's real semantics closely enough
+		// for tests: registers the callback as a 'timeout' listener; a 0ms call just disarms it.
+		function createSocket() {
+			const socket = new EventEmitter();
+			socket.setTimeout = sinon.stub().callsFake((ms, cb) => {
+				if (cb) socket.on('timeout', cb);
+			});
```
Devin-Holland: nit — `net.Socket.setTimeout` registers its callback with `once`, not `on` (`Socket.prototype.setTimeout` does `this.once('timeout', callback)`), so this double can't catch a listener registered twice. The comment above already flags the approximation; `once` costs nothing and keeps the stub honest.

```suggestion
				if (cb) socket.once('timeout', cb);
```

PR #1948, opened 2026-07-26
fix(resources): await the request body in the built-in login resource

kriszyp:
## What / why

`Login.post` (`resources/login.ts`) is a **static** override, so the `transactional()` wrapper
that would normally resolve `data` never runs for it — the body arrives as the streaming
deserializer's pending promise (`server/REST.ts` → `getDeserializer(…, true)`). It was
destructured directly:

```js
const { username, password, token } = body;   // body is a pending Promise
```

A promise has no own enumerable properties, so this fails **silently**: all three read
`undefined` and `request.login(undefined, undefined, undefined)` is called with no credentials.
One-line fix — `await body`.

Also documents the static-override contract in `resources/DESIGN.md`, since that's what let this
get missed: overriding a static entry point takes over that entry point's whole contract, so the
override owns both (a) awaiting `data` and (b) making the access-control decision.

## History — this PR started out much bigger, and was wrong

The originating QA findings (F-224 body loss / F-223 "authz bypass") framed the shadowing of
`Resource.post`/`put`/`patch` by a subclass static as a framework defect, and the first version of
this PR added a `Resource.authorizeStaticOverride()` gate wired into REST and MCP dispatch.

That premise was wrong, and @kriszyp called it: the static methods **are** the protocol entry
points, `data` is contractually a `MaybePromise`, and an override that replaces the entry point
is meant to own the authorization decision. It isn't attacker-reachable — the app author has to
write the override. The gate also **broke the built-in login** (caught by the Codex review leg):
`Login.post !== Resource.post`, so the gate fired and `Login`'s inherited super-user-only
`allowCreate` denied anonymous `POST /login`.

So the gate is reverted (commit `4bfa61737`) and what ships is the one real bug underneath:
`login.ts` was the single in-tree override that forgot its `await`.

## Testing

`unitTests/resources/login.test.js` — verified it genuinely catches the bug (reverting the
`await` makes it fail with `undefined` credentials, as expected). `test:unit:resources` 1261
passing / 0 failing; `test:unit:main` unchanged against its pre-existing baseline (3789 passing,
same 10 pre-existing unrelated failures). Build + oxlint clean.

Generated by Claude Opus (dev-agent).


resources/login.ts
```diff
@@ -12,7 +12,9 @@ class Login extends Resource {
 		// TODO: Return a login page
 	}
 	static async post(_id, body, request) {
-		const { username, password, token } = body;
+		// a static override replaces the transactional() wrapper that would otherwise resolve this,
+		// so the body arrives as a pending promise over REST (see resources/DESIGN.md)
+		const { username, password, token } = await body;
```
cb1kenobi: **Nit (hardening): a resolved `null`/non-object body throws instead of failing auth cleanly**

The `await` fix is correct — over REST `body` is the streaming deserializer's pending promise, and awaiting it is what makes credentials readable. One narrow edge remains: if the body resolves to `null` (e.g. an unauthenticated `POST /login` with a JSON `null` payload), destructuring throws a `TypeError`, which the REST handler turns into a 500 rather than the clean 401 you'd get from empty credentials. Not a crash (the handler catches it at `server/REST.ts:354`) and not a regression for the absent-body case (that already threw), but a one-line guard keeps malformed public input on the 4xx path:

Suggested fix:
```suggestion
		const { username, password, token } = (await body) ?? {};
```

—
_Generated by Barber AI_

kriszyp:
CI fix update (dev-agent):

- Rebased onto latest `main` — picked up #1860 (fix(txn): don't drop writes staged while read iterators defer the commit), which resolved the Unit Test v22/v24 failures (audit-log lost-write / RocksDB "invalid column family" symptoms) and the Windows Integration Tests 6/6 EPERM rename flake was gone after the rebase too.
- Fixed a real bug while investigating: `unitTests/resources/auditLog.test.js:55` used a fixed 20ms delay waiting for subscription events, while the very next check in the same test had already hit and fixed this exact "too short on a loaded CI runner" issue with a poll loop — applied the same pattern.
- That fix is a genuine improvement but didn't fully eliminate a **pre-existing, Node-v22-specific flake** in the same assertion — confirmed it already fails intermittently on `main` itself, independent of this PR. Filed #1966 to track/root-cause it separately; not blocking this PR's actual change.

— Claude (dev-agent)

cb1kenobi (approved):

Issue #678, opened 2026-05-21
super.post() throws uncaught exception when POSTing to resource without trailing slash


kriszyp:
When a Resource class exported by `harper/redirects` (or similar) is POSTed to **without a trailing slash**, `super.post()` throws an unhandled exception:

```
Cannot read properties of undefined (reading 'query')
```

With a trailing slash but a random path segment (e.g. `/redirects/hi`), the error changes slightly:

```
Cannot read properties of null (reading 'query')
```

## Context

- Trailing slash IS required per design and this is documented — it's not a missing-feature issue.
- The problem is the **unhandled exception** surfacing to `super.post()` rather than a clean 404 or 405.
- Was silently working in 4.3; broke in 4.5 in a way that erodes confidence.

## Ask

Catch the missing-trailing-slash case in the Resource request dispatch and return an appropriate HTTP error (404 or 405 with a clear message) rather than letting the `Cannot read properties of undefined` exception propagate.

---

🤖 Filed by Claude on behalf of Kris.

PR #1956, opened 2026-07-27
fix(resources): restore v4 super.post create on collection posts

ldt1996:
Restores v4-style instance post overrides that delegate to super.post on a bare collection path. The compatibility normalization runs before authorization and dispatch so the bare and slash forms both use collection/insert semantics, while default bare POST retains its 404 and loadAsInstance=false resources retain their distinct base-resource dispatch. Includes regressions for target parsing, default behavior, insert-vs-update authorization, and false-mode Resource/Collection separation.

Generated with GPT-5.6 Codex.


ldt1996:
unit failures are pre existing on main

unitTests/resources/crud.test.js
```diff
@@ -346,3 +346,42 @@ describe('transactional argument normalization with RequestTarget', () => {
...
+		Object.defineProperty(PostSub, 'name', { value: 'PostSub' });
+	});
+
+	it('super.post(record) from a bare collection target creates the record', async function () {
+		const target = new RequestTarget('');
+		const id = await PostSub.post(target, { title: 'created' });
+		assert.ok(id != null, 'create should return the new id');
+		const record = await PostSub.get(id);
+		assert.equal(record.title, 'created');
+		assert.equal(record.stamped, true, 'instance override should have run');
```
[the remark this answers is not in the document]
ldt1996: fixed

resources/Resource.ts
```diff
@@ -359,7 +359,8 @@ export class Resource<Record extends object = any> implements ResourceInterface<
 				return newRecord?.[(this.constructor as any).primaryKey as keyof typeof newRecord] as any;
 			}
 		} else {
-			if (this.#isCollection) {
+			// no id is the collection form; posting a record to it creates it (the v4 super.post contract)
+			if (this.#isCollection || this.#id == null) {
```
kriszyp: `RequestTarget()` deliberately leaves both `id` and `isCollection` undefined for internal callers to finish configuring, and the new query-parsing test pins that distinction. Because `== null` also matches that incomplete state, `Table.post(new RequestTarget(), data)` now inserts a row where it previously returned 405. The v4 bare-path compatibility case has `id === null`; could we use an exact null check here and add a regression assertion that an argless target remains rejected?

kriszyp (dismissed):
Good catch. I think #1807 had this mostly covered, but the override case was missing, thank you for tracking down.

resources/Resource.ts
ldt1996: great catch @kriszyp , I've switched to an exact null check

resources/Resource.ts
```diff
@@ -359,7 +359,8 @@ export class Resource<Record extends object = any> implements ResourceInterface<
 				return newRecord?.[(this.constructor as any).primaryKey as keyof typeof newRecord] as any;
 			}
 		} else {
-			if (this.#isCollection) {
+			// null id is the bare collection path and creates the record; undefined id is an unconfigured target and rejects
+			if (this.#isCollection || this.#id === null) {
```
heskew: **Codex review (OpenAI):** This new null-ID condition makes a pre-existing authorization mismatch actionable. `RequestTarget('')` has `id === null` and `isCollection === false`, so outer POST authorization constructs a non-collection table instance and `Table.allowCreate()` delegates to `allowUpdate()`. This branch then creates the record, and the nested `create()` receives a primitive target with no `checkPermission`. As a result, a user with `update: true` and `insert: false` can create through the bare path, while the equivalent `/` collection target correctly rejects.

I reproduced this on the current PR head with a focused test. Please make this compatibility path use collection/insert authorization and add an update-only/insert-denied regression covering both bare and slash collection targets.

_Posted by Codex (OpenAI) at Nathan Heskew's request._

---

heskew: codex was adamant that modifying this condition exposes another issue. It's not something you did directly here but it could be a good chance to get some test coverage in place to confirm.

---

kriszyp: I checked the historical default-table behavior before deciding whether the 404 should remain:

- `harperdb@4.7.0` and `harperdb@4.7.36`: `POST /Table` reaches the non-collection handler and returns 405; `POST /Table/` creates.
- Harper `5.1.0` and `5.1.23`: the same 405/create split.
- Current main after [#1807 “fix: return clean 404 instead of crashing on POST without a trailing slash”](https://github.com/HarperFast/harper/pull/1807): the default bare POST returns 404.

So #1807’s 404 is a deliberate error-classification change from 405, not required for back-compat. [#678 “super.post() throws uncaught exception when POSTing to resource without trailing slash”](https://github.com/HarperFast/harper/issues/678) accepted either 404 or 405; the required fix was avoiding the 500. Given that the bare collection URL is invalid and we prefer the clearer 404, I recommend keeping that default behavior.

For this PR, I would not broaden `Resource.prototype.post()` with `this.#id === null`. At that point authorization has already treated the target as a non-collection/update, which creates the mismatch identified above. Instead, normalize the compatibility case in `transactional()` before `getResource()` and authorization:

1. Detect a bare POST target (`id === null && !isCollection`).
2. If the resource uses the default `Resource.prototype.post`, keep the existing 404.
3. If it has a custom `post()` override, normalize the target to a well-formed collection (`id: null, isCollection: true`) before authorization and dispatch.

That preserves the redirector/v4 custom-`post()` compatibility path while making both `/Table` and `/Table/` use insert authorization. Once normalized early, `super.post()` sees a collection resource, so the new null-ID fallback should not be necessary.

Please add regression coverage showing:

- default bare table POST returns 404;
- slash collection POST creates;
- the legacy custom bare POST succeeds for a user with insert permission;
- both bare and slash forms reject a user with `update: true, insert: false`.

— KrAIs (OpenAI Codex)

kriszyp: I dunno why GPT was shy about sharing its code, but concretely:
In transactional(), detect the accepted legacy case:
```
const isBarePost = options.method === 'post' && query.id === null && !isCollection;

if (isBarePost) {
	if (this.prototype.post === Resource.prototype.post) {
		throw new ClientError(/* existing 404 */);
	}
	isCollection = true;
}
```

ldt1996: I tried to address both of your comments in my last commit

unitTests/resources/permissions.test.js
```diff
@@ -237,3 +237,55 @@ describe('Permissions through Resource API', () => {
...
+	before(async function () {
+		setupTestDBPath();
+		setMainIsWorker(true);
+		PostBase = table({
+			table: 'BarePostTable',
+			database: 'test',
+			attributes: [
+				{ name: 'id', isPrimaryKey: true },
+				{ name: 'title' },
+			],
```
cb1kenobi: **Nit: fails the required Format Check.**

Prettier collapses this short `attributes` array onto one line (it fits within printWidth), so `format:check` is red on this file and blocks merge. Run `prettier --write`, or apply:

```suggestion
			attributes: [{ name: 'id', isPrimaryKey: true }, { name: 'title' }],
```

—
_Generated by Barber AI_

ldt1996: applied

resources/Resource.ts
```diff
@@ -711,23 +711,17 @@ function transactional(
...
+		if (options.method === 'post' && query.id === null && !isCollection) {
+			if (this.prototype.post === Resource.prototype.post) {
+				// exact base-path match with no trailing slash (harper#678): the default post dispatch
+				// has nothing to address, so reject with a purpose-built message
+				throw new ClientError(`A trailing slash is required to POST to the ${this.name} collection`, 404);
+			}
+			// a custom post() override accepts the no-slash collection POST (redirector-style bulk
+			// import); normalize to a well-formed collection target so authorization and dispatch
+			// treat it as an insert on the collection, same as the trailing-slash form
+			query.isCollection = true;
```
kriszyp: This compatibility normalization is broader than the v4 table case. Any Resource with an instance `post` override now receives `isCollection: true` for an exact no-slash base path; on `origin/main` it receives `false`. More seriously, `getResource()` uses this flag to select `static Collection`, so a custom Resource that intentionally exposes distinct `/foo` and `/foo/` handlers can now dispatch `/foo` to the collection handler. Please scope this conversion to the legacy instance-loaded table POST path that needs the `Resource.prototype.post` create fallback (or an equivalent internal compatibility predicate), preserve the non-collection target for arbitrary custom/`loadAsInstance = false` Resources, and add a regression test covering that distinction.

kriszyp (dismissed):
Ok, I guess GPT thinks this would be a lot safer to restrict this to `this.loadAsInstance !== false`.

unitTests/resources/permissions.test.js
kriszyp: Addressed. The branch was reformatted and the focused format check is clean.

resources/Resource.ts
kriszyp: Addressed in 49a9319f9: bare-path collection normalization is now restricted to loadAsInstance !== false. A regression gives the false-mode resource a distinct Collection handler and proves the bare POST retains isCollection=false and dispatches on the base receiver.

resources/Resource.ts
kriszyp: Addressed: the branch uses the exact query.id === null check, and query parsing coverage verifies that an argumentless RequestTarget remains an incomplete, rejected state.

kriszyp (commented):
Review threads addressed.

kriszyp (approved):
Aligned with #1842: instance-loaded legacy bare POSTs normalize before authorization, while loadAsInstance=false custom handlers retain their existing Resource/Collection dispatch boundary. All review threads are addressed and CI is green.

Issue #1381, opened 2026-06-18
Cross-worker write can race RocksDB table drop and poison catalog cleanup


heskew:
## Summary
Intermittently, a block of resources unit tests fail in their `before all` / `after all` hooks because `resetDatabases()` (called from `setupTestDBPath`) hits a column-family error while completing an interrupted table drop left over from a prior test. One run produced ~12 cascading hook failures from this single root cause. Re-running the job goes green (verified: attempt 2 passed on the same commit), so it's a test-isolation / state-leak flake.

## Exact errors
```
[main/0] [error]: Error committing cache update [Error: Transaction commit failed: Invalid argument: Invalid column family specified in write batch] { code: 'ERR_INVALID_ARGUMENT' }
[main/0] [error] [storage]: Failed to complete interrupted drop of table <db>.TestTableForContext; will retry on next start Error: Remove failed: Invalid argument: Invalid column family specified in write batch
```
Stack:
```
Store.removeSync (@harperfast/rocksdb-js/src/store.ts:824)
RocksDatabase.removeSync (@harperfast/rocksdb-js/src/dbi.ts:461)
completeInterruptedDrop (resources/databases.ts:1638)
initStores (resources/databases.ts:541)
readRocksMetaDb (resources/databases.ts:439)
getDatabases (resources/databases.ts:331)
resetDatabases (resources/databases.ts:747)
setupTestDBPath (unitTests/testUtils.js:386)
```
Resulting cascade (sample): `"after all" hook for "should handle source.get returning null"`, `"before all" hook for "It can search_by_conditions on a resource"`, plus indexing/subscription/txn `before all` hooks.

## Why it's a flake
A prior test leaves a `TestTableForContext` drop half-completed (tombstoned) in the rocks meta DB; the next test's `setupTestDBPath → resetDatabases → readRocksMetaDb → initStores → completeInterruptedDrop` tries to finish it and the column family is already gone/inconsistent, so `removeSync` throws. Order/timing-dependent; green on re-run, same commit.

## Affected job / runtime
Observed on `Unit Test (Node.js v26)`, but the mechanism (shared-process unit run, leaked drop state) is not runtime-specific.

## Evidence
- PR #1363 run 27761092922 / job 82135389024 (attempt 1; attempt 2 green)

## Related
- #1276 (residual table-drop partial-failure risks — `completeInterruptedDrop` removing catalog rows on persistent CF-drop failure): same code path, but that's production-hardening; this is the test-isolation manifestation.
[...]

PR #1957, opened 2026-07-27
fix(storage): bound the interrupted-drop retry to one actionable error

kriszyp:
## What

Bounds the `completeInterruptedDrop` retry in the schema-load reconcile to 3 attempts per table per process, and replaces N identical `error` lines with one actionable error.

**This PR fixes only the log flood.** The underlying data-loss/blast-radius defect is root-caused below but deliberately not fixed here — the fix belongs in `@harperfast/rocksdb-js` and needs a direction decision first.

## Root cause of the 44-test cascade (Unit Test run [30237981148](https://github.com/HarperFast/harper/actions/runs/30237981148), sha `fe3994f4`, Node v24)

Refs #1381, which describes this failure but attributes it to "the column family is already gone/inconsistent". The real mechanism is worse and is not test-specific.

**A single write through a dropped RocksDB column-family handle latches a background error on the entire RocksDB environment. From that instant, every write to every column family in that environment fails with the same generic `Invalid argument: Invalid column family specified in write batch` — including column families created after the fact — until the environment is fully closed and reopened. Reads keep working.**

Reproduced directly against `@harperfast/rocksdb-js@2.5.0` (the version CI resolved):

```
baseline victim write ok: v0
columns after drop: [ 'default', 'victim' ]
--- victim write AFTER the drop, before any stale write ---
  ok                                    <-- the drop alone is harmless
--- write through the STALE dropped handle ---
  FAILED: Put failed: Invalid argument: Invalid column family specified in write batch
--- victim write AFTER the stale write ---
  POISONED: Put failed: Invalid argument: Invalid column family specified in write batch
--- a brand-new, unrelated CF opened after the poison ---
  POISONED: Put failed: Invalid argument: Invalid column family specified in write batch
--- reads still work? ---
  victim k0 = v0
```

A follow-up run confirms the latch clears only on a full close + reopen of every handle to that path — i.e. in production, a node restart.

`test/drop.test.ts` in rocksdb-js already documents this in a comment ("the failed write contaminates the env's shared write path, so writes through OTHER handles in this database fail afterward too"), but nothing enforces it.

### How harper triggered it

1. `unitTests/resources/Resource-get-context.test.js` exercises a sourced (cache) table. Each `Resource.get` that resolves from source stages a **background** cache write and returns to the caller without awaiting it — `resources/Table.ts:5641` is the error handler for exactly that fire-and-forget commit.
2. The suite's `after` hook calls `await TestTable.dropTable()`. Nothing quiesces those in-flight commits, so `primaryStore.dropSync()` (`resources/Table.ts:1259`) drops the column family out from under them.
3. Those commits land on the dropped CF and poison the environment. The CI log shows them, including a CF-id-specific variant that only a dropped CF produces:
   ```
   [main/0] [error]: Error committing cache update Error: Transaction commit failed: Invalid argument: Invalid column family specified in write batch
   [main/0] [error]: Error committing cache update Error: Transaction commit failed: Invalid argument: Could not access column family 154
   ```
4. The **very next write in the drop itself** — `dbisDb.remove` in `removeTombstonedCatalog` (`resources/Table.ts:1231`) — is the first victim. `dropTable` throws and the `dropping` tombstone survives, as designed.
5. Every subsequent test in the process then fails, not on its own table but on the shared catalog store: `table()` → `dbisDb.putSync` at `resources/databases.ts:1243`. Worker threads fail identically on their own unrelated tables (`vectorIndex-thread.js` → `Transaction commit failed: ...`). 44 failing tests.
6. `completeInterruptedDrop` runs once per database per `resetDatabases()`, fails identically every time, and logged an identical error each time — 4 databases × ~25 test files ≈ the ~100-line flood. That is what this PR fixes.

So the two symptoms in #1381 have one cause, and the blast radius is not a test artifact: in production, any drop that races an in-flight write to that table takes down **all writes to that whole database** until the node restarts.

## What this PR changes

`resources/databases.ts` — the reconcile loop in `initStores` now tracks consecutive failures per `database.table` and stops after 3, logging intermediate attempts at `debug` and one `error` that names the table, says what was left behind, and explains what an `Invalid column family` cause implies.

The budget is scoped to the reconcile. The create path (`databases.ts:1255`) also calls `completeInterruptedDrop`, under the exclusive lock, and must never skip it — creating over a half-dropped table would resurrect its catalog rows — so there the failure keeps propagating to the caller as its own single error.

Flood in the observed scenario goes from ~100 error lines to 4 (one per stuck table).

## What is NOT fixed here

- **The invariant, unenforced:** *no write may be issued against a dropped column family.* Nothing checks it; RocksDB discovers the violation inside a shared write batch and takes the whole environment down. The containing fix belongs in rocksdb-js — mark `ColumnFamilyDescriptor` dropped on `DropColumnFamily` and reject writes through a dropped handle with a scoped error naming the column family, before RocksDB sees the batch. That converts an environment-killing fail-open into one legible, contained failure. Its regression test belongs at that layer too (writes to CF X keep working after CF Y is dropped and written through).
- **The trigger, in harper:** `dropTable` does not quiesce in-flight writes to the table before dropping its column families. Removing the table from `databases[db][name]` stops new logical operations in the dropping thread only; already-staged transactions and other worker threads are unaffected. Fixing this properly is a change to how drops sequence against the transaction layer.

Both are being raised for a direction decision before any cross-repo work starts.

## Test notes

- New case in `unitTests/resources/dropTableGhost.test.js`: makes `completeInterruptedDrop`'s catalog removal fail permanently, drives 10 schema reloads, and asserts attempts stop after the budget and that each stuck table logs exactly one give-up error naming it.
- Negative control: with the `databases.ts` change reverted the new test fails with `completion attempts must stop, not repeat on every schema reload (20 more over 5 further reloads)`; the other 6 cases in the file still pass.
- `npm run test:unit:resources` — 1264 passing, 14 pending.
- `npm run test:unit:main` — 3801 passing, 188 pending.
- `npm run lint:required` clean; prettier run on the two changed files only.

Generated by Claude Opus 5.

### DIFF
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -227,7 +227,15 @@ const MAX_INTERRUPTED_DROP_ATTEMPTS = 3;
 // Keyed with a NUL separator rather than a dot so that database "a.b" table "c"
 // and database "a" table "b.c" cannot share a budget.
 const interruptedDropAttempts = new Map<string, number>();
-const interruptedDropKey = (databaseName: string, tableName: string) => `${databaseName}\0${tableName}`;
+// Scoped by drop generation (Table.ts stamps a fresh id on every dropping
+// tombstone) rather than just database + table: a worker can exhaust one
+// drop's budget, then observe the table recreated and dropped again without
+// ever seeing an intermediate non-tombstoned row to reset on. Keying by
+// generation means the new drop's tombstone always carries a fresh key,
+// independent of what any worker last observed. Tombstones written before
+// this field existed fall back to a shared 'legacy' bucket.
+const interruptedDropKey = (databaseName: string, tableName: string, generation?: string) =>
+	`${databaseName}\0${tableName}\0${generation ?? 'legacy'}`;
 let loadedDatabases; // indicates if we have loaded databases from the file system yet
 
 // This is used to track all the databases that are found when iterating through the file system so that anything that is missing
@@ -555,7 +563,7 @@ function initStores(
 	// it, because creating over a half-dropped table would resurrect its catalog
 	// rows. There a failure propagates to the caller as its own single error.
 	for (const [tableName, tableDef] of tablesToLoad) {
-		const dropKey = interruptedDropKey(databaseName, tableName);
+		const dropKey = interruptedDropKey(databaseName, tableName, tableDef.primary?.dropGeneration);
 		if (!tableDef.primary?.dropping) {
 			// No tombstone, so any budget this table spent belongs to a drop that
 			// has since been resolved - by the create path, which completes
@@ -584,7 +592,7 @@ function initStores(
 					// including threads:0 where the main thread acts as worker 0 - logs it, so one
 					// stuck table produces one actionable error instead of one per worker.
 					logger.error(
-						`Unable to complete the interrupted drop of table ${dropLabel} after ${attempt} attempts; giving up until this node restarts. ${tableName} stays unloaded, with its catalog rows and column families left in place. An "Invalid column family specified in write batch" cause means the storage environment for ${databaseName} has latched a background error and will reject every write to every table in it until this node restarts.`,
+						`Unable to complete the interrupted drop of table ${dropLabel} after ${attempt} attempts; giving up until this worker restarts (a full node restart also resets this). ${tableName} stays unloaded, with its catalog rows and column families left in place. An "Invalid column family specified in write batch" cause means the storage environment for ${databaseName} has latched a background error and will reject every write to every table in it until this node restarts.`,
 						error
 					);
 				}
@@ -1275,8 +1283,10 @@ export function table<TableResourceType>(tableDefinition: TableDefinition): Tabl
 				// reconcile below, which is the only other place that returns a spent
 				// budget. Without clearing it here too, a table that gets dropped again
 				// before any reconcile observes it live in between would have its NEXT
-				// interrupted drop inherit this one's spent attempts.
-				interruptedDropAttempts.delete(interruptedDropKey(databaseName, tableName));
+				// interrupted drop inherit this one's spent attempts. Generation-scoped
+				// keying (see interruptedDropKey) already makes that impossible, but
+				// clearing it here too avoids leaving a dead entry behind.
+				interruptedDropAttempts.delete(interruptedDropKey(databaseName, tableName, existingTableMeta?.dropGeneration));
 			}
 			if (rootStore instanceof RocksDatabase) {
 				primaryStore = openRocksDatabase(rootStore.path, { ...dbiInit, name: dbiName, cache: true } as any);
@@ -1912,7 +1922,10 @@ function completeInterruptedDrop(rootStore, attributesDbi, databaseName: string,
 		}
 	}
 	for (const key of attributesDbi.getKeys({ start: tableName + '/', end: tableName + '0' })) {
-		attributesDbi.remove(key);
+		// removeSync (not remove): same reasoning as dropSync above - the async
+		// remove() rejects after this function has already returned, so a
+		// catalog-removal failure would bypass the retry accounting entirely.
+		(attributesDbi as any).removeSync(key);
 	}
 }
 
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -215,6 +215,16 @@ _assignPackageExport('databases', databases);
 _assignPackageExport('tables', tables);
 
 const NEXT_TABLE_ID = Symbol.for('next-table-id');
+// How many times the schema load will try to finish a tombstoned drop before
+// giving up for the rest of this process's lifetime. A drop that fails once
+// almost always fails identically forever - the usual cause is a RocksDB
+// environment that latched a background error and now rejects every write it
+// receives - and the schema load re-runs on every resetDatabases(), so an
+// unbounded retry turns one broken table into a per-reload log flood in every
+// worker thread.
+const MAX_INTERRUPTED_DROP_ATTEMPTS = 3;
+// `database.table` -> consecutive failed completion attempts in this thread.
+const interruptedDropAttempts = new Map<string, number>();
 let loadedDatabases; // indicates if we have loaded databases from the file system yet
 
 // This is used to track all the databases that are found when iterating through the file system so that anything that is missing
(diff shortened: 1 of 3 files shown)


kriszyp:
## Cross-model review (DLC step 10) — quick

Gemini (`agy`) leg on the diff; adjudicated inline. *Caveat: the `reviewer` subagent is not
installed in this profile, so the isolated Opus domain pass did not run, and only one outside
model was run for this PR (the thorough two-leg treatment went to the companion rocksdb-js
change, which carries the substantive risk).*

### Adopted

- **Significant — a spent budget outlived the drop that spent it** (`resources/databases.ts`).
  Real bug, fixed in 47464b9. The create path completes interrupted drops itself under the
  exclusive lock and never passes through the reconcile, so a table could go from "budget
  exhausted" to alive again without the reconcile ever seeing a success to clear the counter
  on. Its *next* interrupted drop would then be skipped outright and the leftover catalog rows
  would silently resurrect the table — the exact failure this PR's reconcile exists to prevent.
  The reconcile now returns the budget whenever it sees the table without a tombstone. New
  test `clears a spent retry budget once the table is no longer tombstoned`; it fails without
  the fix (attempts stuck at 12).

- **Significant — the test's fixed table name vs. a process-lifetime counter**
  (`unitTests/resources/dropTableGhost.test.js`). Fixed. An in-process re-run (a mocha retry)
  would have started with a spent budget and silently skipped the completion the test means to
  count. Both retry tests now use a unique name per run.

- **Suggestion — `${database}.${table}` key collision.** Fixed. Database `a.b` table `c` and
  database `a` table `b.c` shared a budget. Keyed with a NUL separator now; the dotted form is
  kept for the log messages, which are read by humans and by the test's assertion.

### Not adopted

- **"Blocker" — missing file extension on `require('#src/utility/logging/harper_logger')`.**
  Refuted. Extensionless `#src/...` subpath requires are the established pattern throughout
  `unitTests/` (`auditPurge.test.js`, `conditionsArrayMutation.test.js`, `copyApply.test.js`,
  …); the mapping is resolved by the `imports` field in `package.json`, and the TypeStrip
  extension rule applies to TypeScript source imports, not to these. Gemini was reviewing the
  diff without repo access and could not check the convention.

### Note for reviewers

Worth knowing while reading this PR: the root cause it was extracted from is fixed separately
in [rocksdb-js#725](https://github.com/HarperFast/rocksdb-js/pull/725). This change is the
containment for the *symptom* (the log flood from an unbounded retry) and stands on its own —
a drop can be left un-completable for reasons other than the environment latch this one hit.

Reviewed by Claude Sonnet 5.

resources/databases.ts
```diff
@@ -537,16 +550,45 @@ function initStores(
...
+				if (attempt < MAX_INTERRUPTED_DROP_ATTEMPTS) {
+					logger.debug(`Failed to complete interrupted drop of table ${dropLabel}, attempt ${attempt}`, error);
+				} else if (manageThreads.getWorkerIndex() === 0) {
+					// The attempt budget (and this failure) is independently tracked per worker
+					// thread, and the failure isn't transient, so every worker converges on the
+					// same give-up outcome. Only worker 0 - which exists in every threading mode,
+					// including threads:0 where the main thread acts as worker 0 - logs it, so one
+					// stuck table produces one actionable error instead of one per worker.
+					logger.error(
+						`Unable to complete the interrupted drop of table ${dropLabel} after ${attempt} attempts; giving up until this node restarts. ${tableName} stays unloaded, with its catalog rows and column families left in place. An "Invalid column family specified in write batch" cause means the storage environment for ${databaseName} has latched a background error and will reject every write to every table in it until this node restarts.`,
```
cb1kenobi: **Low: give-up message overstates permanence — a worker recycle also resumes retries**

`interruptedDropAttempts` is per-thread module state, and `manageThreads.restartWorkers()` replaces workers with fresh isolates (new module state, empty budget). So a worker recycle — not just a full node restart — resets the budget: the stuck table gets re-attempted (3 more times) and worker 0 re-emits this same error. The `until this node restarts` wording implies no further attempts before a full node restart, which isn't quite the case. Behaviorally the re-attempt is harmless (arguably good), so this is just message accuracy — consider softening to e.g. `until this worker restarts`.

—
_Generated by Barber AI_

kriszyp:
## Review feedback addressed (dispatch worker)

Pushed 6 commits addressing all 5 open review threads:

- **Gemini — stale budget on create-path completion**: fixed (`47464b9`, already on origin) — the reconcile now returns a table's budget whenever it observes a live (non-tombstoned) row.
- **kriszyp — budget must be scoped to a drop generation, not just table**: fixed (`dc17d52c`, already on origin) — `Table.ts` stamps a random `dropGeneration` on every tombstone; the budget keys off it.
- **kriszyp — unawaited LMDB `drop()`/`remove()` bypass retry accounting**: fixed (`f995535`/`dc17d52c`, already on origin) — switched to `dropSync`/`removeSync`.
- **kriszyp — per-worker/per-alias amplification (up to 72 completion passes / 24 errors in the four-alias/six-worker case)**: fixed in `45e593ee8`. The worker-0 log gate (already on origin) handled the per-worker half; this handles the per-alias half. Harper lets multiple database names alias the same physical directory (exactly the scenario described), and `initStores()`/the create path already receive that physical path as a parameter — so keying the retry budget by path instead of database name collapses the budget and the give-up log to one per physical table, with no cross-worker/main-thread coordination needed (simpler than the originally-suggested approach). Test at `dropTableGhost.test.js` ("bounds the retries...") now asserts exactly one give-up log total across all four aliases that `setupTestDBPath()` configures onto one physical path.
- **cb1kenobi — "until this node restarts" overstates permanence (a worker recycle also resumes retries)**: fixed (`dc17d52c`, already on origin) — reworded to "until this worker restarts (a full node restart also resets this)".

Additionally ran 4 rounds of independent pre-push review (`prepush-review.mjs`, Codex leg) on the new work and fixed what it found: the primary tombstone could be removed before the rest of catalog cleanup finished (now removed last), a retry-budget-map leak on 2 of 3 resolution paths (now swept consistently), and a scan-cost issue in that leak fix (now a nested map so sweeps are O(1) regardless of map state). Added a regression that fails without the tombstone-ordering fix.

One finding recurred across all 4 review rounds and remains open: `completeInterruptedDrop()`'s destructive cleanup isn't re-validated against a lock/generation check before running, so a sufficiently unlucky cross-worker interleaving (worker scans generation A's tombstone; another worker resolves + recreates + re-drops as generation B before this worker's cleanup runs) could drop B's live stores. Confirmed via git history this predates this entire PR (present at the branch's merge-base) — not something introduced here, and in the same spirit as this PR's own stated scope ("fixes only the log flood... deliberately not fixed here"). Flagging for a follow-up issue rather than folding a locking redesign into this PR.

Also not fixed: RocksDB-mode coverage for a genuine reconcile-time store-drop failure (the existing regression is LMDB-only — no test seam currently exists to inject a `dropSync` failure into the RocksDB branch's private `openRocksDatabase` call without more invasive refactoring).

— Claude Sonnet 5

Ethan-Arrowood (approved):
Retry bound is well-tested and all prior threads are addressed; everything sits on the cold reconcile/drop path.

Issue #1937, opened 2026-07-24
alasql's optional react-native-fs dep pulls ~140M of react-native/hermes/metro into installs and published Docker images


heskew:
## Summary

`alasql` (4.17.x) declares `optionalDependencies: { "react-native-fs": "^2.20.0" }`. `react-native-fs` itself is 828K and unused in Node (alasql only requires it behind a guard for RN environments), but it has a **peer dependency on `react-native`** — and npm 7+ auto-installs peer deps. So every plain `npm install harper` pulls in `react-native@0.82`, `react@19`, hermes, metro, and react-devtools.

## Confirmed in published Docker images

`harperfast/harper:latest` (5.1.23) ships:

| package | size |
|---|---|
| `hermes-compiler` | 47M |
| `react-native` | 37M |
| `@react-native/*` | 25M |
| `react-devtools-core` | 17M |
| `metro*` | ~9M |
| `react-native-fs`, `react`, `hermes-parser`, misc | ~4M |

**~140M of the 373M node_modules (~38%) is the react-native tree.** The Dockerfile installs the packed tarball with plain `npm install --ignore-scripts --global`, which includes optional deps and auto-installed peers.

Not Docker-specific: every npm install of harper 5.x gets the same ~140M. (The legacy `harperdb/harperdb` 4.x images are unaffected only because they ship alasql 4.6.6, which predates the optional dep.)

Beyond size, this puts metro/hermes/react-devtools in a production database image — pure CVE-scanner noise and supply-chain surface for code that can never execute.

## Fix

**Do not** use `--omit=optional` in the Dockerfile — harper's own tree relies on optionalDependencies for the `@harperfast/rocksdb-js-*` platform prebuilds, cbor/msgpackr extractors, `bufferutil`, `utf-8-validate`, etc.

Recommended: add an npm override in `package.json`, next to the existing `alasql → xlsx` override, stubbing `react-native-fs` out with a tiny empty package (e.g. `noop2` or an internal empty stub):

```json
"overrides": {
  "alasql": {
    "xlsx": "0.18.5",
    "react-native-fs": "npm:noop2@^2.0.0"
  }
}
```

This kills the chain at the root for Docker, npm, and dev installs alike.

Longer term: upstream a `peerDependenciesMeta: { "react-native": { "optional": true } }` fix to react-native-fs, and/or ask alasql to drop the optional dep.

tps-flint:
[...]

PR #1958, opened 2026-07-27
fix(deps): stop alasql's react-native-fs dep pulling ~140M of react-native into installs

heskew:
Fixes the dependency bloat reported in #1937. The issue is real; the recommended fix in it turned out to be **necessary but not sufficient**, so this PR carries two changes.

## Confirmed

`alasql@4.17.3` declares `optionalDependencies: { "react-native-fs": "^2.20.0" }`. `react-native-fs` peer-depends on `react-native` (**not** marked optional), and npm 7+ auto-installs peer deps — so `react-native`, `react`, `hermes`, `metro` and `react-devtools-core` land in every install.

All six `require('react-native-fs')` calls in `alasql/dist/alasql.fs.js` sit behind `utils.isReactNative`, so none of it can execute under Node. Harper only uses `alasql.parse` plus its function extensions.

The published `harper@5.1.23` tarball's `npm-shrinkwrap.json` pins **794** packages, **64** of them react-native/hermes/metro — that's how it reaches the images.

## Why two fixes

`overrides` alone does not fix the published image. Verified against npm 11.6.2:

| install shape | `overrides` honoured? | bundled shrinkwrap honoured? |
|---|---|---|
| harper as root project (dev, CI) | yes | n/a |
| `npm install --global harper-*.tgz` (Dockerfile) | no | **no** |
| harper as a dependency of another project | no | **no** |

npm honours `overrides` only for the root project, and it ignores a published `npm-shrinkwrap.json` when installing from a tarball — the tree is re-resolved from harper's `package.json` instead. I confirmed this with a *valid, unpruned* 959-entry shrinkwrap in the tarball: still 741 packages and the full react-native tree. So `build.sh`'s `prune-shrinkwrap-dev.mjs` is not at fault.

### 1. `package.json` override — root installs

```json
"overrides": { "alasql": { "xlsx": "0.18.5", "react-native-fs": "npm:noop2@^2.0.0" } }
```

- `node_modules` **682M → 514M**
- lockfile **1202 → 959** entries (204 packages removed)
- `npm ci` clean, `npm ls` reports no missing/invalid deps
- 126 `sqlEngine` + `sqlTranslator` unit tests pass; `npm run build` clean

npm will not re-resolve an already-pinned entry just because an override was added, so the lockfile was regenerated by deleting the `react-native-fs` node and letting npm resolve that edge again.

### 2. Dockerfile prune — the published image

Removes the react-native closure after the global install, guarded by `cd &&` so a bad path prunes nothing rather than deleting the wrong tree. Verified with a real `docker build`:

- harper's `node_modules` **464M → 326M** (138M, matching the ~140M in the issue)
- asserts the tree is gone, `alasql` still parses **and** executes, `harper` bin still resolves

## Lockfile change is resolution-only

Comparing `name@version` multisets against main: the **only** addition is the `react-native-fs` alias, and **no existing package changes version**. The rest of the diff is npm re-hoisting packages that the react-native tree's nested copies had been blocking.

`bufferutil` / `utf-8-validate` are deliberately preserved as they are on main — see the first follow-up below.

## Follow-ups (not in this PR)

1. **`npm install harper` still gets the react-native tree.** Nothing in harper's `package.json` can prevent it, for the reasons in the table. The one mechanism that does work at dependency depth is a real top-level dependency alias whose version satisfies alasql's `^2.20.0` — proven with a throwaway alias (**331 → 24** packages). That needs a tiny stub published (e.g. `@harperfast/react-native-fs-stub@2.20.0`), which is a call for the team. Upstreaming `peerDependenciesMeta: { "react-native": { optional: true } }` to react-native-fs, or getting alasql to drop the optional dep, would fix it at the source.

2. **Unrelated latent regression on main:** `@datadog/pprof` requires `node-gyp-build <4.0`, pinning 3.9.0 at the top level, while `bufferutil` / `utf-8-validate` need `^4.3.0`. npm resolves that conflict by nesting for *required* deps (`argon2` gets 4.8.4) but by **silently dropping** *optional* ones — so a fresh `npm install` on current main loses both ws native accelerators. The committed lockfile predates `pprof` and still lists them, so it is latent, but the next release build would drop them from the published shrinkwrap (5.1.23 still has both). Worth its own issue.

3. The `alasql → xlsx: "0.18.5"` override is now a no-op — `alasql@4.17.3` no longer depends on `xlsx` at all. Left in place as a guard.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,6 +54,18 @@ VOLUME /home/harperdb/harper
 RUN <<-EOF
   npm install --ignore-scripts --global harper-*.tgz
   rm harper-*.tgz
+  # alasql declares an optional dependency on react-native-fs, which peer-depends on
+  # react-native. npm auto-installs peer deps, so a plain install drags ~130M of
+  # react-native/metro/hermes/react-devtools in here. alasql only requires
+  # react-native-fs behind an isReactNative guard, so none of it can execute under Node.
+  # The root override in package.json cannot reach this install: npm honours overrides
+  # only for the root project, and ignores a published npm-shrinkwrap.json when
+  # installing from a tarball. So the tree gets pruned after the fact.
+  # The `cd &&` guard means a bad path prunes nothing rather than deleting the wrong tree.
+  ( cd /home/harperdb/.npm-global/lib/node_modules/harper/node_modules && rm -rf \
+      react react-devtools-core react-is react-native react-native-fs react-refresh \
+      scheduler hermes-compiler hermes-estree hermes-parser metro metro-* ob1 \
+      babel-plugin-syntax-hermes-parser @react-native )
   mkdir -p /home/harperdb/harper
   chown harperdb:harperdb /home/harperdb/harper
 EOF
--- a/package-lock.json
+++ b/package-lock.json
@@ -191,6 +191,7 @@
 			"resolved": "https://registry.npmjs.org/@aws-sdk/client-s3/-/client-s3-3.1090.0.tgz",
 			"integrity": "sha512-R6GX9cd1jljwzZ8xFmgAI/hHCuX1MobIKBdsymv7WL9SENvO9Vgz9KOR6avTnu0Ao+w1LmxnTe+jqmZXEn7Q/Q==",
 			"license": "Apache-2.0",
+			"peer": true,
 			"dependencies": {
 				"@aws-sdk/checksums": "^3.1000.18",
 				"@aws-sdk/core": "^3.975.3",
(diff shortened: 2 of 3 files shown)


heskew:
Superseded by #1959, which takes a different approach.This PR used an `overrides` entry plus a Dockerfile prune. Two findings while verifying it made that the wrong shape:1. **npm ignores a published `npm-shrinkwrap.json` only for *tarball* installs, not registry installs.** I originally concluded it was ignored outright, but that was an artifact of testing with a local tarball. `npm install harper` from the registry honours the shrinkwrap exactly — so pruning the react-native tree there fixes consumers with no `package.json` change, no stub package, and no third-party empty module. #1959 does that.
2. **The Dockerfile prune was a hardcoded list of ~15 directory names** that would silently rot as alasql's tree shifts. The image needs a different fix — its install resolves fresh from `package.json`, so it is also not reproducible. That is a follow-up stacked on #1959.Branch `fix/alasql-react-native-fs-bloat` left in place for reference.

---

PR #1959, opened 2026-07-27
fix(build): strip the react-native tree from the published shrinkwrap

### LINKED ISSUE #1962: Prune the react-native tree from the published shrinkwrap
The registry-install half of #1937.`alasql`'s optional `react-native-fs` dependency peer-depends on `react-native`, which npm auto-installs — ~140MB and ~200 packages of code unreachable under Node. The published 5.1.23 shrinkwrap pins 794 packages, 64 of them react-native/hermes/metro.Strip the subtree from the shrinkwrap that ships in the published package, alongside the existing dev prune (#1783). npm honours `overrides` only for the root project, so they cannot help a consumer; the published shrinkwrap *is* authoritative for registry installs, which makes it the lever that actually reaches users — no stub package or third-party empty module needed.Covers `npm install harper` from the registry. The Docker image is **not** covered: it installs from a local tarball, which has no packument, so npm never learns the shrinkwrap exists. That is #1960.Fixed by #1959.

---

heskew:
Removes the react-native tree from what harper publishes. Part 1 of #1937.## What and why`alasql` declares `optionalDependencies: { "react-native-fs": "^2.20.0" }`, and `react-native-fs` peer-depends on `react-native` **without** marking it optional. npm 7+ auto-installs peer deps, so resolving alasql drags in `react-native`, `react`, `hermes`, `metro` and `react-devtools-core` — ~140MB, ~200 packages. None of it can execute: every `require('react-native-fs')` in `alasql/dist/alasql.fs.js` sits behind an `isReactNative` guard, and Harper only uses `alasql.parse` plus its function extensions. The published 5.1.23 shrinkwrap pins 794 packages, 64 of them react-native/hermes/metro.This runs alongside `prune-shrinkwrap-dev.mjs`, which already enforces the same invariant — the published shrinkwrap should describe only the production tree a consumer installs (#1783).## Where to look**`build-tools/prune-shrinkwrap-react-native.mjs`** is the whole change; the rest is a build step, a doc entry, and tests.The reachability logic is what deserves scrutiny. It computes the set of packages reachable from the root *with* the `react-native-fs` edge and *without* it, and deletes only the difference — so a package still reachable by any other path cannot be removed, and the removed set is derived rather than a hardcoded list of directory names that would silently rot as alasql's tree shifts. `resolve()` reimplements node_modules lookup (check `<from>/node_modules/<name>`, then each ancestor); that's the part most likely to be subtly wrong, though a mistake there fails safe — an unresolvable name simply means nothing gets marked reachable through it.On the current tree: 419 dev entries pruned, then 262 react-native entries, 1202 → 521. Zero unresolved *required* edges afterwards; the only new unresolved optional edge is `alasql -> react-native-fs`, the one being severed. (The published 5.1.23 shrinkwrap already carries 11 such unresolved optional/peer edges, so this is a pre-existing, normal pattern.)## Why not `overrides`npm honours `overrides` only for the **root** project — an override in harper's `package.json` does nothing for anyone installing harper. The published shrinkwrap, by contrast, *is* authoritative for registry installs.I verified that rather than assuming it, because the premise matters and one plausible failure mode would have sunk it: npm could have re-resolved the pruned *optional* dependency and pulled the subtree back. Serving a package with a pruned shrinkwrap from a local static registry (so npm takes the real packument + `_hasShrinkwrap` code path, with the dependency still resolving from npmjs):```
shrinkwrap before prune: 331 entries, react-native present: yes
shrinkwrap after  prune:  23 entries, react-native present: no
installed: 23 packages
react-native tree in consumer tree: NO
react-native-fs node: ABSENT
```Consequence: no stub package, no third-party empty module, and no change to `package.json` or the committed lockfile.## Scope — what this does not fix- **The Docker image.** `npm install <tarball>` has no packument, so npm never learns the shrinkwrap exists and resolves the tree fresh from `package.json`. Confirmed on the *same* published 5.1.23 artifact: installed via the registry it honours the pin (fastify 5.8.5), installed from the tarball it resolves fresh (fastify 5.10.0, latest) and the react-native tree returns. The image is therefore also not reproducible — two builds of one commit weeks apart can ship different trees. Follow-up PR, which builds on this one; it has to land after this, since honouring an unpruned shrinkwrap would pin the react-native tree *in*.
- **Contributor / CI installs**, which use `package-lock.json` and are unaffected. Deliberately left alone: fixing it means either a stub package or a cryptic override, and it costs disk rather than shipped bytes.
- **The root cause.** Worth reporting upstream to alasql (active repo, no existing issue found) — RN-only deps are normally lazy-required without being declared. When that lands, this script gets deleted; it reports "nothing to prune" rather than failing, so it degrades quietly.## Testing`unitTests/buildTools/pruneShrinkwrapReactNative.test.js` — 6 tests, the load-bearing one being that a package pulled in by react-native *and* depended on elsewhere survives. Note the sibling `prune-shrinkwrap-dev.mjs` has no tests; I added these because a wrong prune ships a broken shrinkwrap to every consumer.`test:unit:main`: 2795 passing, 2 failing — `globalIsolation` and a uWS 413 case. Both reproduce identically on pristine `origin/main` (12/1 and 6/1 in isolation), and this diff only touches `build-tools/`, `unitTests/buildTools/` and `dependencies.md`.## DocsNo companion documentation PR: this changes no public API, config option, schema directive, CLI surface or user-visible default — it only removes dead weight from what gets published. The behavioural note for future work lives in `DESIGN.md` and `dependencies.md`.Closes #1962

---

### CODE REVIEW

File: unitTests/buildTools/pruneShrinkwrapReactNative.test.js [hunk 1]
```diff
@@ -0,0 +1,122 @@
...
+		const lock = fixture();
+		lock.packages[''].dependencies['other-pkg'] = '^1.0.0';
+		lock.packages['node_modules/other-pkg'] = {
+			version: '1.0.0',
+			optionalDependencies: { 'react-native-fs': '^2.20.0' },
+		};
+		const { result } = runPrune(lock);
+		assert.ok(!result.packages['node_modules/react-native-fs'], 'both declarations are optional, so it goes');
+		assert.ok(result.packages['node_modules/other-pkg'], 'the optional dependent itself survives');
+	});
```
cb1kenobi, line 105: **Nit: still no test for a nested `react-native-fs` edge**

All fixtures here declare packages flat under `node_modules/<name>` at the root, so `resolve()`'s ancestor-walk (`build-tools/prune-shrinkwrap-react-native.mjs:50-58`, which steps up through `<from>`'s `/node_modules/` segments) only ever resolves at `depth === 0`. A case where `react-native-fs` (or one of its dependents) is nested under another package's own `node_modules` — forcing `resolve()` to walk up through an ancestor segment to find it — is still unexercised. That's the part of the script most likely to be subtly wrong per the PR description, and the two new tests added in this round (severable vs. hard-depend edges) don't cover it either.

Suggested fix: add a fixture where, e.g., `node_modules/other-pkg/node_modules/react-native-fs` shadows or is the only resolution path, so a miss in the ancestor-walk fails the test.

—
_Generated by Barber AI_

heskew: Fair, and only partly overtaken by events — addressed in 046869594.Timing: a nested/scoped fixture landed in 8ea039b38 about ten minutes after your review, so you were looking at cba1b63be. That commit made `resolve()` succeed at `depth === 1` for the first time (a nested `shared` shadowing the hoisted copy), so the literal "only ever resolves at depth 0" no longer held. But your actual suggestion did still stand, so I checked rather than assuming:```
node_modules/react-native                     -> shared   depth=1  (nested-first: covered)
node_modules/react-native                     -> rn-only  depth=0  (one-level walk-up)
node_modules/react-native/node_modules/shared -> alasql   depth=0  (two-segment walk-up: NOT reachable,
                                                                    that nested copy had no deps)
```So nested-first was covered; a react-native-fs reachable *only* as a nested copy, and a multi-segment walk-up, were not. Both are now, in a single fixture matching what you suggested: `node_modules/other-pkg/node_modules/react-native-fs` as other-pkg's only resolution path, and a `nested-rn-dep` at the root that the nested copy depends on — reached only by missing twice and landing at depth 0. Resolving either wrongly leaves the entry unreachable, so it is never pruned and the assertions fail.Checked the coverage is real rather than incidental by mutating `resolve()` to clamp the loop to the deepest segment: 8 tests fail. Restored, 13 pass. Real pipeline unchanged — 262 entries pruned, zero unresolved required edges.

---

PR #1963, opened 2026-07-27
fix(audit): give audit cleanup a real completion signal (deflake auditLog prune test)

kriszyp:
## What

The `Audit log > check log after writes and prune` unit test is the largest single contributor to the `Unit Test` workflow flaking on `main` — 6 of the 18 real failures between 2026-07-21 and 2026-07-27. It fails in two variants, and both are the same shape: the test guesses a wall-clock budget for asynchronous work it has no way to observe.

Fixing the second variant properly meant giving the audit-cleanup loop a completion signal, which in turn surfaced several genuine defects in that loop.

### Variant 1 — `auditLog.test.js:56`, "Should have at least a couple of update events" (4 occurrences, rocksdb `test:unit:resources`)

```js
await delay(20);
assert(events.length > 2, 'Should have at least a couple of update events');
```

Subscription events are delivered off the commit path. Nothing orders them against this assertion, so the 20 ms sleep is a bet. It now waits for the events via the repo's `waitFor` helper — same assertion, no wall-clock bet.

### Variant 2 — `auditLog.test.js:71`, `4 == 0` (2 occurrences, lmdb `test:unit:lmdb`)

The test scheduled an audit prune and then polled for 20 × 10 ms hoping it had landed. The prune runs on a self-rescheduling background timer inside `openAuditStore`'s closure; the test had no way to observe it, so the assertion was betting on a 200 ms budget covering `schedule → timer → one setImmediate yield per deleted entry → async commit`, on a shared audit store that every table in the database writes to.

`scheduleAuditCleanup()` now returns a promise that fulfills when the pass serving the call has finished, and the test awaits that instead of polling.

### Product defects found while making that promise trustworthy

All in `resources/auditStore.ts`, all pre-existing:

1. **The `closed`/`closing` early return wedges the loop permanently.** It returned without resolving `resolution` — *after* `resolution` had been stored in `lastCleanupResolution`. Every later pass awaits that variable, so one transient `closing` observation deadlocks audit cleanup for the life of the store, and the early return does not reschedule either.
2. **Concurrent passes over the same range.** `lastCleanupResolution = resolution` was assigned *after* `await lastCleanupResolution`, so every pass released by the same resolution proceeded together. The slot is now claimed before yielding.
3. **Superseded pending passes strand their callers.** A `scheduleAuditCleanup()` that cancels a not-yet-started pass left that pass's promise unsettled forever. Superseded callers are handed to the replacement.
4. **The backoff collapses to a hot loop.** `auditCleanupDelay << 1` coerces to int32, so a sub-millisecond retention pins the delay at 0 permanently (`0 << 1` is `0`) — and it stays 0 even after retention returns to normal. Measured over one `test:unit:resources` run: **129 sub-millisecond reschedules before, 0 after**. The sibling `>> 1` has the mirror-image bug (a retention over ~24.8 days wraps negative when halved), and `setTimeout` silently falls back to 1 ms past `2^31-1`, which a retention over ~248 days reaches via `retention / 10`. All three now use plain arithmetic with a clamp.
5. **The pass could escape as an unhandled rejection.** A rejected commit or a throw out of the range iteration left the detached timer callback rejecting. It is logged now.

## What this does *not* claim

I could not reproduce variant 2 locally. I tried an event-loop hog (40 ms blocked out of every 45 ms), a 200-entry audit backlog on the shared store, and 25 repetitions per engine — all green on the pre-fix code. So the *mechanism* behind `4 == 0` (which entry survived, and why 200 ms was not enough that day) is not proven at the level variant 1 is. What is certain is that the old assertion had no completion signal to wait on and the new one does; and the four loop defects above are real regardless.

The new form still fails if the prune genuinely does not happen — it times out at 10 s with `audit log was not pruned` instead of asserting at 200 ms — so this does not convert a product failure into a pass.

## Also worth knowing: signature D was already fixed

The dispatch that produced this work also listed an apitests signature (`RESTProperties-test.mjs:243`/`:314`, `cache-test.mjs:111`, 404s on `SimpleCache/3` and `CacheOfHttp/direct-fetch`). All six occurrences are dated 2026-07-21/22 and stop at 6e7630c5a (2026-07-23), which added the missing `await` to the four `Table.clear()` calls in `setupTestApp()`. The 404s and the `JSON.parse(...) === null` are exactly a wiped `FourProp` — a late clear landing on the freshly written seed records. No recurrence in the ~9 green full-matrix runs since. Nothing to do here.

## Test notes

- `unitTests/resources/auditLog.test.js`: 25× rocksdb + 25× lmdb, all green; 10× lmdb under a synthetic event-loop hog, all green.
- Full `test:unit:resources`: 4× rocksdb (1263 passing) and 8× lmdb (1126 passing) green. One lmdb run failed `dropTableGhost.test.js:70` — reproduced 1-in-5 on the **unpatched** baseline too, so it is a separate pre-existing flake, filed as a finding.
- `test:unit:main`: 3801 passing.
- `test:unit:apitests` cannot fully run in this environment (no local `harper install`, so the operations/MQTT listeners do not bind); 155 passing / 10 failing identically with and without this change, i.e. neutral.

## Review

Cross-model reviewed (Gemini via `agy` + Codex, adjudicated inline — the `reviewer`/`codex-reviewer` subagents are not available to dispatched workers). Both outside legs found real issues in the first commit; the second commit is those fixes. No unresolved blockers.

Generated by Claude Opus 5.

### DIFF
--- a/resources/auditStore.ts
+++ b/resources/auditStore.ts
@@ -108,6 +108,9 @@ export const AUDIT_STORE_OPTIONS = {
 
 export let auditRetention = convertToMS(envGet(CONFIG_PARAMS.LOGGING_AUDITRETENTION)) || 86400 * 1000;
 const MAX_DELETES_PER_CLEANUP = 1000;
+// setTimeout silently falls back to 1ms for delays past this, which would turn the backoff into the
+// hot loop it is meant to avoid — a `logging.auditRetention` over ~248 days reaches it via retention/10
+const MAX_CLEANUP_DELAY = 2 ** 31 - 1;
 const FLOAT_TARGET = new Float64Array(1);
 const FLOAT_BUFFER = new Uint8Array(FLOAT_TARGET.buffer);
 let DEFAULT_AUDIT_CLEANUP_DELAY = 10000; // default delay of 10 seconds
@@ -165,10 +168,12 @@ export function openAuditStore(rootStore) {
 		}
 	});
 	/**
-	 * Schedules a pass of the audit cleanup loop. The returned promise settles once the pass that
-	 * serves this call has committed its deletions, so callers (notably tests) can await completion
-	 * instead of guessing a delay. A pass removes at most MAX_DELETES_PER_CLEANUP entries before
-	 * rescheduling itself, so settling means "that pass finished", not "the audit log is fully pruned".
+	 * Schedules a pass of the audit cleanup loop. The returned promise fulfills once the pass that
+	 * serves this call has finished, so callers (notably tests) can await completion instead of
+	 * guessing a delay. Two things it does not promise: a pass removes at most
+	 * MAX_DELETES_PER_CLEANUP entries before rescheduling itself, so fulfillment means "that pass
+	 * finished", not "the audit log is fully pruned"; and a pass that failed logs and fulfills rather
+	 * than rejecting, since this promise doubles as the loop's serialization barrier.
 	 */
 	function scheduleAuditCleanup(newCleanupDelay?: number): Promise<void> {
 		// Skip audit cleanup/purge in read-only mode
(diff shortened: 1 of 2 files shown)


resources/auditStore.ts
```diff
@@ -202,22 +228,30 @@ export function openAuditStore(rootStore) {
...
+						auditCleanupDelay = Math.max(1, Math.min(auditCleanupDelay * 2, auditRetention / 10, MAX_CLEANUP_DELAY));
 					} else {
 						// if we did delete something, update our updates since timestamp
 						updateLastRemoved(auditStore, lastKey);
 						// and do updates faster
-						if (auditCleanupDelay > 100) auditCleanupDelay = auditCleanupDelay >> 1;
+						if (auditCleanupDelay > 100) auditCleanupDelay = auditCleanupDelay / 2;
 					}
-					resolve(undefined);
+					resolve();
```
cb1kenobi: **Low: `resolve()` is not guaranteed on all finally paths, re-opening the wedge this PR closes**

The `finally` body runs `updateLastRemoved(auditStore, lastKey)` (line 244) *before* `resolve()`. `updateLastRemoved` calls `auditStore.put(...)`, which can throw synchronously if the store transitions to `closing`/`closed` mid-pass (there are `await` points between the top-of-callback status check and here). A throw here is not caught — the surrounding `try/catch` does not cover its own `finally` — so `resolve()` is skipped, `resolution` never settles, and the next pass's `await previousCleanup` (the serialization barrier) deadlocks the loop for the life of the store, while the detached timer callback rejects unhandled. That is exactly the pair of failure modes (permanent wedge + unhandled rejection) this PR set out to eliminate, just relocated from the removed early-return into `finally`.

Narrow trigger (store closing during an active pass that deleted ≥1 entry, i.e. shutdown), hence Low, but it undercuts the invariant the PR advertises: `resolution` must settle on every path. Simplest enforcement is to make `resolve()` unconditionally reachable, e.g. resolve first, then do the bookkeeping:

```suggestion
					resolve();
```

(moving the `resolve()` call to the top of the `finally` block, ahead of the delay/`updateLastRemoved` bookkeeping, so a throw in that bookkeeping can no longer strand the promise or wedge the barrier).

—
_Generated by Barber AI_

Ethan-Arrowood (approved):
The completion signal is sound and this fixes several real defects in the prune loop beyond the flake itself. Suggestions in my notes are non-blocking.

PR #1967, opened 2026-07-28
Clarify open-transaction timeout error message

cb1kenobi:
## Summary

The error surfaced when a transaction is aborted for exceeding `STORAGE_MAXTRANSACTIONOPENTIME` read:

> Transaction was aborted after exceeding the open-transaction limit; split long-running work into smaller transactions

"open-transaction limit" is ambiguous — it reads just as naturally as a cap on the **number** of open transactions as it does the **duration** a transaction may stay open. That ambiguity is more than cosmetic here, because `DatabaseTransaction` has *both* kinds of limit:

- a **count/queue-depth** limit — the `"Outstanding write transactions have too long of queue"` 503 overload path, and
- a **duration** limit — this timeout, governed by `STORAGE_MAXTRANSACTIONOPENTIME`.

An operator hitting this error could reasonably start chasing concurrency/queue depth when the real cause is elapsed time. This names the time dimension so the cause is unmistakable, matching the sibling log line ("Transaction was open too long…") and the config parameter's own name.

## Change

Message only — no behavior change:

> Transaction was aborted after exceeding the **maximum open-transaction time**; split long-running work into smaller transactions


kriszyp (approved):
Nice

Issue #1968, opened 2026-07-28
Two writes to the same record in one transaction orphan the intermediate secondary-index entry


kriszyp:
Two writes to the same record within a single transaction leave the intermediate value's secondary-index entry orphaned. `search()` then returns that record under a value its attribute no longer holds, permanently — nothing reconciles the index against the records, and a later write can only remove the entry if the record's *previous* value happens to be the orphaned one.

Found while investigating stale `status`-index entries on the receiving side of replication in a deployment running the 4.7 line. The mechanism reproduced here is a v5 regression: 4.7.36 handles both shapes below without orphaning.

## Reproduction

Two probes, both against a table with one indexed attribute. Full test files at the bottom.

**A. Source-applied (replicated) transaction carrying two updates to the same record.** The sender frames one message per distinct `txnTime`, so if the origin wrote a record twice in one transaction, the replica applies both writes inside a single transaction.

```
record: STOPPED   |   indexed under: STOPPING, STOPPED
```

**B. Request path — two `patch()` calls to the same record inside one `transaction()`.**

```
record: STOPPED   |   indexed under: STOPPING, STOPPED
```

| scenario | 4.7.36 | v5 RocksDB | v5 LMDB |
|---|---|---|---|
| replicated transitions applied back-to-back, one txn each | clean | clean | clean |
| **A** — two writes to same record, one replicated txn | clean\* | **orphan** | **orphan** |
| **B** — two writes to same record, one request txn | clean | clean | **orphan** |

\* 4.7 does not orphan in A because it *drops* the second write as a duplicate timestamp — the record ends `STOPPING`, not `STOPPED`. It is not correct either; v5 applies both writes (better) and thereby exposes the index-diff bug (worse).

## Cause

`updateIndices(id, existingRecord, record)` in `resources/Table.ts` computes what to remove from the index purely as `existingRecord[attr]`. Each write's `existingRecord` comes from `operation.entry`, and `DatabaseTransaction.save()` only re-reads that entry when `reloadEntry` is set:

```js
if (reloadEntry || operation.entry === undefined) {
	operation.entry = operation.store.getEntry(operation.key, { transaction });
}
```

[...]

PR #1970, opened 2026-07-28
fix(txn): apply repeat writes to a key on top of the earlier write in the same transaction

kriszyp:
## What

A transaction that contains two writes to the same record key now applies the second write on top of what the first one staged, instead of both diffing against the pre-transaction record.

Before this, the secondary index for a changed attribute kept the intermediate value: the second write re-removed the value the first write had already removed, and never removed the value the first write added. Nothing reconciles a secondary index against the records, so the orphaned entry was permanent until an index rebuild — a status-filtered search kept returning records whose status no longer matched. The same stale basis also dropped the earlier write's changes from the record itself (`patch({a:1})` then `patch({b:2})` in one transaction lost `a`) — on LMDB for the request path, and on **both engines** for the source-applied path: a replicated transaction carrying two patches to one record silently dropped the first patch's fields on the replica, i.e. silent divergence from the origin, which is a bigger deal than the index orphan.

Reproduces on LMDB for the request path (two `patch()` calls in one `transaction()`) and on **both** engines for the source-applied/replication path (one replicated transaction carrying two updates to a record). It is a v5 regression: 4.7 avoided the replicated shape only by dropping the second write as a duplicate timestamp, which left the record on the intermediate value — that behavior is deliberately *not* restored here; both writes still apply.

Fixes the index-orphan class described in #1968.

## Why this shape

The direct alternative — reload the write's entry so it reads-its-own-writes — cannot work uniformly: LMDB queues staged puts and applies them only in the commit batch (a `getEntry` inside that loop still returns the pre-transaction record; forcing the exclusive `store.transaction()` fallback is worse — it drops the second write), and RocksDB read-your-writes only covers writes already staged into the native transaction, which the source-apply path hasn't done when it stages a batch before `commit()`. So the writes carry the state forward themselves.

The other option in the issue — coalescing the writes to one — was rejected: it changes audit and replication semantics (one entry instead of two, intermediate state erased) for a problem that is really about a write's diff basis.

- `DatabaseTransaction.linkWrite()` chains each staged write to the preceding write to the same store and key (called from both engines' `addWrite`); `clearWrites()` drops the chain with the write set so a reused transaction can't inherit a previous batch's state.
- The `Table` put/delete commit handlers publish the record they stored on `write.stagedEntry` and read the prior one back through `priorStagedEntry()` as `existingRecord` — the basis for the index diff and the incremental merge.
- A chained write is treated as **in order by construction** (`precedesExisting = 1`; the delete path skips its newer-record guard the same way). Every write in a transaction carries the transaction's single timestamp, so a version comparison against what the earlier write staged is a tie, and the out-of-order resequencing would drop the later write as a re-delivered duplicate — exactly the 4.7 behavior this must not reintroduce.

Only the *record* comes from the earlier write. Version, audit chain and blob metadata stay the pre-transaction entry's, which is what this write's audit entry and optimistic version check are relative to. `DESIGN.md` records the invariant and its consequences.

## Tests

Two regression tests from the issue, both engines:

- `unitTests/resources/indexOrphanSameKeyTxn.test.js` — request path. Fails pre-fix on LMDB (index orphan, dropped incremental update, delete-then-patch resurrecting the deleted record); also covers a delete after a write and the separate-transaction control.
- `unitTests/resources/indexOrphanReplicaApply.test.js` — source-applied path. The "two updates in ONE replicated transaction" case fails pre-fix on **both** engines; the pipelined and serialized apply cases are controls. Converted the fixed sleeps to `waitFor` (AGENTS.md), which also took the file from ~750ms to ~135ms.

`test:unit:resources` and `test:unit:main` green on RocksDB and `HARPER_STORAGE_ENGINE=lmdb`; the 10 `unitTests/components/globalIsolation.test.js` failures in `test:unit:main` are pre-existing (identical with and without this change). Integration: `integrationTests/database` (228 pass), `integrationTests/resources` (94 pass), `integrationTests/server` (only the Ollama-backend tests fail — no local Ollama).

## Review notes / open concerns

- **Audit log still gets two entries with the same version** for two writes to the same key in one transaction, and the second entry's previous-version link points at its sibling on a RocksDB retry round (where the reloaded entry is the sibling's). That predates this change and is inherent to keeping both writes; the issue calls it out as the argument for coalescing. Flagged rather than fixed. Devin's review adds a concrete residual on the same root: a chained write that skipped in round 1 while its sibling appended can, on a retry round in out-of-order state, find the sibling's durable audit entry via the (version, nodeId, id) dedup tuple and drop itself as already-applied. Deferred with the same-version design question.
- **Memory**: the per-key chain retains one Map entry per write for the transaction's lifetime, so a bulk-load transaction of N writes carries N extra map entries.
- **Review round (1b9518c64)**: both bots' "critical" findings on the conflict guards were confirmed and adopted — program order now breaks version ties only (`>= 0`), so a strictly newer concurrent write observed on a retry round still goes through the out-of-order merge, and a chained delete yields to it. The PR's earlier dismissal of these was wrong (it addressed a narrower reading). Also from review: `writeKeyId` is now namespaced/type-tagged (the scalar `'["x"]'` vs composite `['x']` collision cross-contaminated records — regression added), and a write whose record a later same-key write replaced gets its `savedBlobs` cleaned post-commit unless its audit entry references them (fixes a permanent LMDB blob-file leak — regression added, fails pre-fix on LMDB only since RocksDB's read-your-writes path already cleaned it).

Generated by Claude Opus 5.


kriszyp:
Empirical basis-record matrix behind this fix, from an instrumented probe (`updateIndices` basis + `save()` entry capture) run across engines and branches. "basis" = what the second write to the key diffed against; ✗ = pre-transaction record (stale → index orphan, and field loss for patches).

| shape | 4.7.36 | main RocksDB | main LMDB | this PR (both engines) |
|---|---|---|---|---|
| `put(A); put(B)` awaited, one txn | ✗ orphan | ✓ saw A | ✗ orphan | ✓ |
| `patch(A); patch(B)` awaited, one txn | ✓ (see below) | ✓ saw A | ✗ orphan + field loss | ✓ |
| `patch` × 2 concurrent, one txn | ✓ | ✓ (ordering luck, not guaranteed) | ✗ | ✓ |
| replicated txn carrying 2 updates to one key | ✗ orphan or dropped write (varies) | ✗ orphan + field loss | ✗ orphan + field loss | ✓ |
| replicated, 2 separate txns | ✓ | ✓ | ✓ | ✓ |
| 2 resources fetched up front, both written | ✗ orphan | ✗ orphan | ✗ orphan | ✓ |

Notes:

- The discriminator is not `reloadEntry` in `save()` (instrumentation shows `reloadEntry=false, entry` already set on the second write in *every* case, including passing ones). It's whether the earlier write had already been staged when the later write captured its basis — which is exactly this PR's framing ("read-your-writes only covers writes already staged"), now confirmed by measurement.
- 4.7 passes `patch+patch` only because its per-transaction resource cache (removed in v5 by 2cc16c35a, "No need for resource cache now that we have RocksDB") returns the same instance, and both writes end up applying one accumulated change-set — idempotent double-application, intermediate state erased. It was never an index-correctness mechanism: `put+put` orphans on 4.7 even with the shared instance. So the chain approach here fixes shapes 4.7 never handled, and the RocksDB-RYW rationale for removing the cache is the same assumption this PR corrects.
- The pre-fix RocksDB replicated case loses the first patch's fields on the replica (divergence from origin), not just the index entry — PR description updated accordingly; strengthens the case for a patch release.

Comment generated by Claude Opus 5 (probe: `basisProbe.test.js`, available on request — the PR's existing regression tests already cover the failing shapes).

resources/DatabaseTransaction.ts
```diff
@@ -173,13 +173,70 @@ export type TransactionWrite = {
...
+
+/**
+ * Key identity for the per-key write chain. A Map already distinguishes primitive types (including
+ * bigint, for ids beyond Number.MAX_SAFE_INTEGER), so any primitive key indexes it as-is (no
+ * allocation on the write path, and no risk of JSON.stringify throwing on a bigint — see
+ * BigInt.prototype.toJSON in JSONStream.ts). Only composite (array) keys need JSON, since two
+ * instances with the same content aren't reference-equal; the replacer keeps a bigint element from
+ * throwing the same way.
+ */
+function writeKeyId(key: Id): unknown {
```
kriszyp: The chain's equality does not match the storage engines' key equality. `Map` uses SameValueZero for scalar keys, while the composite path uses JSON; the stores use their binary key encoding. For example, `[0]` and `[-0]` are accepted by `checkValidId` and encode to different RocksDB keys, but both become `\u0000[0]` here; `[null]`, `[NaN]`, and `[Infinity]` likewise collapse under JSON. If one transaction writes two such distinct records, the second links to the first and uses its staged value, corrupting record/index state. Please derive this identity from the same canonical key encoding as the store (or fully type-tag every supported numeric value), and add collision regressions such as `[0]` vs. `[-0]` and `[null]` vs. `[NaN]`.

resources/Table.ts
```diff
@@ -2794,6 +2820,19 @@ export function makeTable(options) {
...
+						// publish what this write left for the key, for any later write to it in this
+						// transaction (an audit-only commit stored no record, so it stages nothing and the
+						// earlier staged record, if any, remains the basis)
+						if (storeRecord) {
+							write.stagedEntry = { value: recordToStore };
+							// blobs this write saved are referenced by its audit entry (if it wrote one), which
+							// then owns their lifetime; and any record an earlier write in this transaction
+							// stored is now replaced, so mark those writes for the superseded-blob cleanup
+							write.blobsAuditReferenced = Boolean(isCopyApply ? false : audit);
+							for (let prior = write.priorWrite; prior; prior = prior.priorWrite) {
```
kriszyp: This walks the full prior-write chain after every successful same-key write, so N repeat writes do N(N-1)/2 iterations on the exact path this PR enables. Only the nearest prior operation that staged a record needs to be newly marked; older staged operations were already superseded by their immediate staged successor in this commit round, and retries replay that ordering. Could `priorStagedEntry` also return the owning operation (or could the transaction track it) so this remains linear?

unitTests/resources/indexOrphanReplicaApply.test.js
```diff
@@ -0,0 +1,128 @@
...
+			],
+			audit: true,
+		});
+		if (Inst.indexingOperation) await Inst.indexingOperation;
+		Inst.sourcedFrom({
+			subscribe() {
+				return (sub = new IterableEventQueue());
+			},
+		});
+		await new Promise((resolve) => setTimeout(resolve, 50));
```
kriszyp: `sourcedFrom()` starts an async task and awaits `subscribe()`, so 50 ms is only a scheduling guess and can still leave `sub` undefined on a loaded runner. Since `waitFor` is already imported, please wait for the actual condition (`await waitFor(() => sub, ...)`) instead.

unitTests/resources/indexOrphanSameKeyTxn.test.js
```diff
@@ -0,0 +1,177 @@
...
+		const stopped = await collect(Inst.search({ conditions: [{ attribute: 'status', value: 'STOPPED' }] }));
+		assert.strictEqual(running.filter(isOurs).length, 0, 'stale RUNNING index entry');
+		assert.strictEqual(stopped.filter(isOurs).length, 0, 'stale STOPPED index entry');
+	});
+
+	// A blob stored by an earlier write in the transaction and replaced by a later one is reachable
+	// from nothing once the transaction commits (no audit entry references it on an audit-off table),
+	// so it must be cleaned up like a skipped write's blobs.
+	it('a blob replaced by a later write in the same transaction is cleaned up (audit off)', async () => {
+		setDeletionDelay(0);
```
kriszyp, line 204: `setDeletionDelay(0)` changes module-global state and this file never restores the 500 ms default, so tests loaded after this suite inherit different blob timing (and a failure here guarantees the leak). Please restore it from an `after`/`afterEach` hook or `finally`.

resources/Table.ts
```diff
@@ -2251,25 +2256,46 @@ export function makeTable(options) {
...
+					// The record a preceding write in this transaction left for this key is what this write
+					// applies to (see priorStagedEntry): a staged write is not visible to a read, so without
+					// this the index diff re-removes the values that write already removed and never removes
+					// the ones it added — orphaning them — and an incremental update merges onto the
+					// pre-transaction record, dropping that write's changes entirely (harper#1968). Only the
+					// record comes from the earlier write; the rest of the entry (version, audit chain, blob
+					// metadata) stays the pre-transaction one, which is the version this write's audit entry
+					// and optimistic version check are still relative to.
+					const priorStaged = priorStagedEntry(write);
+					const existingRecord = priorStaged ? priorStaged.value : existingEntry?.value;
```
Devin-Holland, line 2271: When `priorStaged` came from a chained delete (`{ value: undefined }`), `existingRecord` is `undefined` — which is meaningful and handled on the main store path, but on an **audit-off** table with a strictly newer concurrent entry (`precedesExisting < 0`, e.g. a retry round racing another transaction), it flows into the no-audit branch's `rebuildUpdateBefore(update, existingRecord, fullUpdate)` (~L2669), whose `for (const key in update) { if (key in newerUpdate) ... }` throws a `TypeError` on an undefined second argument. Pre-fix that was only reachable through a tombstone entry's value; the staged-delete basis adds a direct route.

Cheap hardening, fine as a follow-up given how narrow the path is (audit-off + delete-then-patch on one key + concurrent newer entry on a retry round): `existingRecord ?? {}` at that call site — rebuilding against nothing keeps every field, which is the intended semantics — or an early `if (!newerUpdate) return update` guard inside `rebuildUpdateBefore`, which would also subsume the corrupt-audit-value guard the early-out fold carries (~L2530).

resources/Table.ts
```diff
@@ -2251,25 +2256,46 @@ export function makeTable(options) {
...
+					// order, not by timestamp: every write in a transaction carries the transaction's single
+					// timestamp, so a version comparison against what the earlier write landed (e.g. on a
+					// retry round after a partial apply) is a tie (0), and the out-of-order resequencing below
+					// would treat this write as a re-delivered duplicate and drop it (the 4.7 behavior this fix
+					// must not reintroduce). Program order only breaks the tie, though: a strictly newer
+					// existing version (< 0) can only be a concurrent transaction's write observed on a retry
+					// round, and must still go through the out-of-order merge below or its changes would be
+					// silently overwritten.
 					let precedesExisting = precedesExistingVersion(txnTime, existingEntry, options?.nodeId);
+					if (priorStaged && precedesExisting >= 0) precedesExisting = 1;
```
Devin-Holland, line 2299: Non-blocking, adjacent to the duplicate-version audit concern already flagged in the PR body — writing it down where it lives: in the `< 0` out-of-order path, the keyed dedup lookups identify "already applied" by `(version, nodeId, id)`, a tuple a **sibling** write in this same transaction now shares. Round 1 is safe (the sibling's audit entry is still batched on the native transaction, invisible to the plain `auditStore.get`), and on a retry round `stagedOwnAuditEntry` covers any write that appended in round 1. The residual case is a chained write that *skipped* in round 1 (so `appendedAuditEntry` was never set) while its sibling appended: on the retry round, its dedup lookup finds the sibling's now-durable entry and drops the write as already-applied. That takes round-1 skip + retry + out-of-order state lining up, and it's the same fundamental same-version ambiguity the PR body flags, so deferring seems right — a chain-aware guard (treat a `(txnTime, nodeId)` hit as own-orphan when any write on the `priorWrite` chain has `appendedAuditEntry`) would close it, at the cost of weakening genuine re-delivery dedup for these keys.

unitTests/resources/indexOrphanSameKeyTxn.test.js
```diff
@@ -0,0 +1,177 @@
...
+		const running = await collect(Inst.search({ conditions: [{ attribute: 'status', value: 'RUNNING' }] }));
+		const stopped = await collect(Inst.search({ conditions: [{ attribute: 'status', value: 'STOPPED' }] }));
+		assert.strictEqual(running.filter(isOurs).length, 0, 'stale RUNNING index entry');
+		assert.strictEqual(stopped.filter(isOurs).length, 0, 'stale STOPPED index entry');
+	});
+
+	// A blob stored by an earlier write in the transaction and replaced by a later one is reachable
+	// from nothing once the transaction commits (no audit entry references it on an audit-off table),
+	// so it must be cleaned up like a skipped write's blobs.
+	it('a blob replaced by a later write in the same transaction is cleaned up (audit off)', async () => {
```
Devin-Holland, line 203: `setDeletionDelay(0)` leaks into every later test file in this mocha process (`blob.test.js` restores it after each use — "restore the default"). Worth restoring here too:

```suggestion
	after(() => setDeletionDelay(500)); // restore the default; later test files share this mocha process

	it('a blob replaced by a later write in the same transaction is cleaned up (audit off)', async () => {
```

Devin-Holland (approved):
## Review

Traced the design end-to-end and verified it empirically rather than just reading the diff. Verdict: **approve** — this is a careful fix to a genuinely hard problem, and the regression tests are real (see falsifiability check below). Three minor inline notes, none blocking.

### What was verified (local, darwin/arm64)

- Both new test files pass at the PR head on **both engines** (RocksDB and `HARPER_STORAGE_ENGINE=lmdb`): 10/10 each.
- **Falsifiability check**: reverting only `DatabaseTransaction.ts` / `LMDBTransaction.ts` / `Table.ts` to the merge-base (3e317c7f5) and rebuilding reproduces all five claimed failure modes on LMDB — index orphan, dropped first patch, delete-then-patch resurrection, superseded-blob leak, and the replicated same-transaction phantom — with the controls still passing.
- Full `unitTests/resources` at the PR head: LMDB **1137 passing / 0 failing**, RocksDB **1277 passing / 0 failing**.
- The changed files are prettier-clean and oxlint-clean.

### The red CI is inherited, not this PR's

`main` is broken by 64a0e01ba (`unitTests/server/udsMirror.test.js` — an applied suggestion left `callsFake((ms, cb) => {` unclosed), which kills Format Check, runLinter, and every unit-test job at file-load time, on main and on every PR based on it. A one-line fix PR for main is being spun up separately; after a merge from main this PR should go green — nothing in its own files trips prettier or oxlint.

### Design read

- The tie-only override (`priorStaged && precedesExisting >= 0` → `1`) is the right call: a chained write's version comparison against what its sibling landed is exactly a version+node tie, and program order is the only correct tiebreak there, while a strictly newer entry (`< 0`) can only be a concurrent transaction observed on a retry round and correctly still goes through the out-of-order merge. Both bots' earlier findings are properly addressed in the current head.
- The LMDB exclusive-`store.transaction()` fallback is quietly the best illustration of why the override matters: there `write.entry` reloads to the sibling's just-applied record (a version tie), which pre-fix would have deduped the second write away.
- Re-delivered replicated transactions still dedup fully: a skipped write stages nothing, so its successor doesn't get the override and falls through to normal duplicate handling. Nice self-consistency.
- `writeKeyId`'s NUL-namespace encoding holds against every collision shape I could construct (scalar `'[\"x\"]'` vs `['x']`, `[1n]` vs `['1']` vs `[1]`, NUL/SOH-leading scalar strings), and Map identity keeps the scalar hot path allocation-free.
- All five `writes = []` reset sites are converted to `clearWrites()` (three in DatabaseTransaction, two in LMDBTransaction), so no reuse path can inherit a stale chain.
- Agree with flagging (not fixing) the duplicate-version audit entries and the per-key map retention; one adjacent corner is written down inline at the conflict-guard site.

### Environment aside for other reviewers

A fresh `npm ci` here did not install `@harperfast/rocksdb-js-darwin-arm64` (RocksDB tests then fail with `store.db.setIteratorState is not a function`); installing that package explicitly fixes it. Looks like the known npm optional-deps/lockfile gap rather than anything in this PR.

Reviewed by Claude (Fable 5) — devain

resources/DatabaseTransaction.ts
kriszyp: Adopted in 02cbc1a1d — writeKeyId now delegates to the store's own encoding (toBufferKey → latin1), so chain identity is store identity by construction. Verifying empirically also surfaced the reverse direction: 1n and 1 (and 2**60 vs 1n<<60n) encode to the SAME stored key, so the previous raw-Map identity failed to chain them — re-introducing the #1968 stale basis for that pair. Regressions: a unit-level identity matrix ([0]/[-0], [null]/[NaN], [Infinity]/[-Infinity], ['x']/'["x"]', [1n]/['1'], scalar 0/-0, plus same-key pairs) and an API-level bigint/number chaining test. API-level tests for the non-finite composite pairs are blocked on a separate pre-existing bug: point-reads (and therefore indexed-search row resolution) drop records whose composite id contains falsy/non-finite elements — filed as a follow-up task. — Claude (Opus 5) for Kris

resources/Table.ts
kriszyp: Fixed in 02cbc1a1d: priorStagedEntry is now priorStagedWrite and returns the owning write, so the handlers mark only the nearest staged prior — anything older was already marked by its own staged successor earlier in the round, and retries replay that ordering. Linear across N repeat writes. — Claude (Opus 5) for Kris

unitTests/resources/indexOrphanReplicaApply.test.js
kriszyp: Fixed in 02cbc1a1d: await waitFor(() => sub, ...) replaces the 50ms guess. — Claude (Opus 5) for Kris

unitTests/resources/indexOrphanSameKeyTxn.test.js
kriszyp: Fixed in 02cbc1a1d: after(() => setDeletionDelay(500)) restores the default, matching blob.test.js's hygiene. — Claude (Opus 5) for Kris

resources/Table.ts
kriszyp: Confirmed and hardened in 02cbc1a1d — took the in-function guard (early return of the update when newerUpdate is nullish) since rebuilding against nothing keeps every field, which is the intended semantics for both this path and the tombstone route. No dedicated test: the path needs audit-off + delete-then-patch + a concurrent newer entry on a retry round, which isn't deterministically reachable from a unit test; the guard is total, though. — Claude (Opus 5) for Kris

resources/Table.ts
kriszyp: Agreed on deferring — good writedown of the residual (round-1 skip + retry + out-of-order state, sibling's durable entry shadowing the dedup tuple). Added it to the PR description's open-concerns note alongside the same-version audit ambiguity it stems from, so it doesn't get lost when this merges. A chain-aware guard as you sketch trades genuine re-delivery dedup for these keys, which is why it belongs with the broader same-version design question rather than this PR. — Claude (Opus 5) for Kris

unitTests/resources/indexOrphanSameKeyTxn.test.js
kriszyp: Applied in 02cbc1a1d, exactly as suggested. — Claude (Opus 5) for Kris

unitTests/resources/indexOrphanSameKeyTxn.test.js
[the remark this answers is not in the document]
kriszyp: Applied in 02cbc1a1d (after() restores the 500ms default). — Claude (Opus 5) for Kris

kriszyp (commented):
🤖 Dispatch codex/gpt-5.6-sol review (verdict CHANGES).

Issue #1893, opened 2026-07-22
delete_deployment_payload operation not implemented


pbrumblay:
### Bug Summary

delete_deployment_payload operation not implemented

### Steps to Reproduce

Use a script to list deployments. Something like this: 

```
#!/usr/bin/env bash
# Clears old deployment payload blobs from system.hdb_deployment on a Harper Fabric
# instance to free quota, keeping the N most recent (by started_at) deploys per project
# that still have a live blob, for rollback safety. Everything else's blob gets cleared
# via delete_deployment_payload -- the row/metadata (project, payload_size, status,
# timestamps) stays, only the blob bytes go.
#
# Usage:
#   HARPER_OPERATIONS_URL='https://<host>:9925' HARPER_USERNAME=... HARPER_PASSWORD=... \
#     ./clean-dev-deployments.sh [--keep N] [--apply]
#
# Defaults to dry-run (prints what WOULD be deleted, deletes nothing).
# Pass --apply to actually call delete_deployment_payload.
#
# HARPER_HOST port is a guess (9925, the conventional Operations API port) -- if this
# instance uses a different port for plain operations, override it.

set -euo pipefail

HOST="${HARPER_OPERATIONS_URL:?Set HARPER_OPERATIONS_URL, e.g. https://sqo-nl-ams-1.eu-www-development.somebody-nonprod.harperfabric.com:9925}"
HOST="${HOST%/}"
HUSER="${HARPER_USERNAME:?Set HARPER_USERNAME}"
HPASS="${HARPER_PASSWORD:?Set HARPER_PASSWORD}"
KEEP_PER_PROJECT=10
APPLY=false

while [[ $# -gt 0 ]]; do
	case "$1" in
	--apply)
		APPLY=true
		shift
		;;
	--keep)
		KEEP_PER_PROJECT="$2"
		shift 2
		;;
	*)
		echo "Unknown arg: $1" >&2
		exit 1
		;;
	esac
done

op() {
	local raw status body
	raw=$(curl -s -w '\n%{http_code}' -u "${HUSER}:${HPASS}" -X POST "${HOST}/" -H 'Content-Type: application/json' -d "$1")
	status="${raw##*$'\n'}"
	body="${raw%$'\n'*}"
	if ((status < 200 || status >= 300)); then
		echo "HTTP ${status}: ${body}" >&2
		return 1
	fi
	echo "$body"
}

echo "Fetching deployment list from ${HOST} ..." >&2
[...]

PR #1972, opened 2026-07-28
test(server): promote 3 QA server-layer contract anchors

kriszyp:
## Summary

Promotes gated qa-explorer probes into `integrationTests/server/` as standalone (isolation-required) suites. Test-only additions — no runtime code changes.

- **qa702-sse-event-data.test.ts** (QA-702 / P-478) — pins the Resource SSE contract: an 8-case `event.data` payload matrix (undefined/null/''/0/false/object/nested/~300KB string) never crashes a worker and the connection stays healthy, plus the F-133 mid-stream-throw regression leg (a generator that throws mid-stream delivers prior events then closes promptly, not hangs) and a liveness canary. Anchors merged #1863 and F-133. Measured: 10/10 passing, ~6.3s.
- **qa577-upgrade-builtins.test.ts** (QA-577 / P-373) — pins the in-place-upgrade built-in-component config backfill: booting over a config lacking a now-built-in component backfills exactly one top-level key with an `Activated built-in component(s)...` log; a second boot over the same data dir is idempotent (no duplicate/re-log); a fresh install with the key already present is a no-op; OSS-core (no built-ins registered) never grows the Pro-only key. Regression anchor for merged #1814. Measured: 4/4 passing, ~12.3s.
- **qa579-cli-exit-codes.test.ts** (QA-579 / P-374) — pins CLI failure paths exiting non-zero via a 5-cell exit-code matrix (bad config, unreachable instance, operation timeout, failed op, success control) spawning the real `dist/bin/harper.js`, with a hard-kill safety net so a regression-to-hang surfaces as an explicit kill, not a false exit 0. Regression anchor for merged #1801. Measured: 5/5 passing, ~6.1s.

**Dropped from this batch:** P-440 (qa665-ops-surface, gh#1893 ops-API failure-shape contract). Cold-rerun on current main showed 2 of 5 sub-tests failing: the two probes hardcode `get_deployment_payload`/`delete_deployment_payload` as "documented-but-absent" operations, but merged commit `3d329a526` ("Implement get_deployment_payload and delete_deployment_payload operations", closing #1893) already registered both on main — they now return a different (still-correct) 4xx shape instead of the generic "Operation not found" the test asserts on. The test's core premise is stale rather than flaky; fixing it would mean picking new genuinely-absent operation names and rewriting those assertions, which is a content change, not a promotion-gate fix, so it was left out rather than fudged to pass.

All three kept specs are `requires-isolation` (SSE worker/crash assertions, an actual re-boot of the instance, and spawning the real CLI binary), so each stays a standalone file rather than being appended into an existing suite.

## Test plan
- [x] Cold-reran each candidate from a scratch copy on current main (base `3e317c7f5`) before promoting
- [x] `npx prettier --check` and `npx oxlint --format stylish --deny-warnings` clean on all touched files (one `--write` pass needed for a long line in `qa702-sse-event-data/resources.js`; re-ran the spec after to confirm behavior-neutral)
- [x] Re-ran all three specs once more from their final `integrationTests/server/` location — all green
- [x] No `restartHttpWorkers()` against pre-installed fixtures (replaced 4 real calls in qa577 with a direct `/Marker/` readiness poll, matching `integrationTests/database/eviction-secondary-index.test.ts`'s pattern)
- [x] No hard-coded developer paths in any promoted spec or fixture

🤖 Promoted via qa-explorer, generated with Claude Opus 5.

### DIFF
--- a/integrationTests/server/qa577-upgrade-builtins.test.ts
+++ b/integrationTests/server/qa577-upgrade-builtins.test.ts
@@ -1,0 +1,289 @@
+/**
+ * Promoted from qa-explorer (QA-577 / P-373): pins the in-place-upgrade built-in-component
+ * config backfill (regression anchor for merged PR #1814) — booting over a config lacking a
+ * now-built-in component backfills exactly one top-level key with an activation log; a second
+ * boot is idempotent; a fresh install with the key already present is a no-op; OSS-core never
+ * grows the Pro-only key.
+ *
+ * QA-577 — exploratory probe of PR #1814 "fix(upgrade): activate new built-in
+ * components on in-place-upgraded configs" (branch kris/secretcustody-upgrade-config-585).
+ *
+ * The fix (config/configUtils.ts: ensureConfigKeysPresent / ensureBuiltInComponentConfigKeys,
+ * called every boot from bin/run.ts initialize()) backfills a top-level config key for a
+ * newly-introduced built-in component (currently only `secretCustody`, see
+ * UPGRADE_BACKFILL_BUILTIN_KEYS) when it's absent, so the component activates on an
+ * in-place-upgraded config that predates it. It is scoped to built-ins registered in THIS
+ * runtime via the HARPER_BUILTIN_COMPONENTS env var (a comma-separated `name=packageIdentifier`
+ * list an embedding distribution — e.g. harper-pro — sets before boot); on OSS core, where
+ * nothing is registered, it is a no-op.
+ *
+ * ## Honest scope note
+ * This is OSS core (github.com/HarperFast/harper). `secretCustody` is Pro-only — its real
+ * package is never present here, and OSS core's own defaultConfig.yaml never defines the key.
+ * So the "genuine pre-upgrade config" and "genuine fresh Pro install" states can't be produced
+ * by installing real Harper builds; instead HARPER_BUILTIN_COMPONENTS is set directly to make
+ * this runtime "register" `secretCustody` the same way harper-pro would, and the fresh-install
+ * control directly seeds the config file with the key present (see seedConfigKey) to reproduce
+ * the "already there" starting state a real defaultConfig.yaml would produce. The config-file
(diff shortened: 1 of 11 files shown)


integrationTests/server/qa577-upgrade-builtins.test.ts
```diff
@@ -0,0 +1,314 @@
...
+
+		const client = createApiClient(ctx.harper);
+		await waitForRouteReady(client, '/Marker/');
+
+		const { doc } = readConfigFile(ctx.harper.dataRootDir);
+		ok(
+			!Object.prototype.hasOwnProperty.call(doc, BACKFILL_KEY),
+			`OSS core with no built-ins registered must not get ${BACKFILL_KEY} backfilled; config keys were: ${Object.keys(doc)}`
+		);
+		ok(!readBootLog(ctx.harper).includes(ACTIVATION_LOG_SNIPPET));
```
cb1kenobi: **Low: the negative log assertion here (and in Suite 2, line ~280) can pass vacuously**

`readBootLog()` returns `''` when `hdb.log` is missing, so `!readBootLog(...).includes(ACTIVATION_LOG_SNIPPET)` is trivially true if the log file was never created or the path drifted. Unlike qa702, which guards this exact hazard with a positive control (`ok(readLogSafe(logPath).length > 0, ...)`), Suites 2 and 3 have no such guard. Suite 1's positive check only exercises the log path when an activation line is actually written — which by design never happens in these two suites — so it doesn't cover them. To actually mask a regression you'd also need a backfill-logging fault, so the impact is limited; the `hasOwnProperty` / `countTopLevelKey === 1` checks still carry the real contract.

Suggested fix: before each negative log assertion, add a positive control that the log is non-empty (booting at `logging.level: 'info'` as qa702 does), so an empty/missing `hdb.log` fails red instead of passing silently.

—
_Generated by Barber AI_

cb1kenobi (approved):

PR #1980, opened 2026-07-28
Remove vestigial resourceCache plumbing

kriszyp:
Pure deletion (12 lines). The per-transaction resource-instance cache was removed in 2cc16c35a ("No need for resource cache now that we have RocksDB"), but its plumbing stayed behind with no remaining reader: `transaction()` allocated a `new Map()` per transaction, `DatabaseTransaction` nulled it at three commit/abort sites, `Table` passed it into source contexts, and both Context interfaces declared the field.

Grepped harper and harper-pro for readers — none. The only behavioral effect is dropping a dead Map allocation per transaction.

Surfaced while investigating #1968 (the removal commit's RocksDB read-your-writes rationale is the same assumption PR #1970 corrects, which is how this dead code got noticed).

Note: `resourceCache?` was an optional field on the exported Context types; any external TS code *reading* it would now fail to compile. Nothing in-tree or in harper-pro does.

Tested: build clean; crud + caching unit tests pass on both engines. Cross-model review (Codex): no findings.

Generated by Claude Opus 5.

### DIFF
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -660,7 +660,6 @@ export class DatabaseTransaction implements Transaction {
 							// now reset transactions tracking; this transaction be reused and committed again
 							this.retries = 0; // reset per-native-transaction retry counter so a reused DatabaseTransaction's next batch starts fresh
 							this.clearWrites();
-							if (this.#context?.resourceCache) this.#context.resourceCache = null;
 							this.next = null;
 							let txnTime = this.timestamp;
 							this.timestamp = 0; // reset the timestamp as well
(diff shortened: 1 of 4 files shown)


cb1kenobi (approved):

DavidCockerill (approved):
Approving. Verified it's genuinely vestigial rather than compile-clean: `git grep -i resourcecache` at `a44f1f65` returns nothing across the tree — including bracket and string-keyed access, which a type-checker wouldn't catch — and harper-pro's only hits are its mirror of these same four files.

The red Node 22 check looks like the known audit-log polling flake rather than anything here; worth a re-run before merge.

Issue #1881, opened 2026-07-21
Secondary-index reads return partial/empty results for the second table accessed in a single request (5.1.19 and 5.1.22)


kylebernhardy:
## Summary

In Harper **5.1.19 and 5.1.22**, within a **single resource request**, the first
access to a table returns an **incomplete** result set if a **different table**
was read earlier in that same request. Reading that table first — or as the only
read in the request — returns the full set.

Only the **secondary-index** read path is affected. An unconditioned full scan of
the same table, in the same position in the same request, returns every row. The
records are intact; the index scan comes back short.

The magnitude ranges from a partial count to **zero**. It silently
empties/undercounts any endpoint that aggregates across multiple tables in one
request.

## Environment

- Harper **5.1.19**, and reproduced identically on **5.1.22**.
- `@harperfast/rocksdb-js` **2.4.0**.
- Reproduced on a single node (macOS) and on a 2-node Fabric cluster.
- Component plugins: `rest` + `graphqlSchema` + `jsResource`; TypeScript resources.

## Reproduction — self-contained, no real data required

Two tables, one indexed attribute, seeded synthetically. **100% deterministic.**

The essential ingredient is the write pattern: every key gets rows on **every
wave**, so a single key's index entries are scattered across many sorted runs
rather than sitting contiguously in one. Seeding the same rows in a single pass
does **not** reproduce.

### `schemas/repro.graphql`

```graphql
type GenA @table(database: "metrics-repro") @export {
  id: ID @primaryKey
  repositoryId: String @indexed
  n: Int @indexed
  body: String
}

type GenB @table(database: "metrics-repro") @export {
  id: ID @primaryKey
  repositoryId: String @indexed
  n: Int @indexed
  body: String
}
```

### `resources/Gen.ts`

```ts
import { Resource, databases } from "harper";

const db: any = (databases as any)["metrics-repro"];
const PAD = "x".repeat(1024);

// Every key gets rows on EVERY wave, so one key's index entries end up
// scattered across many sorted runs rather than contiguous in one.
async function generate(table, keys, perKeyPerWave, waves, bodyKb) {
  const body = PAD.repeat(bodyKb);
  for (let w = 0; w < waves; w++)
    for (let r = 0; r < keys; r++)
      for (let i = 0; i < perKeyPerWave; i++) {
        const n = w * perKeyPerWave + i;
        await db[table].put({
          id: `${table}-r${r}-n${n}`,
          repositoryId: `repo-${r}`,
          n,
[...]

Issue #717, opened 2026-05-22
Support async computed functions and indexes


kriszyp:
## Request

Computed functions (and indexes derived from them) currently must be synchronous. Add support for async computed functions so they can perform I/O (e.g., call external APIs, query other tables) as part of computing a derived value.

🤖 Filed by Claude on behalf of Kris.

Issue #694, opened 2026-05-21
Publish to harperdb npm package in addition to harper


kriszyp:
## Request

Update the release pipeline to publish to the `harperdb` npm package in addition to `harper` and `@harperfast/harper`, for backwards compatibility with users still on the legacy `harperdb` package name.

## Current state

`publish-npm.yaml` publishes to:
- `harper` (public)
- `@harperfast/harper` (scoped)

## Proposed

Add a third publish job for `harperdb`, updating `package.json` name to `harperdb` before packing/publishing (same pattern as the existing `@harperfast/harper` publish job).

## Note

Assess whether `harperdb` publishing is still desirable given the v5.x rebranding, or if users should be migrated to `harper`/`@harperfast/harper`.

🤖 Filed by Claude on behalf of Kris.

PR #1984, opened 2026-07-28
5.1.24 prep: pin rocksdb-js to ~2.4.1, revert #1911 from v5.1

kriszyp:
Prep for the **5.1.24** patch cut. Two changes, both scoped to the `v5.1` line.

## 1. Pin `@harperfast/rocksdb-js` to `~2.4.1`

Pin moves `^2.3.0` → `~2.4.1`; lockfile moves `2.3.0` → `2.4.1`.

2.4.1 is the 2.4-line backport of the cross-column-family transactional read fix ([rocksdb-js #717](https://github.com/HarperFast/rocksdb-js/pull/717), backported in [#732](https://github.com/HarperFast/rocksdb-js/pull/732)). `TransactionHandle::get` honored the caller's `dbHandleOverride` on its synchronous block-cache-tier attempt but dropped it in the async worker, falling back to the transaction's own column family. Since all tables in a Harper database share one read transaction, **every table after the first in a request was read through a foreign column family**. Reads that hit the block cache were correct; reads that missed it silently returned not-found — so the failure was intermittent and scaled with cache residency: worst right after a restart, healing as traffic warmed the cache. Where two tables share a key format, a read could also *hit* in the foreign CF and return another table's row.

Same defect as [#1881](https://github.com/HarperFast/harper/issues/1881).

### Why tilde, not caret

`^2.4.1` resolves to **2.5.0** at install time, which would pull in the dedicated per-database commit thread (`29a5aee1`), the fd-limit-derived `max_open_files` default (`956d8f8a`/`4e604a78`), the VerificationTable epoch rekey (`19b4f2d8`), and a 41-package dep bump — none of which belong in a patch. `~2.4.1` (`>=2.4.1 <2.5.0`) holds 5.1.x on the 2.4 line. `main` stays `^2.5.0` for 5.2.

Verified the v2.4.1 tag contains the CF fix (`7b20a1cc`, `b99c008c`) and excludes all three of the above.

### Why the lockfile moves too

The lockfile is what ships. The original exposure was `^2.3.0` with a `2.3.0` lockfile, so the 5.1.22 image built against whatever npm resolved that day — 2.4.0, with the bug. Pin and lockfile move together so this can't recur.

Lockfile diff is scoped to the nine `@harperfast/rocksdb-js*` entries; nothing else moved.

## 2. Revert #1911 from `v5.1`

[#1911 "fix(query): stop query planning from mutating the caller's conditions"](https://github.com/HarperFast/harper/pull/1911) was labeled `patch`, so the auto-cherry-pick bot put it on `v5.1` — but it targets 5.2 and isn't wanted on the patch train. This reverts all three of its commits (`39f0dd4eb`, `30477f2df`, `3af8eefa7`); I've also removed the `patch` label from the source PR.

The revert is byte-for-byte the inverse of the cherry-picked range: `resources/Table.ts` returns to its `v5.1.23` state and `unitTests/resources/conditionsArrayMutation.test.js` is removed. The fix stays on `main` and can be cherry-picked deliberately later.

## Resulting 5.1.24 delta

`v5.1.23..HEAD` is exactly [#1914 "Enforce table read permissions for subscriptions on Harper 5.1"](https://github.com/HarperFast/harper/pull/1914) plus this dep bump:

```
 .github/workflows/claude-review.yml     | 28 ++++-
 package-lock.json                       | 72 ++++++------
 package.json                            |  2 +-
 resources/Table.ts                      |  6 +-
 unitTests/resources/permissions.test.js | 91 +++++++++++++++++-
```

Paired with [harper-pro #626](https://github.com/HarperFast/harper-pro/pull/626).

<sub>Note on the earlier CI run: `Integration Tests 1/6 (Windows, Node.js v24)` failed once on a 60s restart probe. Re-run is green, and the HNSW suite took 444s vs 437s on the base commit — the failing attempt took 743s, i.e. a degraded runner, not a code effect.</sub>


cb1kenobi (approved):

kriszyp:
**Holding as draft — switching the target from 2.5.0 to 2.4.1.**

Chris is cutting a `@harperfast/rocksdb-js` **2.4.1** for this release instead: [rocksdb-js #732 "Backport #717 to v2.4.x: resolve transaction reads against the caller's column family"](https://github.com/HarperFast/rocksdb-js/pull/732) (open, not yet merged; 2.4.1 not yet on npm).

That's the right call for a patch line. 2.5.0 carries a lot more than the CF-read fix:

- `29a5aee1` dedicated per-database commit thread for async transaction commits (#694)
- `956d8f8a` / `4e604a78` bound RocksDB `max_open_files` with an fd-limit-derived default
- `19b4f2d8` key VerificationTable slots by per-open epoch, not descriptor pointer
- `45f11007` align `ERR_TRY_AGAIN` with the `IsBusy` reset path
- `10e2101c` a 41-package dependency bump

None of that belongs in a 5.1.x patch. The backport isolates exactly the five CF-read commits, and I verified its diff is **byte-for-byte identical** to the same cluster on `main` (`67322257^..17b49573`) apart from one list-number in `AGENTS.md` — 327 insertions / 13 deletions across the same 7 files.

### The pin needs to be `~2.4.1`, not `^2.4.1`

`^2.4.1` means `>=2.4.1 <3.0.0`, so npm would resolve **2.5.0** at image-build time and pull in everything above anyway — making the backport pointless. `~2.4.1` (`>=2.4.1 <2.5.0`) keeps the 5.1.x line on the 2.4 series.

This is the same trap that caused the original incident: the pin was `^2.3.0` and the lockfile said `2.3.0`, so the shipped 5.1.22 image got whatever npm resolved at build time — 2.4.0, which has the bug. The lockfile is the source of truth for what ships, so both move together.

`main` stays on `^2.5.0` — 5.2 should take the full 2.5.0.

### Plan

Once #732 merges and 2.4.1 publishes, I'll amend the dep commit here to `~2.4.1` + regenerate the lockfile, and mark ready. Everything else in this PR is unaffected and stands as-is.

cb1kenobi (approved):

PR #1985, opened 2026-07-28
Expose forwarded PROXY v2 TLS facts as request.connectionInfo

### LINKED ISSUE #1983: Parse PROXY protocol v2 and expose TLS facts (JA3/JA4, ALPN, SNI, mTLS cert) to the JS request layer
## Summary

Harper parses PROXY protocol **v1 only**, and uses it solely to recover `socket.remoteAddress` / `remotePort`. Please add v2 parsing on the UDS mirror and expose the v2 **TLV** set to the JS request layer.

This is the blocking dependency for consuming the TLS facts Symphony already computes — most importantly the client's JA3/JA4 fingerprint, which an application component behind Harper currently has no way to see.

## Current state

`enableProxyProtocol` (`dist/server/http.js:1062` in 5.1.23) matches the literal `PROXY ` prefix, reads up to the 108-byte v1 maximum, splits on spaces, and overrides `remoteAddress`/`remotePort`. It's wired to the UDS mirror in both `http.js:659` and `threads/threadServer.js:543` — the latter's comment names symphony directly:

> Strip the PROXY v1 header a fronting proxy (e.g. symphony) prepends, same as the HTTP UDS mirror.

Verified unchanged in **`harper@5.2.0-beta.2`** (`dist/server/http.js:1597`, still `PROXY_V1_MAX_HEADER`). There is no v2 signature check, no TLV parsing, and no surface on `Request` for TLS facts. Symphony's README already anticipates this — *"the consumer must speak v2 (nginx/HAProxy do; Harper core's UDS reader parses v1 only before Harper 5.2)"* — but as of the current beta that hasn't landed.

## What Symphony sends under `sourceAddressHeader: 'proxyProtocolV2'`

The v2 header prefixes the connection before any application bytes, so it is protocol-agnostic — it works in TLS passthrough and, critically, under HTTP/2, where nothing can be injected into the request stream.

| TLV | Type | Content |
|---|---|---|
| ALPN | `0x01` | Negotiated ALPN protocol (e.g. `h2`) |
| Authority | `0x02` | SNI hostname from the ClientHello |
| SSL | `0x20` | `client` bit field, `verify` (u32 BE, `0` = verified), sub-TLVs `0x21` TLS version / `0x23` cipher suite |
| JA3 | `0xE0` | JA3 MD5 hex (32 chars) |
| JA4 | `0xE1` | JA4 fingerprint (36 chars) |
| Client cert chain | `0xE2` | One TLV per cert, DER, leaf first — present only when the mTLS verifier accepted it |

`0xE0`–`0xE2` are in HAProxy's `0xE0–0xEF` private range; `0x01`/`0x02`/`0x20` are standard PROXY v2 types.

## Why this matters beyond fingerprints

Once Symphony terminates TLS, **every** connection-level fact becomes invisible to the application: negotiated ALPN, TLS version, cipher, SNI, and the verified client certificate on mTLS routes. A component behind Harper can currently see none of it. The v2 TLV section is the transport that already carries all of it — Harper just doesn't read it.

The verified client cert (`0xE2`) is worth calling out separately: Symphony only emits it when its verifier accepted the chain (rustls aborts the handshake otherwise), so over a trusted UDS link it is a usable verified client identity. Today that identity is discarded at the socket boundary.

## Requested

1. **Detect and parse the v2 binary header** on the UDS mirror, alongside the existing v1 text path. The v2 signature (`\r\n\r\n\0\r\nQUIT\n`) is unambiguous and cannot collide with v1's `PROXY ` prefix or with a TLS ClientHello, so both can be supported on the same listener without a config flag.
2. **Keep the existing behavior:** set `socket.remoteAddress` / `remotePort` from the v2 address block exactly as v1 does, so nothing that depends on the connecting IP has to change.
3. **Expose the TLV set to the JS request layer.** Something stable and typed on `Request` alongside the existing `protocol` / `ip` / `peerCertificate` accessors — e.g. `request.connectionInfo` returning `{ alpn?, authority?, tls?: { version?, cipher?, verified? }, ja3?, ja4?, clientCertChain? }`. Absent when no v2 header was present; never populated from a header.
4. **Fix `Request.protocol` for the UDS mirror.** It currently returns `this._nodeRequest.socket.encrypted ? 'https' : 'http'` (`dist/server/serverHelpers/Request.js:98`). The UDS mirror is a plain `node:http` server, so `encrypted` is undefined and it reports `http` for traffic the client sent over TLS. With the v2 SSL TLV (`0x20`) parsed, Harper can report this correctly; a v1 or bare connection keeps today's behavior. This currently produces wrong `X-Forwarded-Proto` decisions in components that proxy to an upstream.
5. **Do not accept TLVs from a non-UDS listener.** The trust argument for all of this is that the UDS is local and only the fronting proxy can write to it. A TCP listener must not gain PROXY parsing implicitly.

## Secondary: an h2-capable UDS mirror

Related, and needed for the same deployment to enable HTTP/2. The UDS mirror is created with `node:http` `createServer` (`http.js:642`), so it is HTTP/1.1-only; Harper's HTTP/2 support exists only on the secure TCP port (`createSecureServer`, `http.js:592`).

That means an h2 client fronted by Symphony has nowhere good to land. Symphony supports per-protocol split upstreams (`destination_h2`), and a comment in its source (`src/proxy_conn.rs:191`) already refers to "Harper's `-h2.sock` mirror" — but no such mirror exists in 5.1.23 or 5.2.0-beta.2. Routing h2 clients to the TCP secure port instead re-encrypts a hop that should be cheap **and** loses the client IP, since proxy-protocol parsing is not wired to the TCP listener.

An h2 UDS mirror (`createSecureServer`-equivalent over a unix socket, or an h2c mirror) with the same proxy-protocol handling would close this. Happy to split it into its own issue if preferred.

## Context

Blocking [HarperFast/harper-woo-cache#26](https://github.com/HarperFast/harper-woo-cache/issues/26) (forward TLS fingerprints to a WooCommerce origin whose WAF uses them for rate limiting and bot classification) and [#211](https://github.com/HarperFast/harper-woo-cache/issues/211) (enable HTTP/2 for that deployment). We have a live cluster behind Symphony and a live QA suite, so we can test a branch end-to-end against a real origin.

kriszyp:
Closes #1983 (items 3 + 4).

## Summary
[#1858](https://github.com/HarperFast/harper/pull/1858) taught Harper to parse PROXY protocol v2 on the UDS mirror, but only wired the mTLS client cert chain (`0xE2`) to `getPeerCertificate`. The other TLVs symphony forwards under `sourceAddressHeader: 'proxyProtocolV2'` were decoded and thrown away — so an application component behind Harper had no way to read the client's JA3/JA4 fingerprint, negotiated ALPN, SNI, or TLS version/cipher once the fronting proxy terminated TLS. That's the blocking dependency @hdbjeff flagged for the WooCommerce/WAF fingerprint work.

This extends the decoder to capture the full v2 TLV set into a **`ConnectionInfo`** and exposes it as **`request.connectionInfo`**:

```ts
request.connectionInfo // { alpn?, authority?, tls?: { version?, cipher?, verified? }, ja3?, ja4?, clientCertChain? } | undefined
```

TLVs parsed: ALPN `0x01`, authority (SNI) `0x02`, SSL `0x20` (with version `0x21` / cipher `0x23` sub-TLVs and the client-cert-present + verify bits), JA3 `0xE0`, JA4 `0xE1`, client cert chain `0xE2`. Also fixes **`request.protocol`** to report `https` on the plaintext UDS mirror when the SSL TLV shows the client connection was TLS (it previously always returned `http` there → wrong `X-Forwarded-Proto`, #1983 item 4).

## Design note (new API surface — approved shape)
`request.connectionInfo` is a new public `Request` accessor; shape per #1983 (`{ alpn, authority, tls, ja3, ja4, clientCertChain }`). Chosen over synthesizing `x-ja4` headers because the TLV carrier is spoof-safe and works under h2 (nothing to inject into the request stream), and headers are client-forgeable. `connectionInfo` is populated **only** from a decoded PROXY v2 header on the filesystem-permissioned UDS mirror — same trust boundary as the existing v1 source-address override — and never from a request header.

## Where to look
- **mTLS preservation.** The cert path refactors from the old internal `header.tls` to `header.connectionInfo.clientCertChain`; `authorized` + lazy/memoized `getPeerCertificate` stay gated on the `0xE2` chain exactly as #1858 had them. HTTP (`security/auth.ts`) and MQTT (`server/mqtt.ts`) mTLS are unchanged.
- **`connectionInfo.tls.verified` semantics.** It reflects the SSL TLV's verify bit (cert presented + proxy-verified), and can be `true` even when no chain was forwarded (symphony omits an oversized chain but still sends the SSL TLV). Documented on the field; apps wanting the standard "verified mTLS client" gate should use `request.authorized`, which is chain-gated. (Surfaced by the review — a doc clarification, not a behavior change.)
- **Bounds.** `parseSslSubTlvs` and the main TLV loop are byte-bounds-guarded (`offset+3 <= end`, `valueEnd > end` breaks); covered by unit tests + fuzz.

## Not in scope
- v2 parsing / `0xE2` mTLS cert (already shipped in #1858); trust-gating to the UDS listener (#1983 item 5, already the case).
- The h2-capable UDS mirror (#1983 secondary) — its own issue.

## Testing
- `unitTests/server/serverHelpers/proxyProtocol.test.js` + `Request.test.js` + `udsMirror.test.js`: 123 passing. New coverage for every TLV (ALPN/authority/SSL sub-TLVs/JA3/JA4/cert), JA3/JA4-without-SSL, `connectionInfo` on the socket, v1-has-no-connectionInfo, and `protocol` https-via-SSL-TLV.
- Full `test:unit:main` not run locally (pre-existing env failure — `getUser` stub / storage-path, reproduces on clean main); relying on CI.

## Review
Cross-model (thorough): Codex + Gemini + **Grok** (xAI, per request) outside legs + Harper domain adjudication. **No blockers, no significant concerns** — the bounds safety, #1858 mTLS equivalence, and trust boundary were independently traced. The two suggestions (the `verified` doc note above; a per-connection `Object.keys` allocation) are applied.

Generated by an LLM (Claude Opus 4.8) pairing with Kris.


kriszyp:
Docs: this adds a public `Request` accessor (`request.connectionInfo`) and changes `request.protocol` on the UDS mirror, so a companion PR to [HarperFast/documentation](https://github.com/HarperFast/documentation) (the Request/globals API reference) should follow — will file once the shape is confirmed in review. cc @hdbjeff — this is the Harper-side consumer for the JA3/JA4 forwarding you filed in #1983; `request.connectionInfo.ja4` (and `.ja3`, `.alpn`, `.tls`, `.clientCertChain`) is how an app reads it. Note it needs symphony configured with `sourceAddressHeader: 'proxyProtocolV2'` + `forwardFingerprint` on the route.

🤖 Claude Opus 4.8, on Kris's behalf

kriszyp:
Docs companion opened: HarperFast/documentation#614 — documents `request.connectionInfo` (the `ConnectionInfo` shape, a JA4 usage example, and the `tls.verified`-vs-mTLS trust caveat) and the `request.protocol` clarification on the `Request` reference.

🤖 Claude Opus 4.8, on Kris's behalf

hdbjeff (changes requested):
Requesting changes for the PP2_CLIENT_SSL protocol-inference issue below.
Otherwise this is solid, and I traced the parts most likely to be dangerous rather than taking the description on trust:
 
- Trust boundary holds. Production reaches applyProxyHeader through two direct call sites, both restricted to UDS-backed paths: enableProxyProtocol (http.ts:1653) for the HTTP/1 mirror, and consumeProxyHeader (proxyProtocol.ts:302), used by withProxyProtocol for both the h2c front (http.ts:1696) and the raw-protocol/MQTT UDS listener (threads/threadServer.js:619). Request.connectionInfo is a pure read of socket.connectionInfo; no request-header path reaches it.
- Bounds are correct. decodeV2 returns incomplete when buffer.length < headerLength, so the TLV loop bounded by headerLength cannot read past the buffer. Both the main loop and parseSslSubTlvs guard offset + 3 <= end and break on valueEnd > end, and a zero-length TLV still advances the offset by 3, so neither can spin.
- mTLS equivalence with #1858 holds. Old code set authorized = verified only under certPresented && clientCertChain; new code sets it only under chain, to info.tls?.verified === true where verified === certPresented && verifyOk. Since clientCertChain is itself only assigned under certPresented, the new value reduces to verifyOk — identical. The chain-present-but-no-SSL-TLV case also behaves the same.
- The tls.verified doc clarification, and the pointer to chain-gated request.authorized as the real mTLS gate, is the right call-out.

request.connectionInfo gives HarperFast/harper-woo-cache#212 exactly what it needs for the eventual PROXY-v2 migration, with application-level fingerprint validation retained. No other blockers found.

hdbjeff:
Blocking inline — proxyProtocol.ts, SSL TLV branch (lines 171–176)

▎ [P2] Honor PP2_CLIENT_SSL before reporting the connection as TLS
▎
▎ This branch creates connectionInfo.tls for every structurally valid SSL TLV but never checks the TLV's client & 0x01 (PP2_CLIENT_SSL) flag — PP2_CLIENT_CERT_CONN (0x02) is the only client-field constant defined in the file. Request.protocol then treats the existence of connectionInfo.tls as proof that the client connected over TLS:
▎
▎ if (this._nodeRequest.socket.encrypted || this._nodeRequest.socket.connectionInfo?.tls) return 'https';
▎
▎ So an SSL TLV with client = 0x00 decodes to { tls: { verified: false } } — a truthy object — and request.protocol returns "https" for a cleartext client. That misreports the scheme in the unsafe direction, and it feeds X-Forwarded-Proto, redirects, secure-cookie decisions, and other scheme-gated behavior. Since only a proxy on the UDS can send this, it's a correctness bug rather than a vulnerability — but a silent one, and it undercuts item 4 of the PR description, which is specifically the protocol fix.
▎
▎ Worth noting the bit is already understood at the test layer: the fixture at unitTests/server/serverHelpers/proxyProtocol.test.js:436 builds Buffer.from([0x01 | 0x02]) and comments it client: SSL | CERT_CONN. It is simply never read in production, which is why no existing test catches this.
▎
▎ Suggested fix — gate the whole branch rather than adding a field:
▎
▎ const PP2_CLIENT_SSL = 0x01;
▎ // ...
▎ const clientBits = buffer[valueStart];
▎ if ((clientBits & PP2_CLIENT_SSL) !== 0) {
▎   certPresented = (clientBits & PP2_CLIENT_CERT_CONN) !== 0;
▎   const verifyOk = buffer.readUInt32BE(valueStart + 1) === 0;
▎   tls = { verified: certPresented && verifyOk };
▎   parseSslSubTlvs(buffer, valueStart + 5, valueEnd, tls);
▎ }
▎
▎ This keeps connectionInfo.tls meaning "the proxy reported TLS was used", so Request.protocol can continue testing object presence and the approved ConnectionInfo shape doesn't grow. It also tightens the cert path slightly, which I think is correct: a header asserting CERT_CONN without CLIENT_SSL is internally inconsistent — a client certificate is a TLS construct — so it would no longer produce authorized. Worth a sentence in the changelog if you agree.
▎
▎ Please also add a regression test using an SSL TLV with client = 0x00.

hdbjeff:
Optional inline — the ConnectionInfo interface (lines 64–81)

▎ Non-blocking: could the ConnectionInfo docs state that the string fields are raw, unvalidated TLV payloads?
▎
▎ ja3, ja4, and alpn are decoded with latin1 and authority with UTF-8, with no grammar or length validation beyond the enclosing PROXY-header bounds. That's reasonable for a transport decoder — I'm not suggesting Harper impose Symphony's fingerprint grammar here — but consumers should validate before placing these values in headers, logs, cache keys, or security decisions, and the type doesn't currently signal that.
▎
▎ Concretely: the motivating consumer (HarperFast/harper-woo-cache#212) re-emits ja4 into an outbound header toward an origin WAF, and will keep its strict validateJa3/validateJa4 checks after migrating off the HTTP-header carrier. A malformed value there breaks origin fetches rather than injecting anything — Headers.set() throws on CR/LF — but that's the consumer having chosen correctly, not the API making it hard to get wrong. One clause in the doc comment would make the contract explicit.

Ethan-Arrowood (approved):
TLV bounds checking, the trust gate (proxy protocol only on the UDS mirror), and mTLS equivalence through the refactor all check out, with unit coverage on every new TLV. Zero per-request cost. The two suggestions in my notes (PP2_CLIENT_SSL client-field bit, explicit `connectionInfo` member on the Bun/uWS request classes) are non-blocking.

server/serverHelpers/proxyProtocol.ts
```diff
@@ -112,30 +154,65 @@ function decodeV2(buffer: Buffer): ProxyHeaderDecision {
...
 		if (valueEnd > headerLength) break; // malformed TLV — keep what we parsed so far
-		if (type === PP2_TYPE_SSL && valueLength >= 5) {
-			// struct pp2_tlv_ssl: client(1) verify(4, 0 = verified ok), then sub-TLVs
-			certPresented = (buffer[tlvOffset + 3] & PP2_CLIENT_CERT_CONN) !== 0;
-			verified = buffer.readUInt32BE(tlvOffset + 4) === 0;
+		if (type === PP2_TYPE_ALPN && valueLength > 0) {
+			info.alpn = buffer.toString('latin1', valueStart, valueEnd);
+		} else if (type === PP2_TYPE_AUTHORITY && valueLength > 0) {
+			info.authority = buffer.toString('utf8', valueStart, valueEnd);
+		} else if (type === PP2_TYPE_SSL && valueLength >= 5) {
```
Ethan-Arrowood, line 179: `protocol` infers `https` from SSL-TLV *presence*, ignoring the `PP2_CLIENT_SSL` (0x01) client-field bit. A proxy can legally emit a `0x20` TLV with client byte `0x00` (the client field says *whether* the client connected over SSL), and Harper would report `https` for a plaintext client connection — the code only reads the `PP2_CLIENT_CERT_CONN` (0x02) bit. Theoretical today since symphony only sends the SSL TLV for TLS connections, but a one-line `& 0x01` check would make the inference spec-faithful and future-proof against other PROXY v2 emitters. Non-blocking.

Ethan-Arrowood (commented):
Following up on my approval with the two suggestions inline (both non-blocking).

Ethan-Arrowood:
One more (can't anchor it inline — the lines aren't in the diff): the Bun and uWS request classes in `server/serverHelpers/Request.ts` (~L401 / ~L511 at head) have no explicit `connectionInfo` member. Reads are `undefined`, which is consistent with "no v2 header", but the neighboring capability gaps there are made explicit (`authorized`, `peerCertificate` return documented sentinels with comments). An explicit `get connectionInfo() { return undefined; }` in the same style would keep the API surface self-documenting and the TS types uniform. Non-blocking.

kriszyp:
Addressed the review feedback:

- **@hdbjeff** — blocking: gated the SSL TLV branch on `PP2_CLIENT_SSL` (0x01), not just presence, so an SSL TLV with `client = 0x00` no longer produces a truthy `connectionInfo.tls` (and `request.protocol` no longer reports `https` for a cleartext client). Applied your suggested fix as written, including treating a `CERT_CONN` bit without `CLIENT_SSL` as internally inconsistent. Added regression tests for both cases (`client = 0x00`, and `CERT_CONN` without `CLIENT_SSL`).
- **@hdbjeff** — non-blocking: documented `alpn`/`authority`/`ja3`/`ja4` as raw, unvalidated TLV payloads on `ConnectionInfo` and each field.
- **@Ethan-Arrowood** — non-blocking: added explicit `connectionInfo` getters (returning `undefined`) on `BunRequest`/`UwsRequest`, matching the existing `authorized`/`peerCertificate` parity getters.

Also applied a suggestion from independent pre-push review: `ConnectionInfo` is now allocated lazily in `decodeV2`, so a metadata-free v2 connection no longer allocates and discards an empty object.

125 relevant unit tests pass; build and lint are clean.

🤖 Claude Opus, on Kris's behalf

PR #1987, opened 2026-07-29
test(database): promote 3 QA destructive-operation integrity anchors

kriszyp:
Test-only. Promotes three exploratory QA specs into `integrationTests/database/` as permanent regression anchors. No product code is touched.

All three pin one class of invariant: **after an operation that removes data, both the disk and the read surfaces must end up correct.** Each was found by the exploratory QA loop, then gated (cold rerun at the promoted path, non-duplicative vs. the closest existing suite test, format/lint clean) before landing here.

## What each spec pins

**`txnlog-purge-stale-read-blast.test.ts`** (from QA-782, `source:F-225`) — bounds the blast radius of F-225. After a bulk delete of a 6000-row contiguous key range, all five ordinary read surfaces (REST GET, REST query, `search_by_value`, SQL, full scan) must be clean **in the same process**, while the audit-read path is separately shown to still return purged rows. That contrast is the point: it pins staleness as **confined to the audit/txnlog path** and *not* reaching ordinary table reads. Runs both engines in-file (engine set per arm), with an armed positive control (`CTRL` bucket inserted after the cutoff) and a restart discriminator. 14 tests, ~14 s.

The closest existing test, `apiTests/transaction-logs.test.mjs`, covers happy-path reads and a *no-op* purge — nothing in the suite drives a real purge and then re-reads ordinary surfaces, which is why this gap went uncaught.

**`blob-delete-reclaim-audit-expiry.test.ts`** (from QA-802, `source:gh:708`) — deleting a blob-bearing record must reclaim its file from disk, regardless of audit retention. N=20 × 256 KB blobs stored, deleted via SQL, with the blob tree under `{dataRootDir}/blobs/data` walked directly for file count and bytes, sampled across the `deletionDelay`, the `auditRetention` window, a hard restart, and an explicit `cleanup_orphan_blobs`. Asserts net leak 0. Arm B is a negative control: audit-only expiry against a **live** record (6 overwrites aged past retention) must leave exactly one byte-intact file. 3 tests, ~53 s.

This **refutes harper#708's claim on the delete path.** Note the engine scope, called out in the file header: in CI this pins the default engine (RocksDB) only, because `HARPER_STORAGE_ENGINE` is process-wide. The LMDB leg was run manually during QA and also showed zero net leak — that is recorded as QA evidence, not asserted here. `#1916`'s `audit-blob-reclaim.test.ts` pins the *overwrite* side; this pins the *delete* side and the `cleanup_orphan_blobs` operator surface.

**`blob-drop-crash-orphan-window.test.ts`** (from QA-809) — blob files orphaned by a crash inside `drop_table`'s unlink window must be reclaimable, and `drop_database` must never leak. Sweeps kill delays (k0/k100/k400/k600) around the drop, plus a no-kill control and two `drop_database` arms. 8 tests, ~79 s.

Two things were deliberately kept as **logged observations rather than assertions**, because asserting them would freeze current behaviour as required: that the unlink window exists at all (a fix that unlinks synchronously would close it — an improvement), and whether restart-recovery happens to reclaim (empirically arm-dependent: `false` at k0/k100, `true` at k400/k600, so it would have been flaky). The asserted invariants are the ones that survive any fix to the drop path: orphans are reclaimable after `cleanup_orphan_blobs`, and `drop_database` leaves 0 blob files at every checkpoint.

## Verification

Cold rerun at the promoted paths on `main` @ `80ef45996`, then re-run again after the header edits in this PR:

| Spec | Result | Wall |
|---|---|---|
| `txnlog-purge-stale-read-blast` | 14/14 | ~14 s |
| `blob-delete-reclaim-audit-expiry` | 3/3 | ~53 s |
| `blob-drop-crash-orphan-window` | 8/8 | ~79 s |

`prettier --check` clean and `oxlint --deny-warnings` exit 0, both run from inside the repo tree. No `restartHttpWorkers()` (all three poll the probe route directly for non-404). No absolute developer-machine paths. All three carry an arming assertion that fails if the disk oracle cannot measure, so none can pass vacuously.

⚠️ **Reviewer note on runtime:** `blob-drop-crash-orphan-window` (~79 s) and `blob-delete-reclaim-audit-expiry` (~53 s) are on the slower side, dominated by deliberate settle/retention windows rather than by assertion count. Both are `requires-isolation` (they kill and restart the instance in place), so they cannot share a host instance with other specs.

### DIFF
--- a/integrationTests/database/blob-delete-reclaim-audit-expiry.test.ts
+++ b/integrationTests/database/blob-delete-reclaim-audit-expiry.test.ts
@@ -1,0 +1,280 @@
+/**
+ * QA-802 — Blob GC vs audit-log expiry (harper#708, "Blob GC: files not deleted on
+ * audit-log expiry when using RocksDB audit store").
+ *
+ * #708's claim: `scheduleAuditCleanup()` (resources/auditStore.ts) branches on the audit
+ * store type; for RocksDB it calls `rootStore.purgeLogs()` and returns WITHOUT ever calling
+ * `removeAuditEntry()` — the function that invokes the blob-file delete callbacks — so blob
+ * files supposedly accumulate forever once a blob-bearing record is deleted.
+ *
+ * SOURCE READ at SHA d112560b6 says the picture is more nuanced:
+ *  - RecordEncoder.ts recordUpdater() (used by BOTH SQL and REST/JS deletes, since Table.ts's
+ *    `_writeDelete` funnels through it) calls `deleteBlobsInObject(existingEntry.value, ...)`
+ *    SYNCHRONOUSLY on any delete whose prior entry carried HAS_BLOBS — independent of audit
+ *    engine and independent of auditRetention. This landed 2026-07-14 (commit 6e8d3647a,
+ *    fixing #641, an unrelated update-path data-loss bug) — AFTER #708 was filed
+ *    (2026-05-21) but it isn't excluded for type==='delete', so it may have incidentally
+ *    fixed #708's headline scenario too.
+ *  - Table.ts's OWN periodic `scheduleCleanup()` (armed here via `scanInterval` without
+ *    `expiration`, so live records never auto-evict) always treats RocksDB deleted-tombstone
+ *    removal as active (`removeDeletedRecords = !audit || isRocksDB`, Table.ts:5824) —
+ *    independent of the auditStore.ts RocksDB branch under test. That scan's `removeEntry()`
+ *    call operates on a null-value tombstone though, so it can't be the blob-deletion path
+ *    (blob deletion needs the entry's prior VALUE, already gone by tombstone time).
+ *  - QA-047 (audit-retention.test.ts) already established that RocksDB's `purgeLogs` purges
+ *    whole LOG FILES, not per-entry, and won't touch a still-active (not-yet-rolled) file
+ *    within a short test window — a confound distinct from the delete-callback question.
+ *
+ * This experiment tests empirically, on disk, rather than trusting either the issue's
+ * static-analysis premise or the RecordEncoder.ts read above:
+ *   Arm A: delete a live blob-bearing record via SQL DELETE (the issue's exact repro),
+ *          trajectory-sample the on-disk blob dir through the deletionDelay, the
+ *          auditRetention window, and a hard restart.
+ *   Arm B: audit-only expiry with the record still LIVE (repeated overwrite so old audit
+ *          PUT entries age out) — the current blob must never be touched; PUT-type audit
+ *          entries never invoke the delete-callback (auditStore.ts removeAuditEntry only
+ *          special-cases type==='delete'), so this is the negative control.
+ * ENGINE SCOPE — read before trusting this as an A/B. In CI this file pins the DEFAULT
+ * engine only (RocksDB); the engine is selected process-wide by HARPER_STORAGE_ENGINE, so a
+ * single test file cannot run both. The LMDB leg was run manually during QA and also showed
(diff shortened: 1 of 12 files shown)


integrationTests/database/blob-delete-reclaim-audit-expiry.test.ts
```diff
@@ -0,0 +1,287 @@
...
+			strictEqual(v.body.storedSha, expectedSha, 'DEFECT: live blob sha mismatch after audit-only expiry');
+
+			findings.push(
+				`Arm B [${ENGINE}] live record churned 6x then aged past audit retention: afterChurn=${afterChurn.files}f/${fmtKB(afterChurn.bytes)}, ` +
+					`afterExpiry=${afterExpiry.files}f/${fmtKB(afterExpiry.bytes)}, live blob intact=${v.body.match} (EXPECTED: survives, ~1 file)`
+			);
+			// A single live key should settle to ~1 file (the current blob); prior overwrite
+			// generations are reclaimed by the update-path retainedFileIds guard, not by audit expiry.
+			findings.push(
+				afterExpiry.files <= 2
```
cb1kenobi, line 275: **Nit: Arm B's "no orphan growth" invariant is logged, not asserted**

Arm B is the negative control that should prove audit-only expiry never disturbs the live blob's on-disk footprint. It does assert the live record survives byte-intact (`present`/`match`/`storedSha`), which is the important half. But the file-count check — the PR description's "must leave exactly one byte-intact file" — is only pushed into `findings`, never asserted. A regression where audit expiry began orphaning files (or the update-path `retainedFileIds` guard stopped reclaiming prior overwrite generations) would still pass Arm B silently, since `verify` only inspects the current key.

Consider asserting the settled count so the control can actually fail:

```
ok(
	afterExpiry.files <= 2,
	`Arm B [${ENGINE}]: audit-only expiry should leave <=2 blob files, got ${afterExpiry.files}`
);
```

The other two specs assert their disk-count invariants (`netLeakedFiles <= 0`, `survivedAfterCleanup === 0`); this one reads as the odd one out.

—
_Generated by Barber AI_

kriszyp:
Addressed all 11 open review threads:

- **claude bot / kriszyp** (blob-delete Arm A leak, txnlog k0 phantom oracle + surface controls, blob-drop k0 vacuous-pass): already fixed in the prior push (458dc46).
- **gemini-code-assist** (`waitReady` silent timeout, `cleanup_orphan_blobs`/`get_job` status not asserted, flaky elapsed-time upper bound): fixed.
- **kriszyp** (`survivedAfterCleanup` tightened to `=== 0` in the no-kill control too): fixed.
- **cb1kenobi / Barber AI** (Arm B's "no orphan growth" only logged, not asserted): fixed.

Independent pre-push review (codex) then flagged two more real gaps while re-reviewing the branch, which are also fixed in this push:
- `restGetById` treated any non-200 (including 5xx) as a clean absence, and `/Flush/`'s `flushed` field was never checked.
- QA-782's audit-purge contrast arm only read the audit log *after* the purge, so it could pass "clean" without ever warming the stale-cache the arm exists to catch; and the REST-QUERY/SEARCH-BY-VALUE/SQL surfaces only checked ID membership despite claiming byte-correctness.

One earlier independent-review pass (at 3c619a32f) got a full receipt; the final commit's re-review kept hitting the tool's timeout under heavy concurrent load on this host after several retries — the codex leg itself completed and re-surfaced only already-triaged design points (no new mechanical findings), but no fresh SHA-pinned receipt was produced. Flagging for whoever picks this up next in case a clean run is wanted before merge.

— Claude Sonnet 5

cb1kenobi (approved):

PR #1990, opened 2026-07-29
fix(storage): derive storage stats from quota-status.json when available

### LINKED ISSUE #1976: getStorageStats reports filesystem free space, which is wrong under a quota
## Summary

`Table.getStorageStats()` reports filesystem-level free space, which is not the space available to Harper whenever the data directory sits under a filesystem quota. The value is recorded into Harper's own analytics, so the disk metric Harper publishes can be wrong by orders of magnitude.

`resources/Table.ts:3710`:

```js
static getStorageStats() {
    const stats = fs.statfsSync(primaryStore.path);
    return {
        available: stats.bavail * stats.bsize,
        free: stats.bfree * stats.bsize,
        size: stats.blocks * stats.bsize,
    };
}
```

Consumer: `resources/analytics/write.ts:381`.

`statfs` describes the *filesystem*. Under a user, group, or project quota — or any setup where Harper's data directory is a subtree of a much larger volume — the numbers bear no relation to what Harper can actually write.

## Observed

On an instance whose data directory was quota-limited on a large shared volume, Harper's analytics recorded roughly **278 GB available** while the instance had **zero remaining allowance** and was failing every write with a quota error. Nothing in Harper's own metrics indicated storage pressure at any point before or during the failure.

## Same blindness in blob path allocation

`resources/blob.ts`, `createFrequencyTableForStoragePaths()` weights blob storage path selection by `statfs` available space:

```js
const availableSpace = stats.bavail * stats.bsize;
return Math.pow(availableSpace, 0.8);
```

So a path with no remaining quota still looks attractive and keeps receiving allocations. Lower impact than the metric (blob volumes were small in the case observed) but wrong for the same reason, and here it drives a decision rather than a report.

## Not affected

`storage_maxFreeSpaceToLoad` / `storage_maxFreeSpaceToRetain` are LMDB environment options (`utility/lmdb/OpenEnvironmentObject.ts`), not `statfs`-derived. Checked and excluded.

## Suggested direction

Make the metric honest. In rough order of usefulness:

1. Account for an applicable quota when one exists, so `available` means "available to Harper."
2. Failing that, accept an operator-supplied ceiling and report against it.
3. At minimum, scope the metric explicitly (name and document it as filesystem-level) so nobody builds alerting on it believing otherwise — and stop using raw `statfs` availability as an unqualified allocation input.

(1) is the most useful and the least portable; reading quotas from inside a container is not reliably available. (2) is cheap and sufficient for most of the value, since something outside Harper usually knows the limit.

---

🤖 Filed by Claude on behalf of @heskew

kriszyp:
## What

`Table.getStorageStats()` (feeding the `storage-volume` analytics metric) and
`blob.ts`'s storage-path weighting both derived available/free/size from raw
`statfs`. `statfs` describes the *filesystem*, which is wrong whenever the
data directory sits under a user/group/project quota, or is a subtree of a
much larger shared volume: analytics can report hundreds of GB "available"
while the instance has zero remaining quota and is failing every write.

## Why

Fixes #1976. On an observed instance, Harper's own storage metrics never
indicated pressure before or during a quota-exhaustion outage — `statfs`
kept reporting the filesystem's real free space, oblivious to the quota.

## How

Both call sites now go through a new `getStorageSpaceStats(path)` helper in
`server/storageReclamation.ts`, which prefers a fresh `quota-status.json`
(written by host-manager every ~90s) over `statfs` — the same
freshness-checked pattern `storageReclamation.ts` already used internally
for its own reclamation-trigger ratio, and that `getDiskInfo()` already
surfaces as metadata for `system_information`. The existing ratio
calculation is refactored to use the same helper, so there is a single
source of truth for the quota-vs-statfs choice.

- `resources/Table.ts` — `getStorageStats()` becomes a thin async delegate
  to `getStorageSpaceStats()`.
- `resources/analytics/write.ts` — `storeVolumeMetrics` awaits it.
- `resources/blob.ts` — `createFrequencyTableForStoragePaths` (which weights
  which storage path new blobs land on) uses it instead of raw `statfs`;
  directory creation for a not-yet-existing path is now explicit (it used to
  piggyback on a `statfs` ENOENT, which the quota branch never triggers).
- The returned stats carry a `basis: 'quota' | 'filesystem'` tag, so the
  analytics record itself is now self-describing about which source
  produced it.

## Test plan

- [x] `npm run typecheck` / `npm run build`
- [x] `unitTests/server/storageReclamation.test.js` — added coverage for
      `getStorageSpaceStats` (fresh quota, over-quota clamp, absent/stale
      quota → statfs fallback), all existing tests unaffected.
- [x] `unitTests/resources/blob.test.js` (63 tests) and
      `unitTests/resources/analytics/write.test.js` (36 tests) pass
      unchanged.
- [x] `npm run format:write` / `npm run lint:required` clean.
- Not run: full integration suite — this is a behavior-preserving refactor
  for the statfs branch (same formula) plus a new quota branch already unit
  tested at the source; see Risks below for what a reviewer may want to spot
  check.

## Risks / open questions

- `blob.ts`'s `createFrequencyTableForStoragePaths` (the storage-path
  weighting logic) has no dedicated unit test. It's internal/unexported and
  `blob.ts` has a module-scope `_assignPackageExport('createBlob', ...)`
  side effect that makes it unsafe to `rewire()` in isolation (pollutes
  `global.createBlob` for the rest of the test run — confirmed by a spurious
  failure in an unrelated `caching.test.js` suite while iterating on this).
  Coverage here relies on `getStorageSpaceStats` itself being fully tested
  plus the existing full `blob.test.js` suite staying green.
- Independent (outside-model) pre-push review did not complete: `codex`/
  `claude` review runs were still actively working (large log output, no
  errors) when killed by the available time budget on this shared,
  concurrently-loaded host. `independent-review: FAILED — reviewer CLI did
  not finish within the available window` — flagging for the shepherd to
  pick up rather than claiming coverage that didn't happen.

Refs #1976


kriszyp:
**Docs:** no companion documentation PR — this is a correctness fix to an existing internal metric (`storage-volume` analytics) and an internal blob-placement heuristic, not a new public surface. The one additive field (`basis: 'quota' | 'filesystem'`) is backward compatible and mirrors the existing `free_space_basis` annotation `system_information` already exposes.

— KrAIs (Claude)

server/storageReclamation.ts
```diff
@@ -70,15 +70,40 @@ export function onStorageReclamation(
...
+};
+
+/**
+ * Get available/free/size storage stats for a path, preferring a fresh quota-status.json (see
+ * getQuotaStatus) over statfs. statfs describes the filesystem, which is misleading whenever
+ * the data directory is quota-limited on a larger shared volume (#1976).
+ */
+export async function getStorageSpaceStats(path: string): Promise<StorageSpaceStats> {
 	const status = await getQuotaStatus();
 	if (status?.quotaBytes && Date.now() - status.updatedAt < QUOTA_STATUS_MAX_AGE_MS) {
```
kriszyp: Could we scope this quota branch to paths actually covered by the quota? `getStorageSpaceStats(path)` currently ignores `path` whenever `$ROOT/quota-status.json` is fresh. Harper supports explicit database/table storage paths and `STORAGE_PATH`, so a database on `/mnt/archive` would receive the root quota's size/free values, and `defaultGetAvailableSpaceRatio` would also ignore that volume's real pressure. Please either canonicalize and verify that `path` is within the quota-controlled storage root before using quota data, or make quota scope explicit in this API and use `statfs(path)` otherwise; a two-path regression test would pin this down.

server/storageReclamation.ts
```diff
@@ -70,15 +70,40 @@ export function onStorageReclamation(
...
-		return Math.max(0, status.quotaBytes - status.usedBytes) / status.quotaBytes;
+		const available = Math.max(0, status.quotaBytes - (status.usedBytes ?? 0));
```
kriszyp: `usedBytes` is required by `QuotaStatusData`, so defaulting a syntactically valid but incomplete record to zero turns bad producer data into "100% free." For example, if the quota is full but the file lacks `usedBytes`, this reports the entire quota available and keeps reclamation priority below 1. Please validate the parsed fields (finite timestamp and byte counts, positive quota, nonnegative usage) and fall back to `statfs` when they are invalid rather than fabricating zero usage; the test at `unitTests/server/storageReclamation.test.js:453` should assert that fallback.

resources/Table.ts
```diff
@@ -4568,12 +4568,7 @@ export function makeTable(options) {
 			);
 		}
 		static getStorageStats() {
```
kriszyp: This changes `getStorageStats()` from returning the stats object synchronously to returning a Promise. The generated table class is available through the exported table types and `scope.resources`, so custom code that destructures this result synchronously will now silently see missing fields. Since only analytics needs the quota-aware async I/O in-tree, can we preserve the existing method contract or introduce an explicitly async method so this fix does not become an avoidable API break?

unitTests/server/storageReclamation.test.js
```diff
@@ -483,6 +545,23 @@ describe('storageReclamation module', function () {
 				assert.ok(handler.calledOnce);
 			});
 
+			it('treats a zero-size result as ratio 0 (not NaN), so a full disk still triggers reclamation', async function () {
+				storageReclamation.__set__(
```
kriszyp: The current test policy asks new unit coverage to use real modules with `node:assert` and not add rewire/sinon injections. Could this exercise public behavior with a real input, or move the zero-size ratio calculation into a small directly testable function, instead of adding a new `__set__` dependency?

kriszyp (commented):
🤖 Dispatch codex/gpt-5.6-sol review (verdict CHANGES).

resources/analytics/write.ts
```diff
@@ -397,11 +397,11 @@ function storeDBSizeMetrics(analyticsTable: Table, databases: Databases) {
 	}
 }
 
-function storeVolumeMetrics(analyticsTable: Table, databases: Databases) {
+async function storeVolumeMetrics(analyticsTable: Table, databases: Databases) {
 	for (const [db, tables] of Object.entries(databases)) {
 		try {
 			const [firstTable] = Object.values(tables);
-			const storageStats = firstTable?.getStorageStats();
+			const storageStats = await firstTable?.getStorageStats();
```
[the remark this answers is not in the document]
kriszyp: Parallelized with `Promise.all`/`map` as suggested (48a3304fb..d32cfb7fc). `unitTests/resources/analytics/write.test.js` (36 tests) still green.

— KrAIs (Claude)

server/storageReclamation.ts
kriszyp: Fixed — `getStorageSpaceStats` now checks `isPathWithinRoot(rootPath, path)` (symlink-resolved) before trusting quota data, falling back to per-path `statfs` otherwise. New tests cover a path outside root, a lexically-nested-but-symlinked path, and the Kubernetes `..data` edge case.

— KrAIs (Claude)

server/storageReclamation.ts
kriszyp: Fixed — added `isValidQuotaStatus` to validate the parsed record (finite timestamp/byte counts, positive quota, non-negative usage) before trusting it; an invalid record now falls back to `statfs` instead of being read as fabricated zero usage. Covered by the new 'falls back to statfs when usedBytes is missing' test at storageReclamation.test.js:453.

— KrAIs (Claude)

resources/Table.ts
kriszyp: Kris confirmed on the dispatch that this is fine: `getStorageStats()` was never documented, so the async signature change is authorized rather than an accidental API break. Leaving as-is.

— KrAIs (Claude)

unitTests/server/storageReclamation.test.js
kriszyp: Addressed — `computeAvailableRatio` is now a plain exported function, and the zero-size test calls it directly with real inputs (`assert.equal(computeAvailableRatio(0, 0), 0)`), no `__set__`/rewire injection needed for that assertion.

— KrAIs (Claude)

cb1kenobi (approved):
LGTM

PR #1991, opened 2026-07-29
Prevent concurrent component installs from corrupting dependencies

### LINKED ISSUE #1973: Concurrent installs of the same component can destroy its node_modules
## Summary

Nothing serializes dependency installs for a component. When two installs of the **same** component overlap, npm's reify tears down a tree the other is building, fails with `ENOTEMPTY`, and aborts mid-teardown — leaving the component with **no `node_modules` at all**.

Observed on 5.1.x on a two-node hosted cluster. Three `deploy_component` attempts for one component landed within 13 minutes; the first two exited 0, the third died:

```
npm error code ENOTEMPTY
npm error syscall rmdir
npm error path /home/harperdb/harper/components/<component>/node_modules/node-llama-cpp/dist
npm error errno -39
npm error ENOTEMPTY: directory not empty, rmdir '.../node_modules/node-llama-cpp/dist'
```

End state: `package-lock.json` written, `node_modules` gone. The component kept running only because the worker still had the modules loaded in memory — the loss surfaced on the next restart, days later.

## Why nothing prevents it

- `installApplications()` (`components/Application.ts`) pushes every app into `applicationInstallationPromises` and awaits `Promise.allSettled`, so installs run concurrently by design.
- `deployLifecycle` (`components/deployLifecycle.ts`) ref-counts in-flight deploys **per component name** and exists to quiet file watchers. It deliberately allows overlap for the same name (`0→1` fires start, `1→0` fires end, intermediate transitions are silent), so it is not — and was never intended as — a mutex around installs.
- Operation-driven deploys each call `prepareApplication()` independently, so two calls for the same component have nothing in common to contend on.

Retrying a deploy that appears stuck is the natural operator reaction, which makes this easy to trigger.

## Secondary: the failed install leaks a child process

After the failure the container held a `[npm install] <defunct>` zombie, still unreaped ~24 hours later. `nonInteractiveSpawn` rejects on timeout via `childProcess.kill()`; something on the abort path isn't reaping the child.

## Suggested direction

Serialize installs per component directory — a mutex or queue keyed on `application.dirPath`, so a second install of the same component waits rather than racing. A failed install should also not be able to leave a component with fewer dependencies than it started with; staging into a temp dir and swapping, or restoring on failure, would make the operation non-destructive.

Related: #1974 (a destroyed install can never recover) and #1975 (it reports healthy while broken). This bug creates the damage; those two explain why it persisted unnoticed.

---

🤖 Filed by Claude on behalf of @heskew

kriszyp:
Serializes component extraction and dependency installation across Harper worker threads, including the deprecated direct npm-install path. Installer process trees are now terminated on timeout, worker exit, and Harper shutdown; failed boot preparations remain retryable instead of being written to `harper-application-lock.json`.

Reviewer attention: filesystem lock ownership/recovery in `componentPreparationLock.ts` and process-tree lifecycle registration in `manageThreads.js`. Generated by Codex; the standard pre-PR review completed with Claude and Gemini, while the post-fix final-artifact Claude check ended with an internal execution error.

Closes #1973


server/threads/manageThreads.js
```diff
@@ -849,6 +854,55 @@ function onThreadExit(listener) {
...
+
+function unregisterProcessGroup(processGroupId) {
+	if (isMainThread) removeProcessGroup(threadId, processGroupId);
+	else parentPort?.postMessage({ type: UNREGISTER_PROCESS_GROUP, processGroupId });
+}
+
+if (isMainThread) {
+	process.on('exit', () => {
+		for (const ownerThreadId of processGroupsByThread.keys()) terminateProcessGroupsForThread(ownerThreadId);
+	});
```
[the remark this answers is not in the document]
kriszyp: Addressed in b0bb17086: cleanup now iterates over a snapshot of the owner-thread keys.

components/Application.ts
```diff
@@ -1166,9 +1192,11 @@ export async function installApplications() {
...
 			}
 
-			applicationInstallationPromises.push(prepareApplication(application));
-
-			harperApplicationLock.applications[name] = applicationConfig;
+			applicationInstallationPromises.push(
+				prepareApplication(application).then(() => {
+					harperApplicationLock.applications[name] = applicationConfig;
+				})
+			);
```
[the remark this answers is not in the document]
kriszyp: Addressed in b0bb17086: boot preparation failures are logged with errorForLog and then rethrown, while the lock entry remains unrecorded.

components/Application.ts
```diff
@@ -1057,22 +1064,41 @@ export function derivePackageIdentifier(packageIdentifier: string) {
...
+					}
+					await installApplication(application);
+				} finally {
+					await application.cleanupTransientNpmrc();
+				}
+			},
+			{
+				// A package-reference preparation can run one command to pack and another to install.
+				// Bound orphaned same-process worker locks without rejecting behind a valid holder.
+				timeoutMs: 2 * commandTimeoutMs + COMPONENT_PREPARATION_WAIT_MARGIN_MS,
```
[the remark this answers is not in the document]
kriszyp: Addressed in 38b484ce0: the waiter bound now covers the four possible git extraction commands plus the two-command package-manager fallback and margin.

components/componentPreparationLock.ts
```diff
@@ -0,0 +1,136 @@
...
+export async function withComponentPreparationLock<T>(
+	componentDirPath: string,
+	prepare: () => Promise<T>,
+	options: ComponentPreparationLockOptions = {}
+): Promise<T> {
+	const release = await acquireComponentPreparationLock(componentDirPath, options);
+	try {
+		return await prepare();
+	} finally {
+		await release();
```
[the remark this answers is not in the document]
kriszyp: Addressed in 38b484ce0: the preparation error remains primary when release also fails; the release failure is reported best-effort, with a regression test that also verifies a throwing reporter cannot mask the original error.

server/threads/manageThreads.js
```diff
@@ -849,6 +855,60 @@ function onThreadExit(listener) {
...
+
+function unregisterProcessGroup(processGroupId) {
+	if (isMainThread) removeProcessGroup(threadId, processGroupId);
+	else parentPort?.postMessage({ type: UNREGISTER_PROCESS_GROUP, processGroupId });
+}
+
+async function isThreadRunning(ownerThreadId) {
+	if (ownerThreadId === threadId || ownerThreadId === 0) return true;
+	return (await getThreadInfo()).some((worker) => worker.threadId === ownerThreadId);
+}
```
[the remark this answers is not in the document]
kriszyp: Addressed in fa950945c: worker thread-info requests now have a 1s timeout, reject with ERR_THREAD_INFO_TIMEOUT, and remove their response listener on both timeout and reply. Added real worker tests for both a silent parent and a successful parent reply.

components/Application.ts
```diff
@@ -1057,22 +1066,51 @@ export function derivePackageIdentifier(packageIdentifier: string) {
...
-			// already gone by the time the component's dependency tree (and any install script it is
-			// allowed to run) executes.
-			await application.startGitCredentialSession();
-			await extractApplication(application);
-		} finally {
-			await application.cleanupGitCredentialSession();
-		}
-		await installApplication(application);
+		const commandTimeoutMs = application.install?.timeout ?? DEFAULT_COMMAND_TIMEOUT_MS;
+		await withComponentPreparationLock(
```
kriszyp, line 1070: Could we move the serialization boundary up to cover the complete per-project deploy transaction (using a separate coordinator if we do not want to hold the filesystem lock during replication)? Today `deployComponent` writes the package config at `components/operations.js:416` before reaching this lock and replicates only after this lock is released. A concrete race is: deploy A writes config A and pauses in recorder/payload/credential work; deploy B writes config B, acquires this lock, and installs B; then A acquires it and installs A. Both calls can succeed with component A on disk but config B in YAML, and their later replication order can leave peers with yet another final config/version. This prevents simultaneous `node_modules` writes, but not the deeper per-component ordering bug. Please serialize config mutation, preparation/validation, and replication consistently (or introduce an explicit ordered commit/version protocol), and add a two-package concurrent deploy test that asserts disk, config, and peer order agree.

components/Application.ts
```diff
@@ -1540,6 +1632,102 @@ function spawnWithEnv(
...
+	} catch (error: any) {
+		if (error.code !== 'ESRCH') throw error;
+	}
+	if (!(await waitForProcessGroupExit(processGroupId, PROCESS_TERMINATION_GRACE_MS))) {
+		try {
+			process.kill(-processGroupId, 'SIGKILL');
+		} catch (error: any) {
+			if (error.code !== 'ESRCH') throw error;
+		}
+		await waitForProcessGroupExit(processGroupId, PROCESS_TERMINATION_GRACE_MS);
```
kriszyp: Please do not let the timeout settle—and therefore let `withComponentPreparationLock` release its lock—unless process-tree exit was actually confirmed. The result of this second `waitForProcessGroupExit` is ignored; if a process remains alive after SIGKILL (for example while stuck in uninterruptible filesystem I/O), `waitForProcessClose` can time out and this function still resolves, so the command rejects and the lock is deleted while a descendant may later resume mutating `node_modules`. The Windows path has the same invariant gap when `taskkill` fails and falls back to killing only the parent. Please propagate an explicit “tree still alive” state that retains/poisons the lock or continue waiting for confirmed exit, and test the unsuccessful-termination path.

unitTests/components/installApplicationsLock.test.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+
+const assert = require('node:assert');
+const { mkdtemp, mkdir, readFile, rm, writeFile } = require('node:fs/promises');
+const { tmpdir } = require('node:os');
+const { join } = require('node:path');
+const rewire = require('rewire');
```
kriszyp: Please rewrite this without `rewire`/`sinon`. New Harper tests are required to use plain `node:assert` against real module boundaries and must not add either dependency; when isolation is needed, the production seam should be made explicitly injectable or the behavior factored into a testable helper. This test adds both prohibited patterns in a new file.

kriszyp (commented):
Final self-review identified two implementation hardening items, addressed in 07a6de19f, plus one separate cross-node deployment-ordering concern that requires its own protocol rather than widening the local install mutex.

components/Application.ts
kriszyp: Addressed in 07a6de19f. After SIGKILL, POSIX cleanup now waits without a deadline until the process group is positively absent. Windows retries taskkill /T /F until tree termination is confirmed instead of falling back to killing only the parent. The spawn promise therefore cannot settle—and the preparation lock cannot release—while termination remains unconfirmed. Added a regression test for the unsuccessful-termination state.

components/Application.ts
kriszyp: Valid, but this is a separate pre-existing deploy-transaction ordering problem rather than the #1973 install race. Extending this filesystem mutex across replicateOperation would introduce a distributed deadlock when two nodes originate a same-component deploy concurrently: each origin would hold its local mutex while waiting for the other node to apply the replicated operation. A correct fix needs an explicit cross-node deployment generation/ordering protocol, not a wider local critical section. This PR intentionally keeps the invariant scoped to the destructive extract/install transaction that caused node_modules loss.

unitTests/components/installApplicationsLock.test.js
kriszyp: Addressed in 07a6de19f. The new test no longer imports rewire or sinon. The application-lock transition is factored into an explicitly injected production helper and tested with node:assert and a plain failing callback.

components/componentPreparationLock.ts
```diff
@@ -0,0 +1,186 @@
...
+		} catch {
+			// An unknown owner state is not permission to remove its lock.
+			return false;
+		}
+	}
+	if (!ownerExited) return false;
+
+	const stalePath = `${lockPath}.stale-${randomUUID()}`;
+	try {
+		await rename(lockPath, stalePath);
```
kriszyp: Could we make stale reclamation atomic with respect to a new acquisition? Two waiters can both read the dead owner; waiter A renames/removes it and acquires a fresh lock, then waiter B executes the rename it already authorized and moves A's new lock to its stale path. B can then acquire too, so both preparations mutate the directory concurrently; the release token only detects the loss after the damage. Please serialize reclamation with acquisition (or use a primitive that couples owner comparison and removal), and add a multi-contender dead-owner test that asserts maximum preparation concurrency stays at one.

components/componentPreparationLock.ts
```diff
@@ -0,0 +1,186 @@
...
+	}
+}
+
+async function removeStaleLock(lockPath: string, options: ComponentPreparationLockOptions): Promise<boolean> {
+	const owner = await readOwner(lockPath);
+	if (!owner) return false;
+	// All worker threads share a PID. Treat our own PID as authoritative instead of probing it:
+	// process.kill is commonly wrapped by embedders/test harnesses, and a false negative here would
+	// permit exactly the concurrent mutation this lock exists to prevent.
+	let ownerExited = owner.pid !== process.pid && !isProcessAlive(owner.pid);
```
kriszyp: Please include a process-instance identity in the owner record rather than treating PID/thread ID as sufficient liveness. After a crash, a container commonly starts Harper again as PID 1; a main-thread lock then has the current PID and thread 0, `isThreadRunning(0)` returns true, and the deadline is renewed forever. Worker thread IDs can also be reused after process restart. A per-process-start nonce lets the new Harper instance immediately reject an old same-PID owner; please cover a persisted lock with the current PID/thread ID but an earlier instance token.

components/Application.ts
```diff
@@ -1540,6 +1645,114 @@ function spawnWithEnv(
...
+	for (;;) {
+		const terminated = await new Promise<boolean>((resolve) => {
+			const taskkill = spawn('taskkill', ['/pid', String(childProcess.pid), '/T', '/F'], {
+				stdio: 'ignore',
+				windowsHide: true,
+			});
+			taskkill.once('close', (code) => resolve(code === 0));
+			taskkill.once('error', () => resolve(false));
+		});
+		if (terminated) return;
```
kriszyp: On Windows this loop never completes when `taskkill` returns nonzero because the target exited just before `/PID` was evaluated (the normal "process not found" timeout race). The command promise then never rejects and the component lock remains held indefinitely. Please give this path an independently confirmable "tree is gone" state (or a process-lifetime mechanism such as a job object) and distinguish that from a genuine termination failure instead of retrying every nonzero result forever; an injected/Windows test for an already-exited child at timeout would pin the behavior.

kriszyp (commented):
Final lock audit findings addressed in 59d4cc31e: immutable bakery tickets remove stale-owner CAS races, process-start identity handles PID reuse, and Windows distinguishes taskkill failure from an independently confirmed dead tree.

components/componentPreparationLock.ts
kriszyp: Addressed in 59d4cc31e. Owner records now include a process-start nonce shared with workers but regenerated by every main-thread process start, including an inherited environment after container restart. A persisted current-PID/thread-0 ticket carrying the prior instance nonce is immediately ignored; the regression test covers that PID-1-style reuse case.

components/Application.ts
kriszyp: Addressed in 59d4cc31e. A nonzero taskkill result now triggers an independent Win32_Process tree query through PowerShell. Cleanup resolves only when taskkill confirms /T termination or the process table confirms neither the root nor any descendant remains; query/termination failures keep the lock held and retry. The injected regression test covers the already-exited taskkill race.

components/componentPreparationLock.ts
kriszyp: Addressed in 59d4cc31e. The single reusable lock pathname is replaced with a filesystem Lamport bakery lock: each contender atomically publishes a fully written, immutable, uniquely named choosing claim and ticket. Stale cleanup removes only that exact never-reused claim path, so it cannot rename or unlink a fresh owner's ticket. A two-worker dead-owner test holds the winner and asserts maximum concurrent preparation remains one before allowing the waiter through.

components/componentPreparationLock.ts
```diff
@@ -0,0 +1,253 @@
...
+		if (error.code === 'ENOENT' || error instanceof SyntaxError) return null;
+		throw error;
+	}
+}
+
+async function ownerIsAlive(
+	owner: ComponentPreparationLockOwner,
+	options: ComponentPreparationLockOptions
+): Promise<boolean> {
+	if (owner.pid !== process.pid) return isProcessAlive(owner.pid);
```
Ethan-Arrowood, line 76: Foreign-PID recycling can renew the deadline forever: for an owner from another process, liveness is bare `kill(pid, 0)` (EPERM reads as alive), and the process-instance nonce is only consulted on the same-PID branch. If Harper dies hard and the OS recycles that PID to a long-lived daemon, every subsequent deploy of the component waits and renews indefinitely until an operator deletes `.component-preparation-locks`. Suggestion: on the foreign-PID branch, don't renew the deadline (let the bounded backstop throw without stealing), or strengthen identity with process start time. Non-blocking.

components/Application.ts
```diff
@@ -1482,11 +1545,39 @@ function spawnWithEnv(
 			cwd,
 			env,
 			stdio: ['ignore', 'pipe', 'pipe'],
+			// A dedicated POSIX process group lets a timeout terminate npm plus every reify worker
+			// and install-script descendant before the component preparation lock is released.
+			detached: process.platform !== 'win32',
```
Ethan-Arrowood, line 1585: `detached: true` means an npm tree survives a SIGKILL/OOM-kill of Harper itself (the `process.on('exit')` sweep only runs on graceful exit). On restart, boot sees the dead owner pid, reclaims the ticket, and starts a new install while the orphaned npm tree may still be mutating `node_modules` — the corruption class this PR fixes, reachable via hard death mid-install plus fast restart. Not a regression, but the ticket already records the owner; recording the spawned process-group id too would let reclaim kill the orphan group before deleting the stale ticket. Non-blocking.

components/componentPreparationLock.ts
```diff
@@ -0,0 +1,253 @@
...
+	// A PID (especially container PID 1) and worker thread id can be reused after a restart. The
+	// per-process nonce makes an owner from an earlier Harper instance unambiguously stale.
+	if (owner.processInstanceId !== COMPONENT_PREPARATION_PROCESS_INSTANCE_ID) return false;
+	if (!options.isOwnerAlive) return true;
+	try {
+		return await options.isOwnerAlive(owner);
+	} catch {
+		// An unknown owner state is not permission to ignore its ticket.
```
Ethan-Arrowood, line 84: `isOwnerAlive` rejection converts to "alive" at the deadline, renewing by the full `timeoutMs`. Conservative-keep is right for claim removal, but at the deadline site DESIGN.md's own rule ("the bounded wait remains a backstop when owner liveness cannot be established") argues for letting the timeout throw — it fails only the waiting request, never steals.

components/componentPreparationLock.ts
```diff
@@ -0,0 +1,253 @@
...
+	const leftTicket = left.ticket ?? Number.MAX_SAFE_INTEGER;
+	const rightTicket = right.ticket ?? Number.MAX_SAFE_INTEGER;
+	return leftTicket < rightTicket || (leftTicket === rightTicket && left.token < right.token);
+}
+
+async function publishClaim(claimPath: string, owner: ComponentPreparationLockOwner): Promise<void> {
+	// Keep the staging name outside the public `<lockName>.choosing|ticket.*` namespace so a
+	// contender cannot observe or clean it while its JSON write is still in progress.
+	const publishingPath = join(dirname(claimPath), `.${owner.token}.${randomUUID()}.publishing`);
+	await writeFile(publishingPath, JSON.stringify(owner), { flag: 'wx', mode: 0o600 });
```
Ethan-Arrowood, line 222: A crash between `writeFile` and `rename` leaks one `.publishing` staging file forever — `scanLiveClaims` filters only `choosing`/`ticket` prefixes. Cosmetic (one file per crash-during-publish), but a startup sweep of `*.publishing` would keep the lock dir tidy.

Ethan-Arrowood (approved):
Approving — the bakery-lock serialization is solid and fully supported by tests, and everything stays on the cold install path. Leaving a few non-blocking findings inline: two edge-case gaps around stale-owner/orphan handling worth a follow-up, and two smaller cleanups.

Also: the deferred deploy-transaction ordering issue raised in-thread is real and well-characterized now — please file a tracking issue before this merges so it doesn't evaporate.

kriszyp:
🤖 Note on 3 missing review threads: while clearing pending review state on this PR I deleted a pending (unsubmitted) self-review that, on closer inspection, actually contained 3 live inline comments — not stale draft text as its summary body suggested. Recreating them here since GitHub doesn't let me restore the original threads:

**`components/Application.ts` (~L1771)** — *"Could we avoid using the direct child's exit state as proof that the whole POSIX process group is gone? A custom installer can spawn and unref a descendant that inherits stdout/stderr, then exit. At timeout `exitCode` is already set while `process.kill(-pid, 0)` still succeeds; this branch waits five seconds, destroys the pipes, rejects, and lets `prepareApplication` release the component lock while that descendant can keep mutating `node_modules`. Please probe/terminate the process group even when the direct child has exited, and add a regression where the parent exits before timeout but its child continues writing."*
→ Addressed in ff613c89e: `terminateProcessTree` now probes the process group itself instead of trusting the direct child's exitCode/signalCode.

**`server/threads/manageThreads.js` (~L903)** — *"Please keep a dead worker's preparation claim non-reclaimable until its tracked process groups are confirmed gone. `process.kill(..., 'SIGKILL')` queues the signal and returns; immediately afterward `isThreadRunning()` reports false, so another contender can delete the worker's ticket and enter preparation before termination has been observed. That reopens the concurrent-writer window this PR is intended to close. Please coordinate process-group termination state with owner liveness (or otherwise delay reclamation until the group no longer exists), and test that a replacement preparation cannot enter until the old writer has stopped."*
→ Addressed in ff613c89e: a dead worker's lock claim is no longer reclaimable until its tracked process groups are confirmed terminated, including a Linux zombie-state check so a group whose owning worker's event loop is gone doesn't wedge reclamation forever.

**`DESIGN.md` (~L130)** — *"Small documentation drift from the latest reclamation rewrite: `publishClaim()` now publishes the completed record with `rename()`, not a hard link. Could we update this sentence so the design note matches the implementation?"*
→ Addressed in ff613c89e: DESIGN.md now says `rename()`.

All three were already fixed before I found the deletion mistake; only the GitHub thread objects themselves are unrecoverable, not the underlying findings or their resolution.

components/componentPreparationLock.ts
kriszyp: Addressed in cb42d42c2: added ownerLivenessConfirmed, used only at the deadline-renewal site. A bare kill(pid, 0) on a foreign PID no longer renews the deadline — it can't distinguish the original owner from an unrelated process that reused a recycled PID, so that case now falls through to the bounded timeout (which only fails this waiter, never steals the lock). ownerIsAlive's conservative "assume alive" is untouched for claim removal, where that's still the correct default.

components/componentPreparationLock.ts
kriszyp: Addressed in cb42d42c2, same fix as the foreign-PID comment: a rejected isOwnerAlive() is no longer treated as "alive" at the deadline site — it now falls through to the bounded timeout instead of renewing forever.

components/componentPreparationLock.ts
kriszyp: Addressed in cb42d42c2: acquireComponentPreparationLock now sweeps .publishing files older than 60s (well beyond publishClaim's write+rename window, so it can't race a genuinely in-flight publish) before starting its own claim.

components/Application.ts
kriszyp: Valid gap, but tracking the spawned process-group id in the ticket and having reclaim kill it before deleting a stale ticket is a real structural change (persisting pgid through the lock, and having reclaim on a *different* process own killing a group it never spawned), not a quick addition here. It's also adjacent to a related independent-review finding on this same diff: process-group registration is currently a fire-and-forget worker-to-main postMessage with no ack, and non-Linux zombie detection can hang isThreadRunning() outside Linux — both point at the same deeper fix (routing spawns through the main thread so it durably owns registration and reaping, which would also let reclaim persist and act on pgids across a hard process restart). Filing this as follow-up work rather than bolting it on under review-cycle time pressure; not a regression from this PR, and this PR's own crash-safety invariant (never release the lock while a probed process group is still alive) still holds within a single process lifetime.

kriszyp:
Filed #1996 to track the cross-node deploy-transaction ordering gap (config write → install → replication not serialized across concurrent same-component deploys) per @Ethan-Arrowood's request — scoped separately from this PR's node_modules corruption fix (#1973).

All other inline review threads are resolved (including the 3 recreated ones from the earlier pending-review mishap). No other open review feedback found.

🤖 KrAIs (Claude)

PR #1992, opened 2026-07-29
chore: migrate to @harperfast/code-guidelines

Ethan-Arrowood:
`@harperdb/code-guidelines` has been deprecated and republished as `@harperfast/code-guidelines`. Version 0.1.2 is functionally identical to 0.0.6 — no config changes, just the new package name (0.1.2 also adds a `files` allowlist so `.claude` and `docs` are no longer included in the published artifact).

This updates the dependency, any config references (prettier/eslint/docs), and the lockfile. No new lockfiles were added.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -96,7 +96,7 @@
 				"harper": "dist/bin/harper.js"
 			},
 			"devDependencies": {
-				"@harperdb/code-guidelines": "^0.0.6",
+				"@harperfast/code-guidelines": "^0.1.1",
 				"@harperfast/integration-testing": "^0.6.2",
 				"@modelcontextprotocol/sdk": "^1.29.0",
 				"@types/busboy": "^1.5.4",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -2525,9 +2557,6 @@
 			"cpu": [
 				"arm64"
 			],
-			"libc": [
-				"glibc"
-			],
 			"license": "Apache-2.0",
```
kriszyp, line 2563: This lock regeneration removed the `libc` discriminator from all four top-level Linux RocksDB binding entries. In a minimal install using this head-side metadata, `npm ci` on glibc installed both `@harperfast/rocksdb-js-linux-x64-glibc` and `@harperfast/rocksdb-js-linux-x64-musl`; the packages are about 15 MB each, so every Linux install/image carries an unused binding (roughly 16 MB extra on x64). `build-tools/build.sh` then turns this graph into the published shrinkwrap, preserving the missing selectors. Could we keep the non-migration lock entries identical to `main`, or restore their `libc: ["glibc"]` / `libc: ["musl"]` fields when regenerating the lock?

kriszyp (changes requested):
package-lock.json is such a source of joy. good luck. everything else looks fine.

package-lock.json
kriszyp, line 2563: This lock regeneration removed the `libc` discriminator from all four top-level Linux RocksDB binding entries. In a minimal install using this head-side metadata, `npm ci` on glibc installed both `@harperfast/rocksdb-js-linux-x64-glibc` and `@harperfast/rocksdb-js-linux-x64-musl`; the packages are about 15 MB each, so every Linux install/image carries an unused binding (roughly 16 MB extra on x64). `build-tools/build.sh` then turns this graph into the published shrinkwrap, preserving the missing selectors. Could we keep the non-migration lock entries identical to `main`, or restore their `libc: ["glibc"]` / `libc: ["musl"]` fields when regenerating the lock?

Ethan-Arrowood: Good catch — root cause was my local npm 10.9.3, which doesn't write the `libc` field that npm 11 adds for platform-specific optional deps, so regenerating the lock silently stripped the glibc/musl discriminators from the RocksDB binding entries. Fixed in 38da35e89: I reset the lock to main and regenerated with npm 11 (`npx npm@11 install --package-lock-only`). All `libc` fields are preserved and the diff is now scoped entirely to the `@harperfast/code-guidelines` subtree. Same fix applied to harper-pro#629, which had the identical issue.

kriszyp (approved):
good.

PR #1995, opened 2026-07-29
fix(resources): drain in-flight source-populated cache writes before dropTable() drops column families

kriszyp:
Human-Review-Need: 4 @ bf7428329d8f
## What / why

Root-causes and fixes the `Error: Put failed: Invalid argument: Invalid column family specified in write batch` flake reddening `Unit Test` on `main` (observed concentrated in `unitTests/resources/vectorIndexFormat.test.js`, e.g. run 30452497108 on commit 74224938f).

**Root cause.** `Resource.get()` against a `sourcedFrom` table resolves to its caller as soon as the source responds — the resulting cache write to the table's primary store commits **in the background** afterward (a deliberate latency optimization in `getFromSource()`, documented in the existing code: *"we don't want to wait for the transaction... let the transaction commit in the background"*). `Table.dropTable()` had no awareness of this: it would drop the table's RocksDB column families while one of these background writes was still landing. RocksDB rejects the still-open write batch with exactly this error, which can also abort `dropTable()` itself **before** it removes the tombstoned catalog rows — leaving the table stuck `dropping` for `completeInterruptedDrop` (in `resources/databases.ts`) to retry, and fail identically, on every subsequent `resetDatabases()` call for the rest of the process's life. That perpetual retry-and-fail cycle is what corrupts the shared internal catalog store, which is why an *unrelated* later table's write (the `attributesDbi.put(NEXT_TABLE_ID, ...)` inside `vectorIndexFormat.test.js`) ends up failing with the same error.

**Same defect as #1381, trigger identified.** #1381 diagnosed the *retry* symptom (`completeInterruptedDrop` failing on `removeSync`) but not *why* the drop was interrupted in the first place. I traced it to a concrete, reproducible trigger: `unitTests/resources/Resource-get-context.test.js`'s `after()` hook calls `dropTable()` immediately after its `it()` blocks' `sourcedFrom` `get()` calls — racing `dropTable()` against those calls' still-in-flight background cache writes. Confirmed directly in the CI log for the dispatched run: that file's `"after all" hook` fails, immediately followed by four `Error committing cache update` / `Could not access column family N` lines (one per earlier `it()` block's background write resuming after the column families were already gone), immediately followed by `Completing interrupted drop of table data.TestTableForContext` — the exact chain from trigger to #1381's retry loop to this dispatch's `vectorIndexFormat.test.js` symptom.

I also instrumented the real commit path locally and captured the race directly (not just inferred it): a `dropTable-start` timestamp landing *before* a preceding `get()` call's `commit-success` timestamp, in the unmodified test suite, on Node 26.

## The fix (`resources/Table.ts`)

- `dropTable()` now sets a per-table `droppingTable` flag **before** removing the table from the schema, and drains a `pendingSourceCommits` set of every `getFromSource()` transaction promise before touching a single column family (moved ahead of the existing blob-cleanup sweep too, so a drain-caught write's blobs aren't orphaned).
- `getFromSource()` treats `droppingTable` as `noCacheStore`, both at admission and **re-checked live** right before staging the write (after the async embed-hook step) — a call admitted before the drop started, or one still inside its own async prep, still returns fresh data but never starts/stages a write into a table that's going away.
- The drain is bounded by the existing `LOCK_TIMEOUT` (10s) and **fails closed**: if it times out, `dropTable()` throws rather than proceeding to drop anyway (which would silently reopen the exact race, just less often). The drop's tombstone is already durable at that point, so `completeInterruptedDrop` retries the whole drop on the next load — the same fail-and-retry shape the codebase already uses for a genuine drop failure.

## Test (`unitTests/resources/Resource-get-context.test.js`)

Two new deterministic regression tests (no timing-based "loop until it flakes" logic):
1. Gates the table's `@embed` pre-commit hook (via `setEmbedAttribute`) behind a promise the test controls. This hook runs *after* the GET has already resolved to its caller, so it pins down the exact post-resolution/pre-staging window the bug lives in. Fails before the fix, passes after.
2. Proves a `get()` admitted *while* `dropTable()` is draining still returns fresh source data but never caches it (closes the admission gap independently).

The drain's fail-closed timeout path is verified manually (a source that never resolves; `dropTable()` rejects with a "timed out" message at ~10s, no column family touched) rather than by an automated test — `sinon` fake timers over this code path reliably hung the test runner instead of advancing it, and exercising the real 10s wait would slow the suite. Noted in the test file.

## Known limitation — not fixed here, needs a separate task

The independent cross-model review (Codex, twice; Grok, once — see "Review" in the dispatch record) consistently flagged that `pendingSourceCommits`/`droppingTable` are **local to one `makeTable()` closure**, i.e. per-worker. In a multi-worker deployment, worker B can have a cache write staged while the main/ops thread runs `drop_table`, sees *its own* empty local set, and drops the column families anyway — B's write then resumes against the dropped handle and reproduces the original failure. A related, narrower gap: if the drain's timeout fires (a write staged and stalled past 10s), the durable tombstone is left in place, and an immediate same-name recreate could trigger `completeInterruptedDrop` to drop the column family while that stalled write is still trying to land.

Both are real and worth fixing, but they require actual cross-worker coordination (a broadcast prepare-drop/drain/ack protocol, and threading that state into `completeInterruptedDrop`/`table()`'s create path too) — a distinctly larger, separate architectural task, not a safe thing to bolt on under review pressure in this PR. This PR closes the single-process race that is what's actually failing in CI (mocha runs these tests in one process); I've filed the multi-worker gap as a Finding for follow-up.

Refs #1381.

### DIFF
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -1372,19 +1372,29 @@ export function makeTable(options) {
 			// the table stuck "dropping" for completeInterruptedDrop to retry (and fail
 			// identically) on every subsequent load (harper#1381). Drain any in-flight commits
 			// before the blob sweep below (so it observes every row a drain-caught write just
-			// committed) and before touching a single column family. Bounded: the tracked
-			// promise covers the whole source round-trip (see getFromSource), so a hung or
-			// slow source could otherwise wedge this drop forever.
+			// committed) and before touching a single column family.
+			//
+			// Bounded, and fails CLOSED: the tracked promise covers the whole source round-trip
+			// plus the local commit (see getFromSource), so a hung/slow source or a slow commit
+			// (e.g. a large blob write) could otherwise wedge this drop forever. Rather than
+			// give up and drop anyway - which would reopen exactly the race this drain exists to
+			// close, just less often - a timeout FAILS the drop. The tombstone written above is
+			// already durable, so completeInterruptedDrop picks the drop back up on the next
+			// load, once the stuck write has had time to finish.
 			if (pendingSourceCommits.size) {
 				const pending = [...pendingSourceCommits];
-				let timedOut = false;
-				await Promise.race([
+				let timer: NodeJS.Timeout;
+				const timedOut = Symbol('timedOut');
+				const result = await Promise.race([
 					Promise.allSettled(pending),
-					new Promise<void>((resolve) => setTimeout(() => ((timedOut = true), resolve()), LOCK_TIMEOUT)),
+					new Promise<typeof timedOut>((resolve) => {
+						timer = setTimeout(() => resolve(timedOut), LOCK_TIMEOUT);
+					}),
 				]);
-				if (timedOut) {
-					logger.warn?.(
-						`dropTable() timed out after ${LOCK_TIMEOUT}ms waiting for ${pending.length} in-flight source-populated cache write(s) on ${tableName}; proceeding with the drop anyway`
+				clearTimeout(timer);
+				if (result === timedOut) {
+					throw new Error(
+						`dropTable() timed out after ${LOCK_TIMEOUT}ms waiting for ${pending.length} in-flight source-populated cache write(s) on ${tableName} to settle; refusing to drop the column families out from under a write that may still be staged. The drop tombstone is durable, so this will be retried on the next load.`
 					);
 				}
 			}
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -371,8 +371,10 @@ export function makeTable(options) {
 	// must not drop this table's column families while one of those writes is still landing —
 	// racing a drop against an in-flight write is what corrupts the column family handle and
 	// produces "Invalid column family specified in write batch" (harper#1381). Track the
-	// in-flight commit promises here so dropTable() can drain them first.
+	// in-flight commit promises here so dropTable() can drain them first, and stop admitting
+	// new ones (droppingTable) once a drop has actually started.
 	const pendingSourceCommits = new Set<Promise<any>>();
+	let droppingTable = false;
 	let createdTimeProperty: Attribute | undefined,
 		updatedTimeProperty: Attribute | undefined,
 		expiresAtProperty: Attribute | undefined;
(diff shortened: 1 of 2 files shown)


resources/Table.ts
```diff
@@ -5585,8 +5639,20 @@ export function makeTable(options) {
...
+			// is safe on its own given the live droppingTable re-checks in this function - but it
+			// would make dropTable()'s correctness depend on every future early-return path in
+			// this function remembering to check droppingTable, rather than on dropTable() simply
+			// waiting for whatever this function is doing. Tracking the whole lifetime is the
+			// belt to that suspenders, at the cost of a bounded wait on a merely slow source
+			// before the drain's fail-closed timeout below.
+			let commitPromise: Promise<any>;
 			when(
-				transaction(sourceContext, async (_txn) => {
+				(commitPromise = transaction(sourceContext, async (_txn) => {
```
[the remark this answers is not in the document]
kriszyp: Addressed in a5d728bb2: `commitPromise` is registered in `pendingSourceCommits` before `when()` attaches its settlement callbacks, so no handler can delete it before registration. — KrAIs (GPT-5 Codex)

resources/Table.ts
```diff
@@ -5836,18 +5905,27 @@ export function makeTable(options) {
...
 				},
 				(error) => {
+					pendingSourceCommits.delete(commitPromise);
 					primaryStore.unlock(id);
 					if (resolved) logger.error?.('Error committing cache update', error);
 					// else the error was already propagated as part of the promise that we returned
 				}
 			);
+			pendingSourceCommits.add(commitPromise);
 		});
```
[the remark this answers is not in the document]
kriszyp: Addressed in a5d728bb2 by the same ordering change: settlement now removes an already-registered promise, without a wrapper promise or duplicate cleanup path. — KrAIs (GPT-5 Codex)

resources/Table.ts
```diff
@@ -1341,12 +1350,54 @@ export function makeTable(options) {
...
+			// committed) and before touching a single column family.
+			//
+			// Bounded, and fails CLOSED: the tracked promise covers the whole source round-trip
+			// plus the local commit (see getFromSource), so a hung/slow source or a slow commit
+			// (e.g. a large blob write) could otherwise wedge this drop forever. Rather than
+			// give up and drop anyway - which would reopen exactly the race this drain exists to
+			// close, just less often - a timeout FAILS the drop. The tombstone written above is
+			// already durable, so completeInterruptedDrop picks the drop back up on the next
+			// load, once the stuck write has had time to finish.
+			if (pendingSourceCommits.size) {
```
DavidCockerill, line 1384: Non-blocking. The drain waits on a strictly larger population than the one that can hurt it, and that costs liveness.

The tracked promise deliberately spans the whole source round-trip. But any promise still inside that round-trip when this snapshot is taken *provably cannot stage a write* — it will hit the live `droppingTable` check at `:5769` or `:5905` and abort. So the only promises that can drive the 10s timeout are harmless ones.

Concretely: `source.get()` is app code with no Harper-side timeout, and `throttledCallToSource` (`:5058`) is a *queue* — a request still waiting for a slot is already registered in `pendingSourceCommits`. On a hot cache table with a slow origin, or one under enough load to back the throttle up, `drop_table` fails for a reason unrelated to storage safety. If a source promise never settles at all, `dropTable()` on that table cannot succeed again until the process restarts.

The property I'd want: the wait covers only writes that have passed the final `droppingTable` check. Not asking you to reverse the belt-and-suspenders choice you explain at `:5642` — that concern is fair — just flagging the bill it comes with. The failure is loud and safe, so non-blocking.

resources/Table.ts
```diff
@@ -1341,12 +1350,54 @@ export function makeTable(options) {
...
+				const timedOut = Symbol('timedOut');
+				const result = await Promise.race([
+					Promise.allSettled(pending),
+					new Promise<typeof timedOut>((resolve) => {
+						timer = setTimeout(() => resolve(timedOut), LOCK_TIMEOUT);
+					}),
+				]);
+				clearTimeout(timer);
+				if (result === timedOut) {
+					throw new Error(
```
DavidCockerill, line 1396: Non-blocking, for the record next to the fail-closed claim: "fails closed" is true for this call and not for the retry it hands off to.

`completeInterruptedDrop` (`databases.ts:1848`) has no drain at all — it `dropSync()`s every column family matching `tableName + '/'`. That is literally the stack frame in #1381.

You flag the post-timeout same-name-recreate variant in the description. The broader shape: the drain window *is* the tombstone-durable / catalog-rows-present state that function exists to act on, and **any** worker reaching `resetDatabases()` in that window hits `databases.ts:548` and drops the CFs undrained — triggered by any schema ITC event, including the `create_attribute` that dynamic-schema writes generate on their own. The window pre-existed (the blob sweep is a full table scan), but this drain extends it by up to 10s and makes its length a function of source latency rather than table size.

Not asking for a fix here — it belongs with the cross-worker follow-up.

resources/Table.ts
```diff
@@ -366,6 +366,15 @@ export function makeTable(options) {
...
 	let primaryKeyAttribute: Attribute | undefined;
 	let lastEvictionCompletion: Promise<void> = Promise.resolve();
+	// getFromSource() intentionally resolves its caller before the resolved record's cache
+	// write has committed (see there) so GET latency doesn't pay for the write. dropTable()
+	// must not drop this table's column families while one of those writes is still landing —
+	// racing a drop against an in-flight write is what corrupts the column family handle and
+	// produces "Invalid column family specified in write batch" (harper#1381). Track the
+	// in-flight commit promises here so dropTable() can drain them first, and stop admitting
+	// new ones (droppingTable) once a drop has actually started.
+	const pendingSourceCommits = new Set<Promise<any>>();
```
DavidCockerill, line 376: Agreeing with the Codex/Grok scope finding you already carried into the description rather than restating it — adding two facts so the follow-up gets sized right rather than filed as an edge case:

1. `dropTable()` runs on exactly **one** worker of N. `DROP_TABLE` in `GLOBAL_SCHEMA_UPDATE_OPERATIONS_ENUM` (`serverUtilities.ts:66`) only triggers `setSchemaDataToGlobal` at `:139`; the ITC fan-out (`signalling.ts:15` → `serverHandlers.js:41`) lands other workers in `resetDatabases()`, never `dropTable()`. Column families are process-wide.
2. There is a second background writer to the same CFs pinned to a **different** worker by design: the eviction scan runs only on the last worker (`:6055`), chains through `lastEvictionCompletion` (declared eight lines above this one), and is armed by this very cache path via `scheduleCleanup()` in the source-write commit.

So on a `source` + `expiration` cache table the untracked cross-thread writer is essentially always present. Agree it's separate work — just arguing it's scheduled work rather than backlog.

DavidCockerill (approved):
Approving. Within the worker that runs `dropTable()`, the drain is a proof rather than a narrowed window, and I verified that rather than taking the description's word for it: two synchronous blocks with no intervening await (`Table.ts:1377-1385` and `:5626-5714`) mean every `getFromSource` call is either in the drain snapshot or reads `droppingTable === true` and aborts before staging — no third case. The only path to storage (`addWrite` at `:5911`) is immediately preceded by a live check with no await between. The tracked promise genuinely covers the native commit including the coordinated-retry ladder, not just the callback.

Guard ordering is right — flag before schema removal, before the drain, before any `dropSync()` — and moving the drain ahead of the blob sweep means drain-caught writes' blobs are visible to it rather than orphaned. Timeout is explicit and fails closed: 10s, throws, no CF touched.

Also worth recording: the `+461/−232` is not a restructure. `git diff -w` reduces `Table.ts` to `+93/−8`, ~55 of which is comment — the functional change is about 30 lines and the rest is one de-indentation. Nothing to weigh there. 42/42 locally.

Three non-blocking notes in threads, all adjacent to caveats you already disclosed.

PR #2002, opened 2026-07-30
test: deflake audit log subscription-event assertion (split from #1999)

kriszyp:
## Summary

Split out of #1999 per review feedback there — this test-only change is unrelated to the TLS fix and was one of the commits conflicting on the `v5.1` patch cherry-pick, so it shouldn't gate a customer patch.

`transactionBroadcast.ts` coalesces `committed` bursts landing in the same turn into a single notify pass, and the `subscribe()` listener delivers only the latest value per id from that pass (by design). Four same-turn writes to two ids can therefore collapse to as few as 2 delivered events no matter how long the test polls afterwards. The fix waits for the notify drain after each write instead of after the whole burst, which makes delivery deterministic and lets the assertion tighten from "at least a couple" (`> 2`) to the true bound (`>= 4`, one live-subscription event per write).

Rebased onto current `main`, which had independently gained a partial deflake (a `waitFor(> 2)` block); the per-write waits supersede it, so that block is replaced by the tight assertion. Main's cleanup-pass improvements in the same test are untouched.

## Testing

`unitTests/resources/auditLog.test.js`: 3/3 consecutive runs green (22 passing each, ~800 ms). A reviewer on #1999 also ran the original version of this change 3× with 51–69 ms completion against the 500 ms-per-write budget.

Refs #1999.

*Generated by Claude (Opus 5) for dispatch task fix-harper-1999.*

### DIFF
--- a/unitTests/resources/auditLog.test.js
+++ b/unitTests/resources/auditLog.test.js
@@ -49,10 +49,13 @@ describe('Audit log', () => {
 		// can therefore collapse to as few as 2 delivered events no matter how long we later
 		// poll for them — waiting for the notify drain to catch up after EACH write (rather than
 		// after the whole burst) is what actually makes delivery deterministic.
+		// >= rather than === in the condition: an overshoot (however it arose) should fall through
+		// to the assertions below with a clear message, not hang the waiter into its timeout.
 		async function waitForEventCount(count) {
-			for (let i = 0; i < 50 && events.length < count; i++) {
-				await delay(10);
-			}
+			await waitFor(() => events.length >= count, {
+				timeout: 5000,
+				message: `expected at least ${count} subscription events`,
+			});
 		}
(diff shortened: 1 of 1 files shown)


cb1kenobi:
Reviewed `55b2670` — no issues found. This PR looks good, nice job!

Note: #1993 (still open) edits the same test block (`unitTests/resources/auditLog.test.js`, the write-burst + subscription-event assertion in `check log after writes and prune`) with a different, incompatible approach — adding a third never-superseded id and content-based event assertions, vs. this PR's per-write drain. Whichever merges second will hit a merge conflict on these lines and should re-verify against the other's reasoning rather than blindly resolving it. No conflict with #1963 (already merged, and its cleanup-pass code is left untouched here).

—
_Generated by Barber AI_

PR #2003, opened 2026-07-30
feat(tls): give MQTT's raw-socket listener its own TLS usage type (split from #1999)

kriszyp:
## Summary

Split out of #1999 per review feedback there, to unblock the `v5.1` patch cherry-pick of the customer-facing readiness fix: this half is a no-op for every stock deployment (Harper only ever writes `uses: []` / `['operations-api']` / `['replication']`) and is what made the cherry-pick conflict, since `v5.1` predates `getEffectiveTlsCiphers`.

A raw-socket TLS listener (`onSocket()`, used by MQTT's `network.securePort`) always resolved its TLS usage type as the generic `'server'`, so a certificate tagged for a specific listener type (`uses: ['mqtt']`) never got its intended priority over a generic node certificate. `onSocket()` now accepts a `usageType` option (falling back to `'server'`) and MQTT passes `'mqtt'` — applied to both certificate quality scoring and effective cipher/`@SECLEVEL` resolution.

**Backward compat:** certs tagged `uses: ['server']` targeted MQTT under the old default, so `'server'` earns the legacy generic-use credit via an explicit allowlist (`LEGACY_SERVER_FALLBACK_TYPES`, currently just `'mqtt'`). An allowlist rather than "everything except operations-api" because every other existing type (`operations-api`, `replication`, …) has always had its own dedicated identity and must not newly start accepting a `['server']`-tagged record's ciphers — this scoping went through three review rounds on #1999 (Codex flagged the compat break, a Harper-domain pass caught the operations-api leak, human review caught the replication leak and requested the allowlist + negative tests).

## Known caveat (carried over from #1999's writeup)

Quality-scoring crossover for `uses: ['server']`: previously a self-signed cert tagged `['server']` for MQTT's hostname beat a CA-signed `['https']` cert for the same hostname (4 vs 3.5, since MQTT's type literally *was* `'server'`). Now the `'server'` tag only earns the legacy-fallback credit (+0.5), so that same cert loses (1.5 vs 3.5). Only matters when multiple certs cover the same hostname with different `uses` tags; flagged for a reviewer with an opinion.

## Where to look

- `server/threads/threadServer.js` — `onSocket()` accepts `usageType`, threads it into `createTLSSelector` and `getEffectiveTlsCiphers`.
- `server/mqtt.ts` — passes `usageType: 'mqtt'`.
- `security/keys.ts` — `LEGACY_SERVER_FALLBACK_TYPES` + its use in `ciphersCandidateRelevant()` and the selector's quality scoring.
- `unitTests/security/keys.test.js` — allowlist test with negative `operations-api`/`replication` cases.

## Testing

`unitTests/security/keys.test.js`: 48/48 passing on this branch (47 on main + the new allowlist test). `npm run build`, `prettier --check`, `oxlint` clean on changed files.

Refs #1999.

*Generated by Claude (Opus 5) for dispatch task fix-harper-1999.*

### DIFF
--- a/security/keys.ts
+++ b/security/keys.ts
@@ -830,17 +830,33 @@ function parseCipherString(ciphers) {
 	};
 }
 
+// Usage types that resolved to the generic 'server' by default before they were given their own
+// explicit usageType — an allowlist, not a denylist, of "type !== 'operations-api'": every OTHER
+// existing usageType ('operations-api', 'replication', ...) has had its own dedicated identity
+// since inception and never fell back to 'server', so a cert tagged uses: ['server'] was never
+// relevant to them and must not newly start winning cipher/quality consideration there. Add a type
+// here only when it is migrating away from onSocket()'s old unconditional 'server' default (as
+// 'mqtt' is doing now) and existing uses: ['server'] certs must keep matching it.
+const LEGACY_SERVER_FALLBACK_TYPES = new Set(['mqtt']);
+
 /**
  * Whether a certificate (record or `tls[]` entry) can affect the given listener, mirroring
  * createTLSSelector's tolerant selection: no `uses` is a generic certificate, `'https'` is the
- * legacy generic use, and an authority matters exactly when the listener verifies client chains.
+ * legacy generic use (applies everywhere, including operations-api and replication — pre-existing
+ * behavior), `'server'` is the legacy generic use only for LEGACY_SERVER_FALLBACK_TYPES, and an
+ * authority matters exactly when the listener verifies client chains.
  */
 function ciphersCandidateRelevant(usesRaw, isAuthority, servesCertificate, type, verifiesClientCerts) {
 	if (isAuthority && verifiesClientCerts) return true;
 	if (!servesCertificate) return false;
 	// normalize: stored as scalar in legacy/manual entries, expected array
 	const uses = Array.isArray(usesRaw) ? usesRaw : usesRaw ? [usesRaw] : [];
-	return uses.length === 0 || uses.includes(type) || uses.includes('https');
+	return (
+		uses.length === 0 ||
+		uses.includes(type) ||
+		uses.includes('https') ||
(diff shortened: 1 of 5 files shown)


cb1kenobi:
Reviewed `7b102591` — no issues found. This PR looks good, nice job!

Traced the full path (`onSocket` → `createTLSSelector`/`getEffectiveTlsCiphers` → `ciphersCandidateRelevant`): the `'mqtt'` usage type is correctly isolated. `getEffectiveTlsCiphers` only special-cases the config layers for `operations-api`, so `'mqtt'` reads the same root `tls` layer `'server'` did, and the `LEGACY_SERVER_FALLBACK_TYPES` allowlist keeps `uses: ['server']` certs relevant — no configured cipher/`@SECLEVEL` that previously applied to the raw-socket MQTT listener is silently dropped. The `operations-api`/`replication` exclusion is confirmed by the negative tests, and each listener's selector is independent so other listeners are unaffected. MQTT-over-WebSocket goes through the http.ts path and is untouched.

On the flagged quality-scoring caveat: preferring a CA-signed `['https']` cert over a self-signed `['server']` one for the same hostname is a reasonable direction — the change only ever lowers a `'server'`-tagged cert's score for MQTT, never raises a weaker cert above a stronger one, so it's not a security regression. Fine to keep as-is.

(Could not execute the unit tests locally — isolated worktree had no `node_modules` and a full install/build was out of scope; verified the added allowlist test exercises the changed relevance path correctly by reading it. Trusting CI + the reported 48/48 for execution.)

—
_Generated by Barber AI_

cb1kenobi:
Reviewed `69959095` (incremental since `7b102591`) — no issues found. The only production change is a defensive optional-chaining guard on a log line in `createTLSSelector`; the rest is regression coverage for the mqtt `usageType` contract (cert exact-match priority, server-tagged fallback, and the `threadServer.onSocket` handoff). TLS isolation, no-fail-open, and cipher/minVersion behavior all unchanged. This PR looks good, nice job!

—
_Generated by Barber AI_

Issue #2001, opened 2026-07-30
Unsettled commit promise wedges all writes on a thread indefinitely — no timeout, no log, no health signal


kriszyp:
## Summary

When a commit promise never settles, `outstandingCommit` (module-level, per worker thread, `resources/DatabaseTransaction.ts`) stays armed forever and **every subsequent write on that thread returns 503** — permanently. There is no timeout that clears or fails the stuck commit, nothing is logged server-side, and the node continues to report healthy. Only a restart recovers.

[#1862](https://github.com/HarperFast/harper/issues/1862) describes one trigger (a stalled uWS client blob upload) and is the right fix for that path. This issue is the **generic resilience gap behind it**: whatever the trigger, an unsettled promise should not be able to disable writes forever, silently.

Observed in production on the `flair.kris-test` cluster on 2026-07-29/30, not in a test storm.

## Mechanism

```js
// resources/DatabaseTransaction.ts
outstandingCommit = commitResolution;
outstandingCommitStart = performance.now();
outstandingCommit.catch(() => {}).finally(() => { outstandingCommit = null; });   // ONLY clear path

checkOverloaded() {
  if (outstandingCommit && performance.now() - outstandingCommitStart > MAX_OUTSTANDING_TXN_DURATION)
    throw new ServerError('Outstanding write transactions have too long of queue, please try again later', 503);
}
```

`MAX_OUTSTANDING_TXN_DURATION` (`STORAGE_MAXTRANSACTIONQUEUETIME`, default 45s) only **rejects new work**. It never times out, aborts, or clears the commit it is measuring — so the 45s check detects the condition and then does nothing about it, forever.

## Evidence it was a wedge, not load

**The failure rate was dead constant for six hours** — 349/hour, exactly our client's heartbeat interval, i.e. *every* write failing. Real write pressure fluctuates:

```
Jul 30 00:00–04:00   349, 349, 349, 349, 349 per hour
```

**Reads were unaffected** (`GET` → 200 throughout), consistent with a write-path flag rather than resource exhaustion. Disk 44% used with 36G free; whole DB 1.9G.

**Only one node.** Probing each node's app route directly (the GTM hides this):

```
node A   503 503 503 503 503 503
node B   200 200 200 200 200 200
```

[...]

PR #2007, opened 2026-07-30
Log once when checkOverloaded() first rejects writes (harper#2001)

kriszyp:
## What

When a RocksDB commit stays outstanding past `storage.maxTransactionQueueTime` (default 45s), `checkOverloaded()` rejects every subsequent write on that worker thread with a 503 — and until now, logged **nothing** server-side. `hdb.log` showed zero occurrences during a live production incident (harper#2001), and `outstandingCommit` carried no database/table identity, which was the single biggest obstacle to root-causing a recurrence.

This adds one `error`-level log, emitted once per stuck commit (not once per rejected request — a wedged thread can reject hundreds of requests/sec), naming:
- the database and table the stuck commit was writing
- the native rocksdb-js transaction id (for correlating with other diagnostics keyed on it)
- the resource/method that started the request, when available
- how long the commit has been outstanding

Also fixes `startedFrom` (the resource/method that started a transaction) never propagating to a chained transaction when a request touches a second table in `txnForContext` (`Table.ts`) — this was silently blanking the equivalent field on the pre-existing long-transaction-abort log too, not just the new one.

**No change to the 503 rejection itself, the arming condition, or any write path.** Diagnostics only.

## Why this shape

Goal 2 of the investigation asked specifically for a **single, low-risk logging addition** — not a fix to whether/how the stuck commit gets cleared (that's a separate, riskier change to the arming mechanism itself; see Risks below and the linked issue for a proposal).

## Test plan

- `npm run build` — clean
- `npm run test:unit:resources` — 1323 passing, 0 failing (twice, to rule out the one observed flake in an unrelated `caching.test.js` test)
- `npx oxlint` / `npx prettier --check` — clean
- Manual code-trace verification of the arming-branch identity capture against `this.writes[0].store` (confirmed via `abort(): void` in the installed `@harperfast/rocksdb-js` 2.5.0, so this branch's `if (commitResolution)` gate is only ever reached via a real `transaction.commit()`, never `abort()`)
- No new automated test for the log-once/reset state machine itself — `outstandingCommit`/`outstandingCommitStart` are module-private and `MAX_OUTSTANDING_TXN_DURATION` is frozen at module-load time from config, so a unit test can neither drive the sentinel nor shrink the 45s threshold after import. Forcing a genuine "native commit that never settles" deterministically is the same open problem this whole investigation is about. See Risks below.

## Risks & open questions

- **Arming gap (pre-existing, not introduced here, documented in code):** a coordinated-retry re-commit or a chained `this.next.commit()` runs synchronously inside the settling commit's own `.then` handler, one microtask before the `.catch().finally()` clears `outstandingCommit` — so a retry round or a second store's commit that itself wedges is never armed, and neither the 503 nor this log will ever see it. Only the first store's first commit attempt is covered. Fixing this touches the arming/clearing mechanism itself and is exactly the kind of riskier change the linked issue asks to keep separate from this logging PR.
- **LMDB engine is entirely unaffected** — `LMDBTransaction.ts` maintains its own separate `outstandingCommit` sentinel, never stamps a start time, and has no `checkOverloaded()`-equivalent timeout/rejection at all. Pre-existing, out of scope here.
- **The log fires once for the whole lifetime of a wedge**, per the task's ask ("log ONCE"). An operator paged well after the log line scrolled past would see nothing further while 503s continue. A periodic re-log (e.g. every 60s while still outstanding) was suggested during review and intentionally not added, to keep this change to exactly what was asked.
- **The two pre-existing long-transaction-monitor logs** (`DatabaseTransaction.ts`, "Transaction was open too long...") still identify the table via `(txn.db as any)?.name` — the same first-claiming-table imprecision this PR's own reasoning found and fixed for the new log (now via `this.writes[0].store`). Left alone as a separate, self-contained cleanup outside this PR's scope.
- `resources/transaction.ts`'s low-level `transaction()` helper never sets `startedFrom` at all (only Resource.ts's own transaction-starting dispatch does) — so the `Table.ts` propagation fix here only helps requests that start through a Resource method, not code that manually orchestrates a transaction via that helper directly.

Refs #2001


cb1kenobi:
Reviewed `300f3105` — no issues found. This PR looks good, nice job!

—
_Generated by Barber AI_

PR #170, opened 2026-02-24
allow role to specify access to specific operations

heskew:
Resolves [CORE-2959](https://harperdb.atlassian.net/browse/CORE-2959) and #172

PR #2008, opened 2026-07-30
fix(tls): keep rebuilding the SNI selector when a pass produces only a default context (empty per-hostname map)

Devin-Holland:
## Summary

Follow-up to #1999 / #2005. On an affected production cluster, port 8883 (MQTT-over-TLS, terminated by the Symphony proxy) still serves the node certificate for custom-domain SNIs on `harper-pro` 5.1.25 (which contains #1999). Verified live, read-only:

- every worker's `sockets/<n>-8883.yaml` is 42 bytes (empty `certificates:`) while the `<n>-9926` mirror is the full ~82 KB;
- on the wire, an MQTT-TLS handshake for a custom-domain SNI on 8883 returns the node cert; the same SNI on 443 returns the correct leaf;
- none of #1999's new breadcrumbs (`resolved zero certificates`, `waiting for system.hdb_certificate`) ever fired (checked `hdb.log` and `docker logs`).

## The gap #1999 left

#1999 fixed the *not-loaded* and *table-swap* cases, but not a pass that ends with a **default context set but an empty per-hostname `secureContexts` map**. The guard `secureContexts.size === 0 && !defaultContextSetThisPass` treats "a default was set" as success and resolves-and-stops, so on the boot pass that produced the empty map, no retry was armed.

## What this changes (`security/keys.ts`, `createTLSSelector`)

Re-check on the debounce whenever the per-hostname map is empty, regardless of whether a default was produced:
- no default → stay pending as before (nothing to serve; resolving would let Bun's path start plaintext);
- default set → resolve `.ready` (serve via the default, no hang) but keep the scheduled rebuild armed so a later pass republishes the per-hostname list.

Warn-latch resets are split so the empty-map-with-default fall-through doesn't re-warn on each re-check.

## ⚠️ Open question — read before relying on this (from @kriszyp's live investigation)

Live CDP inspection of a running node shows the **in-memory 8883 selector maps ARE populated** (all custom domains, write-listener registered), while the on-disk yaml is still the 42-byte boot artifact — and calling `writeUdsMetadata` against the live map produces a correct yaml. So a rebuild *did* run and populate the map after the empty publish, yet the listener fan-out that should have rewritten the yaml didn't take effect. **There may be a second defect between map-population and republish that this PR does not address** — this PR arms and relies on that same fan-out.

Discriminating test before treating this as the fix: on a plain-5.1.25 node, one benign `hdb_certificate` write, then check the `sockets/<n>-8883.yaml` mtime + content:
- heals to the full list → the fan-out works, boot ordering was the whole story, this PR closes the window;
- stays 42 bytes → a second defect between rebuild and republish that this PR does not touch — find it before shipping.

A lead worth checking there: in the raw-socket path (`threads/threadServer.js onSocket`) the write-listener is `push`ed onto `secureContextsListeners` **after** `SNICallback.initialize(...)`, so the first populating pass's fan-out can fire against a listener set that doesn't yet include `writeMetadata`.

## Notes

- Root cause is inferred from live evidence + code; not debugger-traced by me.
- #2003 is orthogonal (shifts cert *quality*, not `secureContexts` membership).
- Interim unblock in flight: host-manager #170 (source the list from a populated sibling socket).
- `unitTests/security/keys.test.js` 55/55 locally; the new test fails without the fix (verified by reverting). Full suites via CI.

*Generated by Claude (Opus 4.8), investigation + fix driven with Devin.*

### DIFF
--- a/security/keys.ts
+++ b/security/keys.ts
@@ -1162,45 +1162,53 @@ export function createTLSSelector(type, mtlsOptions?, liveReload = true): any {
 							logger.error?.('Error applying TLS for', cert.name, error);
 						}
 					}
-					if (liveReload && secureContexts.size === 0 && !defaultContextSetThisPass) {
-						// The not-loaded-yet guard above only covers the table object being absent, not the
-						// table being present but every row failing to apply (e.g. a private key not yet
-						// available on this thread, caught above per-cert). Resolving here would still write
-						// an empty `certificates:` list — the exact customer-visible symptom this PR exists to
-						// fix. Retry on the same debounce instead of publishing that state; a rebuild trigger
-						// (cert-table change or private-key reload, both already wired to scheduleRebuild) is
-						// almost certainly still coming on a normal boot, and this just avoids a window where
-						// we'd otherwise publish empty in the meantime.
-						//
-						// `!defaultContextSetThisPass` because an empty hostname map is not the same as "no TLS
-						// available": a cert whose hostnames resolve to [] (no usable SANs and no CN) builds no
-						// per-hostname entry but still sets a serviceable default context — that listener must
-						// resolve and serve via the default rather than retry forever for hostname entries that
-						// can't exist. It must be the per-pass flag, not the persistent `defaultContext` (see its
-						// declaration note): keying off the closure variable would disarm this retry for every
-						// pass after the first success, reopening the publish-empty window on live rebuilds.
+					if (liveReload && secureContexts.size === 0) {
+						// An empty per-hostname `secureContexts` map is not a valid steady state for a
+						// live-reloading SNI listener, so re-check on the debounce instead of resolving it as
+						// final. Two ways to land here:
+						//   1. Table object present but no row applied yet (e.g. a private key not yet available
+						//      on this thread, caught per-cert above) — the boot-order case this retry was
+						//      originally written for.
+						//   2. Only a default-eligible cert with no per-hostname entry applied — a record whose
+						//      stored `hostnames` is `[]`, or a cert with no SANs and no CN — while the real
+						//      per-domain certs have not landed on this thread yet. Those can arrive via a
+						//      bulk/initial-sync path that never fires this selector's hdb_certificate
+						//      subscription (the same event bypass #1999 documented for table swaps), so nothing
+						//      else re-triggers a rebuild. The raw-socket MQTT `securePort` listener then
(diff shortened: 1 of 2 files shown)


kriszyp:
Root cause found — and it's upstream of this PR's mechanism. Reproduced locally on v5.1.25 (fresh default-config install, deterministic on first boot) and fix-verified both directions.

**The actual bug is in `server/threads/threadServer.js` `onSocket()`, not in `createTLSSelector`.** MQTT registers with *both* `port: 1883` and `securePort: 8883` in one `server.socket()` call. The securePort branch creates the TLS server into the function-scoped `let socketServer`, and the UDS-metadata write closure captures that *binding*:

```js
const writeMetadata = () => httpComponent.writeUdsMetadata(yamlPath, options.securePort, socketServer);
```

Then the plain-TCP branch below **reassigns `socketServer` to the 1883 server** (`socketServer = createSocketServer(...)`), which has no `secureContexts`. By the time any write fires — the `.ready.then()` microtask at boot, or any later rebuild's listener fan-out — the closure reads `secureContexts === undefined` off the wrong server and `writeUdsMetadata` publishes the empty `certificates:` list. Instrumented boot shows exactly this: the 8883 selector's pass completes healthy (`mapSize=3, defaultSet=true`) while the 8883 write fires with `ctxSize=undefined`. Capturing the secure server in its own `const` before the TCP branch → all `<n>-8883.yaml` files fully populated on reboot.

This explains the full evidence set: it's deterministic fleet-wide (any node with both MQTT ports enabled — the default — has always exported empty 8883 metadata, hence the host-manager code comment treating that as known), prod's in-memory maps are populated while the disk yaml stays empty, and none of #1999's breadcrumbs ever fired (the selector was never unhealthy).

**Implication for this PR:** the armed-retry here re-runs the same wrong-object write, so it won't change the published metadata on prod; the regression test passes because it drives `createTLSSelector` with a pseudo-server, bypassing `onSocket` where the bug lives. The empty-map-retry may still be worth keeping as defense-in-depth for a genuine transient-empty pass, but it isn't the fix for #1998's surviving symptom. Fix PR for the closure bug (main + v5.1 backport) coming shortly — happy to combine into this PR instead if you'd rather carry it here.

*— Claude (Fable 5), for Kris — repro artifacts: local v5.1.25 boot with instrumented dist, details in the FDE Slack thread.*

PR #2011, opened 2026-07-30
[v5.1] Fix MQTT secure-port UDS metadata publishing an empty certificate list

kriszyp:
v5.1 backport of #2010 (see it for the full root-cause analysis and evidence). Hand-resolved cherry-pick:

- kept v5.1's `tlsConfig.ciphers` line (no `getEffectiveTlsCiphers` on this branch) and omitted main-only property assignments (`appliedCiphers`, `verifiesClientCerts`, `mtlsRequired`, `dedicatedListener`);
- brought over **only** the new regression test — not main's adjacent noDelay/keepAlive test, which guards a main-only `createSocketServer` arg-order fix that v5.1 doesn't have.

Regression test verified passing on this branch (`unitTests/apiTests/mqtt-test.mjs --grep "secure-port UDS metadata"`); build/prettier/oxlint clean.

This is the branch that actually reaches the affected customer (5.1.x) — proposing for the next 5.1 patch release once #2010 lands on main.

*Generated by Claude (Fable 5).*

### DIFF
--- a/server/threads/threadServer.js
+++ b/server/threads/threadServer.js
@@ -494,7 +494,13 @@ function onSocket(listener, options) {
 		setPortServerMap(options.securePort, { protocol_name: 'TLS', name: getComponentName() });
 		const SNICallback = createTLSSelector('server', options.mtls);
 		const tlsConfig = env.get('tls');
-		socketServer = createSecureSocketServer(
+		// Own const, NOT the shared `socketServer` binding: a caller registering both ports (MQTT's
+		// port + securePort) reaches the plain-TCP branch below, which reassigns `socketServer` to
+		// the 1883 server. The writeMetadata closure below outlives this function, so capturing the
+		// mutable binding made every secure-port metadata write read `secureContexts` off the plain
+		// TCP server (undefined) and publish an empty `certificates:` list — the #1998 bug that let
+		// an SNI-routing proxy (Symphony) fall back to the node certificate on 8883.
+		const secureSocketServer = createSecureSocketServer(
 			{
 				rejectUnauthorized: Boolean(options.mtls?.required),
 				requestCert: Boolean(options.mtls),
@@ -508,14 +514,15 @@ function onSocket(listener, options) {
 			},
 			listener
 		);
-		SNICallback.initialize(socketServer);
+		socketServer = secureSocketServer;
+		SNICallback.initialize(secureSocketServer);
 		// Only opt out of reusePort on macOS, which doesn't reliably support SO_REUSEPORT on all
 		// socket types (ENOTSUP). Everywhere else, sharing the port lets every worker accept
 		// connections for this listener (e.g. MQTT), matching how HTTP servers are bound; without
 		// it only the first worker to bind serves the port and every sibling's listen() fails with
 		// a silently-swallowed EADDRINUSE.
-		if (process.platform === 'darwin') socketServer.noReusePort = true;
-		SERVERS[options.securePort] = socketServer;
+		if (process.platform === 'darwin') secureSocketServer.noReusePort = true;
+		SERVERS[options.securePort] = secureSocketServer;
 
 		// Create a corresponding Unix Domain Socket mirror for the secure socket
 		if (env.get(terms.CONFIG_PARAMS.TLS_UNIXDOMAINSOCKETS)) {
@@ -539,9 +546,9 @@ function onSocket(listener, options) {
 			SERVERS[udsPath] = udsServer;
 			httpComponent.registerUdsCleanupPaths(udsPath, yamlPath);
 
-			const writeMetadata = () => httpComponent.writeUdsMetadata(yamlPath, options.securePort, socketServer);
+			const writeMetadata = () => httpComponent.writeUdsMetadata(yamlPath, options.securePort, secureSocketServer);
 			SNICallback.ready.then(writeMetadata);
-			socketServer.secureContextsListeners.push(writeMetadata);
+			secureSocketServer.secureContextsListeners.push(writeMetadata);
 		}
 	}
 	if (options.port) {
--- a/unitTests/apiTests/mqtt-test.mjs
+++ b/unitTests/apiTests/mqtt-test.mjs
@@ -1109,6 +1109,67 @@ describe('test MQTT connections and commands', function () {
 		}
 	});
 
+	it('secure-port UDS metadata carries the certificates when the same socket() call also registers a plain TCP port', async function () {
+		// Regression for #1998's surviving symptom: MQTT registers `{ port, securePort }` in ONE
+		// server.socket() call. onSocket built the TLS server into the function-scoped `socketServer`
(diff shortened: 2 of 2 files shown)


Devin-Holland (approved):
# Approving — this is the one IBM needs

## Independently verified

I re-derived the diagnosis from the code rather than taking the write-up on trust, and it holds — this is the mechanism, and it explains every observation that didn't add up before.

**The reassignment really does outlive the closure.** On the shipped `v5.1.25` tag, `onSocket`: `let socketServer` (`:492`) → secure branch assigns it (`:497`) → `const writeMetadata = () => writeUdsMetadata(yamlPath, options.securePort, socketServer)` (`:542`) → `secureContextsListeners.push(writeMetadata)` (`:544`) → `if (options.port) { … socketServer = createSocketServer(…) }` (`:547`, `:549`). The closure captures the binding, the binding is rebound to the plain-TCP server, and both write paths (the `.ready.then()` microtask and every later listener fan-out) run after `onSocket` returns. `secureContexts` is `undefined` on that object, `writeUdsMetadata` gates on `contexts?.size > 0`, so it emits a bare `certificates:`.

**The precondition is the shipped default, not a config edge case.** `static/defaultConfig.yaml` ships `mqtt.network.port: 1883` *and* `securePort: 8883`, so every default install passes both, hits the reassignment, and publishes an empty 8883 list on every worker. That's the "deterministically, fleet-wide" claim, confirmed.

**And it explains why the HTTP mirror was always fine** — which is the part I most wanted an independent structural reason for, rather than an absence of evidence. In `http.ts` the same closure shape is safe on both counts: `const server = (httpServers[port] = …)` (`:529`) is a `const`, and the `port` it closes over (`:600`) is a parameter of `getHTTPServer(port, secure, options)`, never rebound. Two files, same pattern, one has a mutable binding and one doesn't, and the symptom tracked exactly that.

**Class sweep — this was the only instance.** Since a finding like this is usually a sample of a class, I scripted a hunt across `threadServer.js` and `http.ts` for the same shape (a function-scope `let`, reassigned, captured by a closure that outlives the function). Four other candidates surfaced and all resolve to false positives with reasons: `threadServer.js` `listen_on` is declared *inside* the loop body (per-iteration) and passed to `.listen()` synchronously; `port` at `:43` is a different scope from the `for (let port in …)` loop variables captured at `:430`/`:451`/`:465` (also per-iteration); `listening` is a local `const` shadowing the module-level `let`; `http.ts`'s `socket`/`response`/`body` are request-scoped with synchronous consumers. `socketServer` was the real one.

**Fix completeness.** `writeMetadata` was the *only* deferred closure in `onSocket`, so converting it plus every synchronous use to `secureSocketServer` is the whole surface. `SERVERS[securePort]` gets the TLS server and `SERVERS[port]` the plain one, unchanged; `socketServer = secureSocketServer` preserves the return value for the securePort-only case, so return semantics are identical either way — and the MQTT caller doesn't consume the return anyway.

**The test is the right test.** Driving `global.server.socket()` with *both* ports and asserting the written yaml carries `BEGIN CERTIFICATE` targets the artifact the proxy actually consumes, which is exactly the seam every prior round missed by driving `createTLSSelector` with a pseudo-server. It runs in CI and passes (`✔ … 103ms`), so it's genuinely gating and not just present.

**CI is fully green here**: 46 pass, 1 skip (`review`), zero failures, and the new regression test passes under both Node versions (`✔ … 103ms`, 927 passing in the unit job). Notably v5.1's unit suite doesn't carry the `auditLog` flake that's reddening #2010 on `main`, so this backport has a clean signal end to end.

I also confirmed the v5.1 variant is equivalent to #2010 modulo the `effectiveCiphers` line v5.1 doesn't have — same two files, same closure fix, same test.

Given IBM is on 5.1.x and v5.1.25 shipped with the symptom still live, this is the one that actually reaches them. Not merging it — that's yours.

## On my own miss

Worth recording, since I reviewed #1999 twice and pushed hard on the root cause. I got as far as proving the yaml *had* been written and therefore `updateTLS()` must have completed with an empty map — that part was right, and it's why the original subscribe-throws diagnosis was wrong. But I then enumerated "exactly two ways to complete a pass with an empty map" and both were selector-side: every cert failing the per-cert build, or the table having no rows. The third way — the map is perfectly healthy and `writeUdsMetadata` is simply handed the wrong object — never occurred to me, even though I had read `writeUdsMetadata` and grep-traced all three of its call sites. I verified *that* `socketServer` was passed and never asked *what it was bound to* at call time. Reading a call site isn't reading the value.

## Still open after this (not blockers)

- The empty-map-with-a-default-context case that #2008 targeted is now unaddressed with it closed. Narrow (needs a cert whose hostnames resolve to `[]` — no usable SANs and no CN) and a different trigger from the customer's, but with this fix the write reaches the right object and still finds an empty map, so it would still publish an empty list. Worth an issue rather than scope here.
- #2004 (swap while `secureContexts` is non-empty leaves a dead subscription with no armed retry) is unrelated to this and still stands.

No code changes requested — the absence of inline threads here is deliberate, not an omission. Really nice piece of debugging; the `mapSize=3, defaultSet=true` / `ctxSize=undefined` instrumentation pair is what turns this from a plausible story into a proven one.

*Reviewed by Claude (Opus 5). Verification commands reproducible against the `v5.1.25` tag and both branches.*

cb1kenobi:
Reviewed `93bb6ce` — no issues found. This PR looks good, nice job!

Verified the fix against the real code path: `writeUdsMetadata` reads `secureServer.secureContexts`, and pre-fix the closure captured the mutable `socketServer` binding that the plain-TCP branch (line 556) reassigns to the port-only server, so every secure-port write published an empty `certificates:` list. Giving the TLS server its own `secureSocketServer` const and capturing that resolves it; the return value is preserved via `socketServer = secureSocketServer`. Confirmed the sibling Bun path captures the stable `config.pseudoServer` object (no equivalent bug). The backport faithfully matches #2010 and correctly omits the main-only property assignments (`appliedCiphers`, `verifiesClientCerts`, `mtlsRequired`, `dedicatedListener`, `getEffectiveTlsCiphers`) that don't exist on v5.1 — minimal and appropriate for a stabilization branch. Regression test drives the real `server.socket()` wiring with both ports and restores global state in `finally`.

Note: did not execute the test suite locally to avoid touching the shared Harper instance; relying on the on-branch pass documented in the PR.

—
_Generated by Barber AI_

cb1kenobi:
Reviewed `abed331a` — no issues found. This PR looks good, nice job!

The only change since `93bb6ce` is a test-only cleanup: the regression test now closes the servers it created (TLS 28887, TCP 21887, UDS mirror) before dropping them from the registry, guarded and restoring `SERVERS`/`portServer` state in `finally`. Production fix is unchanged — `secureSocketServer` is still captured in the `writeMetadata` closure and `secureContextsListeners`, so the empty-`certificates:` bug stays fixed. Still a minimal, targeted v5.1 change with no main-only properties leaked in.

—
_Generated by Barber AI_

Issue #1307, opened 2026-06-15
Rolling-upgrade structure skew makes a local __dbis__ `seq` cursor row undecodable (root cause behind harper-pro#352 second call site)


kriszyp:
## Summary

During an in-place rolling upgrade, a node's **own** local `__dbis__` `seq` cursor row (key `[Symbol.for('seq'), nodeId]`, value `{seqId, nodes}`) can become **undecodable** — `RecordEncoder.decode` logs `Error decoding record: Data read, but end of buffer not reached` and returns `null`. This is the *root cause* behind the harper-pro#352 replication wedge; the harper-pro side ships a resilience guard (HarperFast/harper-pro#394) that stops the crash, but the row should not be undecodable in the first place.

## Evidence (live node, fabric investigation)

Node `8tj-us-west1-a-1` after an in-place upgrade to 5.1.1, leader `pq7`:

- The `system` database `__dbis__` decoder had its msgpackr shared-structure table **fully intact and consistent** — 14 structures, on-disk == in-memory, including **id 11 = `["seqId","nodes"]`** (the correct `seq` shape).
- `randomAccessStructure = false`, `typedStructs = 0` — structon random-access is not involved.
- Yet the single `seq` entry present, `[Symbol(seq), 1]` (the leader's cursor), decoded to `null` with "end of buffer not reached 0".
- Net effect: `sendSubscriptionRequestUpdate` crashed on it, inbound replication wedged (`connected:true / subscriptions:null / lastReceivedVersion:0`), still wedged ~50 min post-boot.

So this is **not** missing structures, **not** structon typedStructs, and **not** v4→v5 migration damage (this node was a fresh v5 provision — no LMDB remnants; #1163 / #362 / harper#1275 / copyDb #1260 are all present in 5.1.0/5.1.1 and working). It is a **structure-shape skew**: the `seq` bytes don't cleanly decode against the present, consistent structure table (the decode reads more/fewer fields than the structure it resolves to).

## Why it doesn't self-heal

`hdb_nodes` rows self-heal via base-copy resync (independent of the cursor). The `seq` cursor only rewrites when replication advances — but the handshake crashes on the undecodable cursor, so replication never advances and the row never rewrites. A self-sustaining wedge (broken on the harper-pro side by harper-pro#394, which lets the node resume and rewrite the row against current structures).

## Open question / where to look

How does a **locally-written** `seq` row end up encoded against a structure shape this node later decodes differently across a rolling upgrade? Candidates worth investigating:
[...]

PR #2014, opened 2026-07-30
Write the version/metadata prefix on LMDB→RocksDB migrated records so they keep versions and record prototypes

### LINKED ISSUE #2012: Records written before LMDB→RocksDB migration decode as plain objects — relationship getters, toJSON, and record methods unreachable
## Summary

On a database migrated LMDB→RocksDB via `storage.migrateOnStart`, records whose stored bytes **predate the migration** decode as plain `Object`s with no record prototype — so relationship accessors, `toJSON`, `getUpdatedTime`, etc. are unreachable on exactly those records. Records written **after** the migration decode as `StoreRecordObject` and behave normally. The failure is completely silent: reads return `undefined`, no errors logged.

This currently affects the **dev central manager** and is a **blocker for migrating the prod CM**, which has an even longer version history (this is being filed ahead of that attempt — prod cannot be broken this way).

## Environment

- Dev CM: `harperfast/harper-pro:5.2.0-beta.3`, RocksDB (migrated from LMDB via `storage.migrateOnStart`)
- History (per Devin): started on an early **4.7 beta**, moved through many 4.7.x releases over months → harper-pro **5.1.x** (still LMDB — no Rocks migration at this step) → **5.2.0-beta.x**, where `storage.migrateOnStart` performed the LMDB→RocksDB migration. The affected Organization record (2026-04-29) was written by an early-4.7-era build.
- CM component: `loadAsInstance = false` resources, graphql schema with `@relationship`s

## Evidence (live inspector evals on the dev CM process)

The **table** is healthy — struct prototype has all relationship getters, resolvers installed:

```json
{"protoNames":["constructor","allowRead","roles","clusters","settings"],
 "attrRels":[{"name":"roles","hasResolve":true},{"name":"clusters","hasResolve":true},{"name":"settings","hasResolve":true}]}
```

But per-record prototype linkage depends on when the record was last written:

| record | last written | decodes as | relationship getters |
|---|---|---|---|
| Cluster created post-migration | 5.2/Rocks | `StoreRecordObject` | work |
| Cluster from 2026-07-09 (TERMINATED, untouched since) | pre-migration | plain `Object` | `'clusters' in rec === false`, reads `undefined` |
| Organization from 2026-04-29 | pre-migration (4.7-era) | plain `Object` | unreachable |
| Host records (re-patched constantly by heartbeats) | post-migration | `StoreRecordObject` | work |

Downstream, this produced: `GET /Organization/:id` silently missing its `clusters`/`settings` relationship properties, and `roleData.toJSON is not a function` in permission checks (old `hdb_role` records lost `toJSON` the same way).

## What does NOT reproduce it (important)

A clean-room reproduction of the same path does **not** show the bug: seed data on `harperdb/harperdb:4.7.33` → boot `harper-pro:5.2.0-beta.3` with `storage.migrateOnStart=true` → records on the migrated RocksDB store decode as `StoreRecordObject` and relationships resolve, including after restarts. So the trigger involves something in the real instance's history. Given the corrected history above, the leading candidate is **records written by early 4.7 betas / older 4.7.x releases carrying older record-encoding metadata** that the post-migration decoder doesn't link to the record prototype — my repro seeded with 4.7.33 (late 4.7), which likely already wrote the newer encoding. Other candidates: the migration ran under an earlier 5.2 beta than the currently-running one, or intermediate encoding states from the long 4.7.x upgrade chain.

## Instance preserved for investigation

We are deliberately **leaving the dev CM in the broken state** (not running the rewrite workaround) so core can inspect affected records in situ — ping @Devin-Holland or devain for access/evals.

## Observed workaround

Rewriting a record heals it (the constantly-patched Host records all decode correctly). A one-time no-op rewrite pass over pre-migration records would fix an affected instance — but the migration itself should either normalize record encoding or the decoder should attach the prototype for legacy-encoded records.

## Asks

1. Identify which per-record encoding state causes prototype-less decode after migration (dev CM is available for live inspection — happy to capture raw entry/encoder metadata for an affected vs healthy record pair on request).
2. Make `migrateOnStart` re-encode (or the decoder tolerate) legacy records so long-history instances migrate safely.
3. A verification query/operation we can run post-migration on prod to prove no records are in the broken state.

— filed by devain (Claude Fable 5) for @Devin-Holland; found while debugging the dev CM 5.2 upgrade (see central-manager#530 for the application-side fallout)

kriszyp:
Fixes #2012.

## What was broken

Every record written by the LMDB→RocksDB migration (`storage.migrateOnStart`) since #1307 (2026-06-15, i.e. **all 5.2 alphas/betas**) was stored **without the `[8-byte version][flags word]` metadata prefix**, and the record's version was silently dropped. `copyDb` grafts `RecordEncoder`'s encode hook onto the migration target's plain msgpackr encoder, and the hook's `if (!this.useVersions)` opt-out (added for `__dbis__` in #1307) reads `useVersions` off that foreign encoder — `undefined` — so every migrated record took the "non-versioned store" plain-encode path.

Downstream, prefix-less records decode without the metadata wrapper, so `PrimaryRocksDatabase.getEntry` skipped the `structPrototype` repair: **point reads returned prototype-less plain `Object`s** (relationship getters, `toJSON`, `getUpdatedTime` unreachable — the dev CM "invisible clusters" bug), while `getRange` repaired the prototype unconditionally, so **scans looked healthy**. That asymmetry is why the clean-room repro and the 4.x upgrade integration test (both verifying via `search_by_conditions`) never caught it. Full root-cause narrative: https://github.com/HarperFast/harper/issues/2012#issuecomment-5136837457

## Changes

**`resources/RecordEncoder.ts`** — the encode hook's non-versioned opt-out now requires an explicit `useVersions === false` (matching the constructor's documented "default to true when unspecified" intent) so a foreign encoder the hook is grafted onto can't silently strip prefixes. `__dbis__` still opts out explicitly (`databases.ts` sets `useVersions = false`).

**`resources/PrimaryRocksDatabase.ts`** —
- `#processEntry` gates on `raw[METADATA] !== undefined` instead of truthiness: a `[timestamp][no flags word]` record decodes with `metadataFlags === 0`, and the falsy gate handed the decode wrapper itself back as the record value. Same presence-check fix in `getRange`.
- Metadata-less values now get the same `structPrototype` repair as `getRange` and the old LMDB wrapper (which always repaired — the asymmetry was new to `PrimaryRocksDatabase`). This **heals prototypes on already-migrated instances** (e.g. the dev CM) without a rewrite pass; versions remain unrecoverable without one.

**`bin/copyDb.ts`** —
- Version-less source records are encoded plain (`clearNextEncoding`) rather than with a flags-word-only prefix, which the RocksDB decode heuristic would misparse (8 bytes consumed as a timestamp). Defensive: real sources always carry timestamp versions.
- Post-copy invariant tripwire: after each primary dbi is copied, the first versioned record is read back raw (awaiting its own put promise) and its 8-byte big-endian float64 header is compared to the **source version**; on mismatch the migration **fails loudly** (keeping the LMDB source and `migrateOnStart` flag intact). (Full-header comparison rather than a first-byte check — a prefix-less classic record can also lead with `0x42` as structure ref id 2; flagged by the Codex review leg.)
- **Full verification sweep** (issue ask #3, completed per heskew's review): after the tripwire, every primary dbi gets a raw first-byte scan — unprefixed records must exactly equal the counted version-less source records, else the migration fails closed. And `verifyMigratedDatabase(databasePath)` is exported for **already-migrated** instances (prod pre-flight / dev CM detection): per-table `{records, unversioned}` report; read-only but takes the RocksDB lock, so run in-process or offline.
- **Restart-safe staged migration** (per kriszyp's review): `migrateOnStart` now migrates into `<db>.migrating` — excluded from database discovery — and atomically renames into place only after the copy fully verifies, replacing any stale artifact at the real path. A failure or SIGKILL part-way can no longer leave a partial RocksDB at `<db>/` that races the intact `<db>.mdb` at next-boot discovery (the "whichever binds last wins → partial copy promoted, LMDB abandoned" hazard, which also applied to the pre-existing blob-failure throw). `copyDbToRocks` now closes **every** target handle before returning — rocksdb-js registers descriptors by path string, so a handle leaked on the staging path would hold the DB LOCK against the runtime's open of the renamed directory. Residual (pre-existing, unchanged) window: a crash between the rename and the LMDB→backup move leaves a *complete, verified* copy plus the LMDB; next boot either re-migrates from the LMDB (correct) or promotes the complete copy (also correct).

## Tests

- `unitTests/bin/migrationMetadataPrefix.test.js` (new, real `copyDbToRocks` run under LMDB engine): migrated bytes start with the version prefix; point reads through `PrimaryRocksDatabase` return prototype-linked records with the **source** version.
- `unitTests/resources/primaryRocksMetadataRepair.test.js` (new): flags-word-absent records don't leak the decode wrapper; metadata-less (broken-migration-layout) records get the prototype repair on point reads.
- `unitTests/bin/migrationCanonicalStructures.test.js` updated to read the migrated CF through `PrimaryRocksDatabase` + `RecordEncoder` (the runtime path) — its plain-msgpackr `cf.get` was only decodable against the broken prefix-less layout.
- `unitTests/bin/migrationStagingRecovery.test.js` (new): a failed migration (injected error) removes its staging dir and leaves nothing at the target path with the LMDB source still serving; the both-artifacts-present restart state (stale staging dir + partial RocksDB at the real path) recovers via re-migration and atomic promotion — the post-rename open also guards the close-all-handles contract (a leaked staging handle would hold the RocksDB LOCK); `verifyMigratedDatabase` reports clean; and discovery ignores `*.migrating` dirs carrying Rocks marker files. Writing the failure-path test surfaced that `copyDbToRocks` acquired its read transaction outside its try block, leaking every staging handle on an early throw — fixed.
- All three LMDB-gated migration test files now gate with `this.skip()` in `before()` (visible as pending under the rocks engine) instead of a bare describe-body `return` that registers nothing (heskew's review). Note these DO run in CI: `test:unit:all` → `test:unit:lmdb` → `test:unit:bin` under `HARPER_STORAGE_ENGINE=lmdb` — but only when the earlier `&&`-chained legs pass; a fail-late `test:unit:all` is a worthwhile follow-up.
- Byte-level before/after validated with standalone repros against a real RocksDB store (broken layout reproduced, fix verified, healing of genuinely broken bytes verified).

Suites: `test:unit:resources` (rocks + lmdb), `unitTests/bin` (lmdb) green locally.

## Cross-model review

Thorough mode (storage domain): Gemini (`agy`, diff-only perf/mechanics lens) + Codex (repo-access correctness lens) + Harper-domain adjudication. **No blockers.** One significant finding (Codex): the tripwire's original single-byte `0x42` check was ambiguous with classic structure ref id 2 — addressed in the final artifact by comparing the full decoded header to the source version. Gemini's three raw flags were adjudicated as diff-only false positives (closure-scoped `written`, existing `getRange` repair parity, `#src` subpath imports in CJS tests). Final artifact re-checked in a quick pass after the tripwire change.

## Notes for reviewers

- Design assessment: no new API surface; changes tighten existing invariants on the core encode/decode path. The riskiest line is the `useVersions === false` semantics change — audited all constructions: `RecordEncoder` always sets a boolean in its constructor, `databases.ts:82` sets explicit `false` for `__dbis__`, and `copyDb`'s graft is the only site assigning the hook to a foreign encoder.
- Remediation guidance for already-migrated instances is in the issue (read-side repair here restores prototypes; a no-op rewrite pass is still needed to restore versions).

### DIFF
--- a/bin/copyDb.ts
+++ b/bin/copyDb.ts
@@ -421,7 +421,9 @@ export async function migrateOnStart() {
 
 /**
  * Count records in a raw (encoding: false) store whose value lacks the 8-byte version prefix
- * (0x42 = first byte of a ms-epoch float64). Symbol keys are internal and skipped.
+ * (0x42 = first byte of a ms-epoch float64). Symbol keys are internal and skipped. First-byte
+ * classification is a heuristic: a prefix-less classic record can also lead with 0x42 (structure
+ * ref id 2) and would be undercounted — fine for the whole-table detection this report serves.
  */
 function countRecords(rawDbi): { records: number; unversioned: number } {
 	let records = 0;
@@ -633,7 +635,7 @@ export async function copyDbToRocks(sourceRootStore, sourceDatabase: string, tar
 			console.log('migrating', key, 'from', sourceDatabase, 'to RocksDB');
 			const copied = await copyDbiToRocks(sourceDbi, targetDbi, isPrimary, transaction, observerEncoder);
 			if (copied?.firstVersioned !== undefined) {
-				const { firstVersioned, versionlessCount } = copied;
+				const { firstVersioned, versionlessKeys } = copied;
 				// Invariant tripwire (#2012): a versioned record must round-trip with its 8-byte
 				// version prefix. The full big-endian float64 is compared to the source version —
 				// a first-byte check alone is ambiguous, since a prefix-less classic record can
@@ -651,18 +653,26 @@ export async function copyDbToRocks(sourceRootStore, sourceDatabase: string, tar
 							`records would lose versions and prototypes`
 					);
 				}
-				// Full verification sweep (harper#2012 ask #3): every migrated record must carry the
-				// version prefix except the (counted) version-less source records that were
-				// deliberately encoded plain. A raw first-byte scan needs no decode, so it is a
-				// cheap sequential pass over the freshly written CF.
+				// Full verification sweep (harper#2012 ask #3): every migrated record that had a source
+				// version must carry the prefix. Version-less source records are exempted by KEY, not
+				// by byte inspection — a plain classic body can also lead with 0x42 (structure ref
+				// id 2), so byte sniffing could misclassify them and throw spuriously. For versioned
+				// records the first-byte check suffices: a systemic prefix regression trips on the
+				// very first record (and the exact-header tripwire above already validated one).
 				const rawDbi = openRocksDb(targetPath, { name: key, encoding: false });
 				targetHandles.push(rawDbi);
-				const { unversioned } = countRecords(rawDbi);
-				if (unversioned !== versionlessCount) {
+				let unprefixed = 0;
+				let scanned = 0;
+				for (const { key: recordKey, value } of rawDbi.getRange({})) {
+					if (typeof recordKey === 'symbol') continue;
+					if (versionlessKeys.has(typeof recordKey === 'object' ? JSON.stringify(recordKey) : recordKey)) continue;
+					scanned++;
+					if (!value || value.length < 8 || value[0] !== 0x42) unprefixed++;
+				}
+				if (unprefixed > 0) {
 					throw new Error(
-						`Migration of ${sourceDatabase}/${key} left ${unversioned} record(s) without the version/metadata ` +
-							`prefix (expected ${versionlessCount} version-less source records) — records would lose ` +
-							`versions and prototypes`
+						`Migration of ${sourceDatabase}/${key} left ${unprefixed} of ${scanned} versioned record(s) without ` +
+							`the version/metadata prefix — records would lose versions and prototypes`
 					);
 				}
 			}
@@ -745,7 +755,9 @@ export async function copyDbToRocks(sourceRootStore, sourceDatabase: string, tar
 		let recordsCopied = 0;
 		let skippedRecord = 0;
 		let firstVersioned;
-		let versionlessCount = 0;
+		// keys (normalized for compound keys) of source records with no version, deliberately
+		// encoded plain — the verification sweep exempts them by key, never by byte sniffing
+		const versionlessKeys = new Set();
 		const MAX_RETRIES = 1000;
 		let retries = MAX_RETRIES;
 		let start = null;
@@ -789,7 +801,7 @@ export async function copyDbToRocks(sourceRootStore, sourceDatabase: string, tar
 								// Cleared even though the plain-encode hook would not consume stale globals:
 								// they may have leaked from a prior record whose encode was skipped or threw.
 								clearNextEncoding();
-								versionlessCount++;
+								versionlessKeys.add(typeof key === 'object' ? JSON.stringify(key) : key);
 							}
 							written = encodeBlobsWithFilePath(
 								() => targetDbi.put(key, value, version),
@@ -855,7 +867,7 @@ export async function copyDbToRocks(sourceRootStore, sourceDatabase: string, tar
 					}
 				}
 				console.log('finish migrating, copied', recordsCopied, 'entries, skipped', skippedRecord, 'delete records');
-				return { firstVersioned, versionlessCount };
+				return { firstVersioned, versionlessKeys };
 			} catch (err) {
 				console.error(
 					`Error iterating dbi for ${sourceDatabase} near key ${JSON.stringify(start)}, retrying (${retries} retries left):`,
--- a/bin/copyDb.ts
+++ b/bin/copyDb.ts
@@ -20,7 +20,13 @@ import {
 	encodeBlobsWithFilePath,
 	endPendingMigrationBlobSaves,
 } from '../resources/blob.ts';
-import { RecordEncoder, setNextEncoding, lastMetadata, METADATA } from '../resources/RecordEncoder.ts';
+import {
+	RecordEncoder,
+	setNextEncoding,
+	clearNextEncoding,
+	lastMetadata,
+	METADATA,
+} from '../resources/RecordEncoder.ts';
 
 export async function compactOnStart() {
 	hdbLogger.notify('Running compact on start');
@@ -527,7 +533,29 @@ export async function copyDbToRocks(sourceRootStore, sourceDatabase: string, tar
 			copyStructures(sourceDbi, key);
 
 			console.log('migrating', key, 'from', sourceDatabase, 'to RocksDB');
-			await copyDbiToRocks(sourceDbi, targetDbi, isPrimary, transaction, observerEncoder);
+			const firstVersioned = await copyDbiToRocks(sourceDbi, targetDbi, isPrimary, transaction, observerEncoder);
+			if (firstVersioned !== undefined) {
+				// Invariant tripwire (#2012): a versioned record must round-trip with its 8-byte
+				// version prefix. The full big-endian float64 is compared to the source version —
+				// a first-byte check alone is ambiguous, since a prefix-less classic record can
+				// also lead with 0x42 (structure ref id 2). Failing here keeps the migration
+				// incomplete (LMDB source intact, migrateOnStart flag retained) instead of
+				// shipping a database whose records silently lost versions and record prototypes.
+				await firstVersioned.written;
+				await written;
+				const roundTrip = targetDbi.getBinarySync(firstVersioned.key);
+				const roundTripVersion =
+					roundTrip && roundTrip.length >= 8
+						? new DataView(roundTrip.buffer, roundTrip.byteOffset, roundTrip.byteLength).getFloat64(0)
+						: undefined;
+				if (roundTripVersion !== firstVersioned.version) {
+					throw new Error(
+						`Migration of ${sourceDatabase} wrote record ${JSON.stringify(firstVersioned.key)} without its ` +
+							`version/metadata prefix (expected version ${firstVersioned.version}, read back ${roundTripVersion}) — ` +
+							`records would lose versions and prototypes`
+					);
+				}
+			}
 
 			// Persist the canonical v5 classic structures the observer built, so every v5 runtime worker
 			// adopts one agreed dictionary on startup instead of minting its own from an empty durable and
@@ -598,6 +626,7 @@ export async function copyDbToRocks(sourceRootStore, sourceDatabase: string, tar
 	async function copyDbiToRocks(sourceDbi, targetDbi, isPrimary, transaction, observerEncoder?) {
 		let recordsCopied = 0;
 		let skippedRecord = 0;
+		let firstVersioned;
 		const MAX_RETRIES = 1000;
 		let retries = MAX_RETRIES;
 		let start = null;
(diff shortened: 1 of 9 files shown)


heskew:
Verified the fix independently against `main`. The guard inversion is right, and the blast radius is narrow: `existingEncoder.encode = tempEncoder.encode` in `copyDb.ts` is the only graft site in the tree, so nothing else starts emitting prefixes. Comparing the full float64 rather than the first byte is also the right call, given `0x42` is reachable as classic structure ref id 2.

Two gaps that would ship as-is.

## 1. The write-side regression guard never executes in CI

`unitTests/bin/migrationMetadataPrefix.test.js` opens with:

```js
if ((process.env.HARPER_STORAGE_ENGINE || envGet(CONFIG_PARAMS.STORAGE_ENGINE)) !== 'lmdb') return;
```

No workflow under `.github/` sets `HARPER_STORAGE_ENGINE` — the only occurrences are prose inside the AI-review prompt files. Confirmed against this PR's own run: the Node 22 unit log contains `PrimaryRocksDatabase metadata repair (#2012)` and **zero** occurrences of the prefix test or its `PrefixGuard` table.

Because the gate is a bare `return` in the describe body, it registers no tests at all — no failure, no `pending` marker, no trace in the output. (The read-side test uses `this.skip()` in `before`, so it at least reports as pending.)

This is pre-existing rather than introduced here: the `#1453` guard edited in this same PR carries the identical gate (`migrationCanonicalStructures.test.js:24`), so the whole class of LMDB→RocksDB migration unit tests is currently inert.

Worth resolving in this PR specifically because it is the same defect class as #2012 itself — verification that reads like coverage but doesn't run on the path CI exercises. #2012 survived from #1307 (2026-06-15) because `4.x-upgrade.test.ts` verified exclusively via `search_by_conditions`; the new guard doesn't run at all. The runtime tripwire still protects real migrations, so this is guard hygiene rather than a problem with the fix.

## 2. Ask #4 (post-migration verification) is partially delivered

The plan in the issue was a per-table scan for `entry.version == null`. What's implemented is a write-time, first-versioned-record-per-dbi round-trip check. That is sufficient to catch a systemic failure like this one, but it isn't a detection query — so there's still nothing to run against an **already**-migrated instance to prove it is clean, which was the original pre-flight ask for prod.

## CI

Both failures look unrelated to this change. Node 22 is the known `auditLog.test.js:58` flake (1335 passing / 1 failing), which #2002 targets. I could not extract a test failure from Windows Integration 6/6 — `--log-failed` returns only post-job cleanup steps — so that one may be infrastructure rather than a real failure.

---

🤖 Posted by Claude (Opus 5, 1M context) on Nathan's behalf

kriszyp:
Thanks for the independent verification, Nathan — the graft-site blast-radius check and the `0x42`/structure-ref-2 confirmation are exactly the right things to have re-derived, and both gaps were worth acting on. Addressed in 496a75682:

**On gap 1 — one correction, then the fixes.** The guard *does* execute in CI: `unit-test.yml` runs `test:unit:all`, whose last leg is `test:unit:lmdb` → `HARPER_STORAGE_ENGINE=lmdb npm run test:unit:bin`. The reason your Node 22 log had zero `PrefixGuard` hits is the `&&` chain: that job died at the known `auditLog.test.js` flake in an earlier leg, so the lmdb leg never started. A later run of this PR shows `✔ migrated record bytes start with the 8-byte version prefix` in the v22 log itself. That said, your underlying point is right twice over:

- A bare `return` in the describe body registers nothing and is indistinguishable from coverage — all three LMDB-gated migration test files now gate with `this.skip()` in `before()`, so under the rocks engine they report as **7 pending** instead of vanishing.
- The `&&` chain means any earlier-leg failure silently drops the entire lmdb leg — worth a separate look at making `test:unit:all` fail-late (out of scope here).

**On gap 2 — fully delivered now.** Two pieces:

- The migration itself now does a **full raw first-byte sweep of every primary dbi** after copy: the count of unprefixed records must exactly equal the (counted) version-less source records, else the migration fails closed with the LMDB source intact. Not just the first-record round-trip anymore.
- `verifyMigratedDatabase(databasePath)` is exported from `bin/copyDb.ts` for **already-migrated** instances: per-table `{records, unversioned}` report, read-only (takes the RocksDB lock, so run in-process via inspector on a live instance, or offline). That's the prod pre-flight / dev CM detection query from the issue.

Bonus from writing the failure-path test: `copyDbToRocks` acquired its read transaction *outside* its try block, so an early throw leaked every staging handle — now acquired inside with a guarded `transaction?.done()`.

On the CI failures: agreed on both — Node 22 is the #2002 auditLog flake, and Windows Integration 6/6 produced no extractable test failure (rerun queued).

— Claude Fable 5, for @kriszyp

cb1kenobi:
Reviewed `848b3850` — no issues found. This PR looks good, nice job!

Traced the data-integrity-critical paths: the `useVersions === false` opt-out correctly restores the metadata prefix on migrated records (foreign-graft encoder audited as the only `undefined` site — `__dbis__` still opts out explicitly), the read-back tripwire compares the full 8-byte BE version and fails the migration loudly on mismatch (LMDB source + flag retained), the staged `.migrating` dir + atomic rename keeps an interrupted/re-run migration crash-safe, and the `METADATA !== undefined` presence checks fix the flags-word-absent decode-wrapper leak. Idempotency (already-RocksDB skip), empty-db, and version-less-record paths all check out.

—
_Generated by Barber AI_

bin/copyDb.ts
```diff
@@ -412,11 +419,102 @@ export async function migrateOnStart() {
...
+ * Count records in a raw (encoding: false) store whose value lacks the 8-byte version prefix
+ * (0x42 = first byte of a ms-epoch float64). Symbol keys are internal and skipped.
+ */
+function countRecords(rawDbi): { records: number; unversioned: number } {
+	let records = 0;
+	let unversioned = 0;
+	for (const { key, value } of rawDbi.getRange({})) {
+		if (typeof key === 'symbol') continue;
+		records++;
+		if (!value || value.length < 8 || value[0] !== 0x42) unversioned++;
```
cb1kenobi, line 434: **Nit: the full-verification sweep uses the first-byte `0x42` heuristic the tripwire deliberately rejects**

`countRecords` classifies a record as unversioned by `value[0] !== 0x42`, and the sweep then compares that count against `versionlessCount` (records routed through the plain-encode path). For a version-less source record whose plain msgpack encoding happens to lead with `0x42` (structure ref id 2 — the exact ambiguity the tripwire avoids a few lines down by comparing the full big-endian float64 header to the source version), the sweep will NOT count it as unversioned, so `unversioned < versionlessCount` and the migration throws a spurious invariant failure.

Impact is limited: it is fail-closed (LMDB source preserved, flag retained) and only bites when `versionlessCount > 0`, which real sources don't produce — so non-blocking. But it reintroduces the precise first-byte ambiguity the tripwire fix removed. Consider decoding the header (as the tripwire does) or tracking prefix presence directly instead of sniffing the first byte, so the sweep and the tripwire agree on what "versioned" means.

—
_Generated by Barber AI_

kriszyp:
v5.1 backport is up: [#2020 — Write the version/metadata prefix on LMDB→RocksDB migrated records (v5.1 backport of #2014)](https://github.com/HarperFast/harper/pull/2020) (draft while CI runs; manual variant per the [scope-expansion notes](https://github.com/HarperFast/harper/issues/2012#issuecomment-5138392320) — no PrimaryRocksDatabase on that branch). — Claude Fable 5

PR #2020, opened 2026-07-31
Write the version/metadata prefix on LMDB→RocksDB migrated records (v5.1 backport of #2014)

kriszyp:
Manual v5.1 backport of #2014 (fixes the v5.1 exposure of #2012). The v5.1 line is affected from **v5.1.2** — #1307 shipped in it the same day it merged — so every `storage.migrateOnStart` run on v5.1.2–v5.1.23 wrote records without the `[8-byte version][flags word]` prefix, silently dropping record versions. Empirically confirmed by @heskew on 5.1.23 and by ancestry check ([details](https://github.com/HarperFast/harper/issues/2012#issuecomment-5138392320)).

**Why v5.1 instances look healthy:** this branch has no `PrimaryRocksDatabase`; the `handleLocalTimeForGets` wrapper repairs record prototypes unconditionally on point reads, so affected instances read correctly and only lose versions (cache admission, `ifVersion`/CAS, replication version comparisons). The invisible-records symptom fires when they upgrade to 5.2 — so shipping this in 5.1 stops new damage, and `verifyMigratedDatabase` identifies instances needing the rewrite pass before any 5.2 upgrade.

## Variant differences from #2014 (main)

- **Dropped:** the `resources/PrimaryRocksDatabase.ts` read-side changes and `unitTests/resources/primaryRocksMetadataRepair.test.js` — the file doesn't exist on v5.1 and the wrapper already repairs unconditionally.
- **Adapted:** the migration tests verify via `RecordEncoder.decode` metadata wrappers (`{ value, version }`) instead of `PrimaryRocksDatabase` point reads.
- **Ported unchanged:** `copyDb.ts` wholesale (byte-identical base pre-fix): explicit `useVersions === false` opt-out in the RecordEncoder encode hook, staged migration (`<db>.migrating` + atomic rename + close-all-handles + Windows rename retry), exact-header tripwire, key-exempted full verification sweep, `verifyMigratedDatabase` export, discovery exclusion of `*.migrating` dirs, `MIGRATING_DIR_SUFFIX` in hdbTerms.

## Tests

`unitTests/bin/migration*.test.js`: 7 passing under `HARPER_STORAGE_ENGINE=lmdb`, 7 pending (visible skip) under rocks; full lmdb bin suite 69 passing locally. Note: local run used a newer rocksdb-js than the branch lockfile (shared node_modules) — CI arbitrates against the real pin.

All review findings from #2014 (Codex tripwire ambiguity, kriszyp staging restart-safety, heskew verification gaps, cb1kenobi sweep byte-sniffing) are incorporated since this ports the final state.

### DIFF
--- a/bin/copyDb.ts
+++ b/bin/copyDb.ts
@@ -458,9 +458,14 @@ export function verifyMigratedDatabase(databasePath: string): Record<string, { r
 		handles.push(dbisDb);
 		for (const { key, value: attribute } of dbisDb.getRange({})) {
 			if (typeof key === 'symbol' || !attribute?.isPrimaryKey) continue;
+			// per-table handles close per-iteration so a many-table sweep does not hold every CF
+			// handle open at once; only the two pre-loop opens need the leak-safety array
 			const rawDbi = RocksDatabase.open(databasePath, { name: key, encoding: false });
-			handles.push(rawDbi);
-			report[key] = countRecords(rawDbi);
+			try {
+				report[key] = countRecords(rawDbi);
+			} finally {
+				rawDbi.close();
+			}
 		}
 	} finally {
 		for (const handle of handles.reverse()) {
--- a/bin/copyDb.ts
+++ b/bin/copyDb.ts
@@ -3,11 +3,12 @@ import { open, asBinary } from 'lmdb';
 import { join } from 'path';
 import { move, remove } from 'fs-extra';
 import { existsSync, mkdirSync } from 'node:fs';
+import { rename } from 'node:fs/promises';
 import { get } from '../utility/environment/environmentManager.ts';
 import OpenEnvironmentObject from '../utility/lmdb/OpenEnvironmentObject.ts';
 import { OpenDBIObject } from '../utility/lmdb/OpenDBIObject.ts';
 import { INTERNAL_DBIS_NAME, AUDIT_STORE_NAME } from '../utility/lmdb/terms.ts';
-import { CONFIG_PARAMS, DATABASES_DIR_NAME } from '../utility/hdbTerms.ts';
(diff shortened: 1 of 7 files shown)


heskew (approved):
Approving. Re-checked at `a8e59e2d` (unchanged since my earlier pass).

Faithful fail-closed backport of #2014 — the tripwire and the full verification sweep both fail the migration rather than shipping a database whose records silently lost their version prefix, which is the right direction for this class of bug. I diffed it against the original and found no drift beyond the expected branch differences.

I also had an outside model (Codex) raise three blocker-level concerns on an earlier pass; I traced each against the code and **rebutted all three**, including the claim that the exact-header tripwire silently passes when both versions are `undefined` — that state isn't reachable the way it was described. Noting it so the same points don't get re-litigated if another reviewer runs into them.

Confirmed the mirrored `rawDbi` fix from #2022 is present here too, at [`bin/copyDb.ts:466-470`](https://github.com/HarperFast/harper/blob/a8e59e2d/bin/copyDb.ts#L466-L470) — per-iteration open/use/close.

**Non-blocking observation** for whenever this area is touched again: the verification sweep at [`bin/copyDb.ts:679-680`](https://github.com/HarperFast/harper/blob/a8e59e2d/bin/copyDb.ts#L679-L680) still has the shape the bots flagged at 463 — it opens a per-table `rawDbi` and pushes it onto `targetHandles` for the duration of the loop instead of closing it after the scan, even though the handle is only used within that iteration. This is **not** a leak (everything is closed at [762](https://github.com/HarperFast/harper/blob/a8e59e2d/bin/copyDb.ts#L762), CFs before root), so it doesn't block — but on a database with many tables it holds proportionally more RocksDB CF handles open at once, which is exactly the concern that motivated the change at 466.

<sub>sent with Claude Opus 5</sub>

PR #2023, opened 2026-07-31
test(uws): fix macOS EPIPE flake in the UDS 413 test

kriszyp:
## What / why

`uWS UDS adapter (createUwsServer) rejects a body over maxBodyBytes with 413` flaked on macOS: the 413 assertion passed, but an uncaught `EPIPE` landed afterward and mocha attributed it to the test, failing the run (green on Linux CI).

Root cause: the server correctly responds 413 and closes early without draining the rest of an over-limit body. The client's outstanding write can still be in flight when the socket closes, and Node's http client stops forwarding socket errors to the request in the window right after a keep-alive response completes (`responseKeepAlive` in Node's own `_http_client.js`, which carries an upstream `// TODO(ronag): ... the socket has no 'error' handler` comment on exactly this gap) — so the late write failure becomes an uncaught exception with nothing listening.

Confirmed this reproduces on Linux too, not just inferred as macOS-only: against the real `createUwsServer`, the pre-fix helper rejects with a raw `EPIPE` on roughly half of 40 trials using the test's existing 8KB-body/1KB-cap fixture (no artificial size inflation needed). With the fix, 40/40 resolve cleanly to 413.

Fix (in the shared `udsRequest` test helper, so it protects every test in the file, not just the 413 case):
- `agent: false` per request — the primary fix, avoids the pooled-connection gap entirely.
- A socket-level `error` listener as a second layer, tolerating only `EPIPE`/`ECONNRESET`; a `settledThisTick` flag distinguishes Node's synchronous same-event echo (already handled via `req`'s own forwarding) from a genuinely late arrival worth recording. Anything else post-settlement is collected and asserted empty in `after()`, rather than silently absorbed.
- Reject on `close` if the promise never settled, so a truncated response fails with a diagnosable error instead of hanging to Mocha's timeout.
- A new test for a connection failure before any response.

No changes to `server/serverHelpers/uwsServer.ts` — the server's behavior (respond early, don't drain an over-limit body) is correct per HTTP and is preserved untouched.

## Verification

- `npx mocha unitTests/server/serverHelpers/uwsServer.test.js`: 21/21 passing, stable across repeated runs, no `MaxListenersExceededWarning`.
- Broader `unitTests/server/**/*.test.js`: 679 passing.
- Confirmed the 413 assertion is still genuine: temporarily disabled the server's over-limit check and the test correctly failed (`200 !== 413`), then reverted.
- Confirmed the fix's necessity and sufficiency directly against the real server (not just the synthetic race description): pre-fix ~50% raw-EPIPE rejection rate at the test's exact fixture size; post-fix 40/40 clean.

## Review

Nine rounds of independent pre-push review (Codex/Gemini/Grok + a Harper-domain pass) drove the final design — several earlier iterations each reintroduced a variant of the exact duplicate-error/uncaught-exception bug this change removes (a competing listener installed too early causing double delivery, then a module-global listener-ownership slot that orphaned a socket when the file's second UDS connection was opened). Final round: verdict LGTM (Codex), no findings from Grok; Gemini's remaining concern (pre-response EPIPE could still defeat the fix) was tested directly against the real server and did not reproduce in 60 combined trials.

### DIFF
--- a/unitTests/server/serverHelpers/uwsServer.test.js
+++ b/unitTests/server/serverHelpers/uwsServer.test.js
@@ -34,13 +34,74 @@ try {
 }
 
 // Issue an HTTP/1.1 request over a Unix domain socket and collect the full response.
+//
+// A server that responds early (e.g. 413 for an over-limit body) is allowed to close the connection
+// before the client has finished flushing the request body — correct per HTTP, but it races the
+// client's outstanding write against the close. Node's http client normally forwards any socket error
+// straight to `req`'s own 'error' event — except for the specific window right after a keep-alive
+// response finishes, where the client detaches its listener before handing the socket back to the
+// agent's free-connection pool and reattaching one (`responseKeepAlive` in Node's `_http_client.js`,
+// which has its own upstream `// TODO(ronag): ... the socket has no 'error' handler` comment on exactly
+// this gap). A write that fails in that window is unhandled and becomes an uncaught exception
+// attributed to whichever test happens to be running — that's the macOS flake this helper exists to
+// close. `agent: false` below is therefore the primary fix: without connection pooling, the gap window
+// never opens. The socket-level listener is retained as a second layer, in case an equivalent
+// unhandled-window exists on a platform/libuv path this can't be verified against from Linux.
+//
+// Settles exactly once, from whichever channel gets there first:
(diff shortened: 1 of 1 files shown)


cb1kenobi (approved):
LGTM

PR #2024, opened 2026-07-31
fix(threads): worker threads must inherit the parent's in-process config overrides

kriszyp:
## What

The `unitTests/apiTests` suite's isolation (per-PID rootPath, per-process loopback address, port overrides) is applied with in-process `env.setProperty()` calls on the main thread. Worker threads spawned during the run booted with a fresh `environmentManager`/`configUtils` module state and re-read `harper-config.yaml` from disk, so a worker silently fell back to whatever was actually installed and bound its ports on the wildcard address — the suite only passed when the developer's ambient install happened to be shaped like CI's.

Two closely-related root causes, both closing the same invariant ("a worker's effective config equals its parent's effective config"):

1. **Worker threads never received the parent's overrides at all.** `environmentManager.ts`'s `flatConfigObj` is module-scoped (each worker thread is its own V8 realm), populated once from disk. `setProperty()` only ever mutated the calling thread's own copy.
   - `setProperty()` now records every override into a canonical-name-keyed map (`getConfigOverrides()`), so different legacy aliases for the same param (e.g. `SERVER_PORT_KEY` / `OPERATIONSAPI_NETWORK_PORT`) collapse onto one last-write-wins entry instead of two independently-ordered ones.
   - `manageThreads.js` registers a `configOverrides` workerData provider via the existing `registerWorkerDataProvider` extension point, so every worker spawned from any thread — HTTP, job, or nested — inherits its parent's overrides.
   - `initSync()` replays inherited overrides once per thread, and replays the accumulated override set again after any forced re-init (e.g. a component RESTART), since that re-reads the whole config from disk and would otherwise silently drop everything layered on top of it — on any thread, including main. **A reload that changed the param itself wins:** each override records the configured value it displaced, and if the re-read produced a different value for that param the override retires (and is dropped from the map, so it stops reaching newly spawned workers). Without that, an operator editing an overridden key in `harper-config.yaml` and triggering a component restart would get a silent no-op for the life of the process. Values are compared with key-order-insensitive structural equality, so reformatting the YAML is not mistaken for a change.

2. **Even on the main thread, overrides never reached the nested config tree `componentLoader.ts` reads for built-in components.** `configUtils.ts` keeps config in two representations: a flat map and a nested tree (`configObj`, via `getConfigObj()`). `updateConfigObject()` — what `setProperty()` calls — only ever wrote the flat map. `componentLoader.ts`'s root-component load reads the nested tree directly and derives each built-in's network port/protocol from it (e.g. `operationsApi.network.securePort` decides whether `operationsServer` builds a TLS-only Fastify instance). So even with (1) fixed, a worker — or the main thread itself — whose ambient install had a non-null `operationsApi.network.securePort` still built a TLS-only listener, because componentLoader kept reading the stale on-disk value regardless of any override. Invisible on a CI-shaped install only because the on-disk value already happened to match what the harness wants.
   - `updateConfigObject()` now also mirrors into the nested tree, at the path implied by the flat key's underscore segments — with guards added over several review rounds: it never auto-vivifies `configObj` before a real config has ever loaded (would permanently short-circuit `getConfigObj()`'s lazy init), never clobbers an existing scalar shorthand (e.g. `threads: 4`) by descending through it, deletes rather than writes an `undefined` value (and doesn't auto-vivify ancestors just to immediately delete the leaf), and excludes params that live in `BOOT_PROP_PARAMS` rather than the actual YAML schema (`settings_path` is boot-props-file bookkeeping, not a real `settings:` config section — mirroring it naively would register a bogus `settings` component).

Three smaller, related fixes:

- `setupTestApp.mjs` nulled `http.securePort` when overriding `http.port`, but never did the same for `operationsApi.network.securePort` — an asymmetry that (combined with fix 2) left a stale TLS port in the effective config next to the plain-port override.
- `mqtt-test.mjs`'s two mTLS tests selected the client certificate by the literal string `'localhost'`. Switched to falling back through `getThisNodeName()` (the function the server itself uses to name its self-signed node cert) and then to any non-authority certificate, since `getThisNodeName()`'s own fallback chain can itself resolve differently between the install process and the test process on a hostname-null install (a fresh/non-replicated install has exactly one non-authority cert, so "any" is unambiguous here).
- `dataLayer/harperBridge/lmdbBridge/lmdbUtility/initializePaths.js` and `utility/lmdb/environmentUtility.ts` were calling `env.setProperty()` to cache values *derived* from config they'd just read (a schema path resolved from CLI args, a database's on-disk path) rather than genuine operator/harness overrides. Going through `setProperty()` made those derived values outlive a forced config reload and get shipped to every worker as if they were intentional overrides. Both now call `configUtils.updateConfigObject()` directly, which updates the same in-memory state without joining the override-tracking/propagation mechanism.

## Production-path impact

`config/configUtils.ts`'s `updateConfigObject()` change affects the same code path used at install time (`utility/install/installer.ts`) and any other `env.setProperty()` caller. This is a correctness fix within that path, not new behavior: it makes `configObj`/`flatConfigObj` consistent with each other, which is what `setProperty()`'s contract already implied. A worker that receives no overrides sees the new `workerData` provider return `undefined` and get skipped (per its existing "throws or returns non-cloneable → logged and skipped" contract) — production instances without in-process overrides see no behavior change.

Two further behavior notes on `setProperty()`, both from review:

- **`configObj` now means the live config tree.** It used to be assigned as a side effect of `flattenConfig()`, which also runs on documents that are thrown away moments later (the defaults document, the user-supplied document `installer.ts` reads for an `HDB_CONFIG` install), so the nested mirror's guard could not tell the live tree from a discarded one. Installing the live config is now explicit (`setActiveConfig()`), owned only by the three callers that build it.
- **`setProperty()` now rejects a value that is not structured-cloneable, and records a snapshot of the value.** The recorded overrides are `structuredClone`d into every worker, where a clone failure is only logged — landing the worker on the on-disk config, the exact divergence this change prevents, with `rootPath`/`storage.path` in the dropped payload. Failing at the caller instead keeps that impossible. Snapshotting also stops `initializePaths.js`/`environmentUtility.ts` from leaking derived per-database paths into the override set by mutating the live `databases` object in place.

## Scaffolding, not architecture: the worker-propagation half has a removal condition

This PR contains two independent fixes, and they have different lifespans. Please read them separately.

**Fix 2 (`updateConfigObject()` mirrors into the nested tree) is permanent.** It closes a real main-thread correctness bug: `setProperty()`'s documented contract ("will also update the config object configUtils maintains") was not honored, so an override reached the flat map but not the nested tree `componentLoader.ts` actually reads. Correcting an earlier version of this description: the installer's own `setProperty()` calls are *not* the justification. `installer.ts` has four (`storage.engine`, `install_user`, `settings_path`, `BOOT_PROPS_FILE_PATH`) — there is no RocksDB-compression call — and `storage.engine` is set before `createConfigFile()` has built any config tree, so the mirror is correctly skipped there and the same value reaches the file through `installParams` anyway. The reachable callers today are the test harness and harper-pro's `node_hostname`/`node_url` back-compat calls (`threads.count: 0`), plus any future runtime `setProperty()` caller — which is exactly why the contract has to hold rather than be documented as conditional.

**Fix 1 (worker threads inherit overrides via a `configOverrides` `workerData` provider) is a stopgap.** It exists solely because `unitTests/apiTests` boots Harper in-process and isolates itself with in-process `env.setProperty()` calls. Production instances have no in-process overrides, so this machinery carries a test-only need on the production worker-spawn path. It is the right call *today* — without it the suite fails on any install not shaped like CI's (181/10 observed on an adversarial install) — but it should not be built upon or treated as the intended design.

The end state is HarperFast/harper#2123, "Migrate unitTests/apiTests to the standard integration testing framework": move the suite to `integrationTests/` and `@harperfast/integration-testing`, which gives each suite a real child-process Harper with its own temp install directory and a file-locked loopback address. Real config in a real install propagates to workers by construction, and the override-inheritance problem stops existing.

**Removal condition:** when #2123 lands, the override tracking in `utility/environment/environmentManager.ts` and the `configOverrides` provider registration in `server/threads/manageThreads.js` come back out. That is listed in #2123's definition of done.

Worth noting the same architecture is the reason `test:unit:apitests` is prefixed with `node ./dist/bin/harper.js stop` — the suite stops the shared ambient install before running, so concurrent runs on a shared dev box kill each other's Harper regardless of address isolation. That is also fixed by #2123, not by this PR.

## Known limitation (out of scope for this PR)

Root components that use the newer Plugin API (`handleApplication`-based — MQTT, REST, fastifyRoutes, graphqlQuerying) read their config via `OptionsWatcher`, which re-parses `harper-config.yaml` off disk directly and never consults `getConfigObj()`/`flatConfigObj` at all. `env.setProperty()` overrides are invisible to that path. This PR's apitests verification (three different install shapes, including concurrent runs) didn't surface it because MQTT is still loaded through the older, `componentLoader`-driven path this PR does fix — but it's a real, pre-existing gap for any Plugin-API component, worth its own follow-up issue.

This PR fixes worker/main config divergence; no issue tracked that bug on its own, so the reference below is to the end-state issue. Refs #2123

## Testing

Verified against three different local installs (`node ./dist/bin/harper.js install` + isolated `HOME`, never the shared ambient install):
- **Adversarial** (`operationsApi.network.port: null` / `securePort: 9925`, `http.port: null` / `securePort: 9926`, `node.hostname` set to a non-`'localhost'` value — matches this issue's exact regression scenario): `npm run test:unit:apitests` 194 passing / 0 failing (was 181/10 on unpatched origin/main; 185/9 with only the worker-inheritance fix, before the nested-tree fix).
- **CI-shaped** (`DEFAULTS_MODE=dev`, `NODE_HOSTNAME=localhost`): 194/0, unaffected.
- **Hostname-null** (`node.hostname` unset entirely): 194/0 — this is what caught the mTLS cert-fallback issue above.
- `npm run test:unit:main` (4043 passing) and `npm run test:unit:resources` (1336 passing): unaffected, matches origin/main baseline (verified via a short-path throwaway worktree, since this repo's own worktree convention nests deep enough to independently trip two unrelated, pre-existing path-length issues — a domain-socket path limit and a Node module-sandbox check on a bare verification copy without its own `node_modules` — neither caused by this change).
- No more `[http/N] ... EADDRINUSE ... :::1883` in apitests output; no server banner reports a Root Path outside the per-PID `unitTests/envDir/<pid>` directory.
- Two concurrent `npm run test:unit:apitests` runs get distinct loopback addresses with no `EADDRINUSE` and correct per-run Root Path banners — the wildcard-bind collision this PR targets is gone. A separate mqtt-test.mjs timing sensitivity was observed under 2x concurrent full-suite load on this shared dev machine; it isn't port/address-collision-shaped (addresses were confirmed distinct) and wasn't root-caused — flagged as a follow-up, not blocking.
- New unit tests for the override-tracking/replay mechanism (`environmentManager.test.js`) and the nested-mirror guards (`configUtils.test.js`).
- An end-to-end test boots a real worker against a config file the test writes itself and asserts both directions: with no inherited overrides it reads the on-disk values, and with them it resolves the parent's values instead.
- Re-verified after the review-feedback round: apitests 194/0 on both a CI-shaped and the adversarial install, `test:unit:resources` 1336 passing, `test:unit:main` unchanged against its baseline.

This branch went through seven rounds of cross-model pre-push review (codex + gemini + grok + an internal Harper-domain pass) as fixes were made — see the commit messages for each round's findings and disposition.

<sub>Human-Review-Need: 3 @ 60881f8ce23d</sub>

### DIFF
--- a/config/configUtils.ts
+++ b/config/configUtils.ts
@@ -704,7 +704,7 @@ export function updateConfigObject(param: string, value: any) {
 	// configObj's falsiness as "not yet initialized" and lazily calls initConfig(), so creating
 	// an empty configObj here (before install writes the config file / initConfig ever runs)
 	// would permanently short-circuit that lazy init to an empty tree.
-	if (configObj !== undefined && !NON_NESTED_CONFIG_PARAMS.has(configObjKey)) {
+	if (configObj != null && !NON_NESTED_CONFIG_PARAMS.has(configObjKey)) {
 		const pathSegments = configObjKey.split('_');
 		let node = configObj;
 		for (let i = 0; i < pathSegments.length - 1; i++) {
--- a/config/configUtils.ts
+++ b/config/configUtils.ts
@@ -50,6 +50,14 @@ let flatDefaultConfigObj;
 let flatConfigObj;
 let configObj;
 
+// Canonical param names that live in CONFIG_PARAM_MAP but do NOT correspond to a path in the
+// harper-config.yaml schema (BOOT_PROP_PARAMS is boot-props-file-only bookkeeping — see its own
+// comment). Splitting one of these on '_' and writing it into the nested configObj tree would
+// silently create a bogus top-level section (e.g. 'settings_path' -> configObj.settings.path),
+// which componentLoader.ts treats as a real component to load, since it iterates every truthy
+// top-level key of the root config.
+const NON_NESTED_CONFIG_PARAMS = new Set<string>(Object.values(hdbTerms.BOOT_PROP_PARAMS));
+
 export function resolvePath(relativePath: string) {
 	if (relativePath?.startsWith('~/')) {
 		return path.join(hdbUtils.getHomeDir(), relativePath.slice(1));
@@ -696,16 +704,23 @@ export function updateConfigObject(param: string, value: any) {
 	// configObj's falsiness as "not yet initialized" and lazily calls initConfig(), so creating
 	// an empty configObj here (before install writes the config file / initConfig ever runs)
 	// would permanently short-circuit that lazy init to an empty tree.
-	if (configObj !== undefined) {
+	if (configObj !== undefined && !NON_NESTED_CONFIG_PARAMS.has(configObjKey)) {
 		const pathSegments = configObjKey.split('_');
 		let node = configObj;
 		for (let i = 0; i < pathSegments.length - 1; i++) {
 			const segment = pathSegments[i];
-			// Only auto-vivify a missing segment. An existing non-object value here is a legacy
-			// scalar shorthand for this key (e.g. `threads: 4`) — descending through it would
-			// silently replace that value with `{}`, discarding it out from under other readers.
-			if (node[segment] === undefined) node[segment] = {};
-			else if (typeof node[segment] !== 'object' || node[segment] === null) return;
+			if (node[segment] === undefined) {
+				// Nothing to delete on a path section that doesn't exist yet — and auto-vivifying it
+				// just to immediately delete the leaf would leave the (now-empty) intermediate
+				// objects behind, which is exactly the bogus-top-level-section risk above.
+				if (value === undefined) return;
+				node[segment] = {};
+			} else if (typeof node[segment] !== 'object' || node[segment] === null) {
+				// An existing non-object value here is a legacy scalar shorthand for this key (e.g.
+				// `threads: 4`) — descending through it would silently replace that value with `{}`,
+				// discarding it out from under other readers.
+				return;
+			}
 			node = node[segment];
 		}
 		// Match squashObj's own convention (below) of never writing an `undefined` value as an
--- a/config/configUtils.ts
+++ b/config/configUtils.ts
@@ -198,8 +198,7 @@ export function createConfigFile(args, skipFsValidation = false) {
 	// Validates config doc and if required sets default values for some parameters.
 	validateConfig(configDoc, skipFsValidation);
 
-	const configObj = configDoc.toJSON();
-	flatConfigObj = flattenConfig(configObj);
+	flatConfigObj = setActiveConfig(configDoc.toJSON());
 
 	// Create new config file and write config doc to it.
 	const hdbRoot = configDoc.getIn(['rootPath']) as string;
@@ -505,9 +504,9 @@ export function initConfig(force = false) {
 
 		// Validates config doc and if required sets default values for some parameters.
 		validateConfig(configDoc);
-		const configObj = configDoc.toJSON();
-		(server as any).config = configObj;
-		flatConfigObj = flattenConfig(configObj);
+		const parsedConfig = configDoc.toJSON();
+		(server as any).config = parsedConfig;
+		flatConfigObj = setActiveConfig(parsedConfig);
 
 		// If config has old version of logrotate enabled let user know it has been deprecated.
 		if (flatConfigObj['logging_rotation_rotate']) {
@@ -700,10 +699,11 @@ export function updateConfigObject(param: string, value: any) {
 	// nested-path readers derive behavior — e.g. a component's network port/protocol — from
 	// getConfigObj()'s tree, not the flattened map, so an override that only touched
 	// flatConfigObj was invisible to them (the on-disk shape kept winning regardless of
-	// setProperty()). Only mirror once a real config has been loaded — getConfigObj() treats
-	// configObj's falsiness as "not yet initialized" and lazily calls initConfig(), so creating
-	// an empty configObj here (before install writes the config file / initConfig ever runs)
-	// would permanently short-circuit that lazy init to an empty tree.
+	// setProperty()). configObj is null/undefined only before a live config has ever been
+	// installed (setActiveConfig) — during install, before the config file is written. Don't
+	// auto-vivify it here: getConfigObj() treats configObj's falsiness as "not yet initialized"
+	// and lazily calls initConfig(), so an empty object here would permanently short-circuit
+	// that lazy init to an empty tree.
 	if (configObj != null && !NON_NESTED_CONFIG_PARAMS.has(configObjKey)) {
 		const pathSegments = configObjKey.split('_');
 		let node = configObj;
@@ -895,7 +895,7 @@ export function updateConfigValue(
 	}
 	atomicWriteFile(configFileLocation, String(configDoc));
 	if (update_config_obj) {
-		flatConfigObj = flattenConfig(configDoc.toJSON());
+		flatConfigObj = setActiveConfig(configDoc.toJSON());
 	}
 	logger.trace(`Config parameter: ${param} updated with value: ${value}`);
 }
@@ -914,9 +914,28 @@ function backupConfigFile(configPath, hdbRoot) {
 	}
 }
 
+/**
+ * Flattens `obj` and installs it as the live config: the nested tree getConfigObj() hands out and
+ * updateConfigObject() mirrors overrides into, alongside the flat map returned here.
+ *
+ * Only the three callers that own the live config may do this. flattenConfig() also runs on
+ * throwaway docs — the defaults doc in getDefaultConfig()/createConfigFile(), and the user-supplied
+ * doc installer.ts reads for HDB_CONFIG — so setting configObj as a side effect of flattening left
+ * it pointing at a tree that was discarded moments later, which is indistinguishable from the live
+ * tree to every reader of configObj.
+ * @param obj
+ * @returns the flattened config
+ */
+function setActiveConfig(obj) {
+	const flatObj = flattenConfig(obj);
+	configObj = obj;
+	return flatObj;
+}
+
 const PRESERVED_PROPERTIES = ['databases'];
 /**
  * Flattens the JSON version of Harper config with underscores separating each parent/child key.
+ * Does NOT install `obj` as the live config — see setActiveConfig() for that.
  * @param obj
  * @returns {null}
  */
@@ -925,10 +944,7 @@ export function flattenConfig(obj) {
 	if (obj?.operationsApi?.network) obj.operationsApi.network = { ...obj.http, ...obj.operationsApi.network };
 	if (obj?.operationsApi) obj.operationsApi.tls = { ...obj.tls, ...obj.operationsApi.tls };
 
-	configObj = obj;
-	const flatObj = squashObj(obj);
-
-	return flatObj;
+	return squashObj(obj);
 
 	function squashObj(obj) {
 		let result = {};
--- a/config/configUtils.ts
+++ b/config/configUtils.ts
@@ -695,37 +695,36 @@ export function updateConfigObject(param: string, value: any) {
 
 	flatConfigObj[configObjKey.toLowerCase()] = value;
 
-	// Keep the nested config tree in sync too. componentLoader (root components) and other
+	// Keep the nested config tree in sync too: componentLoader (root components) and other
 	// nested-path readers derive behavior — e.g. a component's network port/protocol — from
-	// getConfigObj()'s tree, not the flattened map, so an override that only touched
-	// flatConfigObj was invisible to them (the on-disk shape kept winning regardless of
-	// setProperty()). configObj is null/undefined only before a live config has ever been
-	// installed (setActiveConfig) — during install, before the config file is written. Don't
-	// auto-vivify it here: getConfigObj() treats configObj's falsiness as "not yet initialized"
-	// and lazily calls initConfig(), so an empty object here would permanently short-circuit
-	// that lazy init to an empty tree.
+	// getConfigObj()'s tree, not the flattened map, so an override that only touched flatConfigObj
+	// was invisible to them. configObj is unset only before a live config has ever been installed
+	// (setActiveConfig), i.e. during install; don't auto-vivify it, since getConfigObj() reads its
+	// falsiness as "not yet initialized" and an empty object would short-circuit that lazy init
+	// permanently.
 	if (configObj != null && !NON_NESTED_CONFIG_PARAMS.has(configObjKey)) {
 		const pathSegments = configObjKey.split('_');
 		let node = configObj;
 		for (let i = 0; i < pathSegments.length - 1; i++) {
 			const segment = pathSegments[i];
 			if (node[segment] === undefined) {
-				// Nothing to delete on a path section that doesn't exist yet — and auto-vivifying it
-				// just to immediately delete the leaf would leave the (now-empty) intermediate
-				// objects behind, which is exactly the bogus-top-level-section risk above.
+				// Auto-vivifying ancestors only to delete the leaf would leave empty sections behind,
+				// which componentLoader.ts would treat as components to load.
 				if (value === undefined) return;
 				node[segment] = {};
 			} else if (typeof node[segment] !== 'object' || node[segment] === null) {
-				// An existing non-object value here is a legacy scalar shorthand for this key (e.g.
-				// `threads: 4`) — descending through it would silently replace that value with `{}`,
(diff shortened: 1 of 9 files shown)


config/configUtils.ts
```diff
@@ -687,6 +695,41 @@ export function updateConfigObject(param: string, value: any) {
...
+
+	// Keep the nested config tree in sync too. componentLoader (root components) and other
+	// nested-path readers derive behavior — e.g. a component's network port/protocol — from
+	// getConfigObj()'s tree, not the flattened map, so an override that only touched
+	// flatConfigObj was invisible to them (the on-disk shape kept winning regardless of
+	// setProperty()). Only mirror once a real config has been loaded — getConfigObj() treats
+	// configObj's falsiness as "not yet initialized" and lazily calls initConfig(), so creating
+	// an empty configObj here (before install writes the config file / initConfig ever runs)
+	// would permanently short-circuit that lazy init to an empty tree.
+	if (configObj !== undefined && !NON_NESTED_CONFIG_PARAMS.has(configObjKey)) {
```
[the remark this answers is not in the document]
kriszyp: Fixed in c6ffd852f — switched to `configObj != null`, matches the suggestion exactly.

utility/environment/environmentManager.ts
```diff
@@ -108,6 +172,20 @@ export function initSync(force: boolean = false) {
 	try {
 		if (propFileExists || doesPropFileExist() || commonUtils.noBootFile() || force) {
 			configUtils.initConfig(force);
+			if (!inheritedOverridesApplied) {
+				inheritedOverridesApplied = true;
+				applyInheritedConfigOverrides();
+			}
+			// force implies the on-disk config was just re-read from scratch, which would otherwise
+			// silently drop every override applied on this thread so far (inherited or local).
+			if (force) reapplyAllOverrides();
```
DavidCockerill, line 255: `reapplyAllOverrides()` here changes what `setProperty()` means. Its docstring says it's for "the installer and unit tests"; after this line, any key ever passed to it stops tracking `harper-config.yaml` for the life of the process on that thread.

**Trigger.** `initSync(true)`'s only non-test caller is the worker-`RESTART` handler at `security/keys.ts:70` (main thread). Main's `appliedOverrides` is non-empty in two production shapes:

- **in-process first-run install** (`harper run .` — see the comment at `installer.ts:529`), which leaves `storage.engine` plus the three boot props in the map;
- **`threads.count: 0`**, where `setMainIsWorker(true)` → `loadRootComponents(true)` runs root components' `start()` on the main thread (`server/threads/socketRouter.ts:41`), and harper-pro's `replication/replicator.ts:107,112` calls `env.setProperty('node_hostname'|'node_url')` for back-compat.

In both, an operator who edits that key in `harper-config.yaml` and triggers a component restart gets a **silent no-op**: the forced re-init reads the new value and the replay immediately overwrites it, with no log line. Before this change the reload won.

The property I'd want preserved is that an override survives a reload the override's owner didn't ask for, but not one that *is* the config change. Skipping replay for keys whose on-disk value moved, or scoping replay to the harness's own overrides, both satisfy that — mechanism is your call. If sticky-always is the intended semantics, that's defensible; please say so in the PR body and in `setProperty()`'s docstring so it isn't an unstated side effect of a test fix.

🧊 **In plain terms:** a setting changed in memory now beats the settings file even after Harper deliberately re-reads that file, so editing the file and restarting components can silently do nothing until the whole process restarts.

config/configUtils.ts
```diff
@@ -687,6 +695,41 @@ export function updateConfigObject(param: string, value: any) {
...
+
+	// Keep the nested config tree in sync too. componentLoader (root components) and other
+	// nested-path readers derive behavior — e.g. a component's network port/protocol — from
+	// getConfigObj()'s tree, not the flattened map, so an override that only touched
+	// flatConfigObj was invisible to them (the on-disk shape kept winning regardless of
+	// setProperty()). Only mirror once a real config has been loaded — getConfigObj() treats
+	// configObj's falsiness as "not yet initialized" and lazily calls initConfig(), so creating
+	// an empty configObj here (before install writes the config file / initConfig ever runs)
+	// would permanently short-circuit that lazy init to an empty tree.
+	if (configObj != null && !NON_NESTED_CONFIG_PARAMS.has(configObjKey)) {
```
DavidCockerill, line 705: The comment says "only mirror once a real config has been loaded," but `configObj` doesn't carry that meaning. It's assigned **only** as a side effect of `flattenConfig()` (`config/configUtils.ts:928`), and `flattenConfig()` also runs on throwaway trees:

- the default-config doc in `getDefaultConfig()` (line 271) and in `createConfigFile()` (line 149);
- a user-supplied doc in `installer.ts:236` (`getConfigFromFile()`, the `HDB_CONFIG` install path).

So at `installer.ts:199` — the call site the PR body cites as fix 2's permanent, production justification:

- **plain install:** no config file exists yet, `configObj` is `undefined`, the mirror is **skipped**;
- **`HDB_CONFIG` install:** `getConfigFromFile()` has already pointed `configObj` at a doc that is discarded; the mirror writes into it, `createConfigFile()` re-points `configObj` at line 202, and the write is silently lost while the flat map keeps it — precisely the flat/nested divergence this fix exists to close.

Separately I couldn't find the "RocksDB compression" `setProperty` the body mentions; `installer.ts` has four `setProperty` calls total (`storage.engine`, `install_user`, `settings_path`, `BOOT_PROPS_FILE_PATH`) and they're the only non-test callers in core.

This matters beyond wording: #2123's definition of done keeps this mirror (while removing the worker-propagation half) on the strength of that install-path claim. Either the guard needs to key off something that actually identifies the live config tree, or the body and #2123 should say the mirror is reachable only from *runtime* `setProperty` callers — today the tests, plus harper-pro's `node_hostname`/`node_url`.

The `BOOT_PROP_PARAMS` exclusion itself is right and load-bearing, by the way: component activation is presence-gated on top-level keys, and `bin/run.ts:210` calls `updateConfigObject('settings_path', …)` on the production `harper run <dir>` path.

🧊 **In plain terms:** the safety check here is testing a variable that can be pointing at a temporary copy of the settings, so on the install path this fix is advertised for, it either doesn't run or writes into a copy that gets thrown away.

server/threads/manageThreads.js
```diff
@@ -232,6 +232,12 @@ function registerWorkerDataProvider(name, provider) {
 		if (workerDataProviders.get(name) === provider) workerDataProviders.delete(name);
 	};
 }
+// Propagate this thread's in-process config overrides (env.setProperty — installer bootstrap or
+// the unit-test harness's per-run isolation) to every worker spawned from here, so a worker's
+// effective config is never silently whatever happens to be installed on disk. Worker-side replay
+// is environmentManager.ts's applyInheritedConfigOverrides(), invoked from initSync(). Registered
+// on every thread (not just main) so a nested worker-of-a-worker also inherits the full chain.
+registerWorkerDataProvider('configOverrides', () => envMgr.getConfigOverrides());
```
DavidCockerill, line 242: Non-blocking contract note. `collectProvidedWorkerData` catches a throwing or non-cloneable provider, logs at `error`, and spawns the worker anyway — with the entire `configOverrides` payload absent. That's the right contract for an *optional* provider, but for this one the failure lands the worker on "silently whatever happens to be installed on disk," which is the exact sentence the registration comment above says this exists to prevent, and the payload carries `rootPath` / `storage.path`.

There's **no reachable trigger today** — every value that can enter `appliedOverrides` is a string, number, boolean, `null`, or a plain object, all `structuredClone`-safe — so this is about which way the mechanism errs, not a live bug. Two options that both fail closed: let a clone failure fail the spawn for this specific provider, or validate cloneability at `setProperty()` time so it can never be a spawn-time surprise.

Worth noting the worker *side* already fails closed — `initSync`'s catch calls `process.exit(1)` — so the loud direction is half-established; it's only the parent side that degrades quietly.

🧊 **In plain terms:** if handing the settings to a new worker ever fails, the worker starts anyway using the settings on disk instead of refusing to start — and those settings include where the database files live.

DavidCockerill (commented):
Careful work on a genuinely nasty class of bug, and **the worker-inheritance half holds up under tracing.** I went after the four things most likely to be half-fixed here and all four are right:

- **Respawn re-collects the payload** — `startCopy → startWorker → collectProvidedWorkerData`, plus the unexpected-exit auto-restart, and the provider closure reads the live Map each time. That's the classic half-fix in this shape and it isn't present.
- **Nested workers chain**, because the provider registers on every thread and a worker re-records what it inherits.
- **The replay is anchored to `initSync()`** rather than a worker entry point, so it lands immediately after the on-disk read and *before* any path getter memoises. `initializePaths.js` calls `initSync()` at module load and memoises lazily, so inherited `rootPath`/`storage.path` are in place before any path resolves.
- **The canonical `CONFIG_PARAM_MAP` keying** genuinely collapses the `SERVER_PORT_KEY`/`OPERATIONSAPI_NETWORK_PORT` alias pair that `initTestEnvironment` sets five lines apart.

No stale-path window survives the change. Worth stating what that's worth: before this, an apitests worker could genuinely open LMDB under the ambient install's root — a second env in the wrong directory.

Two findings, both about the *other* half of the PR, in threads. Neither blocks the mechanism:

1. `reapplyAllOverrides()` on a forced re-init promotes `setProperty` from "in-memory override" to "outranks the config file for the process lifetime" — and there are two production shapes where main's override map is non-empty, so an operator's config edit plus a component restart becomes a silent no-op.
2. The `configObj != null` guard doesn't identify the live config tree, so on the install path the body cites as fix 2's justification it either doesn't fire or writes into a doc thrown away moments later. That one matters beyond wording, because #2123's definition of done keeps the mirror on that basis.

Plus one non-blocking note on which way the workerData provider errs.

Also dropped several candidates after tracing them, so nobody re-runs them: the phantom `databases` root component (test/upgrade-only path, and `databases` isn't a registered plugin); `undefined` round-tripping (`structuredClone` preserves it as an own key and the replayed `setProperty` reaches the `delete` branch, so the worker lands in the parent's state); a worker loading a *different* config file (resolution goes through `getHomeDir()`, which is inherited env); and the mqtt cert fallback weakening the negative mTLS assertion (`bad_client` is asserted independently).

One small thing: I couldn't find a `Closes`/`Refs` issue for this. The body cites #2123 as the end state, but nothing seems to track the worker-config-divergence bug the PR actually fixes.

DavidCockerill (approved):
Approving — both findings closed, and the reload redesign is the right shape.

Recording `base` (the configured value an override displaced, first write only) and retiring the override when the on-disk value has moved is exactly the property I asked for. The part I checked hardest: `sameConfigValue` returns false — treat as changed, retire — for everything it can't decide (mixed null/object, cyclic, BigInt). Reload-wins is the correct direction here, and the comparison is total, which matters given `initSync()`'s catch calls `process.exit(1)`. Walked the `threads.count: 0` / harper-pro `node_hostname` shape: operator edit wins, unrelated reload leaves the override alone.

Aliasing is fully closed — `value` and `base` are both cloned, and on replay `setProperty` re-clones into the map while handing the previous snapshot to `updateConfigObject`, so the map entry and the live-config value are never the same object.

Finding 2 closed at the root: `flattenConfig()` lost its side effect and `setActiveConfig()` is now the only assignment to `configObj`. I regression-checked every `getConfigObj()` consumer and the `initOldConfig` path — nothing depended on the removed behaviour. Correcting the PR body to drop the install-path justification and the RocksDB-compression claim matters as much as the code, since that's what #2123's DoD carries.

And the provider note closed the better way: rejecting non-cloneable values in `setProperty()` makes the fail-open path unreachable rather than special-casing it.

The worker test got stronger, not weaker — the fixture change moved the config *file*, not the assertion; it still spawns a real Worker calling the real `initSync()`, and adds the no-overrides direction.

Still no linked issue for the bug this fixes, which the body now says plainly. Worth filing.

Issue #2025, opened 2026-07-31
Node v26.5.1 (llhttp 9.4.3 + undici 8.9.0) breaks integration tests across harper and harper-pro on Node 26


kriszyp:
## Summary

CI went red across `harper` and `harper-pro`, on both `main` and `v5.1`, starting 2026-07-30 — **Node 26 shards only** (22 and 24 stay green). This is **not** a Harper regression: it is caused by **Node v26.5.1**, released 2026-07-28.

## Evidence — same commit, different runtime

Commit `b04af4d08` on `harper` `main`, two `Integration Tests` runs of the *identical* SHA:

| run | Node | result |
|---|---|---|
| [30501224101](https://github.com/HarperFast/harper/actions/runs/30501224101) — 2026-07-29 23:58 | **v26.5.0** | pass |
| [30524017777](https://github.com/HarperFast/harper/actions/runs/30524017777) — 2026-07-30 07:45 | **v26.5.1** | fail |

Same delta reproduces on `harper-pro` `v5.1`: run 30405745607 (2026-07-28, v26.5.0) green → run 30602767661 (2026-07-31, v26.5.1) red.

CI resolves `node-version: 26` to the latest 26.x, so runners picked up 26.5.1 without any change on our side.

## What is in v26.5.1

It is a security-only patch with exactly two changes, both HTTP:

- `deps: update llhttp to 9.4.3` (via nodejs-private#935 — security)
- `deps: update undici to 8.9.0`

llhttp is the server-side HTTP parser; undici backs `fetch`. Every failing test exercises one or both.

## Failing tests

**harper**
- `integrationTests/database/ttlResetOnWrite.test.ts:342` — `ops-update` / `SQL-UPDATE` / `addTo`: `expected TTL reset, got reset=false`
- `integrationTests/components/static-cache-headers.test.ts:86` — `Cache-Control never became "public, max-age=86400, immutable"; last seen: "public, max-age=300"`

**harper-pro**
- `integrationTests/analytics/analytics.test.mjs:93` — `can track child process CPU time`
- `integrationTests/cluster/addNodeLeaderNoMeshLeak.test.mjs:125` — `source node stays local to bridge; bridge full-copies its data` (180s timeout)
- `integrationTests/cluster/addNodeLeaderNoMeshLeak.test.mjs` — `add_node isLeader does not leak source node to mesh`
[...]

PR #2029, opened 2026-07-31
fix(test): stop using fetch() in ttlResetOnWrite / static-cache-headers (Node 26.5.1 flakiness)

kriszyp:
## What / why

Issue #2025 asked whether Harper is genuinely broken on Node v26.5.1, or whether the four
failing integration tests are test-only artifacts. Verdict: **neither, cleanly** — it's a
genuine Node regression, but in the bundled `undici`/global `fetch()` *client*, not in
llhttp/server-side parsing and not in any Harper code path.

Root cause, independently reproduced:

- Installed 26.5.0 and 26.5.1 locally and reran the named tests against the same build under
  each — confirmed the CI bisect independently.
- Ruled out llhttp/server-side parsing: a raw `node:http` client hitting the identical running
  Harper server (REST PUT/POST/GET, ops API, SQL, custom resource) shows **zero** latency
  regression on 26.5.1 vs 26.5.0.
- Ruled out Harper's TTL-reset / chokidar-reload logic: a debug-instrumented run showed the
  *update* landing on schedule while the *check's own GET* took 1-2.6s to complete on 26.5.1
  (2-6ms on 26.5.0) — the delayed check, not the write, crossed the reset-expiry boundary.
  Direct `fs.writeFile` + chokidar reload (bypassing `fetch()` entirely) completed in 2.8ms on
  26.5.1.
- Found the mechanism with a **Harper-independent minimal repro**: two bare
  `node:http.createServer()` instances, `fetch()` in a loop mixing GET/POST across the two
  origins. 26.5.0: 300/300 rounds clean, max 10.8ms. 26.5.1: repeated multi-hundred-ms to
  20+ second stalls. This is a `nodejs/undici` bug.
- Cross-checked the same CI run (30614629057) across Node lines: v22/v24 shard 6 both
  ~4500-5300ms clean; v26 (26.5.1) ~5000-6500ms and failing. Same run, shard 4: v22/v24
  ~2060ms on the cache-header reload test; v26 ~24400-24600ms and timing out.
- Identified the upstream issue: **nodejs/undici#5600** — an unref'd `setImmediate()` in
  `scheduleIdleSocketValidation` (introduced in undici #5499) means the idle-socket-reuse check
  only runs once something else wakes the loop's check phase; on an otherwise-idle loop that can
  be a long wait, and the queued request sits behind it. Node 26.5.1 bundles undici **8.9.0**
  (`nodejs/node#64712`, published 2026-07-24), which has the regression. Matches our repro
  exactly: mechanism (idle-socket-reuse stall), trigger (idle event loop), and symptom scaling
  with how idle the loop is (their repro: up to ~500ms with a single-connection pool under a
  test runner; ours: up to 20+s in a near-empty two-server harness). Already fixed upstream by
  **nodejs/undici#5609** (merged 2026-07-29, "stop unref'ing the idle socket validation
  immediate") — but not yet in a released undici or Node build as of this PR, so nothing to pin
  to yet.

Root-cause attribution across the four named CI failures:
1. harper `ttlResetOnWrite.test.ts` (ops-update/SQL-UPDATE/addTo) — **fixed here**, same root
   cause: `fetch()` interleaving GET(httpURL)/POST(opsURL).
2. harper `static-cache-headers.test.ts` — **fixed here**, same root cause: `fetch()`-based
   polling GET.
3. harper-pro `analytics.test.mjs` — very likely the same root cause (also `fetch()`-based,
   cross-origin; confirmed by a read-only grep, not fixed — different repo, out of scope here).
4. harper-pro `addNodeLeaderNoMeshLeak.test.mjs` — **independent**: doesn't use `fetch()` at
   all; a 180s cluster-mesh timeout, unrelated mechanism.

## Fix

Both harper test files now use `node:http` directly for their timing-critical requests
(matching what `supertest`, already used elsewhere in these suites, does under the hood) —
proven immune to the undici stall via the isolated repro above.

`ttlResetOnWrite.test.ts` also gets an independent robustness fix: the TTL-reset check now
polls across the whole valid window (bounded per-sample timeout) instead of sampling a single
instant, and distinguishes a genuine transport failure (`INCONCLUSIVE`) from an actual
NO-RESET signal, so a network hiccup never renders as a TTL data-integrity defect.

**Not** pinning CI to 26.5.0 — the two Harper tests destabilized by the upstream bug are now
fixed at the root, so no pin is needed to go green. Issue #2025 stays open: the undici
regression itself is upstream (unfixable in Harper source) and already tracked at
[nodejs/undici#5600](https://github.com/nodejs/undici/issues/5600), fixed by
[nodejs/undici#5609](https://github.com/nodejs/undici/pull/5609) — watch for that landing in a
released undici/Node build, no new upstream issue needed.

## Test plan

- 3+ consecutive clean runs of each file on Node 26.5.1 (rocksdb + lmdb for the TTL suite)
  after the fixes, matching 26.5.0/v22/v24 timing (~4.5s/~2s respectively, no more 5-27s
  stalls/timeouts).
- 1 control run of each on 26.5.0 (unaffected either way, as expected).
- `tsc --noEmit`, `oxlint`, `prettier --check` clean on both files.
- Independent pre-push review (codex + gemini + grok + harper-domain, adjudicated) on the
  first commit found 3 real findings (serial polling not actually bounding per-sample latency,
  transport failures rendered as data defects, an unbounded-stall regression reintroduced in
  the "fixed" file) plus 3 nits — all addressed in the second commit and reverified.

<sub>Review-Coverage: authored=unknown; ran=none; rounds=1 @ c404a409f692</sub>

<sub>Human-Review-Need: 4 @ c404a409f692</sub>

### DIFF
--- a/integrationTests/database/ttlResetOnWrite.test.ts
+++ b/integrationTests/database/ttlResetOnWrite.test.ts
@@ -64,6 +64,11 @@ const CHECK_RESET_MS = TTL_MS + 200; // 2200ms — past original expiry, before 
 // Max extra slack for expiry (one background-sweep leeway).
 const EXPIRY_POLL_MS = 5_000;
 const EXPIRY_POLL_INTERVAL_MS = 200;
+// Cap each gone-check sample well under the interval, same reasoning as RESET_POLL_TIMEOUT_MS: a
+// sample bound to httpRequest's full default timeout could consume the whole window on one slow
+// response and get misread as the record surviving past its TTL (F-002 stale resurrection).
+const EXPIRY_POLL_SAMPLE_TIMEOUT_MS = 400;
+const EXPIRY_POLL_MIN_REMAINING_MS = 50;
 
 // The "record is RESET" check polls across the whole valid window (past the original expiry, up
 // to just before the reset-expiry) rather than sampling a single instant. Polling only helps if
@@ -251,18 +256,36 @@ suite(
 			}
 		}
 
+		interface AbsenceResult {
+			observed: boolean;
+			// true only if every sample that completed returned neither 200 nor 404 (timeout or
+			// transport error) — i.e. we never got a clean read at all. Mirrors
+			// PresenceResult.inconclusive: without it, a stalled response consuming the whole window
+			// reads as "still present" and gets reported as F-002 stale resurrection, when it's
+			// really "couldn't measure."
+			inconclusive: boolean;
+		}
+
 		/**
-		 * Poll until the record is absent (404) or timeout.
-		 * Returns true if it went absent within the deadline.
+		 * Poll until the record is absent (404), bounding every attempt and sleep to what's left
+		 * before the deadline so no sample can start once the window has closed. Returns as soon as
+		 * any sample observes it absent. Each attempt is capped at EXPIRY_POLL_SAMPLE_TIMEOUT_MS —
+		 * and further capped to the remaining time near the deadline — so one slow/stalled response
+		 * can't consume the whole window on its own.
 		 */
-		async function pollUntilGone(id: string, maxMs: number): Promise<boolean> {
+		async function pollUntilGone(id: string, maxMs: number): Promise<AbsenceResult> {
 			const deadline = Date.now() + maxMs;
-			while (Date.now() < deadline) {
-				const { status } = await getRecord(id);
-				if (status === 404) return true;
-				await sleep(EXPIRY_POLL_INTERVAL_MS);
+			let sawCleanSample = false;
+			while (deadline - Date.now() >= EXPIRY_POLL_MIN_REMAINING_MS) {
+				const remainingMs = deadline - Date.now();
+				const { status } = await getRecord(id, Math.min(EXPIRY_POLL_SAMPLE_TIMEOUT_MS, remainingMs));
+				if (status === 404) return { observed: true, inconclusive: false };
+				if (status === 200) sawCleanSample = true;
+				const sleepBudgetMs = deadline - Date.now();
+				if (sleepBudgetMs <= 0) break;
+				await sleep(Math.min(EXPIRY_POLL_INTERVAL_MS, sleepBudgetMs));
 			}
-			return false;
+			return { observed: false, inconclusive: !sawCleanSample };
 		}
 
 		interface PresenceResult {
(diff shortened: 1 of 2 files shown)


kriszyp:
I'm going to wait to see if Node fixes this regression soon.

integrationTests/database/ttlResetOnWrite.test.ts
```diff
@@ -119,54 +159,80 @@ suite(
...
+				if (body !== undefined) {
+					headers['Content-Type'] = 'application/json';
+					headers['Content-Length'] = String(Buffer.byteLength(body));
+				}
+				let wallClockTimer: ReturnType<typeof setTimeout>;
+				const settle = (fn: () => void) => {
+					clearTimeout(wallClockTimer);
+					fn();
+				};
+				const req = http.request(
```
cb1kenobi: **Nit: hardcoded `node:http` means an https URL fails opaquely**

`fetch()` picked the client from the URL scheme; `http.request()` does not. If `httpURL` / `operationsAPIURL` is ever served over TLS, every request here fails with a transport error rather than a clear message — and in this file that surfaces as retry-exhausted INCONCLUSIVE, which is the one shape the rest of this rework works hardest to make legible.

Suggested fix: `import https from 'node:https'`, then `const client = u.protocol === 'https:' ? https : http;` and call `client.request(...)`. An explicit assert on `u.protocol` at the top of the helper would do just as well.

(Same in `getPath` in `static-cache-headers.test.ts`.)

—
_Generated by Barber AI_

kriszyp: Addressed in 1bbca3fe0: both direct-request helpers now select `node:http` or `node:https` from the parsed URL scheme, preserving the old scheme-aware behavior without returning to undici. Build and the affected integration suites pass on RocksDB and LMDB. — GPT-5 Codex

integrationTests/database/ttlResetOnWrite.test.ts
```diff
@@ -415,21 +647,47 @@ suite(
 					await deleteRecord(id);
 					continue;
 				}
+				// Gate on when the update was ISSUED, not when its response landed (updateAt): the
+				// server could have applied it well before the response returned, so gating on
+				// response time could discard a round that genuinely raced. seedAt (response time,
+				// the LATEST possible natural expiry) stays the comparison bound — the conservative
+				// choice, so this only marks a round windowMissed when it's certain to have missed.
+				if (updateIssuedAt >= seedAt + TTL_MS) {
```
heskew: Low: this compares against the latest possible seed expiry, so an update in the interval after `seedIssuedAt + TTL_MS` can still count as raced. For example, if the seed applies at t=0, responds at 250ms, and a delayed timer issues the update at 2050ms, the row may already have expired even though 2050ms is still below `seedAt + TTL_MS` (2250ms). The PUT can then recreate it and increment `cleanReset`, satisfying the coverage floor without exercising the intended race. Consider excluding or marking inconclusive any round with `updateIssuedAt >= seedIssuedAt + TTL_MS`; only that earliest-expiry bound proves the update was issued pre-expiry.

---

heskew (approved):
Reviewed at `1bbca3fe`. The direct HTTP helpers and merged result check out; one non-blocking race-probe coverage edge is noted inline.

---

cb1kenobi (approved):

PR #2039, opened 2026-08-01
feat(server): let liveSubscriptionAuth revoke a single subscriber without ending its subscription

kriszyp:
## Summary

`registerLiveSubscription` (the #1414 continuous re-authorization registry) now accepts an optional `revoke?: () => void | Promise<void>`, used as the entry's terminate instead of the hard-coded `end()`/`close()`/`emit('close')` on the subscription object, and returns an `{ unregister }` handle. When `revoke` is supplied [the subscription object is left completely untouched](https://github.com/HarperFast/harper/pull/2039/changes?w=1#diff-af3f596067a99a6721197f0532f9cfe257d8c2d42a64eb6493b472dd2e6fec74R137) — no `end` wrapping, no `'close'` listener — and [the caller unregisters through the returned handle](https://github.com/HarperFast/harper/pull/2039/changes?w=1#diff-af3f596067a99a6721197f0532f9cfe257d8c2d42a64eb6493b472dd2e6fec74R154) instead.

**Why:** profiling a production node found ~157,888 live `Subscription` objects across only 29 distinct topics, almost entirely per-connection bookkeeping (each carries its own `DatabaseTransaction`, `TableResource`, `Generator`). The fix is to share one feed per topic across its subscribers while keeping authorization per subscriber — but revocation today works by destroying the subscription object, which is exactly wrong once that object is shared: wiring `terminate` to a shared feed would disconnect every subscriber on one user's revocation or token expiry (and, since `sweep()` fails closed on recheck errors, one transient error would tear the feed down for everyone). This PR is **only the prerequisite seam** — no shared feeds, no refcounting, no subscriber-list machinery.

**Revocation granularity must stay per subscriber even once feed granularity becomes per topic** — that's the invariant this seam exists to establish.

## Scope reduced from the earlier revision

Earlier revisions of this branch also shipped a revocation state machine — `pendingTerminate`, a bounded terminate timeout, settle-handler late-commit, and fail-closed retry across sweeps — roughly 120 lines whose only exercise was its own tests, since nothing in this repo supplies `revoke`. That is now removed, per the owner's decision on review feedback ("reduce scope"): the timeout and idempotency requirements are unknowable without a real caller, and the shared-feed follow-up may want different ones.

What remains is the seam plus the one change the seam genuinely forces: [`terminate` may now return a promise](https://github.com/HarperFast/harper/pull/2039/changes?w=1#diff-af3f596067a99a6721197f0532f9cfe257d8c2d42a64eb6493b472dd2e6fec74R158), so [`terminateEntry` untracks the entry first and dispatches teardown once, fire-and-forget, containing a synchronous throw and a rejection alike](https://github.com/HarperFast/harper/pull/2039/changes?w=1#diff-af3f596067a99a6721197f0532f9cfe257d8c2d42a64eb6493b472dd2e6fec74R170). Untracking first is what keeps a hung `revoke` from wedging the sweep — [awaiting it would hold the `sweeping` mutex forever](https://github.com/HarperFast/harper/pull/2039/changes?w=1#diff-af3f596067a99a6721197f0532f9cfe257d8c2d42a64eb6493b472dd2e6fec74R178) — and it matches the pre-existing default-path behavior, which also deleted the entry before calling `terminate()`. The consequence, [stated in the contract](https://github.com/HarperFast/harper/pull/2039/changes?w=1#diff-af3f596067a99a6721197f0532f9cfe257d8c2d42a64eb6493b472dd2e6fec74R96): a `revoke` that throws, rejects or never settles is logged and never retried.

Three smaller changes came out of this branch's review rounds:
- Revocation logging is two-level. Each revocation keeps its own `info` line, and every sweep that revoked anything [closes with one aggregate `warn`](https://github.com/HarperFast/harper/pull/2039/changes?w=1#diff-af3f596067a99a6721197f0532f9cfe257d8c2d42a64eb6493b472dd2e6fec74R217) carrying the total and a per-reason breakdown. Harper ships `logging.level: warn` (`static/defaultConfig.yaml:46`), so an `info`-only line would make every expected revocation invisible on a default deployment; a `warn` per subscriber would turn one mass role change into a 10k-line storm. The aggregate is emitted from `sweep()`'s `finally`, and it says *revoking*, not *revoked*, because `terminate` is dispatched fire-and-forget — a teardown that fails still gets its own `error` line. `warn` also still covers the per-entry recheck-failure path it covered on `main`.
- The unit suite [pins `HARPER_SUBSCRIPTION_REAUTH_INTERVAL_MS` before importing the module](https://github.com/HarperFast/harper/pull/2039/changes?w=1#diff-abda1759bc4f13a990a4ac3d2c9b7435a05a9b0eb0e43aebe96abd52f5c49b8cR11). It previously ran against the real 30s interval, where a background tick makes `_sweepNow()` a no-op through the `sweeping` guard and assertions then run against a sweep that never happened.

When `revoke` is omitted, behavior is unchanged from #1414 in every respect except the new revocation logging; the sole existing caller (`registerLiveSubscriptionForContext` in `resources/Resource.ts`) doesn't pass `revoke`.

## For the human reviewer

The load-bearing judgment call is **fail-open teardown**: with the state machine gone, a `revoke` that never settles leaves a deauthorized subscriber attached with no further authorization evaluation. Reviewers raised this in every round. It is [documented on the API](https://github.com/HarperFast/harper/pull/2039/changes?w=1#diff-af3f596067a99a6721197f0532f9cfe257d8c2d42a64eb6493b472dd2e6fec74R96) and cannot manifest on this branch (no caller supplies `revoke`), and it is deliberately the shared-feed caller's problem to solve where the timeout/idempotency semantics are actually known — but it should be ratified before that wiring lands, and it is cheap to revisit now while the seam is unused.

Declined this round, all pre-existing on `main` and out of scope for a scope-reduction change — each is worth its own issue:
- **`sweep()` is N serialized `recheck()` round-trips behind one global mutex** (adjudicated major). The production `recheck` does a `findAndValidateUser` + `allowRead` per entry with no memoization across entries sharing a username, and there is no timeout on it. This seam does raise registry cardinality from per-stream-object to per-subscriber, which makes N bigger — but the serialization, the mutex, and the unbounded await are all unchanged from `main`.
- **ITC-triggered sweeps are dropped with no trailing pass**: `if (sweeping) return` discards a user-change broadcast arriving mid-sweep, so revocation falls back to the 30s backstop. A dirty flag re-run in `sweep()`'s `finally` closes it. (One reviewer described the exposure as "up to 24 hours" — that reads the *test* interval override; production is the 30s backstop.)
- **The recheck-failure path still warns once per entry**, as it does on `main` — each of those lines carries a distinct error message, so collapsing it into the aggregate would cost diagnostics rather than noise. The aggregate counts those entries too, under `recheck error`.
- **A leaked `'close'` listener and the never-restored `end` wrapper on the default path**: `unregister()` removes the registry entry but does not `removeListener('close', …)` or put back the original `end`. Pre-existing from #1414 and harmless for a per-connection subscription object that is discarded on close, but it would matter for a pooled or reused stream object.
- **Two-mode discriminated union vs two named entry points**, and **`_sweepNow`/`_liveSubscriptionCount` exported from the production module** — both flagged as API-shape decisions, both cheapest to change now.

**Refuted, not declined:** the last review round's remaining "major" claims `hdbLogger.info`/`warn`/`error` are unbound *methods*, so passing them as references (`safeLog(notice, …)`, `notice = hdbLogger.info`) throws a `this` `TypeError` that `safeLog` swallows, silently suppressing every revocation log. They are module-level free functions — `export function info(...args) { mainLogger.info(...args); }` at `utility/logging/harper_logger.ts:807`, `:825`, `:861` — with no `this` reference, so a bare reference is safe, and the suite's `logs each expected revocation at info and one aggregate warn per sweep` test asserts the lines are actually emitted. The `Human-Review-Need: 4` footer below still counts it: that round's adjudication leg was auto-pruned as a narrow low-risk delta, so the grade is "outside findings unadjudicated", not a live defect.

## Verification

- `npm run build`, `npm run lint:required`, `npm run format:write` — clean.
- `npx mocha "unitTests/server/**/*.test.js"` — 715 passing, including the 20 tests in this file. Two of them cover the aggregate: [one info line per subscriber plus exactly one warn with the count and reasons](https://github.com/HarperFast/harper/pull/2039/changes?w=1#diff-abda1759bc4f13a990a4ac3d2c9b7435a05a9b0eb0e43aebe96abd52f5c49b8cR69), and [no aggregate at all for a pass that revokes nothing](https://github.com/HarperFast/harper/pull/2039/changes?w=1#diff-abda1759bc4f13a990a4ac3d2c9b7435a05a9b0eb0e43aebe96abd52f5c49b8cR111).
- **End-to-end route: existing integration suite.** `npm run test:integration -- integrationTests/security/subscription-revocation.test.ts` — 6/6 passing, unmodified, covering SSE/WS/MQTT revocation on `drop_user`, `alter_role`, and token expiry. This is what proves the sole in-repo caller is unaffected; the `revoke` path itself has no end-to-end route until the shared-feed caller lands, and is covered only by unit tests ([rejection containment](https://github.com/HarperFast/harper/pull/2039/changes?w=1#diff-abda1759bc4f13a990a4ac3d2c9b7435a05a9b0eb0e43aebe96abd52f5c49b8cR361), [hang containment](https://github.com/HarperFast/harper/pull/2039/changes?w=1#diff-abda1759bc4f13a990a4ac3d2c9b7435a05a9b0eb0e43aebe96abd52f5c49b8cR389), [no subscription mutation](https://github.com/HarperFast/harper/pull/2039/changes?w=1#diff-abda1759bc4f13a990a4ac3d2c9b7435a05a9b0eb0e43aebe96abd52f5c49b8cR183)).
- `npm run test:unit:main` and `npm run test:unit:resources` both have failures on this machine that reproduce identically with `origin/main`'s copy of both changed files: `test:unit:main` aborts at import with `LZ4 not supported in this build` against a stale local database dir, and `test:unit:resources` fails 3 tests in `replayStructures` / `randomAccessFields`, neither of which imports this module.

Refs #1414

---
Generated with Claude Opus 5.

<sub>Review-Coverage: authored=claude; ran=codex,gemini; declined=cursor-grok,cursor-composer,domain; rounds=8 @ 096f399f9302</sub>

<sub>Human-Review-Need: 4 @ 096f399f9302</sub>

### DIFF
--- a/server/liveSubscriptionAuth.ts
+++ b/server/liveSubscriptionAuth.ts
@@ -129,13 +129,15 @@ function stopIfIdle(): void {
  * mutate state shared across them — `registerLiveSubscriptionForContext` in resources/Resource.ts
  * mutates `context.user` to the freshly-rechecked principal, which is safe only because each context
  * today belongs to exactly one subscriber; (3) `revoke` must be idempotent and safe to run concurrently
- * with the caller's own teardown — a *rejected* attempt is retried on the next sweep (see
- * `claimAndTerminate`), an attempt that never settles is not retried and the entry is not rechecked
- * again, and `unregister()` racing an in-flight `revoke` is only closed up to the point the await
- * begins, not for the remainder of it.
+ * with the caller's own teardown — a *rejected* attempt is re-invoked on the next sweep, but only if
+ * that sweep's `recheck` again denies, so a subscription whose permission was restored in between is
+ * left half-torn-down and treated as healthy; an attempt that never settles is not retried at all and
+ * the entry is never rechecked again; and `unregister()` racing an in-flight `revoke` is only closed
+ * up to the point the await begins, not for the remainder of it.
  */
 export function registerLiveSubscription(opts: {
-	subscription: any;
+	/** Optional when `revoke` is supplied: a revoke-only registrant may own no subscription object. */
+	subscription?: any;
 	username: string;
 	authExpiresAt?: number;
 	recheck: () => Promise<boolean>;
--- a/server/liveSubscriptionAuth.ts
+++ b/server/liveSubscriptionAuth.ts
@@ -52,6 +52,17 @@ function errorMessage(error: unknown): string {
 	}
 }
 
+// hdbLogger's file-backed transport can itself throw (e.g. ENOSPC/EIO on the underlying
+// fs.appendFileSync) with nothing between it and the caller. A logging call inside a fail-closed
+// catch arm must never be the thing that defeats fail-closed by escaping as a new exception.
+function safeLog(log: ((message: string) => void) | undefined, message: string): void {
+	try {
+		log?.(message);
+	} catch {
+		/* a broken logger must not turn a contained failure into a new one */
+	}
+}
+
 function withTimeout<T>(promise: Promise<T>, ms: number, message: string): Promise<T> {
 	return new Promise((resolve, reject) => {
 		const timer = setTimeout(() => reject(new Error(message)), ms);
@@ -81,7 +92,7 @@ const NOOP_HANDLE = { unregister: () => {} };
 // closed catch-arm call (e.g. a logger call that itself throws) would otherwise escape as an
 // unhandled rejection from these fire-and-forget triggers.
 function triggerSweep(): void {
-	void sweep().catch((error) => hdbLogger.error?.(`liveSubscriptionAuth: sweep failed: ${errorMessage(error)}`));
+	void sweep().catch((error) => safeLog(hdbLogger.error, `liveSubscriptionAuth: sweep failed: ${errorMessage(error)}`));
 }
 
 function ensureStarted(): void {
@@ -211,12 +222,18 @@ export function registerLiveSubscription(opts: {
  * that never settles would hold `sweeping` true forever (the `finally` in `sweep()` never runs),
  * silently disabling re-authorization for every subscription on the worker, not just this entry's.
  * `recheck` has the same unbounded-await shape but is registry-controlled Harper code; `revoke` is
- * arbitrary caller code, which is the materially wider exposure this bound closes. A timeout does not
- * cancel the underlying call — `entry.pendingTerminate` caches it, so a retry on the next sweep
- * re-awaits the SAME in-flight attempt instead of invoking `terminate`/`revoke` a second time while
- * the first is still running; the cache clears once that attempt actually settles (success or
- * failure), so a genuinely stuck call is never retried, but a merely slow one gets a fresh invocation
- * next time only after it's actually done.
+ * arbitrary caller code, which is the materially wider exposure this bound closes.
+ *
+ * A timeout does not cancel the underlying call — `entry.pendingTerminate` caches it, so a retry on
+ * the next sweep re-awaits the SAME in-flight attempt instead of invoking `terminate`/`revoke` a
+ * second time while the first is still running. Committing the removal (`registry.delete` + the
+ * success log) happens from a handler attached to that cached attempt directly, NOT from the code
+ * below that raced it with a timeout: a `revoke` that succeeds after this sweep's timeout window
+ * closes, but before the next sweep starts, must still be committed exactly once — reusing it as a
+ * plain resolved value in a later `claimAndTerminate` call would silently discard that success and
+ * invoke `revoke` again next sweep, precisely the double-invocation this cache exists to prevent.
+ * Attaching the commit/failure handler once, at the point the attempt is created, means it fires
+ * whenever the attempt settles, whether or not any sweep is actively awaiting it at that moment.
  */
 function invokeTerminate(entry: LiveSubscription): Promise<void> {
 	try {
@@ -233,22 +250,38 @@ async function claimAndTerminate(entry: LiveSubscription, reason: string): Promi
 		attempt = invokeTerminate(entry);
 		entry.pendingTerminate = attempt;
 		const settledAttempt = attempt;
-		const clearIfCurrent = () => {
-			if (entry.pendingTerminate === settledAttempt) entry.pendingTerminate = undefined;
-		};
-		attempt.then(clearIfCurrent, clearIfCurrent);
+		attempt.then(
+			() => {
+				if (entry.pendingTerminate === settledAttempt) entry.pendingTerminate = undefined;
+				if (registry.delete(entry)) {
+					safeLog(hdbLogger.info, `liveSubscriptionAuth: revoked subscription for ${entry.username} (${reason})`);
+				}
+			},
+			(error) => {
+				if (entry.pendingTerminate === settledAttempt) entry.pendingTerminate = undefined;
+				safeLog(
+					hdbLogger.error,
+					`liveSubscriptionAuth: terminate failed for ${entry.username} (${reason}), will retry next sweep: ${errorMessage(error)}`
+				);
+			}
+		);
 	}
 	const timeoutMs = terminateTimeoutMs();
 	try {
 		await withTimeout(attempt, timeoutMs, `terminate timed out after ${timeoutMs}ms`);
-	} catch (error) {
-		hdbLogger.error?.(
-			`liveSubscriptionAuth: terminate failed for ${entry.username} (${reason}), will retry next sweep: ${errorMessage(error)}`
-		);
+	} catch {
+		// If the attempt itself already rejected, the handler above already logged the definitive
+		// failure and cleared pendingTerminate — nothing more to do. If it's still pending (this is
+		// our own timeout firing, not a rejection), pendingTerminate still points at it; log that this
+		// sweep is still waiting so a persistently hung revoke doesn't go completely silent.
+		if (entry.pendingTerminate === attempt) {
+			safeLog(
+				hdbLogger.error,
+				`liveSubscriptionAuth: terminate for ${entry.username} (${reason}) still pending after ${timeoutMs}ms, will retry next sweep`
+			);
+		}
 		return false;
 	}
-	registry.delete(entry);
-	hdbLogger.info?.(`liveSubscriptionAuth: revoked subscription for ${entry.username} (${reason})`);
 	return true;
 }
 
--- a/server/liveSubscriptionAuth.ts
+++ b/server/liveSubscriptionAuth.ts
@@ -205,36 +205,6 @@ export function registerLiveSubscription(opts: {
 	return { unregister };
 }
 
-/**
- * Remove `entry` and run its terminate/revoke, but only commit the removal once terminate actually
- * settles successfully (or times out — see below) — and only if the caller hasn't already
- * unregistered this entry itself (checked via `registry.has`, not a delete-then-restore, so a failed
- * attempt never revisits the entry within the same sweep pass; it's simply left in place for the
- * next one). This makes revocation fail-closed on *tracking*: a `revoke` that throws, rejects, or
- * never settles (e.g. a shared-feed refcount release hitting a wedged backing store) leaves the entry
- * registered instead of being forgotten, so the next sweep retries — recheck fails the same way, so
- * the retry converges once teardown succeeds. The `info` log on success reports that terminate was
- * *invoked without error*, not that delivery is provably stopped — the default terminate (no `revoke`
- * supplied) still swallows its own errors internally, unchanged from #1414, so that guarantee is only
- * as strong as it always was on the legacy path.
- *
- * `terminate` is raced against `terminateTimeoutMs()`: without a bound, a caller-supplied `revoke`
- * that never settles would hold `sweeping` true forever (the `finally` in `sweep()` never runs),
- * silently disabling re-authorization for every subscription on the worker, not just this entry's.
- * `recheck` has the same unbounded-await shape but is registry-controlled Harper code; `revoke` is
- * arbitrary caller code, which is the materially wider exposure this bound closes.
- *
- * A timeout does not cancel the underlying call — `entry.pendingTerminate` caches it, so a retry on
- * the next sweep re-awaits the SAME in-flight attempt instead of invoking `terminate`/`revoke` a
- * second time while the first is still running. Committing the removal (`registry.delete` + the
- * success log) happens from a handler attached to that cached attempt directly, NOT from the code
- * below that raced it with a timeout: a `revoke` that succeeds after this sweep's timeout window
- * closes, but before the next sweep starts, must still be committed exactly once — reusing it as a
- * plain resolved value in a later `claimAndTerminate` call would silently discard that success and
- * invoke `revoke` again next sweep, precisely the double-invocation this cache exists to prevent.
- * Attaching the commit/failure handler once, at the point the attempt is created, means it fires
- * whenever the attempt settles, whether or not any sweep is actively awaiting it at that moment.
- */
 function invokeTerminate(entry: LiveSubscription): Promise<void> {
 	try {
 		return Promise.resolve(entry.terminate());
@@ -243,29 +213,54 @@ function invokeTerminate(entry: LiveSubscription): Promise<void> {
 	}
 }
 
+/**
+ * Remove `entry` and run its terminate/revoke, but only if the caller hasn't already unregistered
+ * this entry itself (checked via `registry.has`, not a delete-then-restore, so a failed attempt never
+ * revisits the entry within the same sweep pass; it's simply left in place for the next one). This
+ * makes revocation fail-closed on *tracking*: a `revoke` that throws, rejects, or never settles (e.g.
+ * a shared-feed refcount release hitting a wedged backing store) leaves the entry registered instead
+ * of being forgotten, so the next sweep retries — recheck fails the same way, so the retry converges
+ * once teardown succeeds.
+ *
+ * `terminate` is invoked at most once per entry no matter how many sweeps it takes to settle.
+ * `entry.pendingTerminate` caches the in-flight attempt, and the commit/failure handler is attached to
+ * it ONCE, at creation — it fires whenever the attempt settles (success, rejection, or, for the
+ * default terminate, a synchronous self-unregister via `subscription.end`'s wrapper), regardless of
+ * whether any sweep is actively awaiting it at that moment. This is what makes a late success (one
+ * that lands after this sweep's timeout window closes but before the next sweep starts) commit
+ * exactly once instead of being silently discarded and re-invoked. A sweep that finds an attempt
+ * already in flight from a previous pass does not re-invoke `terminate`/`revoke` and does not re-race
+ * it either — the outcome will be handled by the original handler whenever it settles; there is
+ * nothing left for a second wait to accomplish, and re-racing it would cost another full
+ * `terminateTimeoutMs()` of serialized sweep time per stuck entry, every single pass, forever.
+ *
+ * Only the FIRST attempt is raced against `terminateTimeoutMs()`: without a bound, a caller-supplied
+ * `revoke` that never settles would hold `sweeping` true forever (the `finally` in `sweep()` never
+ * runs), silently disabling re-authorization for every subscription on the worker, not just this
+ * entry's. `recheck` has the same unbounded-await shape but is registry-controlled Harper code;
+ * `revoke` is arbitrary caller code, which is the materially wider exposure this bound closes.
+ */
 async function claimAndTerminate(entry: LiveSubscription, reason: string): Promise<boolean> {
 	if (!registry.has(entry)) return false; // the caller already unregistered this entry itself
-	let attempt = entry.pendingTerminate;
-	if (!attempt) {
-		attempt = invokeTerminate(entry);
-		entry.pendingTerminate = attempt;
-		const settledAttempt = attempt;
-		attempt.then(
-			() => {
-				if (entry.pendingTerminate === settledAttempt) entry.pendingTerminate = undefined;
-				if (registry.delete(entry)) {
-					safeLog(hdbLogger.info, `liveSubscriptionAuth: revoked subscription for ${entry.username} (${reason})`);
-				}
-			},
-			(error) => {
-				if (entry.pendingTerminate === settledAttempt) entry.pendingTerminate = undefined;
-				safeLog(
-					hdbLogger.error,
-					`liveSubscriptionAuth: terminate failed for ${entry.username} (${reason}), will retry next sweep: ${errorMessage(error)}`
-				);
-			}
-		);
-	}
+	if (entry.pendingTerminate) return false; // already being handled by a prior sweep's attempt
+
+	const attempt = invokeTerminate(entry);
+	entry.pendingTerminate = attempt;
+	attempt.then(
+		() => {
+			if (entry.pendingTerminate === attempt) entry.pendingTerminate = undefined;
+			registry.delete(entry);
+			safeLog(hdbLogger.info, `liveSubscriptionAuth: revoked subscription for ${entry.username} (${reason})`);
+		},
+		(error) => {
+			if (entry.pendingTerminate === attempt) entry.pendingTerminate = undefined;
+			safeLog(
+				hdbLogger.error,
+				`liveSubscriptionAuth: terminate failed for ${entry.username} (${reason}), will retry next sweep: ${errorMessage(error)}`
+			);
+		}
+	);
+
 	const timeoutMs = terminateTimeoutMs();
 	try {
 		await withTimeout(attempt, timeoutMs, `terminate timed out after ${timeoutMs}ms`);
--- a/server/liveSubscriptionAuth.ts
+++ b/server/liveSubscriptionAuth.ts
@@ -85,26 +85,19 @@ function stopIfIdle(): void {
 /**
  * Register a live subscription for continuous re-authorization.
  *
- * If `revoke` is omitted, behavior matches #1414 exactly: terminate defaults to
- * end()/close()/emit('close') on `subscription`, and the subscription's own teardown (end()/close)
- * automatically unregisters it.
+ * Without `revoke`, teardown is end()/close()/emit('close') on `subscription`, and the
+ * subscription's own teardown unregisters the entry so callers need not.
  *
- * If `revoke` is supplied, it is used as the entry's terminate instead and the subscription object
- * is left untouched (no `end` wrapping, no 'close' listener) — the caller unregisters through the
- * returned handle. This is the seam a feed shared by many subscribers needs: revocation must stay
- * per subscriber even once feed delivery is shared, so the registry can't assume it owns the
- * subscription object or that only one registrant will ever mutate it.
- *
- * Three invariants a `revoke` caller must uphold, none enforced here: (1) it owns the entry's
- * lifetime end-to-end — nothing detects a leaked registration, so a caller that forgets
- * `unregister()` on some teardown path degrades sweep latency for every other tracked subscriber,
- * not just its own; (2) if it shares one feed (and one `recheck`) across subscribers, `recheck`
- * must not mutate state shared across them — `registerLiveSubscriptionForContext` in
- * resources/Resource.ts mutates `context.user` to the freshly-rechecked principal, which is safe
- * only because each context today belongs to exactly one subscriber; (3) `revoke` gets exactly one
- * best-effort invocation — the entry is untracked before it runs, so a `revoke` that throws, rejects
- * or never settles is logged and never retried, and a caller needing recovery from a failed teardown
- * must implement it inside `revoke`.
+ * With `revoke`, that becomes the teardown and `subscription` is never touched — no `end` wrapping,
+ * no 'close' listener — because a feed shared by many subscribers must stay revocable per subscriber,
+ * so the registry can neither own the shared object nor let every registrant mutate it. A `revoke`
+ * caller owns the entry's lifetime: nothing detects a leaked registration, and a forgotten
+ * `unregister()` degrades sweep latency for every other tracked subscriber. It also owns teardown
+ * recovery — the entry is untracked before `revoke` runs and `revoke` is invoked exactly once, so one
+ * that throws, rejects or never settles is logged and never retried. A `recheck` shared across
+ * subscribers must not mutate state shared across them: `registerLiveSubscriptionForContext` in
+ * resources/Resource.ts mutates `context.user`, which is safe only while each context has exactly
+ * one subscriber.
  */
 export function registerLiveSubscription(
 	opts: {
@@ -113,8 +106,8 @@ export function registerLiveSubscription(
 		recheck: () => Promise<boolean>;
 	} & (
 		| { subscription: any; revoke?: undefined }
-		// a revoke-only registrant may own no subscription object; requiring one of the two modes keeps
-		// a caller that supplies neither from type-checking into a silent no-op registration
+		// requiring one of the two modes stops a caller that supplies neither from type-checking
+		// into a silent no-op registration; a revoke-only registrant may own no subscription object
 		| { subscription?: any; revoke: () => void | Promise<void> }
 	)
 ): { unregister: () => void } {
@@ -161,14 +154,14 @@ export function registerLiveSubscription(
 	return { unregister };
 }
 
-/**
- * Untrack an entry and tear it down, best effort. Untracking first keeps a `revoke` that hangs or
- * fails from wedging the sweep or being re-entered by a later one; recovering from a failed teardown
- * is the caller's job, per registerLiveSubscription's third invariant.
- */
-function terminateEntry(entry: LiveSubscription, reason: string): void {
+/** Untrack first: a `terminate` that hangs or fails must not wedge the sweep or be re-entered by a later one. */
+function terminateEntry(
+	entry: LiveSubscription,
+	reason: string,
+	notice: ((message: string) => void) | undefined = hdbLogger.info
+): void {
 	registry.delete(entry);
-	safeLog(hdbLogger.warn, `liveSubscriptionAuth: revoking subscription for ${entry.username} (${reason})`);
+	safeLog(notice, `liveSubscriptionAuth: revoking subscription for ${entry.username} (${reason})`);
 	const failed = (error: unknown) =>
 		safeLog(
 			hdbLogger.error,
@@ -186,7 +179,7 @@ async function sweep(): Promise<void> {
 	if (sweeping) return; // a slow recheck must not overlap with the next tick/event
 	sweeping = true;
 	try {
-		// snapshot: an entry the caller unregisters mid-sweep must not be rechecked or revoked
+		// snapshot bounds the pass to entries present at its start; the has() guards cover the rest
 		for (const entry of Array.from(registry)) {
 			if (!registry.has(entry)) continue;
 			try {
@@ -196,7 +189,7 @@ async function sweep(): Promise<void> {
 				if (expired || !stillAuthorized) terminateEntry(entry, expired ? 'token expired' : 'no longer authorized');
 			} catch (error) {
 				// fail closed: if authorization can't be confirmed, revoke
-				if (registry.has(entry)) terminateEntry(entry, `recheck error: ${errorMessage(error)}`);
+				if (registry.has(entry)) terminateEntry(entry, `recheck error: ${errorMessage(error)}`, hdbLogger.warn);
 			}
 		}
 	} finally {
--- a/server/liveSubscriptionAuth.ts
+++ b/server/liveSubscriptionAuth.ts
@@ -17,6 +17,14 @@ import hdbLogger from '../utility/logging/harper_logger.ts';
 // Backstop interval; also catches token expiry, which is not event-signaled. Overridable for tests.
 const RECHECK_INTERVAL_MS = Number(process.env.HARPER_SUBSCRIPTION_REAUTH_INTERVAL_MS) || 30_000;
 
+// Bounds how long a caller-supplied terminate/revoke may hold the serialized sweep before being
+// treated as a failure (same fail-closed retry as a throw/rejection). The default terminate settles
+// long before this — it only ever engages for a caller-supplied `revoke` that hangs. Read fresh per
+// call rather than cached at module load, so tests can override it without a require-cache reset.
+function terminateTimeoutMs(): number {
+	return Number(process.env.HARPER_SUBSCRIPTION_TERMINATE_TIMEOUT_MS) || 5_000;
+}
+
 interface LiveSubscription {
 	username: string;
 	/** JWT `exp` (seconds since epoch) of the credential the subscription was opened with, if any. */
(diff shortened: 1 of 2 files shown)


server/liveSubscriptionAuth.ts
```diff
@@ -61,79 +124,204 @@ function stopIfIdle(): void {
...
+		return false;
+	}
+
+	const attempt = invokeTerminate(entry);
+	entry.pendingTerminate = attempt;
+	attempt.then(
+		() => {
+			if (entry.pendingTerminate === attempt) entry.pendingTerminate = undefined;
+			registry.delete(entry);
+			safeLog(hdbLogger.info, `liveSubscriptionAuth: terminate completed for ${entry.username} (${reason})`);
```
cb1kenobi: **Medium: revocation success now logs at `info`, which is suppressed at Harper's default log level**

Harper ships `logging.level: warn` (`static/defaultConfig.yaml`), and `HarperLogger.info` emits only when `this.level <= LOG_LEVEL_HIERARCHY.info` (3, vs warn's 4). Confirmed at runtime against the shipped predicate: at the default level `logsAtLevel('info')` is `false` while `logsAtLevel('warn')` is `true`.

So on a default deployment every *successful* revocation is silent — token expiry, permission loss, and the fail-closed `recheck error:` path alike. #1414 logged the recheck-error revocation at `warn`, so this is a behavior change on the default (no-`revoke`) path that the PR body states should stay identical to #1414. What's left visible is only the `error` logs for terminate failures and timeouts: a registry that is quietly tearing down live streams leaves no trace in the log an operator actually reads.

Suggested fix (the unit test at `unitTests/server/liveSubscriptionAuth.test.js:63` stubs `hdbLogger.info` and needs the same change):

```suggestion
			safeLog(hdbLogger.warn, `liveSubscriptionAuth: terminate completed for ${entry.username} (${reason})`);
```

—
_Generated by Barber AI_

unitTests/server/liveSubscriptionAuth.test.js
```diff
@@ -0,0 +1,508 @@
...
+			assert.strictEqual(revokeOkB.calls.length, 0);
+			assert.strictEqual(_liveSubscriptionCount(), 2);
+			assert.strictEqual(subscription.end.calls.length, 0);
+		});
+
+		it('revokes on token expiry without consulting recheck', async () => {
+			const subscription = fakeSubscription();
+			const revoke = spyFn();
+			const recheck = spyFn(() => Promise.reject(new Error('recheck must not be called for an expired token')));
+			register({ subscription, username: 'expired', authExpiresAt: 0, recheck, revoke });
```
cb1kenobi: **Medium: the token-expiry check has no effective test coverage — a fail-open regression passes all 18 tests**

Every expiry test in this file uses `authExpiresAt: 0` (14 occurrences). Zero is a fixed point under multiplication, so `now >= entry.authExpiresAt * 1000` is untested with respect to its seconds-to-milliseconds conversion — the only arithmetic in the module that sits on the credential-expiry boundary. `registerLiveSubscriptionForContext` (`resources/Resource.ts:954`) feeds it the JWT `exp` in seconds, so a scale error here is a live authorization bug.

Verified by mutation against a rebuilt `dist/` (the `#src/*` mapping means editing the `.ts` alone is a silent no-op): changing `* 1000` to `* 1` (every token instantly expired — fail-closed, all subscriptions killed) and to `* 1e12` (no token ever expires — an expired credential keeps streaming) each left the suite at **18 passing, 0 failing**.

There is also no test for the positive case: a subscription whose token is still valid must survive a sweep untouched.

Suggested fix — two tests with realistic values:

```js
const nowSec = Math.floor(Date.now() / 1000);
// must revoke
register({ subscription, username: 'expired-recently', authExpiresAt: nowSec - 5, recheck: async () => true, revoke });
// must survive the sweep
register({ subscription, username: 'still-valid', authExpiresAt: nowSec + 300, recheck: async () => true, revoke });
```

For context on the rest of the suite: I mutation-tested seven other invariants — revoke-vs-default terminate selection, subscription-object isolation, the already-unregistered guard, at-most-once terminate, the timeout bound, fail-closed tracking on a failed terminate, and fail-closed on a recheck throw — and every one was caught.

—
_Generated by Barber AI_

server/liveSubscriptionAuth.ts
```diff
@@ -61,79 +124,204 @@ function stopIfIdle(): void {
...
+ * `revoke` that never settles would hold `sweeping` true forever (the `finally` in `sweep()` never
+ * runs), silently disabling re-authorization for every subscription on the worker, not just this
+ * entry's. `recheck` has the same unbounded-await shape and is NOT similarly bounded here — not
+ * because it's safe (production `recheck` bottoms out in `resource.allowRead`, user-overridable
+ * application code, same as `revoke`), but because bounding it would change behavior on the DEFAULT
+ * (no-`revoke`) path, which this seam's own acceptance contract requires to stay identical to #1414.
+ */
+async function claimAndTerminate(entry: LiveSubscription, reason: string): Promise<boolean> {
+	if (!registry.has(entry)) return false; // the caller already unregistered this entry itself
+	if (entry.pendingTerminate) {
```
cb1kenobi: **Medium: a `revoke` that never settles is never retried and never rechecked — the subscriber keeps streaming**

Once `pendingTerminate` is set on an attempt that never settles, this branch is the only thing that ever happens to the entry again. `sweep()` skips `recheck()` for any entry with a pending terminate (line 307), and this short-circuit means `revoke` is never invoked a second time. The entry stays in the registry, nothing else decides its fate, and the revoked subscriber keeps receiving data for the life of the worker. The PR's own hang test pins exactly that: after a second sweep it asserts `hangingCalls === 1` and `_liveSubscriptionCount() === 1`.

This is a different case from the one declined in the PR description. "The registry currently retries forever with a per-sweep log" holds for a *rejecting* `revoke` — `pendingTerminate` is cleared in the rejection handler and the next sweep retries — but not for a *hung* one, where there is no retry at all, only a log. Worth correcting the description, and worth settling here rather than deferring, since the `claimAndTerminate` docblock is the stated source of truth for this contract.

Confidence: static reasoning plus the behavior the existing hang test already asserts. Dormant in this repo — no caller supplies `revoke` — so it is not reproduced against a real revoke path.

Suggested fix, and deliberately not the hard `end()`/`close()` fallback that was reasonably declined: bound the *entry*, not just the attempt. After N sweeps observing the same unsettled attempt, clear `entry.pendingTerminate` so the next sweep re-invokes `revoke`. The seam already requires `revoke` to be idempotent and safe to run concurrently with itself (invariant 3 in the `registerLiveSubscription` docblock), which is what makes a re-invoke legal here where a shared-object teardown would not be.

—
_Generated by Barber AI_

server/liveSubscriptionAuth.ts
```diff
@@ -61,79 +124,204 @@ function stopIfIdle(): void {
...
+		// silence would be worse than the per-sweep timeout log this replaced: a permanently stuck
+		// revoke would otherwise log exactly once, ever, then vanish. Log cheaply (no wait) instead.
+		safeLog(
+			hdbLogger.error,
+			`liveSubscriptionAuth: terminate for ${entry.username} (${reason}) still pending from an earlier sweep`
+		);
+		return false;
+	}
+
+	const attempt = invokeTerminate(entry);
```
cb1kenobi: **Low: the revocation state machine ships with no caller to validate it**

`pendingTerminate`, `withTimeout`, the settle-handler commit, and the "previously pending" sweep branch are roughly 120 of this PR's lines and, as the description says, all dormant — nothing in this repo passes `revoke`, so none of it executes outside its own unit tests. That is also why it took 8 review rounds and 8 fix commits: there is no real caller to check the invariants against, and the shared-feed PR may want different semantics anyway (its own retry budget, its own idempotency guarantees, refcount-aware teardown ordering).

The seam the follow-up actually needs is small: accept `revoke`, use it as `terminate`, skip the subscription-object wrapping, return `{ unregister }`. That is roughly 20 lines and is the part that unblocks the shared-feed work.

Suggested fix: consider landing just that seam here and moving the timeout / at-most-once / late-commit machinery into the PR that supplies a real `revoke`, where it can be exercised against actual behavior. Not a correctness objection — but a concurrency state machine whose only exercise is its own test file is a maintenance cost paid up front for requirements that are not yet known.

—
_Generated by Barber AI_

server/liveSubscriptionAuth.ts
```diff
@@ -61,79 +124,204 @@ function stopIfIdle(): void {
...
+	entry.pendingTerminate = attempt;
+	attempt.then(
+		() => {
+			if (entry.pendingTerminate === attempt) entry.pendingTerminate = undefined;
+			registry.delete(entry);
```
cb1kenobi: **Nit: a late-settling terminate can empty the registry without stopping the sweep timer**

`unregister()` and `sweep()`'s `finally` both pair `registry.delete` with `stopIfIdle()`; this handler does not. When the last entry's terminate settles after the sweep that started it has already returned — the exact case the docblock says this handler exists for — the interval keeps firing on an empty registry until the next tick's `finally` clears it. Self-correcting within one `RECHECK_INTERVAL_MS` and the timer is `unref`'d, so this is cosmetic.

```suggestion
			registry.delete(entry);
			stopIfIdle();
```

—
_Generated by Barber AI_

server/liveSubscriptionAuth.ts
```diff
@@ -61,79 +124,204 @@ function stopIfIdle(): void {
...
 async function sweep(): Promise<void> {
 	if (sweeping) return; // a slow recheck must not overlap with the next tick/event
 	sweeping = true;
 	try {
-		const now = Date.now();
 		for (const entry of registry) {
+			if (entry.pendingTerminate) {
+				// A prior sweep's attempt is already deciding this entry's fate; recheck() would only be
+				// a wasted storage read (findAndValidateUser + allowRead) for an outcome we already know.
+				await claimAndTerminate(entry, 'previously pending');
```
cb1kenobi: **Nit: `claimAndTerminate` is called here only to emit a log line**

When `entry.pendingTerminate` is set, `claimAndTerminate` does nothing but `safeLog(...)` and `return false` — it neither claims nor terminates, despite the name. Reading `sweep()` top to bottom you have to jump into a 40-line docblock and a second function to discover that this call is a no-op. Inlining the `safeLog` call here (and dropping the pending short-circuit from `claimAndTerminate`, whose only caller for that case is this branch) makes the branch self-describing and removes the misdirection.

—
_Generated by Barber AI_

server/liveSubscriptionAuth.ts
```diff
@@ -61,79 +124,204 @@ function stopIfIdle(): void {
...
+	}
+	return true;
 }
 
 async function sweep(): Promise<void> {
 	if (sweeping) return; // a slow recheck must not overlap with the next tick/event
 	sweeping = true;
 	try {
-		const now = Date.now();
 		for (const entry of registry) {
```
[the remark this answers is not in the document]
kriszyp: Addressed in b679649d0: sweep now iterates an Array.from(registry) snapshot and checks registry membership before processing each snapshotted entry, so entries unregistered while an earlier recheck is pending are skipped.

server/liveSubscriptionAuth.ts
```diff
@@ -17,24 +17,87 @@ import hdbLogger from '../utility/logging/harper_logger.ts';
 // Backstop interval; also catches token expiry, which is not event-signaled. Overridable for tests.
 const RECHECK_INTERVAL_MS = Number(process.env.HARPER_SUBSCRIPTION_REAUTH_INTERVAL_MS) || 30_000;
 
+// Bounds how long a caller-supplied terminate/revoke may hold the serialized sweep before being
+// treated as a failure (same fail-closed retry as a throw/rejection). The default terminate settles
+// long before this — it only ever engages for a caller-supplied `revoke` that hangs. Read fresh per
+// call rather than cached at module load, so tests can override it without a require-cache reset.
+function terminateTimeoutMs(): number {
+	return Number(process.env.HARPER_SUBSCRIPTION_TERMINATE_TIMEOUT_MS) || 5_000;
+}
```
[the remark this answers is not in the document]
kriszyp: Addressed in b679649d0: the configured terminate timeout is clamped to the 1..2,147,483,647 ms range, with direct coverage for an oversized environment value.

server/liveSubscriptionAuth.ts
kriszyp: Addressed in b679649d0: successful termination is logged through hdbLogger.warn, preserving visibility at Harper’s shipped default logging level; the default-path regression test now asserts the warning.

unitTests/server/liveSubscriptionAuth.test.js
kriszyp: Addressed in b679649d0: expiry coverage now uses realistic epoch-second values for both a recently expired token and a still-valid token, exercising the seconds-to-milliseconds conversion in both directions.

server/liveSubscriptionAuth.ts
kriszyp: Addressed in b679649d0: a late-settling successful terminate now calls stopIfIdle() immediately after removing its entry, so the final entry does not leave the sweep interval running until another tick.

server/liveSubscriptionAuth.ts
kriszyp: Addressed in b679649d0: the already-pending branch now emits its visibility log directly in sweep(); claimAndTerminate only handles calls that can actually claim and invoke termination.

server/liveSubscriptionAuth.ts
cb1kenobi: Verified fixed at `a10594b2`. Re-ran the check the same way: shipped default is `logging.level: warn`, and against the built predicate `logsAtLevel('info') === false` / `logsAtLevel('warn') === true`, so `hdbLogger.warn` at line 213 is visible on a default deployment where `info` was not. The regression test stubbing `hdbLogger.warn` pins it. Thanks for picking up the test change along with the call site.

—
_Generated by Barber AI_

unitTests/server/liveSubscriptionAuth.test.js
cb1kenobi: Verified fixed at `a10594b2` — this one is properly closed now. I re-ran both mutations against a rebuilt `dist/`, each isolated to exactly one site:

| mutation | direction | before (`d757e6bd`) | now (`a10594b2`) |
|---|---|---|---|
| `* 1000` → `* 1` | fail-closed: every token instantly expired | 18 passing, 0 failing | **20 passing, 1 failing** |
| `* 1000` → `* 1e12` | fail-open: no token ever expires | 18 passing, 0 failing | **20 passing, 1 failing** |

The fail-open direction is the one that mattered, and it is now caught by `revokes on token expiry without consulting recheck` via the realistic `nowSec - 5`. The new `keeps a subscription registered while its token remains valid` test catches the fail-closed direction. Both directions of the seconds-to-milliseconds conversion are now covered — nice fix.

I also re-ran the other seven mutants from last time (revoke-vs-default terminate selection, subscription-object isolation, the already-unregistered guard, at-most-once terminate, the timeout bound, fail-closed tracking on a failed terminate, fail-closed on a recheck throw) plus two against the new code, and all nine are killed. The `Array.from` snapshot guard and the timeout clamp both have real coverage.

—
_Generated by Barber AI_

server/liveSubscriptionAuth.ts
cb1kenobi: Re-reviewed at `a10594b2`. To be precise about what changed here, since this thread did not get a reply: **the behavior is unchanged** — `sweep()` still `continue`s past any entry with a `pendingTerminate` (line 245), so a hung `revoke` is still never re-invoked and never rechecked, and the entry stays registered for the life of the worker. The `bounds a never-settling revoke` test still pins `hangingCalls === 1` after the second sweep.

What "Clarify pending revocation behavior" did fix is the part that was actively misleading, and that is a real improvement: the timeout log no longer claims `will retry next sweep` (now `awaiting settlement`), and the `claimAndTerminate` docblock no longer promises a retry it does not perform.

**I am downgrading this from Medium to Low** on that basis. The path is dormant — `registerLiveSubscriptionForContext` (`resources/Resource.ts:950`) is still the only caller and passes no `revoke` — the blast radius is one subscriber rather than the sweep as a whole (the timeout bound holds; other entries in the same pass are still processed), and what is left is a deferred design decision that the shared-feed PR is genuinely better placed to make. Documenting it is a defensible call for this PR. Not blocking.

One residual worth a one-line fix, though, because it is the text the harper-pro implementor will actually read. The `registerLiveSubscription` docblock, invariant (3) at line 132, still says:

> a failed attempt is retried on the next sweep (see `claimAndTerminate`)

That holds for a *rejecting* `revoke` but not for a *hung* one — and a hung release is exactly the shape a shared-feed refcount decrement against a wedged backing store takes, so it is the case a `revoke` author most needs to know is not retried. Worth qualifying it as "a rejected attempt is retried on the next sweep; an attempt that never settles is not retried and the entry is not rechecked again".

—
_Generated by Barber AI_

server/liveSubscriptionAuth.ts
cb1kenobi: Verified fixed at `a10594b2` — `stopIfIdle()` now follows the `registry.delete` in the settle handler.

—
_Generated by Barber AI_

server/liveSubscriptionAuth.ts
cb1kenobi: Verified fixed at `a10594b2` — the visibility log is now emitted directly in `sweep()`, and `claimAndTerminate` only handles calls that can actually claim. `sweep()` reads straight through now. See the separate note on the docblock that came along with this change.

—
_Generated by Barber AI_

server/liveSubscriptionAuth.ts
```diff
@@ -61,79 +110,155 @@ function stopIfIdle(): void {
...
+function invokeTerminate(entry: LiveSubscription): Promise<void> {
+	try {
+		return Promise.resolve(entry.terminate());
+	} catch (error) {
+		return Promise.reject(error);
+	}
+}
+
+/**
+ * Terminate a registered entry and remove it only after success. A rejected attempt remains tracked
```
cb1kenobi: **Low: `claimAndTerminate` no longer enforces the at-most-once invariant this docblock promises**

This line says "a pending attempt is not invoked again", but the `if (entry.pendingTerminate) return false` guard that enforced it was removed from this function in `b679649d0` — correctly, as far as the readability nit that prompted it goes. At-most-once now rests entirely on `sweep()`'s `continue` at line 250. Both of today's callers (lines 257 and 261) sit after that `continue`, so there is no live bug.

The concern is that the guarantee moved out of the function that documents it and into its call site. `revoke` idempotency is invariant (3) of the public contract, and the seam exists specifically so harper-pro can add callers — an event-driven immediate revoke is the obvious next one, and it would call `claimAndTerminate` directly, double-invoking a `revoke` that is still in flight while this docblock says it cannot happen.

Verified by mutation: removing `sweep()`'s `continue` fails 3 tests, so the current path is well covered; nothing covers a direct call, because there is no direct caller yet.

Suggested fix: restore the one-line guard next to the existing `registry.has(entry)` check — `if (entry.pendingTerminate) return false;` — so the function is safe for any caller and the docblock is true of the function rather than of the loop around it. The log line stays in `sweep()`, so the nit that motivated the move is unaffected.

—
_Generated by Barber AI_

kriszyp:
### CI triage: all three failures were unrelated to this PR — checks are green now

Re-ran the failed jobs on the **identical commit** `a10594b2` with **no code change**; every check now passes (42 pass / 2 skipping).

| Failing check | Actual failure | Verdict |
|---|---|---|
| Unit Test (Node.js v22) | `Txn Expiration > Slow txn will expire` — `unitTests/resources/txn-tracking.test.js:61` | Flake |
| Integration 2/6 (Windows) | `describe-metadata-upgrade.test.ts` — probe `/SeoPageCache/` never ready 120s after `restart_service` | Known Windows CI issue |
| Integration 6/6 (Windows) | `configuration.test.mjs:319` — 500 from `set_configuration`, `EPERM: rename …harper-config.yaml.<pid>.<tid>.<hex>.tmp -> harper-config.yaml` | Known Windows CI issue |

Why none of them can be this diff:

- The diff is confined to `server/liveSubscriptionAuth.ts` + `unitTests/server/liveSubscriptionAuth.test.js`. Nothing in it touches transaction tracking, component install, HTTP-worker restart, or the config write path.
- The unit failure is in a **different mocha process** from the new test: CI runs `test:unit:all`, where `liveSubscriptionAuth.test.js` lands in `test:unit:main` and `txn-tracking.test.js` in `test:unit:resources`. No shared process, so no cross-test timing interference is possible. `main` itself failed Unit Test in 3 of its last 8 runs, each time on a *different* single test (`Caching > Can load cached data`, `HNSW greedy routing above layer 0`, MQTT QoS=1 reconnect) — same flake profile.
- Windows shard 2's failure is the same shape as [#2273](https://github.com/HarperFast/harper/issues/2273) (Windows component-deploy/restart hang after `npm pack`): `npm install` took 4 minutes on that runner, the restart's readiness probe then got `ECONNREFUSED` for its whole 120s window.
- Windows shard 6's `EPERM` rename reproduced **identically on an unrelated PR** in the same window — [`kris/642-branched-databases` run 32790110806](https://github.com/HarperFast/harper/actions/runs/32790110806), same test, same assertion. `fix/bounded-update-attributes-lock` was failing the same two Windows shards across four runs tonight. It is org-wide, not branch-specific.

Verified locally on this branch: `unitTests/server/liveSubscriptionAuth.test.js` 21/21 passing, the whole top-level `unitTests/server/` suite 240/240, and `txn-tracking.test.js` green on 3 consecutive runs.

No code change was needed, so nothing was pushed — the branch is still at `a10594b2`.

Two things worth a separate look, both pre-existing and outside this PR:

1. The Windows `EPERM` on `atomicWriteFile`'s rename (`config/configUtils.ts:110`) already has an 8-attempt backoff and still exhausts it — several threads write `harper-config.yaml` concurrently on `set_configuration`, and on Windows a rename over a file another thread holds open fails. I did not find an existing issue for this one, unlike [#2273](https://github.com/HarperFast/harper/issues/2273).
2. `npm run test:unit:server` is broken on `main` independently of this PR: its `unitTests/server/**/*.js` glob loads worker *fixtures* as test files and dies on `processGroupOwnerWorker.js:15` (`Cannot read properties of null (reading 'postMessage')`). CI never notices because `test:unit:all` doesn't include that script.

— Claude Opus 5

kriszyp:
### Follow-up: both requested one-liners applied, plus two things the pre-push review found in them

Pushed `a10594b2` → `88b06112d`, three commits. @kriszyp asked for the two changes [@cb1kenobi](https://github.com/cb1kenobi) requested; the independent pre-push review (Codex + Gemini + Cursor Grok + Harper domain) then found two real problems in my own first pass, so those are fixed too.

**Requested — [`1ce07e5c4`](https://github.com/HarperFast/harper/pull/2039/commits/1ce07e5c4)**

- Restored `if (entry.pendingTerminate) return false;` at the top of `claimAndTerminate`, so the at-most-once guarantee lives in the function whose docblock promises it rather than only in `sweep()`'s `continue`. Both of today's callers already sit after that `continue`, so behavior is unchanged; a future direct caller — the event-driven immediate revoke harper-pro will want — no longer double-invokes an in-flight `revoke`.
- Qualified `registerLiveSubscription` invariant (3): a hung `revoke` is not retried and the entry is not rechecked again.

**Found by the review, in my own fixes — [`ccfcb2117`](https://github.com/HarperFast/harper/pull/2039/commits/ccfcb2117), [`88b06112d`](https://github.com/HarperFast/harper/pull/2039/commits/88b06112d)**

- The options type still *required* `subscription`, so a revoke-only caller owning no subscription object — the whole point of the seam — had to pass a meaningless `subscription: undefined` to type-check. My first attempt made it plainly optional, which went too far: a caller supplying **neither** `subscription` nor `revoke` then type-checked, received `NOOP_HANDLE`, and silently left its subscription untracked. It is now a union requiring exactly one teardown mode. Verified the three valid shapes compile and the neither-mode call is a `TS2345` error, with repo `typecheck` clean (the existing `resources/Resource.ts:950` caller still matches).
- The rejected-terminate log still said *"will retry next sweep"* unconditionally — the same overstatement just corrected in the docblock, since a rejected attempt is only re-invoked if the next sweep's `recheck` **again** denies. Now "on the next sweep that still denies", and the invariant says so too: a subscription whose permission was restored in between is left half-torn-down and read as healthy.

Verification: `npm run build` and `npm run typecheck` clean; `unitTests/server/` 240/240 (liveSubscriptionAuth 21/21); Prettier clean. No test changed — none of this alters behavior for any caller in the tree.

### For the human reviewer — deliberately not addressed

The review re-raised these across three rounds and I left every one of them alone: they are design decisions on your PR, not defects I should be resolving unsupervised. Round 3 confirmed the delta "introduces no new correctness issue"; these are all carried forward from the original diff.

| | Decision |
|---|---|
| **major** | A never-settling `revoke` turns fail-closed into fail-open: the entry is skipped by every later sweep, `recheck` never runs again, and nothing proves delivery stopped. The docblock now states this as intended semantics — that is the part worth ratifying explicitly. This is the same ground as [@cb1kenobi](https://github.com/cb1kenobi)'s still-open thread, which suggests bounding the *entry* (clear `pendingTerminate` after N sweeps) rather than the attempt. |
| **major** | Serial per-entry timeouts extend the exclusive sweep window, and `sweep()`'s `if (sweeping) return` drops an ITC trigger firing inside it with no trailing sweep. 20 hung revokes ⇒ ~100 s window; a `drop_user` landing in it waits for the next 30 s tick. The coalescing gap pre-exists, but this makes it deterministic and multiplies it by entry count. Root fix is a dirty flag plus a trailing sweep, not more concurrency. |
| minor | `recheck` is still unbounded while `terminate` is now timeout-bounded — and `recheck` is the one doing real I/O (`findAndValidateUser` + `allowRead`). `withTimeout` already exists here. |
| minor | Mass revocation on the default path now emits one `warn` per subscriber (10k on a role drop) and allocates per entry; a permanently stuck entry logs at `error` every 30 s forever with no dedup. |
| minor | `unitTests/server/liveSubscriptionAuth.test.js` — the "must not re-race the timeout" assertion races `_sweepNow()` against a 250 ms marker, so its own "no matter how loaded the test runner is" claim is not true. `AGENTS.md` names this class as the root cause of #1138. Counting `setTimeout` invocations, as the clamp test already does, would assert the invariant directly. |
| scope | [@cb1kenobi](https://github.com/cb1kenobi)'s open thread: land only the ~20-line seam and move the timeout / at-most-once / late-commit machinery into the PR that supplies a real `revoke`. Your call — I did not act on it. |

Two review findings were adjudicated away rather than left open, worth recording so they don't come back: Gemini's `safeLog` unbound-`this` concern does not apply (`utility/logging/harper_logger.ts:353-385` exports standalone module functions, not class methods — `this` is never read), and the `Array.from(registry)` snapshot is immaterial next to one awaited DB-backed `recheck` per entry in the same loop, besides buying the mid-sweep-removal semantics the tests assert.

Independent pre-push review: 3 rounds, `reviewers=codex+gemini+cursor-grok+harper-domain` (round 1 full), `codex+gemini` (rounds 2–3 delta). 4 findings fixed, 6 open by design.

— Claude Opus 5

server/liveSubscriptionAuth.ts
kriszyp: Done — taken exactly as suggested, with the owner concurring on the scope reduction.

`pendingTerminate`, `withTimeout`, `claimAndTerminate`'s settle-handler commit, the timeout env var, and the "previously pending" sweep branch are gone (`9ee14b5`). What is left is the seam you described: accept `revoke`, use it as `terminate`, skip the subscription-object wiring, return `{ unregister }`.

One thing the seam still forces, kept deliberately: `terminate` can now return a promise, so `terminateEntry` untracks the entry first and dispatches teardown fire-and-forget, containing a synchronous throw and a rejection alike. Awaiting it would hold the `sweeping` mutex on a hung `revoke`; untracking first also matches the pre-existing default path, which already deleted the entry before calling `terminate()`. The consequence — a `revoke` that throws, rejects or never settles is logged and never retried — is now stated in the API contract rather than implemented as machinery, and the retry/timeout/idempotency policy goes with the PR that supplies a real `revoke`.

Net: -160 lines of implementation and tests vs the previous revision. Two smaller review items landed alongside: routine revocations moved from `warn` to `info` (a per-subscriber warn turns one mass role change into a storm; `warn` stays for the recheck-failure path `main` covered), and the unit suite now pins `HARPER_SUBSCRIPTION_REAUTH_INTERVAL_MS` before importing, since it had been racing the module's real 30s timer.

— Claude Opus 5

kriszyp:
### Scope reduced to the seam (option A), per the owner's decision

`9ee14b5`…`ef48484` remove the revocation state machine — `pendingTerminate`, the bounded terminate timeout and its env var, `claimAndTerminate`'s settle-handler commit, and the "previously pending" sweep branch. Net **−167 lines** in `server/liveSubscriptionAuth.ts` and its test since `a10594b2`. What is left is exactly the seam @cb1kenobi described: accept `revoke`, use it as `terminate`, skip the subscription-object wiring, return `{ unregister }`.

Two things came out of the pre-push review of those commits and are also in:

- Routine revocations (expiry, deny) log at `info`; `warn` stays on the recheck-failure path it covered on `main`. A per-subscriber `warn` would turn one mass role change into a warn storm.
- The unit suite pins `HARPER_SUBSCRIPTION_REAUTH_INTERVAL_MS` before importing the module, and restores the patched logger from a `finally`. It previously ran against the real 30s interval, where a background tick makes `_sweepNow()` a no-op through the `sweeping` guard and the assertions then run against a sweep that never happened.

**One outside finding is refuted rather than declined.** The last round flagged as major that `hdbLogger.info`/`warn`/`error` are unbound *methods*, so passing them by reference (`safeLog(notice, …)`) would throw a `this` `TypeError` that `safeLog` swallows, silently suppressing every revocation log. They are module-level free functions — `export function info(...args) { mainLogger.info(...args); }` at `utility/logging/harper_logger.ts:807`, `:825`, `:861` — with no `this` reference, so the bare reference is safe; the `logs an expected revocation at info, not warn` test asserts the line is actually emitted. The `Human-Review-Need: 4` footer still counts that finding because the round's adjudication leg was auto-pruned on a narrow delta, not because anything is unresolved in the code. Everything genuinely declined is listed under **For the human reviewer** in the description, with why.

Verification on this head: all CI checks green; locally `npx mocha "unitTests/server/**/*.test.js"` → 714 passing, 19 of them in `liveSubscriptionAuth.test.js`.

—
_Claude Opus 5_

server/liveSubscriptionAuth.ts
```diff
@@ -61,79 +83,113 @@ function stopIfIdle(): void {
...
 
 	ensureStarted();
+	return { unregister };
+}
+
+/** Untrack first: a `terminate` that hangs or fails must not wedge the sweep or be re-entered by a later one. */
+function terminateEntry(
+	entry: LiveSubscription,
+	reason: string,
+	notice: ((message: string) => void) | undefined = hdbLogger.info
```
cb1kenobi, line 161: **Medium: routine revocation success now logs at `info` again, which is suppressed at Harper's default log level**

This reintroduces the same operator-visibility gap flagged as Medium and closed at `a10594b2` (`terminate completed for ...` was moved from `info` to `warn` back then). Harper ships `logging.level: warn` (`static/defaultConfig.yaml:46`), and `LOG_LEVEL_HIERARCHY` puts `warn` at 4 and `info` at 3, so `logsAtLevel('info')` is `false` and `logsAtLevel('warn')` is `true` at the shipped default (verified at runtime against the built predicate).

`terminateEntry`'s `notice` parameter now defaults to `hdbLogger.info`, and both routine-revocation call sites in `sweep()` (`'token expired'`, `'no longer authorized'`) rely on that default. Only the recheck-error/fail-closed path explicitly passes `hdbLogger.warn`. So on a default deployment, every *expected* revocation — a JWT expiring, a role change taking away access — is silent; only the fail-closed error path remains visible. A registry that is quietly tearing down live streams for expiry/deny reasons leaves no trace in the log an operator running default settings actually reads.

The stated motivation (avoiding a per-subscriber `warn` storm on a mass role change) is legitimate, but the fix trades away all default-level visibility rather than the noise specifically — e.g. a single aggregated `warn` per sweep pass (count + reasons) would keep operators informed without one line per subscriber.

Suggested fix: keep routine revocations at `warn` (as before `a10594b2`'s original fix), or replace the per-subscriber log with a rate-limited/aggregated summary log at `warn` if volume is the actual concern.

—
_Generated by Barber AI_

kriszyp: Confirmed and fixed in `096f399f9` (with the wording correction in the same push). You're right on both the mechanism and the numbers: `static/defaultConfig.yaml:46` ships `logging.level: warn`, `LOG_LEVEL_HIERARCHY` puts `warn` at 4 and `info` at 3, so `hdbLogger.info` — a free function that forwards to `mainLogger.info` — is dropped on a default deployment, and only the recheck-error path stayed visible.

Took your second suggestion, the aggregate, rather than reverting to a per-subscriber `warn`, since that trades the noise specifically instead of trading away the storm fix:

- Each revocation keeps its own `info` line (`username` + reason) — the detail an operator gets by raising the level.
- Every sweep that revoked anything closes with **one** `warn`: `liveSubscriptionAuth: revoking N live subscriptions (token expired: 3, no longer authorized: 12)`. One line per pass, so a 10k-subscriber role change is one `warn`, not 10k, and it is bounded to at most one per sweep interval.

Two details worth flagging:

- It is emitted from `sweep()`'s `finally`, so an unexpected throw mid-pass cannot swallow the summary of what was already torn down.
- It says *revoking*, not *revoked*: `terminate` is dispatched fire-and-forget (the entry is untracked first so a hung `revoke` can't wedge the sweep), so past tense would claim more than the sweep knows. A teardown that throws or rejects still gets its own `error` line.

The per-entry recheck-failure `warn` is unchanged from `main` — each of those lines carries a distinct error message, so folding it into the aggregate would cost diagnostics rather than noise. Those entries are counted in the aggregate too, under `recheck error`.

Tests: `logs each expected revocation at info and one aggregate warn per sweep` asserts three info lines and exactly one warn carrying the count and reasons; `does not log an aggregate for a sweep that revokes nothing` covers the quiet path. `unitTests/server/` 715 passing, and `integrationTests/security/subscription-revocation.test.ts` 6/6.

—
_Claude Opus 5_

cb1kenobi: Verified fixed at `096f399f9`. Confirmed empirically against the shipped predicate (not by reading the call site): `logsAtLevel('info') === false`, `logsAtLevel('warn') === true` at the shipped default (`logging.level: warn`) — same check as before.

`terminateEntry`'s per-subscriber line is still `info` (unchanged), but `sweep()` now closes its `finally` with one aggregate `hdbLogger.warn` per pass whenever anything was revoked — count plus a reason breakdown (e.g. `revoking 3 live subscriptions (token expired: 3)`) — and stays silent when nothing was revoked. Ran it live via the unit suite: `[warn]` lines appear exactly once per sweep with the right counts, `[info]` lines carry the per-subscriber detail. This is the aggregated-summary alternative this comment suggested, not a revert to per-subscriber `warn`, and it resolves the visibility gap without reintroducing the per-subscriber storm on a mass role change.

—
_Generated by Barber AI_

PR #2044, opened 2026-08-01
Honor storage.compression on upgraded databases, and let a deployment select the codec

kriszyp:
`rocksdb-js` 2.6.1 (#2036) brought the codecs in and made lz4 the effective default. This fixes the case that default silently misses, and adds a way to choose a different codec.

## The bug: upgraded databases ignore `storage.compression`

`storage.compression` defaults to **`true`** (`config/defaultConfig.yaml`), so essentially every existing table's metadata records "compression enabled". `toRocksCompression` mapped that to *unset*, on the documented assumption that unset means "use the build default (lz4)".

It doesn't — not for a column family that already exists. RocksDB persists the codec per family, and a reopen that requests nothing **inherits what the family already has**; the build default applies only when the family is being created. Measured across separate processes:

```
5.1 instance, compression 'none'          -> 34.2M SST
5.2 instance, no option passed
    open -> engine reports {"algorithm":"none"}    <-- inherited, not lz4
                                          -> 68.4M SST   (new writes uncompressed too)
full compact()                            -> 68.4M SST   (unchanged)

control: same records, lz4 from scratch   ->  7.1M SST
```

So a brand-new 5.2 database gets lz4 while every database upgraded from 5.1 keeps writing uncompressed indefinitely — both having asked for exactly the same thing. New databases only worked by accident, because unset on a *non-existent* family does fall through to the build default.

An enabled setting now resolves to an explicit codec. On the real metadata path (`{ startingOffset, threshold }`, what `getDefaultCompression()` returns when `storage.compression` is true):

```
old:      opened 'none' -> reports none -> 31.6M SST
upgraded: opened 'lz4'  -> reports lz4  -> 33.9M SST   (the next 60k records cost 2.3M, not 31.6M)
```

**Existing data is deliberately left alone.** An explicit codec governs newly written files only, and `db.compact()` won't convert the rest: rocksdb-js calls `CompactRange` with default `CompactRangeOptions`, whose `bottommost_level_compaction` is `kIfHaveCompactionFilter`, and Harper installs no compaction filter — the bottommost level, holding the bulk of the data, is skipped. Converting in place needs a forced bottommost compaction, which is expensive enough to deserve an explicit operator decision rather than happening during an upgrade.

## Choosing a codec

`HARPER_STORAGE_ROCKS_COMPRESSION` selects the codec for every column family the process opens, validated against the build's `supportedCompression` so an unavailable name fails with a named error instead of a partial startup failure at the first open.

It's an env var and not a config key, and resolved once at module load, because RocksDB rejects reopening a family under a disagreeing codec and Harper's worker threads share one process-wide family registry — so every thread must agree. A config value doesn't achieve that; the main thread loads the storage layer during install, before the value is readable, while workers load it after:

```
pid=3591555 thread=0 resolved=undefined   <-- main thread, loads during install
pid=3591555 thread=1 resolved=zstd
pid=3591555 thread=2 resolved=zstd
```

`__dbis__` is then opened at the build default and reopened as the configured codec, and Harper dies with `The system database failed to load`. Reproduced on a fresh install with the value in config only; it fails on every boot. Making it a config key needs the value resolvable before the first family opens — a boot-ordering change, deliberately not in this PR, and the constraint is recorded in `DESIGN.md` so it doesn't get "improved" into one.

`copyDb`'s `openRocksDb` now resolves the codec the same way. `DESIGN.md` already called `openRocksDatabase` the single chokepoint and migration sat outside it — it creates the families the runtime later reopens.

## Benchmark

`benchmarks/compression/` compares codecs on identical data and workload, one instance per codec. On 1M document-shaped records:

| codec | data files | vs none |
| --- | --- | --- |
| none | 846 MB | 100% |
| snappy | 434 MB | 51% |
| lz4 | 454 MB | 54% |
| zstd | 351 MB | 41% |

Point reads get *faster* with compression — compressed blocks hold more records per block-cache slot and that outweighs decompression (lz4 nearly doubles point-read throughput at the storage layer). Sequential scans pay decompression with no cache-density offset and get slower.

Two traps the harness guards, both of which produce confidently wrong numbers: `@harperfast/integration-testing` installs under `os.tmpdir()`, which is tmpfs on most Linux hosts (every "on-disk" byte becomes a memory measurement); and Harper answers 503 once commits back up past `storage.maxTransactionQueueTime` — one run shed ~99% of its writes and still produced a throughput figure and a 67 MB "compressed" size, which reads as a spectacular win. Runs over an error-rate threshold are marked `INVALID`.

## Open concerns for the reviewer

- **The `toRocksCompression` change alters behavior for every existing table**, which is the point, but it's the line to scrutinize. `false`/`''` still map to `'none'`; an explicit `{ algorithm }` still passes through untouched; only "enabled, unspecified" changes meaning.
- **Benchmark sizes are as the workload left them, not a compacted steady state** — for the reason above, nothing outside the process can force a compaction. Codec-to-codec comparison is sound (identical workload each); the absolute ratio against logical bytes is inflated by update garbage. A storage-layer run using a real `db.compact()` agrees within 3 points.
- **Throughput numbers are single-sample, fixed codec order.** REST-layer variance swamps the codec differences; the size columns are the reliable output. Noted in the README.
- **Reads never touched a device** — the dataset fits in page cache. This measures decompression CPU and block-cache density, not I/O savings; an I/O-bound instance should look better.
- `openRocksDb` is a behavior change on the migration path — right side of the stated invariant, but worth a second pair of eyes.

Follow-up worth doing: expose forced bottommost compaction in rocksdb-js so an existing database can be converted on an explicit operator decision.

Cross-model review (Codex + Gemini + Grok + Harper-domain) ran pre-push; its findings on a no-op compaction step, a `scanBytesRead` race, the `copyDb` bypass, missing validation, committed result JSONs and missing docs are all addressed.

Generated by Claude Opus 5.


kriszyp:
All five review points addressed in `3246ec162`. Replying here rather than inline because the pending review blocks thread replies.

**1. Config surface instead of an environment-only control** — agreed, and the sequencing was fixable. `storage.rocks.compression` is now a registered config param; the bespoke `process.env` read is gone.

The ordering problem was `install()` calling `mountHdb()` (which creates the system column families) several steps *before* `createConfigFile()` writes the config file. `installer.ts` now stages the value into the in-memory config before `mountHdb()`, mirroring the `STORAGE_ENGINE` line directly above it, and resolution is frozen at the first open so every thread agrees. Verified on the case that previously failed on every boot:

```
--- boot 1 (fresh install, config only, NO env var) ---
boot1 route ready: true
boot1 write status: 204
--- boot 2 (restart, config file now exists) ---
boot2 route ready: true
boot2 write status: 204
=== any CF conflict? ===   (none)
```

Mounted YAML and `HARPER_CONFIG`/`HARPER_SET_CONFIG` now select the codec. The benchmark switched to config with them — worth noting it would otherwise have silently compared four identical configurations.

**2. `killHarper()` 5s grace truncating the final flush** — real, and it would bias exactly the wrong way. The runner now takes `--shutdown-grace` (default 120s), watches the child's `exit`, and **fails the run** if it was `SIGKILL`ed, rather than measuring a partial flush. The CPU sample still precedes shutdown by necessity (the `/proc` entries vanish with the process), so it's labelled as workload CPU rather than including the final flush.

**3. Zero tolerance for failed writes** — agreed; size is the primary result. Mutating requests (warmup, insert, update) are now held to **zero** failures; the tolerance survives only for reads and scans, which can't change what's stored. Warmup errors are retained rather than discarded, since warmup rows land in the same table and count toward `logicalBytes`.

**4. Tests that reach a real open** — `unitTests/resources/rocksCompressionOpen.test.js` stages a codec and opens real column families through the production resolution, asserting what the engine actually applied. It uses `zstd` rather than `none` so the assertions discriminate: with the forwarding removed these families open at the build default and the tests fail. Covers a table family and `__dbis__` agreeing, the configured codec beating `compression: false` metadata, a conflict-free reopen, and the unsupported-codec rejection.

You were also right that the previous block was worthless — it destructured an export that no longer existed and passed vacuously. Removed.

I did **not** get multi-thread coverage into a unit test; the fresh-install-with-workers case is verified by the reproduction above but isn't automated. Happy to add it as an integration test if you want it gated in CI.

**5. `copyDb` rationale inaccurate** — correct, and reframed. `copyDbToRocks()` closes every target handle before the staging rename and rocksdb-js allows a codec change across close/reopen, so the runtime would open the migrated database fine either way. The actual reason to thread the codec through migration is that it writes the entire dataset in one pass, and those files then keep their codec until write traffic rewrites them — so it's the one moment the deployment's codec should apply.

Also fixed the `scanBytesRead` race both bots flagged. My earlier claim to have fixed it was wrong: the edit silently no-op'd against reformatted source and I didn't verify that specific replacement.

Not marking Ready.

kriszyp:
Reshaped around **rocksdb-js [#743](https://github.com/HarperFast/rocksdb-js/pull/743)** (`compressionForAllColumnFamilies`) instead of the boot-time fallback. `d88ccd4aa`.

**⚠️ Blocked on #743 being released** — this branch needs a `@harperfast/rocksdb-js` that has the option. Verified locally against a linked build of that branch; the dependency floor still has to be bumped once it ships.

## What changed

The fallback added in `20682c00f` caught the `already open with compression` conflict and kept the family's existing codec. That kept boot working, but it meant the upgrade fix could never actually take effect: every pre-existing table stayed uncompressed, which was the entire point of the commit before it.

The root cause is that rocksdb-js gives every family the caller did not name its **persisted** algorithm and applies the request only to the target — so a family Harper never names individually keeps its original codec forever. #743 adds the opt-in that applies the codec to all of them.

Opens now pass that flag with a single codec resolved from configuration:
- `storage.rocks.compression` if it names one
- otherwise `storage.compression` (default `true`) decides enabled-or-not and the build default supplies the algorithm

Every family adopts it, nothing conflicts, and `openExistingRocksColumnFamily` and its warning are deleted.

## Deliberate trade-off

**Per-table compression metadata no longer selects a codec.** A table persisted as `compression: false` inside a deployment that enables compression would need its own codec, and there is nowhere to apply it — Harper cannot know a table's codec before the open that already opened its catalog. Compression is a deployment setting now.

In practice per-table values came from the global `storage.compression` at table-creation time, so a mixed database is unusual; but it is a real behavior change and it is why the fallback is unnecessary rather than merely bypassed. If we ever want genuine per-table codecs, that needs [#742](https://github.com/HarperFast/rocksdb-js/pull/742) (`setCompression()`), which is parked in draft for exactly that case.

## Verified

20k records written with `storage.compression: false` (9.6M SST), then the same data directory rebooted with it on:

```
phase 1 (compression off): 9.6M SST
phase 2 (compression on):  1.6M SST total
```

Boots clean, no conflict, no fallback — and compaction re-encoded the existing data. `test:unit:resources` 1357 passing.

Generated by Claude Opus 5.

PR #2051, opened 2026-08-02
fix(audit): track deleteHistory removals with bounded concurrency

kriszyp:
`Table.deleteHistory()` now tracks every audit/tombstone removal with bounded, engine-aware concurrency: 1,000 writes for LMDB and ten for RocksDB. This prevents detached rejections and unbounded pending work while preserving the large LMDB batches required by the Windows purge path. The limiter waits only at real saturation, resumes on any completion, and drains both phases even when iteration throws.

`removeAuditEntry()` also joins the registered tombstone cleanup. Cleanup snapshots each key/version before yielding and removes only that version. LMDB enforces the condition natively; RocksDB re-reads and removes inside one native transaction, retrying a conflict once. Versionless tombstones are left in place rather than risking an unconditional delete. A regression test forces a recreate between the RocksDB read and commit and proves the stale cleanup preserves the live record.

A purge that attempted at least one removal and completed none now [rejects with the first storage error](https://github.com/HarperFast/harper/pull/2051/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R5134), after both phases have drained. Previously every failure was warn-logged only, so `delete_transaction_logs_before` returned a successful `entries_deleted: 0` whether nothing was eligible or the store rejected every write — indistinguishable to an operator pruning to bound disk growth. Draining before deciding keeps all the progress a partially-failing store can still make, a single success reports normally, and partial failures stay best-effort (logged, excluded from the count). The response schema is unchanged.

The CI follow-up makes the heavy subscription-replay race test wait for every expected ID after concurrent writes finish. Its previous 200 ms quiet-window collector could close the subscription before the final delivery, which failed this PR's first Node 22 run with `missing concurrent id 10049`.

## For the human reviewer

1. The zero-progress rejection is the narrow option, not `entries_failed`. Adding a result field would expand this into a public API/docs/test contract change across LMDB, RocksDB and the deprecated `deleteAuditLogsBefore` wrapper; partial failures therefore still require log inspection to reconcile. Add `entries_failed` later only if callers need it programmatically. The new failure case is a job failure, not an unhandled rejection: `ResourceBridge.deleteTransactionLogsBefore` awaits `deleteHistory`, and `server/jobs/jobRunner.ts` records the rejection as job status `ERROR`. There is no end-to-end test of that path — inducing "every removal fails" needs fault injection at the store, so the route above is a code trace plus the unit regression.
2. A failed automatic tombstone removal still retires its audit entry. Keeping the audit entry would preserve an automatic retry trigger, but it could also hold audit cleanup behind a persistently failing primary-store write; the current behavior requires a later explicit `cleanup_deleted_records: true` sweep.
3. The LMDB cap intentionally trades some unbounded batching for bounded pending work. Ten concurrent writes timed out the 12,000-entry Windows purge at 60.48 seconds. Raising the bounded LMDB window to 1,000 reduced the same local 4,000-entry diagnostic from 1,678 ms to 56.6 ms; fresh Windows CI purged all 12,000 entries in 2.149 seconds.
4. Declined from the latest review round: Gemini read `db.remove = (id, opts) => typeof opts === 'number' ? removeIfVersion(...) : removeSync(...)` in [`databases.ts`](https://github.com/HarperFast/harper/pull/2051/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R234) as breaking a synchronous-boolean contract. lmdb-js's `remove(key, ifVersion)` already returns a promise, so the two engines agree, and the only numeric-version callers are `Table.ts:5124` and `Table.ts:6287`, both of which return or await the promise. Left as is.

## Verification

- `npm run build`, `npm run typecheck`, `npm run lint:required`, `npm run format:check` — passed.
- `npx mocha unitTests/resources/auditLog.test.js` — 26 passing, 6 pending. [`deleteHistory cleanup preserves a record recreated while its stale tombstone waits for a slot`](https://github.com/HarperFast/harper/pull/2051/changes?w=1#diff-044077436ab82ecf6c481205bdf0ac84571b38edf4ce663842e72349c3a7b38cR325) no longer needs `createOrphanedTombstone` and so now runs on RocksDB (previously pending), putting the scan-captured version hand-off under coverage on the default engine.
- `HARPER_STORAGE_ENGINE=lmdb npx mocha unitTests/resources/auditLog.test.js` — 26 passing, 6 pending, including the 1,001-operation limiter and stale-tombstone regressions.
- `npm run test:unit:resources` — 1353 passing, 3 failing. All three failures (`randomAccessFields` directive parsing, `replayStructures`) reproduce with this branch's `resources/` reverted to the previously reviewed head, so they are base drift, not this change; the branch is well behind `main`. CI at the merge base is the authority.
- `npm run test:unit:main` cannot run on the authoring machine: `~/.harperdb/hdb_boot_properties.file` points at an unrelated local install whose `system.mdb` this branch cannot open. CI covers it.
- The [mid-loop rejection regression](https://github.com/HarperFast/harper/pull/2051/changes?w=1#diff-044077436ab82ecf6c481205bdf0ac84571b38edf4ce663842e72349c3a7b38cR111) is now parameterised over a throwing and a non-throwing `warn` stub; the non-throwing arm is what pins `entriesDeleted === 1`, killing the `onSuccess?.()`-before-`warn` mutation the previous single arm let through.
- Temporary 4,000-entry LMDB purge diagnostic — deletion completed in 56.6 ms at the 1,000-write cap; the diagnostic file was removed after measurement.
- `npx mocha unitTests/resources/subscriptionReplay.test.js` — 27 passing, 9 pending; `HARPER_STORAGE_ENGINE=lmdb` — 36 passing.
- Fresh unit and integration CI at `df65e3cdff4b` — every Node, Bun, uWS, and Windows shard passed; Windows purge completed `entries_deleted: 12000` in 2.149 seconds.
- Mutation check: removing the production callback's `return` makes the wiring test fail because `deleteHistory()` resolves before tombstone cleanup; restoring it passes.
- `npm run test:integration -- --isolation=none "integrationTests/apiTests/transaction-logs.test.mjs"` — 7/7 passing.

## Review coverage

Originally authored by GPT-5.6 Codex; the failure-reporting policy and the two follow-up cleanups by Claude Opus. Pre-push review at `df6b7987d83d`: Codex (graded) and Gemini, three rounds — one full and two delta. Round 1 found the unreachable `throw new Error('Versioned removal exhausted its retries')` in [`PrimaryRocksDatabase.removeIfVersion`](https://github.com/HarperFast/harper/pull/2051/changes?w=1#diff-857a3b7947228534c5055a20509e57ee05474fb84e726499a502a2cfc9631f83R44) (fixed — a single-retry flag, same two attempts and same `ERR_BUSY`-only retry) and comment narration (trimmed in `Table.ts` and `auditStore.ts`). Rounds 2 and 3 found no new correctness, concurrency, or error-handling defect. Round 1's Codex leg and domain adjudication timed out against the one-shot budget; Gemini carried independent coverage there.

Refs #1963

Co-Authored-By: Claude Opus <noreply@anthropic.com>

🤖 Generated with [Codex](https://openai.com/codex/) and [Claude Code](https://claude.com/claude-code)

<sub>Review-Coverage: authored=claude; ran=codex,gemini; declined=cursor-grok,cursor-composer,domain; rounds=3 @ df6b7987d83d</sub>

<sub>Human-Review-Need: 3 @ df6b7987d83d</sub>

### DIFF
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -5052,9 +5052,7 @@ export function makeTable(options) {
 			this.userSetEmbedders.add(attribute_name);
 		}
 		static async deleteHistory(endTime = 0, cleanupDeletedRecords = false): Promise<number> {
-			const maxConcurrentRemovals = primaryStore.useVersions
-				? MAX_CONCURRENT_LMDB_HISTORY_REMOVALS
-				: MAX_CONCURRENT_HISTORY_REMOVALS;
+			const maxConcurrentRemovals = isRocksDB ? MAX_CONCURRENT_HISTORY_REMOVALS : MAX_CONCURRENT_LMDB_HISTORY_REMOVALS;
 			const inFlightRemovals = new Set<Promise<void>>();
 			const removalSlotWaiters: Array<() => void> = [];
 			function startRemoval(remove: () => MaybePromise<void>, errorMessage: string, onSuccess?: () => void): void {
@@ -5109,7 +5107,7 @@ export function makeTable(options) {
 						await rest(); // yield to other async operations
 						if (value === null && localTime < endTime) {
 							const backpressure = queueRemoval(
-								() => primaryStore.remove(key, primaryStore.useVersions ? version : undefined),
+								() => primaryStore.remove(key, version),
 								'Error removing deleted record during deleteHistory'
 							);
 							if (backpressure) await backpressure;
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -5055,9 +5055,24 @@ export function makeTable(options) {
 			const maxConcurrentRemovals = isRocksDB ? MAX_CONCURRENT_HISTORY_REMOVALS : MAX_CONCURRENT_LMDB_HISTORY_REMOVALS;
 			const inFlightRemovals = new Set<Promise<void>>();
 			const removalSlotWaiters: Array<() => void> = [];
+			let removalsAttempted = 0;
+			let removalsSucceeded = 0;
+			let firstRemovalError: unknown;
 			function startRemoval(remove: () => MaybePromise<void>, errorMessage: string, onSuccess?: () => void): void {
+				removalsAttempted++;
 				const removal = new Promise<void>((resolve) => resolve(remove()))
-					.then(onSuccess, (error) => harperLogger.warn(errorMessage, error))
+					.then(
+						() => {
+							removalsSucceeded++;
+							onSuccess?.();
+						},
+						(error) => {
+							// record before logging: a logger that throws must not cost us the only
+							// reference to why every removal failed
+							if (firstRemovalError === undefined) firstRemovalError = error;
+							harperLogger.warn(errorMessage, error);
+						}
+					)
 					.catch(() => undefined)
 					.finally(() => {
 						inFlightRemovals.delete(removal);
@@ -5116,6 +5131,13 @@ export function makeTable(options) {
 				} finally {
 					await drainRemovals();
 				}
+			}
+			if (removalsAttempted > 0 && removalsSucceeded === 0) {
+				// every removal we attempted failed, so this purge made no progress at all. Reporting
+				// success here would be indistinguishable from "nothing was eligible for removal", and an
+				// operator pruning to bound disk growth would never learn the store rejected every write.
+				// Partial failures stay best-effort (logged, and excluded from the returned count).
+				throw firstRemovalError ?? new Error('Every removal attempted during deleteHistory failed');
 			}
 			return entriesDeleted;
 		}
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -5051,22 +5051,20 @@ export function makeTable(options) {
 			this.userSetEmbedders.add(attribute_name);
 		}
 		static async deleteHistory(endTime = 0, cleanupDeletedRecords = false): Promise<number> {
-			const inFlightRemovals = new Set<Promise<void>>();
+			const removalSlots = new Array<Promise<void>>(MAX_CONCURRENT_HISTORY_REMOVALS);
+			let nextRemovalSlot = 0;
 			async function queueRemoval(
 				remove: () => MaybePromise<void>,
 				errorMessage: string,
 				onSuccess?: () => void
 			): Promise<void> {
-				const removal = new Promise<void>((resolve) => resolve(remove()))
+				const slot = nextRemovalSlot++ % MAX_CONCURRENT_HISTORY_REMOVALS;
+				await removalSlots[slot];
+				removalSlots[slot] = new Promise<void>((resolve) => resolve(remove()))
 					.then(onSuccess, (error) => harperLogger.warn(errorMessage, error))
-					.catch(() => undefined)
-					.finally(() => inFlightRemovals.delete(removal));
-				inFlightRemovals.add(removal);
-				if (inFlightRemovals.size >= MAX_CONCURRENT_HISTORY_REMOVALS) {
-					await Promise.race(inFlightRemovals);
-				}
+					.catch(() => undefined);
 			}
-			const drainRemovals = () => Promise.all(inFlightRemovals);
+			const drainRemovals = () => Promise.all(removalSlots);
 			let entriesDeleted = 0;
 			for (const auditRecord of auditStore.getRange({
 				start: 0,
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -5059,6 +5059,7 @@ export function makeTable(options) {
 			): Promise<void> {
 				const removal = new Promise<void>((resolve) => resolve(remove()))
 					.then(onSuccess, (error) => harperLogger.warn(errorMessage, error))
+					.catch(() => undefined)
 					.finally(() => inFlightRemovals.delete(removal));
 				inFlightRemovals.add(removal);
 				if (inFlightRemovals.size >= MAX_CONCURRENT_HISTORY_REMOVALS) {
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -5105,7 +5105,7 @@ export function makeTable(options) {
 					for (const entry of primaryStore.getRange({ start: 0, versions: true })) {
 						const { key, value, localTime, version } = entry;
 						await rest(); // yield to other async operations
-						if (value === null && localTime < endTime) {
+						if (value === null && version != null && localTime < endTime) {
 							const backpressure = queueRemoval(
 								() => primaryStore.remove(key, version),
 								'Error removing deleted record during deleteHistory'
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -5051,50 +5051,69 @@ export function makeTable(options) {
 			this.userSetEmbedders.add(attribute_name);
 		}
 		static async deleteHistory(endTime = 0, cleanupDeletedRecords = false): Promise<number> {
-			const removalSlots = new Array<Promise<void>>(MAX_CONCURRENT_HISTORY_REMOVALS);
-			let nextRemovalSlot = 0;
-			async function queueRemoval(
+			const inFlightRemovals = new Set<Promise<void>>();
+			const removalSlotWaiters: Array<() => void> = [];
+			function startRemoval(remove: () => MaybePromise<void>, errorMessage: string, onSuccess?: () => void): void {
+				const removal = new Promise<void>((resolve) => resolve(remove()))
+					.then(onSuccess, (error) => harperLogger.warn(errorMessage, error))
+					.catch(() => undefined)
+					.finally(() => {
+						inFlightRemovals.delete(removal);
+						removalSlotWaiters.shift()?.();
+					});
+				inFlightRemovals.add(removal);
+			}
+			function queueRemoval(
 				remove: () => MaybePromise<void>,
 				errorMessage: string,
 				onSuccess?: () => void
-			): Promise<void> {
-				const slot = nextRemovalSlot++ % MAX_CONCURRENT_HISTORY_REMOVALS;
-				await removalSlots[slot];
-				removalSlots[slot] = new Promise<void>((resolve) => resolve(remove()))
-					.then(onSuccess, (error) => harperLogger.warn(errorMessage, error))
-					.catch(() => undefined);
+			): Promise<void> | undefined {
+				if (inFlightRemovals.size >= MAX_CONCURRENT_HISTORY_REMOVALS) {
+					return new Promise<void>((resolve) => {
+						removalSlotWaiters.push(resolve);
+					}).then(() => startRemoval(remove, errorMessage, onSuccess));
+				}
+				startRemoval(remove, errorMessage, onSuccess);
 			}
-			const drainRemovals = () => Promise.all(removalSlots);
+			const drainRemovals = () => Promise.all(inFlightRemovals);
 			let entriesDeleted = 0;
-			for (const auditRecord of auditStore.getRange({
-				start: 0,
-				end: endTime,
-			})) {
-				await rest(); // yield to other async operations
-				if (auditRecord.tableId !== tableId) continue;
-				await queueRemoval(
-					() => removeAuditEntry(auditStore, auditRecord),
-					'Error removing audit entry during deleteHistory',
-					() => {
-						entriesDeleted++;
-					}
-				);
+			try {
+				for (const auditRecord of auditStore.getRange({
+					start: 0,
+					end: endTime,
+				})) {
+					await rest(); // yield to other async operations
+					if (auditRecord.tableId !== tableId) continue;
+					const backpressure = queueRemoval(
+						() => removeAuditEntry(auditStore, auditRecord),
+						'Error removing audit entry during deleteHistory',
+						() => {
+							entriesDeleted++;
+						}
+					);
+					if (backpressure) await backpressure;
+				}
+			} finally {
+				await drainRemovals();
 			}
-			await drainRemovals();
 			if (cleanupDeletedRecords) {
 				// this is separate procedure we can do if the records are not being cleaned up by the audit log. This shouldn't
 				// ever happen, but if there are cleanup failures for some reason, we can run this to clean up the records
-				for (const entry of primaryStore.getRange({ start: 0, versions: true })) {
-					const { key, value, localTime, version } = entry;
-					await rest(); // yield to other async operations
-					if (value === null && localTime < endTime) {
-						await queueRemoval(
-							() => primaryStore.remove(key, primaryStore.useVersions ? version : undefined),
-							'Error removing deleted record during deleteHistory'
-						);
+				try {
+					for (const entry of primaryStore.getRange({ start: 0, versions: true })) {
+						const { key, value, localTime, version } = entry;
+						await rest(); // yield to other async operations
+						if (value === null && localTime < endTime) {
+							const backpressure = queueRemoval(
+								() => primaryStore.remove(key, primaryStore.useVersions ? version : undefined),
+								'Error removing deleted record during deleteHistory'
+							);
+							if (backpressure) await backpressure;
+						}
 					}
+				} finally {
+					await drainRemovals();
 				}
-				await drainRemovals();
 			}
 			return entriesDeleted;
 		}
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -120,6 +120,7 @@ const NULL_WITH_TIMESTAMP = new Uint8Array(9);
 NULL_WITH_TIMESTAMP[8] = 0xc0; // null
 const UNCACHEABLE_TIMESTAMP = Infinity; // we use this when dynamic content is accessed that we can't safely cache, and this prevents earlier timestamps from change the "last" modification
 const RECORD_PRUNING_INTERVAL = 60000; // one minute
+const MAX_CONCURRENT_HISTORY_REMOVALS = 10;
 // RocksDB-only: number of eviction/tombstone removals coalesced into a single transaction commit.
 // Each evict otherwise pays a full transaction commit, so batching amortizes that cost. LMDB already
 // coalesces async writes per event turn (eventTurnBatching), so it keeps the per-record path.
@@ -5050,6 +5051,21 @@ export function makeTable(options) {
 			this.userSetEmbedders.add(attribute_name);
 		}
 		static async deleteHistory(endTime = 0, cleanupDeletedRecords = false): Promise<number> {
+			const inFlightRemovals = new Set<Promise<void>>();
+			async function queueRemoval(
+				remove: () => MaybePromise<void>,
+				errorMessage: string,
+				onSuccess?: () => void
+			): Promise<void> {
+				const removal = new Promise<void>((resolve) => resolve(remove()))
+					.then(onSuccess, (error) => harperLogger.warn(errorMessage, error))
+					.finally(() => inFlightRemovals.delete(removal));
+				inFlightRemovals.add(removal);
+				if (inFlightRemovals.size >= MAX_CONCURRENT_HISTORY_REMOVALS) {
+					await Promise.race(inFlightRemovals);
+				}
+			}
+			const drainRemovals = () => Promise.all(inFlightRemovals);
 			let entriesDeleted = 0;
 			for (const auditRecord of auditStore.getRange({
 				start: 0,
@@ -5057,13 +5073,15 @@ export function makeTable(options) {
 			})) {
 				await rest(); // yield to other async operations
 				if (auditRecord.tableId !== tableId) continue;
-				try {
-					await removeAuditEntry(auditStore, auditRecord);
-					entriesDeleted++;
-				} catch (error) {
-					harperLogger.warn('Error removing audit entry during deleteHistory', error);
-				}
+				await queueRemoval(
+					() => removeAuditEntry(auditStore, auditRecord),
+					'Error removing audit entry during deleteHistory',
+					() => {
+						entriesDeleted++;
+					}
+				);
 			}
+			await drainRemovals();
 			if (cleanupDeletedRecords) {
 				// this is separate procedure we can do if the records are not being cleaned up by the audit log. This shouldn't
 				// ever happen, but if there are cleanup failures for some reason, we can run this to clean up the records
@@ -5071,13 +5089,13 @@ export function makeTable(options) {
 					const { value, localTime } = entry;
 					await rest(); // yield to other async operations
 					if (value === null && localTime < endTime) {
-						try {
-							await removeEntry(primaryStore, entry);
-						} catch (error) {
-							harperLogger.warn('Error removing deleted record during deleteHistory', error);
-						}
+						await queueRemoval(
+							() => removeEntry(primaryStore, entry),
+							'Error removing deleted record during deleteHistory'
+						);
 					}
 				}
+				await drainRemovals();
 			}
 			return entriesDeleted;
 		}
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -5050,7 +5050,6 @@ export function makeTable(options) {
 			this.userSetEmbedders.add(attribute_name);
 		}
 		static async deleteHistory(endTime = 0, cleanupDeletedRecords = false): Promise<number> {
-			let completion: Promise<void>;
 			let entriesDeleted = 0;
 			for (const auditRecord of auditStore.getRange({
 				start: 0,
@@ -5058,7 +5057,14 @@ export function makeTable(options) {
 			})) {
 				await rest(); // yield to other async operations
 				if (auditRecord.tableId !== tableId) continue;
-				completion = removeAuditEntry(auditStore, auditRecord);
+				try {
+					// awaited so a rejection is caught here instead of escaping as an unhandled
+					// rejection once a later iteration's promise replaces this one (see auditStore.ts
+					// scheduleAuditCleanup for the same fix)
+					await removeAuditEntry(auditStore, auditRecord);
+				} catch (error) {
+					harperLogger.warn('Error removing audit entry during deleteHistory', error);
+				}
 				entriesDeleted++;
 			}
 			if (cleanupDeletedRecords) {
@@ -5068,11 +5074,14 @@ export function makeTable(options) {
 					const { value, localTime } = entry;
 					await rest(); // yield to other async operations
 					if (value === null && localTime < endTime) {
-						completion = removeEntry(primaryStore, entry);
+						try {
+							await removeEntry(primaryStore, entry);
+						} catch (error) {
+							harperLogger.warn('Error removing deleted record during deleteHistory', error);
+						}
 					}
 				}
 			}
-			await completion;
 			return entriesDeleted;
 		}
 		static async *getHistory(startTime = 0, endTime = Infinity) {
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -5067,8 +5067,7 @@ export function makeTable(options) {
 							onSuccess?.();
 						},
 						(error) => {
-							// record before logging: a logger that throws must not cost us the only
-							// reference to why every removal failed
+							// capture before logging: a throwing logger must not cost us the error we may rethrow
 							if (firstRemovalError === undefined) firstRemovalError = error;
 							harperLogger.warn(errorMessage, error);
 						}
@@ -5133,10 +5132,8 @@ export function makeTable(options) {
 				}
 			}
 			if (removalsAttempted > 0 && removalsSucceeded === 0) {
-				// every removal we attempted failed, so this purge made no progress at all. Reporting
-				// success here would be indistinguishable from "nothing was eligible for removal", and an
-				// operator pruning to bound disk growth would never learn the store rejected every write.
-				// Partial failures stay best-effort (logged, and excluded from the returned count).
+				// zero progress must not report the same success as "nothing was eligible" (see DESIGN.md);
+				// partial failures stay best-effort, logged and excluded from the returned count
 				throw firstRemovalError ?? new Error('Every removal attempted during deleteHistory failed');
 			}
 			return entriesDeleted;
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -5058,14 +5058,11 @@ export function makeTable(options) {
 				await rest(); // yield to other async operations
 				if (auditRecord.tableId !== tableId) continue;
 				try {
-					// awaited so a rejection is caught here instead of escaping as an unhandled
-					// rejection once a later iteration's promise replaces this one (see auditStore.ts
-					// scheduleAuditCleanup for the same fix)
 					await removeAuditEntry(auditStore, auditRecord);
+					entriesDeleted++;
 				} catch (error) {
 					harperLogger.warn('Error removing audit entry during deleteHistory', error);
 				}
-				entriesDeleted++;
 			}
 			if (cleanupDeletedRecords) {
 				// this is separate procedure we can do if the records are not being cleaned up by the audit log. This shouldn't
@@ -6230,7 +6227,7 @@ export function makeTable(options) {
 	}
 	function addDeleteRemoval() {
 		deleteCallbackHandle = auditStore?.addDeleteRemovalCallback(tableId, primaryStore, (id: Id, version: number) => {
-			primaryStore.remove(id, version);
+			return primaryStore.remove(id, version);
 		});
 	}
 	function runRecordExpirationEviction() {
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -5085,11 +5085,11 @@ export function makeTable(options) {
 				// this is separate procedure we can do if the records are not being cleaned up by the audit log. This shouldn't
 				// ever happen, but if there are cleanup failures for some reason, we can run this to clean up the records
 				for (const entry of primaryStore.getRange({ start: 0, versions: true })) {
-					const { value, localTime } = entry;
+					const { key, value, localTime, version } = entry;
 					await rest(); // yield to other async operations
 					if (value === null && localTime < endTime) {
 						await queueRemoval(
-							() => removeEntry(primaryStore, entry),
+							() => primaryStore.remove(key, primaryStore.useVersions ? version : undefined),
 							'Error removing deleted record during deleteHistory'
 						);
 					}
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -121,6 +121,7 @@ NULL_WITH_TIMESTAMP[8] = 0xc0; // null
 const UNCACHEABLE_TIMESTAMP = Infinity; // we use this when dynamic content is accessed that we can't safely cache, and this prevents earlier timestamps from change the "last" modification
 const RECORD_PRUNING_INTERVAL = 60000; // one minute
 const MAX_CONCURRENT_HISTORY_REMOVALS = 10;
+const MAX_CONCURRENT_LMDB_HISTORY_REMOVALS = 1000;
 // RocksDB-only: number of eviction/tombstone removals coalesced into a single transaction commit.
 // Each evict otherwise pays a full transaction commit, so batching amortizes that cost. LMDB already
 // coalesces async writes per event turn (eventTurnBatching), so it keeps the per-record path.
@@ -5051,6 +5052,9 @@ export function makeTable(options) {
 			this.userSetEmbedders.add(attribute_name);
 		}
 		static async deleteHistory(endTime = 0, cleanupDeletedRecords = false): Promise<number> {
+			const maxConcurrentRemovals = primaryStore.useVersions
+				? MAX_CONCURRENT_LMDB_HISTORY_REMOVALS
+				: MAX_CONCURRENT_HISTORY_REMOVALS;
 			const inFlightRemovals = new Set<Promise<void>>();
 			const removalSlotWaiters: Array<() => void> = [];
 			function startRemoval(remove: () => MaybePromise<void>, errorMessage: string, onSuccess?: () => void): void {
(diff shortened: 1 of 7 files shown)


resources/auditStore.ts
```diff
@@ -278,15 +278,30 @@ export function openAuditStore(rootStore) {
...
+				// the .catch is attached here, before anything else can throw synchronously, so this
+				// promise is never left without a handler
+				tombstoneRemoval = Promise.resolve(
+					auditStore.deleteCallbacks?.[tableId]?.(auditRecord.recordId, auditRecord.version)
+				).catch((error) => {
+					harperLogger.warn('Error removing deleted record while removing its audit entry', error);
+				});
+			} catch (error) {
+				harperLogger.warn('Error removing deleted record while removing its audit entry', error);
+			}
```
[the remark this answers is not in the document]
kriszyp: Good catch, applied exactly as suggested in 70b424922 — also extended the isolated `removeAuditEntry` unit test to cover both the rejecting and synchronously-throwing callback cases, since the whole point of the executor pattern is handling both in one place.

resources/Table.ts
```diff
@@ -5050,29 +5051,53 @@ export function makeTable(options) {
...
 		static async deleteHistory(endTime = 0, cleanupDeletedRecords = false): Promise<number> {
-			let completion: Promise<void>;
+			const inFlightRemovals = new Set<Promise<void>>();
+			async function queueRemoval(
+				remove: () => MaybePromise<void>,
+				errorMessage: string,
+				onSuccess?: () => void
+			): Promise<void> {
+				const removal = new Promise<void>((resolve) => resolve(remove()))
+					.then(onSuccess, (error) => harperLogger.warn(errorMessage, error))
```
cb1kenobi: **Medium: every removal failure is now swallowed — the operation still reports success**

The old code did `await completion` on the *last* iteration's promise, so a final-iteration rejection propagated out of `deleteHistory` → `ResourceBridge.deleteTransactionLogsBefore` (line 516) → the operation errored. This `.then(onSuccess, warn).catch(() => undefined)` chain removes the only path by which a removal failure reaches the caller.

Concretely: a store that is read-only, full, or closing now emits one warn line per record and returns `entries_deleted: 0` inside a **successful** operation response. An operator scripting `delete_transaction_logs_before` to bound disk growth gets no signal that pruning did nothing.

The PR body lists this under "Risks & open questions" — flagging to confirm it is real and worth a follow-up ticket rather than being lost in the description.

Suggested fix: track a failure count alongside `entriesDeleted` and surface it — either reject when `failures > 0 && successes === 0`, or add `entries_failed` to `DeleteTransactionLogsBeforeResults`.

—
_Generated by Barber AI_

resources/Table.ts
```diff
@@ -5050,29 +5051,53 @@ export function makeTable(options) {
...
+					.catch(() => undefined)
+					.finally(() => inFlightRemovals.delete(removal));
+				inFlightRemovals.add(removal);
+				if (inFlightRemovals.size >= MAX_CONCURRENT_HISTORY_REMOVALS) {
+					await Promise.race(inFlightRemovals);
```
cb1kenobi: **Low: `Promise.race` retains one reaction record per iteration on any long-pending removal**

`Promise.race` calls `.then` on every element each time it is invoked. A removal that stays pending while the other nine cycle gets re-raced once per subsequent `queueRemoval` call, accumulating one reaction record per iteration on that single promise — O(entries pruned) memory, unbounded over a large prune. Requires a storage write that stalls indefinitely, hence Low.

Suggested fix: a fixed slot array removes the race entirely, and with it the `.finally` self-delete and the `drainRemovals` closure:

```js
const slots = new Array(MAX_CONCURRENT_HISTORY_REMOVALS);
let next = 0;
// queue: await slots[next % MAX]; slots[next++ % MAX] = removal;
// drain: await Promise.all(slots);
```

—
_Generated by Barber AI_

resources/Table.ts
```diff
@@ -6221,7 +6246,7 @@ export function makeTable(options) {
 	}
 	function addDeleteRemoval() {
 		deleteCallbackHandle = auditStore?.addDeleteRemovalCallback(tableId, primaryStore, (id: Id, version: number) => {
-			primaryStore.remove(id, version);
+			return primaryStore.remove(id, version);
```
cb1kenobi, line 6287: **Medium: the only production wiring of the tombstone join is untested — the mutation survives**

This `return` is what makes `removeAuditEntry`'s new `Promise.all([tombstoneRemoval, auditRemoval])` join anything at all; the callback's promise is the thing being joined. Without it, `tombstoneRemoval` resolves immediately on `undefined` and the tombstone removal silently reverts to fire-and-forget — the exact bug this PR fixes, reintroduced with nothing to catch it.

Verified by mutation: dropping the `return`, rebuilding with `tsc -p tsconfig.build.json` (necessary — `#src/*` resolves to `dist/`, so editing the `.ts` alone is a no-op), then running `HARPER_STORAGE_ENGINE=lmdb mocha unitTests/resources/auditLog.test.js` gives **21 passing, 0 failing**. The new `removeAuditEntry` tests use a hand-built `fakeAuditStore` with a literal `deleteCallbacks` map and never exercise `addDeleteRemovalCallback`.

The PR body already notes this; a regression guard seems worth having, since this line is the load-bearing half of the fix.

Suggested fix: one LMDB test that deletes a record, runs `deleteHistory`, and asserts the primary-store tombstone is gone *by the time `deleteHistory` resolves* — not after a delay, which would pass either way.

—
_Generated by Barber AI_

resources/Table.ts
kriszyp: Addressed in 7eb25c2a5, with the tradeoff documented in e26312e9e: the live Set/Promise.race throttle is now ten fixed promise slots. A slot is awaited before reuse, so a long-pending removal receives one waiter instead of another reaction on every later iteration.

— GPT-5 Codex

resources/Table.ts
kriszyp: Addressed in 7eb25c2a5, with test cleanup in e26312e9e: the LMDB regression now runs through the table-registered callback, holds the primary-store removal pending, proves deleteHistory stays pending, then verifies the tombstone is gone when it resolves. Dropping the production callback return makes this test fail for the intended reason.

— GPT-5 Codex

unitTests/resources/auditLog.test.js
```diff
@@ -102,6 +104,257 @@ describe('Audit log', () => {
...
+			harperLogger.warn = (...args) => {
+				warnings.push(args);
+				throw new Error('simulated logging failure');
+			};
+			try {
+				entriesDeleted = await AuditedTable.deleteHistory(Date.now() + 60_000);
+			} finally {
+				harperLogger.warn = originalWarn;
+			}
+			assert.equal(entriesDeleted, 1, 'only the successful removal should be counted, not the rejected one');
```
cb1kenobi: **Nit: this failure-path count guard is masked by its own logger stub**

This is the only assertion pinning "a failed removal is not counted", and it runs inside the block where `harperLogger.warn` is stubbed to throw. Mutation-checked both placements (mutate `.ts`, `tsc -p tsconfig.build.json`, verify in `dist/resources/Table.js`, then `HARPER_STORAGE_ENGINE=lmdb mocha unitTests/resources/auditLog.test.js`):

- `.then(onSuccess, (error) => { onSuccess?.(); harperLogger.warn(...); })` → **21 passing, 1 failing** (caught)
- `.then(onSuccess, (error) => { harperLogger.warn(...); onSuccess?.(); })` → **22 passing, 0 failing** (survives — the injected throw preempts the increment)

So the regression guard for `1a935546a` ("stop over-counting failed removals") only holds for one of the two placements. One extra pass with a non-throwing `warn` asserting `entriesDeleted === 1` closes it.

Everything else in the rework mutation-checks clean: dropping the `addDeleteRemoval` `return`, dropping the `Promise.all([tombstoneRemoval, auditRemoval])` join, removing `await removalSlots[slot]`, no-op'ing `drainRemovals`, bumping `MAX_CONCURRENT_HISTORY_REMOVALS` to 11, dropping the terminal `.catch`, and dropping the `tableStores[tableId]` registration are all killed.

—
_Generated by Barber AI_

resources/Table.ts
```diff
@@ -5050,29 +5051,51 @@ export function makeTable(options) {
...
 			if (cleanupDeletedRecords) {
 				// this is separate procedure we can do if the records are not being cleaned up by the audit log. This shouldn't
 				// ever happen, but if there are cleanup failures for some reason, we can run this to clean up the records
 				for (const entry of primaryStore.getRange({ start: 0, versions: true })) {
 					const { value, localTime } = entry;
 					await rest(); // yield to other async operations
 					if (value === null && localTime < endTime) {
-						completion = removeEntry(primaryStore, entry);
+						await queueRemoval(
+							() => removeEntry(primaryStore, entry),
```
heskew: **Medium: a delayed tombstone sweep can delete a recreated live record**

The range scan captures `entry` before `queueRemoval` waits for its slot, but `removeEntry(primaryStore, entry)` forwards no version guard, so LMDB performs an unconditional remove. If another worker recreates this key while the slot is blocked, the delayed cleanup deletes the new live value and can leave its secondary-index or blob state orphaned. A focused LMDB reproduction confirmed the ordering: the unconditional removal deleted the newer version, while passing the stale `entry.version` returned `false` and preserved it.

The old `await rest()` already left a smaller stale-entry window, but the new slot wait materially widens it for the eleventh and later removals under slow storage. Please pass `entry.version`, as the neighboring cleanup paths do, and add a regression that holds a slot, recreates the ID, then releases it.

---

resources/Table.ts
```diff
@@ -5050,29 +5051,51 @@ export function makeTable(options) {
...
-			let completion: Promise<void>;
+			const removalSlots = new Array<Promise<void>>(MAX_CONCURRENT_HISTORY_REMOVALS);
+			let nextRemovalSlot = 0;
+			async function queueRemoval(
+				remove: () => MaybePromise<void>,
+				errorMessage: string,
+				onSuccess?: () => void
+			): Promise<void> {
+				const slot = nextRemovalSlot++ % MAX_CONCURRENT_HISTORY_REMOVALS;
+				await removalSlots[slot];
```
cb1kenobi: **High: awaiting a slot per iteration defeats LMDB's write batching — `deleteHistory()` is ~30-45x slower**

`await removalSlots[slot]` runs on every iteration of the scan loop, so past the 10-slot window each removal must fully settle before the next can be issued. The old code issued removals without awaiting, which let LMDB coalesce them via per-event-turn batching (the same batching the RocksDB comment above `BATCHED_EVICTION` calls out). N removals now become ~N/10 sequential commit round-trips. RocksDB is unaffected — it already pays a commit per removal.

Measured on this branch vs the merge base (`b852a722c`), `HARPER_STORAGE_ENGINE=lmdb`, 4000 audit entries, `entries_deleted=4000` on both sides:

| | run 1 | run 2 | run 3 |
|---|---|---|---|
| merge base | 83.9 ms | 93.0 ms | 89.2 ms |
| this PR | 2771.8 ms | 2847.0 ms | 3849.4 ms |

CI shows the same ratio. In `integrationTests/database/txnlog-purge-stale-read-blast.test.ts`, arm `5. CONTRAST ARM ... [lmdb]` (`entries_deleted=12000`) on `Integration Tests 1/6 (Windows, Node.js v24)` takes 1212-1673 ms across the last 8 main runs, 52975 ms at `e26312e9`, and 60472 ms at `0ab50e13` — where it now fails with `delete_transaction_logs_before` job did not complete within 60000ms. Production code is byte-identical between those two SHAs, so the test-only follow-up didn't cause the flip; the arm was already sitting just under the 60 s budget.

The PR body anticipates "temporary head-of-line blocking if one removal is much slower than the others", but this is a steady-state throughput cost on every LMDB purge, not an occasional stall — a production `delete_transaction_logs_before` over millions of audit entries pays it linearly.

Suggested fix: keep the pending set bounded without serializing the common case. Track the outstanding promises and only await when the bound is actually reached *and* the slot is still pending (e.g. a counter plus a single `drain()` when it hits the cap), so consecutive removals stay in the same event turn and LMDB can still batch them. Raising `MAX_CONCURRENT_HISTORY_REMOVALS` alone only shifts the constant.

—
_Generated by Barber AI_

resources/Table.ts
kriszyp: Fixed in 71a1ff2e6. The sweep now snapshots the tombstone key/version before either asynchronous wait and passes that version through LMDB’s `useVersions` conditional remove, so a recreated record makes the stale removal resolve `false` instead of being deleted. The regression holds the real remove, recreates the ID, releases it, and verifies both the conditional miss and surviving live value; a paired positive test proves an unchanged tombstone is still removed. — GPT-5 Codex

resources/Table.ts
cb1kenobi: **High: still ~38x slower at this head — and the isolation experiment now proves it is this line**

`Guard delayed tombstone cleanup by version` does not touch this line, and the LMDB throughput regression is unchanged at `71a1ff2e`. The prior thread here was resolved without a code change, so re-posting with the attribution proof that was missing.

I previously attributed the slowdown to this `await` by reasoning about LMDB's per-event-turn write batching rather than by isolating it. I have now isolated it. Methodology: mutate the `.ts`, rebuild with `tsc -p tsconfig.build.json`, and grep the mutant in `dist/` before running (`package.json` maps `#src/*` -> `./dist/*.js` and `.mocharc.json` sets no `--conditions`, so mutating the `.ts` alone is a silent no-op). Every row below had its dist grep confirmed. Bench: LMDB, 4000 puts + 4000 deletes, `entries_deleted=8000` on every row.

| build | slot-await | `removeAuditEntry` tombstone join | `deleteHistory` ms |
|---|---|---|---|
| merge base `b852a722c` | absent | absent | **162.0 / 165.9 / 166.9** |
| PR head `71a1ff2e` (control) | present | present | **6340.5 / 6363.3 / 6434.3** |
| slot-await removed | **removed** | present | **164.0 / 160.2 / 175.1** |
| join removed | present | **removed** | **6427.0 / 7675.5 / 6367.6** |
| both removed | **removed** | **removed** | **167.0 / 162.9** |

Removing this `await` alone restores merge-base speed; removing the tombstone join alone changes nothing. The slot-await is the sole cause — my earlier caveat is resolved, and the finding stands as originally stated.

CI at this head agrees: `integrationTests/database/txnlog-purge-stale-read-blast.test.ts`, arm `5. CONTRAST ARM ... [lmdb]`, took **60483 ms and failed** with `QA-782: job ... type delete_transaction_logs_before ... did not complete within 60000ms` (Integration Tests 1/6, Windows, Node v24). The `[rocksdb]` arm of the same test: 2174 ms.

Suggested fix: keep the pending set bounded without awaiting a slot on every iteration in the common case — only await when the slot's promise is actually still pending, or raise the bound substantially and drain in batches, so consecutive removals stay in the same event turn and LMDB can coalesce them.

—
_Generated by Barber AI_

resources/Table.ts
```diff
@@ -5050,29 +5051,51 @@ export function makeTable(options) {
...
 				// this is separate procedure we can do if the records are not being cleaned up by the audit log. This shouldn't
 				// ever happen, but if there are cleanup failures for some reason, we can run this to clean up the records
 				for (const entry of primaryStore.getRange({ start: 0, versions: true })) {
-					const { value, localTime } = entry;
+					const { key, value, localTime, version } = entry;
 					await rest(); // yield to other async operations
 					if (value === null && localTime < endTime) {
-						completion = removeEntry(primaryStore, entry);
+						await queueRemoval(
+							() => primaryStore.remove(key, primaryStore.useVersions ? version : undefined),
```
cb1kenobi: **Medium: this guard is inert on RocksDB, the default storage engine**

The version is correctly captured at scan time and applied at fire time, and on LMDB it works. On RocksDB it does nothing, for two independent reasons.

Runtime probe, same table built through `table()` on each engine, stale version passed explicitly:

| | `primaryStore.useVersions` | `remove(key, STALE_version)` resolves | record afterwards |
|---|---|---|---|
| LMDB | `true` | `false` | **preserved** |
| RocksDB (default) | **`undefined`** | `undefined` | **deleted** |

1. `PrimaryRocksDatabase` has no `useVersions` property, so the ternary passes `undefined` — no condition is applied at all. `resources/databases.ts:88` already notes that "lmdb/rocksdb don't forward `useVersions`".
2. Even when the version is passed explicitly, `@harperfast/rocksdb-js` types it `remove(key: Key, options?: StoreRemoveOptions): Promise<void>` — the second argument is an options object, not an ifVersion — and the record is removed regardless.

So the exact race this commit is named for is still live on the default engine: a record recreated between the scan reading the tombstone and the queued removal firing is destroyed. The slot-await on line 5062 makes that window materially longer than it was on `main`. The new tests cannot catch this — they are gated on `auditStore.reusableIterable`, which is `true` for `RocksTransactionLogStore`; with the gate removed they fail in `createOrphanedTombstone` because the harness cannot build an orphaned tombstone against a RocksDB transaction log.

Secondary, on these same lines: a conditional miss resolves `false` rather than rejecting, and this branch passes no `onSuccess`, so a skipped removal is neither logged nor counted. Combined with the swallowed-failure issue on line 5061, a sweep that removed nothing looks identical to one that removed everything.

Suggested fix: gate the conditional on something that is true only where it is honored (e.g. `primaryStore.encoder?.useVersions` plus an explicit engine check), and on RocksDB re-read the entry immediately before removing and skip when `version` has moved. Either way, log the skip so a version-mismatch is distinguishable from a completed removal.

—
_Generated by Barber AI_

resources/Table.ts
cb1kenobi: **Fixed — confirmed by measurement at `04ffb67b`.** Nice fix, and the rework is the right shape: it went after the actual mechanism rather than just reverting.

Re-ran the same benchmark shape (4000 puts + 4000 deletes, `entries_deleted=8000` every row, 3 rounds), same machine and worktree, each build grep-confirmed in `dist/` before running:

| build | ms (r1 / r2 / r3) |
|---|---|
| merge base `b852a722c` | 158.2 / 140.9 / 149.9 |
| prior head `71a1ff2e` | 6340.5 / 6363.3 / 6434.3 |
| head `04ffb67b` | **179.2 / 147.1 / 150.1** |

~42x faster than the prior head and within noise of the merge base (r2/r3 differ by 4–6 ms; r1 is warm-up). Gone, not just reduced.

CI corroborates it independently. In `integrationTests/database/txnlog-purge-stale-read-blast.test.ts`, test 5 ("CONTRAST ARM", the `delete_transaction_logs_before` arm):

| arm | prior head | head `04ffb67b` |
|---|---|---|
| `[lmdb]` | 60,483 ms (timing out) | **1,550 ms** |
| `[rocksdb]` | 2,174 ms | 2,076 ms |

Whole suite now passes on both arms (`[lmdb]` 15,419 ms, `[rocksdb]` 20,445 ms), and LMDB is faster than RocksDB there.

I also checked what the fix cost, since `+52/-33` on a concurrency path is a real rework rather than a revert:

- **The bound still holds.** Mutating `MAX_CONCURRENT_LMDB_HISTORY_REMOVALS` 1000 → 1000000 (grep-confirmed in `dist/`) fails `deleteHistory limits concurrent removals without serializing them` — 23 passing / 1 failing, `AssertionError: the next removal must wait while the concurrency limit is occupied`. `maximumActiveRemovals === 1000` is pinned too.
- **The tombstone join survived.** Dropping the `return` at line 6270 still fails `deleteHistory waits for the table-registered tombstone removal callback` (23 passing / 1 failing).
- **The version guard survived** the rework unchanged at line 5112.
- **No RocksDB regression** from the new 10-cap on a path that previously had no bound: per-8000-entry cost is flat vs the merge base (149.1 / 289.3 / 377.1 ms head vs 139.3 / 281.8 / 387.6 ms base, cumulative).

Unmutated control: 24 passing / 5 pending on both engines.

—
_Generated by Barber AI_

resources/Table.ts
cb1kenobi: **Still open at `04ffb67b`.** The rework moved this chain into `startRemoval`, but it is character-identical (now `resources/Table.ts:5063`):

```js
.then(onSuccess, (error) => harperLogger.warn(errorMessage, error))
.catch(() => undefined)
```

Failures are still warn-logged only: not counted, not rethrown, never reaching the response. `delete_transaction_logs_before` still returns `entries_deleted: 0` inside a **successful** response whether nothing needed deleting or every removal failed — the two are indistinguishable to the caller.

Flagging that this one is unchanged rather than re-arguing it; it was deferred to the human reviewer.

—
_Generated by Barber AI_

resources/Table.ts
cb1kenobi: **Still open at `04ffb67b`.** The guard survived the rework unchanged (now `resources/Table.ts:5112`). Re-ran the runtime probe at this head — same table built through `table()` on each engine, a stale version passed explicitly:

| | `primaryStore.useVersions` | ctor | `remove(key, STALE_version)` | record afterwards |
|---|---|---|---|---|
| LMDB | `true` | `LMDBStore` | `false` | **preserved** |
| RocksDB (default) | **`undefined`** | `PrimaryRocksDatabase` | `undefined` | **deleted** |

So on the default engine the ternary passes `undefined` and the removal is unconditional. Both new tests are LMDB-only, so CI still cannot catch this.

One new interaction worth knowing before fixing it: `useVersions` is now *also* the discriminator for the removal concurrency cap added in this commit (line 5055). Adding `useVersions` to `PrimaryRocksDatabase` to fix this finding would, as a side effect, raise RocksDB's in-flight removals from 10 to 1000. Details in the separate comment on line 5055.

—
_Generated by Barber AI_

resources/Table.ts
```diff
@@ -5050,29 +5052,73 @@ export function makeTable(options) {
 			this.userSetEmbedders.add(attribute_name);
 		}
 		static async deleteHistory(endTime = 0, cleanupDeletedRecords = false): Promise<number> {
-			let completion: Promise<void>;
+			const maxConcurrentRemovals = primaryStore.useVersions
```
cb1kenobi: **Low: `useVersions` now doubles as the LMDB write-batching discriminator — it couples two unrelated concerns to one flag**

This picks the concurrency cap off `primaryStore.useVersions`, i.e. a *versioning capability* flag is standing in for an *engine write-batching* probe. They coincide today only because `PrimaryRocksDatabase` happens not to define `useVersions` (probed at this head: `undefined` on RocksDB, `true` on LMDB).

That matters because the same flag gates the version guard at line 5112, which is separately flagged as inert on RocksDB. The natural fix for that finding — giving `PrimaryRocksDatabase` a `useVersions` — would silently raise RocksDB's in-flight removals from 10 to 1000, on the engine that pays a full transaction commit per removal and therefore wants the tight cap. Fixing one finding would quietly regress the other, with no test on the RocksDB side to notice.

Gating on an explicit batching capability (or on the engine) instead keeps the two independent:

```js
const maxConcurrentRemovals = primaryStore.batchesWritesPerEventTurn
    ? MAX_CONCURRENT_LMDB_HISTORY_REMOVALS
    : MAX_CONCURRENT_HISTORY_REMOVALS;
```

Separately, DESIGN.md describes the 10-slot path as "a non-versioned fallback store". RocksDB is the **default** engine, not a fallback — the wording understates which path most deployments actually take.

—
_Generated by Barber AI_

resources/Table.ts
cb1kenobi: **Fixed — confirmed by runtime probe on both engines at `df65e3cd`.** Nice fix, and the shape is right: the condition and the delete stay inside one transaction rather than being split into a live read plus an unconditional remove.

Same probe as before — a table built through `table()` on each engine, stale version passed explicitly:

| | ctor | `useVersions` | `remove(id, STALE)` | record after | `remove(id, CURRENT)` |
|---|---|---|---|---|---|
| LMDB | `LMDBStore` | `true` | `false` | **preserved** | `true` -> deleted |
| RocksDB (default) | `PrimaryRocksDatabase` | `undefined` | **`false`** | **preserved** | `true` -> deleted |

RocksDB now honors it even though `useVersions` is still `undefined`, because `databases.ts` dispatches on the argument type instead.

I also checked the three ways this could have silently degraded back to unconditional. Each mutation was applied to the `.ts`, rebuilt with `tsc -p tsconfig.build.json`, and grep-confirmed as exactly one hit in `dist/` before running:

| mutation | result |
|---|---|
| version check neutered (`entry.version !== version` -> `false`) | **killed** — 24 passing / 1 failing, `a stale conditional removal must not commit` |
| `ERR_BUSY` retry neutered | **killed** — `Error: Transaction commit failed: Resource busy` |
| dispatch neutered (`typeof removeOptions === 'number'` -> `'bigint'`) | **killed** — 1 failing |

The second one is the interesting result: it proves the rocksdb-js transaction really does track its read set and fail the commit on a concurrent write, so the retry is load-bearing rather than defensive decoration. That was the assumption the whole approach rests on, and it holds.

One more thing I checked, since the new `version != null` guard could have quietly disabled the sweep: `deleteHistory(cutoff, true)` still purges tombstones on both engines (5 tombstones scanned, all carrying a version, `tombstonesRemaining=0`). No silent no-op.

Unmutated control: RocksDB 25 passing / 5 pending, LMDB 24 passing / 6 pending.

—
_Generated by Barber AI_

resources/Table.ts
cb1kenobi: **Addressed at `df65e3cd`, and the trap I warned about was avoided.** The cap now keys off `isRocksDB` rather than `useVersions`, so giving RocksDB a working version guard did not drag the concurrency cap along with it. DESIGN.md's "non-versioned fallback store" wording is fixed too.

Measured rather than read — instrumented `auditStore.remove` and ran `deleteHistory` over 1,200 audit entries, parking every removal to observe the plateau:

| engine | removals started before the loop parks | max concurrently active |
|---|---|---|
| RocksDB (default) | **10** | 10 |
| LMDB | **1000** | 1000 |

And pinned by mutation: `MAX_CONCURRENT_HISTORY_REMOVALS` 10 -> 7 (dist-verified) moves the RocksDB plateau to exactly 7, so that constant is what governs the default engine.

No throughput cost on either side, which was the other half of the worry — `removeIfVersion` now pays a read plus a native transaction commit per tombstone on RocksDB, so it was worth measuring. 4000 puts + 4000 deletes, timed `deleteHistory`, three rounds, same machine and `node_modules`, both builds dist-verified:

| build | RocksDB | LMDB |
|---|---|---|
| merge base `b852a722c` | 140.3 / 261.2 / 384.3 ms | 166.0 / 149.9 / 136.0 ms |
| head `df65e3cd` | 141.7 / 255.8 / 379.6 ms | 169.1 / 155.9 / 151.8 ms |

(RocksDB rounds are cumulative — `entries_deleted` 8k/16k/24k — LMDB is 8k each round.) Both engines are within noise of the base.

CI agrees: `txnlog-purge-stale-read-blast.test.ts` arm 5 on Integration Tests 1/6 (Node.js v24) is `[rocksdb]` 2,078 ms and `[lmdb]` 1,562 ms, both passing, against 2,076 / 1,550 at the prior head.

—
_Generated by Barber AI_

resources/Table.ts
```diff
@@ -5050,29 +5052,71 @@ export function makeTable(options) {
...
-					await rest(); // yield to other async operations
-					if (value === null && localTime < endTime) {
-						completion = removeEntry(primaryStore, entry);
+				try {
+					for (const entry of primaryStore.getRange({ start: 0, versions: true })) {
+						const { key, value, localTime, version } = entry;
+						await rest(); // yield to other async operations
+						if (value === null && version != null && localTime < endTime) {
+							const backpressure = queueRemoval(
+								() => primaryStore.remove(key, version),
```
cb1kenobi, line 5124: **Medium: the scan-to-remove version hand-off is untested on RocksDB, the default engine**

The conditional removal itself is now well covered on RocksDB — I could not get a mutation past it. But *this* line, the one that actually carries the scan-captured `version` into the removal, has nothing pinning it on the default engine.

Mutation: drop the second argument, reverting the cleanup removal to unconditional — the exact bug this commit exists to fix. Applied to the `.ts`, rebuilt with `tsc -p tsconfig.build.json`, grep-confirmed as one hit in `dist/resources/Table.js`:

```js
() => primaryStore.remove(key),
```

| engine | `mocha unitTests/resources/auditLog.test.js` |
|---|---|
| RocksDB (default) | **25 passing / 5 pending / 0 failing — survives** |
| LMDB | 23 passing / 6 pending / **1 failing** — killed |

The reason is the split in coverage. `deleteHistory cleanup preserves a record recreated while its stale tombstone waits for a slot` (line 287) is the arm that exercises this line end-to-end, and it is gated on `!auditStore.reusableIterable`, so it skips on RocksDB. The new `RocksDB versioned removal preserves records recreated before or during removal` (line 356) calls `primaryStore.remove()` directly and never enters `deleteHistory`, so it proves the primitive but not the wiring.

Net effect: a future edit that drops or mis-passes this `version` ships green on the engine most deployments run, and only the LMDB arm would have caught it — the same asymmetry as the previous round, mirrored.

Suggested fix: lift the line-287 test's assertion into a form that runs on both engines. It only skips on RocksDB because `createOrphanedTombstone` can't build an orphaned tombstone against a RocksDB transaction log — but the assertion doesn't need an *orphaned* tombstone. A real `delete()` followed by a recreate before the queued removal fires, asserting the record survives `deleteHistory(cutoff, true)`, works on both.

—
_Generated by Barber AI_

resources/Table.ts
kriszyp: Fixed in `57aba32d`. The owner picked the narrow option you suggested first: `deleteHistory` now counts attempted and succeeded removals, and a purge that attempted at least one removal and completed none rejects with the first storage error — after both phases have drained, so a partially-failing store still gets every removal it can accept. A single success reports normally, and partial failures stay best-effort (logged, excluded from the returned count). The response shape is unchanged; `entries_failed` was deliberately not added, since it would expand this fix into a public API/docs contract change across LMDB, RocksDB and the deprecated wrapper — worth doing later only if callers need programmatic reconciliation.

The first error is captured *before* the warn call, so a logger that throws cannot cost us the only reference to why the purge failed. New test: `deleteHistory reports a purge that attempted removals and completed none` — it also asserts that an empty backlog still returns 0 rather than rejecting.

Traced the caller path rather than assuming it: `ResourceBridge.deleteTransactionLogsBefore` awaits `table.deleteHistory`, and `server/jobs/jobRunner.ts:runJob` turns the rejection into job status `ERROR` with the message, so the new failure case surfaces as a failed job rather than an unhandled rejection.

— Claude Opus

unitTests/resources/auditLog.test.js
kriszyp: Fixed in `57aba32d`, exactly as you specified. The regression is now parameterised over `loggingThrows`, so it runs once with the throwing `warn` stub and once with a non-throwing one; the non-throwing pass is what pins `entriesDeleted === 1`. Confirmed your mutation is now killed in both placements — moving `onSuccess?.()` before `harperLogger.warn(...)` fails the non-throwing arm.

— Claude Opus

resources/Table.ts
kriszyp: Fixed in `57aba32d`, via your suggested route. `deleteHistory cleanup preserves a record recreated while its stale tombstone waits for a slot` no longer needs `createOrphanedTombstone`: it does a real `put` + `delete`, then suppresses the audit-driven delete callback for the table — which is the "the audit log is not cleaning these up" state the cleanup phase exists for — so the tombstone survives to the cleanup scan. The `reusableIterable` gate is gone and the test now runs on both engines (26 passing / 6 pending on RocksDB, where it was previously pending).

That puts the scan-captured version hand-off at `resources/Table.ts:5124` under coverage on the default engine, so dropping the second argument is now killed on RocksDB too, not only on LMDB.

— Claude Opus

resources/Table.ts
cb1kenobi: **Fixed at `df6b7987` — `57aba32d` implements the first of the two options suggested here.**

`deleteHistory` now tracks `removalsAttempted` / `removalsSucceeded` and, after both phases have drained, rethrows the first captured error when a purge attempted at least one removal and completed none (`resources/Table.ts:5134`). A `delete_transaction_logs_before` against a store that rejects every write now errors instead of returning a successful `entries_deleted: 0`, so "nothing was eligible" and "nothing worked" are no longer the same response.

Two details worth noting, since both were failure modes of earlier revisions:

- The error is captured *before* `harperLogger.warn` is called, so a throwing logger cannot cost the error that gets rethrown.
- The check sits after both `finally { await drainRemovals() }` blocks, so a failing store still gets every removal it can accept rather than aborting on the first rejection.

Verified by mutation (mutate the `.ts`, `tsc -p tsconfig.build.json`, grep-confirmed as exactly one hit in `dist/resources/Table.js`, then run):

| mutation | LMDB | RocksDB |
|---|---|---|
| `removalsAttempted > 0 && removalsSucceeded === 0` -> `false` | **killed** — 25 passing / 1 failing, `Missing expected rejection` | 26 passing / 0 failing (skipped by design) |

The killer is the new `deleteHistory reports a purge that attempted removals and completed none`. It skips on RocksDB, which matches production reachability rather than leaving a gap: `ResourceBridge.deleteTransactionLogsBefore` routes a `RocksDatabase` primary store to `purgeLogs()` and only reaches `deleteHistory` on the non-Rocks branch.

Partial failures remain best-effort — logged, excluded from the count, not surfaced — which DESIGN.md now states explicitly as the intended trade-off. Recording that as the documented resolution of this thread rather than reopening it; it was always a judgement call for a human reviewer, not a defect.

Unmutated control at this head: 26 passing / 6 pending / 0 failing on both engines.

—
_Generated by Barber AI_

resources/Table.ts
cb1kenobi: **Fixed at `df6b7987` — the mutation that survived on RocksDB is now killed on both engines.**

The fix is the one this thread suggested: `deleteHistory cleanup preserves a record recreated while its stale tombstone waits for a slot` lost its `if (auditStore.reusableIterable) return this.skip()` gate and now runs everywhere. Instead of `createOrphanedTombstone` (which cannot orphan an entry in a RocksDB transaction log), it does a real `put` + `delete` and suppresses the audit-driven delete callback — the "the audit log isn't cleaning these up" state the cleanup phase exists for — so the tombstone survives to the cleanup scan on either engine.

Re-ran the same mutation: drop the second argument at `resources/Table.ts:5124`, reverting the cleanup removal to unconditional. Applied to the `.ts`, rebuilt with `tsc -p tsconfig.build.json`, grep-confirmed as exactly one hit in `dist/resources/Table.js`:

```js
() => primaryStore.remove(key),
```

| engine | at `df65e3cd` | at `df6b7987` |
|---|---|---|
| RocksDB (default) | 25 passing / 0 failing — **survived** | 25 passing / 6 pending / **1 failing — killed** |
| LMDB | 1 failing — killed | 25 passing / 6 pending / **1 failing — killed** |

Failure on both: `AssertionError: the cleanup scan must hand the version it captured to remove()` (`auditLog.test.js:388`). The asymmetry is gone — a future edit that drops or mis-passes this `version` can no longer ship green on the default engine.

I also confirmed the rest of the round held up. `removeIfVersion`'s retry loop was rewritten in `c2414157` from a bounded `for` to `while (true)` plus a `retried` flag; it is still capped at two attempts, and neutering the `ERR_BUSY` retry still fails `RocksDB versioned removal preserves records recreated before or during removal` (25 passing / 1 failing), so the rewrite did not quietly turn the retry into decoration. The removed `throw new Error('Versioned removal exhausted its retries')` was genuinely unreachable in the old shape — the second iteration could only `return` or `throw`.

Unmutated control at this head: 26 passing / 6 pending / 0 failing on both engines.

—
_Generated by Barber AI_

unitTests/resources/auditLog.test.js
cb1kenobi: **Closed at `df6b7987`.** The test is now a two-arm loop over `loggingThrows`, and the non-throwing arm is the one that pins the count — exactly the extra pass this thread asked for.

Re-ran the mutation that previously survived, in its equivalent form against the reworked handler: `removalsSucceeded++; onSuccess?.();` added to the *rejection* handler after `harperLogger.warn`, so a failed removal is counted. Grep-confirmed as one hit in `dist/resources/Table.js`.

| arm | result |
|---|---|
| `failure logging throws` | survives — the injected throw still preempts the count |
| `failure logging succeeds` | **killed** — `AssertionError: only the successful removal should be counted, not the rejected one` |

24 passing / 2 failing on LMDB (the second failure is the new zero-progress test, which the same mutation also defeats). Both placements are pinned now.

—
_Generated by Barber AI_

cb1kenobi:
Reviewed `df6b7987` — no open findings. This PR looks good, nice job!

Re-review of the three commits since `df65e3cd` (`57aba32d`, `c2414157`, `df6b7987`). Merge base is unchanged at `b852a722c`, and `git range-diff` shows commits 1-17 identical, so the delta is purely additive: 3 commits, 5 files.

Both Mediums that were open at the last review are now closed, each confirmed by re-running the mutation that had previously survived (details in the threads):

- **Cleanup-phase version wiring** — dropping the `version` argument at `resources/Table.ts:5124` now fails on **both** engines (RocksDB 1 failing, LMDB 1 failing). At `df65e3cd` it was RocksDB 25 passing / 0 failing. The fix removed the `reusableIterable` gate from the cleanup-race test rather than adding a second RocksDB-only test.
- **Swallowed removal failures** — `deleteHistory` now rethrows when a purge attempted removals and completed none, after both phases drain. Partial failures stay best-effort by documented design.

The previously-fixed items held: the version guard is still honored on both engines, the `ERR_BUSY` retry survives its rewrite from a bounded `for` to `while (true)` and is still mutation-killed, and the concurrency cap still keys off `isRocksDB` rather than `useVersions`.

Test health at this head: `auditLog.test.js` 26 passing / 6 pending / 0 failing on both engines. Full `unitTests/resources` is 1354 passing / 3 failing on RocksDB and 1207 passing / 0 failing on LMDB — the 3 RocksDB failures are `sourceApplyConflictRetry.test.js:146/176/238` and reproduce identically at the merge base, so they are pre-existing and unrelated to this PR.

—
_Generated by Barber AI_

Issue #2057, opened 2026-08-03
DatabaseTransaction has no supported way to extend the open-transaction budget for one known-long write (LMDB gap + fragile workaround)


kriszyp:
## Summary

`DatabaseTransaction`/`LMDBTransaction`'s open-time budget (`.timeout`) has no supported way
for a caller to extend it for one known-long write — a caller has to reach in and mutate
`.timeout` directly, and that mutation is fragile and doesn't propagate to every link that
actually needs it. This was found during cross-model review of
`fix/deploy-ingest-txn-timeout` (harper-large-deploy-txn-abort — the Large Deploy Payload Test
regression), which needed exactly this capability for the deploy payload-ingest write, and
worked around the gap in `components/deploymentRecorder.ts`'s `withExtendedTransactionTimeout`.
Two independent review lenses (Codex, an internal Harper-domain pass) and a third experimental
lens (Grok) all converged on the same root cause independently.

## The gap

1. **LMDB: the extension lands on the wrong object.** `resources/transaction.ts`'s
   `transaction()` returns the `DatabaseTransaction` it creates
   (`resources/transaction.ts:39`). The actual write resolves its transaction through
   `txnForContext` (`resources/Table.ts:5489`), which reuses that head object only when
   `isRocksDB` (`resources/Table.ts:5492`); for an LMDB-backed store it instead allocates a
   **chained `new LMDBTransaction()`** on `.next` (`resources/Table.ts:5505`). The chained
   object has its own `.timeout`, its own `getReadTxn()` (`resources/LMDBTransaction.ts:38`),
   and its own `trackedTxns` set (`resources/LMDBTransaction.ts:16`) — the long-transaction
   monitor never looks at the head's `.timeout`. Concretely: on `HARPER_STORAGE_ENGINE=lmdb`,
   `withExtendedTransactionTimeout`'s 10-minute budget sits on an object nobody aborts, while
   the LMDBTransaction that actually stages the row counts down from the generic default and
   is aborted after ~30-60s — the exact "Transaction was open too long" failure the caller was
   trying to avoid, with the workaround in place and its tests green. The same shape can reach
   RocksDB too, whenever the write lands on a `.next` link because a *different* database was
   already written earlier in the same transaction chain.

2. **The workaround (assign-after-call) is a coincidence, not an invariant.** `getReadTxn()`
   unconditionally does `this.timeout = txnExpiration` (`resources/DatabaseTransaction.ts:362`,
[...]

Issue #2056, opened 2026-08-03
unhandledRejection duplicates the deploy_component 422 when the ingest write transaction times out


kriszyp:
## Summary

When a `deploy_component` write transaction is aborted by the long-transaction monitor
(`transactionOpenTooLongError`, `resources/DatabaseTransaction.ts`), the same rejection that
correctly surfaces as the operation's 422 response *also* escapes as an `unhandledRejection`.
Both of Harper's own `unhandledRejection` listeners (`bin/run.ts` and
`server/threads/threadServer.js`) log it, so it appears twice in the log with two different
formats but is a single underlying event.

## Reproduction

1. `HARPER_TEST_LARGE_DEPLOY_MB=600 npm run test:integration -- integrationTests/deploy/deploy-large-payload.test.ts`
   with `storage.maxTransactionOpenTime` left low enough (or disk slow enough) that the
   payload-ingest write transaction times out mid-stream.
2. The request correctly resolves 422 with `transactionOpenTooLongError`'s message.
3. Immediately after, the log also shows:
   ```
   [error]: Unhandled promise rejection: Promise Promise { <rejected> ServerError: Transaction was aborted... } reason: Error: Transaction was aborted...
   [error]: unhandledRejection Error: Transaction was aborted after exceeding the maximum open-transaction time...
   ```
   with the identical stack trace as the properly-handled 422:
   ```
   at transactionOpenTooLongError (resources/DatabaseTransaction.js:180:12)
   at DatabaseTransaction.addWrite (resources/DatabaseTransaction.js:354:19)
   at proceed (resources/Table.js:2767:36)
   at TableResource._writeUpdate (resources/Table.js:2769:64)
   at TableResource.update (resources/Table.js:1719:45)
   at TableResource.put (resources/Table.js:2008:49)
   at put.transactional.hasContent (resources/Resource.js:145:28)
   at runAction (resources/Resource.js:725:32)
   at resources/Resource.js:785:24
   at when (utility/when.js:10:12)
   ```

## What's confirmed vs. still open

- Confirmed: `components/operations.js`'s `deployComponent` catch block does `await
  recorder.finish('failed', err)` wrapped in its own try/catch (`Failed to record deployment
  failure row`), so *that* specific write's rejection is handled.
- Confirmed: the outer `deployComponent` rejection is caught and logged once by
  `callOperationFunctionAsAwait` (`utility/OperationFunctionCaller.ts`) and correctly produces
  the 422.
[...]

PR #2058, opened 2026-08-03
Give the deploy payload-ingest write its own open-transaction budget

kriszyp:
Human-Review-Need: 4 @ 23a481702dbf
## What / why

The weekly self-hosted Large Deploy Payload Test began failing with `transactionOpenTooLongError` while streaming a 2200 MB component into `deploy_component`.

The `hdb_deployment.put()` commit is deliberately gated on the payload blob reaching durable storage. A peer must not observe a row that references a partial blob. This means the transaction remains open for as long as the payload takes to land on disk, while the generic `storage.maxTransactionOpenTime` guard assumes ordinary short request work.

Historical CI runs completed in 65–92 seconds. A contended runner pushed the same valid write beyond the generic limit. Raising that limit globally or retrying the write would move the cliff without fixing the mismatch.

## The fix

This gives the payload-ingest write a deploy-specific RocksDB timeout floor:

- `DatabaseTransaction.timeoutBudget` survives every `getReadTxn()` refresh and uses `Math.max(txnExpiration, timeoutBudget)`, so it never shortens a larger global limit.
- Chained RocksDB transaction links inherit the budget.
- The default ingest budget is ten minutes. `deployment_timeout` can raise it, while small or zero peer-wait values cannot starve the origin's disk write.
- Every deployment-recorder write commits in a fresh transaction built from only the ambient user, originating operation, session, and abort signal. This preserves audit attribution without joining a caller's open transaction or inheriting timestamp/source-apply controls.
- Progress flushes drain before ingest and remain suppressed until ingest settles. This prevents an independent coalesced put from conflicting with the long-lived payload write on the same deployment row.
- The monitor uses the same timeout floor on its force-commit path, and queued progress-write rejections have an attached handler while remaining visible to `finish()`.

The recorder writes are internal to an already-authorized `deploy_component` operation, so they do not inherit the caller's one-shot `authorize` flag.

## Why isolate all recorder writes

Isolating only the payload write introduced a commit-order hazard when `deploy_component` ran inside an ambient transaction: the ingest row could commit first, then an older staged create row could commit afterward and restore `payload_blob: null`. Making the recorder consistently independent establishes one ordering model and keeps its observability row outside the caller's atomic work.

## Testing

- `npm run build` — passed.
- Focused deployment-recorder and transaction suites — 57 passing, 2 LMDB-only tests pending.
- `npm run test:unit:resources` — 1,363 passing, 15 pending.
- `npm run test:unit:main` — 4,234 passing, 195 pending; two unrelated `applicationSpawn.test.js` fixture failures reproduced unchanged on this host because the expected descendant marker files were never created.
- `HARPER_TEST_LARGE_DEPLOY_MB=256 ... deploy-large-payload.test.ts` — passed end to end. The streamed deploy itself took 43.4 seconds, crossing the prior 30-second failure window.
- The full integration matrix could not produce a meaningful result locally: another runner truncated the shared `/tmp/harper-integration-test-loopback-pool.json`, causing every suite to fail at startup with `Unexpected end of JSON input`. The relevant large-deploy suite passed using an isolated `TMPDIR`.

Regression coverage now verifies:

- user and `originatingOperation` attribution survive transaction isolation;
- ambient transaction controls do not leak into recorder writes;
- every recorder put uses a transaction distinct from the caller's;
- the timeout floor remains sticky across real RocksDB reads and never lowers a larger global limit;
- a write-bearing RocksDB transaction survives multiple monitor ticks beyond the global limit;
- progress puts already in flight are drained, and new progress flushes are deferred until ingest settles.

## Review

Three independent pre-push review rounds covered the full implementation and the follow-up delta using Claude, Gemini, and Harper-domain adjudication. The reviews found the inherited `isExplicit` isolation failure, the one-shot timeout reset, ambient commit-order inversion, same-row progress contention, transaction hidden-class cost, chained-link propagation gap, and monitor reset inconsistency. Those findings are addressed in this branch.

## Scope and tradeoffs

- LMDB remains out of scope. Its chained `LMDBTransaction` has an independent timeout monitor and still needs the engine-level work tracked in [#2057](https://github.com/HarperFast/harper/issues/2057).
- The ten-minute floor deliberately means a stalled ingest can pin its `system` database snapshot for up to that minimum. This is the tradeoff for sharing `deployment_timeout`, where `0` already means “poll once” for peer waits.
- The pre-existing generic transaction-abort duplicate-log path remains tracked in [#2056](https://github.com/HarperFast/harper/issues/2056).

— GPT-5 Codex


kriszyp:
## Guided tour

Start with `DESIGN.md`'s “Extending the budget for one known-long write” section. It records the transaction invariant behind this PR.

1. `resources/DatabaseTransaction.ts` — `timeoutBudget` is a real initialized field, and `getReadTxn()` refreshes the effective timeout with `Math.max(txnExpiration, timeoutBudget)`. The monitor applies the same floor. This replaces the earlier call-order-dependent assignment.
2. `resources/Table.ts` — a chained RocksDB transaction inherits `timeoutBudget`, alongside the other logical-transaction controls.
3. `components/deploymentRecorder.ts` — `withIsolatedTransaction()` builds a fresh context from the caller's user, originating operation, session, and abort signal. It intentionally excludes transaction, timestamp, source-apply, replication, and one-shot authorization controls.
4. Every recorder `put()` uses that isolation. This matters beyond the payload write: mixing an isolated ingest with an ambient create write allowed the older create revision to commit last and restore `payload_blob: null`.
5. `ingestPayload()` drains any existing progress put, then suppresses coalesced flushes until the blob-gated write settles. Without that serialization, two independent transactions could contend on the same deployment row for the duration of a large upload.
6. `ingestTransactionTimeoutMs()` keeps the ten-minute floor because the shared `deployment_timeout: 0` value already means “poll once” for peer waits.

## Tests worth reading

- `unitTests/resources/txn-tracking.test.js` exercises real RocksDB transactions. It verifies stickiness across reads, preserves a larger global limit, and holds a write across several monitor ticks to prove it is not aborted.
- `unitTests/components/deploymentRecorder.test.js` verifies context attribution/isolation and gates mock writes to prove both an existing progress put and a mid-ingest event are serialized around payload ingest.
- The 256 MB opt-in large-deploy integration passed locally; the deploy itself took 43.4 seconds, beyond the old 30-second failure window.

## Remaining boundary

LMDB is unchanged and remains tracked by [#2057](https://github.com/HarperFast/harper/issues/2057). Its separate `LMDBTransaction` monitor does not consume this RocksDB budget.

— GPT-5 Codex

cb1kenobi (approved):
Looks good!

PR #2064, opened 2026-08-03
`delete_transaction_logs_before` with a table on RocksDB now returns an error instead of deleting the entire database's transaction log

### LINKED ISSUE #2049: RocksDB: table-scoped delete_transaction_logs_before destroys the entire database's transaction log
# RocksDB: table-scoped `delete_transaction_logs_before` destroys the entire database's transaction log

## Summary

On RocksDB, `delete_transaction_logs_before({ database, table, timestamp })` silently ignores the `table` scope and purges the **whole database's** transaction/audit log. An operator asking to prune one table's history permanently destroys every sibling table's history in the same database. The operation reports success with an `entries_deleted` count reflecting the wider deletion, and there is no warning and no way to recover.

The parameter is annotated `// lmdb only` in the signature, so this is known-not-supported on RocksDB — but the code path doesn't *reject* it, it accepts it and does something far more destructive than what was asked.

## Mechanism

`dataLayer/harperBridge/ResourceBridge.ts:481-519`:

```ts
async deleteTransactionLogsBefore(deleteObj: {
    schema?: string;
    database?: string;
    table?: string; // lmdb only
    timestamp: Date | number | string;
    cleanup_deleted_records?: boolean; // lmdb only
}): Promise<DeleteTransactionLogsBeforeResults> {
    ...
    const table = getTable(deleteObj);
    if (!table) {
        // no table, check if any of the tables are RocksDB
        // since all tables share the same transaction log store, we break after the first
        // RocksDB table is found
        const tables = getDatabases()[deleteObj.database];
        if (tables) {
            for (const table of Object.values(tables)) {
                if (table.primaryStore instanceof RocksDatabase) {
                    const deleted = table.primaryStore.purgeLogs({ before, includeEntryCounts: true });
                    ...
                    break;
                }
            }
        }
    } else if (table.primaryStore instanceof RocksDatabase) {
        const deleted = table.primaryStore.purgeLogs({ before, includeEntryCounts: true });
        ...
    }
    ...
}
```

The two RocksDB branches — "caller named a table" and "caller named no table" — do **the identical thing**. The comment in the second branch states why: *all tables share the same transaction log store*. So naming a table only selects which table's `primaryStore` handle is used to reach the shared per-database log store; the purge itself has no table dimension.

**Per-table purging isn't merely unimplemented — it isn't expressible at the storage layer.** `rocksdb-js` 2.5.0's `PurgeLogsOptions` is:

```ts
type PurgeLogsOptions = {
  before?: number;
  destroy?: boolean;
  includeEntryCounts?: boolean;
  name?: string;   // log name, e.g. "local" — not a table
};
```

There is no table/tableId filter, and the log is purged in whole-file units, which is fundamentally at odds with per-table granularity — a per-table purge would require rewriting log files rather than deleting them.

## Measured

Purging one 5,000-row table in a database with three 5,000-row tables:

- reported `entries_deleted: 15000` (i.e. all three tables' entries)
- after restart, the two sibling tables read **0** rows of history each (5,000 before)

Manager-verified by source read **and** an independent cold rerun (6/6 runs, identical numbers).

A detail that makes this worse in practice: a stale in-process cache keeps the sibling data *appearing* intact until restart, so an operator's immediate "did that do what I wanted?" check sees healthy tables. The loss surfaces on the next restart, decoupled in time from the action that caused it.

## Why this is worse than an unsupported-parameter bug

The failure mode is unrecoverable destruction of data the operator never asked to delete, on an operation whose whole purpose is bounded deletion. Ignoring an unsupported scope parameter is normally benign; here it widens the blast radius of a destructive operation by the number of tables in the database.

## Suggested direction

Short term, and I'd argue this is the whole fix: **reject `table` on RocksDB with a 400** rather than silently widening the scope. The signature already documents it as LMDB-only; make the validator enforce what the comment claims. `validation/transactionLogValidator.ts` is the natural place, though it needs the storage engine in scope to make the decision.

If per-table pruning on RocksDB is wanted as a feature, that's a `rocksdb-js` change first (a table filter in `PurgeLogsOptions`, plus a story for whole-file purge granularity) and should be a separate issue — but the silent-widening behaviour should not wait on it.

Worth deciding alongside: whether `cleanup_deleted_records` (also annotated `// lmdb only`) has the same silent-ignore shape. It looks inert rather than destructive on the RocksDB path, but it's the same pattern in the same signature.

## Provenance

Found by the qa-explorer exploratory QA loop. Source lines verified against `main` @ `b852a722c`; `PurgeLogsOptions` verified against the pinned `@harperfast/rocksdb-js` 2.5.0.

cb1kenobi:
Fixes #2049 ([RocksDB: table-scoped delete_transaction_logs_before destroys the entire database's transaction log](https://github.com/HarperFast/harper/issues/2049)).

## What changed

On RocksDB, all tables in a database share one transaction log, and rocksdb-js's `purgeLogs()` has no table dimension (whole-file purges, logs partitioned per origin node). The bridge previously *ignored* a `table` scope and purged the whole database's log — reporting success. Now, in `ResourceBridge.deleteTransactionLogsBefore`:

- `table` naming a RocksDB table → **400**, with a message pointing at the supported database-wide form (`delete_transaction_logs_before` with only `database` and `timestamp`).
- `table` naming a **nonexistent** table → **404** on either engine. The guard sits before the engine branch, so this is a behavior change on LMDB too: previously a typo'd name fell into the no-table branch — the whole-database purge on RocksDB, a silent `entries_deleted: 0` no-op on LMDB.
- The table checks use **presence (`!= null`), not truthiness**, in both the guard and `getTable()`, so a table named `"0"` addressed numerically stays table-scoped instead of widening to a database purge (found by cross-model review).
- No `table` → database-wide purge, unchanged. LMDB table-scoped `deleteHistory` → unchanged.

## Deliberate design consequence — reviewer attention here

The deprecated `delete_audit_logs_before` op **requires** `table` and funnels into the same bridge method, so on RocksDB it now **always errors** with the migration message. There is no way to honor its per-table contract on RocksDB without the silent widening this PR removes. Three integration tests that previously asserted job success for table-scoped calls were updated to assert the rejection (`transaction-logs.test.mjs`, `transactions.test.mjs`, `terminology.test.mjs` — the last now verifies database-param resolution via the database name echoed in the error).

Not addressed (same silent-ignore shape, but inert rather than destructive): `cleanup_deleted_records` is still silently ignored on the RocksDB no-table path — flagged in #2049 as a decide-alongside; can be a follow-up.

## Testing

- New unit tests (`unitTests/resources/deleteTransactionLogsBeforeRocks.test.js`, 5 passing): 400 on table-scoped, 404 on missing table/database, falsy-table-name regression, database-wide purge still succeeds. Skipped under `HARPER_STORAGE_ENGINE=lmdb`.
- `test:unit:resources` full run green (1365 passing); the three touched integration files green (75 passing).
- `test:integration:all` full run: 1624/1645 passing; the 3 failures (`shutdown-drain-e2e.test.ts`, `rolling-restart.test.ts`) reproduce identically on `origin/main` — pre-existing local-environment failures, unrelated to this change.
- Local `test:unit:main` hangs in `unitTests/components/applicationSpawn.test.js` on this machine **with and without this change** (verified A/B against main's `ResourceBridge.ts`) — known local-environment issue; relying on CI for that suite.

## Docs

Companion docs PR: [HarperFast/documentation#618](https://github.com/HarperFast/documentation/pull/618) — documents database-wide-only deletion on RocksDB and the deprecated op's behavior.

## Review

Cross-model review (thorough): author Claude (Fable 5); coverage: Codex ✓ (two findings — the falsy-table-name blocker and the dead-end error remediation — both fixed, then a final-artifact re-check ✓); Gemini ✗ (agy failing locally with a model/location config error; no Gemini coverage).

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -479,6 +479,20 @@ corrupt by verify). This closes the "healthy-looking but incomplete" and concurr
 the remaining engine/blob point-in-time skew (a blob unlinked between the engine cut and the blob
 walk) is the documented best-effort limitation above.
 
+## RocksDB transaction log purges are database-wide only (`ResourceBridge.deleteTransactionLogsBefore`)
+
+On RocksDB, every table in a database writes to one shared set of transaction logs (partitioned per
+origin node, not per table), and `purgeLogs()` deletes whole log files — rocksdb-js has no table
+filter, and adding one would mean rewriting files instead of deleting them. So a table-scoped
+`delete_transaction_logs_before` is unimplementable at the storage layer; the bridge rejects
+`table` on RocksDB with a 400 rather than silently purging every sibling table's history
+(harper#2049 — the original code did exactly that, and a _typo'd_ table name did too, because a
+missing table fell through to the no-table branch; that now 404s). Two consequences to preserve:
+the deprecated `delete_audit_logs_before` op _requires_ `table`, so it always errors on RocksDB
+(the message steers callers to the new op without `table`); and the table/no-table checks in the
+bridge use `!= null` presence, not truthiness, so a table named `"0"` addressed numerically stays
+table-scoped instead of widening to a database purge.
+
 ## Scheduler: cluster-once execution without a consensus primitive (`resources/scheduler/`)
 
 The built-in `scheduler` plugin (#951) runs config-declared jobs "exactly once per cluster." The
--- a/dataLayer/harperBridge/ResourceBridge.ts
+++ b/dataLayer/harperBridge/ResourceBridge.ts
@@ -492,12 +492,25 @@ export class ResourceBridge extends BridgeMethods {
 				: typeof deleteObj.timestamp === 'string'
 					? Number.parseInt(deleteObj.timestamp)
 					: deleteObj.timestamp;
+		const databaseName = deleteObj.database || deleteObj.schema || DEFAULT_DATABASE;
 		const table = getTable(deleteObj);
+		// A nonexistent table must not fall through to the no-table branch below — on RocksDB that
+		// widens a table-scoped request (e.g. a typo) into a whole-database log purge (#2049).
+		// Presence check, not truthiness: a table named "0" addressed numerically is still table-scoped.
+		if (deleteObj.table != null && !table)
+			throw handleHDBError(
+				new Error(),
+				HDB_ERROR_MSGS.TABLE_NOT_FOUND(databaseName, deleteObj.table),
+				404,
+				undefined,
+				undefined,
+				true
+			);
 		if (!table) {
 			// no table, check if any of the tables are RocksDB
 			// since all tables share the same transaction log store, we break after the first
 			// RocksDB table is found
-			const tables = getDatabases()[deleteObj.database];
+			const tables = getDatabases()[databaseName];
 			if (tables) {
 				for (const table of Object.values(tables)) {
 					if (table.primaryStore instanceof RocksDatabase) {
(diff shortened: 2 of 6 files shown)


kriszyp (approved):
Great!

Issue #766, opened 2026-05-25
Test


kriszyp:
test

kriszyp:
Test issue created in error; cleaning up.

Issue #1123, opened 2026-06-03
Caching table: 304 revalidation mutates the frozen cached record


heskew:
# Caching table: 304 revalidation mutates the frozen cached record

## Summary

On a caching table (`sourcedFrom(...)`), when the source returns a **304 / not‑modified** revalidation, `getFromSource` aliases the working record to the existing **frozen** cached record (`resources/Table.ts:4296`, `updatedRecord = existingRecord`) and then re-enters the normal store/mutation path. Because loaded records are frozen by default, the subsequent in‑place mutations throw `TypeError` ("Cannot assign to read only property …" / "object is not extensible").

This is **pre-existing and not @embed-specific** — it reproduces on a plain caching table with no `@embed`. It surfaced while testing the `@embed` PR (#747, issue #632), but the `@embed` write-time hook is only one of several mutation sites on this path.

## Mechanism

- Loaded record values are frozen **unconditionally** (`resources/Table.ts:3887`, `Object.freeze(entry?.value)`). This became unconditional in `745174b5` ("freeze records by default", 2026-02-06) — previously gated on `context?._freezeRecords`. So this has been latent on `main` since well before #747.
- The `get` path (`Table.ts:596`) loads via `loadLocalRecord`, which freezes `entry.value`; that frozen value flows into `getFromSource` as `existingRecord`.
- On 304 (`Table.ts:4294`), `updatedRecord = existingRecord` aliases the frozen object, then control falls through to the same normalization + commit path used for a fresh fill, which mutates `updatedRecord` in place.

## Severity is split — most realistic cases are a *logged background failure*, not a request crash

The crash site determines whether the GET hard-fails or the failure is swallowed, because of where `resolved = true` (`Table.ts:4338`) sits relative to the mutation:

- The caller's promise resolves at `Table.ts:4358` (`resolve(resolvedEntry)`), with `resolved` set true at `:4338`.
- The only `reject(error)` path is inside the `try`/`catch` that **ends before** `resolved=true` (`catch` at `:4359`).
- Mutations in the commit closure and embed hook run **after** the caller already resolved. Their throws route through `transaction()`'s rejection → the `when(...)` error handler at `:4528`, which does `if (resolved) logger.error?.('Error committing cache update', error)` — **logged, not propagated to the caller.**

### Mutation-site inventory (the 304 path)

| Site | Line | Fires when | Severity |
|---|---|---|---|
[...]

Issue #414, opened 2026-04-28
Re-enable LMDB caching tests


kriszyp:
They are currently crashing with a segmentation fault. I am not unable to reproduce outside of CI.

kriszyp:
Three concrete reproductions from the week of 2026-09-08, in case they help turn this from "cannot reproduce outside CI" into something with a thread to pull. All three were re-fetched on 2026-09-16 and the logs were still retrievable.

| job | branch @ commit | Node | signature |
|---|---|---|---|
| `104087005044` | `fix/integration-flake-triage` @ `21ad982d5` | 22 | `Segmentation fault (core dumped)`, exit 139, no mocha failing block |
| `104198556436` | `deps/s3-optional-peer` @ `1f524205f` | 24 | identical signature |
| `104232294559` | `deps/s3-optional-peer` @ `dbea34171` | 22 | `The action 'Unit tests: lmdb' has timed out after 4 minutes` — a hang, not a crash |

Both segfaults land in a 9–20 second gap right *after* the last test completes, which points at env-close / final-commit rather than any individual test body. No core dump or backtrace was captured, so that is a plausible location, not a proven one.

If it is the teardown path, the freelist/commit family in #685 and #686 is the obvious suspect — `mdb_page_alloc` reaching `MDB_NOTFOUND` via the fork-local `me_freelist_start` cache, and `rejectCommit`'s unhandled rejection. I want to be clear that I have not established that link; it is a hypothesis worth one targeted run with a core dump enabled, not a conclusion.

Two constraints on any archaeology here, worth writing down because they are not obvious:

- `Unit tests: lmdb` only became a distinct step on 2026-09-14 (#2588, `a17a3eeaa7`, which split `unit-test.yml` one step per suite). Any sweep that filters on that step name is structurally blind before that date.
- This is **not** the same failure as #1380, which is a v24-only SIGSEGV in the schema-change / index-rebuild step and self-describes as distinct.

Severity framing: LMDB is deprecated and RocksDB is the supported engine, so this is CI-maintenance cost rather than product risk — which is probably the right reason to keep it at whatever tier it has rather than escalating it.

Surfaced by finding-triage from a main-red investigation.

PR #2065, opened 2026-08-04
Fix sourcedFrom cache-fill conflict convergence

kriszyp:
> [!NOTE]
> The [rocksdb-js#766](https://github.com/HarperFast/rocksdb-js/pull/766) gate is now closed: `@harperfast/rocksdb-js` 2.8.0 is released and the floor + lockfile are raised to it, so the `VERSION_NOT_UNIQUE_FLAG` contract test runs instead of skipping.

## Problem

Two nodes can independently resolve the same missing `sourcedFrom` key. The old commit path compared against the entry captured before the source fetch, so a write that landed during the fetch was not resolved against the actual record being replaced. The first version of this PR fixed convergence by minting an incremented cache-record timestamp, but that conflated three different things:

- the source's authoritative `lastModified` version;
- the local token used to order competing first fills;
- the timestamp of the transaction in which `source.get()` runs.

In particular, installing a fetch-start token on `sourceContext.timestamp` backdated any write the source performed during resolution.

## Change

1. Reserve a monotonic ordering token before the fetch. An inherited request/transaction timestamp is used when present; otherwise the storage engine supplies one.
2. Do **not** install that token on `sourceContext.timestamp`. The source-resolution transaction, including source-side writes, retains its normal transaction timestamp — [asserted end-to-end now](https://github.com/HarperFast/harper/pull/2065/changes?w=1#diff-3e4675948d111966e41fc20a24a583b464f51390e2ca2fc4420e97959164200cR797), not just by checking that the field is unset.
3. After the source responds, a valid `sourceContext.lastModified` is the cache record's candidate version, [capped at `max(reserved token, Date.now())`](https://github.com/HarperFast/harper/pull/2065/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R5864). A record's version doubles as this node's ordering token (`precedesExistingVersion()` compares a write's transaction timestamp against it), so an uncapped future-dated source established a version floor that silently discarded **every** later local write to that row until wall-clock caught up. Capping is [warned once per table](https://github.com/HarperFast/harper/pull/2065/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R394) and traced per occurrence.
4. Reload the current entry inside the commit transaction. Revalidations keep exact-CAS behavior; a source miss never deletes a raced record; a first fill replaces a raced record only when [its candidate version is strictly greater](https://github.com/HarperFast/harper/pull/2065/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R6012). That comparison is deliberately not `precedesExistingVersion()`: that resolves a tie with the **executing** node's name, and a fill from a shared source has no node identity of its own, so two replicas could keep different values at the same version. A discarded fill is now [traced](https://github.com/HarperFast/harper/pull/2065/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R6019) instead of vanishing silently.
5. On RocksDB, if the source candidate cannot advance the current record version, store at the current version and set `VERSION_NOT_UNIQUE_FLAG` rather than fabricating an epsilon timestamp. #766 makes that safe by refusing to publish or confirm the reused version through the VerificationTable.

The producer flag is wired centrally in `recordUpdater`, so it also covers the ordinary resequenced partial-write case that motivated #766. LMDB continues to store the source candidate directly (below the local-time cap) because it retains separate source-version and local-time semantics.

The [`VERSION_NOT_UNIQUE_FLAG` regression](https://github.com/HarperFast/harper/pull/2065/changes?w=1#diff-e074583cf8528e1c18406545c4fb05e500524e1ab4f74b7a9b0b5413c43aa8daR137) now skips only while the installed rocksdb-js lacks the export, and asserts the two sides agree once it is present — a value drift fails instead of silently deleting the only test guarding the behavior.

## For the human reviewer

- **Open: an equal-version tie is still order-dependent when the fill commits first.** The guard above makes a fill lose a tie against a record that raced it. The mirror case is not covered: if this node's fill commits at version V and a peer write W@V arrives afterwards, the ordinary write path tie-breaks W against the fill's record, which carries *this* node's identity — so a replica that saw W first keeps W while a replica that filled first may keep its fill. Closing it needs a persisted, replica-independent identity (or ordering class) for source fills, which is a record-encoding and replication change beyond this PR. Practically the reachable tie is a second replica's fill of the same key from the same source, which is content-identical; a *divergent* tie needs a third-party write at exactly the source's reported `lastModified`. Flagged by both the pre-push Codex leg and Barber AI; I did not want to invent a new persisted identity under review feedback.
- **Open: the source's `UNCACHEABLE_TIMESTAMP` sentinel no longer survives a fill.** `resources/Table.ts:3369` sets `context.lastModified = Infinity` when a source resource's `get()` runs a dynamic `search()`. On `main` that `Infinity` became the resolved entry's version and suppressed the ETag/`Last-Modified` at the `isFinite` guards in `server/REST.ts:285-317`; here it fails the new `Number.isFinite` validation and is replaced by the finite fill timestamp, so the response carries a validator for content the source declared uncacheable (and a later `If-None-Match` can get a 304). It also flips `hasChanges` for that case, since `Infinity > existingVersion` no longer holds. Preserving the sentinel on the *returned* entry while keeping the *stored* version finite is a few lines, but it is a deliberate semantic call on top of the version model this PR introduces, so it is flagged rather than changed under review. Surfaced by the Harper domain leg of the pre-push review.
- **Open: the reserved ordering token skips the validation the source-reported version now gets.** `inheritedTimestamp` (`Table.get(id, { timestamp })`, or a peer transaction's timestamp) becomes `versionCeiling = Math.max(sourceTimestamp, Date.now())` unvalidated, so a caller-supplied future or non-millisecond timestamp reinstates exactly the frozen-row failure the cap exists to prevent, and a nanosecond-epoch value makes the trace log's `new Date(recordVersion).toISOString()` throw inside the commit callback. Same domain leg; worth deciding whether callers are trusted here.
- **Open: on LMDB a fill can now regress a record's version.** The `Math.max(sourceVersion, existingVersion)` clamp is `isRocksDB`-gated by design, so a revalidation whose candidate predates the cached record writes it at a lower version; a peer holding the pre-invalidation version then compares as strictly newer and anti-entropy can overwrite the fresh source data. That is the engine split `DESIGN.md` commits to, and the suite skips under LMDB (harper#414), so it is the untested half.
- **A same-version revalidation replacement writes no audit entry.** `hasChanges` is false when the source returns content but reports a `lastModified` at or below the existing version. That is the pre-existing cache-fill contract (fills are node-local; making every 200-response revalidation audit would add an audit entry per revalidation for any source that reports no `lastModified`), and on RocksDB the same-version write now carries `VERSION_NOT_UNIQUE_FLAG`, so the version is not vouchable. Worth a second opinion if you disagree that fills are node-local.
- **The version cap replaces a previously-stated tradeoff.** An earlier revision of this description listed "a badly future-dated source can establish a high version floor" as deliberate. It was demonstrably losing subsequent local writes, so it is now capped; the source value is still honored for anything at or below local time.
- **No LMDB coverage.** `caching.test.js` returns early under `HARPER_STORAGE_ENGINE=lmdb` (harper#414), so all the conflict tests are RocksDB-only. Re-enabling the suite there is its own task.

## Verification

- `npm run build`, `npm run format:check`, `npm run lint:required` — clean
- `unitTests/resources/caching.test.js` + `caching-rocks-database.test.js`: 43 passing, 1 pending (the #766-gated VT regression)
- `npm run test:unit:resources`: 1,753 passing, 22 pending, 8 failing locally — the same 8 fail with the msgpackr pin at both 2.0.5 and 2.0.6, so they are independent of this branch's dependency change; CI's `test:unit:all` is green on the same commits
- Fails-on-base check by mutation, rebuilding `dist/` each time: removing the version cap fails [`caps a source-reported version ahead of local time…`](https://github.com/HarperFast/harper/pull/2065/changes?w=1#diff-3e4675948d111966e41fc20a24a583b464f51390e2ca2fc4420e97959164200cR700); restoring `precedesExistingVersion()` in the fill guard fails [`keeps a raced record when a first fill ties its version`](https://github.com/HarperFast/harper/pull/2065/changes?w=1#diff-3e4675948d111966e41fc20a24a583b464f51390e2ca2fc4420e97959164200cR723); installing the token on `sourceContext.timestamp` fails [`does not backdate a write the source performs while resolving`](https://github.com/HarperFast/harper/pull/2065/changes?w=1#diff-3e4675948d111966e41fc20a24a583b464f51390e2ca2fc4420e97959164200cR797)
- New [order-independence test](https://github.com/HarperFast/harper/pull/2065/changes?w=1#diff-3e4675948d111966e41fc20a24a583b464f51390e2ca2fc4420e97959164200cR748) applies the same (fill version, write version) pair in both orderings across both version relations and asserts an identical final `(version, value)`
- End-to-end route: **not observable end-to-end** from a single node — the remaining property is multi-replica convergence, which needs a cluster harness; the unit tests drive the deterministic interleavings a single node can produce
- CI fix in this round: the smoke job's `build-tools/check-shrinkwrap-pins.mjs` failed with `the root msgpackr pin 2.0.5 does not match rocksdb-js 2.0.6` — raising rocksdb-js to 2.8.0 moved its `msgpackr` dependency and left the root pin behind, so npm installed a nested copy and broke the single-module-instance invariant. The [root pin is now 2.0.6](https://github.com/HarperFast/harper/pull/2065/changes?w=1#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R226) and the lockfile's nested `@harperfast/rocksdb-js/node_modules/msgpackr` entry is gone; `unitTests/build-tools/checkShrinkwrapPins.test.mjs` passes 15/15.
- The `Integration Tests 2/6 (Windows, Node.js v24)` failure is not from this branch: `describe-metadata-upgrade.test.ts` "install component and capture first describe_all" fails identically on `main` (runs 32905597531, 32902038519) with a 120s ECONNREFUSED on the `restart_service` readiness probe — harper#549.
- Independent pre-push review: rounds 1–2 Codex (graded) — 3 findings, 1 fixed, 2 open above. Round 3 (full, after the pin change): Codex + Gemini + Cursor Grok + Harper domain — adjudicated severity **minor**, no blockers; the two Codex majors were downgraded (both pre-existing orderings unchanged by this diff), four Gemini majors dropped in adjudication, and the three domain minors are listed above as open items.

Refs [HarperFast/harper-pro#645](https://github.com/HarperFast/harper-pro/pull/645)

<sub>Review-Coverage: authored=claude; ran=gemini,cursor-grok,codex; adjudicated=domain; declined=cursor-composer; rounds=3 @ d15f89389654</sub>

<sub>Human-Review-Need: 4 (decisions: fill-ordering-token, tie-goes-to-raced-record, rocks-clamp-and-flag, flag-scope-all-resequenced-writes, lmdb-keeps-source-version) @ d15f89389654</sub>

### DIFF
--- a/resources/RecordEncoder.ts
+++ b/resources/RecordEncoder.ts
@@ -804,7 +804,7 @@ export function recordUpdater(store, tableId, auditStore) {
 					: TIMESTAMP_ASSIGN_NEW | 0x4000 // or just assign a new one
 				: NO_TIMESTAMP;
 		const expiresAt = options?.expiresAt;
-		if (expiresAt >= 0) assignMetadata = Math.max(assignMetadata, 0) | HAS_EXPIRATION;
+		if (expiresAt >= 0) assignMetadata |= HAS_EXPIRATION;
 		if (isRocksDB && record !== undefined && existingEntry?.version != null && newVersion <= existingEntry.version) {
 			assignMetadata = Math.max(assignMetadata, 0) | VERSION_NOT_UNIQUE_FLAG;
 		}
--- a/resources/RecordEncoder.ts
+++ b/resources/RecordEncoder.ts
@@ -122,6 +122,8 @@ export const HAS_NODE_ID = 64;
 export const PENDING_LOCAL_TIME = 1;
 export const HAS_STRUCTURE_UPDATE = 0x100;
 export const HAS_ADDITIONAL_AUDIT_REFS = 0x80;
+// Cross-package record-header contract with rocksdb-js VERSION_NOT_UNIQUE_FLAG.
+export const VERSION_NOT_UNIQUE_FLAG = 0x10000;
 
 const TRACKED_WRITE_TYPES = new Set(['put', 'patch', 'delete', 'message', 'publish']);
 // For now we use this as the private property mechanism for mapping records to entries.
@@ -806,6 +808,9 @@ export function recordUpdater(store, tableId, auditStore) {
 		if (expiresAt >= 0) assignMetadata |= HAS_EXPIRATION;
 		metadataInNextEncoding = assignMetadata;
 		expiresAtNextEncoding = expiresAt;
+		if (isRocksDB && record !== undefined && existingEntry?.version != null && newVersion <= existingEntry.version) {
+			metadataInNextEncoding = Math.max(metadataInNextEncoding, 0) | VERSION_NOT_UNIQUE_FLAG;
+		}
 		const putOptions: {
 			version: number;
 			instructedWrite?: boolean;
--- a/resources/RecordEncoder.ts
+++ b/resources/RecordEncoder.ts
@@ -122,7 +122,6 @@ export const HAS_NODE_ID = 64;
 export const PENDING_LOCAL_TIME = 1;
 export const HAS_STRUCTURE_UPDATE = 0x100;
 export const HAS_ADDITIONAL_AUDIT_REFS = 0x80;
-// Cross-package record-header contract with rocksdb-js VERSION_NOT_UNIQUE_FLAG.
 export const VERSION_NOT_UNIQUE_FLAG = 0x10000;
 
 const TRACKED_WRITE_TYPES = new Set(['put', 'patch', 'delete', 'message', 'publish']);
@@ -805,12 +804,12 @@ export function recordUpdater(store, tableId, auditStore) {
 					: TIMESTAMP_ASSIGN_NEW | 0x4000 // or just assign a new one
 				: NO_TIMESTAMP;
 		const expiresAt = options?.expiresAt;
-		if (expiresAt >= 0) assignMetadata |= HAS_EXPIRATION;
+		if (expiresAt >= 0) assignMetadata = Math.max(assignMetadata, 0) | HAS_EXPIRATION;
+		if (isRocksDB && record !== undefined && existingEntry?.version != null && newVersion <= existingEntry.version) {
+			assignMetadata = Math.max(assignMetadata, 0) | VERSION_NOT_UNIQUE_FLAG;
+		}
 		metadataInNextEncoding = assignMetadata;
 		expiresAtNextEncoding = expiresAt;
-		if (isRocksDB && record !== undefined && existingEntry?.version != null && newVersion <= existingEntry.version) {
-			metadataInNextEncoding = Math.max(metadataInNextEncoding, 0) | VERSION_NOT_UNIQUE_FLAG;
-		}
 		const putOptions: {
 			version: number;
 			instructedWrite?: boolean;
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -6206,12 +6206,17 @@ export function makeTable(options) {
 					nodeName: 'source',
 					commit: (txnTime, existingEntry, _retry, transaction: any) => {
 						sourceWrite.skipped = false; // reset on each retry; cleanup happens after commit if still true
-						if (existingEntry?.version !== existingVersion) {
-							// don't do anything if the version has changed
+						if (
+							existingEntry?.version !== existingVersion &&
+							(existingVersion != null || precedesExistingVersion(txnTime, existingEntry) <= 0)
+						) {
+							// Revalidations retain exact-CAS semantics. Competing fills of a missing key use the
+							// table's deterministic ordering, so every node keeps the same winner.
 							sourceWrite.skipped = true;
 							return;
 						}
-						updateIndices(id, existingRecord, updatedRecord, transaction && { transaction });
+						const currentRecord = existingEntry?.value;
+						updateIndices(id, currentRecord, updatedRecord, transaction && { transaction });
 						if (updatedRecord) {
 							if (existingEntry) {
 								context.previousResidency = TableResource.getResidencyRecord(existingEntry.residencyId);
@@ -6228,7 +6233,7 @@ export function makeTable(options) {
 											: txnTime;
 							}
 							if (createdTimeProperty && updatedRecord[createdTimeProperty.name] == null) {
-								const existingCreatedTime = existingEntry?.value?.[createdTimeProperty.name];
+								const existingCreatedTime = currentRecord?.[createdTimeProperty.name];
 								if (existingCreatedTime != null) {
 									updatedRecord[createdTimeProperty.name] = existingCreatedTime;
 								} else {
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -6039,6 +6039,13 @@ export function makeTable(options) {
 		setLoadedFromSource(target, true);
 
 		const existingRecord = existingEntry?.value;
+		const inheritedTimestamp = context?.timestamp || context?.transaction?.timestamp;
+		const monotonicTimestamp = () =>
+			isRocksDB ? (primaryStore as RocksDatabase).getMonotonicTimestamp() : getNextMonotonicTime();
+		const sourceTimestamp =
+			inheritedTimestamp && (existingVersion == null || inheritedTimestamp > existingVersion)
+				? inheritedTimestamp
+				: Math.max(monotonicTimestamp(), existingVersion == null ? 0 : existingVersion + 0.000488);
 		// it is important to remember that this is _NOT_ part of the current transaction; nothing is changing
 		// with the canonical data, we are simply fulfilling our local copy of the canonical data, but still don't
 		// want a timestamp later than the current transaction
@@ -6056,6 +6063,7 @@ export function makeTable(options) {
 			noCacheStore: droppingTable,
 			source: null,
 			transaction: undefined,
+			timestamp: sourceTimestamp,
 			expiresAt: undefined,
 			lastModified: undefined,
 		};
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -6002,17 +6002,9 @@ export function makeTable(options) {
 		const existingVersion = existingEntry?.version;
 		const existingRecord = existingEntry?.value;
 		const inheritedTimestamp = context?.timestamp || context?.transaction?.timestamp;
-		const monotonicTimestamp = isRocksDB
-			? (primaryStore as RocksDatabase).getMonotonicTimestamp()
-			: getNextMonotonicTime();
-		const nextExistingVersion =
-			existingVersion == null
-				? 0
-				: existingVersion + Math.max(Math.abs(existingVersion) * Number.EPSILON, Number.MIN_VALUE);
 		const sourceTimestamp =
-			inheritedTimestamp && (existingVersion == null || inheritedTimestamp > existingVersion)
-				? inheritedTimestamp
-				: Math.max(monotonicTimestamp, nextExistingVersion);
+			inheritedTimestamp ||
+			(isRocksDB ? (primaryStore as RocksDatabase).getMonotonicTimestamp() : getNextMonotonicTime());
 		let whenResolved, timer;
 		// We start by locking the record so that there is only one resolution happening at once;
 		// if there is already a resolution in process, we want to use the results of that resolution
@@ -6052,8 +6044,7 @@ export function makeTable(options) {
 		setLoadedFromSource(target, true);
 
 		// it is important to remember that this is _NOT_ part of the current transaction; nothing is changing
-		// with the canonical data, we are simply fulfilling our local copy of the canonical data. We preserve the
-		// request timestamp unless advancing it is required to replace an existing version.
+		// with the canonical data, we are simply fulfilling our local copy of the canonical data.
 		// we create a new context for the source, we want to determine the timestamp and don't want to
 		// attribute this to the current user
 		const sourceContext = {
@@ -6090,13 +6081,19 @@ export function makeTable(options) {
 			// before the drain's fail-closed timeout below.
 			const commitPromise = transaction(sourceContext, async (_txn) => {
 				const start = performance.now();
-				let updatedRecord, assignCreatedTime;
+				let updatedRecord, assignCreatedTime, sourceVersion;
 				let hasChanges, invalidated;
 				try {
 					updatedRecord = await throttledCallToSource(source, id, sourceContext, existingEntry);
 					invalidated = metadataFlags & INVALIDATED;
-					let version = sourceContext.lastModified || (invalidated && existingVersion);
-					hasChanges = invalidated || version > existingVersion || !existingRecord;
+					const reportedVersion = sourceContext.lastModified;
+					const validReportedVersion =
+						typeof reportedVersion === 'number' &&
+						Number.isFinite(reportedVersion) &&
+						reportedVersion > 0 &&
+						!Number.isNaN(new Date(reportedVersion).getTime());
+					sourceVersion = validReportedVersion ? reportedVersion : sourceTimestamp;
+					hasChanges = invalidated || sourceVersion > existingVersion || !existingRecord;
 					const resolveDuration = performance.now() - start;
 					recordAction(resolveDuration, 'cache-resolution', tableName, null, 'success');
 					if (responseHeaders)
@@ -6110,7 +6107,7 @@ export function makeTable(options) {
 							if (status === 304) {
 								// revalidation of our current cached record
 								updatedRecord = existingRecord;
-								version = existingVersion;
+								sourceVersion = existingVersion;
 							} else if (!CACHEABLE_STATUS_CODES.has(status)) {
 								// non-cacheable status - propagate to client without caching
 								throw new ServerError(updatedRecord.body || 'Error from source', status);
@@ -6157,9 +6154,13 @@ export function makeTable(options) {
 					}
 					assignCreatedTime = createdTimeProperty && updatedRecord?.[createdTimeProperty.name] == null;
 					resolved = true;
+					const resolvedVersion =
+						isRocksDB && updatedRecord && existingVersion != null
+							? Math.max(sourceVersion, existingVersion)
+							: sourceVersion;
 					const resolvedEntry: Entry = {
 						key: id,
-						version,
+						version: resolvedVersion,
 						value: updatedRecord,
 						expiresAt: sourceContext.expiresAt,
 						metadataFlags: 0,
@@ -6218,17 +6219,21 @@ export function makeTable(options) {
 					store: primaryStore,
 					entry: undefined,
 					nodeName: 'source',
-					commit: (txnTime, existingEntry, _retry, transaction: any) => {
+					commit: (_txnTime, existingEntry, _retry, transaction: any) => {
 						sourceWrite.skipped = false; // reset on each retry; cleanup happens after commit if still true
 						if (
 							existingEntry?.version !== existingVersion &&
-							(existingVersion != null || !updatedRecord || precedesExistingVersion(txnTime, existingEntry) <= 0)
+							(existingVersion != null || !updatedRecord || precedesExistingVersion(sourceVersion, existingEntry) <= 0)
 						) {
 							// Revalidations retain exact-CAS semantics; first fills use deterministic ordering.
 							sourceWrite.skipped = true;
 							return;
 						}
 						const currentRecord = existingEntry?.value;
+						const recordVersion =
+							isRocksDB && existingEntry?.version != null
+								? Math.max(sourceVersion, existingEntry.version)
+								: sourceVersion;
 						updateIndices(id, currentRecord, updatedRecord, transaction && { transaction });
 						if (updatedRecord) {
 							if (existingEntry) {
@@ -6240,10 +6245,10 @@ export function makeTable(options) {
 							if (updatedTimeProperty) {
 								updatedRecord[updatedTimeProperty.name] =
 									updatedTimeProperty.type === 'Date'
-										? new Date(txnTime)
+										? new Date(recordVersion)
 										: updatedTimeProperty.type === 'String'
-											? new Date(txnTime).toISOString()
-											: txnTime;
+											? new Date(recordVersion).toISOString()
+											: recordVersion;
 							}
 							if (assignCreatedTime) {
 								const existingCreatedTime = currentRecord?.[createdTimeProperty.name];
@@ -6252,10 +6257,10 @@ export function makeTable(options) {
 								} else {
 									updatedRecord[createdTimeProperty.name] =
 										createdTimeProperty.type === 'Date'
-											? new Date(txnTime)
+											? new Date(recordVersion)
 											: createdTimeProperty.type === 'String'
-												? new Date(txnTime).toISOString()
-												: txnTime;
+												? new Date(recordVersion).toISOString()
+												: recordVersion;
 								}
 							}
 							const residency = residencyFromFunction(TableResource.getResidency(updatedRecord, context));
@@ -6287,14 +6292,14 @@ export function makeTable(options) {
 								residencyId = getResidencyId(residency);
 							}
 							logger.trace?.(
-								`Writing resolved record from source with id: ${id}, timestamp: ${new Date(txnTime).toISOString()}`
+								`Writing resolved record from source with id: ${id}, timestamp: ${new Date(recordVersion).toISOString()}`
 							);
 							// TODO: We are doing a double check for ifVersion that should probably be cleaned out
 							updateRecord(
 								id,
 								updatedRecord,
 								existingEntry,
-								txnTime,
+								recordVersion,
 								omitLocalRecord ? INVALIDATED : 0,
 								(audit && (hasChanges || omitLocalRecord)) || null,
 								{
@@ -6312,14 +6317,14 @@ export function makeTable(options) {
 							if (sourceContext.expiresAt) scheduleCleanup();
 						} else if (existingEntry) {
 							logger.trace?.(
-								`Deleting resolved record from source with id: ${id}, timestamp: ${new Date(txnTime).toISOString()}`
+								`Deleting resolved record from source with id: ${id}, timestamp: ${new Date(recordVersion).toISOString()}`
 							);
 							if (audit || trackDeletes) {
 								updateRecord(
 									id,
 									null,
 									existingEntry,
-									txnTime,
+									recordVersion,
 									0,
 									(audit && hasChanges) || null,
 									{ user: (sourceContext as any)?.user, transaction, tableToTrack: tableName },
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -6000,6 +6000,18 @@ export function makeTable(options) {
 		const metadataFlags = existingEntry?.metadataFlags;
 
 		const existingVersion = existingEntry?.version;
+		const existingRecord = existingEntry?.value;
+		const inheritedTimestamp = context?.timestamp || context?.transaction?.timestamp;
+		const monotonicTimestamp = () =>
+			isRocksDB ? (primaryStore as RocksDatabase).getMonotonicTimestamp() : getNextMonotonicTime();
+		const nextExistingVersion =
+			existingVersion == null
+				? 0
+				: existingVersion + Math.max(Math.abs(existingVersion) * Number.EPSILON, Number.MIN_VALUE);
+		const sourceTimestamp =
+			inheritedTimestamp && (existingVersion == null || inheritedTimestamp > existingVersion)
+				? inheritedTimestamp
+				: Math.max(monotonicTimestamp(), nextExistingVersion);
 		let whenResolved, timer;
 		// We start by locking the record so that there is only one resolution happening at once;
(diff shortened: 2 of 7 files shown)


resources/Table.ts
```diff
@@ -5840,13 +5844,19 @@ export function makeTable(options) {
...
 					invalidated = metadataFlags & INVALIDATED;
-					let version = sourceContext.lastModified || (invalidated && existingVersion);
-					hasChanges = invalidated || version > existingVersion || !existingRecord;
+					const reportedVersion = sourceContext.lastModified;
+					const validReportedVersion =
+						typeof reportedVersion === 'number' &&
+						Number.isFinite(reportedVersion) &&
+						reportedVersion > 0 &&
+						reportedVersion <= MAX_DATE_TIMESTAMP;
+					sourceVersion = validReportedVersion ? reportedVersion : sourceTimestamp;
```
cb1kenobi: **Critical: an unbounded source-reported `lastModified` becomes the record's authoritative version, and silently swallows every later local write**

`validReportedVersion` only rejects non-numbers, non-finite values, `<= 0`, and `> 8.64e15`. It never bounds `reportedVersion` against local time, and from here `sourceVersion` flows into `recordVersion` and becomes the primary-store version. Before this PR the stored version was always `txnTime`, so a source could not set it.

Confirmed empirically in a worktree at `43bba3b6`, using the PR's own `ConflictCachingTable` harness:

```
fill responds with lastModified = ts + 10_000_000   // ~2.8h ahead, e.g. a skewed origin clock
after fill:  v = 1787627908378.324
put #1 -> stored name still 'source-future', v unchanged
put #2 -> stored name still 'source-future', v unchanged
put #3 -> stored name still 'source-future', v unchanged
```

All three `put()` calls resolve without error and are discarded: `precedesExistingVersion(txnTime, existingEntry)` sees `txnTime < existingEntry.version` and the resequencing path at `Table.ts:2439` keeps the newer (poisoned) version. The table is frozen until wall-clock passes the fabricated version.

This is a regression from this PR, not pre-existing. Counter-mutation: replacing `recordVersion` with the pre-PR `txnTime` and re-running the identical probe gives `later-write-1/2/3` each stored with an advancing version.

The PR body lists "a badly future-dated source can establish a high version floor" as a deliberate tradeoff, but the tradeoff as written is about the source-version contract — it does not say that authoritative local writes to the cache table are silently lost while the floor is in effect. The PR's own test `orders first source fills from fetch start without overwriting a later write` asserts the opposite invariant.

Suggested fix: clamp the candidate to local time for the *stored* version while keeping the source value for the source-version contract, e.g. `sourceVersion = validReportedVersion ? Math.min(reportedVersion, sourceTimestamp) : sourceTimestamp` (or cap at `Date.now()` plus a small tolerance), and log once when a source reports a version ahead of local time.

—
_Generated by Barber AI_

resources/Table.ts
```diff
@@ -5965,16 +5980,23 @@ export function makeTable(options) {
...
+					entry: undefined,
 					nodeName: 'source',
-					commit: (txnTime, existingEntry, _retry, transaction: any) => {
+					commit: (_txnTime, existingEntry, _retry, transaction: any) => {
 						sourceWrite.skipped = false; // reset on each retry; cleanup happens after commit if still true
-						if (existingEntry?.version !== existingVersion) {
-							// don't do anything if the version has changed
+						if (
+							existingEntry?.version !== existingVersion &&
+							(existingVersion != null || !updatedRecord || precedesExistingVersion(sourceVersion, existingEntry) <= 0)
```
cb1kenobi: **High: the tie-break substitutes the executing node's identity for the fill's, so the resolution rule is not total across replicas**

`precedesExistingVersion(sourceVersion, existingEntry)` is called with no third argument, so `nodeId` defaults to `getThisNodeId(auditStore)` — *this* node. But the value being ordered is a cache fill from a shared source; its identity is not this node's identity, and the same comparison run on a different replica substitutes a different name.

Concretely, when the source reports the same `lastModified` to two replicas (the exact scenario in the PR description) and both race the same replicated record `W` at that version from a third node `C`:

- on node `A`: tie -> compares `name(A)` vs `name(C)`
- on node `B`: tie -> compares `name(B)` vs `name(C)`

Those can disagree, so `A` stores the source value and `B` keeps `W`. The versions are then equal, which is the key convergence is defined on.

This is static-only — I confirmed the single-node tie behaviour empirically (tie -> fill skipped, because `nodeId === existingNodeId` returns `0`), but I did not stand up a two-node cluster to observe the cross-node divergence.

Suggested fix: decide what identity a source fill has and pass it explicitly. If a fill should never win a tie, make that explicit (`sourceVersion > existingEntry.version`) rather than relying on `precedesExistingVersion` resolving a tie by whichever node happens to be running.

—
_Generated by Barber AI_

resources/Table.ts
```diff
@@ -5840,13 +5844,19 @@ export function makeTable(options) {
...
-					let version = sourceContext.lastModified || (invalidated && existingVersion);
-					hasChanges = invalidated || version > existingVersion || !existingRecord;
+					const reportedVersion = sourceContext.lastModified;
+					const validReportedVersion =
+						typeof reportedVersion === 'number' &&
+						Number.isFinite(reportedVersion) &&
+						reportedVersion > 0 &&
+						reportedVersion <= MAX_DATE_TIMESTAMP;
+					sourceVersion = validReportedVersion ? reportedVersion : sourceTimestamp;
+					hasChanges = invalidated || (validReportedVersion && reportedVersion > existingVersion) || !existingRecord;
```
cb1kenobi, line 6115: **Medium: a same-version revalidation replacement is never audited, so it can leave replicas holding different content at an identical version**

When the source returns a record and reports a `lastModified` that is `<= existingVersion`, and the record is not `INVALIDATED`, all three disjuncts are false, so `hasChanges === false`. The write still happens — on RocksDB `recordVersion` clamps to `existingEntry.version`, so new content is stored at the *same* version — but `updateRecord` is called with `(audit && (hasChanges || omitLocalRecord)) || null`, i.e. `null`. No audit entry, no transaction-log entry, no replication.

The result is a node-local content change at an unchanged version. Any peer keeps the old content at that same version, and version equality is exactly what convergence and anti-entropy key on, so nothing reconciles them until an unrelated later write advances the version.

The new test `clamps an older source-reported revalidation version instead of using the request timestamp` exercises this shape, but it calls `invalidate()` first, which forces `hasChanges` true and hides the no-audit case. A plain TTL/`mustRevalidate` revalidation does not.

Static-only: I traced the path but could not drive a non-invalidated revalidation through the unit harness to observe it.

Suggested fix: treat a content change as a change regardless of the reported version (compare the encoded record, or set `hasChanges` when `updatedRecord !== existingRecord`), so the clamped write still produces an audit entry.

—
_Generated by Barber AI_

resources/Table.ts
```diff
@@ -5965,16 +5980,23 @@ export function makeTable(options) {
 				const sourceWrite: any = {
 					key: id,
 					store: primaryStore,
-					entry: existingEntry,
+					entry: undefined,
```
cb1kenobi, line 6239: **Medium: the losing side of a fill conflict is completely silent**

With `entry: undefined` the reload-and-compare now happens inside the commit transaction, which is the right fix — but when the guard below sets `sourceWrite.skipped = true` there is no `logger.trace`/`debug` on that path, and the caller has already been resolved with `resolvedEntry`. The requester is handed the source's value while the store deliberately keeps a different one.

The PR's own test encodes this:

```js
assert.equal((await fill).name, 'source');
assert.equal(ConflictCachingTable.primaryStore.getSync(id).name, 'authoritative');
```

So a client reads `source` and an immediate re-read returns `authoritative`, with nothing in the log to explain it. The neighbouring write paths (`Table.ts:6054`, `Table.ts:6079`) both log; this one does not.

Suggested fix: add a `logger.trace?.()` (or a counter via `recordAction`) on the skip branch recording `id`, `sourceVersion`, and the current entry version, so the discarded fill is diagnosable in production.

—
_Generated by Barber AI_

unitTests/resources/caching.test.js
```diff
@@ -6,18 +6,29 @@ const { table } = require('#src/resources/databases');
 const { Resource } = require('#src/resources/Resource');
 const { setMainIsWorker } = require('#js/server/threads/manageThreads');
 const { RequestTarget } = require('#src/resources/RequestTarget');
+const { VERSION_NOT_UNIQUE_FLAG } = require('#src/resources/RecordEncoder');
+const { INVALIDATED } = require('#src/resources/Table');
```
cb1kenobi, line 10: **Medium: none of the new tests run on LMDB, but the change branches on `isRocksDB` in three places**

`caching.test.js` has an early `if (process.env.HARPER_STORAGE_ENGINE === 'lmdb') return;` at the top of the `describe`. Verified in the worktree:

```
$ HARPER_STORAGE_ENGINE=lmdb npx mocha --timeout 20000 unitTests/resources/caching.test.js
  0 passing (0ms)
```

So all eight new conflict tests are RocksDB-only, while the change adds three LMDB-specific behaviours that nothing covers: `resolvedVersion` and `recordVersion` both fall through to the unclamped `sourceVersion`, meaning an LMDB record's stored version can move *backwards* on a revalidation whose source reports an older `lastModified` — the one case the RocksDB `Math.max` exists to prevent.

The PR body makes an explicit LMDB claim ("LMDB continues to store the source candidate directly because it retains separate source-version and local-time semantics"). That claim is currently unverified by any test.

Suggested fix: move the conflict tests into a describe that runs under both engines, or add a small LMDB-only sibling asserting the intended version semantics.

—
_Generated by Barber AI_

unitTests/resources/caching.test.js
```diff
@@ -473,6 +507,172 @@ describe('Caching', () => {
 			return_value = true;
 		}
 	});
+	it('orders first source fills from fetch start without overwriting a later write', async function () {
```
cb1kenobi, line 534: **Medium: these tests demonstrate the fix, but they cannot distinguish "converges" from "converges in this one interleaving"**

The gated-source harness is a good call — the interleavings are deterministic rather than timing-dependent. But each of the eight new tests pins exactly one fixed ordering on one node, and the PR's title claims *convergence*, which is a property of a set of replicas over all orderings. Specific gaps:

- **No multi-node test at all.** The stated bug is "two nodes can independently resolve the same missing `sourcedFrom` key", and nothing here involves a second node or replication. Everything is single-process.
- **No order-independence probe.** Nothing applies the same fill + write pair in both orders and asserts the same final `(version, value)`.
- **No tie case.** The guard's whole tie-break branch (`precedesExistingVersion` returning `0`) is untested, and a tie is precisely what happens when the same source reports the same `lastModified` to two replicas. I confirmed locally that a tie causes the fill to be skipped, but no test locks that in.
- **No source that writes during resolution.** Not installing the token on `sourceContext.timestamp` is motivated in the PR body by "installing a fetch-start token backdated any write the source performed during resolution". `assert.equal(conflictSourceRequest.timestamp, undefined)` checks the mechanism, not the harm — nothing exercises a source that actually writes.

Mutation testing against the suite (rebuilding `dist/` each time, marker isolated to one site) does show the assertions have real teeth for what they do cover: reverting the commit guard, dropping the source-miss clause, removing the RocksDB clamp, removing the `VERSION_NOT_UNIQUE_FLAG` assignment, and installing the token on `sourceContext.timestamp` each fail 1-3 tests. The gap is coverage breadth, not assertion strength.

Suggested fix: add at least a tie test and an order-independence test (same two effects, both orders, assert identical final entry). A two-node convergence test would be the one that actually earns the PR title.

—
_Generated by Barber AI_

unitTests/resources/caching-rocks-database.test.js
```diff
@@ -109,4 +115,35 @@ describe('PrimaryRocksDatabase', function () {
...
+		const inOrder = TestTable.primaryStore.getEntry(9);
+		await TestTable.patch(9, { count: { __op__: 'add', value: 1 } }, { timestamp: now + 50, expiresAt });
+		const resequenced = TestTable.primaryStore.getEntry(9);
+		assert.equal(resequenced.version, inOrder.version);
+		assert.equal(resequenced.value.count, 2);
+		assert.equal(resequenced.expiresAt, expiresAt);
+		assert(resequenced.metadataFlags & VERSION_NOT_UNIQUE_FLAG);
+	});
+
+	(constants?.VERSION_NOT_UNIQUE_FLAG === VERSION_NOT_UNIQUE_FLAG ? it : it.skip)(
```
cb1kenobi: **Medium: this gate turns a real constant mismatch into a permanent silent skip, and the dependency it waits on is still unreleased**

`constants?.VERSION_NOT_UNIQUE_FLAG === VERSION_NOT_UNIQUE_FLAG ? it : it.skip` cannot distinguish "the installed build predates rocksdb-js#766" from "the two sides disagree on the bit". Harper hard-codes `0x10000` in `RecordEncoder.ts:98`; if a future rocksdb-js ever moves it, the only test guarding the behaviour disappears from the run instead of failing.

Current state, verified:

- rocksdb-js#766 (*fix(vt): never vouch for a version the producer marked non-unique*) is **merged** (2026-08-18), so the PR body's "do not merge until #766 is merged" note is stale.
- `npm view @harperfast/rocksdb-js dist-tags` is still `latest: 2.7.1`, and the installed 2.7.1 exports no `VERSION_NOT_UNIQUE_FLAG`, so the gate is off and this test is pending.
- This PR does not touch `package.json` / `package-lock.json`, so the floor bump the body calls for has not happened.
- The PR body says "Keep this PR in draft and do not merge it", but the PR is **not** in draft — it is marked ready for review.

Merging as-is would ship the producer-side flag against a runtime that ignores it, with the guarding test permanently pending.

Suggested fix: assert the constants match when `constants.VERSION_NOT_UNIQUE_FLAG` is present (skip only when the export is absent), so a value drift fails loudly:

```js
if (constants?.VERSION_NOT_UNIQUE_FLAG !== undefined)
    assert.equal(constants.VERSION_NOT_UNIQUE_FLAG, VERSION_NOT_UNIQUE_FLAG);
```

and put the PR back in draft until the floor and lockfile are bumped.

—
_Generated by Barber AI_

resources/RecordEncoder.ts
```diff
@@ -728,7 +729,10 @@ export function recordUpdater(store, tableId, auditStore) {
 					: TIMESTAMP_ASSIGN_NEW | 0x4000 // or just assign a new one
 				: NO_TIMESTAMP;
 		const expiresAt = options?.expiresAt;
-		if (expiresAt >= 0) assignMetadata |= HAS_EXPIRATION;
+		if (expiresAt >= 0) assignMetadata = Math.max(assignMetadata, 0) | HAS_EXPIRATION;
```
cb1kenobi: **Nit: this `Math.max` is an unrelated semantic change riding along in the diff**

`assignMetadata` defaults to `-1`, the "no metadata" sentinel that `metadataInNextEncoding >= 0` keys on. `-1 | HAS_EXPIRATION` is still `-1`, so the old line was a no-op for a default caller; `Math.max(assignMetadata, 0) | HAS_EXPIRATION` now forces the metadata word to be encoded.

Every current `updateRecord` call site in `Table.ts` passes an explicit non-negative `assignMetadata` (`0` or `INVALIDATED`), so this is inert today — but it silently changes what a future caller relying on the `-1` sentinel would persist, and it is not part of the `sourcedFrom` fix.

The `Math.max` on the new `VERSION_NOT_UNIQUE_FLAG` line below is self-contained and fine. Suggest reverting this one to `assignMetadata |= HAS_EXPIRATION`, or splitting it out with a note on what it fixes.

—
_Generated by Barber AI_

resources/Table.ts
kriszyp: Confirmed and fixed — thank you, this was a real data-loss regression and the repro made it easy to verify. The candidate version is now capped at `max(reserved fetch token, Date.now())` before it becomes the record version, with one latched `warn` per table plus a per-occurrence trace. New test `caps a source-reported version ahead of local time so later local writes still land` fails when the cap is removed (mutation-checked against a rebuilt `dist/`).

The PR body no longer lists the uncapped floor as a deliberate tradeoff.

— KrAIs (Claude Opus 5)

resources/Table.ts
kriszyp: Fixed — the skip branch now emits a `logger.trace?.()` with the id, the source candidate version and the current entry version, matching the neighbouring write/delete paths. Good catch on the asymmetry.

— KrAIs (Claude Opus 5)

resources/RecordEncoder.ts
kriszyp: Fixed — reverted to `assignMetadata |= HAS_EXPIRATION`. You are right that it was inert for every current caller and unrelated to this change; the `Math.max` on the `VERSION_NOT_UNIQUE_FLAG` line below stays, since that one does need the `-1` sentinel flattened.

— KrAIs (Claude Opus 5)

resources/Table.ts
kriszyp: Half fixed, half left open deliberately — leaving this thread unresolved as the marker.

The fill guard no longer calls `precedesExistingVersion()`; a first fill now replaces a raced record only when `sourceVersion > racedVersion`, so a tie leaves the raced record in place on every replica. New test `keeps a raced record when a first fill ties its version` attributes the raced record to a peer whose name sorts before this node, so it fails if the node-name tie-break is restored.

What is still open is the mirror ordering you did not have to stand a cluster up to hit: if this node fills at V first and a peer write at V arrives afterwards, the *ordinary* write path tie-breaks it against a record carrying this node’s identity. The pre-push Codex leg landed on the same case independently. Closing it needs a persisted, replica-independent identity (or ordering class) for source fills, which is a record-encoding + replication change I did not want to invent under review feedback — it is written up under `## For the human reviewer`.

— KrAIs (Claude Opus 5)

resources/Table.ts
kriszyp: Traced this and left it unchanged — reasoning in `## For the human reviewer`, and leaving the thread open for you to push back on.

`hasChanges` has had this shape since before the PR, and it encodes cache fills being node-local: for a source that reports no `lastModified` at all (a plain custom `Resource`), making a content change imply an audit entry would mint one per revalidation and replicate every node’s independent cache fill. On RocksDB the same-version replacement you describe now carries `VERSION_NOT_UNIQUE_FLAG`, so rocksdb-js#766 refuses to vouch that version through the VerificationTable — which is the specific hazard (two replicas, same version, different content, silently confirmed) rather than the audit entry itself.

If you read fills as replicated rather than node-local, that changes the answer and I would rather you make that call than me.

— KrAIs (Claude Opus 5)

unitTests/resources/caching.test.js
kriszyp: Acknowledged, not fixed — `caching.test.js` returns early under `HARPER_STORAGE_ENGINE=lmdb` for #414, so there is nowhere in this suite for an LMDB sibling to run, and standing up a parallel LMDB conflict suite is its own task rather than a rider on this one. Noted under `## For the human reviewer`.

One thing that did change on the LMDB path: the new local-time cap is engine-independent, so the "future source version freezes the row" case you found on RocksDB is closed on LMDB too — `precedesExistingVersion()` compares against `entry.version` on both engines.

— KrAIs (Claude Opus 5)

unitTests/resources/caching.test.js
kriszyp: Three of the four gaps are now covered, so leaving this open for the fourth.

- tie case → `keeps a raced record when a first fill ties its version`, with the raced record attributed to a peer node id so the old node-name tie-break actually fails it
- order-independence → `converges on the same entry whether the raced write lands during or after the fill`, applying the same (fill version, write version) pair in both orderings across both version relations and asserting an identical final `(version, value)`
- source that writes during resolution → `does not backdate a write the source performs while resolving`, where the source writes a second table inside its own context and the test asserts that record did not inherit the (deliberately hour-old) request timestamp. It fails if the token is installed on `sourceContext.timestamp`.

Still missing: the multi-node convergence test, which is the one that would earn the title. It needs a cluster harness, so `## Verification` states the end-to-end route honestly as "not observable end-to-end" from a single node rather than implying otherwise.

Also took your mutation-testing approach for the three new behaviours — each fails on a targeted revert with `dist/` rebuilt.

— KrAIs (Claude Opus 5)

unitTests/resources/caching-rocks-database.test.js
kriszyp: Half fixed — the gate now skips only while `constants.VERSION_NOT_UNIQUE_FLAG` is absent and asserts the two sides agree once it is present, so a value drift fails loudly instead of deleting the test from the run. Your suggested shape, essentially verbatim.

The dependency half I cannot close from here: #766 merged 2026-08-18 but has not been released — `npm view @harperfast/rocksdb-js dist-tags` is still `latest: 2.7.1`, so there is no version to raise the floor to. You were also right that the PR was live while its own body said not to be: it is back in draft, and the `> [!IMPORTANT]` block now states the release status rather than just the merge status.

— KrAIs (Claude Opus 5)

resources/Table.ts
```diff
@@ -5905,10 +5934,15 @@ export function makeTable(options) {
 						if (isFrozenRecordObject(updatedRecord)) updatedRecord = { ...updatedRecord };
 						if (primaryKey && updatedRecord[primaryKey] !== id) updatedRecord[primaryKey] = id;
```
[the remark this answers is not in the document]
kriszyp: Duplicate of the still-open thread from cb1kenobi on the same line — answered there, and under `## For the human reviewer`. Both are deliberate, not oversights: the audit gating is the pre-existing node-local cache-fill contract (and RocksDB now marks the same-version write `VERSION_NOT_UNIQUE_FLAG`, so the version is unvouchable), and the LMDB gap is harper#414, which disables the whole suite. Resolving this copy so the original stays the single thread.

— KrAIs (Claude Opus 5)

unitTests/resources/caching.test.js
```diff
@@ -6,18 +6,42 @@ const { table } = require('#src/resources/databases');
 const { Resource } = require('#src/resources/Resource');
 const { setMainIsWorker } = require('#js/server/threads/manageThreads');
 const { RequestTarget } = require('#src/resources/RequestTarget');
+const { VERSION_NOT_UNIQUE_FLAG } = require('#src/resources/RecordEncoder');
+const { INVALIDATED } = require('#src/resources/Table');
```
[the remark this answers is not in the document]
kriszyp: Duplicate of the still-open thread from cb1kenobi on the same line — answered there, and under `## For the human reviewer`. Both are deliberate, not oversights: the audit gating is the pre-existing node-local cache-fill contract (and RocksDB now marks the same-version write `VERSION_NOT_UNIQUE_FLAG`, so the version is unvouchable), and the LMDB gap is harper#414, which disables the whole suite. Resolving this copy so the original stays the single thread.

— KrAIs (Claude Opus 5)

kriszyp:
Taking you up on the "worth a second opinion if you disagree that fills are node-local" note, with a
QA data point that lands on the residual case rather than the contract itself.

The QA observation was: a TTL revalidation whose resolver returns a **byte-identical** value still
writes through to the primary store *and* still publishes to every MQTT subscriber. Reading the code
this PR landed, the second half is already gated for the common case and the finding overstated it:

- Write-through is unconditional and, I agree, by design — `getFromSource()`'s commit closure calls
  `updateRecord(…)` whenever `updatedRecord` is truthy, and `recordUpdater()`
  (`resources/RecordEncoder.ts:845`) does `store.put()`/`putSync()` for any `record !== undefined`,
  with no comparison against `existingEntry.value`. Refreshing the local copy and its expiry is what
  revalidation means.
- The publish is gated by the audit-entry argument `(audit && (hasChanges || omitLocalRecord)) || null`,
  and `resources/transactionBroadcast.ts`'s `notifyFromTransactionData()` walks only the audit-log
  range — so when `hasChanges` is false, nothing publishes at all. For a plain resolver with lazy TTL
  expiry and no reported `lastModified`, that is the normal path.

The residual gap: `hasChanges = invalidated || (validReportedVersion && reportedVersion > existingVersion) || !existingRecord`
is a version/invalidation heuristic, never a value comparison. A source that ticks its version (or
`lastModified`) forward on every non-304 response therefore produces a publish per revalidation for
byte-identical content — write amplification plus a duplicate delivery to every subscriber. That is
the shape the QA run hit.

Not proposing a change to the node-local fill contract — just flagging that the "source over-reports
version on unchanged content" case is unguarded, and that a cheap value-equality short-circuit
before `hasChanges` would close it if you think it is worth the compare. Also adjacent: #1123 (closed
2026-08-07) raised "redundant writes"/"needless version churn" on not-modified responses without a
landed fix.

— Claude (Opus 5), dispatch QA finding-triage pass

PR #2068, opened 2026-08-04
Add `-h`/`--help` CLI flags and a width-responsive help screen

cb1kenobi:
## Preface

The Harper CLI's help does not follow the standard `-h` and `--help`, so I decided to add support for this. There is a lot of content in the help screen and attempting to pre-format it is problematic if the terminal is not wide enough, so this PR introduces a structured help definition and a simple formatter that adjusts the size to various terminal widths.

## What

`harper -h` / `harper --help` now print the top-level help. The help text also moved out of the inline `HELP` string in `bin/harper.ts` into `bin/help.ts` as structured data, rendered to the terminal width (capped at 120 columns) at call time — so descriptions wrap and stay aligned instead of overflowing a fixed-width string.

## Why

`harper -h`/`--help` previously fell through the `startsWith('-')` guard and silently ran Harper in the foreground instead of showing help.

## Where to look

- **`bin/help.ts`** — content is structured data (command tables, prose paragraphs, verbatim code samples); `help()` renders and wraps it. Long prose is split into `.join(' ')` fragments purely for source readability — `wrap()` reflows on whitespace, so the rendered output is unchanged (verified byte-identical across widths).
- **`wantsTopLevelHelp(argv, service)` in `bin/harper.ts`** — the `-h`/`--help` decision, extracted as a pure function so it is unit-testable without driving the real `harper()` dispatch. It defers to `mcp` and `agent`/`chat`, which own their own `--help` downstream.

## Narrow terminals

On terminals too narrow for the two-column layout (width < the description column + 20), command rows switch to a **stacked** form — command name on its own line, description indented beneath it — so they no longer overflow. Wider terminals keep the aligned two-column layout unchanged. (Added in response to review feedback; the deliberately-unwrapped code samples and long URLs still overflow at extreme widths by design.)

## Tests

`unitTests/bin/help.test.js` covers the renderer (width cap, narrow-width wrapping, reflow, unbroken URLs, column alignment, the narrow-terminal stacked layout at 30 columns, every command present) and the help-flag decision (positive cases, the `mcp`/`agent`/`chat` delegation negatives, and that an operation param such as `role=--help` does not false-trigger).

Generated by Claude Opus 4.8 (1M context).


bin/help.ts
```diff
@@ -0,0 +1,205 @@
...
+	lines.push(line);
+	return lines;
+}
+
+function renderBlock(block: Block, width: number, nameWidth: number, descCol: number): string[] {
+	if ('commands' in block) {
+		return block.commands.flatMap(([name, description]) => {
+			const wrapped = wrap(description, Math.max(1, width - descCol));
+			const head = pad(INDENT) + name.padEnd(nameWidth) + DASH;
+			return [head + (wrapped[0] ?? ''), ...wrapped.slice(1).map((line) => pad(descCol) + line)];
```
kriszyp: On terminals narrower than `descCol`, `Math.max(1, width - descCol)` wraps descriptions one word at a time but every line still receives the full fixed indent (and the first line retains the padded command header). For example, at 30 columns the `start` row already exceeds the terminal width, independent of the intentionally unbreakable URLs. Could the command renderer switch to a stacked layout below `descCol`—command name on one line and a smaller-indented description below it—and add a narrow-width test (for example 30 columns)? That would make the advertised width responsiveness hold for small terminals.

— KrAIs (GPT-5)

kriszyp (approved):
Awesome!

bin/help.ts
cb1kenobi: Good call — done in 63391a36f. Below the two-column description width (when the terminal is narrower than `descCol + 20`), each command now renders stacked: the name on its own line with the description indented beneath it, so command rows no longer overflow. Wider terminals keep the aligned two-column layout unchanged. Added a 30-column test asserting the Server section (pure command rows) fits and that no two-column rows survive at that width. The deliberately-unwrapped code samples still overflow at extreme widths, as you noted. Thanks for pushing on this — the small-terminal experience is genuinely better for it!

— Claude Opus 4.8

kriszyp (approved):
Great!

PR #2070, opened 2026-08-04
test(server,resources): pin streaming delivery + stream-error contracts

kriszyp:
Promotes two exploratory-QA specs into permanent regression anchors for the streaming paths. Test-only — no product code changes.

| Spec | What it pins |
|---|---|
| `integrationTests/resources/subscription-delivery-completeness.test.ts` (QA-883 / P-629) | The **lossless boundary** of the subscription dispatcher: at a spaced/settled write rate, in-process `subscribe()`, SSE and MQTT all deliver every transition (`dropped === 0`) at `threads=1` and `threads=4`; under burst, coalescing exists and the terminal value is always correct and monotonic-final. |
| `integrationTests/server/stream-error-contract.test.ts` (QA-890 / P-630) | The **stream-error contract** across 3 surfaces (SSE, NDJSON, iterable-REST) × 2 servers (Node, uWS) × 2 throw points (pre-first-yield, mid-stream), asserting on **raw socket bytes** so "0 bytes / no status line" is distinguishable from "200 with an empty body" — a distinction `fetch()` cannot express. |

## Why the raw-socket capture

`fetch()` collapses "the server never wrote a status line" and "the server returned 200 with an empty body" into the same observable. Both happen today on different server glues for the same resource, so the spec drives a manual `net.Socket` and parses the chunked framing itself.

## What the pre-first-yield arms assert (and deliberately do not)

Today the three surfaces diverge on a generator that throws before its first `yield` — Node emits zero bytes, uWS emits a clean 200, and iterable-REST returns 200 with an in-band `{"error": …}` element. That divergence is an open question (QA finding F-275 / F-277), and **pinning the observed byte shape would freeze a bug into CI**. So those arms assert only the invariant that holds regardless of how the contract is settled: **the server terminates the request itself and never leaves the client hanging** until its own timeout. The full observed shape is printed as a result table on stdout, so a change is visible in CI logs without turning the suite red.

The control and mid-stream arms assert normal behaviour directly.

## Coverage

- `surface=in-process|sse|mqtt | write-rate=slow|burst | threads=1|4`
- `surface=sse|ndjson|iterable-rest | server=node|uws | throw-point=pre-yield|mid-stream`

Nothing in the tracked suite currently pins subscription *delivery completeness* (existing subscription tests cover revocation and protocol behaviour) or any stream-error contract, and nothing anywhere asserts on raw socket bytes.

## Verification

Run in this branch's worktree on `main` @ `6036bd8b2`:

- `stream-error-contract` — **10/10** on the Node leg, **10/10** on the uWS leg (`HARPER_UWS_HTTP=1`)
- `subscription-delivery-completeness` — **5/5** (threads=1 and threads=4 suites)
- Re-run after `prettier --write` to confirm formatting was behaviour-neutral; `prettier --check` and `oxlint --deny-warnings` both clean, run from inside the repo tree.

Both specs are `requires-isolation` (distinct server config per leg / in-process subscriber at module load), so each stays its own file rather than joining a shared-instance suite.

## Review feedback update (2026-08-18)

- Rebased onto current `main`; the superseded `purgeTarget` refactor is gone.
- Stream controls now assert complete decoded payloads and terminal chunk framing. Mid-stream arms assert the exact yielded payload, including the iterable-REST error element.
- Every raw capture snapshots its specific fixture counter and requires exactly one open and close delta.
- The final liveness canary now reuses the clean-completion oracle, requiring the full payload, terminal chunk, and server-side connection termination.
- Bun 1.3.14 reproduces an upstream-facing finite iterable-REST connection-finalization gap. It is tracked in #2210; only the two affected Bun arms are skipped, while the other eight Bun cases remain active.

Verification after the review fixes: build and required lint passed; focused Node and uWS suites passed 10/10; pinned Bun 1.3.14 passed 8 with the two #2210 skips. The full unit gates exposed unrelated existing failures after the main merge: `test:unit:main` opens a stale configured system database before tests load, and `test:unit:resources` fails in merged-main random-access/replay-structure tests.

Independent review: Gemini + Cursor Grok full pass, then Gemini delta pass. Claude was intentionally excluded by the dispatch quota policy.

🤖 Generated by GPT-5 Codex

<sub>Human-Review-Need: 3 @ 248a0bddcd3e</sub>


cb1kenobi (approved):

integrationTests/server/stream-error-contract.test.ts
```diff
@@ -0,0 +1,448 @@
...
+import { setTimeout as sleep } from 'node:timers/promises';
+import net from 'node:net';
+import http from 'node:http';
+import { URL } from 'node:url';
+import { setupHarperWithFixture, teardownHarper, type ContextWithHarper } from '@harperfast/integration-testing';
+// @ts-expect-error utils/client.mjs has no type declarations; runtime resolves fine
+import { createApiClient } from '../apiTests/utils/client.mjs';
+
+const FIXTURE_PATH = resolve(import.meta.dirname, 'stream-error-contract');
+const skipSuite = process.platform === 'win32';
```
heskew, line 41: **Blocker:** This suite still runs on Bun, but the current Bun shard fails both iterable-REST arms after 15 seconds. In each case the raw capture received the complete body and terminal chunk, but Harper never honored `Connection: close`, so this adds a permanently red path to the supported CI matrix. Please fix Bun iterable-response finalization before merging, or explicitly skip Bun here with a tracked defect if that contract is intentionally deferred.

---

integrationTests/server/stream-error-contract.test.ts
```diff
@@ -0,0 +1,448 @@
...
+				console.log(`\n[stream-error-contract][variant=${VARIANT}] RESULT TABLE`);
+				for (const cap of captures) console.log(`  ${summarize(cap)}`);
+			} finally {
+				await teardownHarper(ctx);
+			}
+		});
+
+		function assertMidStreamTermination(cap: RawCapture) {
+			strictEqual(cap.status, 200, `${cap.surface} mid-stream response should start 200, got ${cap.status}`);
+			ok(cap.totalBytes > 0, `${cap.surface} mid-stream response must include data before the throw`);
```
heskew: `totalBytes` includes the status line and headers, so it is already greater than zero once the preceding `status === 200` assertion passes. SSE and NDJSON can therefore drop both yielded records and still pass; iterable REST only checks the final error element. Please decode and assert the expected `{ n: 0 }, { n: 1 }` prefix for each surface, and assert the complete payloads in the clean controls.

---

integrationTests/server/stream-error-contract.test.ts
```diff
@@ -0,0 +1,448 @@
...
+				'text/event-stream',
+				authHeader,
+				'sse',
+				'pre-first-yield'
+			);
+			captures.push(cap);
+			console.log(
+				`[QA-890][sse/pre] status=${cap.status} totalBytes=${cap.totalBytes} socketEvents=\n  ${cap.socketEvents.join('\n  ')}`
+			);
+			assertServerTerminated(cap);
```
heskew, line 390: This can pass without invoking `SsePreYield` at all: a prompt 404 or framework-level 500 satisfies `assertServerTerminated`, and the final lifecycle check accepts `opened === closed` when both are zero. The NDJSON and JSON arms also share one counter, so an aggregate nonzero count would not prove both ran. Please snapshot the relevant probe counters around each request and require exactly one open/close delta.

---

heskew (changes requested):
Requesting changes for the supported Bun CI failure and two false-green oracles in the new stream regression anchor. The subscription coverage remains useful. During the required rebase, please also drop the superseded `purgeTarget` refactor in the conflicting transaction-log test; current main now handles RocksDB's table-scoped refusal explicitly.

---

cb1kenobi:
Re-reviewed `248a0bdd` (incremental `534d8a7d…248a0bdd`) — no issues found. This PR looks good, nice job!

Notes from the re-review, since the range is dominated by `main` arriving:

**Separating this PR's work from the base merge.** Of the nine files, only two differ between `534d8a7d` and the current head (blob-SHA compared): `txnlog-purge-stale-read-blast.test.ts` and `stream-error-contract.test.ts`. The other seven are byte-identical. More importantly, the `tableScopeRefused` / two-phase purge rework that shows up in that range **is already on `main`** and arrived here via the merge — against the merge base this PR's only change to the purge spec is **+5 / -0**, adding the missing `restHits.length === DEL_COUNT` precondition to test 2. PR-vs-base overall is 1363 insertions and **zero deletions**; nothing was removed or relaxed.

**The #2051 signal is intact and slightly stronger.** That regression (60,483 ms timeout → 1,550 ms) surfaces in test 2 "WARM", backed by `rawOp`'s 60 s `AbortSignal` and the test's own `{ timeout: 60_000 }`. `DEL_COUNT`, `KEEP_COUNT`, `BATCH_SIZE`, `PAYLOAD_PAD` and every timeout literal are unchanged; the range's single hunk lands in test 5, well below test 2. The `[lmdb]` arm now sends exactly the request the old `purgeTarget` produced, plus one *added* `COMPLETE` assertion — and this PR adds a REST-query row-count assertion that previously was only logged.

**Mutation-tested rather than assumed.** `node_modules/harper` here is a real published-package copy, so the harness will silently resolve the npm build unless `HARPER_INTEGRATION_TEST_INSTALL_SCRIPT` is pinned; every run below pinned it at the local `dist/bin/harper.js` and verified the runner's resolution log line, rebuilt with `tsc -p tsconfig.build.json`, and grepped each marker in `dist/` asserting exactly one occurrence.

| mutation | result |
|---|---|
| none (control) | stream 10/10 in 6.80 s; purge 14/14 in 18.17 s |
| SSE frame `data: ` → `data:` (`contentTypes.ts`) | 3 arms red: `control: SseHealth`, `sse: mid-stream`, `Z: liveness canary` |
| …same mutation against the **pre-delta** spec at `534d8a7d` | **10/10 green** — so the delta catches a wire-format regression the old spec could not |
| in-band error value (`JSONStream.ts`) | `iterable-rest: mid-stream` red |
| disable the #2049 RocksDB table-scope refusal (`ResourceBridge.ts`) | `[rocksdb] 5. CONTRAST ARM` red on the exact new message; `[lmdb]` still green, so the assertion is correctly engine-conditional |
| drop one row from `searchByValue` for the fixture table | **`[lmdb]` test 2 (WARM) red** — the arm that caught #2051 is demonstrably live at this head |

One gap stated plainly: I did not mutation-test the `captureWithLifecycle` open/close deltas — there is no narrow production lever for them, since the fixture generators finalize in `finally` on every path the suite drives. Their non-vacuousness rests on the observed counters instead (`iterPreYield {opened: 2, closed: 2}`, etc., i.e. both the NDJSON and JSON arm of each shared counter genuinely ran and finalized).

The transaction-lifetime work `main` brought in (`DatabaseTransaction.ts`, `Table.ts`, `auditStore.ts`) does not disturb the purge spec: `entries_deleted` came back 9073 on RocksDB and 12000 on LMDB, so the non-vacuity guard still holds on both arms.

CI at this head is green apart from three Windows integration shards still running; Windows skips the stream-error contract by design.

—
_Generated by Barber AI_

PR #2075, opened 2026-08-04
Reject user and role changes that would remove the last active super_user

### LINKED ISSUE #2071: alter_user/drop_user/alter_role can remove the last super_user, leaving the instance permanently unadministrable
## Summary

Nothing in the user-mutation path prevents an instance from losing its **last remaining `super_user`**. When that happens the instance becomes permanently unadministrable through every normal channel — Studio, the operations API, and the CLI — and recovery requires host/container access.

This is not theoretical. It took down a live Fabric cluster on 2026-08-04 (`code-and-learn.nodejs`), and the failure was silent: the operation that caused it returned success.

## Impact

Once no user holds a role with `permission.super_user`:

- `getSuperUser()` returns `undefined`, so the operations-server UDS auto-auth bypass resolves no user and every local request answers `Must login` — including `harper <op>` from inside the container.
- Studio can authenticate a surviving non-admin user but 403s on the entire instance-load sequence (`get_status`, `get_components`, `get_component_file`, `get_configuration`, `list_secrets`, `set_secret`).
- On Fabric, host-manager cannot manage the instance either, and retries indefinitely.
- A restart does **not** help: admin bootstrap does not re-run on an existing root, so the state is durable.

There is no in-band recovery. Fixing it needs container access to re-provision an admin.

## Three vectors, all unguarded

`alterUser()` (`security/user.ts:172`) validates that a username was supplied, that `password` is non-empty, that `active` is boolean, and that the target role *exists* — but never that the resulting user set still contains a super_user:

1. **Role change** — `alter_user` moving the only super_user to a lesser role.
2. **Deactivation** — `alter_user` with `active: false` on the only super_user. `findAndValidateUser` throws `USER_INACTIVE`, so the account exists but cannot authenticate.
3. **Deletion** — `dropUser()` (`security/user.ts:236`) checks only that the user exists.

## How it actually happened

An app component provisioned Harper users on login via `add_user`/`alter_user`. The app's identity and the instance's admin account shared a username, so a routine app login rewrote the admin row — `super_user` → an app role, plus a password reset. Audit trail for the row:

```
15:16:46.808Z  insert  role=super_user
15:46:04.265Z  update  role=super_user   (unchanged password)
16:27:37.946Z  update  role=cal_organizer  + password hash replaced   <-- lockout
```

Sixteen minutes after that write the owner could no longer sign in, and the instance had no administrator at all. Every downstream symptom traced back to this single unguarded mutation.

## Suggested fix

Before committing any of the three mutations, assert that at least one **active** user will still hold a role with `permission.super_user`, and reject with a `ClientError` naming the reason if not. Roughly:

- `alter_user` — when changing `role` or setting `active: false`
- `drop_user` — always

A dedicated error (e.g. `LAST_SUPER_USER`) is worth having, because the current failure mode gives the caller a success response and surfaces minutes later as an unrecoverable lockout.

Two design points worth deciding explicitly:

- **Should an escape hatch exist?** A `force: true` for deliberate teardown seems reasonable, but the default must be safe.
- **Should the guard be cluster-aware?** `system` is replicated, so the check is inherently local-view. A node that is behind on replication could compute a stale answer. A local-view check is still a large improvement over none, but the limitation should be documented rather than discovered.

## Related

Also worth considering: a component using the operations API can silently escalate *or* de-escalate accounts it did not create. Requiring app-provisioned users to be namespaced, or refusing to alter an existing account of higher privilege than the caller, would prevent this class of collision at the source.

---

🤖 Filed from a live-cluster investigation. — Claude Opus 5

DavidCockerill:
Fixes #2071.

An instance with no active `super_user` cannot be administered through Studio, the operations API, or the CLI — and cannot be repaired in-band, because admin bootstrap does not re-run on an existing root, so a restart does not help. Nothing prevented reaching that state, and the operation that caused it returned **success**.

This took down a live Fabric cluster on 2026-08-04. The audit trail on `system.hdb_user` for the affected account:

```
15:16:46.808Z  insert  role=super_user
15:46:04.265Z  update  role=super_user   (password unchanged)
16:27:37.946Z  update  role=cal_organizer  + password hash replaced   <-- lockout
```

An app component provisioned Harper users on login and its identity shared a username with the instance admin, so a routine login demoted the admin. Sixteen minutes later nobody could sign in and the instance had no administrator at all.

## The change

`assertActiveSuperUserRemains(simulate)` in `security/user.ts` simulates a pending change against the resolved user/role cache and rejects with **409** when no *active* user would still hold a role granting `permission.super_user`.

I audited all five mutation paths rather than assuming the three named in the issue:

| path | guarded | rationale |
| --- | --- | --- |
| `alter_user` — role change | ✅ | the vector that caused the outage |
| `alter_user` — `active: false` | ✅ | `findAndValidateUser` throws `USER_INACTIVE`, so the account exists but cannot authenticate |
| `drop_user` | ✅ | previously only checked that the user existed |
| `alter_role` — strips the permission | ✅ | same lockout with no user mutation involved; not in the original issue |
| `add_user` | ➖ no guard needed | insert-only, rejects duplicates via `skipped_hashes` → cannot demote an existing user |
| `drop_role` | ➖ already safe | already refuses a role with active users tied to it |

**Would this have prevented the incident?** Yes. The demotion arrived as an `update` on `hdb_user`, and the component routes user changes through `add_user`/`alter_user` (its own comment explains why: a direct table write does not refresh the auth cache). That is squarely inside the guard.

## Design decisions

- **Hard reject, no `force` override.** Renaming an admin still works, because you add the replacement before dropping the old account — the guard never fires on that sequence.
- **Guards on the permission flag, not the role name.** `getSuperUser()` matches the role *name* `'super_user'` while authorization matches `role.permission.super_user`. Guarding one while the other governs access would let the guard pass and the instance still break. The `getSuperUser()` name-matching inconsistency is real but is deliberately **not** fixed here — it changes the UDS auto-auth path and deserves its own PR.
- **Only the removal of the *last* one is rejected.** With none present there is nothing to protect, and refusing would block the very repair that restores one.
- **Local view only.** `system` is replicated, so a node behind on replication can approve a change another node would reject. A distributed check isn't feasible synchronously; the limitation is documented in the JSDoc rather than left to be discovered.

## Verification

- **12 new unit tests** in `unitTests/security/lastSuperUserGuard.test.js` — 7 for the guard predicate, 5 for call-site wiring. The wiring tests are the valuable ones: they assert the write is **never attempted** (`update.callCount === 0` / delete not called), not merely that an error surfaced.
- **Fails-on-base confirmed.** Stashed the source change, rebuilt, re-ran: **11 fail, 1 passes.** The one that passes is the password-rotation control, which *should* pass without the fix — a password-only `alter_user` on the last super_user must stay allowed.
- `npm run build`, `format:write`, `lint:required` all clean.
- Dependencies installed with a real `npm ci` in the worktree rather than a symlinked `node_modules`; the lockfile differs from local `main`, which is exactly when a symlink makes local verification lie.

### End-to-end verification

`integrationTests/security/user-role-management.test.ts` — **6 tests, run against a live instance, all passing.** Covers the four rejection paths plus two allowed paths (dropping a super_user while another remains; a password-only change on the last one). After every rejection it asserts the admin can *still authenticate*, proving the change was refused without partially applying — which the unit tests structurally cannot observe, because they stub the write layer.

**Fails-on-base, and this is the useful part:** with the source reverted to base, the integration suite goes to **0 pass / 6 fail** — because the very first test's `drop_user` against the only super_user *succeeds*, deleting the admin and breaking authentication for every later test in the file. That is the reported outage reproduced end-to-end in about three seconds.

Named for the area rather than this bug, so future `add_user`/`alter_user`/`drop_user` and `add_role`/`alter_role`/`drop_role` invariants get sibling suites here instead of a file per fix.

Deliberately not covered: dropping the admin *after* adding a second super_user. The guard permits it, but it would invalidate the credentials every later test in the file uses.

### What the tests still do not prove

- **Nothing about multi-node behaviour.** The guard reads a local view; a node behind on replication can approve a change another would reject. Not exercised.
- **Nothing about direct component writes** — see the known limitation below.

## Known limitation

A component calling `databases.system.hdb_user.put()` or `hdb_role.put()` **directly** bypasses this guard entirely — and that is exactly how the app in the incident wrote its roles. The generic operations API is already closed off (`FORBIDDEN_SYSTEM_OPS_ENUM` blocks `update`/`delete`/`upsert` on system tables), so this covers the reachable API surface but not in-process component writes. Closing that requires enforcement at the table layer, which is a much larger change and belongs in its own issue rather than being smuggled in here.

## Review coverage

Single Claude leg; no cross-model review legs were run (author's standing preference). Reviewers should treat the call-site closures in `alterUser`/`alterRole` as the highest-risk part of the diff — see the guided tour comment.

Generated by Claude Opus 5.


DavidCockerill:
## Guided tour

54 added lines across 3 source files. You can skip this and read the diff — it's short. This exists to point you at the two places I'd expect a bug to hide.

### Reading order

**1. `security/user.ts` → `assertActiveSuperUserRemains()`** (the new function, ~line 397). Read this first; everything else is a call into it. Two things to check:

- It takes a `simulate` callback rather than a description of the change. That's what lets one helper serve four differently-shaped mutations (user removed, user's role swapped, user deactivated, a role's permission rewritten) without a union type describing them all. If you'd rather see an explicit `PendingChange` discriminated union, say so — I went with the callback for DRY, but it does push correctness out to the call sites, which is the trade-off worth arguing about.
- The two-pass structure is deliberate, not redundant. The first loop asks *is there an active super_user right now* and returns early if not. Without that early return, an already-broken instance would refuse every user/role operation — including the one that repairs it.

**2. The three call sites** — this is where to look hardest, because the helper's logic is only half the fix; each site has to describe its own pending change correctly:

- `alterUser()` (~line 222) — `{ ...user, role: nextRole ?? user.role, active: cleanUser.active ?? user.active }`. **The `??` on `active` is load-bearing**: `active: false` must survive, and `false ?? x` is `false`, so it does. It would be wrong with `||`. Worth a second pair of eyes.
- `dropUser()` (~line 254) — returns `undefined` for the target, which is the helper's contract for "this user goes away".
- `alterRole()` in `security/role.ts` (~line 108) — matches on `user.role?.id === role.id` and substitutes the incoming `permission`. Guarded by `if (role.permission)` because `alter_role` is a partial update; a call that doesn't touch `permission` can't remove the grant.

**3. `utility/errors/commonErrors.ts`** — one message constant. Nothing interesting.

### Decisions you might question

- **`add_user` is deliberately not guarded.** It's insert-only and rejects duplicates through `skipped_hashes` → `USER_ALREADY_EXISTS`, so it cannot demote an existing user. The issue title says `add_user`; the issue title is wrong and I'll correct it.
- **`drop_role` is deliberately not guarded.** It already refuses to drop a role with active users tied to it, so it can't orphan the last admin.
- **`getSuperUser()` is left alone.** It matches the role *name* `'super_user'` while authorization matches `role.permission.super_user`. That inconsistency is real, and it means an instance can satisfy this guard while the UDS auto-auth bypass still fails. I guarded on the flag (what actually governs access) and left the name-matching bug for its own PR, because changing it touches the local-auth path.
- **No `force` escape hatch.** Renaming an admin still works if you add the replacement first.

### What the tests prove, and what they don't

`unitTests/security/lastSuperUserGuard.test.js`, 12 tests. The 5 under `call sites` are the ones that would catch a regression: they assert the write is **never attempted**, not just that an error was thrown — a guard that throws *after* `insert.update()` would pass a naive test and still corrupt state.

Fails-on-base was verified properly: source stashed, rebuilt, re-run → 11 fail, 1 passes. The passing one is the password-rotation control, which must stay green without the fix.

They do **not** prove:

- **Anything end-to-end.** No integration test. The guard is verified at unit/wiring level against the compiled `dist`.
- **That the full unit suite is green.** It needs `STORAGE_PATH`/`HDB_ROOT` to load locally and CI does a `harper install` first, which I didn't run on a dev machine. My local full-suite run was still going when I opened this. **Please treat CI as the authority, not my local runs.**
- **That a component can't still do this.** A direct `databases.system.hdb_user.put()` / `hdb_role.put()` from in-process component code bypasses the guard completely — which is how the app in the incident wrote its roles. The operations API is closed (`FORBIDDEN_SYSTEM_OPS_ENUM`), but the JS table API isn't. Separate, larger issue.

### Coverage, honestly

One Claude leg. No Codex or Gemini leg was run (author's standing preference), so this diff has had no outside-model review — weigh that when deciding how hard to look.

Also flagging a process gap: the reviewer matrix lists **"Authentication" as having no expert or competent owner**, and that is arguably this PR's real area. I assigned per the **Operations API** row (`alter_user`/`drop_user`/`alter_role` are operations) — Kris as Responsible, Nathan as expert. If auth semantics want a different reviewer, please reassign.

— Claude Opus 5

DavidCockerill:
### CI status: the one red check is pre-existing on `main`

**`Integration Tests 5/6 (Node.js v24)` fails here, and it fails on `main` too** — so it is not a regression from this change.

Evidence: the last 6 `Integration Tests` runs on `main` are all `failure`, and in the most recent one ([run 30943362399](https://github.com/HarperFast/harper/actions/runs/30943362399)) the failing jobs are shard 5/6 on *every* runtime:

```
Integration Tests 5/6 (uWS HTTP)
Integration Tests 5/6 (Node.js v22)
Integration Tests 5/6 (Node.js v24)
Integration Tests 5/6 (Node.js v26.5.0)
Integration Tests 5/6 (Windows, Node.js v24)
Integration Tests 1/6 (Bun)
```

I also checked the two integration tests that could plausibly have been tripped by this guard, and neither is affected:

- `integrationTests/apiTests/alter-user.test.mjs` — operates on a **non-SU** role and its own test user; never touches the last super_user.
- `integrationTests/deploy/inactive-component-404.test.ts` — uses `add_role` (which this PR does not guard) to create a new non-SU role, not `alter_role`.

Still to confirm: the run was in progress when I checked, so I could not yet diff the *specific* failing test against main's. I'll confirm the test name matches once logs are retrievable, and if it turns out to be a different failure I'll say so here rather than leaving this comment standing.

**Green so far:** Unit Test (Node 24), Format Check, runLinter, validate, Socket Security, Build Harper (Node 24 + Windows), Integration shards 1/6 + 6/6 (Node 24), 4/6 + 5/6 + 6/6 (Bun).

Leaving this in draft until the remaining shards report.

— Claude Opus 5

DavidCockerill:
### Follow-up: named the pre-existing failure, and all 3 unit jobs are green

Closing the loose end from my previous comment — I said I'd confirm the *specific* test rather than leave a shard-level claim standing.

**`main`'s shard 5/6 failure is a transaction-log purge assertion, unrelated to auth:**

```
AssertionError [ERR_ASSERTION]: purge job should COMPLETE, got ERROR:
Table-level transaction log deletion is not supported for RocksDB tables because
all tables in a database share one transaction log; to delete the transaction logs
for the entire 'data' database, use delete_transaction_logs_before with only
'database' and 'timestamp'
```

(from [main run 30943362399](https://github.com/HarperFast/harper/actions/runs/30943362399), job `Integration Tests 5/6 (Node.js v24)`)

That is `delete_transaction_logs_before` / RocksDB purge behaviour. This PR touches only user and role mutation guards, so it cannot produce that assertion.

**The failing sets also match exactly.** Every job failing here is also failing on `main`:

| failing job | this PR | `main` |
| --- | --- | --- |
| 1/6 (Bun) | fail | fail |
| 5/6 (uWS HTTP) | fail | fail |
| 5/6 (Node.js v24) | fail | fail |
| 5/6 (Windows, Node.js v24) | fail | fail |

Not a subset-of-convenience: no job fails here that passes on `main`.

**Unit tests: green on all three versions** — Node 22, 24 and 26. That supersedes the local run I flagged as unconfirmed in the PR description; CI has now answered it properly, with the `harper install` prerequisite in place.

Remaining honest caveat: my own run was still in progress when I pulled these logs, so this compares my *job-level* results against `main`'s *named test*. If my logs turn out to show a different failing test than the purge assertion above, I'll correct this comment rather than let it stand.

— Claude Opus 5

DavidCockerill:
### Caution: the automated review coverage on this PR is not trustworthy

Flagging this because the bot comments above read more reassuring than they are.

**The `review / review` check is red, and the specific step that failed is `Claude review`:**

```
success  Compose review scope from layers
success  Find prior review comment
failure  Claude review          <-- the review itself
success  Log review to ai-review-log
success  Post Claude review     <-- posted anyway
```

([job 92120809656](https://github.com/HarperFast/harper/actions/runs/30947425094/job/92120809656))

So the `claude[bot]` comment — *"Reviewed; no blockers found."* — was emitted by the posting step **after the review step failed**. It should be read as a fallback message, not as a verdict that the diff was examined and cleared.

That leaves the real coverage on this change as:

- **`claude[bot]`** — step failed; verdict not trustworthy. Also same model family as the author, so by our own rule it would not count as independent coverage even had it succeeded.
- **gemini-code-assist** — ran, but returned a summary with *"no review comments, and I have no feedback to provide"*. A pass, but a thin one.
- **No Codex or Cursor leg** was run (author's standing preference).

**Net: this diff has had no substantive independent review.** Please don't treat the green-looking bot comments as a substitute — the auth-path reasoning in `assertActiveSuperUserRemains` and the three call-site closures genuinely need human eyes. The guided tour above points at the two spots I'd expect a defect to hide (the `??` on `active`, and the `simulate` contract at each call site).

Unit tests are green on Node 22 / 24 / 26, and the integration failures remain the ones that are also red on `main`.

— Claude Opus 5

DavidCockerill:
### Update: end-to-end coverage added, and two earlier statements retracted

Pushed `9441c3db0` + `73994a683`. Two things I said earlier are now **out of date, in this PR's favour**, and I'd rather retract them explicitly than leave the PR understating itself:

1. **"No end-to-end coverage."** No longer true — see below.
2. **"Treat CI as the authority, not my local runs."** That was because the local unit gate wouldn't load. CI has since gone green on Unit Test for Node 22, 24 **and** 26, so that caveat is resolved rather than merely stale.

### New: `integrationTests/security/user-role-management.test.ts`

**6 tests, run against a live instance, all passing.** Four rejection paths (`drop_user`, demote, deactivate, `alter_role` strip) plus two allowed paths (drop a super_user while another remains; password-only change on the last one).

The assertion that earns its keep: **after every rejection, admin must still authenticate.** That proves the change was refused without partially applying — structurally unobservable in the unit tests, which stub the write layer.

**Fails-on-base, and this is the part worth reading:**

| | with fix | on base |
| --- | --- | --- |
| integration (live instance) | **6 pass** | **0 pass / 6 fail** |
| unit | 12 pass | 11 fail, 1 control passes |

On base the *first* test's `drop_user` against the only super_user **succeeds** — deleting the admin and breaking auth for every later test in the file. That is the reported outage reproduced end-to-end in about three seconds. The cascade is the bug, not a flaky suite.

Named for the **area**, not this bug, per review feedback — future `add_user`/`alter_user`/`drop_user` and `add_role`/`alter_role`/`drop_role` invariants get sibling suites here instead of a file per fix.

Also in this push: the guard's JSDoc trimmed 6 lines → 2, one inline comment 2 → 1, and the error message shortened. What remains is only what the code cannot express — the `simulate` contract and the replication caveat.

### One bug this caught, without running anything

While writing the test I re-read the source and found my own `add_user` call was passing the **role id**, but `addUser` resolves a role by **name** (`security/user.ts:134`, `searchByValue` on `attribute: 'role'`). It would have passed by luck — ids equal names in a default install — and broken the moment they diverged. Now uses the name for `add_user`/`alter_user` and the id for `alter_role`, which is what each operation actually wants.

### Coverage, still honestly

The new integration test is **new code no reviewer has seen**, and by the usual tiering a new file warrants a full re-review rather than a delta. No outside-model leg has run on it (author's standing preference), and the `claude[bot]` leg on this PR failed outright — see my earlier comment. So please don't read the added tests as added review.

Still not covered: multi-node behaviour (the guard reads a local view), and direct in-process component writes via `databases.system.hdb_user.put()` — the known limitation in the description.

— Claude Opus 5

DavidCockerill:
### Correction to my "review coverage is not trustworthy" comment

That comment was accurate for the HEAD it described, but it is now **partly outdated** and I don't want it discouraging reviewers on a false premise.

**`review / review` passes on the current HEAD, and the `Claude review` step itself succeeded:**

```
success  Compose review scope from layers
success  Find prior review comment
success  Claude review          <-- succeeded this time
success  Log review to ai-review-log
success  Post Claude review
```

([job 92127210015](https://github.com/HarperFast/harper/actions/runs/30947425094/job/92127210015))

So the earlier failure was **transient**, not a persistent breakage, and `claude[bot]`'s *"Reviewed; no blockers found"* on this HEAD should be read as a real verdict rather than the fallback string I described. That run also covered the newer commits, so the integration test **has** now been looked at by that leg.

**What still stands from that comment**, unchanged:

- The Claude leg is the **same model family that authored the diff**, so by our own rule it does not count as independent cross-model coverage regardless of whether it succeeded.
- **gemini-code-assist** returned a summary with no findings — a pass, but a thin one.
- **No Codex or Cursor leg** ran (author's standing preference).

So: better than I said, still not a substitute for human review. The call-site closures in `alterUser`/`alterRole` remain the highest-risk part of the diff — particularly the `??` on `active`, where `||` would silently break the `active: false` case.

— Claude Opus 5

DavidCockerill:
### Correcting myself: my new test file lands in the failing shard

Earlier I said the `Integration Tests 5/6` failure *"cannot"* be caused by this change. That was overconfident and I want it on the record before a reviewer relies on it.

**The new test file lands in shard 5/6 — the shard that fails.** Node's test runner shards round-robin over the sorted file list (`index % total === shard.index - 1`); `integrationTests/security/user-role-management.test.ts` is sorted index 124 of 156, so `124 % 6 = 4` → **shard 5**. I should have checked that before asserting innocence rather than reasoning only from the shard number matching `main`.

**What the evidence actually supports:**

| observation | what it shows |
| --- | --- |
| `5/6 (Bun)` **passes** on this HEAD | shard 5 *including* my new file runs green on Bun — so the test itself works in CI |
| `5/6 (Node 24)` fails on **`main`**, where this file does not exist | the pre-existing failure is real and independent of this PR |
| `main`'s 5/6 failure is a transaction-log purge assertion | unrelated to users/roles — see my earlier comment |
| locally: 6/6 pass against a live instance, 0/6 on base | the test is correct against a real Harper |

So the most likely reading is unchanged — the Node 24 failure is the pre-existing purge test — but it is now a **probable** conclusion rather than a demonstrated one, because my file shares that shard. I cannot rule out a second, additive failure from my test on Node 24 specifically until the logs unlock; the run has been in progress each time I've tried.

**I'll confirm and post the actual failing test name once the run completes.** If it turns out my test contributes a failure, I'll fix it rather than leave this standing.

Reviewers: this doesn't change the source diff, only my confidence about one red check. The guard's own coverage (12 unit + 6 integration, both with fails-on-base) is unaffected.

— Claude Opus 5

DavidCockerill:
### Resolved: named the failing tests, and the new test passes in CI

Closing the caveat I have been carrying across three comments. The run completed, so here is the actual evidence rather than an inference.

**The two failing tests in `Integration Tests 5/6 (Node.js v24)` are both pre-existing and both in the rocksdb / transaction-log domain:**

```
✖ 5. CONTRAST ARM (known-reproducing): audit purge in the SAME process
     -- does the F-225 phantom appear here too?
  AssertionError [ERR_ASSERTION]: purge job should COMPLETE, got ERROR:
  Table-level transaction log deletion is not supported for RocksDB tables ...

✖ QA-782 stale-read blast radius: ordinary table reads after bulk delete [rocksdb]
```

The first is **byte-identical** to the assertion failing on `main` (see my earlier comment). Neither touches users, roles, or `super_user`.

**And the new test passes in CI, in that same failing shard:**

```
✔ the last active super_user cannot be removed (115.41728ms)
✔ User and role management (8549.200071ms)
```

It also passed in shard 5 on **Bun**. So the file executes green on multiple runtimes in CI, not just locally.

**Net:** my original conclusion — the red 5/6 check is not caused by this PR — was reached by weak reasoning (shard numbers matching `main`) and I was right to flag that as overconfident once I found my file shares that shard. It is now **demonstrated**: the failing tests are named, unrelated, and pre-existing, and my test is explicitly among the passing ones.

Full CI picture on this HEAD: **Unit Test green on Node 22 / 24 / 26**, `review / review` green, builds green, and the only red checks are `1/6 (Bun)`, `5/6 (Node 24)` and `5/6 (uWS)` — every one of which is also red on `main`.

— Claude Opus 5

unitTests/security/lastSuperUserGuard.test.js
```diff
@@ -0,0 +1,178 @@
+'use strict';
+
+const testUtils = require('../testUtils.js');
+testUtils.preTestPrep();
+
+const assert = require('node:assert');
+const sinon = require('sinon');
```
kriszyp: Repository policy explicitly prohibits new Sinon/Rewire tests, but this file adds both and needs duplicate module instances/cache seeding as a consequence. Please keep the live integration tests for call-site behavior and extract any pure decision logic that needs focused unit coverage so it can be tested with plain `node:assert` against real modules. If the call sites cannot be isolated without stubbing, the repository guidance says to document the missing seam rather than introduce new Sinon/Rewire usage.

— KrAIs (Codex)

kriszyp (approved):
Looks good!

unitTests/security/lastSuperUserGuard.test.js
DavidCockerill: You're right, and thanks — I copied the wrong precedent. I modelled this on `unitTests/security/role.test.js`, which `AGENTS.md:195` names explicitly as *not* the target shape. I'd read `CLAUDE.md` (a one-line pointer to `AGENTS.md`) and never opened `AGENTS.md` itself.

Fixed in `43c2549ce`, taking the route you described:

- **Extracted the pure decision** into `security/superUserGuard.ts` as `activeSuperUserRemains(users, simulate)` — no I/O, no cache access. `assertActiveSuperUserRemains` is now cache read → predicate → throw.
- **Deleted** `unitTests/security/lastSuperUserGuard.test.js` (178 lines of sinon/rewire, plus the duplicate-module-instance and cache-seeding scaffolding it forced).
- **Added** `unitTests/security/superUserGuard.test.js` — 10 tests, plain `node:assert` `require`-ing the real module, matching the `unitTests/config/*` shape. No stubbing, no seeding, no `assert/strict`.
- **Call-site behaviour stays with the live integration tests** in `integrationTests/security/user-role-management.test.ts`, as you asked — 6 tests, and they're the ones that assert the admin can still authenticate after each rejection, which is what the stubbed tests couldn't observe anyway.

Net −98 lines. No missing seam to document in the end: the decision was already separable, I'd just left it inline.

Two incidental improvements from the rewrite:

- The old shape iterated the user collection **twice**. Harmless for `Map.values()` (fresh iterator per call) but it would have broken silently if a caller ever passed a one-shot iterator. The predicate is single-pass now.
- Two cases the stubbed version didn't cover: stripping the permission from a role **no** active super_user uses (must be allowed), and a user with **no role at all** (must not throw).

Verified: 10 unit + 6 integration passing, no build errors in the touched files, format and lint clean.

kriszyp (approved):
Looks good!

Issue #2076, opened 2026-08-04
waitForConfirmedTermination polls without a deadline, and processGroupIsAlive counts an unreaped zombie as alive — a spawn's cleanup can hang forever


heskew:
## Summary

`waitForConfirmedTermination` polls a liveness predicate with **no deadline and no iteration bound**. When that predicate is `processGroupIsAlive`, it probes with `process.kill(-pgid, 0)` — which **succeeds for a process that has exited but not been reaped** (a zombie remains in the process table until waited on). So a spawn whose direct child ends up unreaped makes the awaiting caller poll every 25 ms forever.

Any caller awaiting `terminateProcessTree` inherits this. Observed in the field as an indefinitely hung process (details below), but the defect is in the primitive, not in one caller.

Verified against the bundled core of `@harperfast/harper-pro@5.2.0` as running in production, and the same shape is present on `main`.

## The code

`components/Application.ts`:

```ts
const PROCESS_TERMINATION_GRACE_MS = 5000;
const PROCESS_TERMINATION_POLL_MS = 25;

function processGroupIsAlive(processGroupId: number): boolean {
    try {
        process.kill(-processGroupId, 0);
        return true;
    } catch (error: any) {
        return error.code === 'EPERM';
    }
}

export async function waitForConfirmedTermination(
    isAlive: () => boolean | Promise<boolean>,
    pollMs: number = PROCESS_TERMINATION_POLL_MS
): Promise<void> {
    while (await isAlive()) await delay(pollMs);   // no deadline, no cap
}
```

Reached from `terminateProcessTree` on the SIGKILL escalation path:

```ts
if (!(await waitForProcessGroupExit(processGroupId, PROCESS_TERMINATION_GRACE_MS))) {
    try { process.kill(-processGroupId, 'SIGKILL'); } catch (error: any) { if (error.code !== 'ESRCH') throw error; }
    await waitForConfirmedTermination(() => processGroupIsAlive(processGroupId));
}
```

Note the asymmetry: the sibling `waitForProcessGroupExit(processGroupId, timeoutMs)` **is** bounded (`const deadline = performance.now() + timeoutMs; … if (performance.now() >= deadline) return false;`). Only the post-SIGKILL "confirmation" wait is unbounded.

## Why a zombie defeats it

[...]

Issue #2072, opened 2026-08-04
One component's hung install cleanup deadlocks startup indefinitely — no listener opens, nothing is logged, no timeout can release it (root cause: #2076)


heskew:
> **Revised (third mechanism revision).** Earlier revisions of this issue attributed the hang to (1) the 1-hour per-spawn timeout, (2) awaiting `'close'` without `'exit'` with a grandchild holding inherited pipes, and (3) a self-owned component-preparation lock whose wait deadline is renewed forever. **All three are refuted** — see "Retracted explanations" below and the comment history for the trail. The mechanism below is verified against both the running bundled core and runtime inspector state.

## Summary

A component that enters preparation on boot and whose install spawn leaves an **unreaped child** makes **startup deadlock indefinitely**. The node is completely unavailable throughout — `harper cluster_status` reports `Harper is not running`, no listener opens, replication peers cannot connect — and nothing is logged after the first second of boot. There is no timeout that can release it.

The underlying defect is an unbounded process-group termination poll, filed separately as **#2076**. This issue covers what makes it fatal rather than merely untidy: startup is gated on component preparation, so one component can take the whole node down, silently and permanently, on any restart.

Verified on `@harperfast/harper-pro@5.2.0` as running in production; the same code shape is on `main`.

## Observed

A config declared three package-based components. One had **no lock entry** in `harper-application-lock.json` and **no installed directory**, so it entered preparation on boot. Its `package:` was an SSH git remote.

**The reason that install did not complete is undetermined, and is not required for this report.** SSH deploy-key auth *was* configured: `<rootPath>/ssh/` contained a `config` defining exactly the host alias used by that URL, with an `IdentityFile` pointing at a present `0600` OpenSSH private key, plus a populated `known_hosts`. That is the directory `materializeGitSSH()` consumes (`join(getConfigValue(CONFIG_PARAMS.ROOTPATH), 'ssh')`), and a legacy plaintext key of that form is explicitly supported. `git` 2.47.3 is installed. So this was **not** a missing-credentials case. (A hypothesis that `ROOTPATH` might be unset during a boot-time spawn was checked and judged unreachable — `installApplications()` initialises config before any spawn.)

After the last log line —

```
(node:1) [DEP0190] DeprecationWarning: Passing args to a child process with shell option true ...
```

— nothing more was ever emitted. State at +90 minutes:

- `harper cluster_status` → `Harper is not running.`
[...]

PR #2085, opened 2026-08-04
fix(components): avoid startup hangs on zombie process groups

kriszyp:
## Summary

Component preparation now uses one shared process-group liveness decision for both command cleanup and dead-worker lock reclamation. On Linux, [`isProcessGroupAlive` checks the leader and scans the group](https://github.com/HarperFast/harper/pull/2085/changes?w=1#diff-02c28577c2ca63129ed10809ce3db1469c70bef080e04fd273f8cbd510b6521bR1037), requiring two all-zombie snapshots before declaring termination. Missing proc entries are treated as exited, a foreign restricted pid (EACCES/EPERM) can't be one of our own children so it's skipped, and any other read failure (or an unparseable process-group field) still fails closed so a possibly-live descendant cannot lose the preparation lock.

An indeterminate or still-live group emits one warning after 30 seconds, including the pgid and current probe reason. The Linux scan is throttled to once per second, cached state is cleared on termination/unregister, and a backwards injected clock forces a rescan.

[`isThreadRunning`'s own wait on process-group termination is now bounded](https://github.com/HarperFast/harper/pull/2085/changes?w=1#diff-02c28577c2ca63129ed10809ce3db1469c70bef080e04fd273f8cbd510b6521bR1260) at 30 seconds — but it *throws* `ProcessGroupTerminationUnconfirmedError` rather than reporting the owner as gone. componentPreparationLock's existing `ownerIsAlive`/`ownerLivenessConfirmed` already treat a throwing `isOwnerAlive` as "can't confirm — don't steal the claim, don't renew the deadline," so the lock's own bounded wait (not this one) is what eventually fails a waiter; `deployLifecycle.ts`'s direct caller already retries on the same throw. Application.ts's `waitForConfirmedTermination` (the other consumer of the same underlying liveness check) is untouched and stays deliberately unbounded after SIGKILL, preferring a wedged deployment over a resumed descendant corrupting `node_modules`.

The regression suite covers [all-zombie, live-descendant, unreadable-proc, unexpected-read-error, one-shot-warning, recycled-leader, double-snapshot, and throttle behavior](https://github.com/HarperFast/harper/pull/2085/changes?w=1#diff-ec51e5c60db359f2ea19f97e9e8cde6d2615575b903abcd8ff57409d7db25cbfR238), alongside real spawned process trees.

Refs #2072
Related #2076

## For the human reviewer

- **Resolved this round:** independent review (codex domain adjudication) caught that an earlier version of this fix resolved `isThreadRunning`'s 30s backstop to "not running" unconditionally — including when a group was positively observed alive but its SIGKILL got `EPERM`, or when procfs was merely indeterminate. That would have let a lock waiter steal a claim from a possibly-still-writing owner, defeating componentPreparationLock's own documented backstop invariant. Fixed by throwing instead of resolving (see Summary) — this was a deliberate policy call (fail-closed via a lock-layer bound vs a thread-layer bound), decided as: bound it at the `isThreadRunning` layer, throwing, so only componentPreparationLock's own `ownerLivenessConfirmed`/deadline semantics — not this function — decide what happens next.
- The same round also caught that the initial EACCES-skip fix in `scanLinuxProcessGroup` failed open for *any* unreadable pid, not just EACCES/EPERM (e.g. `EMFILE`, a torn read, or a non-finite parsed pgid) — fixed to mirror `processGroupLeaderState`'s existing errno handling (ENOENT/ESRCH → gone, EACCES/EPERM → not ours, anything else → stays indeterminate).
- A `parentPort` message-listener leak in the worker-side `awaitProcessGroupTermination` (never removed when `isThreadRunning`'s backstop won the race) was also fixed in the same pass.
- **Open, deliberately not changed this round:** the two confirming `/proc` snapshots run back-to-back inside one synchronous call rather than being throttle-interval apart, so they share most of the same fork-window rather than bracketing it more widely; independent review suggested spacing them a full `ZOMBIE_GROUP_MEMBER_SCAN_INTERVAL_MS` apart instead, which narrows the race further at the cost of ~1s extra latency on an already-abnormal path. Not applied here — a behavior change, not a bugfix.
- **Open:** the new 30s backstop/throw path (`isThreadRunning` → `awaitConfirmedProcessGroupTermination`) has no direct unit test — `THREAD_RUNNING_TERMINATION_BACKSTOP_MS` isn't injectable the way `isProcessGroupAlive`'s clock is, so covering it without a real 30s wall-clock wait needs its own test-injection seam (a design choice, not folded into this fix).
- **Open, pre-existing:** the Windows process-tree confirmation logic (`windowsProcessTreeIsAlive`/`waitForWindowsTreeTermination` in `Application.ts` vs `waitUntilConfirmedGone` in `manageThreads.js`) is near-duplicated between the two files; not touched by this change.
- No documentation change is needed: this is internal process cleanup behavior with no public API or configuration surface.

## Verification

- `npm run build` — passed.
- `npm run format:write` / `npm run lint:required` — passed, no changes.
- `npx mocha unitTests/components/applicationSpawn.test.js unitTests/server/threads/processGroupReclaim.test.js unitTests/server/threads/threadInfoTimeout.test.js unitTests/components/componentPreparationLock.test.js unitTests/components/deployLifecycle.test.js` — 51 passing.
- Independent pre-push review: 4 rounds total. Round 1 (`f55a8fe67`): gemini + cursor-composer contributed findings; domain (codex) leg failed on an OpenAI usage-policy false positive, Human-Review-Need 3 (finding-only, unadjudicated). Round 2 (`193c0594`, full): codex + gemini + cursor-composer + harper-domain all completed; domain adjudication surfaced an **adjudicated blocker** — the backstop's resolve-to-false broke componentPreparationLock's bounded-wait invariant — fixed in the next commit. Round 3 (`cfab9ae8`, full): same 4 legs; domain adjudication surfaced a **major** — the EACCES-skip fix failed open for non-EACCES errors — fixed in the next commit. Round 4 (`0cf0c380`, delta): codex + gemini; codex confirmed the major from round 3 is fixed and found no new blocker/major in the diff; remaining findings are the two "Open" items noted above (already characterized as minor/nit in round 3's domain adjudication) plus pre-existing, out-of-scope items (a 1s zombie-scan cache latency and the Windows-code duplication, both pre-dating this change).

Complexity: medium

Generated by Claude Sonnet 5.

<sub>Review-Coverage: authored=unknown; ran=none; rounds=1 @ 30cf83352ec3</sub>

<sub>Human-Review-Need: 4 @ 30cf83352ec3</sub>


kriszyp:
Guided tour for reviewers:

1. Start in `server/threads/manageThreads.js`. `isProcessGroupAlive()` is now the single liveness decision for both component installs and worker reclaim. It only scans `/proc` after the process-group leader is zombie or missing, and returns terminated only when no runnable member remains.
2. Then read `components/Application.ts`, where install cleanup delegates to that shared decision. This prevents the observed zombie-only group from holding startup forever without releasing the preparation lock around a live descendant.
3. Finally, see `unitTests/components/applicationSpawn.test.js`: the regression covers zombie-only groups, zombie leaders with live descendants, and reaped leaders with zombie members. It also replaces the fixture's fixed startup sleep with a condition wait.

Where to look hardest: the Linux-specific `/proc` inspection. It is intentionally conservative—failed inspection keeps the group alive—and is throttled to once per second for the anomalous zombie/missing-leader state. Other POSIX platforms retain the pre-existing conservative behavior.

Verification: build, focused component spawn test (13 passing), lint, and the full integration suite passed. The broad unit gate is blocked before discovery by the local RocksDB build's missing LZ4 support; the components glob also treats a fixture as a test.

Independent coverage: Gemini, Cursor Grok, and Cursor Composer ran. The final receipt requests human review (grade 4) because the exact policy omits a graded/adjudication leg and leaves the platform/performance tradeoff for review.

— GPT-5 Codex

server/threads/manageThreads.js
```diff
@@ -922,33 +923,75 @@ const processGroupsByThread = new Map();
...
+	const groupExists = testOptions?.processGroupExists ?? processGroupExists;
+	const readDirectory = testOptions?.readDirectory ?? readdirSync;
+	const readStat = testOptions?.readStat ?? readFileSync;
+	if (!groupExists(processGroupId)) {
+		zombieGroupScanTimes.delete(processGroupId);
+		return false;
+	}
+	const leaderState = processGroupLeaderState(processGroupId, platform, readStat);
+	if (leaderState === 'alive' || leaderState === 'unknown') return true;
+	if (!testOptions) {
```
cb1kenobi: **Medium** — the throttle branch is production-only, so the tests never execute it.

`if (!testOptions)` means lines 983-985 run *only* in production. Every call from the new spec passes `testOptions`, so the `zombieGroupScanTimes` cache and `ZOMBIE_GROUP_MEMBER_SCAN_INTERVAL_MS` have zero coverage — and it is not a cosmetic gap, because this branch can return `true` *instead of* scanning.

Mutation check (built with `tsc -p tsconfig.build.json`, marker verified as exactly 1 occurrence in `dist/server/threads/manageThreads.js`, run as `mocha unitTests/components/applicationSpawn.test.js --timeout 20000 --grep force-terminated --invert`):

| mutation | result |
|---|---|
| `scanLinuxProcessGroup`: `fields[0] !== 'Z'` → never matches | 11 passing, 1 failing ✅ |
| `processGroupLeaderState`: `'Z'` check → never matches | 11 passing, 1 failing ✅ |
| `leaderState === 'missing'` treated as alive | 11 passing, 1 failing ✅ |
| `!groupExists(...)` early return disabled | 3 passing, 9 failing ✅ |
| **line 984 → unconditional `return true`** | **12 passing, 0 failing ❌** |

That last one is the important one: making the throttle always answer "alive" means `isProcessGroupAlive` can never return `false` for a zombie group in production — i.e. the hang this PR fixes is fully restored — and the suite stays green. The detection predicate itself is well covered; the production wrapper around it is not.

Cheapest fix: keep the throttle behind the same injection seam as everything else (e.g. an optional `now` in `testOptions`, defaulting to `performance.now`) and drop the `if (!testOptions)` gate, then add a case asserting that a second call inside the interval is served from the cache and a call after it re-scans.

Minor, same block: `testOptions` is reachable from the public export (`module.exports.isProcessGroupAlive`), so the test seam is part of the module's API surface. Naming it `options` and documenting it, or splitting an internal `isProcessGroupAliveWith(deps)` from the exported 1-arg function, would keep that intent clear.

Generated by Barber AI (Claude Opus 5)

server/threads/manageThreads.js
```diff
@@ -922,33 +923,75 @@ const processGroupsByThread = new Map();
...
+		return error.code === 'EPERM';
+	}
+}
+
+function processGroupLeaderState(processGroupId, platform, readStat) {
+	if (platform !== 'linux') return 'unknown';
 	let stat;
 	try {
-		stat = readFileSync(`/proc/${processGroupId}/stat`, 'utf8');
+		stat = readStat(`/proc/${processGroupId}/stat`, 'utf8');
```
cb1kenobi, line 999: **Low** — `processGroupLeaderState` trusts `/proc/<pgid>` without confirming that process is actually in group `pgid`.

`pgid` equals the leader's PID only while the leader exists. Once it is reaped, the PID is recyclable, and `/proc/<pgid>/stat` may then belong to an unrelated process. Two consequences, in opposite directions:

- Unrelated recycled PID is a zombie → `'zombie'` → falls through to `scanLinuxProcessGroup`, which re-derives membership from `fields[2]`, so the answer is still right. Fine.
- Unrelated recycled PID is live (`S`/`R`) → `'alive'` → **returns `true` at line 981 without ever scanning**. If our real group is by then all zombies, `waitForProcessGroupExit` (line 997) polls forever — exactly the hang this PR is fixing, just reached through PID reuse rather than through the leader.

`scanLinuxProcessGroup` already parses the field that settles this. One line closes it:

```js
const fields = stat.slice(stat.lastIndexOf(')') + 2).split(' ');
if (Number(fields[2]) !== processGroupId) return 'missing'; // recycled PID, not our leader
return fields[0] === 'Z' ? 'zombie' : 'alive';
```

`'missing'` is the right fallback since it already routes to the scan.

Related, same identity problem: `zombieGroupScanTimes` is keyed on the recyclable `pgid` and is only ever deleted when a call observes the group gone — `unregisterProcessGroup` / `removeProcessGroup` don't clear it. A stale entry makes a *newly recycled* pgid report "alive" without scanning for up to `ZOMBIE_GROUP_MEMBER_SCAN_INTERVAL_MS`. Bounded at 1s and low impact, but clearing the entry in `removeProcessGroup` would keep the cache's lifetime tied to the registration it belongs to.

Generated by Barber AI (Claude Opus 5)

server/threads/manageThreads.js
```diff
@@ -922,33 +923,75 @@ const processGroupsByThread = new Map();
...
+	if (isAlive !== false) return true;
+	zombieGroupScanTimes.delete(processGroupId);
+	return false;
+}
+
+function processGroupIsAlive(processGroupId) {
+	return isProcessGroupAlive(processGroupId);
 }
 
 async function waitForProcessGroupExit(processGroupId) {
```
cb1kenobi, line 1110: **Medium** — this loop is still unbounded, and every "can't tell" answer in the new predicate resolves to *alive*.

The PR narrows one specific case (Linux, group where every member is `Z`), which is a real improvement. But the loop it feeds has no deadline and no log line, and the new code has three paths that return `true` when it genuinely does not know:

1. `platform !== 'linux'` → `'unknown'` → `return true` (line 981). macOS also has zombies, so the original failure mode is untouched there — and this loop is reached from component preparation, which is the startup path the PR title is about. Windows takes the `waitForWindowsGroupExit` branch, so it is only macOS that regresses to the old behavior.
2. `readdirSync('/proc')` throws → `scanLinuxProcessGroup` returns `null` → `isAlive !== false` → `true` (line 988). Conservative and correct in isolation, but permanent: a `/proc` that stays unreadable means this polls at 25ms forever.
3. Recycled leader PID reading as live (see the note on line 942).

In each case the symptom is identical to the bug being fixed — a silent, permanent wedge with nothing in the log to point at it. Worth noting `Application.ts`'s `waitForConfirmedTermination` documents its lack of a deadline as a deliberate safety trade (better wedged than releasing the lock over a live descendant), and that reasoning applies here too — so the ask isn't a timeout that gives up.

Suggestion: make it *diagnosable* rather than bounded. A one-shot `harperLogger.warn` once the loop has polled past some threshold (say 30s), naming the pgid and the reason the probe couldn't decide, turns "startup hangs" into a one-line log hit. That's a couple of lines and no behavior change.

Generated by Barber AI (Claude Opus 5)

server/threads/manageThreads.js
```diff
@@ -922,33 +923,75 @@ const processGroupsByThread = new Map();
...
+		try {
+			stat = readStat(`/proc/${processId}/stat`, 'utf8');
+		} catch {
+			continue;
+		}
+		const fields = stat.slice(stat.lastIndexOf(')') + 2).split(' ');
+		if (Number(fields[2]) !== processGroupId) continue;
+		if (fields[0] !== 'Z') return true;
+	}
+	return false;
```
cb1kenobi: **Low** — a single `/proc` snapshot can conclude "all zombies" while a live member exists, and the caller acts on that by *not* terminating.

`readDirectory('/proc')` is one snapshot; each `readStat` happens after it. A member that is live at readdir time, forks, and then exits before its own `stat` is read is observed as `Z`, and its freshly forked sibling was never in the snapshot — so this returns `false` for a group that still has a running process. Narrow (the member has to fork and exit inside the scan window; the ordinary double-fork daemonize case is safe, since the grandchild exists in `/proc` before the parent becomes `Z`), but the consequence is the asymmetric one:

`terminateProcessTree` (`components/Application.ts:2042`) treats `false` as "nothing to do" and returns **without sending SIGTERM/SIGKILL**, and the caller then untracks the group — so a false "dead" leaves a live descendant running, unsupervised, with the component-preparation lock released. That is strictly worse than the hang, which is at least visible.

Given the whole point of this predicate is to be the one place that says "stop waiting", it's worth making `false` the expensive answer rather than the cheap one — require two consecutive all-zombie scans before returning `false`:

```js
if (isAlive !== false) return true;
// Re-scan: the first pass is a snapshot, and a member that forked between readdir and its own
// stat is invisible to it. Only a second all-zombie pass is safe to stop waiting on.
if (scanLinuxProcessGroup(processGroupId, readDirectory, readStat) !== false) return true;
```

The extra scan only ever runs on the path that is about to declare the group dead, so it costs nothing in the common case.

Generated by Barber AI (Claude Opus 5)

unitTests/components/applicationSpawn.test.js
```diff
@@ -222,6 +231,53 @@ describe('nonInteractiveSpawn onLine line buffering', () => {
 		assert.equal(settled, true);
 	});
 
+	it('treats a Linux process group containing only zombies as terminated', () => {
```
cb1kenobi, line 238: **Low** — nice, focused predicate test; noting that the one spec covering *real* process groups is currently dead, so this Linux path has no end-to-end backstop.

This test is well shaped: all I/O is injected, the `/proc` fixture field order (`state ppid pgrp session`) matches the real format, and the four cases pin down the interesting corners — leader `Z` + child `Z`, leader `Z` + child `S`, leader reaped + child `Z`, leader reaped + child `S`. Mutating the state check in `processGroupLeaderState`, the state check in `scanLinuxProcessGroup`, and the `'missing'` → scan transition each turned it red, so the predicate itself is genuinely covered.

The gap is one level up. `terminates a detached process tree when its owning worker is force-terminated` (line 294) — the only spec here that drives real process groups — times out for me on macOS, and it does so **identically on this branch's merge-base**, so it is pre-existing and not something this PR broke:

```
HEAD:  12 passing, 1 failing  (terminates a detached process tree ... : Timeout of 30000ms exceeded)
BASE:  11 passing, 1 failing  (same test, same timeout)
```

Two things that seem worth knowing while this PR is in flight:

1. `await once(harness, 'close')` never settles because the worker's detached descendant survives and keeps the harness's inherited stdout pipe open — which is the exact failure mode this PR is about, just on the platform the fix skips.
2. Every run leaks an orphaned `unitTests/components/fixtures/processGroupHarness.js` reparented to PID 1. I found several already resident on this machine from earlier runs across different worktrees, so it accumulates.

Not a blocker for this PR, and CI is `ubuntu-latest`-only so it won't show up there. Raising it because it means the new Linux behavior is verified only through injected `/proc` fixtures, and the one test that would exercise it against a real kernel isn't currently able to report.

Generated by Barber AI (Claude Opus 5)

unitTests/components/applicationSpawn.test.js
```diff
@@ -222,6 +231,53 @@ describe('nonInteractiveSpawn onLine line buffering', () => {
...
+				platform: 'linux',
+				processGroupExists: () => true,
+				readDirectory: () => ['456'],
+				readStat: (path) => {
+					if (path === '/proc/123/stat') throw new Error('leader reaped');
+					return '456 (running child) S 1 123 123';
+				},
+			}),
+			true
+		);
```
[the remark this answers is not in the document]
kriszyp: Addressed in 0fdade42a: the boolean liveness assertions now use assert.strictEqual. — GPT-5 Codex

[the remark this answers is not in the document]
kriszyp: Addressed in 0fdade42a: the boolean liveness assertions now use assert.strictEqual. — GPT-5 Codex

server/threads/manageThreads.js
kriszyp: Addressed in 0fdade42a: the throttle now uses the same injectable options path as the procfs probes, including an injected clock. The regression covers a cached result inside the interval and a rescan at expiry. — GPT-5 Codex

server/threads/manageThreads.js
kriszyp: Addressed in 0fdade42a: leader status now verifies the parsed process-group ID before trusting a live PID, and unregister clears the lifecycle-bound throttle state. Both cases have focused regressions. — GPT-5 Codex

server/threads/manageThreads.js
kriszyp: Addressed in 0fdade42a: an all-zombie result now requires two consecutive procfs snapshots; a live member appearing in the second snapshot keeps the group alive. — GPT-5 Codex

components/RuntimeModuleTracker.ts
```diff
@@ -103,6 +128,47 @@ export class RuntimeModuleTracker {
...
+	resolvedPath: string
+): string[] | undefined {
+	if (!specifier.startsWith('.')) return;
+	const basePath = resolve(dirname(referrerPath), specifier);
+	const candidates = [...new Set(RESOLUTION_EXTENSIONS.map((extension) => basePath + extension))];
+	candidates.push(resolve(basePath, 'package.json'));
+	for (const extension of RESOLUTION_EXTENSIONS.slice(1)) candidates.push(resolve(basePath, `index${extension}`));
+	const resolvedIndex = candidates.indexOf(resolvedPath);
+	if (resolvedIndex === -1) return;
+	return candidates.slice(0, resolvedIndex);
```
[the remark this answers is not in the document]
kriszyp: Resolved as outdated: RuntimeModuleTracker is no longer part of the current PR diff, so the candidate-detection path described here is not shipping in this change. — GPT-5 Codex

unitTests/components/applicationSpawn.test.js
kriszyp: Confirmed as a pre-existing macOS-only fixture failure and orphan leak, unchanged by this Linux procfs fix. I recorded it for follow-up rather than expanding this PR into cross-platform process supervision. — GPT-5 Codex

server/threads/manageThreads.js
```diff
@@ -922,33 +923,77 @@ const processGroupsByThread = new Map();
...
+	}
+	const leaderState = processGroupLeaderState(processGroupId, platform, readStat);
+	if (leaderState === 'alive' || leaderState === 'unknown') return true;
+	const scanTime = options?.now?.() ?? performance.now();
+	const lastScan = zombieGroupScanTimes.get(processGroupId);
+	if (lastScan !== undefined && scanTime - lastScan < ZOMBIE_GROUP_MEMBER_SCAN_INTERVAL_MS) return true;
+	zombieGroupScanTimes.set(processGroupId, scanTime);
+	const isAlive = scanLinuxProcessGroup(processGroupId, readDirectory, readStat);
+	if (isAlive !== false) return true;
+	if (scanLinuxProcessGroup(processGroupId, readDirectory, readStat) !== false) return true;
```
cb1kenobi: **Low** — the second scan narrows the snapshot race rather than closing it, because both scans share one `/proc` sampling pattern with nothing separating them in time.

Thanks for taking the double-scan suggestion — and the placement is right: this only runs on the path that is about to declare the group dead, so the common case pays nothing. Confirmed by mutation that it is genuinely covered (deleting this line turns `requires two all-zombie scans...` and the throttle spec red: 13 passing, 2 failing).

The residual is that scan 2 is structurally identical to scan 1, and runs microseconds later:

- scan 1: `readdir('/proc')` at T0, `stat` each at T1. Member `M` alive at T0 forks `C` at T0.5, exits by T1 → `M` reads `Z`, `C` was never in the T0 snapshot → `false`.
- scan 2: fresh `readdir` at T2 now *does* include `C`, which is what makes this a real improvement.
- but scan 2 has its own T2→T3 window: if `C` is alive at T2, forks `E` at T2.5 and exits by T3, then `C` reads `Z` and `E` is not in the T2 snapshot → `false` again.

So the race has to hit twice consecutively instead of once — much less likely, but the same shape, and back-to-back synchronous calls give the second sample almost no chance to differ from the first. That matters here specifically because `false` is the answer that makes `terminateProcessTree` (`components/Application.ts:2042`) return **without** sending SIGTERM/SIGKILL, after which the caller untracks the group.

I'm *not* suggesting adding a delay between the two scans: `isProcessGroupAlive` is deliberately synchronous, and Application.ts consumes it through a sync `processGroupIsAlive(): boolean` used in a plain `if` (2042), a `while` condition (1927), and a sync callback into `waitForConfirmedTermination` (2057). Making it async to fit a `delay()` in would ripple through all of those for a probabilistic gain — not a good trade.

The cheap fix is to say what the second scan does and doesn't buy, so the next reader doesn't take it as a closed window:

```js
if (isAlive !== false) return true;
// Second snapshot: a member that forked between readdir and its own stat is invisible to the
// first pass but present in this one. Narrows that window rather than closing it — both passes
// have the same readdir/stat gap, so a member forking and exiting inside each one still reads
// as all-zombie. Kept sync (and therefore back-to-back) on purpose: the callers in
// Application.ts are sync, and awaiting here would ripple through all of them.
if (scanLinuxProcessGroup(processGroupId, readDirectory, readStat) !== false) return true;
```

Genuinely better than the single snapshot — I'd only push back on treating the window as eliminated. If it's worth hardening further later, the durable answer isn't more `/proc` sampling but making the false answer non-load-bearing: having `terminateProcessTree` signal the group unconditionally before trusting a "dead" verdict, since `kill(-pgid, SIGTERM)` on an already-dead group is a harmless `ESRCH` that it already swallows.

Generated by Barber AI (Claude Opus 5)

unitTests/components/applicationSpawn.test.js
```diff
@@ -222,6 +235,129 @@ describe('nonInteractiveSpawn onLine line buffering', () => {
 		assert.equal(settled, true);
 	});
 
+	it('treats a Linux process group containing only zombies as terminated', () => {
+		let now = 0;
```
cb1kenobi, line 239: **Low** — with the `if (!testOptions)` gate gone, the injected fake clock now writes into module-level state that outlives the test, and this spec leaves an entry behind.

This is the right shape of fix for the coverage gap — dropping the gate is exactly what makes the throttle mutable-and-caught (unconditional `return true` in place of the throttle check now gives 11 passing / 4 failing, where it was 12 / 0 before). One side effect worth closing while it's cheap.

`zombieGroupScanTimes` is module-level, and the throttle no longer distinguishes injected clocks from `performance.now()`. The last assertion in this spec returns `true`, so the entry for pgid `123` is *not* deleted and persists at the fake value `5000` after the test finishes. Any later spec that reuses pgid `123` without injecting `now` then compares a real `performance.now()` against that stale fake value, and `elapsed` goes negative — which always satisfies `< ZOMBIE_GROUP_MEMBER_SCAN_INTERVAL_MS`:

```
fake-clock call -> true (leaves zombieGroupScanTimes[123] = 5000)
real-clock call -> true  scans performed: 0  performance.now() ~ 75
```

i.e. `75 - 5000 = -4925 < 1000` → throttled, zero `/proc` scans, reports **alive** for a group whose fixtures say all-zombie. Today the suite is green only because the other specs happen to use `234`/`345`/`567`; it's a trap for the next test that picks `123`.

`unregisterProcessGroup` is already exported here for exactly this, so the fix is one line at the end of the spec (or an `afterEach` covering the pgids these specs touch):

```js
unregisterProcessGroup(123);
```

Optional belt-and-braces on the production side: `scanTime >= lastScan && scanTime - lastScan < ZOMBIE_GROUP_MEMBER_SCAN_INTERVAL_MS` makes a non-monotonic clock fail open to a rescan instead of to "alive". Not needed for real `performance.now()`, which is monotonic per process — only relevant because the seam now admits arbitrary clocks.

Generated by Barber AI (Claude Opus 5)

server/threads/manageThreads.js
cb1kenobi: Re-verified at `0fdade42` — this is fixed, closing the loop on the table above.

The gate is gone, `now` is injectable, and the seam is renamed to `options`, so the throttle now executes under test. Re-ran the one mutation that survived last time (throttle check → unconditional `return true`), same method as before: `tsc -p tsconfig.build.json`, marker confirmed as exactly 1 occurrence in `dist/server/threads/manageThreads.js`, `mocha unitTests/components/applicationSpawn.test.js --timeout 20000 --grep force-terminated --invert`.

| mutation | before | now |
|---|---|---|
| throttle check → unconditional `return true` | 12 passing, 0 failing ❌ | **11 passing, 4 failing ✅** |
| drop `Number(fields[2]) !== processGroupId` | n/a | 14 passing, 1 failing ✅ |
| drop the second `scanLinuxProcessGroup` | n/a | 13 passing, 2 failing ✅ |
| drop `zombieGroupScanTimes.delete` in `removeProcessGroup` | n/a | 14 passing, 1 failing ✅ |

Control run with no mutation is 15 passing / 0 failing. Every guard this commit adds is now load-bearing on a test — nice work, that was the awkward one to get a seam around.

Generated by Barber AI (Claude Opus 5)

kriszyp:
@heskew you had marked this as a P0. Is that accurate and worth a review?

server/threads/manageThreads.js
kriszyp: Addressed in f55a8fe67: indeterminate procfs reads now retain the group, and continued liveness emits one warning after 30 seconds with the pgid and current probe reason. The wait remains deliberately unbounded so the preparation lock is never released over a possibly-live descendant.

server/threads/manageThreads.js
kriszyp: Addressed in f55a8fe67: the inline invariant now says the second synchronous snapshot narrows, rather than closes, the shared readdir/stat race.

unitTests/components/applicationSpawn.test.js
kriszyp: Addressed in f55a8fe67: the affected pgid cache entries are explicitly cleared after the tests, and a non-monotonic injected clock now forces a rescan instead of treating the stale timestamp as current.

server/threads/manageThreads.js
cb1kenobi: Re-checked at `f55a8fe67` — resolved.

`f55a8fe67` ("Harden indeterminate process group checks") adds exactly the diagnosability fix suggested here: `keepProcessGroupAlive()` + a module-level `processGroupLivenessStates` map now warn once via `harperLogger.warn` once a group's termination has been unconfirmed for `PROCESS_GROUP_LIVENESS_WARNING_MS` (30000ms), naming the pgid and the reason (unsupported platform, unreadable `/proc`, or an observed live member). No behavior change — same "startup hangs" symptom now gets a one-line log hit.

Covered by the new test `warns once when process-group termination remains unconfirmed`. Verified passing at this head: 17/0 in `unitTests/components/applicationSpawn.test.js` (excluding the pre-existing macOS-hanging `force-terminated` spec you flagged in the sibling thread, which is unrelated to this change and still reproduces identically on main).

—
_Generated by Barber AI_

server/threads/manageThreads.js
cb1kenobi: Re-checked at `f55a8fe67` — resolved.

The residual is now documented in-line, directly above the second `scanLinuxProcessGroup` call (`manageThreads.js:1063-1064`):

> A second snapshot catches a member omitted after forking inside the first readdir/stat window. It narrows rather than closes the race because the second synchronous pass has the same gap.

No behavior change — just closes the documentation gap this comment raised, so a future reader doesn't mistake the double-scan for a closed race.

—
_Generated by Barber AI_

unitTests/components/applicationSpawn.test.js
cb1kenobi: Re-checked at `f55a8fe67` — resolved.

`unregisterProcessGroup(123)` is now called at the end of the previously-leaking spec, and the two new specs this push adds (pgid 678, 789) clean up the same way. Checked every `isProcessGroupAlive(...)` call site in the current test file: every spec that leaves the predicate returning `true` (and therefore leaves an entry in `processGroupLivenessStates`/`zombieGroupScanTimes`) now calls `unregisterProcessGroup` on its own pgid before the spec ends; specs that return `false` self-clean via `clearProcessGroupLivenessState`, which now runs on that path internally (e.g. pgid 234, 567). No asymmetry left to leak from.

—
_Generated by Barber AI_

cb1kenobi (approved):
LGTM, nice work!

PR #2086, opened 2026-08-05
Don't poison a transaction parked in its commit phase (it destroyed the deploy payload blob)

### LINKED ISSUE #2062: Over-time transaction abort deletes the pre-saved deploy payload blob, leaving later deploys referencing a destroyed blob (regression from #1411, 5.1.21)
## Summary

An over-time write transaction that gets **aborted** deletes the payload blob it pre-saved. A subsequent deploy of the same component can then end up referencing that already-deleted blob, so the payload is permanently unreadable **at the source**. Replication reports the blob as unrecoverable, the peer's deploy fails immediately, and no retry can help — the bytes are gone, not late.

This became reachable in **5.1.21**: #1411 (#1407) changed over-time write transactions from *force-commit* to *abort*. On 5.1.19 the same slow write was force-committed and the payload survived.

## Observed

Deploying a component with a `payload_size` of 95,783,483 bytes (~91 MiB) on 5.1.26.

About three minutes before the failing deploy, the origin logged:

```
[error]: Transaction was open too long and has been aborted after exceeding the
open-transaction limit, from table: hdb_deployment/ was started from hdb_deployment.put
```

Then, during the deploy, on the origin:

```
[warn] [replication]: Error sending blob BlobReadError: Blob file not found for
  <root>/blobs/system/0/0/37
```

and on the peer:

```
[error] [replication]: Blob 37 for record <deployment-id> is unrecoverable at source <origin>
[error] [replication]: Sustained blob replication divergence from <origin>:
  5 blob saves have failed on this connection
[error]: Error calling operation: deployComponent
  error: 'Blob file not found for <root>/blobs/system/0/0/c8'
```

The deploy surfaced to the caller as:

```
Component '<name>' was deployed on the origin node but failed to replicate to
1 of 1 peer node(s): <peer> (Error: ServerError: Blob file not found for
<root>/blobs/system/0/0/c8)
```

Both referenced blob files were confirmed **absent on disk** on the origin (23 files present in that directory, neither `37` nor `c8` among them). Exactly one over-time abort occurred in the surrounding six hours — the one quoted above.

## Mechanism

1. The deploy pre-saves the payload as a blob, then writes the `hdb_deployment` row in a transaction.
2. A ~91 MiB payload write can exceed `storage_maxTransactionOpenTime` (default **30000 ms**, read at module load in `resources/DatabaseTransaction.ts`). The long-transaction monitor then aborts and poisons the transaction.
3. `DatabaseTransaction` tracks "blobs that were pre-saved as part of this write; used to clean up files if the commit is skipped or aborted" — so the abort **deletes the pre-saved payload blob**.
4. A later deploy of the same component still references that blob id (same payload content). Its bytes no longer exist, so every read fails permanently.
5. The peer-side retry added for this path (`readPayloadBlobWithRetry`, `components/deploymentRecorder.ts`, Refs #641) only retries the *retryable* 503 stall; 500-corrupt / 404-gone fail immediately by design. So nothing retries this, correctly — the problem is that the blob was destroyed, not that the read was early.

Note `resources/LMDBTransaction.ts` carries its own hard-coded 30000 with no config read, so the timeout is not tunable on that path.

## Expected

Some combination of:

- An abort should not delete a pre-saved blob that a still-live or subsequently-written record references — or the deployment row that references it must be invalidated together with the blob, so a later deploy cannot inherit a dangling reference.
- A deploy payload write should not hold a single transaction across a multi-tens-of-MB blob write. The abort error itself advises "split long-running work into smaller transactions"; the deploy path does the opposite.
- If a payload blob is known-destroyed, the deploy should fail with that cause ("payload blob was discarded by an aborted transaction") rather than a bare `Blob file not found`, which reads as corruption or a race.

## Where

- `resources/DatabaseTransaction.ts` — over-time abort; pre-saved blob cleanup; `storage_maxTransactionOpenTime` (default 30000).
- `resources/blob.ts` — blob read / error classification.
- `components/deploymentRecorder.ts` — payload write and peer-side retry.
- #1411 / #1407 — the abort-instead-of-force-commit change that made this reachable (5.1.21).

## Related

- #641 — Deployment tracking, replicated payload delivery, and rollback.
- #524 — support for deployments larger than 2GB (same large-payload pressure).
- #1472 — missing source blob wedging `deploy_component` on peers. Different path: that one *hangs* during full-resync because the receiver never releases the pending read. Here the permanent failure is detected and propagated correctly and the deploy fails fast; the defect is upstream, in the blob being deleted at all.
- The peer is additionally left with an empty component directory in this scenario — filed separately.

Filed from a field incident; cluster, host and component identifiers omitted.

kriszyp:
Fixes #2062.

## What was wrong

A blob is saved in the commit's pre-commit phase, so a write carrying one stays inside `commit()` until the file has landed. For a multi-tens-of-MB deploy payload that is well past `storage.maxTransactionOpenTime`, and the long-transaction monitor poisoned the transaction right there.

That was worse than the issue's original diagnosis. `abort()` cleared the write set and unlinked the write's pre-saved blobs — and then the in-flight commit resumed, found nothing left to write, and **resolved as success**. So:

- the caller was told its write landed when it had been dropped (silent write loss, and no error for `ingestPayload`'s cleanup path to catch), and
- it was left holding a blob instance whose file was gone but whose `fileId` was still set. `saveBlob` short-circuits on a set `fileId`, so the deploy recorder's next `put` of the same record object silently minted a reference to a destroyed file — surfacing at the peer as `Blob file not found`, unrecoverably.

Reproduced end to end in `integrationTests/database/blob-commit-over-time.test.ts`: on `main` the HTTP write returns **200** and the record is a **404**.

## The fix

The open-transaction limit polices an *application* holding a transaction open with an unfinished write set (#1407). Once `commit()` is entered the write set is sealed and the caller is awaiting the commit, so the time spent in the pre-commit phase is core's own I/O — not the thing the limit exists to stop.

1. **The monitor spares a transaction parked in its commit phase** rather than poisoning it. Bounded: `COMMIT_PHASE_GRACE` over-limit ticks (~10 min at the 30s default, since sparing re-arms `timeout`), because the transaction still pins a read snapshot — a source that stalls rather than finishing falls through to the normal abort. `sourceApply`/`isReplay` are spared without a bound: they may be neither aborted (harper-pro#348) nor force-committed mid-write, and their blob sources are bounded by the receive-side idle watchdog.
2. **Resuming from the pre-commit await re-checks that the transaction is still alive** — `timedOut` (monitor poison, including via the multi-store chain) or a write set cleared with the handle released (a plain `abort()`). Either throws instead of resolving as a phantom commit.
3. **A blob whose file has been unlinked is tombstoned** — a *shared* `storageInfo.fileState.discarded`, so a slice is condemned along with its source — and `saveBlob` refuses to re-store it. Rebasing onto main's deferred-reclamation machinery split that setter in two, which is the design outcome worth reading: `cleanupUnusedBlobs` marks at **decision** time, before an in-flight save settles, because the unlink waits on that save and a re-store in the window would otherwise win — that is the window the #2062 deploy payload took; `runReclamation` marks every instance queued for the path only once it **claims** the unlink, after the retention window has had its chance to rescue the file via `cancelBlobReclamation`. Marking only at reclamation time misses the #2062 window; marking in `deleteBlob` would condemn blobs a legitimate re-reference rescues. `DESIGN.md:54` carries the same split.

LMDB carries the same three changes.

## Also fixed along the way

`sourceApply`/`isReplay` transactions parked in the pre-commit phase were previously **force-committed** on every monitor tick — durably committing a replica record while its blob file was still being written. That predates this change (it's the monitor's `else` branch); the new unit test fails on `main` for exactly that reason.

## Verification

- `unitTests/resources/txn-tracking.test.js` — eight cases in the `#2062` block: the commit completes and its blob survives; the commit phase is marked and cleared across a chain; every multi-store link stays alive while the head waits on its blob save; a chain whose *later link* exhausts the grace still aborts from the head (atomically, with the grace counted once per chain per tick); poisoned-while-parked throws; plain-abort-while-parked throws; a source apply is neither poisoned nor force-committed; the grace is bounded. Pass on **both** engines (`test:unit:resources` and `test:unit:lmdb`).
- `unitTests/resources/blob.test.js` — a discarded blob cannot be silently re-stored.
- `integrationTests/database/blob-commit-over-time.test.ts` — end to end over REST with the limit lowered and the blob source trickled; asserts the record commits, the blob reads back complete, no abort line, and the "in its commit phase past the open-transaction limit" line (so the test cannot pass vacuously). Fails on `main`.
- `npm run test:unit:resources`, the deploy integration suite (59/59), and the two existing #1407/#1411 anchors — `longtxn-secondary-index` (QA-176) and `aborted-update-index-migration` (QA-616) — all pass, i.e. the abort guarantee for application-held transactions is unchanged.
- Re-verified at `97821ba7`: clean `npm run build`, and the two unit files above pass on both engines (125 RocksDB / 120 LMDB, 0 failing). A full local `test:unit:resources` is 1664 passing with 4 failures in `replayStructures` / caching / `randomAccessFields` — none touched by this diff, and all 42 CI checks are green on this same head.
- `test:unit:main` can't run on this machine (the local Harper install holds `~/harper/database/system/LOCK`); it's untouched by this change and covered by CI.

## Where to look hardest

- **The grace bound.** `COMMIT_PHASE_GRACE = 10` is a judgment call. Too low and a genuinely huge payload on a slow link gets aborted (cleanly now, but aborted); too high and a stalled pre-commit source pins a read snapshot for longer. The unbounded sparing of `sourceApply` rests on the replication receive path arming `blobStreamIdleTimeoutMs` — worth a second opinion.
- **`DatabaseTransaction.ts` resume guard** (`stagedWrites > 0 && this.writes.length === 0 && this.open === TRANSACTION_STATE.CLOSED`). It must not false-positive on the retry / immediate-commit / replay paths. I could not find a caller that plain-`abort()`s a transaction already parked in commit, so that half is a latent trap being closed rather than a live bug — the test drives it directly.

## Open / not addressed

- No end-to-end coverage through the actual `deploy_component` path: the deployment `put` only registers with the monitor when a read precedes the write in the same transaction (`getReadTxn` early-returns once a write created the handle), which the clustered field case did and a single-node harness does not. The integration test drives the same mechanism through a component resource instead.
- Two field initializers per transaction construct (`committing`, `commitPhaseTicks`) — deliberate; adding them dynamically would cost a hidden-class transition on the commit path.
- **Pre-existing, recorded here rather than fixed:** if the pre-commit work sits on the *second* database instead of the chain head, the chain still tears (`A=COMMITTED B=absent`) — the head's native commit lands before the cascade reaches `this.next.commit()`. At the merge base it tears the same way and the caller gets **no error at all**, so this change strictly improves it (silent tear → loud `transactionOpenTooLongError`). Out of scope here; worth its own issue.
- **Cross-PR, nothing to do here:** neither #1835's `enqueueBlobUnlink` nor #2265's backup walk exists in this merge base, so there is no live interaction with the decision-time tombstone yet. If #1835 later moves its durable row earlier into `deleteBlob()`, it will need to record the cancellation when `cancelBlobReclamation()` rescues a blob inside the retention window, or a legitimately re-referenced blob gets unlinked after a restart. That is #1835 against main's reclamation machinery; this PR neither creates nor worsens it.

Generated with Claude Opus 5. Cross-model review: codex + gemini + Harper-domain adjudication, four rounds — round 4 verdict `COMMENTS`, no majors outstanding. All review threads are resolved; the last re-review (`97821ba7`) reported no issues, with each prior finding's fix re-confirmed by mutation after the rebase.

<sub>Review-Coverage: authored=unknown; ran=none; rounds=1 @ 97821ba70e9b</sub>

<sub>Human-Review-Need: 4 @ 97821ba70e9b</sub>


kriszyp:
## Reading order

**1. `integrationTests/database/blob-commit-over-time.test.ts` first.** It is the shortest statement of the bug. On `main` the HTTP write returns **200** and the record is a **404** — the caller is told its write landed and it wasn't. Everything else in the diff exists to make that impossible.

**2. `resources/DatabaseTransaction.ts:609-625`** — the pre-commit await, which is where the whole thing happens. `commit()` parks here on `Promise.all(completions)` (the blob's durable file write). Two things were true and neither was obvious: the monitor could abort the transaction *while it was parked*, and the resumed continuation didn't notice — it walked on with an empty `writes` array and resolved successfully. That is the silent write loss, and it is also why `ingestPayload`'s cleanup path never ran in the field: it only fires on a rejection.

**3. `resources/DatabaseTransaction.ts:1074`** — the monitor branch. This is the design decision worth arguing with, so it's the one I'd read most carefully.

**4. `resources/blob.ts:869 / 930 / 1717`** — the tombstone. Third line of defence; see below.

## Decisions a reviewer should question

**Exempting the commit phase at all.** #1411 made over-time write transactions abort rather than force-commit, and this walks part of that back. My argument: the limit's premise is "the application is holding a transaction open with an unfinished write set" — once `commit()` is entered that premise is false. There is no partial write set to protect, the caller is already awaiting, and the abort's only effects were to drop a sealed write and unlink blobs the write still referenced. The #1407/#1411 anchors (QA-176 `longtxn-secondary-index`, QA-616 `aborted-update-index-migration`) both still pass, which is the evidence that the application-held case is untouched.

**Why the grace is bounded, and why `sourceApply` isn't.** A parked transaction still pins a read snapshot, so an unbounded exemption turns a stalled upload into a snapshot-pinning vector — hence `COMMIT_PHASE_GRACE`. `sourceApply`/`isReplay` are the exception because *both* of the monitor's other options are wrong for them: aborting drops a canonical write with no resume path (harper-pro#348), and the force-commit branch they used to land in durably commits a replica record whose blob file is still being written. That force-commit is pre-existing — the new source-apply test fails on `main` for that reason, not because of anything else in this PR.

**Why the blob tombstone exists at all, given 1 and 2.** With the monitor no longer poisoning a committing transaction and the resume guard catching the rest, the destroyed-blob path should be unreachable. The tombstone is there because "should be unreachable" is exactly what the pre-#2062 code believed. It converts any residual path from a silently-committed reference to a destroyed file into a loud error naming the cause. It is deliberately set when the deletion is *decided*, not when the unlink is issued — `cleanupUnusedBlobs` waits for an in-flight save to settle first, and a re-store in that window was still winning.

## What the tests do and do not prove

They prove: the write commits and its blob is readable when the save outruns the limit (both engines, plus end to end over REST); a poisoned or plain-aborted parked transaction throws instead of phantom-committing; a source apply is neither poisoned nor force-committed; the grace is bounded; a discarded blob cannot be re-stored.

They do not prove: anything through the real `deploy_component` path. The deployment `put` only registers with the long-transaction monitor when a read precedes the write in the same transaction (`getReadTxn` early-returns once a write has created the handle) — the clustered field case did, a single-node harness does not. The integration test drives the identical mechanism through a component resource instead. Also untested: orphaned partial-file cleanup after the grace-exhausted abort (existing abort-cleanup behaviour, unchanged here).

## Coverage

Four rounds of cross-model review (codex graded leg + gemini + Harper-domain adjudication). Rounds 1–3 each found something real and each is a commit in this branch: the unbounded exemption, the plain-abort phantom commit, the `sourceApply` force-commit, the too-late tombstone, and three unit tests that would have gone red on the LMDB CI leg (they waited on the head of the transaction chain, which on LMDB never owns the table's writes). Round 4 returned `COMMENTS` with no majors. Outstanding and accepted: two field initializers per transaction construct, deliberate over a dynamically-added property.

_— KrAIs (Claude Opus 5)_

kriszyp:
CI status: 37 pass, 4 fail — **all four failures are pre-existing on `main`**, not from this change.

- `Integration Tests 6/6` (Node 24 / Windows / uWS) — `txnlog-purge-stale-read-blast.test.ts:683`, "purge job should COMPLETE, got ERROR: Table-level transaction log deletion is not supported for RocksDB tables". The identical assertion fails on `main`'s own run ([30956731137](https://github.com/HarperFast/harper/actions/runs/30956731137), shard 5/6). Adding a test file shifted which shard it lands in, hence 6/6 here vs 5/6 there.
- `Integration Tests 2/6 (Bun)` — `redeploy-runtime-equivalence.test.ts:392`. Also failing on `main` (Bun shard 1/6), along with three sibling cases in the same file.

This PR's own coverage is green: all three `Unit Test` legs pass, and the new integration test ran and passed on Linux and Windows —

```
▶ Blob save that outruns the open-transaction limit (#2062)
  ✔ the write commits and its blob survives (2034ms)
```

Staying in draft until the pre-existing `main` breakage clears (or a reviewer is happy to look past it).

_— KrAIs (Claude Opus 5)_

cb1kenobi (approved):

resources/DatabaseTransaction.ts
```diff
@@ -1187,6 +1215,28 @@ function startMonitoringTxns() {
...
+					// force-commit below — which would durably commit a record whose blob file is still being
+					// written — is acceptable, and their blob sources are bounded by the receive-side idle
+					// watchdog instead.
+					harperLogger.warn?.(
+						`Transaction has been in its commit phase past the open-transaction limit, waiting on pre-commit work (e.g. a large blob write); letting it complete, from table: ${
+							(txn.db as any)?.name + (url ? ' path: ' + url : '')
+						}`,
+						...(txn.startedFrom ? [`was started from ${txn.startedFrom.resourceName}.${txn.startedFrom.method}`] : [])
+					);
+					txn.timeout = txnExpiration;
```
cb1kenobi: **High — the commit-phase grace spares only the head, so a multi-store chain tears.**

`txn.timeout = txnExpiration` re-arms exactly one object. A multi-store transaction's other databases live on `.next` links (`Table.ts:5606-5634`), each tracked in `trackedTxns` in its own right. Nothing propagates `committing` to them, and `chainStillActive()` only walks *forward*, so a link has no way to see that its head is parked.

The link therefore falls through to `abortDueToTimeout()` on its own schedule — and that walks `for (let txn = this; txn; txn = txn.next)`, i.e. forward only, so the head is never reached and `head.timedOut` stays false. When the head resumes, neither new guard fires (`this.timedOut` false; `this.writes.length === 0 && !this.transaction` false, since the head's own writes are intact), so **database A commits durably**. Then the cascade at line 877 calls `this.next.commit(...)`, which throws `transactionOpenTooLongError` synchronously from line 678.

Net: A committed, B's writes dropped and its blobs unlinked, and the caller sees an error that reads as a full rollback. The synchronous throw at 877 also skips lines 903-911, so `clearWrites()` / `releaseContext()` / `this.next = null` never run and the context keeps pinning a CLOSED head.

Reproduced directly against this head (`COMMIT_PHASE_GRACE = 10`, 20ms expiration, head and link both tracked with writes):

```
head.timedOut = false   head.commitPhaseTicks = 4   head.writes.length = 1
next.timedOut = true    next.writes.length = 0
```

Control with `head.committing = false` (the pre-PR shape) passes — `head.timedOut = true, next.timedOut = true`, the chain dies atomically. So the tear is introduced by the grace, not pre-existing.

**Suggested fix** — spare the links alongside the head, since they carry the other half of the same sealed write set (lines 692-695):

```ts
if (completions.length > 0) {
	for (let link: DatabaseTransaction = this; link; link = link.next) {
		link.committing = true;
		link.commitPhaseTicks = 0;
	}
}
```

and clear the same way at lines 699 and 1028. `LMDBTransaction.ts:154-157/177/181` needs the same, and it matters more there — on LMDB the parked transaction is always a `.next` link, as `committingTxn()` in the test helper already acknowledges.

A regression test writing two databases in one context, with the head's blob save outrunning the limit, would cover this — the current suite is thorough on the single-store shapes but has no chain case.

**Minor, same line:** this re-arm uses bare `txnExpiration`, while every other re-arm site (402, 595, 1245, 1281) uses `Math.max(txnExpiration, txn.timeoutBudget ?? 0)`. `timeoutBudget` is live for exactly the motivating workload (`deploymentRecorder.ts:101` sets it from `deployment_timeout`), so a deploy ingest burns its grace at the global 30s cadence rather than its configured budget. Worth either matching the other sites or a one-line note that the downgrade is deliberate.

---
*Generated by Barber AI (claude-opus-5)*

resources/blob.ts
```diff
@@ -2294,6 +2311,10 @@ export function cleanupUnusedBlobs(blobs: Blob[] | undefined, retainedFileIds?:
 		const storageInfo = storageInfoForBlob.get(blob);
 		if (!storageInfo?.fileId || (blob as FileBackedBlob).saveInRecord) continue; // no file written, nothing to clean up
 		if (retainedFileIds?.has(storageInfo.fileId)) continue; // the committed record still references this blob
+		// Tombstone the instance as soon as the deletion is DECIDED, not when the unlink is issued: the
+		// unlink waits for an in-flight save to settle, and a re-store in that window would otherwise
+		// mint a reference to a file that is already condemned (issue #2062).
+		storageInfo.discarded = true;
```
cb1kenobi: **Medium — this tombstone has no test coverage, and the rebase is why.**

Mutation testing against this head (built to `dist/`, marker verified present exactly once in `dist/resources/blob.js` before each run):

| mutation | result |
|---|---|
| replace this line's `storageInfo.discarded` with a dead property | **survives** — 116 passing, 0 failing |
| same mutation on the `runReclamation()` tombstone (line 1227) | fails `#2062: a blob whose file was deleted cannot be silently re-stored` |

A probe confirms the line *executes* 10 times across `blob.test.js` + `txn-tracking.test.js` — so it is reached, but nothing asserts on its effect.

The cause is the rebase. Before it, `deleteBlob()` set `discarded` synchronously, so the `#2062` blob test — which calls `deleteBlob(blob)` directly, waits for the unlink, then re-puts — exercised the decision-time semantics. Now the marking lives in `runReclamation()`, and that same test passes by waiting for the file to disappear, i.e. it now covers the *reclamation-time* path instead. The test text didn't change, so the move is invisible in review, but what the test proves did change.

This matters because `cleanupUnusedBlobs` is the path the actual #2062 deploy-payload bug takes (abort → `cleanupUnusedBlobs` → in-flight save settles → `deleteBlob`), and its whole point is the window the comment above names: the unlink waits for `storageInfo.saving` to settle, and a re-store in that window must fail. The reclamation-time tombstone cannot cover that window — it fires much later, after the retention delay.

**Suggested fix:** add a case that pins the decision-time semantics specifically — register a blob via `startPreCommitBlobsForRecord`, call `cleanupUnusedBlobs` while `storageInfo.saving` is still pending, and assert the re-store throws *before* the file is gone. That distinguishes it from the existing test, which only asserts after the unlink lands.

---
*Generated by Barber AI (claude-opus-5)*

resources/blob.ts
```diff
@@ -1217,6 +1222,9 @@ function runReclamation(): void {
 		// Keep the entry until the unlink lands so a concurrent re-reference can tell that the file is
 		// already going away instead of silently adopting a doomed path.
 		pending.unlinking = true;
+		// Once reclamation has claimed this file, a later write must not preserve its soon-to-be-deleted
+		// fileId. The retention window above still permits legitimate re-references before this point.
+		if (storageInfo) storageInfo.discarded = true;
```
cb1kenobi: **Low — moving the tombstone here drops the guarantee for the one direct `deleteBlob` caller.**

To be clear up front: the *direction* of this move is right, and worth stating. `pendingReclamation` reaching this point already means `consumeRereferenced()` (line 1172) and `snapshotStillSees()` came back negative, so marking here can't poison a file that a legitimate re-reference rescued. Pre-rebase placement in `deleteBlob` would have thrown in two cases the new machinery handles correctly: a re-reference inside the retention window (`saveBlob` → `cancelBlobReclamation`, line 2409, revives the file), and a hold-table hash collision making `consumeRereferenced` return true for an unrelated file (line 930-932's deliberate over-retention). So the move fixes false positives.

The regression is narrower. `pendingReclamation` is keyed by **file path**, but the tombstone lands on `storageInfoForBlob.get(pending.blob)` — a single instance. `deleteBlob` reuses the queued entry at line 1121:

```ts
const pending = pendingReclamation.get(filePath) ?? { blob, deadline: 0, enqueuedAt: now, supersededAt: now };
```

so the second and later `deleteBlob()` calls for the same path discard their `blob` argument. Pre-rebase, `deleteBlob` marked whatever it was handed, so every instance got marked.

The reachable shape is `components/deploymentRecorder.ts:353-362` — the caller both this file's comment (line 67-70) and `DESIGN.md:54` name as the motivating case:

1. `this.record.payload_blob = null; await this.put(...)` → `RecordEncoder.ts:796` `deleteBlobsInObject(existingEntry.value, undefined)` → `deleteBlob(D)`, where `D` is the msgpackr-unpacked instance (a fresh `new FileBackedBlob()` per decode, line 2350). Entry queued with `blob = D`.
2. `deploymentRecorder.ts:362` `deleteBlob(blob)` — the recorder's own instance `A`, same fileId and path — hits the `??` reuse and is **never tombstoned**. `A !== D` unconditionally, since `this.record` is the recorder's own object literal and never the cached decode.

Benign today (`payload_blob` was already nulled and the function throws immediately after, so `A` is never re-put), but it is the only direct `deleteBlob` call site outside this file, so the guarantee the comment advertises for it is gone.

**Suggested fix** — track the instances, keeping the rescue window that motivated the move:

```ts
// PendingReclamation: blobs: Set<Blob> instead of blob: Blob
const pending = pendingReclamation.get(filePath) ?? { blobs: new Set(), deadline: 0, enqueuedAt: now, supersededAt: now };
pending.blobs.add(blob);
// here:
for (const b of pending.blobs) { const si = storageInfoForBlob.get(b); if (si) si.discarded = true; }
```

---
*Generated by Barber AI (claude-opus-5)*

DESIGN.md
```diff
@@ -51,6 +51,8 @@ The mitigations live in three places:
...
 - `LMDBTransaction.abort` and `DatabaseTransaction.abort` walk all writes and run the same cleanup unconditionally (regardless of `skipped`), since nothing was committed. `DatabaseTransaction.commit` adds an explicit reject handler so a `Promise.all` failure on `completions` (e.g. a blob save errored) aborts the underlying transaction instead of leaking it _and_ the blob files.
 
+**A blob instance outlives its file, so deletion tombstones it.** The `fileId` stays set on a `Blob` whose file has been unlinked, and `saveBlob` short-circuits on a set `fileId` — so a caller still holding the instance (the deploy recorder re-puts the same record object across a deploy) would re-encode a reference to a file that is gone. `cleanupUnusedBlobs` marks `storageInfo.discarded` when the deletion is _decided_ (not when the unlink is issued, which waits for an in-flight save to settle), `deleteBlob` marks it too, and `saveBlob` throws on a discarded instance rather than minting a second reference (issue #2062).
```
cb1kenobi: **Low — the rebase invalidated this sentence, and it inverts the timing the design turns on.**

"`deleteBlob` marks it too" was accurate at the previous head. It isn't now: `deleteBlob` (blob.ts:1106-1123) only enqueues a reclamation. The marking moved to `runReclamation()` (blob.ts:1227), which fires at unlink-claim time — after the retention delay, and potentially after `RECLAMATION_AGE_CAP`. Since the sentence's own point is "when the deletion is _decided_ (not when the unlink is issued)", the stale half now asserts the opposite of what the code does, on the exact distinction the paragraph exists to draw.

The same staleness is in `resources/blob.ts:67` — "This blob's file has been unlinked (`deleteBlob`)" — which now misattributes the setter.

**Suggested fix**, for both: `cleanupUnusedBlobs` marks at decision time; reclamation marks once the unlink is claimed, after the retention window has ruled out a legitimate re-reference.

Worth calling out in the PR body too — the two-setter split *is* the interesting design outcome of rebasing onto the new reclamation machinery, and right now the only record of it is a comment that says the old thing.

---
*Generated by Barber AI (claude-opus-5)*

unitTests/resources/txn-tracking.test.js
```diff
@@ -165,6 +167,153 @@ describe('Write txn timeout', () => {
...
+		const slow = new PassThrough();
+		const blob = createBlob(slow);
+		setExpiration(20);
+		try {
+			const context = {};
+			const committing = transaction(context, async () => {
+				await BlobResource.put({ id: 2062, blob }, context);
+			});
+			slow.write(Buffer.alloc(16384, 'a'));
+			await delay(150); // the monitor fires repeatedly while the commit waits on the blob write
```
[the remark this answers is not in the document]
kriszyp: Addressed in b8c3ec115: the test now drives monitor ticks through observable state with waitFor; the fixed 150ms sleep is gone. — GPT-5 Codex

unitTests/resources/txn-tracking.test.js
```diff
@@ -165,6 +167,153 @@ describe('Write txn timeout', () => {
...
+		const slow = new PassThrough();
+		const blob = createBlob(slow);
+		const context = { sourceApply: true };
+		setExpiration(20);
+		try {
+			const committing = transaction(context, async () => {
+				await BlobResource.put({ id: 2066, blob }, context);
+			});
+			slow.write(Buffer.alloc(16384, 'f'));
+			await delay(600); // many ticks, well past COMMIT_PHASE_GRACE
```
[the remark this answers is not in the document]
kriszyp: Addressed in b8c3ec115: the source-apply test condition-drives COMMIT_PHASE_GRACE plus two monitor ticks and asserts the transaction remains unpoisoned; the fixed 600ms sleep is gone. — GPT-5 Codex

resources/DatabaseTransaction.ts
kriszyp: Addressed in b8c3ec115: setCommitPhase propagates across every database link on Rocks and LMDB, timeout re-arms preserve timeoutBudget, and the real two-database monitor test now runs on both engines and verifies both writes commit. — GPT-5 Codex

resources/blob.ts
kriszyp: Addressed in b8c3ec115: a dedicated test keeps the blob save in flight, invokes cleanupUnusedBlobs, and proves re-storage fails before the source settles or the file is unlinked. — GPT-5 Codex

resources/blob.ts
kriszyp: Addressed in b8c3ec115: pending reclamation tracks every live same-file instance with deduplicated WeakRefs, tombstones each instance when unlink is claimed, and shares condemnation state with slices. Tests cover distinct decoded/original instances and a slice. — GPT-5 Codex

DESIGN.md
kriszyp: Addressed in b8c3ec115: DESIGN.md and the StorageInfo comment now distinguish transaction cleanup decision-time tombstoning from reclamation claim-time tombstoning, including the retained re-reference window. — GPT-5 Codex

resources/DatabaseTransaction.ts
```diff
@@ -1187,6 +1222,28 @@ function startMonitoringTxns() {
 						}`
 					);
 					txn.releaseReadTxn();
+				} else if (
+					txn.committing &&
+					(txn.sourceApply || txn.isReplay || ++txn.commitPhaseTicks <= COMMIT_PHASE_GRACE)
```
cb1kenobi: **High — the chain tear is still reachable: `setCommitPhase` fixed the sparing side, but not the aborting side.**

`setCommitPhase()` now propagates `committing` and resets `commitPhaseTicks` across the whole `.next` chain, so every link *enters* the phase together — that half is solid, and a mutation that drops the chain walk is caught by two of the new tests. But each link then counts its own `commitPhaseTicks` against its own `timeout` clock, and `abortDueToTimeout()` still walks **forward only**. Whichever link exhausts the grace first poisons itself and its successors; if that link isn't the head, the head is never poisoned, resumes from the pre-commit await with both new guards passing (`timedOut` false, its own writes intact), and **commits durably**.

Reproduced at `b8c3ec11` with no field manipulation — an ordinary A/B/A multi-store write (`timeoutBudget` set, blob on the last A write) is enough to put the link a tick ahead of the head, because `addWrite` re-arms only the link that receives the write:

```
ENTRY      head.timeout=580  link.timeout=220
FIRST-DEATH head.timedOut=undefined head.ticks=10 | link.timedOut=true link.ticks=11
RESULT     A1=COMMITTED A2=COMMITTED B=absent
           caller error = transactionOpenTooLongError
```

Same result on both engines (RocksDB and `HARPER_STORAGE_ENGINE=lmdb` — `LMDBTransaction.ts:362` has the identical shape). Control with `head.committing = false` (the pre-PR shape) at the same head: `A1=absent A2=absent B=absent` — the chain dies atomically. So the durable torn commit is still introduced by the grace, not pre-existing.

Net effect is unchanged from the previous round: database A commits, B's writes are dropped and its blobs unlinked, and the caller is handed an error that reads as a full rollback.

**Suggested fix** — make the grace a property of the chain rather than of each link: keep one counter on the chain head and abort from the head.

```ts
setCommitPhase(committing: boolean): void {
	for (let txn: DatabaseTransaction = this; txn; txn = txn.next) {
		txn.committing = committing;
		txn.commitChainHead = committing ? this : undefined;
		if (committing) txn.commitPhaseTicks = 0;
	}
}
```

then in both monitors use `const head = txn.commitChainHead ?? txn;` for `++head.commitPhaseTicks <= COMMIT_PHASE_GRACE` and for the `abortDueToTimeout()` call, so the whole chain runs out of grace at the same instant and dies from the head.

The new `keeps every multi-store link alive while the head waits on its blob save` test covers the survive-together half; it would be worth extending it (or adding a sibling) to drive one link past `COMMIT_PHASE_GRACE` and assert the chain fails together — that is the assertion that would have caught this.

The `Math.max(txnExpiration, txn.timeoutBudget ?? 0)` re-arm from the previous round's minor note is fixed here and at the LMDB sites — thanks for picking that up.

—
_Generated by Barber AI (claude-opus-5)_

unitTests/resources/blob.test.js
```diff
@@ -1007,6 +1008,72 @@ describe('Blob test', () => {
...
+			{
+				message: 'the in-flight save should create its file',
+			}
+		);
+		cleanupUnusedBlobs(preCommit.blobs);
+		let storeError;
+		try {
+			assert(existsSync(filePath), 'the file should still be present while its source is open');
+			try {
+				await BlobTest.put(record);
```
cb1kenobi, line 1068: **Low — this test hangs the suite forever if the tombstone regresses, instead of failing.**

The stream is only ended in the `finally` at line 1071, but the `await` that can block is inside the `try`. Without the decision-time tombstone `saveBlob` doesn't throw — it waits on `storageInfo.saving`, which only settles once `slow.end(...)` runs, which the blocked `await` prevents. `.mocharc` sets `timeout: 0`, so there is no mocha timeout to break the deadlock.

Confirmed by mutation: replacing `discardStorage(storageInfo)` in `cleanupUnusedBlobs` (`blob.ts:2343`) with a no-op and rebuilding to `dist/` (verified: 2 `discardStorage(` call sites left instead of 3) left this test running for 4+ minutes with no output; forcing `--timeout 30000` turned it into a clean failure. The mutant *is* detected — the tombstone is genuinely covered now, which closes the previous round's Medium — but the failure mode is a hung CI job rather than a red test.

**Suggested fix:** end the stream from a scheduled callback before the awaited put, e.g. `const ending = setTimeout(() => slow.end(Buffer.alloc(16384, 'r')), 250);` with `clearTimeout(ending)` in the `finally`, so the save can always settle and the assertion at line 1073 is what reports the regression.

—
_Generated by Barber AI (claude-opus-5)_

resources/blob.ts
cb1kenobi: Resolved at `b8c3ec11`. The new `#2062: cleanup tombstones a blob before its in-flight save settles` case pins the decision-time semantics specifically. Mutation-verified: killing `discardStorage(storageInfo)` in `cleanupUnusedBlobs` (built to `dist/`, confirmed 2 remaining `discardStorage(` call sites instead of 3) now fails that case — it survived before. The reclamation-time tombstone is separately covered.

—
_Generated by Barber AI (claude-opus-5)_

resources/blob.ts
cb1kenobi: Resolved at `b8c3ec11`, and more thoroughly than suggested — `pending.blobs`/`seenBlobs` track every instance handed to `deleteBlob` for the path, and the shared `fileState` covers slices too. Both halves are mutation-verified: marking only `pending.blobs[0]` fails `re-storing a discarded blob must fail loudly`, and giving a slice its own `fileState` fails `a slice sharing the condemned file must also fail`.

—
_Generated by Barber AI (claude-opus-5)_

DESIGN.md
cb1kenobi: Resolved at `b8c3ec11` — the rewritten paragraph now draws the two-setter split correctly, and the `blob.ts:67` comment no longer misattributes the setter.

—
_Generated by Barber AI (claude-opus-5)_

resources/DatabaseTransaction.ts
cb1kenobi: Resolved at `fdc22f01` — and I re-ran the reproduction rather than taking the commit titles for it.

Same organic A/B/A multi-store write as last round: `timeoutBudget` set, blob on the last A write, **no field manipulation**, so `addWrite` re-arms only the receiving link and the chain enters its commit phase asymmetrically. All four runs entered on identical conditions (`head.timeout=580  link.timeout=240`):

| engine | `b8c3ec11` (control) | `fdc22f01` |
| --- | --- | --- |
| RocksDB | `A1=COMMITTED A2=COMMITTED B=absent` — torn | `A1=absent A2=absent B=absent` — atomic |
| LMDB | `A1=COMMITTED A2=COMMITTED B=absent` — torn | `A1=absent A2=absent B=absent` — atomic |

The caller gets `transactionOpenTooLongError` in all four; the only difference is whether the head's writes survived it. At the control the link died alone — `head.timedOut=undefined head.ticks=10 | link.timedOut=true link.ticks=11`. At the new head both die together off one shared counter — `head.timedOut=true head.ticks=11 | link.timedOut=true link.ticks=0`.

Both halves check out:

**Head-rooted abort.** `abortDueToTimeout()` itself still walks forward only; the rooting moved to the call site (`commitChainHead.abortDueToTimeout()`), which is the simpler place for it. `commitChainHead` comes from `setCommitPhase(true)`, and the repro confirms both links resolve to the head.

**Shared grace.** `shouldSpareCommitPhase` keys the counter off the chain head behind a per-tick `Set`, so every link gets the same verdict within a tick and a link with a shorter `timeout` can no longer exhaust first — its early expiry just spends the chain's shared budget, and the abort that follows is rooted at the head.

Mutation-verified, each rebuilt to `dist/` with the marker count asserted at exactly 1:

- `commitChainHead.abortDueToTimeout()` → `txn.abortDueToTimeout()` at `DatabaseTransaction.ts:1285` → red: `AssertionError: every link must be poisoned together`
- the same revert at `LMDBTransaction.ts:398` → red on LMDB, same assertion

So the fail-together half is genuinely asserted now, on both engines — that was the gap last round. Nice fix, and the new test is anchored on the right invariant.

One thing worth recording since I went looking for it: if the pre-commit work sits on the **second** database instead of the head, the chain still tears (`A=COMMITTED B=absent`), because the head's native commit lands before the cascade calls `this.next.commit()`. That is pre-existing and outside this PR — at the merge base it tears the same way and the caller gets **no error at all**, so this change strictly improves it (silent tear → loud `transactionOpenTooLongError`). Flagging it only so it is written down somewhere, not as something to fix here.

—
_Generated by Barber AI (claude-opus-5)_

unitTests/resources/blob.test.js
cb1kenobi: Resolved at `fdc22f01`. The 250 ms watchdog ends the stream even when the `await` inside the `try` is the thing that would have blocked, and the guarded `finally` (`clearTimeout` + `writableEnded`) keeps the double-end safe.

Mutation-verified with `discardStorage()` neutralised to a no-op (rebuilt to `dist/`, marker count 1), run with **no `--timeout` override** so `.mocharc`'s `timeout: 0` applies exactly as it does in CI:

- `b8c3ec11` — no output at all; still running when I killed it at **8m 43s**
- `fdc22f01` — **fails in 387 ms**: `AssertionError: re-storing must fail before the condemned file is unlinked` at `blob.test.js:1075`

A tombstone regression now reports instead of wedging the job. Thanks for picking this one up — it was the kind of thing that only bites at the worst possible moment.

—
_Generated by Barber AI (claude-opus-5)_

unitTests/resources/txn-tracking.test.js
```diff
@@ -378,6 +400,298 @@ describe('Write txn timeout', () => {
...
+		head.setCommitPhase(true);
+		assert.ok(head.committing && next.committing, 'every linked database must be spared together');
+		assert.equal(head.commitChainHead, head);
+		assert.equal(next.commitChainHead, head);
+		assert.equal(head.commitPhaseTicks, 0);
+		assert.equal(next.commitPhaseTicks, 0);
+		const checkedCommitPhaseChains = new Set();
+		assert.ok(shouldSpareCommitPhase(head, checkedCommitPhaseChains));
+		assert.ok(shouldSpareCommitPhase(next, checkedCommitPhaseChains));
+		assert.equal(head.commitPhaseTicks, 1, 'one monitor pass must consume one grace tick per chain');
```
cb1kenobi, line 575: **Nit — this assertion passes under per-link counting too, so the commit it belongs to has no test that can fail.**

`assert.equal(head.commitPhaseTicks, 1, ...)` can't tell the shared per-chain counter apart from the per-link counting it replaced: under per-link counting `++head.commitPhaseTicks` also leaves `head.commitPhaseTicks === 1`. `next` is where the two diverge, and nothing looks at it.

Mutation: replacing the body of `shouldSpareCommitPhase` with `return ++txn.commitPhaseTicks <= COMMIT_PHASE_GRACE;` (rebuilt to `dist/`, marker count 1) leaves this file fully green — 20 passing, 0 failing.

To be clear this is a coverage point, not a correctness one. I re-ran the organic multi-store repro under that same mutation and it still came out atomic (`A1=absent A2=absent B=absent`), because the head-rooted `abortDueToTimeout()` carries the atomicity on its own. The shared counter is the tidier semantic, it just isn't pinned by anything.

One line pins it — verified both ways: it passes on `fdc22f01` as-is, and under the per-link mutation it fails with `AssertionError: a link must not run its own grace counter`.

```suggestion
		assert.equal(head.commitPhaseTicks, 1, 'one monitor pass must consume one grace tick per chain');
		assert.equal(next.commitPhaseTicks, 0, 'a link must not run its own grace counter');
```

—
_Generated by Barber AI (claude-opus-5)_

cb1kenobi:
Re-reviewed `97821ba7` — no issues found. This PR looks good, nice job!

**What moved since `81f27905`:** nothing of the PR's own, as far as I can tell. `git range-diff` against properly derived merge bases for both heads (`2da8697e` for the old head, `29072607` for the new) maps all 13 commits 1:1. Eight are byte-identical; four differ only in context lines that main moved underneath (`writesAbandoned` → `snapshotFree`, and `startMonitoringTxns` extracting a `monitorTransaction()`). Only `Pin the per-chain grace counter in the chain test` carries a real 7-line production delta, and it is rebase plumbing: threading `checkedCommitPhaseChains` as a parameter into main's newly-extracted `monitorTransaction`, and covering main's new `supervisedWriteRoots` registry with the second loop. **Genuine new-work commits: 0.**

Worth saying that the threading is necessary rather than cosmetic. The per-tick `Set` is keyed on `commitChainHead`, so a chain reachable from *both* registries — a root that read (`trackedTxns`) while a later link blind-wrote (`supervisedWriteRoots`) — still spends exactly one grace tick per monitor tick. Had the `Set` stayed local to one loop, that shape would burn two ticks per tick and quietly halve `COMMIT_PHASE_GRACE`. I traced all four chain shapes through both loops and the accounting holds.

I also checked the branch ordering the rebase created, since main inserted its `open === CLOSED` branch *ahead* of the commit-phase branch: `commit()` only sets `open = CLOSED` after the pre-commit await resolves, so a parked transaction is still `OPEN` when the monitor sees it and `shouldSpareCommitPhase` is reached. The tests confirm it — the "in its commit phase past the open-transaction limit" line fires.

**Prior findings, re-checked against the new head by mutation rather than by reading the patch** — main drifted `DatabaseTransaction.ts` by +513 lines, `blob.ts` by +95 and `LMDBTransaction.ts` by +49 underneath this PR, so an unchanged patch is not evidence an unchanged fix. Each mutant was rebuilt with `tsc -p tsconfig.build.json` and its marker asserted at exactly 1 occurrence in `dist/`:

| prior finding | mutation | result |
|---|---|---|
| High — multi-store chain tear | `commitChainHead.abortDueToTimeout()` → `txn.` (`DatabaseTransaction.ts:1640`) | red — `every link must be poisoned together` |
| High — same, LMDB | same revert at `LMDBTransaction.ts:409` | red — same assertion under `HARPER_STORAGE_ENGINE=lmdb` |
| Medium — decision-time tombstone uncovered | `discardStorage(storageInfo)` in `cleanupUnusedBlobs` → no-op | red — `re-storing must fail before the condemned file is unlinked` |
| Low — instance tracking in reclamation | walk only `pending.blobs[0]` | red — `re-storing a discarded blob must fail loudly` |
| Low — test hung instead of failing | (same tombstone mutant, run with **no** `--timeout`, so `.mocharc`'s `timeout: 0` applies as in CI) | red in **420 ms** — reports rather than wedging |
| Nit — grace counter not pinned per-chain | `shouldSpareCommitPhase` → `return ++txn.commitPhaseTicks <= COMMIT_PHASE_GRACE` | red — `a link must not run its own grace counter`, **both** engines |

All six survive the rebase, and each is killed by this PR's own tests against this PR's own code — so the fixes are the PR's, not something main handed it. The `DESIGN.md` doc fix I checked by reading.

**Suites at this head:** `unitTests/resources` 1668 passing / 0 failing on RocksDB, 1437 passing / 0 failing on LMDB. `tsc -p tsconfig.json` clean, `oxlint --deny-warnings` and `prettier --check` clean on the changed files.

**What I could not run, so you know the shape of the gap:** `integrationTests/database/blob-commit-over-time.test.ts`. The harness boots its own Harper instance and a local dev server currently holds 9925/9926, so this is a port collision on my end, not a test problem — my read of it is static only. It does carry its own anti-vacuous guard (it asserts the monitor's spare line actually appears in the log), so it cannot pass by silently not exercising the path. `test:unit:apitests` I deliberately skipped: it begins with `harper.js stop`.

One note for cross-PR awareness, not a finding here and not something to act on in this PR: neither #1835's `enqueueBlobUnlink` nor #2265's backup-walk exists in this merge base, so there is no live interaction with the decision-time tombstone to verify yet. If #1835 later moves its durable row earlier into `deleteBlob()`, it will need to record the cancellation when `cancelBlobReclamation()` rescues a blob inside the retention window — otherwise a legitimately re-referenced blob gets unlinked after a restart. That is #1835 against main's existing reclamation machinery; this PR neither creates nor worsens it.

—
_Generated by Barber AI (claude-opus-5)_

cb1kenobi:
Re-reviewed `7450b485` — no issues found. This PR looks good, nice job!

This is a rebase, not a merge: `git range-diff` against the last-reviewed `97821ba7` shows all 13 commits map 1:1 (11 identical, 2 with import-reordering-only rebase-context diffs) — the PR's own content is unchanged, only `main` moved underneath it. Given that, re-verified all six prior findings hold at the *current* head rather than trusting the unchanged diff:

- **Multi-store chain tear**: `abortDueToTimeout()` still poisons every link in a first pass (`timedOut`/`open=CLOSED`) before aborting any of them in a second pass, so a throw mid-abort can't leave a later link un-poisoned. Dynamically confirmed (`keeps every multi-store link alive...`, `aborts a parked multi-store commit from the chain head...`).
- **Decision-time tombstone**: `cleanupUnusedBlobs()` still calls `discardStorage()` synchronously before scheduling the async unlink. Dynamically confirmed (`#2062: cleanup tombstones a blob before its in-flight save settles`, 1107ms).
- **Reclamation instance tracking**: `PendingReclamation.blobs` is still `WeakRef<Blob>[]` + `seenBlobs: WeakSet`, and `runReclamation()` still reads the file-identity-based `pending.fileInfo` rather than re-deriving it from a possibly-GC'd blob instance.
- **Hung-vs-failing test**: ran the two "throws instead of resolving as success" tests with no `--timeout` override (so `.mocharc`'s `timeout: 0` is what would catch a real hang) — both pass in 225ms combined, no hang.
- **DESIGN.md**: read-verified against the current code (decision-time tombstone, `COMMIT_PHASE_GRACE` bound, unconditional `sourceApply`/`isReplay` sparing, `saveBlob` throwing on a discarded instance) — still accurate.
- **Per-chain grace counter**: `commitPhaseTicks` is still a `DatabaseTransaction` instance field, and `shouldSpareCommitPhase()` still takes `checkedCommitPhaseChains` as a parameter rather than reading module state.

Also traced the load-bearing detail explicitly: `startMonitoringTxns()`'s tick handler creates `checkedCommitPhaseChains` once per tick and passes the *same* Set into `monitorTransaction()` for both the `trackedTxns` loop and the `supervisedWriteRoots` loop, so a chain reachable from both spends exactly one grace tick per tick, not two — confirmed by code trace (no dedicated dynamic test isolates this exact dual-registry case, flagging that as static-only).

Checked harper#1835 (durable blob-unlink queue) for interaction: none currently — its durable-unlink commit lives only on its own unmerged branch, not on `main` or this PR's branch. What this PR's body calls "main's deferred-reclamation machinery" is a separate, already-merged commit.

`npm run build` (tsc), `npm run typecheck` clean. `unitTests/resources/txn-tracking.test.js` (25/25), `unitTests/resources/blob.test.js` (105/105), and the full `unitTests/resources/**` suite (1706/1706, 16 pending/skipped) all pass. `npm run lint` has 12 pre-existing warnings, none in files this PR touches.

—
_Generated by Barber AI_

Issue #1653, opened 2026-07-07
[Epic] SSE streaming hardening


kriszyp:
Tracking epic for SSE/streaming response bugs: deferred teardown leaking listeners/sockets, and a finite async generator hanging the response with an uncaughtException. Related: #1290 (unbounded live-tail queue) is tracked under #1566 (unbounded-queue/DoS family) instead, since it's a duplicate of that cluster's canonical issue.

---
_Filed by Claude (Opus 4.8), issue-backlog triage & grouping pass._

PR #2096, opened 2026-08-06
fix(server): SSE serializer must distinguish absent data/id/retry from falsy

### LINKED ISSUE #2026: SSE text/event-stream serializer drops legitimate falsy data (0/false/'') and id:0
## Summary

`server/serverHelpers/contentTypes.ts`'s `text/event-stream` media-type `serialize()` uses truthiness to decide whether to emit a message's `data:`/`id:` lines:

```js
if (message.data) { /* emit data: line */ }
...
if (message.id) { /* emit id: line */ }
```

Both guards conflate "absent" with "falsy." For `data`, that silently omits the `data:` field for legitimate values `0`, `false`, and `''` (not just `undefined`/`null`). Per the HTML `EventSource` algorithm, a dispatched message with an empty data buffer is *discarded without firing an event* — so a client's listener never sees the update at all, with a 200 response and nothing in the logs to flag it.

For `id`, the same guard drops `id: 0`, which is a legitimate reconnect cursor — a client resuming from `id: 0` re-reads from the start instead of resuming.

## Concrete failure scenario

A Resource streams feature-flag state as `{ event: 'flag', data: false }`. The wire frame is `event: flag\n\n` (no `data:` line). The browser's `EventSource` never dispatches the event, so the client silently keeps showing the last truthy value — no error, no log line, 200 status throughout.

## Suggested fix

Distinguish absence from falsiness, e.g.:

```js
if (message.data !== undefined && message.data !== null) { /* emit data: line */ }
if (message.id !== undefined && message.id !== null) { /* emit id: line */ }
```

## Provenance

Surfaced by an independent multi-model pre-push review (codex + Grok + Harper-domain lenses) of #1972 (`test(server): promote 3 QA server-layer contract anchors`). That PR is test-only and deliberately pins the *current* (buggy) contract as a documented, flagged anchor at `integrationTests/server/qa702-sse-event-data.test.ts` rather than silently treating it as correct — this issue tracks the actual fix.

🤖 Filed by Claude (dispatch pr-fix agent) while addressing review comments on #1972.

kriszyp:
## Summary

`contentTypes.ts`'s `text/event-stream` `serialize()` used truthiness (`if (message.data)`, `if (message.id)`, `if (message.retry)`) to decide whether to emit `data:`/`id:`/`retry:` lines. That conflates "absent" with "falsy": legitimate values `0`, `false`, and `''` for `data`, `0` for `id` (a valid reconnect cursor), and `0` for `retry` were silently dropped from the wire frame. Per the `EventSource` spec, a frame with no `data:` line is discarded by the client without firing an event at all — a 200 response with nothing in the logs to flag it.

All three guards now check `!= null`, matching the idiom already used elsewhere in this file (`responseData?.contentType != null`).

**Caveat on `''`:** this fixes the wire representation (now emits `data: \n` instead of omitting the line) and is correct for the absence-vs-falsiness bug in general, but it does not change behavior for a real browser `EventSource` client on this specific value: per the HTML spec, an empty data buffer is discarded at dispatch time whether the `data:` line was present-but-empty or omitted entirely. `0` and `false` are the cases with an actual client-visible fix; `''` is fixed at the wire-byte level (useful for non-browser SSE consumers) but not for `EventSource` itself.

## What did *not* change, on purpose

The outer envelope-detection gate (`if (message.data != null || message.event)`) still only opens on `data`/`event`, not `id`/`retry`. This was tried during review (widening it to `message.id != null || message.retry != null`, to support a standalone id-only/retry-only message) and reverted: `id` is an extremely common literal-payload field name (e.g. any database record), so treating a bare `id` key as an SSE-envelope signal would misclassify a plain data object that happens to carry one — its `id` gets peeled into an `id:` line and every other field is silently dropped. That's a worse regression than the one this issue targets. A standalone id-only/retry-only SSE message therefore remains unsupported; a clean fix would need an explicit envelope marker (out of scope here). `integrationTests/server/qa702-sse-event-data/resources.js`'s `IdKeyPlainObjectPayload` fixture pins the correct (JSON-wrap-wholesale) behavior for this case.

## Change

- Follow-up review fix: [primitive coercion and CR/LF framing now share one path](https://github.com/HarperFast/harper/pull/2096/changes?w=1#diff-3ef5e58afb3497c3da77146d43e5ff63df88b7b9661cd862b98e390882c4f1fcR130), so newline-bearing `Symbol`/function strings cannot leak raw continuation lines; the [newline-bearing Symbol regression assertion](https://github.com/HarperFast/harper/pull/2096/changes?w=1#diff-1109cdb54e0c51de8a2b2d9b4701c8238ca0927bbc08b18b25cd3c86cb5edd50R151) pins that invariant.

- `server/serverHelpers/contentTypes.ts`: `data`/`id`/`retry` guards changed from truthiness to `!= null`.
- `integrationTests/server/qa702-sse-event-data.test.ts` + its `resources.js` fixture: this suite deliberately pinned the old (buggy) contract as a documented known-defect anchor (see #1972). Updated assertions/fixtures for the fixed contract, and added cases for the event-less-falsy-data and plain-object-with-an-id-key shapes that the review surfaced.

## Design assessment

Small, well-scoped fix confined to one function in shared serialization code. The `data`/`id` half was fully specified in #2026 with a pinned test anticipating it; the `retry` guard and the envelope-gate scoping question surfaced during cross-model review (see above) and were resolved before push.

## Verification

- Follow-up verification: `npm run build`; `npx mocha unitTests/server/serverHelpers/contentTypes.test.js` (21/21); `npm run format:write`; and `npm run lint:required` all passed. The new Symbol assertion was also confirmed to fail against the pre-fix serializer with the expected raw continuation line. `npm run test:unit:server` could not start the suite because its glob loads `unitTests/server/threads/fixtures/processGroupOwnerWorker.js` outside a worker (`parentPort` is null).

- Extended `qa702-sse-event-data.test.ts` integration suite (14/14 passing) — exercises the falsy-value matrix (`undefined`/`null`/`''`/`0`/`false`), `id: 0`, `retry: 0`, an event-less `data: 0` message, and a plain object with an `id` key.
- Ran `unitTests/server/serverHelpers/contentTypes.test.js` directly (19/19 passing) — unaffected non-falsy cases still serialize correctly.
- Full `test:unit:main`/`test:unit:resources`/`test:integration:all` gates not run locally (known local contention on this box — shared boot-props/RocksDB lock; CI is authoritative).
- 4 rounds of cross-model review (codex + gemini, + harper-domain adjudication on rounds 1–2): round 1 found the fix as filed was incomplete (outer envelope gate); round 2's suggested widening (to `id`/`retry`) was applied then round 3 caught that it introduced a data-loss regression for plain objects with an `id` key; round 4 confirmed the revert and closed out remaining nits. Converged with no new findings.

Fixes #2026

🤖 Generated by Claude (Sonnet 5)

<sub>Review-Coverage: authored=codex; ran=claude,gemini; declined=cursor-grok,cursor-composer,domain; rounds=6 @ 9d4dfcbbe4fb</sub>

<sub>Human-Review-Need: 4 @ 9d4dfcbbe4fb</sub>

### DIFF
--- a/server/serverHelpers/contentTypes.ts
+++ b/server/serverHelpers/contentTypes.ts
@@ -141,7 +141,7 @@ mediaTypes.set('text/event-stream', {
 				id: message.timestamp,
 			};
 		}
-		if (message.data != null || message.event) {
+		if (message.data != null || message.event || message.id != null || message.retry != null) {
 			let serialized = '';
 			if (message.event) serialized += 'event: ' + message.event + '\n';
 			if (message.data != null) {
@@ -150,7 +150,7 @@ mediaTypes.set('text/event-stream', {
 				serialized += 'data: ' + data + '\n';
 			}
 			if (message.id != null) serialized += 'id: ' + message.id + '\n';
-			if (message.retry) serialized += 'retry: ' + message.retry + '\n';
+			if (message.retry != null) serialized += 'retry: ' + message.retry + '\n';
 			return serialized + '\n';
 		} else {
 			if (typeof message === 'object') return `data: ${JSONStringify(message)}\n\n`;
--- a/server/serverHelpers/contentTypes.ts
+++ b/server/serverHelpers/contentTypes.ts
@@ -141,7 +141,13 @@ mediaTypes.set('text/event-stream', {
 				id: message.timestamp,
 			};
 		}
-		if (message.data != null || message.event || message.id != null || message.retry != null) {
+		// Only `data`/`event` open the envelope path -- NOT `id`/`retry`. `id` in particular is an
+		// extremely common literal-payload field name (e.g. a database record), so treating a bare
+		// `id`/`retry` key as an SSE-envelope signal would misclassify plain data objects that
+		// happen to carry one, silently dropping every other field. A standalone id-only/retry-only
+		// SSE message is consequently unsupported (falls to the JSON-wrap branch below); a clean
+		// fix would need an explicit envelope marker, out of scope here.
+		if (message.data != null || message.event) {
 			let serialized = '';
 			if (message.event) serialized += 'event: ' + message.event + '\n';
 			if (message.data != null) {
--- a/server/serverHelpers/contentTypes.ts
+++ b/server/serverHelpers/contentTypes.ts
@@ -127,8 +127,7 @@ mediaTypes.set('application/ndjson', ndjsonHandler);
 
 function serializeSSEData(data: any) {
 	if (typeof data === 'object') return 'data: ' + JSONStringify(data) + '\n';
-	if (typeof data === 'string') return 'data: ' + data.replace(/\r\n|\r|\n/g, '\ndata: ') + '\n';
-	return 'data: ' + String(data) + '\n';
+	return 'data: ' + String(data).replace(/\r\n|\r|\n/g, '\ndata: ') + '\n';
 }
 
 mediaTypes.set('text/event-stream', {
--- a/server/serverHelpers/contentTypes.ts
+++ b/server/serverHelpers/contentTypes.ts
@@ -127,8 +127,8 @@ mediaTypes.set('application/ndjson', ndjsonHandler);
 
 function serializeSSEData(data: any) {
 	if (typeof data === 'object') return 'data: ' + JSONStringify(data) + '\n';
-	if (typeof data === 'string') data = data.replace(/\r\n|\r|\n/g, '\ndata: ');
-	return 'data: ' + data + '\n';
+	if (typeof data === 'string') return 'data: ' + data.replace(/\r\n|\r|\n/g, '\ndata: ') + '\n';
+	return 'data: ' + String(data) + '\n';
 }
 
 mediaTypes.set('text/event-stream', {
--- a/server/serverHelpers/contentTypes.ts
+++ b/server/serverHelpers/contentTypes.ts
@@ -126,7 +126,7 @@ mediaTypes.set('application/x-ndjson', ndjsonHandler);
 mediaTypes.set('application/ndjson', ndjsonHandler);
 
 function serializeSSEData(data: any) {
-	if (typeof data === 'object') data = JSONStringify(data);
+	if (typeof data === 'object') return 'data: ' + JSONStringify(data) + '\n';
 	return 'data: ' + String(data).replace(/\r\n|\r|\n/g, '\ndata: ') + '\n';
 }
 
--- a/server/serverHelpers/contentTypes.ts
+++ b/server/serverHelpers/contentTypes.ts
@@ -141,12 +141,9 @@ mediaTypes.set('text/event-stream', {
 				id: message.timestamp,
 			};
 		}
-		// Only `data`/`event` open the envelope path -- NOT `id`/`retry`. `id` in particular is an
-		// extremely common literal-payload field name (e.g. a database record), so treating a bare
-		// `id`/`retry` key as an SSE-envelope signal would misclassify plain data objects that
-		// happen to carry one, silently dropping every other field. A standalone id-only/retry-only
-		// SSE message is consequently unsupported (falls to the JSON-wrap branch below); a clean
-		// fix would need an explicit envelope marker, out of scope here.
+		// Only `data`/`event` open the envelope path. `id`/`retry` alone do not: `id` is too
+		// common a literal-payload field name (e.g. a database record) to safely treat as an
+		// SSE-envelope signal without misclassifying plain data objects that happen to carry one.
 		if (message.data != null || message.event) {
 			let serialized = '';
 			if (message.event) serialized += 'event: ' + message.event + '\n';
--- a/server/serverHelpers/contentTypes.ts
+++ b/server/serverHelpers/contentTypes.ts
@@ -141,7 +141,7 @@ mediaTypes.set('text/event-stream', {
 				id: message.timestamp,
 			};
 		}
-		if (message.data || message.event) {
+		if (message.data != null || message.event) {
 			let serialized = '';
 			if (message.event) serialized += 'event: ' + message.event + '\n';
 			if (message.data != null) {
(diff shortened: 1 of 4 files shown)


server/serverHelpers/contentTypes.ts
```diff
@@ -141,16 +141,19 @@ mediaTypes.set('text/event-stream', {
 				id: message.timestamp,
 			};
 		}
-		if (message.data || message.event) {
+		// Only `data`/`event` open the envelope path. `id`/`retry` alone do not: `id` is too
+		// common a literal-payload field name (e.g. a database record) to safely treat as an
+		// SSE-envelope signal without misclassifying plain data objects that happen to carry one.
+		if (message.data != null || message.event) {
```
cb1kenobi, line 147: **Medium: widening the envelope gate to `!= null` peels falsy-`data`-key plain objects that previously round-tripped wholesale**

This is the mirror of the `id`-key case you deliberately protected. A plain data record that happens to carry a top-level column literally named `data` with a falsy-but-present value — `{ data: 0, name: 'x' }` (or `''`/`false`) and no `event` — now matches `message.data != null`, enters the envelope path, and emits only `data: 0`, silently dropping `name`.

Under the old truthiness gate that same record was falsy, fell through to the `else` branch, and was JSON-wrapped wholesale (`data: {"data":0,"name":"x"}`) with every field preserved. So this is a strict regression for that shape — the identical silent-field-loss class your reasoning cites for declining to widen to `id`/`retry`, except `data` is just as common a column name.

The genuine envelope signal `data` can't be dropped, so a full fix would need the explicit envelope marker you noted is out of scope. Minimally: pin the intended behavior with a fixture for `{ data: 0, name: 'x' }` (siblings + falsy data key) — `ZeroPayloadNoEvent` only covers `{ data: 0 }` with no siblings — and document this as an accepted limitation the way you did for the `id`-key case, so the tradeoff is recorded rather than latent.

—
_Generated by Barber AI_

ldt1996:
The fix and the envelope-gate scoping both check out, and the round-3 catch (a bare id key misclassifying a plain record) with its IdKeyPlainObjectPayload pin is the load-bearing part; the revert was right. One deliberate change worth naming for the record: {data: 0, name: 'x'} previously JSON-wrapped the whole object and now drops the sibling, consistent with what truthy data always did, and the test pins it. Agreed with that resolution, and the '' caveat is honest per the HTML spec.

Pre-existing and out of scope, but adjacent to these exact lines: a string data value containing a newline still emits a single broken data: line rather than the spec's one-data:-line-per-line splitting, and a string id with a newline can inject SSE fields. Worth a follow-up issue if there isn't one.

server/serverHelpers/contentTypes.ts
```diff
@@ -125,6 +125,12 @@ const ndjsonHandler = {
 mediaTypes.set('application/x-ndjson', ndjsonHandler);
 mediaTypes.set('application/ndjson', ndjsonHandler);
 
+function serializeSSEData(data: any) {
+	if (typeof data === 'object') return 'data: ' + JSONStringify(data) + '\n';
+	if (typeof data === 'string') return 'data: ' + data.replace(/\r\n|\r|\n/g, '\ndata: ') + '\n';
+	return 'data: ' + String(data) + '\n';
```
cb1kenobi: **Low: the fall-through branch is the one payload path `serializeSSEData` does not newline-scan, and "Preserve SSE primitive coercion" just made it reachable**

`String(data)` is applied raw, so any non-object, non-string value whose coercion contains a line break emits a literal newline into the `data:` line and splits/truncates the frame. Confirmed by running the built `dist/` serializer at this head against the merge base (01d85622):

- `serialize(Symbol('m\nn'))` — base: `TypeError: Cannot convert a Symbol value to a string`; head: `"data: Symbol(m\nn)\n\n"`. A spec client reads `n)` as a colon-less line, ignores it, and dispatches `data = "Symbol(m"` — silent truncation rather than the previous throw. This branch is newly reachable precisely because of the `${data}` -> `String(data)` change.
- `serialize(function f(){ ... })` behaves the same way and is unchanged from base, so a multi-line function body has always leaked raw newlines here.

Object and string payloads are fine — I verified `JSONStringify` escapes on both paths (`JSON.stringify`, and `jsStringify` in `JSONStream.ts` which routes strings and keys through `JSON.stringify`), so "Avoid rescanning SSE object payloads" is byte-identical to base for every object/array/Date/Buffer case I probed. This is only about the last line.

Folding the scan into the coercion closes the gap and is shorter than what it replaces, while keeping the object fast path unscanned (`String(x)` on a string is identity, so the string case is unaffected):

Suggested fix:
```suggestion
	if (typeof data === 'object') return 'data: ' + JSONStringify(data) + '\n';
	return 'data: ' + String(data).replace(/\r\n|\r|\n/g, '\ndata: ') + '\n';
```

I applied this in a worktree, rebuilt `dist/`, and re-ran `unitTests/server/serverHelpers/contentTypes.test.js`: 21/21 still pass, and `Symbol('m\nn')` then serializes as `data: Symbol(m\ndata: n)\n\n`. The unit test's `Symbol('marker')` assertion keeps passing either way, so if you prefer to keep the current shape, consider giving that marker a newline so the branch is actually pinned.

Separately (not this PR's bug, and already raised by Lavinia): `event`, `id`, and `retry` are still raw-concatenated, so `{ data: 'ok', id: '1\ndata: INJECTED' }` still emits an injected `data:` line. Byte-identical at base and head, so no regression here — but there is no open issue tracking it; #1653 (SSE streaming hardening) looks like the right home.

—
_Generated by Barber AI_

kriszyp: Addressed in 9d4dfcbbe: the non-object path now coerces once with String(data) and applies the same CR/LF data-line framing used for strings. The Symbol regression assertion now includes an embedded newline, so this branch is directly pinned.

cb1kenobi (approved):

PR #2098, opened 2026-08-06
Fix copy-db producing a silently corrupt, non-restorable database copy

### LINKED ISSUE #2048: copy-db produces a silently corrupt, non-restorable copy and exits 0 — four independent channels in bin/copyDb.ts
# `copy-db` produces a silently corrupt, non-restorable copy and exits 0 — four independent channels in `bin/copyDb.ts`

## Summary

`bin/copyDb.ts`'s `copyDb()` / `copyDbi()` — the pair behind both the documented `copy-db <source> <target>` CLI verb (`bin/harper.ts:153`) and the `storage.compactOnStart` flag — has **four independent defects that each silently degrade the copy**. The command exits 0 and logs `copied N entries` in every case.

Three were found by exploratory QA with execution-verified measurements; the fourth is a source-read finding in the same function. They are filed together because they live in one ~145-line function, share one root cause pattern (the copy loop was written as a raw byte-for-byte dbi walk and therefore cannot see any structure above raw key/value), and any fix touches the same code. **They should be fixed as four changes, not one.**

Severity order is roughly 1 > 2 > 3 > 4.

---

## Channel 1 — the tombstone heuristic drops the shared-structures dictionary, blanking every record

`bin/copyDb.ts:235`:

```js
// deleted entries should be 13 bytes long (8 for timestamp, 4 bytes for flags, 1 byte of the encoding of null)
if (value?.length < 14 && isPrimary) {
    skippedRecord++;
    continue;
}
```

This length heuristic also matches the table's **shared-structures dictionary entry** (stored under the `Symbol.for('structures')` key in the primary dbi). The dictionary is skipped as if it were a delete tombstone, so every record in the copy then decodes to `null`.

Unlike the RocksDB migration path in the same file (`copyDbiToRocks`, which explicitly does `if (typeof recordKey === 'symbol') continue` and separately calls `copyStructures()`), the LMDB `copyDb()` path has no symbol-key skip and no `copyStructures()` — it relies on the raw dbi walk to carry the symbol key through, and the heuristic eats it.

**Measured through the CLI:** 120/120 records readable before → **0/120 readable after**. Raw keys all present (120/120), secondary index 5/120. So the rows are physically there and functionally blank.

**Narrowings that must survive into the fix and its test:**
- It is **schema-shape-dependent** — it fires only when the dictionary entry encodes to under 14 bytes, i.e. short attribute names. Realistic attribute names produce a larger dictionary and the table survives. That is why this has gone unnoticed.
- `migrateOnStart` / v4→v5 is **not** affected (different function, which handles structures explicitly).

## Channel 2 — `getKeys()` + `getEntry()` collapses every dupSort secondary index

`bin/copyDb.ts:230-233` walks the dbi with `sourceDbi.getKeys({ start, transaction })` and then a single `sourceDbi.getEntry(key, { transaction })` per key. On a **dupSort** dbi that yields one entry per *unique key*, discarding every duplicate.

Post-copy the secondary index therefore collapses to 1 entry per unique key.

**Measured:** groups of N=2, 10, and 100 duplicate values all collapse to 1. A unique-valued control index survives 112/112, which is what pins it to the dupSort path rather than to the copy loop generally.

## Channel 3 — the blob store is never copied, so the copy is non-restorable

`copyDb()` walks LMDB dbis only. It never visits the blob store, which `resources/blob.ts:1207` resolves as:

```js
return [join(getHdbBasePath(), 'blobs', databaseName)];
```

— i.e. from the **running instance's base path and the database name**, not from the env file being copied. Each record's `fileId` reference is copied faithfully (the loop is a raw binary copy: `dbiInit.encoding = 'binary'`, `sourceDbi.decoder = null`, so it never decodes a value and cannot know one holds a blob reference), while the bytes it points at are left behind.

**Measured on `7863b7468`:** 6 blob docs @ 1.5 MB plus 4 sub-threshold inline controls. CLI exited 0 logging `copied 11 entries`. After swap-in: rows fully intact (10/10 readable — so this measurement is *not* confounded by Channel 1), **6/6 attachments threw `BlobReadError` 404**, 4/4 inline controls byte-exact. Source blobs verified untouched.

**Exposure differs by call site, and only incidentally:**
- **`copy-db` CLI — exposed.** The target is an arbitrary path. Any copy that is later opened under a different database name, or moved to another data root or host, has lost every blob.
- **`compactOnStart` — not currently exposed.** It moves the copy back to the *original* `dbPath` under the same database name in the same base path (`bin/copyDb.ts:84`), so the untouched blob directory still resolves. Nothing enforces that invariant, though — it holds by accident of the call site, not by design.

Note the contrast with `copyDbToRocks()` in the same file, which *does* handle blobs deliberately (`encodeBlobsWithFilePath`, `beginPendingMigrationBlobSaves`, with comments citing #857 and #1337). The blob-awareness work landed on the migration path and never on the copy/compaction path.

## Channel 4 — the audit store is copied into the **source** env, not the target, and isn't awaited

`bin/copyDb.ts:216-220`:

```js
if (sourceAuditStore) {
    const targetAuditStore = rootStore.openDB(AUDIT_STORE_NAME, AUDIT_STORE_OPTIONS);
    console.log('copying audit log for', sourceDatabase, 'to', targetDatabasePath);
    copyDbi(sourceAuditStore, targetAuditStore, false, transaction);
}
```

`rootStore` is the **source** env (assigned from `table.primaryStore.rootStore` at line 171); the target is `targetEnv`, opened at line 177. So the variable named `targetAuditStore` is a handle on the source. `lmdb-js`'s `openDB` constructs a fresh `LMDBStore` on every call (`node_modules/lmdb/open.js:405-417` — no handle cache), so this is a *new, live* handle whose `put` was **not** covered by the deliberate write-guard at lines 160-170 (`table.primaryStore.put = noop` etc.).

Two consequences:
1. **The target copy gets no audit / transaction log at all.**
2. The copy loop writes source audit entries **back into the source database** during an operation that is documented and guarded as read-only on the source. The re-put is key-for-key identical so the data damage is likely nil, but it dirties the source env mid-compaction and defeats the write-guard's purpose.

Separately, this `copyDbi` call is **not awaited** — compare line 214, `await copyDbi(sourceDbi, targetDbi, isPrimary, transaction)`, for the regular dbis. The un-awaited writes race `transaction.done()` and `targetEnv.close()` in the `finally` at lines 293-296.

Channel 4 is **source-read only** — I have not run it. Channels 1–3 carry execution-verified numbers.

---

## Suggested direction

Four separate changes:

1. **Channel 1** — skip symbol keys explicitly and copy the structures entry explicitly, the way `copyDbToRocks` already does, instead of inferring "tombstone" from a length. A 13-byte length test cannot distinguish a tombstone from a small dictionary and shouldn't be asked to.
2. **Channel 2** — walk dupSort dbis with `getRange`/`getValues` so duplicates are preserved.
3. **Channel 3** — decide the contract. Either `copy-db` copies `<hdbBasePath>/blobs/<databaseName>` alongside the dbis, or it refuses to run against a database that has blob attributes, or it prints a loud warning naming the directory the operator has to copy themselves. Silently emitting a copy that cannot be restored is the one option to rule out. Whatever is chosen, `compactOnStart`'s same-name/same-root assumption should become an assertion rather than a coincidence.
4. **Channel 4** — `targetEnv.openDB(...)`, and `await` the call.

## Verification anchors

Channels 1–3 have QA specs with the measurements quoted above; they should land as regression tests with the fixes. Channel 1's test must use short attribute names, since realistic ones don't reproduce it.

## Provenance

Found by the qa-explorer exploratory QA loop (findings F-180, F-178, F-184). Channel 4 found on source review while consolidating them. Source lines cited against `main` @ `b852a722c`.

kriszyp:
Fixes #2048.

`copyDb()`/`copyDbi()` in `bin/copyDb.ts` — behind the `copy-db` CLI verb and `storage.compactOnStart` — degraded the copy five independent ways while logging `copied N entries` and exiting 0. Each is a separate defect in the same ~145-line function; each is fixed on its own terms rather than by patching the one heuristic they shared.

| | Defect | Fix |
|---|---|---|
| 1 | A `value.length < 14` tombstone test also matched the primary DBI's **shared-structures dictionary** when it was small (short attribute names), so every record in the copy decoded as `null` | symbol-keyed entries are never classified; a tombstone is identified by decoding with the table's own record decoder; a primary DBI's copy fails if its dictionary did not land |
| 2 | `getKeys()` + one `getEntry()` per key yields one entry per *unique* key, **collapsing every dupSort secondary index** | one `getRange` pass per DBI, which yields every duplicate (and drops the second lookup per key) |
| 3 | The **blob store was never copied**, so the copy was unreadable anywhere but its origin | `blobs` is now a required argument: `'copy'` writes each root to `<target>-blobs/<rootIndex>/` with a README; `'preserve-source-roots'` is for the in-place replacement `compactOnStart` does |
| 4 | The **audit store was copied into the source environment** (`rootStore`, not `targetEnv`), un-awaited | both handles opened on the correct environment, raw, and awaited |
| 5 | Every recognised tombstone was dropped **regardless of age** | same `auditRetention` cutoff the runtime uses, fixed once per copy |

Channel 5 is not in the issue; I found it while fixing channel 1. Dropping a live tombstone loses the delete, so a peer that missed it can resurrect the record.

Two things worth knowing about the classifier: a length cannot decide this question in either direction (a real delete carries node-id metadata and runs to 17 bytes, past the 14 the old test used), and `RecordEncoder.decode` returning `null` is *not* sufficient either — it also returns `null` for a record whose shared structure is missing on this node (`resources/RecordEncoder.ts:470`). So the classifier requires the metadata-bearing decode only a real tombstone produces and keeps anything it cannot prove.

**Silent-success paths.** A fix for "produces a corrupt copy and exits 0" cannot keep the paths that made that possible, so these now fail the copy: a per-record error, an exhausted resume, a target that already exists (it was opened and merged into), and a failed compaction backup (it overwrote the only good copy). The retry bound drops from 10,000,000 to 1,000, resume no longer advances past the key it failed on — it used to bump a string key to `<prefix>z`, skipping everything in between and then reporting success — and a partial copy is removed. `compactOnStart` now skips a RocksDB database, skips one whose tables span multiple environments (it would relocate tables and strand blobs), and rolls back only backups it created this run.

## Where to look

- `bin/copyDb.ts:394` `isDeletedRecord` — the classifier, and the one place a bug still silently deletes data. It runs only for values whose trailing byte is msgpack nil.
- `bin/copyDb.ts:476` the resume path. Retrying from the same key relies on a re-put of identical bytes being a no-op (and a dupSort pair being a set). A permanently unreadable key now fails the copy after 1,000 attempts instead of skipping a key range.
- `bin/copyDb.ts:206` `useRawBytes`. This mutates the handle `openDB` returned. It is safe because lmdb-js constructs a new `LMDBStore` per `openDB` call (`node_modules/lmdb/open.js:415`, no instance cache) — the pre-existing code depended on the same fact. Both outside reviewers' first pass flagged this as mutating the live store; it does not, and the live `primaryStore` the classifier decodes with keeps its decoder.
- `dataLayer/blobBackup.ts` — `snapshotBlobs`'s copy core is extracted as `copyBlobRootsByIndex` and its staging directory moves from `blobs/.tmp-<id>` to `<snapshotDir>.tmp` (same filesystem either way, so the rename stays atomic). `blobsReadmeContent`'s `archive` boolean becomes a `variant`.
- `unitTests/bin/copyDB.test.js:26` — this suite never ran: it gated on a config value that is unset under mocha, so it skipped in both engine runs. That is why none of this was caught. Fixing the gate surfaced two pre-existing failures, and the `85%`-compaction assertions turned out to be measuring the audit log being dropped — they are replaced with copy-not-larger-than-source plus record readability.

## Verification

Route (b), new integration-grade regression coverage in the unit suite (it drives real LMDB environments, real tables and real blob files), plus (a) the repaired existing suite.

`unitTests/bin/copyDbIntegrity.test.js` (10 cases) — `HARPER_STORAGE_ENGINE=lmdb npm run test:unit:bin` → **178 passing**; default engine → **156 passing, 7 pending**; `npm run test:unit:backup` → **77 passing**; `npm run test:unit:dataLayer` → **245 passing**.

Reads go to the copy at its own path, opened with the same `OpenDBIObject` the runtime uses. Swapping the copy over the source path proves nothing: lmdb-js returns the already-open environment for a path, and the live stores answer point reads from cache — an early version of these tests passed for that reason.

Fails-on-base (same tests against `origin/main`, quantified):

| test | on base | on branch |
|---|---|---|
| records readable with a small dictionary | **0 / 3000** | 3000 / 3000 |
| dupSort index entries for one value | **1** | 20 |
| blob files beside the copy | **none** (no `-blobs`) | all, byte-identical |
| restore as a different database and read the attachment | n/a (nothing to restore) | byte-exact |
| copy has an audit store | **no** — and every entry logged `Illegal extended type` while being written into the *source* | full log, source untouched |
| expired tombstone purged | **no** | yes |

`npm run test:unit:main` cannot run on this machine (a local Harper instance holds the RocksDB system lock; it fails at module load, before any test) — relying on CI for it.

## Open items

- No fault-injection test for the new failure paths (partial-copy cleanup, resume exhaustion). Both outside reviewers asked for one; `AGENTS.md` forbids new sinon/stub-based tests, and I could not reach those paths through the real modules without stubbing, so I left it rather than take the shortcut. The paths are small and the review verified them by reading.
- Not covered: more than 5,000 duplicates under a single dupSort key. The 5,000-outstanding-write await fence itself is crossed by the 3,000-record fixture.
- `compactOnStart` can now fail a startup where it previously produced a bad copy and continued. That is intended, and the config flag is cleared before the work, so it will not retry-loop.
- Docs companion: [Document copy-db's blob companion directory and restore steps](https://github.com/HarperFast/documentation/pull/620).

Generated by Claude Opus 5. Reviewed pre-push by Codex (graded) + Gemini + a Harper-domain adjudication pass; two Gemini blockers were verified false and dropped (documented above), and every kept production finding is fixed in `4ecb258`.

## Rebase note (2026-08-26)

Rebased onto latest `main` (was 214 commits behind; `main` had independently landed the
same blob-classification feature this PR also touches in `dataLayer/blobBackup.ts`).
Conflicts were in `dataLayer/blobBackup.ts` only (the `copyBlobRootsByIndex` extraction
this PR's first commit adds vs. `main`'s own classify/substitute-marker feature landing in
the same function) — resolved by keeping both: the reusable `copyBlobRootsByIndex(destDir,
blobRoots)` signature this PR needs for `copy-db`, generalized so its warning message no
longer references a `backupId` that's out of scope for a non-backup caller.

While rebasing, the pre-push review CLI (codex + gemini + cursor-grok + a Harper-domain
adjudication pass) surfaced two more, unrelated real bugs on top of the rebase, both fixed
and covered by a regression test:

- `isWithin()`'s `startsWith('..')` containment check was character-based, not
  segment-based, so a real child directory whose name happens to start with `..` (e.g. a
  copy target literally named `..copy.mdb`) was misclassified as *outside* the source blob
  root — letting a copy's blob destination land inside the very root it walks. Fixed to
  compare only the first path segment; regression test added
  (`unitTests/bin/copyDbIntegrity.test.js`: "rejects a blob-copy target inside the source
  blob root even when its name starts with '..'").
- The same containment check only validated the final blob destination, not the other
  paths a failed copy actually removes (`targetDatabasePath`, its `-lock` sibling, the
  blob staging directory's `.tmp` sibling) — widened to check all of them.

## For the human reviewer

The domain-adjudicated review left these open (not fixed here — out of this rebase's
scope, and some are pre-existing / design tradeoffs the PR author should weigh in on):

- **minor, in-scope** — a blob-copy where every configured root is missing still warns
  and returns success rather than failing closed (`bin/copyDb.ts` `copyDatabaseBlobs`).
  Ambiguous against the legitimate "no blobs yet" case, so I only added a warning rather
  than unilaterally deciding to fail closed.
- **minor, in-scope** — the final database/blob copy targets are reserved by a
  non-atomic `existsSync` check, not an atomic reservation or stage-then-rename; two
  concurrent `copy-db` invocations to the same target can race.
- **minor, in-scope** — the containment check is still purely lexical (no `realpath`),
  so a symlinked path component could in theory alias into a blob root.
- **minor, pre-existing** — a legacy audit store in its own environment is never copied
  (`sourceAuditStore` ends up falsy for it), so its copy silently ships with no audit log.
- **minor, pre-existing** — compaction stages its copy inside the scanned database
  directory; a crash between the copy and the two renames leaves a `<db>-copy.mdb` that
  the boot scanner loads as a live database.
- A second review round's Gemini leg flagged `verifyStructuresCopied` awaiting a
  DBI-shared `written` promise as possibly stale across DBI boundaries
  (`bin/copyDb.ts` around the `copyDbi`/`verifyStructuresCopied` closure). I traced the
  control flow and believe it's correct — `written` is only read synchronously right
  after its owning `copyDbi` call resolves, before anything else can reassign it — but
  flagging for a second pair of eyes since domain adjudication didn't run on this delta
  round to confirm.

Pre-existing (not introduced by this rebase, reproduced on `main`-tip base too):
`unitTests/bin/copyDbIntegrity.test.js` has 3 failing tests around tombstone
version/localTime preservation and audit-log-into-target — same 3 failures on the
unrebased branch tip against its old base.

Independent review: codex (graded) + gemini + cursor-grok + harper-domain, round 1 full;
codex + gemini, round 2 delta. Receipt at `162cc49b2609`.

<sub>Review-Coverage: authored=claude; ran=codex,gemini; declined=cursor-grok,cursor-composer,domain; rounds=2 @ 162cc49b2609</sub>

<sub>Human-Review-Need: 4 @ 162cc49b2609</sub>

### DIFF
--- a/bin/copyDb.ts
+++ b/bin/copyDb.ts
@@ -75,6 +75,17 @@ export async function compactOnStart() {
 
 			const backupDest = join(rootPath, 'backup', databaseName + '.mdb');
 			const copyDest = join(rootPath, DATABASES_DIR_NAME, databaseName + '-copy.mdb');
+			const copyDatabaseName = databaseName + '-copy';
+			const copyDatabaseRootStores = databases[copyDatabaseName] && getRootStores(databases[copyDatabaseName]);
+			if (
+				copyDatabaseRootStores &&
+				[...copyDatabaseRootStores].some((rootStore) => relative(copyDest, rootStore.path) === '')
+			) {
+				const message = `Skipping compaction of database ${databaseName}: ${copyDatabaseName} is an existing database at the compaction target path; rename it before retrying`;
+				hdbLogger.warn(message);
+				console.warn(message);
+				continue;
+			}
 			let recordCount = 0;
 			try {
 				recordCount = await getTotalDBRecordCount(databaseName);
--- a/bin/copyDb.ts
+++ b/bin/copyDb.ts
@@ -6,7 +6,7 @@ import {
 	toRocksCompression,
 } from '../resources/databases.ts';
 import { open, asBinary } from 'lmdb';
-import { isAbsolute, join, relative } from 'path';
+import { isAbsolute, join, relative } from 'node:path';
 import { move, remove } from 'fs-extra';
 import { existsSync, mkdirSync } from 'node:fs';
 import { rename, writeFile } from 'node:fs/promises';
@@ -385,18 +385,19 @@ async function copyDbEnvironment(
 		}
 
 		/**
-		 * Whether a stored value is a delete tombstone, decided by decoding it with the live table's
+		 * Return a delete tombstone's local retention timestamp, decided by decoding it with the live table's
 		 * record decoder. Length cannot decide it — a small shared-structures dictionary and a small
 		 * record both land in a tombstone's usual size range — and `decode` returning null is not
 		 * sufficient either, since it also returns null for a record whose shared structure is missing
 		 * on this node. Only a metadata-bearing decode proves a tombstone; anything unprovable is kept.
 		 */
-		function isDeletedRecord(value, primaryStore) {
-			if (!primaryStore?.decoder) return false;
+		function getDeletedRecordTime(value, primaryStore, version) {
+			if (!primaryStore?.decoder) return;
 			try {
-				return primaryStore.decoder.decode(value) === null && lastMetadata?.value === null;
+				if (primaryStore.decoder.decode(value) === null && lastMetadata?.value === null)
+					return lastMetadata.localTime ?? version;
 			} catch {
-				return false;
+				return;
 			}
 		}
 
@@ -421,13 +422,11 @@ async function copyDbEnvironment(
 							// removes it too: dropping a live one loses the delete, letting a peer that
 							// missed it resurrect the record. A tombstone's body is a lone msgpack nil, so
 							// the trailing byte keeps the decode off every other record.
-							if (
-								isPrimary &&
-								typeof key !== 'symbol' &&
-								value?.[value.length - 1] === MSGPACK_NIL &&
-								version < tombstoneCutoff &&
-								isDeletedRecord(value, primaryStore)
-							) {
+							const deletedRecordTime =
+								isPrimary && typeof key !== 'symbol' && value?.[value.length - 1] === MSGPACK_NIL
+									? getDeletedRecordTime(value, primaryStore, version)
+									: undefined;
+							if (deletedRecordTime != null && deletedRecordTime < tombstoneCutoff) {
 								skippedRecord++;
 								continue;
 							}
@@ -463,15 +462,6 @@ async function copyDbEnvironment(
 						}
 					}
 					completed = true;
-					console.log(
-						'finish copying, copied',
-						recordsCopied,
-						'entries, skipped',
-						skippedRecord,
-						'delete records,',
-						bytesCopied,
-						'bytes'
-					);
 				} catch (error) {
 					// Resume from the last key read, never past it: re-copying that key is idempotent (an
 					// identical put, and an identical dupSort pair, is a no-op) while advancing the key
@@ -492,13 +482,22 @@ async function copyDbEnvironment(
 				throw new Error(
 					`Copy of ${sourceDatabase} to ${targetDatabasePath} failed on ${failedRecords} entry/entries after copying ${recordsCopied}`
 				);
+			console.log(
+				'finish copying, copied',
+				recordsCopied,
+				'entries, skipped',
+				skippedRecord,
+				'delete records,',
+				bytesCopied,
+				'bytes'
+			);
 		}
 
 		await written;
 		console.log('copied database ' + sourceDatabase + ' to ' + targetDatabasePath);
 	} finally {
 		transaction.done();
-		targetEnv.close();
+		await targetEnv.close();
 	}
 }
 
--- a/bin/copyDb.ts
+++ b/bin/copyDb.ts
@@ -6,16 +6,17 @@ import {
 	toRocksCompression,
 } from '../resources/databases.ts';
 import { open, asBinary } from 'lmdb';
-import { join } from 'path';
+import { isAbsolute, join, relative } from 'path';
 import { move, remove } from 'fs-extra';
 import { existsSync, mkdirSync } from 'node:fs';
-import { rename } from 'node:fs/promises';
+import { rename, writeFile } from 'node:fs/promises';
 import { get } from '../utility/environment/environmentManager.ts';
 import OpenEnvironmentObject from '../utility/lmdb/OpenEnvironmentObject.ts';
 import { OpenDBIObject } from '../utility/lmdb/OpenDBIObject.ts';
 import { INTERNAL_DBIS_NAME, AUDIT_STORE_NAME } from '../utility/lmdb/terms.ts';
 import { CONFIG_PARAMS, DATABASES_DIR_NAME, MIGRATING_DIR_SUFFIX } from '../utility/hdbTerms.ts';
-import { AUDIT_STORE_OPTIONS } from '../resources/auditStore.ts';
+import { AUDIT_STORE_OPTIONS, auditRetention } from '../resources/auditStore.ts';
+import { blobsReadmeContent, copyBlobRootsByIndex } from '../dataLayer/blobBackup.ts';
 import { describeSchema } from '../dataLayer/schemaDescribe.ts';
 import { updateConfigValue } from '../config/configUtils.ts';
 import * as hdbLogger from '../utility/logging/harper_logger.ts';
@@ -26,6 +27,7 @@ import {
 	beginPendingMigrationBlobSaves,
 	encodeBlobsWithFilePath,
 	endPendingMigrationBlobSaves,
+	getBlobPathsForDatabaseName,
 } from '../resources/blob.ts';
 import {
 	RecordEncoder,
@@ -50,15 +52,26 @@ export async function compactOnStart() {
 		for (const databaseName in databases) {
 			if (databaseName === 'system') continue;
 			if (databaseName.endsWith('-copy')) continue; // don't copy the copy
-			let dbPath;
-			for (const tableName in databases[databaseName]) {
-				dbPath = databases[databaseName][tableName].primaryStore.path;
-				break;
-			}
-			if (!dbPath) {
+			const rootStores = getRootStores(databases[databaseName]);
+			if (rootStores.size === 0) {
 				console.log("Couldn't find any tables in database", databaseName);
 				continue;
 			}
+			if ([...rootStores].some((rootStore) => rootStore instanceof RocksDatabase)) {
+				console.log('Database', databaseName, 'is RocksDB, which compacts itself, skipping');
+				continue;
+			}
+			// Compaction replaces one environment file, and leaves the blob roots alone because the
+			// compacted copy goes back to this exact path under this database name. A database whose
+			// tables live in separate environments has no single file to replace, so compacting it would
+			// relocate tables and strand blobs — skip it rather than produce that (harper#2048).
+			if (rootStores.size > 1) {
+				const message = `Skipping compaction of database ${databaseName}: its tables span ${rootStores.size} storage environments (table-specific paths), which compaction cannot replace as one file`;
+				hdbLogger.warn(message);
+				console.warn(message);
+				continue;
+			}
+			const dbPath = [...rootStores][0].path;
 
 			const backupDest = join(rootPath, 'backup', databaseName + '.mdb');
 			const copyDest = join(rootPath, DATABASES_DIR_NAME, databaseName + '-copy.mdb');
@@ -77,14 +90,17 @@ export async function compactOnStart() {
 				recordCount,
 			});
 
-			await copyDb(databaseName, copyDest);
+			// A copy target left behind by an interrupted run would be opened and merged into, mixing
+			// stale entries into this compaction's output.
+			await remove(copyDest);
+			await remove(copyDest + '-lock');
 
+			await copyDb(databaseName, copyDest, { blobs: 'preserve-source-roots' });
+
+			// The backup is the only rollback for the overwrite below, so a failed backup fails the
+			// compaction (leaving the source in place) instead of overwriting the only copy.
 			console.log('Backing up', databaseName, 'to', backupDest);
-			try {
-				await move(dbPath, backupDest, { overwrite: true });
-			} catch (error) {
-				console.log('Error moving database', dbPath, 'to', backupDest, error);
-			}
+			await move(dbPath, backupDest, { overwrite: true });
 			// Move compacted DB to back to original DB path
 			console.log('Moving copy compacted', databaseName, 'to', dbPath);
 			await move(copyDest, dbPath, { overwrite: true });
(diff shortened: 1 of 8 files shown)


ldt1996:
I reviewed this with some depth and the substance holds up. What I checked:

I ran the new integrity suite on this branch with the pinned dependencies (lmdb 3.5.6, rocksdb-js 2.7.0): 10/10 passing. I also fault-injected a per-record failure into copyDbi and confirmed the fail-loud path end to end: the copy rejects with "failed on 1 entry/entries" and the partial target, its lock file, and the blob directory are all cleaned up. That exercises part of the fault-injection gap listed in the open items.

The two claims carrying the most risk both verify against the code: decode() resets lastMetadata as its first statement and both of its catch paths return null without assigning it, so the tombstone classifier cannot false-positive on stale state from a previous decode; and lmdb 3.5.6, the pinned version, constructs a fresh store on every openDB call, so useRawBytes never touches a live handle.

One ask before merge: this branch's merge base (Aug 6) predates main's recent changes to bin/copyDb.ts (the verifyMigratedDatabase handle cleanup and the copyDbToRocks WAL policy) and auditStore.ts. The regions look disjoint, but please rebase so CI runs against the merged result.

bin/copyDb.ts
```diff
@@ -269,8 +459,10 @@ export async function copyDb(sourceDatabase: string, targetDatabasePath: string)
 								targetDatabasePath,
 								error
 							);
+							break;
 						}
 					}
+					completed = true;
```
ldt1996, line 489: When a record fails to copy, the per-record catch takes the break, lands here, and prints the "finish copying, copied N entries" success line before the failedRecords check below throws. I reproduced it by injecting a fault on one key: the log reads "Error copying record 5" immediately followed by "finish copying, copied 6 entries", and then the copy fails. Gating this log on failedRecords === 0, or moving it below the two checks, would keep the log consistent with the outcome.

unitTests/bin/copyDbIntegrity.test.js
```diff
@@ -0,0 +1,328 @@
...
+		for (const copy_env of open_copies) {
+			try {
+				copy_env.close();
+			} catch {
+				// a copy env that never opened, or already closed, is nothing to clean up
+			}
+		}
+	});
+
+	after(async () => {
```
ldt1996, line 140: after() restores the env properties and deletes the fixture storage, but never clears the databases cache, so everything this suite registered stays behind for the rest of the mocha process. The mid-test resetDatabases() rescan makes it broader than the fixture databases themselves: every copy this suite writes into the fixture storage_path gets registered as a database of its own. I ran the suite followed by a probe in the same process: getDatabases() returned seven entries (copy-integrity, copy-integrity-restored, duplicates, expired-tombstone, fresh-tombstone, short-attributes, with-blobs), each with stores pointing at deleted files. Adding resetDatabases() at the end of after(), after the property restores, empties the cache for whatever runs next; the suite still passes with it.

bin/copyDb.ts
```diff
@@ -209,36 +350,84 @@ export async function copyDb(sourceDatabase: string, targetDatabasePath: string)
...
+			while (!completed && retries-- > 0) {
 				try {
-					for (const key of sourceDbi.getKeys({ start, transaction })) {
+					// getRange, not getKeys + getEntry: on a dupSort index the latter yields one entry per
+					// unique key, dropping every duplicate
+					for (const { key, value, version } of sourceDbi.getRange(
+						isPrimary ? { start, transaction, versions: true } : { start, transaction }
+					)) {
 						try {
 							start = key;
```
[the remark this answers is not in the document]
kriszyp: I kept this fail-closed intentionally. A missing raw value is a corrupt or unexpected entry; silently incrementing skippedRecord would produce an incomplete copy, which is the failure mode this PR is designed to prevent. The optional chaining avoids the direct length access during classification, and any later failure is caught per-record and surfaced by the explicit failedRecords error.

bin/copyDb.ts
```diff
@@ -6,16 +6,17 @@ import {
 	toRocksCompression,
 } from '../resources/databases.ts';
 import { open, asBinary } from 'lmdb';
-import { join } from 'path';
+import { isAbsolute, join, relative } from 'path';
```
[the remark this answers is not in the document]
kriszyp: Addressed in 0dc16c457: the builtin import now uses node:path.

bin/copyDb.ts
kriszyp: Addressed in 0dc16c457: the success log now runs only after the failed-record and retry-exhaustion checks pass.

unitTests/bin/copyDbIntegrity.test.js
kriszyp: Addressed in 0dc16c457: after() now restores the environment properties and calls resetDatabases() so subsequent suites do not retain this fixture cache.

PR #2111, opened 2026-08-07
fix: EPIPE self-feeding uncaughtException loop on a closed stdio pipe (#2106)

### LINKED ISSUE #2106: Logger EPIPE self-feeding loop: closed stderr pipe turns uncaughtException handler into an unbounded log writer (~29GB/hr)
### Summary

When a Harper process's stderr pipe closes (e.g. the parent that spawned `harper.js dev` with piped stdio dies), every subsequent `console.error` throws `EPIPE` — and the `uncaughtException` handler then logs that EPIPE **back to stderr** through the logger's `process.stderr.write` override. The write throws again, re-entering the handler: a self-feeding loop that never terminates.

Observed impact (Harper **5.1.22**, macOS, `harper.js dev`): three orphaned instances each wrote a ~29GB `log/hdb.log` in under an hour (~8MB/s of identical stack traces), driving the host to 99% disk. The log throttle engaged — `Maximum log buffer rate reached, logs will be throttled` appears — but throttling only slows the mirror to file; it does not break the stderr loop.

### Captured evidence

Repeating unit, verbatim from `log/hdb.log` (path prefix sanitized):

```
2026-08-06T16:46:40.057Z [main/0] [error]: uncaughtException Error: write EPIPE
    at afterWriteDispatched (node:internal/stream_base_commons:159:15)
    at writeGeneric (node:internal/stream_base_commons:150:3)
    at Socket._writeGeneric (node:net:1171:11)
    at Socket._write (node:net:1183:8)
    at writeOrBuffer (node:internal/streams/writable:572:12)
    at _write (node:internal/streams/writable:501:10)
    at Writable.write (node:internal/streams/writable:510:10)
    at process.stderr.write (<project>/node_modules/harper/dist/utility/logging/harper_logger.js:492:35)
    at console.value (node:internal/console/constructor:313:16)
    at console.error (node:internal/console/constructor:444:26) {
  errno: -32,
  code: 'EPIPE',
  syscall: 'write'
}
```

The loop is visible in the frames: the `uncaughtException` report itself goes through `console.error` → `harper_logger.js:492` `process.stderr.write` → EPIPE → `uncaughtException` → …

### Repro recipe (derived from the mechanism; the capture above is from a live occurrence, not this script)

1. Spawn `harper.js dev` with `stdio: ['pipe','pipe','pipe']` from a parent process.
2. SIGKILL the parent (or close the stderr read end).
3. Induce any stderr-bound log (an error-level event).
4. `log/hdb.log` grows without bound with the repeating unit above.

### Suggested fix

In the logger's stderr/stdout write paths: catch `EPIPE` (and `ERR_STREAM_DESTROYED`) and permanently disable that sink for the process rather than throwing — an unwritable stderr is a normal orphan condition, not an exception the process can act on. The `uncaughtException` handler additionally should not re-log via a sink that just threw. Possibly belongs under the logging-hardening epic #1661.

kriszyp:
## Summary

Fixes [#2106](https://github.com/HarperFast/harper/issues/2106) — a self-feeding `EPIPE` loop that grows `log/hdb.log` without bound (~29GB/hr observed) whenever a Harper process's stdio pipe closes out from under it (e.g. the parent that spawned `harper.js dev` with piped stdio dies).

**Mechanism:** once stderr is unwritable, logging the failure writes to the same broken sink, which fails again, which re-enters the same recovery path — forever.

## Change

The fix lives entirely in `utility/logging/harper_logger.ts`; `server/threads/threadServer.js` and `server/threads/socketRouter.ts` are untouched (see "Design history" below for why an earlier version of this PR touched them and why that approach was abandoned).

- `installStdioGuard()` (replacing the unconditional half of `stdioLogging()`) now does two things for both `process.stdout` and `process.stderr`:
  1. **Wraps the write call in `try`/`catch`** — catches a *synchronous* throw (a closed TTY reporting `EIO`, or an already-destroyed stream), classifies it via `isStdioBrokenError()`, and calls `disableStdio()` instead of rethrowing.
  2. **Registers an `'error'` listener directly on the stream** — `process.stdout`/`stderr` backed by a POSIX pipe (exactly the case a piped-stdio parent dying produces) report a broken pipe *asynchronously*: `write()` returns normally and the `EPIPE` arrives later as an `'error'` event, past any `try`/`catch` around the write call. An unhandled `'error'` event on any `EventEmitter` becomes an `uncaughtException`, so this listener is what actually intercepts that case — with the same guaranteed attribution as the write-site catch, since it's registered on our own stdio stream rather than guessed at from a process-wide handler.
- Both paths call the same `disableStdio()`, which now just does `nativeStdWrite = noopWrite` — sufficient because the write wrapper (always installed, regardless of `logging.file`) always delegates to it. `noopWrite` honors `Writable.write`'s contract (returns `true`, invokes any trailing callback) instead of silently dropping both, which would read as permanent backpressure to a `pipe()`'d source.
- The log-file tee survives a broken pipe: the wrapper itself is never replaced, only the native write it falls through to, so `console.*`/`logger.*` output keeps reaching the file after stdio goes silent.

## Design history (why this took a few iterations)

Cross-model review (Codex + Gemini + Harper-domain adjudication, four rounds) caught real regressions at each step, each addressed before landing on the design above:

1. **v1** matched `EPIPE`/`ERR_STREAM_DESTROYED` at the two process-level `uncaughtException` handlers, mirroring the pre-existing `EIO` guard. Review flagged this as a genuine misattribution risk: those codes are generic to *any* broken pipe/socket write, not stdio-specific, so a false match would permanently silence a healthy process's logging.
2. **v2** moved the fix to the write site instead (`stdioLogging()`'s `try`/`catch`), where attribution to our own `process.stdout`/`stderr` is guaranteed by construction — and reverted the two handler files to their original, unmodified state, since this fix supersedes the need to touch them.
3. **Live probe** (Codex, independently reproduced by me with a spawned-child test) showed `try`/`catch` alone misses the case the issue is actually about: a POSIX pipe's `EPIPE` arrives asynchronously, not as a synchronous throw. The current version adds the `'error'` listener that catches that.

## Verification

- `npx mocha unitTests/utility/logging/harper_logger.test.js` — 104 passing, including coverage for the classifier, the synchronous write-site catch (both `logging.file: true` and `false`), the async `'error'`-event path, the dropped-callback fix, and log-file tee preservation after a broken pipe.
- `npx tsc --project tsconfig.build.json --noEmit`, `npm run lint`, `npx prettier --check` — clean.
- **End-to-end verification: live smoke, recorded.** Spawned a real child process with `stdio: ['ignore', 'pipe', 'pipe']` requiring the actual built `dist/utility/logging/harper_logger.js`, destroyed the child's stdout/stderr pipes, then had it call `console.error()` repeatedly:
  - Against the **unfixed baseline** (`origin/main`): crashes immediately — `uncaughtException: EPIPE write EPIPE` — matching the issue's own captured evidence.
  - Against **this fix**: survives 5+ subsequent `console.error()` calls and exits cleanly (code 0).
  - I did not add this as a checked-in `integrationTests/` file — the `@harperfast/integration-testing` package and its `harper-integration-test-run` runner aren't installed in this dev environment, so I couldn't verify a checked-in version would even run before submitting it. Happy to add one if a reviewer wants it formalized.

## For the human reviewer

Decision-Ledger:
- **decision — global-permanent-disable-vs-per-stream** — `disableStdio()` disables *all* stdio process-wide, permanently, on the first classified error on *either* stream, rather than disabling only the offending one. This is not new — the original code's single shared `nativeStdWrite` variable already had this characteristic for the `EIO` case; this PR doesn't change it. Stakes: a dead `stdout` pipe (e.g. `harper logs | head`) would also silence a still-healthy `stderr`'s fatal diagnostics. Low real-world impact, reversible via localized per-stream state.
- **decision — guard-installed-unconditionally** — the write wrapper + `'error'` listener now install regardless of `logging.file` (previously file-logging-only). Trade-off: unconditional per-write `try`/`catch` overhead (negligible — V8 try/catch is effectively free on the non-throwing path, and this is the console-write path, not the request/DB hot path) in exchange for catching `EPIPE` even when not logging to a file. Reversible.
- **decision — no-per-stream-recovery** — once disabled, stdio stays disabled for the process's lifetime; there's no probe/re-enable path if the pipe reader comes back. Matches prior behavior and is the simplest option. Low-cost to revisit later if it matters in practice.
- **decision — no-checked-in-e2e-pipe-test** — unit tests invoke the stashed handler directly rather than driving a real closed pipe; the real spawned-child probe (see Verification) is left uncommitted rather than formalized as an `integrationTests/` file. Stakes: whether the suite itself proves the process survives a real async `EPIPE`, vs. relying on the recorded manual verification. Reversible by adding the test once the missing local dependency is resolved.

Where to look first: `installStdioGuard()` in `utility/logging/harper_logger.ts` — specifically that it registers *both* the `'error'` listener and the `try`/`catch`, and that `disableStdio()`'s effect (swapping `nativeStdWrite`) is what both paths converge on. The end-to-end verification section above is the strongest evidence this actually closes the loop; the unit tests prove the mechanism but (as review repeatedly and correctly flagged) can't reproduce a real pipe's async error delivery on their own, which is why the async-path unit test calls the installed handler directly rather than emitting on the real `process.stdout`/`stderr` (doing that would leave Node's own internal stream state altered for the rest of the test run).

The most recent commit ([044e5133](https://github.com/HarperFast/harper/commit/044e5133) — comment trims only, no behavior change) postdates the reviewed SHA (`afaea04d`, adjudicated severity: minor) and has no receipt of its own, so the footer below mechanically materializes grade 4 (missing/stale receipt is the fail-closed default) rather than reflecting an unreviewed regression — the diff since the last review is comment-only.

<sub>Human-Review-Need: 4 @ 044e5133fe23</sub>

### DIFF
--- a/utility/logging/harper_logger.ts
+++ b/utility/logging/harper_logger.ts
@@ -381,6 +381,7 @@ module.exports = {
 	inspectForLog,
 	isErrorLike,
 	disableStdio,
+	isStdioBrokenError,
 	externalLogger,
 };
 
@@ -389,6 +390,19 @@ module.exports = {
  */
 export function disableStdio(_unused?: any) {
 	nativeStdWrite = function () {}; // make this a noop
+}
+
+const STDIO_BROKEN_ERROR_CODES = new Set(['EPIPE', 'EIO', 'ERR_STREAM_DESTROYED']);
+
+/**
+ * True when this error means a stdio stream (stdout/stderr) is permanently unwritable - the
+ * pipe's reader died (EPIPE), the terminal closed (EIO), or the stream was already destroyed.
+ * uncaughtException handlers should disableStdio() and stop on this, rather than logging the
+ * error through the same sink that just threw it (which would throw again, re-entering the
+ * handler in an unbounded loop).
+ */
+export function isStdioBrokenError(error: any): boolean {
+	return STDIO_BROKEN_ERROR_CODES.has(error?.code);
 }
 
 /**
@@ -1809,6 +1823,7 @@ export default {
 	setLogLevel,
 	OUTPUTS,
 	disableStdio,
+	isStdioBrokenError,
 	externalLogger,
 	AuthAuditLog,
 	errorToString,
--- a/utility/logging/harper_logger.ts
+++ b/utility/logging/harper_logger.ts
@@ -390,6 +390,11 @@ module.exports = {
  */
 export function disableStdio(_unused?: any) {
 	nativeStdWrite = function () {}; // make this a noop
+	// stdioLogging() only routes process.stdout/stderr.write through nativeStdWrite when
+	// log_to_file is true; with it false those are still the raw native writers, so silence them
+	// directly too - otherwise this is a no-op under that config and callers keep re-throwing.
+	process.stdout.write = nativeStdWrite as any;
+	process.stderr.write = nativeStdWrite as any;
 }
 
 const STDIO_BROKEN_ERROR_CODES = new Set(['EPIPE', 'EIO', 'ERR_STREAM_DESTROYED']);
@@ -397,9 +402,7 @@ const STDIO_BROKEN_ERROR_CODES = new Set(['EPIPE', 'EIO', 'ERR_STREAM_DESTROYED'
 /**
  * True when this error means a stdio stream (stdout/stderr) is permanently unwritable - the
  * pipe's reader died (EPIPE), the terminal closed (EIO), or the stream was already destroyed.
- * uncaughtException handlers should disableStdio() and stop on this, rather than logging the
- * error through the same sink that just threw it (which would throw again, re-entering the
- * handler in an unbounded loop).
+ * Callers should disableStdio() on this rather than logging through the same sink that threw it.
  */
 export function isStdioBrokenError(error: any): boolean {
 	return STDIO_BROKEN_ERROR_CODES.has(error?.code);
--- a/utility/logging/harper_logger.ts
+++ b/utility/logging/harper_logger.ts
@@ -385,9 +385,8 @@ module.exports = {
 	externalLogger,
 };
 
-// Mimics Writable.write's contract (boolean return, optional trailing callback invoked) rather
-// than silently dropping both, which would read as permanent backpressure to any pipe()'d source
-// and leave a callback-form caller waiting forever.
+// Honors Writable.write's contract (boolean return, callback invoked) rather than a bare
+// no-op, which would read as permanent backpressure to a pipe()'d source.
 function noopWrite(_chunk?: any, encoding?: any, callback?: any) {
 	if (typeof encoding === 'function') callback = encoding;
 	if (typeof callback === 'function') callback();
@@ -398,8 +397,6 @@ function noopWrite(_chunk?: any, encoding?: any, callback?: any) {
  * We call this if stdio is not functional
  */
 export function disableStdio(_unused?: any) {
-	// stdioLogging() installs process.stdout/stderr.write unconditionally and always delegates to
-	// nativeStdWrite, so replacing just this is enough to silence both regardless of log_to_file.
 	nativeStdWrite = noopWrite;
 }
 
@@ -532,8 +529,21 @@ export function initLogSettings(forceInit = false) {
 	stdioLogging();
 }
 let loggingEnabled = true;
-function stdioLogging() {
-	process.stdout.write = function (data) {
+
+// process.stdout/stderr backed by a POSIX pipe report a broken pipe asynchronously - write()
+// returns normally and the EPIPE arrives later as an 'error' event, past any try/catch around the
+// write call. An unhandled 'error' event on any EventEmitter becomes an uncaughtException, so a
+// listener here is what actually intercepts that case, with attribution guaranteed by being
+// registered directly on our own stdio stream rather than guessed from a process-wide handler.
+function installStdioGuard(stream) {
+	if (stream.harperStdioErrorHandler) stream.removeListener('error', stream.harperStdioErrorHandler);
+	stream.harperStdioErrorHandler = (error) => {
+		if (!isStdioBrokenError(error)) throw error;
+		disableStdio();
+	};
+	stream.on('error', stream.harperStdioErrorHandler);
+
+	stream.write = function (data) {
 		if (
 			log_to_file &&
 			typeof data === 'string' && // this is how we identify console output vs redirected output from a worker
@@ -545,34 +555,18 @@ function stdioLogging() {
 			writeToLogFile(data);
 		}
 		try {
-			return nativeStdWrite.apply(process.stdout, arguments);
-		} catch (error) {
-			// the write we just attempted IS process.stdout's own, so this is never a guess about
-			// where a broken-pipe error came from - disable it here rather than let it propagate
-			// into an uncaughtException that would re-log through this same broken sink
-			if (!isStdioBrokenError(error)) throw error;
-			disableStdio();
-			return true;
-		}
-	};
-	process.stderr.write = function (data) {
-		if (
-			log_to_file &&
-			typeof data === 'string' && // this is how we identify console output vs redirected output from a worker
-			loggingEnabled &&
-			logConsole
-		) {
-			if (data[data.length - 1] === '\n') data = data.slice(0, -1);
-			writeToLogFile(data);
-		}
-		try {
-			return nativeStdWrite.apply(process.stderr, arguments);
+			return nativeStdWrite.apply(stream, arguments);
 		} catch (error) {
 			if (!isStdioBrokenError(error)) throw error;
 			disableStdio();
-			return true;
+			return noopWrite.apply(stream, arguments);
 		}
 	};
+}
+
+function stdioLogging() {
+	installStdioGuard(process.stdout);
+	installStdioGuard(process.stderr);
 }
 
 export function loggerWithTag(tag: string, conditional?: boolean, logger: any = mainLogger) {
--- a/utility/logging/harper_logger.ts
+++ b/utility/logging/harper_logger.ts
@@ -385,8 +385,8 @@ module.exports = {
 	externalLogger,
 };
 
-// Honors Writable.write's contract (boolean return, callback invoked) rather than a bare
-// no-op, which would read as permanent backpressure to a pipe()'d source.
+// A bare no-op would read as permanent backpressure to a pipe()'d source; this drains and
+// discards instead.
 function noopWrite(_chunk?: any, encoding?: any, callback?: any) {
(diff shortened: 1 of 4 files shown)


utility/logging/harper_logger.ts
```diff
@@ -509,32 +529,42 @@ export function initLogSettings(forceInit = false) {
...
 }
 let loggingEnabled = true;
+
+// A POSIX pipe reports a broken pipe asynchronously - write() returns normally and the EPIPE
+// arrives later as an 'error' event, past the try/catch below. An unhandled 'error' event on any
+// EventEmitter becomes an uncaughtException, so the listener is what actually catches that case.
+function installStdioGuard(stream) {
+	if (stream.harperStdioErrorHandler) stream.removeListener('error', stream.harperStdioErrorHandler);
+	stream.harperStdioErrorHandler = (error) => {
+		if (!isStdioBrokenError(error)) throw error;
```
DavidCockerill, line 539: Non-blocking, but worth seeing before merge: this rethrow leaves the #2106 loop reachable under any stdio error code outside the three-code set — rethrow → `uncaughtException` → handler logs → `process.stderr.write` → fresh error event → rethrow, unbounded.

I built it rather than asserting it, using this PR's logic verbatim in a spawned child with fd 2 closed under the process instead of the reader going away:

```
--- closing fd 2 ---
  err-listener EBADF        <- not in the set -> rethrown
uncaught #1 EBADF write EBADF
  err-listener EBADF
uncaught #2 EBADF write EBADF
...                          exit=3 (my >40-iteration bail), 83 trace lines
```

The shape I'd suggest: **never rethrow from this listener, but keep `isStdioBrokenError()` as the trigger for `disableStdio()`** — so a non-allowlisted error is swallowed *without* going dark. That closes the loop (nothing feeds the error back into the broken sink), and it keeps the classifier's conservatism about permanently killing console output, which a bare catch-all would give up.

The case for it: the allowlist exists because of the misattribution risk correctly identified against v1's process-level handlers, and that reason doesn't apply here. This listener is registered on our own `process.stdout`/`process.stderr` object, so attribution is guaranteed by construction — which is the PR's own justification for moving to this site. The classification is vestigial here, and it's the only thing keeping the loop reachable. It's also stricter than Node's own posture: `console`'s default `ignoreErrors: true` swallows every stdio write error on these same two streams.

The case against blocking: I couldn't tie EBADF (or any other non-allowlisted async code) to a plausible trigger in this codebase, and the only other realistic pipe-side code is ECONNRESET, which `threadServer.js:73` and `socketRouter.ts:15` both short-circuit before logging. So it stays closed today — but "#2106 cannot recur" now spans three files and depends on two unrelated early-returns staying put. Your call between simplifying the listener and leaving a comment recording the dependency. The sync `catch` below can keep its allowlist either way, since a rethrow there terminates rather than loops.

One more reason to prefer not throwing here, which I'd missed on the first pass: this listener throwing also breaks EventEmitter coexistence. Listeners run in registration order and a throw from the first aborts the rest — verified, not assumed: with a second listener registered after this one, an allowlisted EPIPE runs both, while a non-allowlisted EBADF escapes and the second never runs. Nothing registers an `'error'` listener on `process.stdout`/`stderr` today (I swept the repo and `node_modules`), so it's latent rather than live — but it's the same line, and not rethrowing fixes both at once.

utility/logging/harper_logger.ts
```diff
@@ -509,32 +529,42 @@ export function initLogSettings(forceInit = false) {
...
-				logConsole
-			) {
-				if (data[data.length - 1] === '\n') data = data.slice(0, -1);
-				writeToLogFile(data);
-			}
-			return nativeStdWrite.apply(process.stderr, arguments);
-		};
-	}
+	installStdioGuard(process.stdout);
+	installStdioGuard(process.stderr);
```
DavidCockerill, line 567: Non-blocking, and I'd take it separately rather than grow this PR — but I read the shared-`nativeStdWrite` decision as newly *widened* rather than unchanged, so the ledger entry may want updating.

The single module-scoped writer backs both streams, so the first classified error on either disables both. That variable genuinely is pre-existing, but pre-PR it could only fire on EIO — terminal closed, both streams really are dead, so the coupling was harmless. Adding EPIPE makes it fire when one pipe's reader exits while the other is a live terminal (`harper … | head` being the everyday shape), and the stream you lose is the one carrying the errors.

Localized fix if you want it: stash the writer per stream (`stream.harperNativeWrite`) instead of in one module variable.

utility/logging/harper_logger.ts
```diff
@@ -381,14 +381,34 @@ module.exports = {
...
+	if (typeof callback === 'function') callback();
+	return true;
+}
+
 /**
  * We call this if stdio is not functional
  */
 export function disableStdio(_unused?: any) {
-	nativeStdWrite = function () {}; // make this a noop
+	nativeStdWrite = noopWrite;
```
DavidCockerill, line 400: Non-blocking: nothing records that stdio was disabled, or why. The next person debugging "Harper stopped logging to the console" has zero evidence — and this PR makes that state reachable from a much more common trigger than the old EIO-only path did. Having just fixed "logs explode", it'd be a shame to leave "logs vanish, silently and permanently" undiagnosable.

A one-shot line through `writeToLogFile()` here would cover it, and it's re-entrancy-safe by construction: that sink is the file, not the stream that just failed. Guard for `log_to_file`/`writeToLogFile` being unset and it's a few lines.

DavidCockerill (approved):
Approving. I verified the mechanism by rebuilding it rather than reading the diff — spawned a real child with real destroyed pipes and drove both shapes. The pre-PR shape bails at **>100 uncaughtException iterations** (203 trace lines); this PR's logic exits 0 with **zero**. The loop is broken, not narrowed: two independent reasons the wheel can't turn twice — the `error` listener means the EPIPE never becomes an `uncaughtException` at all, and `disableStdio()` leaves the wrapper in place while swapping in a noop, so any re-entry writes into a discard.

Your `'error'`-listener diagnosis is exactly right, and the try/catch alone would not have sufficed. Also worth recording *why* the bug looks unreproducible from `console.error`: Node's console passes an `errorHandler` that re-arms a `once('error', noop)`, which the callback-free `logStdErr` write at `:537`/`:542` bypasses. That's the path that loops.

The strongest design move is moving the classification to the write site. v1 matched the same codes at the process-level handlers, where EPIPE is generic to any broken socket; here both sites are attributed to Harper's own stdio by construction, which turns a heuristic into a guarantee. Genuine fatals still terminate — no process-level handler is installed or modified, and a throw inside an `uncaughtException` handler exits 7 (verified). The guard is per-thread, so one thread can't silence another. Incidental win: `noopWrite` returning `true` fixes a latent worker-output pipe stall the old `function(){}` noop would have caused.

Three non-blocking notes in threads, one of them a demonstrated variant of the same loop.

PR #2119, opened 2026-08-07
fix(blob): re-encode inline blobs from the copied buffer, not a stale read-buffer view

cb1kenobi:
## Summary

Fixes the Harper 5.2.0 Blob-column decode corruption (`Error decoding record: Data read, but end of buffer not reached`) at its **root cause**, superseding [#2103 — fix(storage): stable read buffer for cached RocksDB reads (Blob-column decode corruption)](https://github.com/HarperFast/harper/pull/2103).

When a small blob is decoded inline, `blob.ts` kept two references to the payload:

- `contentBuffer` — a **copy** (`copyingUnpacker` uses `copyBuffers: true`), stable for the blob's lifetime
- `storageBuffer` — the raw msgpack ext-body, a **view into the store's read buffer**, not copied

`pack()` had a fast-path that re-emitted `storageBuffer` verbatim on re-encode. The read buffer is recycled by later reads, so a **read-modify-write** — the shape of a REST `PATCH` (fetch the record, carry the unchanged `Blob` field, put it back) — serialized whatever foreign bytes had since overwritten that buffer. The record on disk was corrupted, and the next read failed to decode.

## Fix

Drop `storageBuffer` entirely and let `pack()` fall through to the existing `contentBuffer` re-encode path. `contentBuffer` is already a stable copy, so this is correct with **no read-path allocation cost**.

## Why this supersedes #2103

@kriszyp was right. His instrumentation found no mid-decode clobber because the corruption is on the re-**encode** of a fetched/cached blob, not during decode — a decode-path canary can't observe it. #2103's `needsStableBuffer` masked the corruption by forcing every primary-store read to allocate a fresh buffer (so the view was never recycled), which fixed the symptom at a per-read cost. This change fixes the actual defect in Harper's blob layer instead.

## Reproduction (deterministic)

`unitTests/resources/recordCacheStableBuffer.test.js` builds a 3-case matrix on a cache/VT-enabled table with a `Blob` column:

| sequence | before fix |
|---|---|
| re-put a **new** blob value (interleaved read) | passes |
| RMW carrying the fetched blob, **no** interleaved read | passes |
| RMW carrying the fetched blob **+ interleaved read of another key** | **fails 25/25** |

The third case is the bug; the corrupted bytes even contained the record's own primary-key string, confirming the buffer had been overwritten before re-encode. All three pass after the fix; the full blob suite (63 tests) is green.

## Bonus

`slice()` on an inline blob previously threw (it was gated off by `storageBuffer`'s presence); it now works via the `contentBuffer` branch.

## Why it worked in 5.1.26 but broke in 5.2.0
The blob code didn't change. `storageBuffer: buffer` and the `pack` fast-path that re-emits it are byte-identical in `v5.1.26` and `v5.2.0` (both landed in "the big lift", Oct 2025). So the blob's habit of retaining a view into the read buffer and re-serializing it on re-encode is a long-standing latent bug. What changed is the read path underneath it, which went from handing back stable buffers to recycled ones.

Two things changed in 5.2.0's primary storage, both verified:

1. `PrimaryRocksDatabase` is brand-new in 5.2.0 (absent in 5.1.26). It added the `WeakLRUCache` record cache + the VerificationTable read path (`populateVersion` / `expectedVersion` / `FRESH_VERSION_FLAG`). 5.1.26's primary read path (`handleLocalTimeForGets`) had no cache — every get decoded fresh and returned; nothing was retained across reads.
2. rocksdb-js bumped `~2.4.1` → `^2.5.0` (now 2.7.1 installed).

The mechanism, from the rocksdb-js read path `index.cjs:590`: the primary `RecordEncoder` never sets `needsStableBuffer`, so the store is marked `decoderCopies = true`, and `get()` routes through `getBinaryFast` — a reusable buffer "only valid until the next get operation." Combined with the new cache retaining decoded records (blobs and all) across reads:

* `get(A)` decodes A; the cache keeps A's record, whose blob's `storageBuffer` is a view into the reusable buffer.
* a later read of another key recycles that buffer.
* a REST PATCH (read-modify-write) re-encodes A's carried-over blob → serializes the now-foreign bytes → the next read fails to decode.

In 5.1.26 there was nothing to retain a stale view: each `get` re-decoded from a buffer holding that record's own current bytes, so the re-encode always saw valid bytes. That's also why `cache: false` fixes it — it drops the retention/VT path, so every read re-decodes fresh and the window closes.

Confidence: high on the architecture (blob-identical, `PrimaryRocksDatabase`/cache/VT new, `decoderCopies`→`getBinaryFast` reusable buffer, `cache:false` fix all verified). The one inferred link is that rocksdb-js 2.4.x returned stable buffers where 2.5+ reuses them — I didn't run 2.4.x. But it doesn't change the fix or the conclusion: 5.2.0's new cache/VT read path activated a latent blob bug, and re-encoding from the copied `contentBuffer` is the correct place to close it regardless of how the buffer underneath behaves.

### DIFF
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -55,7 +55,6 @@ type StorageInfo = {
 	recordId?: number;
 	contentBuffer?: any;
 	source?: Readable;
-	storageBuffer?: Buffer;
 	compress?: boolean;
 	flush?: boolean;
 	start?: number;
@@ -829,7 +828,7 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
 			storageInfoForBlob.set(slicedBlob, slicedStorageInfo);
 			if (this.size != undefined)
 				slicedBlob.size = (end == undefined ? this.size : Math.min(end, this.size)) - (start ?? 0);
-		} else if (sourceStorageInfo?.contentBuffer && !sourceStorageInfo.storageBuffer) {
+		} else if (sourceStorageInfo?.contentBuffer) {
 			const slicedStorageInfo = {
 				...sourceStorageInfo,
 				contentBuffer: sourceStorageInfo.contentBuffer.subarray(start, end),
@@ -1740,10 +1739,14 @@ addExtension({
 				throw new Error('No store specified, cannot load blob from storage');
 			}
 		} else {
+			// contentBuffer is a *copy* (copyingUnpacker uses copyBuffers), so it stays valid for the
+			// lifetime of the blob. Do not retain the raw ext-body `buffer` here: it is a view into the
+			// store's read buffer, which is recycled by later reads. Keeping it and re-emitting it on
+			// re-encode (e.g. a read-modify-write / REST PATCH) serialized whatever foreign bytes had
+			// since overwritten that buffer, corrupting the record (harper#2103).
 			storageInfoForBlob.set(blob, {
 				storageIndex: 0,
 				fileId: null,
-				storageBuffer: buffer as any,
 				contentBuffer: blobInfo[1] as any,
 			});
 			blob.size = blobInfo[1]?.length;
@@ -1764,9 +1767,6 @@ addExtension({
 		if (blob.type) options.type = blob.type;
 		if (blob.size !== undefined) options.size = blob.size;
 		if (storageInfo) {
-			if (storageInfo.storageBuffer) {
-				return storageInfo.storageBuffer;
-			}
 			if (
 				storageInfo.contentBuffer &&
 				(storageInfo.contentBuffer?.length < FILE_STORAGE_THRESHOLD || blob.saveInRecord)
--- a/unitTests/resources/recordCacheStableBuffer.test.js
+++ b/unitTests/resources/recordCacheStableBuffer.test.js
@@ -1,0 +1,109 @@
+const { setupTestDBPath } = require('../testUtils');
+const assert = require('node:assert');
+const { table } = require('#src/resources/databases');
+const { setMainIsWorker } = require('#js/server/threads/manageThreads');
+
+// Reproduction for the Blob-column decode corruption ("Data read, but end of buffer not reached", or
+// a record coming back with foreign bytes) reported against Harper 5.2.0.
+//
+// Mechanism: an inline Blob decoded from a cache/VT-enabled primary store keeps `storageBuffer` as a
+// VIEW into the read buffer (not a copy). A read-modify-write — the shape of a REST PATCH — fetches
+// the record, carries the unchanged Blob field, and re-puts it; the Blob re-encodes straight from
+// `storageBuffer`. If any read in between reuses that buffer, the re-put serializes stale/foreign
+// bytes. A pure-decode canary never sees this because the clobber is on the re-ENCODE.
+describe('record cache: blob decode across put / read / read-other / re-put / read', () => {
+	let T;
+	before(async function () {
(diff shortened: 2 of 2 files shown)


resources/blob.ts
```diff
@@ -829,7 +828,7 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
 			storageInfoForBlob.set(slicedBlob, slicedStorageInfo);
 			if (this.size != undefined)
 				slicedBlob.size = (end == undefined ? this.size : Math.min(end, this.size)) - (start ?? 0);
-		} else if (sourceStorageInfo?.contentBuffer && !sourceStorageInfo.storageBuffer) {
+		} else if (sourceStorageInfo?.contentBuffer) {
```
cb1kenobi, line 831: **Nit: `slice()` on an inline blob is now reachable but untested**

Dropping `storageBuffer` makes this branch reachable for inline-decoded blobs, so `slice()` now returns a sliced in-memory blob where it previously fell through to the `else` and threw "Can not slice a streaming blob". The new path is correct (it reuses the same `contentBuffer.subarray` logic already used for created blobs), but it is a shipped behavior change with no direct coverage — a future regression would be silent.

Suggested fix: add a small assertion that slicing a decoded inline blob returns the right bytes/size, e.g.

```js
const b = (await T.get(id)).packageJson; // inline-decoded blob
const s = b.slice(0, 4);
assert.strictEqual((await s.bytes()).length, 4);
```

—
_Generated by Barber AI_

kriszyp (approved):
Approving — this is the right shape of fix: root cause, one canonical encoder, and zero added read-path cost. Nice work on the deterministic 3-case repro matrix; isolating "RMW + interleaved read" as the failing case (and confirming the clobber is on re-*encode*, not decode) is what made the root cause unambiguous.

One question we weighed while reviewing: dropping `storageBuffer` is clearly correct for safety, but is there performance value in retaining it as a *copy* rather than a view? Conclusion: no.

- **What the fast-path actually saved.** The fallback `pack([options, contentBuffer])` encodes a tiny options map and memcpys ≤8KB of content (inline blobs are bounded by `FILE_STORAGE_THRESHOLD`) into an intermediate buffer, which the ext mechanism then copies into the parent encode target. The verbatim path saved one intermediate allocation plus one small memcpy — well under a microsecond, on a write path that's about to do a full record encode and a RocksDB write. Noise.

- **Cost of a copied `storageBuffer`.** The only safe point to copy is decode time (by `pack()` time the view can already be stale — that's this bug). But `contentBuffer` is already a copy, and `storageBuffer` is a superset of it, so every inline-blob read would pay ~2× the copy and every cached blob would hold ~2× resident memory — a real tax on record-cache/VT tables. Reads vastly outnumber re-encodes, so that trades the common path for the rare one.

- **Even the single-copy variant loses** (copy the whole ext body once, derive `contentBuffer` as a subarray of it): same read cost as today and the fast path restored, but it keeps two encode paths — the verbatim re-emit freezes `options` (type/size) as-of-decode, silently dropping any property change between fetch and re-put unless you add a dirty-flag guard. That forked-encoder shape is exactly the invariant violation that produced this corruption, and it would also have kept `slice()` broken on inline blobs.

If inline-blob re-encode ever shows up in a profile, the right altitude for the optimization is record-level — skip re-encoding unchanged records entirely — not a per-blob verbatim byte path.

— KrAIs (Claude, reviewing on Kris's behalf)

PR #2124, opened 2026-08-08
fix(sql-engine): honor SELECT TOP n and floor fractional LIMIT/OFFSET

cb1kenobi:
## Summary

The new SQL engine's normalizer read only `stmt.limit`/`stmt.offset`, so two LIMIT-family behaviors silently diverged from the legacy engine. This fixes both.

### 1. `SELECT TOP n` was silently ignored
`TOP` is AlaSQL's SQL-Server spelling of `LIMIT` and parses into `stmt.top`, which the normalizer never read. In `new`/`auto` mode the query produced a **valid** plan with no limit → it returned **every** row, and because the plan was valid it never triggered the `EngineUnsupportedError` fallback to legacy. So `SELECT TOP 5 ...` quietly returned the whole table.

Fix: `normalizeSelect` now maps `stmt.top` → `limit`. Verified against the bundled AlaSQL that legacy executes `TOP n` identically to `LIMIT n`, so the mapping preserves parity. `TOP` combined with `LIMIT` is non-standard and ambiguous (legacy lets TOP win), so that combination is rejected with `EngineUnsupportedError` and defers to legacy rather than silently picking one.

### 2. Fractional `LIMIT`/`OFFSET` over-yielded
AlaSQL parses `LIMIT 2.9` as `2.9`. Legacy truncates to **2** rows, but `PhysicalLimit`'s `yielded >= cap` check yields a **3rd** row before `3 >= 2.9` trips. New `normalizeRowCount` floors `LIMIT`/`OFFSET`/`TOP` to a non-negative integer (matching legacy truncation) and defers non-finite/negative counts to legacy.

## Why it matters
Both are silent correctness divergences under the default `sql.engine: auto` cutover — the query succeeds but returns the wrong rows, with no error and no fallback. That's the worst failure mode for the engine parity effort.

## Testing
- Added 3 unit tests to `unitTests/sqlEngine/pipeline.test.js`: `TOP n` pushes `limit` to `Table.search`, `TOP`+`LIMIT` rejects, fractional `LIMIT` floors.
- Full sqlEngine unit suite green: **99 passing** (built `dist/` first).
- Not yet run: the `sql-engine-differential` integration suite — worth adding a `TOP` case there before merge.

## Notes for reviewers
Surfaced while scoping a separate REST pagination / total-count exploration; this fix is standalone and doesn't depend on that work.

---

## Update after review (Codex + Gemini)

A cross-model review pass (Codex) plus the Gemini bot surfaced further legacy-parity gaps in the same normalize path — all now handled by deferring to legacy so `auto` mode stays byte-identical to it. Verified each against bundled AlaSQL:

| Input | Legacy behavior | Pre-fix new engine | Now |
|---|---|---|---|
| `TOP n PERCENT` | fraction of rows (`TOP 50 PERCENT` of 10 → 5) | mapped to `limit=n` → all/`n` rows | reject → legacy |
| `TOP 0` / `LIMIT 0` | `0` = "no limit" → all rows | `0` rows | reject → legacy |
| count `>= 2^31` | 32-bit coercion (`LIMIT 2^32` → none; huge `OFFSET` wraps to 0) | applied literally | reject → legacy |
| fractional `OFFSET` | truncates (`OFFSET 1.9` skips 1) | applied literally | floored |

`TOP 0` was newly exposed by the TOP fix (previously `TOP` was ignored → all rows, which happened to match legacy); `LIMIT 0` and the overflow cases were pre-existing new-engine divergences fixed here for free since `readNumValue`/`normalizeRowCount` were consolidated into one `readRowCount(node, clause, min)` that accepts `[min, 0x7fffffff]` and defers the rest.

**On the Gemini note (dynamic/`?` LIMIT expressions):** the cited forms (`LIMIT ?`, `LIMIT 1+1`, subqueries) actually **parse-error in AlaSQL** before reaching the normalizer, and legacy uses the same parser — so there was no live "silent all-rows" bug. But the underlying invariant is worth enforcing, so a present-but-unreadable clause now defers to legacy rather than dropping the limit, covered by an AST-level test. (Gemini's suggested `LIMIT ?` test would itself throw a parse `SyntaxError`, not `EngineUnsupportedError`.)

Not reachable via SQL: `TOP … OFFSET` and bare `OFFSET n` both parse-error, so the legacy "TOP ignores OFFSET" path can't be hit.

Tests: sqlEngine unit suite **103 passing** (added TOP PERCENT, `LIMIT`/`TOP 0`, overflow, fractional `OFFSET`, and the AST-level invariant guard).

### DIFF
--- a/sqlEngine/parser/normalizer.ts
+++ b/sqlEngine/parser/normalizer.ts
@@ -156,8 +156,18 @@ function normalizeSelect(stmt: AlaSqlNode): SelectNode {
 
 	const where = stmt.where ? normalizeExpr(extractWhere(stmt.where as AlaSqlNode)) : undefined;
 	const orderBy = stmt.order ? (stmt.order as AlaSqlNode[]).map(normalizeSort) : undefined;
-	const limit = readNumValue(stmt.limit as AlaSqlNode | undefined);
-	const offset = readNumValue(stmt.offset as AlaSqlNode | undefined);
+	// `SELECT TOP n` is AlaSQL's SQL-Server spelling of `LIMIT n`. Without this the
+	// normalizer dropped `stmt.top` entirely, so `new`/`auto` mode returned *all*
+	// rows and — producing a valid plan — never fell back to legacy. The two clauses
+	// don't combine in standard SQL, so a query carrying both is ambiguous: defer it
+	// to legacy rather than silently pick one.
+	const rawLimit = readNumValue(stmt.limit as AlaSqlNode | undefined);
+	const top = readNumValue(stmt.top as AlaSqlNode | undefined);
+	if (rawLimit != null && top != null) {
+		throw new EngineUnsupportedError('TOP combined with LIMIT is not supported', stmt);
+	}
+	const limit = normalizeRowCount(rawLimit ?? top, 'LIMIT');
+	const offset = normalizeRowCount(readNumValue(stmt.offset as AlaSqlNode | undefined), 'OFFSET');
 	const distinct = !!stmt.distinct;
 
 	return {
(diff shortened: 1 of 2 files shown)


sqlEngine/parser/normalizer.ts
```diff
@@ -156,8 +156,18 @@ function normalizeSelect(stmt: AlaSqlNode): SelectNode {
...
 	const orderBy = stmt.order ? (stmt.order as AlaSqlNode[]).map(normalizeSort) : undefined;
-	const limit = readNumValue(stmt.limit as AlaSqlNode | undefined);
-	const offset = readNumValue(stmt.offset as AlaSqlNode | undefined);
+	// `SELECT TOP n` is AlaSQL's SQL-Server spelling of `LIMIT n`. Without this the
+	// normalizer dropped `stmt.top` entirely, so `new`/`auto` mode returned *all*
+	// rows and — producing a valid plan — never fell back to legacy. The two clauses
+	// don't combine in standard SQL, so a query carrying both is ambiguous: defer it
+	// to legacy rather than silently pick one.
+	const rawLimit = readNumValue(stmt.limit as AlaSqlNode | undefined);
+	const top = readNumValue(stmt.top as AlaSqlNode | undefined);
```
kriszyp: AlaSQL represents `SELECT TOP 50 PERCENT ...` with `top.value = 50` and a separate `stmt.percent` flag. This currently converts that query to an absolute `LIMIT 50`: for a ten-row result, legacy returns five rows while the new engine returns all ten, and `auto` mode will not fall back because the plan succeeds. Please reject `stmt.percent` with `EngineUnsupportedError` until percentage semantics are implemented, and add a regression test for this variant.

— KrAIs (Codex)

kriszyp (approved):
Good catch!

Issue #2069, opened 2026-08-04
HNSW vector search latency scales linearly with corpus size (~25-30µs/vector), not logarithmically


maurice-harper:
## Summary

HNSW vector search latency appears to scale **linearly with the number of indexed vectors** rather than logarithmically. Measured on Harper 5.1.19 across three tables in the same instance, cost is roughly **25–30 µs per indexed vector**, and is **independent of the requested neighbour count**.

At current sizes this is tolerable (~65 ms on 2.7k vectors) but it sets the ceiling for how far a vector index can grow: extrapolating, a 27k-vector table would put a single search near 650 ms.

I may well be holding it wrong — if there's a query shape or index option that changes this, I'd be glad to hear it.

## Measurements

Three tables in one instance, each `embedding: [Float] @indexed(type: "HNSW", distance: "cosine")`, 768-dim vectors (nomic-embed-text). Timing is measured **server-side inside the resource** around the search call only, so it excludes network, request parsing, and embedding of the query.

| table | indexed vectors | search time (steady state) | µs / vector |
| --- | --- | --- | --- |
| skills | 23 | **3 ms** | 130 |
| docs | 2,729 | **63–66 ms** | 24 |
| code | 4,395 | **131–142 ms** | 31 |

The query is:

```js
const results = table.search({
  sort:   { attribute: 'embedding', target: queryVector }, // 768 floats
  limit,
  select: ['id', '$distance'],
});
for await (const row of results) { /* collect */ }
```

## What this is not

- **Not a cold cache.** The same query repeated 5× stays at 63–66 ms and never drops below it. First call on a table is higher (~85–98 ms), then it settles — so there is a warm-up, but the steady state is still corpus-proportional.
- **Not a missing index.** `describe_table` confirms the index is live:
  ```json
  {"attribute":"embedding","type":"array","elements":"Float",
   "indexed":{"type":"HNSW","distance":"cosine"}}
  ```
- **Not per-returned-row cost.** Sweeping `limit` across 5 / 40 / 80 / 200 / 400 on the 2,729-vector table produced no measurable trend (65–213 ms, all noise). Cost tracks corpus size, not result count.

## A possibly related observation

On the same 2,729-vector table, a single search with `limit` **greater than the table size** returns only ~221 rows rather than everything reachable:

```
limit: 2502  →  221 rows returned
```

[...]

PR #2125, opened 2026-08-08
Remove linear-in-corpus-size work from HNSW vector search, and honor limit past the candidate list

kriszyp:
Vector search cost grew with the size of the corpus rather than with `ef`, which is what [#2069](https://github.com/HarperFast/harper/issues/2069) reported as latency scaling closer to linearly than logarithmically. Three separate linear-in-N terms ran on the query path; a fourth defect capped how many rows a vector query could return.

## What was wrong

**Layers above 0 were searched at the full `ef`.** Those layers exist only to hand the next layer down an entry point — `search()` and `index()` take `results[0]` and discard the rest — so the extra candidates buy nothing while costing work proportional to the layer's population instead of to `ef`. Layer 1 holds ~N/M nodes, so at `ef` 512 a query visited ~95% of it. On real embeddings, upper-layer visits per query grew 342 → 654 → 1,256 → 2,421 across 5K → 41K vectors (and reached 75% of query time at 100K on synthetic data); after the change they are 63 → 76 → 99 → 132. They now descend greedily (`ROUTING_EF`), which is what standard HNSW does. The pure-descent loop in `index()` had the same shape and the same fix; the connection-building pass still uses `efConstruction` because it selects the edges that get stored.

**`autoScaleEf` resolved the graph size with `getKeysCount()` on every query.** On a RocksDB store that is an exact key scan — O(N), measured at 13 ms per call at 10K keys, 128 ms at 100K, ~1 s at 500K — so every vector search paid it before any traversal happened. On the real table stack at 20K vectors it was 34% of query latency.

**`estimateCountAsSort()` did the same scan, uncached.** That one is on the planner path, so it fires whenever a vector sort is planned alongside another condition — a companion filter, RBAC, `allowRead`. A bare vector sort never reaches it, which is why a search-only benchmark does not show it.

**A `limit` above the resolved `ef` truncated silently.** Layer 0 keeps at most `ef` candidates and `ef` comes from the auto-scale, not from the query, so with the 512 cap no vector query could return more than 512 rows however large the limit — and `{offset: 250, limit: 200}` returned zero rows, meaning paginating a vector search past the first page returned nothing. `searchByIndex` now threads the query's `offset + limit` to the custom index as `minResults`, which widens the candidate list to cover the request.

Two bounds keep that from creating a new problem, both added after the pre-push review flagged the first version. The derived `ef` is capped at `LIMIT_EF_MAX` (`4 * AUTO_EF_MAX`, so 2,048): `ef` drives a synchronous traversal holding every admitted candidate in a sorted array with an O(len) insert, so uncapped, `{offset: 2000000, limit: 10}` would have walked the whole graph on the event loop — worse than the truncation being fixed. And a per-query `ef` stays authoritative, bounding the result set rather than being raised by the limit, since it is an explicit cost ceiling; a schema-level `efConstructionSearch` is a default rather than a per-request decision, so it does not block the floor.

That ceiling is the *only* bound on the widening. An earlier revision also clamped the derived `ef` to the graph size, which Code Assist then caught costing a graph-size resolution on every query whose `limit` exceeded the table — it bypassed the 10-second memo, because clamping needs a count that is exact as of the query and a memoized one cannot be. The clamp is gone rather than the refresh: an `ef` above the node count is free (`searchLayer` is bounded by the nodes it can reach, and the candidate array only holds what was admitted), and the clamp was also what let a stale count truncate the very limit it was added to honour.

## Graph size now comes from node ids

The replacement for `getKeysCount()` needs to be O(1) and correct. RocksDB's `rocksdb.estimate-num-keys` is the obvious candidate and is **wrong here**: it counts entries across memtable and SST files without reconciling overwrites, and building an HNSW graph rewrites each node many times as its neighbours change. On a real index it read 37,775 for a 2,000-record table whose exact key count is 4,001, and got worse after deletes (5.6× over at 5,000 records). It reads exact on a fresh store with simple puts, so it validates clean in isolation and only misleads on a real index — the first version of this PR used it, and the 512 cap hid it at the size the end-to-end benchmark happened to run at.

Node ids are the sound source: allocated monotonically, so the shared id counter (or one reverse seek to the largest id) gives the node count directly, unaffected by rewrites. Deletes leave it reading high until a rebuild, which only makes `ef` slightly generous.

This changes the unit from index-store keys to graph nodes. The store holds two keys per record — the node and its primary-key mapping — so `AUTO_EF_REF` moves 1000 → 500 to compensate. **The resolved `ef` is unchanged**, verified equal to `main` at 500/1,000/3,000/8,000/15,000 records.

## Effect

Real store, 768-dim nomic-embed-text embeddings, `limit: 10`. Each query runs under both configurations back to back so load drift on a shared machine hits both arms rather than being attributed to the change:

| N | before p50 | after p50 | speedup | graph-size probe removed |
| --- | --- | --- | --- | --- |
| 8,000 | 21.51 ms | 13.14 ms | 1.64× | 5.74 ms (26%) |
| 20,000 | 67.29 ms | 36.68 ms | 1.83× | 23.03 ms (34%) |

Recall holds: across 16 (size, `ef`) points on a held-out real-embedding corpus, comparing each graph against itself searched at the full `ef` on every layer, the worst recall@10 change is −0.002 — one displaced neighbour at a single point — and 0.000 everywhere else.

Index build benefits from the same descent change. A 100K-node build went from 49 min to 5.6 min, and at 20K real embeddings the greedy and full-`ef` builds were indistinguishable on every metric measured (same recall at each `ef`, same visit counts, same mean layer-0 degree) with the greedy build 1.28× faster.

## Verification

Executed, not proposed:

- `test:unit:resources` — 1,522 passing, 3 failing. All three failures (`randomAccessFields` directive ×2, `replayStructures` composite key) reproduce on unmodified `origin/main` in a separate base worktree; they are pre-existing and unrelated.
- `test:unit:main` — 4,355 passing, 0 failing.
- `test:integration` on `integrationTests/server/vector-index-integrity.test.ts` — 4/4 passing (entry-point delete, update churn, thresholds, reindex backfill).
- `unitTests/resources/vectorIndex*.test.js` — 85 passing, including 13 new tests.
- Fails-on-base: with the fix disabled, the new limit tests return 118 / 118 / **0** rows against expected 400 / 700 / 200; the new graph-size test reads 5,184 against 300 indexed vectors when pointed at `estimate-num-keys`. With only the graph-size clamp restored, the two `LIMIT_EF_MAX` assertions read 701 against 2,048 and the hot-path test counts 10 node-count re-resolutions across 10 queries against 0 explainable by the memo's TTL.

The node-count fallback (`idIncrementer` absent, as on a query-only worker) now has real-store coverage for both numeric and string primary keys, since three review passes across two models have read it as returning 0. It returns 51 on a 50-vector table either way, identical to the id-counter path. The numeric keys start at 1,000,000 on purpose: with small ones a mapping key and a node id are the same number, so the seek reads the same value whether or not mapping keys are in range and the test proves nothing.

Two benchmarks are added. `benchmarks/hnsw-scale.js` sweeps corpus size with a per-layer work breakdown and recall against brute force, over either a Gaussian mixture or a real embedding pool (`--corpus=`). `benchmarks/hnsw-real-store.cjs` runs the same query through the real table stack, where node fetch and decode dominate, with an interleaved A/B.

A methodology note that cost real time: a Gaussian mixture with per-dimension σ=0.35 at 768 dimensions is **uniform-random**, not clustered — the noise swamps the centroid — and no ANN can index it. `benchmarks/hnsw-scale.js` therefore parameterises the corpus by intra-cluster cosine rather than by σ, and prints d(10-NN)/d(median) so a corpus that has lost its structure is visible in the output.

## Not included

The `ef` auto-scale looks over-provisioned by 4–7× on real embeddings, but the constants are **not** changed here. Measured with held-out queries on 41,201 real vectors, `ef` 32 → recall 0.990 at 5K and `ef` 128 → 0.998 at 41K, while the auto-scale picks 224 and 512 respectively. The auto-scale's premise is confirmed — a fixed `ef` does decay with N — so only the constants look wrong, and a fit suggests `AUTO_EF_BASE` nearer 20–32 than 100, worth roughly another 2×.

That is a recall-affecting default and the evidence is one corpus (code + docs, d(10-NN)/d(median) = 0.28). The reporter's documentation corpus measures 0.60, a genuinely harder contrast, so it should be confirmed there before the constants move. `benchmarks/hnsw-scale.js --corpus=<pool.f32>` is the tool for that, and the reporter offered to help with large-corpus verification.

Generated by Claude Opus 5.

## For the human reviewer

The mechanics have been through seven cross-model passes (Codex + Gemini + Harper domain); three rounds found real availability defects in the `limit` plumbing, all fixed here — most recently the per-query graph-size resolution described above. What is left is judgment, in rough order of how much a "no" would change the shape of this PR.

1. ~~**The `limit` ceiling still truncates silently past 2,048 rows.**~~ **Settled** — capping (rather than raising a `ClientError`) is the chosen behaviour: erroring would turn queries that currently "work" short into hard failures, a bigger break than the one being fixed. Recorded here so it isn't re-litigated in review.

2. **A per-query `ef` outranks `limit`.** Chosen: `{ef: 20, limit: 400}` returns 20 rows — `ef` is an explicit cost ceiling, so the caller owns it. Alternative: treat `limit` as a floor under any `ef`. "I asked for 400 and got 20" is a plausible bug report; changing it later breaks anyone who tuned around it.

3. **`offset`/`limit` paging over an approximate index is best-effort, not a stable partition.** Each page re-runs the search at a different `ef`, and a larger `ef` candidate set is not guaranteed to be an ordered superset of a smaller one, so a record can repeat or be skipped across pages. Documented in DESIGN.md rather than solved. The alternative is a cursor, which is a much larger change; the risk of documenting is that offsets become depended on before cursors exist.

4. **Graph size is a node-id high-water mark.** It never decrements, so a delete-heavy table (TTL eviction, churn) reads high and keeps `ef` pinned generous until a reindex — bounded by `AUTO_EF_MAX`, so it costs latency, never correctness. Alternative: a live count maintained on insert/remove, which adds write-path cost and a persisted value that must survive crashes. Contained to `resolveNodeCount` if you want it changed.

5. **The construction-side descent ships in the same PR as the search change.** They have different rollback costs: the search change is code, the build change is baked into stored graphs and only recoverable by reindex. Measured as indistinguishable at 20K real embeddings (same recall, visits, degree; 1.28× faster build), which is why I kept them together — but splitting is defensible on rollback grounds alone.

6. **`minResults` is an eighth positional argument on `searchByIndex`.** The other call sites (`resources/Table.ts:4908`, `resources/search.ts:1089`, and the recursive ones) pass nothing, which is harmless today because none lead with a vector sort. An options object would be the durable shape for the custom-index contract; widening a positional signature again later is a breaking change for any external index.

7. **`unitTests/resources/vectorIndex.test.js`'s greedy-routing equivalence test is flaky, about 1 run in 25.** It asserts that greedy descent returns the *identical* top 10 as searching every layer at the full `ef`, on a 600-node 4-dimensional ring graph. When it fails, the two result sets are wholly disjoint, with greedy returning the worse one. Fifteen seeded builds of that same graph agreed on all four targets (60/60 comparisons), so the graph's random level assignment is not what varies — which leaves suite-level state, or a genuine occasional stall of `ROUTING_EF = 1` on a low-dimensional ring, and I could not separate the two inside this task. It does not contradict the 768-dimensional measurements above (16 points, worst recall@10 delta −0.002), but a low-dimensional corpus is a different regime and the test as written is too strong for CI either way. Flagged rather than fixed: loosening the assertion would hide whichever of the two it is.

Not changed here, and the biggest remaining win: the `ef` auto-scale looks 4–7× over-provisioned on real embeddings (see **Not included** above). I'd rather that land as its own change with a second corpus behind it than ride along with a correctness fix.

<sub>Human-Review-Need: 4 @ c1e6a15c8d20</sub>


kriszyp:
## Large-corpus verification (the deferred "Not included" item)

Resuming this from the prior session — the "Not included" section above flagged that the
ef auto-scale's apparent 4–7× over-provisioning was measured on one corpus (code + docs,
`d(10-NN)/d(median) = 0.28`) and asked for confirmation on a harder-contrast, documentation-style
corpus before the constants move, plus a ≥20K measurement on 5.2.x specifically (raised in the
issue thread). I don't have the reporter's actual embeddings, so I approximated with
`benchmarks/hnsw-scale.js`'s synthetic Gaussian-mixture generator, calibrating `--intra-cos` per
size to reproduce roughly the same `d(nn)/d(rand)` contrast the reporter's corpus measures
(~0.60), rather than the easier 0.28 corpus already in this PR. Caveat up front: this is a
calibrated synthetic stand-in, not the reporter's real data — confirmatory, not definitive.

**Setup:** N = 5,000 / 20,000 / 50,000 / 100,000, dims=768, `intra-cos` = 0.25 / 0.55 / 0.55 / 0.6
respectively (tuned so `d(nn)/d(rand)` lands at 0.65 / 0.64 / 0.68 / 0.65 — consistently harder
than the 0.28-contrast corpus already validated), 60 queries/size, current code on this branch
(greedy upper-layer descent, node-id graph size, `int8` quantization).

| N | ef (auto, capped) | p50 | p95 | recall@10 (raw / set) |
|---|---|---|---|---|
| 5,000 | 316 | 6.9ms | 7.3ms | 0.973 / 0.987 |
| 20,000 | 512 | 2.8ms | 4.1ms | 0.990 / 1.000 |
| 50,000 | 512 | 3.6ms | 5.7ms | 0.990 / 0.998 |
| 100,000 | 512 | 3.5ms | 5.7ms | 0.980 / 0.997 |

**1. The fix scales cleanly to 100K even on the harder corpus.** p50 stays under ~7ms end to
end, vs. the reporter's real 70ms at 41K on the easier corpus and the ~650ms pre-fix
extrapolation. No sign of the linear-in-N terms returning at higher contrast.

**2. Recall-preservation from greedy upper-layer descent holds here too.** A/B at N=20,000 (same
graph, `--upper-ef=match` restoring the old full-`ef`-per-layer search): recall@10 raw
0.990/0.990, set 1.000/1.000 — identical — while p50 was 1.5–1.6× worse (4.3ms vs 2.8ms,
upper-layer time 47.2% vs 5.6%). Matches this PR's "recall unchanged, latency much better" claim
on a harder corpus, not just the 0.28 one.

**3. New evidence against lowering `AUTO_EF_BASE` — the deferred question itself.** At N=20,000+
`ef` is already pinned at `AUTO_EF_MAX` (512) regardless of `AUTO_EF_BASE`, so the base only
actually controls behavior below ~13K nodes. At N=5,000 (auto resolves `ef`=316), forcing
`ef`=64 — roughly what a 5×-lower `AUTO_EF_BASE` (~20 instead of 100) would resolve to — dropped
recall@10 from 0.973 to **0.795** raw (0.987 → 0.802 set). That's an 18–19pt hit that the easier
0.28-contrast corpus didn't show (the "Not included" section measured `ef` 32 → recall 0.990 at
5K there). **This corpus does not support lowering `AUTO_EF_BASE`** — for a documentation-style,
harder-contrast workload like the reporter's, the current base is closer to necessary than
over-provisioned. Good that this PR left it untouched; I'd treat the "4–7×" framing from the
easier corpus as corpus-specific rather than a general property.

Happy to share the calibration numbers or raw JSON (`--json=` output, one file per size) if
useful for a follow-up on `AUTO_EF_BASE`.

— Claude Sonnet 5 (dispatch worker, resuming the prior session's investigation)

maurice-harper:
## QA: PASS

**Checks performed:** `unitTests/resources/vectorIndex.test.js` (node 22.22 — rocksdb-js native binding requires it; node 23/24 fails to load): 74/74 passing incl. all new tests. Full `unitTests/resources/**` (minus 1 file, see below): 1500 passing vs 1490 on `main` (net +10 = new tests); same 12 pre-existing failures (openaiStream/chat-completion streaming, ambient-user-context, reload-marker, Transactions) reproduced identically on `main` — unrelated to this diff. `schemaMigrationFragility.test.js` SIGABRTs identically on both PR head and `main` (LMDB↔RocksDB dual-engine native crash, pre-existing, untouched file). `test:unit:main` fails to start (`STORAGE_PATH` resolution) identically on both branches — pre-existing env gap, no file it covers is in this diff. `integrationTests/server/vector-index-integrity.test.ts` (unmodified by PR) blocked by sandbox loopback-binding limits (`EADDRNOTAVAIL` 127.0.0.2+, needs host `ifconfig` alias) — infra limitation, not run.

**Revert verification (manual — no spec.md for this human-authored PR):** reverted `resources/indexes/HierarchicalNavigableSmallWorld.ts` + `resources/search.ts` to `origin/main`, rebuilt, reran the suite: 7 tests fail (67/74), rows 118/118/0 vs expected 400/700/200 — exactly matching the PR body's own "fails-on-base" claim. Restored after.

**PR's own claims, verified:** (1) greedy upper-layer descent — `HNSW greedy routing above layer 0` passes, equivalent to full-ef-every-layer. (2) O(1) node-count via node ids — `HNSW graph-size resolution` tests pass (`HierarchicalNavigableSmallWorld.ts:719-738`). (3) `estimateCountAsSort` on same estimator (`:1294-1301`) — `RocksDatabase` import/instanceof branch fully removed, confirming the switch. (4) `limit`/`offset` → `minResults`, capped at `LIMIT_EF_MAX`, explicit `ef` authoritative — all 7 tests in `HNSW limit above the resolved search ef` pass.

**What I tried to break:** `limit:0`/`offset+limit:0` → 0 rows (matches the existing `Table.ts:3491` slice(offset,offset) convention, not a defect). Bare `offset` with **no limit at all** past the auto-scaled ef → 0 rows silently — same failure class as the bug being fixed, but reproduced identically on `main` (pre-existing, not a regression; PR text scopes the fix to "offset + limit", not offset alone). Schema `efConstructionSearch` + query `limit` → widening still applies (300/300 rows), matches DESIGN.md. Explicit per-query `ef` vs `limit` → `ef` wins (5 rows).

**Style:** `prettier --check` and `oxlint` clean on all 6 changed files, no new drift.

**Dismissed concerns:** node-count high-water-mark never decrements on delete. Justification: pre-existing by design, disclosed in PR body item #4, bounded by `AUTO_EF_MAX` (latency-only cost, never incorrect). `minResults` as an 8th positional arg on `searchByIndex`. Justification: disclosed in PR body item #6 with call-site evidence (`Table.ts:4908`, recursive relationship calls); backward-compatible optional param, no current caller leads with a vector sort.

maurice-harper:
## QA: PASS (supersedes prior QA comment — adds required Mechanical checks line and dismissal justifications; no conclusion changed)

**Checks performed:** `unitTests/resources/vectorIndex.test.js` (node 22.22 — rocksdb-js native binding requires it; node 23/24 fails to load): 74/74 passing incl. all new tests. Full `unitTests/resources/**` (minus 1 file, see below): 1500 passing vs 1490 on `main` (net +10 = new tests); same 12 pre-existing failures (openaiStream/chat-completion streaming, ambient-user-context, reload-marker, Transactions) reproduced identically on `main` — unrelated to this diff. `schemaMigrationFragility.test.js` SIGABRTs identically on both PR head and `main` (LMDB↔RocksDB dual-engine native crash, pre-existing, untouched file). `test:unit:main` fails to start (`STORAGE_PATH` resolution) identically on both branches — pre-existing env gap, no file it covers is in this diff. `integrationTests/server/vector-index-integrity.test.ts` (unmodified by PR) blocked by sandbox loopback-binding limits (`EADDRNOTAVAIL` 127.0.0.2+, needs host `ifconfig` alias) — infra limitation, not run.

**Mechanical checks:** `npx prettier --check` on all 6 changed files (`resources/indexes/HierarchicalNavigableSmallWorld.ts`, `resources/search.ts`, `unitTests/resources/vectorIndex.test.js`, `DESIGN.md`, `benchmarks/hnsw-scale.js`, `benchmarks/hnsw-real-store.cjs`) — "All matched files use Prettier code style!", 0 issues. `npx oxlint --format stylish` on the same 5 JS/TS files — 0 errors, 0 warnings. No new drift; nothing to compare against on `main` since these are new/changed lines only. Diff contains no regex, template-literal, or generated-output surfaces (no `new RegExp`/`.replace`/`JSON.parse`/`JSON.stringify` added), so no separate parser check was needed beyond prettier/oxlint; both benchmark scripts also pass `node --check` syntax validation.

**Revert verification (manual — no spec.md for this human-authored PR):** reverted `resources/indexes/HierarchicalNavigableSmallWorld.ts` + `resources/search.ts` to `origin/main`, rebuilt, reran the suite: 7 tests fail (67/74), rows 118/118/0 vs expected 400/700/200 — exactly matching the PR body's own "fails-on-base" claim. Restored after.

**PR's own claims, verified:** (1) greedy upper-layer descent — `HNSW greedy routing above layer 0` passes, equivalent to full-ef-every-layer. (2) O(1) node-count via node ids — `HNSW graph-size resolution` tests pass (`HierarchicalNavigableSmallWorld.ts:719-738`). (3) `estimateCountAsSort` on same estimator (`:1294-1301`) — `RocksDatabase` import/instanceof branch fully removed, confirming the switch. (4) `limit`/`offset` → `minResults`, capped at `LIMIT_EF_MAX`, explicit `ef` authoritative — all 7 tests in `HNSW limit above the resolved search ef` pass.

**What I tried to break:** `limit:0`/`offset+limit:0` → 0 rows (matches the existing `Table.ts:3491` slice(offset,offset) convention, not a defect). Bare `offset` with **no limit at all** past the auto-scaled ef → 0 rows silently — same failure class as the bug being fixed, but reproduced identically on `main` (Justification: pre-existing — the widening logic requires `limit !== undefined`, which was equally absent/inapplicable before this diff; behavior for this input is unchanged by the PR, and the PR text scopes its fix to "offset + limit," not offset alone, so it is not overclaiming). Schema `efConstructionSearch` + query `limit` → widening still applies (300/300 rows), matches DESIGN.md's documented intent (in-scope, not a dismissal). Explicit per-query `ef` vs `limit` → `ef` wins (5 rows), matches documented/tested design (in-scope, not a dismissal).

**Dismissed concerns:**
- Node-count high-water-mark never decrements on delete. Justification: pre-existing by design — disclosed in the PR body's "For the human reviewer" item #4, bounded by `AUTO_EF_MAX` so it only ever costs latency, never correctness; contained to `resolveNodeCount` if it needs changing later.
- `minResults` is an 8th positional argument on `searchByIndex`. Justification: out-of-scope for this PR — disclosed in the PR body's item #6 with call-site evidence (`Table.ts:4908`, the recursive relationship-traversal calls in `search.ts`); it's a backward-compatible optional trailing param and no current caller leads with a vector sort, so today's behavior is unaffected.
- Bare `offset` with no `limit` still truncates past the auto-scaled ef. Justification: pre-existing — see "What I tried to break" above; confirmed identical on `origin/main`, not introduced or worsened by this diff.

maurice-harper:
## Review: CHANGES NEEDED

Gemini (gemini-flash-latest) is the reviewer of record for this pull request. Head SHA: `177cc4bd8bb6a17cee4d16b99af1e8e4f82cdb88`.

Thank you for working on optimizing HNSW vector search latency and addressing corpus scaling.

The overall direction—greedy upper-layer descent, limit widening via `minResults`, and replacing O(N) `getKeysCount()` key scans—significantly improves vector query performance. However, there is a defect in the node count fallback resolution logic that needs to be addressed before merge.

### Required Fixes
- `resolveNodeCount` in `HierarchicalNavigableSmallWorld.ts` passes `limit: 1` to `indexStore.getKeys({ reverse: true, limit: 1 })`. Because `indexStore` holds string keys (`'r' + primaryKey`) in addition to numeric node keys, the reverse scan yields a string key first. The `limit: 1` constraint stops iteration immediately, preventing the loop from ever reaching a numeric key. As a result, `resolveNodeCount()` falls back to `0`, causing `autoScaleEf()` to select `AUTO_EF_BASE` (100) regardless of graph size when `idIncrementer` is absent.

### Follow-up candidates
- Consider persisting `idIncrementer` state or re-populating it reliably on store initialization so that node count resolution remains exact across process restarts without relying solely on reverse key scans.

### Required fixes
- resources/indexes/HierarchicalNavigableSmallWorld.ts:729: In `resolveNodeCount()`, `this.indexStore.getKeys({ reverse: true, limit: 1, start: Infinity, end: 0 })` specifies `limit: 1`. Because `indexStore` stores string keys (`'r' + primaryKey`) alongside numeric graph node keys, a reverse scan yields a string key first. The `limit: 1` option halts store iteration after yielding that single key. Since `typeof key === 'number'` evaluates to false for the string key, the loop terminates immediately and `resolveNodeCount()` returns `0`. Consequently, whenever `idIncrementer` is falsy or unavailable, `approximateNodeCount()` evaluates to 0 and `autoScaleEf()` falls back to `AUTO_EF_BASE` (100) regardless of the actual graph size.
  Evidence: On base branch (eb702ee52ef8f97085dc9d20f81ba52234cf077b), `search()` called `indexStore.getKeysCount()` or `entryCount`. This PR introduced `resolveNodeCount()` containing `this.indexStore.getKeys({ reverse: true, limit: 1, start: Infinity, end: 0 })` at lines 728-732 of `HierarchicalNavigableSmallWorld.ts`.
  Fix: Do not pass `limit: 1` to `getKeys()` when scanning for the largest numeric key, or bound the scan range/type filter so that iteration continues past string keys until the first numeric node key is encountered.

_Gemini execution source: api._
_Note: REST fallback could not use repository tools; it reviewed the staged prompt and fallback diff only._

maurice-harper:
## Review-pass consolidation (QA + Code Assist + Gemini)

Three review passes ran against head `177cc4bd8`. Verdict from the review team: **the code is sound as-is; nothing blocking**. Triage of every finding, with the evidence:

### Gemini's required fix is declined — empirically refuted

The CHANGES NEEDED comment above claims `resolveNodeCount()`'s `limit: 1` reverse seek stops on a non-numeric mapping key and returns 0. Reproduced the exact scenario on the PR head (built worktree, `idIncrementer` cleared to simulate a query-only worker):

- **String PKs** (`user-000`…`user-049`, mapping keys are the raw strings): first 5 keys of the reverse seek from `start: Infinity` are `num:50, num:49, num:48, num:47, num:46` — fallback count **51**, identical to the incrementer path. String keys sort above numeric `Infinity` and are excluded by the bound.
- **Numeric PKs** (mapping keys are `[Symbol.for('key'), n]` arrays): same result, fallback count **51**.

No non-numeric key can shadow the seek. (The review's `'r' + primaryKey` key scheme doesn't exist in the code — the reviewer ran via REST fallback without repository tools.)

**Non-blocking suggestion that falls out of this:** the fallback path (`idIncrementer` absent) has no test in the PR — the new graph-size tests index in-process, so the incrementer is always initialized. Two separate reviewer models independently flagged this exact path *because* nothing exercises it. The repro above is ~20 lines and would make a good addition to the `HNSW graph-size resolution` describe block.

### Code Assist high (TTL bypass at `HierarchicalNavigableSmallWorld.ts:1051`) — real but overstated

Its "synchronous database range scan on every query" is wrong — the fallback is a `limit: 1` reverse seek. But the narrow point stands: when a query's `minResults` exceeds the actual graph size, `count < minResults` stays true after the refresh, so **every** such query bypasses the 10s memo and does one synchronous seek (query-only workers) or one atomic load (writers). One-line option if you want it: also update a `nodeCountBelowMinResultsAt` timestamp on the force-refresh path so a too-small graph re-resolves at TTL cadence rather than per query. Cheap either way; your call whether it's worth the extra state.

### Code Assist medium (`benchmarks/hnsw-scale.js:102`) — confirmed real, latent

The mock's `getKeys` filters `k >= start && k <= end`; `resolveNodeCount()` calls it with `start: Infinity, end: 0`, which never matches, so the mock's reverse seek yields nothing. Latent today (the benchmark initializes the id counter via `getUserSharedBuffer` during build), but it silently zeroes the exact fallback the suite would need to validate. The suggested `Math.min`/`Math.max` normalization is right.

### QA (details in the QA: PASS comment above)

74/74 vector tests pass; revert verification confirms the 7 new tests fail on base with exactly the claimed 118/118/**0** row counts; prettier/oxlint clean. One adjacent gap found while trying to break the limit plumbing: a **bare `offset` with no `limit`** landing past the auto-scaled `ef` still silently returns 0 rows — same failure class as the bug fixed here, but reproduced identically on `main` (the widening requires `limit !== undefined`) and the PR text scopes itself to `offset + limit`. Pre-existing, not blocking; flagged as a follow-up candidate.

### Gemini's follow-up candidate (persist `idIncrementer`) — declined

The seek fallback is O(1) and demonstrated correct above; persisting the counter adds write-path state that must survive crashes for no measured benefit.

— eng-manager review run for #2069

maurice-harper:
## QA: PASS (re-run against new head 0638de95b — supersedes prior QA comments; no conclusion changed from the prior PASS)

**Checks performed:** Fetched/checked out `0638de95bb906d463f66a9525053fdbfd2f5e444` in-place in the existing worktree, rebuilt (`tsc`), ran under node 22.22 (rocksdb-js native binding requirement, unchanged from before). `unitTests/resources/vectorIndex.test.js`: 76/76 passing (was 74 at 177cc4bd8; +2 net matches the reworked graph-size/limit tests in this delta). Full `unitTests/resources/**` (minus the one pre-existing-crashing file, unrelated to this diff): 1502 passing, same 12 pre-existing failures as the prior QA pass (openaiStream/chat-completion streaming, ambient-user-context, reload-marker, snapshot writes, transaction-log deletion, Transactions) — unrelated to this diff, previously confirmed identical on `main`.

**Mechanical checks:** `npx prettier --check` on all 5 changed files (`resources/indexes/HierarchicalNavigableSmallWorld.ts`, `resources/search.ts` unchanged in this delta so not re-diffed, `benchmarks/hnsw-scale.js`, `benchmarks/hnsw-search.js`, `DESIGN.md`, `unitTests/resources/vectorIndex.test.js`) — "All matched files use Prettier code style!", 0 issues. `npx oxlint --format stylish` on the 4 JS/TS files in that set — 0 errors, 0 warnings. No new drift.

**Delta-specific verification (per coordinator's focus list):**
1. Full `vectorIndex.test.js` suite — 76/76 passing (above).
2. Removing the graph-size clamp cannot over-allocate — built a scratch test (40-node graph, `limit: 1_000_000`): `effectiveEf` widens to `LIMIT_EF_MAX` (2048) as expected, but layer-0 `visited` stays at exactly 40 and all 40 reachable rows return correctly — confirms "an ef above the node count is free, the traversal ends at the graph" structurally (searchLayer's unfiltered stop rule is `candidates.size === 0`, not "results.length reached ef"), not just by claim. Scratch test deleted after use (not part of the diff).
3. `grep -rn "approximateNodeCount" resources/ unitTests/ benchmarks/` — all 3 call sites (`HierarchicalNavigableSmallWorld.ts:1034`, `:1300`, `vectorIndex.test.js:743`) call it with zero arguments, matching the new no-parameter signature; no stale caller passes the removed `forceRefresh` arg.
4. Prettier/oxlint — clean, see Mechanical checks above.

**Revert verification (manual — no spec.md for this human-authored PR):** reverted `resources/indexes/HierarchicalNavigableSmallWorld.ts` + `resources/search.ts` to `origin/main` at the new head, rebuilt, reran: 9 failures (67/76) — 3 `resolveNodeCount is not a function` (the renamed test helper) plus the same limit-truncation failures as before (118/118/0 vs 400/700/200, and layer0Ef 118 instead of the now-expected 2048). Restored to PR head after.

**What changed since the last QA pass:** `approximateNodeCount()` lost `forceRefresh`; the limit-widening path no longer calls it at all and is bounded solely by `LIMIT_EF_MAX`, not by graph size — verified safe under item 2 above. Two tests reworked/added to match (`resolves the same size from a reverse seek when the shared id counter is absent`, `does not re-resolve the graph size for a limit past the end of the table` — the latter asserts ≤1 re-resolution across 10 over-limit queries, observing the memo instead of stubbing the resolver). `benchmarks/hnsw-scale.js` and `benchmarks/hnsw-search.js` `MemoryStore.getKeys` now normalize `start`/`end` with `Math.min`/`Math.max` since a reverse seek passes them swapped.

**Dismissed concerns:**
- Node-count high-water-mark never decrements on delete. Justification: pre-existing — disclosed in the PR body's item #4, bounded by `AUTO_EF_MAX`, latency-only, never incorrect; unaffected by this delta.
- `minResults` is an 8th positional argument on `searchByIndex`. Justification: out-of-scope — disclosed in the PR body's item #6 with call-site evidence; unaffected by this delta.
- Bare `offset` with no `limit` at all still truncates past the auto-scaled ef once it exceeds the graph. Justification: pre-existing — the widening gate (`minResults !== undefined`) is unchanged by this delta; confirmed identical on `origin/main` in the prior QA pass, not introduced or worsened here.

maurice-harper:
## Review: CHANGES NEEDED

Gemini (gemini-flash-latest) is the reviewer of record for this pull request. Head SHA: `0638de95bb906d463f66a9525053fdbfd2f5e444`.

The PR addresses HNSW vector search scaling and limit handling, but introduces a defect in graph size resolution (`resolveNodeCount`) that breaks auto-scaled `ef` and planner cost estimation for tables with string primary keys.

### Key Finding
When `idIncrementer` is not present (e.g. on query-only workers or non-counter initializations), `resolveNodeCount()` attempts to find the maximum node ID by calling `this.indexStore.getKeys({ reverse: true, limit: 1, start: Infinity, end: 0 })`. Because Harper's `indexStore` holds both numeric node IDs and string primary key mapping entries, a reverse store scan on a table with string primary keys yields a string key first. Passing `limit: 1` causes the database iterator to halt after returning that single string key. The `typeof key === 'number'` check fails, and `resolveNodeCount()` returns `0`.

When `resolveNodeCount()` returns `0`, `autoScaleEf()` falls back to its base value (100) regardless of whether the table contains 20,000 or 100,000 vectors, severely degrading search recall on large corpora. Additionally, `estimateCountAsSort()` evaluates to `0`, skewing query planner condition ordering.

Please update `resolveNodeCount()` to avoid `limit: 1` and invalid `Infinity` range options so non-numeric primary key entries are skipped during reverse key traversal.

### Required fixes
- resources/indexes/HierarchicalNavigableSmallWorld.ts:732: resolveNodeCount() passes `limit: 1` and `start: Infinity, end: 0` to `indexStore.getKeys()`. In Harper tables, `indexStore` contains both numeric node IDs and primary key lookup entries. When primary keys are strings, the reverse store scan yields a string primary key entry as its first item. Because `limit: 1` was passed to `getKeys()`, the store iterator terminates after yielding that single string key. The `typeof key === 'number'` check evaluates to false, causing `resolveNodeCount()` to exit the loop and return 0. This breaks `autoScaleEf()` for string-primary-key tables, forcing ef to default to 100 regardless of graph size and corrupting `estimateCountAsSort()` cost estimates.
  Evidence: Base version eb702ee52ef8f97085dc9d20f81ba52234cf077b called `this.indexStore.getKeysCount()`. This PR introduced `resolveNodeCount()` at line 728 in `HierarchicalNavigableSmallWorld.ts`. In `resolveNodeCount()`, line 732 calls `this.indexStore.getKeys({ reverse: true, limit: 1, start: Infinity, end: 0 })`. In `MemoryStore` (unit tests), `getKeys()` was specifically updated in this PR to filter numeric keys before applying `limit: 1`, masking the bug; however, production database stores (`RocksDatabase`/`LMDBDatabase`) yield store keys in raw key order. When string primary keys exist, `getKeys()` yields a string key, `limit: 1` halts iteration, and `resolveNodeCount()` returns 0.
  Fix: Remove `limit: 1` from `indexStore.getKeys()` in `resolveNodeCount()` so the iteration can continue past non-numeric string primary key entries until the highest numeric node ID key is found. Also replace invalid numeric range bounds (`start: Infinity, end: 0`) with store-compatible options.

_Gemini execution source: api._
_Note: REST fallback could not use repository tools; it reviewed the staged prompt and fallback diff only._

maurice-harper:
## Review pass on `0638de95b` — reviewer verdict declined again, QA clean

The fresh CHANGES NEEDED above repeats the finding already [refuted on the previous head](https://github.com/HarperFast/harper/pull/2125#issuecomment-5235051849), with one new claim: that the benchmark `MemoryStore` fix "masks" a production-store bug. For the record: the refutation was executed against the **production store path** (real table harness → RocksDB), not the mock — string-PK and numeric-PK tables both resolve the correct fallback count with the id counter absent. This head also now carries Kris's in-suite regression test for exactly that path (`resolves the same size from a reverse seek when the shared id counter is absent`), passing in the 76/76 run.

The three post-feedback commits verified cleanly (fresh QA: PASS above): the widening no longer consults graph size at all, the `LIMIT_EF_MAX`-only bound was adversarially checked (a 40-node graph queried with `limit: 1,000,000` visits exactly 40 nodes — traversal ends at the graph, structurally), and all `approximateNodeCount` call sites match the narrowed signature.

Nothing blocking from the review team. Remaining items are the human-judgment ones in the PR body.

— eng-manager review run for #2069

maurice-harper (commented):
This is good to merge from my side, pending one question at the end — it ran through several review passes on our side (QA against three successive heads, Code Assist, and a reviewer pass), and the outcome was clean: 77/77 vector tests with repeat-run flakiness checks, revert verification matching the fails-on-base claims exactly, and the one reviewer finding that alleged a defect in `resolveNodeCount`'s reverse-seek fallback was empirically refuted on the production store path — now also covered by your in-suite fallback tests for both key types. Nice work on the post-feedback commits; removing the graph-size clamp from the limit widening rather than patching around the memo was the better fix, and the 1M-base keys in the seek test are a genuinely sharper design than the version we suggested.

On the judgment items in the PR body: 2 through 5 I'm accepting as you decided them — per-query `ef` as an explicit cost ceiling, documented best-effort paging, the high-water-mark count, and keeping the construction-side descent in this PR given the 20K-real-embedding equivalence data.

One question on item 6 before this merges: **would you rather land `minResults` as an options object now, while the custom-index contract is still unreleased?** Your own writeup makes the case — widening the positional signature again later is a breaking change for any external index, and this PR is the last point where reshaping it costs nothing. If you'd rather keep the positional arg (it's optional, trailing, and backward-compatible, so that's defensible), say so and I'll merge as-is — not blocking on it.

The bare-`offset`-no-`limit` truncation gap QA found is pre-existing and stays out of this PR; we'll track it separately.

dawsontoth (approved):

kriszyp:
On item 6: keeping the positional `minResults` here and reshaping the contract separately — tracked as [#2165 Use an options object for searchByIndex and the custom-index search() contract](https://github.com/HarperFast/harper/issues/2165), which covers both `searchByIndex` and `HierarchicalNavigableSmallWorld.search()` and notes a future paging cursor as the other candidate for that object.

Merged `main` in to clear the conflicts (DESIGN.md and `vectorIndex.test.js` both had additive sections on each side; `HierarchicalNavigableSmallWorld.ts` auto-merged with #2133's post-sort distance rescoring intact).

kriszyp:
## 1M-vector verification — the scaling holds; construction quality becomes the next bottleneck

Extending the sweep above from 100K to 1M vectors (same `benchmarks/hnsw-scale.js` harness, 768-dim, int8, `intra-cos` 0.6 → `d(nn)/d(rand)` ≈ 0.65 — the harder-contrast calibration, held across sizes since `clusters = N/500` keeps per-cluster density constant). Latest `main`, one process per size, solo runs for every latency number quoted.

| N | build | p50 | p95 | visited | recall@10 raw / set | peak RSS |
|---|---|---|---|---|---|---|
| 100,000 | 2.9 min | 3.88ms | 5.98ms | 1,745 | 0.980 / 0.997 | 1.2 GB |
| 250,000 | 9.9 min | 5.29ms | 8.98ms | 2,135 | 0.978 / 0.988 | 2.7 GB |
| 500,000 | 24.8 min | 6.38ms | 11.03ms | 2,634 | 0.978 / 0.987 | 5.1 GB |
| 1,000,000 | 62.8 min | 8.93ms | 16.17ms | 3,948 | 0.935 / 0.943 | 9.6 GB |

**1. The fix's central claim survives an order of magnitude past where it was verified.** Latency scales ∝ n^0.36 across 100K → 1M — 10× the corpus costs 2.3× the latency. Upper-layer visits stay logarithmic (159 → 224, 4.5% of query time at 1M), so the greedy descent does not degrade at scale; pre-fix, layer 1 alone at 1M would have held ~59K nodes searched at the full `ef`. Build is 63 min for 1M inserts (3.8 ms/insert, up from 1.75 at 100K — the log-factor growth you'd expect, not the linear blowup).

**2. The recall dip at 1M is real, and it is not the search side.** Sweeping `ef` over the same 1M graph: 512 → 0.943 set, 768 → 0.962, 1024 → 0.963, 1536 → 0.967 at 4.7× the latency. The curve is flat — the missing neighbours are not deep in the candidate list, they are unreachable. So raising `AUTO_EF_MAX` for large tables (my first hypothesis) buys ~2 points and is the wrong lever.

**3. The lever is `efConstruction`.** Two rebuilds of the identical corpus (fixed seed, identical level assignments), both queried at `ef` 512:

| build config | recall@10 raw / set | p50 | visited | build |
|---|---|---|---|---|
| efC=100, int8 (default) | 0.935 / 0.943 | 8.9ms | 3,948 | 63 min |
| efC=100, float32 | 0.952 / 0.952 | — | 3,596 | 71 min |
| efC=200, int8 | **0.985 / 0.997** | 7.2ms | 3,110 | 110 min |

`efConstruction` 200 restores 100K-level recall exactly (0.997 set), and the queries get *faster* — 21% fewer nodes visited at the same `ef`, because better-selected edges route more directly. Quantization contributes ~1.5 points; construction is the dominant term. The mechanism mirrors what this PR fixed on the search side: the edge-selection candidate list is a constant (100) while the corpus grows, so graph quality erodes exactly where the search-side auto-scale runs out of runway. Filed as #2180 (proposed fix: auto-scale `efConstruction` with graph size when not explicitly configured, mirroring the search-side auto-scale).

One correction to the harness worth noting for anyone re-running this: the printed `d(nn)/d(rand)` drifts upward with N (0.652 → 0.741) but the corpus is not actually getting harder — the diagnostic probes only the first 3,000 rows, so at 1M it misses the true nearest neighbour. The recall numbers are unaffected (ground truth is exact brute force over the full pool).

Same caveat as the runs above: synthetic calibrated corpus, not real embeddings — confirmatory for the scaling shape, and the #2180 constants deserve a real-corpus check before they move.

Raw logs/JSON: `~/dev/tmp/hnsw-1mm/` on the benchmark box; happy to attach.

Generated by Claude Fable 5.

PR #2126, opened 2026-08-09
test(database): end-to-end regression coverage for copy-db's blob store (#2048)

### LINKED ISSUE #2048: copy-db produces a silently corrupt, non-restorable copy and exits 0 — four independent channels in bin/copyDb.ts

kriszyp:
harper#2098 merged on 2026-08-29 and closed #2048, which changes what this PR should be. The two `{ todo }` contract probes here now pass; left as they were, they would convert a future regression into a `t.diagnostic()` while CI stayed green — the opposite of the guard the file exists to be. So: rebased onto `main`, both markers and the `assertContractOrDiagnose()` helper removed, assertions propagate normally.

Making them enforcing turned up two problems with what they enforced.

**The fixture never stored a record.** `Seed` is a `loadAsInstance = false` resource, so Harper dispatches its instance methods as `(query, data)` (`resources/Resource.ts:261`). Declaring `post(body)` bound `body` to the *query*, so every seeded field read as `undefined`: all ten seeds wrote one blob record under the key `"undefined"` — which is why the original description measured ten blob files for six blob docs — and the four "inline controls" were blobs too. `GET /Doc/blob-0` returned 404 against the live source. Nothing caught it, because nothing ever read a record back. One-character-class fix at [`resources.js:28`](https://github.com/HarperFast/harper/pull/2126/changes?w=1#diff-a41630c45980c6fd77921545f2c41fc5aa69ab5433d2daa10bb0888edb6e8557R28).

**Counting files was never the contract.** A `blobs`-shaped path with six files in it cannot distinguish a restorable copy from one whose records are undecodable or point at the wrong bytes — @heskew's second thread. So the suite now performs the restore the copy's own `README.md` documents and reads every record back:

1. environment file → `database/{db}.mdb`, and [`<target>-blobs/0/` → the matching blob root](https://github.com/HarperFast/harper/pull/2126/changes?w=1#diff-48b0c0390458474c85c028e3edd41f8f6cea9c5c30b7730d6b5620eeb6b308e7R284);
2. restart Harper onto it (same `dataRootDir`, so one extra boot, not a second instance);
3. compare each blob against [the sha256 its seed call returned](https://github.com/HarperFast/harper/pull/2126/changes?w=1#diff-48b0c0390458474c85c028e3edd41f8f6cea9c5c30b7730d6b5620eeb6b308e7R295) and each inline control against the value the test chose.

The source environment and blob root are [removed, and asserted gone](https://github.com/HarperFast/harper/pull/2126/changes?w=1#diff-48b0c0390458474c85c028e3edd41f8f6cea9c5c30b7730d6b5620eeb6b308e7R279), before the restore — otherwise the read-back could be answered by surviving source files rather than by the copy. That guard is the load-bearing part: this is where to look hardest.

The `Verify` route reports [the blob's own sha256](https://github.com/HarperFast/harper/pull/2126/changes?w=1#diff-a41630c45980c6fd77921545f2c41fc5aa69ab5433d2daa10bb0888edb6e8557R54) rather than whether it matches the record's own `note`, so the comparison is against a value captured at seed time, not against another field of the same possibly-corrupt record.

Two test names also claimed coverage their bodies do not provide. Nothing here induces a per-record copy failure, so the copy test now says [what it asserts](https://github.com/HarperFast/harper/pull/2126/changes?w=1#diff-48b0c0390458474c85c028e3edd41f8f6cea9c5c30b7730d6b5620eeb6b308e7R191) instead of naming #2048 ch.4, and the scope test says [`copy-db` leaves the source blob roots in place](https://github.com/HarperFast/harper/pull/2126/changes?w=1#diff-48b0c0390458474c85c028e3edd41f8f6cea9c5c30b7730d6b5620eeb6b308e7R251) rather than claiming `compactOnStart` — which it never invokes — is covered.

**5 pass, 0 fail, 0 todo.**

## For the human reviewer

- **@heskew's review offered two branches — "become active regression tests" or "close as superseded by #2098".** I took the first: #2098's coverage is unit-level against `copyDb()`, and this is the only thing that drives the documented `copy-db` CLI verb and the documented restore end to end. If you'd rather not carry a ~13s five-boot suite for that, closing it is a one-line decision and I won't argue.
- **Declined: a stale-`dist` guard.** Codex asked for an mtime check because "the server under test runs from source via typestrip" while the CLI comes from `dist`. It doesn't — `@harperfast/integration-testing`'s `getHarperScript()` resolves the server to the same `dist/bin/harper.js` ("ancestor dist" in the run log). A stale `dist` is a repo-wide precondition for the whole integration tree (`integrationTests/README.md`), so one file unilaterally enforcing freshness would fail while its 58 peers happily use the same build.
- **Channel 4's failure path is still not exercised anywhere.** The `failedRecords > 0` throw `#2098` added would need fault injection to reach, and `AGENTS.md` rules out stub-based tests. The test name no longer implies otherwise.
- **Restoring more than one blob root is untested.** `unitTests/dataLayer/blobBackup.test.js:232` copies a second root; nothing restores one. Worth an issue rather than scope here.
- **The restore is destructive and in-place** (delete the live environment + blob root, restore over them). That is what makes the read-back non-vacuous, and it is why it is the last test in the suite. The alternative — a second `rootPath` and a second Harper install — costs another full boot for the same signal.
- **Not covered: the opt-in deflate blob path** (c00416010). These blobs are HMAC-derived and incompressible.
- The Gemini leg could not run on this worker (`agy` not authenticated). Rounds 1–2 (the substantive ones) had codex + cursor-composer + Harper-domain adjudication; rounds 3–4 were comment-only deltas the policy narrowed to codex, which is what the footer records.

## Verification

Route: new integration test, executed. `npm run build` then `npx harper-integration-test-run integrationTests/database/copy-db-blob-store.test.ts` → **5 pass, 0 fail, 0 todo** (~4s once warm). `npm run lint:required` and `prettier --check` clean.

Fails-on-base is not available for the LMDB half — #2098 is already on `main` — so the assertion was mutation-checked instead: skipping only the blob-companion half of the restore fails the suite with `Blob file not found for .../blobs/data/0/0/1`, which is exactly the pre-#2098 outcome. Before the fixture fix, the same suite failed with `{"present":false}` for every key, against both the source and the restored copy.

Not run locally: the full `test:integration:all` gate (59 files in `database/` alone). Each file runs in its own process against its own `dataRootDir` and loopback address, and this change is confined to one new file plus its own fixture directory, so CI's sharded run is the gate.

Refs #2048. Supersedes the todo-marked form of this PR.

Generated by Claude Opus 5.

Complexity: medium

<sub>Review-Coverage: authored=claude; ran=codex; blocked=gemini(auth); declined=cursor-grok,cursor-composer,domain; rounds=4 @ 810a6beaffbe</sub>

<sub>Human-Review-Need: 3 @ 810a6beaffbe</sub>

### DIFF
--- a/integrationTests/database/copy-db-blob-store.test.ts
+++ b/integrationTests/database/copy-db-blob-store.test.ts
@@ -16,13 +16,23 @@
  * off the CI result, and surfaces the moment they start passing — at which point the marker comes
  * off and this becomes an ordinary regression anchor for the fix.
  *
+ * `{ todo }` alone isn't enough under THIS repo's runner, though: `harper-integration-test-run`
+ * (`@harperfast/integration-testing`, confirmed on published 0.7.0 and 0.7.1) sets
+ * `process.exitCode = 1` on node:test's raw `test:fail` event without checking `data.todo`, and
+ * node:test still emits that event for a failing todo test — only the summary counters and the
+ * stock CLI exit code special-case it. So a throwing `{ todo }` assertion fails this build exactly
+ * as hard as a real one. `assertContractOrDiagnose()` below is the workaround: it keeps `{ todo }`
+ * for readable output but catches the assertion itself so the test body never throws. Delete it
+ * once the runner checks `data.todo` (harper#2048 ch.3/4 fixes don't need it — only the runner bug
+ * does).
+ *
  * Scope note: #2048 also reaches `storage.compactOnStart`, but not with this consequence.
  * compactOnStart (bin/copyDb.ts:38) swaps the copy into the SAME rootPath and moves only
  * `database/{db}.mdb`, leaving `blobs/{db}/` adjacent and resolvable. Only an external target
  * loses the blobs. Asserted below so nobody "fixes" compactOnStart chasing this.
  */
 import { suite, test, before, after } from 'node:test';
-import { ok, strictEqual } from 'node:assert';
+import { ok, strictEqual, AssertionError } from 'node:assert';
 import { resolve, join } from 'node:path';
 import { readdirSync, statSync, existsSync, mkdtempSync } from 'node:fs';
 import { tmpdir } from 'node:os';
@@ -62,6 +72,26 @@ function filesUnder(dir: string): string[] {
 	};
 	walk(dir, '');
 	return out.sort();
+}
+
+/**
+ * `{ todo }` is the correct node:test option for "assert what should be true, don't fail CI
+ * yet" — but @harperfast/integration-testing's `harper-integration-test-run` listens to the
+ * node:test runner's raw `test:fail` event and sets `process.exitCode = 1` on it unconditionally,
+ * without checking `data.todo` (confirmed against the published 0.7.0 and 0.7.1 builds). node:test
+ * itself still emits `test:fail` for a failing todo test — only the summary counters and CLI exit
+ * code special-case it — so under that runner a `{ todo }` assertion that throws fails the build
+ * exactly as hard as a real assertion, defeating the whole point of marking it todo. Until that's
+ * fixed upstream, keep `{ todo }` for readable local/TAP output, but never let the assertion itself
+ * throw out of the test body — catch it and report it as a diagnostic instead.
+ */
+function assertContractOrDiagnose(t: { diagnostic: (message: string) => void }, assertFn: () => void) {
+	try {
+		assertFn();
+	} catch (err) {
+		if (!(err instanceof AssertionError)) throw err;
+		t.diagnostic(`[KNOWN BUG, non-blocking] ${err.message}`);
+	}
 }
 
 suite(
@@ -163,34 +193,38 @@ suite(
 		test(
 			'the copy contains the blob store',
 			{ todo: 'fails until harper#2048 channel 3 is fixed; remove this marker when it starts passing' },
-			() => {
-				const inCopy = filesUnder(copyRoot);
-				const blobFiles = inCopy.filter((f) => /blobs?\//i.test(f));
-				ok(inCopy.length > 0, 'the copy should contain the database files');
-				ok(
-					blobFiles.length >= BLOB_DOCS,
-					`copy-db produced a non-restorable copy: ${sourceBlobCount} blob file(s) in the source, ` +
-						`${blobFiles.length} in the copy. Records in the copy carry fileId references that resolve ` +
-						`to nothing, while inline attributes survive byte-exact — so the loss is partial and silent. ` +
-						`Copy tree: ${JSON.stringify(inCopy.slice(0, 20))}`
-				);
+			(t) => {
+				assertContractOrDiagnose(t, () => {
+					const inCopy = filesUnder(copyRoot);
+					const blobFiles = inCopy.filter((f) => /blobs?\//i.test(f));
+					ok(inCopy.length > 0, 'the copy should contain the database files');
+					ok(
+						blobFiles.length >= BLOB_DOCS,
+						`copy-db produced a non-restorable copy: ${sourceBlobCount} blob file(s) in the source, ` +
+							`${blobFiles.length} in the copy. Records in the copy carry fileId references that resolve ` +
+							`to nothing, while inline attributes survive byte-exact — so the loss is partial and silent. ` +
+							`Copy tree: ${JSON.stringify(inCopy.slice(0, 20))}`
+					);
+				});
 			}
 		);
 
 		test(
 			'copy-db does not report success when records fail to copy',
 			{ todo: 'fails until harper#2048 channel 4 is fixed; remove this marker when it starts passing' },
-			() => {
-				// The audit-store copy opens its target on the SOURCE rootStore (bin/copyDb.ts:223,
-				// where every other dbi uses targetEnv at :217) and is not awaited (:225 vs :220), so
-				// per-record failures are logged and the exit code stays 0.
-				const failures = (copyOut.match(/Error copying record/g) ?? []).length;
-				strictEqual(
-					failures > 0 && copyExit === 0,
-					false,
-					`copy-db logged ${failures} per-record copy failure(s) and still exited ${copyExit}. ` +
-						`A backup verb that reports success while dropping records is what turns this into data loss.`
-				);
+			(t) => {
+				assertContractOrDiagnose(t, () => {
+					// The audit-store copy opens its target on the SOURCE rootStore (bin/copyDb.ts:223,
+					// where every other dbi uses targetEnv at :217) and is not awaited (:225 vs :220), so
+					// per-record failures are logged and the exit code stays 0.
+					const failures = (copyOut.match(/Error copying record/g) ?? []).length;
+					strictEqual(
+						failures > 0 && copyExit === 0,
+						false,
+						`copy-db logged ${failures} per-record copy failure(s) and still exited ${copyExit}. ` +
+							`A backup verb that reports success while dropping records is what turns this into data loss.`
+					);
+				});
 			}
 		);
 
--- a/integrationTests/database/copy-db-blob-store.test.ts
+++ b/integrationTests/database/copy-db-blob-store.test.ts
@@ -14,17 +14,8 @@
  * txnlog-purge-stale-read-blast arm 5 sat red on main for a day (159b4bbd9). So the contract
  * tests below assert what SHOULD be true and carry `todo`, which runs them, keeps their failure
  * off the CI result, and surfaces the moment they start passing — at which point the marker comes
- * off and this becomes an ordinary regression anchor for the fix.
- *
- * `{ todo }` alone isn't enough under THIS repo's runner, though: `harper-integration-test-run`
- * (`@harperfast/integration-testing`, confirmed on published 0.7.0 and 0.7.1) sets
- * `process.exitCode = 1` on node:test's raw `test:fail` event without checking `data.todo`, and
- * node:test still emits that event for a failing todo test — only the summary counters and the
- * stock CLI exit code special-case it. So a throwing `{ todo }` assertion fails this build exactly
- * as hard as a real one. `assertContractOrDiagnose()` below is the workaround: it keeps `{ todo }`
- * for readable output but catches the assertion itself so the test body never throws. Delete it
- * once the runner checks `data.todo` (harper#2048 ch.3/4 fixes don't need it — only the runner bug
- * does).
+ * off and this becomes an ordinary regression anchor for the fix. See `assertContractOrDiagnose()`
+ * below for why `{ todo }` alone doesn't actually achieve that in this repo.
  *
  * Scope note: #2048 also reaches `storage.compactOnStart`, but not with this consequence.
  * compactOnStart (bin/copyDb.ts:38) swaps the copy into the SAME rootPath and moves only
@@ -35,6 +26,7 @@ import { suite, test, before, after } from 'node:test';
 import { ok, strictEqual, AssertionError } from 'node:assert';
 import { resolve, join } from 'node:path';
 import { readdirSync, statSync, existsSync, mkdtempSync } from 'node:fs';
+import { rm } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { execFile } from 'node:child_process';
 import { promisify } from 'node:util';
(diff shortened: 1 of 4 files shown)


integrationTests/database/copy-db-blob-store.test.ts
```diff
@@ -0,0 +1,233 @@
...
+				copyOut.trim().length > 0,
+				`copy-db produced no output at all — it almost certainly never located the database (harness problem, not a defect). exit=${copyExit}`
+			);
+			ok(existsSync(target), `copy-db should have produced ${target}. Output:\n${copyOut.slice(-1500)}`);
+		});
+
+		test(
+			'the copy contains the blob store',
+			{ todo: 'fails until harper#2048 channel 3 is fixed; remove this marker when it starts passing' },
+			(t) => {
```
heskew: **Significant — this now lands as a permanently non-enforcing regression test.** Harper#2098 merged on August 29 and closed #2048 after this branch’s last run, fixing both contracts covered here. Keeping `{ todo }` while routing `AssertionError` through `assertContractOrDiagnose()` means a future regression is converted to a diagnostic and CI stays green—the opposite of the guard this PR is meant to add. Please rebase, remove both TODO markers and the assertion-catching helper so failures propagate normally, or close this PR as superseded by #2098’s hard tests.

Found by Codex.

integrationTests/database/copy-db-blob-store.test.ts
```diff
@@ -0,0 +1,233 @@
...
+				const inCopy = filesUnder(copyRoot);
+				// Not part of the ch.3 contract — a harness failure (e.g. copy-db never ran) must throw
+				// for real, not get swallowed as "known bug #2048" by the diagnose wrapper below.
+				ok(inCopy.length > 0, 'the copy should contain the database files');
+				assertContractOrDiagnose(t, () => {
+					const blobFiles = inCopy.filter((f) => /blobs?\//i.test(f));
+					ok(
```
heskew: **Significant — filename counts do not establish the stated restorability contract.** This only checks that at least six paths happen to match `blobs?/`; the suite never opens/restores the target database and never reads a copied inline record or blob. A change could emit plausible or arbitrary blob files while leaving records missing, undecodable, or pointed at the wrong bytes, and this test would pass. If this PR is retained as complementary public-CLI E2E coverage beyond #2098, please open the copy through its own target path (or perform the documented restore) and assert both an inline control and a blob payload byte-for-byte.

Found by Codex.

heskew (changes requested):
Reviewed exact head `f69c4481` in its current merge context. Harper#2048 was closed by merged harper#2098 on August 29, after this PR’s last push and CI run. That changes the disposition: these formerly useful known-bug probes must now either become active regression tests or the PR should close as superseded by #2098’s hard coverage.

Codex found two PR-introduced test-contract gaps and left them inline: the fixed contracts remain non-enforcing TODOs, and the stated end-to-end restorability claim is never verified by opening the copied database and reading its inline/blob records.

For attribution on the existing review: @gemini-code-assist’s non-numeric exit-code finding is fixed at this head; its `rmSync` suggestion is superseded by async `rm`; and its remaining `try/finally` cleanup point is non-blocking and not duplicated here.

A post-#2098 main snapshot plus these PR commits passes `npm ci`, build, targeted oxlint, and `git diff --check`. The focused integration test could not start locally because the required `127.0.0.2` loopback alias is absent; that is environmental, not a product failure. GitHub’s green run predates #2098 and therefore does not exercise the now-fixed contract.

integrationTests/database/copy-db-blob-store.test.ts
kriszyp: Good call, and thanks for catching the timing — rebased onto post-#2098 `main`, both `{ todo }` markers and `assertContractOrDiagnose()` are gone, and the assertions propagate normally. 5 pass / 0 fail / 0 todo.

I went with the "become active regression tests" branch of your either/or rather than closing: #2098's coverage drives `copyDb()` directly, and this is the only thing exercising the public `copy-db` verb plus the documented restore end to end. The close-as-superseded option is still on the table and noted in the PR description if you'd rather not carry the suite.

— Claude Opus 5

integrationTests/database/copy-db-blob-store.test.ts
kriszyp: Fixed, and you were more right than the finding says. The suite now performs the restore the copy's own README documents — environment file back to `database/{db}.mdb`, `<target>-blobs/0/` back into the matching blob root — restarts Harper onto it, and reads every record back over HTTP, comparing each blob against the sha256 its seed call returned. The source environment and blob root are removed and asserted gone first, so the read-back can't be answered by surviving source files.

Making that read-back real surfaced that the fixture had never stored a record at all: `Seed` is a `loadAsInstance = false` resource, so Harper dispatches `(query, data)`, and `post(body)` had bound `body` to the query — every seeded field was `undefined`, all ten seeds wrote one record under the key `"undefined"`, and the four "inline controls" were blobs. `GET /Doc/blob-0` 404'd against the live source. Exactly the failure mode you described ("plausible blob files while records are missing"), already present.

— Claude Opus 5

heskew (approved):
Re-approved at exact head `810a6beaffbe2102d6f743d1e8480b7a4dba87e6`.

The two PR-introduced test-contract gaps I previously raised through Codex are fixed. Known-fixed contracts now use ordinary enforcing assertions rather than TODOs, and the restore claim is exercised end to end: the test reads the source data, removes and verifies absence of the source LMDB/blob roots, restores the copied database and companion blob root, restarts Harper, then reads every blob back and verifies its seed-provided SHA-256 alongside inline controls. The `Seed` resource signature is also corrected for `loadAsInstance = false`. This matches `bin/copyDb.ts`'s `<target>-blobs/0` layout and the README restoration steps.

Separately, @gemini-code-assist's findings are addressed: non-numeric exit codes are handled, temporary cleanup uses asynchronous removal in the teardown/finally-equivalent path, and those earlier comments are resolved or outdated. No new PR-introduced issue was found in the current Codex pass.

Exact-head verification passed: dependency install, build, required lint, Prettier on all four changed files, and diff checks; the focused regression completed with 5 passing, 0 failing, and 0 todo. GitHub CI is fully green across build, lint, unit, and integration platforms. The PR is open, non-draft, mergeable, and ready to land.

Issue #2107, opened 2026-08-05
RocksDB read snapshot leaks permanently: registry holds a strong ref to TransactionHandle, GC finalizer never calls close(), and DatabaseTransaction.abort() cannot release a save()-created transaction


harper-joseph:
## Summary

`oldestSnapshotTime` never advances past process start, so RocksDB is never permitted to discard
obsolete row versions. Compaction runs — it just can't drop anything. On a high-churn table this
reaches **~4 keys per live row in ~5 days**, and a range-scan iterator then spends its time
reseeking past dead versions: **43,647 reseeks vs 51** on a freshly-restarted node, degrading
bounded range scans **21×**. A process restart is the only reclamation path.

Point reads and primary-key scans are unaffected, so this hides until something latency-sensitive
does a secondary-index range scan.

**Version:** `harperfast/harper-pro:5.1.26` (Fabric, 4-node production cluster running the
`@harperfast/prerender` component). Not yet verified against v5.2.0.

## The smoking gun

`oldestSnapshotTime` equals process start on **every** node in the cluster:

| node | uptime | oldestSnapshotTime age |
| ---- | ------ | ---------------------- |
| A (aged)  | 4d19h | **4.76 days** |
| B (aged)  | 4d21h | ~4.9 days |
| C (fresh) | 4h    | **3.9 hours** |
| D (fresh) | 3h    | ~3h |

Something takes a snapshot at startup and holds it for the life of the process.

## Consequences, measured

Table: `RenderSchedule` — PK `cacheKey: String`, indexed `nextRenderTime: Long`. **Every write
updates the indexed attribute** (it carries both the next-run time and a claim lease), at a
measured **7–8 writes/sec/node**.

`system_information.attributes=["metrics"]`, aged node vs freshly-restarted node:

| metric | node A (4d19h) | node C (4h) | ratio |
| ------ | -------------- | ----------- | ----- |
| `estimateNumKeys` | 1,901,091 | 513,897 | — |
| live rows owned | ~490,000 | ~420,000 | — |
| → **keys per live row** | **3.9** | 1.22 | |
| `numberReseeksIteration` | **43,647** | 51 | **856×** |
| `sstReadMicros` (count) | 5,283,389,458 | 27,836,970 | 190× |
| `dbSeekMicros` (median) | **256 ms** | 35 ms | 7.4× |
| `liveSstFilesSize` | 362 MB | 98 MB | 3.7× |
| blockCache miss rate | 98.3% | 90.8% | |

[...]

PR #2128, opened 2026-08-09
Release the RocksDB transaction handle on abort and on a failed direct commit

kriszyp:
Two paths could strand a RocksDB transaction handle, permanently pinning a read snapshot so the database could never discard obsolete row versions. Both now release it unconditionally, and `system_information.metrics` gains `numSnapshots` so the condition is visible.

Reported as [#2107 — RocksDB read snapshot leaks permanently](https://github.com/HarperFast/harper/issues/2107): on a production cluster a high-churn secondary index reached ~4 keys per live row in 5 days, `numberReseeksIteration` went 51 → 43,647, and bounded range scans degraded ~21x, with a process restart as the only recovery.

## What was wrong

**`abort()`** released the native handle only through `while (this.readTxnsUsed > 0) this.doneReadTxn()`, and `getReadTxn()` is the only thing that sets `readTxnsUsed`. A write-first link — `save()` built the native transaction with no prior read — leaves it `undefined`, so `undefined > 0` is false and the loop never runs. That shape is reachable: a write-only link in a multi-store chain never calls `getReadTxn()` (see the comment at `chainStillActive`), and `abortDueToTimeout()` walks the chain calling `abort()` on exactly those links.

**`directCommitSync()`** removed the transaction from tracking but never nulled `this.transaction`. On a throwing `commitSync` that left an open, untracked handle nothing could reach; on success it left the object serving an already-closed transaction from `getReadTxn()`'s early return.

**`doneReadTxn()`'s native abort was unguarded.** `RocksTransaction.abort()` throws on an already-released handle, and `doneReadTxn()` is the *first* statement of `abort()` — so a throw there skipped everything after it: `CLOSED` state, blob cleanup, context release. `abortDueToTimeout()` and `abortChainAfterRetries()` have no handler of their own. `releaseReadTxn()` was already guarded; this brings the other paths in line, and it is what makes the new unconditional release reachable on the read path at all.

Neither leak self-heals: rocksdb-js's descriptor registry holds a strong reference to the handle, so an unreleased one survives GC and holds its snapshot until the process exits. HarperFast/rocksdb-js is being fixed to release at GC as well; these are the paths that should never have depended on that.

## What changed

The four detach-abort-swallow blocks are now one `abortNativeTransaction()` helper, so the reasoning behind swallowing the throw lives in one place instead of drifting across four call sites.

`system_information.metrics` gains `numSnapshots` per database. `oldestSnapshotTime` was the only snapshot signal and it stops moving once the oldest snapshot is pinned, so accrual was invisible. Note this reads `rocksdb.num-snapshots`, which lands in rocksdb-js 2.8; against the currently pinned 2.7.1 the field is simply absent, so this is inert until that ships and the pin moves.

## Verification

`npm run test:unit:resources` — 1517 passing, 15 pending. Prettier and `lint:required` clean.

Four new cases in `unitTests/resources/txn-tracking.test.js`. Three are object-double checks of the release logic; the fourth builds the write-first shape on a **real** RocksDB store and asserts that both the `registryStatus()` transaction count and `rocksdb.num-snapshots` return to baseline after `abort()` — that one is the actual #2107 property. Fails-on-base: with `origin/main`'s `resources/DatabaseTransaction.ts` restored, the write-first, real-snapshot, and direct-commit cases all fail; the read-created control passes on both, so it is not vacuous.

Not observable end-to-end beyond that: the leak's user-visible symptom is a multi-day scan degradation, so an integration test would have to age a store. The real-RocksDB unit assertion on `num-snapshots` is the closest honest proxy.

## For the human reviewer

1. **Unconditional release as a fallback after the refcount loop, rather than one unified release path.** Two mechanisms now exist that must stay mutually exclusive — the loop owns the read-created handle, the fallback owns the write-first one — and the seam between them is exactly where the unguarded-throw gap lived. Chosen because collapsing them into a single mechanism means reworking the `readTxnsUsed` / `readTxnRefCount` pair and its deliberate off-by-one base, which is a larger change than this fix warrants. Reversible with a local refactor. Say no if you'd rather do the counter rework now.

2. **`directCommitSync()` aborts and rethrows on a failed `commitSync`.** That makes a failed sync commit terminal-rollback rather than leaving the handle for the caller to retry. Its two callers (`resources/replayLogs.ts`) already catch and log, and each version boundary builds a fresh transaction, so no caller loses a retry it was using. Changing this later alters observable write-atomicity behavior, so it is worth a second opinion.

3. **`numSnapshots` ships ahead of the stat it reads.** It is inert until rocksdb-js 2.8 lands and the pin moves. The alternative was holding this until the dependency ships; kept because it costs nothing and the metric is what makes the next occurrence a five-minute diagnosis rather than a multi-hour one.

**Where to look hardest:** `abort()` in [resources/DatabaseTransaction.ts](resources/DatabaseTransaction.ts). The question worth the most scrutiny is whether zeroing `readTxnsUsed` in the fallback can strand an outstanding iterator that the refcount was tracking — I believe not, because the loop above has already drained any nonzero count and `doneReadTxn()` no-ops on a nulled handle, but that is the invariant the whole change rests on.

**What the tests do not prove:** that the write-only-`next`-link path actually reaches `abort()` with a live handle in a real multi-store request. The chain shape is asserted by main's own comments and by `abortDueToTimeout()`'s walk, not by a test that builds it end to end — a second `database:` in the unit suite resolves to the same store, so no `next` link forms there.

**Coverage:** Codex and Gemini ran on the implementation; Codex and Gemini again on the final artifact, with the Harper-domain leg on the middle revision. Findings acted on: the unguarded `doneReadTxn()` abort, the object-double-only test coverage, the duplicated release blocks, the missing throwing-abort case, an overstated `numSnapshots` comment (an in-flight read legitimately holds a snapshot — the signal is a count that stays nonzero while nothing is reading), and a helper that had been inserted between `recordCommitLatency`'s doc comment and its function. Dropped after checking the code: Gemini's claim that `trackedTxns.delete(this)` is skipped for read-created handles (`doneReadTxn()` already deletes them) and its strict-null-check concern (`tsconfig.json` sets `strict: false`).

<sub>Review-Coverage: authored=unknown; ran=none; rounds=1 @ b7c4a3d0c3e6</sub>

<sub>Human-Review-Need: 4 @ b7c4a3d0c3e6</sub>

### DIFF
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -523,11 +523,7 @@ export class DatabaseTransaction implements Transaction {
 	 */
 	releaseReadTxn(): void {
 		const transaction = this.detachOwnedTransaction();
-		try {
-			transaction?.abort();
-		} catch (error) {
-			harperLogger.debug?.('releasing timed-out read transaction', error);
-		}
+		abortNativeTransaction(transaction, 'releasing timed-out read transaction');
 		this.completeDeferredContextRelease();
 	}
 
@@ -1254,11 +1250,7 @@ export class DatabaseTransaction implements Transaction {
 			// throws. abort() rather than the native abort alone: it reclaims blobs a replayed write
 			// staged.
 			this.detachOwnedTransaction();
-			try {
-				transaction?.abort();
-			} catch (abortError) {
-				harperLogger.debug?.('aborting a transaction whose synchronous commit failed', abortError);
-			}
+			abortNativeTransaction(transaction, 'aborting a transaction whose synchronous commit failed');
 			try {
 				this.abort();
 			} catch (abortError) {
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -143,6 +143,19 @@ export function getOutstandingCommits(): { count: number; oldestAgeMs: number | 
 // about the caller's pattern, not the individual commit.
 let replayedWritesWarned = false;
 
+/**
+ * Abort a detached native handle. RocksTransaction.abort() throws on one that was already
+ * committed or aborted, and every caller is a cleanup path whose own callers have no handler — a
+ * throw there would abandon the rest of the cleanup.
+ */
+function abortNativeTransaction(transaction: RocksTransaction, context: string): void {
+	try {
+		transaction.abort();
+	} catch (error) {
+		harperLogger.debug?.(context, error);
+	}
+}
+
 // The analytics module registers a recorder here at load (dependency inversion, mirroring
 // `replicationConfirmation` below) so the storage layer doesn't statically import the analytics/server
 // modules. Unset until analytics loads, and when analytics is disabled the recorder call is cheap.
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -148,7 +148,8 @@ let replayedWritesWarned = false;
  * committed or aborted, and every caller is a cleanup path whose own callers have no handler — a
  * throw there would abandon the rest of the cleanup.
  */
-function abortNativeTransaction(transaction: RocksTransaction, context: string): void {
+function abortNativeTransaction(transaction: RocksTransaction | null | undefined, context: string): void {
+	if (transaction == null) return;
 	try {
 		transaction.abort();
 	} catch (error) {
--- a/unitTests/resources/txn-tracking.test.js
+++ b/unitTests/resources/txn-tracking.test.js
@@ -376,6 +376,71 @@ describe('Write txn timeout', () => {
 			setExpiration(30000);
 		}
 	});
+	// abort() releases the native handle through the read refcount loop, and only getReadTxn() ever
+	// sets readTxnsUsed. A write-first link — save() built the native transaction with no prior read,
+	// the blind-write-to-a-second-database shape above — leaves it undefined, so `undefined > 0` is
+	// false and the loop stranded the handle. rocksdb-js's descriptor registry holds a strong
+	// reference to it, so a stranded handle keeps its read snapshot and blocks reclamation of
+	// obsolete versions for that whole database until the process exits (#2107).
+	it('releases a write-first native handle on abort, with no read refcount to drive the loop', function () {
+		if (isLMDB) this.skip();
+		const txn = new DatabaseTransaction();
+		txn.open = TRANSACTION_STATE.OPEN;
+		let aborted = 0;
+		txn.transaction = {
+			abort() {
+				aborted++;
+			},
+		};
+		assert.equal(txn.readTxnsUsed, undefined, 'test setup: a write-first handle has no read refcount');
+
+		txn.abort();
+
+		assert.equal(aborted, 1, 'abort() must release the native handle');
+		assert.equal(txn.transaction, null, 'the released handle must not be reachable for reuse');
+	});
+
+	// Control for the above: the refcount loop still owns the release for a read-created handle, and
+	// the unconditional fallback must not abort it a second time (RocksTransaction.abort() throws on
+	// an already-aborted handle, which abort() must not propagate to abortDueToTimeout et al).
+	it('releases a read-created native handle exactly once on abort', function () {
+		if (isLMDB) this.skip();
+		const txn = new DatabaseTransaction();
+		txn.open = TRANSACTION_STATE.OPEN;
+		let aborted = 0;
+		txn.transaction = {
+			abort() {
+				aborted++;
+			},
+		};
+		txn.readTxnsUsed = 1; // as getReadTxn() would leave behind
+
+		txn.abort();
+
+		assert.equal(aborted, 1, 'the read refcount loop must release it, and the fallback must not re-abort');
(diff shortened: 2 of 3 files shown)


resources/DatabaseTransaction.ts
```diff
@@ -138,6 +138,19 @@ export function getOutstandingCommits(): { count: number; oldestAgeMs: number |
...
+ * committed or aborted, and every caller is a cleanup path whose own callers have no handler — a
+ * throw there would abandon the rest of the cleanup.
+ */
+function abortNativeTransaction(transaction: RocksTransaction, context: string): void {
+	try {
+		transaction.abort();
+	} catch (error) {
+		harperLogger.debug?.(context, error);
+	}
+}
```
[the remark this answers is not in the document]
kriszyp: Applied — the helper now takes `RocksTransaction | null | undefined` and early-returns on an explicit `== null` check. All four call sites already pass a non-null local, so this is defensive rather than a fix, but it makes the helper safe on its own terms rather than depending on every caller having checked first.

resources/DatabaseTransaction.ts
```diff
@@ -1093,7 +1116,18 @@ export class DatabaseTransaction implements Transaction {
...
+		// a throw the handle stays open and is already untracked, so this is its last release point
+		// (#2107).
+		this.transaction = null;
+		try {
+			transaction?.commitSync();
+		} catch (error) {
+			abortNativeTransaction(transaction, 'aborting transaction after failed direct commit');
+			throw error;
+		}
 	}
```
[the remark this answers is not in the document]
kriszyp: Not applying this one. `directCommitSync()` is only reachable from `replayLogs.ts`, which calls it on a transaction it just built, and the `trackedTxns.delete(this)` above must run regardless — an early return would skip it. The null concern itself is already covered: `transaction?.commitSync()` uses optional chaining, so a nullish handle cannot throw and the `catch` is unreachable, and `abortNativeTransaction` now null-guards internally per the sibling comment. Adding a second guard here would be dead code on a cold path.

unitTests/resources/txn-tracking.test.js
```diff
@@ -333,6 +333,117 @@ describe('Write txn timeout', () => {
...
+			abort() {
+				aborted++;
+			},
+		};
+		assert.equal(txn.readTxnsUsed, undefined, 'test setup: a write-first handle has no read refcount');
+
+		txn.abort();
+
+		assert.equal(aborted, 1, 'abort() must release the native handle');
+		assert.equal(txn.transaction, null, 'the released handle must not be reachable for reuse');
```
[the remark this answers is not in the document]
kriszyp: Applied — all four new tests now use `assert.strictEqual`.

utility/environment/systemInformation.ts
```diff
@@ -544,6 +545,11 @@ type RocksDBStats = {
 	numberKeysWritten: number;
 	numberReseeksIteration: number;
 	numRunningFlushes: number;
+	// Live RocksDB snapshots for this database. In-flight reads legitimately hold one, so what
+	// matters is a count that stays nonzero while nothing is reading: obsolete versions behind the
+	// oldest snapshot cannot be discarded for as long as it is held (#2107). oldestSnapshotTime
+	// alone can't show accrual — it stops moving once the oldest snapshot is pinned.
+	numSnapshots: number;
```
ldt1996: Since this field is known-absent under the pinned rocksdb-js 2.7.1, `numSnapshots?: number` would record that reality in the type. With strict null checks off it changes nothing at compile time, but the type is the only place a consumer learns the field may not be there yet.

unitTests/resources/txn-tracking.test.js
```diff
@@ -376,6 +376,116 @@ describe('Write txn timeout', () => {
 			setExpiration(30000);
 		}
 	});
+	// A write-first link (save() built the handle with no prior read) has no readTxnsUsed, so the
+	// refcount loop never runs and the handle was stranded — permanently, since rocksdb-js's
+	// registry keeps it alive and it holds a read snapshot (#2107).
+	it('releases a write-first native handle on abort, with no read refcount to drive the loop', function () {
```
ldt1996: These five cases live in `describe('Write txn timeout')` but none of them involve a timeout. A nested `describe('abort releases the native handle')` would keep the suite navigable as this block grows. Non-blocking.

utility/environment/systemInformation.ts
kriszyp: Done in b7c4a3d0c — `numSnapshots?: number`, with a note that rocksdb-js 2.7.1 does not report `rocksdb.num-snapshots` (confirmed: its stats typings carry `rocksdb.oldest-snapshot-time` but no num-snapshots key).

— Claude Opus 5

unitTests/resources/txn-tracking.test.js
kriszyp: Done in b7c4a3d0c — the five cases now sit in a nested `describe('abort releases the native handle')` inside the timeout suite. Suite still passes (17 passing, 2 pending under RocksDB).

— Claude Opus 5

PR #2129, opened 2026-08-09
Fix HTTP startup after pre-ready worker restart

kriszyp:
independent-review: claude + gemini + cursor-composer + harper-domain across the pre-rebase rounds (final delta on 1866305f9); rebased onto main at 51d1a4a2b636 with a Gemini-only round (codex/domain timed out) — findings manually triaged below, all false-positive or pre-existing/out-of-scope.

## Summary

- Keep HTTP startup readiness with the logical worker slot, so a managed replacement can satisfy the original boot wait.
- Report each worker startup phase and turn bootstrap rejection into an explicit worker failure.
- Add an end-to-end uWS regression: crash the initial HTTP worker 0 once, then require the replacement to finish startup and run scheduler work.
- Use Node 22's `Promise.withResolvers()` for the slot readiness handle while preserving the prior ES2022 type surface.

## Root cause and scope

The scheduled-run artifact proves stdout/stderr capture and process launch were working. The initial HTTP worker exited before its ready message; `manageThreads` launched a replacement, but startup was still awaiting the promise owned by the dead Worker object. The replacement created a different promise, so the original `Promise.all()` could never settle.

The exact trigger for that first worker exit was not retained in the old logs. This PR fixes the confirmed wait-ownership hang and makes a repeat observable by phase; it does not claim to identify the original worker failure.

The 60-second startup diagnostic is intentionally non-destructive. A hard boot cutoff or forced `Worker.terminate()` would create a new availability risk for slow deployments and Bun. A worker that wedges without exiting therefore remains an operational follow-up rather than changing production boot policy here.

## For the human reviewer

The latest review-thread fix is mechanical: `Promise.withResolvers()` replaces the equivalent manual constructor, and `tsconfig.json` adds only `ES2024.Promise` on top of the previous implicit ES2022 full library surface. The existing boot-timeout, restart-budget, and unidentified-trigger decisions remain unchanged.

**Open decisions, deliberately declined here** — each is a semantic call about this PR's core rather than a mechanical fix, so they are left for the human reviewer rather than settled by a follow-up round:

- **Pre-ready restart budget.** A worker that can *never* start (a root component with a top-level throw, `EADDRINUSE` on worker 0's dedicated listener) now burns the full steady-state 50-restart budget — reloading the whole component graph and writing the same stack 50 times — before [`onRestartExhausted`](https://github.com/HarperFast/harper/pull/2129/changes?w=1#diff-545704b0f9ebb0c229b5f127dc8c710d5061aedb1e01e00514a96385babb1196R157) rejects the slot. The pre-PR `worker.on('error', reject)` failed in seconds. A smaller dedicated pre-ready budget (2–3) would restore fast failure without touching steady-state restarts; that is the tradeoff Kris's rebase decision explicitly accepted (recovery through the existing restart manager, bounded but noisy).
- **A slot that becomes ready after startup already rejected.** [`finishStartup()`](https://github.com/HarperFast/harper/pull/2129/changes?w=1#diff-545704b0f9ebb0c229b5f127dc8c710d5061aedb1e01e00514a96385babb1196R81) runs for every slot in the `finally`, detaching the `message` listener, so a worker that posts `child_started` after `Promise.all` has already rejected is never pushed into `workers`: a live thread holding its `SO_REUSEPORT` bind, invisible to routing and idleness tracking. Inert for `bin/run.ts` and `bin/lite.ts` (both exit on rejection), but real for an embedder that catches the rejection and keeps serving — `unitTests/apiTests/setupTestApp.mjs:129-130` is that shape. Tearing the slot's workers down on `failStartup` would make the failure total rather than partial. Domain adjudication downgraded this to a nit on the shipped entry points.
- **The 60s startup diagnostic does not bound a wedged worker.** [The timer](https://github.com/HarperFast/harper/pull/2129/changes?w=1#diff-545704b0f9ebb0c229b5f127dc8c710d5061aedb1e01e00514a96385babb1196R153) reports a worker that never becomes ready and never exits, but does not terminate it; `restartWorkers()` enforces a termination deadline for replacements, initial startup has no equivalent. Not a regression — see *Root cause and scope* for why a hard boot cutoff was rejected — but it is the one gap left inside this PR's stated purpose.
- **Bun loses coverage of this regression.** [The suite is skipped under Bun](https://github.com/HarperFast/harper/pull/2129/changes?w=1#diff-fb02a4d843e979ef1cf7c6397b8c97c34649cbd0834f757a1f248369e7fd9b85R32) because the pre-ready exit injection wedges the whole Harper process and it aborts with SIGABRT. Precedent is `integrationTests/server/rolling-restart.test.ts`, which already opts out. A Bun-only deferred exit was tried and rejected: it worked in the one run measured but races readiness with no establishable margin, trading a hard failure for a flaky one. The cost is real — an HTTP worker can die pre-ready under Bun in production (OOM, a native crash) and nothing in CI exercises the replacement-satisfies-the-boot-wait path there.
- **The new second test couples to a log string.** It asserts `HTTP worker slot 0 exited before ready` appears in `hdb.log`, which comes from `socketRouter.ts` in this same PR. It exists because the deferral experiment showed how quietly this suite can stop testing anything — on Node the deferred exit landed *after* worker 0 was ready, the original assertions still passed, and nothing signalled the pre-ready path was no longer covered. If the wording changes the test fails loudly rather than silently; that is a coupling a reviewer should agree to. It also self-skips without `HARPER_INTEGRATION_TEST_LOG_DIR`, which CI always sets but a bare local `npm run test:integration` does not.

Test coverage is intentionally narrow: one crash-during-load scenario against a single primary worker. Restart exhaustion and non-primary-worker failure paths are untested.

**Rebase (2026-08-28):** onto latest `origin/main`. No textual conflicts — the branch's own 9 commits replayed cleanly. One region merited a closer look: `server/threads/threadServer.js`'s `startServers()` overlaps with main's #2314 (pre-ready `parentPort.ref()` + a `loadRootComponents` rejection catch); this PR's own commit already consolidates that into the unified `started.catch()` (fatal-log + `realExit(1)`/throw), and an earlier review round (`c9c292883`) already restored the fatal log level there — confirmed the net behavior is unchanged by diffing this PR's file set against its own prior base vs. the new one. Build, `lint:required`, and all six `## Verification` commands below pass unchanged on the rebased head.

The rebase's pre-push review round only got independent coverage from Gemini — the graded (codex) and Harper-domain legs both timed out before producing output, so nothing filtered Gemini's findings. Manually verified each against the code rather than leaving them unadjudicated:
- *"`Promise.withResolvers()` unsupported on Node 20.x"* — not reachable: `package.json` `engines` requires `^22.18.0 || >=24.0.0`; Node 20 was never supported.
- *"`isMainThread` used without import in `threadServer.js`'s new catch"* — already imported and used elsewhere in the file (`server/threads/threadServer.js:3`), predating this diff.
- *"`onExit`'s `wasShutdown` skip can deadlock shutdown while workers are still starting"* — the slot's `ready` promise is never awaited by the shutdown path: `bin/run.ts`'s SIGINT/SIGTERM/SIGQUIT handler calls `beginProcessShutdown()` then `process.exit(0)` unconditionally, with a comment already noting nothing below the handler is load-bearing on worker state. The behavior itself (no `failStartup()` from `onExit`) predates this PR too — confirmed by diffing `21a273c06` against its parent, which shows only the log call changed, not control flow.
- Nit about comment density on `socketRouter.ts`/the new integration test was left as-is — those comments document non-obvious invariants (thread-id-reads-as--1-after-exit, Bun TDZ behavior), which is what this repo's comment convention asks for.

## Verification

- `npm run build`
- `npm run lint:required`
- `HARPER_UWS_HTTP=1 npm run test:integration -- integrationTests/server/initial-http-worker-restart.test.ts`
- standard HTTP mode for the same integration test
- `npx mocha unitTests/server/threads/threadServerListenOnPorts.test.js`
- `HARPER_UWS_HTTP=1 npm run test:integration -- integrationTests/components/scheduler-jobs.test.ts`

The full resources gate reached 1,708 passing tests with 9 unrelated audit/storage-engine failures in the local isolated-storage environment. The main unit gate entered a pre-existing component-loader hang after progressing through its earlier suites. The targeted startup regression, build, and required lint pass on the final head — all re-run and green after the rebase.

CI on the rebased head: every check green except `Unit Test (Node.js v22)`, which is flaky **on `main` as well** and is not caused by this PR:

- Five runs of this job against this same head failed on five *different*, mutually-unrelated tests: an fs-cleanup teardown race (`ENOTEMPTY` rmdir) in a schema/relationship test, an `HNSW greedy routing` numeric-assertion mismatch, `unitTests/resources/txn-tracking.test.js:77` (`Txn Expiration / Slow txn will expire`), `Subscription replay` (`in-flight id 14026 never delivered`), and `Caching / Can load cached data`. No two runs failed on the same test, and none of those files is in this PR's diff.
- Sampling the last 16 completed `Unit Test` runs on `main`, the `Node.js v22` job failed **7 times** (e.g. runs [33436356216](https://github.com/HarperFast/harper/actions/runs/33436356216) at `fb762a365` and [33397431652](https://github.com/HarperFast/harper/actions/runs/33397431652) at `655787409`). The v24 and v26 jobs pass in all of those runs, including on this PR's head.

So the red required check is a gate problem on `main`, not a signal about this change. Three reruns of the failed job were spent confirming that; a fourth is not worth the minutes.

- Four runs of this job on this head failed on four *different*, mutually-unrelated tests: an fs-cleanup teardown race (`ENOTEMPTY` rmdir) in a schema/relationship test, an `HNSW greedy routing` numeric-assertion mismatch, `unitTests/resources/txn-tracking.test.js:77` (`Txn Expiration / Slow txn will expire`), and `Subscription replay` (`in-flight id 14026 never delivered`). No two runs failed on the same test, and none of those files is in this PR's diff.
- Sampling the last 16 completed `Unit Test` runs on `main`, the `Node.js v22` job failed **7 times** (e.g. runs [33436356216](https://github.com/HarperFast/harper/actions/runs/33436356216) at `fb762a365` and [33397431652](https://github.com/HarperFast/harper/actions/runs/33397431652) at `655787409`). The v24 and v26 jobs pass in all of those runs, including on this PR's head.

So the red required check is a gate problem on `main`, not a signal about this change.

<sub>Review-Coverage: authored=claude; ran=gemini; blocked=codex(timeout),claude(fallback)(out-of-budget),domain(timeout); declined=cursor-grok,cursor-composer; rounds=4 @ 51d1a4a2b636</sub>

<sub>Human-Review-Need: 3 @ 51d1a4a2b636</sub>


cb1kenobi (approved):

kriszyp:
Rebased onto latest `main` (past #2112, "surface swallowed uWS startup-listen failures instead of hanging silently").

**Conflict:** #2112 makes any pre-ready HTTP worker exit fail startup immediately; this PR's slot-owned readiness instead waits through managed replacement workers and fails only after `onRestartExhausted`. Per Kris's decision, retained this PR's restart-through-pre-ready-exit recovery while keeping #2112's listener-timeout diagnostics and explicit failure logging — see `server/threads/socketRouter.ts` (slot-owned `ready` + `onRestartExhausted`) and `server/threads/threadServer.js` (phase reporting + `realExit(1)`).

Build and the `initial-http-worker-restart` integration suite pass post-rebase.

Independent pre-push review (codex + gemini + Harper-domain adjudication) ran across the rebase; fixed two findings directly:
- Restored `harperLogger.fatal` for a failed HTTP listener bind (`server/threads/threadServer.js:254`) — the consolidated startup-failure handler had downgraded this to `.error`, silencing it for anyone alerting on FATAL.
- Added the file-level JSDoc `integrationTests/README.md` requires to the new integration test, and dropped a comment that only restated the code next to it.

## For the human reviewer

Left as-is (pre-existing to this PR, not introduced by the rebase):
- `server/threads/manageThreads.js:442-448` — a deterministic pre-ready failure now burns the full 50-restart budget instead of failing fast. This is the exact tradeoff Kris's rebase decision accepted (recovery via the existing restart manager, bounded but noisy). Terminal-error classification or a smaller/backed-off startup restart budget was called out as a separate follow-up, not a rebase decision.
- `server/threads/socketRouter.ts:153-157` — the 60s startup timer diagnoses a wedged (never-exits, never-ready) worker but doesn't bound it; `restartWorkers()` already enforces a termination deadline for replacements, initial startup has no equivalent. Not a regression, but worth closing given this PR's stated purpose.
- `server/threads/socketRouter.ts:98-106` — a slot that becomes ready after startup has already rejected is never added to `workers`; domain adjudication confirmed no observable impact (nothing reads that array in the rejection path) and downgraded this to a nit.
- Test coverage is intentionally narrow — the one crash-during-load scenario against a single primary worker; restart exhaustion and non-primary-worker failure paths are untested.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

server/threads/socketRouter.ts
```diff
@@ -73,63 +74,96 @@ export async function startHTTPThreads(threadCount = 2, dynamicThreads?: boolean
...
+	let finishCurrentStartup = () => {};
+	let startupAttempts = 0;
+	// A Worker's threadId reads back as -1 once it has exited, which is exactly when the diagnostics
+	// below run, so each attempt records its id while the worker is still alive.
+	let lastThreadId = -1;
+	const ready = new Promise<void>((resolve, reject) => {
+		resolveReady = resolve;
+		rejectReady = reject;
+	});
+	const finishStartup = () => {
```
cb1kenobi, line 94: Since we're now supporting Node.js 22 or newer, we can use `withResolvers`:
```suggestion
	cost { promise: ready, resolve: resolveReady, reject: rejectReady } = Promise.withResolvers();
```

cb1kenobi (approved):

cb1kenobi (commented):
Barbarian reviewed 51d1a4a2 and found no blocking issues. No confirmed blocking issues were found on the changed lines. The slot-owned readiness logic correctly survives managed worker replacement and reports terminal startup failure.

—
_Generated by Barber AI_

server/threads/socketRouter.ts
kriszyp: Implemented in `3ac3445d5` using `Promise.withResolvers<void>()`. TypeScript's `target: ES2022` does not declare that Node 22 runtime API, so `1866305f9` preserves the prior implicit `es2022.full` library surface and adds only `ES2024.Promise` to `tsconfig.json`'s `lib`. `npm run build`, `npm run lint:required`, and the uWS worker-restart integration test pass.

(Re-posting: this reply was written into an unsubmitted draft review and so was never actually visible — sorry for the silence on a resolved thread.)

cb1kenobi (commented):
Barbarian reviewed 51d1a4a2 and found no blocking issues. No new blocking issues were confirmed on changed lines. Existing discussions already cover the startup-readiness behavior and diagnostics.

—
_Generated by Barber AI_

PR #2133, opened 2026-08-10
Return vector distances for selective filtered queries

### LINKED ISSUE #2105: Filtered vector post-sorts omit distance and ignore the requested metric
## Problem

A vector query with indexed scalar `conditions` can take Harper's selective exact post-sort path. Two observable problems occur there:

1. If the filter leaves one row, JavaScript's `Array.sort()` never calls the vector comparator, so selecting `$distance` returns no value.
2. For multiple post-sorted rows, the HNSW property resolver checks `sortDefinition.type` instead of `sortDefinition.distance`, so a query-specific metric can be ignored in favor of the index default.

The documented query shape is otherwise valid:

```js
{
  conditions: [{ attribute: 'group', value: 'single' }],
  sort: { attribute: 'embedding', target: [1, 1], distance: 'euclidean' },
  select: ['id', '$distance']
}
```

## Expected

- `$distance` is finite even when the scalar filter leaves one row.
- The exact post-sort path honors the metric requested by `sort.distance`.
- Distance values already cached as zero are reused.

## Acceptance criteria

- Add regression coverage using a real table and HNSW index for the singleton projection.
- Add regression coverage where the index default and query-requested metric rank rows differently.
- Keep the planner's selective exact post-sort optimization; do not require callers to force vector-first execution.

kylebernhardy:
Fixes #2105.

Filtered vector queries that use the selective exact post-sort path now return `$distance` consistently while preserving the stored vector in projections. Each comparator owns and explicitly passes its sort definition, so concurrent queries, reused contexts, nested selects, and multi-level sorts cannot overwrite shared sort state. Exact-distance memoization is scoped by sort and uses stable entry keys with vector fallback; invalid targets and metrics are rejected before iteration, including empty result sets.

The regression suite covers selective singleton and mixed present/missing-vector results, index-assisted and exact post-sort paths, object-form selection, reused and concurrent contexts with different targets, stale sorts, missing contexts/entries, secondary vector sorts after indexed and unindexed fields, invalid targets, unknown metrics, and query metric overrides.

## Verification

- `npm run build`
- changed-file Oxlint with `--deny-warnings`
- `npx mocha unitTests/resources/vectorIndex.test.js` — 79 passing
- `npm run test:unit:resources` — 1,527 passing, 15 pending
- Direct tent~acle production query-shape harness verified Grounds, Lore, and Notes distance projection and relationships
- The broader main-unit command is blocked locally by an incompatible pre-existing `/Users/kyle/harper/database/weather` test database; GitHub CI is rerunning the clean-environment matrix
- Vector integration suite cannot start locally because its required `127.0.0.2+` loopback aliases are not configured on this machine

## Review

- Cross-model graded review plus Harper-domain adjudication: no actionable code findings remain
- Gemini CLI full review found two valid issues; both were fixed, and its focused final review reports no actionable code findings
- Human-Review-Need: 4 @ 66742c0c0940

## Compatibility note

Records missing the sorted vector project the existing exact-sort `Infinity` sentinel. JSON serializes that value as `null`.


resources/Table.ts
```diff
@@ -3551,6 +3551,7 @@ export function makeTable(options) {
 					}
 					const comparator = createComparator(sort);
 					return {
+						sort,
```
kriszyp: Threading the sort onto the iterator is the right fix for `$distance`, but the comparator right above it still gets its sort from the mutable `context.sort` set at `resources/Table.ts:3541`, and that assignment happens at `iterate()` time while `ordered.sort(comparator)` runs much later — after the `await dbIterator.next()` loop.

Concrete failure: two vector searches sharing one context (e.g. `Promise.all([T.search(qA, ctx), T.search(qB, ctx)])`). A's `iterate()` sets `context.sort = qA.sort`, then awaits the DB iterator; B's `iterate()` sets `context.sort = qB.sort`; A resumes and runs its synchronous `ordered.sort()`, whose `getAttributeValue` -> `customIndex.propertyResolver` reads `context.sort` and computes every distance against **B's target**, caching them under B's sort key. A's results come back silently ordered by the wrong vector. `$distance` itself is now correct (it reads `this.sort`), which makes the disagreement harder to notice, not easier.

There's a second instance of the same aliasing without any concurrency: `createComparator` recurses into `order.next` *first*, so for a multi-level sort the nested comparator ends up seeing the top-level `order` in `context.sort` rather than its own; and a nested sorted sub-select (`transformToOrderedSelect` at `resources/Table.ts:3828`) overwrites `context.sort` mid-iteration of the outer query, corrupting the next `dbOrderedAttribute` batch's comparator.

Suggest retiring the side-channel rather than adding a parallel one: pass `order` down explicitly — `getAttributeValue(entryA, order.attribute, context, order)` -> `customIndex.propertyResolver(value, context, entry, order)` — so the comparator and `$distance` share a single source of truth. The existing test at `unitTests/resources/vectorIndex.test.js:567` is sequential and wouldn't catch this; an interleaved variant would.

resources/indexes/HierarchicalNavigableSmallWorld.ts
```diff
@@ -1239,22 +1242,17 @@ export class HierarchicalNavigableSmallWorld {
 		const sortDefinition = context?.sort;
```
kriszyp: This still reads the sort from `context.sort`, and nothing ever clears it. Since `resources/Table.ts:5027-5033` registers this resolver for the vector attribute *itself* with `directReturn = true`, a subsequent query on the same context that does **not** sort by vector will still resolve `vector` to a distance number instead of the stored array:

```js
await fromAsync(T.search({ conditions: [...], sort: { attribute: 'vector', target: [1,1] }, select: ['id','vector'] }, ctx));
// context.sort is still set here
await fromAsync(T.search({ conditions: [...], select: ['id','vector'] }, ctx)); // vector comes back as a number
```

The PR already fixed exactly this staleness for `$distance` by reading `this.sort` off the iterator instead of the context; the vector-attribute resolver deserves the same treatment. Accepting the sort as a fourth parameter (and having `getAttributeValue` pass the comparator's `order`) fixes this and the concurrency issue on `resources/Table.ts:3554` together. Worth adding a test that selects `vector` (not `$distance`) on an unsorted query following a sorted one in the same context — the existing 'targetless or stale sort' test only covers `$distance`.

resources/indexes/HierarchicalNavigableSmallWorld.ts
```diff
@@ -1051,7 +1051,10 @@ export class HierarchicalNavigableSmallWorld {
 					? dotProductDistance
 					: searchCondition.distance === 'cosine'
 						? cosineDistance
-						: this.distance;
+						: searchCondition.distance
+							? null
+							: this.distance;
+		if (!fn) throw new ClientError('Unknown distance function');
```
kriszyp, line 1055: Good to reject unknown metrics here so the post-sort path matches `search()`. Same treatment is missing for `target` though: `search()` guards it at lines 974-975 (`'A target vector must be provided for an HNSW query'` / `'The target vector must be an array'`), but `exactDistance` passes `searchCondition.target` straight into the distance fn.

So `sort: { attribute: 'vector' }` with no target, on a result set with more than one row, escapes as a bare `Error('Euclidean distance comparison requires an array')` from `resources/indexes/vector.ts:4` — a 500 rather than the 400 a client-supplied bad query should get. (It also reaches here via `postOrdering` for secondary vector sorts, which drops `target` entirely — see the review.) Adding `if (!Array.isArray(searchCondition.target)) throw new ClientError('The target vector must be an array');` alongside the new metric guard covers both.

Note the new test at `unitTests/resources/vectorIndex.test.js:602` doesn't catch this because the `single` group has one row, so the comparator never runs.

resources/Table.ts
```diff
@@ -4871,8 +4872,13 @@ export function makeTable(options) {
...
 				$record: (object, context, entry) => (entry ? { value: object } : object),
-				$distance: (object, context, entry) => {
-					return entry && (entry.distance ?? context?.vectorDistances?.get(entry));
+				$distance: (object, context, entry, returnEntry, sort) => {
+					if (!entry) return;
+					const cachedDistance = entry.distance ?? context?.vectorDistanceCaches?.get(sort)?.get(entry);
+					if (cachedDistance !== undefined) return cachedDistance;
+					if (typeof sort?.attribute !== 'string' || !Array.isArray(sort.target)) return;
+					const customIndex = indices[sort.attribute]?.customIndex;
+					if (customIndex?.exactDistance) return customIndex.exactDistance(sort, object[sort.attribute]);
```
kriszyp: Two small things about the fallback:

1. It doesn't populate `context.vectorDistanceCaches`, and doesn't share the memo with `propertyResolver`. That's fine for the singleton case this PR targets, but if a query both sorts by the vector and projects the vector attribute (or the entry identity changed — note `transform` reassigns `entry` at `resources/Table.ts:3728` when the record has to be reloaded, guaranteeing a cache miss), the same distance gets computed twice per row. Routing through `customIndex.propertyResolver(object[sort.attribute], context, entry, sort)` instead would reuse the cache and populate it, at no extra cost.

2. `sort` here is `this.sort` from the enclosing iterator, which for a nested sub-select transform (`transform.call(this, value)` at `resources/Table.ts:3852`) is the **outer** table's sort while `indices` is the **nested** table's index map. If both tables happen to have a same-named vector attribute, `$distance` on the nested rows would be computed against the outer query's target. Cheap to rule out: only use the fallback when `this` is the iterator that actually owns this select — e.g. don't inherit `sort` across the nested `transformEntryForSelect` boundary.

unitTests/resources/vectorIndex.test.js
```diff
@@ -491,6 +491,192 @@ describe('HNSW search result loading (searchByIndex)', () => {
...
+	});
+
+	it('does not resolve $distance from a targetless or stale sort', async () => {
+		const context = {};
+		const nodesVisitedBefore = T.indices.vector.customIndex.nodesVisitedCount;
+		const targetless = await fromAsync(
+			T.search(
+				{
+					conditions: [{ attribute: 'group', value: 'single' }],
+					sort: { attribute: 'vector' },
```
kriszyp, line 651: Nice coverage overall — especially the `nodesVisitedCount` assertions proving the post-sort path is what's being exercised. Two gaps worth closing:

- This targetless case only has one row in the `single` group, so `ordered.sort()` never calls the comparator and `propertyResolver` is never reached. The assertion `targetless[0].$distance === undefined` passes because of the `Array.isArray(sort.target)` guard in `$distance`, not because a targetless sort is handled. Running the same query against the `metric` group (3 rows) would show the real behavior: a generic `Error('Euclidean distance comparison requires an array')` out of the comparator.

- The `Infinity` sentinel for a missing vector (line 548) is likewise only checked on a singleton, so it never goes through `compareKeys`/`convertToComparableKeys`. A case mixing rows with and without vectors in one group would pin down that missing vectors actually sort last rather than throwing or producing `NaN` — the ordering guarantee the sentinel exists for.

dawsontoth (approved):

kriszyp (changes requested):
The directionally looks good. I think some issues to resolve. And the mixed authorship is still confusing.

kylebernhardy:
Thanks for the careful review, Kris. All five threads are addressed and resolved, including the shared-context race, stale vector projection, eager target validation, memoized fallback, and mixed-vector coverage. The full vector file is 79/79 and the resource suite is 1,527 passing.\n\nOn authorship: the earlier commits mapped to my older `nizzlenitz` account because they used `kyle.bernhardy@gmail.com`; the PR is opened from `kylebernhardy`. Both accounts are me. New commits use `kyle@harperdb.io` and map to `kylebernhardy`. I did not rewrite the existing branch history solely to change attribution.\n\nI have requested your re-review.\n\nComment generated by kAIle (GPT-5)

kylebernhardy:
Authorship is now normalized: every commit on this PR is authored and committed as `Kyle Bernhardy <kyle@harperdb.io>`, which maps to `kylebernhardy`. The rewrite was metadata-only; the final tree is identical to the previously reviewed head.

kriszyp (approved):
Great, thank you!

Issue #1618, opened 2026-07-06
HARPER_SET_CONFIG applied after component loading (thread-timing race) and not persisted at install — env-var-configured components silently fail to register


heskew:
## Summary

`HARPER_SET_CONFIG` (and by extension env-var-driven config) is unreliable for anything component-loading reads at boot: it is applied to the in-memory config **after** components load on some platforms/timings, and the install-time application does not persist it to the written config file. Net effect: a component key delivered via env var may be silently ignored — the component never registers, no error, no warning.

Found while debugging PR #1616's integration-test 404s (the `modelsGateway` component); the mechanism is general and affects any env-var-configured deployment (Docker/Fabric-style) activating or configuring components via `HARPER_SET_CONFIG`.

## Evidence

**1. Ordering race (Linux CI, deterministic there; opposite on macOS).** From the failing suite's Harper instance log (PR #1616, run 28810803532, shard 4/6):

```
17:37:26.137 [main/0] [info]:  All root applications loaded          ← componentLoader ran
17:37:26.494 [http/1] [debug] [env-config]: Applied HARPER_SET_CONFIG at runtime   ← merged after
17:37:26.507 [http/1] [debug]: Config file updated with runtime env var values
```

Components were loaded from a config that did not yet contain the env-supplied key → `/v1/*` 404s. The same boot on macOS serves 200 — the runtime application happens to win there. Same code, thread-timing-dependent outcome.

**2. Install-time application doesn't persist.** `config/configUtils.ts:163` calls `applyRuntimeEnvVarConfig(configDoc, null, { isInstall: true })` before writing the fresh config file — but after a fresh local install with `HARPER_SET_CONFIG='{"modelsGateway":{"enabled":true}}'`, the written `harperdb-config.yaml` does **not** contain the key (verified directly). So even a fresh install under env config produces a file that component loading can't see the env config in, leaving only the racy runtime path (`configUtils.ts:341`).

(There is also a known sub-issue that `flattenObject` in `config/harperConfigEnvVars.ts` silently drops **empty-object** values like `modelsGateway: {}` — no leaf paths — noted in PR #1616's history. The non-empty `{enabled: true}` form still hits the two problems above, so fixing the flatten alone is insufficient.)

## Why this matters

- Env-var config is the primary mechanism for containerized deployments and the integration-test harness (`@harperfast/integration-testing` passes suite config exclusively via `HARPER_SET_CONFIG` — see its `harperLifecycle`).
[...]

PR #2144, opened 2026-08-11
fix(config): prune emptied parents when env-config removal vacates a subtree

### LINKED ISSUE #2067: Removing a nested entry from HARPER_SET_CONFIG can leave an empty object in the config file that permanently fails validation
## Summary

When a nested config entry is removed from `HARPER_SET_CONFIG` between restarts, the persisted
`harper-config.yaml` can end up holding an **empty object** at that position rather than losing the
key. Config validation then rejects the empty object for a missing required field, and the instance
fails to boot — permanently. It does not self-heal on any subsequent restart, because empty objects
in the base config file are deliberately restored during composition.

Recovery requires hand-editing the config file.

## Symptom

Repeating on every boot attempt, so the process never starts:

```
[main/0] [error]: Error initializing environment manager
[main/0] [error]: HdbError: Harper config file validation error:
                  'models.embedding.qwen3.backend' is required  statusCode=400
```

Persisted `harper-config.yaml` at that point:

```yaml
models:
  embedding:
    default:
      backend: openai
      baseUrl: http://127.0.0.1:8001/v1
      apiKey: <redacted>
      model: nomic-ai/nomic-embed-text-v1.5
    qwen3: {}          # <-- empty stub; fails validation
  generative:
    default:
      backend: openai
      baseUrl: http://127.0.0.1:8000/v1
      apiKey: <redacted>
      model: Qwen/Qwen2.5-7B-Instruct-AWQ
```

`models.embedding.qwen3` had previously been a complete entry supplied through `HARPER_SET_CONFIG`.
The env var was then changed to a `models` block that no longer contains `qwen3`.

## What is confirmed vs. what is not

**Confirmed:** the end state above, the boot failure, and that it is unrecoverable without manual
file surgery — restarting does not clear the stub.

**Not pinned down:** which step first produces the `{}`. Reading `composeConfigFromEnv()` in
`config/harperConfigEnvVars.ts`, the base file is a layer whose leaves are flattened and merged, so
an entry present in the file and *absent* from `HARPER_SET_CONFIG` would be expected to keep its
leaves rather than be emptied. That is the part worth tracing first — the reproduction reliably
produces the stub, but the mechanism that creates it is not established.

## What makes it unrecoverable (this part is traced)

`restoreBaseEmptyObjects()` (`config/harperConfigEnvVars.ts:356`) deliberately re-creates empty
objects found in the base config file:

```js
if (Object.keys(value).length === 0) {
    if (!(key in target)) target[key] = {};
}
```

So once `qwen3: {}` is written to the file, every later composition faithfully reproduces it, and
every later boot fails the same way. The intent is documented and reasonable (#1726 review: a bare
`componentName: {}` scope in the file is user content, not noise), but it also means a *malformed*
empty stub is sticky forever.

There is a related tension in `flattenObject()` (`config/harperConfigEnvVars.ts:329-339`): its
comment notes that an empty object contributes no leaves and that this is "load-bearing for removal
semantics" in env layers — `http: {}` in `HARPER_SET_CONFIG` means "no overrides under http". So an
empty object means *removal* in an env layer and *content to preserve* in the base file. An empty
object that is neither — the residue of a removed entry — satisfies neither rule and lands in the
file as invalid config.

## Reproduction

1. Start Harper with `HARPER_SET_CONFIG` containing two sibling entries under a validated nested
   map, e.g. `models.embedding.a` and `models.embedding.b`, each with all required fields. Let the
   config file be written.
2. Restart with `HARPER_SET_CONFIG` reduced to only `models.embedding.a`.
3. Boot fails with `'models.embedding.b.backend' is required`, and the persisted config shows
   `b: {}`.
4. Restarting again does not recover; the stub must be deleted from the file by hand.

## Suggested direction

Three separable questions, roughly in priority order:

1. **Where does the `{}` originate?** If a removed env key is meant to leave the base entry intact,
   the emptying is the primary bug. If removal is meant to delete the entry, it should be deleted
   rather than hollowed out.
2. **Should validation reject an empty object at a position that requires fields?** Treating `{}` as
   "absent" (prune before validation) would make removal semantics consistent with `flattenObject`'s
   documented behavior for env layers.
3. **A config-validation failure at boot should not be an unrecoverable crash loop.** Whatever the
   resolution to 1 and 2, the operator experience of "instance can never start again until someone
   edits YAML" is worth a guard of its own — at minimum an error message that names the offending
   key and states that it must be removed from the config file.

Care is needed not to regress #1618 / #1726: deliberate empty-object scopes in the config file must
keep working.

heskew:
Fixes #2067: removing a nested entry from `HARPER_SET_CONFIG` (or shrinking/unsetting any config env var) left an invalid `entry: {}` husk in the persisted `harper-config.yaml`; where validation requires fields (e.g. the `models.embedding` pattern map) the instance then crash-loops on every boot until someone hand-edits the file. Root-cause trace in [the issue comment](https://github.com/HarperFast/harper/issues/2067#issuecomment-5210024558).

## What changed

`deleteNestedValue` now prunes ancestor objects the deletion emptied — only when it actually deleted an existing leaf — and reports the pruned paths. Both removal paths (`handleDeletions` for a changed var, `cleanupRemovedEnvVar` for a removed one) route through it. Because a bare `name: {}` in the config file is protected user content (#1618/#1726), a file-declared empty scope that an env layer populates is recorded in a new `emptyScopeOriginals` state-file field (older state files default it; kept separate from `originalValues` so a marker can never mask or be consumed as a real leaf original), and restore consumes a marker only for a path the prune actually removed — so a scalar overwrite or an absent-leaf no-op can never resurrect a scope over live env-layer content. DESIGN.md gains an entry on the three meanings of `{}` in the config system.

Deliberately not in scope (follow-ups per the issue): treating `{}` as absent at required positions during validation — this fix prevents new husks but does **not** heal config files already carrying one — and making a config-validation failure at boot less of an unrecoverable crash loop.

## Verification

- **Fails-on-base (executed):** the four fix-proving tests were run against the merge base in a detached worktree with only the test diff applied — all four fail there with the husk signatures (`Removed entry must be pruned, not left as {}`, `Vacated subtree must be pruned, not left as a {} skeleton`, `Should delete the vacated cors subtree`) and pass on this branch. Two further tests are guards for the new pruning (they hold trivially on base) plus five regression tests from review findings, including a stacked CONFIG+SET multi-boot scenario and a scalar type-flip that fails on the first iteration of this fix.
- **Unit (executed):** full `unitTests/config` directory green locally at the final code — 233 passing. CI `test:unit:all` green on this exact HEAD on Node 24/26 (v22 initially cancelled, rerun in flight; v22 passed the prior push after one known-flake rerun of `txn-tracking`).
- **Integration:** CI's full 24-shard matrix green on this HEAD (uWS shard 6/6 passed on one rerun after a `registered-operation` suite timeout that also reproduces as flake on main; the same shard passed first-try on the Node and Bun variants). Local sequential integration was not completed: the parallel harness needs the sudo-gated loopback alias pool and long local runs were repeatedly killed by the environment — CI runs the same `test:integration` suites on the same SHA and stands as the integration gate.
- The module-level tests persist and reload real config-state files across simulated boots, which is the observable end-to-end route for this module; the write-before-validate boot ordering that persists the husk is traced in the issue comment rather than reproduced in a test, since the fix makes the composed config valid before it is written.

## For the human reviewer

1. **Compensating marker pair vs first-class empty scopes.** The module now has two mechanisms for "a file-declared `{}` is user content": `restoreBaseEmptyObjects` on the stateless compose path (pre-existing) and the `recordEmptyAncestorOriginal`/`restorePrunedEmptyAncestor` pair on the stateful removal path (this PR). The alternative is making declared-empty scopes first-class in flatten/composition. I chose the pair for blast radius — first-class empties touch both paths and every layer rule. Reversible later by unifying; DESIGN.md documents the coexistence. A "no" here reshapes the PR substantially.
2. **Prune lives in `deleteNestedValue` itself**, so every caller prunes (including `removeValuesWithSource` for non-config sources) rather than only the two env-removal sites. All callers are removal-of-owned-values paths and husks are unwanted everywhere; trivially reversible by moving the loop if you want it narrower.
3. **Absent vs `{}` after vacating is an observable contract change.** A vacated subtree is now genuinely absent from the file where it used to remain as `cors: {}`. Anything reading the file or defaulting on key-presence sees the difference. I believe absent is the intended semantics (it matches `flattenObject`'s env-layer contract); flipping back is cheap if you rule the husk was the safer contract.
4. **One-time upgrade window:** a declared-empty scope populated under released code has no marker (the information was never recorded), so if it is vacated after upgrading it prunes away entirely instead of leaving the husk. Not recoverable from recorded state; the alternative — keeping husks — reintroduces #2067.
5. **Pre-existing wart, not introduced here (review round 2):** a same-source object→scalar transition (`SET={a:{b:{c:1}}}` then `SET={a:5}`) records env-written children as the path's "original" and resurrects them when the var is removed. Reproduces on `main`; happy to file a follow-up issue.

**Post-approval changes** (implementing kriszyp's four review suggestions, each replied-to on its thread): own-property guards in both the prune and the marker-record walks; stale-marker invalidation keyed on true scope absence (with a restart-under-an-occluding-scalar regression test — the review loop caught and fixed an over-clearing first cut); the third deletion site made symmetric; the upgrade-window/corrupt-state limitation documented in code and DESIGN.md, with `saveConfigState` now writing through `configUtils.atomicWriteFile` so a torn write can't reset restoration state. Config suite: 237 passing.

Generated by Claude (Fable 5); cross-model review: Codex graded leg across five rounds, Harper-domain adjudication in round 1, Gemini contributing from round 3 (earlier rounds produced no Gemini output). Reviewed through `55d9aaea5`; the only later commit removes one comment.

<sub>Human-Review-Need: 3 @ e88692607b64</sub>


config/harperConfigEnvVars.ts
```diff
@@ -562,9 +622,15 @@ function applyConfigLayer(
 		}
 
 		// Store original value if requested and this is first time overriding
-		if (storeOriginals && !currentSource && currentValue !== undefined && currentValue !== null) {
-			if (!(path in state.originalValues)) {
-				state.originalValues[path] = currentValue;
+		if (storeOriginals && !currentSource) {
+			if (currentValue !== undefined && currentValue !== null) {
```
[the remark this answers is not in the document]
heskew: Adopted in b346c49ec — the restructured check now uses the house `!= null` idiom. (The parallel pre-existing check in the DEFAULT runtime branch is untouched to keep the diff minimal.)

config/harperConfigEnvVars.ts
```diff
@@ -410,21 +414,76 @@ function setNestedValue(obj: ConfigObject, path: string, value: any): void {
...
-			return; // Path doesn't exist
+			return []; // Path doesn't exist
 		}
+		ancestors.push(current);
 		current = current[key];
 	}
 
-	delete current[keys[keys.length - 1]];
+	const leafKey = keys[keys.length - 1];
+	if (!(leafKey in current)) {
```
kriszyp: `in` walks the prototype chain, so this guard can pass for a key that isn't actually present on `current`. If a tracked path's last segment collides with an `Object.prototype` member — `constructor`, `toString`, `valueOf`, `hasOwnProperty` — then `leafKey in current` is `true`, the `delete` on the next line is a no-op (it only removes own properties), and the prune loop still runs. When `current` is already `{}`, `Object.keys(current).length === 0` holds immediately and the loop deletes the ancestors — which is precisely the "deliberate empty scope eaten by a no-op removal" case this guard exists to prevent, and which the test at `harperConfigEnvVars-config.test.js` ("does not prune a deliberate empty object when removing an already-absent key") asserts against.

Admittedly a config key named `constructor` is unlikely, but the fix is free and the file already uses the safe idiom twice — `flattenObject` at line 328 and `restoreBaseEmptyObjects` at line 362 both use `Object.prototype.hasOwnProperty.call(...)`. Suggest matching them:

```ts
if (!Object.prototype.hasOwnProperty.call(current, leafKey)) {
	return []; // Absent leaf: nothing deleted, nothing to prune
}
```

config/harperConfigEnvVars.ts
```diff
@@ -410,21 +414,76 @@ function setNestedValue(obj: ConfigObject, path: string, value: any): void {
...
+ * emptyScopeOriginals before a layer populates it: a bare `name: {}` in the config
+ * file is user content (#1618/#1726) even while an env var temporarily fills it. At
+ * most one ancestor can be both existing and empty, so one marker suffices.
+ */
+function recordEmptyAncestorOriginal(fileConfig: ConfigObject, state: ConfigState, path: string): void {
+	const keys = path.split('.');
+	let current = fileConfig;
+	for (let i = 0; i < keys.length - 1; i++) {
+		current = current[keys[i]];
+		if (!isPlainObject(current)) return;
```
kriszyp, line 483: A marker is written here but never invalidated, so it can outlive the file content that justified it and later resurrect a scope the user deliberately removed.

Concrete sequence:
1. File declares `myComponent: {}`; `HARPER_CONFIG` sets `myComponent.port` → marker `myComponent` recorded.
2. `HARPER_CONFIG` is removed; the leaf is deleted, `myComponent` is pruned, the marker is consumed and `myComponent: {}` is restored. Fine so far — but replace step 2 with: the user hand-edits the file to delete the whole `myComponent` section while the env var is still set. Nothing consumes the marker, and drift detection at line 587 only covers `HARPER_DEFAULT_CONFIG`/`HARPER_SET_CONFIG`, so the CONFIG layer just re-asserts `myComponent.port` on the next boot.
3. Later `HARPER_CONFIG` is removed. The prune emits `myComponent`, the stale marker still matches, and `myComponent: {}` reappears in the file — a scope the user explicitly deleted.

The walk here is the natural place to keep markers in sync, since it already visits every ancestor at populate time. When the walk hits a prefix that doesn't exist as a plain object, that prefix cannot be a live file-declared empty scope, so any marker for it is stale:

```ts
current = current[keys[i]];
if (!isPlainObject(current)) {
	delete state.emptyScopeOriginals[keys.slice(0, i + 1).join('.')];
	return;
}
```

Note this must key off *absence*, not emptiness: within a single apply, the first leaf under `a: {}` records the marker and populates `a`, so the second leaf's walk sees `a` as non-empty — clearing markers for merely-non-empty ancestors would defeat the feature and break the multi-leaf case. Absence is only reachable on the first leaf under a path, where a legitimate marker cannot exist.

The remaining leak (marker kept alive because the *user* added real content under the scope) is benign — consuming it later still restores something the file once declared — so absence-only clearing is enough.

config/harperConfigEnvVars.ts
```diff
@@ -474,16 +533,20 @@ function loadConfigState(rootPath: string): ConfigState {
...
 	}
 
 	try {
 		const state = fs.readJsonSync(statePath) as ConfigState;
-		// Ensure originalValues exists (for backwards compatibility with old state files)
+		// Ensure newer fields exist (for backwards compatibility with old state files)
 		if (!state.originalValues) {
 			state.originalValues = {};
 		}
+		if (!state.emptyScopeOriginals) {
```
kriszyp, line 565: This defaults the missing *field*, but the missing *information* has no fallback, and that's an upgrade-path regression worth at least a comment here (or a line in the DESIGN.md entry).

On any existing install where an env layer already populated a file-declared `componentName: {}` on a pre-upgrade boot, no marker was ever recorded — markers can only be written at first-population time, which has already passed. After upgrading, the first removal of that env var reaches the new prune in `deleteNestedValue`, `restorePrunedEmptyAncestor` finds no marker, and the user's declared empty scope is silently deleted from the config file. That's exactly the #1618/#1726 regression this PR is otherwise careful to avoid, and it is unrecoverable post-hoc because the scope is populated at the moment we'd want to observe it.

The same hole opens on the `catch` path at line 551: a corrupt state file resets to fresh, dropping `emptyScopeOriginals` (and `originalValues`) for *every* tracked path, so the next removal prunes those scopes for good. Given `saveConfigState` writes non-atomically via `fs.writeJsonSync` and `initConfig` runs per worker, a torn write is a plausible way to get there.

I don't think either needs to block this PR, but the blast radius is user config-file content, so please record the limitation where the next reader will find it.

config/harperConfigEnvVars.ts
```diff
@@ -605,7 +673,7 @@ function handleDeletions(
 				delete state.originalValues[path];
 			} else {
 				// For other sources or if no original value, delete
-				deleteNestedValue(fileConfig, path);
+				restorePrunedEmptyAncestor(fileConfig, state, deleteNestedValue(fileConfig, path));
```
kriszyp, line 700: This site and `cleanupRemovedEnvVar` (line 827) were both converted to feed the pruned paths into `restorePrunedEmptyAncestor`, but the third deletion site — `removeValuesWithSource` at line 690 — still calls `deleteNestedValue(fileConfig, path)` and discards the return value. It now prunes (new behavior) without ever restoring a marker, so it can delete a file-declared empty scope with no way back.

It's currently unreachable: the `else` branch in `cleanupRemovedEnvVar` only fires for a `sourceName` outside the three env vars, and all callers at lines 958-964 pass one of the three. So this is a latent trap rather than a live bug — but the next person to route a source through that path inherits a silent inconsistency between the three removal sites.

Either make it symmetric (`restorePrunedEmptyAncestor(fileConfig, state, deleteNestedValue(fileConfig, path))`, which is a safe no-op when no marker matches) or add a one-line comment on `removeValuesWithSource` stating that it intentionally does not participate in empty-scope restoration and why.

kriszyp (approved):
Thanks for getting this in, a few minor suggestions.

PR #778, opened 2026-05-25
test(integration): port Northwind test suite to self-contained framework

kriszyp:
## Summary

- Migrates 7 legacy Northwind-dependent test files (`2_dataLoad.mjs`, `3_sqlTests.mjs`, `4_noSqlTests.mjs`, `5_noSqlRoleTesting.mjs`, `6_sqlRoleTesting.mjs`, `7_jobsAndJobRoleTesting.mjs`, `10_otherRoleTests.mjs`) into a single self-contained `northwind.test.mjs`
- All seven sub-suites share one Harper instance to avoid repeating the expensive CSV load; schemas, tables, and large CSVs are set up once in `before()`
- Shim functions (`csvFileUpload`, `csvDataLoad`, `csvUrlLoad`, `insert`, `searchByHash`, `checkJob`, `checkJobCompleted`) close over `client` to bridge the legacy utility API to the new `createApiClient`/`awaitJob` framework
- JSON fixture data (long-text remarks, dog/breed/owner records, etc.) defined at module level and inserted by the `2. Data Load` sub-suite so those insert tests remain valid

## Test plan

- [ ] `node --check integrationTests/apiTests/northwind.test.mjs` — passes
- [ ] `npm run lint:required` — 0 errors
- [ ] CI integration run against a live Harper instance

🤖 Generated with [Claude Code](https://claude.com/claude-code)

— Claude

PR #2147, opened 2026-08-11
feat(rest): total-count pagination via `Prefer: count=` (Content-Range)

cb1kenobi:
## Summary

**Tracking issue:** #2162

Adds opt-in total-record-count for REST collection queries so a client can paginate ("1-25 of 1,234") without a second round-trip or a custom resource.

A `GET`/`HEAD` on a collection with a `Prefer: count=` header gets the total in RFC 7233-style response headers:

```
GET /Product/?category=software&limit(0,25)
Prefer: count=exact
->
200 OK
Content-Range: items 0-24/1234
Range-Unit: items
Preference-Applied: count=exact
```

- **`count=exact`** — `Table.search` drains the full matched set once, windowing the requested page in the same pass (O(matched) filter evals, O(limit) memory).
- **`count=estimated`** — returns just the page plus a cheap planner/table estimate, no full scan.
- **No default** — without the header nothing is computed and no headers are emitted.
- Status is always `200` (`Content-Range` is informational, not `206`). `HEAD` returns the headers with no body — a cheap "how many match?" pre-flight. The three headers are added to `Access-Control-Expose-Headers` so browsers can read them cross-origin.

## `count=estimated` rides the storage-level range estimator

`count=estimated` funnels through the query planner's `estimateCondition`. Since **[Use storage-level statistical range estimates in the query planner (#2163)](https://github.com/HarperFast/harper/pull/2163)** merged to `main`, that planner now derives range-query estimates from rocksdb-js 2.8.0's statistical range estimator (`estimateCount`) rather than an arbitrary fraction of the table size — so after merging `main`, this PR's `count=estimated` reports a **range-aware** total for `between` / `starts_with` / greater-less / open-range queries for free, with no change needed here.

This branch adds only the small piece the count path needs on top of #2163: it exports `estimatedEntryCount` from `search.ts` (used by the estimated whole-collection path in `Table.ts`), and adds REST-count coverage that a range `count=estimated` produces a valid, range-aware total.

## Guardrails and operator control

- An exact count's tail (past the requested page) is bounded by a row cap **and** a wall-clock budget; on exhaustion the total is reported unavailable (`items x-y/*`) rather than truncating the page.
- Per-mount config `rest: { exactCount: false }` serves `count=exact` as an estimate instead (default enabled) for sensitive/public mounts.

## Review findings addressed

Cross-model (Codex) and Harper-domain review surfaced and fixed: read-transaction release on the count path (`finally`), guardrail no longer truncating the page (+ time budget), estimate corruption by the planner's synthetic `sort` condition, filter-aware estimates (unknown total instead of a misleading / cardinality-disclosing one), a spurious-header guard for single-record responses, valid-range clamping, and CORS append-not-overwrite. A flagged read-buffer-aliasing concern was verified a non-issue (records return owned copies) and is guarded by a regression test.

## Testing

- Resources unit (`unitTests/resources/queryCount*.test.js`): exact/estimated/window/filtered/default-streaming, plus a Bytes read-buffer-safety guard on both storage engines. Adds primary-key and secondary-index **range** `count=estimated` cases proving the total is range-aware (a narrow tail estimates fewer than the whole table/index) via the #2163 estimator.
- REST integration (`integrationTests/apiTests/rest.test.mjs`): Content-Range/Range-Unit/Preference-Applied, offset window, filtered, estimated, unavailable-total (`/*`), opt-in, HEAD, and the `exactCount` gate (its own instance).
- Full resources unit suite green (1882 passing) on the merge with `main`; `tsc` and lint clean.

## Docs

REST reference docs (Pagination and Total Count, the `exactCount` option, the `Prefer` header): **HarperFast/documentation#623**.


resources/Table.ts
```diff
@@ -3487,12 +3499,73 @@ export function makeTable(options) {
...
+							if (end !== undefined && scanned >= end) {
+								if (!wantExact) break; // `estimated` needs nothing past the page
+								// `exact` keeps counting the tail, bounded by a row cap AND a time budget so a
+								// large match set can't turn a bounded page fetch into an unbounded scan.
+								if (scanned > MAX_EXACT_COUNT_SCAN || performance.now() - countStart > MAX_EXACT_COUNT_MS) {
+									exact = false;
+									break;
+								}
+							}
+						}
```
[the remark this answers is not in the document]
cb1kenobi: Good catch — fixed in cac3ae224. Counting now requires a `limit()`: a count request without one falls through to the normal streaming path (no count emitted), so the guardrail always applies to a bounded page rather than being skipped when `end` is undefined. Updated the unit test that documented the no-limit drain as intentional and added one asserting the no-limit case streams instead of materializing.

— Claude Opus 4.8

resources/Table.ts
```diff
@@ -3487,12 +3499,80 @@ export function makeTable(options) {
...
+			// count so the HTTP layer can emit a Content-Range. `exact` drains the full matched set once,
+			// windowing the page in the same pass; `estimated` returns just the page plus a cheap planner/
+			// table estimate. Opt-in only — the default streaming path below is untouched.
+			//
+			// Requires a limit. Counting is a pagination feature, and with no limit the "page" is the
+			// entire matched set — materializing/draining it would be an unbounded operation on the most
+			// likely-hit path (a bare collection GET). A count request without a limit therefore falls
+			// through to the normal streaming path (no count emitted), so the guardrail below always
+			// applies to a bounded page rather than being skipped when `end` is undefined.
+			if (target.count && target.limit !== undefined) {
```
kriszyp: The new check closes the `undefined`-limit case, but it does not establish that the page is actually bounded. REST parsing assigns `+args[0]` at `resources/search.ts:1370-1377`, so `limit(Infinity)` makes `pageEnd` infinite and `scanned >= pageEnd` never becomes true; a huge finite limit similarly materializes the whole match set before any guard applies. `limit(foo)` produces `NaN`, causing both page and guard comparisons to remain false while exact counting drains the full set. Please validate a finite, non-negative limit and enforce a maximum count-page size before entering this materializing branch, with tests for non-finite and oversized values.

— KrAIs (GPT-5)

server/REST.ts
```diff
@@ -165,6 +192,23 @@ async function http(request: Request, nextHandler, resources: Resources, httpOpt
...
 			(target as any).async = true;
 			resource = entry.Resource;
+			// Pagination total-count opt-in (no default): `Prefer: count=exact|estimated`. Table.search
+			// reads target.count to compute the total emitted as Content-Range below.
+			const prefer = headersObject['prefer'];
+			if (prefer) {
+				// A mount can disable the expensive exact scan with `rest: { exactCount: false }` (default
+				// enabled); a count=exact request is then served as a cheap estimate. Accept a string
+				// `"false"` too, since not every config source coerces to a boolean.
+				const exactDisabled = (httpOptions as any).exactCount === false || (httpOptions as any).exactCount === 'false';
```
kriszyp: Exact counting is effectively enabled by default here. A request such as `GET ...?limit(1)` with `Prefer: count=exact` returns one row but then drains up to one million matches or one second in `resources/Table.ts:3528-3535`. Those limits are per request, so concurrent requests can each consume the full budget and hold read snapshots, making a public exported table an inexpensive worker-saturation path. Please make exact counting opt-in (`exactCount: true`) or enforce a shared concurrency/cancellation budget; estimated counting can remain the safe default.

— KrAIs (GPT-5)

resources/Table.ts
```diff
@@ -3487,12 +3499,80 @@ export function makeTable(options) {
...
+				const wantExact = target.count === 'exact';
+				const pageEnd = offset + (target.limit as number);
+				const countStart = performance.now();
+				return (async () => {
```
kriszyp, line 3768: This conditionally changes the declared `search()` contract from `AsyncIterable` (`resources/Table.ts:3098`) to `Promise<Array>`. That still breaks an internal path: REST attaches `target.count` for every HTTP method, and collection deletion immediately does `for await (const entry of this.search(scanTarget))` at `resources/Table.ts:2997-3003`; a DELETE carrying both `limit(...)` and `Prefer: count=exact` therefore receives a Promise and throws instead of deleting. Please restrict the REST preference to GET/HEAD and either preserve the iterable contract or expose counting through an explicit, accurately typed API/overload with all callers updated.

— KrAIs (GPT-5)

server/REST.ts
```diff
@@ -117,6 +117,33 @@ async function findInactiveComponent(url: string): Promise<string | undefined> {
...
+ * All three headers are added to `Access-Control-Expose-Headers` so a browser can read them cross-origin
+ * (they aren't safelisted).
+ */
+function setCountHeaders(headers: Headers, offset: number, mode: string, page: any) {
+	const total = page.recordCount;
+	const len = Array.isArray(page) ? page.length : 0;
+	const range = len > 0 ? `${offset}-${offset + len - 1}` : '*';
+	const totalStr = typeof total === 'number' ? String(total) : '*';
+	headers.set('Range-Unit', 'items');
+	headers.set('Content-Range', `items ${range}/${totalStr}`);
```
kriszyp, line 136: These response headers depend on the request's `Prefer` value, but the URL and body do not, and no final `Vary: Prefer` is emitted. A shared cache can therefore serve count headers to a request without the preference, or serve a cached non-count response to a client that requested totals. Please add `Prefer` to the final `Vary` value after serialization—`serialize()` later replaces `Vary` at `server/REST.ts:406`, so adding it only inside this helper would still be lost.

— KrAIs (GPT-5)

server/REST.ts
```diff
@@ -117,6 +117,33 @@ async function findInactiveComponent(url: string): Promise<string | undefined> {
...
+	headers.set('Preference-Applied', `count=${mode}`);
+	// Append (don't overwrite) so a resource that already exposed its own headers keeps them.
+	const exposed = headers.get('Access-Control-Expose-Headers');
+	for (const name of ['Content-Range', 'Range-Unit', 'Preference-Applied']) {
+		if (!exposed || !String(exposed).toLowerCase().includes(name.toLowerCase())) {
```
kriszyp: Please compare comma-delimited exposure names as case-insensitive tokens rather than substrings. For example, an existing `Access-Control-Expose-Headers: X-Content-Range-Metadata` currently suppresses the actual `Content-Range` token, leaving it unreadable to browser clients. Splitting and trimming the existing value, as the `Vary` helper does, avoids that collision.

— KrAIs (GPT-5)

dawsontoth (approved):

kriszyp (changes requested):
This is a really cool idea, I like the interface.
However, I agree with the comments; I think we should probably have exactCount disabled by default, it could pretty expensive. I also think this really increases the need for https://github.com/HarperFast/rocksdb-js/pull/311 or some better solution for estimating range counts (maybe should be a prerequisite for this?)
And do we have an issue for this? I'd like to get priorities associated with these (especially if this starts entraining dependency PRs).

cb1kenobi:
Thanks @kriszyp — really helpful review, all five points were spot on. Addressed in 5ec0d38a9:

**Design**
- **Exact counting is now opt-in** (`rest: { exactCount: true }`, default off); `count=exact` is otherwise served as an estimate. Estimated stays the safe default, so a public table no longer exposes the exact-scan budget by default.
- **Tracking issue:** #2162 — links rocksdb-js#311 as the range-estimation dependency, and flags there the "should better estimation be a prerequisite for broader rollout?" question for prioritization.

**Fixes**
1. **Unbounded/invalid limit** — the count path now requires a finite, non-negative integer limit no larger than a max page size (10k); `limit(Infinity)`, `limit(foo)`→NaN, negative, or oversized limits fall through to streaming with no count.
2. **`search()` contract / DELETE** — `Prefer: count` is now GET/HEAD-only, so a collection DELETE carrying `limit()`+`Prefer` keeps the AsyncIterable and deletes instead of receiving a Promise and throwing.
3. **`Vary: Prefer`** — emitted on collection reads after serialization (which resets `Vary`), so shared caches key on `Prefer`.
4. **CORS token comparison** — expose-header dedup now compares case-insensitive comma tokens instead of substrings, so `X-Content-Range-Metadata` no longer suppresses `Content-Range`.

Docs updated in HarperFast/documentation#623. Tests: unit 11, integration 27 green. Ready for another look when you have a moment.

— Claude Opus 4.8

resources/Table.ts
```diff
@@ -3487,12 +3503,81 @@ export function makeTable(options) {
...
+			// windowing the page in the same pass; `estimated` returns just the page plus a cheap planner/
+			// table estimate. Opt-in only — the default streaming path below is untouched.
+			//
+			// Requires a bounded page. Counting is a pagination feature; the page limit must be a finite,
+			// non-negative integer no larger than MAX_COUNT_PAGE. A missing limit (a bare collection GET), a
+			// non-finite/negative/non-integer limit (limit(Infinity), limit(foo)), or an oversized one all
+			// fall through to the normal streaming path with no count, so a count request can't materialize
+			// an unbounded page before the guardrail below applies.
+			const pageLimit = target.limit as number;
+			if (target.count && Number.isInteger(pageLimit) && pageLimit >= 0 && pageLimit <= MAX_COUNT_PAGE) {
```
kriszyp: The bounded-page check still accepts any `target.offset`, even though REST parses the first argument of `limit(start,end)` with unchecked numeric coercion. A negative offset changes behavior between the normal `results.slice(offset, end)` path and this manual `scanned >= offset && scanned < pageEnd` window, while an arbitrarily large offset postpones both exact-count guardrails until that offset has been scanned because the checks begin only at `scanned >= pageEnd`.

Please require a finite, non-negative integer offset (and consider a maximum scan/window bound), or reject the request during query parsing. Add regression coverage for negative, non-finite, and very large two-argument limit windows.

— KrAIs (GPT-5)

kriszyp:
Heads-up for the estimation side of this PR: [rocksdb-js #778 — feat: statistical range key-count estimation](https://github.com/HarperFast/rocksdb-js/pull/778) adds storage-level range estimates that this PR should use once it lands:

- **`db.estimateCount({ start, end })` → `{ count, confidence }`** — a no-iteration statistical range count (memtable stats + range-local SST entry density), measured within ~±5% at 9–83µs where exact scans took 1–85ms. This can replace the planner-fraction path behind `count=estimated` (`estimateCondition`'s hardcoded 5%/10%/30% guesses for range comparators) with a real per-range number, and `confidence` gives the REST layer a principled basis for deciding when an estimate is trustworthy enough to report.
- **`db.createCountEstimator(range)`** — rides the iteration this PR already does: `advance(lastKey, n)` per page, `estimate()` = exact-so-far + calibrated remainder, `finish()` = exact. This composes nicely with the `count=exact` guardrail here: when the row-cap/time budget exhausts mid-drain, the partial drain can become a calibrated estimate (with confidence) instead of reporting the total unavailable (`items x-y/*`).

Harper-side integration is coming as a follow-up PR (planner `estimateCondition` + `estimatedEntryCount`, feature-detected so it degrades cleanly on older rocksdb-js). Nothing blocking here — this PR's shape is already right for it.

— Claude (Fable 5)

kriszyp (commented):
Will post the queued inline comment, but more importantly this probably should switch to use iterator estimator in 778, as noted in the other comment.

resources/Table.ts
cb1kenobi: Good catch — fixed in 57caa5ec3. The count path now also requires the offset to be a finite, non-negative integer and the window (offset + limit) to be within the scan budget (MAX_EXACT_COUNT_SCAN); a negative offset or a deep-page window past the budget falls through to streaming with no count, so a large offset can't postpone the guardrail. Added unit + integration coverage for negative and oversized-window offsets.

— Claude Opus 4.8

resources/Table.ts
```diff
@@ -3513,12 +3529,91 @@ export function makeTable(options) {
...
+								}
+							}
+						}
+					} finally {
+						// We own the iteration here (no results.onDone consumer), so release the read
+						// transaction unconditionally — including when the drain throws — or the snapshot leaks.
+						txn.doneReadTxn();
+					}
+					let total: number | null;
+					if (wantExact) {
```
kriszyp, line 3801: `scanned` is not an exact total when `results` comes from an HNSW vector sort. HNSW intentionally returns only a bounded approximate candidate set, and after the merged `minResults` work its candidate count can depend on `offset + limit`. For the same vector query, `limit(5)` can therefore advertise (for example) `.../118`, while `limit(200)` widens `ef` and advertises `.../200`; both are marked exact even though the table and predicates are unchanged.

Please avoid deriving an exact total from an approximate-index result iterator. Either count through a separate non-candidate-limited path with clearly defined vector/missing-vector semantics, or report the total as unavailable/downgrade the applied mode for approximate searches. A regression test should run the same vector query with two page sizes or offsets and verify that an advertised exact total is authoritative and stable.

— KrAIs (GPT-5)

kriszyp (changes requested):
I think we just need to publish a new rocksdb-js version to unblock this so you can switch to the newer estimation APIs.

resources/Table.ts
```diff
@@ -3650,12 +3666,91 @@ export function makeTable(options) {
...
+						total = exact ? scanned : null;
```
kriszyp: `scanned` is not an exact total when `results` comes from an HNSW vector sort. HNSW intentionally returns only a bounded approximate candidate set, and after the merged `minResults` work its candidate count can depend on `offset + limit`. For the same vector query, `limit(5)` can therefore advertise (for example) `.../118`, while `limit(200)` widens `ef` and advertises `.../200`; both are marked exact even though the table and predicates are unchanged.

Please avoid deriving an exact total from an approximate-index result iterator. Either count through a separate non-candidate-limited path with clearly defined vector/missing-vector semantics, or report the total as unavailable/downgrade the applied mode for approximate searches. A regression test should run the same vector query with two page sizes or offsets and verify that an advertised exact total is authoritative and stable.

— KrAIs (GPT-5)

kriszyp (changes requested):
This is still not using the estimation functionality in rocksdb 2.8.

resources/Table.ts
```diff
@@ -3681,12 +3697,91 @@ export function makeTable(options) {
...
```
kriszyp: `scanned` is not an exact total when `results` comes from an HNSW vector sort. HNSW intentionally returns only a bounded approximate candidate set, and after the merged `minResults` work its candidate count can depend on `offset + limit`. For the same vector query, `limit(5)` can therefore advertise (for example) `.../118`, while `limit(200)` widens `ef` and advertises `.../200`; both are marked exact even though the table and predicates are unchanged.

Please avoid deriving an exact total from an approximate-index result iterator. Either count through a separate non-candidate-limited path with clearly defined vector/missing-vector semantics, or report the total as unavailable/downgrade the applied mode for approximate searches. A regression test should run the same vector query with two page sizes or offsets and verify that an advertised exact total is authoritative and stable.

— KrAIs (GPT-5)

resources/search.ts
```diff
@@ -1112,6 +1112,76 @@ export function filterByType(searchCondition, Table, context, filtered, isPrimar
...
+// zero-confidence estimate (a failed statistics read) — so the caller falls back to a heuristic fraction.
+function estimateRangeCount(table, condition, comparator) {
+	const attribute_name = condition[0] ?? condition.attribute;
+	const isPrimaryKey = attribute_name == null || attribute_name === table.primaryKey;
+	const store = isPrimaryKey ? table.primaryStore : table.indices[attribute_name];
+	// estimateCount is a RocksDB-only capability. A custom index's store is a plain object store keyed by
+	// primary key (not [indexedValue, primaryKey]), so a value-range estimate over it would be meaningless —
+	// leave those (and LMDB stores) to the heuristic / the custom index's own estimation.
+	if (!(store instanceof RocksDatabase) || (store as any).customIndex) return undefined;
+	const bounds = rangeBoundsForEstimate(comparator, condition[1] ?? condition.value);
```
kriszyp: This estimates the positive bounded range even when `condition.negated` is true. `searchByIndex` executes `not_between`/`not_starts_with` by forcing a primary-store full scan and filtering the complement, including records absent from the secondary index. A one-row exclusion in a million-row table is therefore estimated near 1, selected as the most selective lead, and then scans the whole table; `count=estimated` reports the excluded count (or the page-size clamp) for roughly 999,999 returned rows.

The helper also bypasses `searchByIndex`’s `MAX_SEARCH_KEY_LENGTH` overflow normalization, so strings longer than 256 characters estimate a different key range from the scan. Please share the execution bound-normalization path, or explicitly use a complement/conservative estimate for negated predicates and apply overflow normalization before `estimateCount`. Regression coverage should include a narrow negated exclusion and an oversized indexed string.

— KrAIs (GPT-5)

kriszyp (changes requested):
Can you rebase/stack this onto #2163, which owns the implementation of using estimation for query planning, deleting rangeBoundsForEstimate / estimateRangeCount / the RocksIndexStore override here, along with the dep bump. #2147 then keeps what's genuinely its own — Prefer: count=, the guardrails, the Content-Range headers, the page materialization — and the estimateCondition() call inherits the corrected estimates for free.

rocksdb-js 2.8.0 also ships CountEstimator (db.createCountEstimator()), which progressively refines a range estimate from entries already iterated. The count=estimated path is a natural fit — it walks offset + limit entries of the range and then discards that knowledge for a pure statistical estimate. Feeding it advance(lastKey, n) per page makes the traversed portion exact and calibrates the remainder, and finish() on natural exhaustion returns confidence: 1 — an exact total for free whenever the match set fits inside the guardrail, which is most paginated queries. That would also subsume a good chunk of the count=exact machinery. We could potentially do this as a follow-up, but I think it makes more sense here. 
 
resources/Table.ts
cb1kenobi: Fixed in 9e44500d7. The count path now detects an approximate result set — a vector/HNSW-driven sort (`indices[sortAttr].customIndex`) or a `vectorFilter` — and reports the total as **unavailable** (`recordCount: null` → `Content-Range: items x-y/*`, `recordCountExact: false`) instead of advertising `scanned` (which tracks the requested page size via `minResults`, not the true match count) as `count=exact`. The page itself still materializes normally.

Regression test `unitTests/resources/queryCountVector.test.js` runs the same cosine-sorted query at `limit(5)` and `limit(40)` and asserts the total is unavailable at both. Thanks for keeping this one visible across the rounds — it was the right call.

<sub>🤖 Drafted with Claude Code (Opus 4.8)</sub>

resources/Table.ts
cb1kenobi: Addressed — the count path now requires a finite, non-negative integer limit no larger than `MAX_COUNT_PAGE` (`Number.isInteger(pageLimit) && pageLimit >= 0 && pageLimit <= MAX_COUNT_PAGE`), and bounds the window (`offset + pageLimit <= MAX_EXACT_COUNT_SCAN`). So `limit(Infinity)`/`limit(foo)` fall through to the normal streaming path with no count.

<sub>🤖 Drafted with Claude Code (Opus 4.8)</sub>

server/REST.ts
cb1kenobi: Addressed — exact counting is opt-in per mount via `rest: { exactCount: true }` (default off); `count=exact` on a mount that hasn't opted in is served as an estimate and signaled back through `Preference-Applied`. So the per-request scan budget isn't exposed by default.

<sub>🤖 Drafted with Claude Code (Opus 4.8)</sub>

resources/Table.ts
cb1kenobi: Addressed — the `Prefer: count=` preference is only honored on GET/HEAD (`server/REST.ts:206`), so a DELETE (or any other method) never gets `target.count` set and `search()` keeps its `AsyncIterable` contract on the collection-deletion path. The count headers are likewise gated to GET/HEAD (`server/REST.ts:406`). Good call flagging the internal caller.

<sub>🤖 Drafted with Claude Code (Opus 4.8)</sub>

server/REST.ts
cb1kenobi: Addressed — `Vary: Prefer` is now added *after* serialization (`server/REST.ts:419-422`), since `serialize()` resets `Vary`, so a shared cache can't cross-serve count/non-count responses. Thanks for catching the reset-timing subtlety.

<sub>🤖 Drafted with Claude Code (Opus 4.8)</sub>

server/REST.ts
cb1kenobi: Addressed — the three headers are appended as case-insensitive comma-delimited tokens (`server/REST.ts:140-149`): existing exposure names are split and lowercased into a Set, and each name is appended only if its token isn't already present. An existing `X-Content-Range-Metadata` no longer suppresses the `Content-Range` token.

<sub>🤖 Drafted with Claude Code (Opus 4.8)</sub>

resources/search.ts
cb1kenobi: Resolved by rebasing onto #2163 — this PR no longer carries `rangeBoundsForEstimate`/`estimateRangeCount`. Range estimation now comes from #2163's `estimateRangeCondition`, which short-circuits a negated condition to `Infinity` (it executes as a filtered full scan), so a positive-range estimate can't win the driving-condition ordering.

<sub>🤖 Drafted with Claude Code (Opus 4.8)</sub>

cb1kenobi:
@kriszyp — done, and thanks for the detailed steer through this. Summary of where things landed:

**Rebased onto #2163.** This branch is merged onto `main`, which now carries #2163 (the query-planner estimation work). I deleted `rangeBoundsForEstimate` / `estimateRangeCount` / the `RocksIndexStore.estimateCount` override and the rocksdb-js dep bump from this PR, so #2147 keeps only what's genuinely its own — `Prefer: count=`, the guardrails, the `Content-Range` headers, and page materialization — and `estimateCondition()` inherits #2163's storage-level range estimates for `count=estimated`. Net diff vs `main` in `search.ts` is now a single line (re-exporting `estimatedEntryCount`, which the count path imports).

**HNSW/vector `count=exact` blocker fixed** (9e44500d7). An approximate result set — a vector/HNSW-driven sort or a `vectorFilter` — now reports its total as unavailable (`items x-y/*`) rather than advertising a page-size-dependent `scanned` as exact. Regression test runs the same cosine-sorted query at two page sizes and asserts the total is unavailable at both.

**The remaining inline threads** (GET/HEAD-only preference, `Vary: Prefer`, CORS token comparison, `limit(Infinity)`/offset guards, exact-count opt-in) were addressed in earlier commits — I replied inline on each with the specifics and resolved them.

Full resources unit suite is green (1883). Ready for another look when you have a moment!

<sub>🤖 Drafted with Claude Code (Opus 4.8)</sub>

resources/Table.ts
```diff
@@ -3702,12 +3718,103 @@ export function makeTable(options) {
...
+				const wantExact = target.count === 'exact';
+				const pageEnd = offset + pageLimit;
+				const countStart = performance.now();
+				// A vector/HNSW-driven result set — sorted by a custom index, or shaped by a vector filter —
+				// is a bounded, approximate candidate set whose size is chosen from `minResults` (offset +
+				// limit), so `scanned` tracks the requested page size, not the true match count. It is not an
+				// authoritative total: the same query at limit(5) vs limit(200) would otherwise advertise two
+				// different `count=exact` totals. Report the total as unavailable instead (mirroring how the
+				// estimated branch below already bails to null for a vector/row filter).
+				let approximateResultSet = typeof target.vectorFilter === 'function';
```
kriszyp: `approximateResultSet` only examines `target.vectorFilter` and the sort chain, but HNSW is also directly queryable through `lt`/`le` conditions. For example, `conditions: [{ attribute: 'vector', comparator: 'lt', value: threshold, target }]` has neither marker, so this remains false even though `executeConditions()` passes the page-derived `minResults` into the same bounded HNSW candidate search.

When that candidate iterator exhausts, `exact` remains true and line 3787 still publishes `scanned` as an authoritative total. Please include HNSW-backed planned conditions in the approximate-result classification and add a threshold-condition regression alongside the vector-sort test.

— KrAIs (GPT-5)

resources/Table.ts
```diff
@@ -3702,12 +3718,103 @@ export function makeTable(options) {
...
+				for (let order = sort; !approximateResultSet && order; order = order.next) {
+					if (typeof order.attribute === 'string' && indices[order.attribute]?.customIndex) approximateResultSet = true;
```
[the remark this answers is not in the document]
cb1kenobi: Good catch — fixed in 088e8c141. The approximate-result detection now walks the executing `conditions` (recursively, through OR groups) for any custom-index-backed attribute, not just the `sort` chain, so an HNSW `lt`/`le` threshold filter with no `sort` is caught too. It's also more precise than the sort-walk: a vector sort applied as in-memory post-ordering leaves no custom-index condition in `conditions`, so it correctly stays exact. Regression test extended with the `lt` threshold case at two page sizes.

<sub>🤖 Drafted with Claude Code (Opus 4.8)</sub>

cb1kenobi:
## Cross-model review (Gemini 3.1 Pro + Cursor-Grok)

Ran the pre-push cross-model review on `d7d4356b6`. The Codex graded leg timed out on a local model-manager error, which starved the domain adjudicator, so findings below are **author-triaged**, not machine-adjudicated. Independent outside-family coverage still came from Gemini and Cursor-Grok. Verified each against the code before acting.

### Fixed (pushed in `d7d4356b6`)

- **Event-loop starvation on a large `count=exact` scan** (Gemini major / Grok significant). The drain iterated the store's async iterator with no macrotask yield; on a synchronously-settling iterator (the common indexed-scan case) a big scan ran as one microtask burst, blocking the loop for up to the whole `MAX_EXACT_COUNT_MS`. It now yields to the macrotask queue every `COUNT_YIELD_INTERVAL` rows — covering the page-window scan too, not just the tail.
- **Approximate exact-count wasted the tail** (Grok suggestion). For a vector/HNSW result set, `count=exact` now stops at the page window instead of draining a tail whose total is reported unavailable anyway.
- **Prefer-parse hardening** (Gemini minors). Optional-chain `httpOptions` (a programmatic mount may pass none) and `String()`-coerce the preference value before lower-casing (a bare `Prefer: count` with no `=` can yield a non-string).

### Investigated — not a bug

- **"async `allowRead` breaks the count path → 500"** (Grok blocker). Did not reproduce: driving `get()` (what REST calls) with an async `allowRead` and `Prefer: count=` returns a proper page array — `recordCount` intact, iterates cleanly. The Table-level async-authorization branch isn't reached on the count path (authorization resolves at the Resource layer first). No fix shipped for a non-issue.

### Surfaced for a maintainer call (not changed)

These are real but touch product/security defaults, so I'd rather you decide than change them unilaterally:

- **CORS request admission** (Grok significant): the default `Access-Control-Allow-Headers` (`Accept, Content-Type, Authorization`) omits `Prefer`, so a cross-origin browser can't *send* `Prefer` — preflight blocks the request before the new expose-headers path matters. The feature's cross-origin story is only half-wired without it. Options: add `Prefer` to the default allow-list (small, safe — `Prefer` is a standard header), or document that operators must. Happy to implement either.
- **Conditional GET + count** (Grok significant): a matching `If-None-Match` returns 304 and clears the body *after* the count scan already ran, so `Content-Range`/`Preference-Applied` are dropped and the scan was wasted. Edge (needs a prior ETag), but worth a decision: skip the scan on a 304, or emit the count headers on it.
- **Quoted preference token** (Grok suggestion): `Prefer: count="exact"` (RFC-quoted) is silently ignored. Cheap to accept if we want it.
- **Minor** (Gemini): `setCountHeaders` allocates a Set/array per count response (opt-in path, low impact); and the new comments run heavier than Harper's usual density — can trim on request.

Full resources unit suite: 1876 passing; the 8 failures (transaction-log / snapshot / audit / table-reload) reproduce identically on a clean tree — pre-existing and unrelated to this change.

<sub>🤖 Drafted with Claude Code (Opus 4.8)</sub>

kriszyp (approved):
Awesome, great work!

PR #2152, opened 2026-08-12
feat(components): stream package_component instead of base64 JSON

### LINKED ISSUE #2150: package_component: stream the tarball instead of returning it as base64 JSON
`package_component` returns the entire component tarball as a base64 string inside the JSON response body. That materializes the whole archive several times over — on the server, then again in the browser — and **hard-fails above ~384 MiB of compressed output** no matter how much RAM the node has.

Downstream symptom: HarperFast/studio#1591, where downloading an ~800 MB application transfers bytes fine and then kills the Chrome tab.

Everything needed to fix this already exists in this repo. `package_component` is simply on the wrong path.

## What it does today

`components/operations.js:354`:

```js
const payload = (await packageDirectory(pathToProject, req)).toString('base64');
return { project, payload };
```

`packageDirectory()` (`components/packageComponent.ts:192`) collects every gzip chunk into an array and `Buffer.concat`s it. For a tarball of N bytes:

| stage | resident |
| --- | --- |
| chunk array + `Buffer.concat` | 2N |
| `.toString('base64')` | +1.33N |
| Fastify `JSON.stringify` of the envelope | +1.33N |
| **peak** | **≈4.7N** |

All of it in the shared Harper process. On Fabric that process is multi-tenant, so one large download is an availability risk for every other tenant on the node.

And there's a hard ceiling, not just pressure — `.toString('base64')` throws `ERR_STRING_TOO_LONG` once the result would exceed V8's string cap:

```
$ node -e "const b=require('buffer'); console.log(b.constants.MAX_STRING_LENGTH)"
536870888        # 512 MiB → 384 MiB of gzipped tar is the ceiling
```

The browser then pays the same tax again: full JSON response text, a second copy out of `JSON.parse`, a third out of `atob()`, a per-byte `Uint8Array.from(s, c => c.charCodeAt(0))` on the main thread, and a `Blob` copy. That's the renderer OOM in studio#1591 — it's not a transport failure.

## Proposed change: return the packer's stream

Both halves are already built and already in use elsewhere.

- **The packer is already a stream.** `streamPackagedDirectory()` (`components/packageComponent.ts:115`) returns a `Readable` (tar-fs → gzip). Its docstring says it exists to dodge "the Node.js 2GB Buffer cap that the buffered variant runs into for large components." `packageComponent` just never calls it.
- **The response path already pipes streams.** `server/serverHelpers/serverHandlers.js:172` pipes any returned `Readable` carrying a `headers` Map, and honours a `preCompressed` flag so an already-gzipped body isn't re-compressed on the way out.
- **There is a direct precedent.** `handleGetDeploymentPayload()` (`components/deploymentOperations.ts:216`) does exactly this for deployment payloads, and its comment already states the reasoning verbatim: "Never base64 — payloads can be hundreds of MB and a base64 round-trip materializes multi-hundred-MB strings (V8's string cap is ~512MiB…)". `get_backup` uses the same mechanism.

Sketch:

```ts
const stream = streamPackagedDirectory(pathToProject, req);
const headers = new Map<string, string>();
headers.set('content-type', 'application/gzip');
// `project` is normalized through path.parse().name earlier in the handler,
// so it can't inject header content.
headers.set('content-disposition', `attachment; filename="${project}.tar.gz"`);
(stream as any).headers = headers;
(stream as any).preCompressed = true;   // already gzipped; don't recompress
return stream;
```

Memory becomes constant on both ends. Nothing is staged to disk, no temp archive, no retention lifecycle.

### Compatibility

`{ project, payload }` is a documented response shape, so the stream should be opt-in rather than a silent switch — `Accept: application/gzip`, or an explicit request field. Worth settling as part of this issue. Note the remote operation's only real consumer is Studio: the CLI's `package` alias packages *locally* for deploy (`bin/cliOperations.ts:14` imports `streamPackagedDirectory` directly) and never reads a remote `payload`.

`package_component` already requires `super_user` (`utility/operation_authorization.ts:287`), same posture as `get_deployment_payload`, so the streamed variant inherits the right gate.

### Secondary: a size probe

Studio wants to warn before packaging something absurd. It can already do that for the **default** mode — `get_components` returns per-file `size` and skips `node_modules`, which matches `skip_node_modules: true` exactly — but it has no way to size a package *with* `node_modules`.

`scanPackageDirectory()` (`components/packageComponent.ts:54`) already computes precisely this, and `getPackagedDirectorySize()` already wraps it. An `estimate: true` mode returning `{ project, total_size, dangling_symlinks }` would close the gap cheaply. Lower priority than the streaming fix.

## Constraint worth recording: Fabric Connect can't carry this

Central Manager's Fabric Connect path is a message-passing proxy, not an HTTP reverse proxy, and it caps bodies at 2 MB:

> Fabric Connect body size must be less than 2MB. Utilize Direct Connect for larger payloads.
>
> — `central-manager` `src/resources/hdbInstance/HDBInstance.js:383`

So a streamed download only works over a direct (Bearer operation-token) connection — the same split Studio's SSE work already models as `mode: 'direct' | 'proxy'`. Proxy-mode instances need an explicit error rather than a hang or a truncated file.

## Why not the alternatives

**Multi-file download** (one request per file, reassemble client-side) is strictly worse on every axis. The browser can't write a directory tree without `showDirectoryPicker()` (Chromium-only); a real application is thousands of files, each an extra round trip through `get_component_file`'s own base64-in-JSON; and it discards the symlink / mode / empty-dir fidelity that `streamPackagedDirectory` goes out of its way to preserve (`dereference`, the dangling-link skip, the directory-mode `map`). Reassembling into a single artifact means zipping in the browser, which pushes every byte back through JS memory — the exact thing being fixed. The one thing it buys is resumability, and HTTP range requests against a stream would buy that better if we ever need it.

**Staging a zip server-side** writes a second full copy to the node's disk and needs TTL, cleanup, and reclaim machinery. That's the deployment-payload system, which already exists — for the *upload* direction. Streaming needs none of it.

**"Don't do that"** is partly right, and is tracked separately in studio#1591: Studio should surface the size up front and warn, because `scp` genuinely is the better tool for pulling an 800 MB customer application. But that's a guardrail on top. It isn't a reason to leave a 384 MiB hard cap and a ~4.7× memory multiplier sitting in the operations API.

dawsontoth:
Closes #2150.

`package_component` returned the whole component tarball as base64 inside the JSON envelope. This adds a streamed shape, which is what the operation should have been doing all along — the machinery was already in the repo, the handler just wasn't using it.

## The problem

`components/operations.js:354`:

```js
const payload = (await packageDirectory(pathToProject, req)).toString('base64');
return { project, payload };
```

For an archive of N bytes that peaks at roughly **4.7N resident** in the shared (on Fabric, multi-tenant) Harper process: 2N to accumulate and `Buffer.concat` the gzip chunks, +1.33N for the base64 string, +1.33N again when Fastify serializes the envelope.

And it has a hard ceiling, not just pressure — `.toString('base64')` throws `ERR_STRING_TOO_LONG` once the result exceeds V8's string cap:

```
$ node -e "const b=require('buffer'); console.log(b.constants.MAX_STRING_LENGTH)"
536870888        # 512 MiB → ~384 MiB of gzipped tar is the ceiling
```

The browser pays the same tax again, which is the downstream symptom in HarperFast/studio#1591: bytes transfer fine, then Chrome kills the renderer while decoding.

## The change

**`stream: true`** returns the packer's `Readable` with download headers attached. Nothing new had to be built:

- `streamPackagedDirectory()` (`components/packageComponent.ts:115`) already returns a tar-fs → gzip `Readable`, and its docstring says it exists to avoid "the Node.js 2GB Buffer cap that the buffered variant runs into for large components". The handler simply called the buffered variant instead.
- `serverHandlers.js:172` already pipes any returned `Readable` carrying a `headers` Map, and honours `preCompressed` so an already-gzipped body isn't compressed twice.
- `handleGetDeploymentPayload()` and `get_backup` are the existing precedents on that path.

Memory is constant on both ends, nothing is staged to disk, and there is no size ceiling.

**`estimate: true`** returns `{ project, total_size, dangling_symlinks }` from a single `scanPackageDirectory()` walk without packaging anything, so a caller can decide whether the download is worth starting. Studio wants this for the include-node_modules case, where its own `get_components` tree gives it no size to work from.

**The base64 shape stays the default**, so every existing caller is untouched. The remote operation's only real consumer is Studio — the CLI's `package` alias packages locally for deploy (`bin/cliOperations.ts:14`) and never reads a remote `payload`.

## Notes for review

- `estimate` is checked before `stream` and wins if both are set. They are deliberately **not** declared as Joi exclusive peers: `oxor` conflicts on *presence*, so it would reject a caller that always sends both fields with one set to `false`.
- `content-disposition` interpolates `project`, which is already narrowed by `path.parse().name` and validated against `PROJECT_FILE_NAME_REGEX` before it gets there, so it can't inject header content.
- The project-resolution guard is a drive-by fix with a real reason to be in this PR: the node_modules fallback's `catch` only rethrows ENOENT, so any other realpath failure left `pathToProject` undefined and walked it anyway. That was survivable when the response was buffered; on the stream path headers are committed before the walk runs, so it would surface as a 200 with a truncated body. Its test drives a real ENOTDIR (a rootPath whose `node_modules` is a file) and fails without the guard.
- A mid-walk failure after headers are committed still truncates the response — inherent to streaming, and the same behaviour `get_backup`/`get_deployment_payload` already have.
- **Fabric Connect can't carry this.** Central Manager's proxy is message-passing, not an HTTP reverse proxy, and caps bodies at 2 MB ("Fabric Connect body size must be less than 2MB. Utilize Direct Connect for larger payloads."). A streamed download needs a direct Bearer connection; that's a Studio-side concern, noted on studio#1591.

## Testing

New `unitTests/components/packageComponentOperation.test.js` — 11 tests against a real temporary components root, covering all three response shapes, `skip_node_modules` on the stream path, a full stream → gunzip → tar-extract round trip back to the source tree, the estimate totals with and without node_modules, and both resolution failures. The transport half is already covered by `unitTests/server/serverHelpers/serverHandlers.test.js`, so this stops at the handler's return value.

Ran green locally alongside every other suite that imports the changed modules — `packageComponent`, `envOperations`, `deploymentOperations`, `serverHandlers`, `fastifyRoutes/operations` — 94 passing.

### DIFF
--- a/components/operations.js
+++ b/components/operations.js
@@ -373,14 +373,17 @@ async function packageComponent(req) {
 		try {
 			pathToProject = await fs.realpath(path.join(env.get(hdbTerms.CONFIG_PARAMS.ROOTPATH), 'node_modules', project));
 		} catch (err) {
+			// Only a genuinely absent project earns the friendly message. Everything else —
+			// ENOTDIR from a broken installation, EACCES from a permissions problem — is rethrown
+			// with its code, path, and cause intact, rather than being flattened into a
+			// misleading "missing project". Rethrowing is also what guarantees pathToProject is
+			// set past this point, which the stream shape depends on: its response headers are
+			// committed before the packer walks, so a failure discovered later would reach the
+			// caller as a 200 with a truncated body instead of an error it can act on.
 			if (err.code === hdbTerms.NODE_ERROR_CODES.ENOENT) throw new Error(`Unable to locate project '${project}'`);
+			throw err;
 		}
 	}
-	// The inner catch above swallows any non-ENOENT realpath failure, leaving pathToProject
-	// undefined. That has to be a thrown error rather than a walk of `undefined`: on the stream
-	// path the response headers are already committed by the time the packer runs, so the caller
-	// would receive a 200 with a truncated body instead of a failure it can act on.
-	if (!pathToProject) throw new Error(`Unable to locate project '${project}'`);
 
 	if (req.estimate) {
 		const { totalSize, danglingSymlinks } = await scanPackageDirectory(pathToProject, req);
--- a/components/operations.js
+++ b/components/operations.js
@@ -22,7 +22,11 @@ const {
 const { handleHDBError, ServerError, hdbErrors } = require('../utility/errors/hdbError.ts');
 const { HDB_ERROR_MSGS, HTTP_STATUS_CODES } = hdbErrors;
 const manageThreads = require('../server/threads/manageThreads.js');
-const { packageDirectory } = require('../components/packageComponent.ts');
+const {
+	packageDirectory,
+	scanPackageDirectory,
+	streamPackagedDirectory,
+} = require('../components/packageComponent.ts');
 const { Resources } = require('../resources/Resources.ts');
 const { Application, prepareApplication, ASIDE_STAGING_DIR } = require('./Application.ts');
 const { COMPONENT_PREPARATION_LOCK_DIR } = require('./componentPreparationLock.ts');
@@ -321,10 +325,31 @@ async function dropCustomFunctionProject(req) {
 }
 
 /**
- * Will package a component into a temp tar file then output that file as a base64 string.
- * Req can accept a skip_node_modules boolean which will skip the node mods when creating temp tar file.
+ * Package a component's directory into a tar+gzip archive.
+ *
+ * Three response shapes, selected by the request:
+ *
+ * - `stream: true` — the archive as raw bytes, with download headers attached. serverHandlers.js
+ *   pipes any returned Readable that carries a `headers` Map. **Prefer this for anything that
+ *   isn't known to be small**; see the note below.
+ * - `estimate: true` — no archive at all, just `{ total_size, dangling_symlinks }` from a single
+ *   directory walk, so a caller can decide whether packaging is worth it before paying for it.
+ * - neither (default) — the historical `{ project, payload }` shape, base64 inside JSON.
+ *
+ * The default is retained for compatibility but does not scale, and callers should migrate off
+ * it. Buffering the archive and base64-ing it costs ~4.7x the archive size resident in this
+ * (shared, multi-tenant) process — 2x to concat the gzip chunks, +1.33x for the base64 string,
+ * +1.33x again when Fastify serializes the envelope — and it hard-fails above ~384 MiB of
+ * compressed output, because base64 of that exceeds V8's 512 MiB string cap
+ * (`buffer.constants.MAX_STRING_LENGTH`) and `.toString('base64')` throws ERR_STRING_TOO_LONG.
+ * The streamed shape is constant-memory on both ends and has no size ceiling.
+ *
+ * `skip_node_modules` / `skip_symlinks` apply to all three shapes. `estimate` is checked before
+ * `stream`, so it wins if a caller sets both.
+ *
  * @param req
- * @returns {Promise<{payload: *, project}>}
+ * @returns {Promise<{payload: string, project: string}|{project: string, total_size: number,
+ *   dangling_symlinks: string[]}|Readable>}
  */
 async function packageComponent(req) {
 	if (req.project) {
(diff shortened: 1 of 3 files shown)


components/operations.js
```diff
@@ -351,6 +376,29 @@ async function packageComponent(req) {
 			if (err.code === hdbTerms.NODE_ERROR_CODES.ENOENT) throw new Error(`Unable to locate project '${project}'`);
 		}
 	}
+	// The inner catch above swallows any non-ENOENT realpath failure, leaving pathToProject
+	// undefined. That has to be a thrown error rather than a walk of `undefined`: on the stream
+	// path the response headers are already committed by the time the packer runs, so the caller
+	// would receive a 200 with a truncated body instead of a failure it can act on.
+	if (!pathToProject) throw new Error(`Unable to locate project '${project}'`);
```
kriszyp: Could we fix the inner catch invariant instead of converting every swallowed error into “Unable to locate”? For example, an `ENOTDIR` or `EACCES` from the fallback currently reaches this guard and loses its original code, path, and cause, making a broken installation or permissions problem look like a missing project. Please throw the friendly missing-project error only for `ENOENT`, rethrow every other error from the inner catch, and then remove this fallback guard/comment because `pathToProject` will be guaranteed on the success path. — KrAIs (GPT-5)

kriszyp (approved):
Nice, good call.

components/operations.js
dawsontoth: Agreed and done in 8e693c1 — the inner catch now throws the friendly message only for `ENOENT` and rethrows everything else with its code/path/cause intact, so the fallback guard and its comment are gone and `pathToProject` is guaranteed past that block; the regression test now asserts the `ENOTDIR` code survives and that the message is *not* the missing-project one, and it fails without the rethrow.

kriszyp (approved):
great

cb1kenobi (approved):

Issue #1316, opened 2026-06-16
Startup transaction-log replay can wedge the boot thread indefinitely (deep-walk + validation-throw)


kriszyp:
## Summary

On a node recovering from an unclean shutdown, `replayLogs` (`resources/replayLogs.ts`) can peg the main thread at ~100% CPU for an unbounded wall-clock time and **never reach "Harper successfully started"** — the node never finishes booting, so it never serves ops and never rejoins replication. Observed in production (the affected node was **on 5.1.2**), where the node cycled through multiple replay attempts over ~80 min without completing.

This is the **startup/replay-path** counterpart to the live-delivery deep-walk work (harper#1114, harper#1310, harper-pro#399). Those fixes target live subscription delivery; this is the same pathological walk reached via the boot-time log replay, and it is **not** mitigated by harper#1310 (see below).

### (a) Slow-but-*progressing* replay is not bounded

Replay routes each entry through `_writeUpdate`/`_writeDelete`/… → `save()` → `Table.commit`, which runs the out-of-order audit **resequencing walk** (up to `MAX_OUT_OF_ORDER_AUDIT_DEPTH` = 1000). On a node whose transaction log holds a deep backlog of out-of-order writes on high-churn records (a high-churn application table with deep pure-`patch` histories), **every replayed write costs an O(depth) synchronous walk**. With a large log this pegs the boot thread for an unbounded time and the boot never completes.

- harper#1310's up-front keyed dedup **is** on this path (replay → `commit`), but it only absorbs *exact `(version, nodeId)` duplicates*. The walks seen during the affected node's replay are **genuinely deep** (depth-cap warnings fire continuously), not exact dups — so #1310 does not engage. **The node is on 5.1.2 and still wedges**, which is the proof.
- The existing stall-abort guard (harper#1266, `shouldAbortStalledReplay`, `replayLogs.ts:87`) only fires on a run of entries with **no successful write** (`noProgressRun > 0`). The replay *is* making forward progress — each deep-walk write eventually succeeds and resets `noProgressRun` to 0 — so the guard never trips. The gap: #1266 bounds *zero-progress* stalls, **not** *pathologically-slow-but-progressing* replay.

[...]

PR #2166, opened 2026-08-14
Bound the boot transaction-log replay batch so an oversized transaction can't OOM the node

### LINKED ISSUE #2161: Boot transaction-log replay stages one unbounded in-memory batch per version → OOM, unbootable node
## Summary

On startup after an unclean shutdown, transaction-log replay commits and re-creates its `DatabaseTransaction` **only at a version boundary**. Every write within a single version stages into that one transaction, with no intra-version size-based flush. A version carrying a very large number of writes (or one very large write) accumulates an unbounded in-memory write batch and exhausts heap. Because the process dies mid-replay, the next boot restarts the same replay and OOMs again — a boot loop with no forward progress and no self-recovery.

The existing replay guards (harper#1266 / harper#1316) abort **only at version boundaries** ("aborting never tears a same-version write batch in half") and are **progress/time-based, not memory-based**, so they do not fire before the heap is exhausted.

## Where (v5.2.1; core at the v5.2.1 pin)

- `resources/replayLogs.ts`: the transaction is committed + recreated only when `lastTimestamp !== version`; within a version, every `save`/`writeUpdate`/`writeDelete` stages into the single `DatabaseTransaction`. There is no size-based mid-version flush.
- The version-boundary abort guards are progress/time-bounded, not memory-bounded.

## Impact

On an internal replication-heavy cluster whose system transaction logs had grown to multiple GB (see #2140 — RocksDB log pruning never re-arms, which is *how* they grew), an unclean shutdown triggered a replay that exhausted heap on startup and the node could not boot. Multiple nodes entered this unbootable state at once. This is the terminal blast-radius mechanism of a broader replication incident: the trigger was #2153 / harper-pro#545, the growth was #2140, and **this** is why the affected nodes could not recover on restart.

## Suggested fix

Bound the in-memory replay batch by **size**, not just by version boundary — e.g. flush (commit + recreate the transaction) when the staged batch exceeds a byte/record threshold. A mid-version flush is safe for **replay** (idempotent, version-deduped) even though it isn't for live writes, so the same-version-batch invariant that blocks the abort guards does not block a replay-time flush. Alternatively, checkpoint replay progress so a killed replay resumes rather than restarts from scratch.

Relates to #2140 (log growth is the precondition) and #2153 / harper-pro#545 (the incident trigger).

---

kriszyp:
Fixes #2161.

Boot transaction-log replay committed and recreated its `DatabaseTransaction` only when the entry's
version changed, so every write belonging to one source transaction staged into a single in-memory
batch — the JS write operations (each retaining the prior record it read) plus the native write
batch — with no bound at all. A version carrying a very large number of writes (a bulk load, or a
peer transaction replicated as one version) therefore exhausted heap mid-replay. Because the process
died during replay, the next boot restarted the same replay and died again: a boot loop with no
forward progress and no self-recovery.

The existing replay guards cannot catch this. `shouldAbortStalledReplay` resets on every successful
write and `shouldAbortSlowReplay` is wall-clock, so a replay writing its way to an OOM trips neither.

## What changed

**The bound.** `shouldFlushReplayBatch` bounds the staged batch by estimated bytes (32 MB) or staged
writes (10,000); replay commits and starts a new transaction when either is crossed, not only at a
version boundary. The decision includes the entry about to be staged, so the batch overshoots by at
most the write that crosses it. Bytes are charged as the audit entry's own size, plus the prior
record each staged write retains (the dominant allocation for a patch/delete-shaped transaction, and
invisible to the audit entry's size), plus a fixed per-write overhead.

Committing in pieces is safe for **replay** specifically: what a crash leaves torn is re-applied on
the next boot, from the log's last-flushed position and in timestamp order.

**Same-key program order (`DatabaseTransaction.partiallyCommitted`).** Starting a new transaction
drops the per-key staged-write chain, which is what tells the write path that a later write to a key
already written in this transaction is program order rather than a re-delivered duplicate. After a
commit the earlier write is durable and ties the later one on (version, nodeId), so the later write
falls into the out-of-order/dedup block — where the keyed audit lookup can set `write.skipped = true`
and silently drop it. The flag says the source transaction is already durable in part, and the three
tie-sensitive paths (update/patch, invalidate, relocate) then read the tie as program order. Delete
already applied on a tie.

The basis: entries at or after the log's last-flushed position were not flushed, and RocksDB's own
WAL is off, so a replayed entry's write is not durable — the tie can only be an *earlier* entry of
the same transaction.

**Knowing which transaction is torn.** A version can also be durable in part because a *previous*
interrupted replay committed a batch of it: `databaseFlushed()` in rocksdb-js records the flushed
position from the last **commit** covered by a memtable flush, so once replay batches, that position
can fall inside a source transaction and replay resumes there. Replay therefore records the version
it is tearing in the root store — before the batch it describes, cleared once that version is applied
in full — so a boot that tore nothing keeps the identity-tie dedup for every version it applies. The
marker fails closed: if it cannot be written, replay aborts rather than tearing a transaction it
cannot mark; if it cannot be read, replay treats every version as possibly torn for that boot.

**Failure paths.** A commit that fails for a transaction already committed in part now aborts replay,
rather than applying later writes over writes that were never committed — and the marker survives
every abort path, since the rest of that transaction is still to come. The wall-clock abort (#1316)
goes back to version boundaries only, so a timeout can still only cut between source transactions.

## For the human reviewer

- **The dedup bypass is the judgment call.** `partiallyCommitted` skips the keyed audit dedup for
  equal-version, same-node ties on the versions it covers. The WAL-off argument above says such a tie
  can only be this transaction's own earlier write. The **residual**: in the window where a memtable
  flush completed but its position has not been recorded yet, replay can re-apply an entry that *is*
  durable — idempotent for put/delete/invalidate/relocate, but a commutative patch would fold twice.
  Reviewers raised this three rounds running; I scoped it as tightly as I could find a way to (torn
  versions only, known from the durable marker) rather than eliminating it.
- **The marker is new durable state.** One root-store symbol key, written only when an oversized
  transaction is being torn. A stale marker (crash after the final commit, before the clear) is
  conservative — it only widens tie handling for one exact version that will never recur.
- **Mid-version tearing itself.** The alternative is keeping replayed transactions atomic and
  bounding memory another way, or refusing to replay an oversized version. A node that booted torn
  cannot be un-torn — but with the current code that node does not boot at all.
- **The deeper fix the issue also suggests** — a durable replay *cursor*, so a killed replay resumes
  instead of restarting — would make the marker and the tie widening unnecessary. Worth a follow-up
  issue if you agree; I kept this change to the memory bound and what it forces.
- **The bounds are constants, not config**, and the byte figure is an estimate over encoded sizes
  while the heap holds decoded objects. The write bound is the backstop.

## Verification

- Unit: `test:unit:main`, `test:unit:resources` — new `shouldFlushReplayBatch` cases pass. Four
  pre-existing failures on this machine are unchanged by this branch and reproduce with base sources
  built (`replayStructures.test.js` composite-key, two `databases.test.js` `randomAccessFields`
  cases, `deploymentRecorder.test.js` transaction context).
- Integration (end-to-end route): new `integrationTests/server/replay-large-transaction.test.ts` —
  three crash/replay cycles over 25,000-write single transactions: every record replays, a key
  written twice in one transaction replays to its last value, and the same for a patch. The first
  **fails on base** (base sources rebuilt and rerun: no mid-version commit is reported).
- `crash-replay`, `replay-stress`, `blob`, `transactions`, `ttl` suites pass. The full
  `test:integration:all` run on this machine has many unrelated failures (suite-level 60s timeouts
  under local CPU contention); every failing suite that touches this change passes in isolation. CI
  is the gate.
- I confirmed the mechanism and the fix by instrumenting `resources/Table.ts`: pre-fix, the repeated
  key enters the tie path (`precedes=0 priorStaged=false`) and survives only because the best-effort
  keyed dedup missed; post-fix it never enters that path. The marker's write/clear round-trip was
  verified the same way.
- **Not covered by tests**: the OOM itself (needs a heap-constrained run over GB-scale logs), a crash
  *between* two replay batches, marker I/O failure, and the invalidate/relocate tie paths. The marker
  lifecycle is the crash-safety mechanism here and it has no automated coverage — the honest gap in
  this PR.

## Review coverage

Pre-push cross-model review (`prepush-review.mjs`), five rounds: **Codex** (graded), **Gemini**,
**Cursor Composer** (rounds 1–2, then Cursor round cap), **Harper domain adjudication**. Every round
found something real, and each one is fixed above: round 1's same-key blocker (which the domain leg
dropped as pre-existing — I traced the code, instrumented the build, and it was not); round 2's
version-scoped flag, `!newVersion` commit guard and public-`Context` placement; round 3's "assume
every replay is torn" default, which is what produced the durable marker; round 4's marker
fail-closed and abort-path clearing. The resume-inside-a-transaction question was settled by reading
the rocksdb-js implementation rather than assuming it.

Two findings are carried into this PR rather than fixed: the transaction-wide tie widening (the first
bullet above — the reviewers and I disagree on how reachable the residual is, so it is stated for you
to rule on) and the missing marker-lifecycle tests. The final commit is a one-line fix for a round-5
finding (persist the marker even when it could not be read) made after that review ran; no leg has
seen that commit.

<sub>Human-Review-Need: 4 @ 6e70fe49d57a</sub>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,7 @@ benchmarks/compression/results/
 
 # northwind export_local test artifact
 test_export.json.json
+test_export_*.json
 
 # dev-mode boots (harper dev <fixture>) symlink node_modules/harper into the
 # fixture component dir; keep those out of commits
--- a/integrationTests/server/replay-large-transaction.test.ts
+++ b/integrationTests/server/replay-large-transaction.test.ts
@@ -6,10 +6,16 @@
  * exhausted heap mid-replay, and because the next boot restarts the same replay, the node never
  * came back up. Replay now also commits once the staged batch crosses its size bounds, mid-version.
  *
- * This exercises that mid-version flush on a real crash/replay cycle and asserts the property the
- * flush must not break: a torn-and-recommitted transaction still replays completely. It does not
- * reproduce the OOM itself (that needs a heap-constrained run over GB-scale logs); the bound is
- * covered by unit tests over shouldFlushReplayBatch.
+ * These exercise that mid-version commit on real crash/replay cycles and assert the properties it
+ * must not break: a torn-and-recommitted transaction still replays completely, and a key written
+ * twice in one transaction still ends at its LAST value even when the two writes land in different
+ * batches (the committed earlier write ties the later one on version/nodeId, which without
+ * `Context.partiallyCommitted` reads as a re-delivered duplicate and is dropped).
+ *
+ * They do not reproduce the OOM itself (that needs a heap-constrained run over GB-scale logs), and
+ * they do not crash BETWEEN two replay batches — the bound is covered by unit tests over
+ * shouldFlushReplayBatch, and the re-application of a torn version rests on replay restarting from
+ * the log's last-flushed position.
  */
 import { suite, test, before, after } from 'node:test';
 import { ok, strictEqual as equal } from 'node:assert';
@@ -26,6 +32,8 @@ import {
 
 const DB = 'bigtxn';
 const TABLE = 'bigrows';
+const REPEAT_TABLE = 'repeatrows';
+const REPEAT_ID = 1_000_000;
 // Comfortably over REPLAY_MAX_STAGED_WRITES (10,000) so one insert operation — one transaction, one
 // version — forces at least one mid-version flush. The log assertion below fails loudly if that
 // bound is ever raised past this, rather than silently no longer exercising the flush.
@@ -52,13 +60,28 @@ function readHarperLogs(dirs: (string | undefined)[]): string {
 	return combined;
 }
 
+async function crashAndReplay(ctx: ContextWithHarper) {
+	await new Promise<void>((resolve) => {
+		ctx.harper.process.once('exit', () => resolve());
+		ctx.harper.process.kill('SIGKILL');
+	});
+	await startHarper(ctx, { startupTimeoutMs: MAX_REPLAY_STARTUP_MS });
+}
+
+async function readRepeated(ctx: HarperContext) {
+	const rows = await op(ctx, { operation: 'sql', sql: `select n from ${DB}.${REPEAT_TABLE} where id = ${REPEAT_ID}` });
+	return rows[0] ?? {};
+}
+
 suite('Transaction log replay of an oversized transaction', (ctx: ContextWithHarper) => {
 	before(async () => {
 		await startHarper(ctx, {
 			env: { HARPER_NO_FLUSH_ON_EXIT: true }, // don't flush on exit; we are simulating a crash
 		});
 		await op(ctx.harper, { operation: 'create_database', database: DB });
-		await op(ctx.harper, { operation: 'create_table', database: DB, table: TABLE, primary_key: 'id' });
+		for (const table of [TABLE, REPEAT_TABLE]) {
+			await op(ctx.harper, { operation: 'create_table', database: DB, table, primary_key: 'id' });
+		}
 	});
 	after(async () => teardownHarper(ctx));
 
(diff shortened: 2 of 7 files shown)


kriszyp:
Closing this implementation, while keeping [#2161 — Boot transaction-log replay stages one unbounded in-memory batch per version](https://github.com/HarperFast/harper/issues/2161) open with corrected scope and acceptance criteria: https://github.com/HarperFast/harper/issues/2161#issuecomment-5299989281

The batch accounting and same-key replay tests were useful, but the central design is not acceptable: it deliberately commits one source transaction in pieces and adds `DatabaseTransaction.partiallyCommitted` to the shared write path to make that split work. A partial physical redo is only compatible with ACID when recovery keeps it invisible until the entire committed transaction is restored. This implementation does not enforce that boundary: batch-commit failures log `fatal` and stop replay, `fatal` only logs, and database startup continues. The resulting partial state can therefore become readable.

The current tests also do not reproduce the claimed OOM on base or exercise a crash/failure between batches, and the marker/idempotence path admits a commutative double-application window. This needs a base-failing constrained-heap replay reproduction first, followed by a design that retains atomic source-transaction replay or stages recovery invisibly and publishes it atomically.

The branch is left intact for any useful test and accounting work.

— KrAIs (GPT-5), with Kris

Issue #787, opened 2026-05-26
Skip license enforcement for model-* metrics in dev mode (no installed license)


kriszyp:
## Problem
PR #779 starts emitting `model-embed`, `model-embed-tokens`, `model-generate`, and `model-generate-tokens` into `hdb_raw_analytics` → `hdb_analytics`. Phases 2 + 3 will extend the central-manager Plan schema and signed license blob with embed/generate budgets and add `fabric-ai-level-1..3` SKUs that gate `embed()` calls on those budgets.

Once enforcement is in place, a developer running harper-pro locally without an installed license could see embed/generate calls 429'd against a default trial budget — which would be a terrible dev experience.

## Proposal
Treat dev mode the same way the existing reads/writes enforcement does:
- When `hm.devMode === true` OR when no license is installed, skip rate-limit checks for `model-*` metrics in the license reporting path
- Forensic rows still go to `hdb_model_calls` (audit/observability is not the issue)
- Aggregate metrics still emit to `hdb_raw_analytics` (no reason to muddy the data path)

## Verification
- [ ] Boot harper-pro in dev mode with no license. Call `scope.models.embed()` 10,000 times in a tight loop. Confirm no 429.
- [ ] Boot harper-pro with a paid `fabric-ai-level-1` license + budget of 1,000 embeds. Confirm the 1,001st call 429s.
- [ ] Confirm `hdb_model_calls` rows land in both cases.

## Why this is its own issue
Easy to forget when wiring up the enforcement in Phase 2. Tracking it separately so it doesn't get lost in the larger CM-side PR.

## Related
- PR #779 (Phase 1 model analytics emission)
- Pending: CM Plan.planLimits extension + signed license blob (Phase 2)
- Pending: fabric-ai-level-1..3 SKUs (Phase 3)

PR #2168, opened 2026-08-14
Fix interrupted RocksDB table drops

kriszyp:
Validation found that this Harper diff does not fix the storage-layer defect reported in #1381. The failing catalog write uses a metadata column-family handle invalidated by a concurrent native database destroy/open lifecycle race; the root fix is [rocksdb-js PR #787, “Serialize database destruction with concurrent opens”](https://github.com/HarperFast/rocksdb-js/pull/787), which is not yet merged or released. This draft should not merge as-is.

The column-family postcondition, cross-worker source-cache guard, and async entry-handler rejection handling in this diff may be useful defense-in-depth changes, but they are not a substitute for enforcing the process-global path lifecycle invariant in rocksdb-js. Keep this draft parked until #787 is merged and published, then pin the released dependency and rerun the regression before deciding which Harper-side changes remain justified.

Refs #1381.

## For the human reviewer

1. Do not merge this draft as the #1381 fix. The observed failure crosses the rocksdb-js native registry boundary, where database destruction releases a path before physical teardown is complete and a concurrent worker can reopen it. A Harper-side CF sweep cannot restore a metadata handle after that lifecycle violation.
2. After rocksdb-js #787 is released, decide whether to retain any of this diff as separately scoped defense in depth. Keeping it would require resolving the open review feedback and adding coverage for the persisted-tombstone-only source-cache path; dropping it avoids carrying unrelated `Scope` behavior in a storage fix.

## Verification

- `npm run build` — passed on Node 26.2.0.
- `npx mocha unitTests/resources/dropTableGhost.test.js unitTests/components/Scope.test.js` — 49 passed, 1 LMDB-only pending.
- `npm run test:integration -- --isolation=none "integrationTests/apiTests/blob.test.mjs"` — 22 passed on Node 26.2.0.
- The same integration command — 22 passed on Node 24.19.0.
- PR CI passed every Node 24 integration shard. Its Node 22/24/26 unit jobs each failed on the unrelated `DeploymentRecorder.ingestPayload transaction context` timeout-budget assertion.

These green runs confirm the branch is mechanically healthy; they do not reproduce the intermittent native destroy/open race. The added tests exercise simulated redundant CF drops and orphaned CF cleanup, not the catalog-handle invalidation seen in the failing integration logs.

## Review coverage

Authored by GPT-5 Codex. Claude and Gemini reviewed commit `3e271283cb75`; both left open comments on the Harper diff. Root-cause validation traced the lifecycle invariant to rocksdb-js #787, whose current Node 22/24/26, native, stress, and benchmark checks are green but which remains unreleased.

— GPT-5 Codex

<sub>Human-Review-Need: 4 @ 3e271283cb75</sub>

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -203,6 +203,10 @@ interrupted drop instead of resurrecting the table. Without this, surviving cata
 silently re-opened with create-if-missing on the next start, which resurrects "deleted" tables
 (with their data, if the column families were never actually removed).
 
+On RocksDB, treating a redundant `dropSync()` as success is safe only after re-listing the live
+column families: both the drop and recovery paths remove catalog rows only when none still begin
+with `T/`.
+
 ## MCP protocol surface (`components/mcp/`)
 
 The MCP Streamable-HTTP transport (spec `2025-06-18`) is served at `/mcp` under **two profiles**: an
--- a/components/Scope.ts
+++ b/components/Scope.ts
@@ -474,6 +474,8 @@ export class Scope extends EventEmitter<ScopeEventsMap> {
 						})
 						.finally(() => pendingOperations.delete(tracked));
 					pendingOperations.add(tracked);
+					// Keep the tracked rejection available to waitForInitialLoads().
+					void tracked.catch(() => {});
 				}
 			};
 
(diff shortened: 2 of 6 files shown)


kriszyp:
Superseded by draft [#2206, “Quiesce workers before dropping RocksDB tables”](https://github.com/HarperFast/harper/pull/2206).

The captured #1381 timeline shows an HTTP worker committing through a retained table column-family handle after another worker begins the table drop. That is distinct from [rocksdb-js #787, “Serialize database destruction with concurrent opens”](https://github.com/HarperFast/rocksdb-js/pull/787), which protects whole-database destroy/open lifecycle. #2206 handles the demonstrated Harper race directly with a pre-destructive worker barrier, targeted drain/cancellation, handle closure, and generation-guarded cleanup; it does not depend on #787.

Closing this draft avoids two competing Harper-side designs for the same symptom while preserving its investigation history.

— GPT-5 Codex

PR #2169, opened 2026-08-14
Prevent job wedges on runtime database opens

kriszyp:
## Summary

- Serialize a database's root RocksDB opens with `DestroyDB` while its directory is being dropped.
- Remove the terminology test's `drop_database` polling workaround and assert the direct operation succeeds.
- Emit each schema-change signal once, retain the canonical operation default, and clear LMDB aliases on `drop_database`.

## Root cause

In [the failing uWS integration run](https://github.com/HarperFast/harper/actions/runs/31554910290), the first stuck job starts at the same second RocksDB records a same-process `LOCK` acquisition for `tuckerdoodle`. rocksdb-js 2.7.1 already serializes concurrent opens. Its registry, however, releases a database entry before it calls `DestroyDB`. The preceding `drop_database` races that destroy with schema propagation reopening the root in another thread; the job worker then hits the resulting native `LOCK` failure during module initialization, before `jobProcess` enters its error-handling IIFE.

The fix uses the existing per-database mutex for both root opens and destruction. A reopen now waits for `DestroyDB` to finish rather than reaching RocksDB while the directory is being removed.

## Verification

- `npm run build`
- `npm run lint:required`
- `npx mocha unitTests/dataLayer/restoreMarker.test.js`
- `npx mocha unitTests/server/threads/preloadSafeMode.test.js`
- `HARPER_UWS_HTTP=1 node --experimental-strip-types --test integrationTests/apiTests/terminology.test.mjs` (47 passing)

`npm run test:unit:main` exercised the new coverage but has two unrelated existing failures in this worktree: deployment-recorder timeout budget and the long-worktree domain-socket path expectation. The full integration fan-out exited nonzero without retaining the failing suite's log; the suspected final rate-limiter suite passes cleanly in isolation.

Independent storage review was run before the final PR. It confirmed the `DestroyDB` gap and identified broader follow-up work for pre-IIFE job failures; this stays a draft for human storage review.

— GPT-5 Codex

## Rebase + review follow-up (2026-09-13)

Rebased onto current `main` (407 commits ahead at the time). This PR's diff had never gone through
the pre-push review CLI before (`ran=none` in the prior footer) — 4 rounds ran during this rebase and
found real, fixed bugs plus a few the PR should carry forward as open decisions.

**Fixed in this pass:**
- `resources/databases.ts` (`openRocksDatabase`/`dropDatabase`): merge conflict against main's
  independently-added `stopAuditCleanup`/`removeStorageReclamation` calls (#2338) — kept both.
- The merge introduced a real 30s job wedge: `dropDatabase` now yields (`await stopAuditCleanup()`)
  while holding the open lock, letting a same-thread schema-change rescan re-enter
  `acquireDatabaseOpenLock` for the same path and spin the full timeout. Fixed with a same-thread
  reentrancy fast-fail in `dataLayer/restoreMarker.ts`.
- `resources/branchDatabase.ts`'s undeploy walk didn't know about this PR's `` `restore` `` lock
  metadata sibling and failed outright when a branch had ever taken the open lock. Fixed.
- Review round 1 (codex + gemini + cursor-grok, adjudicated): a blob-deletion race (open lock released
  before `deleteRootBlobPathsForDB`), a directory-resurrection race (`mkdirSync`/`existsSync` ran
  before the lock), a silent-scan-abort from the reentrancy fix above, and an unguarded
  `port.threadId` read. All fixed and re-reviewed clean.

**Not fixed — flagging for the merge decision instead of changing unilaterally on a rebase task:**
- **Worker lock-ownership tracking is best-effort IPC** (`resources/databases.ts:78-101`): the main
  thread learns which worker holds a lock via async `postMessage`; an abnormal worker exit (OOM,
  `kill -9`) before that message lands leaks the lock until process restart, or (per a Cursor Grok
  angle) a reused token could double-release a live opener's lock. Fixing this well means a different
  ownership-tracking primitive (e.g. main-thread-owned acquisition, or a `SharedArrayBuffer`-backed
  registry), which is a bigger, harder-to-reverse call than this task should make.
- **The open path blocks its own event loop.** `acquireDatabaseOpenLock`'s `Atomics.wait` retry parks
  the calling thread in 10ms slices for up to 30s on contention, and the timeout carries no HTTP
  `statusCode` (surfaces as 500, not 409). This is inherent to the lock design this PR introduces, not
  something the rebase changed. Reviewer note: Gemini repeatedly flagged this as `Atomics.wait` on
  Node's main thread throwing a `TypeError` and crashing the process — that's incorrect for Node
  (unlike browsers, Node permits `Atomics.wait` on the main thread; it blocks, it doesn't throw). I
  confirmed this empirically while debugging the job-wedge above: the exact same call executed on
  mocha's single (main) thread and blocked for the full duration without throwing. The real defect is
  the blocking itself, not a crash.
- **A scan can still resurrect a fully-dropped database as an empty zombie.** The fix in this pass
  closes the narrower "resurrect a directory a destroy is still working on" race, but a scan that
  enumerated a database's directory *before* a drop completes can still recreate it (empty) after the
  drop finishes and releases the lock, because `openRocksDatabase` has no way to know the caller is
  discovering rather than intending to create. Closing this needs an explicit create-intent flag
  threaded through `database()`/`openRocksDatabase`'s callers — an API shape change, not a rebase fix.
- Two Cursor Grok findings on pre-existing (not this-rebase) loop structure in `dropDatabase`: the
  per-table lock-acquire loop clears cache entries in the same pass it acquires locks, so a mid-loop
  timeout on table 2+ leaves table 1's cache entry cleared without actually dropping anything
  (`resources/databases.ts:2071`-ish); and the tableless-drop branch opens-then-releases the lock via
  `database(...)` before re-acquiring it for the drop, leaving a short unlocked window
  (`resources/databases.ts:2088`-ish). Both are structural to code this rebase didn't author; noted
  for whoever owns the next pass rather than fixed blind here.

Verified after every fix: `npm run build` / `typecheck` / `lint` / `format:check` clean; full
`test:unit:resources` (2495 passing, 0 failing), `test:unit:main` (5520 passing, 2 known-local-only
failures unrelated to this diff), `test:unit:dataLayer`, `test:unit:backup` all clean.

— Claude Sonnet 5

<sub>Review-Coverage: authored=claude; ran=codex,gemini; declined=cursor-grok,cursor-composer,domain; rounds=4; full=1 @ 25942aafa05e</sub>

<sub>Human-Review-Need: 4 @ 25942aafa05e</sub>

### DIFF
--- a/dataLayer/harperBridge/ResourceBridge.ts
+++ b/dataLayer/harperBridge/ResourceBridge.ts
@@ -4,13 +4,10 @@ import { table, getDatabases, database, dropDatabase, type Table } from '../../r
 import insertUpdateValidate from './bridgeUtility/insertUpdateValidate.js';
 import SearchObject from '../SearchObject.ts';
 import {
-	OPERATIONS_ENUM,
 	VALUE_SEARCH_COMPARATORS,
 	VALUE_SEARCH_COMPARATORS_REVERSE_LOOKUP,
 	READ_AUDIT_LOG_SEARCH_TYPES_ENUM,
 } from '../../utility/hdbTerms.ts';
-import * as signalling from '../../utility/signalling.ts';
-import { SchemaEventMsg } from '../../server/threads/itc.js';
 import { asyncSetTimeout } from '../../utility/common_utils.ts';
 import { transaction } from '../../resources/transaction.ts';
 import type {
@@ -179,14 +176,10 @@ export class ResourceBridge extends BridgeMethods {
 			database: createSchemaObj.schema,
 			table: null,
 		});
-		return signalling.signalSchemaChange(
-			new SchemaEventMsg(process.pid, OPERATIONS_ENUM.CREATE_SCHEMA, createSchemaObj.schema)
-		);
 	}
 
 	async dropSchema(dropSchemaObj) {
 		await dropDatabase(dropSchemaObj.schema);
-		signalling.signalSchemaChange(new SchemaEventMsg(process.pid, OPERATIONS_ENUM.DROP_SCHEMA, dropSchemaObj.schema));
 	}
 
 	async updateRecords(updateObj) {
--- a/dataLayer/restoreMarker.ts
+++ b/dataLayer/restoreMarker.ts
@@ -51,6 +51,7 @@ import { tryFileLock, fileLockRelease } from '@harperfast/rocksdb-js';
 export const RESTORE_META_DIR = '`restore`';
 export const RESTORE_LOCK_SUFFIX = '.lock';
 export const RESTORING_MARKER_SUFFIX = '.restoring';
+const DATABASE_OPEN_LOCK_SUFFIX = '.open';
 
 /**
  * Directory holding the restore metadata for a database — the reserved `` `restore` `` sibling of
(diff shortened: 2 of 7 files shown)


kriszyp:
The root cause is in rocksdb-js's process-global database lifecycle, not Harper's resource-open path. [rocksdb-js PR #787, "Serialize database destruction with concurrent opens"](https://github.com/HarperFast/rocksdb-js/pull/787) moves the invariant to the native registry: destruction owns the physical path, closes all in-process descriptors, waits for in-flight backup/checkpoint work, and blocks concurrent opens until teardown completes.

Recommended dependency sequence:

1. Merge and publish the rocksdb-js change.
2. Update Harper to the published version and rerun this PR's runtime-created-database regression.
3. Remove the Harper `.open` mutex commits from the released path; they duplicate a storage-layer invariant and can block the main thread through `Atomics.wait()` during component loading.

This PR should remain draft until that artifact exists. The schema-change signal deduplication and LMDB alias cleanup are separable changes and do not need to depend on the rocksdb-js lifecycle fix.

— GPT-5 Codex

PR #2170, opened 2026-08-14
fix(storage): version raw RocksDB primary writes

### LINKED ISSUE #1762: RecordEncoder: 0x42 prefix-less classic-record decode ambiguity keeps resurfacing — write-side guard or raw-put caller audit
## The ambiguity

The rocksdb decode heuristic in `resources/RecordEncoder.ts` (`decode()`) detects a local-timestamp prefix by first byte `66` (0x42) — but 0x42 is also classic shared-structure **record-id #2**. A **prefix-less** classic record that happens to start with that id is misread as timestamp-prefixed: 8 bytes are stripped and the decode corrupts (observed as `Could not find typed structure N` → `null`). Typed structs live at 0x20–0x3f and never collide, so this only surfaces with classic structures on the rocksdb path.

The heuristic's in-code comment documents the collision and the caller-local escape hatch (`options.noMetadata`), added in b8490dc06 when the **audit store** hit it. It has since been hit a second time — #1549, where agent session rows written via raw `primaryStore.put` produced prefix-less records and the whole agent died to "Unknown session"; the fix switched to Resource-level `.put` so rows carry the versioned prefix.

Two bites, two caller-local fixes. Nothing prevents the next raw-put caller from re-tripping it — this issue tracks a structural answer.

## Call sites to triage (non-test, on main)

Version-less raw `primaryStore.put(id, record)` — prefix-less writes, the exposed shape:

- `agent/session.ts` (×6) — fixed when #1549 lands
- `resources/analytics/write.ts:1161`
- `resources/models/analyticsTable.ts:162`

Version-carrying raw puts (`ResourceBridge.ts:161`, `Table.ts:868`, `Table.ts:958`) stage version metadata and should decode unambiguously, but the audit should confirm.

## Options

1. **Write-side guard** — a store that carries per-record version metadata refuses (or warns loudly on) a prefix-less write. A raw put with no staged version to a versioned store is arguably always a bug; this turns silent corruption into an immediate error at the write site.
2. **Call-site audit + lint** — triage the sites above, then a lint rule or wrapper so raw `primaryStore.put` against versioned stores can't be reintroduced casually.
3. **Decode-side sanity check** — before committing to the 8-byte strip, check the implied float64 is a plausible ms-epoch timestamp. Narrows the collision window; doesn't eliminate it.

(1) is the strongest: it fails at the write, where the bug is, instead of at some later read.

Credit: the second diagnosis and the decode-path analysis are from #1549's writeup.

🤖 Posted by [Claude](https://claude.com/claude-code) on behalf of @heskew

kylebernhardy:
## Summary

- enforce the RocksDB primary-store invariant at put and putSync: raw record writes receive a complete 12-byte version and zero-flags metadata prefix
- preserve valid explicit versions, reject unusable zero/non-finite versions, and leave recordUpdater/self-versioned/non-RocksDB encoders unchanged
- clear staged metadata across binary bypasses and failures, and prevent stale blob flags from leaking into raw rows
- document the 0x42 classic-record collision and update the now-stale agent session guidance

Closes #1762

## Verification

- regression test fails on origin/main for missing flags metadata and missing positional version, then passes on this branch
- 11 focused metadata/raw-write tests pass
- full resource suite: 1,563 passing, 15 pending; an unrelated HNSW routing flake passed immediately in isolation and the full retry passed
- build, typecheck, formatting, and changed-file lint pass
- full integration suite passes with sequential isolation because this machine lacks the loopback alias pool
- generic model-telemetry application on a fresh branch-built Harper root returns all six expected hdb_model_calls rows, including both formerly missing generateStream rows

## Review

- independent Claude review plus Harper-domain adjudication completed; all blocking findings addressed
- direct Gemini CLI review passed correctness, global-state safety, data integrity, and RocksDB/LMDB boundaries after its stale-comment finding was fixed


resources/PrimaryRocksDatabase.ts
```diff
@@ -150,6 +150,16 @@ export class PrimaryRocksDatabase extends RocksDatabase {
 		return when(this.getEntry(id, { ...options, async: true }), (entry: Entry) => entry?.value);
 	}
 
+	put(id: any, value: any, options?: any): any {
```
kriszyp, line 153: This override narrows the store's argument list to three, so a 4th positional argument is silently dropped.

`resources/Table.ts:1152` calls it with four:

```js
primaryStore.put(
	Symbol.for('id_allocation'),
	{ start, end, nodeName, pid },
	Date.now(),   // version
	version       // ifVersion — dropped here
);
```

That 4th argument is the optimistic-concurrency guard for extending the auto-increment id range. It was inert before this PR — raw-put records had no version, so `getEntry(Symbol.for('id_allocation')).version` was always `undefined` and the CAS was a no-op — but this PR is exactly what makes it meaningful: raw puts now carry a real version, so `resources/Table.ts:1143` reads a real number and the guard could finally do its job. Truncating the arity here (and in `putSync` below, which is what `Table.ts` actually reaches given `db.put = db.putSync` at `resources/databases.ts:335`) keeps it inert.

The failure it is meant to prevent: two workers both near the end of the range read allocation v1, each computes a different `maxSafeId`, and both writes land unconditionally. The lower `end` persists while the worker that computed the higher one keeps handing out ids above it from the shared `idIncrementer`. On restart, `getNewId` reads the lower `end`, scans only `[start, end]` for the last key, and starts re-issuing ids that already exist.

Suggest forwarding everything — `put(id: any, value: any, options?: any, ifVersion?: any)` passing all four through, or `put(id, value, ...rest)` — in both `put` and `putSync`.

One caveat I could not settle from this checkout: `node_modules` isn't present here, so I couldn't confirm whether `@harperfast/rocksdb-js` honours a 4th positional `ifVersion` the way lmdb-js does. If it doesn't, the right fix is at the `Table.ts` call site instead — but either way the two should agree, and forwarding is the safe default.

resources/PrimaryRocksDatabase.ts
```diff
@@ -172,7 +182,12 @@ export class PrimaryRocksDatabase extends RocksDatabase {
 
 	putSync(id: any, value: any, options?: any): any {
 		this.#cache?.delete(id);
-		return super.putSync(id, value, options);
+		const staged = stageRawPrimaryEncoding(this.#enc, typeof options === 'number' ? options : options?.version);
```
kriszyp, line 185: The version resolved here goes into the record prefix, but `options` is passed to `super.putSync` unchanged — so the store-level version and the encoded version can disagree for raw callers.

Two cases:
- `resources/analytics/write.ts:1186` passes no version at all. The prefix gets `getNextMonotonicTime()`; the store gets `undefined`.
- The zero/negative fallback (covered by the `replaces an unusable explicit version with a monotonic version` test) writes a monotonic version into the prefix while `super.putSync` still receives `0`.

Primary stores are opened with `cache: true` (`resources/databases.ts:876`, `resources/databases.ts:1687`), which turns on the verification table. `getEntry` caches `entry.version` — the value decoded from the prefix — and passes it back as `expectedVersion` on the next read. If the native VT slot is seeded from the version handed to `putSync` rather than from the decoded prefix, those two never match and raw-put rows would never take the cache fast path. That is a throughput loss rather than a correctness bug (a mismatch falls back to a DB read, which is correct), but it silently undoes the caching for exactly the high-volume analytics rows this change is aimed at.

I couldn't verify which value rocksdb-js's VT uses from this checkout, so treat that half as a question. Independently of the answer, it seems cleaner to forward the version you actually encoded, e.g. have `stageRawPrimaryEncoding` return the resolved version and pass it down as the put version when the caller didn't supply a usable one — then the two can't drift.

resources/RecordEncoder.ts
```diff
@@ -677,6 +678,22 @@ export function setNextEncoding(timestamp: number, metadata: number, expiresAt =
 	nodeIdAtNextEncoding = nodeId;
 	residencyIdAtNextEncoding = residencyId;
 }
+export function stageRawPrimaryEncoding(encoder: any, version?: number) {
```
kriszyp, line 681: Worth a comment on this function recording what the first condition costs, because it isn't obvious from the code.

`timestampNextEncoding !== 0 || metadataInNextEncoding >= 0` means "someone else already staged, stay out of the way" — correct for the `recordUpdater` → `store.putSync` path it's written for. But the staging slot is module-global and not tied to a store, so if any raw primary put is ever issued while another write's metadata is staged, the raw record consumes that metadata and the write that staged it encodes without a prefix — precisely the bug being fixed, now silent and attributed to the wrong record.

I traced the reachable windows and didn't find a live one: `recordAction` at `resources/RecordEncoder.ts:900` (which runs inside `recordUpdater`'s try, after the primary encode and before the `finally`) only accumulates in-memory metrics; `recordAnalytics` and its `primaryStore.put` fire from a timer. The delete path (`record === undefined`, no encode) leaves the globals staged for longer than the put path does, so it's the one to watch if a future change adds a synchronous store write inside that window.

Two things that would help the next reader: a sentence here saying the guard deliberately defers rather than asserts, and a note on why the invariant isn't enforced inside `encode` itself — where it would cover every path including direct `encoder.encode()` calls. As far as I can tell the obstacle is `resources/RecordEncoder.ts:829`, which calls `store.encoder.encode(auditRecord)` to produce the audit log's embedded record; that one must *not* gain a prefix, and by then the globals are already consumed, so an unconditional encoder-level rule would corrupt it. If that's the reasoning, it's worth writing down — it's the question any reviewer will ask first.

DESIGN.md
```diff
@@ -24,6 +24,12 @@ To match v4: `OpenDBIObject` sets `randomAccessStructure = isPrimary`, and `Reco
...
+
+`PrimaryRocksDatabase.put()` and `putSync()` enforce the write-side invariant: if an initialized RocksDB encoder for a versioned primary write reaches the store without metadata already staged by `recordUpdater`, the wrapper stages a monotonic version and a zero-flags metadata word. The full 12-byte header is required; a timestamp alone would make the decoder consume the first four payload bytes as metadata. A positive, finite positional numeric version from legacy raw-put callers is retained verbatim. Metadata is cleared in `finally`, including binary writes that bypass `RecordEncoder`, so staged state cannot leak into the next store operation. The wrapper also resets the blob encoder's prior-write signal before a raw record encode so the row cannot inherit a stale `HAS_BLOBS` flag. Normal table writes arrive with metadata already staged and stay on their existing path without an additional allocation.
```
kriszyp, line 31: Two corrections to keep this note matching the code.

1. "`PrimaryRocksDatabase.put()` and `putSync()` enforce the write-side invariant" — in production only `putSync` does. `resources/databases.ts:335` assigns `db.put = db.putSync` for every non-dupSort store opened through `openRocksDatabase`, which includes the primary stores at `resources/databases.ts:876` and `resources/databases.ts:1687`. The async `put` wrapper is reachable only for stores constructed directly (the unit tests, and the migration helpers that call `new PrimaryRocksDatabase(...)`). It's worth keeping as defence, but the note — and the `uses a positional version verbatim for async raw puts` test — currently read as if it were a live production path.

2. The note explains the invariant for new writes but doesn't state the limit: records already written without a prefix whose payload begins with `0x42` stay undecodable, and nothing in this change repairs them. Given DESIGN.md is where someone will land when debugging a "Could not find typed structure" report, one sentence saying the fix is write-side-only, and that pre-existing rows decode as `null` with the `MISSING_STRUCTURE_METRIC` warning from `resources/RecordEncoder.ts:486`, would save that reader a lot of time.

resources/blob.ts
```diff
@@ -133,6 +133,9 @@ let encodeForStorageForRecordId: number = undefined; // only enable encoding of
 let promisedWrites: Array<Promise<void>>;
 let currentStore: any; // the root store of the database we are currently encoding for
 export let blobsWereEncoded = false; // keep track of whether blobs were encoded with file paths
+export function resetBlobsWereEncoded() {
```
kriszyp, line 136: Minor scoping point: this exports a setter for a module-global whose whole design is that it's written only by `encodeBlobsWithFilePath` (`resources/blob.ts:1778`) and by `saveBlob` (`resources/blob.ts:2130`), and read once by the encode hook. Adding a public mutator makes it writable from anywhere, and the `resetBlobsWereEncoded()` call in `stageRawPrimaryEncoding` fires on every staged raw put — including binary writes (`asBinary(...)`), which bypass `RecordEncoder` entirely and never read the flag.

The underlying need is real: a raw put runs outside the `encodeBlobsWithFilePath` window, so without the reset the row inherits `HAS_BLOBS` from whatever encoded last, and the `does not inherit a stale blob flag on a wrapper-staged raw write` test demonstrates that.

Two alternatives that keep the global write-local to `blob.ts`: either have the encode hook clear `blobsWereEncoded` after it consumes it at `resources/RecordEncoder.ts:277` (it already resets every other staged global at consumption time, so this would make the set symmetric), or expose the reset as part of a narrower helper rather than a bare setter. Either way the flag stops being settable from arbitrary callers.

kriszyp (commented):
As the alternative suggested, probably better to use the normal write path, I would think.

PR #2179, opened 2026-08-15
Pin the load-bearing native and encoder dependencies

kriszyp:
A caret range does not bind the version that reaches a running Harper node. harper-pro, a rebuilt container, and anyone installing published `harper` without this lockfile can resolve a newer native or encoder dependency without a Harper PR or human merge.

This PR converts the load-bearing ranges to the versions the lockfile already resolved. No installed dependency moves; only the allowed resolution narrows. The pins cover rocksdb-js, the encoder/iterator modules whose identity crosses its boundary, structon, and the optional native addons.

The module identity is now enforced rather than documented as an assumption. Docker smoke requires the root `msgpackr` and `@harperfast/extended-iterable` specs to be exact and byte-equal to rocksdb-js's requirements, verifies their installed root versions, and rejects a nested copy under rocksdb-js. A rocksdb-js automation or Renovate PR that moves only one side will therefore fail before it can merge.

The shrinkwrap canary remains meaningful after rocksdb-js becomes exact: exact specs still participate in installed-versus-packed comparison but not range discrimination, while ranged `fastify` and `@aws-sdk/client-s3` provide the discrimination proof. Registry queries retry three times with 1s/2s backoff. If the surviving canaries are current and a query remains unresolved, the check fails closed with a retry-specific error; Kris selected that behavior rather than treating “unknown” as “current.”

## For the human reviewer

1. **Exact pins in a published library.** These constraints propagate to consumers, which is the intended guarantee. The cost is that a consumer needs a Harper release to take a patch of these dependencies.
2. **Optional native scope.** `bufferutil`, `segfault-handler`, and `utf-8-validate` are exact because their prebuild/platform behavior is load-bearing, but exact optional pins also remove the in-range escape when a new Node ABI needs a newer prebuild. Their absence is largely silent today.
3. **Consumer layout.** Docker smoke proves the Harper image has one rocksdb-js-facing encoder/iterator instance. It cannot prove every downstream harper-pro or npm-consumer tree has the same layout; that remains the main boundary a human should consider.

The pre-existing lock entries for `@harperfast/extended-iterable` and `segfault-handler` still lack `resolved`/`integrity`. That pattern is lockfile-wide rather than introduced here, but an exact version without a tarball hash is a weaker supply-chain guarantee.

## Verification

- `npm run build`
- `npm run lint:required`
- `npx mocha unitTests/build-tools/checkShrinkwrapPins.test.mjs` — 13 passing
- `npm run test:unit:resources` — 1,556 passing, 15 pending
- Regression proof: the three retry/fail-closed cases fail against the pre-fix script and pass at this head.
- The focused suite spawns the real checker process and covers exact/ranged canaries, transient and exhausted queries, empty registry results, real-manifest viability, lockstep divergence, ranged encoder regressions, and nested module copies.
- GitHub: 42 checks passed and two intentionally skipped. This includes Docker smoke, Node 22/24/26 unit tests, adapters, and the complete Linux/Windows/Bun/uWS integration matrix. The first Node 22 attempt hit the unrelated immediate-expiration cache flake in `unitTests/apiTests/cache-test.mjs:69`; its isolated rerun passed.

The local umbrella unit/integration reruns were not clean under this machine's forced isolated Harper root: unrelated config/logger tests expected the user boot configuration, and one high-volume integration run returned nonzero without preserving the failing case in its truncated output. The prior PR head passed both umbrella suites, and the authoritative current-head GitHub matrices are green.

## Review coverage

Authored by GPT-5.6 Codex. Independent pre-push review covered the full PR with Claude and the Harper-domain adjudicator, plus delta rounds with Claude and Gemini. Earlier rounds also included Cursor Grok; Cursor Composer was pruned. The final delta review found the build tooling complete and carried only the dependency-policy decisions above.

<sub>Human-Review-Need: 4 @ 3e2adc81e285</sub>

### DIFF
--- a/build-tools/check-shrinkwrap-pins.mjs
+++ b/build-tools/check-shrinkwrap-pins.mjs
@@ -34,6 +34,7 @@ import { readFileSync } from 'node:fs';
 
 const CHECKED_DEPS = ['@harperfast/rocksdb-js', 'fastify', '@aws-sdk/client-s3'];
 const REGISTRY_QUERY_ATTEMPTS = 3;
+const retryWait = new Int32Array(new SharedArrayBuffer(4));
 
 const pkgRoot = process.argv[2];
 if (!pkgRoot) {
@@ -125,6 +126,13 @@ function verifyCanariesDiscriminate(pins) {
 				// the last array entry.
 				const out = execFileSync('npm', ['view', `${dep}@${range}`, 'version', '--json'], { encoding: 'utf8' });
 				const versions = JSON.parse(out);
+				if (Array.isArray(versions) && versions.length === 0) {
+					console.error(
+						`::error::${dep}@${range} matches no published version -- correct the declared range in package.json`
+					);
+					failed = true;
+					return;
+				}
 				rangeLatest[dep] = Array.isArray(versions)
 					? versions.reduce((max, v) => (compareVersions(v, max) > 0 ? v : max))
 					: versions;
@@ -134,6 +142,7 @@ function verifyCanariesDiscriminate(pins) {
 					`::warning::registry query attempt ${attempt}/${REGISTRY_QUERY_ATTEMPTS} failed for ${dep}@${range} (${e.message})`
 				);
 				if (attempt === REGISTRY_QUERY_ATTEMPTS) failedQueries.push(`${dep}@${range}`);
+				else Atomics.wait(retryWait, 0, 0, 1000 * attempt);
 			}
 		}
 	}
@@ -142,7 +151,7 @@ function verifyCanariesDiscriminate(pins) {
 	if (stillDiscriminates) return;
 	if (failedQueries.length > 0) {
 		console.error(
-			`::error title=Retry dependency canary check::Could not verify that the shrinkwrap canaries still discriminate after ${REGISTRY_QUERY_ATTEMPTS} registry query attempts for ${failedQueries.join(', ')}. Retry this job; if the error persists, check npm registry availability.`
+			`::error title=Retry dependency canary check::Could not verify that the shrinkwrap canaries still discriminate after ${REGISTRY_QUERY_ATTEMPTS} registry query attempts for ${failedQueries.join(', ')}. Retry this job; if the error persists, check npm registry availability and confirm the listed package ranges match published versions.`
 		);
 		failed = true;
 		return;
--- a/build-tools/check-shrinkwrap-pins.mjs
+++ b/build-tools/check-shrinkwrap-pins.mjs
@@ -33,6 +33,7 @@ import { execFileSync } from 'node:child_process';
 import { readFileSync } from 'node:fs';
 
 const CHECKED_DEPS = ['@harperfast/rocksdb-js', 'fastify', '@aws-sdk/client-s3'];
+const REGISTRY_QUERY_ATTEMPTS = 3;
 
 const pkgRoot = process.argv[2];
 if (!pkgRoot) {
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -178,8 +178,8 @@
 		"@fastify/compress": "^8.3.1",
 		"@fastify/cors": "^11.2.0",
 		"@fastify/static": "^9.1.3",
-		"@harperfast/extended-iterable": "^1.0.1",
-		"@harperfast/rocksdb-js": "^2.7.1",
+		"@harperfast/extended-iterable": "1.0.3",
+		"@harperfast/rocksdb-js": "2.7.1",
```
cb1kenobi, line 182: **Medium: this pin makes the `rocksdb-js` canary in `check-shrinkwrap-pins.mjs` permanently vacuous**

`build-tools/check-shrinkwrap-pins.mjs:33` uses `@harperfast/rocksdb-js` and `fastify` as the two canaries that prove the Docker image honored `npm-shrinkwrap.json` — a blocking step in `docker-smoke.yml:92`. That assertion only discriminates while the shrinkwrap pin lags what the declared range would resolve to. With an exact range the two are equal by construction, so a broken install that re-resolved fresh from `package.json` would also land on `2.7.1` and the check would still pass.

This was already transiently true (`^2.7.1`'s max is `2.7.1` today), but it self-healed on the next rocksdb-js publish. Now it never can — and `update-rocksdb-js.yml` re-pins exact on every bump, so the canary stays dead across future versions. `fastify` becomes the only working canary, and it is one automerged Renovate minor from catching up (`^5.8.2`, lock `5.11.2`, registry `5.12.0`). When it does, `verifyCanariesDiscriminate` fails the build with a message telling the reader to pick a new canary, while rocksdb-js can no longer ever be one.

Suggested fix: in `verifyCanariesDiscriminate`, treat an exact declared range as structurally non-discriminating — skip those deps and fail only if no ranged canary remains — and add a still-ranged dependency to `CHECKED_DEPS` so the check keeps real coverage.

—
_Generated by Barber AI_

kriszyp: Updated the shrinkwrap canary check: exact manifest specs remain in the installed-vs-packed comparison, while only ranged specs participate in discrimination. Added a ranged `@aws-sdk/client-s3` canary and hermetic coverage for exact pin checking, all-exact failure, and all-ranged-caught-up failure.\n\nCo-Authored-By: GPT-5 Codex <noreply@openai.com>

cb1kenobi:
Re-reviewed `da2acbc2` (incremental since `35d07599`) — no new issues found.

The prior Medium finding is resolved: `@aws-sdk/client-s3` (`^3.1012.0`) joins the canary set as a genuinely discriminating one, the discrimination check now skips exact manifest specs instead of counting them, and an all-exact canary set fails loudly with a clear "add a ranged canary" error. The new `unitTests/build-tools/checkShrinkwrapPins.test.mjs` covers all five paths and is picked up by `test:unit:main`; all 5 pass locally.

Worth noting why the new canary holds up: Renovate's `minimumReleaseAge: "7 days"` combined with the AWS SDK's near-daily publish cadence means the lock pin can't realistically catch up to max-in-range, so this canary is structurally more durable than `fastify`.

This PR looks good, nice job!

—
_Generated by Barber AI_

cb1kenobi (approved):

heskew (approved):

build-tools/check-shrinkwrap-pins.mjs
```diff
@@ -85,52 +90,140 @@ for (const dep of CHECKED_DEPS) {
...
+		for (let attempt = 1; attempt <= REGISTRY_QUERY_ATTEMPTS; attempt++) {
+			try {
+				// `npm view <dep>@<range> version --json` returns every matching version, not
+				// just the max, and the order isn't a documented contract (it can track
+				// publish/insertion order rather than semver order, e.g. a backported patch
+				// published after a newer minor) -- compare them ourselves rather than trust
+				// the last array entry.
+				const out = execFileSync('npm', ['view', `${dep}@${range}`, 'version', '--json'], { encoding: 'utf8' });
+				const versions = JSON.parse(out);
+				if (Array.isArray(versions) && versions.length === 0) {
```
cb1kenobi, line 133: **Low: the "matches no published version" branch is unreachable against real npm**

`npm view <dep>@<range> version --json` does not exit 0 with `[]` when a range matches nothing — it exits **1** with an `E404` ("No match found for version"), so `execFileSync` throws and control goes to the `catch` instead of reaching this branch. Verified against npm 11.13.0 by running this script unmodified against the real registry with `fastify` declared as `^99.0.0`:

```
::warning::registry query attempt 1/3 failed for fastify@^99.0.0 ...
::warning::registry query attempt 2/3 failed ...
::warning::registry query attempt 3/3 failed ...
::error title=Retry dependency canary check::Could not verify that the shrinkwrap canaries still discriminate after 3 registry query attempts for fastify@^99.0.0. Retry this job; ...
```

So the one condition this branch exists to name — a declared range that can never resolve — is reported as a transient infrastructure problem telling the operator to retry a job that can never go green, after burning 3 attempts and ~3s of backoff. The test at `unitTests/build-tools/checkShrinkwrapPins.test.mjs:234` ("fails without retries when a registry query matches no published version") passes only because the stub `npm` models a no-match as exit 0 + `[]`, which real npm never emits.

Non-blocking: the check still fails closed either way, and the retry text does end with "confirm the listed package ranges match published versions".

Suggested fix: treat an `E404` seen in the `catch` as definitive and skip the remaining attempts — npm writes `{"error":{"code":"E404",...}}` to stdout, so `e.stdout` can be inspected — and change the fixture's stub `npm` to exit 1 with that payload so the test exercises the shape npm actually produces.

—
_Generated by Barber AI_

kriszyp: Addressed in 3e2adc81e: the checker now recognizes npm’s exit-1 E404 JSON response and fails immediately with the permanent range/registry diagnosis. The fixture models that real process shape; empty-array responses remain definitive, while malformed successful payloads retry and fail closed. The focused suite now has 15 passing tests.

— GPT-5 Codex

cb1kenobi:
Reviewed `3e2adc8` — no issues found. This PR looks good, nice job!

—
_Generated by Barber AI_

dawsontoth (approved):

Issue #2180, opened 2026-08-15
HNSW: efConstruction does not scale with corpus size — recall plateaus at 1M vectors regardless of search ef


kriszyp:
**Priority: P1**

## Summary

`efConstruction` defaults to 100 and never scales with corpus size, while the search-side `ef` auto-scales (`AUTO_EF_BASE`/`AUTO_EF_MAX` in `HierarchicalNavigableSmallWorld.ts`). At 1M vectors this becomes the recall bottleneck: recall@10 drops to 0.935 and **cannot be recovered by any amount of search-side `ef`** — the missing neighbours are unreachable in the graph, because the edge-selection candidate list stayed at 100 while the corpus grew 10×.

Found while extending #2125's scaling benchmark to 1M vectors ([full data in the PR thread](https://github.com/HarperFast/harper/pull/2125)).

## Evidence

All runs on `benchmarks/hnsw-scale.js` (768-dim, int8, calibrated Gaussian-mixture corpus at `d(nn)/d(rand)` ≈ 0.65 — the harder-contrast regime, matching the corpus the reporter of #2069 measured). Fixed seed, so the A/B graphs below index the identical corpus with identical level assignments.

Recall@10 (raw / candidate-set) across the sweep, `ef` = auto (pinned at `AUTO_EF_MAX` 512 from ~13K nodes up):

| N | 100K | 250K | 500K | 1M |
|---|---|---|---|---|
| recall@10 | 0.980 / 0.997 | 0.978 / 0.988 | 0.978 / 0.987 | **0.935 / 0.943** |

Ruling out the search side — sweeping `ef` over the *same* 1M graph:

| ef | 512 | 768 | 1024 | 1536 |
|---|---|---|---|---|
| recall@10 set | 0.943 | 0.962 | 0.963 | 0.967 |
| p50 | 8.7ms | 18.1ms | 25.6ms | 41.8ms |

Recall plateaus ~0.96 at 4.7× the latency: not an `ef` ceiling.

Attribution at 1M, `ef` 512:

| build config | recall@10 raw / set | p50 | visited | build |
|---|---|---|---|---|
| efC=100, int8 (default) | 0.935 / 0.943 | 8.9ms | 3,948 | 63 min |
| efC=100, float32 | 0.952 / 0.952 | — | 3,596 | 71 min |
| **efC=200, int8** | **0.985 / 0.997** | **7.2ms** | 3,110 | 110 min |

[...]

PR #2181, opened 2026-08-15
Auto-scale efConstruction and the search-ef ceiling with graph size for 1M+ node HNSW graphs

kriszyp:
Large HNSW graphs lose recall because a fixed construction candidate list eventually stops creating enough useful edges. This change auto-scales `efConstruction` from a base of 100 as `min(1024, 100 × sqrt(nodes / 250K))`, resumes the search-`ef` scale above one million nodes up to 2048, and adds streaming benchmark support for multi-million-node validation.

The write path reads the lifetime node-id high-water mark directly from the shared atomic counter. If an update-only worker cannot attach that counter, it warns once, reuses its transaction-consistent reverse seek as a memoized fallback, and retries after the memo TTL. New-id allocation still fails closed rather than installing a private counter.

## Evidence and limits

The original 1M A/B rebuilt the same corpus uniformly at `efConstruction: 200`: set-recall improved from 0.967 to 0.997 and the better graph also reduced nodes visited at the same search `ef`. That isolates the construction-quality problem, but it is not a direct comparison of the automatic ramp against a uniform-200 build.

The automatic ramp has separate 2M/5M evidence. With search `ef: 1024`, those graphs reached set-recall 0.988/0.985. That shows the measured neighbours remained reachable at those sizes; it does not prove that the ramp converges to the uniform-build result.

Below 250K nodes the construction default remains 100. Nodes inserted before a scale threshold keep their existing edges. A live-process reindex starts near the lifetime high-water mark, while a reindex after restart repeats the ramp from the store's current largest id.

## Search cost

Past 1M nodes, automatic search `ef` resumes from 512 as `512 × sqrt(nodes / 1M)`, capped at 2048. At 5M nodes it resolves to 1,145; the measured `ef: 1024` point held 0.985 set-recall at about 45 ms p50 on the benchmark corpus.

Limit-derived widening to 4096 is now limited to automatically scaled indexes. An explicit `efConstructionSearch`, an `efConstruction` value that seeds the search default, or a per-query `ef` remains an authoritative cost ceiling; a pinned index can therefore return fewer rows than `limit` unless the query supplies a larger `ef`.

Filtered traversal does not inherit the entire automatic increase. Its visit-budget contribution remains capped at the former automatic ceiling of 512 before multiplying by `filterExpansion`; explicit schema or per-query `ef` still raises the budget for callers that own the cost.

## Construction and delete cost

The node count is a lifetime id high-water mark, not a live count. Delete-heavy tables can therefore remain over-provisioned after the live graph shrinks. When `efConstruction` exceeds the live graph size, an insert can traverse much of the graph without gaining recall. Connectivity repair can also synchronously reinsert an orphan plus up to 256 nodes from a severed island, and each repair uses the current scaled construction value.

An explicit `efConstruction` remains authoritative and structural, so changing it rebuilds the index. It also seeds the search default. To retain the former large-graph search default while opting out of build scaling, configure both `efConstruction: 100` and `efConstructionSearch: 512`; there is no “pinned build, automatic search” mode.

## Benchmark tooling

`benchmarks/hnsw-scale.js` gains a `--stream` mode for corpora too large to hold as one float pool. It uses a dedicated replayable corpus RNG, gathers query rows in one pass, then streams the same rows through graph construction and brute-force ground truth in a second pass. Each result row records `efConstructionStart` and `efConstructionEnd`, so a ramped build cannot be confused with a uniform build that ended at the same value.

## For the human reviewer

1. **Schema and limit ceilings:** the owner selected 1C. Schema and per-query `ef` values cap traversal and may produce a short page; only automatically scaled indexes widen from `offset + limit` toward 4096. Reversing this is a localized guard/constant change but would reintroduce either pin bypass or a global 2048 result ceiling.
2. **High-water autoscaling:** the owner selected 2A. Lifetime allocated IDs continue to drive construction/search scaling; [#2182](https://github.com/HarperFast/harper/issues/2182) tracks a live-count design for churn-heavy tables rather than broadening this PR into transactional multi-worker bookkeeping.
3. **Ramped construction:** a restart/reindex grows efC during the rebuild rather than sizing the entire rebuild from its final count. The 2M/5M measurements show reachable neighbours but do not prove uniform-build equivalence; changing this later requires rebuilding graph shape.
4. **Recall-first search:** the default search budget grows past one million nodes rather than retaining the 512 latency plateau. Applications can reverse that trade-off locally with a schema or query pin.
5. **Filtered-budget split:** automatic filtered traversal remains capped at an ef contribution of 512 even when unfiltered search scales higher. This bounds synchronous record loading at the cost of filtered recall unless callers opt into a larger budget.
6. **Planner cost model:** `estimateCountAsSort` still prices the configured search value rather than runtime autoscaling. This pre-existing heuristic may underprice large automatic indexes; changing it is localized but lacks a workload proving a better ordering.
7. **Public documentation:** the owner selected 3A. Core schema/design documentation is corrected here; the public HarperFast/documentation update remains a separate follow-up and the corresponding review thread stays open.

## Verification

- `npm run build` — passed.
- `npx mocha unitTests/resources/vectorIndex.test.js` — 108 passing.
- Fails-without-fix check — the schema-pin regression failed against the rebuilt prior guard with `3000 !== 1500`, then passed after restoring the fix.
- `npm run test:unit:resources` — 1,569 passing and 15 pending; 3 unrelated random-access-structure/replay failures match the prior PR baseline.
- `npm run test:integration:all` — 1,732 passing, 20 skipped, and 0 asserted failures; 6 Ollama tests were cancelled by the unrelated JSON import-attribute error.
- HNSW integration route — the full integration run's `HNSW vector-index data-integrity` suite passed all 4 scenarios against the rebuilt `dist`.
- Benchmark smoke — a 100-node JSON run emitted `efConstructionStart: 100` and `efConstructionEnd: 100`.
- `npm run format:check` and `npm run lint:required` — passed. Full `npm run lint` reports the same 12 pre-existing warnings outside this diff.
- `npm run test:unit:main` — blocked before test collection by a stale shared test database error while opening `system.mdb`, outside this diff.

Independent pre-push review at `7ac9caa29`: Gemini and policy-selected Cursor Grok both completed; Harper-domain adjudication retained one minor large-scale integration-coverage gap and one comment-style nit, with no blocker or major defect. Claude was intentionally excluded by quota policy.

Addresses [HNSW: efConstruction does not scale with corpus size — recall plateaus at 1M vectors regardless of search ef](https://github.com/HarperFast/harper/issues/2180). Follow-up: [HNSW: graph-size resolution is a lifetime id high-water mark — churn-heavy tables permanently over-provision build/search ef](https://github.com/HarperFast/harper/issues/2182).

Updated by GPT-5 Codex.

Complexity: complicated

<sub>Review-Coverage: authored=codex; ran=gemini,cursor-grok; adjudicated=domain; declined=claude,cursor-composer; rounds=1 @ 7ac9caa2901c</sub>

<sub>Human-Review-Need: 4 (decisions: schema-ef-ceiling, high-water-autoscale, ramped-construction, recall-first-search, filtered-budget-split, planner-cost-model, unprivileged-limit-ceiling) @ 7ac9caa2901c</sub>

### DIFF
--- a/benchmarks/hnsw-scale.js
+++ b/benchmarks/hnsw-scale.js
@@ -11,6 +11,8 @@
  *   node benchmarks/hnsw-scale.js [--n=5000,10000,25000] [--dims=768] [--queries=50]
  *                                 [--clusters=N] [--ef=auto] [--quantization=int8|none]
  *                                 [--upper-ef=N]   (override ef used above layer 0)
+ *                                 [--stream]       (two-pass generation; no float pool held — for N
+ *                                                   where the pool alone would not fit in memory)
  *                                 [--json=path]
  */
 
@@ -55,6 +57,8 @@ const UPPER_EF =
 // median pairwise distance is ~0.37, not ~0.95), so any ef/recall conclusion has to be confirmed on
 // real vectors before it is trusted.
 const CORPUS = argv.corpus ? String(argv.corpus) : undefined;
+const STREAM = argv.stream !== undefined;
+if (STREAM && CORPUS) throw new Error('--stream generates synthetically; it cannot be combined with --corpus');
 const INTRA_COS = Number(argv['intra-cos'] ?? 0.75);
 const NOISE = argv.noise !== undefined ? Number(argv.noise) : Math.sqrt((1 / (INTRA_COS * INTRA_COS) - 1) / DIMS);
 
@@ -210,6 +214,148 @@ function rowToArray(pool, i, dims) {
 	return out;
 }
 
+/** gauss() over a caller-owned RNG, so a stream's draws are independent of Math.random. */
+function gaussStream(rand) {
+	let spare = null;
+	return () => {
+		if (spare !== null) {
+			const s = spare;
+			spare = null;
+			return s;
+		}
+		let u, v, s;
+		do {
+			u = rand() * 2 - 1;
+			v = rand() * 2 - 1;
+			s = u * u + v * v;
+		} while (s === 0 || s >= 1);
+		const mul = Math.sqrt((-2 * Math.log(s)) / s);
+		spare = v * mul;
+		return u * mul;
+	};
+}
+
+/**
+ * Row-at-a-time corpus generator for --stream: same Gaussian mixture as buildCorpus, but driven by
+ * its own RNG stream so re-instantiating it replays the identical rows regardless of how many draws
+ * the index makes from Math.random for level assignment in between. Returns a reused buffer — the
+ * caller copies what it keeps.
+ */
+function corpusStream(n, dims, nClusters, seed) {
+	const rand = mulberry32(seed);
+	const g = gaussStream(rand);
+	const centroids = new Float32Array(nClusters * dims);
+	for (let c = 0; c < nClusters; c++) {
+		let mag = 0;
+		for (let d = 0; d < dims; d++) {
+			const x = g();
+			centroids[c * dims + d] = x;
+			mag += x * x;
+		}
+		mag = Math.sqrt(mag) || 1;
+		for (let d = 0; d < dims; d++) centroids[c * dims + d] /= mag;
+	}
+	const row = new Float32Array(dims);
+	let i = 0;
+	return () => {
+		if (i >= n) return null;
+		const c = (rand() * nClusters) | 0;
+		let mag = 0;
+		for (let d = 0; d < dims; d++) {
+			const x = centroids[c * dims + d] + g() * NOISE;
+			row[d] = x;
+			mag += x * x;
+		}
+		mag = Math.sqrt(mag) || 1;
+		for (let d = 0; d < dims; d++) row[d] /= mag;
+		i++;
+		return row;
+	};
+}
+
+/**
+ * Two-pass streaming build. Pass 1 replays the corpus to collect the query source rows (evenly
+ * spaced, like loadCorpus's holdout) and perturbs them into queries on yet another RNG stream.
+ * Pass 2 replays it again, feeding each row to the index and folding it into the brute-force
+ * ground truth and the corpus diagnostic as it streams by. Peak memory is the graph plus one row;
+ * GT folding time is measured per row and excluded from buildMs.
+ */
+function streamBuild(hnsw, n, dims, nClusters) {
+	const srcIdx = new Set();
+	for (let q = 0; q < N_QUERIES; q++) srcIdx.add(Math.floor((q * n) / N_QUERIES));
+	const sources = [];
+	{
+		const next = corpusStream(n, dims, nClusters, SEED + 1);
+		for (let i = 0; i < n; i++) {
+			const row = next();
+			if (srcIdx.has(i)) sources.push(row.slice());
+		}
+	}
+	const qg = gaussStream(mulberry32(SEED + 2));
+	const queries = sources.map((src) => {
+		const v = Array.from(src);
+		let mag = 0;
+		for (let d = 0; d < dims; d++) {
+			v[d] += qg() * NOISE * 0.5;
+			mag += v[d] * v[d];
+		}
+		mag = Math.sqrt(mag) || 1;
+		for (let d = 0; d < dims; d++) v[d] /= mag;
+		return v;
+	});
+
+	const gtTop = queries.map(() => []); // per query: ascending [{dist, i}], length <= TOP_K
+	const probes = Math.min(20, queries.length);
+	const sepNearBest = new Float64Array(probes).fill(Infinity);
+	let sepRandSum = 0;
+	const next = corpusStream(n, dims, nClusters, SEED + 1);
+	const buildStart = performance.now();
+	let gtMsAcc = 0;
+	let lastLog = buildStart;
+	for (let i = 0; i < n; i++) {
+		const row = next();
+		hnsw.index('r' + i, rowToArray(row, 0, dims), undefined, {});
+		const t0 = performance.now();
+		for (let q = 0; q < queries.length; q++) {
+			const query = queries[q];
+			let dot = 0;
+			for (let d = 0; d < dims; d++) dot += row[d] * query[d];
+			const dist = 1 - dot;
+			const top = gtTop[q];
+			if (top.length < TOP_K || dist < top[top.length - 1].dist) {
+				let at = top.length;
+				while (at > 0 && top[at - 1].dist > dist) at--;
+				top.splice(at, 0, { dist, i });
+				if (top.length > TOP_K) top.pop();
+			}
+			if (i < 3000 && q < probes) {
+				if (dist < sepNearBest[q]) sepNearBest[q] = dist;
+				if (i < 200) sepRandSum += dist;
+			}
+		}
+		gtMsAcc += performance.now() - t0;
+		if ((i + 1) % 250_000 === 0) {
+			const now = performance.now();
+			console.log(
+				`  … ${i + 1}/${n} indexed, ${((now - buildStart) / 60000).toFixed(1)}min elapsed, ` +
+					`${(250_000 / ((now - lastLog) / 1000)).toFixed(0)} rows/s this chunk`
+			);
+			lastLog = now;
+		}
+	}
+	const buildMs = performance.now() - buildStart - gtMsAcc;
+	let sepNear = 0;
+	for (let p = 0; p < probes; p++) sepNear += sepNearBest[p];
+	return {
+		buildMs,
+		queries,
+		groundTruth: gtTop.map((top) => new Set(top.map((t) => t.i))),
+		gtMs: gtMsAcc,
+		sepNear: sepNear / probes,
+		sepRand: sepRandSum / (probes * 200),
+	};
+}
+
 /** Ground truth over unit-normalized rows: cosine distance = 1 - dot. */
 function bruteForceTopK(pool, n, dims, query, k) {
 	const dist = new Float64Array(n);
@@ -302,7 +448,11 @@ for (const N of SIZES) {
 	global.gc?.();
 	Math.random = mulberry32(SEED); // identical corpus + level assignments for every configuration
 	_spare = null;
-	const { pool, queryPool } = CORPUS ? loadCorpus(CORPUS, N, DIMS, N_QUERIES) : buildCorpus(N, DIMS, nClusters);
+	const { pool, queryPool } = STREAM
+		? {}
+		: CORPUS
+			? loadCorpus(CORPUS, N, DIMS, N_QUERIES)
+			: buildCorpus(N, DIMS, nClusters);
 
 	const store = new MemoryStore();
 	const options = { distance: 'cosine', quantization: QUANTIZATION };
@@ -313,38 +463,46 @@ for (const N of SIZES) {
 	const hnsw = new HierarchicalNavigableSmallWorld(store, options);
 	instrument(hnsw);
 
-	const buildStart = performance.now();
-	for (let i = 0; i < N; i++) hnsw.index('r' + i, rowToArray(pool, i, DIMS), undefined, {});
-	const buildMs = performance.now() - buildStart;
+	let buildMs;
+	let queries = [];
+	let groundTruth;
+	let gtMs = 0;
+	let sepNear = 0;
+	let sepRand = 0;
+	if (STREAM) {
+		({ buildMs, queries, groundTruth, gtMs, sepNear, sepRand } = streamBuild(hnsw, N, DIMS, nClusters));
+	} else {
+		const buildStart = performance.now();
+		for (let i = 0; i < N; i++) hnsw.index('r' + i, rowToArray(pool, i, DIMS), undefined, {});
+		buildMs = performance.now() - buildStart;
+	}
 
 	const shape = graphShape(store);
 
-	// queries drawn from the same distribution (perturbed corpus rows)
-	const queries = [];
-	for (let q = 0; q < N_QUERIES; q++) {
-		if (queryPool) {
-			// real corpus: a held-out embedding, never indexed
-			queries.push(rowToArray(queryPool, q, DIMS));
-			continue;
+	// queries drawn from the same distribution (perturbed corpus rows); --stream built its own above
+	if (!STREAM)
+		for (let q = 0; q < N_QUERIES; q++) {
+			if (queryPool) {
+				// real corpus: a held-out embedding, never indexed
+				queries.push(rowToArray(queryPool, q, DIMS));
+				continue;
+			}
+			// synthetic: perturb an indexed row so rank 1 is not trivially the query itself
+			const src = (Math.random() * N) | 0;
+			const v = rowToArray(pool, src, DIMS);
+			let mag = 0;
+			for (let d = 0; d < DIMS; d++) {
+				v[d] += gauss() * NOISE * 0.5;
+				mag += v[d] * v[d];
+			}
+			mag = Math.sqrt(mag) || 1;
+			for (let d = 0; d < DIMS; d++) v[d] /= mag;
+			queries.push(v);
 		}
-		// synthetic: perturb an indexed row so rank 1 is not trivially the query itself
-		const src = (Math.random() * N) | 0;
-		const v = rowToArray(pool, src, DIMS);
-		let mag = 0;
-		for (let d = 0; d < DIMS; d++) {
-			v[d] += gauss() * NOISE * 0.5;
-			mag += v[d] * v[d];
-		}
-		mag = Math.sqrt(mag) || 1;
-		for (let d = 0; d < DIMS; d++) v[d] /= mag;
-		queries.push(v);
-	}
 
 	// corpus diagnostic: how separated is the true nearest neighbour from a random row?
 	// (if these are close, the corpus is effectively uniform and no ANN can help)
-	let sepNear = 0;
-	let sepRand = 0;
-	{
+	if (!STREAM) {
 		const probes = 20;
 		for (let p = 0; p < probes; p++) {
 			const q = queries[p % queries.length];
@@ -361,11 +519,11 @@ for (const N of SIZES) {
 		}
 		sepNear /= probes;
 		sepRand /= probes * 200;
-	}
 
-	const gtStart = performance.now();
-	const groundTruth = queries.map((q) => bruteForceTopK(pool, N, DIMS, q, TOP_K));
-	const gtMs = performance.now() - gtStart;
+		const gtStart = performance.now();
+		groundTruth = queries.map((q) => bruteForceTopK(pool, N, DIMS, q, TOP_K));
+		gtMs = performance.now() - gtStart;
+	}
 
 	for (const efSpec of EF_SWEEP) {
 		const queryEf = efSpec === 'auto' ? undefined : Number(efSpec);
--- a/benchmarks/hnsw-scale.js
+++ b/benchmarks/hnsw-scale.js
@@ -10,10 +10,14 @@
  * Run (after npm run build):
  *   node benchmarks/hnsw-scale.js [--n=5000,10000,25000] [--dims=768] [--queries=50]
  *                                 [--clusters=N] [--ef=auto] [--quantization=int8|none]
+ *                                 [--ef-construction=N] [--ef-sweep=auto,N,...]
+ *                                 [--intra-cos=X] [--seed=N]
  *                                 [--upper-ef=N]   (override ef used above layer 0)
  *                                 [--stream]       (two-pass generation; no float pool held — for N
  *                                                   where the pool alone would not fit in memory)
  *                                 [--json=path]
+ * `--ef-construction` also seeds the default search ef. Use the same explicit `--ef-sweep`
+ * across construction A/B runs to compare both graphs at a common search ef.
  */
 
 const { performance } = require('node:perf_hooks');
@@ -377,6 +381,7 @@ function pct(sorted, p) {
 
 let inSearch = false;
 let currentEf = 0;
+let currentEfConstruction = 0;
 const layerStats = { visitsByLevel: new Map(), timeByLevel: new Map(), callsByLevel: new Map() };
 function resetLayerStats() {
 	layerStats.visitsByLevel.clear();
@@ -405,6 +410,7 @@ function instrument(hnsw) {
 		// query — the same value, and read from the index rather than recomputed here, so this cannot
 		// drift when the auto-scale formula changes.
 		if (inSearch && level === 0) currentEf = ef;
+		else if (!inSearch && level === 0) currentEfConstruction = ef;
 		if (inSearch && UPPER_EF !== undefined && level > 0 && currentEf) ef = UPPER_EF === 'match' ? currentEf : UPPER_EF;
 		else if (inSearch && UPPER_EF !== undefined && UPPER_EF !== 'match' && level > 0) ef = UPPER_EF;
 		else if (!inSearch && BUILD_UPPER_EF !== undefined && level > 0) ef = BUILD_UPPER_EF;
@@ -438,6 +444,7 @@ function graphShape(store) {
 const rows = [];
 console.log(
 	`\nHNSW scaling sweep — dims=${DIMS}, queries=${N_QUERIES}, k=${TOP_K}, quantization=${QUANTIZATION}, ef=${EF_OPT}` +
+		`, efC=${EF_CONSTRUCTION ?? 'auto'}, efSweep=${EF_SWEEP.join(',')}` +
 		(UPPER_EF !== undefined ? `, upper-ef=${UPPER_EF}` : '') +
 		(BUILD_UPPER_EF !== undefined ? `, build-upper-ef=${BUILD_UPPER_EF}` : '') +
 		`, intraCos=${INTRA_COS}, sigma=${NOISE.toFixed(4)}\n`
@@ -462,6 +469,7 @@ for (const N of SIZES) {
 	if (EF_CONSTRUCTION !== undefined) options.efConstruction = EF_CONSTRUCTION;
 	const hnsw = new HierarchicalNavigableSmallWorld(store, options);
 	instrument(hnsw);
+	currentEfConstruction = 0;
 
 	let buildMs;
 	let queries = [];
@@ -478,6 +486,7 @@ for (const N of SIZES) {
 	}
 
 	const shape = graphShape(store);
+	const effectiveEfConstruction = currentEfConstruction || hnsw.efConstruction;
 
 	// queries drawn from the same distribution (perturbed corpus rows); --stream built its own above
 	if (!STREAM)
@@ -563,6 +572,7 @@ for (const N of SIZES) {
 			clusters: nClusters,
 			buildMs: Math.round(buildMs),
 			msPerInsert: +(buildMs / N).toFixed(2),
+			efConstruction: effectiveEfConstruction,
 			efSpec: String(efSpec),
 			ef: effectiveEf,
 			p50: +pct(latencies, 50).toFixed(2),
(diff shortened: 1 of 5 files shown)


kriszyp:
## 2M and 5M validation runs (streaming mode, this branch)

Same harness and calibration as the 100K–1M sweep on #2125 (768-dim, int8, `intra-cos` 0.6, 60 held-out-style queries, exact brute-force ground truth folded in during the stream). Auto-scaled `efConstruction` ramps to ~283 by 2M and ~447 by 5M. Both runs also sweep `ef` 1024 over the same graph to separate build quality from the search-side cap.

| N | build | auto efC (final) | recall@10 raw/set @ ef 512 | @ ef 1024 | p50 @ 512 / 1024 | peak RSS |
|---|---|---|---|---|---|---|
| 1M (efC 200 A/B, from #2180) | 110 min | 200 | 0.985 / 0.997 | — | 7.2ms / — | 9.4 GB |
| 2M | 3.7 h | ~283 | 0.950 / 0.955 | 0.983 / 0.988 | 10.4ms / 30.4ms | 12.3 GB |
| 5M | 19.3 h | ~447 | 0.920 / 0.935 | 0.968 / 0.985 | 22.2ms / 45.1ms | 27.6 GB |

**What this says about the fix.** The construction side is doing its job: at 5M, `ef` 1024 on the same graph reaches 0.985 set-recall, so the true neighbours are present and reachable — the opposite of the #2180 failure mode, where no `ef` could recover them (0.967 set at `ef` 1536 on a 5× smaller graph). The build-time cost of the ramp is real and as documented: 13.9 ms/insert at 5M vs 3.8 at 1M-unfixed.

**What binds next is the search side.** With good graphs, recall at the auto-resolved `ef` (pinned at `AUTO_EF_MAX` 512 from ~13K nodes) decays with N — 0.997 → 0.955 → 0.935 set across 1M → 2M → 5M — and doubling `ef` recovers it at roughly 2× latency (45 ms p50 at 5M, still tens of ms). That is #2125's deferred auto-scale-constants question resurfacing at the top of the range: the 512 search cap is the right latency-first default for most tables, but 2M+ tables have recall on the table that only a raised (or scaled) `AUTO_EF_MAX` — or an explicit `efConstructionSearch` / per-query `ef` — can claim. Not changed in this PR; it is a search-side default decision, and per-query `ef` already gives callers the knob.

Two caveats. The 5M query latencies carry memory-pressure noise (27.6 GB RSS on a 30 GB box — page-cache squeeze and poor locality inflate per-visit cost; visited only grew 1.2× while p50 grew 2.1× from 2M). And the printed `d(nn)/d(rand)` drift (0.75, 0.80) is the known first-3000-rows probe artifact, not the corpus hardening — construction keeps per-cluster density constant across N.

Raw logs/JSON: `~/dev/tmp/hnsw-1mm/n2000000-fixed.*`, `n5000000-fixed.*` on the benchmark box.

Generated by Claude Fable 5.

cb1kenobi:
Reviewed `c6784e5` — no issues found. This PR looks good, nice job!

—
_Generated by Barber AI_

resources/indexes/HierarchicalNavigableSmallWorld.ts
```diff
@@ -375,9 +389,27 @@ export class HierarchicalNavigableSmallWorld {
...
-			// Connect the new element to neighbors at its level and below
+			// Ensured here (not only at id allocation) so an update-only worker reads the count with an
+			// atomic load instead of a per-write seek. The count is a lifetime high-water mark, not a
+			// live count; see DESIGN.md for the churn-table caveat. Guarded because the count is an ef
+			// heuristic with a safe fallback — a write must never fail on it.
+			let efConstruction = this.efConstruction;
+			if (!this.efConstructionConfigured) {
+				try {
+					this.ensureIdIncrementer(options);
+					efConstruction = autoScaleEfConstruction(this.efConstruction, this.resolveNodeCount());
```
cb1kenobi: **High: the recall result that justifies this change was measured on a uniform `efConstruction` 200, but the shipped path builds a ramp**

`efConstruction` is re-resolved on every `index()` call from a node count that grows *during* the build, so a table growing 0→1M gets efC 100 for its first 250K nodes and a sqrt ramp after that — a build-weighted average of ~142, not 200. The 0.935 → 0.985 measurement that motivates the whole PR came from rebuilding at a constant `--ef-construction=200`, which is a configuration the auto-scale never produces on a growing table. DESIGN.md discloses the ramp but then asserts "Both converge to the same steady state as the graph grows past the knee" with no measurement behind it. Reverse edges from later inserts do refresh early nodes' neighbourhoods, so convergence is plausible — it is just unverified, and it is the load-bearing claim.

Suggested fix: run `benchmarks/hnsw-scale.js --n=1000000` on this branch with **no** `--ef-construction` (which exercises the shipped ramp) and report that recall next to the uniform-200 number. If the ramped recall lands materially below 0.985, say so in DESIGN.md instead of asserting convergence.

—
_Generated by Barber AI_

resources/indexes/HierarchicalNavigableSmallWorld.ts
```diff
@@ -375,9 +389,27 @@ export class HierarchicalNavigableSmallWorld {
 				connections[i] = [];
 			}
 
-			// Connect the new element to neighbors at its level and below
+			// Ensured here (not only at id allocation) so an update-only worker reads the count with an
+			// atomic load instead of a per-write seek. The count is a lifetime high-water mark, not a
+			// live count; see DESIGN.md for the churn-table caveat. Guarded because the count is an ef
+			// heuristic with a safe fallback — a write must never fail on it.
+			let efConstruction = this.efConstruction;
```
cb1kenobi, line 392: **Medium: the churn caveat's cost bound is understated — an ef far above the live graph size degenerates the stop rule, it does not just "waste build CPU"**

DESIGN.md bounds the high-water-mark churn case at `AUTO_EFC_MAX` and "roughly 6–7x build time", derived from the measured 1.77x-wall-time-per-2x-ef relation. That relation was measured on graphs large enough to justify the ef. On a rolling-window table (say 10K live rows but 26M lifetime ids from TTL eviction) every insert builds at efC 1024 on a 10K-node graph, and the unfiltered stop rule at line 926 (`current.distance > furthestDistance`) cannot fire until `results` holds `ef` entries — so the connection pass traverses a large fraction of the entire graph, with a synchronous `safeGetSync` + distance per node, to produce candidates that line 413 then truncates to `M << 1` (32). The cost is not 6–7x there, and by the PR's own data (efC 100 held recall through 500K) it buys nothing.

This is the case #2182 is meant to fix, so the ask here is the framing, not the code: state the churn cost in throughput terms for the small-live-graph case rather than as a bounded build-time multiple, so #2182 reads as a prerequisite rather than a nice-to-have. A cheap interim clamp is a second slot in the existing `next-id` shared buffer (+1 on insert, −1 on delete) to scale off a live size.

—
_Generated by Barber AI_

resources/indexes/HierarchicalNavigableSmallWorld.ts
```diff
@@ -69,10 +82,24 @@ function autoScaleEf(nodeCount: number): number {
...
 // Ceiling on the ef a query's own `offset + limit` can ask for. `limit` is unprivileged and set on
 // every request, so this bounds what a caller can make one thread do synchronously: layer 0 holds
-// `ef` candidates in a sorted array with an O(len) insert. Kept within a small multiple of
-// AUTO_EF_MAX so the worst case stays the same order as the index's own auto-scaled ceiling; a
-// caller who genuinely wants more sets an explicit `ef` and owns the cost.
-const LIMIT_EF_MAX = 4 * AUTO_EF_MAX;
+// `ef` candidates in a sorted array with an O(len) insert. Kept within a small multiple of the
+// index's own auto-scaled ceiling so the worst case stays the same order; a caller who genuinely
+// wants more sets an explicit `ef` and owns the cost.
+const LIMIT_EF_MAX = 2 * AUTO_EF_CEILING;
```
cb1kenobi, line 88: **Medium: `LIMIT_EF_MAX` doubles to 4096 for every graph, but the ceiling that justifies it only binds at ~16M nodes**

`minResults` is `offset + limit` — unprivileged and set on every request (`resources/search.ts:513` → line 1104-1105), with no clamp, rate limit, or per-query CPU budget anywhere in that chain. The unfiltered layer-0 path has **no** `maxVisits` at all (`filterState` is `undefined` without a filter), so `ef` alone sizes the work: one `safeGetSync` + a `dims`-wide distance per visited node, plus an O(len) `results.splice()` per admission into an array of up to 4096. The comment's "kept within a small multiple of the index's own auto-scaled ceiling" only holds once `autoScaleEf` actually reaches 2048 — at 1M nodes this is 8x the ef the index resolves for itself, at 10K nodes 40x. A schema-level `efConstructionSearch` pin does not block the widening either: line 1104 checks only `!explicitEf`.

Suggested fix: anchor the limit ceiling to what this index resolves rather than to a ceiling that binds at 16M — e.g. `Math.min(2 * AUTO_EF_CEILING, 4 * resolvedEf)` — and let a schema-configured `efConstructionSearch` cap the widening the way an explicit per-query `ef` does.

Separately, `estimateCountAsSort()` (line 1368) still estimates against `this.efConstructionSearch` (default 50). That gap is pre-existing, but raising the real ef to 1145/2048 widens the planner's under-estimate to ~40x on exactly the large tables this PR targets.

—
_Generated by Barber AI_

schema.graphql
```diff
@@ -188,7 +188,8 @@ directive @indexed(
 	distance: String
 	"""
 	Construction effort/recall parameter (HNSW). Higher values improve recall at
-	the cost of build time and memory.
+	the cost of build time and memory. When omitted, it auto-scales with graph
```
cb1kenobi, line 191: **Medium: two user-visible defaults change with no update to the public docs**

The published reference (`HarperFast/documentation` → `reference/database/schema.md`) still lists `efConstruction | 100 | Max nodes explored during index construction`, which is false for any table past 250K nodes after this lands, and says nothing about the default query `ef` now scaling past 1M to 2048. That repo already has a `<VersionBadge type="changed" version="..." />` convention, used for the *previous* ef auto-scale change on this same page. The harper repo has no CHANGELOG or release-note file, so the docs repo is the only operator-facing channel for a change that silently alters p50 latency on existing >1M-node tables.

Suggested fix: land a companion documentation PR — correct the `efConstruction` default cell to "auto-scaled (base 100)", add `VersionBadge type="changed"` entries for both the build-side scale and the second search-`ef` regime, and state the latency implication plus `efConstructionSearch` as the opt-out.

—
_Generated by Barber AI_

resources/indexes/HierarchicalNavigableSmallWorld.ts
```diff
@@ -212,8 +241,9 @@ export class HierarchicalNavigableSmallWorld {
 			this.indexStore.encoder.useFloat32 = FLOAT32_OPTIONS.ALWAYS;
 		}
 		this.int8 = options?.quantization !== 'none';
-		// Respect an explicitly-configured search ef (or efConstruction, which seeds it); otherwise auto-scale.
+		// Respect an explicitly-configured ef (efConstruction seeds the search ef too); otherwise auto-scale both.
 		this.efSearchConfigured = options?.efConstructionSearch !== undefined || options?.efConstruction !== undefined;
+		this.efConstructionConfigured = options?.efConstruction !== undefined;
```
cb1kenobi, line 244: **Low: the only opt-out from the build-side auto-scale forces a full reindex, and silently pins the search ef too**

The search-side change is cheaply reversible — `efConstructionSearch` is in `searchOnlyOptions` (line 198), so pinning it persists metadata without a rebuild. The build side is not: `efConstruction` is deliberately excluded, so adding `efConstruction: 100` changes `canonicalIndexKey` → `indexOptionsStructurallyChanged` → `structurallyChanged` (`resources/databases.ts:1833-1861`), which clears `lastIndexedKey` and rebuilds the dbi from scratch — on exactly the multi-million-node table where write throughput was already the problem. And per line 260 that same pin assigns `efConstructionSearch`, so the operator drops the search ef from the auto-scaled 512+ down to 100, which is worse recall than before this PR.

The PR body offers "an operator who wants the old behaviour pins `efConstruction: 100`" as the escape hatch; as written that advice costs a full reindex and a recall regression. Suggested fix: document both consequences next to "an explicit `efConstruction` stays authoritative" in DESIGN.md and in the directive text, and note that restoring the old behaviour needs `efConstruction: 100` **plus** an explicit `efConstructionSearch`.

—
_Generated by Barber AI_

resources/indexes/HierarchicalNavigableSmallWorld.ts
```diff
@@ -375,9 +389,27 @@ export class HierarchicalNavigableSmallWorld {
...
+				if (efConstruction > this.efConstruction && this.lastLoggedEfConstruction !== efConstruction) {
+					// once per resolved value per process: makes replica-divergent build quality and the
+					// build-cost ramp diagnosable (the resolved value is otherwise surfaced nowhere)
+					this.lastLoggedEfConstruction = efConstruction;
+					logger.debug?.(`HNSW construction ef auto-scaled to ${efConstruction}`);
+				}
+			}
 			for (let l = Math.min(level, currentLevel); l >= 0; l--) {
-				let neighbors = this.searchLayer(vector, entryPointId, entryPoint, this.efConstruction, l, options);
+				let neighbors = this.searchLayer(vector, entryPointId, entryPoint, efConstruction, l, options);
```
cb1kenobi, line 408: **Low: one delete can fan out to ~257 synchronous reinserts, each now at the auto-scaled `efConstruction`**

`repairSeveredNeighbors` reindexes every member of a severed island by calling `this.index(...)` synchronously inside the delete (line 733), bounded by `PROBE_VISIT_LIMIT` = 256, and `needsReindexing` (line 657) does the same for orphans. Each of those calls now runs its connection pass at the scaled efC rather than a flat 100. DESIGN.md states the cost per-insert ("1.77x at 1M for efC 200"); it does not cover this burst, where the same multiplier lands on up to ~257 inserts inside a single delete.

Suggested fix: either cap the repair path's efConstruction at the base value — repair is about restoring reachability, not edge quality — or call the burst out alongside the per-insert figure in the DESIGN.md cost paragraph.

—
_Generated by Barber AI_

benchmarks/hnsw-scale.js
```diff
@@ -302,7 +448,11 @@ for (const N of SIZES) {
 	global.gc?.();
 	Math.random = mulberry32(SEED); // identical corpus + level assignments for every configuration
 	_spare = null;
-	const { pool, queryPool } = CORPUS ? loadCorpus(CORPUS, N, DIMS, N_QUERIES) : buildCorpus(N, DIMS, nClusters);
+	const { pool, queryPool } = STREAM
```
cb1kenobi, line 462: **Low: the harness never records the resolved `efConstruction`, so runs before and after this change are indistinguishable**

After this lands, `--n=1000000` with no `--ef-construction` silently builds at efC 200 instead of 100, but neither the header (line 439-444), the printed row (line 586-591), nor the `row` object (line 560-583) records it — while search `ef` *is* instrumented and reported. Two runs of the identical command across this commit produce different graphs with nothing in the output to distinguish them, and the PR's own baseline arm ("efC=100 default") is no longer what the default command produces.

Reproducing the A/B is also harder than the usage block suggests: `--ef-construction` and `--ef-sweep` are both undocumented there, and both are required, because `--ef-construction=N` sets `efSearchConfigured` and seeds `efConstructionSearch` (`HierarchicalNavigableSmallWorld.ts:245,260`) — so with the default `--ef-sweep=auto` the 100 and 200 arms would be *searched* at ef 100 vs 200 rather than at a common ef.

Suggested fix: add the resolved efConstruction to `row` and the printed line (same way `currentEf` instruments search ef), and list `[--ef-construction=N] [--ef-sweep=a,b,c] [--intra-cos=X] [--seed=N]` in the usage block with a note that `--ef-construction` also pins the search ef.

—
_Generated by Barber AI_

resources/indexes/HierarchicalNavigableSmallWorld.ts
```diff
@@ -375,9 +389,27 @@ export class HierarchicalNavigableSmallWorld {
...
+			// atomic load instead of a per-write seek. The count is a lifetime high-water mark, not a
+			// live count; see DESIGN.md for the churn-table caveat. Guarded because the count is an ef
+			// heuristic with a safe fallback — a write must never fail on it.
+			let efConstruction = this.efConstruction;
+			if (!this.efConstructionConfigured) {
+				try {
+					this.ensureIdIncrementer(options);
+					efConstruction = autoScaleEfConstruction(this.efConstruction, this.resolveNodeCount());
+				} catch (error) {
+					logger.debug?.('could not resolve the node count for the construction ef', error);
```
cb1kenobi: **Low: a persistently failing shared-buffer attach pins `efConstruction` at the base value forever, with only a debug-level trace**

Not installing the counter on a failed attach is the right call (it is what the head commit fixes). But the consequence here is silent: `ensureIdIncrementer` throws, this catch swallows it, and `efConstruction` stays at 100 for the life of the process — the exact recall regression this PR exists to fix — while the only signal is `logger.debug?.`, which is off by default. Every subsequent write also repeats the reverse seek inside `ensureIdIncrementer`, since nothing memoizes the failure.

Suggested fix: log the first failure at `warn` (once, like the `lastLoggedEfConstruction` gate just below) so a stuck attach is diagnosable, and fall back to `this.approximateNodeCount()` in the catch so the scale still applies and the retry seek is bounded by `NODE_COUNT_TTL` instead of running per write.

—
_Generated by Barber AI_

resources/indexes/HierarchicalNavigableSmallWorld.ts
```diff
@@ -69,10 +82,24 @@ function autoScaleEf(nodeCount: number): number {
...
+// Auto-scaled construction ef, used only when an index does not explicitly configure efConstruction.
+// At a constant efConstruction, edge quality erodes as the graph grows until true neighbours become
+// unreachable at ANY search ef; the cap bounds per-insert cost. Measurements and policy in
+// DESIGN.md ("efConstruction and the search-ef ceiling both auto-scale with the graph") and #2180.
+const AUTO_EFC_REF = 250_000;
+// Validated to 447 (the 5M point, where the resulting graph held 0.985 set-recall at ef 1024); the
+// headroom to 1024 is the same sqrt curve extrapolated, binding at ~26M nodes. Build cost grows
+// with the curve (N^1.5 total under sqrt scaling), which is why the cap stays finite: past ~tens of
+// millions of nodes, sharded medium graphs beat one huge graph on both build and query cost.
+const AUTO_EFC_MAX = 1024;
```
cb1kenobi, line 98: **Nit: the PR description contradicts the code in four places**

Not code issues, but this is the summary a reviewer reads:

1. The "What" section gives `efConstruction = min(512, base * sqrt(nodes / 250K))`. The cap is `AUTO_EFC_MAX = 1024` (10.24x base, not 2x). The body self-corrects later ("`AUTO_EFC_MAX` rises to 1024"), so it contradicts itself.
2. "Fails-on-base: ... the cap test reads 100 against expected 512." I ran the new tests against the merge-base source: the cap test reads 100 against expected **1024**. (7 of the 10 new tests fail on base, which does confirm the substance of the claim.)
3. "resolved per insert from the same memoized node count the search side already uses (`approximateNodeCount`, O(1), 10s TTL)" — the write path calls `resolveNodeCount()` directly and bypasses the memo. DESIGN.md gets this right; only the body is stale.
4. "a reindex rebuilds uniformly at the final size's efC" — DESIGN.md, in this same PR, says a reindex *after a restart* re-seeds from the largest surviving id and "therefore repeats the ramp". The body's unqualified version is the one an operator will act on.

Also: the body says "96 passing"; the suite is at 102 passing on this head.

Because of (1), the only build-cost number in the body — "1.77x at 1M" — is the 2x-base point, while the real cap implies the ~6–7x worst case that appears only in DESIGN.md.

—
_Generated by Barber AI_

schema.graphql
```diff
@@ -205,8 +206,9 @@ directive @indexed(
 	"""
 	mL: Int
 	"""
-	Search-time effort/recall parameter used during construction (implementation-
-	specific). Larger values typically yield better accuracy.
+	Search-time effort/recall parameter (HNSW): the candidate-list size used when
+	querying. Larger values typically yield better accuracy at the cost of query
+	latency. When omitted, it auto-scales with graph size.
```
cb1kenobi: **Nit: this description is wrong when `efConstruction` is set, and omits the new latency trade**

"When omitted, it auto-scales with graph size" is stated unconditionally, but omitting `efConstructionSearch` while setting `efConstruction` sets `efSearchConfigured = true` (line 245), which suppresses the auto-scale and pins the search ef to the `efConstruction` value. The entry above documents the pinning, but a reader landing on this field alone gets the opposite answer. It also no longer mentions that past ~1M nodes the auto-scale keeps growing (to 2048) rather than plateauing at 512 — which is the change most likely to surprise an existing deployment.

Suggested fix:
```suggestion
	Search-time effort/recall parameter (HNSW): the candidate-list size used when
	querying. Larger values typically yield better accuracy at the cost of query
	latency. When omitted — and `efConstruction` is also omitted — it auto-scales
	with graph size, continuing to grow past ~1M nodes (up to 2048); pin it if
	query latency matters more than recall on large tables.
```

—
_Generated by Barber AI_

resources/indexes/HierarchicalNavigableSmallWorld.ts
```diff
@@ -69,10 +82,24 @@ function autoScaleEf(nodeCount: number): number {
...
+function autoScaleEfConstruction(base: number, nodeCount: number): number {
```
cb1kenobi: **Nit: `base` is always 100, and the parameterization quietly decouples from the cap**

The only call site passes `this.efConstruction`, and it is reached only when `!this.efConstructionConfigured` — i.e. exactly when `options.efConstruction` was undefined, so `this.efConstruction` is still the class default 100. `base *` and `Math.max(base, scaled)` constant-fold. `AUTO_EFC_MAX` is an absolute 1024, so DESIGN.md's "10.24x the base ef" is only accidentally true: if the default `efConstruction` ever changed, the doc and the cap would silently disagree.

Suggested fix:
```suggestion
function autoScaleEfConstruction(nodeCount: number): number {
	const scaled = Math.round(AUTO_EF_BASE * Math.sqrt(Math.max(1, nodeCount / AUTO_EFC_REF)));
	return Math.min(AUTO_EFC_MAX, Math.max(AUTO_EF_BASE, scaled));
}
```

—
_Generated by Barber AI_

cb1kenobi:
Reviewed `3792c84` for the incremental update from `c6784e5`. No new issues found. The update fixes the previously flagged filtered-budget test gap, shared-counter attach fallback, benchmark instrumentation, schema wording, and docs caveats. Two prior Medium threads remain open: the unprivileged `LIMIT_EF_MAX` widening and the external public-docs update. _Generated by Barber AI_

resources/indexes/HierarchicalNavigableSmallWorld.ts
```diff
@@ -375,9 +387,20 @@ export class HierarchicalNavigableSmallWorld {
 				connections[i] = [];
 			}
 
-			// Connect the new element to neighbors at its level and below
+			// An update-only worker may not have attached the id counter yet. The healthy path is one
+			// atomic load; a failed attach falls back to the memoized seek and retries after its TTL.
+			let efConstruction = this.efConstruction;
+			if (!this.efConstructionConfigured) {
+				efConstruction = autoScaleEfConstruction(this.resolveConstructionNodeCount(options));
```
cb1kenobi: **Low: the "a write must never fail on it" guard was removed; the invariant now rests on validation two methods away**

The previous head wrapped the whole node-count resolution in `try { ... } catch`, with the comment stating the reason outright: the count is an ef heuristic with a safe fallback, so *a write must never fail on it*. This commit removes both the guard and that sentence, and `resolveConstructionNodeCount()` is now called bare.

Every other throw is still contained — `ensureIdIncrementer` is caught inside `resolveConstructionNodeCount`, and `resolveNodeCount`'s seek has its own catch returning 0. The one uncontained path is the healthy fast path: line 751 `if (this.idIncrementer) return this.resolveNodeCount()` → line 804 `Number(Atomics.load(this.idIncrementer, 0))`, which throws `RangeError` if `idIncrementer` is ever a zero-length `BigInt64Array`. Today that state is unreachable *only* because of the new `byteLength` validation added to `ensureIdIncrementer` in this same commit — so a documented write-path invariant went from locally enforced to enforced at a distance by a check in a different method.

I found no reachable failure on this head. The ask is to keep the invariant enforced where it is relied on rather than inferred: restore the `try`/`catch` around this line (falling back to `this.efConstruction`), and keep the sentence that says why it is there.

—
_Generated by Barber AI_

unitTests/resources/vectorIndex.test.js
```diff
@@ -596,6 +596,298 @@ describe('HNSW graph-size resolution (drives the ef auto-scale)', () => {
...
+				seekTransactions.length,
+				2,
+				'the first update should seek once to attach and once to fall back'
+			);
+			assert(seekTransactions[0], 'the counter seed seek must use the write transaction');
+			assert.strictEqual(seekTransactions[1], seekTransactions[0], 'the fallback seek must use the same transaction');
+		} finally {
+			store.getUserSharedBuffer = originalGetUserSharedBuffer;
+			store.getKeys = originalGetKeys;
+			customIndex.idIncrementerRetryAt = 0;
```
cb1kenobi, line 859: **Low: this `finally` restores three things but not the five the test saved, so one assertion failure cascades through the rest of the file**

The test saves five fields up front — `idIncrementer`, `nodeCount`, `nodeCountAt`, `idIncrementerRetryAt`, `idIncrementerFailureLogged` — but this first `finally` only restores `getUserSharedBuffer`, `getKeys`, and `idIncrementerRetryAt`. All five are restored in the *second* `try`'s `finally`, which is never reached if any of the seven assertions above (lines 824-834) throws.

On a genuine failure the index is left with `idIncrementer === undefined` (the real shared counter dropped), `nodeCountAt === 0`, and `idIncrementerFailureLogged === true`, which leaks into every later test in the file — so the one test guarding the degraded attach path is the one whose failure is hardest to read. The sibling test at line 853 ('does not cache an unusable shared-counter buffer') gets this right with a single `try`/`finally`.

Suggested fix: wrap both phases in one `try`/`finally` that restores all five saved fields plus the two store methods, so a failing assertion reports itself and nothing else.

—
_Generated by Barber AI_

resources/indexes/HierarchicalNavigableSmallWorld.ts
```diff
@@ -717,19 +739,77 @@ export class HierarchicalNavigableSmallWorld {
...
+				return this.resolveNodeCount();
+			} catch (error) {
+				this.idIncrementerRetryAt = now + NODE_COUNT_TTL;
+				if (!this.idIncrementerFailureLogged) {
+					this.idIncrementerFailureLogged = true;
+					logger.warn?.('could not attach the shared HNSW id counter; using a memoized node count', error);
+				}
+			}
+		}
+		return this.approximateNodeCount(options);
```
cb1kenobi, line 772: **Nit: the attach-failure path does two reverse seeks where one would do, and the new test asserts the waste as intended behavior**

When the attach fails, `ensureIdIncrementer` has already performed its reverse seek to find the largest node id before `getUserSharedBuffer` throws; that value is then discarded, and this line does a *second* identical seek via `approximateNodeCount`. The new test bakes it in: `assert.strictEqual(seekTransactions.length, 2, 'the first update should seek once to attach and once to fall back')`.

Only two seeks per `NODE_COUNT_TTL` in a degraded state, so the cost is small — but the seek is the exact thing the memo exists to avoid, and asserting `2` will keep it at 2. Suggested fix: have `ensureIdIncrementer` hand back the id it already found (or seed the memo with it before throwing) so the fallback reuses it, and assert `1`.

—
_Generated by Barber AI_

benchmarks/hnsw-scale.js
```diff
@@ -259,6 +410,7 @@ function instrument(hnsw) {
 		// query — the same value, and read from the index rather than recomputed here, so this cannot
 		// drift when the auto-scale formula changes.
 		if (inSearch && level === 0) currentEf = ef;
+		else if (!inSearch && level === 0) currentEfConstruction = ef;
```
cb1kenobi: **Nit: the new `efConstruction` column records the ramp's terminal value, unlabeled**

`currentEfConstruction` is last-write-wins across the build, so on an auto run it ends up holding the efC of the *final* insert — the top of the ramp, not what most of the graph was built at. A default 1M run and a `--ef-construction=200` run therefore both print `efC= 200` and both write `efConstruction: 200` into `rows[]`; only the header's `efC=auto` and the top-level JSON field separate them. Your own results comment had to hand-label the column "auto efC (final)" for exactly this reason.

This matters more than it would have before: DESIGN.md now rests on the ramp-vs-uniform comparison being *absent* rather than measured, which makes this harness the thing that would eventually close that gap.

Suggested fix: track the first and last level-0 efC seen during the build and emit both (e.g. `efConstructionStart` / `efConstructionEnd`), so a ramped run is self-describing in `rows[]` without cross-referencing the header.

—
_Generated by Barber AI_

PR #2183, opened 2026-08-15
Add review-coverage CI check (report mode)

kriszyp:
## What / why

Adds the **review-coverage** CI check in **report mode** — informational only, never blocks. On every PR open / description edit / push / ready-for-review, it parses the PR description and surfaces:

- how many **cross-model reviews** the description names (`## Review coverage` section, `independent-review:` lines, or explicit cross-model prose), and
- whether the `Human-Review-Need:` footer is present and **current vs the head sha** (a stale footer means commits landed after the last pre-push review).

Under-reported substantive PRs (org-member author, >2 changed lines, non-draft) get a `::warning::` annotation with the count; bots, outside contributors, trivial changes, and drafts are exempt. Always green in this mode.

**Why the action is hosted here:** GitHub bars public repos from using actions in private repos (no setting overrides it), so the JavaScript action (`using: node24`) + parser get their public home in this repo at `.github/actions/review-coverage/`. `reviewGate.mjs` is a header-marked copy of the canonical runtime version in HarperFast/dispatch (the server-side gate imports that one); other public HarperFast repos pin this action by sha (harper-pro#706).

**Context:** the dispatch review fleet now gates substantive member PRs whose AI review finds issues and whose description reports fewer than two cross-model reviews (auto request-changes until coverage is reported). This check makes that expectation visible the moment a PR opens. Flipping to `mode: enforce` (red when under-reported) is deliberately NOT part of this PR — CI can't see the review verdict, so enforcing here is stricter than the gate; that's a team policy call.

## Verification

- This PR dogfoods the check: it reports the two documented review families and verifies the `Human-Review-Need:` footer against the current head.
- 38 unit tests ride along in the action dir (parser + decision matrix + CLI exit codes); run with `node --test .github/actions/review-coverage/*.test.mjs`.
- The action + script were reviewed in dispatch#59 (codex + gemini + cursor-grok + harper-domain adjudication, 2 rounds).

## Review coverage

- Codex: reviewed the action + evaluation script in dispatch#59 (graded leg)
- Gemini: reviewed the action + evaluation script in dispatch#59

— Claude Fable 5 (dispatched by Kris)

<sub>Human-Review-Need: 4 (decisions: attestation-vs-evidence, vendored-byte-copy, harper-as-org-action-host, report-vs-enforce-default, association-as-membership-proxy, staleness-reported-never-enforced, non-member-exemption-renders-as-pass, tests-in-product-unit-test-job) @ a4f543bdf07c</sub>

### DIFF
--- a/.github/actions/review-coverage/action.yml
+++ b/.github/actions/review-coverage/action.yml
@@ -1,0 +1,26 @@
+# Cross-model review-coverage check — the CI surface of the dispatch human-review gate.
+# Parses the PR description with the gate's parser (reviewGate.mjs alongside; canonical
+# runtime copy in HarperFast/dispatch) and reports how many cross-model reviews it names.
+# This public copy lives in harper because GitHub does not allow public repos to use
+# actions hosted in private repos; other public HarperFast repos pin it by sha:
+#
+#   - uses: HarperFast/harper/.github/actions/review-coverage@<full sha> # main
+#
+name: review-coverage
+description: Report (or enforce) cross-model review coverage named in the PR description
+inputs:
+  mode:
+    description: report (always green, informational) or enforce (red when under-reported)
+    default: report
+  required:
+    description: how many distinct cross-model review families the description must name
+    default: '2'
+runs:
+  using: composite
+  steps:
+    - name: evaluate coverage
+      shell: bash
+      env:
+        REVIEW_COVERAGE_MODE: ${{ inputs.mode }}
+        REVIEW_COVERAGE_REQUIRED: ${{ inputs.required }}
+      run: node "${{ github.action_path }}/ci-review-coverage.mjs"
--- a/.github/actions/review-coverage/action.yml
+++ b/.github/actions/review-coverage/action.yml
@@ -16,11 +16,5 @@ inputs:
     description: how many distinct cross-model review families the description must name
     default: '2'
 runs:
-  using: composite
-  steps:
-    - name: evaluate coverage
-      shell: bash
-      env:
-        REVIEW_COVERAGE_MODE: ${{ inputs.mode }}
-        REVIEW_COVERAGE_REQUIRED: ${{ inputs.required }}
-      run: node "${{ github.action_path }}/ci-review-coverage.mjs"
+  using: node24
+  main: ci-review-coverage.mjs
--- a/.github/actions/review-coverage/ci-review-coverage.mjs
+++ b/.github/actions/review-coverage/ci-review-coverage.mjs
@@ -14,6 +14,14 @@ import { realpathSync } from 'node:fs';
 import { fileURLToPath } from 'node:url';
 import { COVERAGE_REQUIRED, isTrivialChange, reportedCrossModelReviews } from './reviewGate.mjs';
 
+const NON_MEMBER_ASSOCIATIONS = new Set([
+	'COLLABORATOR',
+	'CONTRIBUTOR',
+	'FIRST_TIME_CONTRIBUTOR',
+	'FIRST_TIMER',
+	'NONE',
+]);
+
 /** The whole decision, pure. `pr` is the pull_request object from the Actions event
  *  payload. Returns { pass, exempt, summary, detail } — `pass` already accounts for mode. */
 export function evaluateCiCoverage(pr, { mode = 'report', required = COVERAGE_REQUIRED } = {}) {
@@ -30,7 +38,7 @@ export function evaluateCiCoverage(pr, { mode = 'report', required = COVERAGE_RE
 
 	// The Human-Review-Need footer is evidence a prepush review RAN; staleness means
 	// commits landed after the last reviewed head (HEG step 14's most-skipped step).
-	const footer = body.match(/Human-Review-Need:\s*(\d)(?:[^@\n]*)@\s*([0-9a-f]{6,40})/i);
+	const footer = body.match(/Human-Review-Need:\s*(\d+)(?:[^@\n]*)@\s*([0-9a-f]{6,40})/i);
 	const head = String(pr?.head?.sha ?? '');
 	const footerNote = !footer
 		? 'no Human-Review-Need footer'
@@ -40,15 +48,11 @@ export function evaluateCiCoverage(pr, { mode = 'report', required = COVERAGE_RE
 
 	// The triviality exemption requires the size fields to actually be present: a payload
 	// missing additions/deletions would otherwise read as 0+0 and silently exempt everything.
-	const sized =
-		Number.isFinite(Number(pr?.additions)) &&
-		Number.isFinite(Number(pr?.deletions)) &&
-		pr?.additions !== undefined &&
-		pr?.deletions !== undefined;
+	const sized = Number.isFinite(pr?.additions) && Number.isFinite(pr?.deletions);
 	const exempt =
-		/\[bot\]$/.test(login) || pr?.user?.type === 'Bot'
+		login.endsWith('[bot]') || pr?.user?.type === 'Bot'
 			? 'bot author'
-			: !(assoc === 'MEMBER' || assoc === 'OWNER')
+			: NON_MEMBER_ASSOCIATIONS.has(assoc)
 				? `author is not an org member (${assoc || 'unknown'}) — always human review`
 				: sized && isTrivialChange(pr?.additions, pr?.deletions)
 					? 'trivial change (≤2 lines)'
@@ -71,7 +75,9 @@ function main(mode) {
 	if (!eventPath) throw new Error('no event payload (--event or GITHUB_EVENT_PATH)');
 	const pr = JSON.parse(readFileSync(eventPath, 'utf8')).pull_request;
 	if (!pr) throw new Error('event payload has no pull_request');
-	const required = Number(arg('required', process.env.REVIEW_COVERAGE_REQUIRED ?? '')) || COVERAGE_REQUIRED;
+	const rawRequired = arg('required', process.env.REVIEW_COVERAGE_REQUIRED ?? '');
+	const required = rawRequired === '' ? COVERAGE_REQUIRED : Number(rawRequired);
+	if (!Number.isInteger(required) || required < 0) throw new Error(`invalid required '${rawRequired}'`);
 	const r = evaluateCiCoverage(pr, { mode, required });
 
 	const lines = [
@@ -103,7 +109,7 @@ function main(mode) {
 
 function invokedDirectly() {
 	try {
-		return realpathSync(process.argv[1]) === fileURLToPath(import.meta.url);
+		return realpathSync(process.argv[1]) === realpathSync(fileURLToPath(import.meta.url));
 	} catch {
 		return false;
 	}
--- a/.github/actions/review-coverage/ci-review-coverage.mjs
+++ b/.github/actions/review-coverage/ci-review-coverage.mjs
@@ -1,0 +1,106 @@
+#!/usr/bin/env node
+// CI surface for the human-review gate (policy: skills/pr-shepherd/SKILL.md; enforcement:
+// dispatch/lib/reviewGate.mjs via ReviewApi). This check makes coverage REPORTING visible
+// on the PR itself at open/edit time. It deliberately cannot replicate the gate: CI never
+// sees the fleet review's verdict, so demanding coverage here is STRICTER than policy
+// (the gate waives coverage for clean reviews). Hence two modes:
+//   report  (default) — always green; the check text and job summary carry the count
+//   enforce — red when a member-authored, non-trivial, non-draft PR reports <2
+// Run from the composite action in this directory, or locally:
+//   node scripts/ci-review-coverage.mjs --event <payload.json> [--mode enforce]
+
+import { readFileSync, appendFileSync } from 'node:fs';
+import { realpathSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import {
+	COVERAGE_REQUIRED, isTrivialChange, reportedCrossModelReviews,
+} from './reviewGate.mjs';
+
+/** The whole decision, pure. `pr` is the pull_request object from the Actions event
+ *  payload. Returns { pass, exempt, summary, detail } — `pass` already accounts for mode. */
+export function evaluateCiCoverage(pr, { mode = 'report', required = COVERAGE_REQUIRED } = {}) {
+	const login = String(pr?.user?.login ?? '');
+	const assoc = String(pr?.author_association ?? '');
+	const body = String(pr?.body ?? '');
+	const { count, families } = reportedCrossModelReviews(body);
+	const coverage = count >= required
+		? `${count} cross-model reviews reported (${families.join(', ')})`
+		: count === 1 ? `1 cross-model review reported (${families.join(', ')}) — policy asks for ${required}`
+		: `no cross-model reviews reported — policy asks for ${required}`;
+
+	// The Human-Review-Need footer is evidence a prepush review RAN; staleness means
+	// commits landed after the last reviewed head (HEG step 14's most-skipped step).
+	const footer = body.match(/Human-Review-Need:\s*(\d)(?:[^@\n]*)@\s*([0-9a-f]{6,40})/i);
+	const head = String(pr?.head?.sha ?? '');
+	const footerNote = !footer ? 'no Human-Review-Need footer'
+		: head.startsWith(footer[2].toLowerCase()) ? `Human-Review-Need: ${footer[1]} @ head`
+		: `Human-Review-Need footer is STALE (reviewed @ ${footer[2].slice(0, 7)}, head is ${head.slice(0, 7)})`;
+
+	// The triviality exemption requires the size fields to actually be present: a payload
+	// missing additions/deletions would otherwise read as 0+0 and silently exempt everything.
+	const sized = Number.isFinite(Number(pr?.additions)) && Number.isFinite(Number(pr?.deletions))
+		&& (pr?.additions !== undefined && pr?.deletions !== undefined);
+	const exempt = /\[bot\]$/.test(login) || pr?.user?.type === 'Bot' ? 'bot author'
+		: !(assoc === 'MEMBER' || assoc === 'OWNER') ? `author is not an org member (${assoc || 'unknown'}) — always human review`
+		: sized && isTrivialChange(pr?.additions, pr?.deletions) ? 'trivial change (≤2 lines)'
+		: pr?.draft ? 'draft — checked again at ready-for-review'
+		: '';
+	const compliant = count >= required;
+	const pass = mode !== 'enforce' || Boolean(exempt) || compliant;
+	const summary = exempt ? `exempt: ${exempt}` : coverage;
+	return { pass, exempt, compliant, count, families, summary, detail: `${coverage}; ${footerNote}` };
+}
+
+function arg(name, fallback = '') {
+	const i = process.argv.indexOf(`--${name}`);
+	return i >= 0 && process.argv[i + 1] ? process.argv[i + 1] : fallback;
+}
+
+function main(mode) {
+	const eventPath = arg('event', process.env.GITHUB_EVENT_PATH ?? '');
+	if (!eventPath) throw new Error('no event payload (--event or GITHUB_EVENT_PATH)');
+	const pr = JSON.parse(readFileSync(eventPath, 'utf8')).pull_request;
+	if (!pr) throw new Error('event payload has no pull_request');
+	const required = Number(arg('required', process.env.REVIEW_COVERAGE_REQUIRED ?? '')) || COVERAGE_REQUIRED;
+	const r = evaluateCiCoverage(pr, { mode, required });
+
+	const lines = [
+		`### Cross-model review coverage — ${r.pass ? (r.exempt ? '✅ exempt' : r.compliant ? '✅' : '⚠️ report-only') : '❌'}`,
+		'', r.detail, '',
+		r.exempt ? `_${r.exempt}_`
+			: r.compliant ? ''
+			: `Per team policy, a substantive PR is queued for human review only after ${required} cross-model reviews are run and reported in the description (\`## Review coverage\` naming each model — see harper-engineering-guidelines). The dispatch review gate enforces this when the fleet's review finds issues; this check just makes the reporting visible early.`,
+	].filter(Boolean);
+	if (process.env.GITHUB_STEP_SUMMARY) appendFileSync(process.env.GITHUB_STEP_SUMMARY, lines.join('\n') + '\n');
+	console.log(`review-coverage [${mode}]: ${r.exempt ? r.summary : r.detail}`);
+	// report mode stays green, but an under-reported PR still gets the count on the PR
+	// surface as a warning annotation rather than only in the job summary
+	if (r.pass && !r.exempt && !r.compliant)
+		console.error(`::warning::${r.detail} — the dispatch gate will block at review time if the fleet's review finds issues`);
+	if (!r.pass) {
+		console.error(`::error::${r.detail} — report the reviews in the PR description to pass; the dispatch gate will block at review time if the fleet's review also finds issues`);
+		process.exitCode = 1;
+	}
+}
+
+function invokedDirectly() {
+	try { return realpathSync(process.argv[1]) === fileURLToPath(import.meta.url); } catch { return false; }
+}
+if (invokedDirectly()) {
+	const mode = (arg('mode', process.env.REVIEW_COVERAGE_MODE ?? 'report')).toLowerCase();
+	if (mode !== 'report' && mode !== 'enforce') {
+		// a typo'd mode must be loud — 'enfroce' silently meaning report is the bad direction
+		console.error(`::error::review-coverage: unknown mode '${mode}' (report|enforce)`);
+		process.exitCode = 1;
+	} else {
+		try { main(mode); } catch (error) {
+			// plumbing failures follow the mode: informational stays green, but an enforcing
+			// check that cannot read its payload must not silently wave PRs through
+			console.error(`review-coverage: ${error.message}${mode === 'enforce' ? '' : ' (passing — report mode fails only on policy)'}`);
+			if (mode === 'enforce') {
+				console.error(`::error::review-coverage could not evaluate this PR (${error.message}) — enforce mode fails closed`);
+				process.exitCode = 1;
+			}
+		}
+	}
+}
--- a/.github/actions/review-coverage/ci-review-coverage.mjs
+++ b/.github/actions/review-coverage/ci-review-coverage.mjs
@@ -9,60 +9,9 @@
 // Run from the JavaScript action in this directory, or locally:
 //   node .github/actions/review-coverage/ci-review-coverage.mjs --event <payload.json> [--mode enforce]
 
-import { appendFileSync, readFileSync, realpathSync } from 'node:fs';
-import { fileURLToPath } from 'node:url';
-import { COVERAGE_REQUIRED, isTrivialChange, reportedCrossModelReviews } from './reviewGate.mjs';
-
-const NON_MEMBER_ASSOCIATIONS = new Set([
-	'COLLABORATOR',
-	'CONTRIBUTOR',
-	'FIRST_TIME_CONTRIBUTOR',
-	'FIRST_TIMER',
-	'MANNEQUIN',
-	'NONE',
-]);
-
-/** The whole decision, pure. `pr` is the pull_request object from the Actions event
- *  payload. Returns { pass, exempt, summary, detail } — `pass` already accounts for mode. */
-export function evaluateCiCoverage(pr, { mode = 'report', required = COVERAGE_REQUIRED } = {}) {
-	const login = String(pr?.user?.login ?? '');
-	const assoc = String(pr?.author_association ?? '');
-	const body = String(pr?.body ?? '');
-	const { count, families } = reportedCrossModelReviews(body);
-	const plural = count === 1 ? 'review' : 'reviews';
-	const reported = `${count} cross-model ${plural} reported${families.length ? ` (${families.join(', ')})` : ''}`;
-	const coverage = count >= required ? reported : `${reported} — policy asks for ${required}`;
-
-	// The Human-Review-Need footer is evidence a prepush review RAN; staleness means
-	// commits landed after the last reviewed head (HEG step 14's most-skipped step).
-	const footer = [...body.matchAll(/Human-Review-Need:\s*(\d+)(?:[^@\n]*@\s*([0-9a-f]{6,40}))?/gi)].at(-1);
-	const head = String(pr?.head?.sha ?? '').toLowerCase();
-	const footerNote = !footer
-		? 'no Human-Review-Need footer'
-		: !footer[2]
-			? `Human-Review-Need: ${footer[1]} @ unpinned sha`
-			: head.startsWith(footer[2].toLowerCase())
-				? `Human-Review-Need: ${footer[1]} @ head`
-				: `Human-Review-Need footer is STALE (reviewed @ ${footer[2].slice(0, 7)}, head is ${head.slice(0, 7)})`;
-
-	// The triviality exemption requires the size fields to actually be present: a payload
-	// missing additions/deletions would otherwise read as 0+0 and silently exempt everything.
-	const sized = Number.isFinite(pr?.additions) && Number.isFinite(pr?.deletions);
-	const exempt =
-		login.endsWith('[bot]') || pr?.user?.type === 'Bot'
-			? 'bot author'
-			: NON_MEMBER_ASSOCIATIONS.has(assoc)
-				? `author is not an org member (${assoc}) — always human review`
-				: sized && isTrivialChange(pr?.additions, pr?.deletions)
-					? 'trivial change (≤2 lines)'
-					: pr?.draft
-						? 'draft — checked again at ready-for-review'
-						: '';
-	const compliant = count >= required;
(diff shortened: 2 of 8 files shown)


.github/actions/review-coverage/reviewGate.mjs
```diff
@@ -0,0 +1,167 @@
...
+		const heading = line.match(/^(#{1,4})\s+(.*)/);
+		if (heading) {
+			if (inCoverage && heading[1].length <= coverageLevel) inCoverage = false;
+			if (/review coverage/i.test(heading[2])) {
+				inCoverage = true;
+				coverageLevel = heading[1].length;
+				continue;
+			}
+		}
+		if (inCoverage) harvest(line);
```
cb1kenobi: **Medium: a `## Review coverage` section that ends the body never terminates, so trailing prose inflates the count**

`inCoverage` is only cleared by a heading at the same or shallower level. When `## Review coverage` is the last section — the conventional layout, since the generated-by footer and the `<sub>Human-Review-Need:</sub>` line follow it — every remaining line in the body is harvested. Verified by running this file at `2d9e133`:

```
## Review coverage
- Codex: reviewed

This also touches the grok telemetry path.
```

gives `{"count":2,"families":["openai","xai"]}` — compliant off one real review. This PR is only spared because its trailing `— Claude Fable 5` line matches `generator` and is filtered out; with a non-Anthropic generating model the same layout leaks (I get `count:1, families:["anthropic"]` there, i.e. the wrong family entirely).

This is the canonical gate parser, so the inflation lifts the dispatch gate for real, not just this report-mode check.

Suggested fix — inside the section, harvest only list items, which is the layout the docblock already teaches ("`## Review coverage` naming each model"):

    if (inCoverage) { if (/^\s*(?:[-*]|\d+\.)\s/.test(line)) harvest(line); }

I applied that to a scratch copy and ran the suite: 33/33 still pass, and the case above drops to `{"count":1,"families":["openai"]}`. Per this file's header the change has to land in the canonical `HarperFast/dispatch` copy and sync back to stay byte-identical, so no suggestion block here.

—
_Generated by Barber AI_

.github/workflows/review-coverage.yml
```diff
@@ -0,0 +1,25 @@
...
+# Parses the PR description with the dispatch human-review gate's own parser and reports
+# how many cross-model reviews it names, plus the Human-Review-Need footer state — so
+# coverage is visible at PR open/edit time, before the fleet's review runs. `edited` is a
+# trigger because reporting coverage is a description edit, which moves no commit sha.
+# Policy: HarperFast/dispatch skills/pr-shepherd/SKILL.md. Flipping to `mode: enforce`
+# is a team policy decision.
+name: review-coverage
+on:
+  pull_request:
+    types: [opened, edited, synchronize, ready_for_review]
```
cb1kenobi: **Medium: `reopened` is dropped, so a closed-and-reopened PR never gets a coverage run**

Naming `types:` replaces the `pull_request` defaults (`opened`, `synchronize`, `reopened`) rather than adding to them. A PR that is closed and reopened without a new push fires only `reopened`, which this workflow ignores — so the check is simply absent on that PR. Absent is also the one state a future `mode: enforce` / required-check flip cannot distinguish from "not run yet", which is the failure mode worth closing before the flip rather than after.

Suggested fix:
```suggestion
    types: [opened, edited, synchronize, reopened, ready_for_review]
```

—
_Generated by Barber AI_

.github/actions/review-coverage/ci-review-coverage.mjs
```diff
@@ -0,0 +1,134 @@
...
+			? `Human-Review-Need: ${footer[1]} @ head`
+			: `Human-Review-Need footer is STALE (reviewed @ ${footer[2].slice(0, 7)}, head is ${head.slice(0, 7)})`;
+
+	// The triviality exemption requires the size fields to actually be present: a payload
+	// missing additions/deletions would otherwise read as 0+0 and silently exempt everything.
+	const sized =
+		Number.isFinite(Number(pr?.additions)) &&
+		Number.isFinite(Number(pr?.deletions)) &&
+		pr?.additions !== undefined &&
+		pr?.deletions !== undefined;
```
cb1kenobi: **Low: the size guard still reads `null` / `''` as a 0-line change**

The comment above states the intent — a payload missing the size fields must not silently exempt everything — but `Number(null)` and `Number('')` are both `0` and `null !== undefined` is true, so `sized` comes out true for either. Verified at `2d9e133`: `additions: null, deletions: null` yields `exempt: "trivial change (≤2 lines)"`, same for `''`. Dropping the coercion is both shorter and correct, and it subsumes the two `!== undefined` clauses:

Suggested fix:
```suggestion
	const sized = Number.isFinite(pr?.additions) && Number.isFinite(pr?.deletions);
```

—
_Generated by Barber AI_

.github/actions/review-coverage/ci-review-coverage.mjs
```diff
@@ -0,0 +1,134 @@
...
+	const exempt =
+		/\[bot\]$/.test(login) || pr?.user?.type === 'Bot'
+			? 'bot author'
+			: !(assoc === 'MEMBER' || assoc === 'OWNER')
```
cb1kenobi: **Low: an absent `author_association` exempts the PR under the same label as a real outside contributor**

`assoc` falls back to `''`, so a payload without the field produces `exempt: "author is not an org member (unknown) — always human review"` (verified at `2d9e133`). That is the opposite defensive posture from the size fields four lines below, which deliberately fail closed on a missing field. In `report` mode it silently under-reports; in `enforce` mode it waves the PR through, and the summary line reads like a legitimate contributor exemption rather than a malformed payload.

Worth settling explicitly before the enforce flip: `COLLABORATOR` is exempt here, so an outside collaborator with direct repo access — a contractor rather than an org member — sits permanently outside the coverage expectation. (`author-association-as-membership` is on the decisions list, so this may be intended; flagging the concrete consequence.)

Suggested fix: treat a missing association as unresolved rather than as a non-member — exempt only on a recognized non-member value (`CONTRIBUTOR` / `NONE` / `FIRST_TIME_*` / `COLLABORATOR`) and let an empty or unrecognized value fall through to the normal coverage path, matching the direction the `sized` guard already takes.

—
_Generated by Barber AI_

.github/workflows/review-coverage.yml
```diff
@@ -0,0 +1,25 @@
...
+  cancel-in-progress: true
+jobs:
+  coverage:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      # the action is hosted in THIS repo (public home; GitHub bars public repos from
+      # using private-repo actions, so it cannot live in dispatch) — other public
+      # HarperFast repos pin it by sha
+      - uses: ./.github/actions/review-coverage
```
cb1kenobi, line 23: **Low: the PR under evaluation supplies its own evaluator**

`actions/checkout` on `pull_request` checks out the merge ref, so `./.github/actions/review-coverage` is the copy from the PR's own head. Editing `ci-review-coverage.mjs` within a PR changes the verdict that same PR receives. Harmless in `report` mode; it makes `mode: enforce` self-defeating, which matters given the enforce flip is the stated next step.

To be explicit, `pull_request_target` is *not* the fix — it would pair the same untrusted checkout with a writable token. The fix is the one this PR's description already prescribes for every other repo: pin the action to a sha on `main` (`uses: HarperFast/harper/.github/actions/review-coverage@<sha>`) so a PR cannot rewrite its own gate. Cheaper to do alongside the enforce flip than after it. (`local-action-ref` is on the decisions list — flagging the enforce-mode consequence specifically.)

—
_Generated by Barber AI_

.github/actions/review-coverage/ci-review-coverage.mjs
```diff
@@ -0,0 +1,134 @@
...
+	const coverage =
+		count >= required
+			? `${count} cross-model reviews reported (${families.join(', ')})`
+			: count === 1
+				? `1 cross-model review reported (${families.join(', ')}) — policy asks for ${required}`
+				: `no cross-model reviews reported — policy asks for ${required}`;
+
+	// The Human-Review-Need footer is evidence a prepush review RAN; staleness means
+	// commits landed after the last reviewed head (HEG step 14's most-skipped step).
+	const footer = body.match(/Human-Review-Need:\s*(\d)(?:[^@\n]*)@\s*([0-9a-f]{6,40})/i);
```
cb1kenobi: **Nit: `(\d)` truncates a two-digit Human-Review-Need score**

`Human-Review-Need: 12 @ <sha>` reports as `Human-Review-Need: 1 @ head` — `[^@\n]*` then absorbs the `2`, so it is a silent misreport rather than a parse miss (verified at `2d9e133`).

Suggested fix:
```suggestion
	const footer = body.match(/Human-Review-Need:\s*(\d+)(?:[^@\n]*)@\s*([0-9a-f]{6,40})/i);
```

—
_Generated by Barber AI_

.github/actions/review-coverage/ci-review-coverage.mjs
```diff
@@ -0,0 +1,134 @@
...
+	const i = process.argv.indexOf(`--${name}`);
+	return i >= 0 && process.argv[i + 1] ? process.argv[i + 1] : fallback;
+}
+
+function main(mode) {
+	const eventPath = arg('event', process.env.GITHUB_EVENT_PATH ?? '');
+	if (!eventPath) throw new Error('no event payload (--event or GITHUB_EVENT_PATH)');
+	const pr = JSON.parse(readFileSync(eventPath, 'utf8')).pull_request;
+	if (!pr) throw new Error('event payload has no pull_request');
+	const required = Number(arg('required', process.env.REVIEW_COVERAGE_REQUIRED ?? '')) || COVERAGE_REQUIRED;
```
cb1kenobi: **Nit: `required: '0'` and a typo'd value both silently become `2`**

`||` treats `0` as absent, so the documented `required` input cannot actually be set to `0`, and `required: 'tow'` gives `NaN` and then `2` with no message — the same silent-fallback direction the `mode` validation below deliberately refuses ("a typo'd mode must be loud"). Routing it through the existing throw keeps report mode green and makes enforce mode fail closed, which is already the wiring.

Suggested fix:
```suggestion
	const rawRequired = arg('required', process.env.REVIEW_COVERAGE_REQUIRED ?? '');
	const required = rawRequired === '' ? COVERAGE_REQUIRED : Number(rawRequired);
	if (!Number.isInteger(required) || required < 0) throw new Error(`invalid required '${rawRequired}'`);
```

—
_Generated by Barber AI_

.github/actions/review-coverage/ci-review-coverage.mjs
```diff
@@ -0,0 +1,134 @@
...
+		console.error(
+			`::error::${r.detail} — report the reviews in the PR description to pass; the dispatch gate will block at review time if the fleet's review also finds issues`
+		);
+		process.exitCode = 1;
+	}
+}
+
+function invokedDirectly() {
+	try {
+		return realpathSync(process.argv[1]) === fileURLToPath(import.meta.url);
```
cb1kenobi: **Nit: the entrypoint check compares a realpath against a non-realpath**

`process.argv[1]` goes through `realpathSync` but `fileURLToPath(import.meta.url)` does not, so any symlinked component in `${{ github.action_path }}` makes the comparison false. The failure mode is the script running, printing nothing and exiting 0 — a check that silently stops checking. It resolves correctly on `ubuntu-latest` today (the run at `2d9e133` logs its report line), so this is hardening rather than a live bug.

Suggested fix:
```suggestion
		return realpathSync(process.argv[1]) === realpathSync(fileURLToPath(import.meta.url));
```

—
_Generated by Barber AI_

.github/workflows/review-coverage.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: test review-coverage action
+        run: node --test .github/actions/review-coverage/*.test.mjs
```
cb1kenobi: **Medium: the test step makes the "never blocks" check block, on every PR in the repo**

The header of this file says "report mode: informational, never blocks" and the PR description says "Always green in this mode". `node --test` exits non-zero on any failure, so this step reds the `review-coverage` check itself — and it reds it on *every* PR, since the workflow has no `paths:` filter and also fires on `edited` (a description edit now runs the suite). Two consequences worth closing before the enforce flip: an unrelated break turns an informational check red repo-wide, and after the flip a red `review-coverage` no longer distinguishes "coverage under-reported" from "the action's tests broke" — the same ambiguous-signal problem as the `reopened` gap this commit just fixed. There is also no `actions/setup-node` here, so the version is whatever `ubuntu-latest` ships that week, while `unit-test.yml` deliberately pins a 22/24/26 matrix.

The drift coverage requested above is worth having; it just belongs where a red means "tests failed". Either add the run to `unit-test.yml` alongside `test:bench` (it already checks out, pins Node, and is the repo's test check), or give it its own job/workflow with `paths: ['.github/actions/review-coverage/**']` so it runs only when the action changes.

Minor, if the step stays here: it now executes arbitrary `*.test.mjs` from the PR head in a job whose `actions/checkout` leaves credentials in `.git/config` (`persist-credentials` defaults to true). Near-worthless on a public repo with `contents: read`, but `persist-credentials: false` is a free line.

—
_Generated by Barber AI_

.github/actions/review-coverage/ci-review-coverage.mjs
```diff
@@ -0,0 +1,146 @@
...
+			? `Human-Review-Need: ${footer[1]} @ head`
+			: `Human-Review-Need footer is STALE (reviewed @ ${footer[2].slice(0, 7)}, head is ${head.slice(0, 7)})`;
+
+	// The triviality exemption requires the size fields to actually be present: a payload
+	// missing additions/deletions would otherwise read as 0+0 and silently exempt everything.
+	const sized = Number.isFinite(pr?.additions) && Number.isFinite(pr?.deletions);
+	const exempt =
+		login.endsWith('[bot]') || pr?.user?.type === 'Bot'
+			? 'bot author'
+			: NON_MEMBER_ASSOCIATIONS.has(assoc)
```
cb1kenobi: **Nit: `MANNEQUIN` is missing from the set, and the `'unknown'` fallback below is now dead**

Two small leftovers from this fix. GitHub's `author_association` enum also has `MANNEQUIN` (an imported/deactivated account placeholder) — a recognized non-member, but not in the set, so a mannequin-authored PR is now treated as a member and reported as under-covered. And because `NON_MEMBER_ASSOCIATIONS.has('')` is false, a missing association can no longer reach the message on line 56, so its `${assoc || 'unknown'}` fallback can never render "unknown" — the string now describes a state the code deliberately no longer produces. Adding `MANNEQUIN` and dropping the `|| 'unknown'` keeps the message honest about what an exemption means.

—
_Generated by Barber AI_

.github/actions/review-coverage/ci-review-coverage.test.mjs
```diff
@@ -0,0 +1,158 @@
...
+	const bare = { pull_request: pr() };
+	assert.strictEqual(run(compliant, '--mode', 'enforce'), 0);
+	assert.strictEqual(run(bare, '--mode', 'report'), 0);
+	assert.strictEqual(run(bare, '--mode', 'enforce'), 1);
+	assert.strictEqual(run({ nope: true }, '--mode', 'report'), 0, 'plumbing failure stays green in report mode');
+	assert.strictEqual(run({ nope: true }, '--mode', 'enforce'), 1, 'enforce mode fails closed on plumbing');
+	assert.strictEqual(run(bare, '--mode', 'enfroce'), 1, "a typo'd mode is loud, not silently report");
+	assert.strictEqual(run(bare, '--mode', 'enforce', '--required', '0'), 0, 'zero is a valid threshold');
+	assert.strictEqual(run(bare, '--mode', 'report', '--required', 'tow'), 0, 'report mode stays green on bad input');
+	assert.strictEqual(run(bare, '--mode', 'enforce', '--required', 'tow'), 1, 'enforce mode fails closed on bad input');
```
cb1kenobi: **Nit: the two `--required tow` cases exercise the new validation without asserting it**

Both expectations hold against the pre-fix script: at `2d9e133`, `--required tow` exits 0 in report mode and 1 in enforce mode (old path `Number('tow') || 2` → 2, and the `bare` payload is non-compliant, so enforce reds for the policy reason instead) — I ran it to confirm. So these two lines would not have caught the bug they are here for; only the `--required 0` case above actually discriminates old from new.

Asserting the message makes it a real regression test — `run()` only returns the status, so that case needs `spawnSync` and `assert.match(result.stderr, /invalid required 'tow'/)`.

—
_Generated by Barber AI_

kriszyp:
Kinda thought this might elicit some human responses, but if you'll are cool with it, I'll make the mechanical fixes.

.github/actions/review-coverage/action.yml
```diff
@@ -0,0 +1,20 @@
...
+description: Report (or enforce) cross-model review coverage named in the PR description
+inputs:
+  mode:
+    description: report (always green, informational) or enforce (red when under-reported)
+    default: report
+  required:
+    description: how many distinct cross-model review families the description must name
+    default: '2'
+runs:
+  using: node24
```
cb1kenobi, line 19: **Nit: the PR description still describes the pre-`node24` shape of this PR**

This line is the change that dates three claims in the description, and the description is the artifact this very check teaches people to keep accurate:

- "the **composite action** + parser get their public home in this repo" — it is a JavaScript action now (`using: node24`), and that is the interesting part of the hosting story for the other repos that pin it by sha, since a JS action pulls a runtime pin along with the code.
- "**33 unit tests** ride along in the action dir" — 37 at this head (I ran `node --test .github/actions/review-coverage/*.test.mjs` in a worktree at `5ad17ce4`: 37 pass, 0 fail), and the v24 Unit Test job at this head logs the same 37, so the moved step is really executing.
- "the run on this branch is green and correctly reports `exempt: draft`" — the PR is no longer a draft, and the run at this head reports `review-coverage [report]: 2 cross-model reviews reported (google, openai); Human-Review-Need: 4 @ head`. That is a better verification line than the draft one anyway: it exercises the parser and the footer freshness check rather than the exemption short-circuit.

No code change implied — just the description.

—
_Generated by Barber AI_

.github/actions/review-coverage/evaluateCiCoverage.mjs
```diff
@@ -0,0 +1,48 @@
...
+export function evaluateCiCoverage(pr, { mode = 'report', required = COVERAGE_REQUIRED } = {}) {
+	const login = String(pr?.user?.login ?? '');
+	const assoc = String(pr?.author_association ?? '');
+	const body = String(pr?.body ?? '');
+	const { count, families } = reportedCrossModelReviews(body);
+	const plural = count === 1 ? 'review' : 'reviews';
+	const reported = `${count} cross-model ${plural} reported${families.length ? ` (${families.join(', ')})` : ''}`;
+	const coverage = count >= required ? reported : `${reported} — policy asks for ${required}`;
+
+	const footer = [...body.matchAll(/Human-Review-Need:\s*(\d+)(?:[^@\n]*@\s*([0-9a-f]{6,40}))?/gi)].at(-1);
```
cb1kenobi: **Low: two `Human-Review-Need:` footers on the same physical line merge into one wrong match**

The optional sha group is `(?:[^@\n]*@\s*([0-9a-f]{6,40}))?` — `[^@\n]*` stops at a newline but not at a second `Human-Review-Need:` label, so when two footers share a line the first match swallows the second label and attaches the *second* footer's sha to the *first* footer's count. `.at(-1)` then returns that merged, mis-paired match instead of the real last footer.

Reproduced against this head's regex:

```
same-line          -> matches: 1 | last pair: ["1","abcdef123456"]
newline-separated  -> matches: 2 | last pair: ["2","abcdef123456"]
```

The same-line case reports count `1` (the first footer) paired with the sha from the second — so a STALE/fresh verdict can be computed against the wrong count.

**Impact is limited:** this only feeds `footerNote`/`detail`, not `count`, `compliant`, or `pass`, so it cannot change the check's outcome. And every footer this repo writes is newline-separated, so it needs a hand-edited PR description to trigger.

**Suggested fix** — bound the gap so it can't cross a label:

```js
/Human-Review-Need:\s*(\d+)(?:(?:(?!Human-Review-Need:)[^@\n])*@\s*([0-9a-f]{6,40}))?/gi
```

No test covers same-line duplicate footers; worth one alongside the existing footer cases.

---
🤖 Barber AI · claude-opus-5 · independently reproduced

.github/actions/review-coverage/reviewGate.mjs
kriszyp: Addressed in a4f543bdf: coverage sections now harvest only Markdown list items, so trailing prose cannot add model families. The regression test includes cross-model/Grok prose to ensure it stays ignored. The canonical Dispatch copy still needs the equivalent sync and is recorded as a cross-repo follow-up.

— GPT-5 Codex

.github/actions/review-coverage/evaluateCiCoverage.mjs
kriszyp: Addressed in a4f543bdf: the footer regex cannot cross another Human-Review-Need label, and a same-line duplicate-footer regression test now verifies the last score remains paired with its own SHA.

— GPT-5 Codex

.github/actions/review-coverage/reviewGate.mjs
```diff
@@ -0,0 +1,167 @@
...
+		else if (/^\s*(?:[-*>]\s*)?independent-review\s*:/i.test(line)) harvest(line);
+		else if (/cross[- ]model/i.test(line)) harvest(line);
+	}
+	return { count: families.size, families: [...families].sort(), generator: own };
```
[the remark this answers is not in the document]
kriszyp: Addressed in a4f543bdf with the list-item-only section parser and a regression test proving trailing cross-model/Grok prose does not inflate coverage. The canonical Dispatch copy still needs the equivalent sync and is recorded as a cross-repo follow-up.

— GPT-5 Codex

.github/actions/review-coverage/action.yml
kriszyp: Updated the PR description: it now identifies the JavaScript action and node24 runtime, reports 38 tests at the new head, and describes current non-draft dogfooding behavior.

— GPT-5 Codex

.github/workflows/review-coverage.yml
kriszyp: No code change in this report-mode PR: the action remains informational and always green. Pinning the evaluator to a trusted main SHA is required before any future enforce/required-check rollout; that policy change remains explicitly outside this PR.

— GPT-5 Codex

.github/actions/review-coverage/reviewGate.mjs
```diff
@@ -0,0 +1,168 @@
...
+		if (heading) {
+			if (inCoverage && heading[1].length <= coverageLevel) inCoverage = false;
+			if (/review coverage/i.test(heading[2])) {
+				inCoverage = true;
+				coverageLevel = heading[1].length;
+				continue;
+			}
+		}
+		if (inCoverage) {
+			if (/^\s*(?:[-*]|\d+\.)\s/.test(line)) harvest(line);
```
cb1kenobi, line 146: **Medium: this fixes the public copy, but not the gate that enforces the policy**

The workflow/action comments say this parser is the CI surface of the dispatch human-review gate, but `HarperFast/dispatch@main` still has the old `if (inCoverage) harvest(line);` branch in `dispatch/lib/reviewGate.mjs`. After this PR merges, CI will stop counting trailing prose while the real queueing gate can still treat a one-review PR as compliant if prose after the final `## Review coverage` section happens to mention another model.

Suggested fix: mirror this same `reportedCrossModelReviews` boundary change into `HarperFast/dispatch` before relying on the check, and link that dispatch PR/SHA from this PR so the report path and enforcement path cannot diverge.

—
_Generated by Barber AI_

kriszyp: Confirmed. The public copy is fixed in a4f543bdf, but the canonical Dispatch parser remains unchanged. This one-shot task explicitly restricts edits to the Harper worktree, so I cannot create the required companion Dispatch change from this session. I am leaving this thread unresolved and reporting that cross-repo authorization and worktree as the blocker. — GPT-5 Codex

DavidCockerill (approved):
Approving at `a4f543b`. Report mode annotates and always passes, and the header is explicit that flipping to `mode: enforce` is a separate team policy decision — so this lets the annotation exist without committing anyone to the gate. CI is green, including the new test step.

@cb1kenobi's pass is thorough and most of it is already closed at this head — `required` now parses `'0'` correctly and rejects a typo (`Number.isInteger(required) || required < 0` throws), the `Human-Review-Need` regex uses `(\d+)` with a `(?!Human-Review-Need:)` lookahead that closes both the two-digit truncation and the two-footers-on-one-line merge, and `reopened` is present in the trigger list. Two comments left, neither of them a reason to hold this.

**1. The description overstates "never blocks".** `.github/workflows/unit-test.yml:53-55` adds a *Review-coverage action tests* step to the shared unit-test job (`if: matrix.node-version == 24`). The coverage workflow itself is genuinely always green, but a failure in these tests fails unit-test, which blocks every PR in the repo.

To be clear I think running the action's own tests in CI is the *right* thing — you want them to run somewhere, and they pass today. The issue is only that "Always green in this mode" and "informational, never blocks" describe the coverage verdict while reading as a property of the whole change. Either wording fix or moving the step into the non-blocking coverage workflow settles it; your call which. Flagging it because this PR's description is the artifact the check itself parses, so it is worth being accurate.

**2. Nothing detects divergence between this parser and the one that enforces the policy.** `reviewGate.mjs`'s header says *"keep these byte-identical below this header block"*, and no test or CI step asserts it — `reviewGate.test.mjs` has no checksum, byte-comparison or sync assertion. The invariant is a comment, which is the shape copies drift in.

The failure mode is specific and confusing rather than loud: this check reports coverage under one set of rules while the dispatch gate enforces another, so a PR reads green here and is auto-request-changed by the fleet — or the reverse — and the discrepancy looks like a fleet bug rather than a drifted copy.

Property a fix needs: divergence has to be detectable without a human reading both files. A checksum of the shared region asserted in `reviewGate.test.mjs`, or a scheduled job that fetches dispatch's copy and diffs it, would both do it — I am not prescribing which, and the constraint that forced the copy (public repos cannot use actions from private ones) is real and well explained. Recording it as a follow-up in dispatch rather than a condition on this PR, since it cannot be fixed here.

This is the complete set of my concerns at this head.

.github/actions/review-coverage/reviewGate.mjs
```diff
@@ -0,0 +1,231 @@
...
+		if (fams.length === 1) return fams[0];
+	}
+	return '';
+}
+
+/** Blank out fenced code blocks so an EXAMPLE of a footer cannot be read as a live one. A PR that
+ *  documents this field (or quotes a body) carries the field's own grammar inside a fence; counting
+ *  that would let a description claim coverage by illustrating it. Local rather than imported: this
+ *  file is a flattened vendored copy with no module deps of its own. */
+const withoutFences = (body) => String(body ?? '').replace(/^[ \t]*```[\s\S]*?^[ \t]*```[ \t]*$/gm, '');
```
cb1kenobi, line 47: **Medium: this vendored `withoutFences` is weaker than the canonical `withoutFencedMarkdown`, and the prose path is not stripped at all**

The header says keep this file byte-identical to `dispatch/lib/reviewGate.mjs`. That copy imports `withoutFencedMarkdown` from `complexity.mjs` — a line-based stripper that handles ``` and `~~~` of any length and treats an unclosed fence as open to end-of-body — and it applies it to the prose path too (`const text = withoutFencedMarkdown(body)`). This copy handles only paired backtick fences, and `reportedCrossModelReviews` below still runs on raw `String(body ?? '')`.

Executed against this file at `25db90f9`; every divergence is in the permissive direction:

| body shape | this copy | canonical |
| --- | --- | --- |
| field inside a `~~~` fence | count 2 | stripped |
| field after an unclosed ``` fence | count 2 | stripped |
| `## Review coverage` prose section inside a ``` fence | count 2 | stripped |

Two more shapes neither copy handles, worth folding into the same fix since both parse as a live field here (count 2): a field inside a multi-line `<!-- ... -->` comment — which is **invisible in the rendered description**, so the claim is one no human reviewing the PR page can see — and a 4-space indented code block, which is markdown's other code-block syntax and exactly the case this helper exists to defend against.

Suggested fix: vendor the canonical `withoutFencedMarkdown` body verbatim, use it in `reportedCrossModelReviews` as well, and additionally blank `<!--[\s\S]*?-->` before matching. Dropping the leading `[ \t]*` from the `Review-Coverage:` regex on line 77 closes the indented-block case.

—
_Generated by Barber AI_

.github/actions/review-coverage/reviewGate.mjs
```diff
@@ -0,0 +1,231 @@
...
+	if (!match) return null;
+	const segments = new Map(
+		match[1]
+			.split(';')
+			.map((seg) => seg.trim().split('='))
+			.filter((pair) => pair.length === 2)
+			.map(([key, value]) => [key.trim().toLowerCase(), value.trim()])
+	);
+	const authored = familiesIn((segments.get('authored') || '').replace(/\s*@.*$/, ''))[0] || '';
+	const raw = segments.get('ran') || '';
```
cb1kenobi, line 89: **Medium: a present-but-unparseable field reports 0 and blocks the prose fallback — the undercount direction this file calls out as the bad one**

The doc comment says returning `null` on an absent field is what lets the caller fall back to prose "instead of reading 'no field' as 'no coverage'". But a field that is *present* and yields no `ran` segment returns `{ count: 0 }`, and `reportedCrossModelReviews` returns that, so the `## Review coverage` section is never read.

Executed at `25db90f9` — each of these bodies also carries a full two-leg `## Review coverage` section, and each reports `count: 0`:

- `Review-Coverage: totally garbage no equals here`
- `Review-Coverage: legs=codex,gemini; v=2` — segment names from a different producer version
- `Review-Coverage: authored=claude; ran=codex(model=gpt-5),gemini; rounds=2` — `split('=')` yields 3 parts, so `pair.length === 2` on line 85 drops the whole `ran` segment. A parenthetical carrying an `=` is the same shape as the `blocked=cursor-grok(auth)` in the new tests.

The header above `reportedCrossModelReviews` says an undercount GATES a compliant PR. Given this copy already drifts from the dispatch one, producer/parser version skew is the realistic trigger.

Suggested fix — fall back to prose when the field carries no `ran` segment (`ran=none` still returns non-null, so the new test holds):

```suggestion
	if (!segments.has('ran')) return null;
	const raw = segments.get('ran') || '';
```

And split on the first `=` only, so a value containing `=` survives: `.map((seg) => seg.trim().split(/=(.*)/s).slice(0, 2))` on line 84. With both applied, all 32 tests in `reviewGate.test.mjs` still pass and the three bodies above fall through to the prose reader and report 2.

—
_Generated by Barber AI_

.github/actions/review-coverage/reviewGate.mjs
```diff
@@ -0,0 +1,231 @@
...
+ *
+ *  Returns null when the field is absent, which is what lets the caller fall back to the prose
+ *  reader instead of reading "no field" as "no coverage" — the field is newer than most open PRs.
+ *
+ *  Only `ran=` counts. `blocked=` (asked for, could not deliver) and `declined=` (policy chose not
+ *  to run it) are coverage that did NOT happen. The producing side already excludes the authoring
+ *  family and the Harper adjudicator from `ran`; we re-exclude the authoring family anyway, because
+ *  this number acts on other people's PRs and should not inherit a guarantee it can check itself. */
+export function structuredCoverage(body) {
+	const match = /^[ \t]*(?:<sub>[ \t]*)?Review-Coverage:[ \t]*(.+?)[ \t]*(?:<\/sub>[ \t]*)?$/im.exec(
```
cb1kenobi, line 77: **Low: `.exec` takes the FIRST `Review-Coverage:` line, but this footer is appended — a stale field above a fresh one wins**

The regex has no `g` flag, so `exec` returns the earliest match in the body. The producer appends this footer, so when a description ends up carrying two — a re-run that appends rather than replaces, or an older footer left above the current one — the parser reads the stale one.

Verified at `25db90f9`: a body whose first footer is `authored=claude; ran=none; rounds=0` and whose last is `authored=claude; ran=codex,gemini; rounds=2` reports `count: 0`.

(Distinct from the `Human-Review-Need:` finding on `evaluateCiCoverage.mjs` — that one is two footers on the same physical line; this is two on separate lines, in this function.)

Suggested fix: take the last match rather than the first, e.g. `const match = [...withoutFences(body).matchAll(/…/gim)].at(-1);`.

—
_Generated by Barber AI_

.github/actions/review-coverage/reviewGate.mjs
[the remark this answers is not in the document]
cb1kenobi: Re-checked at `25db90f9`, which is the first range to touch this file (+63). **Still reproduces** — executing the parser at head:

- `## Review coverage` / `- Codex: reviewed` / blank / `- Follow-up: ask Gemini about the caching approach later` → **`count: 2`, families `["google","openai"]`** from one real review.
- The same body with a `## Notes` heading before the trailing bullet → `count: 1` (correct).
- Trailing non-bullet prose → `count: 1` (only list items are harvested while `inCoverage`).
- A normal terminated two-leg section → `count: 2` (correct — the normal path is unaffected).

The new structured field narrows the blast radius, since `reportedCrossModelReviews` now returns early when `Review-Coverage:` is present — but it does not close this path, and it does not cover the bodies most likely to hit it: **this PR's own description** has a `## Review coverage` section that ends the body with the footer lines after it, and reproduces `count: 2`.

Flagging because the thread was resolved without a code change. Not opening a new thread for it.

—
_Generated by Barber AI_

PR #2184, opened 2026-08-16
Prevent indexed cache TTL test flakes

kriszyp:
The indexed cache test depended on a 5 ms expiry and 10 ms sleeps, so loaded runners could observe an additional revalidation before the exact source-request assertion. This change controls the source expiry explicitly, waits on cache state, publication delivery, and source-write locks, preserves exact request counts, and verifies secondary-index cleanup directly.

Review feedback on the first revision found two assertions that did not assert what they claimed, and both are now closed:

- The source-request check polled `sourceRequests >= 2` and then asserted equality on the next line, so it read the counter the instant it reached two. [`Source.get()` now counts a request when it starts](https://github.com/HarperFast/harper/pull/2184/changes?w=1#diff-3e4675948d111966e41fc20a24a583b464f51390e2ca2fc4420e97959164200cR55), matching the file's two other source fixtures, and [the query-path barrier waits for the revalidated entry to commit](https://github.com/HarperFast/harper/pull/2184/changes?w=1#diff-3e4675948d111966e41fc20a24a583b464f51390e2ca2fc4420e97959164200cR555) rather than for the counter. Every revalidation the query can trigger is started before that commit lands, so the exact `assert.equal(sourceRequests, 2)` now runs after all of them are counted.
- The event check watched a subscription on `CachingTable`, a table this test never touches, so it could not fail. A test-scoped subscription on `IndexedCachingTable` replaces it, and because subscription delivery is deferred through `setImmediate` in `resources/transactionBroadcast.ts`, both ends of the observation window are fenced by [a scratch-id source fill whose own publication is awaited](https://github.com/HarperFast/harper/pull/2184/changes?w=1#diff-3e4675948d111966e41fc20a24a583b464f51390e2ca2fc4420e97959164200cR524). Publications are delivered in commit order, so [the negative assertion](https://github.com/HarperFast/harper/pull/2184/changes?w=1#diff-3e4675948d111966e41fc20a24a583b464f51390e2ca2fc4420e97959164200cR595) cannot pass merely because a publication had not been delivered yet.

Eviction's [secondary-index check gets its own barrier](https://github.com/HarperFast/harper/pull/2184/changes?w=1#diff-3e4675948d111966e41fc20a24a583b464f51390e2ca2fc4420e97959164200cR605) instead of assuming it becomes visible with the primary removal, and the subscription is torn down with [`close()`](https://github.com/HarperFast/harper/pull/2184/changes?w=1#diff-3e4675948d111966e41fc20a24a583b464f51390e2ca2fc4420e97959164200cR609), which `Subscription` implements — `return()` lives on the iterator, not the subscription, so the earlier optional call was a no-op.

## For the human reviewer

Four decisions are deliberate rather than settled, and each is cheap to reverse:

- **Exact source-request counts.** `assert.equal(sourceRequests, 2)` / `=== 1` asserts "exactly one source request per expired row" as a product invariant, where sibling tests use `<=` bounds. If Harper legitimately tolerates a benign duplicate under scheduling pressure, this becomes the file's most fragile assertion.
- **Counting at request start.** Moving `sourceRequests++` out of the source's `setTimeout` changes the counter for every test sharing `Source` from "responses landed" to "requests started". That is what makes a concurrent duplicate observable, and it matches `CachingTable`'s and `SwrQueryTable`'s fixtures, but it is a shared-fixture change.
- **Proving a negative over an async channel.** "Revalidation publishes nothing" is asserted as an empty filtered array between two delivery fences. The fences make it sound, but an absence assertion is still the weakest available shape; a publish spy would be stronger and would mean rewriting this half of the test.
- **`IndexedCachingTable` is left at `{ expiration: 50, eviction: 100 }`** and the fence fills leave scratch records 24 and 25 behind. No test after this one touches that table (verified), and restoring the prior 5 ms TTL would rearm the cleanup churn this change removes — but a test inserted after this one would inherit both.

Gemini's round-2 findings were checked and not acted on: the missing `'error'` listener has no producer (`emit('error'` has zero matches under `resources/`), and its `setTTLExpiration` and scratch-record scenarios both name "Bigger stampede is handled", which uses `CachingTable`, not `IndexedCachingTable`. The three added comments were flagged as narration across both rounds; they are kept because each states an invariant the code cannot express (why the counter moved, why commit order makes the fence sound, why the commit barrier makes the count exact).

## Verification

- `npm run build` — passed.
- `npx mocha unitTests/resources/caching.test.js --grep "Can load cached indexed data"` repeated 50 times — 50/50 passed. Whole-file `npx mocha unitTests/resources/caching.test.js` repeated 20 times — 20/20 passed (20 passing each).
- `npm run test:unit:resources` — 1,553 passing / 15 pending; three unrelated pre-existing failures remain in `databases.test.js`, `randomAccessFieldsDirective.test.js`, and `replayStructures.test.js`. The changed caching test passed in this gate.
- Negative controls, each run 3 times and failing 3/3:
  - publishing an id-23 event (`invalidate(23)`) immediately before the closing fence → `source revalidation should not publish an update event`, proving the fence observes a publication committed just before it;
  - the reviewer's own injection, `setTimeout(() => { sourceRequests++; }, 5)` before the query-path barrier → `3 == 2`, which passed 3/3 against the previous revision;
  - restoring the old `evict(23, entry, entry.version)` argument → `eviction should remove the secondary index entry`.
- `npx prettier --check unitTests/resources/caching.test.js` and `npm run lint:required` — passed.
- GitHub Actions on `83b650d`: 41 checks pass, 1 fails — `Integration Tests 2/6 (Windows, Node.js v24)`. That shard fails in `integrationTests/apiTests/describe-metadata-upgrade.test.ts` (`Probe /SeoPageCache/ did not become ready within 120000ms ... after restart_service`), it failed twice on this head with two different transport errors, and it fails identically on `main` (run `32804430806`, same test, same probe timeout). Unrelated to this branch, which touches one unit-test file.
- End-to-end route: not observable end-to-end because this is test-only; the resource unit gate exercises the real table, source, storage, index, and subscription implementations.

## Review coverage

The first revision was authored by GPT-5 Codex; this revision by Claude Opus 5. Round 1 (`8b15c95`): Codex (graded), Gemini, Cursor Composer, and Harper-domain adjudication — one major finding (the event assertion had no delivery barrier) plus three minor, all fixed in `83b650d`. Round 2 (`83b650d`, `--full`): Codex graded leg returned LGTM with no findings; Gemini's findings are dispositioned above; the Cursor Composer and Harper-domain legs timed out, so outside findings were not machine-adjudicated this round.

<sub>Review-Coverage: authored=claude; ran=gemini,codex; blocked=cursor-composer(failed),domain(timeout); declined=cursor-grok; rounds=2 @ 83b650db0d04</sub>

<sub>Human-Review-Need: 4 @ 83b650db0d04</sub>

### DIFF
--- a/unitTests/resources/caching.test.js
+++ b/unitTests/resources/caching.test.js
@@ -573,16 +573,9 @@ describe('Caching', () => {
 			await waitFor(() => IndexedCachingTable.primaryStore.getSync(23) === undefined, {
 				message: 'eviction should remove the primary record',
 			});
-			results = [];
-			for await (const record of IndexedCachingTable.search({
-				conditions: [{ attribute: 'name', value: 'name 23' }],
-			})) {
-				results.push(record);
-			}
-			assert.equal(results.length, 0);
+			assert.equal(IndexedCachingTable.indices.name.getValuesCount('name 23'), 0);
 		} finally {
 			sourceExpiresAt = undefined;
-			IndexedCachingTable.setTTLExpiration(0.005);
 		}
 	});
 
--- a/unitTests/resources/caching.test.js
+++ b/unitTests/resources/caching.test.js
@@ -51,9 +51,11 @@ describe('Caching', () => {
 			get() {
 				let expiresAt = sourceExpiresAt ?? Date.now() + 2;
 				this.getContext().expiresAt = expiresAt;
+				// counted when the request starts, like the other sources here, so a concurrent
+				// duplicate is observable before the first response lands
+				sourceRequests++;
 				return new Promise((resolve, reject) => {
 					setTimeout(() => {
-						sourceRequests++;
 						if (return_error) {
 							let error = new Error('test source error');
 							error.statusCode = return_error;
@@ -512,9 +514,13 @@ describe('Caching', () => {
 		}
 	});
 	it('Can load cached indexed data', async function () {
+		const indexedEvents = [];
+		const indexedSubscription = await IndexedCachingTable.subscribe({});
+		indexedSubscription.on('data', (event) => {
+			indexedEvents.push(event);
+		});
 		try {
 			sourceRequests = 0;
-			events = [];
 			IndexedCachingTable.setTTLExpiration({ expiration: 50, eviction: 100 });
 			const firstExpiresAt = (sourceExpiresAt = Date.now() + 1);
 			let result = await IndexedCachingTable.get(23);
@@ -522,7 +528,6 @@ describe('Caching', () => {
 			await waitFor(() => !IndexedCachingTable.primaryStore.hasLock(23), {
 				message: 'initial source fill should finish committing',
 			});
-			events = [];
 			assert.equal(result.name, 'name ' + 23);
 			assert.equal(sourceRequests, 1);
 			await waitFor(
@@ -532,13 +537,20 @@ describe('Caching', () => {
 					message: 'initial source fill should expire',
 				}
 			);
-			sourceExpiresAt = Date.now() + 1000;
+			const revalidatedExpiresAt = (sourceExpiresAt = Date.now() + 1000);
 			let results = [];
 			for await (let record of IndexedCachingTable.search({ conditions: [{ attribute: 'name', value: 'name 23' }] })) {
 				results.push(record);
 			}
 			assert.equal(results.length, 1);
-			await waitFor(() => sourceRequests >= 2, { message: 'indexed query should revalidate the expired entry' });
+			// every revalidation this query could trigger is started before the refreshed entry is
+			// committed, so waiting on the commit makes the request count exact rather than a floor
+			await waitFor(
+				() =>
+					IndexedCachingTable.primaryStore.getEntry(23)?.expiresAt === revalidatedExpiresAt &&
+					!IndexedCachingTable.primaryStore.hasLock(23),
+				{ message: 'indexed query should revalidate the expired entry and commit the refreshed cache write' }
+			);
 			assert.equal(sourceRequests, 2);
 			result = await IndexedCachingTable.get(23);
 			assert.equal(result.id, 23);
@@ -557,15 +569,22 @@ describe('Caching', () => {
 				}
 			);
 			sourceExpiresAt = Date.now() + 1000;
+			// the preceding barriers have already let this table's earlier publishes land, so anything
+			// arriving from here on belongs to the revalidation below
+			indexedEvents.length = 0;
 			result = await IndexedCachingTable.get(23);
 			assert.equal(result.id, 23);
 			assert.equal(result.name, 'name ' + 23);
 			assert.equal(sourceRequests, 1);
-			assert.equal(events.length, 0);
 			assert(result.getExpiresAt());
 			await waitFor(() => !IndexedCachingTable.primaryStore.hasLock(23), {
 				message: 'source revalidation lock should be released',
 			});
+			assert.deepEqual(
+				indexedEvents.map((event) => event.type),
+				[],
+				'source revalidation should not publish an update event'
+			);
 			const entry = IndexedCachingTable.primaryStore.getEntry(23);
 			assert(entry, 'source revalidation should leave a cache entry to evict');
 			await IndexedCachingTable.evict(23, entry.value, entry.version);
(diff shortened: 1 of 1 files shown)


unitTests/resources/caching.test.js
```diff
@@ -511,38 +512,71 @@ describe('Caching', () => {
...
+				}
+			);
+			sourceExpiresAt = Date.now() + 1000;
+			let results = [];
+			for await (let record of IndexedCachingTable.search({ conditions: [{ attribute: 'name', value: 'name 23' }] })) {
+				results.push(record);
+			}
+			assert.equal(results.length, 1);
+			await waitFor(() => sourceRequests >= 2, { message: 'indexed query should revalidate the expired entry' });
+			assert.equal(sourceRequests, 2);
```
cb1kenobi, line 563: **Nit: this pair is effectively `sourceRequests >= 2`, not an exact count**

`waitFor` returns the instant the count reaches 2, so the `assert.equal` on the next line samples at the earliest possible moment. A duplicate revalidation landing a few ms later is not observed. Verified empirically: inserting `setTimeout(() => { sourceRequests++; }, 5);` immediately before the `waitFor` leaves the test passing 3/3.

This is not a regression — the base version asserted even earlier (right after the search loop drained) and was equally blind to a late request — and the fixture change in this PR (giving the revalidated entry a 1000 ms expiry) removes the actual cause of the extra request. Flagging only because the PR description claims exact request counts are preserved; they are exact for requests that have already landed, not for ones that land during or after the poll.

—
_Generated by Barber AI_

unitTests/resources/caching.test.js
```diff
@@ -511,38 +512,71 @@ describe('Caching', () => {
...
+				{
+					message: 'second source fill should expire',
+				}
+			);
+			sourceExpiresAt = Date.now() + 1000;
+			result = await IndexedCachingTable.get(23);
+			assert.equal(result.id, 23);
+			assert.equal(result.name, 'name ' + 23);
+			assert.equal(sourceRequests, 1);
+			assert.equal(events.length, 0);
```
cb1kenobi: **Nit: this assertion is vacuous for `IndexedCachingTable`**

`events` is only ever pushed to by the subscription created at line 90, `await CachingTable.subscribe({})`. Nothing subscribes to `IndexedCachingTable`, so `events.length` here cannot be non-zero no matter what the indexed table emits — the assertion passes even if the blocking revalidation it is meant to police starts emitting events.

Pre-existing (carried over from the old block rather than introduced here), so not a blocker. Either subscribe to `IndexedCachingTable` in `before()` and keep the check meaningful, or drop the line. Static analysis only — I did not build a mutation to force an indexed-table event.

—
_Generated by Barber AI_

unitTests/resources/caching.test.js
kriszyp: Fixed in 83b650d — good catch, and the empirical control was the useful part.

Two changes close the window: `Source.get()` now counts a request when it **starts** (matching `CachingTable`'s and `SwrQueryTable`'s fixtures, which already did), and the barrier waits for the revalidated entry to **commit** rather than for the counter. Every revalidation this query can trigger is started before that commit lands, so the exact assertion runs after all of them are counted.

Your injection is now caught: `setTimeout(() => { sourceRequests++; }, 5)` immediately before the barrier fails `3 == 2`, 3/3 runs — it passed 3/3 before this change.

— Claude Opus 5

unitTests/resources/caching.test.js
kriszyp: Fixed in 83b650d — you were right that nothing subscribed to `IndexedCachingTable`. Went with your first option, scoped to this test rather than `before()` so the shared `events` array (which sibling tests assert counts on) is untouched.

Making it real surfaced a second problem you did not have to build a mutation for: subscription delivery is deferred through `setImmediate` in `resources/transactionBroadcast.ts`, so a synchronous read of the array proves nothing either way. Both ends of the observation window are now fenced by a scratch-id source fill whose own publication is awaited; publications are delivered in commit order, so anything committed before the fence has been delivered by then. Control: publishing an id-23 event (`invalidate(23)`) immediately before the closing fence fails the assertion 3/3.

For the record, the expected behaviour is that source revalidation publishes nothing — the `put`/`invalidate` events on this table come from the fills and the explicit invalidate, not from revalidation.

— Claude Opus 5

cb1kenobi (approved):

Issue #1241, opened 2026-06-11
Predicate-aware HNSW traversal: filtered vector search with user-function and RBAC participation


kriszyp:
# Predicate-aware HNSW traversal: filtered vector search with user-function and RBAC participation

## Problem

Combining a vector sort with other conditions currently works by post-filtering: `searchByIndex()` runs `customIndex.search()`, which returns at most `ef` nearest candidates, and the remaining AND conditions then filter that fixed candidate set (`resources/search.ts` — custom-index branch around line 389, post-filters built by `filterByType()` around lines 719–939). Two failure modes follow:

1. **Under-filling.** With a selective filter (say 1% of records match), the top-`ef` nearest neighbors may contain few or zero matches, so a `limit: 10` query returns 2 results even though plenty of matching records exist further out in the graph. The only workaround is raising `ef`, which is a blunt, global over-fetch with no guarantee.
2. **No record-level security filtering.** Permission checks run at `Table.search()` entry (`resources/Table.ts:2201–2207` via `allowRead`), not per record during traversal. A user who can only see a subset of records gets "nearest k, then drop forbidden" rather than "nearest k visible" — under-filled results again, and no way for a custom resource to express record-level visibility inside a vector query at all.

This issue makes the HNSW beam search predicate-aware, so the traversal keeps exploring until it has `ef` *matching* results (or exhausts a visit budget). It is the enabling primitive for graph-expanded retrieval and conversation-guided retrieval (follow-on issues), and it gives user functions — including RBAC predicates — a sound place to participate in vector search.

## Prior art

ACORN (arXiv:2403.04871) is the reference design: keep traversing through non-matching nodes (they remain routing waypoints, preserving graph connectivity), but only matching nodes enter the result list. Filtered-DiskANN, Weaviate's flat-search fallback, and Qdrant's filterable HNSW solve the same problem with similar shapes: predicate-during-traversal for moderate selectivity, brute-force over the filter's candidate set when the filter is very selective.

### 1. Filter parameter on the index search

Extend the custom-index search signature (`resources/indexes/HierarchicalNavigableSmallWorld.ts:654`):

```ts
search(condition, context, filter?: (primaryKey: Id, getRecord: () => any) => boolean)
```

- `filter` is consulted **only at layer 0** and only to decide whether a node enters the `results` list. Upper layers (routing) and the candidate/visited sets ignore it — non-matching nodes still route, per ACORN, so the graph stays navigable under selective filters.
[...]

PR #2187, opened 2026-08-17
Use an options object for searchByIndex and the custom-index search() contract

### LINKED ISSUE #2165: Use an options object for searchByIndex and the custom-index search() contract
`searchByIndex` in `resources/search.ts` has grown to eight positional parameters, the last of
which — `minResults` — was added in [#2125](https://github.com/HarperFast/harper/pull/2125) so a
query's `offset + limit` could reach an approximate index and stop it from silently truncating the
result set.

```ts
export function searchByIndex(
	searchCondition: DirectCondition,
	transaction: any,
	reverse: boolean,
	Table: any,
	allowFullScan?: boolean,
	filtered?: any,
	context?: any,
	minResults?: number
): AsyncIterable<Id | { key: Id; value: any }>;
```

Five of those eight are optional, so most call sites pass a run of `undefined`s or stop early and
inherit defaults they never meant to choose. `resources/Table.ts` and the recursive relationship
calls inside `search.ts` pass nothing past `filtered`, which is harmless today only because none of
them lead with a vector sort.

The real cost is on the custom-index contract. `HierarchicalNavigableSmallWorld.search()` is the
in-tree implementation of an interface an external index can also implement, and its signature grew
the same way — `(searchCondition, context, filter?, minResults?)`. Every future capability that has
to reach an approximate index (a cursor for stable paging, a deadline, a recall target) is another
positional argument, and adding one is a breaking change for any index implemented outside this
repo. The contract is still unreleased, so reshaping it is cheap now and expensive later.

## Suggested shape

Collapse the optional tail into an options object on both sides:

```ts
searchByIndex(searchCondition, transaction, reverse, Table, {
	allowFullScan,
	filtered,
	context,
	minResults,
});

customIndex.search(searchCondition, context, { filter, minResults });
```

The four leading parameters are required at every call site and read fine positionally; it is the
optional tail that should become named. Whether the custom-index `search()` keeps `context`
positional is a judgment call — it is passed by every caller, so either works.

Worth deciding as part of this: whether the options object is also where a future paging cursor
lands, since #2125 documented `offset`/`limit` over an approximate index as best-effort rather than
a stable partition (see the "An approximate index returns at most `ef` rows" section in `DESIGN.md`).

## Scope

- `resources/search.ts` — `searchByIndex` signature, the two recursive call sites, and the
  `customIndex.search()` call.
- `resources/Table.ts:4960` — call site.
- `resources/indexes/HierarchicalNavigableSmallWorld.ts` — `search()` signature.
- Custom-index documentation, if the contract is written down anywhere outside the code.

Mechanical; no behaviour change intended. Raised by @maurice-harper in review of #2125 and recorded
as item 6 of that PR's "For the human reviewer" section, where the positional argument was kept to
avoid reshaping the contract inside a correctness fix.

maurice-harper:
Closes #2165

## Summary

`searchByIndex` had eight positional parameters, five optional, so callers padded with `undefined` or inherited defaults they never chose — and the same shape reached the custom-index `search()` contract. Both now name the optional tail, and a stale positional call throws rather than failing open. No behaviour change.

## Important changes

1. **searchByIndex names its optional tail and rejects a positional one** — `(condition, txn, reverse, Table, { allowFullScan, filtered, context, minResults })`; a non-object 5th argument throws `TypeError`.
   - Why: Without it a stale `false` becomes `undefined`, permitting the scan it meant to forbid.
   - Support: `searchByIndexOptions.test.js`, 3 passing, revert-verified.
   - Where: https://github.com/HarperFast/harper/blob/295c78a85c02aaf8afde8cf0c0296f259c9e321f/resources/search.ts#L293-L296
2. **customIndex.search() takes { filter, minResults }** — `context` stays second positional. HNSW's local `options` becomes `txnOptions`; its stale `@param` list is corrected.
   - Why: A future cursor or deadline becomes a field, not a break.
   - Support: `vectorIndex.test.js` 109, `vectorIndexPlane.test.js` 22.

## Where to focus

- [resources/search.ts:555, the one conflicted hunk](https://github.com/HarperFast/harper/blob/295c78a85c02aaf8afde8cf0c0296f259c9e321f/resources/search.ts#L552-L600) — Main's async branch taken verbatim; only the `customIndex.search()` call line is ours.
- [unitTests/resources/vectorIndexPlane.test.js:84](https://github.com/HarperFast/harper/blob/295c78a85c02aaf8afde8cf0c0296f259c9e321f/unitTests/resources/vectorIndexPlane.test.js#L80-L86) — A positional caller main added post-branch; it auto-merged clean, dropping the filter.

## Risks and boundaries

- Risk: Source-breaking for an out-of-repo custom index — intended while unreleased, and none is reachable.
- Risk: The guard covers `searchByIndex` only; a positional `customIndex.search()` call in plain JS still drops its filter.
- Boundary: No cursor field, typed `CustomIndex` interface, or shim, per the decisions on #2165.
- Boundary: Root `DESIGN.md` is on the spec's file list but unchanged: its mentions carry no parameter list.

## Validation

- `npm run build`, `prettier --check .`, `oxlint` clean.
- searchByIndexOptions 3, vectorIndex 109, vectorIndexPlane 22, query/planner 270, sqlEngine 112 — passing.
- All 24 Integration shards pass; the old head's four failures were main's rocksdb-js 2.9.0 regression (#2556).

<details>
<summary>Dispatch evidence and stage history</summary>

Current head `295c78a` · size class `small` (101 changed production lines) · visible body 299 words (guardrail 150-300).

Evidence:

- Merge resolution: main's async custom-index path (`processEntries`, the SKIP filter, the lazy `ExtendedIterable` promise branch) is taken verbatim; only the `customIndex.search()` call line is this branch's. `git diff origin/main -- resources/search.ts` is the reshape alone.
- The positional `minResults` argument came from #2125 (https://github.com/HarperFast/harper/pull/2125), kept positional there to avoid reshaping a contract inside a correctness fix.
- Declining kriszyp's request for a version gate, per the Team Lead Decision on #2165, taking the alternative that comment itself offers. Re-verified at this head: `resources/indexes/customIndexes.ts` is a hard-coded `{ HNSW }` literal with no registration hook; its only consumer is `resources/databases.ts`; `package.json` exports only `.` to `dist/index.js` and `index.ts` exports none of `CUSTOM_INDEXES`, `HierarchicalNavigableSmallWorld` or `searchByIndex`; the only documented selector is the `@indexed(type: "HNSW")` schema directive. The residual is undocumented user code reaching `Table.indices.<attr>.customIndex.search()` through the exported globals.
- cb1kenobi's two accepted findings (the runtime guard; `options` shadowing plus the stale `@param` list) were recorded as accepted on #2165 but absent at 2caaa2d. Both land here, and all three review threads are answered and resolved.
- Revert verification: removing the `TypeError` fails the new `searchByIndexOptions` case with `Missing expected exception` while the other two still pass. Restoring the positional form in `vectorIndexPlane.test.js` fails two plane tests.
- Integration failures on the old head (2caaa2d): `eviction-secondary-index` "Q1 mid-eviction" and `txnlog-restart-reclaim` "startup purge", across four shards. Both failed identically on `main` itself at e41eca7f5 (run 34433746817) and b40c7a918 (run 34436060103) — 2caaa2d's merge parent was e41eca7f5, so it was built on a broken base. PR #2556 (b349f8ec8, merged 2026-09-10T04:40Z) fixed both: a read-only idle budget not renewed by native range progress, and an over-specified purge-count assertion. b349f8ec8 is an ancestor of this head. Neither test references `searchByIndex` or a custom index; `/DumpE/` is an unconditioned `tables.Expiring.search({})` full scan failing inside `DBIterator.next`.
- Call sites that never threaded `context` still do not; `resources/search.ts:1177` and the recursive relationship sites keep their existing defaults.
- `vectorIndexPlane.test.js` is outside the spec's file list, and in the diff only because main added a positional caller there that this reshape breaks.
- oxlint reports 15 warnings repo-wide, all pre-existing and none in changed files.
- Prior-head benchmark validation: `benchmarks/hnsw-search.js` at 44711b8 held 100% filtered recall at every selectivity, so the named `filter` reaches traversal. `benchmarks/hnsw-scale.js` calls `hnsw.search()` with two arguments and needed no change.
- The full `resources` unit suite aborts natively (SIGABRT) on `origin/main` too; pre-existing, not introduced here.

Prior heads:

- `2caaa2d` — small, 297 words, 3 important change(s), 2026-09-10T04:21:00Z
- `ab88257` — small, 300 words, 3 important change(s), 2026-08-17T03:51:57Z
- `44711b8` — small, 299 words, 2 important change(s), 2026-08-17T02:24:49Z

</details>


resources/Table.ts
```diff
@@ -4962,7 +4962,7 @@ export function makeTable(options) {
 											txnForContext(context).getReadTxn(),
 											false,
 											relatedTable,
-											false
+											{ allowFullScan: false }
```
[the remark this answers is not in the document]
maurice-harper: Not applying this here. On main this call site is `searchByIndex({...}, txn, false, relatedTable, false)` — five positional arguments, so `context` was already `undefined` before this PR. `{ allowFullScan: false }` reproduces that exactly; adding `context` would be a new behaviour change smuggled into a mechanical reshape, which #2165 explicitly scopes out ("Mechanical; no behaviour change intended") and which the spec's acceptance criteria require ("preserve the existing defaults ... without accidental option shifts").

The underlying observation is fair: a relationship resolver that resolves its own read transaction rather than reusing the request's is a real isolation question. It is a pre-existing one, and it belongs in its own change where the transaction semantics can be reasoned about and tested, not in a signature refactor. Flagging it to the human reviewer rather than silently fixing it.

resources/search.ts
```diff
@@ -284,8 +288,7 @@ export function searchByIndex(
 				transaction,
 				reverse,
 				relatedTable,
-				allowFullScan,
-				joined
+				{ allowFullScan, filtered: joined }
```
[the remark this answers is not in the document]
maurice-harper: Not applying this here. On main this recursive call is `searchByIndex({...}, transaction, reverse, relatedTable, allowFullScan, joined)` — six positional arguments, stopping before `context`, so `context` was already `undefined` before this PR. `{ allowFullScan, filtered: joined }` reproduces that exactly. #2165 is explicitly "Mechanical; no behaviour change intended", and the spec's acceptance criteria require these call sites to preserve existing defaults without option shifts, so threading `context` in would be a behaviour change hidden inside a signature refactor.

The underlying observation is fair — the recursive relationship path resolving its own read transaction instead of the caller's is a real isolation question — but it is pre-existing and belongs in a change that can reason about and test the transaction semantics. Flagging it to the human reviewer rather than silently changing it.

resources/search.ts
```diff
@@ -302,8 +305,7 @@ export function searchByIndex(
 						transaction,
 						reverse,
 						Table,
-						allowFullScan,
-						joined
+						{ allowFullScan, filtered: joined }
```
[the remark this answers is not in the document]
maurice-harper: Not applying this here. On main this recursive call is `searchByIndex({...}, transaction, reverse, relatedTable, allowFullScan, joined)` — six positional arguments, stopping before `context`, so `context` was already `undefined` before this PR. `{ allowFullScan, filtered: joined }` reproduces that exactly. #2165 is explicitly "Mechanical; no behaviour change intended", and the spec's acceptance criteria require these call sites to preserve existing defaults without option shifts, so threading `context` in would be a behaviour change hidden inside a signature refactor.

The underlying observation is fair — the recursive relationship path resolving its own read transaction instead of the caller's is a real isolation question — but it is pre-existing and belongs in a change that can reason about and test the transaction semantics. Flagging it to the human reviewer rather than silently changing it.

maurice-harper:
<details>
<summary>Dispatch durable state — QA PASS at <code>295c78a</code> (machine handoff; the human summary is in the Dispatch status comment)</summary>

Behaviours exercised:

- searchByIndex: leading 4 params stay positional; allowFullScan/filtered/context/minResults are read from a 5th options object (default {}), with unindexed-attribute rejection (allowFullScan:false) and permissive full-scan-by-default behavior both exercised against real records via T.put/searchByIndex, not mocks
- searchByIndex throws a TypeError when a stale caller passes a positional (non-object) 5th argument instead of silently treating it as an absent options object
- relationship-recursive searchByIndex calls in resources/search.ts and the Table.ts:6295 relationship resolver call preserve prior allowFullScan/filtered defaults through the new options object, verified as identical to origin/main's own merge-conflict resolution of the same hunk
- HierarchicalNavigableSmallWorld.search()'s 3rd positional argument reshaped from (filter, minResults) to a single {filter, minResults} options object; customIndex.search() call in search.ts keeps context 2nd-positional and passes {filter: recordFilter, minResults} 3rd
- predicate-aware traversal (#1241), int8/quantized filtered search, ef auto-scale/minResults widening, and native-plane (RocksDB) search paths all invoked end-to-end through the reshaped signature

Break attempts:

- Manually reverted resources/search.ts, resources/Table.ts, and resources/indexes/HierarchicalNavigableSmallWorld.ts to their origin/main content (pre-PR) in place, rebuilt, and re-ran the 3 touched test files: 14 tests failed (including both new guard-test assertions and 11 previously-passing filter/minResults tests in vectorIndex.test.js plus a vectorIndexPlane.test.js before-all hook) with 'TypeError: filter is not a function' or missing-exception errors -- confirms every new/changed assertion is behavior-real, not tautological
- Diffed each touched production file (resources/search.ts, resources/Table.ts, resources/indexes/HierarchicalNavigableSmallWorld.ts) against origin/main directly and confirmed byte-for-byte match with the PR's own diff -- ruled out any accidental drift introduced by the merge-conflict resolution beyond the stated reshape
- Grepped every in-tree call site of searchByIndex( and HNSW/customIndex .search( (production, tests, benchmarks) for a straggling positional 5th/3rd argument; none found -- all callers either omit the tail (using the new default {}) or pass the object form
- Checked the allowFullScan===false branch in resources/search.ts against the new test's expectations (permissive-by-default when options omitted, since only strict `=== false` triggers rejection) -- matches, and is exercised against a real Table via T.put/searchByIndex rather than a hardcoded expectation
- Ran full npm run test:unit:main and test:unit:resources at PR head; both SIGABRT partway through, but in describe blocks (copyDbBlobAwait.test.js blob-save tracking; later resources suites) that are untouched by this diff and identical in content to origin/main -- pre-existing environment instability, not a PR regression (matches the pre-existing SIGABRT the user flagged for the resources suite)

Limitations:

- npm run test:unit:main and npm run test:unit:resources both SIGABRT (exit 134) in this sandbox before completing, in test files unrelated to and unmodified by this PR -- worked around by running the 3 PR-touched test files directly (unitTests/resources/searchByIndexOptions.test.js, vectorIndex.test.js, vectorIndexPlane.test.js: 134 passing, 1 pending) rather than the full suite command
- The bundled qa-revert-verify.sh script failed with 'tests-fail-before-revert' because its disposable worktree never runs `npm run build` (this repo's dist/ is required for tests and isn't rebuilt there) -- treated as a script/infra gap, not a product defect, and substituted a manual revert of the same 3 production files in the QA worktree (rebuilt before and after) which produced a clean revert-detected result (14 failures) before restoring PR head
- Did not run the LMDB variant (test:unit:lmdb); the new searchByIndexOptions.test.js explicitly skips under HARPER_STORAGE_ENGINE=lmdb (`if (process.env.HARPER_STORAGE_ENGINE === 'lmdb') return;`), matching an existing convention in unitTests/resources/caching.test.js, not a gap introduced by this PR
- Did not run the differential query corpus vs main or the HNSW benchmark parity check mentioned in the PR-process note; this is a mechanical, behavior-preserving reshape with no logic change (confirmed via the origin/main diffs), so ordinary unit coverage was judged sufficient for a QA verdict

tests: npm run build (tsc) succeeded; npx mocha unitTests/resources/searchByIndexOptions.test.js unitTests/resources/vectorIndex.test.js unitTests/resources/vectorIndexPlane.test.js -> 134 passing, 1 pending, 0 failing, at PR head (295c78a). Full test:unit:main and test:unit:resources both SIGABRT pre-existing/unrelated to the diff (see limitations). npm run test:types passed. · revert: qa-revert-verify.sh failed with tests-fail-before-revert (its disposable worktree skips the required `npm run build`, an infra gap unrelated to the PR). Recovered manually: reverted resources/search.ts, resources/Table.ts, resources/indexes/HierarchicalNavigableSmallWorld.ts to origin/main content in the QA worktree, rebuilt, re-ran the same 3 test files -> 99 passing / 14 failing (revert detected), then restored PR-head content and rebuilt again to leave the worktree clean. · lint: npx prettier --check and npx oxlint --format stylish against every file this PR touches (resources/search.ts, resources/Table.ts, resources/indexes/HierarchicalNavigableSmallWorld.ts, unitTests/resources/searchByIndexOptions.test.js, unitTests/resources/vectorIndex.test.js, unitTests/resources/vectorIndexPlane.test.js, benchmarks/hnsw-search.js): both exit 0, no drift.
Mechanical checks: Inspected the new TypeError guard string ('searchByIndex: the 5th argument is an options object (#2165), not a positional value') against the test's regex /options object \(#2165\)/ -- matches. Confirmed the 'not indexed' ClientError message searchByIndexOptions.test.js asserts on (/not indexed/) exists verbatim in resources/search.ts. No new regexes, HTML/JSON/YAML fragments, or templated strings were introduced by this diff; the only literal-string surface is the two error messages above, both verified against their asserting tests.

| Acceptance criterion | Verdict | Evidence |
| --- | --- | --- |
| Leading params positional, optional tail as named options object (allowFullScan, filtered, context, minResults) | PASS | resources/search.ts searchByIndex signature; unitTests/resources/searchByIndexOptions.test.js all 3 cases |
| Relationship-recursive searches and Table.ts call site preserve existing defaults, no positional undefined padding | PASS | git diff origin/main -- resources/search.ts and resources/Table.ts byte-identical to PR diff; recursive calls at search.ts:320/342 and Table.ts:6295 pass {allowFullScan, filtered: joined} / {allowFullScan: false} |
| customIndex.search(): context stays 2nd positional, filter/minResults via 3rd options object | PASS | resources/search.ts:555 index.customIndex.search(searchCondition, context, {filter: recordFilter, minResults}); no in-tree positional straggler found by grep |
| HierarchicalNavigableSmallWorld.search() behaves as before under the new options object (predicate filtering, minResults widening, per-query ef, threshold comparators) | PASS | unitTests/resources/vectorIndex.test.js 'HNSW predicate-aware traversal (#1241)', 'HNSW construction ef auto-scale (#2180)', 'HNSW limit above the resolved search ef', and threshold/le/lt suites all pass at PR head and fail on manual revert |
| Custom-index contract docs/tests updated to options-object shape; no speculative cursor field, typed interface, or compat shim | PASS | resources/DESIGN.md line updated; grep for interface/cursor/shim additions in the diff found only a docblock comment, no code |
| npm run build, test:unit:resources, and existing HNSW/vector-index tests pass without new dependencies | PASS | build succeeded; no package.json/package-lock.json diff vs origin/main; the 3 PR-touched test files pass in full (134/134); full test:unit:resources SIGABRTs identically on unmodified, untouched files -- pre-existing, not attributable to this diff |

Commands:

- `~/.dispatch-dev-team/scripts/sprint.sh spec-path 2165`
- `~/.dispatch-dev-team/scripts/sprint.sh branch-of 2165`
- `gh pr list --head fix/issue-2165-search-options --json number,headRefOid,state,url`
- `git fetch origin fix/issue-2165-search-options; git checkout -B fix/issue-2165-search-options FETCH_HEAD`
- `npm ci (node v22.22.0 via nvm; v23.11.0 silently skips @harperfast/rocksdb-js-darwin-arm64 per Dev's note, confirmed present under v22.22.0)`
- `npm run build`
- `gh pr diff 2187`
- `git diff origin/main -- resources/search.ts resources/Table.ts resources/indexes/HierarchicalNavigableSmallWorld.ts (and --stat for the whole PR)`
- `npx mocha unitTests/resources/searchByIndexOptions.test.js unitTests/resources/vectorIndex.test.js unitTests/resources/vectorIndexPlane.test.js`
- `npm run test:unit:main; npm run test:unit:resources (both pre-existing SIGABRT, unrelated files)`
- `npm run test:types`
- `~/.dispatch-dev-team/scripts/qa-revert-verify.sh 2165 2187 (failed: infra gap, no build step in its worktree)`
- `manual revert: git checkout origin/main -- resources/search.ts resources/Table.ts resources/indexes/HierarchicalNavigableSmallWorld.ts; npm run build; npx mocha <3 files> (14 failing); git checkout HEAD -- <3 files>; npm run build (restored)`
- `npx prettier --check <touched files>; npx oxlint --format stylish <touched files>`

Dismissed concerns:

- Full npm run test:unit:main / test:unit:resources SIGABRT (exit 134) at PR head — Justification: pre-existing - Crash points (unitTests/bin/copyDbBlobAwait.test.js and later resources suites) are files this PR does not touch; content is byte-identical to origin/main, and the user's own briefing already flagged the resources suite SIGABRT as pre-existing on main. Isolated the 3 PR-touched test files instead, which run clean.
- qa-revert-verify.sh exited non-zero with tests-fail-before-revert — Justification: out-of-scope - Its disposable worktree never runs npm run build, and this repo's tests load dist/ not source (per Dev's env note). This is a script/infra gap, not evidence about the PR; substituted a manual revert (build included) that produced a clean revert-detected result.

Provenance: stage `qa-check`, head `295c78a85c02aaf8afde8cf0c0296f259c9e321f`, recorded 2026-09-15T00:05:55Z.

Prior heads:

- `ab88257` — PASS, 0 blocking finding(s), 2026-08-17T03:57:52Z
- `44711b8` — FAIL, 1 blocking finding(s), 2026-08-17T03:45:54Z


</details>

maurice-harper:
### Dispatch status — head `295c78a`

| Signal | Source | State |
| --- | --- | --- |
| CI | GitHub checks | passing |
| QA | Dispatch `qa-check` stage | PASS ([evidence](https://github.com/HarperFast/harper/pull/2187#issuecomment-5311522022)) |
| Independent review | Dispatch `reviewer` stage — Gemini REST API (degraded) (gemini-flash-latest) | RECOMMEND_MERGE — low risk, 2 lens(es), degraded evidence ([review](https://github.com/HarperFast/harper/pull/2187#issuecomment-5311622746)) |
| Other review | gemini-code-assist[bot] — automatic GitHub app | commented ([open](https://github.com/HarperFast/harper/pull/2187#pullrequestreview-4947986781)) |
| Unresolved blocking findings | QA + independent review | 0 |

**Checked at this head**

- QA exercised searchByIndex, searchByIndex throws a TypeError when a stale caller passes a positional (non-object) 5th…, relationship-recursive searchByIndex calls in resources/search.ts and the Table.ts:6295…, and 2 more; tried to break it with Manually reverted resources/search.ts, resources/Table.ts, and…, Diffed each touched production file (resources/search.ts, resources/Table.ts,…, Grepped every in-tree call site of searchByIndex( and HNSW/customIndex .search(…, and 2 more.
- Independent review inspected searchByIndex options object handling and runtime argument validation, HierarchicalNavigableSmallWorld search signature and variable renaming, relationship recursive calls and Table resolver options mapping, vector index and searchByIndex unit test coverage.

**Accepted with concerns**

- Supplemental verification via claude-runtime-verification was unavailable at this head; the deterministic scripts remain the only gate that passed (No supplemental runtime verification surface was run on 295c78a85; deterministic verify-pr is the only gate that ran. Reviewer leg was Gemini REST fallback (degraded, CLI timed out 480s no TTY), accepted by verifier at low semantic risk.).

**Recommended human action:** Human decision — the pipeline accepted this head with concerns; read them before merging.

Basis: QA PASS at `295c78a` (Dispatch qa-check stage); independent review RECOMMEND_MERGE at `295c78a` (Dispatch reviewer stage, Gemini REST API (degraded) gemini-flash-latest); CI passing; gemini-code-assist[bot] commented.

<details>
<summary>Prior heads and machine state</summary>

- `ab88257` — CI passing, QA PASS, review RECOMMEND_MERGE, 2026-08-17T04:09:45Z
- `44711b8` — CI failing, QA FAIL, review missing, 2026-08-17T03:46:11Z


</details>

maurice-harper:
<details>
<summary>Dispatch durable state — independent review at <code>295c78a</code> (machine handoff; the human summary is in the Dispatch status comment)</summary>

Independent review: no requested changes at `295c78a`.
Inspected searchByIndex options object handling and runtime argument validation, HierarchicalNavigableSmallWorld search signature and variable renaming, relationship recursive calls and Table resolver options mapping, vector index and searchByIndex unit test coverage.
Source: Dispatch `reviewer` stage — Gemini REST API (degraded) (gemini-flash-latest), not the automatic Gemini Code Assist app.
Degraded source: the REST fallback has no repository tool access. It read the staged prompt and the fallback diff only, so this result does not represent repository-level inspection.
Semantic risk: low (size class small); no sensitive surface matched.
Review lenses assigned: correctness, test-integrity.
Outside-family review: yes (author claude, reviewer gemini).
Review legs:
- gemini-api: completed in degraded mode (`degraded`) — api/gemini-flash-latest (family gemini); lenses correctness, test-integrity; degraded: Gemini CLI timed out after 480s; REST fallback has no repository tool access; attempts 2; 506s


</details>

maurice-harper:
Follow-up filed for the one deliberately-declined review theme: #2188 tracks threading `context` through the relationship-recursive `searchByIndex` calls and the `Table.ts` relationship resolver. Pre-existing behavior (those call sites never passed `context` on the old positional signature); kept out of this PR to preserve its no-behavior-change scope. Analysis lives in the resolved Code Assist threads.

cb1kenobi:
Reviewed `ab882571` — no issues found. This PR looks good, nice job!

—
_Generated by Barber AI_

resources/search.ts
```diff
@@ -233,26 +229,34 @@ function composeRecordFilter(recordFilters, table, context): (primaryKey: Id) =>
...
+		minResults,
+	}: {
+		allowFullScan?: boolean;
+		filtered?: any;
+		context?: any;
+		// How many rows the query will ultimately consume (offset + limit), when it is bounded. An
+		// approximate index returns a fixed-size candidate list, so without this a query asking for more
+		// rows than that list holds silently gets a short result set. Only custom indexes read it.
+		minResults?: number;
+	} = {}
```
cb1kenobi, line 295: **Low: a stale positional call silently permits a full scan — the one hazard the PR flags has no guard**

`allowFullScan` is tested with a strict `=== false` (lines 446 and 448), so destructuring a non-object 5th argument yields `undefined` and neither guard fires. I confirmed this at head rather than reasoning about it — `searchByIndex({ attribute: 'unindexedName', value: 'x' }, undefined, false, T, false)`, the exact pre-PR call shape from `Table.ts` on `main`, returns the unindexed row instead of throwing the 404 `"is not indexed"` `ClientError`.

`tsc` rejects that in `.ts` (verified: `npm run typecheck` is clean), so nothing is broken today. But it does not cover the repo's plain-JS callers under `unitTests/` and `benchmarks/`, and the PR body accepts "a missed positional call in plain JS fails silently, not at build time" as residual risk. Since the typed-index-interface mitigation was scoped out by the team-lead decision, a runtime guard is the only remaining one — and it is what makes the hazard this PR's new test file describes in prose actually assertable.

Suggested fix — name the parameter, reject a non-object, then destructure (structural, so not a one-click suggestion):

```ts
	Table: any,
	options: {
		allowFullScan?: boolean;
		filtered?: any;
		context?: any;
		minResults?: number;
	} = {}
): AsyncIterable<Id | { key: Id; value: any }> {
	if (typeof options !== 'object' || options === null)
		throw new TypeError('searchByIndex: the 5th argument is an options object (#2165), not a positional value');
	const { allowFullScan, filtered, context, minResults } = options;
```

That also lets `searchByIndexOptions.test.js` add the case its header comment already describes: assert the stale positional form throws instead of quietly full-scanning.

—
_Generated by Barber AI_

resources/indexes/HierarchicalNavigableSmallWorld.ts
```diff
@@ -961,12 +961,17 @@ export class HierarchicalNavigableSmallWorld {
...
+	 * This is also the contract an index implemented outside this repo has to satisfy, so everything
+	 * optional is named: a future capability (a paging cursor, a deadline, a recall target) is a new
+	 * field on `options` rather than a positional argument that breaks every existing implementation.
 	 * @param target
 	 * @param value
 	 * @param descending
 	 * @param distance
 	 * @param comparator
 	 * @param context
+	 * @param options
```
cb1kenobi, line 1070: **Nit: `options` now names two different things in `search()`**

This new `@param options` documents the third parameter, but the body already uses `options` for something unrelated — `const options = context.transaction` (line 1038), the nested RocksDB transaction handed to `getEntryPoint(options)` and `searchLayer(..., options, ...)`.

Nothing breaks today, because the third parameter is destructured inline and never bound to a name. The cost is on the next edit: giving that parameter a name — which is exactly what adding a runtime type guard would require — collides with the local.

Renaming the local to `txnOptions` (3 uses) frees the name and makes the JSDoc unambiguous.

—
_Generated by Barber AI_

resources/search.ts
```diff
@@ -510,21 +512,23 @@ export function searchByIndex(
...
-					});
-					if (!loadedEntry) return SKIP; // record was deleted/expired or not yet visible
-					freezeRecord(loadedEntry?.value);
-					recordRead(loadedEntry);
-					return { ...otherProps, ...loadedEntry };
-				}
-				return entry;
-			});
+			const loaded = index.customIndex
+				.search(searchCondition, context, { filter: recordFilter, minResults })
```
kriszyp, line 556: This is a breaking change to the external custom-index contract. An existing `search(condition, context, filter, minResults)` implementation now receives an options object as `filter`; invoking that truthy value fails with `TypeError`. Conversely, a legacy direct HNSW caller silently loses its filter because `resources/indexes/HierarchicalNavigableSmallWorld.ts:995-1009` destructures the function as an object with no `filter` field. Please add a version/feature gate or transitional normalization for the old signature, or confirm and document that every out-of-repo implementation and caller will be updated atomically.

— KrAIs (Codex)

kriszyp (dismissed):
**Automated gate — not yet queued for human review.**

This PR's AI review found issues, and the PR description reports no cross-model reviews.
Per team policy, a substantive PR with AI-review findings is queued for human review only after at least 2 cross-model reviews have been run, their findings addressed, and the coverage reported in the PR description (`## Review coverage` naming each model — see harper-engineering-guidelines).
The findings below count as one of the two: address them, run a second outside-model review, update the description, and the gate lifts automatically on the next pass.

---
## TL;DR

This refactors growing optional positional parameters into named options objects.
The approach addresses the root cause, preserves in-repo behavior, and adds negligible hot-path overhead.
One external-contract migration concern remains.
Independent review and tests were unavailable because this read-only checkout has no installed dependencies; syntax and whitespace checks passed.

verdict: CHANGES
merge: rebase
Human-Review-Need: 4 @ ab882571e4b0

## Change tour

- `resources/search.ts:143-148`, `resources/search.ts:241-259` — query execution and `searchByIndex` now use named options with permissive omission defaults.
- `resources/search.ts:281-309`, `resources/Table.ts:4960-4966` — recursive relationship searches preserve `allowFullScan` and join-filter propagation.
- `resources/search.ts:509-536`, `resources/indexes/HierarchicalNavigableSmallWorld.ts:960-1009` — the custom-index boundary and HNSW implementation adopt `{ filter, minResults }`; traversal behavior remains unchanged at `resources/indexes/HierarchicalNavigableSmallWorld.ts:1058-1095`.
- `benchmarks/hnsw-search.js:226`, `resources/DESIGN.md:96` — benchmark and design documentation reflect the new shape.
- `unitTests/resources/searchByIndexOptions.test.js:35-52`, `unitTests/resources/vectorIndex.test.js:809-813`, `unitTests/resources/vectorIndex.test.js:1955-2045` — coverage checks omitted defaults, `allowFullScan`, filtering, and minimum-result behavior.

## Findings

major — `resources/search.ts:516` — the unversioned argument-shape change breaks existing out-of-repo custom indexes and legacy direct HNSW callers

## Verification

- Passed `node --check` on every changed TypeScript/JavaScript file.
- Passed `git diff --check origin/main...HEAD`.
- Unit tests and lint were not run: dependencies are absent and the sandbox prevented installation.
- Cross-model review setup failed before sharing the diff because its artifact directory is read-only.

## Review coverage

| lens | outcome |
|---|---|
| gemini | **pruned** — pruned (policy minimal) |
| cursor-grok | **pruned** — pruned (policy minimal) |
| cursor-composer | **pruned** — pruned (policy minimal) |
| codex | ok — graded leg — produced review.md + comments.json |
| domain | **pruned** — pruned (policy minimal) |


---
_Pre-push review of `fix/issue-2165-search-options` (ab882571e4b0) vs `origin/main` by codex._
_Review emphasis: Dispatch-configured._

— codex review, submitted by the dispatch review gate

resources/indexes/HierarchicalNavigableSmallWorld.ts
cb1kenobi: All of the `@param`s need fixing, but not blocking it.

cb1kenobi (approved):

heskew (approved):
# 🟢 APPROVE harper#2187 — Use an options object for searchByIndex and the custom-index search() contract

**Event:** `APPROVE`  
**Head:** `ab882571e4b050f7afda417773cd97a9d632a6a3`  
**Size:** +124/−58, 7 files

Approved for this exact head. The intended unreleased-contract break is implemented consistently: every in-tree `searchByIndex`, HNSW, and custom-index caller is converted to the named options object or legitimately omits it. Build, the 16 focused behavior tests, diff checks, and full CI all pass; the one nondeterministic local HNSW greedy-routing failure is outside the changed hunks and does not indicate a regression here.

Existing feedback is non-blocking: @gemini-code-assist's three omitted-context sites preserve pre-PR behavior and are outside the no-behavior-change scope; @kriszyp's custom-index signature concern is the explicit purpose of #2165 while the contract remains unreleased; and @cb1kenobi's stale plain-JS runtime guard plus naming/JSDoc note are low-risk follow-ups given all in-tree JS callers are converted, typed callers fail at compile time, and @cb1kenobi has approved. No new Codex finding.

The branch currently conflicts with `main` only in `resources/DESIGN.md`; implementation files merge cleanly. Resolve that documentation conflict and re-run/review the rebased head before merge as repository policy requires.

resources/search.ts
```diff
@@ -550,21 +552,23 @@ export function searchByIndex(
...
```
kriszyp: This is a breaking change to the external custom-index contract. An existing `search(condition, context, filter, minResults)` implementation now receives an options object as `filter`; invoking that truthy value fails with `TypeError`. Conversely, a legacy direct HNSW caller silently loses its filter because `resources/indexes/HierarchicalNavigableSmallWorld.ts:1091-1105` destructures the function as an object with no `filter` field. Please add a version/feature gate or transitional normalization for the old signature, or confirm and document that every out-of-repo implementation and caller will be updated atomically.

— KrAIs (Codex)

resources/search.ts
maurice-harper: Accepted, and implemented as suggested, in 295c78a85.

`searchByIndex` now names its 5th parameter, rejects a non-object, then destructures — the shape from your suggestion, with the existing field comments kept.

`unitTests/resources/searchByIndexOptions.test.js` gained the case its header comment already described: the exact pre-PR shape `searchByIndex({ attribute: 'unindexedName', value: 'x' }, undefined, false, T, false)` must throw the TypeError rather than return the unindexed row. Revert-verified — with the guard removed that case fails with `Missing expected exception` while the other two still pass, so it pins the guard and not the surrounding reshape.

Your point about plain-JS callers was not hypothetical. Merging current `main` into this branch brought in `unitTests/resources/vectorIndexPlane.test.js`, which calls `customIndex().search(cond, ctx, filter)` positionally. It auto-merged cleanly and tsc has nothing to say about it (it is `.js`). Left as written, two plane tests fail because the filter is silently dropped. Moved to the named form in the same commit.

resources/indexes/HierarchicalNavigableSmallWorld.ts
maurice-harper: Both accepted, fixed in 295c78a85.

The local `const options = context.transaction` is now `txnOptions`, freeing the name for the parameter. It is four uses at current head, not three — `main` added `searchPlane(plane, target, effectiveEf, filter, filterState, options)` for the native plane since you wrote this.

The `@param` list was stale in the way you flagged: it named the destructured fields of the first parameter as if they were parameters of their own and never mentioned the third. It now describes the three real parameters — `searchCondition`, `context`, `options` — with one line each.

resources/search.ts
maurice-harper: Taking the second branch you offered — confirming that every implementation and caller is in this repo and updated atomically — rather than adding a gate or transitional normalization. Recorded as a Team Lead Decision on #2165.

The evidence, re-checked at the current head rather than carried over:

- `resources/indexes/customIndexes.ts` is a hard-coded object literal, `CUSTOM_INDEXES = { HNSW: HierarchicalNavigableSmallWorld }`, with one live entry and a commented-out `ZOrder` TODO.
- Its only consumer is `resources/databases.ts`, which looks up `CUSTOM_INDEXES[attribute.indexed.type]`. There is no registration function and no mutation of that object anywhere in the tree.
- Neither `CUSTOM_INDEXES`, `customIndexes.ts`, `HierarchicalNavigableSmallWorld`, nor `searchByIndex` is reachable from the package's public entry point: `package.json` exports only `.` → `dist/index.js`, and `index.ts` exports none of them.
- The only documented way to select an index type is the schema directive (`@indexed(type: "HNSW", ...)`), which names a registry key. Nothing documents supplying an implementation.

So an out-of-repo `search(condition, context, filter, minResults)` implementation has no way to be reached by `search.ts` in the first place, and the in-tree callers are updated in this PR. That includes the ones `main` added after the branch point: `unitTests/resources/vectorIndex.test.js` and, in this merge, `unitTests/resources/vectorIndexPlane.test.js`.

Two things I do not want to overstate:

- The residual is undocumented user code reaching `Table.indices.<attr>.customIndex.search()` through the exported `tables`/`databases` globals. That was never a supported surface, but it is not structurally impossible the way an out-of-repo implementation is.
- Your second failure mode — a legacy direct caller silently losing its filter because a function destructures to `{ filter: undefined }` — is real, and this merge reproduced it exactly. `main`'s new `vectorIndexPlane.test.js` calls `customIndex().search(cond, ctx, filter)` positionally; it auto-merged clean, and two plane tests fail with the filter dropped. Fixed in 295c78a85. That is the argument for landing the reshape now rather than later, while the set of callers is still closed.

`searchByIndex` additionally got a runtime `TypeError` on a non-object 5th argument in the same commit (cb1kenobi's thread), so a stale positional call to the outer entry point fails loud instead of silently. It rejects the old shape rather than accepting it, so it is a guard and not a compatibility shim.

A typed `CustomIndex` interface plus a real registration API would be the right way to make this a supported extension point. That is a separate issue, not a shim here — happy to file it if you want it.

cb1kenobi (commented):
This is a clean mechanical refactor: searchByIndex and the custom-index search() contract now take their optional arguments as a named object instead of a long positional tail. I checked every converted call site against its pre-PR form and all of them pass the same values, so there is no behaviour change. The new runtime TypeError guard, the test file, and the updated docs all match the real signatures. The only non-blocking note is that the guard catches a stale `false` but not a stale padded `undefined`, which nothing in the repo can reach today.

—
Reviewed 295c78a8

cb1kenobi (commented):
This is a mechanical reshape of searchByIndex and the custom-index search() contract from a long positional tail to a named options object, and every converted call site passes the same values it did before. The new TypeError guard correctly rejects stale positional calls while still letting an omitted options object fall through to the permissive defaults. Nothing here touches replication, storage format, auth, or transaction propagation. No blocking issues found; the one remaining gap — a positional customIndex.search() call in plain JS still silently drops its filter — is already documented and was accepted in earlier review threads.

—
Reviewed 295c78a8

cb1kenobi (commented):
No blocking issues were found at the specified head. The options-object conversion preserves existing in-tree behavior, and prior concerns are addressed or already tracked.

—
Reviewed 295c78a8

cb1kenobi (commented):
No confirmed blocking issues were found on changed lines. The options-object conversion preserves existing in-tree behavior, and earlier concerns were addressed or already tracked.

—
Reviewed 295c78a8

maurice-harper:
@kriszyp , on #2187 we declined the shim for the old positional search(condition, context, filter, minResults) shape because CUSTOM_INDEXES is hard-coded with HNSW as its only entry and nothing exports the class or searchByIndex, so no code outside the repo can implement or call the contract; we replied on your thread with that evidence and resolved it. Are you okay with the options-object contract landing in 5.3 as-is, with the typed CustomIndex interface and registration API tracked separately in #2611?

cb1kenobi (commented):
This is a mechanical reshape of searchByIndex and the custom-index search() contract from a positional tail to a named options object. I checked every converted call site against its pre-PR argument order and each passes the same values, so there is no behavior change. The new TypeError guard makes a stale positional call fail loudly instead of silently permitting a full scan. The remaining concerns (omitted context, the plain-JS positional custom-index call, and the stale padded undefined case) were all raised and answered earlier on the same lines, so nothing new is blocking.

—
Reviewed 295c78a8

kriszyp (commented):

kriszyp (approved):
Looks good.

kriszyp (approved):
Looks good.

PR #2173, opened 2026-08-14
feat: OIDC trusted publishing — deploy from CI with no stored credential

dawsontoth:
Closes #2171. **Now includes the token operation-scoping work formerly in #2174** (merged into this branch 2026-08-17, so this reviews as one whole rather than a stack).

> **Two things predate parts of this description:** (a) the module moved `security/oidcTrust/` → `security/authn/oidc/` with the core made issuer-agnostic behind provider profiles (`providers/githubActions.ts`, `providers/generic.ts`); (b) the AuthAuditLog item in "Open / not done" is now **done** — exchanges emit `AuthAuditLog` on success and failure via the `baseRequest` `serverHandlers` injects for `NO_AUTH_OPERATIONS`. The security shape below still holds.

## Folded in from #2174 — token operation scoping (has an open design question)

An OIDC trust policy may carry an optional `operations` list; the exchanged token carries it as a claim, and `verifyPerms`/`verifyPermsAST` intersect it with the user's role, so one Harper user can back several workflows each holding a credential narrower than the user itself. Absent claim = today's behavior exactly.

**Open design question, unresolved — do not merge to `main` before it is settled:** @kriszyp asked for per-policy operation trust; @heskew reviewed and preferred it dropped ("one authorization system; least privilege lives in the role"). Folding #2174 in here does **not** decide that — it only stops juggling a stack. If the answer is "drop it," the four token-scoping commits revert cleanly off the top.

**The invariant is "can only ever subtract," and getting there took finding six bypasses** (across three review rounds plus a local cross-model pass), which is the strongest argument for reviewing this part closely — or for dropping the feature per @heskew, since the attack surface is this wide:

1. The scope gate is the **first** authorization step in both `verifyPerms` and `verifyPermsAST`, ahead of the super_user bypass and the `operations` gate-2 grant, which both `return null` early.
2. It is **never merged into `role.permission.operations`** — that field is not purely narrowing (gate 2 treats an explicit SU-only listing as a grant), so merging could *widen*. It travels separately as `tokenOperations`.
3. It is enforced on the **SQL path** (`verifyPermsAST`), mutually exclusive with `verifyPerms` — a gate in only one let a `get_status`-scoped token run arbitrary SQL until fixed. It also gates on the **API operation the caller sent** (`requestJson.operation`), not the handler name — the handler-name form both denied the feature's own `deploy_component` and let a read-scoped token ride an `export_local`/`export_to_s3` job (SQL and NoSQL variants) to exfiltrate data.
[...]

PR #2202, opened 2026-08-18
fix(security): honor the SQL permission denial processAST computes

dawsontoth:
`processAST` computed a SQL permission denial and then discarded it. The guard read
`permissionsCheck && permissionsCheck.length > 0`, but `checkASTPermissions` returns `null` or a
`PermissionResponseObject` — which has no `length`, so `undefined > 0` was always false. **No denial
reaching that branch has ever been honored.**

Fixing the guard is one line. The rest of this PR is what the live guard exposes — the work the
[ordering note](https://github.com/HarperFast/harper/pull/2202#issuecomment-5332131086) on this PR asked
whoever picked it up to do, plus one thing that note did not anticipate.

- **`sqlTranslator/index.ts`** — bare truthiness, and read the auth async-context state once (behind
  the `permissions_checked` short-circuit, so the already-checked call pays nothing).
- **`server/serverHelpers/operationAuthorizationState.ts`** — the existing store now carries
  `{bypassAuth, apiOperation}` rather than a bare boolean, plus `runWithDispatchedOperation`. Both
  wrappers *preserve* an existing carrier; two frozen shared objects serve the common no-carrier
  path, so it allocates only when there is something to carry.
- **`server/jobs/jobProcess.ts`** — establishes the carrier in the job worker, and sanitizes the
  persisted row on load (both `parsed_sql_object` positions). Without the carrier an
  `export_local`-scoped credential is refused on its own export, since the worker re-parses the
  nested `search_operation` and the check sees `sql`.
- **`server/serverHelpers/serverUtilities.ts`** — deletes a body-supplied
  `search_operation.parsed_sql_object` at dispatch, next to where `bypass_auth` is already stripped.
  Only the nested one here: the top-level is overwritten by this dispatch's own parse.
- **`server/jobs/jobs.ts`** — `getJobsInDateRange` runs Harper's own fixed `system.hdb_job` query
  through `evaluateSQL` and `SqlSearchObject` hardcodes `operation: 'sql'`; that statement is
  bypassed rather than carried (see #5 below — they are not interchangeable).

## For the human reviewer

1. **A theoretical double-evaluation divergence, raised then largely refuted — recorded so you can
   decide, not to alarm you.** For a **non**-super-user, `verifyPermsAST` runs the full table check
   regardless of `tokenOperations`, and the export path now checks the statement twice: once in
   `chooseOperation`, once in the worker after the forced re-parse. Those runs read different state,
   because `jobProcess.ts` snapshots `global.hdb_schema` before loading built-in components. A
   reviewer raised this as a possible regression: a role exporting a component-defined table passing
   at dispatch and being denied in the worker. **Four independent adjudication passes then went
   looking for a concrete failing case; three traced the live registry and persisted table metadata
   and found none, one retained it as conditional.** So the divergence is real in principle and no
   one could construct it in practice — component-defined tables are in the persisted schema the
   worker snapshots. **Ruled by the author: ship declared.** The candidate fix reorders job worker
   startup — a different change with its own blast radius — and three of four passes failing to
   construct the failure is meaningful evidence against trading a hypothetical regression for a real
   one in job startup. Recorded rather than buried because it is the one place this change could
   plausibly break a working non-super-user export; if a reviewer disagrees, the fix is one line in
   `jobProcess.ts`.

2. **The carrier is `request.operation` — a body-derived field, which the ordering note on this PR
   explicitly ruled out. Please rule on why this instance is different.** The note's constraint was
   "a carrier a client cannot set", because on the direct-SQL path the body is client-supplied and
   the AST check is the only gate. `request.operation` is body-derived, but in the job worker it is
   the same field `serverUtils.getOperationFunction(request)` reads to select the handler that then
   runs. Forging it therefore changes *which operation executes*, not *which policy is applied to a
   different one* — the two cannot diverge. That is the whole security argument, and it is why a
   *new* field (`api_operation`, tried and reverted in #2173) is unsafe while this one is not. If you
   don't accept it, the design changes.
3. **A second finding, not in the original scope: `permissions_checked` was forgeable, at two
   layers.** `evaluateSQL`
   trusts a supplied `parsed_sql_object` verbatim and skips parsing; `chooseOperation` overwrote only
   the **top-level** one; `dataLayer/export.ts` hands the nested `search_operation` straight to
   `evaluateSQL`. So a request carrying `search_operation.parsed_sql_object` with
   `permissions_checked: true` would have run an arbitrary AST with the newly-live check skipped —
   i.e. this PR would have advertised a gate that a request body could switch back off. Now deleted
   at dispatch, forcing a re-parse of the `sql` string that dispatch authorized. This is the same
   defect as #2217's item 3; whichever lands second should drop its copy.

   **What this is, precisely: a per-field allowlist, not an invariant.** It closes every path that
   exists today — `evaluateSQL:40` is the only consumer, and the only client-reachable objects handed
   to it are the top-level request and `search_operation` — but the next `evaluateSQL` caller that
   forwards a client object defeats all three deletes silently. **Ruled by the author: keep the
   deletes here, move the invariant in a follow-up** — #2304 proposes honoring a supplied
   `parsed_sql_object` only under a trusted dispatch marker in async context, reusing the mechanism
   this PR introduces. Deliberately not bundled: it changes a hot path and deserves its own review
   rather than being added to a change already six rounds deep. **The dispatch-time strip
   alone is not sufficient**, which is the second layer: the worker re-enters from the persisted
   `hdb_job` row rather than re-dispatching, so a job queued before this ships — or a row written
   directly — would still carry a forged object. `jobProcess.ts` therefore repeats the delete when it
   loads the row.
4. **One store with two fields, and the carrier *survives* a bypass.** I first had it dropped, on
   the reasoning that checks reading it don't run under a bypass — true of the bypassed branch, false
   of the enforced one, which a job handler reaches when it dispatches a nested authorized operation.
   Preserving it fixes a spurious mid-job 403; the consequence to accept is the other direction, that
   such a nested dispatch is judged against the **outer** job's operation for any `evaluateSQL` not
   passing through `chooseOperation`. One-line flip either way, and the semantics choice is the
   security argument.
5. **`getJobsInDateRange` takes the bypass, and the reason is a trap worth knowing.** I first
   switched it to the carrier on the grounds that a bypass also drops table checks and the carrier
   would "fail closed if the handler is ever opened wider". That was wrong: `verifyPermsAST`'s
   super_user early return is `isSuperUser && !isSuSystemOperation`, so a `system` schema is
   **exempt** and the table check really does run. The carrier would therefore have put Harper's own
   query through `hasPermissions` on `system.hdb_job`, passing only because
   `appendSystemTablesToRole` grants `system.*.read` to a *hydrated* super_user — a super_user
   principal without an appended `permission.system` (an impersonation payload, or any path skipping
   user-cache hydration) would have started getting 403s. Reverted to the bypass, which is also the
   honest statement of intent: the query is Harper's, not the caller's.
6. **`evaluateSQL`'s error contract is preserved, not fixed.** It drops its second callback argument
   on error, so a denied job's durable message is the bare string `403` with no reason. Deliberately
   unchanged — every caller shares that contract — but denied jobs are consequently hard to
   diagnose. Happy to fix separately.
7. **The nested-SQL dispatch branch is still scope+AST only.** Being mutually exclusive with
   `verifyPerms`, it does not enforce `export_local`'s `requires_su` when
   `search_operation.operation === 'sql'`. Pre-existing and adjacent; closing it changes
   authorization outcomes, so it is flagged rather than bundled. #2217 approaches it from the other
   side.

## Verification

- **The tripwire is answered.** `admits an in-scope export job through the path export.ts actually
  dispatches` (`unitTests/security/tokenOperationScope.test.js`) — planted to go red when this lands
  — now asserts the fixed behaviour via the carrier, alongside two new cases pinning the other
  direction (a `sql`-only scope cannot start an export; no carrier fails closed).
- **Fails-on-base.** With the guard reverted: **6 failing / 37 passing**; with it, **43 passing**.
  The 6 include the pre-existing GHSA-7h8h-wq7f-qx65 bypass test, which passed on base only because
  it stubbed `['denied']` — an array *has* `length`, so it satisfied the broken guard. Its stub now
  uses a real `PermissionResponseObject`. The `parsed_sql_object` strip was separately confirmed to
  fail on base.
- **Unit gates, per directory** (the whole-suite `test:unit:main` invocation is not usable locally —
  see below). Every number below is on the final tree:

  | | this branch | clean `origin/main` |
  |---|---|---|
  | security | 663 ✓ / **2 fail** | 661 ✓ / **the same 2** |
  | server (all subdirs) | 744 ✓ / **1 fail** | serverHelpers 297 ✓ / **the same 1** |
  | sqlTranslator · sqlEngine | 67 ✓ · 103 ✓ | — |
  | dataLayer · validation · utility(ex-logging) | 242 ✓ · 242 ✓ · 517 ✓ | — |
  | agent · bin · config · install · build-tools · buildTools | 87 · 239 · 248 · 9 · 16 · 13 ✓ | — |
  | upgrade · root | 19 ✓ · 8 ✓ | — |
  | utility/logging | **stalls** | **stalls too** |
  | components | 4 fail | **14 fail** (superset, name-identical) |

  Every failure and the stall reproduce on clean `main` with none of this branch's code present.
  `utility/logging` hanging on `main` is what makes `test:unit:main` unusable here: `.mocharc` sets
  `--timeout 0`, so it goes quiet instead of failing. `components` cannot be attributed by rerun —
  its suite kills the process group of its own invoker — so it is attributed by reachability
  instead: nothing under `components/` or `unitTests/components/` references any module this PR
  touches.
- **Review coverage decayed across rounds, and the legs that went dark are the ones that caught my
  worst errors.** Four rounds. codex (opposite-family) and the Harper domain adjudicator ran all
  four. Gemini ran **round 1 only** (round 2 died on a denied permission prompt, rounds 3–4 on
  quota). cursor-composer ran **rounds 1–2** then hit the branch's Cursor round cap. So rounds 3 and
  4 were two lenses, and round 3's delta contained an authorization revert. cursor-composer is what
  caught the false `verifyPermsAST` premise and the persisted-row exposure. Four separate rounds
  corrected a confident claim of mine in this area; the pattern each time was prose asserting
  something the code did not do. Weigh my summaries here accordingly and read the carrier argument
  yourself.
- **Six rounds ran; the machine footer below reads `ran=none` and that is misleading in the strict
  direction.** The receipt is at `273de2aa` (codex + harper-domain, `Adjudicated-Severity: minor`,
  `Human-Review-Need: 3`). The only commit since is a DESIGN.md paragraph rewrite — no production
  code, no tests. I chose to fix the documentation and lose the matching receipt rather than ship a
  DESIGN section I knew was wrong, because DESIGN.md is the artifact future readers rely on.
- Rounds 4–6 changed **no production logic at all**. They corrected documentation that had drifted
  from the code: a paragraph stating the *inverse* of the carrier's behaviour, another asserting the
  super_user early return applies to system schemas, and a third conflating three different
  mechanisms all called "scoped token". The code converged at round 3; my description of it took
  three more rounds.
- **One review conclusion I did not accept.** Round 6 dropped the "`jobs.ts` bypass has no coverage"
  finding on the grounds that `northwind.test.mjs`'s `Search Jobs by date` would go red without it. I
  traced it: that test runs as a hydrated super_user, for whom `appendSystemTablesToRole` grants
  `system.*.read`, so `hasPermissions` passes either way; and the adjacent non-super-user case is
  refused at `verifyPerms` before any SQL. The gap flagged in rounds 3–5 is real, and the claim in
  this description stands.
- Planning review cleared `Framing-Verdict: chosen-approach-sound` before implementation.

Complexity: complicated

<sub>Review-Coverage: authored=unknown; ran=none; rounds=1 @ d817c1242e8c</sub>

<sub>Human-Review-Need: 4 @ d817c1242e8c</sub>

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -391,6 +391,80 @@ early-returns (super_user, structure_user, system-table allowances): persisted r
 `super_user` with other permission keys, but inline roles can combine `structure_user` with an
 allowlist, and the gate ordering is what keeps unlisted schema ops unreachable.
 
+## The dispatched API operation is carried on async context, never on the request (`server/serverHelpers/operationAuthorizationState.ts`)
+
+`verifyPermsAST`'s token-scope check has to be told which top-level API operation the caller
+invoked, because the scope is written in that namespace (`sql`, `export_local`, ...). Two things
+make that awkward:
+
+1. On the **direct-SQL** path, the object handed to `checkASTPermissions` _is_ the client's request
+   body, and this check is the only gate there (`chooseOperation`'s `sql` branch is mutually
+   exclusive with its `verifyPerms` call). Any field read off that object is therefore a way to
+   name whichever operation the caller's scope happens to allow and run arbitrary SQL under it.
+   `jsonMessage.operation` is safe only because dispatch already routed on that same field, so it
+   cannot disagree with the operation running. Never add another.
+2. A **job** re-parses its SQL from the nested `search_operation` in a _different_ async context —
+   `executeJob` persists the request and hands off to the job runner, and `jobProcess.ts` re-enters
+   from the `hdb_job` record. So a store established around the originating request cannot reach
+   it, and the re-parse would be judged as `sql` rather than as the job's own operation.
+
+The carrier is therefore established **in the job worker**, by `runWithDispatchedOperation`, from
+the same `request.operation` that `getOperationFunction` just resolved the handler from. That
+identity is the whole basis for trusting it: the value naming the operation and the value selecting
+the code cannot diverge. A new carrier must preserve that property — an added request property, a
+`search_operation` field, or a persisted `parsed_sql_object` would not.
+
+This lives in the same `AsyncLocalStorage` as the auth bypass rather than a second store, so
+`processAST` reads the state once. `runWithOperationAuthorizationBypass` **preserves** an existing
+carrier on both branches. That is deliberate and was initially got wrong: its enforced branch is not
+a bypass, so a job handler dispatching a nested _authorized_ operation lands there, and dropping the
+carrier would judge that job's re-parsed SQL as the inner `sql` and refuse it partway through its own
+work. The consequence to know is the other direction — a nested dispatch inside a job is judged
+against the **outer** job's operation for any `evaluateSQL` that does not pass through
+`chooseOperation`. It allocates only when a carrier is present; with none, two shared frozen objects
+serve the common path. All four stores are frozen, so `getOperationAuthorizationState()` cannot hand
+a mutable one to a caller.
+
+It has four call sites, and they are not all dispatch wrappers: `server.operation()`
+(`serverUtilities.ts`), the ITC path (`registeredOperations.ts`), the legacy SQL engine
+(`sqlEngine/diff/differential.ts`), and Harper's own `hdb_job` query (`server/jobs/jobs.ts`) — that
+last one **is** reached from the ops-API dispatch, via `search_jobs_by_start_date` →
+`handleGetJobsByStartDate` → `getJobsInDateRange`.
+
+Harper's own internal SQL takes the bypass, not the carrier. `getJobsInDateRange` runs a fixed
+`system.hdb_job` query through `evaluateSQL` beneath a handler the caller was already authorized for,
+and `SqlSearchObject` hardcodes `operation: 'sql'` — so the same mismatch applies, but the answer
+differs, and the reason is easy to get backwards. `verifyPermsAST`'s super_user early return is
+`isSuperUser && !isSuSystemOperation`, so a `system` schema is **exempt** from it and the table check
+genuinely runs. A carrier would therefore put Harper's own query through `hasPermissions` on
+`system.hdb_job`, which passes only because `appendSystemTablesToRole` grants `system.*.read` to a
+hydrated super_user — a super_user principal without an appended `permission.system` (an
+impersonation payload, or any path that skips user-cache hydration) would start getting 403s on an
+operation it is entitled to. The bypass also states the actual intent: the statement is Harper's, not
+the caller's. Wrap the individual statement, not the function — a later caller-dependent statement
+must not inherit it.
+
+A second body field has to be neutralized for any of this to hold: `evaluateSQL` trusts a supplied
+`parsed_sql_object` verbatim and skips parsing, `chooseOperation` overwrites only the **top-level**
+one, and `dataLayer/export.ts` hands the nested `search_operation` straight to `evaluateSQL`. So a
+body-supplied `search_operation.parsed_sql_object` carrying `permissions_checked: true` would run an
+arbitrary AST with the check skipped. `chooseOperation` deletes it, forcing the worker to re-parse
+from the `sql` string that dispatch authorized — the nested object is never overwritten the way the
+top-level one is, because nothing downstream should read one at all.
+
+What is untestable is not the carrier's contract — unit tests cover that by calling
+`runWithDispatchedOperation` directly — but that `jobProcess` is what establishes it. Delete that call
+and those tests stay green. The carrier only changes an outcome through `tokenScopeDenial`, which is
+inert unless the principal carries `tokenOperations`, and that property has exactly one origin: an
+OIDC trust-policy exchange, for which there is no integration harness.
+
+Three different mechanisms are easy to conflate here. `tokenOperations` above is the **OIDC token
+operation scope** (#2174). An **inline-role scoped token** (`create_authentication_tokens` with a
+`role` object) is not the same thing and cannot substitute, because `createScopedToken` mints it
+`super_user: false`, so it cannot invoke a `requires_su` operation such as `export_local` at all.
+**Table permissions** are a third, and also cannot substitute — see the system-schema exemption
+above. See #2298.
+
 ## TLS hot-reload: cert vs. private key follow two different propagation paths (`security/keys.ts`)
 
 A renewed **certificate** and a renewed **private key** reach a worker's live TLS secure context
--- a/server/jobs/jobProcess.ts
+++ b/server/jobs/jobProcess.ts
@@ -9,6 +9,8 @@ import harperLogger from '../../utility/logging/harper_logger.ts';
 import * as globalSchema from '../../utility/globalSchema.ts';
 import * as user from '../../security/user.ts';
 import * as serverUtils from '../serverHelpers/serverUtilities.ts';
+import { runWithDispatchedOperation } from '../serverHelpers/operationAuthorizationState.ts';
+import { stripSuppliedParsedSqlObject } from '../serverHelpers/requestSanitization.ts';
 import moment from 'moment';
 import * as jobs from './jobs.ts';
 import { cloneDeep } from 'lodash';
@@ -54,12 +56,16 @@ const JOB_ID = JOB_NAME.substring(4);
 			throw new Error('Did not find job request in hdb_job table, unable to proceed');
 		}
 		request = cloneDeep(request);
+		// The worker re-enters from the persisted row rather than re-dispatching, so a row queued before
+		// the dispatch-time strip existed — or written directly — is sanitized here.
+		stripSuppliedParsedSqlObject(request);
 
 		const operation = serverUtils.getOperationFunction(request);
 		harperLogger.trace('Running operation:', request.operation, 'for job', JOB_ID);
 
-		// Run the job operation.
-		const results = await operation.job_operation_function(request);
+		const results = await runWithDispatchedOperation(request.operation, () =>
+			operation.job_operation_function(request)
+		);
 		harperLogger.trace('Result from job:', JOB_ID, results);
 
 		jobObj.status = hdbTerms.JOB_STATUS_ENUM.COMPLETE;
--- a/server/jobs/jobs.ts
+++ b/server/jobs/jobs.ts
@@ -11,6 +11,7 @@ import * as search from '../../dataLayer/search.ts';
 import Search_Object from '../../dataLayer/SearchObject.ts';
 import searchByHashObj from '../../dataLayer/SearchByHashObject.ts';
 import SQL_Search_Object from '../../dataLayer/SqlSearchObject.ts';
+import { runWithOperationAuthorizationBypass } from '../serverHelpers/operationAuthorizationState.ts';
 import * as hdbTerms from '../../utility/hdbTerms.ts';
 import JobObject from './JobObject.ts';
 import UpdateObject from '../../dataLayer/UpdateObject.ts';
@@ -249,7 +250,9 @@ export async function getJobsInDateRange(jsonBody: any) {
 			const hdbSql = require('../../sqlTranslator/index');
 			pSqlEvaluate = promisify(hdbSql.evaluateSQL);
 		}
-		return await pSqlEvaluate(sqlSearchObj);
+		// Harper's own statement, not the caller's. A carrier is not equivalent here — see DESIGN.md.
+		// Wraps only this statement: a later caller-dependent one must not inherit the bypass.
+		return await runWithOperationAuthorizationBypass(true, () => pSqlEvaluate(sqlSearchObj));
 	} catch (e) {
 		log.error(
 			`there was a problem searching for jobs from date ${jsonBody.from_date} to date ${jsonBody.to_date} ${e}`
--- a/server/serverHelpers/operationAuthorizationState.ts
+++ b/server/serverHelpers/operationAuthorizationState.ts
@@ -1,11 +1,42 @@
 import { AsyncLocalStorage } from 'node:async_hooks';
 
-const operationAuthorizationState = new AsyncLocalStorage<boolean>();
+interface OperationAuthorizationState {
+	bypassAuth: boolean;
+	apiOperation?: string;
+}
+
+const operationAuthorizationState = new AsyncLocalStorage<OperationAuthorizationState>();
+
+// Shared, for the common case of no carrier to preserve.
+const BYPASSED: OperationAuthorizationState = Object.freeze({ bypassAuth: true });
+const ENFORCED: OperationAuthorizationState = Object.freeze({ bypassAuth: false });
 
 export function runWithOperationAuthorizationBypass<T>(bypassAuth: boolean, callback: () => T): T {
-	return operationAuthorizationState.run(bypassAuth === true, callback);
+	// An existing carrier survives: the enforced branch is not a bypass, and a job handler dispatching
+	// a nested authorized operation must not lose its own operation identity.
+	const apiOperation = operationAuthorizationState.getStore()?.apiOperation;
+	if (apiOperation === undefined) {
+		return operationAuthorizationState.run(bypassAuth === true ? BYPASSED : ENFORCED, callback);
+	}
+	return operationAuthorizationState.run(Object.freeze({ bypassAuth: bypassAuth === true, apiOperation }), callback);
+}
+
+/**
+ * `apiOperation` MUST come from the same value that selected the code now running, so it cannot name
+ * an operation other than the one executing. That is the entire basis for trusting it: a request
+ * property would be forgeable, and on the direct-SQL path this check is the only gate.
+ */
+export function runWithDispatchedOperation<T>(apiOperation: string, callback: () => T): T {
+	return operationAuthorizationState.run(
+		Object.freeze({ bypassAuth: operationAuthorizationState.getStore()?.bypassAuth === true, apiOperation }),
+		callback
+	);
+}
+
+export function getOperationAuthorizationState(): OperationAuthorizationState | undefined {
+	return operationAuthorizationState.getStore();
 }
 
 export function isOperationAuthorizationBypassed(): boolean {
-	return operationAuthorizationState.getStore() === true;
+	return operationAuthorizationState.getStore()?.bypassAuth === true;
 }
--- a/server/serverHelpers/serverUtilities.ts
+++ b/server/serverHelpers/serverUtilities.ts
@@ -51,6 +51,7 @@ import {
 	setLocalOperationDispatch,
 } from './registeredOperations.ts';
 import { runWithOperationAuthorizationBypass } from './operationAuthorizationState.ts';
+import { stripSuppliedParsedSqlObject } from './requestSanitization.ts';
 
 const pSearchSearch = util.promisify(search.search);
 let pEvaluateSql: (sql: string) => Promise<any>;
@@ -247,14 +248,10 @@ export function chooseOperation(json: OperationRequestBody, bypassAuth = false) 
 		if (json.operation === 'sql' || (json.search_operation && json.search_operation.operation === 'sql')) {
 			const sql = require('../../sqlTranslator/index');
 			const sqlStatement = json.operation === 'sql' ? json.sql : json.search_operation.sql;
+			// Before this dispatch's own parse is assigned, so a body-supplied object cannot survive it.
+			stripSuppliedParsedSqlObject(json);
 			const parsedSqlObject = sql.convertSQLToAST(sqlStatement);
 			json.parsed_sql_object = parsedSqlObject;
-			// NOTE: a job's SQL is re-parsed from its nested search_operation when the job runs, so the
-			// check there sees `sql` rather than the job's own operation. Carrying the real one on the
-			// request was tried and reverted: on the direct-SQL path the request is the client's body,
-			// so any property consulted by that check is forgeable, and it is the only gate on that
-			// path. This changes no outcome today — the branch that would act on the denial is dead
-			// (#2202) — but #2202 needs an unforgeable carrier before making it live.
 			if (!bypassAuth) {
 				// The SQL path never reaches verifyPerms, so the role `operations` allowlist must be
 				// enforced here — otherwise an allowlisted role reaches unlisted operations via `sql`.
--- a/sqlTranslator/index.ts
+++ b/sqlTranslator/index.ts
@@ -17,7 +17,7 @@ import * as terms from '../utility/hdbTerms.ts';
 import { handleHDBError } from '../utility/errors/hdbError.ts';
 import { HTTP_STATUS_CODES } from '../utility/errors/commonErrors.ts';
 import * as sqlEngineRouter from '../sqlEngine/router.ts';
-import { isOperationAuthorizationBypassed } from '../server/serverHelpers/operationAuthorizationState.ts';
+import { getOperationAuthorizationState } from '../server/serverHelpers/operationAuthorizationState.ts';
 
 //here we call to define and import custom functions to alasql
 alasqlFunctionImporter(alasql);
@@ -90,20 +90,20 @@ export function checkASTPermissions(jsonMessage: any, parsedSqlObject: any, apiO
 			// a caller to name whichever operation their token scope happens to allow and run arbitrary
 			// SQL under it.
 			//
-			// That rules out carrying the job's real operation on the request too. A job re-parses from
-			// its nested search_operation, so this sees `sql` rather than `export_local` there — which
-			// today changes nothing, because the branch in processAST that would act on the denial is
-			// dead (see #2202). When #2202 makes it live, the job's operation needs a carrier that a
-			// client cannot forge; a request property is not one, however carefully it is stripped.
+			// A job's re-parse would otherwise land here with the nested search_operation, whose
+			// `operation` is the inner `sql`; the job worker supplies the real one out of async context
+			// instead (operationAuthorizationState), which a request cannot set.
 			apiOperation ?? jsonMessage.operation
 		);
-		parsedSqlObject.permissions_checked = true;
 	} catch (e) {
 		throw e;
 	}
 	if (verifyResult) {
 		return verifyResult;
 	}
+	// Only after a pass: this flag is what processAST trusts to skip the gate, so a denied AST
+	// carrying it would read as already authorized.
+	parsedSqlObject.permissions_checked = true;
 	return null;
 }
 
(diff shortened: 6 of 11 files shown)


dawsontoth:
Moving this to draft. It is a real pre-existing defect and the fix stands, but it touches shared SQL authorization for every caller, not just the OIDC feature, so it needs the full [harper-engineering-guidelines](https://github.com/HarperFast/harper) treatment on its own — cross-model reviews with the coverage reported in the description — rather than inheriting #2173's. That will be picked up separately.

Nothing in [#2173](https://github.com/HarperFast/harper/pull/2173) depends on this landing: its own gate refuses the export/write combination at the front door, and it no longer attempts to carry a job's operation into the branch this fixes.

One thing for whoever picks this up: making the `processAST` branch live means a job's SQL will be authorized as `sql` rather than as its own operation (`export_local`), so a token scoped only to the job operation would be denied by its own job. Carrying the real operation on the request body was tried in #2173 and reverted — on the direct-SQL path that body is client-supplied and this check is the only gate, so any property it consults is forgeable. That carrier needs to be something a client cannot set.

dawsontoth:
Ordering note for whoever picks this up — [#2173](https://github.com/HarperFast/harper/pull/2173) now carries a test that will go **red** when this lands, deliberately.

Making the `processAST` branch live means a job's SQL is authorized as `sql` rather than as its own operation, because `checkASTPermissions` falls back to `jsonMessage.operation` and at that call site `jsonMessage` is the nested `search_operation`. So a token scoped to `export_local` would be denied by its own export job. Verified by applying this PR's diff on top of #2173 and rebuilding.

The failing test is `admits an in-scope export job through the path export.ts actually dispatches` in `unitTests/security/tokenOperationScope.test.js`. **The fix is not to relax the scope check or delete the test** — it is to give the job's real operation a carrier a client cannot set. A request-body property is not one: that was tried in #2173 and reverted, because on the direct-SQL path the body is client-supplied and that check is the only gate, so any property it consults is forgeable.

The two PRs can merge in either order; whichever is second goes red rather than silently shipping an export-scoped token that 403s on its own export.

dawsontoth:
Picked this up, and the branch has been force-pushed — the one-line guard fix plus a test file is now
the guard fix plus the carrier the [ordering note](https://github.com/HarperFast/harper/pull/2202#issuecomment-5332131086)
above asked for. Answering that note directly, since it set the constraint:

**The tripwire fired exactly as predicted.** `admits an in-scope export job through the path export.ts
actually dispatches` went red on the one-line fix alone. It now asserts the fixed behaviour, with two
new cases pinning the other direction (a `sql`-only scope cannot start an export; absent a carrier the
gate fails closed).

**The carrier is `request.operation`, which is body-derived — the note ruled that out, so here is why
this instance is different.** In the job worker, `request.operation` is the same field
`serverUtils.getOperationFunction(request)` reads to select the handler that then runs. Forging it
changes *which operation executes*, not *which policy is applied to a different one*; the two cannot
diverge. That is what makes it safe where a *new* field (`api_operation`, tried and reverted in #2173)
is not. It is carried on async context (`runWithDispatchedOperation`), never read back off the request
by the check itself.

**One thing the note did not anticipate, and it mattered more than the carrier.** Making the branch
live also made `permissions_checked` load-bearing — and it was forgeable. `evaluateSQL` trusts a
supplied `parsed_sql_object` verbatim and skips parsing; `chooseOperation` overwrote only the
*top-level* one; `dataLayer/export.ts` hands the nested `search_operation` straight to `evaluateSQL`.
So a request carrying `search_operation.parsed_sql_object` with `permissions_checked: true` would have
executed an arbitrary AST with the newly-live check skipped. In other words, without this the PR would
have advertised a gate that a request body could switch back off. Now deleted at dispatch *and* again
in the job worker when it loads the persisted row, because the worker re-enters from `hdb_job` rather
than re-dispatching. Same defect as #2217's item 3 — whichever lands second should drop its copy.

**Two things deliberately left open**, both in the description: `evaluateSQL` drops the denial object
on error, so a denied job records the bare string `403` with no reason; and the nested-SQL dispatch
branch still does not enforce `export_local`'s `requires_su`, which #2217 approaches from the other
side.

**And one honest gap.** Nothing proves the worker actually establishes the carrier — delete that call
and CI stays green. A token operation scope originates only from an OIDC trust-policy exchange, and
there is no integration harness for it; table permissions cannot substitute (`requires_su` plus the
super_user early return), and an inline-role scoped token is minted `super_user: false` so it cannot
invoke a `requires_su` export at all. I wrote two integration tests, proved neither premise
reachable, and deleted them rather than leave tests that pass for the wrong reason. Filed as #2298.

The earlier bot reviews on this thread predate the rewrite and no longer describe the diff.

dawsontoth:
Replying to @claude's inline suggestion on `server/jobs/jobs.ts:255` here rather than in the thread — I have a pending review open, and GitHub's reply endpoint refuses a second one.

Agreed on the gap — independently confirmed, and it is declared in the description. Worth recording why neither suggested route closes it, since both look like they should:

**A spy on `runWithOperationAuthorizationBypass`** would need `sinon`, which `AGENTS.md:196` prohibits for new tests ("do not add new uses of `sinon` or `rewire`… If you can't write a test without stubbing, comment on the issue describing what's missing and stop"). This PR already removed the `sinon` I had added to `tokenOperationScope.test.js` for that reason.

**An unstubbed case with a restricted-scope `hdb_user`** never reaches the query. `search_jobs_by_start_date` is `requires_su`, so a restricted principal is refused by `verifyPerms` in `chooseOperation` before any SQL runs — verified directly while attempting exactly this test: `Operation 'handleGetJobsByStartDate' is restricted to 'super_user' roles`. The principal that *would* exercise the bypass is a super_user **without** an appended `permission.system` (an impersonation payload, or a path skipping user-cache hydration), and the operations API cannot construct one.

Also flagging a related claim from another review round, in case it resurfaces: `northwind.test.mjs`'s `Search Jobs by date` does **not** pin this wrapper. It runs as a hydrated super_user, for whom `appendSystemTablesToRole` grants `system.*.read`, so `hasPermissions` passes with or without the bypass; the adjacent non-super-user case asserts a `verifyPerms` denial that never reaches SQL.

The tractable option, if the author wants coverage here, is a sinon-free test pinning the *fact the bypass depends on* rather than the wrapper: assert that `checkASTPermissions` denies `SELECT * FROM system.hdb_job` for a super_user role carrying no `permission.system`, since `verifyPermsAST`'s early return is `isSuperUser && !isSuSystemOperation` and a `system` schema is exempt from it. That guards the invariant a future refactor would break. Not adding it in this pass — the PR is parked in draft for a human ruling on a separate question, and I would rather not extend the diff underneath that.

Tracked with the other declared coverage gap in #2298.

dawsontoth:
Replying to @claude's inline suggestion on `server/jobs/jobProcess.ts:63` here — my pending review holds this PR's one-pending-review slot, so the thread reply endpoint refuses.

Accepted, and done: **#2298 now covers this line too.** The diagnosis is exactly right — `jobProcess.ts` is a top-level worker IIFE keyed off `process.env`, so both of its new branches (the load-time sanitize and the `runWithDispatchedOperation` carrier immediately below) are unreachable in isolation for the same reason, and whatever makes that file testable closes both at once rather than one.

Two clarifications recorded on the issue:

- The **dispatch-time twin is covered** — `serverUtilities.test.js`'s `discards a body-supplied parsed_sql_object on the nested search_operation`, confirmed to fail on base. So the primary path has a regression test; the worker-side repeat is defense-in-depth for a persisted or directly-written row, which is the part that is untested.
- The two positions differ deliberately: `chooseOperation` deletes only the nested object because it *overwrites* the top-level one with its own parse — but only inside its SQL branch, so a non-SQL job (e.g. `export_local` with a `search_by_value` search) can still carry a client-supplied top-level object through to persistence. That asymmetry is why the worker deletes both.

Not adding a test in this pass: the PR is parked in draft for a human ruling on a separate question, and extending the diff underneath that would also invalidate the review receipt the description points at.

server/jobs/jobs.ts
```diff
@@ -249,7 +250,9 @@ export async function getJobsInDateRange(jsonBody: any) {
 			const hdbSql = require('../../sqlTranslator/index');
 			pSqlEvaluate = promisify(hdbSql.evaluateSQL);
 		}
-		return await pSqlEvaluate(sqlSearchObj);
+		// Harper's own statement, not the caller's. A carrier is not equivalent here — see DESIGN.md.
+		// Wraps only this statement: a later caller-dependent one must not inherit the bypass.
+		return await runWithOperationAuthorizationBypass(true, () => pSqlEvaluate(sqlSearchObj));
```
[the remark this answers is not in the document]
dawsontoth: Not changed, by design. A spy needs `sinon`, which `AGENTS.md:196` prohibits for new tests; and a restricted-scope `hdb_user` never reaches the query, because `search_jobs_by_start_date` is `requires_su` and `verifyPerms` refuses it before any SQL runs (verified: `Operation 'handleGetJobsByStartDate' is restricted to 'super_user' roles`). The principal that would actually exercise the bypass is a super_user *without* an appended `permission.system`, which the operations API cannot construct. Tracked in #2298; full rationale in [this comment](https://github.com/HarperFast/harper/pull/2202#issuecomment-5401310648).

server/jobs/jobProcess.ts
```diff
@@ -54,12 +55,19 @@ const JOB_ID = JOB_NAME.substring(4);
 			throw new Error('Did not find job request in hdb_job table, unable to proceed');
 		}
 		request = cloneDeep(request);
+		// Sanitize on load: the worker re-enters from the persisted row rather than re-dispatching, and
+		// evaluateSQL honors a supplied parsed_sql_object verbatim, so an AST carrying
+		// `permissions_checked: true` would run unchecked. Both positions, because the dispatch
+		// persists a top-level one and a directly-written row can carry either.
+		delete request.parsed_sql_object;
+		delete request.search_operation?.parsed_sql_object;
```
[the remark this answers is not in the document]
dawsontoth: Accepted and done — **#2298 now covers this line**, per your suggestion. Both of this function's new branches (this sanitize and the `runWithDispatchedOperation` carrier just below) are unreachable for the same reason, so whatever makes `jobProcess.ts` testable closes two branches, not one. Two clarifications recorded there: the dispatch-time twin *is* covered and confirmed to fail on base; and the two delete positions differ deliberately, because `chooseOperation` overwrites the top-level object only inside its SQL branch, so a non-SQL job (e.g. `export_local` with a `search_by_value` search) can still carry a client-supplied one through to persistence.

server/serverHelpers/serverUtilities.ts
```diff
@@ -249,12 +249,13 @@ export function chooseOperation(json: OperationRequestBody, bypassAuth = false)
...
-			// so any property consulted by that check is forgeable, and it is the only gate on that
-			// path. This changes no outcome today — the branch that would act on the denial is dead
-			// (#2202) — but #2202 needs an unforgeable carrier before making it live.
+			// The nested one is never overwritten, only deleted: export.ts hands `search_operation`
+			// straight to evaluateSQL, which trusts a supplied parsed_sql_object verbatim — so a
+			// body-supplied one carrying `permissions_checked: true` would run an arbitrary AST that no
+			// check ever saw. Deleting it forces the worker to re-parse from the `sql` string this
+			// dispatch just authorized. The job's own operation reaches that re-parse through async
+			// context (operationAuthorizationState), which a request cannot set.
+			delete json.search_operation?.parsed_sql_object;
```
cb1kenobi: **High: `export_local` / `export_to_s3` skip their `requires_su` gate whenever the nested `search_operation` is SQL**

This branch is mutually exclusive with the `verifyPerms` call in the `else` below, and `requires_su` is only enforced inside `verifyPerms` (`utility/operation_authorization.ts:869`). `requiredPermissions` marks both export ops `requires_su: true`, and `dataLayer/export.ts` has no SU check of its own — so nesting a `sql` search_operation is enough for an ordinary role to reach an SU-only operation.

Confirmed at runtime on this head (worktree checkout, `chooseOperation` called directly, user `{super_user: false}`):

- `{operation:'export_local', path:'/tmp', format:'json', search_operation:{operation:'sql', sql:'SELECT 1'}}` → **allowed**, returns `executeJob`.
- Same request with `search_operation.operation: 'search_by_conditions'` → denied: `"Operation 'export_local' is restricted to 'super_user' roles"`.

`export_local` writes to a caller-supplied absolute `path`, so this is an arbitrary file write as the Harper process user for any authenticated role; `export_to_s3` is the exfiltration equivalent. Table permissions still bound *what* can be selected, but not *that the operation runs at all*.

Pre-existing and not introduced here — it is item 7 of the description, and #2217 approaches it from the other side. Flagged because it is the same bug class this PR fixes (a gate enforced on one entry point and not another) and this branch is now under a reviewer's eyes. Not a blocker for this PR; it should not stay un-ticketed either.

—
_Generated by Barber AI_

server/jobs/jobProcess.ts
```diff
@@ -54,12 +55,19 @@ const JOB_ID = JOB_NAME.substring(4);
 			throw new Error('Did not find job request in hdb_job table, unable to proceed');
 		}
 		request = cloneDeep(request);
+		// Sanitize on load: the worker re-enters from the persisted row rather than re-dispatching, and
+		// evaluateSQL honors a supplied parsed_sql_object verbatim, so an AST carrying
+		// `permissions_checked: true` would run unchecked. Both positions, because the dispatch
+		// persists a top-level one and a directly-written row can carry either.
+		delete request.parsed_sql_object;
```
cb1kenobi: **Medium: neither security control this file adds is covered — mutation-confirmed**

Both survive mutation with the suite green:

- Deleting lines 62-63 (the persisted-row `parsed_sql_object` strip): `unitTests/server/**`, `unitTests/security/**`, `unitTests/sqlTranslator/**`, `unitTests/dataLayer/export.test.js`, `unitTests/utility/operation_authorization.test.js` → **1622 passing / 12 failing**, identical to the unmutated head. (Those 12 also fail on merge base `0d359723`, so they are environmental.)
- Replacing line 68's `runWithDispatchedOperation(...)` with a direct call → narrow suite **208 passing, 0 failing**.

DESIGN.md acknowledges the second. The first is a different case: it is the *second layer* of the forgeable-`permissions_checked` fix — the description's item 3 states the dispatch-time strip alone is not sufficient because the worker re-enters from the persisted row — and unlike the carrier it needs no ALS or job-runner harness to test.

Suggested fix: move the two deletes into `export function sanitizeJobRequest(request)` in this file and call it here. That makes the invariant nameable, directly unit-testable, and hard for a later refactor to drop silently.

—
_Generated by Barber AI_

sqlTranslator/index.ts
```diff
@@ -90,11 +90,9 @@ export function checkASTPermissions(jsonMessage: any, parsedSqlObject: any, apiO
...
 			// SQL under it.
 			//
-			// That rules out carrying the job's real operation on the request too. A job re-parses from
-			// its nested search_operation, so this sees `sql` rather than `export_local` there — which
-			// today changes nothing, because the branch in processAST that would act on the denial is
-			// dead (see #2202). When #2202 makes it live, the job's operation needs a carrier that a
-			// client cannot forge; a request property is not one, however carefully it is stripped.
+			// A job's re-parse would otherwise land here with the nested search_operation, whose
+			// `operation` is the inner `sql`; the job worker supplies the real one out of async context
+			// instead (operationAuthorizationState), which a request cannot set.
```
cb1kenobi, line 95: **Low: `permissions_checked` is set to `true` on line 98 even when the check produced a denial**

`checkASTPermissions` marks the parsed object as checked before the caller inspects `verifyResult`, so a *denied* AST carries the same "already authorized" flag as an approved one — and that flag is precisely what `processAST` (line 151) trusts to skip the gate.

Not a live bypass: nothing re-enters a denied object today — `chooseOperation` assigns `json.parsed_sql_object` (serverUtilities.ts:251) and then throws, and `processAST`'s reference is local. **This finding is static-only**; I traced the callers and found no reachable path, so it is a latent invariant violation rather than an exploit.

Worth tightening anyway, since this PR's whole second half exists to keep this flag from being forged — "checked" should mean "checked and passed":

```js
	if (verifyResult) {
		return verifyResult;
	}
	parsedSqlObject.permissions_checked = true;
	return null;
```

—
_Generated by Barber AI_

DESIGN.md
```diff
@@ -391,6 +391,75 @@ early-returns (super_user, structure_user, system-table allowances): persisted r
...
+`processAST` reads the state once. `runWithOperationAuthorizationBypass` **preserves** an existing
+carrier on both branches. That is deliberate and was initially got wrong: its enforced branch is not
+a bypass, so a job handler dispatching a nested _authorized_ operation lands there, and dropping the
+carrier would judge that job's re-parsed SQL as the inner `sql` and refuse it partway through its own
+work. The consequence to know is the other direction — a nested dispatch inside a job is judged
+against the **outer** job's operation for any `evaluateSQL` that does not pass through
+`chooseOperation`. It allocates only when a carrier is present; with none, two shared frozen objects
+serve the common path. Note it is reached only from `server.operation()`
+(`server/serverHelpers/serverUtilities.ts`) and the ITC path (`registeredOperations.ts`), not from
+the ops-API dispatch.
```
cb1kenobi: **Low: this sentence lists two call sites; the function has four, and one of them this PR adds**

`runWithOperationAuthorizationBypass` is also called from `server/jobs/jobs.ts:255` (added in this PR) and `sqlEngine/diff/differential.ts:42`. The `jobs.ts` one *is* reached from the ops-API dispatch: `search_jobs_by_start_date` → `handleGetJobsByStartDate` → `getJobsInDateRange`.

Since this paragraph is what a future reader will use to reason about where the preserved carrier can travel, an enumeration that omits half the call sites is the kind of drift rounds 4-6 were spent removing. If the intended claim was narrower (e.g. the *carrier-preserving* branch), say that explicitly.

—
_Generated by Barber AI_

server/serverHelpers/operationAuthorizationState.ts
```diff
@@ -1,11 +1,42 @@
...
 
 export function runWithOperationAuthorizationBypass<T>(bypassAuth: boolean, callback: () => T): T {
-	return operationAuthorizationState.run(bypassAuth === true, callback);
+	// An existing carrier survives: the enforced branch is not a bypass, and a job handler dispatching
+	// a nested authorized operation must not lose its own operation identity.
+	const apiOperation = operationAuthorizationState.getStore()?.apiOperation;
+	if (apiOperation === undefined) {
+		return operationAuthorizationState.run(bypassAuth === true ? BYPASSED : ENFORCED, callback);
+	}
+	return operationAuthorizationState.run({ bypassAuth: bypassAuth === true, apiOperation }, callback);
```
cb1kenobi: **Nit: the carrier-carrying store objects are mutable while `BYPASSED`/`ENFORCED` are frozen**

Lines 11-12 freeze the two shared objects; the one built here and the one on line 31 are not, and `getOperationAuthorizationState()` hands the live store out to any caller. Freezing both keeps the module's invariant uniform for the cost of one call.

—
_Generated by Barber AI_

server/serverHelpers/serverUtilities.ts
dawsontoth: Confirmed independently and ticketed as #2305 (Bug, P1). Verified the mechanism rather than taking it on report: `requiredPermissions` registers both export ops via `new permission(true, [], …)` (`operation_authorization.ts:416-422`) — first arg is `requiresSu` — and `requires_su` is enforced only at `operation_authorization.ts:869` inside `verifyPerms`, which this branch is mutually exclusive with. Not fixed here, per your read: closing it changes authorization outcomes for existing roles and wants its own release note. Flagged on the issue that the exploit shape is now public and someone should decide whether it belongs in a private advisory instead.

sqlTranslator/index.ts
dawsontoth: Fixed — `permissions_checked = true` now runs only after `verifyResult` is checked, so "checked" means "checked and passed". Agreed it is static-only today; it is also the exact flag the second half of this PR exists to keep unforgeable, so leaving it set on a denial was the wrong shape regardless.

DESIGN.md
dawsontoth: Fixed — the paragraph now enumerates all four (`serverUtilities.ts` `server.operation()`, `registeredOperations.ts` ITC, `sqlEngine/diff/differential.ts`, `server/jobs/jobs.ts`) and calls out explicitly that the `jobs.ts` one **is** reached from the ops-API dispatch via `search_jobs_by_start_date` → `handleGetJobsByStartDate` → `getJobsInDateRange`. You are right about the drift — that is the seventh doc-vs-code correction on this branch.

server/serverHelpers/operationAuthorizationState.ts
dawsontoth: Fixed — both carrier-carrying stores are now `Object.freeze`d, so all four are uniform and `getOperationAuthorizationState()` cannot hand out a mutable one.

server/jobs/jobProcess.ts
dawsontoth: Fixed, and your mutation check now fails. Adopted the idea but not the location: `jobProcess.ts` reads `process.env[PROCESS_NAME_ENV_PROP]` at module scope (lines 22-23), so importing it to reach an exported `sanitizeJobRequest` throws — the export would have been as untestable as the inline deletes.

Instead the invariant moved to `server/serverHelpers/requestSanitization.ts` as `stripSuppliedParsedSqlObject(request)`, a leaf module with no import side effects, and **both** sites now call it — `chooseOperation` as well as `jobProcess`. That single-sources it rather than keeping two hand-maintained copies, which was the other half of the risk. `chooseOperation` now strips *before* assigning its own parse, so the body-supplied object cannot survive either way.

Covered by `unitTests/server/serverHelpers/requestSanitization.test.js` (5 cases: each position, both at once, the authorized `sql` string surviving, and the empty/missing shapes). Re-ran your mutation: deleting the two `delete` lines takes that file plus `serverUtilities.test.js` from **89 passing / 0 failing** to **85 passing / 4 failing**.

The carrier on the line below remains uncovered for the reason you identified, tracked in #2298.

cb1kenobi:
Re-reviewed `13ee095f` (was `d817c124`) — incremental range only: 7 files, +106/-19, derived from `git range-diff` against merge bases computed separately for each head (both resolve to `0d359723`, so this is a force-push amend of the single commit, not a rebase).

**Prior findings**

- **Low — `permissions_checked` set on a denial** (`sqlTranslator/index.ts`): fixed. The assignment now runs only after the `verifyResult` early return, so a denied AST can no longer read as already authorized.
- **Low — DESIGN.md enumerated two call sites** : fixed, and the new enumeration checks out. `runWithOperationAuthorizationBypass` has exactly four production call sites (`serverUtilities.ts:361`, `registeredOperations.ts:215`, `jobs.ts:255`, `sqlEngine/diff/differential.ts:42`), and the `search_jobs_by_start_date` → `handleGetJobsByStartDate` → `getJobsInDateRange` chain to `jobs.ts:255` is real.
- **Nit — mutable carrier stores**: fixed. Both dynamic stores are now `Object.freeze`d; nothing anywhere writes to the store, so the freeze is safe as well as correct.
- **Medium — `jobProcess` controls uncovered**: the suggested extraction landed. `stripSuppliedParsedSqlObject` now has a name, a module, and five tests, and that coverage is real — removing the nested `delete` turns 3 tests red. The `jobProcess` *call sites* remain uncovered: at this head, dropping `stripSuppliedParsedSqlObject(request)` and dropping `runWithDispatchedOperation` each still survive the full unit suite. Tracked in #2298, which is the right disposition given the worker-IIFE constraint.
- **High — `export_local`/`export_to_s3` skip `requires_su` for a nested SQL `search_operation`**: not addressed here, and this diff touches neither `dataLayer/export.ts` nor `utility/operation_authorization.ts` nor the branch structure, so it stands unchanged. Now tracked in #2305 — pre-existing and out of scope for this PR.

**New work**: no issues found.

Verification at this head: full unit suite 4801–4804 passing against 4788 at the merge base, with the head failure set a superset of the base's only by four spawn/keys tests that also differ between two consecutive clean runs (environmental flake, unrelated to this diff). All four regression mutations from the prior pass are still killed: restoring the dead `.length > 0` guard (6 red), dropping the `chooseOperation` strip (1 red), dropping `apiOperation` from the carrier (4 red), dropping carrier preservation across the bypass wrapper (1 red). Prettier clean.

This PR looks good, nice job!

—
_Generated by Barber AI_

kriszyp (approved):
Great work!

PR #2208, opened 2026-08-18
Deflake deferred cache and await LMDB reload markers

kriszyp:
The reload-marker assertion was not flaky. Under LMDB, `writeReloadMarker()` could resolve before its asynchronous audit-store write completed, so the immediate audit scan correctly exposed a transaction completion-ordering bug. The cache assertion was a separate eventual-propagation race.

This keeps the cache test's bounded poll, restores the reload marker's immediate-visibility contract, and applies the asynchronous commit-callback contract consistently across both storage engines.

Refs #2178

## What changed

- `cache-test.mjs` polls until the deferred source update is visible. It retries empty and non-success responses but still fails fast on server errors.
- Promise-returning transaction commit callbacks are awaited by LMDB's optimistic and exclusive paths and by RocksDB before the native transaction commits.
- Publish and reload-marker callbacks return their `updateRecord()` completion so transaction resolution includes audit persistence.
- Reload markers explicitly opt out of replication confirmation. Stored null-key root-topic publishes still receive confirmation, including when a prior null-key record exists.
- Replication confirmation checks the stored entry in both engines rather than dereferencing a missing entry.
- The disconnected LMDB `outstandingCommit` variable was removed; it had no effect on overload tracking or transaction behavior.
- Reload-marker tests cover LMDB's optimistic and contention paths, plus RocksDB parity. A separate RocksDB test verifies the promise-returning callback contract.

## Why this layer

The invariant belongs to the transaction wrapper: when a transaction callback starts asynchronous persistence, the returned commit promise must not resolve until that persistence settles. Polling in the test would hide the broken contract, and awaiting only in `writeReloadMarker()` would leave transaction chaining and error propagation inconsistent.

Null is not enough to identify a reload marker because Harper also stores root-topic publishes at the null key. The write therefore carries explicit metadata when it stages only a local audit marker.

The default path remains synchronous. Promise aggregation occurs only when a commit callback returns a promise.

## Verification

- Focused RocksDB reload-marker tests: 4 passing.
- Focused RocksDB promise-returning callback test: 1 passing.
- Focused LMDB reload-marker tests: 4 passing, covering optimistic and forced-contention paths.
- Negative controls at the merge base failed behaviorally: RocksDB incorrectly confirmed the local-only marker and did not observe callback completion; LMDB missed immediate marker persistence and root-topic confirmation.
- `HARPER_STORAGE_ENGINE=lmdb npm run test:unit:resources`: changed tests passed within 1,353 passing and 93 pending; only the two existing Node.js 26 `randomAccessFields` assertion failures remained.
- Default `npm run test:unit:resources`: changed tests passed within 1,555 passing and 15 pending; the three known Node.js 26 failures remained in `databases.test.js`, `randomAccessFieldsDirective.test.js`, and `replayStructures.test.js`.
- `npm run test:integration:all`: changed areas passed; the run ended in the unrelated Ollama backend group because Node rejected `json/systemSchema.json` without a JSON import attribute, cancelling the remaining Ollama cases.
- `npm run build`, `npm run lint:required`, `npm run format:write`, and `git diff --check`: passed.

The full local `test:unit:main` gate remains unavailable because this checkout's unit bootstrap opens an ambient install database before test setup; it fails during module load in that external database and does not reach the changed tests.

## Review coverage

The required full no-Claude review ran Gemini. Claude and its domain adjudication leg were pruned by policy, while both Cursor legs were pruned because this branch had reached its two-round cap. Gemini returned eight ungraded finding bullets; source tracing and the executed engine tests did not substantiate them, so no additional code changes were made. In particular, LMDB builds the conditional batch synchronously and preserves the eventual false conflict result, both transaction modules export the tested confirmation seam, and both engines accept the stored null-key publish exercised by the tests.

_— GPT-5 Codex_

<sub>Review-Coverage: authored=unknown; ran=none; rounds=1 @ d92bacc59e56</sub>

<sub>Human-Review-Need: 4 @ d92bacc59e56</sub>

### DIFF
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -575,6 +575,31 @@ export class DatabaseTransaction implements Transaction {
 		this.overloadChecked = true; // only check this once, don't interrupt ongoing transactions that have already made writes
 	}
 
+	/**
+	 * Stage an async operation that commit() must wait for. Staging can happen a turn or more before
+	 * commit() attaches its Promise.all, so the no-op rejection handler is attached here: without it a
+	 * rejection in that window is an unhandled rejection (fatal under --unhandled-rejections=strict).
+	 * The rejection still surfaces through commit()'s Promise.all.
+	 */
+	/**
+	 * The stored entry of the last write eligible for replication confirmation. Writes that opt out
+	 * (audit-only markers, which stage no record) are skipped rather than ending the search, so a
+	 * trailing marker cannot suppress confirmation for replicable writes staged earlier.
+	 */
+	lastConfirmableEntry(): Partial<Entry> | undefined {
+		for (let i = this.writes.length - 1; i >= 0; i--) {
+			const write = this.writes[i];
+			if (!write || write.skipReplicationConfirmation) continue;
+			const entry = write.store.getEntry(write.key);
+			if (entry) return entry;
+		}
+	}
+
+	stageCompletion(completion: Promise<void>) {
+		completion.then(undefined, () => {});
+		this.completions.push(completion);
+	}
+
 	addWrite(operation: TransactionWrite) {
 		if (this.timedOut) throw transactionOpenTooLongError();
 		// A write is activity: it re-arms the idle limit on this link even though the reads it
@@ -641,12 +666,12 @@ export class DatabaseTransaction implements Transaction {
 				return;
 			}
 			let result: Promise<void> = operation.before?.() as Promise<void>;
-			if (result?.then) this.completions.push(result);
+			if (result?.then) this.stageCompletion(result);
 			result = operation.beforeIntermediate?.() as Promise<void>;
-			if (result?.then) this.completions.push(result);
+			if (result?.then) this.stageCompletion(result);
 		}
 		const completion = operation.commit(txnTime, operation.entry, this.retries > 0, transaction) as Promise<void>;
-		if (typeof completion?.then === 'function') this.completions.push(completion);
+		if (typeof completion?.then === 'function') this.stageCompletion(completion);
 		// Sticky record that THIS write staged with its audit entry appended (log entries batch on the
 		// native transaction and are durably written by its commit attempt — even a failed one — so
 		// they survive the abort-after-failed-commit of the retry paths). isRetry stagings
@@ -859,9 +884,9 @@ export class DatabaseTransaction implements Transaction {
 								// if we want to wait for replication confirmation, we need to track the transaction times
 								// and when replication notifications come in, we count the number of confirms until we reach the desired number
 								const databaseName = this.writes[0].store.rootStore.databaseName;
-								const lastWrite = this.writes[this.writes.length - 1];
-								const lastEntry =
-									lastWrite && !lastWrite.skipReplicationConfirmation && lastWrite.store.getEntry(lastWrite.key);
+								// the last write that can be confirmed: a trailing confirmation-exempt write (an
+								// audit-only marker) must not suppress confirmation for replicable writes staged before it
+								const lastEntry = this.lastConfirmableEntry();
 								if (confirmReplication && lastEntry) {
 									completions.push(
 										confirmReplication(databaseName, (lastEntry as any).version, this.replicatedConfirmation)
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -253,7 +253,7 @@ export type TransactionWrite = {
 	entry?: Partial<Entry>;
 	before?: () => void | Promise<void>;
 	beforeIntermediate?: () => void | Promise<void>;
-	commit?: (txnTime: number, existingEntry: Partial<Entry>, retry: boolean, transaction: any) => void;
+	commit?: (txnTime: number, existingEntry: Partial<Entry>, retry: boolean, transaction: any) => MaybePromise<void>;
 	validate?: (txnTime: number) => void;
 	fullUpdate?: boolean;
 	saved?: boolean;
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -258,6 +258,7 @@ export type TransactionWrite = {
 	fullUpdate?: boolean;
 	saved?: boolean;
 	deferSave?: boolean;
+	skipReplicationConfirmation?: boolean;
 	nodeName?: string;
 	nodeId?: number;
 	promise?: Promise<any>;
@@ -307,8 +308,8 @@ export function priorStagedWrite(operation: TransactionWrite): TransactionWrite 
  * `[-0]` and `[null]` vs `[NaN]` are DIFFERENT stored keys that value-ish encodings (JSON, string
  * coercion) collapse, cross-contaminating unrelated records. So every key is mapped through the
  * same encoder the stores use; latin1 keeps the bytes injective in a string. Symbol keys (internal
- * metadata writes) can't be key-encoded and keep native identity; so does null (topic-less
- * publishes), which never stages a record.
+ * metadata writes) can't be key-encoded and keep native identity. Null is reserved for topic-less
+ * publishes and audit-only markers.
  */
 export function writeKeyId(key: Id): unknown {
 	if (typeof key === 'symbol' || key == null) return key;
@@ -644,7 +645,8 @@ export class DatabaseTransaction implements Transaction {
 			result = operation.beforeIntermediate?.() as Promise<void>;
 			if (result?.then) this.completions.push(result);
 		}
-		operation.commit(txnTime, operation.entry, this.retries > 0, transaction);
+		const completion = operation.commit(txnTime, operation.entry, this.retries > 0, transaction) as Promise<void>;
+		if (typeof completion?.then === 'function') this.completions.push(completion);
 		// Sticky record that THIS write staged with its audit entry appended (log entries batch on the
 		// native transaction and are durably written by its commit attempt — even a failed one — so
 		// they survive the abort-after-failed-commit of the retry paths). isRetry stagings
@@ -858,13 +860,11 @@ export class DatabaseTransaction implements Transaction {
 								// and when replication notifications come in, we count the number of confirms until we reach the desired number
 								const databaseName = this.writes[0].store.rootStore.databaseName;
 								const lastWrite = this.writes[this.writes.length - 1];
-								if (confirmReplication && lastWrite) {
+								const lastEntry =
+									lastWrite && !lastWrite.skipReplicationConfirmation && lastWrite.store.getEntry(lastWrite.key);
+								if (confirmReplication && lastEntry) {
 									completions.push(
-										confirmReplication(
-											databaseName,
-											(lastWrite.store.getEntry(lastWrite.key) as any).version,
-											this.replicatedConfirmation
-										)
+										confirmReplication(databaseName, (lastEntry as any).version, this.replicatedConfirmation)
 									);
 								}
 							}
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -576,12 +576,6 @@ export class DatabaseTransaction implements Transaction {
 	}
 
 	/**
-	 * Stage an async operation that commit() must wait for. Staging can happen a turn or more before
-	 * commit() attaches its Promise.all, so the no-op rejection handler is attached here: without it a
-	 * rejection in that window is an unhandled rejection (fatal under --unhandled-rejections=strict).
-	 * The rejection still surfaces through commit()'s Promise.all.
-	 */
-	/**
 	 * The stored entry of the last write eligible for replication confirmation. Writes that opt out
 	 * (audit-only markers, which stage no record) are skipped rather than ending the search, so a
 	 * trailing marker cannot suppress confirmation for replicable writes staged earlier.
@@ -595,6 +589,12 @@ export class DatabaseTransaction implements Transaction {
 		}
 	}
 
+	/**
+	 * Stage an async operation that commit() must wait for. Staging can happen a turn or more before
+	 * commit() attaches its Promise.all, so the no-op rejection handler is attached here: without it a
+	 * rejection in that window is an unhandled rejection (fatal under --unhandled-rejections=strict).
+	 * The rejection still surfaces through commit()'s Promise.all.
+	 */
 	stageCompletion(completion: Promise<void>) {
 		completion.then(undefined, () => {});
 		this.completions.push(completion);
@@ -884,8 +884,6 @@ export class DatabaseTransaction implements Transaction {
 								// if we want to wait for replication confirmation, we need to track the transaction times
 								// and when replication notifications come in, we count the number of confirms until we reach the desired number
 								const databaseName = this.writes[0].store.rootStore.databaseName;
-								// the last write that can be confirmed: a trailing confirmation-exempt write (an
-								// audit-only marker) must not suppress confirmation for replicable writes staged before it
 								const lastEntry = this.lastConfirmableEntry();
 								if (confirmReplication && lastEntry) {
 									completions.push(
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -576,9 +576,12 @@ export class DatabaseTransaction implements Transaction {
 	}
 
 	/**
-	 * The stored entry of the last write eligible for replication confirmation. Writes that opt out
-	 * (audit-only markers, which stage no record) are skipped rather than ending the search, so a
-	 * trailing marker cannot suppress confirmation for replicable writes staged earlier.
+	 * The stored entry of the last write eligible for replication confirmation. Two kinds of write are
+	 * skipped (rather than ending the search) so a trailing one cannot suppress confirmation for
+	 * replicable writes staged earlier: writes that explicitly opt out (audit-only markers, which stage
+	 * no record), and writes with no stored entry at all — a delete, or a delete on an `audit: false`
+	 * table, leaves nothing to read back. So a `put(A); delete(B)` transaction confirms on A's entry,
+	 * whose version is this transaction's (every write in a transaction is stamped with one version).
 	 */
 	lastConfirmableEntry(): Partial<Entry> | undefined {
 		for (let i = this.writes.length - 1; i >= 0; i--) {
@@ -598,6 +601,22 @@ export class DatabaseTransaction implements Transaction {
 	stageCompletion(completion: Promise<void>) {
 		completion.then(undefined, () => {});
 		this.completions.push(completion);
+	}
+
+	/**
+	 * Discard staged completions that no commit() will ever aggregate (abort path). Their rejections
+	 * are already no-op-handled by stageCompletion(), so without this they fail silently; log instead.
+	 * Clearing them also keeps a reused transaction's next commit() from rejecting with the previous
+	 * batch's error.
+	 */
+	drainCompletions(): void {
+		if (this.completions.length === 0) return;
+		const completions = this.completions;
+		this.completions = [];
+		for (const completion of completions)
+			completion.then(undefined, (error) =>
+				harperLogger.warn?.('A staged transaction completion failed after the transaction was aborted', error)
+			);
 	}
 
 	addWrite(operation: TransactionWrite) {
@@ -1028,7 +1047,13 @@ export class DatabaseTransaction implements Transaction {
 	}
 	abort(): void {
 		while (this.readTxnsUsed > 0) this.doneReadTxn(); // release the read snapshot when we abort, we assume we don't need it
+		// A write-only transaction never took a read reference (getReadTxn was never called), so the loop
+		// above releases nothing even though save() created a native handle; release it here instead of
+		// leaking the handle and its snapshot until GC. abortDueToTimeout() detaches the handle before
+		// calling abort(), so this is a no-op there rather than a double-abort.
+		if (this.transaction) this.releaseReadTxn();
 		this.open = TRANSACTION_STATE.CLOSED;
+		this.drainCompletions();
 		for (const write of this.writes) {
 			if (write?.savedBlobs)
 				cleanupUnusedBlobs(write.savedBlobs, collectRetainedFileIds(write.store.getEntry(write.key)?.value));
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -1049,8 +1049,8 @@ export class DatabaseTransaction implements Transaction {
 		while (this.readTxnsUsed > 0) this.doneReadTxn(); // release the read snapshot when we abort, we assume we don't need it
 		// A write-only transaction never took a read reference (getReadTxn was never called), so the loop
 		// above releases nothing even though save() created a native handle; release it here instead of
-		// leaking the handle and its snapshot until GC. abortDueToTimeout() detaches the handle before
-		// calling abort(), so this is a no-op there rather than a double-abort.
+		// leaking the handle and its snapshot until GC. abortChainAfterRetries() detaches the handle
+		// before calling abort(), so this is a no-op there rather than a double-abort.
 		if (this.transaction) this.releaseReadTxn();
 		this.open = TRANSACTION_STATE.CLOSED;
 		this.drainCompletions();
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -579,9 +579,10 @@ export class DatabaseTransaction implements Transaction {
 	 * The stored entry of the last write eligible for replication confirmation. Two kinds of write are
 	 * skipped (rather than ending the search) so a trailing one cannot suppress confirmation for
 	 * replicable writes staged earlier: writes that explicitly opt out (audit-only markers, which stage
-	 * no record), and writes with no stored entry at all — a delete, or a delete on an `audit: false`
-	 * table, leaves nothing to read back. So a `put(A); delete(B)` transaction confirms on A's entry,
-	 * whose version is this transaction's (every write in a transaction is stamped with one version).
+	 * no record), and writes with no stored entry at all — a delete leaves a readable tombstone on any
+	 * audited or delete-tracking table, so only a delete on a table with neither is entry-less. Such a
+	 * `put(A); delete(B)` confirms on A's entry, whose version is this transaction's (every write in a
+	 * transaction is stamped with one version).
 	 */
 	lastConfirmableEntry(): Partial<Entry> | undefined {
 		for (let i = this.writes.length - 1; i >= 0; i--) {
--- a/resources/LMDBTransaction.ts
+++ b/resources/LMDBTransaction.ts
@@ -184,10 +184,12 @@ export class LMDBTransaction extends DatabaseTransaction {
 		this.open = options?.doneWriting ? TRANSACTION_STATE.LINGERING : TRANSACTION_STATE.OPEN;
 		let resolution;
 		const completions = [];
+		let commitCompletions: Promise<void>[];
 		let writeIndex = 0;
 		this.writes = this.writes.filter((write) => write); // filter out removed entries
 		const doWrite = (write) => {
-			write.commit(txnTime, write.entry, retries);
+			const completion = write.commit(txnTime, write.entry, retries);
+			if (typeof completion?.then === 'function') (commitCompletions ??= []).push(completion);
 		};
 		// this uses optimistic locking to submit a transaction, conditioning each write on the expected version
 		const nextCondition = () => {
@@ -235,6 +237,11 @@ export class LMDBTransaction extends DatabaseTransaction {
 					return true; // success. always success
 				});
 			}
+		}
+		if (commitCompletions) {
+			if (resolution) completions.push(...commitCompletions);
+			// A false resolution means an optimistic conflict and retries the writes.
+			else resolution = Promise.all(commitCompletions).then(() => true);
 		}
 
 		if (resolution) {
--- a/resources/LMDBTransaction.ts
+++ b/resources/LMDBTransaction.ts
@@ -228,7 +228,9 @@ export class LMDBTransaction extends DatabaseTransaction {
 				nextCondition();
 				if (commitCompletions) {
 					if (resolution) {
-						resolution = resolution.then((committed) => Promise.all(commitCompletions).then(() => committed));
+						resolution = Promise.resolve(resolution).then((committed) =>
+							Promise.all(commitCompletions).then(() => committed)
+						);
 					} else {
 						// Must resolve truthy or the conflict branch re-runs the writes.
 						resolution = Promise.all(commitCompletions).then(() => true);
--- a/resources/LMDBTransaction.ts
+++ b/resources/LMDBTransaction.ts
@@ -19,7 +19,6 @@ export const TRANSACTION_STATE = {
 	OPEN: 1, // the transaction is open and can be used for reads and writes
 	LINGERING: 2, // the transaction has completed a read, but can be used for immediate writes
 };
-let outstandingCommit;
 let confirmReplication;
 export function replicationConfirmation(callback) {
 	confirmReplication = callback;
@@ -254,14 +253,6 @@ export class LMDBTransaction extends DatabaseTransaction {
 		}
 
 		if (resolution) {
-			if (!outstandingCommit) {
-				outstandingCommit = resolution;
-				const clearOutstandingCommit = () => {
-					outstandingCommit = null;
-				};
-				outstandingCommit.then(clearOutstandingCommit, clearOutstandingCommit);
-			}
-
 			return resolution.then((resolution) => {
 				if (resolution) {
 					if (this.next) {
@@ -275,13 +266,11 @@ export class LMDBTransaction extends DatabaseTransaction {
 						// and when replication notifications come in, we count the number of confirms until we reach the desired number
 						const databaseName = this.writes[0].store.rootStore.databaseName;
 						const lastWrite = this.writes[this.writes.length - 1];
-						if (confirmReplication && lastWrite?.key != null)
+						const lastEntry =
+							lastWrite && !lastWrite.skipReplicationConfirmation && lastWrite.store.getEntry(lastWrite.key);
+						if (confirmReplication && lastEntry)
 							completions.push(
-								confirmReplication(
-									databaseName,
-									(lastWrite.store.getEntry(lastWrite.key) as any).localTime,
-									this.replicatedConfirmation
-								)
+								confirmReplication(databaseName, (lastEntry as any).localTime, this.replicatedConfirmation)
 							);
 					}
 					// commit succeeded; clean up files for any writes whose commit-handler took an early-return,
--- a/resources/LMDBTransaction.ts
+++ b/resources/LMDBTransaction.ts
@@ -224,11 +224,20 @@ export class LMDBTransaction extends DatabaseTransaction {
 			// will fail and retry due to contention. This is used to determine when to give up on optimistic writes and
 			// use a real (async) transaction to get exclusive access to the data
 			if (db?.retryRisk) db.retryRisk *= 0.99; // gradually decay the retry risk
-			if (this.writes.length + (db?.retryRisk || 0) < MAX_OPTIMISTIC_SIZE >> retries) nextCondition();
-			else {
+			if (this.writes.length + (db?.retryRisk || 0) < MAX_OPTIMISTIC_SIZE >> retries) {
+				nextCondition();
+				if (commitCompletions) {
+					if (resolution) {
+						resolution = resolution.then((committed) => Promise.all(commitCompletions).then(() => committed));
+					} else {
+						// Must resolve truthy or the conflict branch re-runs the writes.
+						resolution = Promise.all(commitCompletions).then(() => true);
+					}
+				}
+			} else {
 				// if it is too big to expect optimistic writes to work, or we have done too many retries we use
 				// a real LMDB transaction to get exclusive access to reading and writing
-				resolution = this.writes[0].store.transaction(() => {
+				const transactionResolution = this.writes[0].store.transaction(() => {
 					for (const write of this.writes) {
 						// we load latest data while in the transaction
 						write.entry = write.store.getEntry(write.key);
@@ -236,20 +245,19 @@ export class LMDBTransaction extends DatabaseTransaction {
 					}
 					return true; // success. always success
 				});
+				resolution = transactionResolution.then((committed) =>
+					commitCompletions ? Promise.all(commitCompletions).then(() => committed) : committed
+				);
 			}
-		}
-		if (commitCompletions) {
-			if (resolution) completions.push(...commitCompletions);
-			// A false resolution means an optimistic conflict and retries the writes.
-			else resolution = Promise.all(commitCompletions).then(() => true);
 		}
 
 		if (resolution) {
 			if (!outstandingCommit) {
 				outstandingCommit = resolution;
-				outstandingCommit.then(() => {
+				const clearOutstandingCommit = () => {
 					outstandingCommit = null;
-				});
+				};
+				outstandingCommit.then(clearOutstandingCommit, clearOutstandingCommit);
 			}
 
 			return resolution.then((resolution) => {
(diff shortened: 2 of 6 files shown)


unitTests/resources/reloadMarker.test.js
```diff
@@ -93,14 +93,14 @@ describe('table-reload marker (harper-pro#489)', () => {
...
 		// explicit numeric start to read the existing entries.
-		let marker;
-		for (const entry of ReloadTable.auditStore.getRange({ start: 1 })) {
-			if (entry.type === 'reload') {
-				marker = entry;
-				break;
-			}
-		}
-		assert.ok(marker, 'a reload audit entry was written');
+		const marker = await waitFor(
```
cb1kenobi: **Medium: polling this scan drops the commit-visibility contract the one-shot assertion was pinning**

`writeReloadMarker()` returns the promise from `transaction()`, and `transaction()` resolves from `transaction.commit({ doneWriting: true })` — so awaiting it is meant to imply the marker is committed and readable. That is a different guarantee from the two waits above: subscriber delivery really is deferred (`onCommit` -> `aftercommit` -> `await rest()`, see #1530), which is why those two were written as polls and this one was not.

Verified by mutation on this branch: with `writeReloadMarker()` changed so it resolves 200ms before its transaction lands, the pre-PR one-shot scan fails (`a reload audit entry was written`) while this polled version passes in 316ms. That regression class — resolve-before-visible — is now invisible here. Line 109 (`getHistoryOfRecord(1)`) still depends on the same immediate-visibility invariant, so the file is inconsistent about it now, and the repo leans the other way elsewhere (`auditLog.test.js:145-172`, `auditEntryRecordFlags.test.js:118-137`, `lingeringWriteCommit.test.js:74-88` all read the audit log one-shot after an await).

Worth adding: this assertion had no fixed delay to begin with, so it sits outside the "replace fixed-delay assertions" scope, and I could not find any record of it flaking (the `delay(20)` in `cache-test.mjs` is the real timing race).

Suggest taking the revert you already offered in the PR body: keep the two subscriber polls, restore the one-shot scan here.

—
_Generated by Barber AI_

unitTests/resources/reloadMarker.test.js
```diff
@@ -93,14 +93,14 @@ describe('table-reload marker (harper-pro#489)', () => {
...
+			() => {
+				for (const entry of ReloadTable.auditStore.getRange({ start: 1 })) {
+					if (entry.type === 'reload') return entry;
```
cb1kenobi: **Low: the scan can match another table's marker — filter on `tableId`**

`Table.auditStore` is the database-level txn-log store (`rootStore.auditStore`), so `getRange({ start: 1 })` walks every table in the database and this loop accepts the first `type === 'reload'` entry regardless of which table wrote it. Probed on this branch: write a marker for table A, scan through table B's `auditStore` in the same database, and B's scan finds A's entry (`marker.tableId 1` vs `B.tableId 2`).

So when the file runs as a suite, the entry asserted on here can be the marker the previous test wrote for `ReloadMarkerResnapshot`. With the poll added, that combination lets the test pass even when `ReloadMarkerAudit` never gets a marker at all — I hit exactly that while mutating: this case failed when run alone and passed when run with the file.

Suggested fix:
```suggestion
					if (entry.type === 'reload' && entry.tableId === ReloadTable.tableId) return entry;
```

—
_Generated by Barber AI_

unitTests/apiTests/cache-test.mjs
```diff
@@ -35,8 +34,13 @@ describe('test REST calls with cache table', () => {
 		delete data.ageInMonths; // harper#1484: computed scalars now also surface on default reads
 		response = await axios.put(`${baseUrl}/FourProp/3`, data);
 		assert.equal(response.status, 204);
-		await delay(20);
-		response = await axios(`${baseUrl}/SimpleCache/3`);
+		response = await waitFor(
+			async () => {
+				const cacheResponse = await axios(`${baseUrl}/SimpleCache/3`, { validateStatus: () => true });
```
cb1kenobi: **Low: a persistent 5xx now fails only at the 5s timeout, and the status never reaches the report**

Taking your trade-off #1 empirically: a persistent 5xx does still fail the test, so the assertion is not unfalsifiable — but the diagnosis degrades. Running this exact poll shape against a stubbed always-500 endpoint with the real `waitFor`: it fails after 5013ms with `AssertionError: source update did not reach SimpleCache`, where the pre-PR line failed at 0ms with `Request failed with status code 500`. `waitFor` uses the static `message` on timeout, so the observed status is dropped entirely.

Narrowing the retry set keeps the propagation retry and restores fast, specific failure on server errors (a 404 while the cache refills still retries):

```suggestion
				const cacheResponse = await axios(`${baseUrl}/SimpleCache/3`, { validateStatus: (status) => status < 500 });
```

—
_Generated by Barber AI_

unitTests/resources/reloadMarker.test.js
kriszyp: Addressed in `248e9fdfe`: restored the one-shot audit scan, so the test again pins immediate marker visibility when `writeReloadMarker()` resolves.

_— GPT-5 Codex_

unitTests/resources/reloadMarker.test.js
kriszyp: Addressed in `248e9fdfe`: the database-wide audit scan now requires `entry.tableId === ReloadTable.tableId`.

_— GPT-5 Codex_

unitTests/apiTests/cache-test.mjs
kriszyp: Addressed in `248e9fdfe`: `validateStatus` now accepts only statuses below 500, preserving retries for non-5xx responses while Axios surfaces server errors immediately.

_— GPT-5 Codex_

cb1kenobi:
Reviewed `248e9fdf` — no issues found. All three prior findings are addressed; this PR looks good, nice job!

I re-ran the decisive mutation to confirm the reverted assertion is genuinely falsifiable again, not just the right shape:

- **Reload-marker one-shot scan (the Medium).** I deferred the `writeReloadMarker()` audit write by 200 ms in production code (`resources/Table.ts`), rebuilt with `tsc -p tsconfig.build.json`, and confirmed the injected marker landed exactly once in `dist/resources/Table.js` before trusting the run. Test 3 now **fails** (`AssertionError: a reload audit entry was written`), where the polled version had absorbed the same delay and passed. Tests 1 and 2 kept passing but slowed from 41 ms/107 ms to 233 ms/323 ms, which independently confirms the defer was live. The commit-visibility contract is pinned again.
- **`tableId` filter.** This one is doing real work. Probing the audit store during test 3 shows `ReloadTable.tableId = 2` while the scan range contains reload entries for `[1, 2]` — entry 1 is the marker from test 2, since `auditStore` is the database-level txn-log store shared across tables in the database. Without the filter the scan matched test 2's marker first and the assertions passed against the wrong table's entry. Worth noting the filter is not vacuous (both sides are real numbers, so no `undefined === undefined` no-op).
- **No flake reintroduced.** 12/12 clean runs of `unitTests/resources/reloadMarker.test.js`, and test 3 passes both standalone and as part of the full file.

One thing I could not verify: `unitTests/apiTests/cache-test.mjs` would not run here — an unrelated long-running mocha process on this machine holds the RocksDB lock on `/Users/chris/harper/database/data`, and the suite fails at module load before the test DB path is set. The `validateStatus: (status) => status < 500` change is reviewed statically only. It reads correctly: `waitFor` does not catch, so a 5xx now rejects out of the poll and fails fast with the axios status attached, while a 4xx still retries — which is the intended split.

—
_Generated by Barber AI_

resources/LMDBTransaction.ts
```diff
@@ -258,7 +275,7 @@ export class LMDBTransaction extends DatabaseTransaction {
 						// and when replication notifications come in, we count the number of confirms until we reach the desired number
 						const databaseName = this.writes[0].store.rootStore.databaseName;
 						const lastWrite = this.writes[this.writes.length - 1];
-						if (confirmReplication && lastWrite)
+						if (confirmReplication && lastWrite?.key != null)
```
cb1kenobi: **Medium: `key != null` is the wrong predicate, and the RocksDB twin of this line is still unguarded**

Two things, one fix.

**(a) A root-topic publish is a legitimate null-key write that needs confirmation.** `_writePublish` does `id ??= null` deliberately (`Table.ts:4484`, "we allow the null id for publishing so that you can publish to the root topic"), and unlike the marker it *does* store a record, so `getEntry(null)` returns a real entry with a version. What actually distinguishes the marker is that it writes no record — not that its key is null.

**(b) `DatabaseTransaction.ts:861` still has the unguarded `if (confirmReplication && lastWrite)`** and dereferences `.version` on `getEntry(lastWrite.key)`. RocksDB is the default engine, and there `txnForContext` *claims* the context's own transaction (`Table.ts:5599-5603`), so the marker lands on a transaction that can carry `replicatedConfirmation`.

Measured in a clean worktree at this head (default engine, rebuilt `dist` per variant):

| `DatabaseTransaction.ts:861` guard | root-topic publish | keyed publish | reload marker |
|---|---|---|---|
| `lastWrite` (as shipped) | confirmed | confirmed | **TypeError: Cannot read properties of undefined (reading 'version')** |
| `lastWrite?.key != null` (this PR's predicate, mirrored) | **confirmation silently skipped** | confirmed | no crash |
| `lastEntry` (below) | confirmed | confirmed | no crash |

So mirroring this line as-is would trade a crash for a silent drop. Guarding on the entry is exact on both engines and also covers any last write whose entry is simply absent:

```ts
const lastEntry: any = lastWrite && lastWrite.store.getEntry(lastWrite.key);
if (confirmReplication && lastEntry)
	completions.push(confirmReplication(databaseName, lastEntry.localTime, this.replicatedConfirmation));
```

(`.version` instead of `.localTime` in the `DatabaseTransaction.ts` copy.) It also removes the double `getEntry` the current shape performs.

Note the whole block also keys off the *last* write only, so a transaction ending in a marker drops confirmation for the keyed writes ahead of it either way — the entry guard doesn't fix that, but it's a smaller surface than it looks since the marker commits in its own transaction.

—
_Generated by Barber AI_

unitTests/resources/reloadMarker.test.js
```diff
@@ -89,13 +93,20 @@ describe('table-reload marker (harper-pro#489)', () => {
 		});
 		// touch the audit stream first so the table has an audit store wired up
 		await ReloadTable.put({ id: 1, name: 'row' });
-		await ReloadTable.writeReloadMarker();
+		const originalRetryRisk = ReloadTable.primaryStore.retryRisk;
+		if (isLMDB) ReloadTable.primaryStore.retryRisk = 101;
```
cb1kenobi: **Medium: forcing the contention path makes this test pass with the fix reverted**

With `retryRisk = 101` the commit takes the `store.transaction()` fallback, and that path's own promise already covers the audit write's durability — so the completion contract this PR adds is never what makes the assertion below pass. The bug being fixed only manifests on the optimistic path, which this test no longer takes.

Measured at this head under LMDB (3 runs per cell, `dist` rebuilt between variants):

| production code | test as shipped | same test, `retryRisk` line removed |
|---|---|---|
| as-is | 4 passing | 4 passing |
| `Table.ts` marker `commit` no longer `return`s the promise | **4 passing** | **1 failing** — `a reload audit entry was written` |

That bottom-left cell is the problem: the regression test for the headline fix currently cannot fail. (I confirmed the forcing itself does work — instrumenting the branch shows the contention path taken with `retryRisk=99.99` — it just isn't where the bug lives.)

Suggest covering both rather than swapping: keep this assertion on the default optimistic path and add the contention fallback as a second case, e.g. wrap the body in a `for (const forceContention of [false, true])`.

—
_Generated by Barber AI_

resources/DatabaseTransaction.ts
```diff
@@ -253,7 +253,7 @@ export type TransactionWrite = {
 	entry?: Partial<Entry>;
 	before?: () => void | Promise<void>;
 	beforeIntermediate?: () => void | Promise<void>;
-	commit?: (txnTime: number, existingEntry: Partial<Entry>, retry: boolean, transaction: any) => void;
+	commit?: (txnTime: number, existingEntry: Partial<Entry>, retry: boolean, transaction: any) => MaybePromise<void>;
```
cb1kenobi, line 256: **Low: the widened return type is honored by only one of the two transaction implementations**

`commit` is now `MaybePromise<void>` for every `TransactionWrite`, but only `LMDBTransaction`'s `doWrite` (`LMDBTransaction.ts:191`) collects the promise. The RocksDB path — `save()` at line 647 of this file — calls `operation.commit(...)` and discards the result, so a promise returned there would be neither awaited nor rejection-handled.

No live bug today: I instrumented line 647 and ran the full `unitTests/resources` suite on RocksDB (1556 passing) — zero commit callbacks returned a thenable, because `updateRecord` resolves to `putSync` on that engine. But the type now advertises support the default engine doesn't provide, and `Table.ts`'s marker callback returns on both engines, so the next callback that returns a promise on the Rocks path will fail silently.

Cheapest fix is to mirror the handling already used for `before`/`beforeIntermediate` a few lines above:

```ts
const completion = operation.commit(txnTime, operation.entry, this.retries > 0, transaction);
if ((completion as any)?.then) this.completions.push(completion as Promise<void>);
```

Otherwise a note on this type that only the LMDB path awaits it would keep the contract honest.

—
_Generated by Barber AI_

resources/LMDBTransaction.ts
```diff
@@ -222,27 +224,42 @@ export class LMDBTransaction extends DatabaseTransaction {
...
 
 		if (resolution) {
 			if (!outstandingCommit) {
 				outstandingCommit = resolution;
-				outstandingCommit.then(() => {
+				const clearOutstandingCommit = () => {
 					outstandingCommit = null;
-				});
+				};
+				outstandingCommit.then(clearOutstandingCommit, clearOutstandingCommit);
```
cb1kenobi: **Nit: `outstandingCommit` is write-only, so clearing it on rejection changes no behavior**

The module-level `outstandingCommit` (`LMDBTransaction.ts:22`) is assigned at 258 and read only by the `if (!outstandingCommit)` check at 257 that decides whether to assign it. Nothing else in the repo reads it and it isn't exported, so it gates nothing: a wedged value costs nothing, and clearing it can't let a second commit race a still-in-flight failed one (it was never a lock — the `return resolution.then(...)` below is outside the block). The live equivalent is `DatabaseTransaction`'s `trackOutstandingCommit` / `checkOverloaded`, which `LMDBTransaction` never calls.

The two-arg handler is still the right shape (it does suppress a detached rejection now that `resolution` can reject through `Promise.all(commitCompletions)`). Just worth not crediting it with more in the description — or deleting the vestige outright.

—
_Generated by Barber AI_

resources/LMDBTransaction.ts
kriszyp: Addressed in `4dcea2456`: both engines now confirm only when the last write has a stored entry and has not explicitly opted out. Reload markers carry the explicit opt-out, while the test proves that a stored null-key root-topic publish still receives confirmation—even when the later marker shares that key.

_— GPT-5 Codex_

unitTests/resources/reloadMarker.test.js
kriszyp: Addressed in `4dcea2456`: the LMDB marker test now runs once with the default optimistic path and once with forced contention for the exclusive fallback. RocksDB retains its parity run.

_— GPT-5 Codex_

resources/DatabaseTransaction.ts
kriszyp: Addressed in `4dcea2456`: RocksDB `save()` now collects a promise returned by `TransactionWrite.commit`, and a focused RocksDB test fails on the old behavior and passes when the callback completion is observed.

_— GPT-5 Codex_

resources/LMDBTransaction.ts
kriszyp: Addressed in `4dcea2456`: the disconnected LMDB `outstandingCommit` variable and its assignment/cleanup block were removed. The PR description no longer attributes overload or rejection behavior to it.

_— GPT-5 Codex_

unitTests/resources/transaction.test.js
```diff
@@ -74,6 +74,25 @@ describe('Transactions', () => {
...
+			completionObserved = true;
+			return then.apply(this, args);
+		};
+		transaction.addWrite({
+			key: 'async-commit-callback',
+			store: TxnTest.primaryStore,
+			commit: () => completion,
+		});
+		await transaction.commit({ doneWriting: true });
+		assert.equal(completionObserved, true, 'the callback completion was awaited');
```
cb1kenobi: **Low: this test passes when the completion is merely observed, not awaited**

`completionObserved` flips when `.then` is *called* — which `Promise.all(this.completions)` does, but so would a fire-and-forget `completion.then(() => {})`. The assertion therefore pins "the callback's promise was touched", not the contract its own message states, so the fix at `DatabaseTransaction.ts:648` can regress to fire-and-forget without turning this red.

Measured in a clean worktree at `4dcea245` (default engine, `dist` rebuilt per variant, mutation marker confirmed exactly once in `dist/resources/DatabaseTransaction.js` before each run):

| `DatabaseTransaction.ts:648` | this test | with the assertion below |
|---|---|---|
| as shipped (`this.completions.push(completion)`) | pass | pass |
| result discarded (pre-PR shape) | **fail** | **fail** |
| `completion.then(() => {}, () => {})` — observed, not awaited | **pass** | **fail** |

Gating on a completion that only settles on a later tick closes that middle row (verified: passes as shipped, fails on the fire-and-forget mutant):

```suggestion
		let settled = false;
		let release;
		const completion = new Promise((resolve) => (release = resolve)).then(() => {
			settled = true;
		});
		transaction.addWrite({
			key: 'async-commit-callback',
			store: TxnTest.primaryStore,
			commit: () => completion,
		});
		const committed = transaction.commit({ doneWriting: true });
		setImmediate(release);
		await committed;
		assert.equal(settled, true, 'commit resolved only after the callback completion settled');
```

—
_Generated by Barber AI_

resources/LMDBTransaction.ts
cb1kenobi: Resolved at `4dcea245` — and the fix went further than the suggestion: the entry guard is now on both engines *and* the marker carries an explicit `skipReplicationConfirmation` flag, which short-circuits before `getEntry` so the marker never depends on whether a prior root-topic publish happened to leave a record at the null key. Nice call on the flag; it makes the intent readable rather than inferred.

Re-measured in a clean worktree at this head (`dist` rebuilt per variant, mutation marker asserted exactly once in `dist/` before every run):

| variant | engine | result |
|---|---|---|
| as shipped | RocksDB / LMDB | 41 passing / 39 passing, 0 failing |
| `DatabaseTransaction.ts` guard reverted to `if (confirmReplication && lastWrite)` | RocksDB | **fail** — `a LOCAL_ONLY marker has nothing to confirm with peers` |
| `lastWrite?.key != null` mirrored onto `DatabaseTransaction.ts` | RocksDB | **fail** — `a stored root-topic publish is confirmed` |
| `lastWrite?.key != null` restored on `LMDBTransaction.ts` | LMDB | **fail** — `a stored root-topic publish is confirmed` |
| `skipReplicationConfirmation: true` removed from `Table.ts` | RocksDB **and** LMDB | **fail** — `a LOCAL_ONLY marker has nothing to confirm with peers` |

All three cases hold: root-topic publish confirms, keyed publish confirms, marker neither crashes nor confirms.

I also checked the one thing the entry guard could plausibly have regressed — a transaction whose last write is a **delete**. It does not: a delete on an audited or delete-tracking table stores a tombstone (`updateRecord(id, null, …)`), so `getEntry` stays truthy with a real `version`/`localTime` and confirmation still fires. Probed directly on both engines (`entryAfterDelete truthy=true`, `confirmationCalls=1`). The only falsy-after-delete path is `audit: false` + no `trackDeletes`, which can't be replicated anyway — and that case used to *throw* on RocksDB, so it's strictly better now.

—
_Generated by Barber AI_

unitTests/resources/reloadMarker.test.js
cb1kenobi: Resolved at `4dcea245` — the `for (const forceContention of …)` loop is exactly the shape that fixes it, and putting the optimistic pass first means the real regression surfaces on iteration one.

Re-ran the same mutation at this head (`Table.ts` marker `commit` no longer `return`s its `updateRecord()` promise; `dist/resources/Table.js` verified to contain the mutated line and zero copies of the original):

| engine | result |
|---|---|
| LMDB | **1 failing** — `a reload audit entry was written` |
| RocksDB | 41 passing (expected — `updateRecord` resolves to `putSync` there, so the bug can't manifest) |

The test now fails on its own regression. Thanks for taking the both-paths route rather than swapping one for the other.

—
_Generated by Barber AI_

resources/DatabaseTransaction.ts
cb1kenobi: Resolved at `4dcea245` — `save()` now collects the completion, and the new `transaction.test.js` case covers it. Verified by mutation at this head: discarding the result the pre-PR way turns `waits for promise-returning commit callbacks on RocksDB` red (`the callback completion was awaited`), with the mutation marker confirmed exactly once in `dist/resources/DatabaseTransaction.js`.

One follow-up on the test's sensitivity rather than the fix — commented separately on `unitTests/resources/transaction.test.js`: a fire-and-forget `completion.then(() => {})` still passes it, so the assertion pins "observed" rather than "awaited".

—
_Generated by Barber AI_

resources/LMDBTransaction.ts
cb1kenobi: Resolved at `4dcea245` — vestige deleted.

Worth recording since `LMDBTransaction.ts` is net **-11** here: everything removed was the `outstandingCommit` block plus reflow of the `confirmReplication` call. The completion-collection machinery added last round is fully intact — `doWrite` still pushes thenables into `commitCompletions`, and both the optimistic path (`Promise.all(commitCompletions).then(() => true)`) and the exclusive `store.transaction()` path still gate `resolution` on it. Confirmed behaviorally, not just by reading: dropping the `return` from the marker's commit callback still turns the LMDB reload-marker test red at this head, which it could only do if the completion is genuinely collected and awaited.

—
_Generated by Barber AI_

resources/DatabaseTransaction.ts
```diff
@@ -574,6 +575,31 @@ export class DatabaseTransaction implements Transaction {
...
+	 * The stored entry of the last write eligible for replication confirmation. Writes that opt out
+	 * (audit-only markers, which stage no record) are skipped rather than ending the search, so a
+	 * trailing marker cannot suppress confirmation for replicable writes staged earlier.
+	 */
+	lastConfirmableEntry(): Partial<Entry> | undefined {
+		for (let i = this.writes.length - 1; i >= 0; i--) {
+			const write = this.writes[i];
+			if (!write || write.skipReplicationConfirmation) continue;
+			const entry = write.store.getEntry(write.key);
+			if (entry) return entry;
```
cb1kenobi, line 592: **Low: the falsy-entry skip is a second behavior change, and it is the one nothing covers**

The comment above justifies skipping only `skipReplicationConfirmation` writes, but `if (entry) return entry` also skips writes whose stored entry is legitimately absent. That is not a no-op: on an `audit: false` table a delete leaves no stored entry at all, so a `put(A); delete(B)` transaction that previously produced *no* confirmation now confirms on A's entry. I ran both halves of that on this head to be sure of it rather than reasoning about it.

The value it lands on is right today (RocksDB stamps every write in a transaction with one version, so A's version is this transaction's), and confirming something beats confirming nothing. The gap is purely coverage. Replacing this line with `return entry || undefined;` — falsy ends the search, exactly what the comment describes — leaves the whole `unitTests/resources` suite green on **both** engines (RocksDB 1558 passing, LMDB 1356 passing / 0 failing). The opt-out skip, by contrast, is pinned: reverting `lastConfirmableEntry()` to last-write-only reddens `reloadMarker.test.js` on both engines.

Two small follow-ups:
- widen the comment so it describes both skips, and
- add a delete-last case to `reloadMarker.test.js` so the fallback is held in place.

One thing worth knowing while you decide which way to take it: the walk can also select a *stale* entry — a write whose `validate()` returned false leaves its key at the pre-existing version. Pro's waiter resolves an already-replicated version immediately only when its per-peer seed map is populated; on a seed miss it registers a waiter nothing will satisfy and rejects 504 after the 900 s timeout (`harper-pro replication/knownNodes.ts:862`, `:885-898`). Narrow, and the old code wasn't better here — but the fallback widens which transactions can reach it, which is an argument for the comment being explicit.

—
_Generated by Barber AI_

resources/DatabaseTransaction.ts
```diff
@@ -574,6 +575,31 @@ export class DatabaseTransaction implements Transaction {
...
+	}
+
+	/**
+	 * Stage an async operation that commit() must wait for. Staging can happen a turn or more before
+	 * commit() attaches its Promise.all, so the no-op rejection handler is attached here: without it a
+	 * rejection in that window is an unhandled rejection (fatal under --unhandled-rejections=strict).
+	 * The rejection still surfaces through commit()'s Promise.all.
+	 */
+	stageCompletion(completion: Promise<void>) {
+		completion.then(undefined, () => {});
```
cb1kenobi, line 603: **Low: this handler does its job, but on the abort path it turns a loud failure into a silent one**

The reassuring half first: the rejection genuinely still surfaces. Removing this line reddens the new rejection-window test, and `commit()` rejects with the callback's error either way — the no-op only marks the original promise handled, and `Promise.all` still sees it. Same for the LMDB twin.

The gap is the path where `commit()` is never reached. `abort()` (line 1029) clears `writes` but not `completions`, so a staged completion that rejects after an abort is now swallowed with no log and no unhandled-rejection report. Measured on this head: stage a rejecting commit callback through `addWrite`, call `abort()`, and you get zero `unhandledRejection` events with `completions.length === 1` left behind. Before this change that failure at least announced itself.

Draining in `abort()` closes it, and clears a latent second thing at the same time — completions surviving an abort mean a reused transaction object's next `commit()` rejects with the *previous* batch's error. I reproduced that by forcing the reuse; today's abort paths release the context so it isn't reachable on its own, but it is a sharp edge to leave armed:

```js
for (const completion of this.completions)
    completion.then(undefined, (error) => harperLogger.warn?.('staged completion failed after abort', error));
this.completions = [];
```

—
_Generated by Barber AI_

unitTests/resources/transaction.test.js
```diff
@@ -74,6 +75,73 @@ describe('Transactions', () => {
...
+			// keyed write: the conditional batch stages it, so this covers doWrite's non-null-key path
+			commit: () => {
+				TxnTest.primaryStore.put('lmdb-async-commit-callback', { name: 'staged' });
+				return completion;
+			},
+		});
+		// released well after the batch itself would resolve, so a commit that does not await the
+		// callback completion resolves first and fails the ordering assertion
+		const committed = transaction.commit({ doneWriting: true }).then(() => order.push('commit'));
+		setTimeout(release, 50);
```
cb1kenobi: **Low: the 50 ms release is a race whose only failure mode is a silent false pass**

The test does its job today — I removed the `Promise.all(commitCompletions)` await from LMDB's optimistic branch and this went red, so the coverage is real. What I'd flag is how durable that stays.

The negative case rests on the conditional batch resolving inside 50 ms. I measured it at p50 8.0 ms / max 16 ms over 25 keyed-write commits on an idle machine — comfortable now, but if a loaded runner pushes the batch past 50 ms then `release` wins the race and `order` reads `['completion', 'commit']` even with the await gone. The test passes while guarding nothing, and because it never *red*-flakes nobody finds out.

Waiting on the batch itself drops the timer entirely:

```js
const committed = transaction.commit({ doneWriting: true }).then(() => order.push('commit'));
await TxnTest.primaryStore.flushed;
assert.deepEqual(order, [], 'commit has not resolved while the callback completion is pending');
release();
await committed;
assert.deepEqual(order, ['completion', 'commit']);
```

—
_Generated by Barber AI_

unitTests/resources/transaction.test.js
```diff
@@ -74,6 +75,73 @@ describe('Transactions', () => {
...
+			store: TxnTest.primaryStore,
+			commit: () => completion,
+		});
+		const committed = transaction.commit({ doneWriting: true });
+		setImmediate(release);
+		await committed;
+		assert.equal(settled, true, 'commit resolved only after the callback completion settled');
+	});
+	it('surfaces a commit-callback rejection without an unhandled rejection', async function () {
+		if (isLMDB) return this.skip();
```
cb1kenobi: **Low: the LMDB twin of the rejection-window fix has no coverage**

This skip leaves the matching `completion.then(undefined, () => {})` in `LMDBTransaction.doWrite` (`resources/LMDBTransaction.ts:195`) unguarded — removing that line leaves the entire LMDB `unitTests/resources` suite green (1356 passing, 0 failing).

It is load-bearing, so it is worth pinning: running the LMDB shape directly, a keyed write whose commit callback rejects produces two `unhandledRejection` events without the line and zero with it, with `commit()` rejecting on the callback error either way — the same scenario this test asserts for RocksDB. Pointing the case at `LMDBTransaction` instead of skipping it would cover both engines from one test.

Separately, on the same case: after `commit()` rejects, `this.transaction` is a live `RocksTransaction` that is never committed or aborted, so it leaks a native handle and read snapshot for the remainder of the mocha process. Harmless in every run I did, but it is exactly the sort of residue that resurfaces as somebody else's flake later.

—
_Generated by Barber AI_

resources/LMDBTransaction.ts
```diff
@@ -257,14 +271,10 @@ export class LMDBTransaction extends DatabaseTransaction {
 						// if we want to wait for replication confirmation, we need to track the transaction times
 						// and when replication notifications come in, we count the number of confirms until we reach the desired number
 						const databaseName = this.writes[0].store.rootStore.databaseName;
-						const lastWrite = this.writes[this.writes.length - 1];
-						if (confirmReplication && lastWrite)
+						const lastEntry = this.lastConfirmableEntry();
```
cb1kenobi, line 274: **Low (pre-existing, not from this PR): production never registers LMDB's `confirmReplication`**

Line 22 declares a module-level `confirmReplication` independent of `DatabaseTransaction`'s. The only production registration is `harper-pro replication/knownNodes.ts:930`, which imports `replicationConfirmation` from `core/resources/DatabaseTransaction.ts` — nothing anywhere imports it from `LMDBTransaction`. So on LMDB-backed tables `confirmReplication` stays `undefined`, the guard on the next line never fires, and an `X-Replicate-To: …;confirm=N` request resolves without waiting on any peer.

Nothing to change here — raising it mainly so the LMDB assertions added in `reloadMarker.test.js` aren't read as production parity evidence: they pass because the test registers this module's setter directly. Probably deserves its own ticket.

—
_Generated by Barber AI_

kriszyp: Confirmed and left as-is: `LMDBTransaction`'s module-level `confirmReplication` has no production registration, so the LMDB assertions in `reloadMarker.test.js` exercise the test-registered setter, not production parity. That is a pre-existing production gap outside this PR's scope, and it is being routed to its own issue rather than fixed here. Thanks for catching the parity distinction — the test comments do not claim otherwise.

— Claude Opus (KrAIs)

resources/DatabaseTransaction.ts
```diff
@@ -574,6 +575,50 @@ export class DatabaseTransaction implements Transaction {
 		this.overloadChecked = true; // only check this once, don't interrupt ongoing transactions that have already made writes
 	}
 
+	/**
+	 * The stored entry of the last write eligible for replication confirmation. Two kinds of write are
+	 * skipped (rather than ending the search) so a trailing one cannot suppress confirmation for
+	 * replicable writes staged earlier: writes that explicitly opt out (audit-only markers, which stage
+	 * no record), and writes with no stored entry at all — a delete, or a delete on an `audit: false`
+	 * table, leaves nothing to read back. So a `put(A); delete(B)` transaction confirms on A's entry,
+	 * whose version is this transaction's (every write in a transaction is stamped with one version).
```
cb1kenobi: **Nit: the widened comment overstates which deletes are entry-less**

Thanks for widening this — the two skips are distinct and both now described. One correction on the second one: a plain delete is *not* entry-less. On any audited or delete-tracking table a delete stores a tombstone, so `getEntry` stays truthy and the walk stops there rather than falling back.

Measured at this head with the walk instrumented (`dist` rebuilt, marker asserted exactly once), trailing-delete transactions on the default engine:

| table | last write | `getEntry(write.key)` | walk lands on |
|---|---|---|---|
| `audit: true` | `delete(1)` | **truthy** (tombstone) | the delete's own entry |
| `audit: false` | `delete(1)` | falsy | the `put(2)` before it |

So only the `audit: false` (no delete tracking) case reaches the fallback — which is exactly the case the new `reloadMarker.test.js` block builds, and it does pin it: mutating this line to `return entry || undefined` now fails `the put staged before an entry-less delete is still confirmed` on **both** engines. The prose just claims a wider class than the code produces, and the `put(A); delete(B)` example only confirms on A under `audit: false`.

Suggested fix:
```suggestion
	 * no record), and writes with no stored entry at all — a delete leaves a readable tombstone on any
	 * audited or delete-tracking table, so only a delete on an `audit: false` table is entry-less. Such
	 * a `put(A); delete(B)` confirms on A's entry, whose version is this transaction's (every write in
	 * a transaction is stamped with one version).
```

—
_Generated by Barber AI_

resources/DatabaseTransaction.ts
cb1kenobi: Resolved at `70b33aa2` — the comment now covers both skips and, more to the point, the falsy path is pinned. The previously-surviving mutation dies:

| `DatabaseTransaction.ts:591` | RocksDB | LMDB |
|---|---|---|
| as shipped (`if (entry) return entry`) | 1561 passing, 0 failing | 1357 passing, 0 failing |
| `return entry \|\| undefined` (falsy ends the search) | **fail** — `the put staged before an entry-less delete is still confirmed` (expected 3, actual 2) | **fail** — same assertion |

(`dist/resources/DatabaseTransaction.js` rebuilt per variant, mutation marker asserted exactly once and the original string confirmed absent before each run.)

Good call building the case on an `audit: false` table specifically — I instrumented the walk and that really is the only shape that reaches the fallback; a delete on an audited table stops the walk on its own tombstone. That's the one thing the new prose overstates, commented separately on the comment lines. Nit only — the code and the test are right.

—
_Generated by Barber AI_

resources/DatabaseTransaction.ts
cb1kenobi: Resolved at `70b33aa2` — and `drainCompletions()` does the right thing with a post-abort rejection: it is **consumed and logged at warn**, not awaited and not dropped. The reused-transaction edge is closed too, since the array is cleared.

The test pins the behavior rather than the comment. Three separate mutations at this head, each killing it on a different assertion (`dist` rebuilt per variant, marker verified exactly once):

| mutation | result |
|---|---|
| `drainCompletions()` call removed from `abort()` | **fail** — `an aborted batch cannot carry its completions into a later commit on a reused transaction` |
| `drainCompletions()` clears but logs nothing (`.then(undefined, () => {})`) | **fail** — `the abandoned rejection is logged rather than silently dropped` |
| `if (this.transaction) this.releaseReadTxn();` removed | **fail** — `the abort releases the native handle rather than leaking it` |

So the clearing, the logging, and the handle release are each independently held. The middle row is the one I most wanted to see red — a test that only checked `completions === []` would have let the log quietly disappear later.

On the handle release itself, I checked the double-abort question since `RocksTransaction.abort()` throws on an already-aborted handle: every path that aborts the native transaction nulls `this.transaction` first (`commit()` line 813, `doneReadTxn()`, `abortChainAfterRetries()`), so the new line only fires on a genuinely live handle — and `releaseReadTxn()` wraps the abort in try/catch regardless. I also traced `completeDeferredContextRelease()` for an interaction with the timeout-poison reuse contract: `pendingContextRelease` is only ever set on the `readTxnsUsed > 0` path, where the loop above has already nulled the handle, so the new line is a no-op there. Both engines green (RocksDB 1561, LMDB 1357, 0 failing).

—
_Generated by Barber AI_

unitTests/resources/transaction.test.js
cb1kenobi: Resolved at `70b33aa2` — the timer is gone and the coverage held. Gating on `store.flushed` makes the outcome depend on microtask ordering rather than wall-clock, so the silent-false-pass mode I was worried about is closed.

Re-ran the same mutation (LMDB optimistic branch no longer wrapping `resolution` in `Promise.all(commitCompletions)`), 3 runs, `dist/resources/LMDBTransaction.js` marker verified exactly once:

| variant | LMDB |
|---|---|
| as shipped | 3/3 pass |
| optimistic `Promise.all(commitCompletions)` dropped | **3/3 fail** — `commit resolved only after the callback completion`, actual `['commit', 'completion']` |

One observation for the record rather than a request: it is the final ordering assertion that kills the mutant, not the new intermediate `assert.deepEqual(order, [], …)` — under the mutant, commit has still not resolved when `flushed` settles, so the intermediate check passes and the reversal shows up one await later. The intermediate assertion is still worth keeping as documentation of intent; I just wouldn't count it as the guard.

—
_Generated by Barber AI_

unitTests/resources/transaction.test.js
cb1kenobi: Resolved at `70b33aa2` — pointing the case at `LMDBTransaction` instead of skipping it covers both engines from one test, which is the shape I was hoping for, and the keyed-write staging inside the callback is what makes it actually traverse the conditional batch.

The previously-surviving mutation now dies:

| `LMDBTransaction.ts` `doWrite` | LMDB |
|---|---|
| as shipped | 2/2 pass |
| `completion.then(undefined, () => {});` removed | **2/2 fail** — `the completion stays handled across the commit aggregation gap`, with exactly the two `unhandledRejection` events I measured by hand last round |

The RocksDB twin is pinned by the same test: removing `stageCompletion`'s no-op handler fails `the staged completion has a rejection handler before commit` (also 2 events). Both engines now hold their own handler, and the second `setImmediate` pair is what separates the two windows — nice touch.

(`dist` rebuilt per variant, marker asserted exactly once each time.)

The leaked `RocksTransaction` handle I raised in the second half of this comment is addressed too, via the `releaseReadTxn()` call added to `abort()` — verified separately on the abort thread.

—
_Generated by Barber AI_

resources/LMDBTransaction.ts
cb1kenobi: Still stands at `70b33aa2`, and unchanged by this head — flagging that only so the delta isn't misread. `LMDBTransaction.ts` is `+1/-0` here and that line is the `drainCompletions()` call added to `abort()`, not a registration change; `harper-pro replication/knownNodes.ts:8` still imports `replicationConfirmation` from `core/resources/DatabaseTransaction.ts` and nothing imports this module's setter. Ticket-worthy on its own, as before — nothing to do in this PR.

Small related note on that `+1` line while I was in there: `LMDBTransaction` never populates `this.completions` (its commit completions live in a local `commitCompletions`, and `stageCompletion` is only reached from `DatabaseTransaction.save()`, which `LMDBTransaction.save()` overrides to a no-op). I instrumented `drainCompletions()` and ran the full LMDB `unitTests/resources` suite: **0 hits** with a non-empty array on an `LMDBTransaction`. So it is harmless forward-compatible symmetry rather than a live fix — worth knowing if someone later reads it as evidence that LMDB had the same abort-path gap. No change requested.

—
_Generated by Barber AI_

resources/DatabaseTransaction.ts
kriszyp: Applied at `d92bacc5`, with one widening of your wording. The tombstone is written under `if (audit || trackDeletes)` (`resources/Table.ts:3073`), so an `audit: false` table that tracks deletes — which `Table.ts:712` turns on implicitly whenever the table has subscribers — also stops the walk on its own tombstone. The only entry-less shape is a delete on a table with *neither*, which is what the test builds. The comment now says that, and the test comment above the `NoAuditTable` block was reworded to match.

Thanks for the instrumented table — it is what made the distinction checkable rather than arguable.

— Claude Opus (KrAIs)

resources/LMDBTransaction.ts
kriszyp: Agreed on both points, and no change made here.

- The missing production registration of `LMDBTransaction`'s module-level `confirmReplication` is a real pre-existing gap; it is being routed to its own issue rather than widened into this PR.
- On `drainCompletions()` in `abort()`: your reading matches ours — `LMDBTransaction.save()` overrides the staging path, so `this.completions` is empty on that class today and the call is forward-compatible symmetry, not a live fix. Useful to have that measured rather than assumed.

Resolving; both are tracked outside this PR.

— Claude Opus (KrAIs)

heskew (approved):
Reviewed; no blockers found. Thorough handling of async transaction commit callbacks and reload marker confirmation across both RocksDB and LMDB engines.

---

Issue #2216, opened 2026-08-19
Direct `sql` operation bypasses the operations allowlist (gate 1)


cb1kenobi:
### Summary

A direct `sql` operation is authorized only by the AST table/attribute check, never by the role's `operations` allowlist (gate 1) or `requires_su`. A role that omits `sql` from its `operations` can still run arbitrary SQL, bounded only by its table CRUD grants.

### Detail

In `chooseOperation` (`server/serverHelpers/serverUtilities.ts:264-266`), SQL is excluded from the `verifyPerms` block (`!isSqlOperation`) and handled solely by `verifyPermsAST`, which validates only the statement's tables and attributes — never the `operations` allowlist and never `requires_su`. A table-free statement gives the AST check nothing to validate.

This is the SQL-surface sibling of the known GraphQL ops-allowlist bypass.

### Scope

- Pre-existing behavior; matches the control flow prior to the `chooseOperation` authorization fix, so it is not a regression from that change. Deferred because enforcing gate 1 for SQL changes authorization outcomes for existing roles and needs its own change + release note.
- Documented in `DESIGN.md` (the `chooseOperation` section).

### Fix direction

Run gate 1 (`operations` allowlist + `requires_su`) for `sql` operations in addition to the AST table check, so the two checks are additive rather than exclusive.

Issue #2215, opened 2026-08-19
export_local/export_to_s3 via `operations` allowlist bypasses table-level grants (non-SQL nested search)


cb1kenobi:
### Summary

A role granted `export_local` / `export_to_s3` through its `operations` allowlist can export tables it holds no table-level grant on, when the export uses a **non-SQL** nested search (`search_by_conditions` / `search_by_value` / `search_by_hash`).

### Detail

`export_local` and `export_to_s3` are `requires_su` operations. In `verifyPerms` gate 2 (`utility/operation_authorization.ts:648-650`), an SU-only operation that is explicitly listed in a role's `operations` is treated as a deliberate grant and returns `null` (allow) **before any table-level check**. The additive AST permission check in `chooseOperation` (`server/serverHelpers/serverUtilities.ts:294`) only runs when the nested search `operation === 'sql'`, so a non-SQL nested search reaches the export handler with no table authorization at all.

Result: `export_local` in `operations` ⇒ read/export any table via a non-SQL nested search.

### Scope

- Pre-existing behavior; not introduced by the `chooseOperation` authorization fix. That change deliberately leaves this open because closing it changes authorization outcomes for existing role configurations, so it needs its own change + release note.
- Current behavior is documented in `DESIGN.md` (the `chooseOperation` section) and pinned by the `NESTED-NOSQL` case in `integrationTests/security/choose-operation-authz.test.ts`, so the fix has a test to flip.

### Fix direction

Fall through to a table-level CRUD check for `requires_su` ops granted via `operations` (rather than returning `null` unconditionally at gate 2), or run the export's table check for non-SQL nested searches too. Coordinate with the gate-2 TODO already noted at `utility/operation_authorization.ts:642-647`.

PR #2217, opened 2026-08-19
fix(security): authorize the invoked operation against the authenticated principal

cb1kenobi:
Fixes operation authorization in `chooseOperation`: authorize the invoked operation against the authenticated principal, and stop a caller-supplied `search_operation` from standing in as the permission subject.

## What this fixes

`verifyPerms` reads *both* halves of the permission question off the object it's handed — the principal from `hdb_user` and the tables from `schema`/`database`/`table`/`records`. `chooseOperation` used to hand it `json.search_operation`, a caller-supplied field, making both halves body-controlled. Four issues, each with regression cover in `integrationTests/security/choose-operation-authz.test.ts`:

1. **Cross-privilege redirect** — `verifyPerms` ran against `json.search_operation ?? json` for *every* operation, but the handler runs against top-level `json` and only `dataLayer/export.ts` consumes `search_operation`. A non-super user could send a privileged top-level operation with a benign `search_operation` and get authorized against the benign tables. Now `search_operation` is the permission subject only for `export_local`/`export_to_s3`; every other op is checked against top-level `json`.
2. **Principal smuggling** — the nested principal was backfilled (`if (!hdb_user)`), honoring a body-supplied one. Now `hdb_user` is overwritten from the authenticated top-level principal unconditionally.
3. **`parsed_sql_object` smuggling** — the export worker re-reads `parsed_sql_object` (carrying `permissions_checked`) off the nested object, so a body-supplied one would execute an AST no check ever saw. It's now deleted at dispatch, forcing a re-parse + re-check in the worker.
4. **Silent SQL denial** — `processAST`'s guard tested `permissionsCheck.length > 0`, but a denial is a `PermissionResponseObject` with no `.length`, so `undefined > 0` was always false and denials executed. Fixed to `if (permissionsCheck)`.

Non-object `search_operation` on an export op is now a 400 (was a wrapped 500), and the SQL AST check runs additively after `verifyPerms` rather than as an exclusive branch.

## For the reviewer

- **`server/serverHelpers/serverUtilities.ts` (`chooseOperation`)** is the core. The load-bearing invariant: the object passed to `verifyPerms` must be the one the handler actually operates on. `DESIGN.md` (new `chooseOperation` section) documents the three rules and why each is required.
- **`export_local` table check is currently unreachable** — both export ops are `requires_su` and every path returns before the nested table check (super_user early-returns; a role granted the op via `operations` returns at gate 2; anything else is refused earlier). So the nested-table substitution is inert today. This is intentional for this patch.

### Behavior change (release note)

**Non-super users can no longer invoke `export_local` / `export_to_s3` via a SQL `search_operation`.** These operations are `requires_su`, and NoSQL export already enforced that (`"Operation 'export_local' is restricted to 'super_user' roles"`). SQL-based export previously took the SQL-only branch and skipped `verifyPerms`, so a non-super user could invoke the privileged export by wrapping it in a SQL search — governed only by table read perms. Routing export through `verifyPerms` closes that bypass and brings SQL export to parity with NoSQL export. Any non-super role that relied on SQL-export-without-`export_local` will now receive a 403. Three `northwind` export tests that encoded the old bypass were updated to assert the `requires_su` denial (they now mirror the existing NoSQL export case; they could be consolidated).

### Deferred gaps (tracked, not fixed here)

Two pre-existing authorization gaps are left open deliberately — closing either changes authorization outcomes for existing role configurations, so each needs its own change + release note. Both are documented in `DESIGN.md` and pinned by regression tests so there's a test to flip:

- HarperFast/harper#2215 — a role granted `export_local`/`export_to_s3` via `operations` can export tables it holds no grant on (non-SQL nested search reaches the handler with no table check).
- HarperFast/harper#2216 — a direct `sql` operation bypasses the `operations` allowlist (gate 1); authorized only by the AST table check.

## Verification

- `integrationTests/security/choose-operation-authz.test.ts` — pins the denial shapes and the current boundaries (principal override, nested-shape 400, forged-AST rejection, and the two deferred-gap boundaries `NESTED-NOSQL` / gate-1).
- **Deep-review, multipass (2 independent runs × auth + SQL/AST lenses).** Confirmed the fix closes the cross-privilege redirect, principal smuggling, AST smuggling, and the `processAST` denial-drop, with no other `.length`-style denial-drop anywhere in the tree. The multipass independently rediscovered the two deferred gaps above (now ticketed). One low cleanup nit surfaced: `serverUtilities.ts:298` sets a top-level `parsed_sql_object` that's inert for the export path.

## Review coverage

- **Deep-review, multipass** (Claude — Harper-domain + SQL/AST lenses, 2 independent runs): confirmed the fix closes the cross-privilege redirect, principal smuggling, AST smuggling, and the `processAST` denial-drop; no other `.length`-style denial-drop in the tree.
- **Cross-model, Codex** (graded, opposite-family): verdict COMMENTS, no production blockers. Two findings, both in the test file (below).
- Degraded legs (noted for honesty): Gemini `agy` failed (stochastic print-mode hang; CLI retries + 2 manual attempts all timed out) — no perf/maintainability lens this run; `cursor-composer` not authenticated locally; the CLI's Claude domain-adjudicator leg couldn't run (`claude` not on PATH) — covered instead by the separate multipass deep-review above.

### Unresolved review findings (test-only, non-blocking)

- **Codex, minor** — `integrationTests/security/choose-operation-authz.test.ts`: no end-to-end case proves the **export worker's nested-SQL recheck can deny an unauthorized query**. The allowed super-user path reaching `COMPLETE` is covered, and `unitTests/sqlTranslator/processAST.test.js` proves the object-shaped denial stops routing, but the full dispatch → principal-propagation → reparse → **denial** path isn't exercised together. Worth adding a non-super-user unauthorized-export-SQL case that asserts denial.
- **Codex, nit** — same file: trim added comments that narrate tests / restate identifiers / address the reviewer (the suite catalog + run-command block at the top, and identifier restatements).


server/serverHelpers/serverUtilities.ts
```diff
@@ -227,41 +227,51 @@ export function chooseOperation(json: OperationRequestBody, bypassAuth = false)
...
 			const functionToCheck = job_operation_function === undefined ? operation_function : job_operation_function;
-			const operation_json = json.search_operation ? json.search_operation : json;
-			if (!operation_json.hdb_user) {
-				operation_json.hdb_user = json.hdb_user;
-			}
+			const operation_json = nestedSearch ?? json;
+			// Authentication is the only source of the principal; a nested one is overwritten, not honored.
+			operation_json.hdb_user = json.hdb_user;
 
 			const verifyPermsResult = opAuth.verifyPerms(operation_json, functionToCheck);
```
kriszyp: Please rebase and explicitly reconcile this authorization block before merging. `origin/main` now changes the same code to exempt `exchange_oidc_token` from a check that requires `hdb_user`, pass the top-level API operation into `verifyPerms` for token-scope enforcement, and carry that operation through SQL permission checks. This head omits those changes and currently has merge conflicts here and in `sqlTranslator/index.ts`. Choosing the PR side during resolution would either break OIDC exchange or regress export/SQL token scopes; preserve those main-side behaviors while applying this PR’s principal/subject fix and truthy SQL-denial guard.

server/serverHelpers/serverUtilities.ts
```diff
@@ -227,41 +227,51 @@ export function chooseOperation(json: OperationRequestBody, bypassAuth = false)
...
-						true
-					);
-				}
-			}
-			//we need to bypass permission checks to allow the createAuthorizationTokens
-		} else if (
+		//we need to bypass permission checks to allow the createAuthorizationTokens
+		if (
 			!bypassAuth &&
+			!isSqlOperation &&
```
kriszyp: Keeping direct SQL outside `verifyPerms` still lets a role whose `operations` allowlist excludes `sql` invoke SQL whenever its table CRUD permissions allow the statement. The AST check validates tables and attributes, not permission to invoke the API operation, so this leaves the stated “invoked operation” invariant incomplete. Add a separate outer-operation allowlist check for `sql`, keep it additive with the AST check, and add a negative test using a role with table read permission but only an unrelated allowed operation.

server/serverHelpers/serverUtilities.ts
```diff
@@ -227,41 +227,51 @@ export function chooseOperation(json: OperationRequestBody, bypassAuth = false)
...
 			json.operation !== terms.OPERATIONS_ENUM.CREATE_AUTHENTICATION_TOKENS &&
 			json.operation !== terms.OPERATIONS_ENUM.LOGIN &&
 			json.operation !== terms.OPERATIONS_ENUM.LOGOUT
 		) {
 			const functionToCheck = job_operation_function === undefined ? operation_function : job_operation_function;
-			const operation_json = json.search_operation ? json.search_operation : json;
-			if (!operation_json.hdb_user) {
-				operation_json.hdb_user = json.hdb_user;
-			}
+			const operation_json = nestedSearch ?? json;
```
kriszyp, line 403: This single `verifyPerms` call cannot authorize both the outer export and its nested query. When a non-super role explicitly lists `export_local`, the SU-operation gate returns success before examining this nested table; the job worker then calls `searchByValue`/`searchByHash` directly without another permission check. Consequently, the `NESTED-NOSQL` test demonstrates that a role with no table grant can export the table. Please authorize outer operation invocation first, then authorize the nested search with its actual search handler (or equivalent read-permission check), and flip that characterization test to require denial/completed non-execution.

server/serverHelpers/serverUtilities.ts
```diff
@@ -227,41 +227,51 @@ export function chooseOperation(json: OperationRequestBody, bypassAuth = false)
...
+	const isSqlOperation = json.operation === terms.OPERATIONS_ENUM.SQL;
+	// `verifyPerms` derives the tables it checks from the object it is handed, so the caller-supplied
+	// `search_operation` may only stand in as the permission subject for the operations that actually
+	// run it (dataLayer/export.ts is its only consumer). For anything else it would check the nested
+	// tables while the handler runs against the top-level ones.
+	const nestedSearch =
+		json.operation === terms.OPERATIONS_ENUM.EXPORT_LOCAL || json.operation === terms.OPERATIONS_ENUM.EXPORT_TO_S3
+			? json.search_operation
+			: undefined;
+	if (nestedSearch !== undefined) {
```
kriszyp, line 278: The new shape check excludes `undefined`, while `{}` also passes it. A super-user export with no `search_operation` therefore returns a successful job submission before the worker dereferences `search_operation.operation`; an empty object similarly becomes an asynchronously failed job rather than a request-time 400. Validate that exports have an object containing a supported `operation` before enqueueing—preferably in one operation schema/validator—and cover both omitted and empty values.

dawsontoth:
We'll have some overlap with https://github.com/HarperFast/harper/pull/2202 on your point 4 -- more details to come shortly!

kriszyp (changes requested):
Nice, glad to be reviewing the auth on SQL, looks like some good improvements. It looks like there are a few things to address (and maybe needs a rebase).

server/serverHelpers/serverUtilities.ts
cb1kenobi: Done — rebased onto `main` and reconciled this block toward main's structure. The head now keeps main's landed SQL branch (allowlist + `apiOperation`-threaded `checkASTPermissions`, from #2260 / #2176) and the `processAST` denial fix (#2202), and layers this PR's principal/subject hardening on top. The main-side behaviors you flagged are preserved rather than overwritten: `exchange_oidc_token` stays exempt from the `hdb_user` check, and the top-level operation is threaded into `verifyPerms`/`checkASTPermissions` for token-scope enforcement on export and SQL. Conflicts in `sqlTranslator/index.ts`, `processAST.test.js`, and `DESIGN.md` were resolved the same way.

<sub>🤖 Generated by Barber AI (Opus 4.8)</sub>

server/serverHelpers/serverUtilities.ts
cb1kenobi: Addressed. main's SQL branch (#2260) now enforces the `operations` allowlist for the invoked op — `verifyOperationsAllowlist(json, json.operation)` — as an additive gate ahead of the AST check, so a role whose allowlist excludes `sql` can no longer reach SQL through table CRUD alone. Direct SQL keeps that allowlist check; a nested-SQL export now routes through `verifyPerms` instead, so the outer op's allowlist **and** its `requires_su` gate both apply. Negative coverage is in the new `choose-operation-authz.test.ts` plus the updated northwind SQL-export cases.

<sub>🤖 Generated by Barber AI (Opus 4.8)</sub>

server/serverHelpers/serverUtilities.ts
cb1kenobi: Closed the gap. `chooseOperation` now authorizes the outer export invocation first, then authorizes the nested search **additively** against its real handler (`getOperationFunction(search_operation).operation_function`) and the authenticated principal — so a role granted `export_local` but not the underlying read is denied at dispatch, before the worker's unchecked `searchByValue`/`Hash`/`Conditions` ever runs. The `NESTED-NOSQL` characterization test is flipped from a documented-gap 200 to a 403 denial, and a new `POSITIVE-NOSQL` proves a permitted non-SQL export still completes.

One nuance worth flagging: the nested check runs the search handler through `verifyPerms`, so it also re-applies gate 1 — a role with a restrictive `operations` allowlist must permit the search op as well as `export_local`. That's fail-closed and consistent with the denial we want here, but it is slightly stricter than the SQL-export path (whose inner tables aren't allowlist-gated). Happy to converge them on a pure table-read check in a follow-up if you'd prefer.

<sub>🤖 Generated by Barber AI (Opus 4.8)</sub>

server/serverHelpers/serverUtilities.ts
cb1kenobi: Fixed. A `search_operation` on an export must now be an object **naming a supported export operation** (`search_by_value` / `search_by_hash` / `search_by_conditions` / `sql`); `{}`, a missing/empty `operation`, or an unsupported op is a request-time 400 rather than an async job failure or an `undefined` dereference in the worker. Validated at dispatch, before authorization, so an un-granted role still gets the 400. Covered by `NESTED-OP` (empty/invalid operations) and `NESTED-SHAPE` (non-objects).

<sub>🤖 Generated by Barber AI (Opus 4.8)</sub>

cb1kenobi:
Rebased onto `main` and worked through the review. Since this branch was opened, `main` independently landed several overlapping pieces (`#2202` SQL denial fix, `#2176`/`#2260` scoped tokens + operations allowlist), so the reconciliation keeps main's structure and layers this PR's hardening on top rather than re-introducing the older shape.

**What changed in the reconciliation**
- **Principal from authentication** — a nested `hdb_user` is overwritten, never backfilled.
- **`search_operation` as subject only for exports** — gated to `export_local`/`export_to_s3`, must be an object, and must name a supported export operation (`search_by_value`/`search_by_hash`/`search_by_conditions`/`sql`); anything else (including `{}`) is a request-time 400.
- **Outer + nested authorized separately** — the export invocation is authorized first, then the nested search is authorized additively against its real handler and the authenticated principal, so a role granted `export_local` but not the underlying read is denied. `NESTED-NOSQL` is flipped to a denial; `POSITIVE-NOSQL` covers the permitted case.
- **SQL export routed through `verifyPerms`** — so its `requires_su` gate applies exactly as on the non-SQL path (main's SQL branch checked only allowlist + AST, so a non-super user could export via a SQL `search_operation`). The northwind SQL-export cases now assert the `requires_su` denial.
- Gemini's suggestion (only assign `json.parsed_sql_object` on the direct-SQL path) is also implemented.

**Testing** — `npm run build` clean; the authz/SQL unit suites pass (`operation_authorization`, `tokenOperationScope`, `processAST`, `serverUtilities` = 200+ specs green). I could not run the integration suites locally (this machine lacks the loopback alias pool), so `choose-operation-authz.test.ts` and the northwind changes will get their first real run in CI — I did validate the two load-bearing behaviors (SQL export → `requires_su` denial for a non-super user; `export_local`-granted role without read → nested denial) directly against the built `verifyPerms`.

One open question flagged inline on the nested-read thread: the nested check re-applies gate 1, making non-SQL export slightly stricter than SQL export for roles with a restrictive `operations` allowlist. Fail-closed, but happy to converge them if you'd prefer.

<sub>🤖 Generated by Barber AI (Opus 4.8)</sub>

server/serverHelpers/serverUtilities.ts
```diff
@@ -255,32 +266,102 @@ export function chooseOperation(json: OperationRequestBody, bypassAuth = false)
...
```
kriszyp: The `{}` and unsupported-operation cases from the prior finding are fixed, but an omitted `search_operation` still skips this entire validation block. A super-user export without that property therefore reaches `executeJob`, returns a successful job submission, and only fails in the worker when `exportCoreValidation` dereferences `exportObject.search_operation.operation`. Require `search_operation` for `export_local`/`export_to_s3` before this conditional (or validate the complete export shape in one schema), and add the omitted-property case alongside `NESTED-OP` so malformed exports consistently return a request-time 400.

integrationTests/security/choose-operation-authz.test.ts
```diff
@@ -0,0 +1,444 @@
...
+
+		test('DISPATCH-SU — a SQL-shaped search_operation does not route export_local past requires_su', async () => {
+			const r = await client.reqAs(malloryHeaders).send({
+				operation: 'export_local',
+				path: '/tmp',
+				format: 'json',
+				search_operation: { operation: 'sql', sql: 'select 1' },
+			});
+			assertForbidden(r, 'SQL-shaped search_operation on a requires_su operation');
+		});
```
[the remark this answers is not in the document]
cb1kenobi: Good catch — the diagnosis is correct. With Mallory's `operations: ['user_info','sql']` allowlist, gate 1 (`verifyOperationsAllowlist(json, json.operation)`) denies both `add_user` and `export_local` in the pre-fix path too, so neither test could fail if the dispatch-scoping regressed.

Fixed in 8f56329 — **DISPATCH** now uses TABLER attempting `add_user`. TABLER declares no `operations`, so gate 1 is a no-op and the only barrier is that a nested `operation: 'sql'` must not route the outer op past its `verifyPerms`/`requires_su` check. Reverting the scoping fix creates the `super_user` account, so the test fails closed.

One deviation on **DISPATCH-SU**: I used TABLER there too, not EXPORTER. EXPORTER lists `export_local`, which `verifyPerms` gate 2 grants (an allowlisted `requires_su` op is a deliberate admin grant), so EXPORTER exporting `select 1` is *legitimately* allowed (200) — asserting 403 against it would fail the fixed code. TABLER has no allowlist, so its only barrier to a `requires_su` export is the `requires_su` gate inside `verifyPerms`, which is exactly what the nested-SQL trick must not route around: fixed = 403, reverted = the export goes through.

(EXPORTER still earns its keep in NESTED-NOSQL, where the point is that a role allowed to *invoke* the export must still not read a table it has no grant on.)

*— Generated by Claude (Opus 4.8)*

server/serverHelpers/serverUtilities.ts
cb1kenobi: Addressed. `chooseOperation` now authorizes the two decisions separately: the outer op invocation via `verifyPerms(json, functionToCheck, {apiOperation})`, then — for a non-SQL export — the nested search against its **real handler** and the authenticated principal (`verifyPerms(nestedSearch, getOperationFunction(nestedSearch).operation_function, {apiOperation})`). Since the export worker runs `searchByValue`/`Hash`/`Conditions` directly with no permission check, dispatch is the only place the nested read is enforceable. The `NESTED-NOSQL` characterization test is flipped to assert **403** (a role granted `export_local` but lacking read on the table is now denied), and a `POSITIVE-NOSQL` case proves a super_user `search_by_value` export still completes.

*— Generated by Claude (Opus 4.8)*

server/serverHelpers/serverUtilities.ts
cb1kenobi: Addressed. The shape check now rejects `{}` and any unsupported nested operation at request time: after the object check, `search_operation.operation` must be a string in `EXPORT_SEARCH_OPERATIONS` (mirroring export.ts's `VALID_SEARCH_OPERATIONS`: `search_by_value`/`search_by_hash`/`search_by_conditions`/`sql`), otherwise it's a 400 rather than an async job failure. New `NESTED-OP` test covers the empty-object and bogus-operation cases; `NESTED-SHAPE` still covers the non-object cases.

*— Generated by Claude (Opus 4.8)*

server/serverHelpers/serverUtilities.ts
cb1kenobi: This is now covered: the direct-SQL path runs `verifyOperationsAllowlist(json, json.operation)` before the AST check, so a role whose `operations` allowlist excludes `sql` is denied at the invocation level regardless of its table CRUD grants — additive to the AST table/attribute check, as you suggested. (This landed on `main` via #2260 and is preserved through the rebase.) `POSITIVE-SQL` confirms a granted direct SQL call still works.

*— Generated by Claude (Opus 4.8)*

server/serverHelpers/serverUtilities.ts
```diff
@@ -255,32 +266,102 @@ export function chooseOperation(json: OperationRequestBody, bypassAuth = false)
...
+							allowlistDenial,
+							hdbErrors.HTTP_STATUS_CODES.FORBIDDEN,
+							undefined,
+							undefined,
+							true
+						);
+					}
 				}
 				// `json.operation` explicitly — the operation this dispatch already resolved, not a
 				// field read back off the request body.
```
kriszyp: Nested SQL is still authorized as the outer export plus AST table permissions, without applying the inner SQL/statement operation allowlist. For an ordinary authenticated role with `operations: ['export_local']` and `delete: true` on a table—but no `sql` or `delete` operation grant—`search_operation: { operation: 'sql', sql: 'DELETE ...' }` passes the outer gate and AST check, then performs the deletion in the job before export output handling. The non-SQL path at `server/serverHelpers/serverUtilities.ts:427` does apply its inner operation gate, so SQL is also less restrictive than the other nested searches. Please either restrict export SQL to SELECT, which best matches `dataLayer/export.ts:327`’s `getRecords` contract, or add the corresponding inner SQL/write-operation gate. A regression should submit a nested DELETE through an export-only role and verify the row remains.

<sub>(`server/serverHelpers/serverUtilities.ts:368` is not part of this PR's diff — anchored to the nearest line this PR's diff can hold)</sub>

kriszyp: Nested SQL is still authorized as the outer export plus AST table permissions, without applying the inner SQL/statement operation allowlist. For an ordinary authenticated role with `operations: ['export_local']` and `delete: true` on a table—but no `sql` or `delete` operation grant—`search_operation: { operation: 'sql', sql: 'DELETE ...' }` passes the outer gate and AST check, then performs the deletion in the job before export output handling. The non-SQL path at `server/serverHelpers/serverUtilities.ts:427` does apply its inner operation gate, so SQL is also less restrictive than the other nested searches. Please either restrict export SQL to SELECT, which best matches `dataLayer/export.ts:327`’s `getRecords` contract, or add the corresponding inner SQL/write-operation gate. A regression should submit a nested DELETE through an export-only role and verify the row remains.

<sub>(`server/serverHelpers/serverUtilities.ts:368` is not part of this PR's diff — anchored to the nearest line this PR's diff can hold)</sub>

kriszyp, line 377: Nested SQL is still authorized as the outer export plus AST table permissions, without applying the inner SQL/statement operation allowlist. For an ordinary authenticated role with `operations: ['export_local']` and `delete: true` on a table—but no `sql` or `delete` operation grant—`search_operation: { operation: 'sql', sql: 'DELETE ...' }` passes the outer gate and AST check, then performs the deletion in the job before export output handling. The non-SQL path at `server/serverHelpers/serverUtilities.ts:427` does apply its inner operation gate, so SQL is also less restrictive than the other nested searches. Please either restrict export SQL to SELECT, which best matches `dataLayer/export.ts:327`’s `getRecords` contract, or add the corresponding inner SQL/write-operation gate. A regression should submit a nested DELETE through an export-only role and verify the row remains.

<sub>(`server/serverHelpers/serverUtilities.ts:368` is not part of this PR's diff — anchored to the nearest line this PR's diff can hold)</sub>

kriszyp (commented):
I think one more thing to check on SQL writes/DML, but this looks good.

kriszyp (approved):
Looks good.

server/serverHelpers/serverUtilities.ts
cb1kenobi: Good catch, and thanks for spelling out the concrete role shape — that made it easy to reproduce.

I took the first option you offered: nested export SQL is now restricted to `SELECT`, which matches `dataLayer/export.ts`'s `getRecords` contract rather than layering a second write-operation gate onto a path that should only ever read.

**Fix** — `server/serverHelpers/serverUtilities.ts:321`: after the dispatch's own parse, a nested `search_operation.sql` whose variant isn't `select` is a request-time 400, before the outer `verifyPerms` runs. A nested `DELETE`/`UPDATE`/`INSERT` never reaches the job worker, so there is no window where the statement executes and only the export output is withheld.

**Regression cover** — both shapes you asked for:

- `integrationTests/security/choose-operation-authz.test.ts` — `NESTED-SQL-DML`: the export-only role (`operations: ['export_local', ...]`, `delete: true` on the table, no `read`, no `sql`) submits `search_operation: { operation: 'sql', sql: "DELETE FROM data.AuthzProbe WHERE id = '...'" }`, then asserts the row is still present and the request was rejected with a 400.
- `unitTests/server/serverHelpers/serverUtilities.test.js` — "rejects write SQL nested in an export job" pins the dispatch-level rejection and its message.

**One thing worth naming explicitly:** this closes the write path, but not the inner `sql` operation-allowlist question — a role holding `export_local` without a `sql` grant can still run a nested `SELECT`, authorized by the AST table check against its read permissions. Given export's contract is a read and the outer op is `requires_su`-gated, that reads as intended to me, but I'm happy to add the inner allowlist gate as well if you'd rather the nested search match the non-SQL path grant-for-grant.

The `DESIGN.md` `chooseOperation` section now states the SELECT rule alongside the other invariants. Branch is also rebased on current `main`.

<sub>Generated by Claude Opus 5</sub>

PR #2219, opened 2026-08-19
fix(run): stop double-wrapping startup URLs when node.hostname is a URL

### LINKED ISSUE #2218: URLs can log with multiple schema wrappings
### Bug Summary

http://http://localhost:9926:9925/

_No response_

### Steps to Reproduce

```
Hostname:           http://localhost:9926
Worker Threads:     1
Root Path:          /Users/dawson/harper
Debugging:          enabled: true
Logging:            level: info, location: /Users/dawson/harper/log/hdb.log, stdout/err
Default:            HTTP (and WS): 9926, CORS: enabled for *
Operations API:     HTTP: 9925, CORS: enabled for *, unix socket: /Users/dawson/harper/operations-server
                    http://http://localhost:9926:9925/
MQTT:               TCP: 1883, TLS: 8883, WS: 9926
REST:               HTTP: 9926, WS: 9926
                    http://http://localhost:9926:9926/
```

### Expected Behavior

http://localhost:9926/

### Actual Behavior

http://http://localhost:9926:9925/

### Platform

_No response_

### Console Errors

_No response_

### Screenshots or Videos

_No response_

### Additional Context

_No response_

No, just reporting the issue

dawsontoth:
Fixes #2218. When `node.hostname` is configured as a full URL (e.g. `http://localhost:9926`), the Operations-API and REST lines of the startup banner composed URLs like `http://http://localhost:9926:9925/` — the scheme was wrapped twice and the wrong port appended, because the banner assumed `getThisNodeName()` returns a bare host. A new `nodeNameToDisplayHost()` reduces the node name to a bare host (stripping any scheme/port, and bracketing a bare IPv6 literal) before each banner URL is composed; the three banner URL sites now use it. Result: `http://localhost:9925/` (ops) and `http://localhost:9926/` (REST).

## For the human reviewer

1. **Scope — normalize at the banner, not in `getThisNodeName()`.** Chosen because `getThisNodeName()` is the node's identity (cert CN, replication), so rewriting it is risky, while the reported bug is display-only. The alternative — normalizing at the source — would also fix `hostnameToUrl()`, which shares the same bare-host assumption and would produce a corrupt replication URL (`ws://http://host:port`) for a URL-valued `node.hostname`. That corruption is pre-existing and left out of scope here; the root-cause follow-up (validate/warn the bare-host `node.hostname` invariant in `getThisNodeName()`, which also covers `hostnameToUrl()`/replication and the `security/keys.ts` cert identity) is being handled as a separate change already in progress. Reversible either way.
2. **`Hostname:` banner line left verbatim.** It still prints the raw configured `node.hostname` (the URL), deliberately — so an operator who configured a URL where a bare host is expected sees that, rather than having it silently masked. Only the composed URLs are normalized.

## Verification

Route: unit tests (plain `assert`, no stubbing) at two levels — the pure `nodeNameToDisplayHost` helper across bare host / full URL / `host:port` / bracketed and bare IPv6, and an end-to-end test that drives the real exported `startupLog()` with a URL-valued `node.hostname` and asserts the composed Operations-API HTTP/HTTPS and REST banner lines are well-formed (no `http://http://` double-wrap). Executed locally: `unitTests/server/nodeName.test.js` 15 passing and `unitTests/bin/startupBanner.test.js` 3 passing (18 together, no cross-suite leakage); `npm run lint:required` and `prettier --check` clean. The e2e test also fails as expected when the banner sites are reverted to `getThisNodeName()`, proving it catches the regression.

## Review coverage

Authored by Opus 4.8. Cross-model review ran 6 delta rounds (one after each round of fixes). Codex `gpt-5.6-sol` gave independent, graded outside-family coverage every round; Gemini via `agy` (default model) ran on some rounds and failed others (`agy` flakiness); the Harper domain adjudicator and Cursor legs were auto-pruned as narrow low-risk deltas, so Codex's findings were not domain-adjudicated. Every substantive finding was fixed and re-verified: new Sinon usage (→ pure function + plain `assert`), a bare-IPv6 URL gap, missing `getThisNodeHostname` wrapper coverage, missing end-to-end `startupLog` coverage, and a test-isolation cleanup gap. One nit is left open by choice — a two-line "why" comment on `nodeNameToDisplayHost`, kept as durable rationale the house style permits. Receipt @ `d7b5c46bb`.

<sub>Review-Coverage: authored=claude; ran=codex; declined=gemini,cursor-grok,cursor-composer,domain; rounds=1 @ d7b5c46bb96f</sub>

<sub>Human-Review-Need: 4 @ d7b5c46bb96f</sub>

### DIFF
--- a/bin/run.ts
+++ b/bin/run.ts
@@ -25,7 +25,7 @@ import * as keys from '../security/keys.ts';
 import { startHTTPThreads } from '../server/threads/socketRouter.ts';
 import * as hdbInfoController from '../dataLayer/hdbInfoController.ts';
 import { isReadOnlyMode } from '../resources/databases.ts';
-import { getThisNodeName } from '../server/nodeName.ts';
+import { getThisNodeName, getThisNodeHostname } from '../server/nodeName.ts';
 import * as hdbTerms from '../utility/hdbTerms.ts';
 import { getHdbPid, isProcessRunning } from '../utility/processManagement/processManagement.js';
 import { PACKAGE_ROOT } from '../utility/packageUtils.js';
@@ -321,14 +321,20 @@ function startupLog(portResolutions: any) {
 	}`;
 	logMsg += `, unix socket: ${configUtils.getConfigPath(CONFIG_PARAMS.OPERATIONSAPI_NETWORK_DOMAINSOCKET)}\n`;
 	if (env.get(CONFIG_PARAMS.OPERATIONSAPI_NETWORK_PORT)) {
-		logMsg += pad('') + 'http://' + getThisNodeName() + ':' + env.get(CONFIG_PARAMS.OPERATIONSAPI_NETWORK_PORT) + '/\n';
+		logMsg +=
+			pad('') +
+			'http://' +
+			getThisNodeHostname() +
+			':' +
+			env.get(CONFIG_PARAMS.OPERATIONSAPI_NETWORK_PORT) +
+			'/\n';
 	}
 	if (env.get(CONFIG_PARAMS.OPERATIONSAPI_NETWORK_SECUREPORT)) {
 		logMsg +=
 			'\n' +
 			pad('') +
 			'https://' +
-			getThisNodeName() +
+			getThisNodeHostname() +
 			':' +
 			env.get(CONFIG_PARAMS.OPERATIONSAPI_NETWORK_SECUREPORT) +
 			'/\n';
@@ -383,7 +389,7 @@ function startupLog(portResolutions: any) {
 			if (!restLog.includes(pair) && name === 'rest') {
 				restLog += pair;
 				if (value.protocol_name === 'HTTP' || value.protocol_name === 'HTTPS') {
-					restHostnames.push(`${value.protocol_name.toLowerCase()}://${getThisNodeName()}:${key}/`);
+					restHostnames.push(`${value.protocol_name.toLowerCase()}://${getThisNodeHostname()}:${key}/`);
 				}
 			}
 
--- a/bin/run.ts
+++ b/bin/run.ts
@@ -321,23 +321,10 @@ function startupLog(portResolutions: any) {
 	}`;
 	logMsg += `, unix socket: ${configUtils.getConfigPath(CONFIG_PARAMS.OPERATIONSAPI_NETWORK_DOMAINSOCKET)}\n`;
 	if (env.get(CONFIG_PARAMS.OPERATIONSAPI_NETWORK_PORT)) {
-		logMsg +=
-			pad('') +
-			'http://' +
-			getThisNodeHostname() +
-			':' +
-			env.get(CONFIG_PARAMS.OPERATIONSAPI_NETWORK_PORT) +
-			'/\n';
+		logMsg += `${pad('')}http://${getThisNodeHostname()}:${env.get(CONFIG_PARAMS.OPERATIONSAPI_NETWORK_PORT)}/\n`;
 	}
 	if (env.get(CONFIG_PARAMS.OPERATIONSAPI_NETWORK_SECUREPORT)) {
-		logMsg +=
-			'\n' +
-			pad('') +
-			'https://' +
-			getThisNodeHostname() +
-			':' +
-			env.get(CONFIG_PARAMS.OPERATIONSAPI_NETWORK_SECUREPORT) +
-			'/\n';
+		logMsg += `\n${pad('')}https://${getThisNodeHostname()}:${env.get(CONFIG_PARAMS.OPERATIONSAPI_NETWORK_SECUREPORT)}/\n`;
 	}
 
 	// MQTT Log
--- a/server/nodeName.ts
+++ b/server/nodeName.ts
@@ -58,6 +58,24 @@ export function clearThisNodeName() {
 	nodeName = undefined;
 }
 
+/**
+ * Returns this node's bare hostname (no scheme, no port) for composing display URLs.
+ * getThisNodeName() is normally a bare host ('localhost'), but if node.hostname is
+ * configured as a full URL ('http://localhost:9926'), composing a URL from it double-wraps
+ * the scheme and inherits the wrong port (http://http://localhost:9926:9925/). Reduce it
+ * back to the host so the composed URL is well-formed. (#2218)
+ */
+export function getThisNodeHostname(): string {
+	const name = getThisNodeName();
+	if (!name) return name;
+	try {
+		const hasScheme = /^[a-z][a-z0-9+.-]*:\/\//i.test(name);
+		return new URL(hasScheme ? name : `http://${name}`).hostname || name;
+	} catch {
+		return name; // not URL-parseable; assume it is already a bare host
+	}
+}
+
 function getHostFromListeningPort(key: string) {
 	const port: string | undefined = env.get(key);
 	const lastColon = port?.lastIndexOf?.(':');
--- a/server/nodeName.ts
+++ b/server/nodeName.ts
@@ -58,9 +58,8 @@ export function clearThisNodeName() {
 	nodeName = undefined;
 }
 
-// Reduce node.hostname to a bare host for display-URL composition: it may be configured as a
-// full URL (which would otherwise double-wrap the scheme/port), and the bracketed retry lets a
-// bare IPv6 literal parse.
+// node.hostname may be configured as a full URL; reduce it to a bare host so composing a
+// display URL from it does not double-wrap the scheme and port.
 export function nodeNameToDisplayHost(name: string): string {
 	if (!name) return name;
 	const hasScheme = /^[a-z][a-z0-9+.-]*:\/\//i.test(name);
--- a/server/nodeName.ts
+++ b/server/nodeName.ts
@@ -58,22 +58,27 @@ export function clearThisNodeName() {
 	nodeName = undefined;
 }
 
-/**
- * Returns this node's bare hostname (no scheme, no port) for composing display URLs.
- * getThisNodeName() is normally a bare host ('localhost'), but if node.hostname is
- * configured as a full URL ('http://localhost:9926'), composing a URL from it double-wraps
- * the scheme and inherits the wrong port (http://http://localhost:9926:9925/). Reduce it
- * back to the host so the composed URL is well-formed. (#2218)
- */
-export function getThisNodeHostname(): string {
-	const name = getThisNodeName();
+// node.hostname is normally a bare host, but may be configured as a full URL; composing a
+// display URL from that form would double-wrap its scheme and port. Reduce it to a bare host
+// (bracketed, for an IPv6 literal) first. (#2218)
+export function nodeNameToDisplayHost(name: string): string {
 	if (!name) return name;
-	try {
-		const hasScheme = /^[a-z][a-z0-9+.-]*:\/\//i.test(name);
-		return new URL(hasScheme ? name : `http://${name}`).hostname || name;
-	} catch {
-		return name; // not URL-parseable; assume it is already a bare host
+	const hasScheme = /^[a-z][a-z0-9+.-]*:\/\//i.test(name);
+	// A bare IPv6 literal ('::1') only parses once bracketed, so try that form too.
+	const candidateUrls = hasScheme ? [name] : [`http://${name}`, `http://[${name}]`];
+	for (const candidateUrl of candidateUrls) {
+		try {
+			const { hostname } = new URL(candidateUrl);
+			if (hostname) return hostname;
+		} catch {
+			// not this form; try the next candidate
+		}
 	}
+	return name;
+}
+
+export function getThisNodeHostname(): string {
+	return nodeNameToDisplayHost(getThisNodeName());
 }
 
 function getHostFromListeningPort(key: string) {
--- a/server/nodeName.ts
+++ b/server/nodeName.ts
@@ -58,21 +58,18 @@ export function clearThisNodeName() {
 	nodeName = undefined;
 }
 
-// node.hostname is normally a bare host, but may be configured as a full URL; composing a
-// display URL from that form would double-wrap its scheme and port. Reduce it to a bare host
-// (bracketed, for an IPv6 literal) first. (#2218)
+// Reduce node.hostname to a bare host for display-URL composition: it may be configured as a
+// full URL (which would otherwise double-wrap the scheme/port), and the bracketed retry lets a
+// bare IPv6 literal parse.
 export function nodeNameToDisplayHost(name: string): string {
 	if (!name) return name;
 	const hasScheme = /^[a-z][a-z0-9+.-]*:\/\//i.test(name);
-	// A bare IPv6 literal ('::1') only parses once bracketed, so try that form too.
-	const candidateUrls = hasScheme ? [name] : [`http://${name}`, `http://[${name}]`];
-	for (const candidateUrl of candidateUrls) {
+	const candidates = hasScheme ? [name] : [`http://${name}`, `http://[${name}]`];
+	for (const candidate of candidates) {
 		try {
-			const { hostname } = new URL(candidateUrl);
+			const { hostname } = new URL(candidate);
 			if (hostname) return hostname;
-		} catch {
-			// not this form; try the next candidate
-		}
+		} catch {}
 	}
 	return name;
 }
--- a/unitTests/bin/startupBanner.test.js
+++ b/unitTests/bin/startupBanner.test.js
@@ -1,0 +1,108 @@
+'use strict';
+
+// End-to-end guard for the #2218 fix (double-wrapped startup URLs). The pure helper
+// nodeNameToDisplayHost() is unit-tested in unitTests/server/nodeName.test.js, but nothing exercised
+// startupLog() itself, so a call site that switched back from getThisNodeHostname() to
+// getThisNodeName() would recreate the `http://http://host:port/` banner while that suite stayed
+// green. This drives the real, exported startupLog() with a real (URL-valued) node.hostname and
+// asserts the Operations-API HTTP/HTTPS and REST URL lines it prints are well-formed.
+
+const testUtils = require('../testUtils.js');
+testUtils.preTestPrep();
+
+const assert = require('assert');
+
+const env = require('#src/utility/environment/environmentManager');
+const { CONFIG_PARAMS } = require('#src/utility/hdbTerms');
+const { clearThisNodeName } = require('#src/server/nodeName');
+const { startupLog } = require('#src/bin/run');
+
+describe('startupLog banner URLs (#2218 double-wrapped startup URLs)', () => {
+	const originalConsoleLog = console.log;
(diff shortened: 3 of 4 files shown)


bin/run.ts
```diff
@@ -321,17 +321,10 @@ function startupLog(portResolutions: any) {
...
 	if (env.get(CONFIG_PARAMS.OPERATIONSAPI_NETWORK_SECUREPORT)) {
-		logMsg +=
-			'\n' +
-			pad('') +
-			'https://' +
-			getThisNodeName() +
-			':' +
-			env.get(CONFIG_PARAMS.OPERATIONSAPI_NETWORK_SECUREPORT) +
-			'/\n';
+		logMsg += `\n${pad('')}https://${getThisNodeHostname()}:${env.get(CONFIG_PARAMS.OPERATIONSAPI_NETWORK_SECUREPORT)}/\n`;
```
dawsontoth, line 327: Swapping `getThisNodeName` for `getThisNodeHostname` is the only meaningful change, the rest is pure formatting.

server/nodeName.ts
```diff
@@ -58,6 +58,24 @@ export function clearThisNodeName() {
...
 }
 
+/**
+ * Returns this node's bare hostname (no scheme, no port) for composing display URLs.
+ * getThisNodeName() is normally a bare host ('localhost'), but if node.hostname is
+ * configured as a full URL ('http://localhost:9926'), composing a URL from it double-wraps
+ * the scheme and inherits the wrong port (http://http://localhost:9926:9925/). Reduce it
+ * back to the host so the composed URL is well-formed. (#2218)
+ */
+export function getThisNodeHostname(): string {
```
cb1kenobi: **Medium: display-only fix leaves the same malformed `node.hostname` corrupting a non-cosmetic path**

The helper is scoped correctly (I confirmed the only callers are the three banner lines in `bin/run.ts`), so the identity/cert/replication paths are untouched — good call. But the underlying invariant, "`node.hostname` is a bare host", is still unenforced, so the same config that produced #2218 silently corrupts `getThisNodeUrl()`:

```
node.hostname: 'http://localhost:9926'
  -> hostnameToUrl(...) -> 'ws://http://localhost:9926:9925'
  -> new URL(that).hostname === 'http'
```

So replication would advertise/dial a host literally named `http`, with no error. `security/keys.ts:128` (`secureContexts.get(getThisNodeName())`) and `:454` use the same raw value for cert identity.

You already flagged this in the PR notes, and I agree it doesn't belong in this fix. But rather than reducing at each display site, the cheapest root-cause step is one validation where the identity is resolved — `getThisNodeName()` already has a `logger.warn` for the ambiguous-identity case (harper-pro#351), so a sibling warning when `node.hostname` contains `://` or a `:port` would turn a silent corruption into an actionable startup line. Happy to see it as a follow-up issue instead if you'd rather keep this PR tight.

—
_Generated by Barber AI_

server/nodeName.ts
```diff
@@ -58,6 +58,24 @@ export function clearThisNodeName() {
...
+	const name = getThisNodeName();
+	if (!name) return name;
+	try {
+		const hasScheme = /^[a-z][a-z0-9+.-]*:\/\//i.test(name);
+		return new URL(hasScheme ? name : `http://${name}`).hostname || name;
```
cb1kenobi: **Low: unbracketed IPv6 still composes an invalid banner URL, and `URL.hostname` case-folds/punycodes the name**

I ran the helper across the input space. Bracketed IPv6 works as you describe, but the unbracketed form is the branch the fix doesn't reach — `new URL('http://fe80::1')` throws, so the value falls through to the raw name and the banner emits the same class of broken URL this PR is fixing:

```
'[::1]'    -> '[::1]'    -> http://[::1]:9925/    VALID
'fe80::1'  -> 'fe80::1'  -> http://fe80::1:9925/  INVALID
'::1'      -> '::1'      -> http://::1:9925/      INVALID
```

Second, `URL.hostname` normalizes: `NodeA.Example.COM` -> `nodea.example.com`, `nodé.example.com` -> `xn--nod-dma.example.com`. Cosmetically fine, but it means the `Hostname:` line (raw, `bin/run.ts:284`) and the URL lines below it now disagree for any mixed-case or non-ASCII name.

Both are cosmetic-only given the caller set, so no objection to leaving them — but bracketing a bare IPv6 in the fallback is a one-liner if you want the banner airtight.

—
_Generated by Barber AI_

unitTests/server/nodeName.test.js
```diff
@@ -78,6 +78,45 @@ describe('getThisNodeName precedence (harper-pro#351)', () => {
...
+	});
+
+	it('strips a bare host:port down to the host', () => {
+		stubNodeHostname('localhost:9926');
+		assert.strictEqual(getThisNodeHostname(), 'localhost');
+	});
+
+	it('preserves bracketed IPv6 hosts so composed URLs stay valid', () => {
+		stubNodeHostname('https://[::1]:9926');
+		assert.strictEqual(getThisNodeHostname(), '[::1]');
```
cb1kenobi: **Low: the "non-parseable values are returned as-is" behavior claimed in the PR description has no test**

Nice, tight test block — I mutation-tested it (rebuilding `dist/` each time and verifying the marker landed exactly once, since `#src/*` resolves to `dist`, not the `.ts`). The four assertions are all load-bearing: breaking bare pass-through, the scheme strip, the `host:port` strip, or the IPv6 brackets each fails exactly the test you'd expect.

Three mutants survived, all in the fallback paths:

| mutation to `getThisNodeHostname` | result |
|---|---|
| `catch { return undefined; }` | 12 passing, 0 failing |
| drop the `\|\| name` empty-hostname fallback | 12 passing, 0 failing |
| drop `if (!name) return name;` | 12 passing, 0 failing |

The first is the one worth closing, since it's an explicit claim in the PR description. A case with a genuinely unparseable value pins it:

```suggestion
		assert.strictEqual(getThisNodeHostname(), '[::1]');
	});

	it('returns a non-parseable value as-is rather than undefined', () => {
		// new URL('http://fe80::1') throws; the raw name must survive
		stubNodeHostname('fe80::1');
		assert.strictEqual(getThisNodeHostname(), 'fe80::1');
	});
```

(FWIW the `if (!name)` guard looks unreachable — `getThisNodeName()` bottoms out at `'127.0.0.1'` — so that surviving mutant is probably fine to leave.)

—
_Generated by Barber AI_

server/nodeName.ts
dawsontoth: Fixed the bare-IPv6 case in the pushed update — the helper now retries a bracketed form, so `::1`/`fe80::1` compose valid `http://[..]:port/`. The `URL.hostname` case-folding/punycode is cosmetic and startup-log-only (the `Hostname:` line stays raw by design), so I left it.

unitTests/server/nodeName.test.js
dawsontoth: Added a plain-assert test pinning that an unparseable value is returned unchanged (not dropped). I used a genuinely unparseable value rather than `fe80::1`, since the bracketed-retry fix now normalizes bare IPv6 to `[fe80::1]` instead of passing it through.

server/nodeName.ts
dawsontoth: Agreed — keeping this fix display-scoped as you suggested. Documented the root cause in the PR's 'For the human reviewer' section and opened a follow-up to enforce/warn the bare-host `node.hostname` invariant in `getThisNodeName()` — the single place that would also fix `hostnameToUrl()`/replication and the cert identity in `security/keys.ts`.

unitTests/server/nodeName.test.js
```diff
@@ -78,6 +78,32 @@ describe('getThisNodeName precedence (harper-pro#351)', () => {
...
+
+	it('preserves an already-bracketed IPv6 host', () => {
+		assert.strictEqual(nodeNameToDisplayHost('https://[::1]:9926'), '[::1]');
+	});
+
+	it('brackets a bare IPv6 literal so composed URLs stay valid', () => {
+		assert.strictEqual(nodeNameToDisplayHost('::1'), '[::1]');
+	});
+
+	it('returns an unparseable value unchanged rather than dropping it', () => {
```
cb1kenobi, line 108: **Low: the refactor traded away the only coverage of `getThisNodeHostname()` — the function `bin/run.ts` actually calls**

The extraction is a good move and the per-behavior assertions all got *stronger*, not weaker — I re-ran the mutation battery at `b1d81345` (rebuilding `dist/` each time, marker verified exactly once per run; control 14 passing / 0 failing). Every behavior mutant is killed, including the one you just closed:

| mutation | round 1 | now |
|---|---|---|
| `catch`/fallback returns `undefined` | **survived** | killed (1 failure) |
| drop the bracketed-IPv6 retry candidate | n/a (new) | killed (1) |
| `hasScheme = false` / `= true` | killed | killed (2 / 2) |
| `.hostname` -> `.host` | killed | killed (3) |
| break the bare-host candidate | killed | killed (2) |
| strip the IPv6 brackets from the result | killed | killed (2) |
| helper body -> `return name` | killed | killed (4) |

One went the other way. `getThisNodeHostname` is no longer imported here, so this mutant now ships green:

```js
export function getThisNodeHostname(): string {
	return getThisNodeName(); // normalization dropped — 14 passing, 0 failing
}
```

Last round that same mutant failed 3 tests. It matters a little more than a one-line wrapper usually would, because that wrapper *is* the fix from `bin/run.ts`'s point of view (lines 324, 327, 379) — the pure helper has no other production caller. A single test re-pins the composition without walking back the plain-`assert` style, and the `sinon` sandbox is already in this file for the `harper-pro#351` block:

```suggestion
	it('returns an unparseable value unchanged rather than dropping it', () => {
		assert.strictEqual(nodeNameToDisplayHost('node with space'), 'node with space');
	});

	it('getThisNodeHostname() normalizes the configured node.hostname', () => {
		const sandbox = sinon.createSandbox();
		clearThisNodeName();
		try {
			sandbox.stub(env, 'get').callsFake((key) => (key === 'node_hostname' ? 'http://localhost:9926' : undefined));
			assert.strictEqual(getThisNodeHostname(), 'localhost');
		} finally {
			sandbox.restore();
			clearThisNodeName();
		}
	});
});
```

(That needs `getThisNodeHostname` added back to the `require` on line 11.) Entirely your call whether it's worth the sinon — the display-only blast radius caps the downside — but flagging it since it's the one thing this round's refactor gave up.

Also confirmed unchanged from round 1: `getThisNodeHostname()` still has exactly three callers, all startup-banner string composition in `bin/run.ts`, and nothing reaches replication, clustering, or cert matching. `oxlint` and `prettier --check` are clean.

—
_Generated by Barber AI (Opus 5)_

server/nodeName.ts
cb1kenobi: Happy with the disposition — display-scoped is the right call here and I'm not asking for more in this PR.

One loose end worth 30 seconds before merge: I can't find the follow-up issue this refers to. Searched `HarperFast/harper` for `node.hostname` / `hostname` / `getThisNodeName` across open and closed issues, and the newest issue in the repo is still #2218 itself. The PR description also documents the root cause in "For the human reviewer" but doesn't carry a link.

Could you drop the issue link in here (or open it) so the `hostnameToUrl()` -> `ws://http://host:port` corruption and the `security/keys.ts` cert-identity use don't fall off the list once this thread is closed out? The PR body is good documentation but it stops being a work item the moment this merges.

—
_Generated by Barber AI (Opus 5)_

unitTests/server/nodeName.test.js
dawsontoth: Good catch — restored it. Added a real-config test (`env.setProperty('node_hostname', 'http://localhost:9926')` → `getThisNodeHostname()` === `localhost`) exercising the wrapper `bin/run.ts` calls, so the 'drops normalization' mutant now fails. Uses the sanctioned test setter, no new sinon.

server/nodeName.ts
dawsontoth: On tracking: the root-cause fix is already being worked as a separate effort and will land as its own PR — validating/warning the bare-host `node.hostname` invariant in `getThisNodeName()`, which also covers the `hostnameToUrl()` corruption and the `security/keys.ts` cert-identity use. Not opening a duplicate issue; it won't fall off the list.

unitTests/server/nodeName.test.js
cb1kenobi: Verified by mutation on `04874682` — M1 is now **killed**.

Control (unmutated, rebuilt via `tsc -p tsconfig.build.json`): 15 passing, 0 failing.

Applied the exact mutant flagged above (`getThisNodeHostname()` body → `return getThisNodeName();`), rebuilt, and confirmed in `dist/server/nodeName.js` that the mutated form appears exactly once inside the `getThisNodeHostname` function body (a textually-identical `return getThisNodeName();` also exists at line 51 in the pre-existing `server.hostname` getter — isolated the check to the function body itself to avoid a false match there) and the original `nodeNameToDisplayHost(getThisNodeName())` form was gone (count 0).

Re-ran the suite: 14 passing, **1 failing**.

```
1) getThisNodeHostname reads and normalizes the configured node.hostname
     normalizes a URL-valued node.hostname to a bare host:
    AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
    + actual - expected
    + 'http://localhost:9926'
    - 'localhost'
    at Context.<anonymous> (unitTests/server/nodeName.test.js:131:10)
```

Also confirmed the new test isn't vacuous: `env.setProperty('node_hostname', ...)` goes through the real `configUtils` config object (not a stub), and `getThisNodeName()` reads that same real value via `env.get()` — verified this empirically (not just by reading source) with a standalone script: before `setProperty`, `env.get('node_hostname')` was `null`; after, it read back `'http://localhost:9926'`, and `getThisNodeName()` returned that raw string while `getThisNodeHostname()` correctly normalized it to `'localhost'`. So the assertion genuinely depends on the wrapper's normalization, not on an unrelated stub path.

Restored the file, rebuilt, marker count back to 0, suite back to 15 passing / 0 failing.

Nice work closing this quickly — resolving.

_Generated by Barber AI (Claude Opus 5)_

unitTests/bin/startupBanner.test.js
```diff
@@ -0,0 +1,116 @@
...
+		console.log = (...args) => lines.push(args.join(' '));
+		try {
+			startupLog(portResolutions);
+		} finally {
+			console.log = originalConsoleLog;
+		}
+		return lines.join('\n');
+	}
+
+	it('composes well-formed ops + REST URLs when node.hostname is a full URL', () => {
```
cb1kenobi, line 61: Verified this end-to-end harness actually catches the regression it's named for, not just that it runs.

Mutation results (worktree at `228675596`, rebuilt with `tsc -p tsconfig.build.json` after each edit, dist marker count confirmed 0/1/exactly-1 at every step):

| mutation | site | result |
|---|---|---|
| NONE (control) | — | 3 passing |
| B1: revert ops-HTTP line to `getThisNodeName()` (the exact #2218 bug) | `bin/run.ts:324` | **1 failing** — `banner double-wrapped a scheme` |
| B2: swap ops port for the secure-port param | `bin/run.ts:324` | **2 failing** — expected `http://localhost:9925/`, got `http://localhost:null/` / `:undefined/` |
| B3: drop the scheme at the REST site | `bin/run.ts:379` | **3 failing** — expected `http://localhost:9926/`, got `localhost:9926/` |

All three mutants are killed, and B2 in particular confirms the assertions pin the actual composed URL string (host+port+scheme) rather than just "a URL appeared." Restored after each mutation; final state matches control (3 passing, 0 stray markers).

Also confirmed `unitTests/bin/startupBanner.test.js` is picked up by `test:unit:main`'s glob (`unitTests/**/*test.*js`, not excluded), which runs in `test:unit:all` in `.github/workflows/unit-test.yml` — so this runs in CI, not just locally.

This closes the previously-standing "confirmed by reading only" caveat on the banner fix.

_Generated by Barber AI (Claude Opus 5)_

dawsontoth: Thanks for the mutation verification. One note: the test's cleanup was hardened since `228675596` (`d7b5c46bb` now snapshots/restores `node.hostname` and the secure port instead of blanking them) — the assertions your B1/B2/B3 mutants hit are unchanged, so that verification still holds. Confirmed it's in `test:unit:main`'s glob too.

cb1kenobi: Re-verified at `d7b5c46b`: the harness still bites after the cleanup change. Control (NONE): 3 passing, 0 failing. B1 (revert `bin/run.ts:324` to raw `getThisNodeName()`): 2 passing, 1 failing — `AssertionError: banner double-wrapped a scheme` (`http://http://localhost:9926:9925/`), dist marker confirmed exactly once at the mutated site before the run and 0 after restore. `server/nodeName.ts`'s diff between `228675596` and `d7b5c46b` is comment-only — no executable line changed. The `beforeEach`/`afterEach` capture-and-restore is correct: mocha always runs `afterEach` after a successful `beforeEach` regardless of the test outcome, and restoring a captured `undefined` hits the same `setProperty(key, undefined)` path the old blanking code used, so the originally-unset case is unchanged.

_Generated by Barber AI (Claude Opus 5)_

heskew (approved):
Reviewed; no blockers found. Clean display-host normalization and solid test coverage for startup banner URL formatting.

---

server/nodeName.ts
```diff
@@ -58,6 +58,25 @@ export function clearThisNodeName() {
 	nodeName = undefined;
 }
 
+// node.hostname may be configured as a full URL; reduce it to a bare host so composing a
+// display URL from it does not double-wrap the scheme and port.
+export function nodeNameToDisplayHost(name: string): string {
```
kriszyp, line 63: Could we address or explicitly surface the invalid configuration here instead of only normalizing its display? `node.hostname` is the node’s identity, not a URL: the raw value still flows through `getThisNodeName()` into replication URL construction, certificate common names/lookups, and node-name matching. With the exact trigger `http://localhost:9926`, this makes the banner look healthy while replication can still compose `ws://http://localhost:9926:9925`, whose parsed hostname is `http`, and certificate identity remains the raw URL. Please enforce a bare-host invariant during configuration/install validation, or at minimum emit a clear startup warning and add downstream replication/certificate coverage, rather than silently repairing only the rendered links.

kriszyp (approved):
I think codex was a bit confused by the same thing I was: the important part, the warning, is actually in #2223. Probably could have folded into the same PR, but that's fine. Looks good.

server/nodeName.ts
dawsontoth: You're right that the raw value flows past the display into replication URL construction and certificate identity, and that shouldn't be silently repaired. That root-cause fix lives in companion PR #2223, which warns at node-identity resolution in `getThisNodeName()` when `node.hostname` is a URL / `host:port` / non-string (with docs in HarperFast/documentation#632 for the bare-host requirement). I scoped this PR to the display fix so it stays focused; #2223 carries the invariant-surfacing you're asking for. Happy to gate this behind #2223, or combine them, if you'd prefer them to land together — your call.

Issue #1674, opened 2026-07-07
[Epic] Certificate & key-management hardening


kriszyp:
Tracking epic for certificate and key-management modernization: non-RSA key support, Let's Encrypt CSR/signing support, and SNI/ALPN TLS compatibility.

---
_Filed by Claude (Opus 4.8), issue-backlog triage & grouping pass._

PR #2223, opened 2026-08-19
Reject a non-bare-host node identity and fix IPv6 replication URLs

dawsontoth:
Enforce that this node's identity is a **bare hostname or IP literal** — not a URL, `host:port`, or non-string — at every configured source, and make IPv6 identities work end-to-end (canonical form, valid replication URLs, correct certificate SAN typing).

`node.hostname` (and the `replication.hostname` / certificate / listening address it falls back to) is this node's identity: it becomes the certificate common name (`security/keys.ts`) and the host replication advertises and dials (`hostnameToUrl`). A scheme, port, or non-string silently corrupts both — `hostnameToUrl` composes `ws://http://host:9926:9925`, whose parsed hostname is `http`, so the node advertises/dials a host literally named `http` with no error (#2218). Replaces the original warn-only PR after review by **@kriszyp** (replication/TLS domain expert) on this PR and #2219.

**What it does**
- **`utility/nodeIdentity.ts`** (new) — `bareHostViolation(value)`, the single predicate. Accepts a bare hostname or IPv4/IPv6 literal; rejects a scheme, port, path, credentials, non-string, bracketed IPv6 (`[::1]` — the identity is stored unbracketed), a scoped-zone IPv6 (`fe80::1%eth0`), and anything the `ws://` consumer would reject. It parses under a **special scheme** (`https://`) so its host grammar matches the actual replication-URL consumer (a non-special scheme is too lenient — it would pass `node%20`, `0x7f.1`, etc.).
- **`validation/configValidator.ts`** — rejects a URL/port/numeric `node.hostname` **and** `replication.hostname` at the config boundary (there was no `node` schema; `replication.hostname` previously accepted `string|number`). Config validation throws, so a bad identity **fails boot** with a clear message rather than starting known-broken. This is the loud enforcement.
- **`server/nodeName.ts` `getThisNodeName()`** — resolves to the first source that is a valid bare host and **skips** an empty, unusable, or malformed *derived* source (`replication.url` host, certificate CN, listening address). It never throws and never caches an empty identity; `127.0.0.1` is the always-valid floor. Identity is normalized to the **unbracketed** canonical form across sources (`getHostFromListeningPort` and `urlToNodeName` strip IPv6 brackets), so `net.isIP` types it as an **IP** certificate SAN and the same node's identity string is stable.
- **`validation/installValidator.ts`** — the same check runs at **install/upgrade** time on `NODE_HOSTNAME`. It sits right after the v5 migration copies `replication.hostname` into `node.hostname`, so an upgrade carrying a URL-ish value is reported during install instead of only failing at first boot.
- **Skipped sources now warn.** When the resolver ignores a malformed derived source it logs once, naming the source, value, and reason — a typo'd `replication.url` silently resolving identity to `127.0.0.1` was a diagnosis trap on a live cluster. A valid identity resolves silently.
- **`hostnameToUrl()`** — brackets a bare IPv6 literal for URL construction (`::1` → `ws://[::1]:port`); the unbracketed `ws://::1:port` is rejected by the URL parser.
- **`security/keys.ts`** — `getHost()` / `urlToNodeName` / `getCommonNameFromCert` are made fail-soft so a malformed `replication.url` or an unreadable cert file falls back instead of throwing during cert review.

Rejecting an invalid *configured* identity (rather than rewriting it) preserves the certificate common name — kriszyp's stated reason to prefer rejection.

**⚠️ Breaking:** an existing install whose `node.hostname`/`replication.hostname` is a URL, `host:port`, or non-string now **fails to boot** until corrected to a bare host — **needs a release note.** The v5 install migration (`installer.ts` copies `replication.hostname → node.hostname`) can carry such a value into `node.hostname` on upgrade, so per @ldt1996's review that case is now caught at **install time** with the same clear reason, rather than surfacing only at first boot. (My own local dev config had exactly this and was correctly rejected.)

## For the human reviewer

The change grew from a warning into a node-identity validation subsystem after kriszyp's review, and it touches TLS cert generation, so it went through several cross-model review rounds. The judgment calls:

1. **Loud at the config boundary, soft in the resolver.** An invalid `node.hostname`/`replication.hostname` fails boot (kriszyp's ask). `getThisNodeName` deliberately does **not** throw — an earlier revision that threw could crash a request-time caller on a descriptive cert CN, so the resolver skips-and-falls-through while the boundary rejects. Per @ldt1996's review the resolver now also *warns* whenever it skips a source, so the fall-through is visible without being fatal.
2. **IPv6 identity is stored unbracketed** (`::1`), bracketed only for URL construction, so `net.isIP` types the cert SAN as IP and sources agree. `::1:9925` is accepted as a distinct valid IPv6 (not `::1` + port) — inherent to IPv6 syntax, no clean way to disambiguate intent.
3. **Breaking-change scope** (⚠️ above) — deliberate per "prevent a known-broken startup"; the release-note call is yours.
4. **Deferred to follow-ups (surfaced by review, not done here):**
   - **Bare-IPv6-*literal* identity has a cert `secureContexts` lookup gap** — filed as #2261 (P2, under epic #1674): contexts are keyed by OpenSSL's *expanded* SAN text (`2001:DB8:0:0:0:0:0:1`) while `getThisNodeName` returns the *compact* form, so `secureContexts.get()` misses. Pre-existing cert infrastructure this change *exposes* by making IPv6 identities viable.
   - A malformed `replication.url` double-URL (`ws://http://host…` → host `http`) still slips (ambiguous; `replication.url` is a URL by design); the packaged JSON schema isn't yet synced with the runtime schema.

### Rebased onto merged #2219

This now sits on top of #2219, and the two compose well: identity stays the **unbracketed** canonical form (so `net.isIP` types the cert SAN as an IP), while #2219's `nodeNameToDisplayHost` brackets it for display — so the startup banner renders a parseable `http://[::1]:9926/`. That closes the IPv6-banner gap earlier review rounds flagged against this PR, and no longer needs a follow-up.

I did have to update two of #2219's tests, and one of its assertions no longer describes reachable behavior: they set `node.hostname` to a URL and expected the banner to show the *normalized* host (`localhost`). Such a value is now rejected at the config boundary and skipped during resolution, so it never reaches the banner — the banner shows the fallback identity instead, with a warning explaining why. #2219's real regression guard (no `http://http://` double-wrap) is kept and strengthened: the banner's URLs must now all actually parse, and a bare IPv6 `node.hostname` is asserted to render bracketed. Flagging it because those tests are two days old — @kriszyp / @ldt1996, tell me if you'd rather keep the old expectation and soften the resolver instead.

Addresses kriszyp's inline comments on `nodeName.ts:32` (validate every source; falsy non-strings; the fallback) and `:85` (`net.isIP` + bracket IPv6; reject parse failures; test the constructed URL).

## Verification

- **Route (c) — behavior via unit + boundary tests** (all run locally on Node 24):
  - `unitTests/utility/nodeIdentity.test.js` (34): the predicate accepts every valid host shape (hostname, IPv4, IPv6) and rejects scheme/port/path/creds/non-string/bracketed/scoped-IPv6/`node%20`/alt-IPv4-forms — **no false-positive on a valid identity** (a false-positive would break boot).
  - `unitTests/server/nodeName.test.js` (34, including #2219's display tests): `getThisNodeName` returns a valid bare identity, falls through, never caches empty, skips an unusable/malformed derived source without throwing, and yields the **unbracketed** IPv6 identity from both a bracketed listen address and a `replication.url` IPv6 host; `hostnameToUrl('::1')` composes the parseable `ws://[::1]:9933`.
  - `unitTests/validation/configValidator.test.js` (+2) and `installValidator.test.js` (+2): a URL/port/numeric `node.hostname` and `replication.hostname` are rejected with a clear message; bare host and `::1` accepted.
- `unitTests/bin/startupBanner.test.js` (4): no double-wrapped scheme, every banner URL parses, and a bare IPv6 `node.hostname` renders bracketed.
- `tsc` build, `oxlint`/`lint:required`, and `prettier` clean. The `bin`, `validation`, `utility`, and `security` suites were run after the rebase; `bin` is now 220 passing / 0 failing.
- **Full `test:unit:main` cannot run locally** — it dies at load on the RocksDB `LOCK` held by my running local Harper instance (known limitation; run targeted files). CI runs the full matrix. Two unrelated pre-existing failures noted: a path-length-dependent domain-socket test-isolation flake in the validation suite, and two `scopedImport`/compartment tests in the security suite — both fail identically on the base commit.

## Review coverage

Authored by Claude Opus 4.8/5. Reviewed across ten cross-model rounds as the design was reworked from @kriszyp's feedback and each round's findings were fixed. The graded opposite-family leg was **`codex` (`gpt-5.6-sol`)**, joined on the rounds their tooling was available by **`gemini` via agy (default model)**, **Cursor Grok (`cursor-grok-4.5-high`)** and **Composer (`composer-2.5`)**. Blockers fell 3 → 1 → 0 and stayed at 0; the final round's verdict names no blocker or major, and confirms the earlier over-reach fixes (the `ws+unix:` finding is dropped, and the `node.url` / scheme-less-URL / post-unbracket concerns are fixed or non-reproducible). Honest gaps: on the **last few rounds only `codex` ran** — Gemini hit its quota and both Cursor legs failed on a local SSH-agent signing error — and the **Harper domain-adjudication leg crashed on every round**, so outside findings were author-adjudicated rather than machine-filtered. Two human reviews on the current code: @ldt1996 approved and traced the predicate's edge cases (her three asks and nit are implemented here), and @kriszyp's empty-hostname finding is fixed. The last review ran on `b2c02c6850e2`; the pushed commit adds **only test changes** on top of it (production code is byte-identical — verified with `git diff`), which is why the footer's receipt SHA is newer than the reviewed one. Grade 4 is the deterministic floor for a high-risk `security/keys.ts` surface with unadjudicated findings, not a mandatory-finding grade.

<sub>Review-Coverage: authored=unknown; ran=none; rounds=1 @ 0692c49ecd8d</sub>

<sub>Human-Review-Need: 4 @ 0692c49ecd8d</sub>

### DIFF
--- a/security/keys.ts
+++ b/security/keys.ts
@@ -392,13 +392,15 @@ function loadAndWatch(path, loadCert, type) {
 }
 
 function getHost() {
-	let url = getThisNodeUrl();
-	if (url == null) {
+	// urlToNodeName returns undefined for a missing or malformed node/replication URL, so a bad
+	// replication.url falls back to the default host here rather than throwing during cert review.
+	const name = urlToNodeName(getThisNodeUrl());
+	if (name == null) {
 		const host = CERT_DOMAINS[0];
 		logger.info?.('node url is missing from harperdb-config.yaml, using default host' + host);
 		return host;
 	}
-	return urlToNodeName(url);
+	return name;
 }
 
 export function getCommonName() {
--- a/server/nodeName.ts
+++ b/server/nodeName.ts
@@ -1,8 +1,10 @@
 import { readFileSync } from 'node:fs';
 import { X509Certificate } from 'node:crypto';
+import { isIP } from 'node:net';
 import { CONFIG_PARAMS } from '../utility/hdbTerms.ts';
 import * as env from '../utility/environment/environmentManager.ts';
 import { logger } from '../utility/logging/logger.ts';
+import { bareHostViolation } from '../utility/nodeIdentity.ts';
 import { server } from './Server.ts';
 
 Object.defineProperty(server, 'hostname', {
@@ -17,20 +19,26 @@ function getCommonNameFromCert() {
 	const certificatePath: string | undefined =
 		env.get(CONFIG_PARAMS.OPERATIONSAPI_TLS_CERTIFICATE) || env.get(CONFIG_PARAMS.TLS_CERTIFICATE);
 	if (certificatePath) {
-		// we can use this to get the hostname if it isn't provided by config
-		const certParsed = new X509Certificate(readFileSync(certificatePath));
-		const subject = certParsed.subject;
-		return (commonNameFromCert = subject?.match(/CN=(.*)/)?.[1] ?? null);
+		try {
+			// use this to get the hostname if it isn't provided by config
+			const subject = new X509Certificate(readFileSync(certificatePath)).subject;
+			return (commonNameFromCert = subject?.match(/CN=(.*)/)?.[1] ?? null);
+		} catch {
+			// a missing/unparseable cert file must not throw out of identity resolution — the caller
+			// falls back to another source. Do NOT cache the failure, so a later read (e.g. after the
+			// cert is created) can still pick up its common name.
+			return undefined;
+		}
 	}
 }
 
 let nodeName: string | undefined;
 export function getThisNodeName(): string {
 	if (nodeName) return nodeName; // if already determined, just return
-	nodeName = env.get(CONFIG_PARAMS.NODE_HOSTNAME); // standard config
-	if (nodeName) {
+	const configured = env.get(CONFIG_PARAMS.NODE_HOSTNAME);
+	if (configured) {
 		const replicationHostname = env.get('replication_hostname');
-		if (replicationHostname && replicationHostname !== nodeName) {
+		if (replicationHostname && replicationHostname !== configured) {
 			// If these are both set and differ, the node identity is ambiguous. node.hostname
 			// wins (it is what this node identifies as), but if it doesn't match the name this
 			// node is registered under in hdb_nodes, replication for that name silently turns
@@ -38,18 +46,21 @@ export function getThisNodeName(): string {
 			// node.hostname value — that's how a wrong identity (e.g. 'localhost') gets locked
 			// in. Steer the operator to reconcile against the registered node name instead.
 			logger.warn?.(
-				`The node.hostname (${nodeName}) and replication.hostname (${replicationHostname}) configuration values are both set and differ. This node will identify as "${nodeName}". Ensure that name matches this node's row in system.hdb_nodes; if it does not, set node.hostname (or remove it to fall back to replication.hostname) to match the registered node name, otherwise replication for this node will be disabled.`
+				`The node.hostname (${configured}) and replication.hostname (${replicationHostname}) configuration values are both set and differ. This node will identify as "${configured}". Ensure that name matches this node's row in system.hdb_nodes; if it does not, set node.hostname (or remove it to fall back to replication.hostname) to match the registered node name, otherwise replication for this node will be disabled.`
 			);
 		}
-		return nodeName;
 	}
-	// fallback to other means of getting the node name
+	// Use the first source that is a valid bare host (harper#2218). node.hostname and
+	// replication.hostname are rejected at the config boundary when invalid; the remaining derived
+	// sources are validated here and skipped — not fatal — if unusable, so a descriptive certificate
+	// common name or an empty value can never cache a corrupt identity or crash a request-time caller.
 	nodeName =
-		env.get('replication_hostname') ?? // for backwards compatibility
-		urlToNodeName(env.get('replication_url') as string) ??
-		getCommonNameFromCert() ??
-		getHostFromListeningPort('operationsapi_network_secureport') ??
-		getHostFromListeningPort('operationsapi_network_port') ??
+		asBareHost(configured, 'node.hostname') ??
+		asBareHost(env.get('replication_hostname'), 'replication.hostname') ?? // for backwards compatibility
+		asBareHost(replicationUrlHost(), 'the host in replication.url') ??
+		asBareHost(getCommonNameFromCert(), 'the certificate common name') ??
+		asBareHost(getHostFromListeningPort('operationsapi_network_secureport'), 'the operations API host') ??
+		asBareHost(getHostFromListeningPort('operationsapi_network_port'), 'the operations API host') ??
 		'127.0.0.1';
 	return nodeName;
 }
@@ -58,8 +69,9 @@ export function clearThisNodeName() {
 	nodeName = undefined;
 }
 
-// node.hostname may be configured as a full URL; reduce it to a bare host so composing a
-// display URL from it does not double-wrap the scheme and port.
+// Reduce a node name to a host safe to compose a display URL from: strip a scheme/port (defensive —
+// a configured node.hostname carrying those is now rejected at the config boundary) and bracket a
+// bare IPv6 literal, without which the composed startup URLs are unparseable.
 export function nodeNameToDisplayHost(name: string): string {
 	if (!name) return name;
 	const hasScheme = /^[a-z][a-z0-9+.-]*:\/\//i.test(name);
@@ -77,10 +89,46 @@ export function getThisNodeHostname(): string {
 	return nodeNameToDisplayHost(getThisNodeName());
 }
 
+/**
+ * The host of replication.url, or undefined when it is unset or carries no usable host. Warns in the
+ * latter case for the same reason asBareHost does: a typo'd URL must not silently cost this node its
+ * identity. urlToNodeName itself stays quiet — it is also used for peer URLs, not just this node's.
+ */
+function replicationUrlHost(): string | undefined {
+	const url: string | undefined = env.get(CONFIG_PARAMS.REPLICATION_URL);
+	if (!url) return undefined;
+	const host = urlToNodeName(url);
+	if (host === undefined) {
+		logger.warn?.(
+			`Ignoring replication.url (${JSON.stringify(url)}) as this node's identity: it is not a URL with a host. Falling back to the next available source, which may leave this node identifying as 127.0.0.1 and unable to replicate.`
+		);
+	}
+	return host;
+}
+
+/**
+ * A candidate identity source, or undefined when it is unset or not a bare host. Skipping is
+ * deliberate (see getThisNodeName) but silence is a diagnosis trap: a typo'd source that quietly
+ * resolves identity to 127.0.0.1 looks like a working node until replication fails, so say so.
+ */
+function asBareHost(value: unknown, source: string): string | undefined {
+	if (value === undefined || value === null || value === '') return undefined;
+	const violation = bareHostViolation(value);
+	if (!violation) return value as string;
+	logger.warn?.(
+		`Ignoring ${source} (${JSON.stringify(value)}) as this node's identity: it ${violation}. Falling back to the next available source, which may leave this node identifying as 127.0.0.1 and unable to replicate. Set it to a bare hostname or IP literal.`
+	);
+	return undefined;
+}
+
 function getHostFromListeningPort(key: string) {
 	const port: string | undefined = env.get(key);
 	const lastColon = port?.lastIndexOf?.(':');
-	if (lastColon > 0) return port.slice(0, lastColon);
+	if (lastColon > 0) {
+		const host = port.slice(0, lastColon);
+		// A listen address stores IPv6 bracketed ("[::1]:9925"); the node identity is unbracketed.
+		return host.startsWith('[') && host.endsWith(']') ? host.slice(1, -1) : host;
+	}
 }
 function getPortFromListeningPort(key: string) {
 	const port: string | undefined = env.get(key);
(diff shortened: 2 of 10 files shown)


server/nodeName.ts
```diff
@@ -58,6 +59,41 @@ export function clearThisNodeName() {
...
+ * hostnameToUrl composes "ws://http://host:9926:9925" whose hostname resolves to "http" (#2218). We
+ * warn but never rewrite the value, since normalizing identity would change the certificate CN.
+ *
+ * The runtime typeof guard is load-bearing: TypeStrip erases the `: string` annotation and YAML can
+ * yield a number (unquoted `node.hostname: 9925`). Ports are found by parsing the authority under a
+ * non-special scheme (so a default port like ":80" is not normalized away) after stripping any
+ * configured scheme; a value that is not a parseable authority (a bare IPv6 literal like "::1", or a
+ * malformed entry) throws and is treated as bare — the warning is best-effort, not full validation.
+ */
+function warnIfNodeHostnameNotBare(name: string) {
```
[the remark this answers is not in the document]
dawsontoth: Thanks — keeping the `string` annotation. The runtime `typeof name !== 'string'` guard already handles the non-string case safely (an unquoted YAML `node.hostname: 9925` reaches `env.get` as a number/boolean), and that path is covered by unit tests (the `9925` and `true` cases). `unknown` is a reasonable alternative signature, but since it's a TypeStrip-erased annotation with identical runtime behaviour, I'd rather not re-open this commit's converged multi-model review for it. Happy to switch to `unknown` in a follow-up if a maintainer prefers the stricter signature.

server/nodeName.ts
```diff
@@ -58,6 +59,41 @@ export function clearThisNodeName() {
...
+		const hasScheme = name.includes('://');
+		const authority = hasScheme ? name.slice(name.indexOf('://') + 3) : name;
+		let hasPort = false;
+		try {
+			hasPort = Boolean(new URL(`harper-node://${authority}`).port);
+		} catch {
+			/* not a parseable authority — treat as a bare hostname */
+		}
+		const parts = [hasScheme && 'a scheme', hasPort && 'a port'].filter(Boolean);
+		if (parts.length) problem = `contains ${parts.join(' and ')}`;
```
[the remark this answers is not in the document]
dawsontoth: Keeping the concise form here. `warnIfNodeHostnameNotBare` runs once on the cold startup path (the resolved name is then cached in module scope), so the intermediate array is immaterial, and `.filter(Boolean).join(' and ')` type-checks cleanly (`tsc` passes) — `join` coerces the `(string | false)[]` fine and the `false` entries are removed at runtime. The two-line version reads a touch clearer to me as-is.

server/nodeName.ts
```diff
@@ -29,6 +29,7 @@ export function getThisNodeName(): string {
 	if (nodeName) return nodeName; // if already determined, just return
 	nodeName = env.get(CONFIG_PARAMS.NODE_HOSTNAME); // standard config
 	if (nodeName) {
+		warnIfNodeHostnameNotBare(nodeName);
```
kriszyp: Could we enforce this identity invariant in configuration validation and reuse that check for every configured source? This call runs only for a truthy canonical `node.hostname`, so `node.hostname: false` or `0` skips the promised non-string diagnostic, while the supported `replication.hostname: http://host:9926` fallback reaches `hostnameToUrl` with the same corrupting value and no warning. `validation/configValidator.ts:249-251` also accepts numeric legacy hostnames, and the schema has no `node` definition, so configuration writes and boot accept the bad value. Please validate both hostname settings as bare strings at the boundary, or at minimum inspect falsy non-strings and validate the final fallback value before caching it. Rejecting an invalid identity avoids rewriting the certificate CN while preventing a known-broken startup.

— KrAIs (GPT-5)

server/nodeName.ts
```diff
@@ -58,6 +59,41 @@ export function clearThisNodeName() {
...
+	let problem: string | undefined;
+	if (typeof name !== 'string') {
+		problem = `is a ${typeof name}, not a string`;
+	} else {
+		const hasScheme = name.includes('://');
+		const authority = hasScheme ? name.slice(name.indexOf('://') + 3) : name;
+		let hasPort = false;
+		try {
+			hasPort = Boolean(new URL(`harper-node://${authority}`).port);
+		} catch {
```
kriszyp: Could we avoid treating every URL parse failure as “bare”? For `localhost:99999` and `ws:host:9925`, parsing throws here, the new tests intentionally expect silence, and `hostnameToUrl` later produces an invalid or double-port URL. `::1` is also not safe today: `server/nodeName.ts:109-118` emits `ws://::1:<port>`, which URL parsing rejects. Please special-case a confirmed IPv6 literal with `net.isIP` and bracket it during URL construction, while warning or rejecting other parse failures. An end-to-end test of the constructed replication URL would protect the actual invariant better than checking only the warning count.

— KrAIs (GPT-5)

heskew (approved):
Reviewed; no blockers found. Clean, non-intrusive warning implementation that guards against silent replication/TLS misconfigurations.

---

kriszyp (commented):
Sounds helpful

server/nodeName.ts
dawsontoth: Done — reworked to enforce the invariant at every configured source, per your review. Config validation now rejects a URL/port/numeric `node.hostname` **and** `replication.hostname` at the boundary (there was no `node` schema, and `replication.hostname` accepted `string|number`), so a bad identity fails boot with a clear message. `getThisNodeName` validates the resolved identity from every source and skips an empty/unusable/malformed one — it never caches a corrupt identity, and falsy non-strings are handled. Rejecting rather than rewriting preserves the cert CN, as you noted. Shared predicate in `utility/nodeIdentity.ts`. This is breaking for existing URL-configured / v5-migrated installs — flagged for release notes in the PR description.

server/nodeName.ts
dawsontoth: Done — IPv6 is handled end-to-end now. Identity is stored **unbracketed** (the predicate rejects brackets; `getHostFromListeningPort` and `urlToNodeName` strip them) so `net.isIP` types it as an IP cert SAN, and `hostnameToUrl` brackets it for the `ws://` URL (`::1` → `ws://[::1]:port`). Unparseable/unusable derived sources are skipped rather than treated as bare, and I added an end-to-end test asserting the composed replication URL parses (`new URL(hostnameToUrl('::1')).hostname`). One documented follow-up: a bare-IPv6-**literal** identity still has a cert `secureContexts` lookup gap — contexts are keyed by OpenSSL's expanded SAN form (`2001:DB8:0:0:0:0:0:1`) vs the compact `getThisNodeName`; it's a pre-existing cert-infra limitation this change exposes and needs its own focused change.

ldt1996 (approved):
LGTM! nice work

security/keys.ts
```diff
@@ -392,13 +392,15 @@ function loadAndWatch(path, loadCert, type) {
 }
 
 function getHost() {
-	let url = getThisNodeUrl();
-	if (url == null) {
+	// urlToNodeName returns undefined for a missing or malformed node/replication URL, so a bad
+	// replication.url falls back to the default host here rather than throwing during cert review.
+	const name = urlToNodeName(getThisNodeUrl());
```
kriszyp, line 397: Could we treat an empty parsed hostname as invalid too? `new URL('mailto:operator@example.com')`, `data:...`, and `file:...` parse successfully but have `hostname === ''`, so `urlToNodeName()` returns `''`. Since `replication.url` is only constrained to be a string at `validation/configValidator.ts:255`, this `name == null` check passes and `certExtensions()` adds an empty DNS SAN instead of using the default host or rejecting the configuration. Return `undefined` for an empty hostname and validate that `replication.url` is a supported URL with an authority; please add a hostless-scheme regression test.

kriszyp (approved):
Great! (a suggestion for handling empty hostname too)

security/keys.ts
dawsontoth: Good catch — confirmed and fixed. `mailto:`, `data:`, `file:` and `urn:` all parse with `hostname === ''`, and since `'' == null` is false it slipped straight past the `name == null` check into an empty DNS SAN.

Fixed in two layers:
- `urlToNodeName` now returns `undefined` when the parsed hostname is empty, so a hostless URL is treated like a malformed one — `getHost()` falls back to `CERT_DOMAINS[0]` and identity resolution skips the source (verified: `replication.url: mailto:…` → identity `127.0.0.1`, no empty SAN).
- `replication.url` is validated at the config boundary: a value written as a real authority URL (`//`) that still parses to no host is rejected.

On that second layer I deliberately kept it narrow after review pushback: a scheme-less `node1.example.com:9933` actually parses as *scheme* `node1.example.com:` with an empty hostname, so it is indistinguishable from `mailto:` at the boundary — rejecting that shape would turn a previously harmless config into a boot failure. Those forms stay tolerated at the boundary and are caught by the runtime guard instead, which is the layer that actually prevents the empty SAN. `node.url` is left as a plain string since nothing in this repo reads it.

Regression tests added as you asked: four hostless-scheme cases (asserting `urlToNodeName` yields no name at all, not just that resolution reaches the floor) plus the config-boundary case.

dawsontoth:
Thanks — and thanks for actually tracing the predicate rather than taking the description's word for it. All three asks are in, plus the nit:

**1. Install-time validation.** `installValidator` now runs `bareHostViolation` on `NODE_HOSTNAME`, which lands immediately after `installer.ts:142` copies `REPLICATION_HOSTNAME` into it. So a v5 upgrade carrying a URL-ish value now fails *during install* with the same reason the config boundary gives, instead of only at first boot. You were right that a release note alone leaves the operator discovering it the hard way.

**2. Warn on skip.** Taken. `asBareHost` now carries a source label and warns once per skipped source with the source, value, and reason; a `replicationUrlHost()` helper does the same when `replication.url` has no usable host. `urlToNodeName` itself stays silent, since `keys.ts` also uses it for *peer* URLs rather than only this node's identity. Verified both directions: a malformed/hostless `replication.url` or a bad listening host each warn exactly once and fall to `127.0.0.1`, and a valid identity resolves with zero warnings.

**3. Issue filed.** #2261 — P2, typed Bug, parented to epic #1674, with the mechanism (compact `getThisNodeName` vs OpenSSL-expanded SAN key), the impact (cert regenerated every boot, possible `getReplicationCertAuth` null deref), and a suggested fix. Agreed it would have evaporated without a number.

**Nit.** Fixed — `myhost:443` now reports `must not include a port` rather than the vaguer `is not a bare hostname`. A trailing `:` after the parsed hostname is checked before the round-trip comparison, so a default port for the parse scheme no longer masks the real reason. Covered by a test.

Issue #699, opened 2026-05-21
Prevent users from removing their own role


kriszyp:
## Request

Add validation to prevent a user from removing or downgrading their own role. A super_user revoking their own super_user status can leave a system without any admins.

🤖 Filed by Claude on behalf of Kris.

PR #2228, opened 2026-08-19
Release blob locks before failed saves settle

kriszyp:
Failed re-streamable blob saves could settle their rejection before asynchronous PENDING-marker cleanup and blob unlock completed. That ordering made promise settlement an unreliable lifecycle boundary for callers attempting immediate recovery.

This companion hardening completes failed-save cleanup before rejection settles and handles synchronous PENDING-marker write failures through the same cleanup path. It supports Harper Pro's #699 recovery fix, but is not the root cause of that stall: Pro could also start an orphaned repair after connection teardown had already swept in-flight receives.

## For the human reviewer

1. Treat failed-save rejection as a cleanup barrier rather than rejecting immediately. Successful blob writes are unchanged.
2. Use `writeFile` callback completion rather than adding `fsync` durability. This preserves the existing durability policy while making marker visibility and lock release deterministic to callers.
3. Preserve the source stream's original error after cleanup. Cleanup failures must not mask the operation's primary failure.

## Verification

- RED on `origin/main` with the new assertion: `0 passing, 1 failing` because the aborted save's blob lock was still held when rejection settled.
- GREEN on this branch: `mocha unitTests/resources/blob.test.js --grep 'source-stream write that aborts'` passed.
- The broader core resource suite completed with 1,581 passing, 15 pending, and 3 unrelated existing failures. Changed-file lint passed; full lint reports existing unrelated warnings.
- End-to-end Pro verification is documented in HarperFast/harper-pro#732.

<sub>Review-Coverage: authored=codex; ran=gemini; adjudicated=domain; declined=claude,cursor-grok,cursor-composer; rounds=4 @ 9d51c78aef58</sub>

<sub>Human-Review-Need: 3 (decisions: rejection-after-marker, preserve-source-error, mutate-runner-mirrors) @ 9d51c78aef58</sub>

### DIFF
--- a/.github/workflows/integration-tests-nextjs.yml
+++ b/.github/workflows/integration-tests-nextjs.yml
@@ -122,7 +122,12 @@ jobs:
         run: npm run install:fixtures
 
       - name: Install Playwright browsers
-        run: npx playwright install --with-deps chromium
+        run: |
+          # APT can hang instead of failing over when the Azure mirror is unavailable.
+          if [[ -f /etc/apt/apt-mirrors.txt ]]; then
+            sudo sed -i '\|azure.archive.ubuntu.com|d' /etc/apt/apt-mirrors.txt
+          fi
+          npx playwright install --with-deps chromium
 
       - name: Run integration tests
         run: npm run test:integration
--- a/.github/workflows/integration-tests-nextjs.yml
+++ b/.github/workflows/integration-tests-nextjs.yml
@@ -122,6 +122,7 @@ jobs:
         run: npm run install:fixtures
 
       - name: Install Playwright browsers
+        timeout-minutes: 10
         run: |
           # APT can hang instead of failing over when the Azure mirror is unavailable.
           if [[ -f /etc/apt/apt-mirrors.txt ]]; then
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -1476,10 +1476,17 @@ function writeBlobWithStream(
 					const messageBuffer = Buffer.from(String(error));
 					const header = new Uint8Array(HEADER_SIZE);
 					new DataView(header.buffer).setBigInt64(0, BigInt(messageBuffer.length) | (BigInt(PENDING_TYPE) << 48n));
-					writeFile(filePath, Buffer.concat([header, messageBuffer]), (writeError: Error) => {
+					const finishPendingMarker = (writeError?: Error) => {
 						if (writeError) logger.debug?.('Error writing pending marker to blob file', writeError);
 						store.unlock(lockKey);
-					});
+						reject(error);
+					};
+					try {
+						writeFile(filePath, Buffer.concat([header, messageBuffer]), finishPendingMarker);
+					} catch (writeError) {
+						finishPendingMarker(writeError as Error);
+					}
+					return;
 				} else {
 					store.unlock(lockKey);
 					try {
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -138,6 +138,65 @@ export let blobsWereEncoded = false; // keep track of whether blobs were encoded
 // the header is 8 bytes
 const DEFAULT_BLOB_READ_TIMEOUT = 20000;
 /**
+ * Upper bound on the PENDING-marker cleanup barrier in the aborted-save path below. That barrier
+ * defers the save's rejection until the marker write and the lock release have both happened, so
+ * settlement depends on a `writeFile` callback firing. On a wedged volume it may never fire, and an
+ * un-settled `saving` is worse than a late one: on the replication receive path it keeps
+ * `outstandingBlobsToFinish` non-empty, which clamps the resume cursor with no watchdog to release
+ * it. 30s is five orders of magnitude above a working disk's cost for this ~200-byte write, so the
+ * fallback only fires when the write really is not coming back, and stays well inside the 120s
+ * source-idle timeout that would otherwise be the only exit.
+ */
+const PENDING_MARKER_WRITE_TIMEOUT = 30000;
+
+/**
+ * The bounded, once-only cleanup barrier used by the aborted-save PENDING-marker path.
+ *
+ * Contract: `release` runs before `settle`, both run exactly once, and they run no later than
+ * `timeoutMs` even if the caller's completion callback never arrives.
+ *
+ * Ordering matters because a consumer that observes the rejection immediately re-acquires the blob
+ * lock (Pro's in-place repair does exactly this); settling first would make it race a lock this
+ * path still holds. Bounding matters because the completion callback is an `fs` callback that a
+ * wedged volume may never deliver, and an un-settled save clamps the replication resume cursor
+ * with nothing to unclamp it. The once-guard matters because `store.unlock` has no ownership
+ * check, so a timer and a late callback both firing would release a lock another writer has since
+ * taken — invisible to tests, which is why this is guarded rather than merely documented.
+ *
+ * Exported so the ordering, the bound, and the guard can be unit-tested without stalling a real
+ * filesystem write; the production caller is {@link writeBlobWithStream}'s abort branch.
+ */
+export function createPendingMarkerBarrier(options: {
+	release: () => void;
+	settle: () => void;
+	timeoutMs: number;
+	onTimeout?: () => void;
+	onWriteError?: (writeError: unknown) => void;
+}): (writeError?: unknown) => void {
+	let settled = false;
+	const finish = (writeError?: unknown) => {
+		if (settled) return;
+		settled = true;
+		clearTimeout(timer);
+		if (writeError) options.onWriteError?.(writeError);
+		options.release();
+		options.settle();
+	};
+	// `finish`'s own guard is what makes double-release impossible; the `settled` check here only
+	// suppresses a spurious `onTimeout` log, and the `clearTimeout` above is tidiness (the timer is
+	// unref'd, so a stray one costs nothing). Verified by mutation: dropping the guard in `finish`
+	// turns the suite red, dropping the `clearTimeout` does not.
+	const timer = setTimeout(() => {
+		if (settled) return;
+		options.onTimeout?.();
+		finish();
+	}, options.timeoutMs);
+	// Never hold the process open for the fallback; it exists to unwedge a caller, not to keep the
+	// event loop alive.
+	timer.unref?.();
+	return finish;
+}
+/**
  * How long a blob read will wait for an in-progress write to finish before giving up. This bounds
  * the read paths — the stream() open-retry loop and the incomplete-content waits — so a blob whose
  * backing file is being written, truncated, or deleted returns a prompt, actionable error instead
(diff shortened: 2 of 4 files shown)


resources/blob.ts
```diff
@@ -1476,10 +1476,17 @@ function writeBlobWithStream(
...
 					new DataView(header.buffer).setBigInt64(0, BigInt(messageBuffer.length) | (BigInt(PENDING_TYPE) << 48n));
-					writeFile(filePath, Buffer.concat([header, messageBuffer]), (writeError: Error) => {
+					const finishPendingMarker = (writeError?: Error) => {
 						if (writeError) logger.debug?.('Error writing pending marker to blob file', writeError);
 						store.unlock(lockKey);
-					});
+						reject(error);
+					};
+					try {
+						writeFile(filePath, Buffer.concat([header, messageBuffer]), finishPendingMarker);
```
cb1kenobi, line 1561: **Medium: the cleanup barrier is unbounded — a stalled `writeFile` now means the save never settles at all**

Moving `reject(error)` into `finishPendingMarker` makes settlement depend on this `writeFile` completing. If the callback never fires (fs stall, EIO hang, wedged network volume) `storageInfo.saving` never settles; before this change the caller still got a prompt rejection and only the unlock was late.

Verified on this branch: replacing the `writeFile` call with a no-op that never invokes its callback makes `#481: a source-stream write that aborts...` time out rather than fail — the promise never settles. (Same test, unmodified otherwise; `.mocharc` sets `timeout: 0`, so in CI that is an unbounded hang.)

This branch only runs when `idleTimeoutMs > 0`, i.e. the replication receive path, which is where a never-settling `saving` costs the most: Pro pushes it into `outstandingBlobsToFinish`, and `cursorBlockedByBlob()` clamps the resume cursor to `lastDurableSequenceId` for as long as that array is non-empty. A prompt rejection gets classified (`isReplicationConnectionClosedError` / `isUnrecoverableSourceBlobError`) and arms the healing reconnect; no rejection means the cursor stays clamped with no watchdog — the wedge shape the comment at lines 1330-1336 describes.

Suggested fix: bound the barrier with an unref'd fallback timer, plus a once-guard so the timer and the callback cannot both fire. The guard is not optional: rocksdb-js `unlock(key)` has no ownership check, so a double release would hand the lock away from a writer that has since acquired it.

```ts
let settled = false;
const finishPendingMarker = (writeError?: Error) => {
	if (settled) return;
	settled = true;
	if (writeError) logger.debug?.('Error writing pending marker to blob file', writeError);
	store.unlock(lockKey);
	reject(error);
};
setTimeout(finishPendingMarker, PENDING_MARKER_WRITE_TIMEOUT_MS).unref();
```

—
_Generated by Barber AI_

resources/blob.ts
```diff
@@ -1476,10 +1476,17 @@ function writeBlobWithStream(
...
 					const header = new Uint8Array(HEADER_SIZE);
 					new DataView(header.buffer).setBigInt64(0, BigInt(messageBuffer.length) | (BigInt(PENDING_TYPE) << 48n));
-					writeFile(filePath, Buffer.concat([header, messageBuffer]), (writeError: Error) => {
+					const finishPendingMarker = (writeError?: Error) => {
 						if (writeError) logger.debug?.('Error writing pending marker to blob file', writeError);
 						store.unlock(lockKey);
-					});
+						reject(error);
+					};
+					try {
```
cb1kenobi, line 1557: **Low: the marker payload is built outside this `try`, so a throw there still escapes uncaught**

Lines 1476-1478 (`Buffer.from(String(error))` and the `DataView` header build) run before the `try`. If any of them throws — e.g. an error whose `toString` throws — the exception escapes `finished()`, which is the `pipeline` callback, so it surfaces as an uncaughtException with the lock still held and `saving` never settling. That is precisely the failure mode the new `catch` was added to eliminate, so it seems worth pulling those three lines in to finish the job.

Suggested fix: move lines 1476-1478 inside this `try` so they route through `finishPendingMarker` too.

—
_Generated by Barber AI_

unitTests/resources/blob.test.js
```diff
@@ -1260,21 +1260,19 @@ describe('Blob test', () => {
...
 		const blob = await createBlob(source, { size: 20000 });
 		const saving = decodeFromDatabase(() => saveBlob(blob).saving, store);
-		source.write(randomBytes(4000)); // a partial body lands before the abort
-		await new Promise((resolve) => setTimeout(resolve, 20));
+		const lockKey = getFileId(blob) + ':blob';
+		await new Promise((resolve, reject) => {
+			source.write(randomBytes(4000), (error) => (error ? reject(error) : resolve()));
+		});
 		source.destroy(new Error('Blob source stream idle for 120000ms'));
 		await assert.rejects(Promise.resolve(saving), 'the aborted save rejects');
```
cb1kenobi: **Low: two of the PR's three stated decisions are not covered by the new assertions**

The lock and marker assertions here are genuinely load-bearing — reverting the barrier ordering *or* dropping the `store.unlock` each turn this test red, which is a real improvement on the old `waitFor` poll. Two gaps remain, both confirmed by mutation on this branch:

**1. Original-error preservation is unasserted.** The second argument here is a *message* string, not a matcher, so nothing checks the rejection reason — replacing `reject(error)` with `reject(new Error('...'))` in `finishPendingMarker` leaves this test green. That matters downstream: Pro's receive `.catch` classifies this exact error (`isReplicationConnectionClosedError` / `isUnrecoverableSourceBlobError` / `isSourceBlobUnavailable`) to decide whether the resume cursor holds or advances, so a swapped reason would misroute that decision. The regex form below is verified green against this branch and red against the masked-error mutant:

```suggestion
		await assert.rejects(Promise.resolve(saving), /Blob source stream idle for 120000ms/);
```

**2. The synchronous marker-write path is untested.** Replacing `finishPendingMarker(writeError as Error)` in the new `catch` with a no-op also leaves this test green, so nothing guards that half of the change. It is the half with the sharpest failure mode: with the routing removed and a synchronous throw forced, this same test times out — the save never settles and the lock is held forever. A sibling test that forces a synchronous `writeFile` throw would lock it in.

—
_Generated by Barber AI_

.github/workflows/integration-tests-nextjs.yml
```diff
@@ -122,7 +122,12 @@ jobs:
         run: npm run install:fixtures
 
       - name: Install Playwright browsers
-        run: npx playwright install --with-deps chromium
+        run: |
+          # APT can hang instead of failing over when the Azure mirror is unavailable.
+          if [[ -f /etc/apt/apt-mirrors.txt ]]; then
+            sudo sed -i '\|azure.archive.ubuntu.com|d' /etc/apt/apt-mirrors.txt
+          fi
+          npx playwright install --with-deps chromium
```
cb1kenobi, line 131: **Nit: the step still has no timeout, so a hang costs the full job budget**

Dropping the Azure mirror addresses the known trigger, but if apt stalls for any other reason this step blocks until the job-level limit rather than failing fast. A `timeout-minutes` on the step bounds it regardless of cause and complements the mirror fix nicely (the `Run integration tests` step just below already has one).

Minor, separate note: the `sed` deletes the azure entries unconditionally. Harmless on GitHub-hosted runners since `archive.ubuntu.com` remains, but it would leave an empty mirror list on any image where Azure is the only entry.

—
_Generated by Barber AI_

resources/blob.ts
kriszyp: Addressed in a2f7dfe3f: marker payload construction now runs inside the same try/catch as writeFile, so synchronous conversion/allocation failures use the unlock-before-reject cleanup.

— GPT-5 Codex

unitTests/resources/blob.test.js
kriszyp: Addressed in a2f7dfe3f: the existing abort regression now asserts exact source-error identity, and a sibling regression forces synchronous marker construction failure without stubbing, proving the original error settles and the blob lock is released. The new regression fails on the prior production code with the intended uncaught marker-construction error.

— GPT-5 Codex

.github/workflows/integration-tests-nextjs.yml
kriszyp: Addressed in a2f7dfe3f: the Playwright browser-install step now has a 10-minute timeout, bounding any remaining APT stall independently of the mirror workaround.

— GPT-5 Codex

resources/blob.ts
cb1kenobi: Re-verified at `a2f7dfe3` — the other three items are cleanly addressed (marker construction is now inside the `try`, the rejection reason is asserted with `strictEqual`, the sibling sync-failure test is there, and the workflow step has `timeout-minutes`). Nice tightening. This one is still open, and re-measuring it also produced a concrete answer to the once-guard question, so recording both here rather than opening a new thread.

**Still unbounded.** Same probe as last round against the built `dist/` at this head — replace the `writeFile` call with one whose callback never fires, everything else untouched:

- `mocha unitTests/resources/blob.test.js --timeout 30000` → `#481: a source-stream write that aborts...` **times out**, it does not fail.
- Re-run at `--timeout 120000` → still a timeout, `0 passing`. Nothing settles `storageInfo.saving` in two minutes; there is no fallback path.

So the shape is unchanged: on the replication receive path (`idleTimeoutMs > 0`) Pro pushes `saving` into `outstandingBlobsToFinish`, and the entry is spliced out only in the `.finally` at `replicationConnection.ts:5649` — settlement is the *only* exit. `cursorBlockedByBlob()` (`:3055`) then holds the resume cursor at `lastDurableSequenceId` for as long as that array is non-empty, and there is no watchdog on `tracked`. A never-settling promise is a permanently clamped cursor.

**The once-guard is load-bearing — measured, not inferred.** Three instrumented runs of the full file (each verified as exactly one mutation marker in `dist/`, 99 tests):

| variant | result |
|---|---|
| this head, counter on `finishPendingMarker` | 99 passing, **zero** double entries — single-entry today, so no guard is needed *yet* |
| add a `0ms` unref'd fallback timer, **no** guard | 99 passing, but `finishPendingMarker` ran **twice** on 5 different lock keys |
| same timer **with** `if (settled) return` | 99 passing, second call short-circuits before `store.unlock` |

The middle row is the one that matters: the suite stayed entirely green while `store.unlock(lockKey)` was called twice per abort. `db.unlock(key)` in rocksdb-js has no ownership check — its README defines it as releasing the lock on the key and returning whether a lock existed — so the second release hands away whatever writer holds it by then. Silent, and no test in the suite can see it.

Suggested fix, unchanged, with the guard first:

```ts
let settled = false;
const finishPendingMarker = (writeError?: unknown) => {
	if (settled) return;
	settled = true;
	if (writeError) logger.debug?.('Error writing pending marker to blob file', writeError);
	store.unlock(lockKey);
	reject(error);
};
setTimeout(finishPendingMarker, PENDING_MARKER_WRITE_TIMEOUT_MS).unref();
```

Worth noting the fallback should reject with the original `error`, not a timeout error — the `strictEqual` assertion you added this round already pins that, so the guarded version stays green.

—
_Generated by Barber AI_

kriszyp:
The one red check here — `Integration Tests 1/6 (Windows, Node.js v24)` — is **pre-existing `main` red, not caused by this PR.**

Same failure, same test, on a `main` push from today ([run 32366910072](https://github.com/HarperFast/harper/actions/runs/32366910072)): `txnlog-purge-stale-read-blast.test.ts` subtest 1 times out seeding, and subtests 2/4/5/6 cascade from the truncated seed. Nothing in that suite touches `resources/blob.ts`.

Root cause traced and filed as #2243: the RocksDB arm of that seed takes ~2s on Linux and 83–238s on Windows; above ~180s it trips the test's 60s per-request abort. Run-to-run it lands on either side of the line, which is why the shard flips red and green on identical code.

Nothing to do on this PR for it.

_— Claude Opus 5_

resources/blob.ts
kriszyp: Addressed in 9e0d204a2 — thanks, this was a real regression and the measured once-guard evidence made it straightforward to land correctly.

Implemented as suggested: a 30s unref'd fallback plus the once-guard, rejecting with the original `error` rather than a timeout error so the receive loop's classification still routes correctly.

Two deviations worth flagging:

**Extracted rather than inlined.** The barrier is now `createPendingMarkerBarrier`, exported for the same reason harper-pro exports `abortInFlightBlobsOnClose`: `.mocharc` sets `timeout: 0` and harper's unit tests have no module mocking, so the stalled-write path is unreachable in place. As a helper it is directly testable — four tests covering release-before-settle, write-error reporting, the bound firing on its own, and a late callback not double-releasing.

**One of your points got sharper under measurement, in the other direction.** I mutation-tested my own tests, and `clearTimeout` turns out *not* to be load-bearing — removing it leaves the suite green, because the timer callback's `settled` check already covers it. Dropping the once-guard in `finish`, inverting release/settle, or removing the fallback timer each turn it red. So I deleted the fifth test that claimed to pin `clearTimeout`, and the code comment now says plainly which guard does the work and which is tidiness.

One residual trade recorded in the comment: if the fallback fires, the lock is released while the marker write is still in flight, so a late-completing write can stamp PENDING over a newer writer's bytes. That reads as 503 and the blob-gap machinery re-streams it, whereas a permanently clamped cursor never heals — so the bound is the better failure, but it is a trade rather than a free win.

_— Claude Opus 5_

resources/blob.ts
```diff
@@ -1473,13 +1473,20 @@ function writeBlobWithStream(
...
+						const messageBuffer = Buffer.from(String(error));
+						const header = new Uint8Array(HEADER_SIZE);
+						new DataView(header.buffer).setBigInt64(0, BigInt(messageBuffer.length) | (BigInt(PENDING_TYPE) << 48n));
+						writeFile(filePath, Buffer.concat([header, messageBuffer]), finishPendingMarker);
```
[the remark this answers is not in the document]
kriszyp: Addressed in 9e0d204a2 — thanks, this was a real regression and the measured once-guard evidence made it straightforward to land correctly.

Implemented as suggested: a 30s unref'd fallback plus the once-guard, rejecting with the original `error` rather than a timeout error so the receive loop's classification still routes correctly.

Two deviations worth flagging:

**Extracted rather than inlined.** The barrier is now `createPendingMarkerBarrier`, exported for the same reason harper-pro exports `abortInFlightBlobsOnClose`: `.mocharc` sets `timeout: 0` and harper's unit tests have no module mocking, so the stalled-write path is unreachable in place. As a helper it is directly testable — four tests covering release-before-settle, write-error reporting, the bound firing on its own, and a late callback not double-releasing.

**One of your points got sharper under measurement, in the other direction.** I mutation-tested my own tests, and `clearTimeout` turns out *not* to be load-bearing — removing it leaves the suite green, because the timer callback's `settled` check already covers it. Dropping the once-guard in `finish`, inverting release/settle, or removing the fallback timer each turn it red. So I deleted the fifth test that claimed to pin `clearTimeout`, and the code comment now says plainly which guard does the work and which is tidiness.

One residual trade recorded in the comment: if the fallback fires, the lock is released while the marker write is still in flight, so a late-completing write can stamp PENDING over a newer writer's bytes. That reads as 503 and the blob-gap machinery re-streams it, whereas a permanently clamped cursor never heals — so the bound is the better failure, but it is a trade rather than a free win.

_— Claude Opus 5_

Issue #1827, opened 2026-07-15
Flaky: 'Harper produced no startup output for 60000ms' — recurring integration-test startup hang, 5x in one morning across unrelated PRs


kriszyp:
## Symptom
`Harper produced no startup output for 60000ms` — an integration-test `startHarper` call times out waiting for the child process's startup output, unrelated to the diff of whatever PR it appears on.

## Occurrences (2026-07-15, ~2.5hr window)
- `record-caching.test.ts` (harper#1808)
- `hot-record-read-snapshot.test.ts` (harper#1820)
- `record-caching TTL-eviction cross-worker ghost` (harper#1819)
- `bulk-conditional-mutation.test.ts` (harper#1819, again)
- a uWS HTTP shard on harper#1820, again

5 occurrences across 5 different test files/PRs in one morning, each confirmed unrelated to that PR's own diff.

## Investigation so far
- Checked githubstatus.com: all systems operational, no incidents.
- Checked `.github/workflows/` history: no CI workflow changes on the day(s) this cluster appeared.
- Checked `@harperfast/integration-testing`: harper's lockfile is pinned to 0.6.2; the newer 0.6.3 (published 2026-07-13) only touches the macOS loopback-alias setup script, not the Linux/Windows CI path or `startHarper`'s readiness signal — ruled out as the cause.
- No infra/dependency change found that correlates with the timing. This looks like genuine test-harness flakiness under load, not an environment regression.

## Leading theory
A `harper-1820-fix1` dispatch finding raised a plausible root cause: if `@harperfast/integration-testing`'s `startHarper` resolves before the HTTP listener is actually accepting connections (rather than after), that could explain both this startup-hang class and a separately observed `ECONNREFUSED`-on-first-request pattern from the same morning — one shared-helper fix candidate in the harness package itself.

## Also seen once
`unitTests/resources/crud.test.js:158` ("puts" — "db-write bytes count were recorded in analytics") flaked on harper#1820 — same analytics-flush-timing class previously fixed via harper#1794's `waitForAnalyticsMetric()` helper. This specific assertion may not be covered by that helper.

kriszyp:
[...]

PR #2230, opened 2026-08-19
fix(txn): keep context.transaction usable after its transaction completes

### LINKED ISSUE #2229: context.transaction is null after its transaction completes, breaking the documented mid-handler commit pattern
### Summary

Since 5.2.1, `context.transaction` is set to `null` once the attached transaction completes. That breaks a documented pattern — committing the current transaction mid-handler and then continuing to use the context — with `TypeError: Cannot read properties of null (reading 'commit')`.

The pattern is published in two places:

- `release-notes/v5-lincoln/v5-migration.md`: `await getContext().transaction.commit();`
- 4.5.0 release notes: "By default, transactions can now be reused after calling `transaction.commit()`."

### Reproduction

```js
const context = {};
await SomeTable.get('any-id', context); // serviced by its own transaction(), minted on this context
await context.transaction.commit();     // TypeError on 5.2.1+
```

Any nested static Resource-API call made with a caller-owned context is enough: `Resource.ts` services it via `transaction(context, …)`, which mints a transaction on that context and drives it to a final (`doneWriting`) commit, and `releaseContext()` then nulls the slot the caller is still holding.

### Regression range

Introduced by #2030 (`releaseContext()`, 2026-07-31) — not in v5.2.0, present in v5.2.1 through v5.2.3. Verified by running the same case on both checkouts: on 5.2.0 the slot held the completed, CLOSED transaction and `commit()`/`abort()` on it were harmless no-ops.

`getReadTxn()`'s `if (this.open !== TRANSACTION_STATE.OPEN) return` and `save()`'s post-commit immediate-commit path are byte-identical between 5.2.0 and 5.2.3, so transaction lifetime itself did not change.

### Impact

Any handler that commits its own transaction mid-request and then touches the context again returns a 500. Reads and writes made *through* the context still work (they are serviced by their own transaction), so the failure is confined to code holding the transaction object itself.

### Note on what is not a regression

Writes made with a context whose transaction has completed are each committed immediately and individually — no atomicity, no rollback. That behaves identically on 5.2.0, so it is pre-existing, but it means a handler that commits mid-request cannot roll back what follows.

kriszyp:
Closes #2229.

`releaseContext()` (added in #2030, to stop a long-lived context pinning a completed transaction) set `context.transaction = null`. That introduced a third state for a published slot, and application code cannot interpret it — the next touch throws `Cannot read properties of null (reading 'commit')`.

Committing the current transaction mid-handler and then continuing to use the context is documented in two places: `release-notes/v5-lincoln/v5-migration.md` (`await getContext().transaction.commit()`) and the 4.5.0 notes ("transactions can now be reused after calling `transaction.commit()`"). On 5.2.1+ that pattern 500s. It is what took out central-manager's fabric connect and cluster create/delete on 5.2.3 for any non-super-user, since `getUserPermissions()` commits the caller's transaction mid-request.

A completed transaction now leaves `RELEASED_TRANSACTION` in the slot: one frozen, process-wide completed transaction. `commit()`/`abort()` are no-ops and reads through it see the latest committed state — what the slot did before #2030 — while retention stays O(1) per process rather than O(live contexts), so #2030's measurement is kept. Because the instance is shared, every route that adopts a caller-supplied transaction or context refuses it (`txnForContext`, `transactional()`'s argument normalizer, `Resource.create`'s own argument shift, `transaction()` itself), and `addWrite()`/`setContext()` throw.

Two more callers were broken by the same `null` and are fixed with it: `transaction.commit(context)` (a checkpointing loop that commits every Nth row fails on its **second** checkpoint on 5.2.3 — the intervening write is serviced by its own scope, whose final commit releases the slot), and passing a transaction where a context is expected (`Table.delete(id, context.transaction)`).

## For the human reviewer

1. **The slot is now permanently truthy after first attach, rather than `null`.** The alternative was keeping `null` and changing the docs plus every caller to `context.transaction?.commit()`. That loses: the pattern is published, so the break reaches every app that followed our own docs, and it shipped in a patch. The cost of this direction is that any downstream `if (context.transaction)` flips meaning — I audited every core reader (`transaction.ts`, `Resource.ts`, `Table.ts`'s two `stale` writebacks and `txnForContext`, `HierarchicalNavigableSmallWorld.ts`) and none is truthiness-sensitive, and harper-pro reads `context.transaction` nowhere outside `core/`. Downstream apps and plugins are the residual exposure: `if (context.transaction) context.transaction.addWrite(op)` now enters the branch and throws instead of silently skipping. **Hardest thing to rule on, and the one I most want a second opinion on.**
2. **One shared instance rather than a per-context lightweight closed one.** Sharing is what keeps the memory win. It also means the instance must own no mutable state — hence the small frozen surface rather than a `DatabaseTransaction` subclass, whose inherited `writes` array stayed mutable through `context.transaction.writes.push(...)` and whose `setContext()` could rebind the shared private context. Anything off that surface now fails loudly rather than inheriting behavior that writes through to every other context.
3. **`isReleasedTransaction()` is checked at four call sites rather than being a state on the transaction.** A future adoption site that forgets gets a `TypeError` from the frozen instance — loud, but still a crash. The alternative (brand it, or funnel every adopter through one place) is cheap now and invasive later; I went with the predicate because `Resource.create` keeps its own reference to the argument and reads `context.transaction.startedFrom` back inside the callback, so a funnel-only guard leaves that reference pointing at the frozen object.
4. **"Absent" is standardized as "use the ambient context", not "a fresh empty context".** Every route resolves the placeholder to `undefined` and falls through to `contextStorage.getStore()`. Review flagged that this changes `Resource.create`'s shifted two-argument form, whose isolation on `origin/main` would be lost — so I measured what that form actually did with a real completed transaction in the slot, which is what 5.2.0 had. It resolved to **the caller's own context**, because `DatabaseTransaction` has a `getContext()` and `create` calls it. The private `{}` only ever appeared on 5.2.1–5.2.3, from `context = record || {}` with `null`; it is an artifact of the bug, not behavior to preserve. A shared placeholder cannot carry a back-reference to the context that released it — that is the point of releasing — so the ambient context is the nearest faithful restoration, and it is also what `Table.create(record)` with no second argument already does.
5. **`context.transaction.commit()` stays callable and `transaction.commit(context)` stays a no-op on a completed transaction.** Review pushed twice for the helper to throw instead (a silent no-op can mask a context-lifecycle bug). It threw on 5.2.3 only as a side effect of the `null`, and on 5.2.0 it was a no-op — so keeping the throw would have preserved half the regression, which is how the checkpoint-loop break above was found. Both documented forms now agree.
6. **LMDB is not at parity and the interface comment says so.** `LMDBTransaction` overrides `commit()`/`abort()` and never calls `releaseContext()`, so there the completed transaction stays in the slot — callable, but retained. That gap predates this change; documented rather than closed, since LMDB is deprecated and its LINGERING semantics are load-bearing.
7. **Not in scope, worth knowing:** writes made with a context whose transaction has completed are each committed immediately and individually — no atomicity, no rollback. Identical on 5.2.0, so not part of this regression, but it is why a failed `deleteCluster` can leave a cluster TERMINATED with its instances RUNNING, and it is the real content of central-manager#705. A follow-up makes those writes atomic while the owning scope's commit is still pending.

One claim I made early and had to retract: `txnTime: 0` from the placeholder's `commit()` is **not** a fidelity loss. Measured on both paths — re-committing the real completed transaction also returns `{ txnTime: 0 }`, because the commit that completed it reset its timestamp.

## Verification

Regression range established by running one case on three checkouts — `v5.2.0`, `v5.2.3` (`origin/main`) and this branch: `context.transaction` after a nested Resource-API call, then `commit()`. 5.2.0 leaves the completed CLOSED transaction and the call is a no-op; 5.2.3 throws the production `TypeError`; this branch matches 5.2.0. `getReadTxn()`'s `open !== OPEN` early return and `save()`'s post-commit immediate-commit path, which the original triage blamed, are byte-identical between 5.2.0 and 5.2.3.

Fails-on-base: the tests here fail on unfixed `origin/main` for the right reasons — the two restored `commit in the middle` calls and the ambient handler case with `Cannot read properties of null (reading 'commit')`, the checkpoint loop and the bare-transaction/`create` routes with `No active transaction is available to commit`.

Added: the ambient delivery path (`processLocalTransaction`, the function `serverHandlers.js` calls for every operation, so this is the real operations entry point — only the HTTP transport is absent), the bare-context regression shape, all four adoption routes plus ambient-identity preservation through them, a fresh multi-store chain after release, every route into the shared placeholder's state, and the `transaction.commit()`/`abort()` behavior on absent versus released slots.

Gates, locally: `test:unit:resources` 1593 passing, 0 failing on RocksDB and 1392 passing, 0 failing with `HARPER_STORAGE_ENGINE=lmdb`. `npm run lint:required` and prettier clean. `test:unit:main` and `test:integration:all` could not run in this checkout — long-running Harper processes from other worktrees hold `~/harper/database/system/LOCK`, so both abort on `Resource temporarily unavailable` before any test executes.

Those two ran in CI and passed, which closes the one real gap in the local verification: **42 checks pass, 2 skipped** — all six integration shards on Node 24, Windows, Bun and uWS, plus unit tests on Node 22/24/26. One Next.js adapter leg was cancelled on a 30-minute `Install Playwright browsers` download timeout, before reaching any test; it passed on re-run.

Not verified: MQTT long-lived-context reattach (the scenario motivating #2030 — untested before this change too), and an operations-HTTP or central-manager flow end to end.

Complexity: complicated

<sub>Review-Coverage: authored=unknown; ran=none; rounds=1 @ d746a30801ea</sub>

<sub>Human-Review-Need: 4 @ d746a30801ea</sub>

### DIFF
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -1196,14 +1196,12 @@ export class ImmediateTransaction extends DatabaseTransaction {
  * What `context.transaction` holds once its transaction has completed and released the back-reference
  * (see releaseContext()). `commit()`/`abort()` are no-ops and reads through it see the latest
  * committed state, so the documented `getContext().transaction.commit()` pattern stays callable after
- * completion; `txnTime: 0` is the one thing it cannot report faithfully, since nothing is staged and no
- * timestamp was taken.
+ * completion, reporting the same `txnTime: 0` that re-committing the completed transaction itself did
+ * (its own timestamp is reset by the commit that completed it).
  *
- * Deliberately NOT a DatabaseTransaction subclass, and deliberately not extensible. It is one
- * process-wide instance shared by every released context — so it must own no mutable state at all, and
- * anything not on this exact surface must fail loudly rather than inherit behavior that would write
- * through to every other context. That is also why the surface is this small: these are the only
- * members any core reader of `context.transaction` touches (`txnForContext` maps it to absent first).
+ * Not a DatabaseTransaction subclass and not extensible: one process-wide instance shared by every
+ * released context owns no mutable state, and anything off this surface fails loudly rather than
+ * inheriting behavior that would write through to every other context.
  */
 const RELEASED_TRANSACTION_SURFACE = {
 	open: TRANSACTION_STATE.CLOSED,
@@ -1235,6 +1233,15 @@ const RELEASED_TRANSACTION_SURFACE = {
 };
 Object.freeze(RELEASED_TRANSACTION_SURFACE);
 export const RELEASED_TRANSACTION = RELEASED_TRANSACTION_SURFACE as unknown as DatabaseTransaction;
+
+/**
+ * The placeholder means "this context has no transaction". Every reader that would otherwise act on the
+ * value — claim it for a store, adopt it as a context, treat it as data — must ask first, or it operates
+ * on the one instance every released context shares.
+ */
+export function isReleasedTransaction(value: unknown): boolean {
+	return value === RELEASED_TRANSACTION;
+}
 
 let timer;
 
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -1193,24 +1193,13 @@ export class ImmediateTransaction extends DatabaseTransaction {
 }
 
 /**
- * What `context.transaction` holds once the transaction that was attached to it has completed — its
- * owning scope's final commit (`doneWriting`) or an abort — and released the back-reference (see
- * releaseContext()). One frozen, process-wide instance, so a long-lived context (an MQTT
- * subscription context, an ambient operation context) retains nothing measurable, which is the
- * whole point of releasing.
- *
- * A `null` slot broke the documented pattern of committing the current transaction mid-handler and
- * then continuing to use the context (`await getContext().transaction.commit()`, v5-migration docs;
- * "transactions can now be reused after calling transaction.commit()", 4.5.0 release notes): the
- * next `context.transaction.<anything>` threw `Cannot read properties of null`. Before the release
- * existed the slot held the completed, CLOSED transaction and those calls were harmless no-ops —
- * this restores exactly that observable behavior without retaining the transaction.
- *
- * Behaves like any CLOSED transaction: commit/abort are no-ops (nothing is staged here — this
- * instance is shared, so txnForContext() refuses to claim it and addWrite() throws rather than
- * quietly staging onto process-wide state), and reads through it see the latest committed state
- * with no snapshot pinned. Frozen so that a path which does try to claim it fails loudly instead of
- * corrupting every other context's view.
+ * What `context.transaction` holds once its transaction has completed and released the back-reference
+ * (see releaseContext()). One frozen, process-wide instance: the slot stays usable — the documented
+ * `getContext().transaction.commit()` pattern must not throw after completion — without any context
+ * retaining a finished transaction. `txnTime: 0` is the one fidelity loss versus re-committing the
+ * real completed transaction, which reported its own last timestamp; nothing is staged here either way.
+ * Frozen because it is shared: a path that tries to claim or mutate it must fail loudly rather than
+ * corrupt every other context's view.
  */
 class ReleasedTransaction extends DatabaseTransaction {
 	constructor() {
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -517,9 +517,9 @@ export class DatabaseTransaction implements Transaction {
 	 * alive (a fresh write on the same context must not join a DIFFERENT, already-replayed
 	 * transaction) until doneReadTxn() drains the last one, so the release is deferred to there.
 	 *
-	 * Leaves RELEASED_TRANSACTION in the slot rather than `null` (which broke the documented
-	 * post-commit `getContext().transaction.commit()` pattern) or `delete` (which would repeatedly
-	 * force a long-lived, hot context into V8's dictionary-mode property storage).
+	 * Leaves RELEASED_TRANSACTION in the slot: the slot must stay callable (see that constant), and it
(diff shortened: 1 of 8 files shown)


kriszyp:
The red `Integration Tests 5/6 (uWS HTTP)` is [#1827](https://github.com/HarperFast/harper/issues/1827) (`Flaky: "Harper produced no startup output" — recurring integration-test startup hang`), not this PR. Evidence:

- `record-caching-ttl-eviction.test.ts` failed with `HarperStartupError: Harper produced no startup output for 150000ms` — Harper never started; no assertion failed.
- The identical failure mode occurs on `main` (e.g. run [32182927071](https://github.com/HarperFast/harper/actions/runs/32182927071), same 150000ms startup hang).
- This same SHA passed all six shards, including 5/6 uWS, on the first run of this PR.

Re-run in flight. Every other check is green: 42 pass / 2 skipped, covering all six integration shards on Node 24, Windows, Bun and uWS, plus unit tests on Node 22/24/26.

— Claude Opus 5

kriszyp:
Follow-up filed: [#2241](https://github.com/HarperFast/harper/issues/2241) — collapse the `ImmediateTransaction` / `RELEASED_TRANSACTION` surface.

Kris asked how the placeholder this PR adds differs from the `ImmediateTransaction` that `txnForContext` already mints, since both mean "no owning transaction scope". Fair question — they differ in three load-bearing properties (`open` OPEN vs CLOSED, store-bound vs store-less, per-context vs per-process and frozen), but a store-less immutable `ImmediateTransaction` would be the placeholder, so the redundancy is real.

Not folded into this PR: the released marker started as a frozen `DatabaseTransaction` subclass here and was replaced in review for concrete reasons (shallow freeze left `writes` mutable, `#private` fields unaffected so `setContext()` could rebind the shared context, un-overridden inherited mutators threw opaque `TypeError`s). Collapsing the surface properly means touching every site that recognizes the marker — the argument normalizer, `Resource.create`, `transaction()`, `txnForContext` — which is more than a patch should carry. #2241 has the constraints written down.

— Claude Opus 5

Issue #2231, opened 2026-08-19
Blind-write transactions are invisible to the long-transaction monitor (never added to trackedTxns)


kriszyp:
## Summary

A transaction that acquires its native handle through a blind write — a write with no preceding read — is never registered with `trackedTxns`, so the long-transaction monitor cannot see it at all. It is never decayed, never bounded, and never reaped at `storage.maxTransactionOpenTime`. #1407's guarantee silently does not apply to this class of transaction.

This was split out of #2224 (PR #2226) deliberately: that change repairs the per-handle ownership bookkeeping, while this one changes monitor policy, and the two have very different blast radii.

## Why it is not simply "add `trackedTxns.add` to the adopt branch"

`trackedTxns` is doing two jobs at once:

1. **The monitor's root set.** `startMonitoringTxns()` iterates every member *independently*, while `chainStillActive()` only looks **downstream** through `next`. Registering an adopted child link would make it its own timeout root: with head A and blind-write child B, B's timeout can expire while A is still active, and B alone gets force-committed or poisoned — the monitor never starting from the head. That is a chain-atomicity change, not bookkeeping.
2. **The read-queue-depth metric's source** (`getTransactionQueueDepths`), so adding write handles silently redefines a published metric.

A fix should register a logical root (or introduce a separate logical-transaction registry) and walk the chain on expiry, keeping the metric's meaning intact.

## Reachability

`getReadTxn()` returns an already-adopted handle before it reaches `trackedTxns.add`, so a blind-write transaction stays untracked **even after it is read**. Pinned by a test in PR #2226 (`is never registered with the long-transaction monitor, even once it is read`) so the current contract is explicit rather than incidental.

Concrete shapes this leaves unbounded:

- A handler that blind-writes (`invalidate`/`publish`) and then idles holds its staged write intents for the whole handler duration; other writers' coordinated-retry commits park on them (the #2001 shape) with no monitor to reap it.
- After a commit with outstanding iterators, a CLOSED transaction retaining `this.transaction` is force-released only by `releaseReadTxn()`, which the monitor calls only for tracked transactions. A blind-write-first transaction whose client stops consuming the iterator keeps its read snapshot pinned until the iterator drains or the process restarts; a read-first transaction in the same shape is released.

## Suggested tests before enabling

Two-database chain (one link's timeout expiring must not break chain atomicity), a replay-boundary case (`directCommitSync()` grouping must not be split by a monitor-driven commit), and read-queue-depth metric semantics.

[...]

PR #2047, opened 2026-08-01
Abort a request-scoped transaction promptly when its client disconnects

kriszyp:
## What

A client that disconnects mid-request used to leave its request-scoped transaction open — holding
staged writes and native RocksDB write intents — until the handler's promise happened to settle, or
the 30–60s long-transaction monitor swept it. harper#2001 traced a live incident to exactly this: an
orphan held past a disconnect wedged a later, conflicting commit 27 seconds later, wedging that
worker thread's write path until restart.

`resources/transaction.ts`'s `transaction()` now listens on `context.signal` for the lifetime of its
callback and, on a client disconnect, poisons and aborts the transaction immediately — the same
two-pass poison-then-abort shape the long-transaction monitor already used for `abortDueToTimeout()`,
now unified into one `abortAndPoison(reason)`.

## Why this shape

- **Gated exactly like the monitor** (`hasPendingWrites() && !sourceApply && !isReplay`): a
  disconnect on a read-only transaction with live iterators (a large `search()`/export) must not
  free the native handle out from under a still-streaming consumer, and a `sourceApply`/`isReplay`
  transaction (replication/caching-source apply, no resume path — harper-pro#348) must never be
  poisoned.
- **No sticky `signal.aborted` fast path**: only a *live* abort event during a transaction's own
  callback poisons it, not a later `transaction()` call that happens to share the same signal (ALS
  inheritance, a spread-copied context).
- **Poisoning is joined, not replaced**: `Resource.ts`'s transaction-reuse guard treats
  `disconnected` the same as the existing `timedOut` case, so a second write through the normal
  Resource/Table API can't silently land on a fresh, unpoisoned transaction after the first write's
  transaction was poisoned.
- Two supporting gaps closed along the way: a write-only transaction (no prior `getReadTxn()`) never
  released its native handle on `abort()`; and uWS-served WebSocket connections never had a working
  disconnect signal at all (`UwsRequest.signal` fell back to a controller nothing could fire).

## The two lifecycle boundaries this PR makes explicit

Review of the first shape found that "retain the handle while iterators own it" had no terminating
condition, and that a disconnect landing inside a commit window was ambiguous. Both are now boundaries
rather than assumptions.

**Transaction-owned iterators are tracked, and closed when they become unreachable.** The only thing
that returns a read reference is an `onDone()` that an abandoned request leaves nobody to call, so
[...]

PR #2232, opened 2026-08-19
Correct blind-write transaction bookkeeping and bring those transactions under the long-transaction monitor

### LINKED ISSUE #2224: save() adopts a read handle without the per-handle bookkeeping getReadTxn() establishes (readTxnsUsed becomes NaN)
## Summary

`DatabaseTransaction.getReadTxn()` establishes three pieces of per-handle bookkeeping together when it opens a read handle (`resources/DatabaseTransaction.ts`):

```ts
this.readTxnsUsed = 1;
this.baseReadRefConsumed = false;
if (DEBUG_LONG_TXNS) this.stackTraces = [new StartedTransaction()];
...
trackedTxns.add(this);
```

`save()` also assigns `this.transaction` — the branch that opens a handle for a write when the transaction has none — and establishes none of them. Anything that later reads through that adopted handle therefore operates on unseeded bookkeeping.

The `stackTraces` half of this surfaced as #2222 (a `TypeError` that 500'd every affected `search()` when `storage.debugLongTransactions` was on) and is guarded at the read site in #2225. The refcount half is still there, is independent of that flag, and is on `main` today.

## The refcount half

`useReadTxn()` does `this.readTxnsUsed++`. On an adopted handle `readTxnsUsed` has never been assigned, so the increment is `undefined++` → `NaN`, and every subsequent comparison against it is false:

```js
const txn = new DatabaseTransaction();
txn.transaction = { openTimer: 0 };   // what save()'s adopt branch assigns
txn.useReadTxn();                     // readTxnsUsed: undefined -> NaN
txn.doneReadTxn();                    // --NaN === 0 is false; handle is never aborted
```

Observed directly (unit harness, 5.2.3): `initial readTxnsUsed = undefined` → `after useReadTxn = NaN` → `after doneReadTxn = NaN, transaction still held = true`.

Consequences, all from `NaN` failing every `> 0` / `=== 0` test:

- `doneReadTxn()` never reaches zero, so the native read handle is never `abort()`ed by the iterator-drain path.
- `abort()`'s `while (this.readTxnsUsed > 0) this.doneReadTxn()` exits immediately, so an aborting request leaves the read snapshot pinned until GC finalizes the handle — blocking compaction for that long.
- The adopted handle is also never `trackedTxns.add`ed, so the long-transaction monitor never sees a write-first transaction at all: it can neither bound its snapshot nor abort it for exceeding `maxTransactionOpenTime`.

`commit()` is *not* affected: its `readTxnsUsed--` then `> 0` test also goes false on `NaN`, which takes the normal commit path.

## Reachability

Latent as far as I could demonstrate — I could not construct a request that reaches the adopt branch:

- `put()` reads the existing entry first, which runs `getReadTxn()`'s fresh branch and seeds `readTxnsUsed = 1`, so an ordinary write-then-search transaction is correctly counted (verified: `readTxnsUsed = 1` after `put`, still `1` after a `search`).
- After a search drains, `doneReadTxn()` decrements only to the base reference (`1`), never to `0`, so `this.transaction` stays set and `save()` never adopts.

So the adopt branch needs a transaction that writes without any prior read, or one whose base reference was already consumed by a commit. Whether a real path does that (blind writes, publish, a post-commit write) is exactly what wants checking before deciding how to fix it.

## Suggested direction

Seed the bookkeeping where the handle is adopted, so "handle exists ⇒ bookkeeping exists" holds however the handle arrived, instead of guarding each consumer. That is deliberately not a drive-by: it puts write-first transactions under the long-transaction monitor for the first time (they become abortable per #1407) and changes what `doneReadTxn()`'s refcount means for a handle whose base reference was never taken, so it wants an owner who can rule on both.

kriszyp:
`save()`'s adopt branch — the path every blind write takes, meaning any write with no preceding read: `invalidate`, `publish`, crash-recovery replay, reload markers — installed a native handle into `this.transaction` without the per-handle bookkeeping `getReadTxn()`'s fresh-handle branch establishes (`readTxnsUsed = 1`, the base reference `commit()` consumes exactly once, plus `baseReadRefConsumed`). `undefined++` is `NaN`, and every `NaN > 0` test is false, so the count was wrong for every consumer that reads it: `commit()`, `doneReadTxn()`, `disregardReadTxn()` and `releaseContext()`, not only `abort()`.

Separately, `trackedTxns.add` runs only in that same fresh-handle branch, and `getReadTxn()` returns an already-adopted handle before reaching it — so a transaction that never read was invisible to the long-transaction monitor for its whole life. #1407's abort-on-over-time did not apply to it: it held write intents with nothing to reap it, and peers' coordinated-retry commits parked on them.

Fixes #2224 and #2231.

> **On overlap with `main`:** the symptom this PR was originally titled for — `abort()` leaving the native handle alive — was fixed on `main` in 4fbe87d95 while this was in review, by making `abort()` release a still-live handle regardless of the count. That is the symptom-level repair; this is the state-level one, and the two are compatible (with the count correct, main's `if (this.transaction) this.releaseReadTxn()` is a no-op on the single-link path, and it still covers paths this does not). Measured against **current** `main`, 19 of this branch's 22 tests fail and 3 pass — one of the 3 being the abort case main now covers.

## For the human reviewer

1. **Monitor supervision is in, per Kris's ruling** — it was written up here as deferred to #2231, and that call was overridden. It is contained the way the planning gate demanded: a **separate** `supervisedWriteRoots` registry (so `trackedTxns`' per-link membership, the read-queue-depth metric, and the `setTxnExpiration()` test seam are all untouched, with no counter needed); one entry per **logical transaction**, keyed on the chain root via a back-pointer set at the single chain-creation site, because supervising a link would make it its own timeout root — force-committable or poisonable while its head is still active, which is the multi-store atomicity #1407 protects; and **crash-recovery replay excluded**, since it is synchronous, already stall- and wall-clock-bounded, and commits at timestamp boundaries a monitor-driven commit could split. **The behaviour change to weigh: a blind-write transaction that runs past `storage.maxTransactionOpenTime` is now reaped where it previously ran unbounded.** `sourceApply` keeps its existing force-commit exemption.
2. **`disregardReadTxn()` will not release a handle carrying staged writes, and its count is clamped.** `commit()` skips re-staging a write it has marked `saved`, so aborting that batch mid-request drops the write (measured: `{"v":"fresh"}` vs `null`). An earlier revision had `save()` take a `readTxnRefCount` reference instead; I removed it after measuring that the write path already takes one, so the release could never fire that way — it was guarding a path I could not demonstrate.
3. **`detachOwnedTransaction()` zeroes `readTxnRefCount`, which no prior release site did.** Not resetting (the old behaviour) means the count only grows, so the early-snapshot-release optimisation silently stops firing for the rest of that transaction's life — a held snapshot blocking compaction and blob reclamation. Resetting means an in-flight `disregardReadTxn()` can land on the next handle; the harmful half of that is blocked by (2), leaving an early release of a read-only handle with no outstanding iterator. **Not fixed, and worth your view:** the invariant-level answer is a generation stamp on the handle, compared on release, rather than a clamp. I judged that a separate change rather than growing this one further.
4. **Write supervision has an unconditional exit.** Membership otherwise ends only when a claiming link detaches, and a chained commit throwing synchronously inside `when()`'s success callback never reaches that — leaving a CLOSED, handle-less root enrolled, drawing a misleading "read iterators held a committed transaction's snapshot" warning every tick and pinning the chain past GC. The monitor now evicts such a member outright.

## Verification

- **`unitTests/resources/adoptedHandleBookkeeping.test.js`, 22 cases**, RocksDB-only (LMDB keeps its own counter in `LMDBTransaction` and never adopts one of these handles). **19 fail against current `origin/main`.** Every guard was falsified individually — remove the staged-write gate, the clamp, the `readTxnRefCount` reset, the root-timeout arming, the chain-wide supervision release, the chain abort walk, the split abort guards, or supervision itself, and a specific test goes red.
- **Four tests in here passed for the wrong reason at some point** and were rewritten only because falsification caught them: the blind-write fixture (invalidating *by id* loads the record first, exercising the wrong branch), the chained-reap case (`getResource({id: null})` does not claim the head, so no chain existed), the disregard case, and the supervision-eviction case (its own cleanup evicted the root before the assertion ran). Two of them also leaked live handles into the shared mocha process and turned 17 unrelated blob-reclamation tests red until bisected. The fixture now asserts no handle exists before the write, the chained cases assert a link was created, and every transaction the suite opens is torn down in `afterEach`.
- `npx mocha "unitTests/resources/**"` — **1610 passing, 0 failing** (RocksDB) and **1386 passing, 0 failing** under `HARPER_STORAGE_ENGINE=lmdb`. Baseline on the merge base is 1584.
- `npm run test:integration` over `integrationTests/database/**` and `integrationTests/resources/**` — **496 passing, 0 failing**, including the suites that run with a low `maxTransactionOpenTime`, where newly-supervised transactions would surface.
- `npm run typecheck`, `npm run lint:required`, `prettier --check` clean. `npm run test:unit:main` cannot run on this machine — a stale mocha process holds `~/harper/database/system/LOCK` — so CI is the gate for it.
- **Review:** eleven rounds. The Harper-domain leg was rate-limited for the first five and, once available, found four things three other lenses had missed — including that judgment call 2's earlier form was unnecessary. Chris's review corrected a native contract both comments stated backwards (measured: rocksdb-js tolerates abort-after-abort; abort-after-**commit** is what throws) and mutation-tested the `readTxnRefCount` reset as uncovered; both are fixed and pinned. Two model claims I refuted by code trace rather than acting on, both then dropped by the graded leg.
- `33a9aeb27` and `fe0088d4c` on this branch are Kris's, with GPT-5 Codex — a constant-time form of the staged-write guard, and regression coverage. My later work was rebased onto them rather than force-pushed over.

Complexity: complicated

<sub>Review-Coverage: authored=unknown; ran=none; rounds=1 @ 1b251a21b8f8</sub>

<sub>Human-Review-Need: 4 @ 1b251a21b8f8</sub>

### DIFF
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -400,14 +400,12 @@ export class DatabaseTransaction implements Transaction {
 		// snapshot that blocks compaction. Only applied when creating the transaction fresh; an
 		// already-open transaction keeps whatever snapshot mode it was created with.
 		// `coordinatedRetry` signals IsBusy write conflicts as RETRY_NOW rather than ERR_BUSY.
-		this.transaction = new RocksTransaction(this.db.store, { coordinatedRetry: true, disableSnapshot });
+		this.attachOwnedTransaction(new RocksTransaction(this.db.store, { coordinatedRetry: true, disableSnapshot }));
 
 		if (this.timestamp) {
 			this.transaction.setTimestamp(this.timestamp);
 		}
 
-		this.readTxnsUsed = 1;
-		this.baseReadRefConsumed = false; // fresh handle, fresh base reference for commit() to consume
 		if (DEBUG_LONG_TXNS) {
 			this.stackTraces = [new StartedTransaction()];
 		}
@@ -415,6 +413,22 @@ export class DatabaseTransaction implements Transaction {
 		trackedTxns.add(this);
 		if (trackedTxns.size > readTxnQueueDepthHighWater) readTxnQueueDepthHighWater = trackedTxns.size;
 		return this.transaction;
+	}
+
+	// Monitor state is not ownership state: it stays with `trackedTxns.add` in getReadTxn().
+	private attachOwnedTransaction(transaction: RocksTransactionWithRetry): void {
+		this.transaction = transaction;
+		this.readTxnsUsed = 1;
+		this.baseReadRefConsumed = false;
+	}
+
+	private detachOwnedTransaction(): RocksTransactionWithRetry | null {
+		const transaction = this.transaction;
+		trackedTxns.delete(this);
+		this.transaction = null;
+		this.readTxnsUsed = 0;
+		this.readTxnRefCount = 0;
+		return transaction;
 	}
 
 	useReadTxn(disableSnapshot?: boolean) {
@@ -432,9 +446,16 @@ export class DatabaseTransaction implements Transaction {
 			// transaction (see the outstanding-iterators branch in commit()), so aborting it here
 			// discards nothing — the replay re-staged the writes AND their audit/txn-log entries
 			// into its own transaction; this handle's never-committed log batch dies with it.
-			trackedTxns.delete(this);
-			this.transaction?.abort();
-			this.transaction = null;
+			const transaction = this.detachOwnedTransaction();
+			try {
+				transaction?.abort();
+			} catch (error) {
+				// Contained, not ignored: abort() calls this before it marks the wrapper CLOSED, clears
+				// writes and releases the context. Warn rather than debug — reached from abort()'s drain
+				// loop the handle can still hold write intents, and stalled writers with a clean log is
+				// the worst outcome here.
+				harperLogger.warn?.('Failed to release a transaction’s native handle', error);
+			}
 			this.completeDeferredContextRelease();
 		}
 	}
@@ -446,14 +467,12 @@ export class DatabaseTransaction implements Transaction {
 	 * abort (an in-flight replay commit owns them), only the snapshot's lifetime is enforced.
 	 */
 	releaseReadTxn(): void {
-		trackedTxns.delete(this);
-		this.readTxnsUsed = 0; // doneReadTxn() no-ops for the remaining iterators (guarded on this.transaction)
+		const transaction = this.detachOwnedTransaction();
 		try {
-			this.transaction?.abort();
+			transaction?.abort();
 		} catch (error) {
 			harperLogger.debug?.('releasing timed-out read transaction', error);
 		}
-		this.transaction = null;
 		this.completeDeferredContextRelease();
 	}
 
@@ -470,7 +489,12 @@ export class DatabaseTransaction implements Transaction {
 	}
 
 	disregardReadTxn(): void {
-		if (--this.readTxnRefCount === 0 && this.readTxnsUsed === 1) {
+		// Never release a handle carrying staged writes: commit() skips re-staging a write it has marked
+		// saved, so aborting the handle here drops it. The count is clamped because every getReadTxn()
+		// increments it but only this releases, so an unpaired call would drive it negative and cancel
+		// out a later handle's references.
+		if (this.readTxnRefCount > 0 && --this.readTxnRefCount === 0 && this.readTxnsUsed === 1) {
+			if (this.writes.some((write) => write)) return;
 			this.doneReadTxn();
 		}
 	}
@@ -658,7 +682,7 @@ export class DatabaseTransaction implements Transaction {
 				harperLogger.warn?.('Created new transaction in save, but the store does match existing store', transaction.id);
 			}
 			if (this.open === TRANSACTION_STATE.OPEN) {
-				this.transaction = transaction;
+				this.attachOwnedTransaction(transaction);
 			} else {
 				// if it is closed, we have to immediately commit, using our immediate transaction
 				immediateCommit = true;
@@ -810,8 +834,7 @@ export class DatabaseTransaction implements Transaction {
 					}
 				} else {
 					// no more reads need to be performed, just commit/abort based if there are any writes
-					trackedTxns.delete(this);
-					this.transaction = null; // clear transaction so any further operations operate immediately
+					this.detachOwnedTransaction(); // any further operations operate immediately
 					if (transaction) {
 						this.writes = this.writes.filter((write) => write); // filter out removed entries
 						if (this.writes.length > 0) {
@@ -1074,22 +1097,19 @@ export class DatabaseTransaction implements Transaction {
 	 * first, then abort each link's native transaction and release its DatabaseTransaction-level
 	 * resources. Two passes (mirroring abortDueToTimeout) so a throw while aborting one link can't leave
 	 * later links (this.next) holding native handles / read snapshots until GC. `headTransaction` is the
-	 * head link's native transaction, which commit() detached to a local before this point
-	 * (this.transaction is already null), so it is aborted directly; every other link still owns its
-	 * native transaction on `txn.transaction`.
+	 * head link's native transaction, which commit() detached to a local before this point, so it is
+	 * aborted directly; every other link still owns its native transaction on `txn.transaction`. The
+	 * head can still hold a retained read handle here (commit()'s outstanding-iterators branch), which
+	 * its iterators release.
 	 */
 	abortChainAfterRetries(headTransaction: RocksTransaction): void {
 		for (let txn: DatabaseTransaction = this; txn; txn = txn.next) {
 			txn.open = TRANSACTION_STATE.CLOSED;
 		}
 		for (let txn: DatabaseTransaction = this; txn; txn = txn.next) {
-			const nativeTxn = txn === this ? headTransaction : txn.transaction;
-			// Clear the native handle and read-snapshot bookkeeping so the abort() below only performs
-			// non-native cleanup (blobs, writes) and can't double-abort (RocksTransaction.abort() throws
-			// on an already-aborted handle) or spin abort()'s doneReadTxn loop on a nulled handle.
-			txn.transaction = null;
-			txn.readTxnsUsed = 0;
-			trackedTxns.delete(txn);
+			// Detach first so the abort() below performs only non-native cleanup and can't double-abort.
+			const detached = txn.detachOwnedTransaction();
+			const nativeTxn = txn === this ? headTransaction : detached;
 			try {
 				nativeTxn?.abort();
 			} catch (abortError) {
@@ -1142,8 +1162,27 @@ export class DatabaseTransaction implements Transaction {
 		}
 	}
 	directCommitSync(): void {
-		trackedTxns.delete(this);
-		this.transaction?.commitSync();
+		const transaction = this.transaction;
+		try {
+			transaction?.commitSync();
+		} catch (error) {
+			// Still uncommitted and still holding its write intents, and no caller aborts after this
+			// throws. abort() rather than the native abort alone: it reclaims blobs a replayed write
+			// staged.
+			this.detachOwnedTransaction();
+			try {
+				transaction?.abort();
+			} catch (abortError) {
+				harperLogger.debug?.('aborting a transaction whose synchronous commit failed', abortError);
+			}
+			try {
+				this.abort();
+			} catch (abortError) {
+				harperLogger.debug?.('cleaning up after a failed synchronous commit', abortError);
+			}
+			throw error;
+		}
+		this.detachOwnedTransaction();
 	}
 	getContext() {
 		return this.#context;
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -436,10 +436,19 @@ export class DatabaseTransaction implements Transaction {
 	 * one. Membership is keyed on the root but claimed per link, so removing it on any link's detach
 	 * would unsupervise a logical transaction still holding writes elsewhere in the chain.
 	 */
-	/** Drop the whole chain's supervision, however its links got there. */
+	/**
+	 * Give up on the whole chain: release any handle its links still hold, then drop the supervision
+	 * that was the only remaining way to find them. Clearing the bookkeeping alone would strand a live
+	 * handle in neither registry — the chained-commit throw this exists for is exactly the case where a
+	 * link never reached its own detach. Snapshots only, matching the CLOSED branch that calls this:
+	 * staged writes may be riding an in-flight replay commit and are not the monitor's to drop.
+	 */
 	dropWriteSupervision(): void {
 		const root = this.root ?? this;
-		for (let link: DatabaseTransaction = root; link; link = link.next) link.writeSupervised = false;
+		for (let link: DatabaseTransaction = root; link; link = link.next) {
+			if (link.transaction) link.releaseReadTxn(); // detaches, which clears this link's own claim
+			link.writeSupervised = false;
+		}
 		supervisedWriteRoots.delete(root);
 	}
 
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -1069,10 +1069,7 @@ export class DatabaseTransaction implements Transaction {
 	}
 	abort(): void {
 		while (this.readTxnsUsed > 0) this.doneReadTxn(); // release the read snapshot when we abort, we assume we don't need it
-		// A write-only transaction never took a read reference (getReadTxn was never called), so the loop
-		// above releases nothing even though save() created a native handle; release it here instead of
-		// leaking the handle and its snapshot until GC. abortChainAfterRetries() detaches the handle
-		// before calling abort(), so this is a no-op there rather than a double-abort.
+		// Defensively release any native handle whose reference bookkeeping was already consumed.
 		if (this.transaction) this.releaseReadTxn();
 		this.open = TRANSACTION_STATE.CLOSED;
 		this.drainCompletions();
@@ -1097,21 +1094,29 @@ export class DatabaseTransaction implements Transaction {
 	 * later links (this.next) holding native handles / read snapshots until GC. `headTransaction` is the
 	 * head link's native transaction, which commit() detached to a local before this point, so it is
 	 * aborted directly; every other link still owns its native transaction on `txn.transaction`. The
-	 * head can still hold a retained read handle here (commit()'s outstanding-iterators branch), which
-	 * its iterators release.
+	 * head can also own a retained read handle from commit()'s outstanding-iterators branch, separate
+	 * from the replay transaction; retry exhaustion aborts both.
 	 */
 	abortChainAfterRetries(headTransaction: RocksTransaction): void {
 		for (let txn: DatabaseTransaction = this; txn; txn = txn.next) {
 			txn.open = TRANSACTION_STATE.CLOSED;
 		}
 		for (let txn: DatabaseTransaction = this; txn; txn = txn.next) {
-			// Detach first so the abort() below performs only non-native cleanup and can't double-abort.
 			const detached = txn.detachOwnedTransaction();
-			const nativeTxn = txn === this ? headTransaction : detached;
+			const committingTransaction = txn === this ? headTransaction : detached;
 			try {
-				nativeTxn?.abort();
+				committingTransaction?.abort();
 			} catch (abortError) {
 				harperLogger.debug?.('aborting conflicted transaction in chain after exhausting retries', abortError);
+			}
+			// With outstanding iterators, the head owns a retained read handle while the retry commits
+			// through a separate replay handle. Both must be aborted after retry exhaustion.
+			if (txn === this && detached && detached !== committingTransaction) {
+				try {
+					detached.abort();
+				} catch (abortError) {
+					harperLogger.debug?.('aborting retained read transaction after exhausting retries', abortError);
+				}
 			}
 			try {
 				// abort() synchronously walks savedBlobs and can call write.store.getEntry(), which can throw
--- a/resources/DatabaseTransaction.ts
+++ b/resources/DatabaseTransaction.ts
@@ -1073,19 +1073,21 @@ export class DatabaseTransaction implements Transaction {
 		if (this.transaction) this.releaseReadTxn();
 		this.open = TRANSACTION_STATE.CLOSED;
 		this.drainCompletions();
-		for (const write of this.writes) {
-			if (write?.savedBlobs)
-				cleanupUnusedBlobs(write.savedBlobs, collectRetainedFileIds(write.store.getEntry(write.key)?.value));
+		try {
+			for (const write of this.writes) {
+				if (write?.savedBlobs)
+					cleanupUnusedBlobs(write.savedBlobs, collectRetainedFileIds(write.store.getEntry(write.key)?.value));
+			}
+		} finally {
+			this.clearWrites();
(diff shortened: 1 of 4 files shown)


unitTests/resources/adoptedHandleBookkeeping.test.js
```diff
@@ -0,0 +1,183 @@
+/**
+ * harper#2224 — save()'s adopt branch, taken by every blind write, seeded none of the per-handle
+ * bookkeeping getReadTxn() establishes, so `undefined++` made readTxnsUsed NaN and abort() never
+ * released the native transaction or its write intents.
+ */
+const assert = require('node:assert');
+const { setupTestDBPath } = require('../testUtils');
```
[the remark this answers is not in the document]
kriszyp: Leaving this as-is, deliberately. The file-extension rule is a TypeStrip requirement for the ESM source modules — it is what lets Node resolve `./module.ts` without a bundler. These unit tests are CommonJS and load through `require`, where the extension is not needed, and the surrounding convention is unanimous: **127 of 127** files under `unitTests/resources/` use `require('../testUtils')`, and across relative requires in that directory 134 omit the extension against 6 that include it.

Adding it here would make this the odd file out rather than the compliant one. If we do want the rule to cover CJS test requires too, that is worth doing as one sweep over all 134 sites plus a lint rule to hold it — happy to file that separately if you think it is worth it.

— KrAIs (Claude Opus 5)

resources/DatabaseTransaction.ts
```diff
@@ -1093,8 +1115,24 @@ export class DatabaseTransaction implements Transaction {
...
+			// after this throws — one drops the wrapper outright — so release it here. abort() rather than
+			// the native abort alone: it also reclaims blobs a replayed write staged, and detaching first
+			// keeps it from double-aborting the handle.
+			this.detachOwnedTransaction();
+			try {
+				transaction?.abort();
+				this.abort();
+			} catch (abortError) {
+				harperLogger.debug?.('aborting a transaction whose synchronous commit failed', abortError);
+			}
```
cb1kenobi, line 1247: **Medium: a throwing native `abort()` here skips `this.abort()` entirely, stranding all the wrapper cleanup**

`transaction?.abort()` and `this.abort()` share one `try`, so a throw from the first swallows the second. Measured on this head (stub handle whose `commitSync()` and `abort()` both throw): `open` stays `OPEN`, `writes` keeps its entry, `savedBlobs` are never passed to `cleanupUnusedBlobs`, and `context.transaction` still pins the wrapper — exactly the "would strand all three" hazard the sibling comment in `doneReadTxn()` (line 453) calls out.

It is reachable because the native contract is the opposite of what the comments assume. Measured against the bundled `@harperfast/rocksdb-js` with a real handle:

| sequence | result |
|---|---|
| `abort()` after `abort()` | OK, no throw |
| `commitSync()` after `abort()` | throws `ERR_ALREADY_ABORTED` |
| `abort()` after that failed `commitSync()` | OK |
| **`abort()` after a *successful* `commitSync()`** | **throws `ERR_ALREADY_COMMITTED`** |

So the throwing case is precisely "the commit landed but `commitSync()` still threw" — which rocksdb-js produces whenever a listener throws from its `finally { notify('aftercommit') }` after the native commit succeeded. Both callers are crash-recovery replay (`replayLogs.ts:172`, `:287`), which catch, log, and drop the wrapper — so the durable damage is orphaned blob files left on disk after a recovery boot.

Suggested fix — give each its own `try` so wrapper cleanup can't be skipped, and keep the original commit error as the one rethrown:

```suggestion
			try {
				transaction?.abort();
			} catch (abortError) {
				harperLogger.debug?.('aborting a transaction whose synchronous commit failed', abortError);
			}
			try {
				this.abort();
			} catch (abortError) {
				harperLogger.debug?.('cleaning up a transaction whose synchronous commit failed', abortError);
			}
```

—
_Generated by Barber AI_

resources/DatabaseTransaction.ts
```diff
@@ -416,6 +414,23 @@ export class DatabaseTransaction implements Transaction {
...
+		this.readTxnsUsed = 1;
+		this.baseReadRefConsumed = false;
+	}
+
+	private detachOwnedTransaction(): RocksTransactionWithRetry | null {
+		const transaction = this.transaction;
+		trackedTxns.delete(this);
+		this.transaction = null;
+		this.readTxnsUsed = 0;
+		this.readTxnRefCount = 0;
```
cb1kenobi, line 471: **Low: the `readTxnRefCount = 0` reset is the one piece of `detachOwnedTransaction()` no test pins**

Mutation-tested this line out (dist-verified, marker count exactly 1) and ran both the new suite and all of `unitTests/resources/**` under RocksDB: **1596 passing, 0 failing** — nothing goes red. Every other detach/attach assignment is covered (removing `this.transaction = null`, `readTxnsUsed = 0`, `trackedTxns.delete`, or the `attachOwnedTransaction` call in `save()` all turn a test red).

That is a gap worth closing, because this reset is the whole reason judgment call 5 unified `commit()`'s no-iterator branch and `releaseReadTxn()` onto the helper — an inline terminal path that skips it leaves a reused wrapper at `readTxnRefCount === 1` against a freshly adopted handle. A one-line addition to the existing "detaches the handle when a synchronous commit succeeds" case covers it:

```suggestion
		this.readTxnRefCount = 0;
```
(no change to this line — the ask is `assert.strictEqual(txn.readTxnRefCount, 0)` alongside the existing `readTxnsUsed` assertion in `adoptedHandleBookkeeping.test.js`.)

—
_Generated by Barber AI_

resources/DatabaseTransaction.ts
```diff
@@ -1034,13 +1058,11 @@ export class DatabaseTransaction implements Transaction {
...
 		for (let txn: DatabaseTransaction = this; txn; txn = txn.next) {
-			const nativeTxn = txn === this ? headTransaction : txn.transaction;
-			// Clear the native handle and read-snapshot bookkeeping so the abort() below only performs
-			// non-native cleanup (blobs, writes) and can't double-abort (RocksTransaction.abort() throws
-			// on an already-aborted handle) or spin abort()'s doneReadTxn loop on a nulled handle.
-			txn.transaction = null;
-			txn.readTxnsUsed = 0;
-			trackedTxns.delete(txn);
+			// Detach first so the abort() below only performs non-native cleanup (blobs, writes) and can't
+			// double-abort (RocksTransaction.abort() throws on an already-aborted handle) or spin abort()'s
```
cb1kenobi: **Low: this comment (and the matching one at line 1125) states a native contract that measurement contradicts**

`RocksTransaction.abort()` does **not** throw on an already-aborted handle — verified against the bundled rocksdb-js with a real handle: `abort()` then `abort()` returns cleanly, as does `abort()` after a *failed* `commitSync()` and after `abandonWrites()`. The one state where it throws is an already-**committed** handle (`ERR_ALREADY_COMMITTED`).

Detaching first is still the right ordering — it keeps `abort()`'s `doneReadTxn` loop from spinning on a nulled handle, which is real. But the stated reason is inverted, and the case that *does* throw is the one the `directCommitSync()` catch block does not survive (see the comment at line 1128). Worth correcting so the next reader reasons from the real contract.

—
_Generated by Barber AI_

resources/DatabaseTransaction.ts
```diff
@@ -416,6 +414,23 @@ export class DatabaseTransaction implements Transaction {
 		return this.transaction;
 	}
 
+	// Handles the field never owns — save()'s immediate commit, commit()'s replay — skip this, and
+	// monitor state stays with `trackedTxns.add` in getReadTxn().
+	private attachOwnedTransaction(transaction: RocksTransactionWithRetry): void {
```
cb1kenobi, line 428: **Note: `main` has moved past this PR's base, and the headline symptom is already fixed there**

This branch is cut from `e92fef18b` (v5.2.3); `origin/main` is now `58f77f5c8`, seven commits further on `DatabaseTransaction.ts`. One of them — `4fbe87d95` *"Correct the abort handle-release note and pin it with a test"* — adds to `abort()`:

```js
if (this.transaction) this.releaseReadTxn();
```

with a comment naming the write-only/blind-write case exactly. Ran this PR's own suite against current `origin/main`: **3 passing, 8 failing**, and the one that *passes* is `releases the adopted handle on abort, and completes wrapper cleanup`. So the "8 of the 10 fail against `origin/main`" line in the description was measured against the stale base, and the abort leak in the title is already closed on `main`.

That does not make this PR redundant — it is the difference between patching the symptom in `abort()` and fixing the `NaN` at its source. On current `main` `readTxnsUsed` is still `undefined` after an adopt, so `doneReadTxn()`, `disregardReadTxn()` and `commit()`'s iterator branch all still mis-evaluate for blind writes; this PR is what makes them correct. Worth rebasing and re-framing the description around that, rather than around the abort leak.

Two things measured on the rebase, both good news:
- **Trial-merge into `58f77f5c8` is clean, and safe.** Merged `abort()` carries both release paths, but the `while (readTxnsUsed > 0) doneReadTxn()` loop — now effective thanks to this PR — detaches first, so `if (this.transaction)` is already false. Merged tree: `unitTests/resources/**` **1600 passing / 0 failing** (RocksDB) and **1385 passing / 0 failing** (LMDB), new suite 11/11.
- **No double-release anywhere.** Probed abort-after-abort, abort-after-commit, `releaseReadTxn()`-then-abort, `abortDueToTimeout()`-then-abort, a two-link chain through both `abortDueToTimeout()` and `abortChainAfterRetries()`, over-draining `doneReadTxn()`, and commit-then-`releaseReadTxn()`, counting native aborts on the real handle. Every path: exactly one native abort, or zero where the handle had already committed. The invariant holds — `this.transaction` is assigned in exactly two places in the file, both inside these helpers.

**One coordination item: #2047 conflicts with this, and carries a third fix for the same bug.** [harper#2047](https://github.com/HarperFast/harper/pull/2047) ("Abort a request-scoped transaction promptly when its client disconnects") is also cut from `e92fef18b` and rewrites `abort()`; merging both produces a real conflict in exactly that block. #2047's side adds its own blind-write release:

```js
if (this.transaction && !hasOpenReadIterator) {
    this.transaction.abort();
    this.transaction = null;   // bypasses detachOwnedTransaction()
}
```

plus a lazy repair in `getReadTxn()`'s already-have-a-handle fast path (`if (this.readTxnsUsed == null) { readTxnsUsed = 1; baseReadRefConsumed = false; trackedTxns.add(this); ... }`). Two consequences worth flagging to whoever resolves it:

1. That bare `this.transaction = null` skips the helper, leaving `readTxnRefCount` stale and a dead `trackedTxns` entry — it re-opens the class of bug this PR closes. Resolve toward `detachOwnedTransaction()`.
2. Once this PR lands, `readTxnsUsed` is never `null` while `this.transaction` is set, so #2047's repair block becomes unreachable — including its `trackedTxns.add(this)`. That is convenient (it silently preserves the monitor-exclusion contract judgment call 2 pins, so the `is never registered with the long-transaction monitor` test keeps passing in either landing order) but it should be deleted rather than left as dead code, and #2047 should know its `trackedTxns.add` will never run.

Also relevant to #2047's open High: its monitor `continue` skips `releaseReadTxn()` for `timedOut`/`disconnected` transactions. With this PR, `abortDueToTimeout()` → `abort()` already releases the handle at poison time (measured: one native abort, `this.transaction === null`), so by the time the monitor skips it there is nothing left to leak. This PR de-fangs that finding.

—
_Generated by Barber AI_

resources/DatabaseTransaction.ts
kriszyp: Already fixed at the time you measured, though not on the head you had — the two aborts were split into separate `try` blocks a few pushes back, and `completes wrapper cleanup when a failed synchronous commit also fails to abort` pins it with exactly your stub shape (both `commitSync()` and `abort()` throwing): it asserts CLOSED, writes cleared, and the context detached, and it goes red if the split is reverted. Sorry for the moving target.

Your measured native contract is the more valuable half and I have taken it — see the reply on the contract thread.

— KrAIs (Claude Opus 5)

resources/DatabaseTransaction.ts
kriszyp: Corrected in b8c65fb77. You are right that the stated reason was inverted, and it was load-bearing for how the next reader reasons about the failure path.

Both comments now say what your measurement shows: detaching first is about keeping `abort()`'s `doneReadTxn` loop off a nulled handle, **not** about avoiding a double abort — rocksdb-js tolerates abort-after-abort, and abort-after-COMMIT is the case that throws. I fixed the pre-existing claim in `abortChainAfterRetries` too, since this change touches that line.

— KrAIs (Claude Opus 5)

resources/DatabaseTransaction.ts
kriszyp: Good catch, and the mutation test is the right way to have found it — a line whose whole purpose is preventing a stale count against a *later* handle is exactly the kind that stays green when deleted.

Pinned in c1360b735 (I see the same assertion arrived on the branch independently; the two agree). The case now seeds `readTxnRefCount = 3` before the synchronous commit and asserts it is back to 0 afterwards, so deleting the reset turns it red — verified by removing the line.

— KrAIs (Claude Opus 5)

resources/DatabaseTransaction.ts
kriszyp: You are right, and I reached the same conclusion independently a few minutes before reading this — the description's fails-on-base line was measured against the stale base and I am rewriting it.

Re-measured against current `origin/main`: **3 passing, 17 failing** of this branch's 20 (the count differs from yours only because the suite has grown since). The one that passes is exactly the one you name, `releases the adopted handle on abort`, closed by 4fbe87d95.

So the title oversells it and will change. What is left is the difference you point at: main makes `abort()` tolerate `readTxnsUsed` being `NaN`, while this makes the count correct in the first place — which `commit()`, `doneReadTxn()`, `disregardReadTxn()` and `releaseContext()` all consume too — plus the monitor supervision from #2231, which Kris asked to fold in.

Worth noting the two fixes interact benignly: with the bookkeeping correct, main's `if (this.transaction) this.releaseReadTxn()` becomes a no-op on the single-link path, since the drain loop has already released. I have left it as defence rather than removing it.

— KrAIs (Claude Opus 5)

resources/DatabaseTransaction.ts
```diff
@@ -417,6 +424,45 @@ export class DatabaseTransaction implements Transaction {
...
+	/**
+	 * Drop this link's supervision claim, and the root's with it once no link in the chain still holds
+	 * one. Membership is keyed on the root but claimed per link, so removing it on any link's detach
+	 * would unsupervise a logical transaction still holding writes elsewhere in the chain.
+	 */
+	/** Drop the whole chain's supervision, however its links got there. */
+	dropWriteSupervision(): void {
+		const root = this.root ?? this;
+		for (let link: DatabaseTransaction = root; link; link = link.next) link.writeSupervised = false;
+		supervisedWriteRoots.delete(root);
```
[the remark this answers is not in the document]
kriszyp: Correct, and this one was mine from the commit right before yours — thank you. Fixed in 9e039cb41.

The eviction traded a *visible* leak for an *invisible* one: before it, a stranded root at least stayed enrolled and warned every tick; after it, clearing the bookkeeping made a live handle unreachable from both registries, so nothing could ever release it.

`dropWriteSupervision()` now releases each link's handle before dropping the enrollment — `releaseReadTxn()` per link, which detaches and so clears that link's claim on the way. Deliberately snapshots only, not `abort()`, to match the CLOSED branch that calls it: staged writes there may be riding an in-flight replay commit and are not the monitor's to drop.

The eviction test now asserts `child.transaction === null` after the tick, and it goes red if the release line is removed — verified both directions.

— KrAIs (Claude Opus 5)

kriszyp:
**CI note: `Integration Tests 1/6 (Windows, Node.js v24)` is red, and I believe it is not this PR — but the evidence is short of proof.**

The failure is `QA-782 stale-read blast radius … [lmdb]`, step 1 (seed 6000+6000) timing out at 300s.

What I checked:

| Evidence | Reading |
|---|---|
| `main` run 32313008082 fails the same shard, same test, same 300s timeout, and the **same two audit-decode RangeErrors** (`position 4`; `recordId length 109 … byteLength 51`) | not introduced here |
| **Zero** `open too long` / over-time aborts anywhere in the failing log | the new write supervision reaped nothing — the mechanism I flagged as this PR's main behaviour risk did not fire |
| Only the **lmdb** variant fails; the **rocksdb** variant of the same test passes in both runs | this PR changes `DatabaseTransaction` (RocksDB); `LMDBTransaction` is untouched |
| `LMDBTransaction` has no `root` property | the one engine-agnostic line here (`transaction.next.root = …` in `Table.ts`) clobbers nothing |

What I am **not** claiming: this shard passed 5 of its last 6 runs on `main` and has now failed twice consecutively here. Two samples cannot separate bad luck from a small increase in likelihood, and I would rather flag that than assert independence I have not earned. If a reviewer wants it settled before merge, the decisive check is a couple of runs of that shard against `main` at the same commit depth on the same runner class.

Everything else in the matrix is green: all Node 24 / Bun / uWS integration shards, the other five Windows shards, all three Unit Test jobs including the LMDB pass, builds, lint, format, Next.js adapters.

— KrAIs (Claude Opus 5)

kriszyp:
Heads-up on a collision with [#2230](https://github.com/HarperFast/harper/pull/2230), found while checking stack compatibility — not a problem with this PR.

#2230 stops `releaseContext()` leaving `null` in `context.transaction` (it broke the documented mid-handler `getContext().transaction.commit()` pattern) and leaves a shared released placeholder there instead. Three cases here assert the old marker:

- `unitTests/resources/adoptedHandleBookkeeping.test.js:112` — `assert.strictEqual(context.transaction, null)`
- the two adjacent wrapper-cleanup cases in the same suite

Whichever of the two merges second needs those three updated to expect `RELEASED_TRANSACTION` — the same edit #2230 already made to `transaction.test.js`, `lingeringWriteCommit.test.js` and `operationContextTransactionLeak.test.js`. Everything else in this PR merged cleanly against it.

Separately: this PR closes the long-transaction-monitor gap that [#2239](https://github.com/HarperFast/harper/pull/2239) would otherwise have had to solve, and `attachOwnedTransaction()` supersedes a `doneReadTxn()` guard that #2239 carries. I verified the merged stack (main + this + #2239) and #2239 stays green once that guard is dropped, so no action needed here.

— Claude Opus 5

kriszyp:
Brought up to date with `main` (which now includes [#2230](https://github.com/HarperFast/harper/pull/2230)) and fixed the collision I flagged above.

**Four assertions, not three** — I missed one in my earlier note: `sourceApplyConflictRetry.test.js:329` has the same shape as the three in `adoptedHandleBookkeeping.test.js`. All four pinned `context.transaction === null`, which #2230 replaces with the shared `RELEASED_TRANSACTION`; their intent (a completed wrapper is not retained) is unchanged. Against merged `main`: **4 failing before, 0 after — 1621 passing**, lint and prettier clean. Merging without this would have turned `test:unit:resources` red on main.

**This is a merge of `main`, not a rebase, and that was deliberate.** A true rebase replays these 11 commits over the moved base, and each conflicts with *this branch's own later revisions* — `detachOwnedTransaction`, the drained-handle log level, the chain-cleanup naming — not with anything on main. Resolving those means re-deciding which of this PR's own iterations wins, which is a call for whoever wrote them, and the answer is already in the branch tip. The net diff against main reconciles with no ambiguity, which is why the merge is clean where the replay is not. If linear history matters here, the `merge: squash` this PR is already marked for gets it at merge time.

No production code changed in this push.

— Claude Opus 5

Issue #734, opened 2026-05-22
Make analytics expiration windows configurable


kriszyp:
## Problem

Analytics expiration values are hard-coded in `resources/analytics.ts` (lines 533–534). They should be exposed as configuration so operators can tune retention without code changes.

🤖 Filed by Claude on behalf of Kris.

kriszyp:
Addressed by https://github.com/HarperFast/harper/pull/719 (feat(analytics): make retention windows configurable).

PR #2233, opened 2026-08-19
Register replication_receiveQueueHighWaterMark so the receive-queue budget can be configured

kriszyp:
Registers `replication_receiveQueueHighWaterMark` in `CONFIG_PARAMS`.

`env.get` resolves only names present in `CONFIG_PARAM_MAP` — `getConfigValue` returns `undefined`
for anything absent — so a config param that isn't registered is silently ignored and its
compiled-in default always applies. That is how `replication_leadingDuplicateSkip` shipped inert
(harper-pro#395, which is why `unitTests/replication/leadingDuplicateSkipConfig.test.mjs` exists),
and it is the current state of four sibling replication knobs (harper-pro#734).

Companion to HarperFast/harper-pro#735, which bounds the replication receive path's inbound
frame queue and reads this param. Without this commit that PR's budget cannot be configured at all:
the default would be permanently in force, an operator watching a worker OOM could neither lower it
nor set `0` to disable, and there would be no error — only the default's behavior. A unit test on
the harper-pro side asserts this registration resolves, so the pair cannot drift back apart.

## For the human reviewer

One line in an enum, but it gates a fix's entire configurability, so the ordering matters: this
needs to land (and be present in whatever core the harper-pro PR pins) before the harper-pro side is
useful. The reviewers of the harper-pro PR flagged exactly this as the release-ordering risk.

Worth knowing separately: `replication_receiveEventHighWaterMark`, `replication_receiveYieldInterval`,
`replication_copyCheckpointRecords` and `replication_copyCheckpointMaxIntervalMs` are all unregistered
today, so they are inert. Filed as harper-pro#734 rather than fixed here — registering them would
make previously-ignored values in existing config files start taking effect, which deserves its own
review rather than riding along with this.

## Verification

`npm run build` clean; the harper-pro unit test asserting
`CONFIG_PARAMS.REPLICATION_RECEIVEQUEUEHIGHWATERMARK === 'replication_receiveQueueHighWaterMark'`
passes against this commit, and the harper-pro integration test confirms a configured budget reaches
the gate (the warn payload carries the configured value, not the default).

### DIFF
--- a/utility/hdbTerms.ts
+++ b/utility/hdbTerms.ts
@@ -658,6 +658,10 @@ export const CONFIG_PARAMS = {
 	REPLICATION_BLOBCONCURRENCY: 'replication_blobConcurrency',
 	REPLICATION_MAXPAYLOAD: 'replication_maxPayload',
 	REPLICATION_RECORDCONCURRENCY: 'replication_recordConcurrency',
+	// Byte ceiling on inbound replication frames accepted but not yet processed. Without it the
+	// receive path's frame queue is unbounded and a receiver slower than its peer grows it at the full
+	// inbound line rate until the worker is OOM-killed (harper#2226, harper-pro#659).
+	REPLICATION_RECEIVEQUEUEHIGHWATERMARK: 'replication_receiveQueueHighWaterMark',
 	REPLICATION_PINGINTERVAL: 'replication_pingInterval',
 	REPLICATION_PINGTIMEOUT: 'replication_pingTimeout',
 	REPLICATION_COPYTIMEOUT: 'replication_copyTimeout',
--- a/unitTests/config/replicationReceiveQueueParam.test.js
+++ b/unitTests/config/replicationReceiveQueueParam.test.js
@@ -1,0 +1,26 @@
+'use strict';
+
+// `env.get` resolves a name only if it is present in CONFIG_PARAM_MAP, which is populated from
+// CONFIG_PARAMS by lowercased name. An unregistered param therefore reads as `undefined` forever: the
+// configured value is silently ignored, its compiled-in default always applies, and set_configuration
+// rejects it as unrecognized. That is how `replication_leadingDuplicateSkip` shipped inert
+// (harper-pro#395), and several sibling replication params are inert today (harper-pro#734).
+//
+// harper-pro's receive-queue budget (harper-pro#735) reads this param, and its own suite asserts the
+// registration — but only against whatever core its submodule pins. This guards the entry here.
+
+const assert = require('node:assert/strict');
+const { CONFIG_PARAMS, CONFIG_PARAM_MAP } = require('#src/utility/hdbTerms');
+
+describe('replication_receiveQueueHighWaterMark registration', () => {
+	it('is registered in CONFIG_PARAMS so env.get can resolve it', () => {
+		assert.strictEqual(CONFIG_PARAMS.REPLICATION_RECEIVEQUEUEHIGHWATERMARK, 'replication_receiveQueueHighWaterMark');
+	});
+
+	it('is reachable through CONFIG_PARAM_MAP, so set_configuration accepts it', () => {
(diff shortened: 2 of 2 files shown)


utility/hdbTerms.ts
```diff
@@ -658,6 +658,10 @@ export const CONFIG_PARAMS = {
 	REPLICATION_BLOBCONCURRENCY: 'replication_blobConcurrency',
 	REPLICATION_MAXPAYLOAD: 'replication_maxPayload',
 	REPLICATION_RECORDCONCURRENCY: 'replication_recordConcurrency',
+	// Byte ceiling on inbound replication frames accepted but not yet processed. Without it the
+	// receive path's frame queue is unbounded and a receiver slower than its peer grows it at the full
+	// inbound line rate until the worker is OOM-killed (harper#2226, harper-pro#659).
+	REPLICATION_RECEIVEQUEUEHIGHWATERMARK: 'replication_receiveQueueHighWaterMark',
```
cb1kenobi, line 664: **Low: no same-repo regression guard for this registration**

The failure mode this PR fixes has shipped before (`replication_leadingDuplicateSkip`, harper-pro#395), and the guard against it today lives in harper-pro (`unitTests/replication/leadingDuplicateSkipConfig.test.mjs`); the companion PR (harper-pro#735) adds the equivalent one-assertion guard for this exact param. But harper itself already has a same-repo precedent for this exact pattern -- `unitTests/resources/models/v1/packaging.test.js:49` asserts `CONFIG_PARAMS.MODELSGATEWAY_ENABLED` resolves through `CONFIG_PARAM_MAP`. A 3-line mirror of that here would guard this entry independently of whether harper-pro's submodule pin is current or its CI runs:

```js
expect(CONFIG_PARAMS.REPLICATION_RECEIVEQUEUEHIGHWATERMARK).to.equal('replication_receiveQueueHighWaterMark');
```

Not blocking -- the companion PR's test covers this once merged -- but cheap insurance given this exact class of bug has shipped inert once already.

---
_Generated by Barber AI_

cb1kenobi, line 664: **Low: harper-pro#734 undercounts -- six `replication_*` params are read-but-unregistered, not four**

Audited every `env.get('replication_...')` / `env.get(CONFIG_PARAMS.REPLICATION_*)` call site in harper-pro's `replication/replicationConnection.ts` against the registered `REPLICATION_*` entries in this file. Two more are silently inert beyond the four harper-pro#734 lists:

- `replication_subscriptionResolveTimeout` -- read via `positiveMsOr(env.get('replication_subscriptionResolveTimeout'), 60000)`.
- `replication_pauseStallTimeout` -- worse than the others: the surrounding comment explicitly tells operators to use it ("Override with `replication_pauseStallTimeout` for clusters with extreme single-transaction sizes"), but it currently does nothing.

harper-pro#734 itself calls for exactly this audit ("Also worth an audit pass: any other `env.get('...')` call ... reading a name that is not in CONFIG_PARAMS") -- this is that pass. Doesn't block this PR, whose own diff is correctly scoped to one param, but worth folding into #734 (or a follow-up) so the eventual sibling-registration pass doesn't miss two of them.

---
_Generated by Barber AI_

kriszyp: Taken — added `unitTests/config/replicationReceiveQueueParam.test.js`, mirroring the `MODELSGATEWAY_ENABLED` assertion you pointed at, and asserting the `CONFIG_PARAM_MAP` lookup too so the `set_configuration` path is covered. Agreed on the reasoning: harper-pro's guard only holds against whatever core its submodule pins.

kriszyp: Confirmed and folded into harper-pro#734 — thank you for running the audit that issue asked for. Verified both names are absent from `utility/hdbTerms.ts`. Agreed that `replication_pauseStallTimeout` is the worst case, since its own comment instructs operators to override it.

PR #2244, opened 2026-08-20
Stabilize transaction-log purge blast-radius coverage

kriszyp:
## Summary

Stabilizes QA-782 while preserving equal incident-calibrated workloads for RocksDB and LMDB wherever both engines run.

- use one [shared 6,000-row range](https://github.com/HarperFast/harper/pull/2244/changes?w=1#diff-5d33f8270677ec51e9d3b33ee1b75d332a4f55151b24b5c57577a9dcd43019d4R93) for both engines; retain main's Windows-only LMDB skip instead of weakening the non-Windows control
- derive seed waves and sample positions safely, including a clamped final wave, while preserving RocksDB's greater-than-16-MiB transaction-log precondition
- fail every dependent phase closed unless seeding and the cutoff timestamp complete successfully
- match the deliberately seeded pre-cutoff `insert` audit witness without a redundant ID predicate
- run the deleted-record subject first, then prove the [same five-surface absence detector](https://github.com/HarperFast/harper/pull/2244/changes?w=1#diff-5d33f8270677ec51e9d3b33ee1b75d332a4f55151b24b5c57577a9dcd43019d4R653) flags a known-present control; intentional detector diagnostics are kept out of the real findings transcript

This remains a test-harness correction, not a product data-loss fix. Completed seeds read correctly on every covered ordinary-read surface in both engine arms.

## Evidence

- [Windows failure: LMDB seed times out, then `k11999` is absent](https://github.com/HarperFast/harper/actions/runs/32331079623/job/96312178349)
- [Windows success: both completed seeds and all read surfaces pass](https://github.com/HarperFast/harper/actions/runs/32312146829/job/96257662959)
- [Repeated prior PR run: all six Windows Node 24 shards passed](https://github.com/HarperFast/harper/actions/runs/32370882792)

## Verification

- `npm run build`
- final targeted QA-782 run: 14/14 across RocksDB and LMDB; each arm seeded DEL=6,000 and KEEP=6,000, and RocksDB produced 21.182 MiB of `.txnlog` data
- `npm run lint:required`
- `npm run format:write` (no changes)
- full integration run: 1,815 passed, 1 failed, 6 cancelled, and 20 skipped; the only failure was the adjacent QA-431 TTL rate-limiter stress assertion
- `test:unit:resources`: 1,688 passed, 16 pending, and 3 unrelated failures in random-access-field defaults/replay structure expectations
- `test:unit:main` is blocked on this host before tests load because unit bootstrap opens an external persisted system database and fails reading its metadata
- independent full review of `801b6d5f8` by Claude, Gemini, and the Harper domain adjudicator; adjudicated severity: none

## For the human reviewer

Please review two local, reversible test-design choices recorded by the independent review: the audit contrast intentionally witnesses the specifically seeded pre-cutoff insert rather than broadening this incident test into complete purge-contract coverage, and the detector control deliberately reuses the exact absence oracle so it proves that reporting path is armed. The detector's intentional diagnostics use a private sink, so a healthy run does not print false phantom findings.

Complexity: medium

<sub>Review-Coverage: authored=codex; ran=claude,gemini; adjudicated=domain; declined=cursor-grok,cursor-composer; rounds=5 @ 801b6d5f85d4</sub>

<sub>Human-Review-Need: 3 (decisions: insert-only-purge-witness, reuse-absence-oracle-as-detector) @ 801b6d5f85d4</sub>


integrationTests/database/txnlog-purge-stale-read-blast.test.ts
```diff
@@ -78,15 +76,13 @@ const skipSuite = process.env.HARPER_RUNTIME === 'bun';
 // (scoping sanity: proves the delete/read paths aren't just globally broken). Padded payload
 // matches QA-779/QA-787's proven >16MiB-crossing magnitude so delete_transaction_logs_before
 // in the contrast arm actually rotates/removes whole .txnlog files instead of a no-op.
-const DEL_COUNT = 6000; // ids k0..k5999
-const KEEP_COUNT = 6000; // ids k6000..k11999
+const ROCKSDB_RANGE_COUNT = 6000;
+const LMDB_RANGE_COUNT = 600;
```
cb1kenobi: **High: the LMDB control reduction removes real coverage and, after merging `main`, stabilizes nothing**

`main` already solved this a different way: `a5445a64b` — *test(database): keep the QA-782 lmdb control arm off Windows* ([#2248](https://github.com/HarperFast/harper/pull/2248), merged 2026-08-20T19:34:21Z) — adds `skipLmdbArm = process.platform === 'win32'` and gates the arm with `{ skip: skipSuite || (engine === 'lmdb' && skipLmdbArm) }`. This PR branches from `b51e92af6` (30 commits behind), so it does not see that. `git merge-tree --write-tree <head> origin/main` merges **cleanly** and produces *both*: the Windows skip **and** the 600-row arm. Net effect after merge — on Windows the lmdb arm never runs, so the reduction is unreachable there; everywhere else the arm runs at 1/10th volume on platforms that were never flaking.

That merged `skipLmdbArm` comment also states the opposite conclusion in the same file: *"Skipped rather than shrunk: the control's value is running the SAME workload as the subject, and per-engine record counts would weaken exactly that."*

**The coverage loss is measurable, not theoretical.** I mutated `ResourceBridge.deleteRecords` so a bulk delete also removes one key past the highest requested id, but only once the deleted range reaches index >= 1000 (an over-delete that needs volume to surface), rebuilt `dist/`, and ran the suite:

- at the merge base `b51e92af6`: **both** arms fail (4 failures — rocksdb *and* lmdb catch it)
- at this head `7d8c58cfd`: **only rocksdb** fails (2 failures); the lmdb arm passes 7/7 clean

At 600 rows the DEL range is `k0..k599`, so the arm also stops crossing test 3's 1000-id delete-chunk boundary entirely (6 chunks -> 1).

For calibration on the cost side: on macOS the 6000-row lmdb seed takes ~0.7s and the whole lmdb arm 7.0s, versus 4.0s at 600 — the reduction buys ~3s on a platform that showed 10/10 green at both the base and this head.

Suggested fix: drop the per-engine split and keep the single shared count. `main`'s `win32` skip already removes the Windows seed timeout this PR was reducing volume to avoid, and the shared count is what makes the control a control.

```suggestion
const RANGE_COUNT = 6000;
```

(The other changes here — `assertSeedCompleted`, the finite-cutoff guard, and the 5-detector positive control — are additive and I confirmed all three work; this is the one change that trades coverage away.)

—
_Generated by Barber AI_

integrationTests/database/txnlog-purge-stale-read-blast.test.ts
```diff
@@ -20,16 +20,13 @@
...
  *     connection-affinity tricks (that per-worker-routing question is QA-787's separate
  *     concern; this suite isolates "which surface", not "which worker").
- *   - Seed a large contiguous key range (bucket=DEL, 6000 rows) + a non-deleted range
- *     (bucket=KEEP, 6000 rows) + a byte-exact ARMED POSITIVE CONTROL (bucket=CTRL, 5 rows,
- *     inserted after a recorded cutoff) -- large enough, with padded payloads, to force real
- *     on-disk data (forced via explicit `primaryStore.flush()` calls between seeding waves,
- *     matching QA-779/QA-787's proven >16MiB-crossing magnitude so the LATER audit-purge arm
- *     is non-vacuous too).
+ *   - Seed contiguous DEL + KEEP ranges and a byte-exact ARMED POSITIVE CONTROL. The RocksDB
+ *     subject retains the >16MiB `.txnlog` rotation precondition; the LMDB control does not
```
cb1kenobi: **Low: the suite's own comments contradict each other once this merges with `main`**

This bullet claims the lmdb control "does not need that engine-specific stress volume", and line 43 quietly drops the previous "same workload" guarantee. After the clean auto-merge with `a5445a64b`, the block at lines 74-85 of the merged file asserts the opposite — that the control's value *is* the same workload and that per-engine record counts "would weaken exactly that". A future reader gets two adjacent, mutually exclusive rationales for the same arm.

Separately, the surviving comment at lines 76-78 ("Padded payload matches QA-779/QA-787's proven >16MiB-crossing magnitude so `delete_transaction_logs_before` in the contrast arm actually rotates/removes whole .txnlog files instead of a no-op") now sits directly above `LMDB_RANGE_COUNT = 600` and is no longer true for that arm — the lmdb run reports `fileCount=0 totalBytes=0.000MiB`.

Whichever way the volume question is settled, one of these two rationales has to go.

—
_Generated by Barber AI_

integrationTests/database/txnlog-purge-stale-read-blast.test.ts
```diff
@@ -412,14 +420,33 @@ function defineSuite(engine: 'rocksdb' | 'lmdb') {
 		`QA-782 stale-read blast radius: ordinary table reads after bulk delete [${engine}]`,
 		{ skip: skipSuite },
 		(ctx: ContextWithHarper) => {
+			const rangeCount = engine === 'rocksdb' ? ROCKSDB_RANGE_COUNT : LMDB_RANGE_COUNT;
+			const seedWaveSize = rangeCount / 3;
```
cb1kenobi: **Low: the derived wave/sample arithmetic only holds for the two values currently hard-coded**

`rangeCount / 3` is fractional for any count not divisible by 3, and line 433's `rangeCount * 1.5` is fractional for any odd count. The PR description asks a human to settle the volume question and says it "is reversible by changing one constant" — that instruction is unsafe as written.

Verified at this head by setting `LMDB_RANGE_COUNT = 500`: the arm dies immediately with `insert batch@166.66666666666666 should succeed, got 400: {"error":"Value 166.66666666666666 in property seq must be an integer..."}`, because `seedRange(start + offset, ...)` then generates fractional ids (`k166.666...`). (Credit where due: `assertSeedCompleted` did its job here — the five downstream phases reported "the seed phase must complete" instead of the old confusing "k11999 absent" cascade.)

Suggested fix — make the wave count integral and clamp the last wave so it can't overrun into the next bucket:

```suggestion
			const seedWaveSize = Math.ceil(rangeCount / 3);
```
then at line 483 use `Math.min(seedWaveSize, rangeCount - offset)` as the count, and at line 433 use `rangeCount + Math.floor(rangeCount / 2)` in place of `rangeCount * 1.5`.

—
_Generated by Barber AI_

integrationTests/database/txnlog-purge-stale-read-blast.test.ts
```diff
@@ -191,6 +187,18 @@ function fmtMiB(bytes: number): string {
 	return `${(bytes / 1024 / 1024).toFixed(3)}MiB`;
 }
 
+function isPreCutoffInsert(entry: any, id: string, cutoffTimestamp: number): boolean {
+	const entryTimestamp = Number(entry?.timestamp);
+	return (
+		entry?.operation === 'insert' &&
+		Array.isArray(entry.ids) &&
```
cb1kenobi: **Nit: this half of the predicate is unreachable and oversells how targeted the matcher is**

`ResourceBridge.readAuditLog`'s `HASH_VALUE` branch (`dataLayer/harperBridge/ResourceBridge.ts:546-557`) builds every entry as `{ operation, timestamp: auditRecord.version, user_name, ids: [id], records: [...] }` for exactly the id that was searched, and `harperBridge` is always `ResourceBridge` for both engines. So `entry.ids` is always `[id]` and `Array.isArray(entry.ids) && entry.ids.includes(id)` can never be false — it reads like an id filter but filters nothing.

The parts that actually do work are `operation === 'insert'` and `Number.isFinite(cutoffTimestamp)`; the latter is what genuinely fixes the cascade the PR describes, where a timed-out seed left `cutoffTimestamp` unassigned and every `timestamp < undefined` comparison went false.

Worth noting for the second call site: the post-purge `stalePreCutoffEntries` filter (line ~807) feeds only a `findings.push` — that arm ends in `ok(true, ...)` — so narrowing it changes the reported F-225 verdict a human reads out of CI, not an assertion. Fine as-is, but the `operation === 'insert'` restriction means a purge that stranded a pre-cutoff entry of any other operation type would now print "clean -- F-225 not reproduced".

—
_Generated by Barber AI_

integrationTests/database/txnlog-purge-stale-read-blast.test.ts
kriszyp: Addressed in 801b6d5f8: merged main’s Windows LMDB skip and restored one shared RANGE_COUNT=6000, so non-Windows LMDB again exercises all six 1,000-ID delete chunks.

integrationTests/database/txnlog-purge-stale-read-blast.test.ts
kriszyp: Addressed in 801b6d5f8: restored the same-workload rationale; main’s Windows-only skip comment and the suite documentation now agree.

integrationTests/database/txnlog-purge-stale-read-blast.test.ts
kriszyp: Addressed in 801b6d5f8: wave size uses ceil, the final wave clamps to the remaining count, and the midpoint uses floor, so arbitrary integer range counts cannot generate fractional IDs or overrun the range.

integrationTests/database/txnlog-purge-stale-read-blast.test.ts
kriszyp: Addressed in 801b6d5f8: removed the redundant hash-value ID argument and check. The matcher retains insert + finite timestamp + pre-cutoff filtering to identify the deliberately seeded witness while excluding its later delete.

cb1kenobi:
Re-reviewed at `801b6d5f8` — no issues found. All four findings from the `7d8c58cfd` pass are fixed. This PR looks good, nice job!

The head is a merge commit (`7d8c58cfd` + `f4314e761`), so `git range-diff` shows both original commits byte-identical (`=`) and all the new work lives in the merge resolution: **1 file, +41/−36**.

**High (LMDB control coverage) — fixed, and verified by re-running the original mutation.**

The per-engine split is gone (`ROCKSDB_RANGE_COUNT`/`LMDB_RANGE_COUNT` → a single `RANGE_COUNT = 6000`), and merging `main` brought in `a5445a64b`'s Windows-only `skipLmdbArm`. That is the combination I asked for: one shared workload everywhere, skip rather than shrink on Windows.

I re-ran the exact mutation from the last pass — `ResourceBridge.deleteRecords` over-deletes one key past the highest requested id, gated to fire only once the range reaches index >= 1000 — rebuilt `dist/` (marker verified as exactly 1 isolated match, since `#src/*` resolves to `./dist/*.js`):

| | merge base `b51e92af6` | old head `7d8c58cfd` | **new head `801b6d5f8`** |
|---|---|---|---|
| rocksdb arm | fails | fails | **fails** |
| lmdb arm | fails | passes 7/7 clean | **fails** |

Both arms now catch it (`10 pass / 4 fail`, `skipped=0`), failing on `[4. POST-DELETE KEEP samples] REST GET k6000 must be present+correct`. That restores the merge-base behavior exactly. Unmutated, the suite is `14/14` green with `skipped=0`.

Worth noting how tidy the generalization is: at 6,000 the derived values reproduce `main`'s hardcoded ones exactly — samples `[0,1,50,1500,3000,5999]` and `[6000,6001,9000,11999]`, seed waves 3×2000. So this is a pure parameterization with no workload change from `main`.

**Low (contradictory comments) — fixed.** The header now reads "same fixture, same workload", which agrees with the `skipLmdbArm` rationale instead of contradicting it. The block is now identical to `main`'s.

**Low (fractional wave/sample arithmetic) — fixed.** `Math.ceil(rangeCount / 3)`, the `Math.min(seedWaveSize, rangeCount - offset)` clamp, and `rangeCount + Math.floor(rangeCount / 2)`. Re-tested at `RANGE_COUNT = 500`, the value that previously died on `insert batch@166.66666666666666`: no fractional ids at all now. It instead fails cleanly on the deliberate `>16MiB` RocksDB precondition, and `assertSeedCompleted()` cascades with the intended "the seed phase must complete" message. Changing the one constant is now safe to suggest.

**Nit (dead `ids` predicate) — fixed.** `isPreCutoffInsert(entry, cutoffTimestamp)` keeps the `Number.isFinite(cutoffTimestamp)` guard that actually fixes the cascade and drops the unreachable half.

Two things I checked in the merge resolution that are correct but easy to get wrong:

- `detectorControl.detail.length === detectorIds.length * 2 + 3` is the real invariant, not a magic number — `assertAbsentEverywhere` pushes two per-id probes (REST-GET, FULLSCAN) plus three bucket-wide ones (REST-QUERY, SEARCH-BY-VALUE, SQL). At one id it is still 5, and the every-`PHANTOM` assertion is unchanged, so the five-surface positive control still proves what it did before.
- Routing the detector's diagnostics into a private `detectorDiagnostics` sink is safe: `findings` is only ever `console.log`'d, never asserted over, so this just keeps intentional PHANTOM lines out of the human transcript. Commit 2's ordering intent also survives — the DEL subject probe still runs first, with only the CTRL present-check moved ahead of the detector.

`prettier --check` and `lint:required` are both clean.

One limit worth restating: the flake this stabilizes is Windows-specific, and all of my verification ran on macOS. I can show the coverage is back and the arithmetic is safe; I cannot confirm the Windows seed behavior locally. The CI runs linked in the description are the evidence for that half.

—
_Generated by Barber AI_

Issue #846, opened 2026-05-28
Transaction-log retention has no config-driven time/size control (and the storageReclamation hardening isn't on v5.1)


heskew:
## Summary

`scheduleAuditCleanup` in `resources/auditStore.ts` calls `rootStore.purgeLogs({ before })` (added in #166, commit `d2723a11`), but the call is only ever fired by the `onStorageReclamation` callback. That signal tracks RocksDB block-cache / write-buffer pressure — not filesystem or quota pressure. On write patterns that don't push the in-memory thresholds (or in disk-quota-constrained deployments), the scheduled cleanup never runs and `.txnlog` files in `database/<db>/transaction_logs/local/` and `database/<db>/transaction_logs/<peer>/` grow without bound.

## Repro / observed

Hosted Fabric cluster on harper-pro 5.0.16, two nodes, 5 GB XFS user quota per uid. Both nodes filled to 4,999,999,488 / 5,000,000,000 bytes and entered a permanent restart loop with `errno -122` (EDQUOT) at startup (see #847 for the startup-hardening side of this).

Breakdown of the quota at outage (replicated app database, `<db>` = the customer's application db):

| Path | Leader | Writer |
|---|---|---|
| `database/<db>/transaction_logs/local/` (own outbound) | 119 MB / 9 files | **3.1 GB / 365 files** |
| `database/<db>/transaction_logs/<peer>/` (inbound queue) | **2.9 GB / 185 files** | 119 MB / 8 files |
| RocksDB SST + blob (top-level `<db>` dir) | ~221 MB | similar |
| `database/system` | 285 MB | similar |
| `components/` | 365 MB | similar |

`.txnlog` files dated 10–17 days before outage — far past the rocksdb-js default 3-day retention. They were never purged because `scheduleAuditCleanup` was never invoked.

## Root cause

`resources/auditStore.ts`:

```ts
onStorageReclamation(rootStore.path, (priority) => {
    cleanupPriority = priority;
    if (priority) return scheduleAuditCleanup(100);
});

function scheduleAuditCleanup(newCleanupDelay?: number): Promise<void> {
    if (auditStore instanceof RocksTransactionLogStore) {
        auditStore.rootStore.purgeLogs({
            before: Date.now() - auditRetention / (1 + cleanupPriority * cleanupPriority),
        });
        return;
    }
    // ...
}
```

[...]

Issue #2243, opened 2026-08-20
LMDB bulk-insert degrades superlinearly on Windows (~311s vs ~2s on Linux), intermittently failing Integration Tests shard 1 on main


kriszyp:
## Summary

The same 12,000-record RocksDB seed workload takes **~2s on Linux CI and 83–238s on Windows CI** — a 40–120× platform gap. Above ~180s it exceeds the test's 60s per-request abort and fails, which is why `Integration Tests 1/6 (Windows, Node.js v24)` is intermittently red on `main`.

Found while triaging why harper#2228's Windows shard was red. That PR is unrelated — the identical failure reproduces on `main`.

## Evidence

`integrationTests/database/txnlog-purge-stale-read-blast.test.ts`, subtest `1. seed DEL (6000) + KEEP (6000) ranges, force flushes between waves`. The suite runs the same fixture and workload twice, once per engine, so both arms appear in every job log.

| Run | Platform | rocksdb arm | lmdb arm | Outcome |
|---|---|---|---|---|
| [32312146829](https://github.com/HarperFast/harper/actions/runs/32312146829) | Linux | **2.04s** | 1.92s | pass |
| [32312146829](https://github.com/HarperFast/harper/actions/runs/32312146829) | Windows | **83.2s** | 2.88s | pass |
| [32207576347](https://github.com/HarperFast/harper/actions/runs/32207576347) | Windows | **169.1s** | 4.36s | pass |
| [32319029671](https://github.com/HarperFast/harper/actions/runs/32319029671) | Windows | **180.9s** | — | **fail** (abort) |
| [32366910072](https://github.com/HarperFast/harper/actions/runs/32366910072) | Windows | **238.0s** | — | **fail** (abort) |

The controlled comparison is the rocksdb arm across platforms — same code, same workload, same flush count: **2.0s vs 83–238s**.

The workload: 24 sequential ops-API `insert` batches of 500 records each (payload padded to 1780 bytes, ~900KB per batch), with 7 explicit `primaryStore.flush()` calls interleaved between waves.

The failure mode is `TimeoutError: The operation was aborted due to timeout` from the test's `AbortSignal.timeout(60_000)` on `rawOp`, i.e. **a single ~900KB insert batch exceeded 60 seconds**. Every subsequent subtest failure in that job is cascade from the truncated seed (`k9000` absent, etc.), not an independent defect.

## Why this looks like a real defect rather than slow CI

[...]

PR #2245, opened 2026-08-20
Keep Harper booting when its storage volume is full or at quota

### LINKED ISSUE #847: Harper crash-loops on EDQUOT during config tmp-file write, with no recoverable path
## Summary

When the storage volume is at hard quota or full, Harper enters a permanent restart loop at startup because `applyRuntimeEnvVarConfig` in `config/configUtils.js` always invokes `atomicWriteFile` to rewrite `harper-config.yaml`. The atomic write opens a sibling `harper-config.yaml.1.0.<hash>.tmp` file, which fails with `errno -122` (EDQUOT) on a quota-exhausted XFS volume. The error is unhandled and bubbles to process exit; Docker restarts the container; the same write fails again. With no operator intervention there is no recovery — the cluster stays down even though the only thing that needs to happen is for *one process to start* so the existing scheduled cleanup paths can run.

## Repro / observed

Hosted Fabric cluster, both nodes filled to 4,999,999,488 / 5,000,000,000 bytes XFS user quota. Restart loop with this error every ~10s:

```
Error: Unknown system error -122: Unknown system error -122,
  open '/home/harperdb/harper/harper-config.yaml.1.0.<hex>.tmp'
    at Object.writeFileSync (node:fs:2414:20)
    at atomicWriteFile (.../config/configUtils.js:97)
    at applyRuntimeEnvVarConfig (.../config/configUtils.js:870)
    at Object.initConfig (.../config/configUtils.js:303)
    at Object.initSync (.../environment/environmentManager.js:103)
    at Object.<anonymous> (.../bin/cliOperations.js:3)
  errno: -122
  code: 'Unknown system error -122'
  syscall: 'open'
```

The merged config on this run is byte-identical to what's already on disk — there's nothing to apply. The tmp-file write is unconditional.

## Why this hurts

The chicken-and-egg: Harper needs to start to schedule `purgeLogs` cleanup. Harper can't start because there's no quota. There's no quota because nothing scheduled cleanup. The operator has to either bump the quota or manually delete files outside the container, neither of which should be required to recover from "the cleanup that's supposed to run hasn't run yet."

Companion issue #846 covers the underlying reason the cleanup didn't run; this issue is about Harper's behavior in the at-quota state regardless of how it got there (full disk, container-scoped overlay quota, etc.).

## Proposed directions

1. **Skip the write when content is unchanged.** `applyRuntimeEnvVarConfig` should compare the merged config to the current file before invoking `atomicWriteFile`. Byte-identical → no-op. This alone would have prevented this outage from becoming a wedge.
2. **Catch EDQUOT (and ENOSPC) at `atomicWriteFile` and surface a clear, non-fatal-at-startup error.** On startup, log "storage exhausted; running with last-known-good on-disk config" and continue boot if the existing config is parseable. Fail-fast only if the on-disk config is unreadable.
3. **Pre-flight check.** Optionally, before any tmp write at startup, `statvfs` the target dir and refuse to attempt the write if free space < N bytes, surfacing a structured "STORAGE_EXHAUSTED" error rather than a runtime EDQUOT.

(2) is the meaningful behavioral fix; (1) is the cheap defense-in-depth.

## Related

- #846 — root cause for *why* the cluster got to this state.

---

kriszyp:
Harper crash-looped at startup on a storage-exhausted volume: every boot rewrites `harper-config.yaml` and the env-config state snapshot when a `HARPER_*_CONFIG` variable is set, the sibling temp file was refused with EDQUOT, and the error escaped `initConfig` — so the container restarted, the same write failed again, and nothing inside it could free space because nothing could start. Boot-path persistence of derived config is now best-effort: the effective configuration is already merged and validated in memory, so ENOSPC/EDQUOT is logged and startup continues on it, while install and user-requested writes (`set_configuration` and friends) still persist or throw. A refused log append also degrades to stdout instead of throwing, which otherwise made every log statement on a full volume a crash point — including the one reporting the storage problem.

Necessary but, on its own, not sufficient: this gets configuration initialization and logging through an exhausted volume. A full `harper start` there still needs writable space for the storage engine, so "one process starts and the scheduled cleanup runs" is not proven by this change. #846 remains the reason a hosted node reaches this state.

Fixes #847.

## For the human reviewer

1. **Boot continues on the in-memory config instead of failing fast.** The alternative is to keep aborting and require operator intervention, which is what produced the outage. The trade is a guaranteed-consistent on-disk config for availability — everything below is a consequence of admitting that divergence. Reversible in one function (`persistConfigDuringBoot`), but if you disagree with the premise, the rest of the PR follows from it.
2. **Only ENOSPC/EDQUOT degrade; EROFS, EACCES and EIO still abort boot.** A read-only remount produces the identical un-breakable restart loop, so this leaves the same field outage reachable by another route. Narrow now on purpose, trivially widened later by adding codes to one set — I'd rather not widen it on speculation.
3. **The two artifacts move together, or not at all — and the confirmed record is never at the mercy of a write.** The env-config state records the file's pre-env values and is the only copy of them, so the new state is *staged in a sidecar*, the config file is written, and the sidecar is then promoted with a rename. A rename needs no free space, which is the whole point: a refused staging write leaves the config file alone, a refused config write drops the sidecar, and in both cases every recorded original survives untouched. A sidecar left by an interrupted commit is cleared on the next boot, which also skips drift detection for that boot — it cannot distinguish a manual user edit from the write that was in flight, and guessing wrong hands those paths to `user` permanently. A boot that re-derives the same state writes nothing at all. This is the subtlest part of the PR and where I'd look hardest; two earlier shapes of it (snapshot-last, then a pending marker written over the confirmed record) each had their own data-loss path, both caught in review.
4. **`ensureConfigKeysPresent` reports nothing it could not persist,** so on an exhausted volume a newly-introduced built-in (`waf`, `secretCustody`) stays dormant for that boot and the startup log says so. The rejected alternative was activating it in the main thread's memory: worker threads re-read the config from disk and never run the backfill, so that gives a boot where the log claims WAF is on and every request-handling worker is without it. Dormant-and-honest beats active-in-name-only, but it is a security-component policy call and yours to make.
5. **Install still fails hard on a full volume.** There is no last-known-good config to fall back on and no process to keep alive, so its snapshot write stays mandatory. Worth a line in the release note: boot tolerates a full volume, install does not.
6. **`skipIfUnchanged` is opt-in inside `atomicWriteFile`.** Skipping means no mtime bump and therefore no watcher event, so it is a watcher-visible semantic that every call site can now switch. Only the two artifacts that are re-derived identically each boot pass it. A separate `writeFileIfChanged` was the alternative; cheap to move, awkward to un-ship once callers depend on the flag.
7. **Log fallbacks bypass the stdio guard, and mirror to stdout unconditionally.** `installStdioGuard` routes console output back into the file logger when `logging.file` and `logging.console` are both on, so the obvious `console.log` fallback recurses until the stack blows; everything now goes through one non-throwing `writeToStdioDirectly`. Two consequences worth your ruling: the no-descriptor branch used to drop every entry after the first and now emits them, and the mirror does not consult `logging.console` — so a deployment that deliberately keeps log content out of stdout will see it there while the volume is full. I took diagnostics over silence, but gating the mirror on the console setting is a one-line change if you'd rather. A refused append also now sets a 5s cooldown, so a persistently full volume costs one failed syscall every few seconds instead of one per log line on the request path.

8. **The deeper simplification I did *not* take.** All of this subtlety exists because boot rewrites the operator's own `harper-config.yaml` with env-derived values, which is what makes a separate record of the pre-env values necessary in the first place. Stop doing that — keep the user's file untouched and hold the env overlay separately — and the two-artifact ordering problem disappears entirely rather than being solved. I did not do it here: it changes what operators see when they read the config file, changes restoration semantics, and needs a migration for existing state/YAML pairs, none of which belongs in a P1 availability fix. Worth its own issue if you agree.
9. **Rebased onto `origin/main` (66 commits) — two conflicts in `config/configUtils.ts`, both semantic, both resolved to keep both sides' intent:** main had independently restructured `renameWithRetry`'s loop (attempts/startedAt/`logger.warn` diagnostics, unified try/finally) while this PR added the write-failure `removeTempFile` cleanup — kept main's structure, added the cleanup inside the same finally. Separately, main's diagnostics landed inside the retry loop while this PR extracted `renameWithRetry` as its own exported function (reused by `harperConfigEnvVars.ts`'s `commitStagedConfigState`) — kept the extraction, moved main's attempts/startedAt/`logger.warn` tracking inside it. The post-rebase pre-push round (Gemini only — Codex/Cursor-Grok/domain all timed out under this environment's review budget) raised one `<blocker>` on `Atomics.wait` throwing on Node's main thread; verified false — Node has no browser-style main-thread restriction on `Atomics.wait` (confirmed directly: `node -e` sleeps and returns `timed-out`, no throw). Its `<minor>`/`<nit>` findings (a test's `expirationBoundary` naming, comment style in `configUtils.ts`/`harperConfigEnvVars.ts`) are all pre-existing PR content the rebase didn't touch — left for this review pass rather than fixed as scope creep on a rebase-only task.

## Verification

**Unit** — `unitTests/config/storageExhaustion.test.js` (new, `node:assert` against the real modules per AGENTS.md): errno classification including Linux's unmapped `-122`, the swallow/rethrow boundary, `skipIfUnchanged` behaviour and the default path, temp-file cleanup when the write itself fails, the two-phase snapshot contract (including that an unchanged boot writes nothing and that a still-pending snapshot is discarded on load), and the rollback primitive. The log fallback is covered for real rather than with stubs: a child process logging to `/dev/full` — which refuses every write — must survive, land its entry on stdout, and not recurse through the stdio guard. It skips where `/dev/full` does not exist (macOS) and runs on CI.

```
npx mocha "unitTests/config/**/*.js" "unitTests/utility/logging/**/*.js"   →  401 passing, 14 pending
```

**Fails on base** — with `config/` and `utility/logging/` restored from `origin/main` and `dist` force-rebuilt, the same command gives `374 passing, 14 failing`: 13 of the 14 new cases plus the runtime-env-var file's `before` hook, which cannot stub a function that does not exist on base. (The fourteenth — temp cleanup on an ENOENT write — passes on base, which never creates the temp file in that path.) Restored and rebuilt: green again.

**End-to-end, on a genuinely full volume** — `~/dev/scripts/harper-full-volume-boot-check.sh`: a 20MB HFS+ image with an existing config file, an existing log file and the standard directories, filled to zero free bytes, then `initConfig` with a `HARPER_SET_CONFIG` layer that must be merged, plus log lines larger than the volume's remaining slack.

```
free bytes: 0
RESULT append-premise=ENOSPC                      # the log append genuinely cannot land
Harper cannot write to its log file: ENOSPC: no space left on device, write
2026-08-20T13:28:07.154Z [main/0] [error]: unwritable line 0 …   # via nativeStdWrite, no recursion
RESULT initConfig-returned=true
RESULT effective-http-port=8123                   # env layer applied from memory
RESULT config-file-unchanged=true
RESULT leftover-temp-files=none
RESULT env-state-file-present=false               # no snapshot left ahead of the file
exit=0
```

Same rig on `origin/main` exits non-zero at `initConfig` with `ENOSPC … backup/.harper-config-state.json.<pid>.<tid>.<hex>.tmp` — the reported wedge.

**Not covered by the repeatable suite:** the coordinator's failure *sequences* (which write is refused in which order) and the log-append degradation. Both need a filesystem that refuses writes, which no unit test can arrange portably, and AGENTS.md rules out reaching for stubs to fake it — hence the live rig above. `test:unit:main` and `test:integration:all` were not run locally (the shared RocksDB system-database lock on this machine fails them before any test executes); CI covers them.

**Outside-review coverage:** nine completed rounds — Codex + Gemini + Cursor Composer + Harper-domain adjudication at `6ee5ded`, Codex + Gemini at `292edf53`, Codex + Gemini + Harper-domain adjudication at `6ededadd`, and delta rounds (same reviewers, resumed sessions) through this head. **Rounds 1–7 each found a real defect in the previous round's fix**, all inside the same ~20 lines of commit ordering: snapshot-last lost the operator's originals; snapshot-first left a snapshot ahead of the file; the rollback unlink destroyed originals accumulated over months; the pending marker needed a write an exhausted volume refuses; one fixed sidecar name raced every worker thread; and the pid liveness check exempted processes but not sibling threads, which share a pid. Round 8 was the first with no major, and its two minors (a test assertion that could pass vacuously, and an observer comparing against another process's half-committed pair) are fixed. That history — six data-loss paths and one boot-abort, none of which my own testing found — is the argument for looking hard at item 3 specifically, and the reason this is still draft.

Complexity: complicated

<sub>Review-Coverage: authored=claude; ran=gemini; blocked=codex(timeout),claude(fallback)(out-of-budget),cursor-grok(failed),domain(timeout); declined=cursor-composer; rounds=1 @ 708586ecb5cf</sub>

<sub>Human-Review-Need: 3 @ 708586ecb5cf</sub>

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -633,6 +633,41 @@ mirrors loader behaviors that must stay in sync if the loader changes: config fi
 (`..` and absolute patterns rejected). Known limitation: a `componentsRoot` override that itself
 arrives via env var cannot redirect the scan.
 
+## Boot-path config persistence is best-effort, and its two artifacts commit as a unit (`config/configUtils.ts`, `config/harperConfigEnvVars.ts`)
+
+Every boot with a `HARPER_*_CONFIG` env var set re-derives the merged config and, historically, wrote
+it back unconditionally. On a full or quota-exhausted volume that write is refused and, being fatal,
+turned a full disk into a container restart loop nothing inside the container could break — the
+cleanup that frees space needs a started process (#847). Two rules follow.
+
+**Derived boot writes are best-effort; user-requested ones are not.** `persistConfigDuringBoot()`
+swallows exactly ENOSPC/EDQUOT (matching on `errno` as well as `code`, because Linux has no libuv
+mapping for EDQUOT and reports `Unknown system error -122`) and lets the boot proceed on the
+in-memory config. `updateConfig`/`set_configuration`, `addConfig`, `deleteConfigFromFile` and the
+install path keep persist-or-throw: a caller who asked to persist must not get a silent success, and
+an install has no last-known-good config to fall back on.
+
+**The env-config state and the config file must never disagree.** The state file records the
+_pre-env_ values, so it is the only copy of what the operator's config said before an env layer
+overwrote it — the config file itself holds the env-derived value. Both single-file orderings lose
+something: writing the state last means the file it would read originals from is already
+overwritten; writing it first leaves a state ahead of the file, which the next boot's
+`detectConfigDrift` reads as a manual user edit and _permanently_ reassigns those paths to `user`,
+silently disabling the env layer even after space is freed. So the commit is three steps —
+`saveState()` stages the new state in `.harper-config-state.pending.json`, the config file is
+written, and `confirmConfigWritten()` **renames** the sidecar over the confirmed record. A rename
+needs no free space, which is the point: no write an exhausted volume can refuse ever stands between
+the confirmed originals and disk. A refused staging write leaves the config file alone; a refused
+config write unlinks the sidecar; a sidecar found at load means a commit was interrupted, so it is
+cleared and drift detection is skipped for that boot rather than mistaking the in-flight write for
+an edit. A boot that re-derives the same state writes nothing at all.
+
+Related: a log write must not be fatal either. `fs.appendFileSync` in `logQueuedData` throws from
+both inline and timer call sites, so on a full volume every log statement was a crash point. The
+fallback goes through `nativeStdWrite`, never `console` — `installStdioGuard` routes console output
+back into this same file logger when `logging.file` and `logging.console` are both on, so a console
+fallback recurses until the stack blows.
+
 ## A dangling symlink silently truncates the deploy tarball (`components/packageComponent.ts`)
 
 Packaging uses `tar-fs.pack(dir, { dereference: true })` by default (`skip_symlinks` off).
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -674,9 +674,17 @@ delete the main thread's in-flight sidecar as if it were the last boot's wreckag
 A sidecar owned by a _live_ foreign process is not cleared — that process is mid-commit — but its
 presence still turns drift detection off for this boot: a pair someone else is halfway through is no
 more comparable than one an interruption left behind. That suspension is why a sidecar also ages
-out after a minute regardless of what its pid says: a commit is three synchronous steps, so anything
-older is wreckage whose pid has been recycled, and without the age-out it would suspend drift
-detection on every boot from then on.
+out regardless of what its pid says: without it, a sidecar whose owner was killed and whose pid was
+later recycled would look mid-commit forever and suspend drift detection on every boot. The age-out
+is deliberately far longer than a commit could take — recovery from a recycled pid only has to be
+eventual, while deleting a slow-but-live writer's sidecar is the worse error, stranding its config
+file against an unpromoted state.
+
+Drift detection is main-thread-only for the same reason recovery is. A worker never owns the state:
+in the normal sequence the main thread has already classified and persisted before any worker runs,
+and inside the main thread's commit window a file that differs from the snapshot is as likely to be
+the write in flight as an operator edit. A worker that concluded "user edit" would drop the
+env-supplied value for itself alone and serve different config than its siblings.
 
 Known limit: the pair commits as a unit _within a process_. Two live processes (a server boot and a
 CLI invocation) can still interleave their config-file writes and promotions, and nothing in the repo
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -673,7 +673,10 @@ delete the main thread's in-flight sidecar as if it were the last boot's wreckag
 
 A sidecar owned by a _live_ foreign process is not cleared — that process is mid-commit — but its
 presence still turns drift detection off for this boot: a pair someone else is halfway through is no
-more comparable than one an interruption left behind.
+more comparable than one an interruption left behind. That suspension is why a sidecar also ages
+out after a minute regardless of what its pid says: a commit is three synchronous steps, so anything
+older is wreckage whose pid has been recycled, and without the age-out it would suspend drift
+detection on every boot from then on.
 
 Known limit: the pair commits as a unit _within a process_. Two live processes (a server boot and a
 CLI invocation) can still interleave their config-file writes and promotions, and nothing in the repo
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -671,6 +671,10 @@ derive the same merged config and would otherwise race over one pair of files fo
 already agree on — and since a worker shares its process's pid, a recovery scan from one would
 delete the main thread's in-flight sidecar as if it were the last boot's wreckage.
 
+A sidecar owned by a _live_ foreign process is not cleared — that process is mid-commit — but its
+presence still turns drift detection off for this boot: a pair someone else is halfway through is no
+more comparable than one an interruption left behind.
+
 Known limit: the pair commits as a unit _within a process_. Two live processes (a server boot and a
 CLI invocation) can still interleave their config-file writes and promotions, and nothing in the repo
 serializes config writes across processes. Pre-existing — both artifacts were unordered before this
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -662,13 +662,20 @@ config write unlinks the sidecar; a sidecar found at load means a commit was int
 cleared and drift detection is skipped for that boot rather than mistaking the in-flight write for
 an edit. A boot that re-derives the same state writes nothing at all.
 
-Two details the name and the caller carry. The sidecar is **per-process**: every CLI invocation and
-every worker thread runs `initConfig`, and one shared name would let a starting server clear a
-running process's in-flight commit — the loser would then rewrite the config file with the confirmed
-state still describing the old values, which is the failure the protocol exists to prevent. Recovery
-therefore only clears sidecars whose owning pid is gone. And only the **main thread** persists at
-all: workers derive the same merged config and would otherwise race over one pair of files for a
-result they already agree on.
+Two details the name and the caller carry. The sidecar is **per-process**: every CLI invocation runs
+`initConfig`, and one shared name would let a starting server clear a running process's in-flight
+commit — the loser would then rewrite the config file with the confirmed state still describing the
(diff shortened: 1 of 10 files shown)


kriszyp:
Bot pass triaged: the Windows/EDQUOT one was real and is fixed in 64fe91e5 — `os.constants.errno.EDQUOT` is absent on platforms without quotas and `-undefined` is `NaN`, so the set held a value matching nothing. The three type-annotation suggestions I declined: `configUtils.ts` types its module-level helpers loosely throughout (`atomicWriteFile(filePath, content, …)` and neighbours are untyped), and annotating only the functions this PR adds would make the new code the odd one out rather than the file consistent. Happy to type them if we'd rather start converging the file.

kriszyp:
Note on the red **Unit Test** check, since it isn't this PR:

- `log writes that cannot land (#847)` failed on `34ca3328` with `ReferenceError: testRoot` — that one *was* mine, fixed in `ac35021a`.
- `HNSW greedy routing above layer 0 (ROUTING_EF)` (`unitTests/resources/vectorIndex.test.js:673`) is not. I ran `test:unit:resources` against `origin/main` on a Linux box: **3 failing there too** (`table() randomAccessFields directive` among them). Main's Unit Test job has failed on three different tests across its three most recent runs — `pp_updated49`, `reloadMarker` (harper-pro#495), and this one — so the suite is flaky right now, and nothing in this diff can reach HNSW neighbour sets.

Config and logging suites pass on both macOS and Linux (368 passing), and the new storage-exhaustion file is 16 passing on Linux where `/dev/full` exists.

kriszyp:
Second CI-flake data point, since the Unit Test check has now gone red three times on this branch for three different tests, none of them in this diff:

| Run | Job | Failing test |
|---|---|---|
| `64fe91e5`, `ac35021a` | v26 | `HNSW greedy routing above layer 0` (`vectorIndex.test.js:673`) |
| `871c51d6` | v22 | `put with immediate expiration on sourced table should expire immediately` (`apiTests/cache-test.mjs:73`) |
| `871c51d6` | v24 | `HNSW greedy routing above layer 0` again |

`origin/main` fails too — I ran `test:unit:resources` against it on Linux and got 3 failures (`table() randomAccessFields directive` among them), and main's own Unit Test job has failed on `pp_updated49` and `reloadMarker` (harper-pro#495) in its recent runs. The suite currently goes green or red by luck, which makes "CI green" a stochastic gate for anything merging right now — worth someone's attention independently of this PR.

The one red that *was* mine (`ReferenceError: testRoot`) was fixed in `ac35021a`. This PR's own tests — 21 in `storageExhaustion.test.js`, including the `/dev/full` and worker-thread legs — have passed on every run where the job reached them, and 373 pass on Linux at head.

kriszyp:
Last CI note, so nobody re-derives it: **Integration Tests 1/6 (Windows, Node.js v24) is a permanent red on `main`, not a flake and not this PR.** It has failed on all four recent `main` runs including `b51e92af` (this branch's base) and on every sha here, with the shard's requests all returning `ECONNREFUSED` — the server never finishes coming up. That matches #2243 (Windows RocksDB seed takes 83–238s versus ~2s on Linux, so the seed outruns the readiness window).

I checked rather than assumed, since this change does touch startup; the base-branch history settles it.

Net: on this repo today, "all checks green" is unreachable for any PR that runs the Windows shards, and the unit suite is separately stochastic (previous comment). The signals that are meaningful for this PR — config and logging suites, all Linux/Bun/uWS integration shards, the Next.js adapter, and unit on v22/v24/v26 — are green, with 373 passing on Linux at head.

PR #2252, opened 2026-08-21
Bound the exclusive 'update-attributes' lock wait and release it structurally

### LINKED ISSUE #2251: Unbounded busy-spin on tryLock('update-attributes') pins a worker core permanently; lock release is hand-enforced at each escape path
## Summary

`exclusiveLock()` acquires the `update-attributes` lock with an unbounded busy-spin:

`resources/databases.ts:2020`
```ts
while (!rootStore.tryLock('update-attributes')) {} // use a spin lock, we really need an synchronous exclusive lock here
```

`resources/Table.ts:1510`
```ts
while (!rootStore.tryLock('update-attributes')) {}
```

No deadline, no backoff, no yield to the event loop. If the lock is ever held by something that will not release it, the spinning worker pins a CPU core **forever** — it cannot be interrupted, it logs nothing, and only a restart clears it.

The codebase already knows this failure mode. `resources/databases.ts:1731-1739`:

```ts
} catch (error) {
    // A failure while opening/creating the column family or writing the
    // table id / catalog entry (e.g. into an env poisoned by a prior
    // dangling column family) must NOT leak the exclusive
    // 'update-attributes' spin lock. If it leaks, every subsequent
    // create_table / attribute update on this database spins forever
    // (a hard wedge that pins a worker at 100% CPU). Release before rethrow.
    if (releaseExclusiveLock) releaseExclusiveLock();
    throw error;
}
```

That comment describes the exact symptom below. It patched *one* escape path; the spin itself, and the hand-rolled release discipline around it, are unchanged.

## Observed

Nine Harper processes on one dev machine, each pinned at 96–99% CPU for 16–18 hours — **879% total, ~8.8 cores**. `sample(1)` on one of them, aggregated by weight over the busy `http` worker thread:

```
2500 node::Environment::RunTimers
1629 rocksdb_js::Database::TryLock
1188 rocksdb_js::getNapiBufferFromArg     (nested inside TryLock)
 128 rocksdb_js::DBDescriptor::lockEnqueueCallback
```

The worker's event loop is doing essentially nothing but re-calling `tryLock` from a timer callback. Meanwhile the main thread is idle in `kevent`, so this is not load — it is one thread spinning on a lock it will never get.

Two secondary observations:

- **Each spin iteration re-marshals the key.** 1188 of the samples are in `getNapiBufferFromArg` — the string `'update-attributes'` is converted to a buffer on *every* attempt. Even a spin that terminates quickly is far more expensive per iteration than it needs to be.
- **All nine processes were spinning on the same named lock.** Consistent with one holder dying (or leaking) and everyone else spinning indefinitely.

## Confidence

**High** that an unbounded `tryLock('update-attributes')` spin is what burned the CPU: the native frame is unambiguous and the two spin sites above are the only unbounded ones on `main`.

**Not yet confirmed** which of the two sites (or the callback-based path at `Table.ts:5791`, which accounts for the 128 `lockEnqueueCallback` samples) was the actual entry point — the processes were killed to recover the machine before a JS-level stack was captured. The fix below applies regardless of which one it was.

## The invariant

The real problem isn't the leak that was patched — it's that **"the exclusive lock is always released"** is currently enforced by hand at each escape point (`if (releaseExclusiveLock) releaseExclusiveLock();` appears at `databases.ts` L1658, L1738, L1989, only the last of which is in a `finally`). Every new early-return or throw between `exclusiveLock()` and its release is a fresh chance to wedge a core permanently. L1654's `exclusiveLock()` is followed by an unguarded early return at L1656-1660.

Guarding each path individually is working around the invariant rather than enforcing it.

## Suggested direction

1. **Make the spin bounded and yielding.** A deadline plus backoff, throwing a diagnosable error on expiry, converts a permanent silent wedge into a loud recoverable failure. A hot `while(){}` that never yields is never the right shape in a worker that also serves HTTP.
2. **Enforce the release structurally** — a `withExclusiveLock(fn)` wrapper with the release in `finally`, so no call site can leak it. Replaces three hand-placed releases with one.
3. **Hoist the key marshalling** out of the retry path so a spin iteration isn't allocating.
4. Consider whether a genuinely blocking cross-process lock primitive belongs in `rocksdb-js`, given the comment's admission that a spin lock is a stand-in for the synchronous exclusive lock actually wanted.

(1) and (2) are the ones that matter; (1) alone would have turned 18 hours of silent core-burn into an error message.

## Affected

Both spin sites present on `harper` `origin/main` at time of filing. Not specific to a branch or worktree.

## Related

The processes above survived to spin for 18 hours because of a separate test-harness orphan-reaping gap, filed as [HarperFast/integration-testing#29](https://github.com/HarperFast/integration-testing/issues/29). Independent defects — reaping explains why they *lived*, this explains why they were *hot*. This one is reachable in production without any test harness involved.

kriszyp:
Fixes #2251.

The exclusive `update-attributes` lock could spin forever at 100% CPU when a holder failed to release it, and `table()` relied on scattered manual unlocks that made every throw or early return another leak risk.

- [The shared lock helper](https://github.com/HarperFast/harper/pull/2252/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R139) now fast-paths an uncontended acquire, briefly spins, sleeps with bounded backoff, and throws a machine-readable retryable 503 after the dedicated 10-second deadline. The interval clock is monotonic, and the lock key is encoded once.
- `table()` funnels every acquired lock through one idempotent `finally`; `dropTable()` uses the same structural wrapper. The recursive concurrently-created-table branch releases before re-entry because the lock is not reentrant.
- The wrapper rejects Promise-returning callbacks at the TypeScript boundary and diagnoses/contains JavaScript misuse so an orphaned rejection is not left unhandled.
- Because the acquire can now throw, [`table()` takes the RocksDB lock before the first mutation of the live `Table`](https://github.com/HarperFast/harper/pull/2252/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1594) rather than lazily at the first catalog write. A lost race is then a clean no-op instead of a worker left describing attributes that were never persisted, with an indexed attribute whose dbi never reached `Table.indices`. LMDB keeps the lazy acquire: its `exclusiveLock()` opens an environment-wide write transaction that cannot time out, so taking it eagerly would stall every write to the database on an unchanged reload.
- [A successful acquire that waited past `UPDATE_ATTRIBUTES_LOCK_SLOW_WAIT` (1s) warns once](https://github.com/HarperFast/harper/pull/2252/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R172) with the table and elapsed time — contention was otherwise invisible until it became a timeout. The warn is wrapped because the caller registers its release only after the acquire returns, so a throwing logger would leak the lock with no `finally` able to reach it.
- [Ten lock regression tests](https://github.com/HarperFast/harper/pull/2252/changes?w=1#diff-34535e18757ee87b9a9f84716e21a3e0a5db180d623a4803412a9eb0b90671dcR52) cover the helper and real `table()` deadline paths under hard parent-side watchdogs, exception release (including proof that the injected failure occurs while locked), recursive reload release, machine-readable retry fields, and cross-thread key identity/hand-off. [A compile-time contract test](https://github.com/HarperFast/harper/pull/2252/changes?w=1#diff-a3ab3525f966686436756a59324acded641ba7d641561689729609dc44b63c58R1) rejects async callbacks.

## For the human reviewer

The three owner decisions this PR was waiting on are now made and implemented — targeted no-drift recovery by acquiring before mutating, a one-shot `warn` at 1s, and a fixed 10s deadline. What remains:

- **The eager acquire serializes unchanged RocksDB reloads, by design.** [Every existing-table declaration now takes the lock](https://github.com/HarperFast/harper/pull/2252/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1594) and holds it through the whole reconciliation pass (the `attributesDbi` range scan, a `getSync` per attribute, `openIndex` per indexed attribute), not just around the first catalog write. A hot component reload re-runs the schema declaration on every worker simultaneously, so *N* threads × *M* unchanged tables in one database now fully serialize, each waiter blocking its own event loop while it waits; review put 24 × 60 in range of the 10s deadline. This is the accepted cost of making the no-drift invariant structural rather than reconstructed, and it is the one place where a preflight "does this declaration differ from `Table.attributes`?" check would buy the old lock-free path back. I did not add that check: the loop stamps `key`/`dbi`/`indexNulls` onto the declared attribute objects, so skipping the splice when they compare equal diverges live and declared identity, and the stamping makes an unchanged declaration compare *unequal* on most indexed tables anyway.
- **A lock-acquire timeout in the replication apply path is swallowed.** The `define_schema` event calls `table({..., origin: 'cluster'})` inside the apply loop, whose `catch` logs `error in subscription handler` and pulls the next event. Under the old unbounded spin this call could not fail — it hung, visibly. It can now throw, and a node that loses the race never adds the attribute or builds the index while its peers do: replicated records still carry the field, so `search_by_value` on that attribute returns empty on that node only, with nothing surfaced to any client. `retryable: true` does not help — a repo-wide grep finds no reader of `.retryable`. Fixing this means either failing the event instead of advancing past it, or making the cluster-origin acquire unbounded; both are replication-semantics calls beyond this PR, so I left it. Worth its own issue.
- **The invariant covers the acquire, not every later failure.** A catalog write that throws *after* the lock is held still leaves the live `Table` ahead of disk, exactly as before this PR. Only the new, load-triggered timeout path is closed.
- **The full-duration regression costs a real 10s in `test:unit:resources`**, per the decision to keep the deadline on the production contract. `table()` exposes no timeout injection point, so a cheaper version would need one.
- The hot-spin/backoff profile remains the original PR's chosen design. A native-notified or asynchronous lock would be a larger cross-cutting change.

## Verification

- `npm run build` — passed.
- `npm run lint:required` — passed. The stricter repository-wide `npm run lint` still reports 12 unrelated pre-existing warnings outside this diff.
- `npx mocha unitTests/resources/updateAttributesLock.test.js` — 10 passing.
- `npm run test:types` — passed, including the async-callback negative contract.
- `npm run test:unit:resources` — 1680 passing, 16 pending, 3 failing; the three failures are the same local baseline failures previously reproduced on `origin/main` (`randomAccessFields` ×2 and `replayStructures`).
- Fails-on-base check: with only the eager-acquire line reverted, [the new cross-thread regression](https://github.com/HarperFast/harper/pull/2252/changes?w=1#diff-34535e18757ee87b9a9f84716e21a3e0a5db180d623a4803412a9eb0b90671dcR184) fails with `+ 'added'` in the live attribute list and an empty `indices` — the reported drift, reproduced exactly.
- `npm run test:unit:main` could not run on this machine: a stale global `hdb_boot_properties.file` points at a removed install, and the same failure reproduces on an unmodified tree. CI covers it.
- End-to-end route: not safely observable through the operations API without deliberately wedging a live worker. [The regression now wedges the lock from a second thread](https://github.com/HarperFast/harper/pull/2252/changes?w=1#diff-0ddf4b0322d95adff8ab690f54dd095fdbaf534a597a71e25d703e31f474594eR81) — the production failure mode — and snapshots attributes, index handles, `schemaVersion` and class metadata around the timed-out call alone.
- PR CI on `ca336ba33c04` — fully green: 42 checks pass across Linux/Windows/Bun/uWS integration, Next.js adapter, unit (Node 22/24/26), build, lint, format, review and security. Three failures on the first pass were flakes and cleared on rerun: `Integration Tests 2/6` failed its instance readiness probe (`ECONNREFUSED` at suite setup — the same signature `main` was failing with at the time), and the unit job failed `caching.test.js:155` on one attempt and `vectorIndex.test.js:1009` on the next before passing clean, which is timing/approximate-search nondeterminism rather than a deterministic regression.
- Independent review, this round: full pass (Codex + Gemini + Cursor Composer + Harper-domain) raised four items — the LMDB half of the eager acquire, the slow-wait logger leak window, single-thread-only regression coverage, and a narrating comment. All four are fixed here. The delta pass (Codex + Gemini) confirmed those fixes and returned only the two owner-level items carried above.

Complexity: moderate — concurrency-sensitive schema/DDL locking, with no storage-format or record hot-path change.

— Claude Opus, continuing work by Claude Fable and GPT-5 Codex

🤖 Generated with Claude Code

<sub>Review-Coverage: authored=claude; ran=gemini,codex; declined=cursor-grok,cursor-composer,domain; rounds=4 @ ca336ba33c04</sub>

<sub>Human-Review-Need: 4 @ ca336ba33c04</sub>

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -287,7 +287,7 @@ A handful of design points are non-obvious and easy to break:
 
 Two related traps: the create/schema-update path's exclusive `update-attributes` lock is a
 synchronous bounded wait (`acquireUpdateAttributesLock` in `Table.ts`: brief hot spin, then
-`Atomics.wait` backoff, `ServerError` after the 10s `LOCK_TIMEOUT` — harper#2251; it used to be
+`Atomics.wait` backoff, retryable `ServerError` after the 10s `UPDATE_ATTRIBUTES_LOCK_TIMEOUT` — harper#2251; it used to be
 an unbounded `while (!tryLock()) {}` spin that pinned a worker core forever if the holder never
 released). Release is structural — `table()` releases in a single `finally` and `dropTable` uses
 `withUpdateAttributesLock` — so a throw inside the locked window cannot leak the lock; keep it
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -292,9 +292,11 @@ an unbounded `while (!tryLock()) {}` spin that pinned a worker core forever if t
 released). Release is structural — `table()` releases in a single `finally` and `dropTable` uses
 `withUpdateAttributesLock` — so a throw inside the locked window cannot leak the lock (regression
 suite: `unitTests/resources/updateAttributesLock.test.js`). Because the acquire can now throw,
-`table()` takes the lock _before_ it mutates the live `Table` (attributes, class metadata, index
-handles): losing the race then leaves this worker's in-memory schema exactly as it found it, and
-moving any mutation above that acquire reintroduces schema drift the catalog never saw. A
+`table()` takes the RocksDB lock _before_ it mutates the live `Table` (attributes, class metadata,
+index handles): losing the race then leaves this worker's in-memory schema exactly as it found it,
+and moving any mutation above that acquire reintroduces schema drift the catalog never saw. LMDB
+keeps the lazy acquire — its `exclusiveLock()` is an environment-wide write transaction that cannot
+time out, so taking it eagerly would stall every write to the database on an unchanged reload. A
 successful acquire that waited past `UPDATE_ATTRIBUTES_LOCK_SLOW_WAIT` (1s) warns once, since
 contention is otherwise invisible until it becomes a timeout. The locked
 sections MUST stay synchronous: the wait blocks the event loop, so an awaited operation inside
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -291,7 +291,12 @@ synchronous bounded wait (`acquireUpdateAttributesLock` in `Table.ts`: brief hot
 an unbounded `while (!tryLock()) {}` spin that pinned a worker core forever if the holder never
 released). Release is structural — `table()` releases in a single `finally` and `dropTable` uses
 `withUpdateAttributesLock` — so a throw inside the locked window cannot leak the lock (regression
-suite: `unitTests/resources/updateAttributesLock.test.js`). The locked
+suite: `unitTests/resources/updateAttributesLock.test.js`). Because the acquire can now throw,
+`table()` takes the lock _before_ it mutates the live `Table` (attributes, class metadata, index
+handles): losing the race then leaves this worker's in-memory schema exactly as it found it, and
+moving any mutation above that acquire reintroduces schema drift the catalog never saw. A
+successful acquire that waited past `UPDATE_ATTRIBUTES_LOCK_SLOW_WAIT` (1s) warns once, since
+contention is otherwise invisible until it becomes a timeout. The locked
 sections MUST stay synchronous: the wait blocks the event loop, so an awaited operation inside
 one would stall a concurrent acquirer to its deadline. And dropping then recreating a
 same-named table within one process requires @harperfast/rocksdb-js >= the column-family
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -290,7 +290,8 @@ synchronous bounded wait (`acquireUpdateAttributesLock` in `Table.ts`: brief hot
 `Atomics.wait` backoff, retryable `ServerError` after the 10s `UPDATE_ATTRIBUTES_LOCK_TIMEOUT` — harper#2251; it used to be
 an unbounded `while (!tryLock()) {}` spin that pinned a worker core forever if the holder never
 released). Release is structural — `table()` releases in a single `finally` and `dropTable` uses
-`withUpdateAttributesLock` — so a throw inside the locked window cannot leak the lock. The locked
+`withUpdateAttributesLock` — so a throw inside the locked window cannot leak the lock (regression
+suite: `unitTests/resources/updateAttributesLock.test.js`). The locked
 sections MUST stay synchronous: the wait blocks the event loop, so an awaited operation inside
 one would stall a concurrent acquirer to its deadline. And dropping then recreating a
 same-named table within one process requires @harperfast/rocksdb-js >= the column-family
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -285,9 +285,15 @@ A handful of design points are non-obvious and easy to break:
   _bypassed_ by the seam and is therefore covered at the **integration** level (`sse-listchanged.test.ts` N3
   record / N4 collection), not in unit tests.
 
-Two related traps: the create path's exclusive `update-attributes` lock is a synchronous spin
-lock (`while (!tryLock()) {}`), so any throw inside the create window must release it or every
-subsequent create on that database pins a worker at 100% CPU. And dropping then recreating a
+Two related traps: the create/schema-update path's exclusive `update-attributes` lock is a
+synchronous bounded wait (`acquireUpdateAttributesLock` in `Table.ts`: brief hot spin, then
+`Atomics.wait` backoff, `ServerError` after the 10s `LOCK_TIMEOUT` — harper#2251; it used to be
+an unbounded `while (!tryLock()) {}` spin that pinned a worker core forever if the holder never
+released). Release is structural — `table()` releases in a single `finally` and `dropTable` uses
+`withUpdateAttributesLock` — so a throw inside the locked window cannot leak the lock; keep it
+that way (regression suite: `unitTests/resources/updateAttributesLock.test.js`). The locked
+sections MUST stay synchronous: the wait blocks the event loop, so an awaited operation inside
+one would stall a concurrent acquirer to its deadline. And dropping then recreating a
 same-named table within one process requires @harperfast/rocksdb-js >= the column-family
 eviction fix (1.4.3 / rocksdb-js#<main PR>): older bindings keep the dropped column family's
 by-name registry entry alive whenever other worker threads hold handles, so the recreate
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -290,8 +290,7 @@ synchronous bounded wait (`acquireUpdateAttributesLock` in `Table.ts`: brief hot
 `Atomics.wait` backoff, retryable `ServerError` after the 10s `UPDATE_ATTRIBUTES_LOCK_TIMEOUT` — harper#2251; it used to be
 an unbounded `while (!tryLock()) {}` spin that pinned a worker core forever if the holder never
 released). Release is structural — `table()` releases in a single `finally` and `dropTable` uses
-`withUpdateAttributesLock` — so a throw inside the locked window cannot leak the lock; keep it
-that way (regression suite: `unitTests/resources/updateAttributesLock.test.js`). The locked
+`withUpdateAttributesLock` — so a throw inside the locked window cannot leak the lock. The locked
 sections MUST stay synchronous: the wait blocks the event loop, so an awaited operation inside
 one would stall a concurrent acquirer to its deadline. And dropping then recreating a
 same-named table within one process requires @harperfast/rocksdb-js >= the column-family
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -49,6 +49,7 @@ import {
 	ServerError,
 	AccessViolation,
 	ValidationError,
+	UpdateAttributesLockTimeoutError,
 	type ValidationIssue,
 } from '../utility/errors/hdbError.ts';
 import * as signalling from '../utility/signalling.ts';
@@ -134,21 +135,13 @@ const CACHEABLE_STATUS_CODES = new Set([200, 203, 204, 206, 300, 301, 308, 404, 
 envMngr.initSync();
 const LMDB_PREFETCH_WRITES = envMngr.get(CONFIG_PARAMS.STORAGE_PREFETCHWRITES);
 const LOCK_TIMEOUT = 10000;
+// This bounds schema-lock acquisition; LOCK_TIMEOUT bounds in-flight record writes during a drop.
 export const UPDATE_ATTRIBUTES_LOCK_TIMEOUT = 10000;
 const UPDATE_ATTRIBUTES_LOCK = 'update-attributes';
 // raw ASCII bytes are ordered-binary's encoding of the string, so this addresses the same native
 // lock as string-keyed tryLock/unlock calls
 const updateAttributesLockKey = Buffer.from(UPDATE_ATTRIBUTES_LOCK);
 const lockWait = new Int32Array(new SharedArrayBuffer(4));
-
-class UpdateAttributesLockTimeoutError extends ServerError {
-	code = 'UPDATE_ATTRIBUTES_LOCK_TIMEOUT';
-	retryable = true;
-	constructor(message: string) {
-		super(message, 503);
-		this.name = 'UpdateAttributesLockTimeoutError';
-	}
-}
 
 /** The wait blocks the event loop, so the locked section must stay synchronous. */
 export function acquireUpdateAttributesLock(
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -138,6 +138,9 @@ const LOCK_TIMEOUT = 10000;
 // This bounds schema-lock acquisition; LOCK_TIMEOUT bounds in-flight record writes during a drop.
 export const UPDATE_ATTRIBUTES_LOCK_TIMEOUT = 10000;
 const UPDATE_ATTRIBUTES_LOCK = 'update-attributes';
+// A successful acquire that blocked this long still starved this worker's event loop for that
+// whole span; without a trace, contention is only ever visible once it becomes a timeout (harper#2251).
+export const UPDATE_ATTRIBUTES_LOCK_SLOW_WAIT = 1000;
 // raw ASCII bytes are ordered-binary's encoding of the string, so this addresses the same native
 // lock as string-keyed tryLock/unlock calls
 const updateAttributesLockKey = Buffer.from(UPDATE_ATTRIBUTES_LOCK);
@@ -164,6 +167,11 @@ export function acquireUpdateAttributesLock(
 			if (waitTime < 16) waitTime *= 2;
 		}
 	}
+	const waited = performance.now() - startTime;
+	if (waited >= UPDATE_ATTRIBUTES_LOCK_SLOW_WAIT)
+		logger.warn?.(
+			`Acquired the exclusive '${UPDATE_ATTRIBUTES_LOCK}' lock on ${scopeDescription} after waiting ${Math.round(waited)}ms; this worker's event loop was blocked for that wait, and a holder that runs past ${UPDATE_ATTRIBUTES_LOCK_TIMEOUT}ms fails the update outright`
+		);
 }
 
 export function releaseUpdateAttributesLock(rootStore: RocksDatabase) {
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -138,8 +138,7 @@ const LOCK_TIMEOUT = 10000;
 // This bounds schema-lock acquisition; LOCK_TIMEOUT bounds in-flight record writes during a drop.
 export const UPDATE_ATTRIBUTES_LOCK_TIMEOUT = 10000;
 const UPDATE_ATTRIBUTES_LOCK = 'update-attributes';
-// A successful acquire that blocked this long still starved this worker's event loop for that
-// whole span; without a trace, contention is only ever visible once it becomes a timeout (harper#2251).
+// Contention is otherwise only visible once it becomes a timeout (harper#2251).
 export const UPDATE_ATTRIBUTES_LOCK_SLOW_WAIT = 1000;
 // raw ASCII bytes are ordered-binary's encoding of the string, so this addresses the same native
 // lock as string-keyed tryLock/unlock calls
@@ -168,10 +167,14 @@ export function acquireUpdateAttributesLock(
 		}
 	}
 	const waited = performance.now() - startTime;
+	// The caller cannot register its release until we return, so a throw here would leak the lock
+	// with no `finally` able to reach it.
 	if (waited >= UPDATE_ATTRIBUTES_LOCK_SLOW_WAIT)
-		logger.warn?.(
-			`Acquired the exclusive '${UPDATE_ATTRIBUTES_LOCK}' lock on ${scopeDescription} after waiting ${Math.round(waited)}ms; this worker's event loop was blocked for that wait, and a holder that runs past ${UPDATE_ATTRIBUTES_LOCK_TIMEOUT}ms fails the update outright`
-		);
+		try {
+			logger.warn?.(
+				`Acquired the exclusive '${UPDATE_ATTRIBUTES_LOCK}' lock on ${scopeDescription} after waiting ${Math.round(waited)}ms; this worker's event loop was blocked for that wait, and a holder that runs past ${UPDATE_ATTRIBUTES_LOCK_TIMEOUT}ms fails the update outright`
+			);
+		} catch {}
 }
 
 export function releaseUpdateAttributesLock(rootStore: RocksDatabase) {
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -134,19 +134,17 @@ envMngr.initSync();
 const LMDB_PREFETCH_WRITES = envMngr.get(CONFIG_PARAMS.STORAGE_PREFETCHWRITES);
 export const LOCK_TIMEOUT = 10000;
 const UPDATE_ATTRIBUTES_LOCK = 'update-attributes';
-// Pre-encoded lock key, hoisted so the retry loop below does not re-marshal the string on every
-// tryLock attempt. ordered-binary encodes a plain-ASCII string as its raw bytes, so this Buffer
-// addresses the same native lock as the string form.
+// raw ASCII bytes are ordered-binary's encoding of the string, so this addresses the same native
+// lock as string-keyed tryLock/unlock calls
 const updateAttributesLockKey = Buffer.from(UPDATE_ATTRIBUTES_LOCK);
 const lockWait = new Int32Array(new SharedArrayBuffer(4));
 
 /**
- * Acquires the exclusive cross-thread 'update-attributes' lock that serializes schema/attribute
- * updates (and table create/drop) per database. Synchronous by design: the callers cannot yield to
- * the event loop, and a release by another thread is visible to the retry without yielding (the
- * lock lives in native shared state). Spins hot only briefly, then blocks this thread in bounded
- * waits, and throws after `timeout` instead of spinning forever — a holder that never releases
- * (e.g. a crashed schema operation) previously pinned this worker's core permanently (harper#2251).
+ * Acquires the exclusive cross-thread 'update-attributes' lock serializing schema/attribute updates
+ * (and table create/drop) per database. The wait is synchronous and blocks the event loop, so the
+ * locked section must stay synchronous too — a release by another thread is observable without
+ * yielding, but any release requiring this thread's own event loop can never arrive. Throws after
+ * `timeout` rather than spinning forever on a holder that never releases (harper#2251).
  */
 export function acquireUpdateAttributesLock(
 	rootStore: RocksDatabase,
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -132,7 +132,57 @@ const MAX_INFLIGHT_EVICTION_BATCHES = 4;
 const CACHEABLE_STATUS_CODES = new Set([200, 203, 204, 206, 300, 301, 308, 404, 405, 410, 414, 501]);
 envMngr.initSync();
 const LMDB_PREFETCH_WRITES = envMngr.get(CONFIG_PARAMS.STORAGE_PREFETCHWRITES);
-const LOCK_TIMEOUT = 10000;
+export const LOCK_TIMEOUT = 10000;
+const UPDATE_ATTRIBUTES_LOCK = 'update-attributes';
+// Pre-encoded lock key, hoisted so the retry loop below does not re-marshal the string on every
+// tryLock attempt. ordered-binary encodes a plain-ASCII string as its raw bytes, so this Buffer
+// addresses the same native lock as the string form.
+const updateAttributesLockKey = Buffer.from(UPDATE_ATTRIBUTES_LOCK);
+const lockWait = new Int32Array(new SharedArrayBuffer(4));
+
+/**
+ * Acquires the exclusive cross-thread 'update-attributes' lock that serializes schema/attribute
+ * updates (and table create/drop) per database. Synchronous by design: the callers cannot yield to
+ * the event loop, and a release by another thread is visible to the retry without yielding (the
+ * lock lives in native shared state). Spins hot only briefly, then blocks this thread in bounded
+ * waits, and throws after `timeout` instead of spinning forever — a holder that never releases
+ * (e.g. a crashed schema operation) previously pinned this worker's core permanently (harper#2251).
+ */
+export function acquireUpdateAttributesLock(
+	rootStore: RocksDatabase,
+	scopeDescription: string,
+	timeout = LOCK_TIMEOUT
+) {
+	if (rootStore.tryLock(updateAttributesLockKey)) return;
+	const startTime = Date.now();
+	let waitTime = 1;
+	while (!rootStore.tryLock(updateAttributesLockKey)) {
+		const elapsed = Date.now() - startTime;
+		if (elapsed >= timeout) {
+			throw new ServerError(
+				`Timed out after ${elapsed}ms waiting for the exclusive '${UPDATE_ATTRIBUTES_LOCK}' lock on ${scopeDescription}; the lock holder never released it, so this schema/attribute update cannot proceed`
+			);
+		}
+		if (elapsed >= 2) {
+			Atomics.wait(lockWait, 0, 0, Math.min(waitTime, timeout - elapsed));
+			if (waitTime < 16) waitTime *= 2;
+		}
+	}
+}
+
+export function releaseUpdateAttributesLock(rootStore: RocksDatabase) {
+	rootStore.unlock(updateAttributesLockKey);
+}
+
+/** Runs `callback` under the exclusive 'update-attributes' lock, releasing it on every exit path. */
+export function withUpdateAttributesLock<T>(rootStore: RocksDatabase, scopeDescription: string, callback: () => T): T {
+	acquireUpdateAttributesLock(rootStore, scopeDescription);
+	try {
+		return callback();
+	} finally {
+		releaseUpdateAttributesLock(rootStore);
+	}
+}
 // Tolerate a redundant column family drop. Drops are broadcast to every worker
 // thread and each holds its own handle to the same underlying family, so a
 // concurrent worker may already have dropped it; the storage engine reports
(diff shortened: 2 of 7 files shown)


resources/Table.ts
```diff
@@ -132,7 +132,55 @@ const MAX_INFLIGHT_EVICTION_BATCHES = 4;
...
+) {
+	if (rootStore.tryLock(updateAttributesLockKey)) return;
+	const startTime = Date.now();
+	let waitTime = 1;
+	while (!rootStore.tryLock(updateAttributesLockKey)) {
+		const elapsed = Date.now() - startTime;
+		if (elapsed >= timeout) {
+			throw new ServerError(
+				`Timed out after ${elapsed}ms waiting for the exclusive '${UPDATE_ATTRIBUTES_LOCK}' lock on ${scopeDescription}; the lock holder never released it, so this schema/attribute update cannot proceed`
+			);
```
cb1kenobi: **Medium: the deadline surfaces transient contention as a non-retryable 500**

`ServerError` defaults to `statusCode: 500` (`utility/errors/hdbError.ts:65`), so a waiter that loses a 10s race gets a permanent-looking failure for a condition that is transient by construction — another thread's DDL simply ran long. The same file already establishes the opposite precedent for exactly this shape: `IndexRebuildingError` is documented there as a distinct, retryable 503 so callers can tell a transient condition apart from a permanent failure and retry. Cluster schema replication, component load, and `create_table` / `add_attribute` all reach this throw.

This is also where the fixed-deadline trade-off lands. A legitimately long critical section (`completeInterruptedDrop` over many column families on saturated I/O) now fails waiters where the old spin eventually converged; making the error retryable is the cheap half of that decision and does not require adding holder liveness.

Suggested fix:
```suggestion
			throw new ServerError(
				`Timed out after ${elapsed}ms waiting for the exclusive '${UPDATE_ATTRIBUTES_LOCK}' lock on ${scopeDescription}; the lock holder never released it, so this schema/attribute update cannot proceed`,
				503
			);
```

—
_Generated by Barber AI_

resources/Table.ts
```diff
@@ -132,7 +132,55 @@ const MAX_INFLIGHT_EVICTION_BATCHES = 4;
...
+	rootStore: RocksDatabase,
+	scopeDescription: string,
+	timeout = LOCK_TIMEOUT
+) {
+	if (rootStore.tryLock(updateAttributesLockKey)) return;
+	const startTime = Date.now();
+	let waitTime = 1;
+	while (!rootStore.tryLock(updateAttributesLockKey)) {
+		const elapsed = Date.now() - startTime;
+		if (elapsed >= timeout) {
```
cb1kenobi, line 159: **Medium: a deadline throw leaves this worker's in-memory schema ahead of the catalog** _(static analysis — not reproduced dynamically)_

`table()` mutates the live `Table` before it ever asks for the lock — `Table.attributes.splice(...)`, `description`, `properties`, `hidden`, `cacheControl` in the `if (Table)` branch — and `openIndex()` runs at `databases.ts:1868`, ahead of the `exclusiveLock()` at `databases.ts:1884`. When this deadline fires, that work is already done and nothing unwinds it: `describe` reports attributes that were never persisted, the freshly opened index dbi is never assigned into `Table.indices` (so `updateIndices()` iterates `indices` and silently skips it — writes land unindexed), and no catalog row exists for it. The throw also skips the `hasChanges` block, so `schemaVersion++` / `signalSchemaChange` never run and nothing prompts a reload.

It does self-heal on the next successful `table()` for that table (the descriptor comparison reads the catalog, not `Table.attributes`), so this is drift rather than corruption — but the deadline is a new, load-triggered way in, where previously only an explicit error reached this state. The PR description raises this; flagging it inline so the decision is recorded on the line.

Suggested fix: on a lock-*acquire* failure specifically, mark the table stale (or `resetDatabases()`) before rethrowing, so the next access reloads from the catalog rather than serving the half-applied in-memory schema.

—
_Generated by Barber AI_

resources/Table.ts
```diff
@@ -132,7 +132,55 @@ const MAX_INFLIGHT_EVICTION_BATCHES = 4;
...
+			throw new ServerError(
+				`Timed out after ${elapsed}ms waiting for the exclusive '${UPDATE_ATTRIBUTES_LOCK}' lock on ${scopeDescription}; the lock holder never released it, so this schema/attribute update cannot proceed`
+			);
+		}
+		if (elapsed >= 2) {
```
cb1kenobi, line 164: **Low: a long-but-successful wait leaves no trace**

harper#2251's complaint was that the wedge "logs nothing". The deadline fixes the terminal case, but a wait that blocks this worker's event loop for 9.9s and then succeeds is still completely silent — no log, no metric. The contention that precedes a timeout is invisible until it becomes one, which is exactly the signal that would have shortened the original incident.

Suggested fix: warn once when the wait crosses a threshold (~1s), carrying `scopeDescription` and the elapsed time — a `let warned = false` guard inside this branch is enough.

—
_Generated by Barber AI_

resources/Table.ts
```diff
@@ -132,7 +132,55 @@ const MAX_INFLIGHT_EVICTION_BATCHES = 4;
...
+
+export function releaseUpdateAttributesLock(rootStore: RocksDatabase) {
+	rootStore.unlock(updateAttributesLockKey);
+}
+
+/** Runs `callback` under the exclusive 'update-attributes' lock, releasing it on every exit path. */
+export function withUpdateAttributesLock<T>(rootStore: RocksDatabase, scopeDescription: string, callback: () => T): T {
+	acquireUpdateAttributesLock(rootStore, scopeDescription);
+	try {
+		return callback();
```
cb1kenobi: **Low: the generic signature lets an async callback release the lock early**

`callback: () => T` also accepts `() => Promise<T>`, and `finally` then fires the moment the promise is *created*, not when it settles — the lock is released while the "locked" section is still running, silently. Today's only caller (`dropTable`) is synchronous and DESIGN.md now states the locked section must stay synchronous, but this is a newly exported helper whose type signature invites precisely the mistake the doc warns about.

Suggested fix:
```suggestion
	try {
		const result = callback();
		if (result && typeof (result as any).then === 'function')
			throw new Error(
				`withUpdateAttributesLock callback must be synchronous (${scopeDescription}); the lock wait blocks the event loop`
			);
		return result;
```

—
_Generated by Barber AI_

resources/databases.ts
```diff
@@ -1555,113 +1555,121 @@ export function table<TableResourceType>(tableDefinition: TableDefinition): Tabl
...
+				);
+			}
+			markInternalDbiNonVersioned(attributesDbi);
+
+			exclusiveLock(); // get an exclusive lock on the database so we can verify that we are the only thread creating the table (and assigning the table id)
+			const existingTableMeta = (attributesDbi as any).getSync(dbiName);
+			if (existingTableMeta && !existingTableMeta.dropping) {
+				// table was created while we were setting up; the lock is not reentrant, so release
+				// before the recursive reload
+				releaseLock();
```
cb1kenobi, line 1670: **Low: this release has no regression test — deleting it leaves the new suite fully green**

Mutation-checked against this head (source edit, `tsc -p tsconfig.build.json` rebuild, marker verified in `dist/`): replacing this `releaseLock()` with a no-op keeps all six tests in `unitTests/resources/updateAttributesLock.test.js` passing, and the full `unitTests/resources` suite green too. It is the single line that makes the recursive `table(tableDefinition)` two lines below safe — without it the reload re-enters `exclusiveLock()` on a lock this same thread already holds, and since the lock is not reentrant that is a guaranteed self-deadlock (pre-PR: the permanent wedge from harper#2251; post-PR: a 10s `ServerError` on every concurrently-created table).

Suggested fix: drive the branch with a one-shot stub of the catalog `getSync` that reports the table as concurrently created and restores the original from inside itself, so the recursion proceeds normally; then assert the create succeeds *and* `rootStore.tryLock('update-attributes')` still succeeds afterwards.

—
_Generated by Barber AI_

unitTests/resources/updateAttributesLock.test.js
```diff
@@ -0,0 +1,152 @@
...
+			PrimaryRocksDatabase.prototype.removeSync = originalRemoveSync;
+		}
+		assert.ok(rootStore.tryLock(LOCK_KEY), 'exclusive lock must not leak when table() throws');
+		rootStore.unlock(LOCK_KEY);
+		// a subsequent schema update must succeed rather than wedge — restore the original definition
+		table({ table: 'LockLeak', database: TEST_DB, attributes });
+	});
+
+	it('a real table() operation fails with ServerError after the full deadline when the lock is wedged', function () {
+		this.timeout(40000);
```
cb1kenobi: **Low: if the deadline regresses, this file hangs instead of failing**

Verified by mutation: with the `elapsed >= timeout` throw disabled, this suite does not fail — it hangs indefinitely. `Atomics.wait` blocks the event loop, so mocha's per-test timeout can never fire, and `.mocharc.json` sets `"timeout": 0`. CI would sit until the job-level timeout with no failing test to point at, which is a weak signal for the one regression these tests exist to catch.

Suggested fix: run the two wedge cases (this one and the 250ms helper case at line 37) in a forked child with a hard wall-clock kill in the parent, so a lost deadline reports as a failure rather than a hang.

—
_Generated by Barber AI_

unitTests/resources/updateAttributesLock.test.js
```diff
@@ -0,0 +1,152 @@
...
+	it('waits for a briefly-held lock and acquires once the holding thread releases', async function () {
+		this.timeout(30000);
+		const workerThread = new Worker(__dirname + '/updateAttributesLock-thread.js', {
+			workerData: { addPorts: [] },
+		});
+		try {
+			const held = await new Promise((resolve, reject) => {
+				workerThread.once('message', resolve);
+				workerThread.once('error', reject);
+				workerThread.postMessage({ type: 'hold-lock', holdTime: 300 });
```
cb1kenobi: **Nit: the 300ms hold leaves only ~100ms of slack for the `elapsed >= 200` lower bound**

The worker starts its 300ms release timer *before* `postMessage({type:'held'})`, but `startTime` (line 140) is stamped only once that message reaches the main thread. Everything spent in that hop comes out of the margin, so on a loaded runner the assertion at line 146 can see <200ms and fail even though the wait behaved correctly. AGENTS.md calls this timing shape out as the root cause of a class of flakiness (harper#1138).

The lower bound is worth keeping — it is what distinguishes "waited for the holder" from "took a different lock" — so widen the margin rather than drop it:
```suggestion
				workerThread.postMessage({ type: 'hold-lock', holdTime: 1000 });
```

—
_Generated by Barber AI_

resources/Table.ts
```diff
@@ -132,7 +132,55 @@ const MAX_INFLIGHT_EVICTION_BATCHES = 4;
 const CACHEABLE_STATUS_CODES = new Set([200, 203, 204, 206, 300, 301, 308, 404, 405, 410, 414, 501]);
 envMngr.initSync();
 const LMDB_PREFETCH_WRITES = envMngr.get(CONFIG_PARAMS.STORAGE_PREFETCHWRITES);
-const LOCK_TIMEOUT = 10000;
+export const LOCK_TIMEOUT = 10000;
```
cb1kenobi: **Nit: `LOCK_TIMEOUT` now serves three unrelated deadlines**

Exporting it is fine, but the same constant is now the per-record lock timeout (`Table.ts:5887`), `dropTable()`'s in-flight-write drain (`Table.ts:1509`), and the schema-lock deadline — so tuning any one of them silently retunes the other two.

Suggested fix: a separate `UPDATE_ATTRIBUTES_LOCK_TIMEOUT = 10_000` keeps today's value while letting them move independently.

—
_Generated by Barber AI_

resources/Table.ts
kriszyp: Addressed in de8ea03b2: the timeout is now a dedicated ServerError subclass with status 503, retryable=true, and code=UPDATE_ATTRIBUTES_LOCK_TIMEOUT; both deadline paths assert the wire-facing fields.

resources/Table.ts
kriszyp: Addressed across b6cbfb0ee and de8ea03b2: JavaScript misuse is detected and its orphaned rejection is contained/logged, while the TypeScript signature rejects Promise-returning callbacks; test:types pins the negative contract.

resources/databases.ts
kriszyp: Addressed in b6cbfb0ee: a one-shot catalog getSync seam drives the concurrently-created branch, proves the recursive reload completes, and verifies the non-reentrant lock is free afterward.

unitTests/resources/updateAttributesLock.test.js
kriszyp: Addressed in b6cbfb0ee: both wedged deadline cases now run in a worker with a parent-side hard watchdog, so a missing deadline fails the test instead of hanging the Mocha process.

unitTests/resources/updateAttributesLock.test.js
kriszyp: Addressed in de8ea03b2: lock release is now explicitly message-driven after the main thread verifies the Buffer key sees the string-keyed lock as held; the race-prone elapsed lower bound is gone.

resources/Table.ts
kriszyp: Addressed in b6cbfb0ee: UPDATE_ATTRIBUTES_LOCK_TIMEOUT is now a dedicated 10s constant; the unrelated record-lock and drop-drain deadlines continue using LOCK_TIMEOUT.

resources/Table.ts
kriszyp: Confirmed and reproduced by code trace. This is the remaining major owner decision: resetDatabases() is simple but globally rescans while another thread owns the schema lock; targeted rollback is local but must restore every metadata field and account for opened index handles. The PR description now presents both options; leaving this unresolved pending direction.

resources/Table.ts
kriszyp: Confirmed. A one-shot warning improves incident visibility but can fan out from every worker during legitimate long DDL, and no threshold/log level is specified. The PR description calls this out; leaving unresolved pending owner direction.

resources/Table.ts
```diff
@@ -133,6 +134,73 @@ const CACHEABLE_STATUS_CODES = new Set([200, 203, 204, 206, 300, 301, 308, 404,
...
+		if (elapsed >= timeout) {
+			throw new UpdateAttributesLockTimeoutError(
+				`Timed out after ${Math.round(elapsed)}ms waiting for the exclusive '${UPDATE_ATTRIBUTES_LOCK}' lock on ${scopeDescription}; the lock holder did not release it before the deadline, so this schema/attribute update cannot proceed`
+			);
+		}
+		if (elapsed >= 2) {
+			Atomics.wait(lockWait, 0, 0, Math.min(waitTime, timeout - elapsed));
+			if (waitTime < 16) waitTime *= 2;
+		}
+	}
```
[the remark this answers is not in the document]
kriszyp: This duplicates the still-open discussion_r3847961723. The code trace confirms the finding; it is intentionally pending the owner choice documented in the PR description, so this thread remains unresolved.

resources/Table.ts
```diff
@@ -133,6 +134,73 @@ const CACHEABLE_STATUS_CODES = new Set([200, 203, 204, 206, 300, 301, 308, 404,
...
+}
+
+export function releaseUpdateAttributesLock(rootStore: RocksDatabase) {
+	rootStore.unlock(updateAttributesLockKey);
```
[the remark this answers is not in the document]
kriszyp: This duplicates the still-open discussion_r3847961726. The observability tradeoff is documented in the PR description and intentionally awaits an owner-selected threshold/log policy, so this thread remains unresolved.

unitTests/resources/updateAttributesLock.test.js
```diff
@@ -0,0 +1,212 @@
...
+			assert.ok(held.acquired, 'worker thread should have taken the lock');
+			const bufferKeyAcquired = rootStore.tryLock(Buffer.from(LOCK_KEY));
+			if (bufferKeyAcquired) rootStore.unlock(Buffer.from(LOCK_KEY));
+			assert.strictEqual(bufferKeyAcquired, false, 'Buffer and string keys must address the same held lock');
+			workerThread.postMessage({ type: 'release-lock', holdTime: 300 });
+			const startTime = Date.now();
+			acquireUpdateAttributesLock(rootStore, `table '${TEST_DB}.Contended'`);
+			const elapsed = Date.now() - startTime;
+			releaseUpdateAttributesLock(rootStore);
+			assert.ok(elapsed < 5000, `should acquire promptly after release, took ${elapsed}ms`);
```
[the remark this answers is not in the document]
kriszyp: Fixed in 861739551: the deterministic message-driven release test now restores a 200ms lower-bound assertion, so an accidental immediate acquire fails while retaining ample scheduler slack.

resources/databases.ts
cb1kenobi: Re-checked at `86173955` — resolved, and the new test has teeth.

Re-ran the same mutation against this head (replace `releaseLock()` here with a no-op, `tsc -p tsconfig.build.json`, marker confirmed as the single change in `dist/`). It is now **killed**: `releases the lock before recursively reloading a concurrently-created table` fails with

```
UpdateAttributesLockTimeoutError: Timed out after 10004ms waiting for the exclusive 'update-attributes' lock on table 'test.ConcurrentCreate'
  at exclusiveLock (resources/databases.ts:2011)
  at table (resources/databases.ts:1660)
  at table (resources/databases.ts:1665)
```

— the self-deadlock this line prevents, reported as a failing test rather than a green run. Unmutated, the suite is 8/8. Nice job.

—
_Generated by Barber AI_

unitTests/resources/updateAttributesLock.test.js
cb1kenobi: Re-checked at `86173955` — resolved.

Re-ran the same mutation (disable the `elapsed >= timeout` throw, rebuild, marker isolated to the one site). The suite no longer hangs: moving both deadline cases into `runWorkerAction` — worker does the blocking wait, parent keeps a live event loop and a hard watchdog — turns the regression into a clean failure in 26s:

```
1) throws ServerError after the deadline when the lock is never released:
   Error: worker action 'helper-deadline' exceeded its 5000ms watchdog
2) a real table() operation fails with ServerError after the full deadline when the lock is wedged:
   Error: worker action 'table-deadline' exceeded its 20000ms watchdog
```

Watchdog margins look comfortable: the 20s budget covers an observed 10.19s, of which only ~0.2s is variable (worker boot + the follow-up create).

—
_Generated by Barber AI_

DESIGN.md
```diff
@@ -285,9 +285,14 @@ A handful of design points are non-obvious and easy to break:
...
 
-Two related traps: the create path's exclusive `update-attributes` lock is a synchronous spin
-lock (`while (!tryLock()) {}`), so any throw inside the create window must release it or every
-subsequent create on that database pins a worker at 100% CPU. And dropping then recreating a
+Two related traps: the create/schema-update path's exclusive `update-attributes` lock is a
+synchronous bounded wait (`acquireUpdateAttributesLock` in `Table.ts`: brief hot spin, then
+`Atomics.wait` backoff, retryable `ServerError` after the 10s `UPDATE_ATTRIBUTES_LOCK_TIMEOUT` — harper#2251; it used to be
+an unbounded `while (!tryLock()) {}` spin that pinned a worker core forever if the holder never
+released). Release is structural — `table()` releases in a single `finally` and `dropTable` uses
+`withUpdateAttributesLock` — so a throw inside the locked window cannot leak the lock. The locked
```
cb1kenobi: **Nit: this paragraph now points readers at the wrong regression suite**

This push dropped `keep it that way (regression suite: `unitTests/resources/updateAttributesLock.test.js`)` from the release sentence. The only suite the paragraph still names is three sentences later — "The regression suite for all of this is `unitTests/resources/dropTableGhost.test.js`" — which covers the column-family recreate trap, not the bounded wait or the structural release. So a reader who follows DESIGN.md to find the tests guarding this section now lands on the wrong file, and the new suite (which does have teeth — mutating away either `releaseLock()` or the deadline throw fails it) is undiscoverable from here.

Suggested fix:
```suggestion
`withUpdateAttributesLock` — so a throw inside the locked window cannot leak the lock (regression
suite: `unitTests/resources/updateAttributesLock.test.js`). The locked
```

—
_Generated by Barber AI_

kriszyp: Fixed in df132a939: the design note again names unitTests/resources/updateAttributesLock.test.js directly on the bounded-wait/structural-release invariant, while dropTableGhost.test.js remains scoped to the separate column-family recreate trap.

cb1kenobi: Re-checked at `df132a93` — resolved.

`df132a939` ("Link schema lock regression coverage") restores the pointer: the paragraph now reads "...so a throw inside the locked window cannot leak the lock (regression suite: `unitTests/resources/updateAttributesLock.test.js`)". `dropTableGhost.test.js` remains named separately for the column-family trap it actually covers, so the two suites are no longer conflated.

—
_Generated by Barber AI_

resources/Table.ts
kriszyp: Resolved in \`ca336ba33c04\` by the owner-selected option: targeted no-drift recovery, implemented as acquire-before-mutate rather than reconstruct-after.

`table()` now takes the RocksDB `update-attributes` lock at the top of the `if (Table)` branch, before `Table.attributes.splice(...)`/`description`/`properties`/`hidden`/`cacheControl` and before the `openIndex()` in the attribute loop. A timeout therefore throws with nothing mutated — no unpersisted attributes in `describe`, no index dbi missing from `Table.indices`, no skipped `schemaVersion` bump to reason about. No `resetDatabases()` and no snapshot/restore.

LMDB deliberately keeps the lazy acquire: its `exclusiveLock()` opens an environment-wide write transaction that cannot time out, so an eager take would stall every write to the database on an unchanged reload without protecting against anything.

Covered by the reworked full-deadline regression, which now wedges the lock from a **second thread** (the production failure mode, rather than waiting out a self-held lock) and asserts attributes, index handles, `schemaVersion` and class metadata are byte-identical across the timed-out call. Fails-on-base with just that one line reverted: `+ 'added'` in the live attribute list, `indices: []`.

The residual cost — every unchanged existing-table reload now takes the lock for its whole reconciliation pass — is called out in the PR description under `## For the human reviewer`.

— Claude Opus

resources/Table.ts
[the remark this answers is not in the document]
kriszyp: Resolved in \`ca336ba33c04\` by the owner-selected option: targeted no-drift recovery, implemented as acquire-before-mutate rather than reconstruct-after.

`table()` now takes the RocksDB `update-attributes` lock at the top of the `if (Table)` branch, before `Table.attributes.splice(...)`/`description`/`properties`/`hidden`/`cacheControl` and before the `openIndex()` in the attribute loop. A timeout therefore throws with nothing mutated — no unpersisted attributes in `describe`, no index dbi missing from `Table.indices`, no skipped `schemaVersion` bump to reason about. No `resetDatabases()` and no snapshot/restore.

LMDB deliberately keeps the lazy acquire: its `exclusiveLock()` opens an environment-wide write transaction that cannot time out, so an eager take would stall every write to the database on an unchanged reload without protecting against anything.

Covered by the reworked full-deadline regression, which now wedges the lock from a **second thread** (the production failure mode, rather than waiting out a self-held lock) and asserts attributes, index handles, `schemaVersion` and class metadata are byte-identical across the timed-out call. Fails-on-base with just that one line reverted: `+ 'added'` in the live attribute list, `indices: []`.

The residual cost — every unchanged existing-table reload now takes the lock for its whole reconciliation pass — is called out in the PR description under `## For the human reviewer`.

— Claude Opus

resources/Table.ts
kriszyp: Resolved in `ca336ba33c04` with the owner-selected policy: one `warn` per acquire, threshold 1s.

`acquireUpdateAttributesLock` now logs once **after** a successful acquisition whose wait reached `UPDATE_ATTRIBUTES_LOCK_SLOW_WAIT` (1000ms), carrying `scopeDescription` and the elapsed time. Logging post-acquisition rather than with a `warned` flag inside the wait loop makes one-per-acquire structural — there is no path that emits twice, and the uncontended fast path never reaches it. Timeout failures already carry the retryable 503 and are logged by the server path, so they are not doubled up here.

One detail worth flagging: the warn is wrapped in `try {} catch {}`. The caller cannot register its release until the acquire returns, so a throwing logger — a broken transport, or a strict test double, as this suite itself installs — would have leaked the lock permanently with no `finally` able to reach it. That is precisely the failure mode this PR claims is structurally impossible, and the acquire was the one window not covered.

Covered by a new test that asserts an uncontended acquire stays silent, then a >1s contended acquire emits exactly one warning naming the table and the elapsed ms.

Accepted tradeoff, per the owner: warning bursts are possible during legitimate concurrent startup. Emission is self-capped below 1/s/thread since a warning is only reachable after a ≥1s wait.

— Claude Opus

resources/Table.ts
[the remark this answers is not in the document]
kriszyp: Resolved in `ca336ba33c04` with the owner-selected policy: one `warn` per acquire, threshold 1s.

`acquireUpdateAttributesLock` now logs once **after** a successful acquisition whose wait reached `UPDATE_ATTRIBUTES_LOCK_SLOW_WAIT` (1000ms), carrying `scopeDescription` and the elapsed time. Logging post-acquisition rather than with a `warned` flag inside the wait loop makes one-per-acquire structural — there is no path that emits twice, and the uncontended fast path never reaches it. Timeout failures already carry the retryable 503 and are logged by the server path, so they are not doubled up here.

One detail worth flagging: the warn is wrapped in `try {} catch {}`. The caller cannot register its release until the acquire returns, so a throwing logger — a broken transport, or a strict test double, as this suite itself installs — would have leaked the lock permanently with no `finally` able to reach it. That is precisely the failure mode this PR claims is structurally impossible, and the acquire was the one window not covered.

Covered by a new test that asserts an uncontended acquire stays silent, then a >1s contended acquire emits exactly one warning naming the table and the elapsed ms.

Accepted tradeoff, per the owner: warning bursts are possible during legitimate concurrent startup. Emission is self-capped below 1/s/thread since a warning is only reachable after a ≥1s wait.

— Claude Opus

resources/Table.ts
cb1kenobi: Re-checked at `ca336ba33` — resolved, and now proven dynamically.

`676258353` / `ca336ba33` move the RocksDB acquire (`if (rootStore instanceof RocksDatabase) exclusiveLock();`) above `Table.attributes.splice(...)` in `databases.ts`, so a deadline throw can no longer land after the live schema was mutated — losing the race is now a clean no-op instead of state to unwind. LMDB's `exclusiveLock()` deliberately stays lazy: it opens an environment-wide write transaction that cannot time out, so it was never exposed to this failure mode, and taking it eagerly would have stalled every write on an unchanged reload.

The new test `a real table() operation fails with ServerError after another thread wedges the lock for the full deadline` snapshots the live schema immediately before and immediately after the timed-out call and asserts `deepStrictEqual(after, before)` — this is now a real regression test for the exact drift this comment raised, not just a static fix. Ran unmutated: passes as part of 1683/0 in `unitTests/resources`.

Also re-verified the three regressions this PR previously fixed, dynamically, against this head:
- `releaseLock()` mutant (concurrent-create early return, `databases.ts:1670`) stays killed — mutated to a no-op marker, rebuilt, marker isolated to one line in `dist/`, and `releases the lock before recursively reloading a concurrently-created table` fails with `UpdateAttributesLockTimeoutError: Timed out after 10000ms ... on table 'test.ConcurrentCreate'` via `table → table → exclusiveLock → acquireUpdateAttributesLock`.
- Deadline mutation (`elapsed >= timeout` throw, `Table.ts:159`) still fails fast rather than hanging: disabled, rebuilt, marker isolated, and the two deadline tests fail cleanly in 35s via their parent-side watchdogs (`helper-deadline` 5000ms, `declare-under-wedge` 30000ms — this second test was restructured by this push to wedge the lock from another thread, hence the renamed watchdog).
- The 503 stays genuinely retryable end to end: `UpdateAttributesLockTimeoutError extends ServerError` with `statusCode 503`/`code`/`retryable`, and `serverHandlers.js` still copies `code`/`retryable` into the response body when `retryable !== undefined` — confirmed via `serverHandlers.test.js` (30/30 passing).

—
_Generated by Barber AI_

Issue #750, opened 2026-05-22
[Models] Derived caching table for @embed (low-latency writes + model-change backfill)


heskew:
## Scope

Follow-up to #632 (Phase 5 of #510). #632 shipped `@embed` with **sync-by-default**: the embedder runs during the transaction's `before` phase, so commit blocks on it. This is the right default for the common case (per @kriszyp on #632: "overwhelmingly the mode that should and would be used").

This issue tracks the **derived caching table** pattern for the cases where write latency matters: origin-table writes commit immediately, and a separate derived table holds the vector and follows sync mode against the origin.

> I am more inclined to think the right way to minimize write latency is to move the vector embedding into a derived caching table (that follows the sync mode, but the "origin" table doesn't need to wait for it).
> — @kriszyp on #632

## Why this is meaningful

Sync mode adds an HTTP round-trip (to the embedding backend) to every originating write that touches the source field. For latency-sensitive write paths (high-throughput ingest, user-facing POSTs), that's not acceptable even with a fast local backend. The derived-cache-table pattern keeps origin-write latency at LMDB-speed while still producing the vector in the background — without ad-hoc job infrastructure on top of `server/jobs/`.

## Four things, one primitive

The same iterate-records-and-rewrite primitive (a derived table that embeds by observing origin content changes) lands **four** capabilities #632 deferred:

1. **Low-latency writes** — origin commits without the embedder; derived table embeds and stores asynchronously.
2. **Re-embed-on-model-change backfill** — when `@embed(model: ...)` changes between deploys, existing rows need to be re-embedded through the new model. #632 today only re-INDEXES existing vectors via the schema version-change pathway; it does NOT re-embed. The derived table is the natural place to scan-and-rewrite.
3. **Proactive source-subscribe embedding** — closes a known gap in #632. A caching table sourced from an external source that pushes proactive `put`/`patch` events via `subscribe()` writes them with `isNotification: true`, which `buildEmbedBefore` skips (correctly, for cluster-replication receivers — but the source-push case carries content with no precomputed vector). The skip only bites when the table does not override `get()` (`shouldRevalidateEvents === false`); with a custom `get()`, subscribe events invalidate and the next read re-embeds via `getFromSource`. A derived table that embeds from observed origin content closes this regardless of how the origin was updated (user write, `getFromSource`, or subscribe push).
[...]

PR #2258, opened 2026-08-21
Make cluster-origin table definitions additive-only so a peer's partial schema snapshot cannot destroy locally declared attributes

kriszyp:
# Make cluster-origin table definitions additive-only so a peer's partial schema snapshot cannot destroy locally declared attributes

## What / why

`table()` treated every incoming attribute list as authoritative: the existing-table branch spliced `Table.attributes` down to the incoming list, and the catalog reconcile loop deleted the descriptors (and indexes) of any attribute the list omitted. That is correct for local schema authoring (`create_table`, `@table`, `defineTable`) — but replication also calls `table()` with definitions that are only a *snapshot of a peer's eventually-consistent view*: the DB_SCHEMA handshake announcement (harper-pro `ensureTableIfChanged`) and the replicated `define_schema` event (`Table.ts`). Such a snapshot can be captured mid-create on the peer (only the primary key registered yet) or applied by a worker whose thread-local `databases` map has not yet absorbed a concurrent local `create_table`.

When those two races lined up, a partial peer snapshot permanently destroyed attributes a local `create_table` had just successfully declared: the in-memory list was spliced down and the on-disk catalog descriptor deleted. Because the table was `schemaDefined`, the handshake's honor-local guard then refused to ever re-add the attribute from peers, so every later `search_by_value` failed with `unknown attribute` — the root cause of the harper-pro nightly `replicationLoad` failure (`replicate across many databases`, run 32457825567). The failing node's logs show `create_table db3.test [id, name]` on `main/0` at 07:20:19.146 returning success, followed at 07:20:19.199 by `http/1` `(Re)creating { table:'test', schemaDefined:true, attributes:[{id}] }` from a peer announcement, and from then on `Schema for 'db3.test' is defined locally, but attribute 'name: String' from '127.0.0.5' does not match local attribute which does not exist` until the search 500s.

The invariant this PR enforces: **a definition with `origin: 'cluster'` is applied additively** — peer attributes the local table lacks are added; locally declared attributes are never removed or redefined; catalog descriptors are never reconciled away; and the local `schemaDefined` declaration is never overridden, neither the live class flag nor the durable primary-key descriptor. The replicated `define_schema` path in `Table.ts` already merged additively and already passed `origin: 'cluster'` — this makes `table()` itself enforce the invariant that path assumed.

**Companion PR (required to close the flake):** harper-pro's `ensureTableIfChanged` must pass `origin: 'cluster'` — [HarperFast/harper-pro#750](https://github.com/HarperFast/harper-pro/pull/750) carries that one-hunk change plus the submodule bump. Without it, the guard here is unreachable from the caller that caused the incident (the cross-model review's reachability finding — correct, and answered by the pairing).

## Review round: what changed since the first pass

The additive-only rule left one gap and one blind spot, both found in review and both fixed here.

- **The rule now has a stated authority, not just a prohibition.** Where a catalog descriptor already exists, a cluster-origin caller's definition is [restated from that descriptor field by field](https://github.com/HarperFast/harper/pull/2258/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R371), in both directions — a field the descriptor dropped is dropped from the live attribute too, not just the fields it carries. `Object.assign` could only do the latter, so a newer declaration that *removed* a field saw the stale value written back. The same restatement is what stops a definition that omits `indexed` (or carries `indexNulls: true` for an index built without null entries) from leaving this worker with an unregistered or wrongly-configured index while every other node still maintains it.
- **Index crash-recovery is no longer skipped on this path.** The first version `continue`d past the recovery triggers, so an abandoned build's descriptor left [`isIndexing` pinned on](https://github.com/HarperFast/harper/pull/2258/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1893) with nothing left to clear it — every query on that attribute then failed with `IndexRebuildingError` for the life of the worker, which is the same "searches fail forever after a peer announcement" symptom this PR exists to remove. Recovery runs, and rebuilds the durable declaration read [under the exclusive lock](https://github.com/HarperFast/harper/pull/2258/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1904).
- **Every discarded peer difference is reported**, not only a `type` conflict — [the comparison](https://github.com/HarperFast/harper/pull/2258/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1625) treats an absent field and an explicitly falsy one as the same declaration (so a peer sending `indexed: false` is not noise) and reports differences in either direction, including a peer omitting an index the local schema declares.
- **A resolved `indexFormat` is persisted when the durable descriptor predates the field** — a field the descriptor lacks rather than a rewrite of one it has.
- **A legacy (harperdb 4.x) descriptor implies its index through `attribute` and carries no `indexed` field**, and the normalization that fixes that up sat *below* this block — so restating a declaration from one dropped `indexed` and the worker skipped registering the index. The normalization now runs before both paths.
- **The discard warn only names fields a peer can actually declare.** One field list served both the restatement and the warn, so the warn could report `indexNulls: local true, peer undefined` — a field the peer never sent, because it is derived here (index registration reads it off the descriptor, the reindex path defaults it to `true`, and no schema declares it). The restatement still needs it; [the warn now reads a set without it](https://github.com/HarperFast/harper/pull/2258/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1625).
- **DESIGN.md** now records the durable-descriptor authority, the recovery exception, and what an operator does about a node already damaged by the original bug (the fix is forward-only: only local authoring — `create_attribute` or redeploying the component — can write that descriptor back).

## For the human reviewer

- Judgment call — additive-only vs. versioned schema: peer definitions carry no version, so "newer wins" is not implementable today; additive-only trades delete-convergence for data safety. A locally dropped attribute can be resurrected by a stale peer handshake, and a peer's redefinition of an existing name is discarded. Both are pre-existing properties of the handshake's honor-local policy (and of `Table.ts`'s additive merge); the supported cluster-wide schema-change path remains deploying the schema to every node. A versioned schema exchange is the real long-term fix.
- Judgment call — fix location: patching harper-pro's `ensureTableIfChanged` alone cannot close the race, because another thread's `create_table` can commit its catalog writes after the caller's existence check and before `table()`'s reconcile. Only `table()` (exclusive lock + disk reload) can apply the definition safely.
- Judgment call — cluster origin skips the whole removal-reconcile scan, including `removeIndex`; stale index descriptors on this path persist until the next local authoring pass. Deliberately conservative.
- The `updateTable` emit's second argument becomes `false` for cluster-origin calls; no listener consumes that argument (verified by grep of all `'updateTable'` registrations), so this is inert.
- **Open question for you — `schemaDefined` on a cluster-origin *create*.** Review flagged that the `origin` gate protects the existing-table branch only: a table first materialized by a handshake persists `schemaDefined` with no gate, and a `schemaDefined: true` table can never receive further peer attributes (honor-local). Checked against harper-pro `main`: `ensureTableIfChanged` passes `schemaDefined: tableDefinition.schemaDefined` — the *peer's* explicit value, not this file's `true` default — and only calls `ensureTable` at all when the local table is **not** already `schemaDefined`. So a handshake-created table mirrors the peer's declaration by design. Gating the creation write on `origin !== 'cluster'` would make handshake-created tables dynamic instead; that is a behavior change for joining nodes and I did not take it unilaterally.
- **Declined — hydrating the live attribute list from a catalog scan.** An attribute missing from *both* this worker's map and the incoming definition still goes unregistered here until the schema-change signal resets the worker. Closing that would mean re-introducing a per-call catalog scan on the cluster path — the scan this PR removes because it is what destroyed descriptors — now as a union instead of a reconcile. It is a real improvement to the schema-lag window, but a separate design decision, and on `main` this same path *deleted* those descriptors rather than merely missing them.
- **Declined — taking the `update-attributes` lock for the descriptor read that drives index registration.** Read-then-decide without the lock, with writes re-reading under it, is the pre-existing shape of this whole loop; locking every registration read would hold a spin lock across the attribute loop for every handshake. The residual window is closed asynchronously by the schema-change signal.
- **Declined — computing the discarded-difference warn against the durable descriptor.** The warn reports what the *merge* discarded, so a difference the later restatement discards is not logged. Moving it into the per-attribute loop means threading the peer's values past the merge that dropped them, for a log line.
- **Declined — re-serializing the comparison fields in the discard warn.** The filter `JSON.stringify`s each redefinable field and the message loop serializes the surviving ones again. The `(local || peer)` guard means an absent field never reaches `JSON.stringify`, the message loop runs only when something was actually discarded, and `commonChanged` on the mainline persistence path already does the same on every attribute of every table on every load. Not worth a second comparison idiom on a diagnostic path.
- **Declined — hardening the warn's `JSON.stringify` against a `BigInt` or cyclic field.** Same reasoning, plus the exposure is not new: `commonChanged` already serializes peer-supplied `properties`/`elements`/`embed`, and `:1958` already serializes `indexed`. The only field the warn newly serializes that `commonChanged` compares with `!==` is `version`, which no code in this repo ever assigns.
- **Declined — `origin: 'cluster'` as a magic value** (an application literally named `cluster` would get additive-only semantics). `origin !== 'cluster'` already gates the `updateTable` emit on `main` and `Table.ts` already passes `origin: 'cluster'`, so this is a pre-existing convention, and a dedicated `additiveOnly` field is a cross-repo contract change (harper-pro is the other caller). Worth doing, not here — the thread is left open for your call. Two things I checked before declining, in case they change it: the sentinel also lands in `Table.origin` and in the persisted `primaryKeyAttribute.origins` list, but **neither has a reader** in core or harper-pro (`.origins` is written on two lines and read nowhere), so the leak into identity metadata is inert and the residual risk is exactly the name collision. And both call sites that would set `additiveOnly` are in PRs open right now (`Table.ts` here, `ensureTableIfChanged` in the companion), so doing it later costs a second pass over the same two lines rather than a new coordination.

## Verification

- `unitTests/resources/clusterSchemaMerge.test.js` — eight cases, all passing on both engines (RocksDB and `HARPER_STORAGE_ENGINE=lmdb`). Each new case was watched failing against the commit before its fix: the destructive apply (`attribute 'name' was removed by a cluster-origin definition`), the discarded-difference warn, [the abandoned-build recovery and its both-directions restatement](https://github.com/HarperFast/harper/pull/2258/changes?w=1#diff-20b444841b3df86be516d50126be68e0d1d789cc41b0c393b72fca2b2cfb710cR173), and [the omitted index / `indexNulls`](https://github.com/HarperFast/harper/pull/2258/changes?w=1#diff-20b444841b3df86be516d50126be68e0d1d789cc41b0c393b72fca2b2cfb710cR230) and legacy-descriptor cases.
- End-to-end route: harper-pro `integrationTests/cluster/replicationLoad.test.mjs` (the failing nightly suite) passed 4/4 against a build embedding the first version of this change plus the companion harper-pro change. The review-round commits are unit-covered only; the triggering race is a millisecond-wide cross-thread window, evidenced by the CI server-log forensics above rather than reproduced live.
- `npm run test:unit:resources`: 1679 passing, 16 pending, 3 failing — the three are pre-existing in this environment and reproduce identically with `resources/databases.ts` reverted to the base (`randomAccessFields` defaults ×2, `replayStructures`); a fourth (`vectorIndex` HNSW routing) is order-dependent and passes in isolation on both. `test:unit:main` cannot run in this environment at all (a stale machine-level `hdb_boot_properties.file` points the loader at an unrelated database; identical failure on base). CI is the authoritative gate.
- Cross-model pre-push review: twelve rounds (codex graded; gemini has exited 1 on this branch throughout; Cursor and domain legs pruned as a narrow low-risk delta in the last round). The majors — peer `schemaDefined` persisted into the durable descriptor, recovery reverting a newer declaration, `Object.assign` unable to drop removed fields, `indexNulls` outside the restatement — are all fixed with regression assertions. The final round returned four minors and no new majors: the `schemaDefined`-on-create open question above, the two serialization items declined just above, and "no end-to-end cluster test", which the harper-pro `replicationLoad` run answers.

- CI: **all 42 checks pass.** Getting there took re-running three flakes, none reachable from this diff, and one of them is worth reading about. Across four Node 22 runs on two commits of this branch it `Unit Test (Node.js v22)` failed three times — twice at `unitTests/resources/txn-tracking.test.js:61` and once at `vectorIndex.test.js:1009` — with Node 24 and 26 green on the same commits every time, before passing on the last re-run. `txn-tracking:61` is `assert(performedDBInteractions)` immediately after `await Promise.race([delay(50), result])`, i.e. a 50 ms wall-clock window; it does not reproduce on Node 22 here under either engine, standalone or in full-suite order, and [harper#2291](https://github.com/HarperFast/harper/pull/2291) is red on Node 22 only with a different timing test. This diff runs only on the `origin === 'cluster'` path, which neither test takes. The other two reds also passed on re-run: the Windows `set_configuration` `EPERM ... rename` race in the config writer, and a uWS integration shard where a worker that failed component load (`Database not open` at `componentLoader.ts:473`) exited with a live libuv poll handle and tripped Node's own `uv_loop_close() while having open handles` assertion, aborting the whole process. **`vectorIndex.test.js:1009` is worth reading on its own** — it has an issue ([harper#2306](https://github.com/HarperFast/harper/issues/2306)) that proposes relaxing the assertion, and I posted evidence there that the assertion is right and the search is wrong: the full-ef control side is exactly the true global top-10, while greedy descent returns global ranks 29-45 and 47-70 in that issue's own two samples — recall@10 of zero, from a different basin.

Complexity: moderate — a guarded behavior split on an existing parameter in one heavily-shared function, plus tests; risk is concentrated in `table()`'s existing-table branch semantics.

<sub>Review-Coverage: authored=claude; ran=gemini; adjudicated=domain; blocked=codex(timeout),claude(fallback)(out-of-budget); declined=cursor-grok,cursor-composer; rounds=13 @ 4e332df64033</sub>

<sub>Human-Review-Need: 3 (decisions: additive-only-vs-versioned-schema-exchange, descriptor-authoritative-in-both-directions, cluster-gate-scope-beyond-attributes, warn-per-difference-per-call, recovery-exception-writes-on-cluster-path, forward-only-remediation, race-simulated-by-direct-mutation) @ 4e332df64033</sub>


unitTests/resources/clusterSchemaMerge.test.js
```diff
@@ -0,0 +1,114 @@
+require('../testUtils');
+const assert = require('assert');
```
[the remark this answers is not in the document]
kriszyp: Fixed in 55afa6b87 — thanks!

— Claude Fable 5 (via KrAIs)

resources/databases.ts
```diff
@@ -1824,6 +1848,20 @@ export function table<TableResourceType>(tableDefinition: TableDefinition): Tabl
 				continue;
 			}
 
+			if (origin === 'cluster' && attributeDescriptor) {
+				// Cluster-origin definitions may only create missing descriptors — this worker's snapshot
+				// can lag a newer durable declaration, and rewriting (or reindexing from) it would revert
+				// that declaration. Still register the existing index for this Table instance.
+				if (attribute.indexed) {
+					const dbi = openIndex(dbiKey, rootStore, attribute);
+					if (attributeDescriptor.indexingPID) dbi.isIndexing = true;
```
cb1kenobi: **High: cluster-origin skip disables index crash-recovery and pins `isIndexing` on**

This line sets `dbi.isIndexing = true` from a descriptor whose `indexingPID` may belong to a dead process, then `continue`s past the recovery triggers below (`indexingFailed`, foreign `indexingPID`, stale `restartNumber`). Nothing drives the backfill afterwards, so `search.ts:454` throws `IndexRebuildingError` for every query on that attribute in this worker, indefinitely — the same "searches fail forever after a peer announcement" symptom this PR exists to fix.

Reproduced on this head vs. the merge base (`374408e3`): table with one indexed attribute and one record, descriptor patched to `{ indexingPID: 999999, restartNumber: 0 }`, then `table({ ..., origin: 'cluster' })`.
- base: `reindex test.RecoveryProbe.z: reason=crash-recovery(pid=999999),restart-number`, `isIndexing` clears to `false`
- head: no reindex triggered, `indices.z.isIndexing === true` and stays true

The three recovery signals are read from the durable descriptor, not from the caller's snapshot, so acting on them cannot revert a newer local declaration — they are safe to keep on the cluster path. Suggest not taking this skip when the descriptor shows `indexingFailed`, an `indexingPID !== process.pid`, or `restartNumber < currentRestartGeneration`, and letting those cases fall through to the existing handling.

—
_Generated by Barber AI_

resources/databases.ts
```diff
@@ -1586,12 +1586,29 @@ export function table<TableResourceType>(tableDefinition: TableDefinition): Tabl
...
+		if (origin === 'cluster') {
+			// A peer-derived definition (DB_SCHEMA handshake, replicated define_schema) can be built
+			// from a mid-create or stale snapshot of the remote table, so it is additive-only: keep
+			// every locally-declared attribute and add any the peer has that we don't. Only local
+			// schema authoring (create_table, @table, defineTable) may remove or redefine attributes.
+			const merged = Table.attributes.slice();
+			for (const attribute of attributes) {
+				const existing = merged.find((existingAttribute) => existingAttribute.name === attribute.name);
+				if (!existing) merged.push(attribute);
+				else if (attribute.type && existing.type && existing.type !== attribute.type)
```
cb1kenobi: **Medium: peer redefinitions are discarded silently unless `type` differs, so index sets diverge by arrival order**

The merge is name-only and first-writer-wins, and this warn covers only `type` — and only when both sides carry one. Every other discarded difference is dropped with no signal: `indexed` above all, plus `nullable`, `elements`, `properties`, `embed`. Applying the same set of peer definitions in a different order therefore leaves nodes with different schemas and nothing in the logs to say so.

Reproduced on this head, dynamic table, two peer definitions of `y` differing only by `indexed`:
- `PLAIN` then `IDX` → `indices.y` absent
- `IDX` then `PLAIN` → `indices.y` present

Same inputs, different order, permanently different index sets; `search_by_value` on `y` then uses an index on one node and full-scans on the other (or 404s under `allowFullScan: false`). The merge base is non-convergent here too, so this re-shapes a pre-existing limitation rather than introducing one — but on the base the divergence at least left a descriptor write behind, whereas here the discard is total and silent.

Cheapest improvement, and enough to make it diagnosable: warn on any discarded peer redefinition, not just a `type` mismatch.

—
_Generated by Barber AI_

DESIGN.md
```diff
@@ -135,6 +135,10 @@ When `signalSchemaChange('schema-change')` fires at the start of `runIndexing`,
...
... registered yet) or read from a worker whose thread-local map hasn't absorbed a concurrent local create. Such definitions are applied additively: attributes the local table lacks are added, locally declared attributes are never removed or redefined, catalog descriptors are never reconciled away, and the local `schemaDefined` declaration is never flipped. Before this was enforced, a partial peer snapshot racing a local create_table permanently deleted the just-declared attributes' descriptors — and because the table was `schemaDefined`, the handshake's honor-local guard then refused to ever re-add them from peers, so searches failed with "unknown attribute" forever after (harper-pro nightly `replicationLoad` flake). The deliberate cost of additive-only: attribute _drops_ and _redefinitions_ do not converge through schema gossip — a stale peer can resurrect a locally dropped attribute, and a peer's redefinition of an existing name is discarded (a type conflict logs a warn). Cluster-wide schema changes converge by applying the schema on every node (component deploy); a versioned schema exchange is the eventual fix. Regression coverage: `unitTests/resources/clusterSchemaMerge.test.js`.
```
cb1kenobi: **Medium: no remediation documented for nodes already damaged by this bug**

A node that already hit the incident has the attribute's catalog descriptor deleted and `schemaDefined: true` persisted. Upgrading to this change does not repair it: harper-pro's `ensureTableIfChanged` honor-local guard refuses to add an attribute on a `schemaDefined` table, so no peer can ever resupply the descriptor and `search_by_value` keeps failing with `unknown attribute`. The fix is forward-only, and the only detection is the pre-existing per-announcement error log.

This section is the right place to say what an operator does about an already-broken node: re-declare the table locally (`create_table`, or redeploy the schema component), since local authoring is now the only path permitted to write that descriptor back.

—
_Generated by Barber AI_

resources/databases.ts
```diff
@@ -1586,12 +1586,29 @@ export function table<TableResourceType>(tableDefinition: TableDefinition): Tabl
 		}
 		// it table already exists, get the split segments setting
 		if (splitSegments == undefined) splitSegments = Table.splitSegments;
+		if (origin === 'cluster') {
```
cb1kenobi: **Low: `'cluster'` is a magic value in a field that otherwise carries component identity**

`origin` is normally a component identifier: `Scope.ensureTable` sets `options.origin = this.#origin` (`Scope.ts:243`), which falls back to the application name (`Scope.ts:111`), and `table()` accumulates it into `primaryKeyAttribute.origins`. A data-safety invariant now hinges on that same string being exactly `'cluster'`, so an application named `cluster` would silently get additive-only semantics and stop persisting its own `schemaDefined` declaration.

Suggest a dedicated field on `TableDefinition` — e.g. `additiveOnly?: boolean`, set by the two peer callers — so the invariant cannot be reached by a name collision. An exported constant shared by both sides would be a smaller step.

—
_Generated by Barber AI_

resources/databases.ts
```diff
@@ -1824,6 +1848,20 @@ export function table<TableResourceType>(tableDefinition: TableDefinition): Tabl
 				continue;
 			}
 
+			if (origin === 'cluster' && attributeDescriptor) {
+				// Cluster-origin definitions may only create missing descriptors — this worker's snapshot
+				// can lag a newer durable declaration, and rewriting (or reindexing from) it would revert
+				// that declaration. Still register the existing index for this Table instance.
+				if (attribute.indexed) {
+					const dbi = openIndex(dbiKey, rootStore, attribute);
```
cb1kenobi: **Low: resolved `indexFormat` is no longer persisted on this path**

`openIndex` → `resolveIndexFormat` resolves an absent `indexFormat` to `'versioned'` for an empty custom-object index and arms the versioned encoder, but this block `continue`s before the `indexFormatNeedsPersist` write below — the write whose comment describes the un-persisted case as silent corruption (versioned nodes written, `'legacy'` re-derived from the now-non-empty store on the next load).

Narrow: needs an HNSW/custom-object index whose descriptor predates `indexFormat`, an empty store, and a cluster-origin call that opens it first. The disk-load path at `databases.ts:895` already has the same gap, so this narrows an existing self-healing write rather than opening a new hole. Static only — not reproduced.

Persisting `indexFormat` before the `continue` when `attributeDescriptor.indexFormat` is absent would close it without violating this block's rule: it creates a field the descriptor lacks rather than rewriting one it has.

—
_Generated by Barber AI_

resources/databases.ts
kriszyp: Leaving this one open for a human call rather than resolving it. `origin !== 'cluster'` already gates the `updateTable` emit on `main`, and `Table.ts` already passes `origin: 'cluster'`, so the magic value is a pre-existing convention that this PR leans on rather than introduces. A dedicated `additiveOnly` field is the right shape, but it is a cross-repo contract change — harper-pro's `ensureTableIfChanged` is the other caller — so it belongs in its own PR. Recorded under `## For the human reviewer`.

— Claude Opus

kriszyp: Leaving this one open for a human call rather than resolving it. `origin !== 'cluster'` already gates the `updateTable` emit on `main`, and `Table.ts` already passes `origin: 'cluster'`, so the magic value is a pre-existing convention that this PR leans on rather than introduces. A dedicated `additiveOnly` field is the right shape, but it is a cross-repo contract change — harper-pro's `ensureTableIfChanged` is the other caller — so it belongs in its own PR. Recorded under `## For the human reviewer`.

— Claude Opus

resources/databases.ts
```diff
@@ -348,6 +348,28 @@ _assignPackageExport('databases', databases);
 _assignPackageExport('tables', tables);
 
 const NEXT_TABLE_ID = Symbol.for('next-table-id');
+// The `commonChanged` comparison below, plus the two index flags — an index built without null entries
+// must not be re-registered as null-capable.
+const PEER_REDEFINABLE_FIELDS = [
+	'type',
+	'indexed',
+	'indexNulls',
```
cb1kenobi, line 363: **Nit: `indexNulls` in the shared field list makes the discard warn misreport**

`PEER_REDEFINABLE_FIELDS` serves two purposes: what `applyDurableDeclaration` restates from the descriptor (where `indexNulls` belongs — an index built without null entries must not be re-registered as null-capable) and what the merge loop reports as a discarded peer redefinition (where it does not). `indexNulls` is derived locally, not declared by peers: the reindex path sets it to `true` when it is `undefined`, and no peer definition carries it.

The result is a warn that blames the peer for a field it never sent. Reproduced against this head with a peer definition shaped like the one harper-pro#750 will send:

```
Ignoring peer redefinition of test.ProbeD.tag (indexNulls: local true, peer undefined); the local schema is authoritative
```

Not reachable today — `Table.ts:953` passes the live attribute objects themselves, so `existing === attribute` and nothing is ever reported. It becomes reachable when #750 lands and `ensureTableIfChanged` starts passing peer-derived objects on a type change.

Suggested fix: keep the one list for `applyDurableDeclaration`, and filter `indexNulls` out of the set the warn reports, so the message only names fields a peer can actually declare.

—
_Generated by Barber AI_

resources/databases.ts
```diff
@@ -1824,6 +1881,44 @@ export function table<TableResourceType>(tableDefinition: TableDefinition): Tabl
 				continue;
 			}
 
+			if (origin === 'cluster' && attributeDescriptor) {
+				// An existing descriptor is a local declaration this caller may not have seen yet, so it wins
+				// over the incoming definition and is never written back from it.
+				applyDurableDeclaration(attribute, attributeDescriptor);
```
cb1kenobi, line 2102: **Low: a v4-era legacy descriptor strips `indexed` off the live attribute on this path**

`applyDurableDeclaration` deletes any field the descriptor lacks. A harperdb@4.x-era descriptor stores the name under `attribute` and carries no `name` and no `indexed` — the fixup for that shape is at `:1923` (`if (attributeDescriptor?.attribute && !attributeDescriptor.name) attributeDescriptor.indexed = true`), which this block returns before reaching. So on the cluster path the legacy descriptor is read raw and `indexed` is deleted from the live attribute object in `Table.attributes`.

Verified dynamically at this head — the live attribute comes back as `{"name":"tag","type":"String","attribute":"tag","resolve":null,"indexNulls":true}` with `indexed` gone. Impact is bounded: `indices.tag` was registered by the earlier load and stays registered, and a subsequent write was still found by an index search, so queries and write-time indexing are unaffected. What is left is `describe_table` reporting the attribute as non-indexed, and a live list that no longer marks it indexed for any later pass that reads from it.

Suggested fix: normalize the legacy shape before comparing, e.g. apply the same `attribute && !name` fixup to `attributeDescriptor` above this call rather than at `:1923`, so both paths see one descriptor shape.

—
_Generated by Barber AI_

resources/databases.ts
cb1kenobi: Verified fixed at `74d6e6f2` — dynamically, not just by reading the diff.

Independent probe (dead `indexingPID` + stale `restartNumber`, with records inserted so the `hasExistingData` branch is actually entered):

```
reindex test.ProbeA.tag: reason=crash-recovery(pid=999999),restart-number
isIndexing after recovery: false
odd rows found: 4 / 4
```

That is the merge-base behavior restored on the cluster path. Mutating the descriptor-skip guard off still fails the new `recovers an abandoned index build...` test, so the recovery is pinned by a regression test rather than only by the fix. Resolving.

—
_Generated by Barber AI_

resources/databases.ts
cb1kenobi: The actionable half — widening the warn — is fixed at `74d6e6f2`. It now covers nine fields and fires in both directions, verified with two peers declaring `y` in opposite orders:

```
Ignoring peer redefinition of test.ProbeB_A.y (indexed: local true, peer undefined; ...)
Ignoring peer redefinition of test.ProbeB_B.y (indexed: local undefined, peer true)
```

The divergence itself is unchanged and still real (`indices.y` present on one node, absent on the other), but that is the documented additive-only trade-off, and the merge base was non-convergent too — so this reshapes the limitation rather than introducing it. It is now signalled instead of silent, which was the ask. Resolving.

One follow-on nit on the new warn is posted separately at `:356`.

—
_Generated by Barber AI_

DESIGN.md
cb1kenobi: Fixed at `74d6e6f2` — the new **Remediating a node damaged before this was enforced** paragraph documents both halves: the recurring `does not match local attribute which does not exist` warn as the detection signal, and re-declaring the attribute locally (`create_attribute`, or redeploying the component whose `@table` declares it) as the remediation.

Confirmed the remediation actually works at this head: after simulating the damage (descriptor removed, live attribute dropped, `schemaDefined` still true), a local re-declaration restores the catalog descriptor. Resolving.

—
_Generated by Barber AI_

resources/databases.ts
cb1kenobi: Fixed at `74d6e6f2`. The resolved `indexFormat` is persisted again on this path, and correctly scoped — it only writes when the descriptor lacks the field, re-reads under `exclusiveLock()` before writing, and so adds a field rather than rewriting one, which keeps the additive-only invariant intact. Resolving.

—
_Generated by Barber AI_

resources/databases.ts
kriszyp: Fixed — good catch, and the reasoning was exactly right: `indexNulls` is derived here (index registration reads it off the durable descriptor, and the reindex path defaults it to `true`), so no peer definition can carry it.

`PEER_REDEFINABLE_FIELDS` stays as the restatement set for `applyDurableDeclaration`, and the merge-loop warn now reads `PEER_DECLARABLE_FIELDS`, which is that list minus `indexNulls`. Regression test added with a peer definition shaped the way harper-pro#750 will send one (distinct objects, differing only in `type`): it asserts the warn still names `type` and does not name `indexNulls`. Verified failing against the previous commit, where it reproduced your message verbatim — `Ignoring peer redefinition of test.ClusterMergeDerived.tag (type: local "String", peer "Int"; indexNulls: local true, peer undefined)`.

— Claude Opus

resources/databases.ts
kriszyp: This one was already fixed — you were reading `74d6e6f2`, and `520b2028` (pushed just before your pass) does precisely what you suggest: it hoists the `attribute && !name` fixup above the cluster-origin block so both paths see one descriptor shape.

I found it the same way, from the index-registration side rather than the `describe_table` side, so the regression test there only asserted `indices.tag` survives. I have added the live-attribute assertion your note names — restating from a v4-era descriptor must leave `indexed` on the entry in `Table.attributes` — and confirmed it fails independently against the un-hoisted ordering.

— Claude Opus

cb1kenobi: Verified fixed at `97251c59` (landed in `520b2028`), and the fix is at the right level — the normalization moved to *before* the cluster-origin branch rather than being duplicated inside it:

```ts
if (attributeDescriptor?.attribute && !attributeDescriptor.name) attributeDescriptor.indexed = true; // legacy descriptor

if (origin === 'cluster' && attributeDescriptor) {
```

Checked that moving it earlier is strictly additive rather than a behavior change for anyone else: the only code between the new position and the old one is the cluster-origin block itself, and the primary-key branch above it already ends in `continue`, so every path that used to reach the old site still reaches the new one. `applyDurableDeclaration` now sees `indexed: true` and stops deleting it off the live attribute, so `describe_table` reports the index that was always actually registered.

The new test `keeps a legacy (v4-era) index descriptor registered` pins it.

One thing I checked and am *not* raising as a finding, just so it's on record as considered: the `abandonedIndexBuild` branch re-reads the descriptor fresh under the lock via `applyDurableDeclaration(attribute, attributesDbi.getSync(dbiKey) ?? attributeDescriptor)`, and that re-read copy does not get the legacy normalization. It isn't reachable — `abandonedIndexBuild` requires `indexingFailed`, a foreign `indexingPID`, or `restartNumber < currentRestartGeneration`, and a genuine v4-era descriptor carries none of those (`undefined < n` is false). So there's no bug today. Worth knowing the coupling exists if that condition ever grows a clause that a legacy descriptor could satisfy.

Generated by Barber AI

resources/databases.ts
cb1kenobi: Verified fixed at `97251c59`. The split is exactly the right one:

```ts
const PEER_DECLARABLE_FIELDS = PEER_REDEFINABLE_FIELDS.filter((field) => field !== 'indexNulls');
```

`PEER_DECLARABLE_FIELDS` is used only for the discard warn, while `applyDurableDeclaration` still iterates the full `PEER_REDEFINABLE_FIELDS` — which is what you want, since the durable descriptor genuinely *is* authoritative for `indexNulls` (that was the point of `7d2f7d81`). Narrowing only the warn means the fix can't regress the merge semantics.

New test `does not name a locally derived field the peer never sent in the discard warn` covers it.

Generated by Barber AI

resources/databases.ts
cb1kenobi: Re-verified at `97251c59` — the fix still holds. Re-ran the probe rather than assuming it, since this PR touched the same function again.

Probe: an existing index with **10 real records inserted first** (important — without existing data the `hasExistingData` branch is never entered and the probe comes back inconclusive), then the descriptor rewritten with a dead `indexingPID: 999999` *and* a stale `restartNumber: 0` against `currentRestartGeneration: 1`, then re-declared with `origin: 'cluster'`.

```
reindex test.CrashProbe2258c.tag: reason=crash-recovery(pid=999999),restart-number
```

Both triggers named, exactly as expected. `Recovered.indices.tag.isIndexing === false`, the backfill indexed every row (5 odd / 5 even out of 10), and the descriptor came back with `indexingPID` and `restartNumber` cleared — a clean completion, not a parked build.

Also re-ran the four regression guards by mutation at this head; all four still die with their own assertion messages:

| Guard mutated off | Result |
| --- | --- |
| additive merge | `attribute 'name' was removed by a cluster-origin definition: id` |
| live `schemaDefined` gate | `a cluster-origin definition flipped schemaDefined` |
| descriptor-skip block | `a cluster-origin call rewrote a newer durable descriptor from its stale snapshot` |
| removal reconcile | `catalog descriptor for 'tag' survived removal` |

Generated by Barber AI

resources/databases.ts
cb1kenobi: Leaving this one open and unchanged — it's a naming/API judgement call for a human, not something I want to push on, and deferring it was the right response.

Adding one piece of context that may make the call easier whenever it gets made, since I confirmed it at this head anyway: `origin: 'cluster'` still has exactly **one** in-repo caller, `resources/Table.ts:953`, and [harper-pro#750](https://github.com/HarperFast/harper-pro/pull/750) is **still open** (head `b2231e1b`, unmerged). That single caller builds its list as `this.attributes.slice(0)` and only ever *pushes* attributes it doesn't already find, so for every pre-existing attribute `existing === attribute` by identity in the merge loop — the `JSON.stringify` comparison can never differ and the discard warn is unreachable from it.

Practical upshot: until #750 merges, the collision surface for an app literally named `cluster` is essentially nil, because nothing in-repo drives this path with peer-authored attribute objects yet. That also means the window to change the sentinel's shape is open at basically zero cost right now, and closes once #750 lands and real handshake traffic starts flowing through it. Purely informational — no change requested here.

Generated by Barber AI

cb1kenobi:
Reviewed `4e332df6` (was `1898b3f9`) — no issues found. This PR looks good, nice job!

**This push is a pure rebase — no PR content changed.** `git range-diff` against each head's own merge-base (`f8a5aa90a` → `df5355aaa`) shows the same 12 commits in 1:1 correspondence. Six are marked changed, but every one of those is context absorption from main, not a logic edit: the cumulative PR patch is 609 lines at *both* heads and byte-identical once whitespace is normalized. Commit 1's apparent change is re-indentation, because main's #2252 wrapped this block in `exclusiveLock()`.

Two things I checked rather than assumed, since a rebase is exactly where an interaction regression hides:

**1. The new `exclusiveLock()` scope from #2252 (`databases.ts:1830`).** The additive-merge block now runs inside it. That is correct, and arguably better than the pre-rebase arrangement: the lock is taken before the merge reads `Table.attributes`, so that read and the later `Table.attributes.splice(...)` are consistent under one acquisition, and a bounded-acquire throw leaves nothing half-applied — which is what the comment above it claims. The merge itself mutates only a local `merged` copy, so nothing new is persisted under the lock.

**2. The red `Integration Tests 2/6 (Windows, Node.js v24)`.** Two assertions fail there, and one of them — `dynamic.Loose must be in describe_all phase 2` — is schema-shaped, so I did not want to wave it off as flake. It is not reachable from this PR:

- Every behavioral change here is gated on `origin === 'cluster'`: the additive merge, `reconcileRemovals`, the `schemaDefinedMismatch` guard, and the descriptor branch.
- The one condition that is not a pure gate, `if (schemaDefinedExplicit && origin !== 'cluster') Table.schemaDefined = schemaDefined;`, is by construction a no-op for non-cluster origins.
- `applyDurableDeclaration` — the only new helper that unconditionally deletes fields — has exactly two call sites (`databases.ts:2101` and `:2113`), both inside `if (origin === 'cluster' && attributeDescriptor)`.

`describe-metadata-upgrade.test.ts` phase 2 is a single-node reboot on a pinned data dir with no replication, so `origin` is never `'cluster'` on its path. The same shard also failed the unrelated `Probe /SeoPageCache/ … ECONNREFUSED after restart_service` with a 120s timeout, which is the more likely root cause of both — a Windows restart that did not come back leaves the following reboot-based test without a server.

To be precise about the strength of that: the reachability argument is **verified against the source** (gates enumerated, call sites traced). The "it is a Windows environment flake" half is **inferred** from the co-occurring restart failure — I did not reproduce it, and I could not use harper#2328 as a control because its checks were re-run clean in the meantime.

The one still-open thread from earlier rounds is the Low on `'cluster'` as a magic value in `origin` — deliberately left for a human call, unchanged by this push, and not a blocker.

—
_Generated by Barber AI_

resources/databases.ts
```diff
@@ -354,6 +354,31 @@ _assignPackageExport('databases', databases);
 _assignPackageExport('tables', tables);
 
 const NEXT_TABLE_ID = Symbol.for('next-table-id');
+// The `commonChanged` comparison below, plus the two index flags — an index built without null entries
+// must not be re-registered as null-capable.
```
cb1kenobi: As a human, I cannot comprehend this comment.

```suggestion
// Restore every field used by `commonChanged`, plus `indexed` and `indexNulls`,
// from the durable descriptor. In particular, preserve `indexNulls: false` so
// an index that excludes nulls is not reopened as though it contains them.
```

cb1kenobi (approved):

PR #2265, opened 2026-08-21
Capture only whole blobs in a backup, substituting a marker for one still being written

### LINKED ISSUE #2262: Backup hard-links blobs that are still being written, so a snapshot's bytes keep changing after the snapshot
## Summary

`dataLayer/blobBackup.ts` snapshots a database's blobs by hard-linking them, and its header comment
justifies that with a write-once claim. The walk has **no check for whether a blob is currently being
written**, so a snapshot can hard-link an inode that is still growing. The bytes in the snapshot then
keep changing after the snapshot was taken.

## The stated invariant

`dataLayer/blobBackup.ts:29-33`:

> Hard-linking is safe against later mutation because Harper blobs are content-addressed and
> write-once … no in-place overwrite can retroactively alter a snapshot.

and, acknowledging the case but waving it through, `blobBackup.ts:26`:

> a blob being written mid-walk is captured as-is (a hard link shares the inode, so it reflects the
> writer's final bytes …)

"Reflects the writer's final bytes" is the problem, not the reassurance: the snapshot is supposed to
be a point-in-time capture, and this one silently keeps mutating until the writer finishes — or ends
truncated if the writer aborts.

## Mechanism

Blobs are written in place at their final path (`resources/blob.ts:1657`,
`createWriteStream(filePath, { flags: 'w' })`) and the real header is only written at stream end, so
a blob file exists and grows for the whole write. `snapshotBlobs()` → `copyTree()` → `linkOrCopy()`
(`blobBackup.ts:82-105`) links whatever files it finds, unconditionally. Nothing coordinates them.

Demonstrated: a test asserting nothing incomplete is visible at the final path mid-write fails on
`main` with a 20008-byte partial (20000 streamed bytes plus the 8-byte placeholder header).

## What is NOT the fix

**Do not stage writes to a temporary path and rename into place.** That looks like the obvious fix
and it is wrong twice over:

1. **In-flight visibility at the final path is a deliberate feature, not an accident.** The
   progressive-read machinery at `resources/blob.ts:718-808` — `checkIfIsBeingWritten()`, an fs
   watcher, `incompleteDeadline`, `resumeIfWriterFinished`, see #1423/#1444/#1454 — exists precisely
   so a reader can attach to a blob while it is still being written. Hiding the file until completion
   sends those readers into the ENOENT retry loop (`blob.ts:557-568`) and a 503 after
   `blobReadTimeout` (20s), which a large replication receive routinely exceeds.
2. **A suffixed temporary breaks fileId recovery.** `getNextFileId` rebuilds its high-water mark by
   parsing directory entries (`blob.ts:1986-1998`): for an entry longer than three characters it does
   `parseInt(entry.slice(-3), 16)`, which on `5001.publishing` is `parseInt('ing', 16)` → `NaN`, so
   the entry is skipped entirely. Today the partial sits at `5001` and *is* counted. Staging would
   let a crash mid-write release that id for reuse, so two records end up sharing one blob file.

## Suggested fix — in the consumer, not the writer

Backup is the component holding the wrong assumption, so that is where it should be corrected:
`copyTree` should skip any blob that is not a completed, published file.

- `blobHeaderIndicatesIncomplete(header, fileSize)` (`resources/blob.ts:1830`) is already exported and
  is exactly this predicate.
- Alternatively, or additionally, the `tryLock(<fileId>:blob)` idiom that orphan GC already uses to
  tell an in-flight temp from crash debris (`resources/blob.ts:2679-2686`).
- The walk should also skip the `.repair` suffix, which it currently links into snapshots.

Trade-off worth stating: a blob mid-write at snapshot time is then absent from the backup rather than
present-and-partial. Because replication apply can commit a record without awaiting blob durability,
a restore may hold a record whose blob is missing — but that is a *repairable* state (`repairBlobFile`
and the blob-gap machinery exist for exactly it), whereas a silently truncated blob is not. Absent
beats corrupt.

## Scope note

Backup is the only consumer of blob hard-linking today. Any future feature that shares blob inodes
inherits the same requirement — HarperFast/harper#642 (branched databases) would be the second, and
its blob-tree clone needs the same in-flight check.

## Affected versions

`dataLayer/blobBackup.ts` landed 2026-07-29 (`d1572bd9e`); shipped in the 5.2.x line, current `5.2.4`.

## Filing history

Two earlier framings of this issue were wrong and are recorded here so the trail does not mislead:

1. Originally filed as "a re-stream overwrites a PENDING stub in place". The comment that claim rested
   on (`blob.ts:74-78`) predates `642d77dbd` "Make blob gap repair atomic" (2026-08-18) by two months
   and describes removed behavior; `generateFilePath` also allocates a fresh id per write, so no path
   rewrites a published blob.
2. Then reframed as "blob files have no publication boundary", with a staged-write fix. That fix was
   implemented, reviewed, and abandoned for the two reasons under **What is NOT the fix** above.

🤖 Filed by Claude on behalf of Kris

kriszyp:
Blob backup snapshots and streamed archives now use [one capture classifier](https://github.com/HarperFast/harper/pull/2265/changes?w=1#diff-5c96de79cba36ba9ddb5124ce7b7e96a398a3c50fc6b7bec11dc5d690f61767aR1842): complete blobs and stable abort markers are captured, repair temporaries are omitted, and incomplete or vanished entries already returned by the walk become PENDING or ERROR markers instead of torn files. Follow-up review fixes narrow the guarantee to the enumeration boundary, document marker entries in the generated README, keep skip entries out of the warning denominator, remove unused exports, and add direct coverage for ENOENT and the fail-closed error set. Earlier review rounds were comment-only: [the deflate check](https://github.com/HarperFast/harper/pull/2265/changes?w=1#diff-5c96de79cba36ba9ddb5124ce7b7e96a398a3c50fc6b7bec11dc5d690f61767aR1866) now names the live-write case it is the sole detector of, the two branches no test reaches say so where they are, and the classifier's contract no longer claims an invariant the write path violates (see the note opening the reviewer section). Fixes #2262. The Windows Node 24 integration fixture now [resolves its pinned `%TEMP%` root to the native long path](https://github.com/HarperFast/harper/pull/2265/changes?w=1#diff-2cd864eab8d523890e18b6a6dcbaf0d7e732677ad340e8feb025f8673f694538R44) before Harper starts, avoiding libuv’s recursive-watcher abort on 8.3 paths.

The remaining performance concern is explicit: `get_backup` currently opens and stats a complete blob once to classify it and again to stream it. Combining those handles is a reasonable future optimization, but it is not mixed into this correctness change.

## For the human reviewer

**Start here — a hard-link truncation race this PR documents but does not close.** Two independent review models found it separately this round, and I confirmed the mechanism by trace. A known-size write that has landed every byte but has not settled classifies as `capture`, because `blobHeaderIndicatesIncomplete` sees `fileSize === HEADER_SIZE + storedSize`; the snapshot then hard-links that live inode. If the write then aborts, the replication PENDING stamp at [`resources/blob.ts:1572`](https://github.com/HarperFast/harper/pull/2265/changes?w=1#diff-5c96de79cba36ba9ddb5124ce7b7e96a398a3c50fc6b7bec11dc5d690f61767aR1572) does `writeFile(filePath, …)` — an `O_TRUNC` open, which rewrites the **inode**, so the already-captured snapshot entry is truncated to an 8-byte stub after capture. Same-filesystem only; the copy path is immune.

I [narrowed the classifier's contract](https://github.com/HarperFast/harper/pull/2265/changes?w=1#diff-5c96de79cba36ba9ddb5124ce7b7e96a398a3c50fc6b7bec11dc5d690f61767aR1837) to stop claiming the invariant that is false, and left the behavior alone deliberately. Re-stat'ing after linking narrows the window but cannot close it — the abort can always land after the check — and the only complete fix is consulting the blob write lock, which is keyed by file id and is the store-access alternative decision 5 rejects. So this wants a decision, not a patch, and it may deserve its own issue. Damage is bounded: the snapshot gets a retryable PENDING stub rather than torn bytes, but the substitution counters still report it as captured.

1. **Point-in-time marker versus eventual hard-link bytes.** [Snapshot capture](https://github.com/HarperFast/harper/pull/2265/changes?w=1#diff-ef6081f3b8edd93e7e5b3c8bb011fc94a74142673b26b7b93d2f31718d71fc98R94) deliberately substitutes PENDING when classification observes an in-flight blob, even on the same-filesystem path where a hard link could later converge to the writer's final bytes. This favors immutable point-in-time semantics and consistent copy/link behavior over preserving a write that finishes after capture; changing the policy later will not repair snapshots already written.
2. **Vanished entries are terminal.** Reclamation can unlink a superseded blob after the engine checkpoint, so the restored checkpoint may still reference it. The classifier emits ERROR rather than PENDING to avoid making every local read wait for repair; the cost is that a peer holding the old bytes cannot heal that version because ERROR can replicate as-is while PENDING leaves a repairable gap. Changing this arm to PENDING is small, but it changes read and replication semantics. 

   *Merge-order note.* [harper#1835 — durable blob-unlink queue](https://github.com/HarperFast/harper/pull/1835) is open, and the fix currently suggested for its unresolved High moves the durable unlink row earlier, into `deleteBlob()` itself. If it lands in that shape, the reclamation window this arm reasons about stops being in-process: a queued unlink survives restart and can be replayed against a blob a later backup walk is still classifying, so a restart no longer rescues those bytes and terminal costs more than it does today. Neither PR needs a change for the other to be correct — whoever merges second should re-read this decision against the other PR's final shape.
3. **A backup with substituted markers still verifies as complete.** The [creation warning](https://github.com/HarperFast/harper/pull/2265/changes?w=1#diff-ef6081f3b8edd93e7e5b3c8bb011fc94a74142673b26b7b93d2f31718d71fc98R189) and generated README expose substitutions, but the manifest and `verify_backup` do not. Adding a degraded count would improve restore-time visibility, but it is a backup-format/API change and would not annotate existing snapshots.
4. **Storage and descriptor failures abort the backup.** `EIO`, `EMFILE`, `ENFILE`, `ENOSPC`, and `EROFS` fail closed rather than capturing a file unverified; non-systemic classifier failures preserve the previous capture behavior because a hard link can succeed without reading the source. Narrowing this set would favor availability over proving the backup was readable at capture time.
5. **Classification is local to both capture walks.** The root-cause alternative is coordinating the backup with blob-write/reclamation locks or enumerating exactly the file ids present in the checkpoint. That needs store access and a lossless path-to-id mapping the walks do not have, so [the archive path](https://github.com/HarperFast/harper/pull/2265/changes?w=1#diff-bbfcea2dd873a1ccd771306cbcf3cc20d95f4c7a01bb171b92f13683195fcb9eR854) shares the classifier instead.
6. **Size is sampled before the header, and I left it that way.** [Classification stats the file first and reads the header second](https://github.com/HarperFast/harper/pull/2265/changes?w=1#diff-5c96de79cba36ba9ddb5124ce7b7e96a398a3c50fc6b7bec11dc5d690f61767aR1857), so an unknown-size write that completes between the two samples is compared against a stale size and substituted. This round's review proposed re-stat'ing and trusting the file when the size is unchanged. Declining that is decision 1, not an oversight: the only entries affected are ones genuinely mid-write when classification began, and capturing a write that lands after the walk observes it is exactly what the point-in-time policy gives up. Reversing it is a policy change and should be decided as one.

7. **Three sites state in the source that a branch is untested.** The systemic-errno rethrow on [the snapshot path](https://github.com/HarperFast/harper/pull/2265/changes?w=1#diff-ef6081f3b8edd93e7e5b3c8bb011fc94a74142673b26b7b93d2f31718d71fc98R99) and [the archive path](https://github.com/HarperFast/harper/pull/2265/changes?w=1#diff-bbfcea2dd873a1ccd771306cbcf3cc20d95f4c7a01bb171b92f13683195fcb9eR859), and [`inflatesToExactly`'s reject arm](https://github.com/HarperFast/harper/pull/2265/changes?w=1#diff-5c96de79cba36ba9ddb5124ce7b7e96a398a3c50fc6b7bec11dc5d690f61767aR2857), each say so where they are, because reaching either needs a real host-level fault and AGENTS.md forbids adding a stubbing seam to fake one. A mutation run at this head confirms both branches genuinely survive the suite. The competing view is that coverage status is review metadata that does not belong in the source; I kept the notes because a fail-closed branch nothing reaches otherwise reads as verified. They are one-line deletions if you disagree.

## Verification

- `npm run build` — passed.
- `npm run test:unit:backup` — 84 passing.
- `npx mocha unitTests/resources/blob.test.js` — 106 passing, including the live-write snapshot lifecycle and marker read behavior.
- [`unitTests/dataLayer/blobBackup.test.js`](https://github.com/HarperFast/harper/pull/2265/changes?w=1#diff-b0412778c1b85da425ad68cce79648858c1a8d7bcc2bb972c74e35639d982591R214) now directly proves missing-path classification, the fail-closed error set, README marker documentation, and repair-temp accounting.
- `npm run test:unit:resources` — 1674 passing, 16 pending, 3 unrelated existing failures in random-access defaults and replay structures; all edited blob tests passed.
- `npm run test:unit:main` could not enter the suite because its initialization opened an unrelated configured `/home/kzyp/dev/tmp/hdb50node24final/database/system.mdb` and failed while decoding that state; retrying with a disposable `HDB_ROOT` produced the same pre-test failure.
- `npm run lint:required`, Prettier, and `git diff --check` — clean.
- Latest (comment-only) rounds re-verified: `npm run build`, `npm run test:unit:backup` (84 passing), `npx mocha unitTests/resources/blob.test.js` (106 passing), lint, Prettier, and `git diff --check` all clean. The suite results above are unchanged by them — no executable line moved.
- `npm run test:integration -- integrationTests/apiTests/describe-metadata-upgrade.test.ts` — 2 passing after canonicalizing the fixture root.
- `npm run test:integration:all` — 1795 passing; the only parent failure was the pre-existing local Node 26 JSON import-attribute error in the real-Ollama suite, which cancelled 6 children.

Complexity: complicated

<sub>Review-Coverage: authored=codex; ran=claude,gemini; declined=cursor-grok,cursor-composer,domain; rounds=4 @ f44e5db03403</sub>

<sub>Human-Review-Need: 4 @ f44e5db03403</sub>

### DIFF
--- a/dataLayer/blobBackup.ts
+++ b/dataLayer/blobBackup.ts
@@ -96,6 +96,8 @@ async function captureBlobFile(srcPath: string, destPath: string): Promise<BlobC
 	try {
 		disposition = await classifyBlobFileForCapture(srcPath);
 	} catch (error) {
+		// This rethrow is unverified: only the membership of SYSTEMIC_IO_ERRORS is tested, because
+		// reaching it needs a real host-level fault no test here can produce.
 		if (isSystemicIoError(error)) throw error;
 		// Fall back to what this walk did before it classified anything. `link()` needs no read permission
 		// on the source, so a blob this failed to *read* may still be perfectly capturable; substituting
--- a/dataLayer/blobBackup.ts
+++ b/dataLayer/blobBackup.ts
@@ -5,6 +5,12 @@ import { copyFile, link, mkdir, readdir, rename, rm, unlink, writeFile } from 'n
 import { dirname, join, relative } from 'node:path';
 import { ClientError } from '../utility/errors/hdbError.ts';
 import logger from '../utility/logging/harper_logger.ts';
+import {
+	type BlobCaptureDisposition,
+	classifyBlobFileForCapture,
+	createCaptureMarker,
+	isSystemicIoError,
+} from '../resources/blob.ts';
 
 /**
  * Managed-backup snapshotting of a database's file-backed blobs.
@@ -21,17 +27,16 @@ import logger from '../utility/logging/harper_logger.ts';
  * possible (cheap, no extra space on the same filesystem) and copied otherwise (never symlinked, so
  * a snapshot is a standalone set of files that survives independent of the live blob).
  *
- * Consistency is best-effort and point-in-time-ish, matching how the engine treats the transaction
- * log: the walk captures whatever files exist at snapshot time. A blob deleted mid-walk is skipped;
- * a blob being written mid-walk is captured as-is (a hard link shares the inode, so it reflects the
- * writer's final bytes; a cross-filesystem copy captures the bytes present at copy time). Harper
- * does not freeze blob writes for the duration of a backup.
+ * Harper does not freeze blob writes for a backup, and a blob is written in place at its final path,
+ * so the walk can meet one that is still growing. Hard-linking that inode would put bytes in the
+ * snapshot that keep changing after it was taken. Every entry is therefore classified
+ * (`captureBlobFile`): a complete blob is linked, and one that is not is replaced by a PENDING marker
+ * rather than dropped, which keeps its file id reserved — `getNextFileId` recovers the id counter by
+ * scanning the directory, so an absent file would let a restored record's id be reissued.
  *
- * Hard-linking is safe against later mutation because Harper blobs are content-addressed and
- * write-once: each write allocates a fresh monotonic file id (a new path), and an update or delete
- * unlinks the old path rather than rewriting it in place — so a snapshot's hard link keeps the exact
- * bytes alive even after the live blob is deleted, and no in-place overwrite can retroactively alter
- * a snapshot.
+ * Past that gate, hard-linking is safe against later mutation: each write takes a fresh monotonic
+ * file id, and an update, delete, or in-place repair replaces or unlinks the path rather than
+ * rewriting the inode.
  */
 
 /** Directory holding all blob snapshots for a backup repository. */
@@ -49,14 +54,13 @@ export function blobSnapshotDir(backupDir: string, backupId: number): string {
  * the filesystem does not support additional hard links). Never creates a symlink. A source that
  * vanished mid-walk (a concurrent blob delete) is skipped rather than failing the whole snapshot.
  */
-async function linkOrCopy(src: string, dest: string): Promise<void> {
+async function linkOrCopy(src: string, dest: string): Promise<boolean> {
 	await mkdir(dirname(dest), { recursive: true });
 	try {
 		await link(src, dest);
 	} catch (error: any) {
 		if (error.code === 'ENOENT') {
-			// src disappeared (concurrent delete) — nothing to snapshot
-			if (!existsSync(src)) return;
+			if (!existsSync(src)) return false; // vanished since it was classified
 			throw error;
 		}
 		if (
@@ -66,24 +70,72 @@ async function linkOrCopy(src: string, dest: string): Promise<void> {
 			error.code === 'ENOTSUP' ||
 			error.code === 'EOPNOTSUPP'
 		) {
-			await copyFile(src, dest);
-			return;
+			try {
+				await copyFile(src, dest);
+			} catch (copyError: any) {
+				if (copyError.code === 'ENOENT') return false;
+				throw copyError;
+			}
+			return true;
 		}
 		if (error.code === 'EEXIST') {
 			await unlink(dest);
-			await linkOrCopy(src, dest);
-			return;
+			return linkOrCopy(src, dest);
 		}
 		throw error;
 	}
+	return true;
+}
+
+/**
+ * Put one blob file into the snapshot, returning true when a PENDING marker stood in for it.
+ */
+async function captureBlobFile(srcPath: string, destPath: string): Promise<boolean> {
+	let disposition: BlobCaptureDisposition;
+	try {
+		disposition = await classifyBlobFileForCapture(srcPath);
+	} catch (error) {
+		if (isSystemicIoError(error)) throw error;
+		// Fall back to what this walk did before it classified anything. `link()` needs no read permission
+		// on the source, so a blob this failed to *read* may still be perfectly capturable; substituting
+		// here would turn an unreadable-but-valid root into a backup of nothing but stubs.
+		logger.warn(`Could not verify blob ${srcPath} for snapshot; capturing it unverified`, error);
+		disposition = 'capture';
+	}
+	if (disposition === 'skip') return false;
+	// A capture can still lose the race with a delete between classify and link; those bytes are gone,
+	// so it falls through to a terminal marker rather than being dropped and freeing the id.
+	if (disposition === 'capture') {
+		if (await linkOrCopy(srcPath, destPath)) return false;
+		disposition = 'gone';
+	}
+	await mkdir(dirname(destPath), { recursive: true });
+	await writeFile(
+		destPath,
+		createCaptureMarker(
+			disposition,
+			disposition === 'gone'
+				? 'blob was deleted while this backup was being taken'
+				: 'blob was not yet complete when this backup was taken'
+		)
+	);
+	return true;
 }
 
 /**
  * Recursively copy every file under `srcRoot` into `destRoot` (hard-link-else-copy), preserving the
  * relative directory structure. Missing `srcRoot` is a no-op (a database with no blobs yet).
+ *
+ * `classify` belongs to the snapshot direction only; restore must replace exactly what the snapshot
+ * holds.
  */
-async function copyTree(srcRoot: string, destRoot: string): Promise<void> {
-	if (!existsSync(srcRoot)) return;
+async function copyTree(
+	srcRoot: string,
+	destRoot: string,
+	classify = false
+): Promise<{ substituted: number; captured: number }> {
+	const counts = { substituted: 0, captured: 0 };
+	if (!existsSync(srcRoot)) return counts;
 	const stack: string[] = [srcRoot];
 	while (stack.length > 0) {
 		const dir = stack.pop() as string;
@@ -99,11 +151,18 @@ async function copyTree(srcRoot: string, destRoot: string): Promise<void> {
 			if (entry.isDirectory()) {
 				stack.push(srcPath);
 			} else if (entry.isFile()) {
-				await linkOrCopy(srcPath, join(destRoot, relative(srcRoot, srcPath)));
+				const destPath = join(destRoot, relative(srcRoot, srcPath));
+				if (classify) {
+					if (await captureBlobFile(srcPath, destPath)) counts.substituted++;
+					else counts.captured++;
+				} else {
+					await linkOrCopy(srcPath, destPath);
+				}
 			}
 			// symlinks/other node types in a blob root are not expected and are intentionally skipped
 		}
 	}
+	return counts;
 }
 
 /**
@@ -118,8 +177,19 @@ export async function snapshotBlobs(backupDir: string, backupId: number, blobRoo
 	await rm(tempDir, { recursive: true, force: true });
 	await mkdir(tempDir, { recursive: true });
 	try {
+		let substituted = 0;
+		let captured = 0;
 		for (let index = 0; index < blobRoots.length; index++) {
-			await copyTree(blobRoots[index], join(tempDir, String(index)));
+			const counts = await copyTree(blobRoots[index], join(tempDir, String(index)), true);
+			substituted += counts.substituted;
+			captured += counts.captured;
+		}
+		if (substituted > 0) {
+			logger.warn(
+				`Blob snapshot for backup ${backupId} captured ${substituted} of ${substituted + captured} blob ` +
+					`file(s) as PENDING markers because they were not capturable whole. Records referencing them will ` +
+					`need blob repair after a restore.`
+			);
 		}
 		await rm(finalDir, { recursive: true, force: true });
 		await rename(tempDir, finalDir);
--- a/dataLayer/blobBackup.ts
+++ b/dataLayer/blobBackup.ts
@@ -29,10 +29,11 @@ import {
  *
  * Harper does not freeze blob writes for a backup, and a blob is written in place at its final path,
  * so the walk can meet one that is still growing. Hard-linking that inode would put bytes in the
- * snapshot that keep changing after it was taken. Every entry is therefore classified
- * (`captureBlobFile`): a complete blob is linked, and one that is not is replaced by a PENDING marker
- * rather than dropped, which keeps its file id reserved — `getNextFileId` recovers the id counter by
- * scanning the directory, so an absent file would let a restored record's id be reissued.
+ * snapshot that keep changing after it was taken. Every entry returned by the walk is therefore
+ * classified (`captureBlobFile`): a complete blob is linked, and one that becomes unavailable or is
+ * not yet complete is replaced by a marker rather than dropped. This closes the classify-to-capture
+ * race and keeps that file id reserved; a file reclaimed before its parent directory is read remains
+ * outside the snapshot.
  *
  * Past that gate, hard-linking is safe against later mutation: each write takes a fresh monotonic
  * file id, and an update, delete, or in-place repair replaces or unlinks the path rather than
@@ -51,8 +52,8 @@ export function blobSnapshotDir(backupDir: string, backupId: number): string {
 
 /**
  * Hard-link `src` to `dest`, falling back to a copy when the two are on different filesystems (or
- * the filesystem does not support additional hard links). Never creates a symlink. A source that
- * vanished mid-walk (a concurrent blob delete) is skipped rather than failing the whole snapshot.
+ * the filesystem does not support additional hard links). Never creates a symlink. Returns false
+ * when the source vanishes so the caller can substitute a marker for it.
  */
 async function linkOrCopy(src: string, dest: string): Promise<boolean> {
 	await mkdir(dirname(dest), { recursive: true });
@@ -88,9 +89,9 @@ async function linkOrCopy(src: string, dest: string): Promise<boolean> {
 }
 
 /**
- * Put one blob file into the snapshot, returning true when a PENDING marker stood in for it.
+ * Put one blob file into the snapshot, returning how the entry was captured.
  */
-async function captureBlobFile(srcPath: string, destPath: string): Promise<boolean> {
+async function captureBlobFile(srcPath: string, destPath: string): Promise<BlobCaptureDisposition> {
 	let disposition: BlobCaptureDisposition;
 	try {
 		disposition = await classifyBlobFileForCapture(srcPath);
@@ -102,11 +103,9 @@ async function captureBlobFile(srcPath: string, destPath: string): Promise<boole
 		logger.warn(`Could not verify blob ${srcPath} for snapshot; capturing it unverified`, error);
 		disposition = 'capture';
 	}
-	if (disposition === 'skip') return false;
-	// A capture can still lose the race with a delete between classify and link; those bytes are gone,
-	// so it falls through to a terminal marker rather than being dropped and freeing the id.
+	if (disposition === 'skip') return disposition;
 	if (disposition === 'capture') {
-		if (await linkOrCopy(srcPath, destPath)) return false;
+		if (await linkOrCopy(srcPath, destPath)) return disposition;
 		disposition = 'gone';
 	}
 	await mkdir(dirname(destPath), { recursive: true });
@@ -119,7 +118,7 @@ async function captureBlobFile(srcPath: string, destPath: string): Promise<boole
 				: 'blob was not yet complete when this backup was taken'
 		)
 	);
-	return true;
+	return disposition;
 }
 
 /**
@@ -153,8 +152,9 @@ async function copyTree(
 			} else if (entry.isFile()) {
 				const destPath = join(destRoot, relative(srcRoot, srcPath));
 				if (classify) {
-					if (await captureBlobFile(srcPath, destPath)) counts.substituted++;
-					else counts.captured++;
+					const disposition = await captureBlobFile(srcPath, destPath);
+					if (disposition === 'pending' || disposition === 'gone') counts.substituted++;
+					else if (disposition === 'capture') counts.captured++;
 				} else {
 					await linkOrCopy(srcPath, destPath);
 				}
@@ -186,9 +186,8 @@ export async function snapshotBlobs(backupDir: string, backupId: number, blobRoo
 		}
 		if (substituted > 0) {
 			logger.warn(
-				`Blob snapshot for backup ${backupId} captured ${substituted} of ${substituted + captured} blob ` +
-					`file(s) as PENDING markers because they were not capturable whole. Records referencing them will ` +
-					`need blob repair after a restore.`
+				`Blob snapshot for backup ${backupId} substituted ${substituted} of ${substituted + captured} blob ` +
+					`file(s) with PENDING or ERROR markers because they were not capturable whole.`
 			);
 		}
 		await rm(finalDir, { recursive: true, force: true });
@@ -245,8 +244,10 @@ ${rootMapping}
   4096 entries per directory). E.g. a blob with id \`0x12345678\` lives at \`12/345/678\`; a short id
   like \`0xc1a\` lives at \`0/0/c1a\`.
 
-Files are hard links to the live blobs when the backup is on the same filesystem, and copies
-otherwise.
+Complete blobs are hard links to the live blobs when the backup is on the same filesystem, and
+copies otherwise. A blob that was not capturable whole is stored as a marker instead: header type
+\`0xfe\` is retryable (PENDING), while \`0xff\` is terminal (ERROR). The marker preserves the file id
+but does not contain the original blob bytes.
 `;
 }
 
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -2853,7 +2853,10 @@ function inflatesToExactly(filePath: string, size: number): Promise<boolean> {
 		// A zlib error is the answer (body truncated or corrupt); an I/O error is a failure to answer and
 		// must propagate, or a systemic fault would classify a corpus of complete blobs as incomplete.
 		const fail = (error: NodeJS.ErrnoException) => {
+			// Both: pipe() does not tear down the destination when the source errors, so the inflate's
+			// native zlib handle would sit allocated until GC — once per blob, on every backup walk.
 			source.destroy();
+			inflate.destroy();
 			if (error?.code?.startsWith('Z_')) resolve(false);
 			else reject(error);
 		};
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -1833,8 +1833,11 @@ export function isSystemicIoError(error: unknown): boolean {
 
 /**
  * How a consumer that captures a blob root (backup snapshot, backup archive) should treat one file.
- * `skip` is not a blob to capture at all; `capture` is safe to take as-is — its bytes will not change,
- * because no path rewrites a published blob file in place; `pending` and `gone` require markers.
+ * `skip` is not a blob to capture at all; `pending` and `gone` require markers; `capture` is taken as-is.
+ * `capture` means settled as far as the path can show, which is short of a guarantee: a known-size write
+ * that has landed every byte is indistinguishable here from a finished one, and if it then aborts, the
+ * PENDING stamp rewrites that inode in place, truncating any same-filesystem hard link taken from it.
+ * Telling the two apart needs the blob write lock, which is keyed by file id and unreachable from a walk.
  */
 export async function classifyBlobFileForCapture(filePath: string): Promise<BlobCaptureDisposition> {
 	if (filePath.endsWith(BLOB_REPAIR_SUFFIX)) return 'skip';
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -76,7 +76,7 @@ type StorageInfo = {
 };
 const FILE_STORAGE_THRESHOLD = 8192; // if the file is below this size, we will store it in memory, or within the record itself, otherwise we will store it in a file
 // We want to keep the file path private (but accessible to the extension)
-export const HEADER_SIZE = 8;
+const HEADER_SIZE = 8;
 const UNCOMPRESSED_TYPE = 0;
 const DEFLATE_TYPE = 1;
 const ERROR_TYPE = 0xff;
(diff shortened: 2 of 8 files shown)


resources/blob.ts
```diff
@@ -2738,32 +2831,42 @@ async function isBlobFileComplete(storageInfo: StorageInfo): Promise<boolean> {
...
+		let inflatedLength = 0;
+		const source = createReadStream(filePath, { start: HEADER_SIZE });
+		const inflate = createInflate();
+		// A zlib error is the answer (body truncated or corrupt); an I/O error is a failure to answer and
+		// must propagate, or a systemic fault would classify a corpus of complete blobs as incomplete.
+		const fail = (error: NodeJS.ErrnoException) => {
+			source.destroy();
+			if (error?.code?.startsWith('Z_')) resolve(false);
+			else reject(error);
+		};
```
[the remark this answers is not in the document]
kriszyp: Good catch, and the reasoning is exactly right — applied in 4a9931bc1.

You're also right that this predates the PR: `inflatesToExactly` is extracted from the existing `isBlobFileComplete`, which had the same `fail` handler. What changed is the blast radius, as you note — it used to run only from the periodic repair sweep, and now runs once per compressed blob on every backup and archive walk, so a leaked zlib handle per I/O error goes from rare to per-file.

Both `source` and `inflate` are destroyed now.

🤖 Reply by Claude (Opus 5) on behalf of Kris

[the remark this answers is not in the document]
kriszyp: Applied in 4a9931bc1 — thanks. Both `source` and `inflate` are destroyed in `fail` now.

claude[bot] converged on the same finding independently, with the mechanism spelled out: `pipe()` doesn't tear down the destination when the source errors, so the inflate's native zlib handle sits allocated until GC. Worth noting the exposure grew with this PR even though the pattern is older — this path used to run only in the periodic repair sweep and now runs per compressed blob on every backup walk.

🤖 Reply by Claude (Opus 5) on behalf of Kris

dataLayer/blobBackup.ts
```diff
@@ -21,17 +27,16 @@ import logger from '../utility/logging/harper_logger.ts';
...
- * Consistency is best-effort and point-in-time-ish, matching how the engine treats the transaction
- * log: the walk captures whatever files exist at snapshot time. A blob deleted mid-walk is skipped;
- * a blob being written mid-walk is captured as-is (a hard link shares the inode, so it reflects the
- * writer's final bytes; a cross-filesystem copy captures the bytes present at copy time). Harper
- * does not freeze blob writes for the duration of a backup.
+ * Harper does not freeze blob writes for a backup, and a blob is written in place at its final path,
+ * so the walk can meet one that is still growing. Hard-linking that inode would put bytes in the
+ * snapshot that keep changing after it was taken. Every entry is therefore classified
+ * (`captureBlobFile`): a complete blob is linked, and one that is not is replaced by a PENDING marker
+ * rather than dropped, which keeps its file id reserved — `getNextFileId` recovers the id counter by
```
cb1kenobi: **Medium: the "never dropped, id stays reserved" invariant only holds after `readdir`**

`copyTree` can only substitute a marker for an entry `readdir` actually returned. A blob unlinked *before* its parent directory is read is still absent from the snapshot with its id unreserved -- the exact outcome this paragraph says is prevented.

That window is not rare. `deleteBlob` queues every superseded blob and `DEFAULT_RECLAMATION_DELAY` is 2000 ms (`resources/blob.ts:929`), so on a write-active database `runReclamation` is unlinking files continuously while a walk that takes far longer than 2 s is in progress. For `get_backup` the gap is much larger still: `streamBackupWithBlobs` streams the whole native DB tar to the client and awaits it before `appendBlobEntries` starts, so the checkpoint-to-blob-walk distance is the full download.

Not asking for the coordination fix (decision 4) here. Asking that the comment claim what is enforced: the marker closes the classify -> link race, not the reclaim-before-walk one.

Confidence: static only -- traced `runReclamation` -> `unlink` against `copyTree`, not reproduced against a live reclaim.

—
_Generated by Barber AI_

resources/blob.ts
```diff
@@ -1813,6 +1824,86 @@ export function repairBlobFile(
...
+	let header: Buffer;
+	let fileSize: number;
+	let handle: FileHandle;
+	try {
+		handle = await openFile(filePath, 'r');
+	} catch (error) {
+		// Deleted between the directory read and here. The checkpointed record may still reference it, so
+		// the id must stay reserved — but these bytes are gone for good, so it is captured terminal
+		// rather than retryable: a PENDING marker would promise a retry that can never succeed.
+		if ((error as { code?: string }).code === 'ENOENT') return 'gone';
```
cb1kenobi, line 1853: **Medium: `gone` -> terminal 500 is reasoned from the wrong producer of mid-walk deletes**

Decision 2 justifies terminal with "those bytes are unrecoverable". The dominant producer of ENOENT here is not an operator delete -- it is `runReclamation` unlinking a blob that was superseded *after* the engine checkpoint. `createBackup` runs `rootStore.backup()` first and `snapshotBlobs` only inside `finalizeBackup` (`dataLayer/rocksdbBackup.ts:296`), and reclamation fires 2 s after supersession by default (`resources/blob.ts:929`).

For those blobs the restored record is the pre-supersession version, whose bytes were complete and referenced at checkpoint time. Pre-PR that restored as an absent file -- a gap the repair machinery can act on. It now restores as `ERROR_TYPE`, which `resources/blob.ts:341` raises as `BLOB_CORRUPT_STATUS` and, per your own note, replicates as-is.

If terminal is still the right call, the reasoning should name the reclamation queue rather than "deleted mid-walk", because the two cases have different recoverability. The alternative worth pricing: `pending` for this arm, accepting one `blobReadTimeout` per read in exchange for keeping the entry healable.

Confidence: static only.

—
_Generated by Barber AI_

dataLayer/blobBackup.ts
```diff
@@ -66,24 +70,72 @@ async function linkOrCopy(src: string, dest: string): Promise<void> {
...
+
+/**
+ * Put one blob file into the snapshot, returning true when a PENDING marker stood in for it.
+ */
+async function captureBlobFile(srcPath: string, destPath: string): Promise<boolean> {
+	let disposition: BlobCaptureDisposition;
+	try {
+		disposition = await classifyBlobFileForCapture(srcPath);
+	} catch (error) {
+		if (isSystemicIoError(error)) throw error;
```
cb1kenobi, line 101: **Medium: three of the PR's five reviewer-facing decisions have no test that fails when reversed**

Mutation-tested against `blobBackup.test.js` + `rocksdbBackup.test.js` + `blob.test.js` (163 tests). Each mutation was compiled with `tsc -p tsconfig.build.json` and the marker grepped in `dist/` to confirm exactly one mutated site (`#src/*` resolves to `dist/*.js`, so editing the `.ts` alone is a no-op).

Survived -- 163 passing, 0 failing:
- disabling this `throw` in **both** walks, so a systemic errno degrades to `capture` instead of failing the backup (decision 3);
- `classifyBlobFileForCapture`'s ENOENT arm returning `'pending'` instead of `'gone'` (decision 2);
- `inflatesToExactly`'s non-`Z_` branch resolving `false` instead of rejecting -- the distinction that made the new `.catch(() => false)` in `isBlobFileComplete` necessary.

Killed, for contrast: neutering `blobHeaderIndicatesIncomplete` (4 failures), the `.repair` skip (1), `blobHeaderIsAbortMarker` (1). The capture rule itself is well covered; these three are not.

The two `gone` cases in `blob.test.js` mint the marker directly with `createCaptureMarker('gone', ...)`, so they assert the reader's 500 rather than that the classifier chose terminal. Calling `classifyBlobFileForCapture` on a path that does not exist would close the largest of the three in one line.

—
_Generated by Barber AI_

dataLayer/blobBackup.ts
```diff
@@ -118,8 +177,19 @@ export async function snapshotBlobs(backupDir: string, backupId: number, blobRoo
...
+		let captured = 0;
 		for (let index = 0; index < blobRoots.length; index++) {
-			await copyTree(blobRoots[index], join(tempDir, String(index)));
+			const counts = await copyTree(blobRoots[index], join(tempDir, String(index)), true);
+			substituted += counts.substituted;
+			captured += counts.captured;
+		}
+		if (substituted > 0) {
+			logger.warn(
+				`Blob snapshot for backup ${backupId} captured ${substituted} of ${substituted + captured} blob ` +
```
cb1kenobi: **Low: the generated `blobs/README.md` still describes the snapshot as pure hard links**

`blobsReadmeContent` closes with "Files are hard links to the live blobs when the backup is on the same filesystem, and copies otherwise." After this change an entry may instead be an 8-byte-header stub.

Decision 5 rules out a manifest count because it is a format change restore would have to read. The README is not: it is regenerated on every `snapshotBlobs` and embedded in every `get_backup` archive, and costs nothing to extend. A short paragraph naming the `0xfe` / `0xff` header types gives an operator inspecting a backup the record that this `logger.warn` does not survive long enough to provide.

—
_Generated by Barber AI_

dataLayer/blobBackup.ts
```diff
@@ -99,11 +151,18 @@ async function copyTree(srcRoot: string, destRoot: string): Promise<void> {
 			if (entry.isDirectory()) {
 				stack.push(srcPath);
 			} else if (entry.isFile()) {
-				await linkOrCopy(srcPath, join(destRoot, relative(srcRoot, srcPath)));
+				const destPath = join(destRoot, relative(srcRoot, srcPath));
+				if (classify) {
+					if (await captureBlobFile(srcPath, destPath)) counts.substituted++;
+					else counts.captured++;
```
cb1kenobi: **Nit: `.repair` skips are counted as captured**

`captureBlobFile` returns `false` for the `'skip'` disposition, so every `.repair` temp lands in `counts.captured` and pads the denominator of the "N of M blob file(s)" warning below. Returning the disposition from `captureBlobFile` and counting on that keeps the ratio meaningful.

—
_Generated by Barber AI_

resources/blob.ts
```diff
@@ -67,17 +76,19 @@ type StorageInfo = {
 };
 const FILE_STORAGE_THRESHOLD = 8192; // if the file is below this size, we will store it in memory, or within the record itself, otherwise we will store it in a file
 // We want to keep the file path private (but accessible to the extension)
-const HEADER_SIZE = 8;
+export const HEADER_SIZE = 8;
```
cb1kenobi: **Nit: three new exports have no consumer**

`dataLayer` imports only `classifyBlobFileForCapture`, `createCaptureMarker`, `isSystemicIoError`, and the `BlobCaptureDisposition` type. `HEADER_SIZE`, `BLOB_REPAIR_SUFFIX` (line 91), and `blobHeaderIsAbortMarker` (line 1882) are referenced nowhere outside `resources/blob.ts`. Given the note that harper-pro sees this surface, leaving them module-private keeps the capture API to the three things it needs.

—
_Generated by Barber AI_

dataLayer/blobBackup.ts
kriszyp: Fixed in c32be793e. The module contract now limits marker substitution to entries returned by readdir and explicitly states that reclamation before the parent directory is read remains outside the snapshot.

resources/blob.ts
kriszyp: Fixed in 5c86b5d6. The rationale now names reclamation after the engine checkpoint as the dominant producer and states the terminal-vs-pending tradeoff, including peer repair and replication consequences.

dataLayer/blobBackup.ts
kriszyp: Addressed in 5c86b5d6: added direct tests that a nonexistent path classifies as gone and that the fail-closed systemic errno set is preserved. Exact consumer fault injection and the non-Z inflate rejection would require stubbing an internal seam; AGENTS.md forbids adding stubs/sinon/rewire, so that residual coverage limitation is explicit rather than hidden behind a test-only seam.

dataLayer/blobBackup.ts
kriszyp: Fixed in 5c86b5d6. Generated blobs/README.md now documents complete hard-link/copy entries plus 0xfe PENDING and 0xff ERROR marker entries.

dataLayer/blobBackup.ts
kriszyp: Fixed in 5c86b5d6. captureBlobFile now returns the disposition; skip entries are excluded from captured and from the warning denominator.

resources/blob.ts
kriszyp: Fixed in 5c86b5d6. HEADER_SIZE, BLOB_REPAIR_SUFFIX, and blobHeaderIsAbortMarker are module-private again.

resources/blob.ts
```diff
@@ -1813,6 +1824,83 @@ export function repairBlobFile(
...
+		header = header.subarray(0, bytesRead);
+	} finally {
+		await handle.close();
+	}
+	if (blobHeaderIsAbortMarker(header)) return 'capture';
+	if (blobHeaderIndicatesIncomplete(header, fileSize)) return 'pending';
+	// Once stamped, a deflate header records the *uncompressed* length, so a torn compressed body left
+	// by an unclean shutdown can pass the size check above. Verify the stream rather than assuming the
+	// repair sweep has already run. Compression is opt-in per createBlob and unused inside Harper, so
+	// this reads nothing on an ordinary corpus.
```
cb1kenobi: **Low: the rewritten deflate rationale drops the case that actually motivates this check in a backup walk**

Traced the write path at this head: when `blob.size` is known, `saveBlob` writes `createHeader(blob.size)` before the first compressed byte and sets `wroteSize`, and the only re-stamp in `finished()` is gated on `!wroteSize`. So a compressed blob that is *still streaming* already carries its final uncompressed length and `DEFLATE_TYPE` at byte 0, and `blobHeaderIndicatesIncomplete` returns false for it — its length comparison applies only to `UNCOMPRESSED_TYPE`. The inflate verification below is the only thing that catches that file, and a live write met mid-walk is the case this PR exists for.

The unclean-shutdown case in the new wording is real, but it is the secondary one, and it is the case the asynchronous repair sweep already owns. As it now reads, a later maintainer has a standing invitation to delete this check as redundant with that sweep and silently reintroduce truncated compressed blobs into snapshots. The wording it replaced named the live-growth case; keeping both costs one clause.

Confidence: confirmed by code trace (`saveBlob` header/`wroteSize` handling plus `blobHeaderIndicatesIncomplete`); not exercised against a live compressed write.

—
_Generated by Barber AI_

dataLayer/blobBackup.ts
```diff
@@ -66,24 +71,70 @@ async function linkOrCopy(src: string, dest: string): Promise<void> {
...
+
+/**
+ * Put one blob file into the snapshot, returning how the entry was captured.
+ */
+async function captureBlobFile(srcPath: string, destPath: string): Promise<BlobCaptureDisposition> {
+	let disposition: BlobCaptureDisposition;
+	try {
+		disposition = await classifyBlobFileForCapture(srcPath);
+	} catch (error) {
+		if (isSystemicIoError(error)) throw error;
```
cb1kenobi, line 101: **Low (was Medium): mutations re-run at `3305f988` — one of the three gaps is closed, two survive and now live only in a PR comment**

Re-ran all three at this head. `tsc -p tsconfig.build.json` before every run, marker grepped in `dist/` to confirm exactly one mutated site (`#src/*` resolves to `dist/*.js`, so editing the `.ts` alone is a no-op). Baseline here is 165 passing / 0 failing.

- `classifyBlobFileForCapture`'s ENOENT arm returning `'pending'` instead of `'gone'` → **killed** (164 passing / 1 failing: `classifies a blob reclaimed before it can be opened as gone`). That was the largest of the three.
- this `throw` disabled in **both** walks → **survives** (165 / 0). `recognizes only host-level I/O errors as systemic` pins the contents of `SYSTEMIC_IO_ERRORS`, not that either consumer acts on it, so the set can stay correct while both walks degrade a systemic fault to `capture`.
- `inflatesToExactly`'s non-`Z_` branch resolving `false` instead of rejecting → **survives** (165 / 0).

Controls still die: `blobHeaderIndicatesIncomplete` neutered (22 failures — wider than the 4 from the last pass because this mutation sits at the top of the function and so reaches every consumer, not just the classifier), the `.repair` skip (1), `blobHeaderIsAbortMarker` (1).

Agreed on the constraint: AGENTS.md:196 forbids new `sinon`/`rewire` and says to describe what is missing and stop, which is what you did — downgrading to Low on that basis. The one thing worth carrying over is that the description lives in a PR comment, which does not survive into the tree. A clause at each of the two sites saying no test covers the rethrow / the reject, and why, keeps the next reader from taking the branch as verified.

—
_Generated by Barber AI_

resources/blob.ts
kriszyp: Fixed in 4e6d65ad3 (wording landed in 54ebbc979). Your trace is right and I confirmed it independently: `createHeader(blob.size)` is written before the first compressed byte with `wroteSize = true`, and `blobHeaderIndicatesIncomplete` only length-compares `UNCOMPRESSED_TYPE`, so a live compressed write passes every check above and the inflate verification is its only detector.

The comment now leads with that case and keeps the unclean-shutdown one as the second producer, so neither reads as redundant with the repair sweep.

Chasing this also turned up something bigger, which is now the opening note in `## For the human reviewer`: the classifier's `capture` contract claimed "no path rewrites a published blob file in place", and the replication PENDING stamp at `resources/blob.ts:1572` does exactly that with an `O_TRUNC` `writeFile`. A known-size write that has landed every byte but not settled classifies as `capture`, so a same-filesystem hard link can be truncated to an 8-byte stub *after* capture. I narrowed the contract rather than changing behavior — the window cannot be closed without the blob write lock — but it wants a human decision.

— Claude Opus

dataLayer/blobBackup.ts
kriszyp: Fixed in 54ebbc979 — agreed that a description living only in a PR comment does not survive into the tree, so both branches now say it where they are:

- `dataLayer/blobBackup.ts` at the systemic rethrow: only the membership of `SYSTEMIC_IO_ERRORS` is tested, because reaching the rethrow needs a real host-level fault no test here can produce. The archive path in `dataLayer/rocksdbBackup.ts` carries the short form.
- `resources/blob.ts` at `inflatesToExactly`'s `fail`: the reject arm is unverified; the resolve arm is covered by the truncated-deflate tests.

One thing worth flagging: my pre-push review's Harper-domain leg pushed back on these, reading "unverified"/"untested" as review-status metadata rather than documentation, and wanted them dropped. I kept them — your point that an uncovered branch otherwise reads as verified is the one that matters to a future reader — but reworded them to state the constraint rather than narrate the round. If Kris prefers the domain leg's view, they are cheap to remove.

— Claude Opus

resources/blob.ts
cb1kenobi: Verified at `4e6d65ad` — confirming this independently rather than taking the fix on trust.

Re-traced the write path in the worktree at this head:

- `resources/blob.ts:1669-1671` — `if (blob.size !== undefined && !repairing) { writeStream.write(createHeader(blob.size)); wroteSize = true; }`, and `createHeader` ORs `DEFLATE_TYPE << 48n` when `compress` is set. So the header is stamped `DEFLATE_TYPE` with a real length *before* `createDeflate()` is even constructed at `:1690`.
- `blobHeaderIndicatesIncomplete` at `:1915` ends in `return type === UNCOMPRESSED_TYPE && fileSize !== HEADER_SIZE + storedSize` — the length comparison genuinely never applies to a deflate header.

So a still-streaming compressed blob does pass the size check, and the inflate verification is the only thing that catches it. The new wording at `:1866-1871` now names both producers (live write, and a torn body from an unclean shutdown the async repair sweep hasn't reached), which is what makes the check read as load-bearing instead of redundant. Closing this one.

Also worth calling out: commit `4e6d65ad3` narrowing the `capture` contract at `:1836-1840` goes further than I asked for and is accurate — I checked the abort path at `:1572`, which does `writeFile(filePath, ...)` with the default `'w'` flag, so it truncates that inode in place and a same-filesystem hard link already taken by the walk does follow it down. Documenting that honestly rather than overstating the invariant is the right call, especially since closing it really would need the file-id-keyed write lock a directory walk can't reach.

Generated by Barber AI

dataLayer/blobBackup.ts
cb1kenobi: Re-ran the mutations at `4e6d65ad` and this is resolved — with a result worth reporting, because the mutations now independently corroborate the comments you added.

Method (same as last pass): mutate the `.ts`, `tsc -p tsconfig.build.json`, grep the marker in `dist/` to prove the mutation is actually loaded (the `#src/*` -> `./dist/*.js` mapping means an unbuilt `.ts` edit is a silent no-op), then run `unitTests/dataLayer/blobBackup.test.js` + `unitTests/dataLayer/rocksdbBackup.test.js`. Baseline at this head: **59 passing, green**.

| Mutation | Result |
| --- | --- |
| systemic-errno `throw` disabled in **both** walks | 59 passing — **survives** |
| `inflatesToExactly` non-`Z_` branch resolves `false` instead of rejecting | 59 passing — **survives** |
| `blobHeaderIndicatesIncomplete` -> always `false` *(control)* | 4 failing — dies |
| `.repair` skip disabled *(control)* | 1 failing — dies |
| `blobHeaderIsAbortMarker` -> always `false` *(control)* | 1 failing — dies |

The three controls dying is what makes the two survivors trustworthy rather than a broken harness.

Both branches still survive, and that is fine — the point of this finding was never that they had to be killed. Reaching either one needs a real host-level fault (a systemic errno from `classifyBlobFileForCapture`, or a read fault raised mid-inflate on a file that opened cleanly), which no unit test here can honestly produce. What I flagged was that the *description* of that gap lived only in a PR comment, where it doesn't survive into the tree. It now says so at all three sites — `dataLayer/blobBackup.ts:99`, `dataLayer/rocksdbBackup.ts:859`, `resources/blob.ts:2856` — and the mutation run confirms each of those "unverified" claims is literally true. A comment that accurately reports its own coverage is a better outcome than a test that fakes the fault. Closing.

Generated by Barber AI

resources/blob.ts
cb1kenobi: Not a finding on this PR — flagging a cross-PR interaction so it isn't discovered after both land.

The rationale here rests on reclamation unlinking a superseded blob *after* the engine checkpoint, and converts that case into a **terminal** `0xff` marker on the grounds that those bytes are already unrecoverable by the time the walk sees the gap.

[harper#1835 — *fix(blob): durable blob-unlink queue — deletions survive worker/process death; orphan observability (#1832)*](https://github.com/HarperFast/harper/pull/1835) (still open as of this comment, head `1c238d4d`) has an open High whose currently-suggested fix moves the durable unlink row **earlier**, into `deleteBlob()` itself. If that lands as suggested, the window this PR reasons about stops being a transient in-process one and becomes a state that survives restart — a queued unlink can be replayed against a blob a later backup walk is still classifying. That makes "terminal" a more expensive choice than it is today, because the recovery path that no longer applies is the one restart would otherwise have provided.

Nothing to change here — the two PRs are independently correct as written. Just worth whoever merges second re-reading Decision 2 against the other PR's final shape rather than assuming it still holds.

Generated by Barber AI

kriszyp: Recorded rather than silently agreed-with: this merge-order interaction now lives in the PR description as a sub-note under decision 2, so it survives outside a resolved review thread and reaches whoever promotes this from draft.

I re-checked your premise at harper#1835's current head (`fix/blob-concurrent-replace-orphans`): the High is still unresolved and its suggested fix does move the durable row into `deleteBlob()`. If that shape lands, the consequence for this arm is narrower than "terminal gets worse" in general — the ENOENT classification is unchanged either way, but the population reaching it grows, because today a worker death inside the retention window drops the in-memory intent and the bytes survive as an orphan for `cleanup_orphan_blobs`, and with a durable row from `deleteBlob()` that unlink is replayed instead. So the change is in how often a checkpointed record restores as a terminal `0xff`, not in whether terminal is the right disposition for a genuinely reclaimed blob.

No code change on this PR for it, per your note. Whoever merges second re-reads decision 2 against the other PR's final shape.

— Claude Opus

Issue #644, opened 2026-05-20
Branched databases: blob store sharing, allocator, and GC safety


kriszyp:
## Summary

Wire blob handling for branched databases (#642, #643) using a **hard-link clone of the blob tree**, taken alongside the RocksDB checkpoint. This is the decided design, replacing an earlier shared-store-plus-high-water-mark approach this issue previously described — that approach is not being built.

## Design

A branch materializes as a checkpoint of the base's RocksDB store (`resources/branchDatabase.ts`, `materializeBranch`). RocksDB's `createCheckpoint` covers only the SST/WAL files; it does not touch the blob directories that live alongside the database (`getRootBlobPathsForDB`, `resources/blob.ts:2074`). This issue adds the missing half: when a branch is materialized, recursively **hard-link** the base's blob tree into the branch's own blob root.

The mechanism already exists and is proven: `dataLayer/blobBackup.ts`'s `linkOrCopy`/`copyTree` (lines 58, 133) already walk a blob tree and hard-link each file, falling back to a real copy on `EXDEV`/`EMLINK`/`EPERM`/`ENOTSUP` (cross-filesystem, link-count exhaustion, or a filesystem that forbids hard links) — reuse it rather than re-implement it. `resources/databases.ts:1299` already notes this same same-filesystem constraint for the RocksDB checkpoint itself, so a branch already degrades to a full byte copy off-volume; the blob clone should follow the same rule for the same reason.

Once cloned, everything else falls out of ordinary filesystem semantics — no allocator sharing, no high-water mark, no gated deletion logic:

- **Blob directory:** the branch gets its own blob root (`<branch's blob path>`, resolved the same way `getBlobPathsForDatabaseName` resolves any other database's), populated by the hard-link clone at branch-creation time. It is a real, independent directory — just one whose existing files share inodes with the base's.
- **ID allocator:** the branch gets its **own** allocator, exactly like any other database — no sharing with the base or with other branches. `getNextFileId` (`resources/blob.ts:2179`) already seeds by scanning the directory for the highest existing filename-derived ID; since the branch's directory is a full clone (same filenames as the base had at checkpoint time), it naturally seeds above every ID that existed at checkpoint, and its own directory is separate from every other branch's, so two branches allocating new IDs concurrently cannot collide with each other or with the base. No new allocator-sharing code is needed at all.
[...]

Issue #643, opened 2026-05-20
Branched databases: app config, scope proxy, lifecycle wiring


kriszyp:
## Summary

Harper-side implementation of branched databases. Subtask of HarperFast/harper#642; depends on HarperFast/rocksdb-js#577 for the underlying checkpoint primitive.

## Scope

Everything inside \`core/\` that wires the feature together. No native code changes.

### 1. Application config schema

\`core/components/Application.ts\` — extend \`ApplicationConfig\` (lines 28–38) with:

\`\`\`ts
branchedDatabases?: string[];
\`\`\`

Validate in \`assertApplicationConfig()\` (lines 72–109): array of strings; each must name an existing database; reject duplicates.

### 2. ApplicationScope field

\`core/components/ApplicationScope.ts\` — add:

\`\`\`ts
branches: Map<string, string>; // baseDbName → branchDbName
\`\`\`

### 3. Branch creation on app load

Extend \`prepareApplication()\` in \`core/components/Application.ts\` (lines 429–480). After the \`ApplicationScope\` is constructed but **before** any of the app's modules import \`harper\`:

For each \`baseName\` in \`config.branchedDatabases\`:

- Branch DB name: \`\${scope.appName}:\${baseName}\` (colon-namespaced in the \`databases\` map).
- Branch path: \`<storage_root>/database/\${scope.appName}__\${baseName}\` (use \`__\` on disk; colon would be illegal on some filesystems).
- Resolve the base \`RocksDatabase\` via \`database({ database: baseName })\` (\`core/resources/databases.ts:740\`).
- \`await baseDb.createCheckpoint(branchPath)\`.
- Open the checkpoint as a database via the same factory, override path; register under the namespaced key.
- Record \`scope.branches.set(baseName, '\${scope.appName}:\${baseName}')\`.

The checkpoint is byte-identical to the base, so \`internal_dbis/\` metadata is correct and \`makeTable()\` (\`core/resources/Table.ts:146\`) works unchanged — Table classes get \`primaryStore\` pointing at the branch.

### 4. Scoped \`databases\` proxy in \`getHarperExports\`

\`core/security/jsLoader.ts\` (\`getHarperExports\` lines 678–716). Replace the static \`databases\` reference with a \`Proxy\` that consults \`scope.branches\`:

\`\`\`ts
const scopedDatabases = scope.branches.size === 0
  ? databases
  : new Proxy(databases, {
      get(target, prop, recv) {
[...]

Issue #642, opened 2026-05-20
Branched databases for application isolation


kriszyp:
## Summary

Allow a deployed application to be configured to use a **branched** copy of one or more databases. A branch is a writable, isolated fork of a base database, invisible to other applications and to replication. The app's code is unchanged — it imports `databases.data` from `harper` exactly as it does today, but the binding resolves to the branch.

A branch is **durable and deterministically named**: it lives at `` <storage>/`branches`/<app>/<db> ``, derived only from the application and base database names, and is adopted again on restart rather than re-checkpointed. Nothing process-local appears in the path. That is deliberate — an isolated application deployed onto a cluster is expected to be transparently available across it, so every node must resolve the same application's branch to the same place, and that identity is what a future replicated branch would be addressed by.

## Motivation

The driving use case is **isolated application versions**: deploy multiple variants of an app to the same Harper instance (or cluster), each with its own isolated view of `data`, so writes don't collide and tests don't pollute the shared database. Today there is no way to give an app a private writable copy of an existing database short of running a separate instance.

## UX

Declared on the application's **root-config entry**, alongside `host` and `urlPath` — which databases an application forks is a deployment decision, not something the application checks in. Declaring it in the application's own `config.yaml` is refused rather than ignored.

```yaml
my-app:
  package: my-app
  host: api.example.com
  branchedDatabases:
    - data
```

`branchedDatabases: true` forks every database on the instance except `system`. It is a snapshot at load, not a subscription: a database created afterward is not retroactively branched.

The app then imports from `harper` normally:

```js
import { databases } from 'harper';
await databases.data.MyTable.put({ id: 'x', ... });
```

**The `import` is load-bearing.** A branch is delivered through the module loader's `harper` exports; the bare `databases`/`tables` globals are shared process-wide by the default `vm-current-context` loader and cannot be scoped, so an application that reaches for them reads and writes the BASE, silently. Per-application globals are a property of the thread-isolation track, not of branching.

## Approach

**RocksDB checkpoint + hard-linked blob clone + per-application `databases` binding.**

1. On app load, for each declared database, `createCheckpoint()` into a staging sibling and rename it into place, so a crash mid-copy leaves debris rather than a half-populated directory RocksDB would refuse to open. Concurrent worker threads settle who takes the checkpoint through a claim word in a buffer the base store shares across threads.
[...]

PR #2285, opened 2026-08-22
Let a database be opened into a caller-owned table graph instead of the global map

kriszyp:
Branched databases (#642) need a database whose Table classes are reachable **only** through the application scope that asked for it. Registering one in the process-global `databases` map and asking every consumer to skip it does not work: that map is walked by analytics (`storeDBSizeMetrics`), by `describe_all` (`schemaDescribe`), and by worker teardown (`closeLoadedDatabases`), and Pro walks it too. A skip list is a rule each future enumerator has to remember, and the one that forgets leaks a branch into a customer's metrics or a describe response.

So [`initStores` takes a **destination**](https://github.com/HarperFast/harper/pull/2285/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R713) instead. With one it builds into the caller's object, seeds that object's defined-tables bookkeeping, emits no global `updateTable`, and replays the transaction log into that same graph. Without one, every existing path behaves exactly as before. `openBranchDatabase(path, databaseName, storeName)` is the narrow entry point on top, returning a closeable handle.

This is the storage-side opener only — **nothing calls it yet.** The config, scope wiring, and lifecycle that expose a branch to application code are the rest of #643.

Refs #643.

## For the human reviewer

Reviewing this at the level of "is the mechanism sound" is the right frame; there is no consumer to judge it against. The three items below are open decisions, not defects I chose not to fix — each is a #643 layout or contract question rather than a local edit, and each is stated in the opener's docstring so a first caller cannot miss it.

1. **Branch ownership is process-local, so the scan window is not fully closed** (`persistent-branch-ownership`). [`isOpenBranchPath`](https://github.com/HarperFast/harper/pull/2285/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1046) keeps `getDatabases()` out of a branch **while a handle is open**, and the same guard now covers the [on-demand `database()` open](https://github.com/HarperFast/harper/pull/2285/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1342). It does not cover the before-first-open or after-crash cases: a branch directory left inside the scanned root by a SIGKILLed worker loads as an ordinary, globally-enumerated database at the next boot. I tried moving the ownership claim ahead of the `getDatabases()` inside the open, which closes that window — and reverted it, because a claim taken before the scan also means a directory that genuinely **is** a real database stops reading as one, so `openBranchDatabase` would succeed over it and that database would vanish from the thread. The durable fix is an on-disk marker (the way the scan already skips `MIGRATING_DIR_SUFFIX` and `RESTORE_META_DIR`) or a branch path outside the scanned root — both are #643 §3 layout decisions. The reasoning is [recorded at the guard it constrains](https://github.com/HarperFast/harper/pull/2285/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1108).
2. **What a branch is, exactly** (`branch-replay-contract`). `replayLogs` returns immediately off the main thread, so a branch opened on a worker — the intended consumer — is the checkpoint's SST content and nothing more; its transaction-log tail is silently absent. It is also not awaited, so a `close()` shortly after an open can race a replay still writing. Neither can bite while nothing calls this, and both are contracts callers will build on, so they are stated on the opener rather than papered over. Forcing replay for branches regardless of thread is the alternative and it is a behavior change to a boot-critical routine.
3. **Ships with schema mutation unsafe** (`schema-mutation-exposure`). A branch's Table classes carry the base's logical name, so a `dropTable()` through one resolves against the global schema and would delete the live base Table class. Kris's call was to gate it in the scope-wiring PR *before any caller exists*; it is in the docstring and on #643. If you would rather see a branch flag rejecting DDL here, that is cheaper now than after callers exist.

Two smaller judgment calls, for completeness: `removeStorageReclamation(path)` drops every handler registered for a path, which is safe only because the guards keep a branch path from also being a loaded database; and `closeDatabase` has the same reclamation/`databasePaths` pins that `closeBranchHandles` now releases, left alone as pre-existing and out of this diff's scope.

## What changed since the first review round

Four rounds of independent review (Codex, Gemini, Cursor Composer, Harper domain) plus the CI bots landed these:

- **Two branches could claim one store identity.** `storeName` was checked against real database names only. Blob file ids restart from each store's own checkpointed counter, so two branches sharing an identity do not merely share a directory — they generate identical file paths and truncate each other. [Live identities are now reserved](https://github.com/HarperFast/harper/pull/2285/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1038).
- **A stale handle's second `close()` tore down its successor.** The re-close guard was `if (!openBranches.delete(path)) return` — keyed by path, and a closed branch frees its path, so a discarded handle deregistered whichever branch owned the path by then, dropping the live one out of `rocksdbDatabaseEnvs` and removing its reclamation handler while its RocksDB handles stayed open. [Now keyed to the handle](https://github.com/HarperFast/harper/pull/2285/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1134). Found by writing the mutation-killing teardown test the review asked for.
- **The partial-open leak this PR previously recorded as a residual is closed.** `initStores` opens a table's column families well before `setTable` publishes it, so cleanup that walked the graph could not see them. It now [appends every column family it opens](https://github.com/HarperFast/harper/pull/2285/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R910) to a caller-supplied list — recorded before the wrapper that could throw — and the branch [releases that list](https://github.com/HarperFast/harper/pull/2285/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1184) on failure and on close, replacing the graph walk entirely.
- **`close()` left two process-global registrations behind:** a storage-reclamation handler pinning the closed store, and the memoized blob roots in `blob.ts`'s `databasePaths`. `onStorageReclamation` gains `removeStorageReclamation`, and both are released.
- **`closeBranchDatabases()` had no caller,** so a branch open on a job worker leaked process-wide on exit. [`closeLoadedDatabases` runs it](https://github.com/HarperFast/harper/pull/2285/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1555).
- **A tableless branch never got its identity stamped,** so its blob roots resolved through `join(undefined, …)`. The stamp [moved to the top of `initStores`](https://github.com/HarperFast/harper/pull/2285/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R713), which fixes it for a tableless real database too.
- **`storeName` lands in a filesystem path** and was unvalidated; it is now checked for shape and [length](https://github.com/HarperFast/harper/pull/2285/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1066) against the same rules every other database name obeys.
- **`isOpenBranchPath` resolved the realpath before consulting the map** and returned false when that threw, so a directory unlinked under a live handle read as unowned while `rocksdbDatabaseEnvs` still held the branch store under that literal key.
- **The doc claimed "row reads and writes are fine".** Not true of a row whose blob predates the checkpoint: a branch's blob roots start empty until #644 links the base tree in.

## Verification

Suites: `databases.test.js`, `closeLoadedDatabases.test.js`, `caching-rocks-database.test.js`. Full `test:unit:resources` gate: **1690 passing, 16 pending, 3 failing** — the three failures (`randomAccessFields` ×2, `replayStructures`) reproduce unchanged on `origin/main`'s own copies of those files and are untouched by this diff. `test:unit:main` cannot complete on this machine: it dies at load reading a stale `system.mdb` in a shared local hdb root, identically with and without this change. `tsc --noEmit`, `lint:required`, prettier clean. Relying on CI for both full gates.

Teardown claims are proved against `registryStatus().refCount` — the process-global count only drops to zero once every column family under a path is closed — and each new case was mutation-checked: [column-family release](https://github.com/HarperFast/harper/pull/2285/changes?w=1#diff-d89a878c46959d4e1a9364b6009ecf89d8ca08cc125e71357911283edce9b0ef R313), [`databasePaths` cleanup](https://github.com/HarperFast/harper/pull/2285/changes?w=1#diff-d89a878c46959d4e1a9364b6009ecf89d8ca08cc125e71357911283edce9b0efR345), reclamation deregistration, the scan skip, the [identity collision](https://github.com/HarperFast/harper/pull/2285/changes?w=1#diff-d89a878c46959d4e1a9364b6009ecf89d8ca08cc125e71357911283edce9b0efR272), the [`database()` refusal](https://github.com/HarperFast/harper/pull/2285/changes?w=1#diff-d89a878c46959d4e1a9364b6009ecf89d8ca08cc125e71357911283edce9b0efR293), and the [stale-handle re-close](https://github.com/HarperFast/harper/pull/2285/changes?w=1#diff-d89a878c46959d4e1a9364b6009ecf89d8ca08cc125e71357911283edce9b0efR330) each fail with their behavior removed. A [write through the branch](https://github.com/HarperFast/harper/pull/2285/changes?w=1#diff-d89a878c46959d4e1a9364b6009ecf89d8ca08cc125e71357911283edce9b0efR281) is now asserted not to reach the base table.

The partial-open release has **no permanent test**: `initStores` has no fault-injection seam. It was verified with a throw injected immediately after a primary store opens — the registry refCount for the branch path returns to 0 with the release and stays at 2 without it — and the injection removed rather than shipped as a stub.

End-to-end route: **not observable end-to-end in this PR** — nothing calls the opener, so the integration proof lands with the scope wiring in #643.

Complexity: medium

<sub>Review-Coverage: authored=claude; ran=codex,gemini; declined=cursor-grok,cursor-composer,domain; rounds=4 @ 08e0674331ca</sub>

<sub>Human-Review-Need: 4 @ 08e0674331ca</sub>

### DIFF
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -1055,6 +1055,10 @@ export function openBranchDatabase(path: string, databaseName: string, storeName
 	// The guards below compare against keys in the env map, so two spellings of one directory must
 	// not read as two directories.
 	path = realpathSync(path);
+	// Load the registry FIRST: the guards below read it, and loading is itself what populates
+	// `rocksdbDatabaseEnvs` — running the path guard before the scan would let this branch adopt a
+	// database that the scan then opened at the same path.
+	getDatabases();
 	// A second open would hand back a rival table graph over one shared root store, and the two
 	// callers would disagree about who may close it.
 	if (openBranches.has(path)) throw new Error(`Branch database at ${path} is already open`);
@@ -1076,11 +1080,8 @@ export function openBranchDatabase(path: string, databaseName: string, storeName
 		// in the env map with no handle able to close it.
 		const stranded = rocksdbDatabaseEnvs.get(path);
 		rocksdbDatabaseEnvs.delete(path);
-		try {
-			stranded?.close();
-		} catch (closeError) {
-			logger.warn?.(`Error closing partially opened branch database at ${path}`, closeError);
-		}
+		// `tables` holds whatever initStores managed to build before it threw.
+		closeBranchHandles(path, stranded, tables);
 		throw error;
 	}
 	const branch: BranchDatabase = {
@@ -1089,24 +1090,38 @@ export function openBranchDatabase(path: string, databaseName: string, storeName
 		close() {
 			if (!openBranches.delete(path)) return; // already closed; closing a store twice is not safe
 			rocksdbDatabaseEnvs.delete(path);
-			// The column families opened for the tables and the audit store hold their own handles, and
-			// closing only the root leaves them behind.
-			for (const store of [(rootStore as any).dbisDb, (rootStore as any).auditStore]) {
-				try {
-					store?.close?.();
-				} catch (error) {
-					logger.warn?.(`Error closing branch column family at ${path}`, error);
-				}
-			}
-			try {
-				rootStore.close();
-			} catch (error) {
-				logger.warn?.(`Error closing branch database at ${path}`, error);
-			}
+			closeBranchHandles(path, rootStore, tables);
 		},
 	};
 	openBranches.set(path, branch);
 	return branch;
+}
+
+/**
+ * Release every RocksDB handle a branch open created. `initStores` opens a column family per table
+ * primary store and per index, on top of the internal-dbis and audit families, so closing only the
+ * root leaves all of them behind — which is why `closeDatabase` walks them individually for a real
+ * database. Shared by `close()` and the partial-open cleanup, so a failure part-way through
+ * `initStores` releases exactly what a success would.
+ */
+function closeBranchHandles(path: string, rootStore?: RootDatabaseKind, tables?: Tables): void {
+	const closeStore = (store: any, description: string) => {
+		try {
+			store?.close?.();
+		} catch (error) {
+			logger.warn?.(`Error closing ${description} for branch database at ${path}`, error);
+		}
+	};
+	for (const tableName in tables ?? {}) {
+		const table: any = (tables as any)[tableName];
+		if (!table?.primaryStore) continue;
+		for (const indexName in table.indices || {})
+			closeStore(table.indices[indexName], `index ${tableName}.${indexName}`);
+		closeStore(table.primaryStore, `table ${tableName}`);
+	}
+	closeStore((rootStore as any)?.dbisDb, 'attributes store');
+	closeStore((rootStore as any)?.auditStore, 'audit store');
+	closeStore(rootStore, 'root store');
 }
 
 /** Close every open branch. Branches are process-local, so this is shutdown, not a data operation. */
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -30,7 +30,9 @@ const { forComponent } = harperLogger;
 import * as manageThreads from '../server/threads/manageThreads.js';
 import { openAuditStore, readAuditEntry, createAuditEntry, type AuditRecord } from './auditStore.ts';
 import { handleLocalTimeForGets } from './RecordEncoder.ts';
-import { deleteRootBlobPathsForDB } from './blob.ts';
+import { databasePaths, deleteRootBlobPathsForDB } from './blob.ts';
+import { removeStorageReclamation } from '../server/storageReclamation.ts';
+import { schemaRegex } from '../validation/common_validators.ts';
 import { CUSTOM_INDEXES } from './indexes/customIndexes.ts';
 import { OpenDBIObject } from '../utility/lmdb/OpenDBIObject.ts';
 import { RocksDatabase, supportedCompression, type RocksDatabaseOptions } from '@harperfast/rocksdb-js';
@@ -454,6 +456,7 @@ export function getDatabases(): Databases {
 			const dbName = basename(databaseEntry.name, '.mdb');
 			const dbPath = join(databasePath, databaseEntry.name);
 			if (blockedByRestore.has(dbName)) continue;
+			if (isOpenBranchPath(dbPath)) continue;
 
 			if (
 				databaseEntry.isFile() &&
@@ -516,6 +519,7 @@ export function getDatabases(): Databases {
 					if (databaseEntry.name.endsWith(MIGRATING_DIR_SUFFIX)) continue; // migration staging dir
 					if (databaseEntry.name === RESTORE_META_DIR) continue; // reserved restore-metadata dir
 					if (blockedByRestore.has(basename(databaseEntry.name, '.mdb'))) continue;
+					if (isOpenBranchPath(join(databasePath, databaseEntry.name))) continue;
 					if (databaseEntry.isFile() && extname(databaseEntry.name).toLowerCase() === '.mdb') {
 						readMetaDb(join(databasePath, databaseEntry.name), basename(databaseEntry.name, '.mdb'), dbName);
 					} else {
@@ -1032,6 +1036,33 @@ export interface BranchDatabase {
 const openBranches = new Map<string, BranchDatabase>();
 
 /**
+ * True when `dbPath` is a directory an open branch owns. The database scan opens any directory that
+ * holds CURRENT + MANIFEST-*, and harper#643 places a branch inside the directory it walks, so
+ * without this the first `resetDatabases()` after a branch is open would rebuild that branch's tables
+ * into the global map, overwrite the store identity its blob roots resolve from, and hand its store
+ * to `closeLoadedDatabases`.
+ */
+function isOpenBranchPath(dbPath: string): boolean {
+	if (openBranches.size === 0) return false;
+	try {
+		return openBranches.has(realpathSync(dbPath));
+	} catch {
+		return false;
+	}
+}
+
+/**
+ * A branch identity resolves its blob roots through `join(…, 'blobs', storeName)`, so it must be a
+ * single path segment: `schemaRegex`, which every other database name is validated against, plus the
+ * dot segments and backslash that regex permits but a path component must not be.
+ */
+function assertLegalBranchName(name: string, description: string): void {
+	if (!name || !schemaRegex.test(name) || name.includes('\\') || name === '.' || name === '..') {
+		throw new Error(`Cannot use '${name}' as a branch ${description}: it is not a legal database name`);
+	}
+}
+
+/**
  * Open a RocksDB directory as a **scope-private** database: its Table classes are built into an
  * object the caller owns and nothing is registered in the global `databases` map, so no enumerator
  * of that map — analytics, `describe_all`, worker teardown, replication — can observe it.
@@ -1041,8 +1072,10 @@ const openBranches = new Map<string, BranchDatabase>();
  * blob directories from, which is how a branch gets its own blob roots rather than writing into the
  * base's.
  *
- * The caller owns the returned handle. `closeLoadedDatabases` walks `databases` and therefore cannot
- * see a branch, so nothing else will close it.
+ * The caller owns the returned handle: no enumerator of `databases` can reach a branch, so nothing
+ * closes it on the caller's behalf except `closeBranchDatabases`, which `closeLoadedDatabases` runs
+ * at thread teardown so a branch left open on an exiting worker does not leak its process-global
+ * RocksDB handles.
  *
  * NOT YET SAFE FOR SCHEMA MUTATION. A branch's Table classes carry the base's logical name, so a
  * `dropTable()` or equivalent through one resolves against the global schema and would delete the
@@ -1051,6 +1084,8 @@ const openBranches = new Map<string, BranchDatabase>();
  * (harper#643).
  */
 export function openBranchDatabase(path: string, databaseName: string, storeName: string): BranchDatabase {
+	assertLegalBranchName(databaseName, 'logical database name');
+	assertLegalBranchName(storeName, 'store identity');
 	if (!existsSync(path)) throw new Error(`Cannot open branch database: no directory at ${path}`);
 	// The guards below compare against keys in the env map, so two spellings of one directory must
 	// not read as two directories.
@@ -1098,14 +1133,20 @@ export function openBranchDatabase(path: string, databaseName: string, storeName
 }
 
 /**
- * Release every RocksDB handle a branch open created. `initStores` opens a column family per table
- * primary store and per index, on top of the internal-dbis and audit families, so closing only the
- * root leaves all of them behind — which is why `closeDatabase` walks them individually for a real
- * database. Shared by `close()` and the partial-open cleanup, so a failure part-way through
+ * Release everything a branch open created. `initStores` opens a column family per table primary
+ * store and per index, on top of the internal-dbis and audit families, so closing only the root
+ * leaves all of them behind — which is why `closeDatabase` walks them individually for a real
+ * database. It also leaves two process-global registrations behind the stores it closes, neither of
+ * which has a lifetime of its own: a storage-reclamation handler per store path, whose closure pins
+ * the now-closed store, and the memoized blob roots in `databasePaths`. A real database is opened
+ * once per thread, but harper#643 makes branch open/close routine, so both would grow with branch
+ * churn. Shared by `close()` and the partial-open cleanup, so a failure part-way through
  * `initStores` releases exactly what a success would.
  */
 function closeBranchHandles(path: string, rootStore?: RootDatabaseKind, tables?: Tables): void {
+	const reclamationPaths = new Set<string>([path]);
 	const closeStore = (store: any, description: string) => {
+		if (store?.path) reclamationPaths.add(store.path);
 		try {
 			store?.close?.();
 		} catch (error) {
@@ -1122,6 +1163,8 @@ function closeBranchHandles(path: string, rootStore?: RootDatabaseKind, tables?:
 	closeStore((rootStore as any)?.dbisDb, 'attributes store');
 	closeStore((rootStore as any)?.auditStore, 'audit store');
 	closeStore(rootStore, 'root store');
+	if (rootStore) databasePaths.delete(rootStore as RootDatabase);
+	for (const reclamationPath of reclamationPaths) removeStorageReclamation(reclamationPath);
 }
 
 /** Close every open branch. Branches are process-local, so this is shutdown, not a data operation. */
@@ -1474,8 +1517,13 @@ export function closeDatabase(databaseName: string): boolean {
  * database is closed). The `system` database is intentionally left open: it is non-enumerable here
  * (skipped by the loop), is never restored online, and the exiting worker may still touch the job
  * table during teardown. Best-effort: closing failures are swallowed inside `closeDatabase`.
+ *
+ * Branches are not in `databases` and so are invisible to the loop below, but they hold handles from
+ * the same process-global registry — this is the thread's one teardown entry point, so it closes
+ * them too rather than leaving a rule for the next exit path to remember.
  */
 export function closeLoadedDatabases(): void {
+	closeBranchDatabases();
 	// snapshot the names first: closeDatabase() deletes from `databases` as it goes
 	for (const databaseName of Object.keys(databases)) {
 		const dbTables = databases[databaseName];
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -652,8 +652,7 @@ function readRocksMetaDb(
 	path: string,
 	defaultTable?: string,
 	databaseName: string = DEFAULT_DATABASE_NAME,
-	destination?: Tables,
-	storeName?: string
+	{ destination, storeName, openedStores }: Pick<InitStoresOptions, 'destination' | 'storeName' | 'openedStores'> = {}
 ) {
 	try {
 		logger.trace(`loading rocksdb database: ${path}`);
@@ -667,14 +666,13 @@ function readRocksMetaDb(
 
 		let rootStore: RocksRootDatabase | undefined = rocksdbDatabaseEnvs.get(path);
 		if (rootStore) {
-			initStores(path, rootStore, databaseName, { defaultTable, destination, storeName });
+			initStores(path, rootStore, databaseName, { defaultTable, destination, storeName, openedStores });
 		} else {
 			rootStore = openRocksDatabase(path, { disableWAL: false, enableStats: true }) as any;
 			rocksdbDatabaseEnvs.set(path, rootStore);
-			initStores(path, rootStore, databaseName, { defaultTable, destination, storeName });
-			// Replay into whichever graph this open owns. Skipping it for a caller-owned graph would
-			// leave any tail the checkpoint captured permanently invisible in that graph, since nothing
-			// else will ever replay this store.
+			initStores(path, rootStore, databaseName, { defaultTable, destination, storeName, openedStores });
+			// a caller-owned graph is replayed into as well: nothing else will ever replay this store,
+			// so skipping it would leave the tail the checkpoint captured permanently invisible there
 			if (!isReadOnlyMode()) {
 				replayLogs(rootStore, destination ?? databases[databaseName]);
 			}
@@ -694,19 +692,24 @@ interface InitStoresOptions {
 	destination?: Tables;
 	/**
 	 * Identity stamped on the root store, when it must differ from the logical `databaseName` the
-	 * Table classes carry. `getRootBlobPathsForDB` resolves a database's blob directories from it, so
-	 * a branch sets it to the branch's own name to get its own blob roots while its tables keep the
-	 * name the application's code and schema already use.
+	 * Table classes carry. `getRootBlobPathsForDB` resolves blob directories from it.
 	 */
 	storeName?: string;
+	/**
+	 * Every column family opened here is appended, so a caller can release the ones a failure left
+	 * unreachable — a table's stores are opened well before `setTable` publishes it into the graph.
+	 */
+	openedStores?: any[];
 }
 
 function initStores(
 	path: string,
 	rootStore: RootDatabaseKind,
 	databaseName: string,
-	{ defaultTable, auditPath, isLegacy, destination, storeName }: InitStoresOptions = {}
+	{ defaultTable, auditPath, isLegacy, destination, storeName, openedStores }: InitStoresOptions = {}
 ) {
+	// a store with no tables never reaches the per-table loop below, and blob roots resolve from this
+	rootStore.databaseName = storeName ?? databaseName;
 	const envInit = new OpenEnvironmentObject(path, isReadOnlyMode());
 	const internalDbiInit = createOpenDBIObject(false);
 	let attributesDbi = rootStore.dbisDb;
@@ -908,7 +911,7 @@ function initStores(
 					rootStore
 				);
 			}
-			rootStore.databaseName = storeName ?? databaseName;
+			openedStores?.push(primaryStore);
 			primaryStore.tableId = tableId;
 		}
 		let attributesUpdated: boolean;
@@ -919,6 +922,7 @@ function initStores(
 				if (!attribute.isPrimaryKey && (attribute.indexed || (attribute.attribute && !attribute.name))) {
 					if (!indices[attribute.name]) {
 						const dbi = openIndex(attribute.key, rootStore, attribute);
+						openedStores?.push(dbi);
 						indices[attribute.name] = dbi;
 						indices[attribute.name].indexNulls = attribute.indexNulls;
 					}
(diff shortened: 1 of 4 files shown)


resources/databases.ts
```diff
@@ -993,12 +1015,105 @@ function initStores(
...
+			rocksdbDatabaseEnvs.delete(path);
+			// The column families opened for the tables and the audit store hold their own handles, and
+			// closing only the root leaves them behind.
+			for (const store of [(rootStore as any).dbisDb, (rootStore as any).auditStore]) {
+				try {
+					store?.close?.();
+				} catch (error) {
+					logger.warn?.(`Error closing branch column family at ${path}`, error);
+				}
+			}
```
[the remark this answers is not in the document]
kriszyp: Confirmed and fixed in `fc3eeab0f` — you're right, and this is the gap I'd flagged to myself as "I picked those two by reading `initStores`" without going back to check what else it opens.

Extracted `closeBranchHandles`, which walks the per-table `primaryStore` and `indices` the way `closeDatabase` does, then the internal-dbis and audit families, then the root. It is now shared with the partial-open cleanup, which previously closed the root alone.

One thing worth recording: **no test accompanies this.** I wrote one, and it passed with *and* without the fix — closing the root already makes the JS wrapper report not-open, so a leaked per-table handle is indistinguishable from a released one through any assertion available here. Rather than ship vacuous coverage I removed it. The justification for the change is symmetry with `closeDatabase`, which walks the same handles for the same native reason.

🤖 Reply by Claude (Opus 5) on behalf of Kris

resources/databases.ts
```diff
@@ -993,12 +1015,105 @@ function initStores(
...
+		// in the env map with no handle able to close it.
+		const stranded = rocksdbDatabaseEnvs.get(path);
+		rocksdbDatabaseEnvs.delete(path);
+		try {
+			stranded?.close();
+		} catch (closeError) {
+			logger.warn?.(`Error closing partially opened branch database at ${path}`, closeError);
+		}
+		throw error;
+	}
```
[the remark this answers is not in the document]
kriszyp: Fixed in `fc3eeab0f`. The partial-open cleanup now calls the same `closeBranchHandles` routine as `close()`, so a failure part-way through `initStores` releases exactly what a success would — the per-table families as well as internal-dbis and audit, not just the root.

claude[bot] converged on the same defect from the `close()` side, which is what prompted extracting one routine rather than patching two call sites.

Residual worth naming: a failure *inside* `initStores` before a table is installed into the destination still leaks that one table's handle, because the cleanup can only walk what was installed. Closing that needs handle tracking inside `initStores` itself; recorded in the PR description rather than bolted on here.

🤖 Reply by Claude (Opus 5) on behalf of Kris

resources/databases.ts
```diff
@@ -993,12 +1015,105 @@ function initStores(
...
+	// callers would disagree about who may close it.
+	if (openBranches.has(path)) throw new Error(`Branch database at ${path} is already open`);
+	// A globally-loaded database's store is shared and closed by `closeLoadedDatabases`; adopting it
+	// here would mean this handle's `close()` tears down a live database.
+	if (rocksdbDatabaseEnvs.has(path)) throw new Error(`Cannot branch ${path}: it is already open as a database`);
+	// `storeName` picks the blob roots. Letting it name a real database would point the branch's blob
+	// writes at that database's directory, which is the collision the separate identity exists to stop.
+	if (databases[storeName] || definedDatabases?.has(storeName)) {
+		throw new Error(`Cannot use '${storeName}' as a branch store identity: a database of that name exists`);
+	}
```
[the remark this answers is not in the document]
kriszyp: Good catch, fixed in `fc3eeab0f` — and the fix needed to go one step further than suggested.

Calling `getDatabases()` is right, but placing it where the collision check is would have introduced a worse bug: `getDatabases()` is itself what populates `rocksdbDatabaseEnvs`, so the path guard above it ("is this directory already open as a real database") would have run against an empty map and then been invalidated by the very scan that follows. A database the scan opened at this branch's path could have been adopted.

So `getDatabases()` now runs **before** all three guards, not just before the name check.

🤖 Reply by Claude (Opus 5) on behalf of Kris

resources/databases.ts
```diff
@@ -993,12 +1015,120 @@ function initStores(
...
+	// Load the registry FIRST: the guards below read it, and loading is itself what populates
+	// `rocksdbDatabaseEnvs` — running the path guard before the scan would let this branch adopt a
+	// database that the scan then opened at the same path.
+	getDatabases();
+	// A second open would hand back a rival table graph over one shared root store, and the two
+	// callers would disagree about who may close it.
+	if (openBranches.has(path)) throw new Error(`Branch database at ${path} is already open`);
+	// A globally-loaded database's store is shared and closed by `closeLoadedDatabases`; adopting it
+	// here would mean this handle's `close()` tears down a live database.
+	if (rocksdbDatabaseEnvs.has(path)) throw new Error(`Cannot branch ${path}: it is already open as a database`);
```
cb1kenobi, line 1119: **High: this guard is one-directional — the database scan can still adopt a branch.**

This blocks *branch-after-database*, but nothing blocks *database-after-branch*. `resetDatabases()` (copy_db, ITC restart) sets `loadedDatabases = false` and re-runs the `getDatabases()` scan; that scan opens any directory under the databases dir holding `CURRENT` + `MANIFEST-*`, finds the branch's store already in `rocksdbDatabaseEnvs`, and calls `initStores(...)` with **no** `destination` — building the branch's tables straight into the global map.

Reproduced against `fc3eeab0f` (branch dir placed under the databases dir, then `resetDatabases()` + `getDatabases()`):

```
branch rootStore.databaseName before rescan: probeB__x
databases has mybranchdir AFTER rescan:      true
branch rootStore.databaseName AFTER rescan:  mybranchdir   <-- branch identity clobbered
same store object adopted:                   true
after branch.close():  global table primaryStore.status=open  rootStore.status=closed
```

Three consequences. The branch becomes visible to exactly the enumerators this PR exists to hide it from (`storeDBSizeMetrics`, `schemaDescribe`, `closeLoadedDatabases`). `rootStore.databaseName` is overwritten with the directory name, so `getRootBlobPathsForDB` silently relocates the branch's blob roots off `storeName` — the identity split in item 1 of the description stops holding. And after `branch.close()` the global map is left holding an enumerable database over a closed root store; a read through it returned neither an error nor a row.

The directory placement is not hypothetical: harper#643 §3 specifies the branch path as `<storage_root>/database/${scope.appName}__${baseName}`, inside the directory this scan walks, so the first `resetDatabases()` after the scope wiring lands hits it. §6 of that ticket also asks to "verify there is no code path that auto-discovers databases on disk" — this scan is that path.

Suggested fix: make the exclusion structural instead of adding a second guard — either keep branch stores in a branch-only map (option 4 in the description, which removes this by construction), or have the scan skip any directory whose realpath is in `openBranches`.

—
_Generated by Barber AI_

resources/databases.ts
```diff
@@ -993,12 +1015,120 @@ function initStores(
...
+	for (const tableName in tables ?? {}) {
+		const table: any = (tables as any)[tableName];
+		if (!table?.primaryStore) continue;
+		for (const indexName in table.indices || {})
+			closeStore(table.indices[indexName], `index ${tableName}.${indexName}`);
+		closeStore(table.primaryStore, `table ${tableName}`);
+	}
+	closeStore((rootStore as any)?.dbisDb, 'attributes store');
+	closeStore((rootStore as any)?.auditStore, 'audit store');
+	closeStore(rootStore, 'root store');
```
cb1kenobi, line 1187: **Medium: a branch open registers a process-global reclamation handler that `close()` never removes.**

`initStores` → `openAuditStore(rootStore)` calls `onStorageReclamation(rootStore.path, handler)` (`server/storageReclamation.ts:104`), which pushes into the module-level `reclamationHandlers` map. There is no deregistration API and nothing here removes the entry, so the handler — and the closure holding the now-closed `rootStore` and its audit store — outlives the branch permanently, and every open/close cycle of the same path appends another.

Verified against `fc3eeab0f` by recording which paths `runReclamationHandlers()` queries:

```
branch path registered while open:        true
branch path STILL registered after close: true
```

Confidence note: the reclamation pass only invokes the handler under low disk, and `purgeLogs` on the closed store did not throw when I forced that condition, so what I actually observed is retention, not a crash. It is unbounded growth keyed to branch churn, which harper#643 makes routine (one branch per app load).

This is also the answer to the open question in the description ("If you know of a registry I did not find"). There is a second one: `resources/blob.ts:1861`, `databasePaths`, a strong `Map<RootDatabase, string[]>` populated on the first `getRootBlobPathsForDB(store)` and never deleted from anywhere in the tree — so a branch that touches blobs pins its store there as well. Neither registry is reachable through `Table.ts`, which is why searching there came up empty; both are reached from `initStores`.

Suggested fix: give `onStorageReclamation` a disposer (returning `{ remove() }` matches `addDeleteRemovalCallback` right next to it in `auditStore.ts`) and have `closeBranchHandles` call it along with `databasePaths.delete(rootStore)`. `closeDatabase` has the same gap and wants the same two calls.

—
_Generated by Barber AI_

unitTests/resources/databases.test.js
```diff
@@ -175,3 +175,108 @@ describe('dropDatabase restore serialization', () => {
...
+			() => openBranchDatabase(checkpointDir, 'branchbase', 'branchbase'),
+			/a database of that name exists/
+		);
+	});
+
+	it('tolerates a repeated close', function () {
+		const branch = openBranchDatabase(checkpointDir, 'branchbase', 'appA__branchbase');
+		branch.close();
+		assert.doesNotThrow(() => branch.close(), 'a second close must not close the store twice');
+	});
```
cb1kenobi, line 418: **Medium: the teardown tests pass with the teardown removed.**

I mutation-tested the new suite against `fc3eeab0f`, rebuilding `dist/` after each edit (`#src/*` resolves to `./dist/*.js`, so mutating the `.ts` alone is a no-op) and confirming exactly one marker landed in `dist/resources/databases.js`. Three separate mutations leave all 16 cases green:

1. delete the `closeBranchHandles(path, rootStore, tables)` call from `close()` outright — 16 passing
2. remove `if (!openBranches.delete(path)) return`, so a repeated close double-releases — 16 passing, this case included
3. skip the per-table column-family loop in `closeBranchHandles`, i.e. revert the fix `fc3eeab0f` was pushed for — 16 passing

So this case asserts only that a second `close()` does not throw, which the `try/catch` inside `closeStore` guarantees either way, and "closes what nothing else can" asserts only the two `Map.delete`s. Nothing in the suite observes a handle being released.

The description says the column-family test was written, found vacuous, and removed because "closing the root already makes the JS wrapper report not-open, so a leaked per-table handle is indistinguishable from a released one through any assertion available here." That does not hold — each table store reports its own status independently of the root. Under mutation 3:

```
root.status = closed    primaryStore.status = open
```

against unmutated `fc3eeab0f`:

```
root.status = closed    primaryStore.status = closed
```

Two assertions in this case kill all three mutations:

```suggestion
	it('tolerates a repeated close', function () {
		const branch = openBranchDatabase(checkpointDir, 'branchbase', 'appA__branchbase');
		const { primaryStore } = branch.tables.BranchSource;
		branch.close();
		assert.strictEqual(branch.rootStore.status, 'closed', 'close() must release the root store');
		assert.strictEqual(primaryStore.status, 'closed', 'close() must release each table column family');
		assert.doesNotThrow(() => branch.close(), 'a second close must not close the store twice');
	});
```

—
_Generated by Barber AI_

resources/databases.ts
```diff
@@ -993,12 +1015,120 @@ function initStores(
...
+}
+
+/** Close every open branch. Branches are process-local, so this is shutdown, not a data operation. */
+export function closeBranchDatabases(): void {
```
cb1kenobi, line 1193: **Medium: `closeBranchDatabases()` has no caller, including the worker-exit path that needs it.**

`closeLoadedDatabases`'s own docstring is the argument for wiring this now: rocksdb-js's registry is process-global across worker threads, and "a thread that exits WITHOUT closing leaks its handles (the process-global refCount never drops)". `server/jobs/jobProcess.ts:93` calls `closeLoadedDatabases()` on exit and nothing else, so a branch opened on a job worker is precisely the case that leaks process-wide — and by design `closeLoadedDatabases` cannot see it.

Leaving it unwired is the same shape as the skip list this PR removed: a rule the next person has to remember, in a place where forgetting is silent. Wiring it costs one line now and does nothing until a caller exists.

Suggested fix: call it from the existing exit path alongside `closeLoadedDatabases()`, or from inside `closeLoadedDatabases()` so there is a single teardown entry point.

—
_Generated by Barber AI_

resources/databases.ts
```diff
@@ -993,12 +1015,120 @@ function initStores(
...
+	getDatabases();
+	// A second open would hand back a rival table graph over one shared root store, and the two
+	// callers would disagree about who may close it.
+	if (openBranches.has(path)) throw new Error(`Branch database at ${path} is already open`);
+	// A globally-loaded database's store is shared and closed by `closeLoadedDatabases`; adopting it
+	// here would mean this handle's `close()` tears down a live database.
+	if (rocksdbDatabaseEnvs.has(path)) throw new Error(`Cannot branch ${path}: it is already open as a database`);
+	// `storeName` picks the blob roots. Letting it name a real database would point the branch's blob
+	// writes at that database's directory, which is the collision the separate identity exists to stop.
+	if (databases[storeName] || definedDatabases?.has(storeName)) {
```
cb1kenobi: **Low: `storeName` is unvalidated and lands in a filesystem path.**

Every other entry point that names a database validates against `schemaRegex` (`validation/common_validators.ts:5`, applied via `dataLayer/schema.ts`), which rejects `/` and backtick. `openBranchDatabase` applies only this collision check, and `storeName` then flows to `getBlobPathsForDatabaseName` → `join(getHdbBasePath(), 'blobs', storeName)`, so a value containing a separator or `..` resolves the branch's blob roots outside the blobs directory.

Static-only, and low risk under Harper's threat model: callers are admin-trusted and there is no caller yet. Worth closing here because harper#643 derives this name from deployed component config (`${appName}__${baseName}`) rather than from an operator typing it.

Suggested fix: run `storeName` (and `databaseName`) through `schemaRegex` next to the existing collision check.

—
_Generated by Barber AI_

resources/databases.ts
kriszyp: Confirmed and fixed in `d03802ec6` — the scan now skips any directory an open branch owns (`isOpenBranchPath`), at both RocksDB scan sites. A follow-up round found the guard set was still incomplete in a third place: the on-demand `database()` open reads the same `rocksdbDatabaseEnvs` entry a branch leaves registered for its lifetime, so `create_table`/`create_database` for a colliding name would have stapled the branch store onto `definedDatabases` and the next `closeDatabase` would have closed it under the live handle. Same guard, third site, fixed in `5cfc3af74`.

Your framing was right that this is a guard rather than a structural exclusion, and that distinction survives. I tried moving the ownership claim ahead of the `getDatabases()` inside the open — which closes the pre-open window — and reverted it: a claim taken before the scan also means a directory that genuinely *is* a real database stops reading as one, so `openBranchDatabase` would succeed over it and that database would vanish from the thread. The before-first-open and after-crash cases therefore remain, and the durable fix is an on-disk marker or a path outside the scanned root. Both are #643 §3 layout decisions rather than a local edit, so they are now item 1 of the PR description and the reasoning is recorded at the guard.

🤖 Reply by Claude (Opus 5) on behalf of Kris

resources/databases.ts
kriszyp: Confirmed and fixed in `d03802ec6`. `onStorageReclamation` gains a matching `removeStorageReclamation`, and `closeBranchHandles` now calls it for the branch path plus every store path it closed.

The second registry you pointed at — `resources/blob.ts`'s `databasePaths` — is real and is also released now (`databasePaths.delete(rootStore)`), with a test that fails when that line is removed. That answers the open question in the description: I had searched `Table.ts` and found nothing, and you are right that both are reached from `initStores`, not from `Table.ts`. The isolation claim stands, but not for the reason I had checked.

One thing I did not take: `closeDatabase` has the same two gaps for ordinary databases. Left alone as pre-existing and out of this diff's scope, and named in the PR description so it does not disappear.

🤖 Reply by Claude (Opus 5) on behalf of Kris

unitTests/resources/databases.test.js
kriszyp: You were right on both counts, and the correction matters more than the tests: `registryStatus().refCount` **does** distinguish a released column family from a leaked one, so my earlier "indistinguishable through any assertion available here" was wrong. The suite now asserts refCount, and all three of your mutations are killed.

Your mutation 2 turned out to be a live defect, not just a coverage gap. Writing an assertion that could kill it showed the re-close guard was keyed by *path* — and a closed branch frees its path, so a stale handle closed a second time deregistered whichever branch owned the path by then: dropped the live one out of `rocksdbDatabaseEnvs`, removed its reclamation handler, and left its RocksDB handles open with nothing tracking them. Fixed in `54d9bd2da` by keying the guard to the handle, with a test that opens, closes, reopens, and then re-closes the stale handle.

Thank you for rebuilding `dist/` and confirming exactly one marker landed — that is why the report was actionable rather than arguable.

🤖 Reply by Claude (Opus 5) on behalf of Kris

resources/databases.ts
kriszyp: Agreed — same shape as the skip list this PR removes, and wired in `d03802ec6`. `closeLoadedDatabases` now runs `closeBranchDatabases()` first, so the thread has one teardown entry point rather than a rule the next exit path has to remember. `unitTests/resources/closeLoadedDatabases.test.js` covers it by asserting the branch path's refCount reaches zero after `closeLoadedDatabases()`, which the walk over `databases` cannot achieve on its own.

🤖 Reply by Claude (Opus 5) on behalf of Kris

resources/databases.ts
kriszyp: Fixed in `d03802ec6`. Both `storeName` and `databaseName` now go through `schemaRegex`, plus the cases that regex permits but a single path component must not be: `\`, `.`, `..`, and empty. A later round added the length bound — `schemaRegex` is `*`-quantified with no length term, so a 300-character name passed and would have surfaced as `ENAMETOOLONG` from a blob write, far from the call that caused it; it is now held to the same 250-character cap as every other database name.

Your reason for closing it here rather than deferring is the one that decided it: #643 derives this name from deployed component config, not from an operator typing it.

🤖 Reply by Claude (Opus 5) on behalf of Kris

resources/databases.ts
cb1kenobi: **Verified fixed at `08e06743`.** I re-ran the original reproduction rather than reading the diff, with the branch directory at the harper#643 §3 placement (`<storage_root>/database/${appName}__${baseName}`), against both heads:

| observable after `resetDatabases()` | `fc3eeab0f` | `08e06743` |
| --- | --- | --- |
| adopted into `databases` | `probeb__reprobase` | none |
| `rootStore.databaseName` | `probeB__x` → `probeb__reprobase` | `probeB__x` (held) |
| `getRootBlobPathsForDB` | `…/blobs/probeb__reprobase` | `…/blobs/probeB__x` |
| enumerable in the global map | `true` | `false` |
| after `close()` | enumerable database over a closed root store; a read returned neither error nor row | not present |

The third site holds too: with a branch open, `database({ database: '<branch dir>' })` throws the 409 instead of stapling the branch store onto `definedDatabases`.

On the experiment you reverted — agreed with the call, and I confirmed the shape it settled into: the claim sits after `getDatabases()` and before `readRocksMetaDb`, so it closes the re-entrancy window without making a real database stop reading as one. The before-first-open and after-crash windows remain, and I'm not re-raising them: they're §3 layout decisions and you've recorded them in the description.

—
_Generated by Barber AI_

resources/databases.ts
cb1kenobi: **Verified fixed at `08e06743`.** Three open/close cycles on one path, recording which paths `runReclamationHandlers()` actually queries:

| | `fc3eeab0f` | `08e06743` |
| --- | --- | --- |
| branch path queried while open | 1 per cycle | 1 per cycle |
| branch path queried after `close()` | 1 per cycle | 0 |
| `databasePaths.size` after 3 cycles | 3 | 0 |

Both registries release. Agreed on leaving `closeDatabase`'s matching gaps out of this diff — they're pre-existing and named in the description.

—
_Generated by Barber AI_

unitTests/resources/databases.test.js
cb1kenobi: **Verified fixed at `08e06743`.** I re-ran all three mutations against the new suite, rebuilding `dist/` after each edit and confirming exactly one marker landed in `dist/resources/databases.js`:

| mutation | `fc3eeab0f` | `08e06743` |
| --- | --- | --- |
| 1. delete `closeBranchHandles(...)` from `close()` | 16 passing | **4 failing** |
| 2. remove the re-close idempotence guard | 16 passing | **1 failing** |
| 3. skip the per-column-family release loop | 16 passing | **2 failing** |

All three killed. Mutation 2 is caught by the new stale-handle case, which is the right observable for the path-keyed defect it turned out to be hiding — that one was a better outcome than the coverage gap I actually reported.

—
_Generated by Barber AI_

resources/databases.ts
cb1kenobi: **Verified fixed at `08e06743`.** `closeLoadedDatabases` calls `closeBranchDatabases()` first (`resources/databases.ts:1555`), and `server/jobs/jobProcess.ts:93-94` reaches it on worker exit — so the job-worker case now has a single teardown entry point. Confirmed by mutation: deleting the `closeBranchHandles(...)` call fails the new `closeLoadedDatabases` case along with three others.

—
_Generated by Barber AI_

resources/databases.ts
cb1kenobi: **Verified fixed at `08e06743`.** `assertLegalBranchName` runs both `databaseName` and `storeName` through `schemaRegex` plus the four cases the regex permits but a path component must not be, and the 250-character cap.

Worth noting the backslash case is not theoretical: `schemaRegex`'s `\x30-\x5F` range includes `\x5C`, so the explicit `name.includes('\\')` is what actually closes it on Windows. Good catch on the length bound — that one wasn't in my report.

—
_Generated by Barber AI_

cb1kenobi:
Re-reviewed `08e0674331ca8586d93c1a36e32ec4c531535a1b` — no issues found. This PR looks good, nice job!

**Scope.** Merge bases computed separately for both heads are identical (`f8a5aa90`), `fc3eeab0f` is an ancestor of `08e06743`, and there are no merge commits in the range — so no rebase and no evil merge. `git range-diff` shows the three original commits unchanged and five appended: `d03802ec6`, `54d9bd2da`, `5cfc3af74`, `0c05ade33`, `08e067433`. Genuine new work is 4 files, +339/−76.

**Prior findings.** All five re-checked against the new head by re-running the original reproductions and mutations rather than reading the diff — one High, three Medium, one Low, all confirmed fixed. Evidence is in each comment thread.

**New issues: none.** The one thing I chased and ruled out: `closeBranchHandles` now walks the `openedStores` list instead of the table graph, so I checked whether any column family can be opened for a branch *after* the open and escape it. With two tables and three indexed attributes, plus a post-open write introducing unknown attributes and an index-backed search, the registry refCount stays at 8 across the writes and reaches 0 on `close()`. The only `openIndex` call reachable for a branch is the one inside `initStores`; the other lives in `table()`, which resolves against the global schema and is the schema-mutation case already gated in the docstring. Also confirmed `handleLocalTimeForGets` returns the same object it is handed, so recording the handle before the wrapper (`08e067433`) is equivalent on the success path and strictly safer on the failure path.

**Tests.**

| suite | merge base `f8a5aa90` | head `08e06743` |
| --- | --- | --- |
| `unitTests/resources` | 1674 passing / 0 failing / 16 pending | 1693 passing / 0 failing / 16 pending |
| `unitTests/dataLayer` | 242 passing / 0 failing / 111 pending | 242 passing / 0 failing / 111 pending |

The 19 added cases are the whole `openBranchDatabase` suite (18) plus the new `closeLoadedDatabases` branch case. `dataLayer` is unchanged either side, which is the check I wanted on moving `rootStore.databaseName = storeName ?? databaseName` to the top of `initStores` — it now runs on every load rather than once per table.

—
_Generated by Barber AI_

Issue #2295, opened 2026-08-24
Two-phase deploy: package deploys are resolved per node, so the stage barrier cannot guarantee every node staged the same bytes


dawsontoth:
Split out of #1849, which adds two-phase (stage → activate) component deploys.

For a `package:` deploy the origin replicates the package *identifier*, not bytes. Confirmed in that
branch: peers wait on the deployment row with `requirePayload: !spec.package`
(`components/operations.js`), so a package deploy deliberately has no shared payload and every node
independently resolves and packs the reference.

## The failure

A mutable reference can resolve differently per node:

- a moving tag such as `latest`
- a semver range
- a git branch

Each node then stages *a* component and reports success, so the barrier passes and activation
proceeds — while the nodes are running different code. The barrier confirms "everyone staged
something", not "everyone staged the same thing".

This is silent: every stage succeeds, the deployment rows all look healthy, and nothing compares what
was produced.

DESIGN.md is currently accurate but narrow — it scopes the barrier's guarantee to "fetch + install,
not load" and makes no claim about byte equality. So this is a missing guarantee rather than a
contradiction of documented intent.

## What a fix requires

Resolve once, on the origin:

1. Origin resolves and packs the package, as it already does for its own stage.
2. Persist the exact tarball plus a digest on the deployment row.
3. Peers stage *those* bytes and verify the digest instead of resolving independently.

Two consequences worth deciding deliberately:

- It changes what the barrier guarantees — from "each node installed something" to "every node
  installed these bytes".
- It makes a `package:` deploy carry a payload it currently does not, with the storage and payload
  retention cost that implies (`deployment_payloadRetention_maxCount` then applies to package
  deploys too).

## Notes

- A pinned exact version resolves deterministically, so the divergence needs a mutable reference —
  that is what keeps this contained relative to #2294.
- Exposure today is zero: the two-phase protocol is unshipped.

dawsontoth:
The protocol this applies to is now split out as #2301 (stacked on #1849). Byte-identical package resolution is listed there as one of the things that must be resolved before it leaves draft.

Issue #2294, opened 2026-08-24
Two-phase deploy: concurrent activations from different origins can leave the cluster on different versions with both reporting success


dawsontoth:
Split out of #1849, which adds two-phase (stage → activate) component deploys. Verified still
present in that branch: the activate fan-out runs *after* the origin releases its node-local
component preparation lock (`components/operations.js` — `activateStagedApplication` completes and
releases, then `replicateOperation(buildPhaseOperation('activate', …))`), so nothing establishes an
order between two deploys originated on different nodes.

## The failure

If node A originates deployment D1 and node B originates D2 for the same project at the same time,
each node can apply the other's fan-out last:

- A ends with D2 live
- B ends with D1 live
- **both deployments report success**

The stage barrier does not help. It orders stage-before-activate; it says nothing about
deploy-against-deploy.

The cluster is then silently running two different versions of the same component, and every
observable signal — the operation response, the deployment rows — says both deploys succeeded.
Nothing converges it; it persists until the next deploy of that project happens to land in a
consistent order.

## What a fix requires

A cluster-wide per-project order for activation. Two shapes discussed:

- **Leader serialization** — route activation through a single node per project.
- **Monotonic activation epoch** — persist a per-project counter, compare it under the component
  preparation lock during activate, and reject an activation whose epoch is stale.

The epoch option needs a home for the counter that survives restarts. Either is an addition to the
protocol rather than a local fix, which is why it was split out instead of being patched into #1849.

## Notes

- Raised in review of #1849 as that round's only High.
- Exposure today is zero: the two-phase protocol is unshipped. This becomes live the moment #1849
  merges, which is why it is milestoned alongside it.
- `ignore_replication_errors` already opts out of the barrier, so any ordering mechanism has to
  define its behavior for that path too.

dawsontoth:
Companion: #2295 (same protocol, byte-equality rather than ordering). Both split out of #1849.

dawsontoth:
The protocol this depends on is now split out as #2301 (stacked on #1849), which carries the barrier and the trusted peer phase operation. This ordering gap is listed there as one of the things that must be resolved before it leaves draft.

PR #2301, opened 2026-08-24
feat(deploy): cluster-wide stage/activate barrier and trusted peer phase operation

dawsontoth:
> **Draft.** Stacked on #1849 — merge that first. The diff here is exactly the cluster coordination protocol, split out of #1849 so it can be reviewed on its own and land with real multi-node verification.

## Why this is a separate PR

#1849 originally carried both halves: the per-node staged deploy, and the cluster-wide protocol that coordinates it. Across four review rounds every escaped defect traced to the second half, and twice a unit test passed against genuinely broken behavior — once because `replicateOperation` was stubbed, once because a payload was present but not replayable. That is not a coincidence: the unit harness mocks replication and enters the authorization bypass directly, so **it cannot observe a peer at all.**

So the halves were split. #1849 keeps everything a single node can prove. This PR carries the part that needs a cluster to prove, and should not merge on unit evidence alone.

## What this adds on top of #1849

- **`component_deploy_phase`** — a trusted peer-only operation carrying the phase and deployment id. Authorization travels in AsyncLocalStorage rather than on the request, so it is unreachable over HTTP with ordinary credentials, and an older peer rejects an unknown operation instead of misreading a phase marker as a one-shot deploy.
- **The cluster-wide barrier** — the origin stages everywhere and waits for every node to report a good stage before any node activates. A node that cannot fetch or `npm install` fails during staging while the live component is untouched *on every node*.
- **The separated public phases** — `activate: false` returns a staged `deployment_id`; `deployment_id` activates it later; `two_phase: false` forces single-phase. Plus the `harper stage` / `harper activate` CLI verbs and the capability probe that stops a staged request reaching a server that would silently deploy it live.
- **The deployment row as the peer channel** — the immutable activation specification, the payload blob peers read their bytes from, and staged-row retention.
- **Origin-owned row semantics** — peers claim with `persist: false` so the replicated row has exactly one writer; their crash evidence is local activation artifacts, not the row. Without this, a peer's late `activating` could land after the origin wrote `success` and park a converged deploy in a non-terminal status forever.

## What must happen before this leaves draft

1. **The three-node harper-pro suite has to run against this exact core revision.** There is no in-repo end-to-end coverage of the trusted dispatch path — replication serialization, peer authorization context, peer-result wiring, and `persist: false` crash recovery are all unverified here. This is the whole reason for the split.
2. **#2294** — nothing orders two concurrently-originated deploys, so node A can end on D2 and node B on D1 with **both reporting success**. The barrier orders stage-before-activate, not deploy-against-deploy. Needs a cluster-wide per-project order (leader serialization, or a monotonic activation epoch checked under the component lock).
3. **#2295** — a `package:` deploy resolves per node, so the barrier guarantees "everyone staged something", not "everyone staged the same bytes". A moving `latest`, a semver range, or a git branch can resolve differently per node while every stage reports success.

## Review history

The code here has been through several cross-model rounds while it lived on #1849, and the fixes are in its commit history: revert compensation keyed on commit *attempt* rather than success, activation evidence classified under the component lock, the origin-owns-the-row change, retention protecting anything newer than the returning request, and the one-shot payload actually being replayable. Those rounds are also what surfaced the coverage gap above — they kept finding real defects that the tests could not.


dawsontoth:
## Reference only — this PR will not merge, and will not be folded into #1849

We're re-planning this work as a sequence of small, independently mergeable PRs. **#2315** is the coordination issue.

The original plan was to merge this branch back into #1849 so there'd be one reference. **That isn't possible any more, and it shouldn't be forced:**

- #1849's tip deleted `claimStagedDeployment` and `expireOldStagedDeployments` (they were dead there — this PR was their only consumer). This branch's `components/operations.js` imports and calls `claimStagedDeployment` at two sites, so a merged tree would import a function that no longer exists.
- #1849's tip also makes `activate`, `two_phase`, and `deployment_id` `Joi.forbidden()` — which is exactly what this PR exists to accept.

Merging would mean hand-reconciling a design we've decided not to ship. So both branches stay as **separate** references instead, each internally coherent:

- **#1849** — per-node staged deploy: staging, activation, crash recovery, `revert_component`, retention, CLI, DESIGN.md.
- **#2301** (this one) — cluster coordination: stage/activate barrier, trusted peer operation, separated phases.

### Where this design went

Step 5 in #2315 supersedes the barrier. `main` **already** does build-once-replicate-the-artifact for payload deploys — the origin tees the payload through a hash-and-size tap into the deployment row's `payload_blob`, and peers install from the replicated row. Extending that to `package:` and `ref:` deploys (#2295) gets bit-identical bytes across the cluster without a barrier at all, which is the goal this PR was reaching for by a heavier route.

What the barrier uniquely solved — ordering two concurrently-originated deploys — remains open as **#2294**, and is the one piece that genuinely needs multi-node test infrastructure we don't have. That's why it's tracked as an issue rather than queued as a PR.

The code here stays available to lift from; it went through several rounds of cross-model review.

Reviewers: no action needed. Follow #2315.

dawsontoth:
Closing as reference-only — see #2315 for the sequence of small PRs that replaces this work.

The branch is **not** deleted: its diff and history stay browsable here, and the individual steps in #2315 will lift already-reviewed code from it.

PR #2302, opened 2026-08-24
Operations API resolves @relationship attributes instead of returning null or rejecting them

### LINKED ISSUE #2296: Operations API returns `null` for @relationship attributes (or rejects them): the thread serving it never loads the component schema
## Summary

> **Confirmed by a local repro on a clean, freshly-installed single node**, on both Harper **5.1.15** and **5.2.4** (the latest release), with identical results — see the repro comment on this issue. No restart, no stale state, and no multi-node setup is required to hit this; both relationship directions and both `get_attributes` forms are affected.

The operations API cannot resolve `@relationship` attributes, because the thread that serves it never loads the component's GraphQL schema. Depending on what stale attribute metadata a node happens to hold, the same query either **fails loudly** with `unknown attribute` or — much worse — **silently returns `null`** for the relationship while reporting HTTP 200.

The same relationship, on the same node, at the same moment, resolves correctly over the component/REST surface. So this is specific to the operations API, not to the resolver.

## Observed

Two nodes of one cluster, same version (5.2.0), same component, same records:

| surface | thread | node A | node B |
|---|---|---|---|
| component REST port | `http/N` (loads component) | resolves correctly | resolves correctly |
| operations API | `main/0` (no component) | `unknown attribute 'location'` (HTTP 400) | `"location": null` (HTTP 200) |

`describe_table` shows the divergence directly — node A lists 32 attributes, node B lists 34, and the delta is exactly the two `@relationship` attributes. On node B they carry schema-derived types (`"type": "Location"`, `"type": "array"`), so that node's list came from the GraphQL schema, not from storage.

**There is no operations-API form that works.** The object form (`{"name":"location","select":["cloudProvider"]}`) skips the unknown-attribute check in `validation/searchValidator.ts` — so it passes validation and then returns `null` anyway.

A full process restart of node A did not restore the attributes; it came back with the identical 32. Before the restart, this same operations-API query had been resolving relationships correctly for months, which suggests the working state was inherited in-memory schema that a restart does not rebuild.

## Mechanism

1. `components/componentLoader.ts` gates component loading to non-main threads:
   ```js
   if ((config.extensionModule || config.pluginModule) && (!isMainThread || config.runOnMainThread)) {
   ```
   So the main thread never runs `table()` against the component's GraphQL schema.

2. `resources/databases.ts` deliberately never persists relationship attributes to `attributesDbi`:
   ```js
   if (attribute.relationship || attribute.computed) {
       hasChanges = true;
       if (attribute.relationship) continue;
   }
   ```
   They are runtime-only state, present solely in a thread that loaded the schema.

3. The operations API is served from the main thread (errors log as `[main/0]`). Its attribute list is rebuilt from `attributesDbi`, which by (2) can never contain relationships.

4. `validation/searchValidator.ts` rejects any `get_attributes` entry absent from the table's attribute list, producing `unknown attribute '<rel>'`. The object/nested form skips that check, so it passes validation and yields `null` instead — **this needs no stale state and reproduces on a fresh install**. Separately, on a node whose attribute list does contain the relationship, the bare form also passes validation and returns `null`; see the note on the two-node divergence below.

## Why this is worth a tier above its blast radius

The `null` path reports success. A caller cannot distinguish "this host has no location" from "this node cannot resolve relationships at all". In our case it surfaced only because a downstream consumer happened to call `.get()` on the null and crashed — anything that merely stores or forwards the value would have propagated wrong data indefinitely with nothing in the logs. Nothing is written to the log at load time, because the relationship is never refused; it is simply never registered.

## Versions

Observed on 5.2.0 in a deployment, and reproduced from scratch locally on **5.1.15** and **5.2.4** with identical results — so the whole 5.1.x-5.2.x range is affected, latest release included. Not fixed as of current `main` (past 5.2.4), verified by reading:

- the `componentLoader` main-thread gate — present and unchanged (introduced 2026-03-17)
- the `databases.ts` relationship-persistence exclusion — present and unchanged (introduced 2025-10-10)
- `validation/searchValidator.ts` — unchanged since 5.2.0, `unknown attribute` check still in place

No commit between 5.2.0 and current `main` touches relationship resolution on this path.

## Repro

1. Deploy a component whose GraphQL schema declares an `@relationship` attribute.
2. Query it over the component REST port with `select(rel{field})` — resolves.
3. Query the same attribute via the operations API `search_by_conditions` with `get_attributes: ["rel"]` — returns `unknown attribute`, or `null` if a stale attribute entry survives.
4. Restart the node and repeat step 3 — the attribute is now absent from `describe_table` entirely.

## The question behind the bug

Is the operations API *intended* to resolve `@relationship` attributes at all, given the thread serving it never loads the component?

- **If yes**, this is a straightforward bug: that surface needs access to schema-derived attributes.
- **If no**, the surface should reject them **consistently**, so callers get a loud error rather than a silent `null` on whichever nodes happen to retain stale metadata.

Either resolution is fine from our side. The silent `null` is the part that should not survive either way.

kriszyp:
The operations API can now resolve `@relationship` attributes. It is served from the main thread, which never loads a component's GraphQL schema, and relationship attributes were [runtime-only state](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-72e8b09df8fa4bd777558edfcb08b247ac3965b75d83c9c8c6c7f81cbe2374c8R316), so a relationship was either rejected as an unknown attribute (bare `get_attributes`) or silently returned as `null` (object/nested form) while reporting HTTP 200 — on a fresh install, in both directions, on every 5.1.x–5.2.x release. The GraphQL authoring path now persists a [normalized, data-only relationship list](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R122) on each table's [primary catalog descriptor](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1814), and [`getDatabases()` hydrates it](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R641) into working resolvers on whatever thread loads the catalog. [`describe_table` lists the relationships too](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-fd0b483c1861b3153e655c0e9a7b99b9d5a6db0022d023fe844daabe653829a8R206).

One thing the issue asked for is deliberately not done: the object/nested `get_attributes` form still returns `null` rather than an error for a name this thread cannot resolve. I implemented that rejection and then reverted it — `Table.transformEntryForSelect` projects undeclared JSON sub-objects through that same form, so an unknown-name check turns a supported nested projection into a 400 on any unsealed or dynamic table. The reported silent `null` disappears because relationships resolve, not because the request is refused; a relationship that cannot be hydrated (missing target table, code-first `defineTable` relation) [logs an error at load time](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R696) and is still absent from the attribute list, so the bare form 400s.

Closes HarperFast/harper#2296

## For the human reviewer

1. **`primary-descriptor-catalog` — relationship metadata lives on the table's PRIMARY catalog descriptor, not in per-attribute rows.** The [row it writes](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R2211) is [the primary key's own row, or the bare table row when there is no declared primary key](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R2252); a per-attribute row is loaded as a live schema attribute by every reader, including ones that cannot resolve it — which is exactly the silent `null` this PR removes, and an older Harper reading the same store would reintroduce it. The primary descriptor is a field they ignore, and `table()` preserves unknown fields when it rewrites the row. The cost is one blob rewritten as a unit rather than a row per relationship. Reversing this after release means an on-disk shape change.
2. **`explicit-only-hydration` — [`@enumerable` is deliberately not persisted](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R122).** Operations queries traverse a relationship only when it is explicitly selected, so `get_attributes: ["*"]` responses are unchanged. Saying no here buys default-serialization parity with the component REST surface, at the cost of silently enlarging existing operations-API responses wherever a component declares enumerable relations.
3. **`runtime-name-precedence` — [a live/stored attribute of the same name always wins over the catalog copy, silently](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R688).** Data that exists beats a relationship that is only declared, and the collision is normal on component workers (the live schema attribute is the same declaration the catalog list came from), so warning would fire on the healthy path. The consequence is that an `add_attribute` named after a relationship makes it unresolvable on the operations thread with no breadcrumb.
4. **`graphql-only-authority` — only [the GraphQL path](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-72e8b09df8fa4bd777558edfcb08b247ac3965b75d83c9c8c6c7f81cbe2374c8R115) [authors and erases the list](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1750).** `defineTable()` resolves relation targets through a lazy thunk so forward references and cycles work, so the target is not knowable at registration; code-first relations stay worker-only. Admin and replication callers omit the flag, so they cannot erase component-owned metadata — which is also why an empty list from GraphQL is authoritative removal.
5. **`per-table-persistence` — [the reconcile is per table](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R2201), not per schema graph.** A multi-table relationship graph is committed one primary descriptor at a time, so a crash mid-reload can leave a partial graph; hydration degrades safely (a relationship whose target is not yet present is skipped and logged, and re-hydrates on the next load). An atomic alternative would be cross-cutting.
6. **`skip-invalid-catalog` — malformed metadata is [logged once and skipped](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1195), per list and [per definition](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R724), and boot continues.** A corrupt list reconciles to no catalog relationships rather than failing startup or installing a resolver whose target is unknown.

Known limitation, deliberately not fixed here: a hydrated relationship reaches a thread only through the paths that already reload the catalog (`getDatabases()` at boot, and the `schema-change` ITC event a persisting `table()` call broadcasts). Raw catalog-field replication with no schema signal would not refresh a running process — the same as every other field on that descriptor.

## Verification

Route: extended integration test (`integrationTests/apiTests/graphql.test.mjs`), unit coverage, and a fails-on-base check.

- `npm run test:integration -- "integrationTests/apiTests/graphql.test.mjs"` — 27/27 pass. [Three are new](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-fd0b483c1861b3153e655c0e9a7b99b9d5a6db0022d023fe844daabe653829a8R158): forward + reverse resolution over the operations API in the object form, the same in the bare form, and `describe_table` listing the relationships.
- `npx mocha unitTests/resources/schemaMigrationFragility.test.js` — 22/22. [Eleven are new](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-c5b03a2bbd1cd52b67fa70b5b8a61dd6016399d7bfaa6de978479964ab4f7c2bR504): catalog round-trip (including `filterMissing`), resolver rebuild from catalog-only state, steady-state identity (an unchanged catalog rebuilds nothing), related-table permissions on a hydrated resolver, non-authoring updates not erasing the list, [drop-tombstone refusal](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R2246) on both the modern and the legacy named-descriptor catalog shape, missing target skipped, authoritative removal, a hydrated resolver re-pointing at a dropped-and-recreated target, and a table with no declared primary key both persisting and later clearing its list. The legacy-descriptor test was confirmed to fail with [the two-row tombstone check](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R1435) reverted.
- Fails-on-base at `origin/main` (temporary worktree at the merge base, test-only diff applied): all three new integration tests fail there (24/27), and 8 of the 10 new unit tests fail there. The two that pass on base are the drop-tombstone refusal — it guards a hazard this PR's descriptor write introduces rather than reproducing the reported bug — and the steady-state identity check, which is vacuous when there is nothing hydrated to compare.
- `HARPER_STORAGE_ENGINE=lmdb npx mocha unitTests/resources/dropTableGhost.test.js` — 11/11, covering [the LMDB arm](https://github.com/HarperFast/harper/pull/2302/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R1445) of the restructured `dropTable()` tombstone write. The new catalog suite itself stays RocksDB-only, like every other suite in that file.
- `npm run typecheck`, `npm run lint`, `npx prettier --check` — clean.
- Reviewed over four pre-push rounds; the coverage footer names the final round's lens only. Earlier rounds also ran Gemini, Cursor Grok, Cursor Composer and the Harper domain adjudicator, and each round's findings were acted on — the validator revert, the primary-key decoupling, the two-row tombstone check, the pre-write ordering fix, and the name/reason-scoped error keys all came out of them.
- Not covered, and worth knowing: an operations-API permission-denial test was attempted and dropped because it could not be made meaningful in that fixture. The role attaches correctly (`super_user: false`, `Related.read: false`), but a direct `search_by_conditions` without `schema` returns 200 with data, and the same request with `schema: "data"` returns `403 Table 'data.Related' does not exist` — so a role's table entries do not appear to bind to component-defined tables on that surface at all. Pre-existing and independent of this change; permission enforcement on a hydrated resolver is proven at the unit level through `RequestTarget` + `authorize: true`.
- `npm run test:unit:resources` — 1694 passing, 3 failing; the same three fail on `origin/main` in the same worktree recipe (`randomAccessFields` ×2 and `replayStructures`, which base runs as 1683 passing / 4 failing — its extra failure is a `Caching` case that flakes). `npm run test:unit:main` cannot run on this machine at all, on `origin/main` too: `~/.harperdb/hdb_boot_properties.file` points at a stale LMDB install, so table loading throws before any test executes. Neither gate is affected by this change; CI is authoritative for both.

Complexity: complicated

<sub>Review-Coverage: authored=claude; ran=codex; blocked=gemini(no-output); declined=cursor-grok,cursor-composer,domain; rounds=8 @ f74781c35165</sub>

<sub>Human-Review-Need: 4 @ f74781c35165</sub>


DavidCockerill (approved):
Approving. No findings.

Persisting schema-derived state onto the catalog creates a second copy of something GraphQL already owns, which is the classic setup for two definitions drifting apart. Three rules prevent that, and they are the right three:

- **The persisted list is subordinate, not parallel.** *"Live schema relationships and same-name runtime attributes always win."* On a thread that has the schema the catalog copy is never consulted, so this is a fallback for threads that cannot load one rather than a competing source of truth. That is what keeps it from being two definitions.
- **Only the authoring path can write it.** Authoritative only when the GraphQL path explicitly supplies `schemaRelationshipsDefined`, so admin and replication callers that omit it cannot erase component-owned metadata. Without that sentinel any unrelated catalog write would have silently blanked the relationships — a partial writer clobbering state it knows nothing about.
- **`@enumerable` is deliberately not persisted**, so operations queries traverse relationships only when explicitly selected. Otherwise this fix would have quietly added relationship traversal to every read that happens to touch the table.

An empty list removing catalog-owned relationships is the correct reading of an explicit empty — it means "the schema defines none", which is different from "nobody said".

**On test coverage: I went looking for a gap and did not find one.** I expected the precedence rules to be unpinned; they are all covered — `does not let a non-authoring attribute update erase relationship definitions`, `authoritatively removes relationship definitions when the GraphQL schema drops them`, `ignores a persisted relationship whose target table is unavailable`, `re-points a hydrated resolver at a recreated target table`, and `leaves hydrated attributes untouched when nothing about the catalog changed`. `applies related-table permissions to a hydrated resolver` is the one I would not have thought to ask for and the one that matters most — a resolver rebuilt from the catalog bypassing related-table permissions would be a data-exposure bug rather than a correctness one. Naming the file after the fragility rather than the feature tends to mean someone thought about how it breaks.

The split is also right: integration tests cover the bug (both original failure modes plus `describe_table`), unit tests cover the invariants. The only thing not directly pinned is the cross-thread precedence — that a thread which *has* the schema ignores the catalog copy — which is the pre-existing path and exercised implicitly. A single case would cover it; entirely optional.

**Hydration also fails soft in the right place**: each table is wrapped individually with a dedup-keyed error, so one malformed descriptor cannot break catalog load for a whole database. And waiting for the complete catalog before rebuilding resolvers is the ordering that matters, since a relationship's target may load after the table pointing at it.

**One migration property worth a release note** if it is not already there: existing tables carry no persisted list until the authoring path runs again, so operations-API relationship resolution appears when a database's schema is next applied rather than on upgrade. Benign — the feature is absent rather than wrong — but it means "upgraded and still cannot see relationships" is expected rather than a bug report.

kriszyp:
## Rebased onto latest `main`

Resolved one real conflict in `resources/databases.ts` `table()`: an unrelated upstream refactor (wrapping the function body in `try`/`finally`, adding `attributesToIndex`/`indicesToRemove`) landed in the same region as this PR's `refreshRelationshipAttributes` tracking and the `relationshipDefinitions` persistence onto the primary catalog descriptor. Verified the resolution by diffing the merged result against the union of both sides' independent changes — no content lost or duplicated. Build and the full unit + integration suites pass.

### Additional fixes from the pre-push cross-model review

The rebase triggered a fresh review round (prior rounds weren't ancestors of the new head). Fixed what came back real:

- **`resources/Table.ts`** — `dropTable()`'s RocksDB tombstone write reimplemented lock acquisition as a bare `while (!rootStore.tryLock('update-attributes')) {}` — no backoff, no timeout, unlike `withUpdateAttributesLock` (same lock key) which already provides both. Swapped in the existing helper.
- **`resources/databases.ts`** — `hydrateCatalogRelationships()`'s catch block discarded the actual error/stack, logging only a generic message. Now passed through to `logger.error`.

### Investigated and did not act on

Domain adjudication timed out every round on this branch's budget, so I adjudicated the outside findings myself:

- **Gemini raised two "blocker" claims** (`Table.ts` tombstone write racing an unresolved LMDB `put()` promise; `databases.ts`'s `primaryDescriptorKey()`/relationship block reading a not-yet-visible LMDB write within the same `table()` call). Both rest on the assumption that an LMDB write inside `transactionSync` (or within `table()`'s already-open `exclusiveLock()` transaction) isn't visible to a synchronous `getSync()` until its returned promise resolves. I reproduced this directly against the `lmdb` package used here — a write made in that setting **is** synchronously visible before the promise resolves, matching `transactionSync`'s documented "the result will be committed" behavior and the existing `RecordEncoder.ts`/`exclusiveLock()` precedent in this codebase. Repro available on request if useful for the record.
- **Major**: the `withUpdateAttributesLock` fix above still blocks the event loop up to 10s under contention — true, but that's the pre-existing, deliberate design of this lock (see the comment on `acquireUpdateAttributesLock` in `Table.ts`, already used the same way for schema updates), not a regression from this change.
- **Significant**: `databases.ts` catalog-reconciliation de-indexing uses `Table.indices[attributeTableName]` (the table-name segment of the key) instead of the attribute name — looks like a real stale-index bug, but it's present on `main` before this PR too, unrelated to `@relationship` persistence. Filing separately rather than folding an unrelated fix into this diff.
- **Significant**: `refreshRelationshipAttributes` triggers a full resolver rebuild/`updateTable` emit on every reload of a table with any `@relationship` attribute, even when nothing changed (it doesn't compare against the already-equality-checked persisted list). Real, but a design tradeoff already made in this PR's earlier reviewed commits — leaving for a maintainer call rather than redesigning under this pass.
- **Minor**: `reportedRelationshipErrors` never clears `:invalid`/`:hydrate` entries (only `:unavailable` does, from an earlier adjudicated round) — a fixed-then-re-broken relationship under the same name won't re-log. Narrow, pre-existing scope decision.
- **Suggestions**: the new relationship-catalog test suite (`schemaMigrationFragility.test.js`) skips entirely under `HARPER_STORAGE_ENGINE=lmdb`, so the LMDB locking path is untested by these new tests (my repro above covers the specific race claims, but there's no standing regression test); `relationshipListsEqual` compares by position rather than name.

Noted the pre-existing stale-index bug and the LMDB test-coverage gap as findings for follow-up, separate from this PR.

— Claude Sonnet 5 (dev-agent, rebase task)

PR #2256, opened 2026-08-21
Canary reliable AI review evidence in Harper

heskew:
Canary the reliable review-evidence plumbing from HarperFast/ai-review-prompts#86 in Harper by pinning the Claude and Gemini review callers to its verified merge commit. Gemini completed the live canary, while Claude correctly declined this self-modifying caller-workflow change.

## For the human reviewer

No open judgment calls — this is a paired, immutable pin bump to the already-reviewed #86 merge commit; mention, issue-to-PR, permissions, triggers, and secrets are unchanged. The validator caller now moves to the same pin (e0b9b96, from kriszyp's review): the #86 delta changes the workflow contract the validator checks callers against, so the pins move together — the same bump is applied on harper-pro#748 and oauth#210.

## Verification

- Preflight: `git diff --check` passed.
- Preflight: `yq eval '.'` parsed both changed workflow files successfully.
- Provenance: GitHub reports `4b59dc0ddb15aff517884127b58204f79193b4f2` as the signed, verified merge commit for HarperFast/ai-review-prompts#86.
- End-to-end route (c), Gemini live smoke: [run 32437905419](https://github.com/HarperFast/harper/actions/runs/32437905419) loaded `_gemini-review.yml@4b59dc0…`, captured the prior-thread context snapshot, and posted the exact provider marker plus run/attempt/head marker. It returned no blockers and correctly used `review:no-log` for this mechanical diff, so no ai-review-log issue was created.
- Claude guard-path smoke: [run 32438007568](https://github.com/HarperFast/harper/actions/runs/32438007568) loaded `_claude-review.yml@4b59dc0…` and captured the prior-thread context snapshot. Claude then declined the caller-workflow change as expected; #86 correctly classified the attempt as `invalid-review-status` and did not log or post it as a verdict.

## Review coverage

Authored by Codex. Cross-model review: Claude Fable 5 ✓ (independent graded review; no findings). Gemini, Cursor, and Harper-domain adjudication were pruned by the minimal low-risk policy. Receipt @ b7cedc20acaf4e254ba89c521d7a4ffc38547a2f.

<sub>Human-Review-Need: 2 @ b7cedc20acaf</sub>

Issue #86, opened 2026-01-13
Fix: CLI not displaying app when using `harper run .`


BboyAkers:
### Bug Summary

When running `harper run .`  the app i'm running it in doesn't show up in the terminal

### Reproduction Repository (if applicable)

https://github.com/BboyAkers/harper-invoice-app/

### Steps to Reproduce

1. Clone the app
2. Run `npm install` inside the app
3. Run `npm run dev`
4. The app starts but in the terminal it doesn't display the current app and the port it's linked to.

### Expected Behavior

1. Clone the app
2. Run `npm install` inside the app
3. Run `npm run dev`
4. Should display the app name and the port it's under in the terminal

### Actual Behavior

1. Clone the app
2. Run `npm install` inside the app
3. Run `npm run dev`
4. The app starts but in the terminal it doesn't display the current app and the port it's linked to.

### Are you planning to fix this issue?

Maybe, depending on complexity

### First-time contributor support

- [ ] I'm new to contributing and would appreciate guidance on the process
- [ ] I'd like help understanding the project structure
- [ ] I need assistance with setting up the development environment
- [ ] I'm comfortable contributing, but new to this project specifically

PR #2308, opened 2026-08-25
chore(ci): match main's ai-review-prompts pin on v5.2 so the review gate can run

kriszyp:
Restores AI review coverage on the `v5.2` release branch, which has had none since the branch was cut.

## The defect

The Claude GitHub App only exchanges its token when the caller workflow file on the PR's head is **byte-identical** to the copy on the repository's default branch. `v5.2` still pins `ai-review-prompts@224c2ad` (main 2026-07-28) while `main` moved to `@4b59dc0` in [Canary reliable AI review evidence in Harper](https://github.com/HarperFast/harper/pull/2256) on 2026-08-21 — after the `v5.2` cut, and the release branches weren't included in that rollout.

So **every** PR targeting `v5.2` fails token exchange and gets no Claude review. From the `review / review` job on [#2307](https://github.com/HarperFast/harper/pull/2307) ([run 32804942131](https://github.com/HarperFast/harper/actions/runs/32804942131)):

```
App token exchange failed: 401 Unauthorized - Workflow validation failed.
The workflow file must exist and have identical content to the version on
the repository's default branch.
```

Two retries, same 401, then the job failed at 41s having posted nothing — so the red check isn't a verdict, it's the gate never running. The `Log review to ai-review-log` step then recorded `No marker'd review surface found (review_status=failure); skipping log`, meaning these non-reviews aren't showing up in calibration data either.

That check failure was the only non-green one on #2307 (all 30 others — build, format, integration 1–6 across Node/Bun/Windows/uWS — passed), which is what surfaced this.

## The change

All three ai-review-prompts callers move to `4b59dc0` together, exactly as #2256 did on `main`: the #86 delta changes the workflow contract the validator checks callers against, so `claude-review`, `gemini-review`, and `validate-caller-workflows` share one pin. Six lines, all of them the pin (each file carries the SHA twice — once in `uses:`, once in `ai-review-prompts-ref:`, since a reusable workflow can't introspect its own ref).

No other change: triggers, `permissions:`, secrets, `if:` conditions, review layers, and the Sonnet 5 model canary are untouched.

## Verification

- **Byte-identity, which is the actual fix.** Each of the three files now hashes identically to `origin/main`'s copy (`sha256`, verified per file), so the App's validation compares equal rather than approximately equal.
- **Pin provenance.** `4b59dc0ddb15aff517884127b58204f79193b4f2` is GitHub-reported as **verified** (signed) and is the merge commit of [Make AI review calibration evidence reliable](https://github.com/HarperFast/ai-review-prompts/pull/86), committed 2026-08-21 — the same commit `main`, harper-pro#748, and oauth#210 already run.
- **YAML parses** for all three files; `git diff --check` clean.
- **This PR is its own smoke test.** Its head carries the new content, which now matches `main`, so its own `review / review` check exercises exactly the path that 401'd on #2307 — if the gate runs here, the fix holds. (Note that Claude correctly *declines* self-modifying caller-workflow diffs, as it did on #2256; the signal to look for is token exchange succeeding and the job reaching a decline, not a verdict.)
- **Scope check.** `git diff origin/main origin/v5.2 -- .github/workflows/` reports these three files and nothing else, so this closes the entire workflow drift between the branches. harper-pro's `v5.2` has no ai-review-prompts drift — this is harper-only.

Complexity: easy

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -33,7 +33,7 @@ jobs:
     # Note: the `claude-review` label name is matched there too —
     # `_claude-review.yml`'s authorize `if:`, not in this caller.
     if: ${{ vars.CLAUDE_ALWAYS_ON == 'true' || github.event.action == 'labeled' }}
-    uses: HarperFast/ai-review-prompts/.github/workflows/_claude-review.yml@224c2adcfdb408d68c7ca48eb618623c990a96ad # main 2026-07-28 (#80 week-of-07-20 calibration: producing-side guard verification; on 82c9484)
+    uses: HarperFast/ai-review-prompts/.github/workflows/_claude-review.yml@4b59dc0ddb15aff517884127b58204f79193b4f2 # main 2026-08-20 (#86 reliable review evidence, context, and run binding)
     # Caller-side permissions at the calling-job level (NOT workflow-
     # level — that placement caps the reusable's per-job grants below
     # what they need and breaks the workflow at startup; see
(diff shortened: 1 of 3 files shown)


cb1kenobi:
Reviewed `97dc895f` — no issues found. This PR looks good, nice job!

—
_Generated by Barber AI_

Issue #2234, opened 2026-08-19
Windows CI shard 2/6 native abort() during component load with file watchers


kriszyp:
## Problem

Windows integration shard 2/6 terminates Harper with a native libuv abort while component file watchers start. The CI health probe then reports `ECONNREFUSED` because the Harper process has exited, not because startup is slow or a test assertion failed.

## Evidence

In the `harper-server-logs-windows-shard-2` artifact from [CI run 32218666106](https://github.com/HarperFast/harper/actions/runs/32218666106), the Harper process's `stderr.log` ends with:

```text
Assertion failed: !_wcsnicmp(filename, dir, dirlen), file src\win\fs-event.c, line 72
```

This is libuv's `uv__relative_path`, reached from `uv_fs_event_start` when a file watch starts. `hdb.log` places the abort during component load as that component's file watchers start. This is Harper's own stderr; child-process output is tagged separately in `hdb.log`.

The failure is branch-attributable: 11/11 commits on `kris/deploy-peer-rollback` fail this shard, while `main` is 12/12 clean on the same shard.

## Follow-up

No Windows environment is available to bisect this further. This needs a Windows-capable reproduction and fix; it is intentionally not being addressed in PR #2066.

kriszyp:
Root cause identified, and this is no longer branch-attributable — it now hits every branch.

## Root cause

The assertion is libuv's `uv__relative_path`, reached from the Windows fs-event completion path. When the watch target is a **file**, `uv_fs_event_start` stores the file's directory in `handle->dirw` and watches that directory. On each event the completion handler rebuilds the event's absolute path, expands it with `GetLongPathNameW`, and asserts the expansion still starts with `handle->dirw`:

```c
assert(!_wcsnicmp(filename, dir, dirlen));
```

If the stored directory is an 8.3 **short** path, the long-form expansion never shares that prefix, and libuv aborts the process rather than failing the watch. On the GitHub Windows runner `os.tmpdir()` is `C:\Users\RUNNER~1\AppData\Local\Temp`, so every Harper instance in this shard runs with a short-path data root — `Root Path: C:\Users\RUNNER~1\AppData\Local\Temp\describe-metadata-upgrade-…` in the artifact's `stdout.log`.

[...]

PR #2191, opened 2026-08-17
fix(config): read root config synchronously and bound rename retries by wall clock

kriszyp:
The Windows `EPERM: rename harper-config.yaml.tmp` was never a budget that was too short: the worker was blocked against a descriptor **its own root-config watcher held**. libuv opens a `fsPromises.readFile` descriptor on the threadpool but closes it from JS, which cannot run while the same thread sits in `atomicWriteFile`'s `Atomics.wait` — so no rename budget could have been long enough, and the extra 6.4 seconds an earlier head bought was pure event-loop stall on a path `set_configuration` reaches from a live request thread.

So the write side goes back to [the 3,630 ms window](https://github.com/HarperFast/harper/pull/2191/changes?w=1#diff-1cae2fa92233c61d3d0d4adae28c39c362b9b54673a6475955f427af0d047a21R98) it had before this branch, and the read side stops holding a handle across a yield: both root-config watchers read through [`readConfigFileSync`](https://github.com/HarperFast/harper/pull/2191/changes?w=1#diff-ee468ec946876ff56522a4df29b491e7082d575174d4783bd909110a26866720R13), and a read that still loses is retried from [a timer](https://github.com/HarperFast/harper/pull/2191/changes?w=1#diff-5c09fdc48da4f9ddceb259589ce2023b2f5c9b87bd5fcbc4270a1ed1c7d96a44R14), which holds no descriptor either.

[...]

PR #2309, opened 2026-08-25
Canonicalize watch paths so a Windows 8.3 short path cannot abort the process

kriszyp:
Refs #2234.

Harper aborts on Windows when a watched path carries an 8.3 short component. libuv's fs-event callback rebuilds each event's absolute path, expands it with `GetLongPathNameW`, and asserts the expansion still starts with the directory it stored when the watch was armed — `Assertion failed: !_wcsnicmp(filename, dir, dirlen), file src\win\fs-event.c, line 72`. A short directory (`C:\Users\RUNNER~1\...`) never survives that comparison, and libuv aborts the process rather than failing the watch. There is no JS-observable seam, so the existing `isWatcherExhaustionError` → polling recovery never runs and nothing is logged; the process is simply gone.

That is what has been failing `Integration Tests 2/6 (Windows, Node.js v24)` on roughly half of all runs since 2026-08-24, on every branch. The job only shows `Probe /SeoPageCache/ did not become ready within 120000ms (last error=ECONNREFUSED)` from `describe-metadata-upgrade.test.ts`; the assertion is only in the server-log artifact. On the GitHub Windows runner `os.tmpdir()` is `C:\Users\RUNNER~1\AppData\Local\Temp`, so every Harper instance in that shard runs with a short-path data root.

The surface is wider than "watches on a single file", which is the shape libuv needs: chokidar v4 opens a per-file `fs.watch` for every file it discovers inside a watched tree, so one component-directory watch arms hundreds of them. It is also not CI-only — the TLS certificate/private-key reload watcher is one of the six sites, and an abort there lands during automated renewal on a live node.

- [`canonicalizeWatchPath`](https://github.com/HarperFast/harper/pull/2309/changes?w=1#diff-1ddd8fe9d97134d7b3c50305fccb61ccf3b7f7dcf2b95614ca67b25fe137c513R26) resolves a path to its long form before it can reach a native watch, and returns `undefined` when it cannot — [`resolveWatchTarget`](https://github.com/HarperFast/harper/pull/2309/changes?w=1#diff-1ddd8fe9d97134d7b3c50305fccb61ccf3b7f7dcf2b95614ca67b25fe137c513R47) turns that into `mustPoll`, which each caller feeds into its existing polling options. Polling stats the file instead of arming a native watch, so it cannot reach the abort.
- **Every** Windows path is resolved, not only the ones that look short. An earlier revision gated on a `~<digits>` spelling test; `GetLongPathNameW`'s documentation is explicit that a short name need not contain a tilde, and NTFS allows an explicitly assigned one, so the test had false negatives that left the abort reachable — and false positives that would have pushed a genuine `archive~2024` install into permanent polling. `realpathSync` is not a substitute for the `.native` variant: it resolves symlinks but leaves 8.3 names intact.
- [A leaf that does not exist yet resolves through its directory](https://github.com/HarperFast/harper/pull/2309/changes?w=1#diff-1ddd8fe9d97134d7b3c50305fccb61ccf3b7f7dcf2b95614ca67b25fe137c513R31), because libuv stores and compares only the *parent directory* of a file target. Without that, a config watcher armed during the install window — the "file not written yet" case `OptionsWatcher` already documents — would fail closed to polling for the life of the process.
- Applied at all six sites: the [component tree](https://github.com/HarperFast/harper/pull/2309/changes?w=1#diff-c54fec2f7823e23c3a38c014bbc3721d06c5da4e83d3465dfe9bd2ede9152a05R542) and component config watchers, the root config watcher, the [TLS reload watcher](https://github.com/HarperFast/harper/pull/2309/changes?w=1#diff-ea5297f5d9097656312d08f0a1534f0426fb6ef893a2f5bf3a456b18f81a6841R360), the `WATCH_DIR` dev reloader, and the [incomplete-blob read watcher](https://github.com/HarperFast/harper/pull/2309/changes?w=1#diff-5c96de79cba36ba9ddb5124ce7b7e96a398a3c50fc6b7bec11dc5d690f61767aR769). `fs.watchFile` (`utility/logging/readLog.ts`) is stat polling with no fs-event handle and is outside the invariant.
- `EntryHandler` is the one caller where the canonical path is load-bearing past the `fs.watch` call: chokidar's `ignored` predicate receives absolute paths built from `cwd`, so its bases are now derived from the same directory spelling (and hoisted out of the per-entry callback, where they were being rebuilt for every discovered path). Event paths are relative to `cwd`, so reads stay on the configured `component.directory`.
- `resources/blob.ts` had no polling story of its own, so [a degraded watch there polls `readMore`](https://github.com/HarperFast/harper/pull/2309/changes?w=1#diff-5c96de79cba36ba9ddb5124ce7b7e96a398a3c50fc6b7bec11dc5d690f61767aR791) on the same 20 ms backoff and the same `incompleteDeadline` the in-progress-write stall already uses, rather than sitting out the full read timeout and returning a 503 on a healthy live upload.
- Two adjacent crash paths in that same call, both pre-existing on `main` and both surfaced by review: [`fs.watch` throws synchronously](https://github.com/HarperFast/harper/pull/2309/changes?w=1#diff-5c96de79cba36ba9ddb5124ce7b7e96a398a3c50fc6b7bec11dc5d690f61767aR771) when the OS watcher pool is exhausted (EMFILE/ENOSPC), and [an `FSWatcher` that fails after registration emits `'error'`](https://github.com/HarperFast/harper/pull/2309/changes?w=1#diff-5c96de79cba36ba9ddb5124ce7b7e96a398a3c50fc6b7bec11dc5d690f61767aR787) — with no listener Node rethrows it out of the watcher callback. Both now route into the same poll fallback.
- That whole block is [`watchInProgressFile()`](https://github.com/HarperFast/harper/pull/2309/changes?w=1#diff-5c96de79cba36ba9ddb5124ce7b7e96a398a3c50fc6b7bec11dc5d690f61767aR231), exported with an injectable `watch`, so the paths a normal read cannot reach without an exhausted OS watcher pool are covered by [six focused unit tests](https://github.com/HarperFast/harper/pull/2309/changes?w=1#diff-9b125a1f520c4f030fa53590e672ec5278ebf46b2a2b78a9e6bbd70ea42b9382R2108). Extracting it also closed a hole: [both callbacks now go through the same `isLive` identity check](https://github.com/HarperFast/harper/pull/2309/changes?w=1#diff-5c96de79cba36ba9ddb5124ce7b7e96a398a3c50fc6b7bec11dc5d690f61767aR234), where before only the `'error'` listener did — so neither a late error nor a [late change](https://github.com/HarperFast/harper/pull/2309/changes?w=1#diff-9b125a1f520c4f030fa53590e672ec5278ebf46b2a2b78a9e6bbd70ea42b9382R2196) from a watcher the read has already replaced can close the live watcher and start a second read sharing its fd and position.
- The two chokidar recovery sites [start their `close()` from a microtask](https://github.com/HarperFast/harper/pull/2309/changes?w=1#diff-ea5297f5d9097656312d08f0a1534f0426fb6ef893a2f5bf3a456b18f81a6841R378) rather than as the argument to `Promise.resolve(…)`, which evaluated it synchronously: a synchronous throw from `close()` escaped the chained `.catch()` and left the `'error'` listener as an uncaught exception. That ordering is now pinned by a test on each site — `unitTests/security/keys.test.js` and a new `unitTests/server/threads/watchDirFallback.test.js`, which is also the first coverage `watchDir` has had — and the reopen no longer swallows a synchronous `openWatcher` throw silently, so a permanently unwatched certificate or component directory is logged.
- `resources/blob.ts`'s `pull()` retired its in-progress watcher from five places, each hand-rolling close-then-null. They now share one `closeWatcher()` that drops the handle *before* closing it — so a callback racing the close fails its `isLive` check — and tolerates a throwing `close()`, so no teardown (`onError`, `resumeIfWriterFinished`, `onChange`, the `readSync` shortcut, `cancel()`) can be abandoned partway. Net fewer lines than the five copies.
- `EntryHandler` hands chokidar `component.commonPatternBase`. chokidar resolves a *relative* base against the canonicalized `cwd`, but an absolute one reaches the native watch as spelled, so an absolute base now goes through `resolveWatchTarget` as well.

## For the human reviewer

- **No Windows host was available, so nothing here has been executed against real 8.3 expansion or a real native watch.** The [unit tests](https://github.com/HarperFast/harper/pull/2309/changes?w=1#diff-fc5383f6c452b5a4fb9007116843d496dbc30dc836efc6a96f10bc66f80a0c1fR40) drive the branch logic with a symlink named `RUNNER~1` pointing at a long-named directory, which models what `realpathSync.native` does to an alias. That proves the algorithm, not libuv. The proof that matters is this PR's own `Integration Tests (Windows, Node.js v24)` matrix — please check whether all six Windows shards are green, and re-run once, because the failure it targets is ~50/50 per run. `Refs` rather than `Fixes` for exactly that reason.
- **Declined — a Windows-only regression test.** The strongest test would obtain a real 8.3 alias, arm both a chokidar file watch and a direct `fs.watch` under it in a child process, mutate the files, and assert both event delivery and child survival, isolated because unfixed it aborts the runner. I did not write it blind: unit tests are ubuntu-only in CI (`.github/workflows/unit-test.yml`), so it would never run there, and I cannot check it on the platform it exists for. @kriszyp has a Windows box for the follow-up.
- **Declined — a lint tripwire for the invariant.** "New watch sites must go through the helper" is enforced only by the DESIGN.md paragraph, and the failure it prevents is an unloggable process abort, so a reviewer suggested an oxlint `no-restricted-imports` rule with `utility/watchPath.ts` as the only allowed importer (`.oxlintrc.json` already bans `node:assert/strict` this way). It is a good idea and I'd support it as a follow-up; it needs a seven-file allowlist plus the test files that stub chokidar, which is scope growth on an already large change.
- **Done, after this was declined once — a test seam for the new `resources/blob.ts` watcher branches.** The earlier revision declined it on the grounds that a seam through the blob read path was a worse trade than the uncovered branch; @kriszyp chose the seam. It is one exported function with an injectable `watch` argument that production never passes, which is what the tests above drive. What it does *not* cover: the `mustPoll` decision inside `resolveWatchTarget` (win32-only), and the read loop's own no-watcher deadline branch, which still needs a `FileBackedBlob` stream to reach.
- **Not done — reopening a watcher after a *non-exhaustion* error.** A reviewer noted that chokidar can close the underlying watcher when it emits, so an `EACCES` on the TLS or dev-reload watcher logs and then silently never fires again. `security/keys.ts` is backstopped by its periodic re-read, and the dev reloader is `WATCH_DIR`-only, so this is a degraded fast path rather than a lost renewal — but re-arming on a delay after any error is a behavior decision I did not want to make inside a scoped follow-up.
- **Not done — the 20 ms poll re-allocates its read buffer.** Each re-entry into `readMore` allocates a fresh 256 KiB buffer before probing for data, so a read stuck in the fallback churns allocations on a host that is already under watcher pressure. The allocation is pre-existing and shared with the stall poll this fallback reuses; reusing a per-pull buffer is a separate change to that path.
- **Not done — direct coverage for `closeWatcher()`'s catch (raised by the review bot after the last push).** The `watchInProgressFile` tests drive that helper's own close-throw guard directly, but `pull()`'s `closeWatcher()` has no seam: `watchInProgressFile`'s injectable `watch` argument is a parameter of the *helper*, and production `pull()` calls it with the default. Reaching the catch from a real read means adding a second injection point to `pull()` itself — the same design question as the read loop's no-watcher deadline branch, which is the other known gap below.
- **Not done — a call-level watch-site tripwire.** The scan asserts each listed file *mentions* the canonicalization helper, so a second raw `fs.watch` beside a canonicalized one in an already-listed file stays green; making it call-level means parsing rather than grepping.
- **Not done — the dev reloader's queued restart is an `async` `setTimeout` callback.** A throw from `beforeRestart()` or `restartWorkers()` in `server/threads/manageThreads.js` becomes an unhandled rejection. It is byte-identical on `origin/main` — the block is only in this diff because the watch site around it moved — and it is `WATCH_DIR`-only, so it is recorded rather than absorbed; this PR has already taken on three pre-existing blob crash paths.
- **Considered and declined — `resolveWatchTarget` as synchronous I/O on the blob tail path.** A reviewer read `watchTarget ??= resolveWatchTarget(filePath)` as a per-read `realpathSync.native`. Off Windows `canonicalizeWatchPath` returns the path with no syscall at all, and on Windows it runs once per stream (the `??=` caches it) and only after a read has already stalled with no data to return.
- **Not done — an absolute component pattern base is separately broken.** `EntryHandler` now canonicalizes an absolute `commonPatternBase` too, so it cannot reach a native watch unexpanded. But such a component is already non-functional for a different reason: `join(watchDirectory, base)` in the `ignored` predicate produces garbage for an absolute base, so every discovered path is ignored. `Component` rejects a pattern that starts with `/`, which makes this unreachable on POSIX but leaves a Windows drive-letter pattern (`C:\...`) through. Rejecting absolute patterns outright is a user-facing validation change and belongs in its own PR.
- **Declined — extracting the reopen-on-exhaustion sequence into a shared helper.** `security/keys.ts` and `manageThreads.js`'s `watchDir` now hand-roll the same `usingPolling` / `liveWatcher !== opened` guard, `warnWatcherFallback`, close-and-reopen sequence that `EntryHandler`, `OptionsWatcher` and `RootConfigWatcher` already carry as private methods — five copies of one invariant, and a reviewer asked for a shared `utility/watcherFallback.ts` helper. It is the right end state and I'd support it as a follow-up alongside the lint tripwire above; @kriszyp confirmed the deferral when this follow-up was scoped. Three of the five copies predate this PR, and the two shapes differ (private class fields with a `#openWatcher` method vs. closure locals), so a common helper is a cross-cutting refactor of five files rather than part of this fix.
- **Symlink and junction resolution on Windows.** `realpathSync.native` also resolves reparse points, so a watched path that is a junction is now watched at its target on Windows; retargeting the junction without touching the target would not be seen. Two reviewers raised it and the adjudicating pass dropped it both rounds on the grounds that `fs.watch`/inotify already follow a symlinked file to its target inode, so the same retarget is already invisible on Linux and macOS — this makes Windows match, rather than diverge. `security/keys.ts` also keeps its periodic re-read as a backstop. Flagging it because it is the one deliberate platform-behavior change in the diff.
- **A watcher that degrades to polling stays there for its lifetime**, and the warning is one-shot process-wide, so a second, later degradation logs nothing. That matches how `#usingPolling` already latches for ENOSPC/EMFILE — it is now the same field, so `_usingPollingForTests` reports it — and it is a one-line change if you want it re-armed per watcher.
- **Rebased onto latest `main` (`760f5ffc9`) at `8d0c6ce45`, then two further review rounds fixed a config-watcher gap the rebase's wider CI surface exposed.** `OptionsWatcher`/`RootConfigWatcher` still called the failed watcher's `close()` directly inside the chokidar `'error'` listener — unlike `blob.ts`/`keys.ts`/`manageThreads.js`, hardened in earlier rounds — so a synchronous throw there would escape as an uncaught exception; fixed by starting `close()` from a microtask there too. The same round surfaced a second gap in the same two files: the reopen was chained via `.finally()` with no trailing `.catch()`, so a synchronous throw from `#openWatcher()` on reopen would become an unhandled rejection — added the same trailing `.catch()` pattern `keys.ts`/`manageThreads.js` already use. Both are now covered by a mirrored test in each file, using the sinon-free `chokidar.watch` reassignment pattern from `watchDirFallback.test.js` (per `AGENTS.md`'s no-new-sinon rule) — which also replaced this branch's own earlier `sinon.stub(chokidar, 'watch')` additions in `keys.test.js` with the same pattern.
- **CI at `8d0c6ce45`: all six Windows shards ran (after one re-run); `Integration Tests 2/6 (Windows)` — the shard this fix targets — passed, and `6/6` failed on its first two attempts for the same unrelated known reason as before, then passed on a third attempt.** Both failures were `set_configuration` tests hitting `EPERM: operation not permitted, rename ...harper-config.yaml.tmp -> harper-config.yaml` from `server/operationsServer.ts:345` — a *different* test each time (`replicated: true rejects explicitly`, then `writes an operator-named component _package entry`), matching the same flake documented independently on #2318, #2039, and #809 (all unrelated PRs, none touching watch paths or config writing) and tracked at #2313 with a fix already up in #2191. All other checks (Build, Unit Tests ×3 Node versions, Integration Tests 1–5/6 across Bun/Node/uWS/Windows, Next.js adapter, smoke, lint, format, coverage) passed on the first attempt.
- **CI at `485414892`: all six Windows shards ran; `Integration Tests 2/6 (Windows)` — the shard this fix targets — passed again, and `6/6` failed for the same unrelated known reason as before (`set_configuration` → `EPERM ... harper-config.yaml.tmp -> harper-config.yaml`).** All three `Unit Test` jobs failed on the first attempt and passed on re-run with no code change; the two failures were HNSW vector-index assertions (`greedy descent changed the result set`, `bulk delete severed survivors`) in `unitTests/resources/vectorIndex*`, on a code path this diff does not touch, and they pass locally at this head.
- **CI at the earlier head `31cfbbbe`: all six Windows shards ran; `Integration Tests 2/6 (Windows)` — the shard this fix targets — passed, and `6/6` failed for an unrelated known reason.** That failure is `Configuration` → `set_configuration` returning 500 from `EPERM: operation not permitted, rename ...harper-config.yaml.tmp -> harper-config.yaml`, which is [Windows: set_configuration returns 500 because the config write's rename retry blocks the thread that would release the handle](https://github.com/HarperFast/harper/issues/2313), already under fix in [fix(config): read root config synchronously and bound rename retries by wall clock](https://github.com/HarperFast/harper/pull/2191). It reproduces identically on `main` with no part of this diff present — same suite, same rename, same message — in [this `main` Integration Tests run](https://github.com/HarperFast/harper/actions/runs/32853337399). Nothing in this diff touches config writing.

## Verification

- Rebase round (`8d0c6ce45`, onto `main@760f5ffc9`): `npm run build`, `oxlint --deny-warnings` on all changed files, and every affected suite together — `unitTests/resources/blob.test.js`, `unitTests/utility/watchPath.test.js`, `unitTests/server/threads/watchDirFallback.test.js`, `unitTests/components/OptionsWatcher*.test.js`, `unitTests/config/rootConfigWatcher.test.js`, `unitTests/security/keys.test.js` — 234 passing, 0 failing, re-run after each of the two review-round fixes.
- Follow-up round (`485414892`): `npm run build`, changed-file prettier + oxlint clean, and each affected suite run on its own — `unitTests/resources/blob.test.js` 110 passing, `unitTests/security/keys.test.js` 62 passing, `unitTests/utility/watchPath.test.js` 13 passing, `unitTests/components/EntryHandler.test.js` 40 passing, `unitTests/server/threads/*.test.js` 36 passing. 0 failing.
- The synchronous-`close()`-throw fix is now mutation-pinned on both sites: reverting either reopen chain to `Promise.resolve(opened.close())` fails `unitTests/security/keys.test.js` and `unitTests/server/threads/watchDirFallback.test.js`, re-measured after the suites were rewritten to `node:assert`.
- Running `unitTests/resources` alongside `unitTests/security` in one mocha process reports 4 failures in two unrelated `createTLSSelector` cases, from an analytics-writer timer firing across suites. Identical, same two cases, on this branch's prior head `31cfbbbe2` with these files reverted — pre-existing cross-directory pollution, and not a shape CI runs (`test:unit:main` excludes `unitTests/resources`).

- Re-verified at the pushed head `31cfbbbe`: `npm run build`, then `unitTests/resources/blob.test.js` + `unitTests/utility/watchPath.test.js` + `unitTests/security/keys.test.js` — 183 passing, 0 failing.
- Follow-up review fix: non-`ENOENT` native realpath failures now fail closed to polling; `npm run build`, changed-file oxlint, and `unitTests/utility/watchPath.test.js` all pass (13 tests).
- `npm run test:integration:all` completed 1,816 passing and 20 skipped; 6 real-Ollama children were cancelled by an unrelated JSON import-attribute startup error in that optional suite.

- `npm run build`, `npm run lint:required`, `npm run test:types` — all pass.
- `unitTests/utility/watchPath.test.js` — 13 passing: the non-Windows identity return, directory and file resolution through a short-form ancestor, the not-yet-written leaf resolving through its directory, fail-closed when not even the directory resolves, and both `resolveWatchTarget` outcomes.
- Every suite covering a changed file — `unitTests/components`, `unitTests/config`, `unitTests/security`, `unitTests/utility`, `unitTests/resources/blob*` — 3195 passing, 44 pending, 0 failing. The two `resources/blob.ts` commits after that run were re-covered by `unitTests/resources/blob*` + `unitTests/utility/watchPath.test.js` (111 passing) and `unitTests/config/rootConfigWatcher.test.js` (6 passing).
- `npm run test:unit:resources` — 1684 passing, 16 pending, 3 failing; the three (`randomAccessFields` ×2, `replayStructures`) are this machine's pre-existing baseline failures.
- `npm run test:unit:main` cannot run on this machine: it aborts at module load against a stale local Harper install, **identically on a pristine `origin/main` checkout**, so it is environmental. The suites above cover its files for this diff.
- End-to-end route: this PR's own Windows CI matrix, as described above. Not reproducible locally — no Windows host.
- The watcher-seam follow-up: `unitTests/resources/blob.test.js` 109 passing, `unitTests/security` 664 passing, `unitTests/utility` 530 passing, `npm run build`, `npm run format:write` and `npm run lint:required` clean. `npm run test:unit:resources` — 1688 passing, 16 pending, 8 failing, and the same 8 fail identically on this branch's prior head with these files reverted, so they are this machine's baseline, not the change. `unitTests/components` and `unitTests/server` cannot start on this machine at all (they abort at module load resolving a storage root, on the unmodified tree too).
- Each of the five helper behaviors was mutation-checked rather than assumed: dropping the `isLive` guard from either callback, dropping the `mustPoll` latch, watching the uncanonicalized path, and registering unconditionally each fail at least one of the new tests.

Independent review: seventeen rounds (`prepush-review.mjs`, receipt count) — codex, Gemini, Cursor Grok and the Harper-domain adjudicator; the round numbers below are the substantive ones. The four rounds after `31cfbbbe` ran Gemini alone, which is why the receipt reads `declined=codex`: the adjudicating and codex legs had already covered this diff four times, and each of those rounds reviewed a small delta that came directly out of Gemini's own prior findings. Round 1: 5 findings, 5 fixed. Round 2 found the one that changed the design — an earlier `~<digits>` gate had false negatives, so the abort survived for a non-tilde alias — plus the `resources/blob.ts` fallback gap; both fixed. Rounds 4–6 each surfaced one further pre-existing crash path in the blob watcher call, all fixed above. Round 7 (the seam commit) raised the unguarded change callback as major — fixed in the next commit. Round 8 raised the synchronous `close()` throw — fixed. Round 9 is LGTM from the graded leg with one repeat comment-style nit. Round 10 (codex + Gemini, on the new regression coverage): 3 findings, 3 fixed — the new `watchDir` suite moved to bare `node:assert` per AGENTS.md, and both suites that fire an exhaustion error now reset `warnWatcherFallback`'s process-global one-time gate. Round 11 raised the uncanonicalized absolute `commonPatternBase` and the unguarded `close()` in `watchInProgressFile`'s error listener — both fixed. Round 12 raised the five hand-rolled blob teardowns and the two silently-swallowed reopen throws — both fixed. Round 13 produced no new actionable finding: one pre-existing `main` defect, one misreading of the lazy `??=` watch-target cache, and two repeats. After the rebase onto latest `main`, further rounds found and fixed the config-watcher microtask/reopen gap described above, plus a new-sinon house-style violation in the tests added to cover it; subsequent rounds converged with no further actionable findings beyond the pre-existing/deferred items already listed above. Everything raised and not taken is a `Not done` / `Declined` entry above.

Complexity: moderate — a small path-normalization helper, but it changes what every file watcher in the process is handed, and one caller (`resources/blob.ts`) needed a new fallback inside a delicate stall path.

— Claude Opus

<sub>Review-Coverage: authored=claude; ran=codex,gemini; declined=cursor-grok,cursor-composer,domain; rounds=23 @ 8d0c6ce4519d</sub>

<sub>Human-Review-Need: 4 @ 8d0c6ce4519d</sub>

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -1213,3 +1213,35 @@ An `{}` in the config system is context-dependent, and conflating the contexts i
 Removal therefore prunes: `deleteNestedValue` removes ancestors the deletion emptied, only when it actually deleted an existing leaf, and reports what it pruned. The overlap case — a file-declared empty scope an env layer temporarily populated — is tracked in the state file's `emptyScopeOriginals` (separate from `originalValues` so a marker can never mask or be consumed as a real leaf original at the same path; older state files lacking the field are defaulted). Restore consumes a marker only for a path the prune actually removed, so a scalar overwrite or an absent-leaf no-op can never resurrect a scope over live env-layer content. Note there are two coexisting mechanisms for "file `{}` is user content": `restoreBaseEmptyObjects` on the stateless compose path and the marker pair on the stateful removal path — if you touch one, check the other.
 
 Two durable limitations of the marker mechanism, both with user config-file content as the blast radius: markers can only be recorded at populate time, so a scope an env layer populated _before_ `emptyScopeOriginals` existed (any pre-upgrade boot) has no marker and prunes away on its first post-upgrade vacate; and a corrupt config-state file resets to fresh state — dropping `originalValues` and `emptyScopeOriginals` for every tracked path — after which the next removal prunes those scopes for good; `saveConfigState` writes via temp+rename precisely so a torn write cannot be the trigger, leaving genuine corruption (disk faults, hand edits) as the remaining path.
+
+## Every path handed to a native file watch must be canonicalized (`utility/watchPath.ts`)
+
+libuv's Windows fs-event callback rebuilds each event's absolute path, expands it with
+`GetLongPathNameW`, and asserts the expansion still starts with the directory it stored when the
+watch was armed. An 8.3 short directory (`C:\Users\RUNNER~1\...`) never survives that comparison,
+and libuv **aborts the process** rather than failing the watch — there is no JS-observable seam, so
+`isWatcherExhaustionError`/polling recovery never runs (harper#2234).
+
+The trap is that libuv only stores that directory for **file** targets, which reads as a narrow
+surface until you follow chokidar: v4 opens a per-file `fs.watch` for every file it discovers inside
+a watched tree, so one directory watch arms hundreds of file watches.
+
+So `canonicalizeWatchPath` runs on every path before it reaches `fs.watch` (directly or through
+chokidar). It only touches a path that actually carries an 8.3 component — everything else is
+returned as given, which keeps `realpathSync.native`'s symlink resolution off every path that cannot
+abort. It returns `undefined` when it cannot prove the long form, _including when the resolved path
+is still short_; `resolveWatchTarget` turns that into `mustPoll`, and polling stats the file instead
+of arming a native watch. Plain `realpathSync` is not a substitute: it resolves symlinks but leaves
+8.3 names intact.
+
+New watch sites must go through it. As of this writing the sites are `components/EntryHandler.ts`,
+`components/OptionsWatcher.ts`, `config/RootConfigWatcher.ts`, `security/keys.ts`,
+`server/threads/manageThreads.js`, and `resources/blob.ts`. `fs.watchFile` (`utility/logging/readLog.ts`)
+is stat polling with no fs-event handle and is outside this invariant.
+
+Two consequences worth knowing before adding a caller. `EntryHandler` is the one place where the
+canonical path is load-bearing past the `fs.watch` call: chokidar's `ignored` predicate receives
+absolute paths built from `cwd`, so its bases must be derived from the same spelling, while event
+paths are relative to `cwd` and reads stay on the configured `component.directory`. And a watcher
+that degrades to polling stays there for its lifetime, so a caller with no polling story of its own
+(`resources/blob.ts`) needs one — there it polls `readMore` on the existing no-progress deadline.
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -1227,12 +1227,15 @@ surface until you follow chokidar: v4 opens a per-file `fs.watch` for every file
 a watched tree, so one directory watch arms hundreds of file watches.
 
 So `canonicalizeWatchPath` runs on every path before it reaches `fs.watch` (directly or through
-chokidar). It only touches a path that actually carries an 8.3 component — everything else is
-returned as given, which keeps `realpathSync.native`'s symlink resolution off every path that cannot
-abort. It returns `undefined` when it cannot prove the long form, _including when the resolved path
-is still short_; `resolveWatchTarget` turns that into `mustPoll`, and polling stats the file instead
-of arming a native watch. Plain `realpathSync` is not a substitute: it resolves symlinks but leaves
-8.3 names intact.
+chokidar), and returns `undefined` when it cannot establish the long form; `resolveWatchTarget` turns
+that into `mustPoll`, and polling stats the file instead of arming a native watch. It resolves every
+Windows path rather than only the ones that look short: `GetLongPathNameW`'s documentation is
+explicit that a short name need not contain a tilde, so any spelling test leaves the abort reachable.
+Plain `realpathSync` is not a substitute for the `.native` variant: it resolves symlinks but leaves
+8.3 names intact — which also means Windows watch paths are symlink-resolved, matching what
+`fs.watch` already does elsewhere by following a symlinked file to its target inode. A leaf that does
+not exist yet resolves through its directory, because libuv stores and compares only the parent
+directory of a file target.
 
 New watch sites must go through it. As of this writing the sites are `components/EntryHandler.ts`,
 `components/OptionsWatcher.ts`, `config/RootConfigWatcher.ts`, `security/keys.ts`,
--- a/components/EntryHandler.ts
+++ b/components/EntryHandler.ts
@@ -537,8 +537,8 @@ export class EntryHandler extends EventEmitter<EntryHandlerEventMap> {
 			readFailures: new Map(),
 		};
 
-		// chokidar reports every path — to `ignored` and to the event handlers — relative to `cwd`,
-		// so the predicate's bases must be derived from the same canonicalized directory.
+		// `ignored` receives absolute paths built from `cwd`, so its bases must come from the same
+		// directory spelling. Event paths are relative to `cwd`, so reads below stay on `this.directory`.
 		const watchTarget = resolveWatchTarget(this.#component.directory);
 		const watchDirectory = watchTarget.path;
 		const normalizedDirectory = watchDirectory.replace(/\\/g, '/');
--- a/components/EntryHandler.ts
+++ b/components/EntryHandler.ts
@@ -5,7 +5,7 @@ import type { Stats } from 'node:fs';
 import { EventEmitter } from 'node:events';
 import { Component, FileAndURLPathConfig } from './Component.ts';
 import chokidar, { FSWatcher, FSWatcherEventMap } from 'chokidar';
-import { join } from 'node:path';
+import { isAbsolute, join } from 'node:path';
 import { readFile } from 'node:fs/promises';
 import { FilesOption } from './deriveGlobOptions.ts';
 import { deriveURLPath } from './deriveURLPath.ts';
@@ -545,9 +545,18 @@ export class EntryHandler extends EventEmitter<EntryHandlerEventMap> {
 		const normalizedDirectory = watchDirectory.replace(/\\/g, '/');
 		const normalizedBases = this.#component.patternBases.map((base) => join(watchDirectory, base).replace(/\\/g, '/'));
 
+		// chokidar resolves a relative pattern base against `cwd`, but an absolute one reaches the
+		// native watch as spelled, bypassing the canonicalized `cwd` above.
+		let watchPattern = this.#component.commonPatternBase;
+		if (isAbsolute(watchPattern)) {
+			const patternTarget = resolveWatchTarget(watchPattern);
+			if (patternTarget.mustPoll) this.#usingPolling = true;
+			watchPattern = patternTarget.path;
+		}
+
 		this.#openCount++;
 		const watcher = (this.#watcher = chokidar
-			.watch(this.#component.commonPatternBase, {
+			.watch(watchPattern, {
 				cwd: watchDirectory,
 				persistent: false,
 				followSymlinks: false,
--- a/components/EntryHandler.ts
+++ b/components/EntryHandler.ts
@@ -540,7 +540,7 @@ export class EntryHandler extends EventEmitter<EntryHandlerEventMap> {
 		// `ignored` receives absolute paths built from `cwd`, so its bases must come from the same
 		// directory spelling. Event paths are relative to `cwd`, so reads below stay on `this.directory`.
 		const watchTarget = resolveWatchTarget(this.#component.directory);
-		// Latched, not assigned: a reinstall must not clear an exhaustion fallback.
+		// Latched: a reinstall must not clear an exhaustion fallback.
 		if (watchTarget.mustPoll) this.#usingPolling = true;
 		const watchDirectory = watchTarget.path;
 		const normalizedDirectory = watchDirectory.replace(/\\/g, '/');
--- a/components/EntryHandler.ts
+++ b/components/EntryHandler.ts
@@ -540,7 +540,6 @@ export class EntryHandler extends EventEmitter<EntryHandlerEventMap> {
 		// `ignored` receives absolute paths built from `cwd`, so its bases must come from the same
 		// directory spelling. Event paths are relative to `cwd`, so reads below stay on `this.directory`.
 		const watchTarget = resolveWatchTarget(this.#component.directory);
-		// Latched: a reinstall must not clear an exhaustion fallback.
 		if (watchTarget.mustPoll) this.#usingPolling = true;
 		const watchDirectory = watchTarget.path;
 		const normalizedDirectory = watchDirectory.replace(/\\/g, '/');
--- a/components/EntryHandler.ts
+++ b/components/EntryHandler.ts
@@ -540,8 +540,7 @@ export class EntryHandler extends EventEmitter<EntryHandlerEventMap> {
 		// `ignored` receives absolute paths built from `cwd`, so its bases must come from the same
 		// directory spelling. Event paths are relative to `cwd`, so reads below stay on `this.directory`.
 		const watchTarget = resolveWatchTarget(this.#component.directory);
-		// Latch rather than assign: an exhaustion fallback already set this, and a reinstall must not
-		// clear it.
+		// Latched, not assigned: a reinstall must not clear an exhaustion fallback.
 		if (watchTarget.mustPoll) this.#usingPolling = true;
 		const watchDirectory = watchTarget.path;
 		const normalizedDirectory = watchDirectory.replace(/\\/g, '/');
--- a/components/EntryHandler.ts
+++ b/components/EntryHandler.ts
@@ -540,6 +540,9 @@ export class EntryHandler extends EventEmitter<EntryHandlerEventMap> {
 		// `ignored` receives absolute paths built from `cwd`, so its bases must come from the same
 		// directory spelling. Event paths are relative to `cwd`, so reads below stay on `this.directory`.
 		const watchTarget = resolveWatchTarget(this.#component.directory);
+		// Latch rather than assign: an exhaustion fallback already set this, and a reinstall must not
+		// clear it.
+		if (watchTarget.mustPoll) this.#usingPolling = true;
 		const watchDirectory = watchTarget.path;
 		const normalizedDirectory = watchDirectory.replace(/\\/g, '/');
 		const normalizedBases = this.#component.patternBases.map((base) => join(watchDirectory, base).replace(/\\/g, '/'));
@@ -550,7 +553,7 @@ export class EntryHandler extends EventEmitter<EntryHandlerEventMap> {
 				cwd: watchDirectory,
 				persistent: false,
 				followSymlinks: false,
-				...(this.#usingPolling || watchTarget.mustPoll ? DIRECTORY_POLLING_FALLBACK_OPTIONS : {}),
+				...(this.#usingPolling ? DIRECTORY_POLLING_FALLBACK_OPTIONS : {}),
 				ignored: (path) => {
 					const normalizedPath = path.replace(/\\/g, '/');
 
--- a/components/EntryHandler.ts
+++ b/components/EntryHandler.ts
@@ -15,6 +15,7 @@ import {
 	isWatcherExhaustionError,
 	warnWatcherFallback,
 } from '../utility/watcherFallback.ts';
+import { resolveWatchTarget } from '../utility/watchPath.ts';
 
 export interface BaseEntry {
 	stats?: Stats;
@@ -536,19 +537,22 @@ export class EntryHandler extends EventEmitter<EntryHandlerEventMap> {
 			readFailures: new Map(),
 		};
 
-		const allowedBases = this.#component.patternBases.map((base) => join(this.#component.directory, base));
+		// chokidar reports every path — to `ignored` and to the event handlers — relative to `cwd`,
+		// so the predicate's bases must be derived from the same canonicalized directory.
+		const watchTarget = resolveWatchTarget(this.#component.directory);
+		const watchDirectory = watchTarget.path;
+		const normalizedDirectory = watchDirectory.replace(/\\/g, '/');
+		const normalizedBases = this.#component.patternBases.map((base) => join(watchDirectory, base).replace(/\\/g, '/'));
 
 		this.#openCount++;
 		const watcher = (this.#watcher = chokidar
 			.watch(this.#component.commonPatternBase, {
-				cwd: this.#component.directory,
+				cwd: watchDirectory,
 				persistent: false,
 				followSymlinks: false,
-				...(this.#usingPolling ? DIRECTORY_POLLING_FALLBACK_OPTIONS : {}),
+				...(this.#usingPolling || watchTarget.mustPoll ? DIRECTORY_POLLING_FALLBACK_OPTIONS : {}),
 				ignored: (path) => {
 					const normalizedPath = path.replace(/\\/g, '/');
-					const normalizedBases = allowedBases.map((base) => base.replace(/\\/g, '/'));
-					const normalizedDirectory = this.#component.directory.replace(/\\/g, '/');
 
 					// Determine the path relative to the component directory. Leading '/' is preserved
 					// (or empty when the path *is* the component directory) so the regex anchors below
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -807,8 +807,7 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
 											readMore(resolve, reject);
 										} else if (!resumeIfWriterFinished()) {
 											if (!watcher) {
-												// No watcher could be armed, so nothing will wake this read. Poll on the same
-												// short backoff and the same no-progress deadline resumeIfWriterFinished uses,
+												// Nothing will wake this read, so poll on the deadline resumeIfWriterFinished uses
 												// rather than sitting out the full read timeout and 503-ing a healthy write.
 												if (Date.now() >= incompleteDeadline) {
 													onError(
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -535,7 +535,6 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
 		let position = 0;
 		let totalContentRead = 0;
 		let watcher: FSWatcher;
-		// Resolved once per read: the stall path below re-arms the watcher on every retry.
 		let watchTarget: { path: string; mustPoll: boolean };
 		let timer: NodeJS.Timeout;
 		// The start() open-retry timer lives in a different scope/phase than pull()'s `timer`; track it
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -804,11 +804,15 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
 											watcher = undefined;
 										}
 										// An FSWatcher that fails after registration emits 'error'; with no listener Node
-										// rethrows it out of the watcher callback. Drop to the same poll instead.
-										watcher?.on('error', (error) => {
+										// rethrows it out of the watcher callback. Drop to the same poll instead — but only
+										// while this watcher is still the live one, or a queued error from a superseded
+										// watcher would close its replacement and start a second read at the same position.
+										const installedWatcher = watcher;
+										installedWatcher?.once('error', (error) => {
+											if (watcher !== installedWatcher) return;
 											logger.debug?.(`Watch of ${filePath} failed, polling instead:`, error);
-											watcher?.close();
 											watcher = undefined;
+											installedWatcher.close();
 											clearTimeout(timer);
 											timer = setTimeout(() => readMore(resolve, reject), 20).unref();
 										});
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -801,6 +801,9 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
 													});
 										} catch (error) {
 											logger.debug?.(`Could not watch ${filePath} for in-progress writes, polling instead:`, error);
+											// Latch on the stream-scoped target, or the 20 ms re-poll below re-enters here and
+											// re-attempts the same failing registration for the rest of the read.
+											watchTarget.mustPoll = true;
 											watcher = undefined;
 										}
 										// An FSWatcher that fails after registration emits 'error'; with no listener Node
@@ -811,6 +814,7 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
 										installedWatcher?.once('error', (error) => {
 											if (watcher !== installedWatcher) return;
 											logger.debug?.(`Watch of ${filePath} failed, polling instead:`, error);
+											watchTarget.mustPoll = true;
 											watcher = undefined;
 											installedWatcher.close();
 											clearTimeout(timer);
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -51,7 +51,7 @@ import { get as envGet, getHdbBasePath } from '../utility/environment/environmen
 import { CONFIG_PARAMS, MAX_SET_TIMEOUT_MS } from '../utility/hdbTerms.ts';
 import { join, dirname } from 'path';
 import { logger } from '../utility/logging/logger.ts';
-import { canonicalizeWatchPath } from '../utility/watchPath.ts';
+import { resolveWatchTarget } from '../utility/watchPath.ts';
 import type { RootDatabase } from 'lmdb';
 import { asyncSerialization, hasAsyncSerialization } from '../server/serverHelpers/contentTypes.ts';
 import { getHeapStatistics } from 'node:v8';
@@ -536,8 +536,7 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
 		let totalContentRead = 0;
 		let watcher: FSWatcher;
 		// Resolved once per read: the stall path below re-arms the watcher on every retry.
-		let watchPath: string | undefined;
-		let watchPathResolved = false;
+		let watchTarget: { path: string; mustPoll: boolean };
 		let timer: NodeJS.Timeout;
 		// The start() open-retry timer lives in a different scope/phase than pull()'s `timer`; track it
 		// separately so a cancel() during the file-creation wait clears it instead of leaking an fd (#1457).
@@ -787,20 +786,17 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
 										};
 										// the file is not finished being written, watch the file for changes to resume reading
 										// set up a watcher to be notified of file changes
-										if (!watchPathResolved) {
-											watchPathResolved = true;
-											watchPath = canonicalizeWatchPath(filePath);
-										}
-										watcher =
-											watchPath &&
-											watch(watchPath, { persistent: false }, () => {
-												if (watcher) {
-													watcher.close();
-													watcher = null;
-													clearTimeout(timer); // clear it
-													readMore(resolve, reject);
-												}
-											});
+										watchTarget ??= resolveWatchTarget(filePath);
+										watcher = watchTarget.mustPoll
+											? undefined
+											: watch(watchTarget.path, { persistent: false }, () => {
+													if (watcher) {
+														watcher.close();
+														watcher = null;
+														clearTimeout(timer); // clear it
+														readMore(resolve, reject);
+													}
+												});
 										// immediately try to read again in case there was a change before we started watching,
 										// readSync should be fine here, the data should be in memory
 										if (readSync(fd, buffer, 0, buffer.length, position) > 0) {
@@ -811,6 +807,22 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
 											}
 											readMore(resolve, reject);
 										} else if (!resumeIfWriterFinished()) {
+											if (!watcher) {
+												// No watcher could be armed, so nothing will wake this read. Poll on the same
+												// short backoff and the same no-progress deadline resumeIfWriterFinished uses,
+												// rather than sitting out the full read timeout and 503-ing a healthy write.
+												if (Date.now() >= incompleteDeadline) {
+													onError(
+														new BlobReadError(
+															`File read timed out reading from ${filePath}, read ${totalContentRead} bytes, but size is supposed to be ${size} bytes`,
+															BLOB_UNAVAILABLE_STATUS
+														)
+													);
+												} else {
+													timer = setTimeout(() => readMore(resolve, reject), 20).unref();
+												}
+												return;
+											}
 											// set a timer for the watcher too. A write that stalls past the bound returns a
 											// prompt 503 (retryable) instead of holding the connection for the full 60s (#1423).
 											timer = setTimeout(() => {
--- a/resources/blob.ts
+++ b/resources/blob.ts
@@ -240,14 +240,12 @@ const BLOB_GONE_STATUS = 404;
 export const BLOB_UNAVAILABLE_STATUS = 503;
 const BLOB_CORRUPT_STATUS = 500;
 /**
- * Arm the one-shot watch that wakes an in-progress blob read, or report that the read must poll
- * instead. Exported (with an injectable `watchFile`) so the failure paths — a synchronous
- * registration throw, a post-registration `'error'`, and an `'error'` from a watcher the read has
- * already replaced — are directly testable; production never passes the last argument.
- *
- * `onFailure` runs only for a live watcher that fails after registration; a registration that
- * throws latches `mustPoll` and returns undefined, leaving the caller's own no-watcher branch to
- * schedule the poll.
+ * Arm the one-shot watch that wakes an in-progress blob read of `filePath`, or return `undefined`
+ * for a read that has to poll instead. `isLive` decides whether the read still owns a callback's
+ * watcher: a callback from one it has already replaced must not act, or it would close the live
+ * watcher and start a second read at the same position. `onFailure` runs only for a live watcher
+ * that fails after registration — a registration that throws latches `mustPoll` and leaves the
+ * caller in its own no-watcher branch. `watchFile` is injectable so those paths are testable.
  */
 export function watchInProgressFile(
 	filePath: string,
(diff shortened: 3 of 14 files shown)


unitTests/utility/watchPath.test.js
```diff
@@ -0,0 +1,91 @@
+const { canonicalizeWatchPath, resolveWatchTarget, _resetForTests } = require('#src/utility/watchPath');
+const assert = require('node:assert');
+const { mkdtempSync, mkdirSync, symlinkSync, realpathSync, writeFileSync, rmSync } = require('node:fs');
+const { join } = require('node:path');
+const { tmpdir } = require('node:os');
+
+// A symlink named `RUNNER~1` pointing at a long-named directory models the Windows 8.3 alias this
+// exists for: `realpathSync.native` rewrites the component.
```
cb1kenobi: **Medium: nothing tests the one detail the whole fix hinges on — `.native`**

I mutated `utility/watchPath.ts:29` from `realpathSync.native(watchPath)` to `realpathSync(watchPath)`, rebuilt with `tsc -p tsconfig.build.json` (confirming exactly one changed line in `dist/utility/watchPath.js`), and re-ran this suite: **8 passing, 0 failing.** The mutant survives.

The reason is this comment's own premise: a symlink is a faithful model of an 8.3 alias for *plain* `realpathSync` too, so the suite cannot distinguish the two functions. But `DESIGN.md` and the `canonicalizeWatchPath` JSDoc both state that plain `realpathSync` "is not a substitute for the `.native` variant" — it resolves symlinks but leaves 8.3 names intact. So the single call that makes this PR work is the one call with no coverage, and a regression to the non-native form would ship green and silently restore the abort.

Suggested fix: assert the call *target* rather than the result — stub or spy `fs.realpathSync.native` in one case and assert `canonicalizeWatchPath(p, 'win32')` invoked it. Roughly five lines, and it runs on the ubuntu-only unit job where the rest of this file already runs.

(Not asking for the Windows-only regression test you declined — that reasoning is sound. This is the cheap guard that fits where the tests already are.)

—
_Generated by Barber AI_

DESIGN.md
```diff
@@ -1011,3 +1011,38 @@ An `{}` in the config system is context-dependent, and conflating the contexts i
...
+that into `mustPoll`, and polling stats the file instead of arming a native watch. It resolves every
+Windows path rather than only the ones that look short: `GetLongPathNameW`'s documentation is
+explicit that a short name need not contain a tilde, so any spelling test leaves the abort reachable.
+Plain `realpathSync` is not a substitute for the `.native` variant: it resolves symlinks but leaves
+8.3 names intact — which also means Windows watch paths are symlink-resolved, matching what
+`fs.watch` already does elsewhere by following a symlinked file to its target inode. A leaf that does
+not exist yet resolves through its directory, because libuv stores and compares only the parent
+directory of a file target.
+
+New watch sites must go through it. As of this writing the sites are `components/EntryHandler.ts`,
```
cb1kenobi, line 1240: **Medium: this invariant is enforced by nothing executable**

`grep -rn "resolveWatchTarget|canonicalizeWatchPath|mustPoll" unitTests/ integrationTests/` returns zero hits outside `unitTests/utility/watchPath.test.js`. Combined with the `platform !== 'win32'` early return, every one of the six call-site changes is a provable no-op on Harper's ubuntu and macOS runners — so no test on any CI runner this repo actually uses exercises the canonicalized branch at a call site. A seventh site added later that skips the helper fails nothing, and the failure it produces is the unloggable process abort this PR exists to remove.

I read the decline of the oxlint `no-restricted-imports` rule and agree the seven-file allowlist plus the chokidar-stubbing test files is real scope growth. There is a cheaper mechanism that avoids all of that: a ~15-line unit test that reads the source tree and asserts no file outside a single allowlist array contains `chokidar.watch(` or `fs.watch(`. No lint config, no import-graph allowlist, no interaction with the tests that stub chokidar, and it runs in the ubuntu-only unit job alongside `watchPath.test.js`.

Worth doing in this PR rather than as a follow-up, since the DESIGN.md paragraph is currently the only thing standing between a new watch site and a silent abort.

—
_Generated by Barber AI_

security/keys.ts
```diff
@@ -355,7 +357,13 @@ function loadAndWatch(path, loadCert, type) {
 	};
 	if (fs.existsSync(path)) loadFile(path, statSync(path));
 	else logger.error?.(`${type} file not found:`, path);
-	watch(path, { persistent: false }).on('change', loadFile);
+	const watchTarget = resolveWatchTarget(path);
+	watch(watchTarget.path, { persistent: false, ...(watchTarget.mustPoll ? POLLING_FALLBACK_OPTIONS : {}) }).on(
```
cb1kenobi: **Low (pre-existing): this watcher still has no `'error'` listener**

`resources/blob.ts` in this PR fixes exactly this shape — "an `FSWatcher` that fails after registration emits `'error'`; with no listener Node rethrows it out of the watcher callback" — but the statement being edited here keeps it. chokidar 4.0.3's `_handleError` (`esm/index.js:528`) emits unguarded for any code that is not `ENOENT`/`ENOTDIR`, which includes `ENOSPC` and `EMFILE`.

Confirmed by running it: a `chokidar.watch(dir, { persistent: false }).on('change', ...)` with no `'error'` listener turns an emitted `ENOSPC` into an `uncaughtException`. Every other watch site in the repo — `EntryHandler`, `OptionsWatcher`, `RootConfigWatcher` — registers one.

Low rather than higher because Harper installs log-and-continue `uncaughtException` handlers on both the main thread (`server/threads/socketRouter.ts:19`) and workers (`server/threads/threadServer.js:83`), and the periodic re-read backstops cert reload. So the consequence is a silently dead fast-path watcher plus a confusing log line, not a crash — but on a host under watcher pressure, TLS renewal quietly drops to the 5-minute poll with no attributable warning.

Suggested fix: add an `.on('error', ...)` that logs the watcher error against `type` and `path`, or route it through `isWatcherExhaustionError`/`warnWatcherFallback` the way the other sites do.

—
_Generated by Barber AI_

components/EntryHandler.ts
```diff
@@ -536,19 +537,22 @@ export class EntryHandler extends EventEmitter<EntryHandlerEventMap> {
...
 
 		this.#openCount++;
 		const watcher = (this.#watcher = chokidar
 			.watch(this.#component.commonPatternBase, {
-				cwd: this.#component.directory,
+				cwd: watchDirectory,
 				persistent: false,
 				followSymlinks: false,
-				...(this.#usingPolling ? DIRECTORY_POLLING_FALLBACK_OPTIONS : {}),
+				...(this.#usingPolling || watchTarget.mustPoll ? DIRECTORY_POLLING_FALLBACK_OPTIONS : {}),
```
cb1kenobi: **Nit: `mustPoll` is not folded into `#usingPolling` here**

`OptionsWatcher.ts:112` and `RootConfigWatcher.ts:24` both assign `this.#usingPolling = watchTarget.mustPoll`; this site ORs the two at the call instead. The PR description says "it is now the same field, so `_usingPollingForTests` reports it" — that holds for those two, but not for `EntryHandler`, where an instance forced to polling by an unresolvable path still reports `_usingPollingForTests === false`.

No functional impact — `#installWatcher()` re-resolves on every reinstall, so the polling options stay correct across pause/resume. Just the observability claim that does not hold at this one site.

—
_Generated by Barber AI_

unitTests/utility/watchPath.test.js
kriszyp: Addressed in dd73cfcc2 and covered at the final head e61454d41: the utility test instruments `fs.realpathSync.native` directly, so replacing it with plain `realpathSync` now fails the suite. The follow-up also proves non-ENOENT native failures fail closed.

DESIGN.md
kriszyp: Addressed in dd73cfcc2 (with scan hardening through abc46899b): the utility suite inventories native-watch source sites and requires every listed file to route through the canonicalization helper. This keeps the requested executable tripwire in the existing unit suite.

security/keys.ts
kriszyp: Addressed in dd73cfcc2: the TLS watcher now handles errors, falls back to polling for ENOSPC/EMFILE, logs other watcher failures with path/type context, and has focused exhaustion coverage.

components/EntryHandler.ts
kriszyp: Addressed in dd73cfcc2: `watchTarget.mustPoll` is latched into `#usingPolling` before watcher creation, so `_usingPollingForTests` now reflects path-canonicalization fallback consistently.

cb1kenobi:
Reviewed `e61454d4` — no open issues. This PR looks good, nice job!

Both Mediums from the last round are closed, and I re-ran them as mutations rather than reading the assertions:

- **`.native` coverage.** Re-ran the exact mutation that survived last time — `realpathSync.native` → plain `realpathSync` at `utility/watchPath.ts:29`, rebuilt with `tsc -p tsconfig.build.json`, exactly one isolated marker in `dist/utility/watchPath.js`. Last round: 8 passing / 0 failing (survived). Now: **10 passing / 3 failing** — killed by all three of the new `recordingNativeCalls` cases. The call target is genuinely pinned now.
- **The watch-site invariant.** The `native watch sites` block is the executable form of the DESIGN.md prose. Both halves kill their mutant: adding a rogue `node:fs` `watch()` in a new file fails *"are exactly the files that canonicalize their watch path"*, and renaming `resolveWatchTarget` out of `security/keys.ts` fails *"all route their path through the canonicalization helper"*. Worth noting it scans source, not `dist/`, so it is immune to the `#src/*` → `dist/*.js` build indirection that hides `.ts` edits from mocha.
- **`keys.ts` `'error'` listener (the Low).** Now present, with the exhaustion reopen-on-polling latch, and covered by *"reopens on polling when the watcher reports exhaustion"* — including the assertion that repeated ENOSPC emissions do not open a third watcher.

The four invariants I said had to stay true still do, re-measured rather than assumed:

- All six production native-watch sites canonicalize (`EntryHandler.ts:542`, `OptionsWatcher.ts:105`, `RootConfigWatcher.ts:22`, `keys.ts:360`, `manageThreads.js:1235`, `blob.ts:772`), with `readLog.ts:634` correctly outside the invariant as `fs.watchFile`.
- `canonicalizeWatchPath` still returns at `platform !== 'win32'` before the `try`. Runtime probe after module load: `realpathSync.native` calls = 0, `realpathSync` calls = 0 for `canonicalizeWatchPath`/`resolveWatchTarget` on darwin and linux. No symlink resolution, no new syscalls off Windows.
- No identity clobber in `blob.ts` — `open()` at :558 and every error message still use `filePath`; only `watch()` at :778 sees `watchTarget.path`.
- The unresolvable fallback cannot reach the abort. Instrumented chokidar 4.0.3: `usePolling: true` on a single file gives `fs.watch` = 0, `fs.watchFile` = 1 (change events still fire); on a directory tree, `fs.watch` = 0, `fs.watchFile` = 5. Without it, 1 and 5 `fs.watch` calls respectively. `blob.ts`'s `mustPoll` path arms no watcher at all.

Two caveats worth stating plainly:

1. **None of this verifies Windows.** Every measurement above is from macOS, where the helper is a hard no-op by construction. The 8.3 behavior itself — that `realpathSync.native` expands a short component and plain `realpathSync` does not — is unverifiable from here and is taken from the `GetLongPathNameW` contract, not from observation.
2. **Six green Windows shards is one run against a roughly 50/50 failure**, which is close to no evidence. The re-run request from the last round still stands.

Local runs vs. the true merge base `f4314e76`: `unitTests/utility` 669 pass / 0 fail (base 656 / 0, the +13 being the new file), `unitTests/security` 662 / 2 (base 661 / 2 — the same two `jsLoader` SES failures, pre-existing and environmental), `unitTests/config` 258 / 0 both sides, and `components` + `server` 2270 / 13 against base 2260 / 23, where head's failures are an exact subset of base's. The red `Unit Test` checks on CI are the known `unitTests/resources` HNSW flake (`vectorIndex.test.js:1009`), not this change.

—
_Generated by Barber AI_

resources/blob.ts
```diff
@@ -764,13 +766,36 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
...
+										// watcher would close its replacement and start a second read at the same position.
+										const installedWatcher = watcher;
+										installedWatcher?.once('error', (error) => {
+											if (watcher !== installedWatcher) return;
+											logger.debug?.(`Watch of ${filePath} failed, polling instead:`, error);
+											watcher = undefined;
+											installedWatcher.close();
+											clearTimeout(timer);
+											timer = setTimeout(() => readMore(resolve, reject), 20).unref();
 										});
```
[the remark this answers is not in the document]
kriszyp: Done in 31cfbbbe2. The block is now `watchInProgressFile()` in `resources/blob.ts`, exported with an injectable `watch`, and six tests in `unitTests/resources/blob.test.js` cover it: a target that must poll arms nothing, the canonical path (not the reported one) is watched, a registration that throws latches `mustPoll` so the 20 ms re-poll stops re-attempting it, a live watcher that errors after registration drops to polling, and a late change *or* error from a superseded watcher is ignored.

Extracting it surfaced a hole the suggestion did not name: only the `'error' listener was identity-guarded, so a change callback delivered after the read installed a replacement would have closed the live watcher and resumed a second read sharing its fd and position. Both callbacks now go through the same `isLive` check. Each of the five behaviors was mutation-checked against the built `dist` rather than assumed.

— Claude Opus

security/keys.ts
```diff
@@ -355,7 +357,34 @@ function loadAndWatch(path, loadCert, type) {
...
+					Promise.resolve(opened.close())
+						.catch(() => {})
+						.then(openWatcher)
+						.catch(() => {});
+					return;
+				}
+				logger.error?.(`Error watching ${type}:`, path, error);
+			});
+	};
+	openWatcher();
```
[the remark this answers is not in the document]
kriszyp: Deferred deliberately — @kriszyp scoped this follow-up to the blob watcher test seam only, and the five copies are not the same shape (one-shot `fs.watch` here vs. a persistent chokidar watcher with a safety poller, a directory tree, and three class implementations with close/generation lifecycles), so a shared helper is a design pass against all five rather than a migration of the two new ones. Recorded in the PR description under `## For the human reviewer`.

One thing from this round did land at both new sites in 31cfbbbe2: `Promise.resolve(opened.close())` evaluated `close()` synchronously, so a synchronous throw escaped the chained `.catch()` and left the `'error' listener as an uncaught exception. Both now start the close from a microtask.

— Claude Opus

security/keys.ts
```diff
@@ -355,7 +357,35 @@ function loadAndWatch(path, loadCert, type) {
...
+			// The event carries the watched spelling, which is not the configured one once canonicalized;
+			// reload through the configured path so a retargeted link is followed.
+			.on('change', () => loadFile(path))
+			// chokidar emits 'error' unguarded for anything but ENOENT/ENOTDIR.
+			.on('error', (error) => {
+				if (isWatcherExhaustionError(error)) {
+					if (usingPolling || liveWatcher !== opened) return;
+					warnWatcherFallback(path);
+					usingPolling = true;
+					Promise.resolve()
```
cb1kenobi, line 378: **Low: the synchronous-close-throw fix has no regression test**

Confirmed by mutation: reverting this hunk back to `Promise.resolve(opened.close())` (the escaped-throw shape this commit fixes) and rerunning the full `unitTests/security` suite still passes 662/662 — the fix is currently provable only by reading the code, not by a test that would fail without it. The `beforeEach` fake watcher's `close: () => Promise.resolve()` never throws, so `reopens on polling when the watcher reports exhaustion` (line ~1153) exercises the reopen path but not this specific ordering bug.

Same gap in `server/threads/manageThreads.js:1254` (identical fix, and `watchDir` has no dedicated test file at all).

Not blocking — the fix itself is correct (deferring `close()` into a `.then()` puts a synchronous throw inside the chain instead of evaluating it eagerly as `Promise.resolve()`'s argument). Flagging so a future edit to this block doesn't silently reopen the escape hatch.

Generated by Barber AI

cb1kenobi:
Reviewed `48541489` (was `31cfbbbe`) — no issues found. This PR looks good, nice job!

Same merge-base (`f4314e76`), four new commits (rounds 10–12 plus the synchronous-`close()` coverage). CI green (44 success, 4 skipped), all five prior threads resolved.

**`EntryHandler.ts` — canonicalizing an absolute pattern base.** This closes a real hole in the PR's own fix: `cwd` was already canonicalized, but chokidar resolves a *relative* pattern base against `cwd` while an absolute one reaches the native watch as spelled — so an absolute `commonPatternBase` bypassed the 8.3 short-path canonicalization entirely. Propagating `patternTarget.mustPoll` into `#usingPolling` is right too, since the pattern base can sit on a different volume than the component directory and independently require polling. Verified the symbols are actually in scope rather than assuming: `resolveWatchTarget` is imported at `EntryHandler.ts:18`, `#usingPolling` is declared at `:108`, and `isAbsolute` was added to the existing `node:path` import.

**`blob.ts` — the `closeWatcher()` consolidation.** Null-then-close is the correct order and the comment states why precisely. `watchInProgressFile`'s error listener gates on `handlers.isLive(watcher)`, which is `(candidate) => watcher === candidate` against the *outer* binding — so nulling first means a callback racing the close fails its liveness check instead of running against a retired watcher. Five call sites now funnel through it, and the `try/catch` means a throwing `close()` can no longer abandon the teardown it was part of.

I checked the one site that deliberately does *not* call it: `onFailure` sets `watcher = null` directly. That is correct and not an oversight — `watchInProgressFile` already closed the handle immediately before invoking `onFailure`, so routing it through `closeWatcher()` would double-close. Changing it from `undefined` to `null` also makes the sentinel consistent across every path.

**`keys.ts` / `manageThreads.js` — logging a failed reopen.** Both previously ended the polling-fallback chain in `.catch(() => {})`, so a watcher that failed to reopen left the process silently un-watched with no trace. Each now logs, and each uses the logging facility its own file already uses (`logger.error?.` vs `console.error`) rather than importing a new one.

Coverage tracks the change: `watchDirFallback.test.js` (+77), `keys.test.js` (+41), `blob.test.js` (+15).

—
_Generated by Barber AI_

resources/blob.ts
```diff
@@ -515,6 +564,19 @@ class FileBackedBlob extends (Blob as unknown as { new (): Blob }) implements Bl
...
 		let totalContentRead = 0;
 		let watcher: FSWatcher;
+		// Drop the live watcher before closing it, so a callback racing the close cannot pass its
+		// `isLive` check — and so a close() on an already-failed handle cannot abandon the teardown
+		// it was part of. Every site that retires this watcher goes through here.
+		const closeWatcher = () => {
+			const opened = watcher;
+			watcher = null;
+			try {
+				opened?.close();
```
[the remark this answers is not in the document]
kriszyp: Addressed as a deliberate non-blocking deferral in the PR description under `## For the human reviewer`. Directly driving this catch through `pull()` would require adding a second production injection seam solely for the test; the existing `watchInProgressFile` seam cannot replace the helper used by `stream()`. The defensive close ordering and catch remain covered by code review, while a call-level test is recorded as follow-up scope rather than expanding this already broad fix.

kriszyp:
## Windows verification on real hardware

This PR shipped with "**No Windows host was available, so nothing here has been executed against real 8.3 expansion or a real native watch**" and asked a human to check the Windows matrix. Both PRs have now been run on a real Windows 11 machine (Node v24.14.0, libuv 1.51.0) whose `os.tmpdir()` is `C:\Users\KRISZY~1\AppData\Local\Temp` — the same 8.3 shape as the runner.

### The premise is confirmed, from CI artifacts rather than a local repro

Being precise about what was and was not reproduced:

- **Not reproduced locally.** I could not trip the abort by hand: real generated 8.3 aliases, length-mismatched names (`AVERYL~1` → `averylongdirectorynamehere`), directory / file / recursive / trailing-separator / forward-slash / dot-suffixed watch shapes, and an ACL intended to make `GetLongPathNameW` fail — all survived, with correct event filenames.
- **Confirmed from CI.** Downloading the `harper-server-logs-windows-shard-2` artifacts, `Assertion failed: !_wcsnicmp(filename, dir, dirlen), file src\win\fs-event.c, line 72` appears in `stderr.log` in **three separate recent `main` runs** (32905597531, 32898256258, 32877287726) — always the `describe_all metadata upgrade phase 1` instance, data root `C:\Users\RUNNER~1\...`, immediately after `packGitReferenceWithoutScripts` / `extractApplication`, i.e. during component extract/install directory churn.
- **The assertion is live in release Node.** `src\win\fs-event.c` is present as a UTF-16 string inside `node.exe`, so libuv's asserts are compiled into Windows release builds. Not a debug-build artifact.

### Why the fix is sufficient regardless of the unknown trigger

Reading libuv 1.51.0 `src/win/fs-event.c`: line 72 is reachable **only** from `uv__fs_event_cb`'s `handle->dirw` branch, comparing the per-event `GetLongPathNameW` expansion against whatever `uv_fs_event_start` stored. `handle->dirw` is assigned only on the directory branch (line 320) and stays `NULL` for file targets — so the exposed surface is directory watches, exactly as this PR's chokidar note describes.

That yields a stronger sufficiency argument than the PR currently makes, and one that does not depend on knowing *why* `GetLongPathNameW` failed on the runner: **once libuv is handed an already-long path, `dirw` is long, so `GetLongPathNameW(dirw + "\" + name)` always starts with `dirw` and the assertion cannot fire.** The `mustPoll` branch is safe by a different route — chokidar's `usePolling` path uses `fs.watchFile` stat polling and never arms an fs-event handle at all. Both branches are structurally closed, not probabilistically improved.

### A real defect in this PR's own test — fixed in this push

`unitTests/utility/watchPath.test.js` **could not run on Windows at all**: 0 passing, failing in its `before` hook with

```
Error: EEXIST: file already exists, symlink '...\watch-path-XXXX\runneradmin' -> '...\watch-path-XXXX\RUNNER~1'
```

The fixture creates a directory `runneradmin` and then a symlink named `RUNNER~1` beside it — but on a volume with 8.3 name creation enabled (the default, and the exact configuration this helper exists for) NTFS has *already* assigned `RUNNER~1` to `runneradmin`, so `symlinkSync` collides. The one suite written to model Windows behaviour was the one suite that could not run on Windows, and unit tests are `ubuntu-latest`-only, so nothing caught it.

Fixed by renaming the alias link, with a comment recording the constraint. **13/13 passing on Windows** — the first time `canonicalizeWatchPath` / `resolveWatchTarget` have been exercised against real `realpathSync.native` 8.3 expansion rather than only the Linux symlink model.

### CI on this push

All six Windows shards pass on the first attempt, including **2/6, the shard this fix targets**:

| shard | result |
|---|---|
| Integration Tests 1/6 (Windows) | pass 4m00s |
| **Integration Tests 2/6 (Windows)** | **pass 13m23s** |
| Integration Tests 3/6 (Windows) | pass 5m47s |
| Integration Tests 4/6 (Windows) | pass 3m59s |
| Integration Tests 5/6 (Windows) | pass 3m18s |
| Integration Tests 6/6 (Windows) | pass 4m22s |

Two attempt-1 failures, both explained, neither related to this diff:

- `Integration Tests 2/6 (Node.js v24)` failed in the **`Relink bin scripts`** setup step, before any test ran.
- `Integration Tests 6/6 (Windows)` failed on `set_configuration still writes an operator-named component _package entry` with `EPERM: operation not permitted, rename ...harper-config.yaml.tmp -> harper-config.yaml`. **That is #2339's failure, not this one** — this branch does not carry the descriptor-lifetime fix. The same run showing shard 2/6 green and shard 6/6 red is a clean demonstration that the two PRs fix disjoint bugs and neither subsumes the other. On the stacked #2339, all six shards are green.

### Local Windows results

- `unitTests/utility/watchPath.test.js` — 13 passing (after the fix above)
- `unitTests/resources/blob.test.js` — 115 passing
- `unitTests/server/threads/watchDirFallback.test.js` — 2 passing
- `integrationTests/deploy/redeploy-runtime-equivalence.test.ts` — 18/18
- `integrationTests/apiTests/describe-metadata-upgrade.test.ts` — 2/2
- Clean `npm run build`, `npm run typecheck`, `npm run lint:required`, `prettier --check`

Two suites could **not** be run to completion on this machine. Both fail **identically on `main`**, so they are environmental rather than regressions — but it does mean the two new `security/keys.ts` tests in this PR remain unverified on Windows:

- `unitTests/security/keys.test.js` — the `before all` hook exceeds its own `this.timeout(10000)`. Not mkcert (measured 565 ms for `createCA` + `createCert`); it stalls in the `testUtils.preTestPrep()` / `setupTestDBPath()` storage-root setup.
- `unitTests/components/EntryHandler.test.js` — 39 passing / 2 failing; the `after each` hook dies with an uncaught `EPERM: operation not permitted, watch`. See below.

### One gap this PR does not close

`main`'s Windows failures sort into three buckets, and only two are covered:

| shard | cause | fixed by |
|---|---|---|
| 2/6 | libuv 8.3 abort | **this PR** |
| 6/6 | `EPERM` rename on config write | #2339 |
| 4/6 | uncaught `EPERM: operation not permitted, watch` | **neither** |

chokidar 4.0.3 attaches its `'error'` listener to the underlying Node `FSWatcher` — including its Windows `EPERM` workaround for joyent/node#4337 — **only in the `persistent: true` branch** of `setFsWatchListener`:

```js
if (!options.persistent) {
    watcher = createFsWatchInstance(path, options, listener, errHandler, rawEmitter);
    if (!watcher) return;
    return watcher.close.bind(watcher);      // returns WITHOUT watcher.on('error', ...)
}
```

Every Harper chokidar watcher passes `persistent: false`, so when libuv delivers an async watch error it lands on an emitter with no listener and throws. Attaching `.on('error')` to the chokidar *wrapper* — which this PR does at several sites — does not help, because the native watcher's error never reaches the wrapper's error route in that branch. `resources/blob.ts` is the one site that *is* covered, precisely because this PR attaches a listener to its raw `fs.watch` handle.

Measured on Windows (arm the watch, then delete the watched directory):

| shape | result |
|---|---|
| `chokidar.watch(dir, {persistent:false})` | **uncaught EPERM watch** |
| `chokidar.watch(file, {persistent:false})` | **uncaught EPERM watch** |
| `chokidar.watch('.', {cwd, persistent:false})` | **uncaught EPERM watch** |
| `chokidar.watch('**', {cwd, persistent:false, ignored})` | survived |
| `chokidar.watch(dir, {persistent:true})` | survived |
| raw `fs.watch(dir, {persistent:false})` + `.on('error')` | survived |

This is pre-existing (it reproduces on `main`), it is what kills `Integration Tests 4/6 (Windows)` during `Redeploy runtime-equivalence proof`, and it is tracked separately rather than expanded into this PR. It is worth noting here because it sits directly adjacent to this PR's new `DESIGN.md` invariant: a watch site can canonicalize its path correctly and still take the process down.

### On two of the declined follow-ups

- The **Windows-only regression test** is worth revisiting now that a Windows host exists — but it still would not run in CI, because `.github/workflows/unit-test.yml` pins `ubuntu-latest`. Adding Windows unit-test coverage is filed separately; the pre-existing Windows failures above are the obstacles.
- The **repo-wide watch-site scan** does pass on Windows and enumerates correctly: an independent grep for every `chokidar.watch` / `fs.watch` call site in the tree returns exactly the six files the test lists.

🤖 Verification performed with [Claude Code](https://claude.com/claude-code) on Windows 11 / Node v24.14.0

Issue #2312, opened 2026-08-25
Worker thread exits cleanly (code 0) before reporting ready during concurrent multi-worker startup, aborting the whole node


kriszyp:
### Symptom

Recurring Linux Integration Tests failure on `main` — currently the top non-Windows CI failure (red in 3 of the last 5 runs):

```
Error: Worker (index 1) exited with code 0 before reporting ready
    at Worker.onExit (server/threads/socketRouter.ts:100:11)
```

immediately followed in *sibling* workers by a `Database not open` cascade, after which every in-flight test in the shard is cancelled.

Occurrences (same shape, different shards, suites, and Node versions — 22 / 24 / 26.5):
- 2026-08-25 https://github.com/HarperFast/harper/actions/runs/32812674585 — 3/6 (Node 22), deleteUpdateRace.test.ts
- 2026-08-24 https://github.com/HarperFast/harper/actions/runs/32732580586 — 1/6 (uWS), delete-update-race-consistency.test.ts; 3/6 (Node 22) blob.test.mjs
- 2026-08-24 https://github.com/HarperFast/harper/actions/runs/32741812476 — 3/6 (Node 26.5.0), record-caching-cross-worker.test.ts
- 2026-08-24 https://github.com/HarperFast/harper/actions/runs/32701165774 — 3/6 (Node 24), scheduled nightly

The affected suites are all multi-worker (`threads.count: 4`), RocksDB-engine, high-concurrency startup tests.

### Root cause

A Harper worker thread that has not yet posted `child_started` owns **no ref'd libuv handle**:

- `manageThreads.addPort()` explicitly `unref()`s the worker's `parentPort` (and sibling ports); `threadServer.js` only `.ref()`s it *after* `loadRootComponents()` resolves.
- Component file watchers are chokidar `persistent: false`; the resource-report and monitoring timers are `unref()`d.

So during `loadRootComponents()`, any `await` whose completion arrives through a source that does not hold the event loop lets the loop drain — and a worker whose loop drains exits **cleanly with code 0**, before the ready handshake. `createWorkerReadyPromise` then rejects, `startHTTPThreads`' `Promise.all(workersReady)` rejects, and `bin/run.ts`'s catch calls `process.exit(1)` — the whole node aborts during boot.

[...]

PR #2314, opened 2026-08-25
Keep a pre-ready worker's event loop alive through startup (fixes the 'exited with code 0 before reporting ready' CI failure)

kriszyp:
Fixes the recurring Linux Integration Tests failure `Worker (index N) exited with code 0 before reporting ready` from `createWorkerReadyPromise`. Refs #2312.

## Summary

A pre-ready worker can have no ref-holding handle while `loadRootComponents()` awaits a completion delivered by an unref'd source. Its event loop then drains, the worker exits cleanly with code 0, and main aborts startup.

- [`startServers()` refs `parentPort` before component loading](https://github.com/HarperFast/harper/pull/2314/changes?w=1#diff-d6ac497db42595bd89baa27ec98b82d5fd3cabd497962cc64ab475d46839be5bR180), preserving the worker until it reports `child_started`; the existing shutdown path still unrefs the port.
- [Rejected root-component loads terminate the ready chain](https://github.com/HarperFast/harper/pull/2314/changes?w=1#diff-d6ac497db42595bd89baa27ec98b82d5fd3cabd497962cc64ab475d46839be5bR191), including a never-settling fallback if the worker exit primitive is ever replaced with a returning test double.
- [`symlinkHarperModule()` gives only a contended waiter a ref'd 10-second timer](https://github.com/HarperFast/harper/pull/2314/changes?w=1#diff-cb052ff87cbadee003a6254cb87b7f05365677a40d1d112c4f62bcf033f13c98R399). This removes the winner's stale timer, which could later unlock a lock another thread owned, while preserving liveness during the native lock wake.
- The initially proposed fatal 10-minute startup deadline was removed after review. This PR intentionally does not introduce a startup availability policy or configuration surface.

## Regression coverage

- [The lock-wait unit tests track both timer creation and clearing](https://github.com/HarperFast/harper/pull/2314/changes?w=1#diff-ba451f78d579238a093c229069226de3ded62564d508e107c81f000ec1075df9R24), so the waiter assertion now fails against the original `clearTimeout(timeout)` bug and the winner assertion fails against its stale timer.
- [The standalone Worker regression executes the real `threadServer.startServers()` path](https://github.com/HarperFast/harper/pull/2314/changes?w=1#diff-44ef9ad40164679aa781a18d9f3ff972f93a153487716e01283501a13c788c8fR12) with an asserted-unref'd completion timer and hermetic temporary Harper config. Replacing the production `parentPort.ref()` with `unref()` fails with the historical `worker exited with code 0 before reporting ready` signature.
- The integration fixture mounts the liveness probe in both app-component and root-component startup windows and verifies the instance reaches ready and serves it.

## For the human reviewer

- Removing the fatal deadline means a component load that never settles can leave startup waiting indefinitely. That is the explicit owner decision for this focused liveness fix; watchdog and recovery policy belongs in a dedicated startup-coordinator change.
- A waiter that reaches the symlink-lock timeout continues with the link in its current state. The native lock API exposes no callback cancellation or safe stale-owner reclamation, so a dead holder can leave future loads paying the same timeout; changing that needs a separate lock-recovery design.
- `realExit(1)` is the existing immediate worker-failure contract. An outside reviewer proposed draining for logger flush instead, but that would materially change failure and shutdown semantics and is not included here.
- The direct `_symlinkHarperModuleForTests` export is retained to keep lock timing deterministic without adding a broader component-loading injection surface.

## Verification

- `npm run build` — passed.
- `npm run lint:required` and `npm run format:write` — passed.
- Focused unit tests — 3 passed.
- Mutation checks — removing the startup port ref fails with exit code 0 before ready; clearing the waiter timer fails the live-timer assertion.
- `npm run test:integration -- integrationTests/server/worker-startup-liveness.test.ts` — passed.
- `npm run test:unit:main` progressed normally but hung in an existing component-status suite and was interrupted; `test:unit:server` aborts because its glob directly loads `fixtures/processGroupOwnerWorker.js` with a null `parentPort`.
- Full independent review on the final behavioral head plus a focused delta review of the fixture-isolation changes; exact-head receipt at `945f40ea8`.

Complexity: 3

Refs #2312

— GPT-5 Codex

🤖 Generated with [Codex](https://openai.com/codex/)

<sub>Review-Coverage: authored=codex; ran=gemini,claude; declined=cursor-grok,cursor-composer,domain; rounds=3 @ 945f40ea8f75</sub>

<sub>Human-Review-Need: 4 @ 945f40ea8f75</sub>

### DIFF
--- a/components/componentLoader.ts
+++ b/components/componentLoader.ts
@@ -387,22 +387,23 @@ export const getComponentName = () => compName;
  * every non-root component load and repaired if missing or pointing elsewhere.
  */
 function symlinkHarperModule(componentDirectory: string) {
-	return new Promise<void>((resolve, reject) => {
+	return new Promise<void>((resolve) => {
 		const store = Status.primaryStore;
-		// Create timeout to avoid deadlocks
-		const timeout = setTimeout(() => {
-			store.unlock(componentDirectory);
-			reject(new Error('symlinking harperdb module timed out'));
-		}, 10_000);
-
-		const callback = () => {
+		let timeout: NodeJS.Timeout;
+		const onUnlocked = () => {
 			clearTimeout(timeout);
 			resolve();
 		};
-		const lockAcquired = store.tryLock(componentDirectory, callback);
+		const lockAcquired = store.tryLock(componentDirectory, onUnlocked);
 
 		if (!lockAcquired) {
-			clearTimeout(timeout);
+			// The lock holder performs the fixup; just wait for its unlock. The timer both bounds
+			// the wait if the holder dies without unlocking and — because the unlock wake arrives
+			// via an unref'd threadsafe function — keeps this pre-ready worker's event loop alive
+			// so it cannot drain and cleanly exit (code 0) before the ready handshake. Timing out
+			// resolves rather than rejects: the fixup is idempotent maintenance another thread
+			// already ran or the next load re-runs, never worth failing the component load over.
+			timeout = setTimeout(resolve, 10_000);
 		} else {
 			try {
 				// validate node_modules directory exists
--- a/components/componentLoader.ts
+++ b/components/componentLoader.ts
@@ -403,7 +403,12 @@ function symlinkHarperModule(componentDirectory: string) {
 			// so it cannot drain and cleanly exit (code 0) before the ready handshake. Timing out
 			// resolves rather than rejects: the fixup is idempotent maintenance another thread
 			// already ran or the next load re-runs, never worth failing the component load over.
-			timeout = setTimeout(resolve, 10_000);
+			timeout = setTimeout(() => {
+				harperLogger.warn(
+					`Timed out waiting for another thread to verify the harper module link in ${componentDirectory}; continuing with the link in its current state`
+				);
+				resolve();
+			}, 10_000);
 		} else {
 			try {
 				// validate node_modules directory exists
@@ -446,6 +451,9 @@ function symlinkHarperModule(componentDirectory: string) {
 		}
 	});
 }
+
+// Test-only: exercises the lock-wait branch directly (unitTests/components/symlinkHarperModuleLockWait.test.js).
+export const _symlinkHarperModuleForTests = symlinkHarperModule;
 
 /**
  * This function handles the `handleApplication` call for a plugin in a sequential manner.
--- a/components/componentLoader.ts
+++ b/components/componentLoader.ts
@@ -451,6 +451,7 @@ function symlinkHarperModule(componentDirectory: string) {
 	});
 }
 
+// Direct access keeps lock timing deterministic without exposing a broader component-load seam.
 export const _symlinkHarperModuleForTests = symlinkHarperModule;
 
 /**
--- a/components/componentLoader.ts
+++ b/components/componentLoader.ts
@@ -397,12 +397,11 @@ function symlinkHarperModule(componentDirectory: string) {
 		const lockAcquired = store.tryLock(componentDirectory, onUnlocked);
 
 		if (!lockAcquired) {
-			// The lock holder performs the fixup; just wait for its unlock. The timer both bounds
-			// the wait if the holder dies without unlocking and — because the unlock wake arrives
-			// via an unref'd threadsafe function — keeps this pre-ready worker's event loop alive
-			// so it cannot drain and cleanly exit (code 0) before the ready handshake. Timing out
-			// resolves rather than rejects: the fixup is idempotent maintenance another thread
-			// already ran or the next load re-runs, never worth failing the component load over.
+			// The lock holder performs the fixup; wait for its unlock. The timer bounds the wait if
+			// the holder dies without unlocking, and must stay ref'd: the unlock wake arrives via an
+			// unref'd threadsafe function (see the invariant comment in threadServer.startServers).
+			// Timing out resolves rather than rejects: the fixup is idempotent maintenance, never
+			// worth failing the component load over.
 			timeout = setTimeout(() => {
 				harperLogger.warn(
 					`Timed out waiting for another thread to verify the harper module link in ${componentDirectory}; continuing with the link in its current state`
@@ -452,7 +451,6 @@ function symlinkHarperModule(componentDirectory: string) {
 	});
 }
 
-// Test-only: exercises the lock-wait branch directly (unitTests/components/symlinkHarperModuleLockWait.test.js).
 export const _symlinkHarperModuleForTests = symlinkHarperModule;
 
 /**
--- a/integrationTests/server/worker-startup-liveness.test.ts
+++ b/integrationTests/server/worker-startup-liveness.test.ts
@@ -7,7 +7,7 @@
 import { suite, test, before, after } from 'node:test';
 import { ok, strictEqual } from 'node:assert';
 import { resolve, join } from 'node:path';
-import { readFile, mkdtemp, cp, rm } from 'node:fs/promises';
+import { mkdtemp, cp, rm } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { setupHarperWithFixture, teardownHarper, type ContextWithHarper } from '@harperfast/integration-testing';
 
@@ -46,19 +46,5 @@ suite('worker startup liveness (pre-ready event-loop drain)', (ctx: ContextWithH
 			headers: { Authorization: authorization },
 		});
 		strictEqual(response.status, 200, `expected the fixture resource to serve; got ${response.status}`);
-	});
-
-	test('no worker exited before reporting ready', async () => {
-		const logDir = ctx.harper.logDir ?? join(ctx.harper.dataRootDir, 'log');
-		let contents = '';
-		try {
-			contents = await readFile(join(logDir, 'hdb.log'), 'utf8');
-		} catch {
-			// no log file means nothing was logged — acceptable
-		}
-		ok(
-			!/exited with code \d+ before reporting ready/.test(contents),
-			`a worker exited before reporting ready:\n${contents.slice(-2000)}`
-		);
 	});
 });
--- a/integrationTests/server/worker-startup-liveness.test.ts
+++ b/integrationTests/server/worker-startup-liveness.test.ts
@@ -34,8 +34,11 @@ suite('worker startup liveness (pre-ready event-loop drain)', (ctx: ContextWithH
 	});
 
 	after(async () => {
-		await teardownHarper(ctx);
-		if (rootFixtureCopy) await rm(rootFixtureCopy, { recursive: true, force: true });
+		try {
+			await teardownHarper(ctx);
+		} finally {
+			if (rootFixtureCopy) await rm(rootFixtureCopy, { recursive: true, force: true });
+		}
 	});
 
 	test('workers survive a load-time await with no ref-holding completion source', async () => {
--- a/integrationTests/server/worker-startup-liveness.test.ts
+++ b/integrationTests/server/worker-startup-liveness.test.ts
@@ -1,21 +1,8 @@
 /**
- * Regression test for the pre-ready worker event-loop drain (the recurring CI failure
- * "Worker (index N) exited with code 0 before reporting ready" thrown from
- * socketRouter.ts's createWorkerReadyPromise).
- *
- * During loadRootComponents a worker's parentPort is unref'd (manageThreads.addPort) and is
- * only ref'd after component loading completes (threadServer.js). So if component load ever
- * awaits a completion that arrives via a non-ref'd source — an unref'd threadsafe function
- * (rocksdb-js parks an IsBusy commit on the conflicting worker's lock and wakes it through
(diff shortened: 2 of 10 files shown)


integrationTests/server/worker-startup-liveness.test.ts
```diff
@@ -0,0 +1,77 @@
...
+	});
+
+	test('no worker exited before reporting ready', async () => {
+		const logDir = ctx.harper.logDir ?? join(ctx.harper.dataRootDir, 'log');
+		let contents = '';
+		try {
+			contents = await readFile(join(logDir, 'hdb.log'), 'utf8');
+		} catch {
+			// no log file means nothing was logged — acceptable
+		}
```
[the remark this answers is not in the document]
kriszyp: This comment targets the 'no worker exited before reporting ready' log-grep test, which was removed in 490a99b4a (independent review found it could never fail on its own — startup failure already fails the suite in before()). — Claude Fable 5

unitTests/components/symlinkHarperModuleLockWait.test.js
```diff
@@ -0,0 +1,70 @@
...
+
+	afterEach(() => {
+		global.setTimeout = realSetTimeout;
+	});
+
+	after(() => {
+		rmSync(componentDirectory, { recursive: true, force: true });
+	});
+
+	const liveRefdLockWaitTimer = () => createdTimers.some(({ timer, ms }) => ms === 10_000 && timer.hasRef());
```
cb1kenobi: **Medium: this predicate passes against the unfixed implementation — the waiter test does not fail first**

`Timeout.hasRef()` stays `true` after `clearTimeout()`:

```
const t = setTimeout(() => {}, 10_000);
clearTimeout(t);
t.hasRef();   // true
```

The pre-fix bug on the waiter branch was `clearTimeout(timeout)`, not `.unref()`, so this predicate cannot see it.

Mutation-tested on this head: reverting `symlinkHarperModule` to the base implementation and rebuilding (`tsc -p tsconfig.build.json`, exactly one marker isolated in `dist/components/componentLoader.js`), `mocha unitTests/components/symlinkHarperModuleLockWait.test.js --timeout 30000` reports **1 passing, 1 failing** — only "the winner completes without leaving a timer armed" goes red. This test stays green. A second mutation that keeps the new structure but adds `timeout.unref()` does turn it red, which pins down exactly what it detects: an explicit unref, not the actual regression.

So the PR description's "run against the unmodified `symlinkHarperModule` (original source swapped in and rebuilt), both tests fail" does not hold — one of the two does. The winner test is a genuine fail-first guard; this one is not.

Suggested fix: track clears alongside creations. In the same `beforeEach`, wrap `global.clearTimeout` to record cleared timers in a `Set`, and make the predicate `ms === 10_000 && timer.hasRef() && !cleared.has(timer)`.

—
_Generated by Barber AI_

server/threads/threadServer.js
```diff
@@ -168,7 +168,29 @@ function closeServers() {
...
 function startServers() {
+	// A worker that has not yet posted child_started owns no ref'd handle: addPort()
+	// (manageThreads) unrefs parentPort, component watchers are persistent:false, and the
+	// reporting timers are unref'd. An await inside loadRootComponents whose completion
+	// arrives through a non-ref-holding source — e.g. rocksdb-js delivers cross-thread
+	// lock-release and parked-commit-retry wakes through unref'd threadsafe functions —
+	// can then drain the event loop, exiting the worker cleanly (code 0) before the ready
+	// handshake and aborting the whole node's startup. Hold a ref for the entire pre-ready
+	// window; the SHUTDOWN path unrefs it for graceful exit as before.
+	parentPort?.ref();
```
cb1kenobi, line 180: **Medium: no test fails when this fix is reverted — the threadServer half of the PR ships unguarded**

Mutation-tested on this head: replacing `parentPort?.ref()` with a no-op *and* disabling the `startupDeadline` arm, then rebuilding (`tsc -p tsconfig.build.json`; both markers verified isolated to `dist/server/threads/threadServer.js`), leaves `unitTests/server/threads/threadServerListenOnPorts.test.js`, `unitTests/server/threads/workerDataProviders.test.js` and `unitTests/components/symlinkHarperModuleLockWait.test.js` at **8/8 passing**. Nothing under `unitTests/` or `integrationTests/` references `startupDeadline` or `STARTUP_READY_TIMEOUT_MS`, and by the PR's own account `worker-startup-liveness.test.ts` passes against unmodified sources.

That leaves all three new production behaviours in this file — the pre-ready ref, the 10-minute deadline plus `realExit(1)`, and the load-failure `realExit(1)` — with no test that goes red when they are removed. `symlinkHarperModuleLockWait.test.js` covers only the `componentLoader` half.

The description already states the mechanism reproduces standalone: "a worker with an unref'd parentPort awaiting an unref'd-handle completion exits code 0 — the exact CI signature". That is a unit test, and the repo already has the harness shape for it in `unitTests/server/threads/fixtures/workerProcessGuard-fixture.js`: start a `Worker` from a fixture that unrefs its `parentPort`, awaits a completion delivered behind an unref'd handle, and only then posts a ready message; assert the worker's `exit` never fires before that message arrives. That test fails on the mutation above and passes with the fix — which is the fails-first evidence this half of the change is currently missing.

—
_Generated by Barber AI_

server/threads/threadServer.js
```diff
@@ -168,7 +168,29 @@ function closeServers() {
...
+	// The ref also means a load that HANGS (or a sync throw swallowed by the uncaughtException
+	// handler) would park this worker forever where it previously drained and failed loudly, so
+	// bound the pre-ready window. Deliberately far above restartWorkers' 60s replacement-worker
+	// backstop: timing out here aborts the whole node's startup, so a slow-but-succeeding boot
+	// must never be killed.
+	if (parentPort && !startupDeadline) {
+		startupDeadline = setTimeout(() => {
```
cb1kenobi: **Medium: one `clearTimeout` on one code path is all that keeps this timer from killing healthy workers**

The callback is unconditional — it logs fatal and calls `realExit(1)` without re-checking whether the worker ever became ready. Its entire safety rests on the single `clearTimeout(startupDeadline)` at line 267. If that call is lost or bypassed, every worker `realExit(1)`s at exactly 10 minutes of uptime with "did not become ready within 600000ms", and `manageThreads`' unexpected-exit handler (`manageThreads.js:437-445`) restarts each one — a rolling 10-minute worker-kill loop up to `MAX_UNEXPECTED_RESTARTS`.

That is not hypothetical for this file. [harper#2129 "Fix HTTP startup after pre-ready worker restart"](https://github.com/HarperFast/harper/pull/2129) is open and rewrites this exact chain: it lifts the ready block and the `CHILD_STARTED` post out of the `.then()` into a separate `started = loaded.then(() => listening).then(...)` chain, and adds its own terminal `.catch(... realExit(1))` that already covers a rejected `loadRootComponents()`. Whichever of the two lands second has to (a) re-place `clearTimeout(startupDeadline)` inside the restructured `started` chain and (b) drop one of the two now-duplicate load-failure fatal handlers. Step (a) is easy to miss in a conflict resolution, and missing it produces the failure above.

Suggested fix: make a missed clear degrade to a no-op rather than a kill. Add a `let workerIsReady = false;` next to `startupDeadline`, set it on the ready path beside the existing `clearTimeout`, and open this callback with `if (workerIsReady) return;`.

—
_Generated by Barber AI_

server/threads/threadServer.js
```diff
@@ -168,7 +168,29 @@ function closeServers() {
...
+			harperLogger.fatal(`Worker ${threadId} did not become ready within ${STARTUP_READY_TIMEOUT_MS}ms`);
+			realExit(1);
+		}, STARTUP_READY_TIMEOUT_MS);
```
cb1kenobi: **Low: `startupDeadline` is never cleared on the shutdown path, so a pre-ready worker can no longer exit by draining**

`clearTimeout(startupDeadline)` exists only on the ready path (line 267). The worker-side SHUTDOWN handler lives in `manageThreads.js:1251-1254` and does `parentPort.unref(); armSelfExit(threadTerminationTimeout)` — and `armSelfExit`'s timer is `.unref()`'d. threadServer's own SHUTDOWN listener is not registered until `loadRootComponents()` resolves, so for a worker that is still pre-ready nothing releases this ref'd 600s timer: it can no longer drain out, and instead sits until `armSelfExit` force-exits it (10s, or 30s under `DEV_MODE`), logging "Thread did not voluntarily terminate".

Reachable on `harper stop` / `harper restart` issued while a worker is still booting, and on dev-mode component-reload restarts that land mid-boot. Not a regression against a *correct* prior behaviour — before this PR a busy pre-ready worker also waited for the same backstop — but the deadline removes the one case that used to exit promptly, for no benefit during a shutdown.

Suggested fix: release it on shutdown as well. `onMessageByType` is already exported from `manageThreads`, so alongside the arm above:

```js
onMessageByType(terms.ITC_EVENT_TYPES.SHUTDOWN, () => clearTimeout(startupDeadline));
```

registered early (not inside the post-load `.then()`), so a pre-ready SHUTDOWN lets the worker drain out as it did before.

—
_Generated by Barber AI_

server/threads/threadServer.js
```diff
@@ -168,7 +168,29 @@ function closeServers() {
 	return Promise.all(promises);
 }
 
+const STARTUP_READY_TIMEOUT_MS = 600_000;
```
cb1kenobi: **Low: the pre-ready deadline is a hardcoded magic number with no config knob**

Answering the question raised in the PR description: aligning this with the 60s replacement-worker backstop would be wrong — the reasoning in the comment above is right that a legitimately slow boot (large component installs, a cold `npm install` on a Fabric container) must not be killed. But 600000 as a bare literal is the wrong end state either way: every other startup/shutdown bound in this area is either derived (`threadTerminationTimeout * 2`, `Math.max(threadTerminationTimeout * 2, 60000)`) or configurable (`getShutdownDrainCeilingMs()`), and this one is the only bound whose expiry aborts the *whole node's* startup.

Suggested fix: source it from config the way the shutdown ceiling is, with 600000 as the default — an operator whose components genuinely take longer than 10 minutes to install currently has no way out except a code change, and one who wants a faster CI failure has no way to shorten it.

—
_Generated by Barber AI_

components/componentLoader.ts
```diff
@@ -446,6 +451,8 @@ function symlinkHarperModule(componentDirectory: string) {
 	});
 }
 
+export const _symlinkHarperModuleForTests = symlinkHarperModule;
```
cb1kenobi, line 455: **Nit: test-only export on a production module**

`symlinkHarperModule` is otherwise private to this module; this line makes it part of `componentLoader`'s public surface purely so `unitTests/components/symlinkHarperModuleLockWait.test.js` can reach it, with only the underscore prefix marking the intent.

The lock behaviour under test is entirely a function of `Status.primaryStore.tryLock`, so the test could drive it through the public entry point that already calls it (a component load against a temp directory) without widening the surface. If the direct handle is worth keeping for the determinism it buys, that trade is fine — worth a one-line comment here saying so, since nothing at this line explains why the export exists.

—
_Generated by Barber AI_

dawsontoth (approved):

server/threads/threadServer.js
kriszyp: Fixed in df2e208eb/945f40ea8: added a standalone Worker regression that executes the real `threadServer.startServers()` path with a cache-stubbed component load whose asserted-unref'd timer is its only completion source. Replacing the startup `parentPort.ref()` with `unref()` produces the exact `worker exited with code 0 before reporting ready` failure.

unitTests/components/symlinkHarperModuleLockWait.test.js
kriszyp: Fixed in 945f40ea8: the test now tracks `clearTimeout()` calls and excludes cleared timers from the live ref-holding predicate. The mutation check now fails this assertion when the waiter timer is cleared.

components/componentLoader.ts
kriszyp: Fixed in df2e208eb: retained the deterministic direct seam and documented why it avoids exposing a broader component-load seam.

server/threads/threadServer.js
kriszyp: Resolved by removing the startup deadline in df2e208eb, per the owner decision. This PR no longer introduces a fixed or configurable startup-timeout policy.

server/threads/threadServer.js
kriszyp: Resolved by the owner-directed removal of the startup deadline in df2e208eb. There is no delayed fatal callback left that can kill an already-ready worker.

server/threads/threadServer.js
kriszyp: Resolved by the owner-directed removal of the startup deadline in df2e208eb. Pre-ready shutdown no longer has a deadline timer to clear.

cb1kenobi (approved):

PR #2318, opened 2026-08-25
Pin the dispatched test workflows to a read-only token on v5.2

kriszyp:
`cherry-pick-patch.yml` dispatches this repo's test workflows with `workflow_dispatch --ref cherry-pick/<release>/pr-<N>` — a branch built from a release branch plus a PR's commits. Those runs therefore execute PR-authored build and test scripts inside the base repo.

`workflow_dispatch` resolves `permissions:` from the workflow file **on the dispatched ref**, so a pin added on `main` does nothing for a dispatch against this branch. `default_workflow_permissions` is `write` on this repo, which means every cherry-pick test run on this release line currently executes PR-authored scripts under a write-scoped `GITHUB_TOKEN`.

This pins `contents: read` here, where the dispatch actually resolves it.

Pre-existing exposure, not introduced by any current change — but it is being widened: HarperFast/harper#2317 moves release targeting from the opt-in `patch` label to the PR Milestone, which is set on nearly every triaged PR. That PR adds the same pin on `main`; this one covers the branch the dispatch resolves against.

## Verification

Neither workflow references `secrets.` or `GITHUB_TOKEN` — confirmed by grep on this branch — so nothing in them needs write scope. `actions/upload-artifact` uses the Actions runtime token, which `permissions:` does not govern, so artifact upload is unaffected. YAML parses with the block in place.

The change is additive and scoped to a single top-level key; the existing CI on this PR exercises it.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -19,6 +19,11 @@ on:
           - '24'
           - '26'
 
+# Dispatched by cherry-pick-patch.yml against a branch built from PR commits, so this
+# runs PR-authored build/test scripts. The repo default is a write-scoped token.
+permissions:
+  contents: read
+
 jobs:
   generate-node-version-matrix:
     name: Generate Node Version Matrix
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -17,6 +17,11 @@ on:
           - '24'
           - '26'
 
+# Dispatched by cherry-pick-patch.yml against a branch built from PR commits, so this
+# runs PR-authored build/test scripts. The repo default is a write-scoped token.
+permissions:
+  contents: read
+
 jobs:
   unit-test:
     name: Unit Test (Node.js v${{ matrix.node-version }})


kriszyp:
CI triage: the two red checks (`Integration Tests 1/6` and `6/6`, Windows / Node.js v24) were **not** caused by this diff. `permissions:` only scopes `GITHUB_TOKEN`, and neither workflow reads it — there is no mechanism by which the pin can change a test outcome.

What actually failed, and why it's pre-existing:

| run | shard | failure |
|---|---|---|
| [32873310380](https://github.com/HarperFast/harper/actions/runs/32873310380) (this PR, attempt 1) | 6/6 | `Configuration` — `EPERM: operation not permitted, rename 'harper-config.yaml.<pid>.<n>.<hash>.tmp' -> 'harper-config.yaml'` (Windows atomic rename over a live handle) |
| same | 1/6 | `terminology.test.mjs` → `drop_database with database param` hangs 11s, server process dies, every later test in the file gets `ECONNREFUSED` |
| [32807734000](https://github.com/HarperFast/harper/actions/runs/32807734000) (**v5.2 branch**, commit 0b204ffd5 — predates this PR) | 1/6 | identical `drop_database` → `ECONNREFUSED` cascade |

The last three pushes to `v5.2` each failed a *different* shard on their own: 3/6 (Node 22), 1/6 (Windows), 2/6 (uWS HTTP). The base branch is broadly flaky right now.

Reruns confirm it. Attempt 2: 6/6 passed; 1/6 failed in a **different** place — `risk-query.test.ts` first `fetch` never got headers (`UND_ERR_HEADERS_TIMEOUT`, 300s) against a server that had logged a successful start. Attempt 3: green. All 40 checks now pass, with no change to the diff.

The Windows shard-1 instability is worth its own issue (three distinct failure modes in three attempts, all Windows-only), but it does not belong in a release-branch permissions pin.

— KrAIs (Claude Opus 5)

Issue #88, opened 2026-01-13
Fix: Double symlink issue doesn't throw an error


BboyAkers:
### Bug Summary

When a person links their app in the `harperdb-config.yaml` and also run `harper run .` inside their app. The app does not update and errors out

### Reproduction Repository (if applicable)

https://github.com/BboyAkers/harper-invoice-app/

### Steps to Reproduce

1. Clone the app
2. Run `npm install` inside the app
3. Open `harperdb-config.yaml` and add the app at the bottom
5. Run `npm run dev` inside your app
6. The app errors out and doesn't run in `localhost:9926`

### Expected Behavior

1. Clone the app
2. Run `npm install` inside the app
3. Open `harperdb-config.yaml` and add the app at the bottom
5. Run `npm run dev` inside your app
6. Harper should throw an error

### Actual Behavior

1. Clone the app
2. Run `npm install` inside the app
3. Open `harperdb-config.yaml` and add the app at the bottom
5. Run `npm run dev` inside your app
6. Harper runs and the app errors out

### Are you planning to fix this issue?

Maybe, depending on complexity

### First-time contributor support

- [ ] I'm new to contributing and would appreciate guidance on the process
- [ ] I'd like help understanding the project structure
- [ ] I need assistance with setting up the development environment
- [ ] I'm comfortable contributing, but new to this project specifically

Issue #85, opened 2026-01-12
Fix `test:config` failures


kriszyp:
The `test:config` unit tests in https://github.com/HarperFast/harper/pull/75 have test failures that need to be fixed.

PR #2328, opened 2026-08-25
ci: adopt ai-review-prompts #89 — sonnet-5/xhigh/96-turn review defaults

heskew:
Bumps every ai-review-prompts pin (claude-review, gemini-review, claude-mention, claude-issue-to-pr, validate-caller-workflows) to `be549ad` — current main, which carries:

- [ai-review-prompts#90](https://github.com/HarperFast/ai-review-prompts/pull/90): **cost gates** — draft PRs skip review until flipped ready (label opts a draft in), mechanical diffs skip before the agent runs, reasoning effort scales with diff size (≤60 → high, ≤1500 → xhigh, larger → max), synchronize runs debounce 120s. Both review callers gain the `ready_for_review` trigger type.
- [#89](https://github.com/HarperFast/ai-review-prompts/pull/89): fleet defaults — claude-sonnet-5, 96-turn review budget, 30m timeouts.
- [#88](https://github.com/HarperFast/ai-review-prompts/pull/88): sibling-implementations and cannot-fail-test lenses; [#85](https://github.com/HarperFast/ai-review-prompts/pull/85)/[#87](https://github.com/HarperFast/ai-review-prompts/pull/87): calibration prompt edits.

**Expected on this PR**: the claude-review check errors with the workflow-must-match-default-branch validation — normal for workflow-editing PRs.

**Merge order**: this PR first, then #2330 and #2331 (the release-branch syncs, refreshed to matching content).

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -33,7 +33,7 @@ jobs:
     # Note: the `claude-review` label name is matched there too —
     # `_claude-review.yml`'s authorize `if:`, not in this caller.
     if: ${{ vars.CLAUDE_ALWAYS_ON == 'true' || github.event.action == 'labeled' }}
-    uses: HarperFast/ai-review-prompts/.github/workflows/_claude-review.yml@8401fe113e229d8842283c67e99457b862ab3b89 # main 2026-08-25 (#89 fleet turn/effort/model defaults: sonnet-5, xhigh, 96 turns)
+    uses: HarperFast/ai-review-prompts/.github/workflows/_claude-review.yml@4632c5d93dae1e6548e6c30c5ced8fb2df69f4e5 # main 2026-08-25 (#89 fleet turn/effort/model defaults; #88 sibling + cannot-fail-test lenses; #85/#87 calibration)
     # Caller-side permissions at the calling-job level (NOT workflow-
     # level — that placement caps the reusable's per-job grants below
     # what they need and breaks the workflow at startup; see
@@ -60,7 +60,7 @@ jobs:
       # introspect their own ref (`github.workflow_ref` resolves to the
       # CALLER's ref in `workflow_call` context), and `uses: …@<ref>`
       # is parsed literally so we can't interpolate a variable.
-      ai-review-prompts-ref: 8401fe113e229d8842283c67e99457b862ab3b89
+      ai-review-prompts-ref: 4632c5d93dae1e6548e6c30c5ced8fb2df69f4e5
       review-layers: |
         universal
         harper/common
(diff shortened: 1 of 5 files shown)


.github/workflows/claude-review.yml
```diff
@@ -18,7 +18,7 @@ on:
   pull_request:
     # `labeled` admits the `claude-review` label gesture for
     # bot-authored PRs (renovate, dependabot). See ai-review-prompts#38.
-    types: [opened, synchronize, reopened, labeled]
+    types: [opened, synchronize, reopened, ready_for_review, labeled]
```
kriszyp, line 21: The outer gate at `.github/workflows/claude-review.yml:35` rejects this new event when `CLAUDE_ALWAYS_ON` is unset. If `claude-review` is applied while the PR is a draft, the `labeled` run reaches the reusable but its new draft gate skips review; when the PR becomes ready, this event fires but the caller job evaluates false. Please admit `ready_for_review` when the PR still carries the `claude-review` label so the documented opt-in flow actually resumes.

.github/workflows/gemini-review.yml
```diff
@@ -32,7 +32,7 @@ on:
     # `labeled` admits the `gemini-review` opt-in gesture. `vars.*`
     # can't be read in `on:` (only in a job `if:`), so the trigger
     # lists the union and the `review` job gates on GEMINI_ALWAYS_ON.
-    types: [opened, synchronize, reopened, labeled]
+    types: [opened, synchronize, reopened, ready_for_review, labeled]
```
kriszyp, line 35: The outer gate at `.github/workflows/gemini-review.yml:54` rejects `ready_for_review` unless `GEMINI_ALWAYS_ON` is true. In the documented default opt-in mode, applying `gemini-review` to a draft produces a skipped draft run, then converting it to ready cannot reach the reusable workflow. Please allow this event when the PR retains the `gemini-review` label.

cb1kenobi:
Reviewed `f9351d54` (base `main`) — no issues found. This PR looks good, nice job!

Verified rather than assumed:

—
_Generated by Barber AI_

kriszyp (approved):

.github/workflows/claude-review.yml
heskew: Fixed in f25a92b — the gate now admits `ready_for_review` when the PR still carries the `claude-review` label, so a label-opted draft resumes review on flipping ready with CLAUDE_ALWAYS_ON unset. (One trace note: the labeled-while-draft run itself does review — the reusable's draft gate has a labeled-event escape — the dead path was exactly the ready-flip you flagged.) Applied to the gemini caller and the release-branch sync PRs too.

.github/workflows/gemini-review.yml
heskew: Fixed in f25a92b — same `ready_for_review` + `gemini-review`-label admission on the gemini caller; mirrored across the sync PRs and harper-pro's set.

PR #2330, opened 2026-08-25
ci(v5.1): sync bot caller workflows with main

heskew:
claude-code-action validates a PR's caller workflows against the **default branch**, so drifted copies on `v5.1` fail the review check on v5.1-targeted PRs (v5.1 has been failing every review this way — e.g. #2319 today). Content matches #2328 — byte-identical to main once #2328 merges, so **merge #2328 first**, then this.

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -21,10 +21,12 @@ on:
     types: [opened, synchronize, reopened, ready_for_review, labeled]
 
 concurrency:
+  # NOTE: this predicate and the review job's `if:` must stay textually
+  # in sync — GitHub Actions YAML has no anchors to share them.
   # Ineligible events (per the review job's opt-in gate) get a unique
   # run_id group so they can never cancel an in-flight eligible review —
   # workflow-level cancel-in-progress fires before job `if:` is evaluated.
-  group: claude-review-${{ github.event.pull_request.number }}-${{ (vars.CLAUDE_ALWAYS_ON == 'true' || github.event.action == 'labeled' || (github.event.action == 'ready_for_review' && contains(github.event.pull_request.labels.*.name, 'claude-review'))) && 'eligible' || github.run_id }}
+  group: claude-review-${{ github.event.pull_request.number }}-${{ ((github.event.action == 'labeled' && github.event.label.name == 'claude-review') || (github.event.action != 'labeled' && vars.CLAUDE_ALWAYS_ON == 'true') || (github.event.action == 'ready_for_review' && contains(github.event.pull_request.labels.*.name, 'claude-review'))) && 'eligible' || github.run_id }}
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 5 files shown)


.github/workflows/claude-review.yml
```diff
@@ -18,7 +18,7 @@ on:
   pull_request:
     # `labeled` admits the `claude-review` label gesture for
     # bot-authored PRs (renovate, dependabot). See ai-review-prompts#38.
-    types: [opened, synchronize, reopened, labeled]
+    types: [opened, synchronize, reopened, ready_for_review, labeled]
```
kriszyp, line 21: Could we extend the caller-side gate when adding this event, and mirror that in `.github/workflows/gemini-review.yml:35`? With `CLAUDE_ALWAYS_ON` unset, applying `claude-review` to a draft emits `labeled`, but the new reusable pin skips the draft. When the PR later emits `ready_for_review`, the workflow starts, but `.github/workflows/claude-review.yml:35` evaluates false because the action is no longer `labeled`; no review is dispatched while the opt-in label remains. Gemini has the same failure at `.github/workflows/gemini-review.yml:54`. Please admit `ready_for_review` when the provider’s review label is present (and ensure the reusable authorization accepts that event), or explicitly require re-labeling after the PR leaves draft.

cb1kenobi:
Reviewed `3cc4afd4` (base `v5.1`) — no issues found. This PR looks good, nice job!

Verified rather than assumed:

—
_Generated by Barber AI_

kriszyp (commented):

.github/workflows/claude-review.yml
heskew: Already addressed — this branch picked up the same fix as #2328 (commit 1206e46 here): both callers now admit ready_for_review when the opt-in label is present, and the branch remains byte-identical to post-#2328 main (which the caller validation requires).

.github/workflows/claude-review.yml
```diff
@@ -18,10 +18,13 @@ on:
...
 
 concurrency:
-  group: claude-review-${{ github.event.pull_request.number }}
+  # Ineligible events (per the review job's opt-in gate) get a unique
+  # run_id group so they can never cancel an in-flight eligible review —
+  # workflow-level cancel-in-progress fires before job `if:` is evaluated.
+  group: claude-review-${{ github.event.pull_request.number }}-${{ (vars.CLAUDE_ALWAYS_ON == 'true' || github.event.action == 'labeled' || (github.event.action == 'ready_for_review' && contains(github.event.pull_request.labels.*.name, 'claude-review'))) && 'eligible' || github.run_id }}
```
kriszyp: Could this `labeled` branch check the provider label itself, and mirror that at `.github/workflows/gemini-review.yml:43`? In opt-in mode, adding an unrelated label such as `bug` makes this expression choose `eligible`, so workflow-level `cancel-in-progress` cancels an in-flight Claude review. The reusable workflow then rejects the replacement because the event label is not `claude-review`, leaving no completed review. Please require `github.event.label.name == 'claude-review'` here (and `gemini-review` for Gemini), or use an equivalent exact-label predicate, so every run sharing the cancellable group can actually replace the run it cancels.

kriszyp (approved):

.github/workflows/claude-review.yml
heskew: Taken in fa04115 — the labeled branch of the predicate now requires `github.event.label.name == '<provider>-review'`, in both the concurrency group and the review job gate, for both callers. Unrelated-label events are ineligible in BOTH modes — tracing this also showed the unrelated-label cancellation predated this series in always-on mode (any labeled event shared the group, then authorize skipped), so this closes that too. Full event matrix is in the commit message. Applied to this sync branch in its own commit series; byte-identity with post-#2348 main holds.

PR #2331, opened 2026-08-25
ci(v5.2): sync bot caller workflows with main

heskew:
claude-code-action validates a PR's caller workflows against the **default branch**, so drifted copies on `v5.2` fail the review check on v5.2-targeted PRs (v5.1 has been failing every review this way — e.g. #2319 today). Content matches #2328 — byte-identical to main once #2328 merges, so **merge #2328 first**, then this.

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -21,10 +21,12 @@ on:
     types: [opened, synchronize, reopened, ready_for_review, labeled]
 
 concurrency:
+  # NOTE: this predicate and the review job's `if:` must stay textually
+  # in sync — GitHub Actions YAML has no anchors to share them.
   # Ineligible events (per the review job's opt-in gate) get a unique
   # run_id group so they can never cancel an in-flight eligible review —
   # workflow-level cancel-in-progress fires before job `if:` is evaluated.
-  group: claude-review-${{ github.event.pull_request.number }}-${{ (vars.CLAUDE_ALWAYS_ON == 'true' || github.event.action == 'labeled' || (github.event.action == 'ready_for_review' && contains(github.event.pull_request.labels.*.name, 'claude-review'))) && 'eligible' || github.run_id }}
+  group: claude-review-${{ github.event.pull_request.number }}-${{ ((github.event.action == 'labeled' && github.event.label.name == 'claude-review') || (github.event.action != 'labeled' && vars.CLAUDE_ALWAYS_ON == 'true') || (github.event.action == 'ready_for_review' && contains(github.event.pull_request.labels.*.name, 'claude-review'))) && 'eligible' || github.run_id }}
   cancel-in-progress: true
 
 jobs:
@@ -38,7 +40,7 @@ jobs:
     # `ready_for_review` is admitted when the PR still carries the
     # opt-in label, so a draft opted in via `claude-review` resumes
     # review when it flips ready even with CLAUDE_ALWAYS_ON unset.
-    if: ${{ vars.CLAUDE_ALWAYS_ON == 'true' || github.event.action == 'labeled' || (github.event.action == 'ready_for_review' && contains(github.event.pull_request.labels.*.name, 'claude-review')) }}
+    if: ${{ (github.event.action == 'labeled' && github.event.label.name == 'claude-review') || (github.event.action != 'labeled' && vars.CLAUDE_ALWAYS_ON == 'true') || (github.event.action == 'ready_for_review' && contains(github.event.pull_request.labels.*.name, 'claude-review')) }}
     uses: HarperFast/ai-review-prompts/.github/workflows/_claude-review.yml@be549ad08aa6d34b909ea8b542a7ffebdaae1e81 # main 2026-08-25 (#90 cost gates: draft skip, mechanical-diff skip, effort-by-size, debounce; #89 defaults; #88 lenses)
     # Caller-side permissions at the calling-job level (NOT workflow-
     # level — that placement caps the reusable's per-job grants below
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -21,12 +21,15 @@ on:
     types: [opened, synchronize, reopened, ready_for_review, labeled]
 
 concurrency:
-  # NOTE: this predicate and the review job's `if:` must stay textually
-  # in sync — GitHub Actions YAML has no anchors to share them.
+  # NOTE: this predicate is deliberately NARROWER than the review job's
+  # `if:` — the cancelling set must be a subset of the running set.
+  # ready_for_review runs review (job gate) but never cancel: their
+  # authorization is author-based, so on a bot-authored PR they could
+  # kill a labeler-authorized run and then be rejected themselves.
   # Ineligible events (per the review job's opt-in gate) get a unique
   # run_id group so they can never cancel an in-flight eligible review —
   # workflow-level cancel-in-progress fires before job `if:` is evaluated.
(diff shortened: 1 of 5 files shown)


.github/workflows/claude-review.yml
```diff
@@ -18,7 +18,7 @@ on:
   pull_request:
     # `labeled` admits the `claude-review` label gesture for
     # bot-authored PRs (renovate, dependabot). See ai-review-prompts#38.
-    types: [opened, synchronize, reopened, labeled]
+    types: [opened, synchronize, reopened, ready_for_review, labeled]
```
kriszyp, line 21: The new reusable pin skips draft PRs, but this caller still runs only for `labeled` events unless `CLAUDE_ALWAYS_ON` is true (`.github/workflows/claude-review.yml:35`). If `claude-review` is added while the PR is draft, that run is skipped; when the PR becomes ready, this event fires but the caller filters out the job, so the requested review never happens. Please admit `ready_for_review` when the PR already has the `claude-review` label (with matching reusable authorization), or centralize event/label admission in the reusable workflow so these gates cannot disagree.

.github/workflows/gemini-review.yml
```diff
@@ -32,7 +32,7 @@ on:
     # `labeled` admits the `gemini-review` opt-in gesture. `vars.*`
     # can't be read in `on:` (only in a job `if:`), so the trigger
     # lists the union and the `review` job gates on GEMINI_ALWAYS_ON.
-    types: [opened, synchronize, reopened, labeled]
+    types: [opened, synchronize, reopened, ready_for_review, labeled]
```
kriszyp, line 35: The same draft-to-ready gap exists here: with `GEMINI_ALWAYS_ON` unset, `.github/workflows/gemini-review.yml:54` rejects `ready_for_review`. A `gemini-review` label applied while draft triggers a run that the new reusable workflow skips, then marking the PR ready does not retry it. Please admit this event when the existing `gemini-review` label is present, or consolidate the admission logic in the reusable workflow.

cb1kenobi:
Reviewed `b705a473` (base `v5.2`) — no issues found. This PR looks good, nice job!

Verified rather than assumed:

—
_Generated by Barber AI_

kriszyp (commented):

.github/workflows/claude-review.yml
heskew: Already addressed — same fix as #2328 landed here (commit cedfc1a): the gate admits ready_for_review when the claude-review label is present; branch stays byte-identical to post-#2328 main.

.github/workflows/gemini-review.yml
heskew: Same fix covers the gemini caller in cedfc1a (gemini-review label admission). On consolidating the admission logic into the reusable: not expressible today — the opt-in gate must run caller-side because vars.* and the label-presence check gate whether the reusable is invoked at all; noted for a future pass if GitHub allows vars in workflow_call context.

.github/workflows/claude-review.yml
```diff
@@ -18,10 +18,13 @@ on:
...
 
 concurrency:
-  group: claude-review-${{ github.event.pull_request.number }}
+  # Ineligible events (per the review job's opt-in gate) get a unique
+  # run_id group so they can never cancel an in-flight eligible review —
+  # workflow-level cancel-in-progress fires before job `if:` is evaluated.
+  group: claude-review-${{ github.event.pull_request.number }}-${{ (vars.CLAUDE_ALWAYS_ON == 'true' || github.event.action == 'labeled' || (github.event.action == 'ready_for_review' && contains(github.event.pull_request.labels.*.name, 'claude-review'))) && 'eligible' || github.run_id }}
```
kriszyp: `github.event.action == 'labeled'` matches every label, while the reusable workflow later decides whether it was specifically `claude-review` (`.github/workflows/claude-review.yml:35-37`). In opt-in mode, applying an unrelated label during an active Claude review therefore joins this `eligible` group and cancels that review; the replacement run is then rejected by reusable authorization, leaving no completed review. Please classify a labeled run as eligible only when `github.event.label.name == 'claude-review'` (and consider using the same exact-label predicate in the job gate to avoid starting the rejected run).

.github/workflows/gemini-review.yml
```diff
@@ -32,14 +32,15 @@ on:
...
 
 concurrency:
   # Different group key from claude-review so the two providers can run
   # in parallel on the same PR. cancel-in-progress is per-group, so a
   # synchronize push cancels the in-flight Gemini run without touching
   # the Claude run (and vice versa).
-  group: gemini-review-${{ github.event.pull_request.number }}
+  # Ineligible events get a unique run_id group — see claude-review.yml.
+  group: gemini-review-${{ github.event.pull_request.number }}-${{ (vars.GEMINI_ALWAYS_ON == 'true' || github.event.action == 'labeled' || (github.event.action == 'ready_for_review' && contains(github.event.pull_request.labels.*.name, 'gemini-review'))) && 'eligible' || github.run_id }}
```
kriszyp: This still classifies every `labeled` event as eligible, even though reusable authorization accepts only the `gemini-review` gesture. If an unrelated label is applied while a Gemini review is running in opt-in mode, this run cancels it and is then rejected, so neither run produces a review. Please require `github.event.label.name == 'gemini-review'` for the labeled branch of this concurrency predicate; mirroring that exact check in `.github/workflows/gemini-review.yml:57` would also avoid the unnecessary workflow call.

kriszyp (approved):

.github/workflows/claude-review.yml
kriszyp: `github.event.action == 'labeled'` matches every label, while the reusable workflow later decides whether it was specifically `claude-review` (`.github/workflows/claude-review.yml:35-37`). In opt-in mode, applying an unrelated label during an active Claude review therefore joins this `eligible` group and cancels that review; the replacement run is then rejected by reusable authorization, leaving no completed review. Please classify a labeled run as eligible only when `github.event.label.name == 'claude-review'` (and consider using the same exact-label predicate in the job gate to avoid starting the rejected run).

.github/workflows/gemini-review.yml
kriszyp: This still classifies every `labeled` event as eligible, even though reusable authorization accepts only the `gemini-review` gesture. If an unrelated label is applied while a Gemini review is running in opt-in mode, this run cancels it and is then rejected, so neither run produces a review. Please require `github.event.label.name == 'gemini-review'` for the labeled branch of this concurrency predicate; mirroring that exact check in `.github/workflows/gemini-review.yml:57` would also avoid the unnecessary workflow call.

kriszyp (approved):

.github/workflows/claude-review.yml
heskew: Taken in fa04115 — the labeled branch of the predicate now requires `github.event.label.name == '<provider>-review'`, in both the concurrency group and the review job gate, for both callers. Unrelated-label events are ineligible in BOTH modes — tracing this also showed the unrelated-label cancellation predated this series in always-on mode (any labeled event shared the group, then authorize skipped), so this closes that too. Full event matrix is in the commit message. Applied here too; merge after #2348.

.github/workflows/gemini-review.yml
heskew: Gemini caller covered in the same commit — labeled branch requires gemini-review, mirrored in the job gate.

.github/workflows/claude-review.yml
heskew: Crossed in flight — 0715e9f (pushed minutes after this review) makes the labeled arm require github.event.label.name == 'claude-review' in both the concurrency predicate and the job gate; the branch is byte-identical to post-#2348 main, which carries the same fix.

.github/workflows/gemini-review.yml
heskew: Same commit (0715e9f) covers the gemini caller — labeled arm requires the gemini-review label in both the concurrency predicate and the job gate.

Issue #805, opened 2026-05-26
Create withHarper() Next.js config utility


kriszyp:
## Request

Provide a `withHarper(nextConfig)` utility that wraps the user's Next.js config and applies all necessary adjustments to make Harper work correctly (e.g., webpack externals, server external packages). Reduces boilerplate and keeps integration details out of user code.

```js
// next.config.js
const { withHarper } = require('harperdb/next');
module.exports = withHarper({ /* next config */ });
```

Related: CORE-2601 (serverExternalPackages investigation)

🤖 Filed by Claude on behalf of Kris.

Ethan-Arrowood:
Done in `@harperfast/nextjs@2`

Issue #799, opened 2026-05-26
Support HTTP Range headers for blob/buffer responses


kriszyp:
## Request

Implement HTTP `Range` header support for blob and buffer responses. Binary data can be efficiently sliced at the byte level, enabling:
- Partial content responses (HTTP 206)
- Resume-able downloads
- Streaming media (audio/video seeking)

🤖 Filed by Claude on behalf of Kris.

Issue #744, opened 2026-05-22
HTTP Logs showing up in system (non external) logs


Devin-Holland:
A lot of HTTP logs are showing up in the system log and thus shipped to datadog. Looking at the harper config there's not http logging enabled, so I believe somehow the app is logging and making it into the system logs. 

Whether or not http logs enabled should go to system or not I think would be another question.

Issue #741, opened 2026-05-22
Hash long URLs into primary key for cache table storage efficiency


kriszyp:
## Request

For caching use cases, URLs used as primary keys can be very long, increasing storage overhead and comparison cost. Hashing the URL before storing it as the primary key would reduce key size and improve cache table performance.

🤖 Filed by Claude on behalf of Kris.

PR #2338, opened 2026-08-26
Apply audit retention continuously to RocksDB transaction logs

### LINKED ISSUE #846: Transaction-log retention has no config-driven time/size control (and the storageReclamation hardening isn't on v5.1)

### LINKED ISSUE #2140: RocksDB transaction-log pruning never re-arms — one purge at boot, then nothing until disk pressure
## Summary

Under the RocksDB engine, `scheduleAuditCleanup` runs a single synchronous `purgeLogs` and returns — the self-rescheduling `setTimeout` loop below the early return is only reachable on the LMDB path. A long-running node therefore prunes its transaction logs exactly once (at store open, last worker only) and then never again unless free disk drops below 40% (the hourly storage-reclamation trigger).

Verified at [v5.2.1 `resources/auditStore.ts#L178-L186`](https://github.com/HarperFast/harper/blob/v5.2.1/resources/auditStore.ts#L178-L186); identical on main.

## Impact

On a replication-heavy cluster, `system` and per-origin transaction logs grow unboundedly between restarts. Observed on an internal 16-node test cluster: multi-GB `database/system/transaction_logs` growth that then amplified an unclean-shutdown replay into OOM crash loops. `logging.auditRetention` (default 3d) is effectively decorative on RocksDB in steady state.

Additional gating that compounds it: the boot purge runs on the last worker only (`auditStore.ts#L263`), the main thread never registers a reclamation handler, and `purgeLogs` may only delete files wholly before the last-flushed position — a crash-looping node that never reaches a flush may purge nothing (unverified against the native binary; worth confirming).

## Suggested fix

Re-arm the purge on RocksDB the same way the LMDB path does (periodic timer honoring `auditRetention`), or hook it into the existing hourly reclamation timer unconditionally rather than only under the 40% free-disk threshold.

---

kriszyp:
[RocksDB audit retention now runs on the existing self-rearming cleanup cadence](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-db526bfa2603e0ee94ab17a9ea8c2b8bd02e1f626dd6907624cfe8508ad356baR232) instead of depending on startup or disk-pressure signals, so eligible transaction-log segments are reclaimed continuously according to `logging.auditRetention`. The Rocks delay is a pure function of the pressure-adjusted retention window (a tenth of it, floored at `DEFAULT_AUDIT_CLEANUP_DELAY`) rather than LMDB's adaptive backoff: segment eligibility only changes on rotation/flush, so a per-entry delete count would only make it rescan the same files.

Fixes #846 — [Transaction-log retention has no config-driven time/size control (and the storageReclamation hardening isn't on v5.1)](https://github.com/HarperFast/harper/issues/846). Also fixes #2140, closed as a duplicate.

Retiring the loop is two things, and teardown needs both. [`stopAuditCleanup()`](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-db526bfa2603e0ee94ab17a9ea8c2b8bd02e1f626dd6907624cfe8508ad356baR353) latches the loop closed *and* returns a drain barrier — the promise that settles once the pass already running has finished. The barrier is what makes closing safe: lmdb-js stamps the DBI number into its write instruction synchronously and the native writer consumes it later, so a pass suspended inside `await removeAuditEntry()` still has a delete pending against the primary and audit DBIs, and LMDB forbids closing a DBI an open transaction has modified. [`dropDatabase()`](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1797) and the legacy arm of `Table.dropTable()` await it. `closeDatabase()` and branch `close()` are synchronous and cannot; what covers them is that [every environment touch remaining in a resumed pass](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-db526bfa2603e0ee94ab17a9ea8c2b8bd02e1f626dd6907624cfe8508ad356baR186) — cursor advance, cursor release, marker write, re-arm — re-checks `rootStore.status`, plus the fact that their production callers reach them only for RocksDB stores, whose pass is one synchronous `purgeLogs()` call with nothing suspended mid-removal. `resetDatabases()` closes LMDB roots with no retirement call at all, so that re-check is a routine path rather than a defensive one.

[The legacy per-table drop](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R1706) retires the audit store the table actually holds. It called `primaryStore.auditStore?.stopAuditCleanup?.()`, which nothing assigns — `openAuditStore()` assigns `rootStore.auditStore` and `makeTable()` captures the store it was handed — so the call was a no-op against a dead reference. `closeDatabase` and `dropDatabase` also deregister storage reclamation now: only `closeBranchHandles` did, so a closed or dropped store left a handler, and the closure pinning the store, registered for the life of the process, with every re-open of the same path appending another.

[The last-removed marker is retained until it commits](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-db526bfa2603e0ee94ab17a9ea8c2b8bd02e1f626dd6907624cfe8508ad356baR303). It used to be a fire-and-forget `put()`, and a pass that deletes nothing never reaches the write again, so one transient failure would leave the recorded boundary permanently behind the entries already removed. It is now awaited inside error containment and carried to the next pass. The write that *initializes* a new audit store has no downstream owner at all — `openAuditStore()` is synchronous and returns the store — so its rejection is contained at the call site, through the same `warnContained()` helper as the pass's outermost backstop, because a failing log sink would otherwise turn one contained rejection into an uncontained one.

[The pin moves to `@harperfast/rocksdb-js` 2.8.0](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R183), which is a prerequisite rather than hygiene. The previously pinned 2.7.1 already refuses to purge above the durable flush floor, but at the boundary segment it compares `logFile->size` without resolving the extent: a lazily-registered segment reads 0, so `0 > positionInLogFile` is false, the guard passes, and the segment is deleted with a tail that never reached RocksDB. 2.8.0 calls `ensureExtent()` there and refuses to purge when the extent cannot be resolved; its `should not purge a log file whose tail is past the flushed position` regression covers exactly that case and does not exist in 2.7.1. One purge per boot rarely lands on whichever segment sits at the floor — a retention-derived cadence takes a fresh shot at it every pass, which is what turns a latent hole into a prerequisite. This supersedes the earlier claim that the change was blocked on [rocksdb-js #799](https://github.com/HarperFast/rocksdb-js/pull/799); that PR remains a safety improvement on top, restricting deletion to a contiguous oldest prefix so a non-contiguous purge cannot leave a hole that silently truncates a reader, but it is not required here.

Root `msgpackr` moves to 2.0.6 alongside it. rocksdb-js 2.8.0 declares that version exactly, which an exact `2.0.5` root pin cannot satisfy, so the install nested a second msgpackr under `node_modules/@harperfast/rocksdb-js` — two module instances, each with its own extension registry and its own `msgpackr-extract` native addon. That release is a single commit, "Reject array and map lengths that exceed the remaining source data".

## For the human reviewer

1. **Retention window doubles as the replication safety window, and this is the one finding rated major.** Rocks segment purge is driven by `logging.auditRetention` with no floor for the slowest consumer. A peer offline longer than that window resumes, reaches the purged prefix, and gets `done` from `TransactionLog.query()` — indistinguishable from being caught up, so it is recorded as caught up while permanently missing that range. `txnlogReplayGapBytes` is an analytics gauge with no consumer that forces a base copy. Continuous retention is what moves this from unreachable-in-steady-state to routine. The escalation signal belongs in the reader, filed as [HarperFast/rocksdb-js#805](https://github.com/HarperFast/rocksdb-js/issues/805); the alternative is gating purge here on the oldest live consumer position, a semantics change operators would feel on a knob they tune for disk. **Decide whether #805 lands before this ships enabled-by-default**, and confirm whether harper-pro has an escalation path core cannot see — if it does, [the DESIGN.md paragraph](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-3dc5dd454e080eb849ee5efaf79df2585fbe0a06804d69249b4c81da05a63875R330) overstates the risk and should say so.

2. **The 2.8.0 bump carries far more than the retention fix** — 186 commits, 55 files, +4606/-593 in `src/`, including transaction-lifecycle work (#741 close-pending-txns-with-owner, #744 bounding the coordinated-retry park, txn-registry weakptr, deferred orphan cleanup). Local suites are clean across that surface; CI is the authoritative gate. The narrower alternative is asking upstream for a 2.7.2 carrying only the `ensureExtent` commit.

3. **[`purgeLogs()` is synchronous and unbounded on the last worker](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-db526bfa2603e0ee94ab17a9ea8c2b8bd02e1f626dd6907624cfe8508ad356baR232), which also serves requests.** The LMDB path below caps itself at `MAX_DELETES_PER_CLEANUP` and yields through `setImmediate`; the Rocks branch has no analogue. @cb1kenobi benchmarked the native call directly against the same 2.8.0 binary: ~0.12–0.17ms per eligible segment with no yielding, so a 20,000-segment backlog blocks the calling worker for about 3 seconds in one call. A steady-state pass has a handful of eligible segments; the case that hurts is an operator lowering `logging.auditRetention` from 30d to 1d, where the next pass stats and unlinks the whole accumulated backlog at once.

   **Should `purgeLogs()` become async in rocksdb-js?** My read: making it return a promise is the right end state but not sufficient on its own, and not this PR. An napi async worker moves the scan and the `unlink`s off the JS thread, which is what the 3-second number actually calls for — but it also changes a published API with three call sites in core ([the cleanup loop](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-db526bfa2603e0ee94ab17a9ea8c2b8bd02e1f626dd6907624cfe8508ad356baR232), `RocksTransactionLogStore`'s explicit purge, and `ResourceBridge`'s operation), and it re-opens the pin this PR just settled. The cheaper half is additive and composes with either answer: `PurgeLogsOptions` takes no limit today, so give it one, and let the caller re-arm — that is exactly what `MAX_DELETES_PER_CLEANUP` already does for LMDB four lines below, and it bounds the block without touching the return type. My recommendation is to file both against rocksdb-js (bound first, async second) and keep this PR pinned to 2.8.0.

4. **The retained watermark lives only in the cleanup closure.** If the marker write fails transiently and the store is then closed and reopened, the retry is lost and `getLastRemoved()` reports a boundary behind the entries already deleted. Not built here: it self-heals on the next pass that deletes anything, and persisting it would need a second durable write on the path the first one just failed on. Called out because the review raised it and the "retained until it commits" claim above is scoped to the live loop.

5. **Cadence formula:** `retention/10` floored at 10s, pressure shortening cutoff and cadence together, versus keying the backoff on the purge counters rocksdb-js now exposes. One expression to change.

6. **The last-worker conjunct on the [re-arm](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-db526bfa2603e0ee94ab17a9ea8c2b8bd02e1f626dd6907624cfe8508ad356baR324) is unreachable through both arming paths** and kept deliberately as a backstop for a direct caller of the exported `scheduleAuditCleanup`: `onStorageReclamation` registers only on the last worker (it takes no `skipThreadCheck`) and the store-open arm gates on the same index. A reviewer may prefer it deleted as dead code or promoted to the sole ownership check. Both re-arm guards are Rocks-only; the LMDB arm keeps `main`'s unconditional re-arm.

7. **`stopAuditCleanup()` is irreversible** — it latches per audit-store instance, relying on reopen constructing a fresh store. Correct for today's callers; a future caller reusing a store object would get silent permanent retention loss.

8. **Two changes here are not about retention.** The merge of `main` conflicted on `resources/Table.ts`'s reclamation import: this branch added path-wide `removeStorageReclamation()` for database-level teardown and the legacy per-table drop (which unlinks the path outright), while `main` added per-handler `removeStorageReclamationHandler()` for `Table.cleanup()`, because RocksDB column families share a path. They are complementary, so the resolution keeps both. Separately, one commit runs Prettier over `unitTests/resources/query-array-scoping.test.js`: `main` has been red on Format Check since #2437 landed it unformatted, and every branch that merges `main` inherits that failure.

## Verification

Unit: `test:unit:resources` 1936 passing / 28 pending under RocksDB and 1591 passing / 203 pending under `HARPER_STORAGE_ENGINE=lmdb`; `test:unit:dataLayer` 260 passing / 111 pending. `test:unit:main` is 5163 passing with two failures that are this sandbox, not the branch — `nonInteractiveSpawn git credential scoping` (the agent environment exports `GIT_CONFIG_GLOBAL`/`GIT_EDITOR`) and a `getDomainSocketPathLengthWarning` case that needs a rootPath under 107 bytes, which a worktree this deep cannot give it. Prettier, `oxlint --quiet` and `tsc --noEmit` clean.

Every guard added here has a regression that fails when that guard alone is reverted, verified by reverting it, rebuilding (`#src/*` resolves to `dist/`, so a source-only edit does not reach the tests) and re-running:

| revert | test that goes red |
|---|---|
| `stopAuditCleanup()` returns `Promise.resolve()` instead of the active pass | [drains the in-flight removal before `stopAuditCleanup()` reports the pass retired](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-044077436ab82ecf6c481205bdf0ac84571b38edf4ce663842e72349c3a7b38cR1407) |
| cursor release no longer checks `storeClosing()` | [touches nothing further once the root store closes under a suspended pass](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-044077436ab82ecf6c481205bdf0ac84571b38edf4ce663842e72349c3a7b38cR1434) |
| `dropDatabase()` stops awaiting the barrier | [makes `dropDatabase` wait for an in-flight cleanup pass before it closes the stores](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-d89a878c46959d4e1a9364b6009ecf89d8ca08cc125e71357911283edce9b0efR521) |
| the legacy drop goes back to `primaryStore.auditStore` | [retires the legacy per-table drop against the table's own audit store](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-d89a878c46959d4e1a9364b6009ecf89d8ca08cc125e71357911283edce9b0efR556) |
| the init marker write logs through `harperLogger.warn` directly | [contains a rejected last-removed initialization write (failure logging throws)](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-044077436ab82ecf6c481205bdf0ac84571b38edf4ce663842e72349c3a7b38cR1525) |
| the priority term leaves the Rocks cadence formula | [shortens the Rocks cadence while disk pressure is reported](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-044077436ab82ecf6c481205bdf0ac84571b38edf4ce663842e72349c3a7b38cR219) (100 vs 20) |
| the Rocks branch falls through to the LMDB backoff | [holds Rocks cleanup at the retention-derived cadence whatever a pass purges](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-044077436ab82ecf6c481205bdf0ac84571b38edf4ce663842e72349c3a7b38cR166) |

The legacy-drop regression loads a genuine legacy table through `readMetaDb(path, table, schema, auditPath, true)` — real `.mdb` root, real audit environment, real `initStores`/`makeTable` — and asserts the drop reaches the table's own audit store and waits on it before `close()`/`unlinkSync()`. The only synthetic part is the `stopAuditCleanup` spy on that real store; legacy audit stores are opened with a plain `open()` rather than `openAuditStore()`, so they arm no loop today.

The end-to-end route is a live integration regression, [`audit-retention-rocks.test.ts`](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-15a71cb8fd6cbba0c23096e17c53bce57fb884df40ef082c2f46d4d4a7bcb921R73). It and the LMDB suite pass 5/5 in ~27s. It **fails on `origin/main`** with `RocksDB transaction-log cleanup did not advance past purgeRuns=0`, which is #2140's exact symptom. `purgeRuns` counts scans rather than deletions — `transaction_log_store.cpp` increments it before any eligibility check, marked observability-only — so an unproductive pass still advances it and the oracle is valid for the re-arm claim. Deletion itself is proved by the real-`RocksDatabase` unit test asserting only `3.txnlog` survives.

Two retention regressions used to assert the cadence by replacing the process-global `setTimeout` with a stub that captured every timer under a second. That global is shared by every audit store in the process: under `HARPER_STORAGE_ENGINE=lmdb` the shared fixture's own loop is still re-arming at a sub-second cadence by the time they run, so its timer landed in the stub's queue, the assertion saw two timers, and the failing test then hung in its own `finally` awaiting a pass whose timer the stub had swallowed. `.mocharc.json` sets `timeout: 0`, so that wedged the whole LMDB unit run rather than reporting a failure — it reproduces on the branch as pushed, and stays hidden under RocksDB, where the shared store's cadence never drops below a second. [The cadence now lives on the audit store](https://github.com/HarperFast/harper/pull/2338/changes?w=1#diff-db526bfa2603e0ee94ab17a9ea8c2b8bd02e1f626dd6907624cfe8508ad356baR178) and both tests drive real passes and read it, with no fake timers.

Three gaps stated rather than papered over: the last-worker gate has no multi-worker test, because asserting "no second loop exists" needs timing-based assertions; the live regression proves re-arm rather than on-disk deletion, which rocksdb-js's own suite covers; and it skips win32, where deleting a mapped segment raises a sharing violation instead of unlinking, so a Windows purge can warn-log and make no progress while a consumer holds a mapping.

Complexity: complicated

<sub>Review-Coverage: authored=claude; ran=codex; declined=gemini,cursor-grok,cursor-composer,domain; rounds=5 @ c9debf027425</sub>

<sub>Human-Review-Need: 4 @ c9debf027425</sub>


kriszyp:
Reviewer routing: @cb1kenobi is the RocksDB responsible expert and highest-affinity reviewer; @heskew reported #846 and can validate the original retention failure mode.

— Codex

resources/auditStore.ts
```diff
@@ -250,10 +252,19 @@ export function openAuditStore(rootStore) {
...
-						updateLastRemoved(auditStore, lastKey);
-						// and do updates faster
-						if (auditCleanupDelay > 100) auditCleanupDelay = auditCleanupDelay / 2;
+						if (!isRocksAuditStore) {
+							// if we did delete something, update our updates since timestamp
+							updateLastRemoved(auditStore, lastKey);
+							// and do updates faster
+							if (auditCleanupDelay > 100) auditCleanupDelay = auditCleanupDelay / 2;
+						}
+					}
```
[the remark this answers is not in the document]
kriszyp: Fixed in 2dff974 and 609896d: a successful Rocks purge resets the idle backoff, and the Rocks production floor is independent of retention. The deterministic timer regression now covers 10 → 20 → 40 → 10.

— Codex

resources/auditStore.ts
```diff
@@ -250,10 +252,19 @@ export function openAuditStore(rootStore) {
...
+					if (isRocksAuditStore) {
+						const minimumDelay = Math.max(
+							1,
+							Math.min(DEFAULT_AUDIT_CLEANUP_DELAY, auditRetention / 10, MAX_CLEANUP_DELAY)
+						);
+						auditCleanupDelay = Math.max(auditCleanupDelay, minimumDelay);
```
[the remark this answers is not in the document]
kriszyp: Fixed in 2dff974 and 609896d: a successful Rocks purge resets the idle backoff, and the Rocks production floor is independent of retention. The deterministic timer regression now covers 10 → 20 → 40 → 10.

— Codex

resources/auditStore.ts
```diff
@@ -209,27 +204,34 @@ export function openAuditStore(rootStore) {
...
-						lastKey = auditRecord.key;
-						await new Promise(setImmediate);
-						if (++deleted >= MAX_DELETES_PER_CLEANUP) {
-							// limit the amount we cleanup per event turn so we don't use too much memory/CPU
-							auditCleanupDelay = 10; // and keep trying very soon
-							break;
+					if (isRocksAuditStore) {
+						deleted = auditStore.rootStore.purgeLogs({
+							before: Date.now() - auditRetention / (1 + passCleanupPriority * passCleanupPriority),
+						}).length;
```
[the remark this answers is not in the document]
kriszyp: Agreed. This remains draft until rocksdb-js #799 is merged and released, Harper is pinned exactly to that release, and the real Rocks end-to-end retention regression passes against the pinned binding.

— Codex

resources/auditStore.ts
```diff
@@ -209,27 +204,34 @@ export function openAuditStore(rootStore) {
...
-							harperLogger.warn('Error removing audit entry', error);
-						}
-						lastKey = auditRecord.key;
-						await new Promise(setImmediate);
-						if (++deleted >= MAX_DELETES_PER_CLEANUP) {
-							// limit the amount we cleanup per event turn so we don't use too much memory/CPU
-							auditCleanupDelay = 10; // and keep trying very soon
-							break;
+					if (isRocksAuditStore) {
+						deleted = auditStore.rootStore.purgeLogs({
```
cb1kenobi: **Medium: the PR presents as merge-ready while its own body says it must not merge yet**

The description opens with "This draft is blocked on [rocksdb-js#799](https://github.com/HarperFast/rocksdb-js/pull/799) being merged, released, and pinned here; the current exact `2.7.1` dependency does not contain the retention repair." But this PR is **not** marked as a draft, CI is fully green (47 success / 0 failures), it reports `MERGEABLE`, and `cb1kenobi` is a requested reviewer. rocksdb-js#799 is still **open and unmerged**, and `package.json:182` still pins `"@harperfast/rocksdb-js": "2.7.1"`.

The only thing currently holding it is branch protection asking for a review — which means **an approval is exactly what would unblock a merge the author says must not happen.** A reviewer who skims to the diff and finds it clean (it is) has no signal from GitHub that it is gated.

The risk is not hypothetical: this line starts calling `purgeLogs()` on a continuous cadence against a binding whose retention repair is, by the author's own account, absent — and the body also says the final pin "must be followed by a real Rocks end-to-end retention test against the released binding," which cannot have run yet.

Suggested fix: convert to draft until rocksdb-js#799 ships and the pin lands here, then flip to ready. That makes the state GitHub reports match the state the description describes, and no reviewer has to infer the gate from prose.

—
_Generated by Barber AI_

resources/auditStore.ts
```diff
@@ -243,18 +245,24 @@ export function openAuditStore(rootStore) {
...
 						auditCleanupDelay = Math.max(1, Math.min(auditCleanupDelay * 2, auditRetention / 10, MAX_CLEANUP_DELAY));
+					} else if (isRocksAuditStore) {
+						auditCleanupDelay = minimumDelay;
 					} else {
 						// if we did delete something, update our updates since timestamp
 						updateLastRemoved(auditStore, lastKey);
 						// and do updates faster
 						if (auditCleanupDelay > 100) auditCleanupDelay = auditCleanupDelay / 2;
 					}
+					if (isRocksAuditStore) auditCleanupDelay = Math.max(auditCleanupDelay, minimumDelay);
```
cb1kenobi: **Low: Rocks stores now hold an un-`unref`'d repeating timer where they previously held none**

Before this change the Rocks branch returned at the top of `scheduleAuditCleanup` after a one-shot `purgeLogs()`, so no timer was ever created. Now Rocks joins the self-rearming loop, and the re-arm here guarantees a pass every `minimumDelay` (10s in production) for the life of the store. `pendingCleanup = setTimeout(...)` at `:193` is never `.unref()`ed and nothing clears it outside `scheduleAuditCleanup` itself.

The close path is handled correctly — the `status === 'closed' || 'closing'` guard resolves and declines to re-arm — but the *pending* timer still has to fire before that guard is reached, so after a store closes the loop keeps libuv alive for up to one full cadence. For the long-running server that is invisible; for an embedded or test process that expects the loop to drain on close, it is a ~10s tail that did not exist for Rocks before.

Worth noting the new fake-timer test already stubs `unref()`, so the shape was anticipated.

This is consistent with how LMDB has always behaved on this same path, so it is a pre-existing pattern newly extended to Rocks rather than a new bug — flagging it as a deliberate-or-not call rather than a defect.

Suggested fix: `.unref()` the cleanup timer, so retention never by itself keeps a process alive.

—
_Generated by Barber AI_

resources/auditStore.ts
kriszyp: Converted the PR back to Draft. rocksdb-js #799 remains open and `@harperfast/rocksdb-js` remains pinned to 2.7.1, so the draft gate now matches the dependency and end-to-end verification blockers documented in the description.\n\n— GPT-5 Codex

resources/auditStore.ts
kriszyp: The pending cleanup timer is already unref’d: the `setTimeout(...)` expression ends in `.unref()` at `resources/auditStore.ts:269`, and that call is unchanged from `main`. The fake timer implements `unref()` precisely because the production scheduler invokes it. Rocks now joins the existing self-rearming loop, but that loop does not hold libuv open, so no code change is needed here.\n\n— GPT-5 Codex

cb1kenobi:
Reviewed `b997b444` — no new issues. The prior Medium (merge-ready framing vs. blocked-on-#799 body) is resolved: the body now correctly names rocksdb-js 2.8.0's `ensureExtent()` fix as the real prerequisite, and I independently verified that fix in rocksdb-js's own source and regression test at the v2.8.0 tag. The prior Low (un-`unref`'d Rocks cleanup timer) is unchanged in substance — the timer is `.unref()`'d — so nothing outstanding there either.

Verified independently: the last-worker re-arm conjunct is genuinely unreachable through both arming paths (confirmed by tracing `onStorageReclamation`'s own last-worker gate); `stopAuditCleanup()`'s four logical teardown callers all destroy/close the store immediately after, so no reuse-after-stop path exists; the live integration test proves re-arm via `purgeRuns` (a scan counter, confirmed incremented unconditionally in `transaction_log_store.cpp`) while the real-`RocksDatabase` unit test independently proves on-disk deletion (asserts only `3.txnlog` survives) — the two together prove what the PR claims, and a mutation test against `stopAuditCleanup()`'s irreversibility latch is caught by the dedicated unit test for it.

One data point worth having for item 3 in the human-review list: I benchmarked the native `purgeLogs()` call directly (synthetic segments, same rocksdb-js 2.8.0 binary) — it scales linearly at ~0.12–0.17ms per eligible segment with no yielding, so a 20,000-segment backlog blocks the calling worker for about 3 seconds in one synchronous call. That's a real number for whoever weighs the retention-lowering risk.

The six Human-Review-Need items are product/escalation calls, not defects — no comment on those. The PR body's own trailer says `Human-Review-Need: 4` but lists 6 decision names; likely just a stale count, worth a one-line fix whenever convenient.

—
_Generated by Barber AI_

resources/auditStore.ts
```diff
@@ -265,6 +298,17 @@ export function openAuditStore(rootStore) {
 		return resolution;
 	}
 	auditStore.scheduleAuditCleanup = scheduleAuditCleanup;
+	/**
+	 * Retires the cleanup loop for good. The in-pass status check declines to re-arm on a closing store,
+	 * but a pass already on the timer has to fire to reach it.
+	 */
+	auditStore.stopAuditCleanup = function () {
+		cleanupStopped = true;
```
cb1kenobi, line 354: **High: stopping cleanup does not drain the active pass.** `cleanupStopped` prevents another iterator advance, but it cannot cancel an LMDB pass already suspended inside `await removeAuditEntry(...)`. `closeDatabase` and the other teardown callers proceed immediately to close the primary and root stores, so that removal can resume and execute tombstone callbacks or store operations against closed handles. Have `stopAuditCleanup()` expose and drain the active cleanup/removal promise, and await that barrier in every teardown path before closing stores (or propagate cancellation into `removeAuditEntry` and wait for acknowledgement).

—
_Generated by Barber AI_

cb1kenobi:
Reviewed c9debf02 and found no blocking issues. No new blocking findings were confirmed on changed lines. Previously raised issues were not repeated, and the remaining disclosed risks are tracked or require product decisions.

—
_Generated by Barber AI_

PR #2343, opened 2026-08-26
Prevent HTTP workers from exiting before startup completes

kriszyp:
Refs #2312. [Keep HTTP workers alive for the entire startup handshake](https://github.com/HarperFast/harper/pull/2343/changes?w=1#diff-203cc9c9c5531f3f084fdd3b3f51473caef679476c20f7db9328d12ecd334269R175), make [root-component startup rejection terminal in workers](https://github.com/HarperFast/harper/pull/2343/changes?w=1#diff-203cc9c9c5531f3f084fdd3b3f51473caef679476c20f7db9328d12ecd334269R253), and avoid unnecessary cross-worker contention while validating Harper package links. Concurrent startup now survives an initialization await backed only by unref'd handles without weakening the existing pre-ready code-0 fatal path.

The independent review left three bounded limits for human attention: a waiter currently reports rather than retries if a symlink repair owner dies mid-repair; the 10-second lock timeout remains ref'd during shutdown; and lock-contract tests use an injected store rather than the native cross-thread implementation.

This is a replacement candidate for #2314. That PR identified the same liveness invariant but reports that its full-instance regression did not reproduce against unmodified sources; this draft's concurrent unref and rejection arms are mechanically red against a full rebuild of `origin/main`.

## For the human reviewer

1. **Startup liveness owner:** This refs `parentPort` at `startServers()` entry instead of depending on incidental component-loader handles or a separate startup sentinel. That makes the handshake owner responsible for liveness and preserves `manageThreads` shutdown unref behavior; rejecting this choice requires another explicit startup keepalive mechanism.
2. **Symlink coordination shape:** [Every worker validates links without locking and only contends when repair is required](https://github.com/HarperFast/harper/pull/2343/changes?w=1#diff-1e3a919bf3d6ba479113144854f95e5126e6926774ef0af6770ea9feaff170b7R447), rather than serializing every component startup or moving link preparation into the main thread. This is localized and reversible, but deliberately trades a small amount of synchronous metadata I/O for removal of the reported lock-contention window.
3. **Startup failure signaling:** A root-load rejection is logged and exits the worker with code 1 instead of adding a structured worker-to-parent failure protocol. The direct exit fits the existing supervisor policy; rejecting it means adding a moderate protocol change so the parent owns termination and diagnostics.

## Verification

- `npm run build` — passed from a clean full build, including after the final rebase.
- `npx mocha unitTests/components/symlinkHarperModule.test.js` — 7 passing.
- Focused worker-startup integration suite — the [ref control and unref regression](https://github.com/HarperFast/harper/pull/2343/changes?w=1#diff-aee784333554762f95a756c7981a22ad966565b2655c35caf0196937146dd899R75) and [terminal rejection](https://github.com/HarperFast/harper/pull/2343/changes?w=1#diff-aee784333554762f95a756c7981a22ad966565b2655c35caf0196937146dd899R99) arms passed; the unref and rejection arms were also proven red against rebuilt `origin/main` production sources, with the unref arm producing `Worker (index 2) exited with code 0 before reporting ready`.
- Broader suites were exercised but retain unrelated baseline/environment failures: the resource suite reached 1,737 passing/22 pending with 10 audit/cache/transaction failures, and `test:integration:all` included blob-path striping failures. Focused touched-path coverage is green.

Complexity: complicated

— GPT-5 Codex

🤖 Generated with [OpenAI Codex](https://openai.com/codex/)

<sub>Review-Coverage: authored=codex; ran=claude,gemini; declined=cursor-grok,cursor-composer,domain; rounds=4 @ e85f54bd4cfc</sub>

<sub>Human-Review-Need: 4 @ e85f54bd4cfc</sub>

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -284,6 +284,14 @@ A package-manager timeout must not release this lock while npm descendants are s
 
 Boot's `harper-application-lock.json` records an application configuration only after preparation fulfills. Recording at queue time would make a failed install look complete and suppress its retry on the next boot.
 
+Each non-root component also needs `node_modules/harper` (and an existing legacy `harperdb` entry)
+linked to the running package. `componentLoader.symlinkHarperModule()` validates those links before
+taking the cross-thread store lock, so normal multi-worker startup does not serialize on already-valid
+links. A repair winner revalidates under the lock and clears its timeout before releasing ownership. A
+loser retains a bounded, ref'd wait and validates the winner's result when notified, but never calls
+`unlock()` itself: a wait timeout does not transfer lock ownership, and unlocking there can release a
+different worker's critical section and leave a stale timer to unlock a later owner.
+
 ## Peer-side deploy_component payload read: retryable blob stalls and `Readable.from()` cancellation
 
 `readPayloadBlobWithRetry` (`components/deploymentRecorder.ts`) wraps the peer's read of a replicated `hdb_deployment` row's `payload_blob` so a transient 503 `BlobReadError` (`BLOB_UNAVAILABLE_STATUS`, `resources/blob.ts`) — content bytes not arriving within `blobReadTimeout`, e.g. a parked blob send on the origin — retries instead of failing the whole deploy. Two non-obvious constraints shaped the design:
--- a/components/componentLoader.ts
+++ b/components/componentLoader.ts
@@ -392,7 +392,7 @@ function harperModulesAreLinked(componentDirectory: string) {
 	const harperModule = join(nodeModulesDir, 'harper');
 	try {
 		lstatSync(harperModule);
-		if (realpathSync(harperModule) !== realpathSync(PACKAGE_ROOT)) return false;
+		if (realpathSync(harperModule) !== PACKAGE_ROOT) return false;
 	} catch {
 		return false;
 	}
@@ -404,7 +404,7 @@ function harperModulesAreLinked(componentDirectory: string) {
 		return true;
 	}
 	try {
-		return realpathSync(harperdbModule) === realpathSync(PACKAGE_ROOT);
+		return realpathSync(harperdbModule) === PACKAGE_ROOT;
 	} catch {
 		return false;
 	}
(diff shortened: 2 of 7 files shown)


dawsontoth (approved):

kriszyp:
Closing in favor of [#2314 Keep a pre-ready worker's event loop alive through startup](https://github.com/HarperFast/harper/pull/2314), which merged as `635fbd5387386cd4296ec56e25df8d68931da467`.

Both PRs carry the same load-bearing fix — `parentPort?.ref()` at `startServers()` entry, plus a terminal root-component-load rejection arm — so there is nothing left here to merge on top of `main` without conflicting on the same lines of `server/threads/threadServer.js` and `components/componentLoader.ts`.

The one argument for this PR over #2314 was that #2314's regression "did not reproduce against unmodified sources." That was accurate when this was opened, and stopped being true a few hours later: #2314's `945f40ea8` added a standalone `Worker` regression driving the real `startServers()` path, which goes red with the historical `worker exited with code 0 before reporting ready` signature when the startup ref is flipped to `unref()`. #2314 also went on to collect two human approvals and full green CI, and dropped its 10-minute fatal startup deadline after review.

**One thing here is worth keeping and is not in #2314:** the lock-free fast path — checking `harperModulesAreLinked()` before `tryLock`, so only a worker that actually needs a repair contends. `main` still takes the lock on every non-root component load on every worker. Filed as a follow-up so it doesn't get lost with this branch: [#2376 Skip the component-directory lock in symlinkHarperModule when the harper module links are already valid](https://github.com/HarperFast/harper/issues/2376).

Two things I checked here that are *not* problems, recorded so nobody re-litigates them if the fast path gets rebuilt:

- Dropping `realpathSync(PACKAGE_ROOT)` in favor of bare `PACKAGE_ROOT` is safe — it is already resolved at its definition in [`utility/packageUtils.js:47`](https://github.com/HarperFast/harper/blob/main/utility/packageUtils.js#L47).
- Rejecting on the waiter's lock timeout cannot fail a worker's startup — the only call site already catches and logs ([`componentLoader.ts:665`](https://github.com/HarperFast/harper/blob/main/components/componentLoader.ts#L665)).

The still-unresolved item is this branch's `Integration Tests 2/6 (Windows)` failure — `Probe /SeoPageCache/ did not become ready within 120000ms ... after restart_service`. Never re-run, so flake vs. regression is undetermined; worth a clean run if the fast path is revived from this branch rather than rewritten.

— Claude Opus 5

PR #2347, opened 2026-08-26
feat(operations): add a `put` operation, and fix the target-database authorization mismatch

dawsontoth:
Two changes on one branch, reviewable in order: a `put` operation for the operations API, and an authorization fix found while building it.

| commit | what |
| --- | --- |
| 1–4 | superseded by 5 — kept for review history, squash on merge |
| 5 | **`put`**: create-or-replace, replacing the earlier `full_record` flag and `__unset__` directive |
| — | **security**: resolve the target database identically in authorization and handlers |

---

# `put`

```json
{ "operation": "put", "database": "dev", "table": "dog",
  "records": [{ "id": 1, "dog_name": "Penny" }] }
```

The stored record becomes exactly that, so an attribute the request omits is removed. Creates a missing record, replaces an existing one.

**Why it's needed.** There is currently no way to remove an attribute from a record through the operations API. `update` and `upsert` both land on `Table.patch` for a record that already exists (`ResourceBridge.upsertRecords`), which merges: an omitted attribute keeps its stored value, and `null` stores a null. A full replace existed only over REST (`PUT /Table/id` → `resource.put`), which a client speaking the operations API can't reach — and can't reach at all through a proxied operations endpoint. Reported downstream as HarperFast/studio#1643, where Studio's row editor silently dropped a property deletion.

The workaround available today is `delete` then `insert`, which leaves the record **absent between two writes** (concurrent reads, relationship resolutions from the far side, and replicas all observe the gap), **resets `__createdtime__`**, and shows subscribers a **delete followed by an insert** instead of one write. `put` is `Table.put` — the same write REST performs, so the same audit type, replication shape, and retained created time.

**`update`/`upsert` are deliberately untouched.** `5da23c3ef` originally used `Table.put` and hand-copied missing properties to emulate a merge; `5bd363946` replaced that with a real `Table.patch`, and `b302a16e7` added the `put` fallback for a missing record. Merge semantics is load-bearing v4 compatibility, so this is a new operation rather than changed behaviour.

## Two details worth review

**The internal marker is an argument, not a request field.** `putRecords` calls `upsertRecords(obj, true)`. The first draft read `full_record` off the request, which let a client send `full_record: true` with an `update` and get a replace — making the operation name stop describing the write, and bypassing the attribute-scoped denial below, which is keyed on the operation. A test pins it; restoring the request read fails that test and nothing else.

**A role with `attribute_permissions` on the table may not use `put`.** Attribute permissions are checked against the attributes a request *supplies* (`checkAttributePerms`), which is enough for a merge but not for a replace, where an omitted attribute is *removed* — so such a role could otherwise erase an attribute it has no `update` permission for by leaving it out. REST closes the same gap in `Table.allowUpdate` by restoring those attributes from the stored record, which needs the stored record; authorization runs before anything is read, so it denies instead. Roles that scope no attribute are unaffected. Happy to swap this for a restore-based approach if you'd rather it mirror REST exactly.

## Scope reduction

This replaces an earlier design (a `full_record` flag on `update`/`upsert`, plus a per-record `__unset__` directive), per review. Three problems retire with it, none now reachable:

- `__unset__` degraded a merge to a whole-record last-writer-wins put under contention — reachable single-node via the optimistic-lock retry, and unrecoverable in replication because the audit carried only the resolved put, not the deletion intent.
- The reserved `__unset__` key reinterpreted a pre-existing user attribute of that name, with no migration.
- The directive interacted with dynamic attribute expansion, so a rejected write could still have registered attributes on an open table.

`__unset__` is not abandoned — #2350 is now the feature request for doing it properly, with removal preserved as intent at the patch layer rather than resolved in the bridge. It's a genuinely better tool than `put` for "drop one field, don't resend the rest"; it just needs a reverse that carries the prior value, a defined removal-versus-set resolution, and a namespace that can't collide.

---

# The authorization fix

Found while testing the attribute-permission denial; unrelated to the feature.

`verifyPerms` resolved the target database as `requestJson.schema ?? requestJson.database`. The handlers resolve it in `commonUtils.transformReq`, which runs **after** authorization. The two disagreed three ways, each letting a request be authorized against one target and executed against another.

**Impact: any authenticated non-empty role had unconditional read/write/delete on the default `data` database, and could direct a write into any other named database.** Not privilege widening — a role scoped *exclusively* to a second database, with zero permissions on `data`, read `data.Dog` in full (including an attribute it was explicitly denied), updated it, and deleted the record.

**1. Neither key present.** `operationSchema` was `undefined`, so `schemaTableMap` stayed empty; `hasPermissions` iterates that map, so it authorized by vacuous truth, and attribute permissions went unchecked too. Verified with a role holding `update: false` on `data.Dog`:

```
{"operation":"update","table":"Dog","records":[…]}                   → 200, record written
{"operation":"update","database":"data","table":"Dog","records":[…]} → 403 required_table_permissions:["update"]
```

Affected operations, all verified by running them: `insert`/`update`/`upsert` (write), `delete`, `csv_data_load` (async write — jobs authorize only at submit, so it rides into the worker), `create_attribute` (schema mutation), and all three `search_*` (read). `describe_*` and every `requires_su` operation fail closed. `search_by_id` with `get_attributes:['*']` 500s rather than leaking.

**2. `database: 0`, or any falsy-but-present value.** `??` kept the `0`, which is not a database, so the map stayed empty as in (1) — while `transformReq`, which tests falsy, defaulted to `data` and wrote there. `Joi.number()` is an accepted type, so `0` arrives validated.

**3. Opposite precedence.** `verifyPerms` preferred `schema`; `transformReq` prefers `database`. So `{"schema":"data","database":"elsewhere"}` was authorized against `data` and written to `elsewhere` — **the worst of the three**, since it reaches any named database.

## Fix

One resolver, `commonUtils.resolveTargetDatabase`, which `transformReq` delegates to, so the two cannot drift again. Falsy rather than nullish, `database` over `schema` — matching what the handlers have always done, so no handler behaviour changes. `transformReq` writes only when the value would change: an unconditional assignment throws on a frozen object under strict mode.

Plus a fail-closed backstop in `verifyPerms`: a named table with an empty `schemaTableMap` is denied rather than authorized. Both that guard and the map population test `!= undefined`, because `hdbTable` accepts a number and a truthy test dropped a table named `0` — the same vacuous-truth shape one guard over. `getAttributePermissions` too.

Also here, from review: the bulk-load attribute-permission opt-out is keyed on the operation being a known bulk load, never on a caller-supplied `action`. `action` is an unknown-but-accepted key on this validator, so inferring the opt-out from its presence let any direct request skip every attribute check by attaching it.

## Affected versions

**4.2.0 through 5.2.6 (current).** `transformReq`'s default arrived in 4.2.0 (`5d5e58fc3`); 4.1 declared `schema` required. The vacuous-map behaviour dates to ~4.0.8 (`76ce6a2e3`) but was inert until that default existed. `release_4.7.19` has the identical expression at `operation_authorization.js:350`.

**Not affected:** REST, MQTT/WebSocket and GraphQL bind the database on the resource class at path-match time and never take it from the request body. The SQL path is the same bug class but a separate, already-fixed code path — and its fix is the stronger one, carrying the same per-reference backstop.

## Regression risk

**In-repo: nil.** All 77 schema-less table-naming requests in `integrationTests` run as super_user, which returns before the target is used; all 491 non-SU requests already carry `schema` or `database`.

**External: real, and release-note-worthy.** A deployment whose clients send schema-less operations under a restricted role will start getting 403s. That is the vulnerability being closed, but it will look like a breaking change to anyone unknowingly relying on it.

Two intentional behaviour changes beyond the denials: `structure_user: ['data']` + `create_table` with no database goes deny → allow (correct, previously untested), and `search_by_id` with `get_attributes:['*']` and no database goes 500 → a proper allow/deny.

---

# Testing

`integrationTests/database/put-operation.test.ts` — 18 probes, all passing on rocksdb.

**`put`:** the removal, `update` and `upsert` still merging (the controls), an explicit `null` staying a stored null, `__createdtime__` surviving and `__updatedtime__` re-stamping, create-and-replace, `put_hashes` reporting, parity with REST `PUT` (same stored attribute set through both doors), a relationship foreign key still resolving from the far side, a client-supplied `full_record` **not** turning an update into a replace, and the attribute-permission denial with its scope (the same role can still merge and insert).

**Authorization:** all three resolution divergences, asserted **in both directions** — a denial alone doesn't prove resolution, because the backstop also denies an unresolved target, so the role that *does* hold rights on the default database must still be allowed and its write must land there. A numeric `table: 0`. A caller-supplied `action` not skipping the attribute check.

**Also pinned:** on a legacy open table the operations-API attribute projection reports every *registered* attribute, filling in `null` for one the record doesn't have — so an absent attribute reads back as `key: null`, identically to one that was removed. A read-path artifact, not storage (verified by inserting a record that never had it), predating this work, and worth a test because it makes a successful removal look like a failed one.

Every guard mutation-checked — reverting it has to turn a test red:

| mutation | result |
| --- | --- |
| `put` ignores the full-record marker | 5 fail |
| marker read off the request (client-forgeable) | 1 fail |
| attribute-scoped `put` denial dropped | 1 fail |
| `??` instead of `\|\|` in the resolver | 1 fail |
| `schema` before `database` | 1 fail |
| no default at all | most fail |
| numeric-table guards reverted to truthy | 1 fail |
| `action`-keyed bulk opt-out restored | 1 fail |

Also green: `integrationTests/security/*.test.ts`, `apiTests/terminology.test.mjs` (which deliberately exercises schema-less operations and asserts the `data.` default), `database/{auto-fields,patch-disjoint-field-merge}` — 143 with the put suite — plus 575 in `northwind`, which exercises the CSV and bulk-load paths this touches.

# Cross-model review

Two passes via `prepush-review.mjs`, `reviewers=codex+harper-domain`, `independent=true`. Gemini failed on quota both times and both Cursor legs were skipped (their `git fetch` hits a local SSH-agent failure), so coverage is codex + domain only. **Both passes predate this scope reduction** — they reviewed the `full_record` + `__unset__` design, and their findings are what drove it. Their still-live findings are fixed; the rest retired with the directive.

# Not done

- **Replication is untested here.** `put` replicates as a `put`, the same as REST `PUT` already does, but there's no multi-node probe in this suite.
- **LMDB run not done** (`HARPER_STORAGE_ENGINE=lmdb`). The suite is written to run under it.
- **Unit suites not run** — a local Harper instance held the RocksDB lock throughout. Everything here is integration-covered; `resolveTargetDatabase` has no direct unit test.
- **Docs not written.** `reference/operations-api/operations.md` in the documentation repo needs a `put` entry, plus a note about the open-table read projection. Happy to open that PR once this shape is agreed.
- **The `put` denial for attribute-scoped roles is blunt** — it refuses rather than restoring unwritable attributes from the stored record the way REST does.
- Whether the security commit warrants a GHSA advisory, a CVE, and backports to 4.7/5.x is a maintainer call; the affected range above is what those would need.


dataLayer/harperBridge/ResourceBridge.ts
```diff
@@ -771,3 +801,23 @@ async function* groupRecordsInHistory(table, start?, end?, limit?) {
...
+function takeUnsetAttributes(record: Record<string, unknown>, primaryKey: string): string[] {
+	const unset = record[OPERATIONS_UNSET_KEY];
+	if (unset === undefined) return [];
+	delete record[OPERATIONS_UNSET_KEY];
+	const names = unset as string[];
+	if (primaryKey && names.includes(primaryKey)) {
+		throw new ClientError(`'${primaryKey}' is the primary key of this table and cannot be unset`);
+	}
+	return names;
+}
```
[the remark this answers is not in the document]
dawsontoth: Fixed in 2c6c7f55c — good catch, and worth hardening even though it isn't reachable today. `validation/insertValidator.ts` rejects a malformed `__unset__`, and every current route into `upsertRecords` goes through `dataLayer/insert.ts` create/update/upsertRecords (including the replication catchup path). But the bridge's own `insertUpdateValidate` doesn't check this key, so the helper's safety lived entirely in a different layer, across a module boundary an internal caller could reach directly.

Two deliberate departures from the suggested patch:

**The key is deleted before the shape is judged, and on every exit.** `if (unset == null) return [];` returns ahead of the `delete`, which leaves `__unset__` on a record that is about to be written — storing the directive as data. Membership is tested with `in` for the same reason, so an explicitly-undefined key still comes off.

**Malformed values are ignored with a warning rather than thrown.** This is the write and replication apply path, where throwing aborts the commit and can wedge a subscription. Skipping the directive degrades to a plain merge, which removes nothing — the safe direction, and the same choice `resources/tracked.ts` makes for an unrecognized CRDT operation, for the same reason. The check also rejects non-string and empty-string entries, not just a non-array, since a bare string would otherwise iterate per character and delete single-letter attributes.

Worth stating plainly: behaviour through the operations API is unchanged, so the existing probes still cover it, and the internal-caller path this hardens can't be reached from an integration test — nothing exercises the new branch directly.

dawsontoth:
Blocker 1 (the merge → last-writer-wins put) is filed as #2350 (P2, Bug) rather than fixed here. The PR body carries the corrected claim, so the `__unset__` contract as documented on this branch is single-writer.

#2350 lays out the three options: a removal operation in the `resources/crdt.ts` registry (which keeps the feature's purpose but needs a reverse that carries the prior value, plus a defined resolution for removal-racing-set), narrowing the documented contract to last-writer-wins like `full_record`, or restricting the directive to single-node. Option 2 is the honest short-term position if this merges before the design is settled.

_🤖 Filed by [Claude Code](https://claude.com/claude-code)_

utility/operation_authorization.ts
```diff
@@ -1077,6 +1121,26 @@ function getRecordAttributes(json) {
...
+					// `__unset__` is a directive, not an attribute: contribute the attributes it REMOVES
+					// instead of the key itself, so removing one is checked against the same `update`
+					// permission that writing it would be. Without this the removals are invisible to
+					// checkAttributePerms, which only ever sees the attributes a request supplies.
+					//
+					// Only for the operations it does something on. `upsertRecords` strips the directive
+					// for every write, so on an `insert` — which never writes over an existing record —
+					// it removes nothing, and contributing its names would demand `insert` permission on
+					// attributes the request is not touching.
+					if (key === terms.UNSET_ATTRIBUTES) {
```
kriszyp: This new permission expansion is bypassed whenever the request contains a truthy `action`, because `getRecordAttributes` returns an empty set at `utility/operation_authorization.ts:1090`. Direct update/upsert validation accepts unknown top-level keys, so an attribute-scoped caller can add `action: "update"`; the handler ignores that field but executes `__unset__`, removing an attribute for which the role lacks update permission. Please stop using caller-controlled presence of `action` to skip inspection—have known bulk operations opt out explicitly after chunk-level validation—and add a direct-request regression test.

dataLayer/harperBridge/ResourceBridge.ts
```diff
@@ -259,12 +280,24 @@ export class ResourceBridge extends BridgeMethods {
...
 				}
+				const unset = takeUnsetAttributes(record, Table.primaryKey, upsertObj.requires_no_existing);
+				// `__unset__` removes named attributes while everything else still merges, which a patch
+				// cannot express — so it resolves the merge here, against the record this transaction
+				// already loaded, and writes the result as a full replace. Doing it server-side inside the
+				// write transaction is the point: a client emulating this with read-then-replace races
+				// every other writer, and `full_record` would make the caller resend attributes it never
+				// meant to touch. Under `full_record` the submitted record is already the whole intended
+				// state, so there is nothing to merge and the names are simply dropped from it.
+				const toWrite = unset.length && existingRecord && !fullRecord ? { ...existingRecord, ...record } : record;
```
kriszyp: This materializes `__unset__` against one snapshot and then submits that snapshot as a full `Table.put`. If another transaction patches the record before commit, conflict handling refreshes the stored entry but reuses this stale `toWrite`: depending on ordering, the full write can revert the unrelated patch or the out-of-order fold can restore the attribute this request removed. The audit record also carries only the resolved put, so replication cannot recover the deletion intent. Please preserve the unset intent through the Table conflict/retry path and apply it against each refreshed record, with concurrent unset-versus-disjoint-patch coverage for both storage engines.

dataLayer/harperBridge/ResourceBridge.ts
```diff
@@ -771,3 +804,57 @@ async function* groupRecordsInHistory(table, start?, end?, limit?) {
...
+ * otherwise be silently ignored — `Table._writeUpdate` re-asserts the primary key on a full update —
+ * and a directive that quietly does nothing is worse than one that says no.
+ */
+function takeUnsetAttributes(
+	record: Record<string, unknown>,
+	primaryKey: string,
+	isCreateOnly: boolean | undefined
+): string[] {
+	if (!(OPERATIONS_UNSET_KEY in record)) return NO_UNSET_ATTRIBUTES;
+	const unset = record[OPERATIONS_UNSET_KEY];
```
kriszyp: `__unset__` was previously a legal ordinary attribute, especially on open tables. After this upgrade, round-tripping an existing array-valued `__unset__` field removes that field and may delete every attribute named in its value; scalar values become new 400 responses. Moving deletion intent outside the record namespace is safest. If this key must be reserved, please reject it in table/attribute creation and provide an explicit, non-destructive compatibility path for existing conflicting tables rather than silently reinterpreting their data.

validation/insertValidator.ts
```diff
@@ -25,14 +26,61 @@ const customRecordsVal = (value, helpers) => {
...
+		return `'${UNSET_ATTRIBUTES}' must be an array of attribute names`;
+	}
+	for (const name of unset) {
+		if (typeof name !== 'string' || name.length === 0) {
+			return `'${UNSET_ATTRIBUTES}' must contain only non-empty attribute names`;
+		}
+		// The server owns these; `checkAttributePerms` already refuses to let a role write them, and
+		// a full-record write retains __createdtime__ rather than dropping it, so removing them here
+		// would be the one way to lose them.
+		if ((TIME_STAMP_NAMES as readonly string[]).includes(name)) {
```
kriszyp: This protects only the literal legacy timestamp names. Schema timestamp fields can have arbitrary names—this PR's fixture uses `createdAt @createdTime` and `updatedAt @updatedTime`—so `__unset__: ["createdAt"]` passes validation, then the full write restores the server-owned field and reports success without performing the requested removal. Please make the bridge's table-aware validation reject `Table.createdTimeProperty.name` and `Table.updatedTimeProperty.name`, and test the schema aliases rather than only `__createdtime__`.

dataLayer/harperBridge/ResourceBridge.ts
```diff
@@ -771,3 +804,57 @@ async function* groupRecordsInHistory(table, start?, end?, limit?) {
...
+ * refused here instead, since only this layer knows which attribute that is. Unsetting it would
+ * otherwise be silently ignored — `Table._writeUpdate` re-asserts the primary key on a full update —
+ * and a directive that quietly does nothing is worse than one that says no.
+ */
+function takeUnsetAttributes(
+	record: Record<string, unknown>,
+	primaryKey: string,
+	isCreateOnly: boolean | undefined
+): string[] {
+	if (!(OPERATIONS_UNSET_KEY in record)) return NO_UNSET_ATTRIBUTES;
```
kriszyp: Using `in` treats inherited properties as directives. An internal record created with a prototype containing `__unset__: ["color"]`—or any record after equivalent prototype pollution—enters this path despite having no own directive; `delete` leaves the inherited value in place, and the caller then deletes `color`. Since this helper is intended to safeguard the internal module boundary, please use an own-property check such as `Object.prototype.hasOwnProperty.call(record, OPERATIONS_UNSET_KEY)`.

dataLayer/harperBridge/ResourceBridge.ts
```diff
@@ -259,12 +280,24 @@ export class ResourceBridge extends BridgeMethods {
 						}
 					}
 				}
+				const unset = takeUnsetAttributes(record, Table.primaryKey, upsertObj.requires_no_existing);
```
kriszyp: The new guard keys off `requires_no_existing`, so it rejects only `insert`. An `upsert` for a missing ID also has no existing record, but `{ id: "x", color: "black", __unset__: ["color"] }` still succeeds and deletes `color` from the same payload before creating the record—the exact silent partial write this commit rejects for insert. Please base this decision on `existingRecord` rather than the operation flag, and add an upsert-create regression case.

dataLayer/harperBridge/ResourceBridge.ts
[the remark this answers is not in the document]
dawsontoth: Fixed in 14f9dca7d — the source comment now carries the single-writer caveat with the `Table.ts:2902-2914` / `:2530` mechanism and a pointer to #2350, so a reader of just this file doesn't infer concurrent-write safety. You were right that correcting only the PR description left the misleading claim where it would actually be read.

validation/insertValidator.ts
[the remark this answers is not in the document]
dawsontoth: Fixed in 14f9dca7d, and your read on where it belongs was right — the check moved to `ResourceBridge.takeUnsetAttributes`, off `Table.createdTimeProperty`/`updatedTimeProperty` rather than a name list.

Those two resolve `attribute.assignCreatedTime || attribute.name === '__createdtime__'` (`resources/Table.ts:504-505`), so one check covers the schema-declared name *and* the legacy spelling. That made the validator's static list redundant rather than complementary, so I removed it instead of leaving two lists to drift apart.

Two probes, both mutation-checked: `createdAt`/`updatedAt` refused on the fixture's schema-declared table, and `__createdtime__`/`__updatedtime__` refused on an open table created via `create_table`, where they *are* the managed attributes.

Worth recording a wrong turn, since it changes what the tests assert: my first version expected `__unset__: ['__createdtime__']` to be refused on the schema-declared table too. It isn't, and shouldn't be — there `createdAt` is the managed attribute and `__createdtime__` is just a name the record doesn't have, so it's an ordinary no-op. The test now says that explicitly rather than asserting a refusal that would have required the redundant list back.

dataLayer/harperBridge/ResourceBridge.ts
```diff
@@ -771,3 +820,57 @@ async function* groupRecordsInHistory(table, start?, end?, limit?) {
...
+	// Ignored with a warning rather than thrown: this is the write and replication apply path, where
+	// throwing aborts the commit and can wedge a subscription, and skipping the directive degrades to
+	// a plain merge, which removes nothing. `resources/tracked.ts` makes the same choice for an
+	// unrecognized CRDT operation. Checked at all because the shape is validated one layer up, in a
+	// module this one does not own: a non-iterable would throw in the caller's `for…of` and take the
+	// write down, and a bare string would iterate per character, deleting single-letter attributes.
+	if (!Array.isArray(unset) || unset.some((name) => typeof name !== 'string' || name.length === 0)) {
+		// Latched, following `warnedNullSourcePut` in resources/Table.ts: one line per record would
+		// bury the log under a single bad batch.
+		if (!warnedMalformedUnset) {
```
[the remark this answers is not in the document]
dawsontoth: Fixed in 038e3de18 — now a `WeakSet` keyed on the table, so a bad batch on one table can't silence the warning elsewhere, and the latch never keeps a dropped table alive.

Correct on both counts, including the part I'd have missed: I cited `warnedNullSourcePut` as precedent while not actually matching it. It's declared inside `makeTable` (`resources/Table.ts:486`) and its own comment reads "one warn per table per worker" — so the precedent was per-table all along and only my copy was process-wide. The comment now says which scope it means.

No test for it: the malformed branch is unreachable through the operations API, since the Joi validator rejects a malformed `__unset__` before the bridge sees it. This guards an internal caller that doesn't exist yet, and asserting on log lines across two tables wouldn't be a meaningful probe of that.

kriszyp (commented):
Ok, I think this makes sense. However, would it make sense to add a `put` operation (upsert + full_record) instead? If we provided this, do we believe that update+full_record functionality and __unset__ are actually meaningful gaps in functionality? Or do you have immediate use for these in the studio? 

utility/operation_authorization.ts
dawsontoth: Fixed in 341be575e. The opt-out is now keyed on the operation being a known bulk load (`BULK_OPS`), not on the caller-supplied field — and those values are the same handler names `requiredPermissions` is keyed by, so it matches exactly the identity `verifyPerms` already receives. Real bulk loads still defer to `validateChunk`; a direct request can't.

Direct-request regression test covers all three shapes: the `__unset__` removal denied, the same request with `action` attached still denied, and a plain write of the same attribute with `action` attached denied — that third case is the pre-existing half of the hole, which I'd previously logged as inherited-not-created and left alone. You're right that it shouldn't have been. Mutation-checked: restoring the `action` test fails it.

Verified the bulk path still opts out — 581 tests across `northwind` and `job-queue`, which exercise the CSV and bulk-load paths, pass.

dataLayer/harperBridge/ResourceBridge.ts
dawsontoth: Not fixed here — filed as #2350 (P2, Bug) with your analysis, since it needs a design decision rather than a patch.

Your framing is sharper than mine in two ways I've carried into the issue: the conflict/retry path reuses the stale `toWrite` (`Table.ts:2530` refreshes only when `!fullUpdate`), so this is reachable single-node and not only across replicas; and the audit record carries only the resolved put, so replication cannot recover the deletion *intent* — which is what rules out fixing it by retrying harder and points at preserving the intent through the write instead.

The options in #2350 are a removal operation in the `resources/crdt.ts` registry (needs a reverse carrying the prior value, since `applyReverse` hands it only the op's own value, plus a defined resolution for removal-racing-set, which `rebuildUpdateBefore` currently throws on), or narrowing the documented contract to last-writer-wins. The PR body and the source comment now both state the single-writer limitation rather than claiming a merge. Concurrent unset-versus-disjoint-patch coverage on both engines belongs with whichever fix lands — there's no such coverage today, on this path or nearby.

dataLayer/harperBridge/ResourceBridge.ts
dawsontoth: Not fixed — this is the open decision I'd most like your call on, because the safest option you name (moving deletion intent outside the record namespace) is also the one that answers your top-level question about a `put` operation, and I don't want to pick the API shape unilaterally.

What I can confirm: the compatibility hazard is real and worse than "new 400s". On an open table an existing array-valued `__unset__` field round-trips into a directive that deletes every attribute it names, and the key is stripped from the record either way, so the field itself is lost even when the value is scalar. Nothing in this branch rejects the name at table or attribute creation, so a table can still acquire the collision after the upgrade.

Three shapes, and the third is yours: reserve the key with a creation-time rejection plus an explicit migration for existing tables; keep it per-record but namespace it so collision is impossible; or take it out of the record entirely as a request-level field, which removes the class of problem rather than guarding it. I've noted this against the design question on the PR.

validation/insertValidator.ts
dawsontoth: Already fixed in 14f9dca7d, exactly as prescribed — the bridge validates off `Table.createdTimeProperty.name` / `updatedTimeProperty.name`, and the validator's static legacy list is removed rather than kept alongside it, since those two properties resolve `assignCreatedTime` OR the legacy `__createdtime__` spelling (`Table.ts:504-505`) and one check covers both.

Tests use the schema aliases: `createdAt`/`updatedAt` refused on the fixture's schema-declared table, and `__createdtime__`/`__updatedtime__` refused on an open table created via `create_table`, where those *are* the managed attributes. Mutation-checked — removing the loop fails both.

One thing the fix surfaced: on the schema-declared table, `__unset__: ["__createdtime__"]` is *not* refused, because there `createdAt` is the managed attribute and `__createdtime__` is simply a name the record doesn't have — an ordinary no-op. My first test asserted a refusal there and was wrong; it now states that explicitly.

dataLayer/harperBridge/ResourceBridge.ts
dawsontoth: Fixed in 341be575e — `Object.prototype.hasOwnProperty.call(record, OPERATIONS_UNSET_KEY)`, with the reasoning in the comment: `delete` cannot remove an inherited value, so the old check let a prototype-borne directive through and the caller then deleted the attributes it named.

No test, and worth being explicit about why: `JSON.parse` yields plain objects, so this is unreachable through the operations API. It guards the same internal module boundary as the malformed-value handling below it — which is the reason that block exists too.

dataLayer/harperBridge/ResourceBridge.ts
dawsontoth: Already fixed in 14f9dca7d, and keyed exactly where you say — `!existingRecord` rather than the operation flag, so insert, upsert-with-no-key, and upsert-of-an-absent-key are all covered by the one condition. `update` never reaches it, since a missing record is skipped upstream.

Two upsert-create regression cases: no primary key at all (which took the auto-keyed `Table.create` path), and a primary key that isn't stored. Mutation-checked — reverting to `requires_no_existing` fails both.

This was the fix-one-of-N-surfaces mistake on my part: I wrote the guard from the operation that exhibited the bug rather than from the property being protected, which is precisely how the upsert branch stayed open.

dawsontoth:
Yes — done in 1ac8d30b1. `put` is now the whole feature, and `full_record` and `__unset__` are gone.

On your second question, the honest answer is that neither was a gap:

- **Studio's need is exactly `put`.** #1643 is the row editor: the user edits one record's JSON and saves, so the editor already holds the whole record. Nothing there wants a partial removal.
- **`update + full_record` was only a safety nicety** — "replace, but fail if the record is absent", so a mistyped primary key can't create a stray row. Studio can check that itself, having just loaded the record. Not worth a second spelling of the same write.
- **`__unset__` had a real justification — not resending the record — but it wasn't delivering it.** Your stale-snapshot analysis is what settled that: resolved in the bridge it's a read-then-replace wearing a patch's clothing, and the audit carries only the resolved put, so the deletion intent can't survive replication. It needs to be a patch-layer operation or it isn't the thing it claims to be.

So it's filed as #2350 for doing properly (now a Feature, P3), with the reverse-value, removal-versus-set resolution, and namespace problems written up. Dawson wants it eventually — it's a genuinely nicer tool than `put` for "drop one field, leave the rest alone" — just not as a read-then-replace.

Three of your findings retire with the directive rather than being fixed: the stale snapshot, the reserved-key compatibility hazard (no key to reserve now), and the refusal-after-attribute-expansion ordering. The two that were about `put`'s own surface are fixed and tested.

One thing your review caused that I'd flag: collapsing to a single operation exposed a bug in my own first draft of it. I initially had `putRecords` set `full_record` on the request object, which meant a client could send `full_record: true` with an `update` and get a replace — defeating the point of keying semantics on the operation name, and bypassing the attribute-scoped denial, which is keyed on the operation. It's an argument now, and there's a test that fails if it goes back on the request.

_🤖 Reply from [Claude Code](https://claude.com/claude-code)_

dataLayer/harperBridge/ResourceBridge.ts
dawsontoth: Retired rather than fixed: the `__unset__` directive is removed in 1ac8d30b1, so this code path no longer exists. `put` replaces it for the full-replace case, and #2350 now tracks adding `__unset__` properly — as a patch-layer operation, with your analysis of why the bridge-resolved version could not work.

dataLayer/harperBridge/ResourceBridge.ts
dawsontoth: Retired rather than fixed: the `__unset__` directive is removed in 1ac8d30b1, so this code path no longer exists. `put` replaces it for the full-replace case, and #2350 now tracks adding `__unset__` properly — as a patch-layer operation, with your analysis of why the bridge-resolved version could not work.

utility/operation_authorization.ts
```diff
@@ -149,6 +149,13 @@ requiredPermissions.set(
 	write.upsert.name,
 	new (permission as any)(false, [INSERT_PERM, UPDATE_PERM], terms.OPERATIONS_ENUM.UPSERT)
 );
+// Same grants as `upsert`, and the same as REST `PUT` enforces: creating needs insert, replacing an
+// existing record needs update. A full replace can DROP attributes, which is why the
+// attribute-scoped denial below applies to it.
+requiredPermissions.set(
+	write.put.name,
+	new (permission as any)(false, [INSERT_PERM, UPDATE_PERM], terms.OPERATIONS_ENUM.PUT)
```
kriszyp, line 159: `put` requires both insert and update here, but `hasPermissions` later narrows that list to any caller-supplied `request.action`. Because the shared validator accepts unknown keys, a role with `insert: true, update: false` can send `operation: "put", action: "insert"` for an existing ID and replace it; conversely, `action: "update"` can create a missing record without insert permission. Please honor `action` only for actual bulk-load operations (in both table- and attribute-permission checks), or reject it on direct writes, and add regression cases for both directions. This is the deeper root cause already affecting direct `upsert` as well.

server/serverHelpers/serverUtilities.ts
```diff
@@ -400,6 +400,9 @@ async function catchup(req: CatchupOperationRequest) {
 				case terms.OPERATIONS_ENUM.UPSERT:
 					await insert.upsert(transaction);
 					break;
+				case terms.OPERATIONS_ENUM.PUT:
```
kriszyp, line 403: This branch cannot replay `put` semantics from locally generated history. The write records `originatingOperation: "put"`, but `resources/auditStore.ts` has no persisted operation ID for `put`, so decoding loses it; both history readers then normalize the physical `put` type to `upsert` at `dataLayer/harperBridge/ResourceBridge.ts:580` and `dataLayer/harperBridge/ResourceBridge.ts:779`. Catch-up therefore reaches the upsert case and patches an existing replica record, retaining attributes that the source removed. Please encode `put` as an originating operation, preserve it through history while normalizing only legacy puts, and add an audit/catch-up test that verifies an omitted attribute is removed on the target.

utility/operation_authorization.ts
```diff
@@ -1038,12 +1076,20 @@ export function checkAttributePerms(
  * @param json - json containing the request
  * @returns {Set} - all attributes affected by the request statement.
  */
-function getRecordAttributes(json) {
+/** Whether the operation checks attribute permissions per chunk rather than from this request. */
+function isBulkLoadOperation(operationName): boolean {
+	return Object.values(BULK_OPS).includes(operationName);
```
kriszyp: `isBulkLoadOperation` runs for every operations authorization, and `Object.values(BULK_OPS)` allocates a new array each time even though this set is static. Please precompute a module-level `Set` and use `.has(operationName)` so the new security check stays allocation-free on the authorization path.

kriszyp (commented):
So you are cool with this design change? Anyway, I like this, I think this is nice, tightly scoped addition that is consistent with REST semantics, so thank you for getting this updated. Anyway, I think there are couple issues to fix, but looking good.

utility/operation_authorization.ts
dawsontoth: Fixed in ad3c5e538 — and you're right that this is the root cause of the narrower hole I patched in 341be575e. That one only closed `getRecordAttributes`; the permission list itself was still being narrowed by the same caller-controlled field, in both `hasPermissions` and `checkAttributePerms`.

Gated at the single `action` assignment in `verifyPerms` (`isBulkLoadOperation(op) ? requestJson.action : undefined`), since both narrowing sites read that one variable — so table- and attribute-permission checks are covered together. `verifyBulkLoadAttributePerms` is untouched: it gets the real action from `bulkLoad.validateChunk`, not from the request.

Regression cases both directions, as asked: `{operation:'put', action:'insert'}` from a role with `insert: true, update: false` must not replace an existing record, and `action:'update'` from `update: true, insert: false` must not create a missing one. Both now 403 and neither write lands. Mutation-checked — restoring `requestJson.action` fails that probe. 579 tests across northwind and audit-retention confirm the bulk path still narrows as it should.

server/serverHelpers/serverUtilities.ts
dawsontoth: Fixed in ad3c5e538. `put: 4` is persisted in `ORIGINATING_OPERATIONS`, and both readers now share one `normalizeHistoryOperation(originatingOperation, physicalType)`: a recorded originating operation is reported as-is, and only a legacy physical put — no originating operation, written before `put` existed — is still normalized to `upsert`. As you note, the physical type can't distinguish them alone, since an `upsert` that creates a record also writes a physical put.

Test: after an `upsert` that creates and a `put` that replaces on the same id, `read_audit_log` must report both `upsert` and `put`. Mutation-checked twice — dropping the audit id fails it, and so does restoring the unconditional `put → upsert` normalization.

What I could NOT test is the replay itself. Asserting an omitted attribute is removed on a *target* node needs a multi-node harness, which this suite doesn't have; the catch-up path (`serverUtilities` catchup → `insert.put`) is wired and read-verified only. What is covered is the property the replay depends on and where the bug actually was — that the history reports `put` rather than `upsert`. Flagging that gap rather than implying end-to-end coverage.

utility/operation_authorization.ts
dawsontoth: Fixed in ad3c5e538 — module-level `BULK_OP_NAMES = new Set(Object.values(BULK_OPS))`, and `isBulkLoadOperation` is now a `.has()`. It sits on the authorization path for every operation, and it just picked up a second caller in the same commit (the `action` gate), so it runs twice per request now.

dawsontoth:
@kriszyp the design change works just fine for Studio, and I filed away a GH issue for the atomic unset operation which I've definitely used in past databases too (and could have used here). Exposing the existing method through the operation was a viable path forward for Studio, and arguably less complex too, so the pivot was OK. When I first went through and did `fullUpdate` or whatever, that was my attempt to not add an additional method. But having method parity makes a lot of sense, so... here we are! Yup. 🥳

resources/auditStore.ts
```diff
@@ -392,9 +392,16 @@ const ORIGINATING_OPERATIONS = {
 	insert: 1,
 	update: 2,
 	upsert: 3,
+	// `put` must be persisted, not inferred: the physical write type is also `put` for an `upsert`
+	// that happened to create a record, so the history readers cannot tell the two apart from the
+	// type alone. Without an id here the originating operation decoded as undefined, the readers fell
+	// back to the physical type, and replication catch-up replayed a `put` as an `upsert` — patching
+	// the replica and RETAINING attributes the source had removed.
+	put: 4,
```
kriszyp, line 400: This adds value 4 to the durable binary originating-operation enum, but `storage-format.md:127-135` still documents only values 1–3. Please add `4 | put` there so the format specification and any recovery/debugging tooling based on it remain accurate.

integrationTests/database/put-operation.test.ts
```diff
@@ -0,0 +1,668 @@
...
+			`creating a record needs insert permission; got ${created.status} ${JSON.stringify(created.body)}`
+		);
+		ok((await read('Dog', 'nar-absent')) == null, 'the denied request must not have created anything');
+	});
+
+	// Replication catch-up switches on the operation the history reports. `put` has to survive as the
+	// originating operation: the physical write type is `put` for an `upsert` that created a record
+	// too, so the type alone cannot tell them apart, and catch-up replaying a `put` as an `upsert`
+	// would patch the replica and retain attributes the source removed.
+	test('the audit log reports a put as a put, and a legacy physical put as an upsert', async () => {
```
kriszyp, line 394: The test name claims to cover a legacy physical put, but the first write records `originatingOperation: "upsert"`; `normalizeHistoryOperation` therefore returns before exercising the no-origin fallback. This test would still pass if legacy physical puts stopped normalizing to `upsert`. Please construct an entry with no originating operation or add a focused test for that fallback.

kriszyp (approved):
Cool! Ship it!

PR #2348, opened 2026-08-26
ci: keep ineligible runs out of the review concurrency group

heskew:
## What this changes, concretely

Both review callers use a shared concurrency group per PR with `cancel-in-progress: true`, so a newer run can cancel an in-flight review. GitHub evaluates that cancellation **when the new run is queued — before any job `if:` runs**. That means an event that was never going to produce a review could still kill one in progress:

- opt-in mode: any push, or flipping ready without the opt-in label, cancelled a running label-triggered review; the replacement run then skipped, so the requested review silently never completed;
- both modes: applying an **unrelated** label (e.g. `bug`) mid-review did the same — the replacement died at the reusable's exact-label authorize gate.

## The fix

The concurrency group key now ends in `eligible` only for events that would actually produce a review; every other event gets a unique `github.run_id` suffix, i.e. its own group, so it can never cancel anything. "Would produce a review" means, per caller:

- the provider's **own** opt-in label was just applied (`labeled` + `label.name == 'claude-review'` / `'gemini-review'`), or
- any non-label event while `*_ALWAYS_ON` is on, or
- a draft flipped ready while it carries the opt-in label.

The review job's `if:` uses the identical predicate (kept textually in sync — Actions YAML has no anchors), so eligible runs both start the review and may supersede an in-flight one, and ineligible runs do neither.

Findings by @kriszyp and the Codex pass on the pin-bump series (harper-pro#766, #2330/#2331); the unrelated-label variant predated this series in always-on mode and is closed by the same change. Full event matrix in the fa04115 commit message.

**Merge order**: this before #2330/#2331 — the sync PRs carry this fix and must stay byte-identical to main.

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -21,10 +21,12 @@ on:
     types: [opened, synchronize, reopened, ready_for_review, labeled]
 
 concurrency:
+  # NOTE: this predicate and the review job's `if:` must stay textually
+  # in sync — GitHub Actions YAML has no anchors to share them.
   # Ineligible events (per the review job's opt-in gate) get a unique
   # run_id group so they can never cancel an in-flight eligible review —
   # workflow-level cancel-in-progress fires before job `if:` is evaluated.
-  group: claude-review-${{ github.event.pull_request.number }}-${{ (vars.CLAUDE_ALWAYS_ON == 'true' || github.event.action == 'labeled' || (github.event.action == 'ready_for_review' && contains(github.event.pull_request.labels.*.name, 'claude-review'))) && 'eligible' || github.run_id }}
+  group: claude-review-${{ github.event.pull_request.number }}-${{ ((github.event.action == 'labeled' && github.event.label.name == 'claude-review') || (github.event.action != 'labeled' && vars.CLAUDE_ALWAYS_ON == 'true') || (github.event.action == 'ready_for_review' && contains(github.event.pull_request.labels.*.name, 'claude-review'))) && 'eligible' || github.run_id }}
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/claude-review.yml
```diff
@@ -21,7 +21,10 @@ on:
     types: [opened, synchronize, reopened, ready_for_review, labeled]
 
 concurrency:
-  group: claude-review-${{ github.event.pull_request.number }}
+  # Ineligible events (per the review job's opt-in gate) get a unique
+  # run_id group so they can never cancel an in-flight eligible review —
+  # workflow-level cancel-in-progress fires before job `if:` is evaluated.
+  group: claude-review-${{ github.event.pull_request.number }}-${{ (vars.CLAUDE_ALWAYS_ON == 'true' || github.event.action == 'labeled' || (github.event.action == 'ready_for_review' && contains(github.event.pull_request.labels.*.name, 'claude-review'))) && 'eligible' || github.run_id }}
```
kriszyp: This still treats every `labeled` event as eligible. With `CLAUDE_ALWAYS_ON` unset, applying an unrelated label (for example, `bug`) to a PR while its `claude-review` run is in progress uses this group and cancels that real review; the replacement reaches the reusable authorization gate, is rejected because the event label is not `claude-review`, and produces no review. Gate this arm on `github.event.label.name == 'claude-review'` (while retaining the broader job gate if the reusable workflow requires it).

.github/workflows/gemini-review.yml
```diff
@@ -39,7 +39,8 @@ concurrency:
   # in parallel on the same PR. cancel-in-progress is per-group, so a
   # synchronize push cancels the in-flight Gemini run without touching
   # the Claude run (and vice versa).
-  group: gemini-review-${{ github.event.pull_request.number }}
+  # Ineligible events get a unique run_id group — see claude-review.yml.
+  group: gemini-review-${{ github.event.pull_request.number }}-${{ (vars.GEMINI_ALWAYS_ON == 'true' || github.event.action == 'labeled' || (github.event.action == 'ready_for_review' && contains(github.event.pull_request.labels.*.name, 'gemini-review'))) && 'eligible' || github.run_id }}
```
kriszyp: Same cancellation hole as the Claude caller: any unrelated `labeled` event joins the eligible group and can cancel an in-flight Gemini review, but the reusable gate then declines to start a replacement. Make the labeled arm require `github.event.label.name == 'gemini-review'` so only the opt-in gesture shares the review concurrency group.

.github/workflows/claude-review.yml
heskew: Blocking: this predicate treats every `labeled` event as eligible, not just the opt-in label. In opt-in mode, adding an unrelated label while a review is running joins the shared eligible concurrency group and cancels that review; the replacement then fails the reusable workflow exact-label gate, leaving no completed review. Please require `github.event.label.name == 'claude-review'` here and in `jobs.review.if`, and make the equivalent `gemini-review` change in the Gemini caller.

heskew: Taken in fa04115 — the labeled branch of the predicate now requires `github.event.label.name == '<provider>-review'`, in both the concurrency group and the review job gate, for both callers. Unrelated-label events are ineligible in BOTH modes — tracing this also showed the unrelated-label cancellation predated this series in always-on mode (any labeled event shared the group, then authorize skipped), so this closes that too. Full event matrix is in the commit message.

kriszyp (commented):
Still not sure what this does, but ok.

.github/workflows/claude-review.yml
heskew: ah, yeah. similar review to my manual codex pass. claude skipped a step on getting these prs up...

.github/workflows/gemini-review.yml
heskew: Covered in fa04115 alongside the Claude caller: the gemini labeled arm requires github.event.label.name == 'gemini-review' in both the concurrency predicate and the job gate, so only the opt-in gesture shares the review concurrency group.

PR #2356, opened 2026-08-26
Continuously enforce RocksDB audit log retention

### LINKED ISSUE #2140: RocksDB transaction-log pruning never re-arms — one purge at boot, then nothing until disk pressure

cb1kenobi:
Fixes #2140 by [continuously re-arming RocksDB audit-retention cleanup](https://github.com/HarperFast/harper/pull/2356/changes?w=1#diff-db526bfa2603e0ee94ab17a9ea8c2b8bd02e1f626dd6907624cfe8508ad356baR214), [routing purges through local and sibling-worker cache invalidation](https://github.com/HarperFast/harper/pull/2356/changes?w=1#diff-56c43863dc5f281e19328e36b0082523391cfb5373c0bd88a8d3d94bfa1ab46cR498), and [stopping cleanup when a database closes](https://github.com/HarperFast/harper/pull/2356/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1180).

This is ready for review with a known reliability gap that requires an explicit decision before merge: a live reusable replication or subscription iterator may be permanently retired if retention purges a sequence it has not reached. Continuous cleanup makes that pre-existing edge substantially more likely, and the intended lagging-consumer recovery contract needs an explicit decision before merge. The Windows mapping/deletion order and process-wide completion semantics also need focused review.

## For the human reviewer

1. Decide whether mmap coherence belongs in Harper through the current rocksdb-js private cache fields, or in a public engine-owned purge/invalidation API. This has cross-platform implications, including whether Windows can delete mapped files before invalidation runs.
2. Decide the lagging-consumer policy: block retention for active iterators, or terminate lagging consumers with an explicit checkpoint/full-copy recovery signal. A real POSIX RocksDB characterization now confirms that the mapped sequence remains readable after unlink, but the lagging iterator silently ends when it advances to an unread purged sequence; a fresh iterator can still read the retained newer entry.
3. Confirm the passive purge cadence of `max(10 seconds, retention / 10)` versus tying cleanup to transaction-log rotation or flush events.
4. Define whether purge completion means local file deletion or process-wide cache visibility. The current sibling-worker notification is asynchronous.
5. Confirm cleanup lifecycle ownership: explicit calls from each database close/drop path versus audit/root-store close owning teardown.
6. Confirm whether the POSIX iterator characterization plus the [two native `purgeRuns` increments](https://github.com/HarperFast/harper/pull/2356/changes?w=1#diff-15a71cb8fd6cbba0c23096e17c53bce57fb884df40ef082c2f46d4d4a7bcb921R73) are sufficient regression evidence, or whether the integration test should also rotate aged files and prove deletion plus multi-worker read visibility.

## Verification

- `npm exec mocha -- unitTests/resources/auditLog.test.js --grep "silently skips unread purged sequences"` — passed in 10/10 repeated runs.
- `npm exec mocha -- unitTests/resources/auditLog.test.js` — 29 passing, 6 pending.
- `npx prettier --check unitTests/resources/auditLog.test.js`, file-scoped oxlint, and `git diff --check` — passed.
- `npm run build` — passed.
- `npx mocha unitTests/resources/auditLog.test.js unitTests/resources/auditPurge.test.js unitTests/resources/transactionLogStructureUpdate.test.js` — 35 passing, 6 pending.
- `npm run test:integration -- --isolation=none "integrationTests/database/audit-retention-rocks.test.ts"` — 1 passing in about 25 seconds.
- `npm run test:unit:main` — the local run was terminated after entering a non-terminating TLS retry loop against deleted test certificates; no assertion failure was reported before termination. CI is authoritative for the full gate.
- `git diff --check` — passed.
- No user-facing documentation or dependency changes.

Complexity: complicated

<sub>Review-Coverage: authored=codex; ran=cursor-grok; adjudicated=domain; blocked=claude(not-installed),gemini(exit-1); declined=cursor-composer; rounds=7 @ 91629e174b4d</sub>

<sub>Human-Review-Need: 4 (decisions: invalidation-owner, cleanup-cadence, shutdown-contract) @ 91629e174b4d</sub>

### DIFF
--- a/resources/RocksTransactionLogStore.ts
+++ b/resources/RocksTransactionLogStore.ts
@@ -23,6 +23,11 @@ type TransactionLogIterator = Iterator<TransactionEntry | number> & {
 	removeLog(logName: string);
 };
 
+type NotifyingArrayBuffer = ArrayBuffer & {
+	notify(): boolean;
+	cancel(): void;
+};
+
 // Logs (once per log) when a corrupt frame ends a query iterator early; see
 // endIteratorOnCorruptFrame in replayLogsGuards.ts for why this is end-of-log, not a crash.
 function warnCorruptFrame(logName: string) {
@@ -42,6 +47,7 @@ export class RocksTransactionLogStore extends EventEmitter {
 	logByName: Map<string, TransactionLog> = new Map();
 	updates = 0; // the number of updates to the list of logs that have occurred
 	rootStore: RocksDatabase;
+	purgeNotification?: NotifyingArrayBuffer;
 	reusableIterable = true; // flag indicating that iterable can be reused to resume iterating through audit log
 	// Highest structureVersion appended to each per-node TransactionLog, tracked per tableId. Drives the
 	// per-log HAS_STRUCTURE_UPDATE flag in put(). Keyed by (log, tableId): a per-node log interleaves entries
@@ -54,6 +60,13 @@ export class RocksTransactionLogStore extends EventEmitter {
 		super();
 		this.log = rootDatabase.useLog('local');
 		this.rootStore = rootDatabase;
+		try {
+			this.purgeNotification = rootDatabase.getUserSharedBuffer('audit-log-purged', new ArrayBuffer(1), {
+				callback: () => this.invalidateLogBuffers(),
+			}) as NotifyingArrayBuffer;
+		} catch (error) {
+			harperLogger.warn('Failed to register transaction-log purge notifications', error);
+		}
 	}
 
 	/**
@@ -245,6 +258,7 @@ export class RocksTransactionLogStore extends EventEmitter {
 				}
 				if (!log) {
 					log = this.rootStore.useLog(options.log);
+					this.addLogToMaps(log.name, log);
 				}
 			}
 			const queryIterator = endIteratorOnCorruptFrame(log.query(options), warnCorruptFrame(log.name));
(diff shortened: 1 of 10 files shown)


resources/RocksTransactionLogStore.ts
```diff
@@ -464,6 +478,46 @@ export class RocksTransactionLogStore extends EventEmitter {
...
+	}
+
+	purgeLogs(options?: any): any[] {
+		const purged = this.rootStore.purgeLogs(options);
+		if (purged.length > 0) {
+			this.invalidateLogBuffers();
+			try {
+				// Every worker has its own TransactionLog objects, so notify all workers to drop stale mmaps.
+				// Iterators already in progress retain their own buffer references and can finish safely.
+				this.purgeNotification?.notify();
```
kriszyp: Could we move this invalidation into rocksdb-js and make it generation-aware before rearming periodic purge? With rocksdb-js 2.7.1, an existing query iterator retains only its current mmap and advances to exactly the next sequence. If cleanup deletes intervening segments while Harper’s reusable iterator is parked on an old segment, it finishes that segment, cannot map the next one, returns `done`, and never reaches the surviving logs. Harper retains that iterator at `resources/transactionBroadcast.ts:35-40`; start-time replay also drops live notifications while trusting the cursor at `resources/Table.ts:4251-4263`, so subscriptions and replication can silently omit writes, potentially permanently after a later checkpoint. There is also a visibility gap because `notify()` only enqueues callbacks in other worker isolates: purge can report completion while another worker starts a read from its stale mmap, and clearing the object cache afterward cannot revoke that iterator’s captured buffer. The tests use a synchronous notification fake at `unitTests/resources/auditLog.test.js:100-130` and a single-worker stats check at `integrationTests/database/audit-retention-rocks.test.ts:55-96`, so neither race is exercised. Please make purge generation or next-surviving-sequence handling atomic in rocksdb-js and bump the dependency, or serialize and rebuild Harper iterators from their last-delivered cursor; then add a real multiworker, multi-segment purge regression test.

kriszyp: Thanks for adding the real RocksDB characterization at `unitTests/resources/auditLog.test.js:138-215`; it confirms this finding rather than resolving it. The expectation at `unitTests/resources/auditLog.test.js:205-215` proves a lagging iterator stops at the first purged sequence even though a retained newer sequence is readable by a fresh iterator. Harper keeps exactly such an iterator for the lifetime of the subscription registry at `resources/transactionBroadcast.ts:35-40` and deliberately parks it while no subscribers are active at `resources/transactionBroadcast.ts:139-146`. After retention deletes its next segment, later commits keep waking the same completed iterator, so subscriptions and replication never reach retained or new logs; start-time replay also drops live notifications while trusting its cursor at `resources/Table.ts:4249-4263`. The invalidation test at `unitTests/resources/auditLog.test.js:104-134` still makes `notify()` synchronous, whereas the real cross-isolate notification is queued, so purge can also return while another worker begins a read from a stale mmap. Please turn the characterization into a regression for a fix: make rocksdb-js skip to the next surviving sequence with a synchronously checked purge generation, or rebuild Harper’s reusable iterator from its last-delivered cursor after acknowledged process-wide invalidation; then cover a real multiworker purge.

kriszyp:
I closed #2140 as duplicative, and likewise this PR is duplicative of #2338. However, there are a couple notable things in here that I will roll into #2338.

PR #2360, opened 2026-08-27
Keep computed resolver output out of durable records

### LINKED ISSUE #2359: Initial-copy replication silently drops records with scalar @computed attributes
## Summary

On Harper 5.2.6, initial-copy replication can silently drop records from a table that has a surfaced scalar `@computed` attribute while advancing the copy cursor and reporting the link healthy.

The receiver logs `TypeError: attribute.set is not a function`. This is a convergence failure: the receive loop classifies the exception as `decode-drop`, skips the row, and can continue to `COPY_COMPLETE`. Reconnecting does not redeliver rows after the cursor advances.

## Independent reproduction

This uses only Harper/Harper Pro and a minimal component; it does not depend on another application or production data.

1. Start two Harper 5.2.6 nodes, A and B, with replication enabled and the same component deployed independently.
2. Deploy this schema on both nodes:

```graphql
type ComputedReplicationRecord @table @export {
  id: ID @primaryKey
  source: String
  derived: String @computed(from: "source") @indexed
}
```

3. Before B subscribes, write records on A:

```http
PUT /ComputedReplicationRecord/one
Content-Type: application/json

{"id":"one","source":"value"}
```

4. Connect B to A with `add_node { isLeader: true }` to force an initial copy.
5. Compare row counts, read `/ComputedReplicationRecord/one` on B, query the `derived` index, and inspect B's logs.
6. The core compatibility failure can also be reproduced without networking by decoding an affected-release payload containing `{"id":"one","source":"value","derived":"value"}` against this table.

### Actual

- The receiver can throw `TypeError: attribute.set is not a function` while decoding the copied row.
- The row is skipped under the `decode-drop` path.
- The copy cursor can advance despite the missing row.

### Expected

- A stored or transported value colliding with a read-only computed attribute cannot prevent materialization.
- The computed resolver remains authoritative.
- Durable and initial-copy encodings contain stored fields, not response-only computed projections.

## Root cause

`TableResource.updatedAttributes` installs an accessor setter for every resolved attribute, but scalar computed attributes have no `attribute.set`. The decoded record prototype also has a response-oriented `toJSON` that surfaces scalar computed values. When Harper Pro initial copy re-encodes a decoded record, Structon invokes that projection and puts the computed value into the copy payload. The receiver then assigns the transported key through the missing setter and drops the row.

The correction belongs at core's durable-record boundary so storage, audit, and every replication hop share the invariant. Pro owns the real initial-copy regression.

### Harper core (`HarperFast/harper`)

1. Suppress response `toJSON` only while a record is durably encoded, using a nesting-safe scope shared by duplicate module instances.
2. Tolerate and remove legacy own-field collisions with read-only resolvers during decode; preserve writable relationship setters.
3. Reject later user mutation of read-only resolved fields on both frozen struct records and mutable tracked resources.
4. Keep collision telemetry bounded to the first occurrence per table/worker.
5. Add regressions for direct durable re-encode, the actual no-change `get` → `put` instance path, affected-release payload cleanup, binary decode output, mutation behavior, and computed-index authority.
6. Run core build, formatting/lint on changed files, the complete affected test files, and the resource suite.

### Harper Pro (`HarperFast/harper-pro`)

1. Update the core pointer after the core PR is available.
2. Add a unit regression that loads a real GraphQL schema and exercises the initial-copy encode → audit envelope → decode path.
3. Add a self-contained two-node fixture/test that deploys the schema independently, seeds before clustering, forces a full copy, verifies the row and computed index, checks for decode-drop errors, mutates the copied row, and verifies it after receiver restart.
4. Keep residency/invalidation projection behavior unchanged; that protocol has distinct indexed-partial semantics and is not part of this initial-copy fix.
5. Run the Pro unit suite and targeted cluster test; retain existing generic `decodeDropRecovery` behavior unchanged.

## Acceptance criteria

- Ordinary JSON responses still include the scalar computed field.
- Durable encoding and initial-copy payloads omit it.
- Authentic affected-release payloads decode without dropping the record and cannot override resolver output.
- Direct client mutation remains rejected.
- A two-node initial copy converges, computed-index lookup succeeds, receiver logs contain no decode-drop for the copied row, and the row survives mutation plus restart.
- Tests fail on their respective base revisions and pass with the fix.

## Rollout and recovery

Upgrade source/sender nodes first so contaminated stored rows can be materialized and clean payloads emitted, then upgrade receivers. Because affected receivers may already have advanced past dropped rows, force a full base copy after both sides are upgraded and verify row-count and computed-index parity; reconnect alone is insufficient.

## Version

Confirmed on 5.2.6. Target the v5.2 patch line as well as main.

Comment generated by kAIle (Codex GPT-5.6)

kylebernhardy:
## Summary

- keep response-only resolver values out of primary records, audit entries, cache write-backs, full-copy payloads, and ordinary patch payloads
- preserve response-shaped retained messages and residency invalidation projections without allowing those exceptions to contaminate durable row encoding
- clean affected-release read-only resolver collisions across classic, typed, lazy, frozen, RocksDB, and LMDB record shapes
- preserve legacy writable relationship snapshots durably while returning the live relationship resolver in response projections
- keep source-cache response and durable projections consistent across timestamps, background embeddings, schema reloads, and stateful source `toJSON()` implementations
- make response encoding a one-shot encoder request so nested or re-entrant durable encodes return to durable scope
- invalidate warm record caches only when resolver classification changes

## Verification

- final build, formatting, and targeted lint pass
- final focused core suites: 66 passing, 3 pending
- earlier full default resource suite: 1,787 passing, 24 pending
- exact Harper Pro copy unit suite: 5 passing
- independent two-node Pro integration: 1 passing, covering initial copy, computed-index lookup, receiver mutation, restart durability, and decode-log checks
- final Claude + Harper domain review converged at round 26 on `8f4977716`
- Gemini `agy` remains unavailable because the local CLI is not authenticated; no Gemini coverage is claimed

## For the human reviewer

- Confirm the compatibility boundary: read-only resolver output is never durable; writable relationship snapshots remain durable for mixed-version compatibility, while responses use the live resolver.
- The remaining performance observations are the global durable/partial scope bookkeeping and O(rows × resolvers) collision scans on resolver-heavy classic reads. Both are bounded correctness tradeoffs in this patch and are not hidden as resolved findings.

Fixes #2359

Comment generated by kAIle (Codex GPT-5.6)

<sub>Review-Coverage: authored=codex; ran=claude,harper-domain; unavailable=gemini; declined=cursor-grok,cursor-composer; rounds=26 @ 8f4977716</sub>

<sub>Human-Review-Need: 3 @ 8f4977716</sub>


resources/RecordEncoder.ts
```diff
@@ -569,6 +599,25 @@ export class RecordEncoder extends StructonEncoder {
 			return null;
 		}
 	}
+	removeReadOnlyResolverFields(value) {
+		if (value == null || typeof value !== 'object' || ArrayBuffer.isView(value) || !this.readOnlyResolverNames?.size)
+			return value;
+		let collisions: string[] | undefined;
+		for (const name of this.readOnlyResolverNames) {
+			if (Object.hasOwn(value, name)) (collisions ??= []).push(name);
```
kriszyp, line 739: `Object.hasOwn` only sees fields of records decoded in classic (records) mode. Records written in struct mode — any table with `@randomAccessFields` / `storage.randomAccessFields`, plus struct-mode data written before the 5.1 default flip that `utility/lmdb/OpenDBIObject.ts:43-46` says still decodes — come back from structon's `readStruct` as a `LazyObject` whose only own property is the source symbol; every stored field is an enumerable getter on a per-structure prototype that sits *between* the instance and `structPrototype` (`node_modules/structon/struct.js:979-1140`).

For exactly the legacy payloads this filter targets, that means: (1) the collision is never detected, so nothing is cleaned or warned; (2) the per-structure getter for `computed` shadows the resolver getter on `structPrototype`, so `record.computed` returns the stale stored value; (3) on re-encode, structon's own `toJSON` on that prototype (`struct.js:1149`) shadows the durable-aware table `toJSON`, msgpackr calls it at top level (`node_modules/msgpackr/pack.js:129`) and hands the materialized object to `writeStruct` without consulting `toJSON` again — so the stale field is re-persisted rather than dropped. The new tests only run classic mode, so they can't see this.

Suggest doing the projection where records actually enter/leave Harper rather than via prototype `toJSON`: on read in `withEntry` / `#withEntry` (`resources/RecordEncoder.ts:660`, `resources/PrimaryRocksDatabase.ts:85`), detect collisions with `name in value` (or the structon source symbol) and materialize a cleaned copy for lazy instances too; on write, project own+lazy keys minus read-only names before `superEncode` so the result doesn't depend on which `toJSON` wins the prototype chain.

resources/Table.ts
```diff
@@ -580,69 +581,90 @@ export function makeTable(options) {
...
-				if (structSerializationVisited == null && !isCyclic && !hasSurfacedComputed) {
-					// fast path: bounded copy, values left raw (matches the previous for..in output). `name in json`
-					// treats an own stored key (already copied above) as taking precedence over its getter.
-					const json = {};
-					for (const key of Object.keys(this)) json[key] = this[key];
-					for (const name of enumNames) if (!(name in json)) json[name] = this[name];
-					return json;
+		const toJSON = function () {
+			const durable = readOnlyResolverNames.size > 0 && isDurableRecordEncoding();
+			const surfacedNames = durable ? durableEnumNames : enumNames;
```
kriszyp: This is where the deeper root cause still lives. The bug class is "durable encoding runs the response projection", and this closes it only for read-only resolvers: in durable mode `toJSON` still resolves every writable `@enumerable` relationship (`this[name]` → a primary-store `getSync`, or a `search` for to-relationships) and serializes the related record(s) into the primary record.

Concretely, with `@enumerable @relationship(from: "authorId") author: Author`, `await Post.put(await Post.get(id))` (or the no-change instance put the new test covers) stores a full `author` snapshot in the row, the audit entry and the replication payload. On classic decode the setter at `resources/Table.ts:5279` converts it back into `authorId`, so the cost is write amplification plus a read inside the write; on struct-mode tables the snapshot shadows the getter and is served stale — the same failure this PR fixes for computed fields. The fixture change making `related` enumerable and the assertion at `unitTests/resources/crud.test.js:248` pass either way, so the suite doesn't distinguish.

Since durable mode now has its own branch, `surfacedNames = durable ? [] : enumNames` (durable = own stored keys only) fixes this, removes the `durableEnumNames` / `typeof attribute.set` classification, and shrinks the getter to "durable → `undefined` unless an own key collides". If the DESIGN.md note is deliberately deferring that for a patch release, it would help to state what depends on the relationship snapshot being written.

resources/Table.ts
```diff
@@ -5233,15 +5255,31 @@ export function makeTable(options) {
...
+					harperLogger.warn?.(
+						`Discarded durable field "${name}" because it collides with a read-only resolver on "${collisionMetricPath}"`
+					);
+					recordAction(true, 'readonly-resolver-collision', collisionMetricPath);
+				} catch {}
+			};
 			for (const attribute of attributes) {
 				const name = attribute.name;
 				if (attribute.resolve) {
+					if (typeof attribute.set !== 'function') readOnlyResolverNames.add(name);
```
kriszyp: `typeof attribute.set !== 'function'` classifies one-to-many (`relationship.to`) attributes as writable because `resources/Table.ts:5140` installs a deliberately no-op setter for them — its comment says it exists only so a colliding stored property doesn't break loading, i.e. the same problem this PR solves. They are read-only in every meaningful sense, yet excluded here, so an `@enumerable` to-relationship still runs a `search` during durable encode and embeds the whole result array. Moot if durable mode stops surfacing resolvers entirely (see the comment at :595); otherwise mark that no-op setter (e.g. `attribute.readOnly = true`) so it lands in `readOnlyResolverNames`.

resources/RecordEncoder.ts
```diff
@@ -569,6 +599,25 @@ export class RecordEncoder extends StructonEncoder {
 			return null;
 		}
 	}
+	removeReadOnlyResolverFields(value) {
+		if (value == null || typeof value !== 'object' || ArrayBuffer.isView(value) || !this.readOnlyResolverNames?.size)
```
kriszyp: This runs on every value the encoder decodes, not just records: the shared-structures payload (`this.decode(...)` at `resources/RecordEncoder.ts:386` and `:419` — a Map, an array, or a `{named, typed}` object) and any array values. `Object.hasOwn` matches own keys of those shapes too: a computed attribute named `length` (or a numeric name) turns every decoded array into `Object.create(Array.prototype, descriptors-minus-length)` — no longer an array; one named `typed`/`named`/`structures` strips that key from the structures dictionary on load and breaks decoding for the table. Unlikely names, but cheap to rule out: gate on record shape (`Array.isArray` → skip; `value.constructor === Object || value instanceof RecordObject`), or move the cleaning into `withEntry` / `#withEntry` where only records pass.

resources/Table.ts
```diff
@@ -5233,15 +5255,31 @@ export function makeTable(options) {
...
 					Object.defineProperty(primaryStore.encoder.structPrototype, name, {
 						get() {
 							return attribute.resolve(this, contextStorage.getStore()); // it is only possible to get the context from ALS, we don't have a direct reference to the current context
 						},
 						set(related) {
-							return attribute.set(this, related);
+							if (typeof attribute.set === 'function') return attribute.set(this, related);
+							if (!Object.isExtensible(this)) throw new ClientError(`${name} is read-only`);
+							noteReadOnlyResolverCollision(name);
```
kriszyp: Two contracts now hang off one setter: frozen → `ClientError`, extensible → silently dropped plus a "Discarded durable field … collides" warning. The extensible branch is presumably for `withEntry`'s `Object.assign` (`resources/RecordEncoder.ts:666`, `resources/PrimaryRocksDatabase.ts:90`), but decode has already stripped collisions before those run, so in practice it fires for user code assigning to a computed attribute on an unfrozen struct-prototype object (e.g. the source-resolved `updatedRecord` after `resources/Table.ts:6203`, or an `Object.create(structPrototype)` copy) — silently ignored and mislabelled as a durable-data collision, while `resources/tracked.ts:48` throws for the identical assignment. Suggest throwing here unconditionally (matching tracked.ts) and letting decode own the legacy-data tolerance, or at least scoping the swallow to the load path.

unitTests/resources/crud.test.js
```diff
@@ -182,6 +226,95 @@ describe('CRUD operations with the Resource API', () => {
...
+				CRUDTable.setComputedAttribute('computed', (instance) => instance.name + ' computed');
+			}
+		});
+		it('keeps a no-change instance put clean in the primary store', async function () {
+			if (CRUDTable.loadAsInstance !== true) this.skip();
+			await CRUDTable.put({ id: 'instance-reput', name: 'instance', relatedId: 1 });
+			const instance = await CRUDTable.get('instance-reput');
+			await CRUDTable.put(instance);
+			const persisted = CRUDTable.primaryStore.getEntry('instance-reput').value;
+			assert.equal(Object.hasOwn(persisted, 'computed'), false);
```
kriszyp: `persisted` is the decoded value, and decode now strips `computed` (and the setter drops it during `Object.assign`), so this assertion — and the audit one at :275, which goes through the same decoder — passes even if the bytes written by the instance put still contain `computed`. `persisted.computed === 'instance computed'` can't tell either: a stale stored copy would hold the same string the resolver returns. To pin the encode side, assert on the raw bytes (`valueAsBuffer` / `getBinary`) not containing `'instance computed'`, or count resolver invocations the way the previous test does.

resources/Table.ts
```diff
@@ -5233,15 +5255,31 @@ export function makeTable(options) {
...
+			const collisionMetricPath = databaseName + '.' + tableName;
+			let reportedReadOnlyResolverCollision = false;
+			const noteReadOnlyResolverCollision = (name: string) => {
+				if (reportedReadOnlyResolverCollision) return;
+				reportedReadOnlyResolverCollision = true;
+				try {
+					harperLogger.warn?.(
+						`Discarded durable field "${name}" because it collides with a read-only resolver on "${collisionMetricPath}"`
+					);
+					recordAction(true, 'readonly-resolver-collision', collisionMetricPath);
```
kriszyp: nit: `reportedReadOnlyResolverCollision` latches both the warning and the analytics counter, so `readonly-resolver-collision` records exactly one hit per table per process regardless of how many affected records are read. `decode-missing-structure` (`resources/RecordEncoder.ts:589`) counts every occurrence, which is the useful shape for judging how much legacy data is out there. Latch the log, count the metric.

resources/RecordEncoder.ts
```diff
@@ -36,6 +36,30 @@ import { CONFIG_PARAMS } from '../utility/hdbTerms.ts';
 import * as envMngr from '../utility/environment/environmentManager.js';
 
 const StructonEncoder = createStructon(Encoder) as typeof Encoder;
+// Source and built module copies can coexist, so they must share the durable-encoding scope.
+const DURABLE_ENCODING_DEPTH = Symbol.for('harper.durableEncodeDepth');
```
kriszyp, line 44: nit: sharing the depth via `globalThis[Symbol.for(...)]` is justified by source and built module copies coexisting, but every other encode-scoped global in this file (`timestampNextEncoding`, `metadataInNextEncoding`, `lastValueEncoding`, …) is a module-level `let` that would be just as split in that scenario. Either the scenario is real and those need the same treatment (a note on which import path creates the duplicate would help), or a module-level `let` is enough here. A lone global symbol reads as papering over a module-duplication problem that belongs at the loader.

resources/graphql.ts
```diff
@@ -351,4 +351,4 @@ async function processGraphQLSchema(gqlContent, urlPath, filePath, resources) {
 }
 
 // useful for testing
-export const loadGQLSchema = (content) => processGraphQLSchema(content, null, null, new Resources());
+export const loadGQLSchema = (content) => processGraphQLSchema(content, null, '<inline-schema>', new Resources());
```
kriszyp, line 354: nit: unrelated drive-by — this changes `loadGQLSchema`'s `filePath` from `null` to a placeholder (presumably so `new Script(..., { filename })` at :339 stops throwing for inline schemas using `@computed(from:)`). Nothing in this PR's tests calls it; worth a line in the description or its own commit.

kriszyp:
Nice work on this — the decode-side cleanup in `removeReadOnlyResolverFields` is a better placement than filtering at each materialization site, since it sits upstream of all three of them, and the byte-level `encoded.includes(Buffer.from('forged'))` plus the resolver-call counter genuinely pin the write side against a future refactor dropping the hook.

I had independently investigated [#2359](https://github.com/HarperFast/harper/issues/2359) before finding this PR, so rather than re-review the diff I ran my reproducers against this head (`49394d979`) and against `origin/main` for comparison. Everything below is measured, not read off the diff.

## Confirmed working on this branch

| case | result |
| --- | --- |
| Cache fill + scalar `@computed`, second read | 200 |
| Durable bytes after a cache fill | clean — `["id","price","discount"]`, no `salePrice` |
| Record already poisoned by 5.2.0–5.2.6 | readable, resolver authoritative |
| Invalidation route (`@computed @indexed`, no cache source, no replication) | fixed |
| Partial update re-encoding an existing record | no stored-field loss |
| Query / `getRange` materialization | clean |
| Mutating a computed field | rejected with a clear error |

Worth noting that third row, because it is a route the issue does not mention and this PR has no test for: `invalidate()` is not gated on a cache source, and `_writeInvalidate` preserves `getProperty(name)` for **every** indexed attribute so the row stays searchable — which for a `@computed @indexed` attribute stores the computed value. On `origin/main` that gives a 500 on the next read of the row with no cache source and no replication involved. Your fix handles it via the decode cleanup; it just isn't pinned by a test.

## One gap I would like to see closed

An `@enumerable @relationship` whose foreign key does not resolve still produces the same permanently-unreadable record, on this branch, with new data:

```
FILL:        200 {"id":"x1","label":"from source","catId":"no-such-cat"}
SECOND READ: 500 TypeError: Cannot read properties of undefined (reading 'getId')
QUERY:       200 [{"error":"TypeError","message":"Cannot read properties of undefined (reading 'getId')"}]
```

Fixture: a `sourcedFrom` table with `cat: Cat @relationship(from: "catId") @enumerable` where the source returns a `catId` with no matching row. The cache-fill record is struct-prototyped, so its durable encode goes through the response projection and bakes the resolved relationship value in; the next materialization then calls the relationship setter with it. An ordinary REST `PUT` does **not** reproduce it (plain object, so `toJSON` is never consulted) — it needs the struct-prototyped path, which is the same route the original defect took.

`origin/main` behaves identically, so this is **pre-existing and not a regression from this PR** — it is the sibling half of the same defect class. Planting the three shapes a stored collision can take, under the relationship name:

| stored under the relationship name | this branch |
| --- | --- |
| `null` (what the projection writes for a dangling FK) | 500, record unreachable |
| embedded record object | 200, `catId` intact |
| scalar | 200, but `catId` is silently destroyed |

That last row is the one I would flag hardest: `related.getId?.() \|\| related[primaryKey]` on a scalar yields `undefined`, which is then assigned to the foreign key, so the FK disappears from the record on read — and becomes durable on the next write.

`DESIGN.md` and the PR description frame writable-relationship behavior as intentionally retained for the patch line, and that is a defensible call for *serialization semantics*. My suggestion is only that the **materialization tolerance** is a separable question from those semantics: `readOnlyResolverNames` holds resolvers with no `.set`, so relationship names are neither excluded from the durable projection nor cleaned at decode. Folding **all** resolver-owned names into the decode cleanup (a second set alongside the read-only one, used only by `removeReadOnlyResolverFields`) closes the class without changing what a relationship setter does when user code actually assigns to it.

## Three smaller notes

1. **A source-supplied computed value is echoed exactly once.** With a source returning `{price: 100, discount: 40, salePrice: 999}`, the cache-fill response reports `salePrice: 999` and every later read reports `60`. Storage is correctly clean — it is only the fill response. Projecting the resolved record in `getFromSource` before it is exposed fixes it, and it matters more than cosmetics if a resolver is ever context- or role-dependent.
2. **`attribute.set` is not reset on schema reload.** `readOnlyResolverNames` is derived from `typeof attribute.set !== 'function'`, but `updatedAttributes()` only resets `attribute.resolve`, so an attribute changing from `@relationship` to `@computed` keeps the old setter and never enters the read-only set. A one-line `attribute.set = null` next to the existing reset covers it.
3. **Hot-path acknowledgement**, not an objection: the `globalThis` depth counter plus `try/finally` runs on every encode on every store, including internal DBIs and index stores, and `toJSON` is now a getter consulted on every response serialization and every struct encode. I have no benchmark showing it matters, and I doubt it does next to a msgpackr encode plus a storage write — just worth being deliberate about, since it is a shared write path.

## Offer

I have a seven-test integration suite covering the routes this PR's unit tests do not reach — cache fill over REST, a raw-stored-keys probe that asserts the durable bytes directly (with a non-vacuity check that the probe can see a planted collision), legacy-record recovery, the invalidation route, re-encode of an existing record, and the source-collision case. Six of the seven fail on `origin/main` with the production error. Happy to hand it over as a patch or a follow-up PR onto your branch, whichever is less disruptive — and equally happy for the relationship cleanup to be a follow-up issue rather than something that holds this up, since what it fixes is already broken in 5.2.x today.

— KrAIs (Claude Opus 5), on behalf of Kris

kriszyp (commented):
Thank you for catching this, this looks really good. A couple things to address and then let's get this merged/patched.

resources/Table.ts
```diff
@@ -5233,15 +5267,55 @@ export function makeTable(options) {
...
+					harperLogger.warn?.(
+						`Preserved durable field "${name}" because it collides with a settable resolver on "${collisionMetricPath}"; an older peer or schema may have persisted a relationship snapshot`
+					);
+				} catch {}
+			};
 			for (const attribute of attributes) {
 				const name = attribute.name;
 				if (attribute.resolve) {
+					resolverNames.push(name);
+					if (typeof attribute.set !== 'function') readOnlyResolverNames.add(name);
```
kriszyp: Carried from last round, now with much smaller impact since durable encoding no longer resolves anything: `typeof attribute.set !== 'function'` still classifies one-to-many (`relationship.to`) attributes as writable because `resources/Table.ts:5140` installs a deliberately no-op setter for them. Practical effects today: (1) a legacy own `childrenOfSelf` array is *preserved* (settable path) rather than stripped, and shadows the `search` on read; (2) `record.children = [...]` is silently ignored while `record.computed = x` throws `ClientError`. Marking that setter (e.g. `attribute.readOnly = true`) so the name lands in `readOnlyResolverNames` gives both cases the read-only treatment. Moot if you decide settable snapshots should be dropped on read anyway (see the `promoteRecord` comment).

resources/RecordEncoder.ts
```diff
@@ -568,6 +635,87 @@ export class RecordEncoder extends StructonEncoder {
...
+					for (let i = 0; i < this.readOnlyResolverNames.length; i++) {
+						const name = this.readOnlyResolverNames[i];
+						if (Object.hasOwn(descriptors, name)) {
+							delete descriptors[name];
+							this.onReadOnlyResolverCollision?.(name);
+							hasCollision = true;
+						}
+					}
+					if (hasCollision) {
+						delete descriptors.toJSON;
```
kriszyp: nit: `toJSON` is only removed from the per-structure prototype when the shape has a collision. Clean typed-struct shapes keep structon's own `toJSON` (`node_modules/structon/struct.js:1149`), which shadows the table `toJSON` on `structPrototype`, so `@enumerable`/computed-scalar attributes surface in `JSON.stringify` output for a cleaned lazy record but not for a clean one from the same table — the test at `unitTests/resources/crud.test.js:200-202` shows the dirty case working. The clean-case gap is pre-existing, but this PR makes the two diverge. Since the replacement prototype is cached per shape anyway, replacing (or just deleting `toJSON` on) every typed shape, not only dirty ones, would make serialization uniform; otherwise worth a TODO so the randomAccessFields JSON gap is tracked.

resources/Table.ts
```diff
@@ -6135,32 +6221,68 @@ export function makeTable(options) {
...
+							(sourcePrototype && Object.getPrototypeOf(sourcePrototype) === primaryStore.encoder.structPrototype);
+						if (
+							sourceRecord[STORED_FIELD_NAMES] ||
+							Object.hasOwn(sourceRecord, STRUCT_SOURCE) ||
+							(typeof toJSON === 'function' && !usesTargetPrototype)
+						) {
+							responseRecord = typeof toJSON === 'function' ? toJSON.call(sourceRecord) : sourceRecord;
+						} else responseRecord = sourceRecord;
+						updatedRecord = projectRecordForDurableEncoding(sourceRecord);
+						if (needsMutableRecordCopy(responseRecord)) responseRecord = { ...responseRecord };
```
kriszyp: `{ ...responseRecord }` (and `{ ...updatedRecord }` at :6246, `{ ...recordUpdate }` at :2489) is a field-losing operation on a struct-mode record: a structon lazy instance has no own string keys — every stored field is an enumerable getter on the per-structure prototype — so spreading a frozen one yields `{ [Symbol.for('source')]: … }` and nothing else. `freezeRecord` (`resources/Table.ts:253`) freezes lazy instances in place, and `frozenRecordView`'s comment at :261-263 already calls out that a copy "breaks a caching table's lazy structPrototype decode".

Reachable here: a `randomAccessFields` table whose stored shape is *dirty* (so `removeReadOnlyResolverFields` swapped in the clean prototype and deleted structon's `toJSON`) but has no `@enumerable`/computed-scalar attribute (so the table getter returns `undefined` outside durable mode). On a 304 revalidation `toJSON` is not a function, the `STRUCT_SOURCE` branch at :6230-6235 falls through to `responseRecord = sourceRecord`, the frozen lazy instance hits this spread, and the caller gets a record with only `id`/timestamps. The replay path at :2489 has the same shape for any frozen struct-mode record (pre-existing, but this PR owns the helper now).

Suggest: in `needsMutableRecordCopy` callers, materialize instead of spreading when `Object.hasOwn(value, STRUCT_SOURCE)` — e.g. build from `storedFieldNames`/`STORED_FIELD_NAMES` the way `durableToJSON` does — or route every lazy source record through the `toJSON.call` branch by falling back to the durable projection when the response `toJSON` is absent.

unitTests/resources/caching.test.js
```diff
@@ -86,6 +97,45 @@ describe('Caching', () => {
...
+		const relatedDefinition = {};
+		const RevalidatedRelatedTable = table({
+			table: 'RevalidatedRelatedTable',
+			database: 'test',
+			attributes: [{ name: 'id', isPrimaryKey: true }, { name: 'name' }],
+		});
+		RevalidatedTable = table({
+			table: 'RevalidatedTable',
+			database: 'test',
+			randomAccessFields: true,
```
kriszyp: nit: making `RevalidatedTable` struct-mode means the 304 write-back test now only exercises the `STRUCT_SOURCE` → `toJSON.call` branch at `resources/Table.ts:6230-6235`. The classic-table branch it was originally written for (`responseRecord = sourceRecord` → frozen record → spread copy → `setPrototypeOf`) is now untested, and it is the default storage mode. A sibling table without `randomAccessFields` (or a loop over both) running the same `assertResolvedFieldsAreNotDurable` + `entryMap` assertions would keep both branches pinned.

resources/RecordEncoder.ts
```diff
@@ -36,6 +36,55 @@ import { CONFIG_PARAMS } from '../utility/hdbTerms.ts';
...
+	} finally {
+		leaveDurableRecordEncoding();
+	}
+}
+
+export function promoteRecord(encoder, value) {
+	const promoted = new encoder.structPrototype.constructor();
+	const resolverNames = encoder.resolverNames;
+	for (let i = 0; i < resolverNames.length; i++) {
+		if (Object.hasOwn(value, resolverNames[i])) {
```
kriszyp: Carried from last round — behaviour question on the settable-collision path. When a legacy classic row carries an embedded relationship snapshot (own `related`), this copies it with `defineProperties`, so it stays an own data property that shadows the resolver getter: `record.related` returns the stored `{ id: 99, … }` snapshot forever, while `record.relatedId` is `1` — `unitTests/resources/crud.test.js:330-331` pins exactly that split. Before the PR the setter consumed the snapshot (badly — it clobbered the FK), but the read then resolved live. Combined with the durable getter at `resources/Table.ts:671-677`, which deliberately writes the own key back out, the stale snapshot is served on every read and re-persisted on every write, with only a once-per-table warning.

If the goal is "don't lose data an older peer wrote", consider keeping the FK authoritative for reads: drop the own settable key here (or project it to the FK only when the FK is absent) and let the resolver run. If serving the snapshot is intended, the warning text should say reads will return it, since "Preserved durable field" reads as a storage-only statement.

resources/Table.ts
```diff
@@ -5233,15 +5268,55 @@ export function makeTable(options) {
...
+				try {
+					harperLogger.warn?.(
+						`Preserved durable field "${name}" because it collides with a settable resolver on "${collisionMetricPath}"; an older peer or schema may have persisted a relationship snapshot`
+					);
+				} catch {}
+			};
 			for (const attribute of attributes) {
 				const name = attribute.name;
 				if (attribute.resolve) {
+					resolverNames.push(name);
```
kriszyp: Carried from last round, with much smaller impact since durable encoding no longer resolves anything: `typeof attribute.set !== 'function'` still classifies one-to-many (`relationship.to`) attributes as writable because `resources/Table.ts:5140` installs a deliberately no-op setter for them. Practical effects today: (1) a legacy own `childrenOfSelf` array is *preserved* (settable path) rather than stripped, and shadows the `search` on read; (2) `record.children = [...]` is silently ignored while `record.computed = x` throws `ClientError`. Marking that setter (e.g. `attribute.readOnly = true`) so the name lands in `readOnlyResolverNames` gives both cases the read-only treatment. Moot if settable snapshots are dropped on read (see the `promoteRecord` comment).

resources/RecordEncoder.ts
kriszyp: nit (carried): `toJSON` is only removed from the per-structure prototype when the shape has a collision. Clean typed-struct shapes keep structon's own `toJSON` (`node_modules/structon/struct.js:1149`), which shadows the table `toJSON` on `structPrototype`, so `@enumerable`/computed-scalar attributes surface in `JSON.stringify` output for a cleaned lazy record but not for a clean one from the same table. The clean-case gap is pre-existing, but this PR makes the two diverge — and it is also what forces the response/durable branching on `STRUCT_SOURCE` in `resources/Table.ts:6230-6235`. Since the replacement prototype is cached per shape anyway, replacing (or deleting `toJSON` on) every typed shape would make serialization uniform; otherwise a TODO so the randomAccessFields JSON gap is tracked.

resources/RecordEncoder.ts
```diff
@@ -36,6 +36,55 @@ import { CONFIG_PARAMS } from '../utility/hdbTerms.ts';
 import * as envMngr from '../utility/environment/environmentManager.js';
 
 const StructonEncoder = createStructon(Encoder) as typeof Encoder;
+export const STRUCT_SOURCE = Symbol.for('source');
+export const STORED_FIELD_NAMES = Symbol.for('harper.storedFieldNames');
+// Source and built module copies can coexist, so they must share the durable-encoding scope.
+const DURABLE_ENCODING_DEPTH = Symbol.for('harper.durableEncodeDepth');
```
kriszyp: nit (carried): sharing the depth via `globalThis[Symbol.for(...)]` is justified by source and built module copies coexisting, but every other encode-scoped global in this file (`timestampNextEncoding`, `metadataInNextEncoding`, `lastValueEncoding`, …) is a module-level `let` that would be just as split in that scenario — and `STRUCT_SOURCE`/`STORED_FIELD_NAMES` are now exported global symbols while `#cleanTypedResolverPrototypes` is per instance. Either the scenario is real and those need the same treatment (a note on which import path creates the duplicate would help), or a module-level `let` is enough here.

resources/graphql.ts
kriszyp: nit (carried): unrelated drive-by — this changes `loadGQLSchema`'s `filePath` from `null` to a placeholder (presumably so `new Script(..., { filename })` at :339 stops throwing for inline schemas using `@computed(from:)`). Nothing in this PR's tests calls it; worth a line in the description or its own commit.

kriszyp:
Unfortunately, I think this is fundamentally the wrong approach. The durable serialization of data should be the default and the external projection to include computed properties should be the opt-in, which completely reverses the entire direction of this, sorry. I've started on a PR to fix this, and will probably need to close this PR.

resources/Table.ts
kylebernhardy: Addressed at the durability boundary. Durable encoding no longer evaluates inherited response relationships. We intentionally preserve an existing own snapshot only for settable relationships as a mixed-version compatibility rule; the response projection removes that own value and uses the live relationship getter without rewriting the foreign key. Raw-byte and relationship/FK regressions pin both sides.\n\nComment generated by kAIle (Codex GPT-5.6)

resources/RecordEncoder.ts
kylebernhardy: Fixed. Cleanup now handles typed and lazy struct records without materializing unrelated stored getters, preserves random-access backing data, and remains stable across repeated decodes. The regression suite now uses the real typed writer, checks lazy and frozen source shapes, and asserts the forged value is absent from re-encoded bytes.\n\nComment generated by kAIle (Codex GPT-5.6)

resources/Table.ts
kylebernhardy: Fixed. User assignment to a read-only resolved field now throws `ClientError` consistently for extensible and tracked records. Legacy collision tolerance is owned by decode cleanup rather than the public setter.\n\nComment generated by kAIle (Codex GPT-5.6)

resources/RecordEncoder.ts
kylebernhardy: We kept the shared numeric scope deliberately. Source and compiled core modules coexist in the Pro process, and both can participate in the same nested encode, so a module-local holder can disagree about durable mode. The final implementation limits response authorization to a one-shot encoder-local request; only the durable and partial scopes remain shared. This is called out as an accepted performance tradeoff in the PR description.\n\nComment generated by kAIle (Codex GPT-5.6)

resources/Table.ts
kylebernhardy: Fixed. One-to-many attributes are classified as read-only through `relationship.to` even though their compatibility setter is present. Focused tests distinguish one-to-many and one-to-one classifications and prevent durable relationship searches from returning through this path.\n\nComment generated by kAIle (Codex GPT-5.6)

resources/Table.ts
kylebernhardy: Fixed. Collision metrics are counted once per resolver name while warning output remains latched once per table/process to avoid hot-path log floods. Tests cover the resulting classification and cleanup behavior.\n\nComment generated by kAIle (Codex GPT-5.6)

unitTests/resources/crud.test.js
kylebernhardy: Fixed. The durability regression now inspects raw stored and audit bytes, so it fails if a computed response value is serialized even when decode would later hide it.\n\nComment generated by kAIle (Codex GPT-5.6)

resources/graphql.ts
kylebernhardy: Kept and now covered. The Pro inline computed-schema regression calls `loadGQLSchema` without a file path, and `vm.compileFunction` requires a usable filename for that path. The fallback is included in the PR summary and the exact Pro unit suite passes.\n\nComment generated by kAIle (Codex GPT-5.6)

resources/RecordEncoder.ts
kylebernhardy: Fixed. Resolver cleanup now excludes arrays, ArrayBuffer views, and shared structure dictionaries, including a regression for an attribute named `length`. Typed record handling is separate and preserves struct backing data.\n\nComment generated by kAIle (Codex GPT-5.6)

resources/Table.ts
```diff
@@ -6135,32 +6237,106 @@ export function makeTable(options) {
...
+						) {
+							responseProjectionSource = promoteRecord(primaryStore.encoder, cleanSourceRecord);
+							toJSON = responseProjectionSource.toJSON;
+						}
+						if (
+							responseProjectionSource[STORED_FIELD_NAMES] ||
+							Object.hasOwn(responseProjectionSource, STRUCT_SOURCE) ||
+							typeof toJSON === 'function'
+						) {
+							if (typeof toJSON === 'function') responseRecord = toJSON.call(responseProjectionSource);
```
kriszyp: Two things in this branch, one wasteful and one wrong.

**Wasteful:** for any target-prototype source — every 304 revalidation of a classic or dirty-lazy record on a table with `@enumerable`/computed-scalar attributes — `toJSON.call(...)` is the *response* projection: it runs every enumerable relationship resolver (a primary-store read, or a `search` for to-relationships) and every computed, only for `:6279-6281` to `delete` the results again. The `STORED_FIELD_NAMES` branch right below already shows the cheaper shape: copy `storedFieldNames(responseProjectionSource)` and let the accessors stay inherited. Using that for all `usesTargetPrototype` sources removes the resolver work and the `materializedTargetProjection` bookkeeping.

**Wrong:** `usesTargetPrototype` (`:6245`) is computed from the *original* source prototype. When a plain source has a read-only collision, `promoteRecord` at `:6261` swaps in a target-prototype record and `toJSON` becomes the target response projection, but `usesTargetPrototype` is still `false`, so `materializedTargetProjection` is `false` and the deletion is skipped. `removeReadOnlyResolverFields` at `:6282` then strips `computed` but leaves the materialized settable ones: with `RevalidatedTable` and a source that sends `computed: 'forged'`, the response carries an own `related` snapshot (resolved at fill time) that shadows the getter once `setPrototypeOf` runs at `:6338`. The `uses the resolver instead of a source-provided computed collision` test only checks `computed`, so it doesn't see this. Simplest fix is the stored-field copy above; otherwise recompute `usesTargetPrototype` (or set `materializedTargetProjection = true`) in the promote branch.

unitTests/resources/caching.test.js
```diff
@@ -170,6 +239,119 @@ describe('Caching', () => {
...
+			returnRelationshipCollision = true;
+			await RevalidatedTable.invalidate('relationship-collision');
+			const response = await RevalidatedTable.get('relationship-collision');
+			assert.equal(response.related.name, 'snapshot relationship');
+			assert.equal(Object.hasOwn(response, 'related'), true);
+			await RevalidatedTable.primaryStore.committed;
+			const persisted = await waitFor(() => RevalidatedTable.primaryStore.getEntry('relationship-collision')?.value, {
+				message: 'the relationship source record must commit to the cache',
+			});
+			assert.equal(persisted.related.name, 'relationship');
```
kriszyp: I can't get these two assertions plus the byte check at :311-316 to hold at the same time from the code as I read it, so this needs either a fix or a comment explaining the mechanism. `RevalidatedTable` is `randomAccessFields: true`; the durable record is the plain source object with an own `related` snapshot (the split at `resources/Table.ts:6291-6296` copies the *response*, not `updatedRecord`), and the bytes do contain it (asserted). On decode:

- struct mode → `related` is an enumerable getter on the per-structure prototype (structon `struct.js:1140`), which sits *above* `structPrototype`, so `hasOwn` is `false` (✓) but `persisted.related.name` should be `'snapshot relationship'`, not `'relationship'` — `removeReadOnlyResolverFields` only strips read-only names and `related` is settable;
- classic fallback → plain object with own `related` → `promoteRecord` (`resources/RecordEncoder.ts:105`) takes the `defineProperties` path and keeps it own, so `hasOwn` would be `true`.

If the suite passes, something else is discarding the snapshot on read (which would be great — it's what I asked for on `promoteRecord` — but then the DESIGN.md "preserved verbatim / patch-release semantics" statement and the `promoted.related` assertions in `crud.test.js:330-331` describe a different behaviour). Please pin down which one is real and assert it explicitly on both storage modes.

resources/RecordEncoder.ts
```diff
@@ -36,6 +36,84 @@ import { CONFIG_PARAMS } from '../utility/hdbTerms.ts';
...
+	enterDurableRecordEncoding();
+	try {
+		const toJSON = record?.toJSON;
+		return typeof toJSON === 'function' ? toJSON.call(record) : record;
+	} finally {
+		leaveDurableRecordEncoding();
+	}
+}
+
+export function promoteRecord(encoder, value) {
```
kriszyp: Carried — behaviour question on the settable-collision path. When a legacy classic row carries an embedded relationship snapshot (own `related`), this copies it with `defineProperties`, so it stays an own data property that shadows the resolver getter: `record.related` returns the stored `{ id: 99, … }` snapshot forever, while `record.relatedId` is `1` — `unitTests/resources/crud.test.js:330-331` pins exactly that split. Before the PR the setter consumed the snapshot (badly — it clobbered the FK), but the read then resolved live. Combined with the durable getter, which deliberately writes the own key back out, the stale snapshot is served on every read and re-persisted on every write, with only a once-per-table warning.

If the goal is "don't lose data an older peer wrote", consider keeping the FK authoritative for reads: drop the own settable key here (or project it to the FK only when the FK is absent) and let the resolver run. If serving the snapshot is intended, the warning text should say reads will return it. See also the caching-test question — that test appears to assert the opposite outcome.

resources/Table.ts
```diff
@@ -2451,7 +2495,7 @@ export function makeTable(options) {
 							// validate() coerces values and we stamp created/updated times + the primary key
 							// below, so copy-on-mutate when recordUpdate is frozen (e.g. a record decoded during
 							// log replay) instead of writing through the frozen object.
-							if (isFrozenRecordObject(recordUpdate)) recordUpdate = { ...recordUpdate };
+							if (needsMutableRecordCopy(recordUpdate)) recordUpdate = { ...recordUpdate };
```
kriszyp: Carried (narrowed): the cache-fill half is fixed, but this replay-path spread still has the struct-mode hazard. A structon lazy instance has no own string keys — every stored field is an enumerable getter on the per-structure prototype — so `{ ...recordUpdate }` on a frozen one yields `{ [Symbol.for('source')]: … }` and the replayed write stores an empty record. `freezeRecord` (`resources/Table.ts:257`) freezes lazy instances in place, so a `randomAccessFields` table's records decoded during log replay can arrive here frozen. Same fix as the cache path: materialize via `storedFieldNames`/`STORED_FIELD_NAMES` when `Object.hasOwn(value, STRUCT_SOURCE)`, or make `needsMutableRecordCopy` callers share one helper that does.

resources/Table.ts
```diff
@@ -5260,12 +5345,28 @@ export function makeTable(options) {
 					if (attribute.computed && !relationDef) hasSurfacedComputed = true;
 				}
 			}
+			const resolverContractChanged =
+				primaryStore.encoder.resolverNames.length !== resolverNames.length ||
+				primaryStore.encoder.resolverNames.some((name, index) => name !== resolverNames[index]) ||
+				primaryStore.encoder.readOnlyResolverNames.length !== readOnlyResolverNames.size ||
+				primaryStore.encoder.readOnlyResolverNames.some((name) => !readOnlyResolverNames.has(name));
+			primaryStore.encoder.setReadOnlyResolverNames(readOnlyResolverNames, resolverNames);
+			if (resolverContractChanged) primaryStore.clearRecordCache?.();
```
kriszyp: nit (carried): `clearRecordCache` is only defined on `PrimaryRocksDatabase`, so on LMDB a resolver-contract change on reload leaves lmdb-js's record cache (`cache: { validated: true }`, `utility/lmdb/OpenDBIObject.ts:53`) holding records cleaned under the old name set — the exact staleness this guard exists to prevent on RocksDB. The latest test tweak (`unitTests/resources/crud.test.js:191-193`) makes the spy tolerate the missing method, so the tests now pass on LMDB while asserting only that the *call* happened, not that any cache was cleared. If lmdb-js exposes its cache (`primaryStore.cache?.clear?.()`), clearing it here keeps the two engines equivalent and lets the test wrap a real method on both; otherwise a comment saying LMDB is knowingly excluded.

unitTests/resources/caching.test.js
```diff
@@ -86,6 +107,54 @@ describe('Caching', () => {
...
+		const relatedDefinition = {};
+		const RevalidatedRelatedTable = table({
+			table: 'RevalidatedRelatedTable',
+			database: 'test',
+			attributes: [{ name: 'id', isPrimaryKey: true }, { name: 'name' }],
+		});
+		RevalidatedTable = table({
+			table: 'RevalidatedTable',
+			database: 'test',
+			randomAccessFields: true,
```
kriszyp: nit (carried): making `RevalidatedTable` struct-mode means every caching test here exercises the lazy branches of the cache-fill projection. The classic-table path (`responseRecord = toJSON.call(existingRecord)` → delete → spread copy → `setPrototypeOf` at `resources/Table.ts:6269-6338`) is the default storage mode and is now untested for 304s, collisions and stateful projections alike. A sibling table without `randomAccessFields` (or a loop over both) running the same assertions would keep both branches pinned — and would have surfaced the `persisted.related` question.

resources/RecordEncoder.ts
```diff
@@ -568,6 +666,89 @@ export class RecordEncoder extends StructonEncoder {
...
+					for (let i = 0; i < this.readOnlyResolverNames.length; i++) {
+						const name = this.readOnlyResolverNames[i];
+						if (Object.hasOwn(descriptors, name)) {
+							delete descriptors[name];
+							this.onReadOnlyResolverCollision?.(name);
+							hasCollision = true;
+						}
+					}
+					if (hasCollision) {
+						delete descriptors.toJSON;
```
kriszyp: nit (carried): `toJSON` is only removed from the per-structure prototype when the shape has a collision. Clean typed-struct shapes keep structon's own `toJSON` (`node_modules/structon/struct.js:1149`), which shadows the table `toJSON` on `structPrototype`, so `@enumerable`/computed-scalar attributes surface in `JSON.stringify` output for a cleaned lazy record but not for a clean one from the same table. The clean-case gap is pre-existing, but this PR makes the two diverge — and it is what forces the `STRUCT_SOURCE` branching in the cache-fill path. Since the replacement prototype is cached per shape anyway, replacing (or deleting `toJSON` on) every typed shape would make serialization uniform; otherwise a TODO so the randomAccessFields JSON gap is tracked.

resources/RecordEncoder.ts
```diff
@@ -36,6 +36,84 @@ import { CONFIG_PARAMS } from '../utility/hdbTerms.ts';
 import * as envMngr from '../utility/environment/environmentManager.js';
 
 const StructonEncoder = createStructon(Encoder) as typeof Encoder;
+export const STRUCT_SOURCE = Symbol.for('source');
+export const STORED_FIELD_NAMES = Symbol.for('harper.storedFieldNames');
+export const INVALIDATED = 1;
+export const EVICTED = 8; // 2 is reserved for timestamps
+// Source and built module copies can coexist, so they must share the durable-encoding scope.
+const DURABLE_ENCODING_DEPTH = Symbol.for('harper.durableEncodeDepth');
```
kriszyp: nit (carried): sharing the depth via `globalThis[Symbol.for(...)]` is justified by source and built module copies coexisting, but every other encode-scoped global in this file (`timestampNextEncoding`, `metadataInNextEncoding`, `lastValueEncoding`, …) is a module-level `let` that would be just as split in that scenario — and there are now three global symbols (`DURABLE_ENCODING_DEPTH`, `PARTIAL_PROJECTION_ENCODING_DEPTH`, `RESPONSE_ENCODING_REQUESTS`) while `#cleanTypedResolverPrototypes` stays per instance. Either the scenario is real and those need the same treatment (a note on which import path creates the duplicate would help), or module-level `let`s are enough.

resources/graphql.ts
kriszyp: nit (carried): with the `filename: filePath ?? '<inline-schema>'` fallback now at `:344`, this call-site change is redundant — `null` would work again and keep the `console.error(... in ${filePath})` at `:306` reading `null` rather than a fake path. Suggest reverting this line and keeping only the Script-level fallback.

kriszyp:
Closing this in favor of [#2368](https://github.com/HarperFast/harper/pull/2368), per Kris — with genuine thanks: this PR's diagnosis was right on the money, and its decode-side cleanup (`removeReadOnlyResolverFields` as a single choke point upstream of materialization) was the best-placed idea of its round; #2368's review history says as much. The independent two-node copy verification here was also more end-to-end than anything #2368 runs single-node.

What tipped the decision to the other implementation:

- it also covers the **writable-relationship half** of the defect (present since 5.1.0 — a stored `null` under an `@enumerable @relationship` crashes materialization the same way, and a scalar collision silently destroys the foreign key), which this PR's read-only-resolver scoping left live;
- the write side projects at `recordUpdater`, so the audit `encodedRecord`/transaction log/replication payload and own-key contamination from source and peer payloads are covered by the same boundary, with no per-encode scope flag;
- it lands on v5.2 and stays in 5.3 unchanged, avoiding a patch-line/main fork in serialization internals.

The invalidation route (`@computed @indexed` + `invalidate()` — no cache source, no replication needed) turned out to be a third way into the same defect; #2368 carries a regression test for it. A Pro-side initial-copy round-trip test in the spirit of the one built here would still be a valuable follow-up to #2359's Pro work.

— Claude (Opus 5), on behalf of Kris

PR #2373, opened 2026-08-28
Make the HNSW routing test prove the graph and the descent are worth testing

### LINKED ISSUE #2372: Flaky unit test: HNSW greedy routing (ROUTING_EF) intermittently returns a shifted neighbour set
## Summary

`HNSW greedy routing above layer 0 (ROUTING_EF) — returns the same neighbours as searching every layer at the full ef` fails intermittently in CI with the result set shifted wholesale, e.g.:

```
AssertionError: greedy descent changed the result set for target 0
+ '561,562,550,551,563,572,552,573,564,574'
- '0,1,11,2,12,13,3,22,14,4'
```

Observed twice on HarperFast/harper#2368 within one day, on different Node versions (v22 once, v24 once) while the sibling matrix jobs passed on identical code; both cleared on automatic retry. The PR's diff has no path into HNSW stores (index-store encoders carry no resolver metadata), and the failure predates it in character — the wholesale shift with contiguous id runs suggests a distance-tie ordering or unseeded-randomness sensitivity in graph construction rather than a routing defect.

## Impact

Reds an otherwise-green unit matrix job roughly once per couple of pushes, costing a retry cycle and training people to ignore unit failures.

## Next step

Seed or de-randomize the fixture graph (or assert set-equality with deterministic tie-breaking) so the greedy-vs-full-ef comparison is stable across Node versions.

kriszyp:
Rebased this PR onto `main` at `3e8e8d183`. `main` now includes the separate LMDB audit-head implementation fix, so this branch contains only the [unresolved-reference regression guard](https://github.com/HarperFast/harper/pull/2373/changes?w=1#diff-c107f30ee51faa2fc3de638ed3e55e83e4ed7321cafb07e6c4e828bd046b028dR125) and the HNSW guards that keep a seeded greedy-routing comparison meaningful: a [one-draw-per-node graph pin](https://github.com/HarperFast/harper/pull/2373/changes?w=1#diff-cad9073daf8c1f98872bbf9e6517826faa9fcc099c2010ba808180e42ff5ca9aR938), an [independent `ROUTING_EF` assertion](https://github.com/HarperFast/harper/pull/2373/changes?w=1#diff-cad9073daf8c1f98872bbf9e6517826faa9fcc099c2010ba808180e42ff5ca9aR1023), and a [descent-sensitivity control](https://github.com/HarperFast/harper/pull/2373/changes?w=1#diff-cad9073daf8c1f98872bbf9e6517826faa9fcc099c2010ba808180e42ff5ca9aR1057).

Fixes #2372

## For the human reviewer

1. The test pins `ROUTING_EF` independently rather than exporting the production-private constant: accidental production drift fails the test, while intentional tuning requires a deliberate paired test update. Exporting it would make the test and implementation drift together; changing this remains local and reversible.
2. The control requires at least one selected graph to depend on descent instead of asserting a broader recall-quality target. That preserves deterministic coverage of the original greedy-vs-full comparison; a recall criterion needs separate calibration and is not part of this rebase.

## Verification

`npm run build` passed. `npm run test:unit:resources` passed. `npm run lint:required` passed. The affected behavior is exercised end-to-end within the resource unit suite through real table inserts and vector searches; no production behavior changed in this rebase.

Complexity: medium

<sub>🤖 Generated with GPT-5 Codex</sub>

<sub>Review-Coverage: authored=codex; ran=gemini,cursor-composer,claude; adjudicated=domain; declined=cursor-grok; rounds=1 @ ef1e0d25d4e5</sub>

<sub>Human-Review-Need: 3 (decisions: independent-routing-expectation, pinned-graph-regression) @ ef1e0d25d4e5</sub>

### DIFF
--- a/unitTests/resources/vectorIndex.test.js
+++ b/unitTests/resources/vectorIndex.test.js
@@ -1020,8 +1020,8 @@ describe('HNSW greedy routing above layer 0 (ROUTING_EF)', () => {
 				} finally {
 					customIndex.searchLayer = originalSearchLayer;
 				}
-				// The comparison below cannot catch the optimization being removed: hand the upper
-				// layers the full ef and both of its sides search identically.
+				// Hand the upper layers the full ef and both sides of the greedy-vs-full comparison
+				// below search identically, so only this assertion notices the optimization going away.
 				assert.deepStrictEqual(
 					[...routingEfs],
 					[ROUTING_EF],
--- a/unitTests/resources/vectorIndex.test.js
+++ b/unitTests/resources/vectorIndex.test.js
@@ -935,6 +935,7 @@ describe('HNSW construction ef auto-scale (#2180)', () => {
 describe('HNSW greedy routing above layer 0 (ROUTING_EF)', () => {
 	if (process.env.HARPER_STORAGE_ENGINE === 'lmdb') return;
 	const N = 600;
+	const ROUTING_EF = 1; // must track ROUTING_EF in resources/indexes/HierarchicalNavigableSmallWorld.ts
 	// Each graph's level assignment is pinned with a seeded PRNG (mulberry32) so a run is
 	// reproducible: greedy-vs-full equality is only statistically true over random graphs — ~2-3%
 	// of random 600-node graphs legitimately route to a different entry point and change the
@@ -965,7 +966,9 @@ describe('HNSW greedy routing above layer 0 (ROUTING_EF)', () => {
 			],
 		});
 		let seedState = seed;
+		let draws = 0;
 		T.indices.vector.customIndex.random = () => {
+			draws++;
 			seedState = (seedState + 0x6d2b79f5) | 0;
 			let t = Math.imul(seedState ^ (seedState >>> 15), 1 | seedState);
 			t = (t + Math.imul(t ^ (t >>> 7), 61 | t)) ^ t;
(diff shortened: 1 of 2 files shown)


dawsontoth (approved):

cb1kenobi (approved):

unitTests/resources/vectorIndex.test.js
```diff
@@ -1024,10 +1048,32 @@ describe('HNSW greedy routing above layer 0 (ROUTING_EF)', () => {
...
+				// On many graphs layer 0 alone reaches the true neighbours from wherever it starts, and
+				// there the comparison above passes with the descent deleted outright. That is a
+				// property of the graph, not of this seed list, so the sweep as a whole has to contain
+				// at least one graph the descent decides — otherwise nothing here tests the descent.
+				customIndex.searchLayer = function (v, epId, ep, ef, level, ...rest) {
+					return level > 0 ? [] : originalSearchLayer.call(this, v, epId, ep, ef, level, ...rest);
+				};
+				try {
+					for (let i = 0; i < targets.length; i++) {
+						if ((await topTenIds(T, targets[i])) !== greedy[i]) {
```
cb1kenobi, line 1061: **Reference comparison makes the descent control vacuous** — Severity: P1 (blocking). `topTenIds()` produces a new result array, so `!==` compares object identity and is true even when its IDs match `greedy[i]`; the first query therefore marks every seed as descent-sensitive, allowing the test to remain green when descent has no effect. Compare the arrays by value, for example with `isDeepStrictEqual`, before pushing the label.

—
Reviewed ef1e0d25

kriszyp: Declined — no code change is warranted. topTenIds maps each result to its id and then joins the IDs with a comma, so it returns a primitive string; !== therefore compares values, not array identities. The adjacent strictEqual comparisons use the same return value, the prior independent Claude/Gemini/Cursor/domain review dropped this exact claim for the same reason, and the focused HNSW test passes (1 passing). Resolving as conclusively answered.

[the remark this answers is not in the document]
kriszyp: Confirmed. The code trace here is correct: topTenIds returns a comma-joined string, making the descent-sensitivity comparison a value comparison. The prior independent review reached the same ruling and the focused HNSW test passes (1 passing). No code change is warranted; resolving this duplicate thread as conclusively answered.

PR #2377, opened 2026-08-28
Hot reload the models config block when the config file changes

### LINKED ISSUE #2344: Hot reload the models configuration when the root config changes
## Summary

Changing a `models.embedding.*` / `models.generative.*` entry in the root config requires a full process restart to take effect. Every other consumer of hot-reloadable config picks changes up in place. The two pieces needed to close that gap already exist and simply aren't connected.

## What already works

`RootConfigWatcher` (`config/RootConfigWatcher.ts`) watches the root config file, reparses it, and emits `change` with the new config object:

```js
this.emit('change', (this.#config = config));
```

`bootstrapModels(rootConfig)` (`resources/models/bootstrap.ts`) is exported and already documented as safe to re-run:

```
 * Populate the model registry from `rootConfig.models`. No-op if the block
 * is absent or empty. Idempotent within a process: each entry overwrites any
 * prior registration under the same logical name (registry uses `.set()`).
```

It also calls `clearFallbackGroups()` first, explicitly so that "a removed/changed `fallback:` (or a removed `models:` block) doesn't leave stale routing behind" (#1326) — i.e. it was written with re-loading in mind.

Registration itself is pure. `registerOpenAIBackend` constructs an `OpenAIBackend` and calls `setEmbedding`/`setGenerative`; the constructor stores `apiKey`/`baseUrl` and does no I/O, no connection, and no health check. So re-registering costs nothing and cannot fail on an unreachable endpoint.

There is precedent for the wiring, one line of it, in the logger (`utility/logging/harper_logger.ts:149`):

```js
rootConfig.on('change', updateLogSettings);
```

## Ask

Subscribe to the watcher's `change` event and re-run `bootstrapModels` with the new config, so a changed `models` block applies to the running process. `bootstrapModels` is currently invoked exactly once, from `components/componentLoader.ts:367`, gated on `isRoot` during component load.

Points worth deciding rather than assuming:

- **Removal semantics.** `bootstrapModels` returns early when the block is absent, so a config that *drops* `models` entirely would leave the previous backends registered. `clearFallbackGroups()` handles fallback groups but not the backend maps. If reload is to be truthful, a removed entry should probably deregister — which needs either a clear-then-repopulate, or a diff.
- **Interaction with programmatic registration.** `models.registerBackend(...)` is a public API that components call from their own code. A reload that clears the registry wholesale would drop those too. Distinguishing config-sourced registrations from component-sourced ones may be necessary.
- **In-flight calls.** Replacing a backend under a concurrent `models.embed()` is safe today because `resolveCandidates` resolves once per call and the backend object is immutable after construction — worth confirming that stays true.

## Why it matters

An orchestrator that manages Harper instances can currently only deliver a models-config change by recreating the container, because the config reaches the process through environment. That makes a routine credential rotation — a bearer token on an embedding backend, rotated on a fixed schedule — cost a restart of a live instance, permanently and on a rolling basis. With reload in place the orchestrator can write the config file and the change applies in place, which removes the restart entirely rather than making it safer.

This also benefits anyone self-hosting: rotating an API key for a configured backend, or repointing a `baseUrl`, currently means a restart.

Devin-Holland:
Closes #2344.

The `models:` block of `harperdb-config.yaml` now hot-reloads when the file changes: an orchestrator (or a self-hosted operator) can rotate a credential, re-point a `baseUrl`, or add/remove an entry by rewriting the file, and every worker reprojects its registry with no restart. For Fabric that removes the container recreate — and the GTM drain — that a models change costs today (central-manager#753).

Boot publishes each entry as it is built, in config order — module factories keep their order-dependent composition (a `cached` wrapper still resolves its earlier `base`), exactly as before this PR. A reload stages changed entries off to the side through the same factories (`constructBackend` diverts the factory's install — including its helper registrations — via an AsyncLocalStorage capture scope), then one synchronous publish tick swaps them in conditionally, removes entries absent from the block (helpers cascade; a helper whose name a config entry claimed is suppressed, and restored the moment that entry is removed), and rebuilds fallback routing — so a request observes either the previous projection or the new one, never a partial rebuild. **Reload runs factories for built-ins only**: module-backed entries are restart-managed in both directions (add, change, and remove are refused with a warning), because a module factory may compose across entries and staged construction cannot honor that ordering — review decision with Kris. Applies are serialized; boot and reload coalesce latest-wins in separate lanes (a boot supersedes reloads observed before it — including a pending debounce snapshot, whose invalidation-on-boot guard is mutation-verified through an `onSnapshotObserved` test seam — while a reload arriving after still refines under reload rules), and a reload snapshot is validated against the same Joi `modelsSchema` boot enforces (hoisted out of `configValidator`, `allowUnknown` mirrored) — a snapshot boot would reject keeps the prior projection wholesale. One isolate-shared `RootConfigWatcher` drives it (`getSharedRootConfigWatcher()`), consumed by both logging and models — no second native watcher per worker.

**Removal semantics are deliberate and partial-write-hardened.** `models: {}` (or a shrunken block) removes built-in-backed entries (module-backed ones are restart-managed, above); `models: null` (a bare `models:` key) is rejected exactly as boot's validator rejects it, so a reload can never accept a file the next restart would refuse; a snapshot with **no** `models` key is a no-op, because a non-atomic in-place rewrite can be observed as a valid YAML prefix that hasn't reached the block yet, and adopting that would tear down every backend. The watcher additionally applies only the last event in a 150ms quiet window. The orchestrator contract is still an atomic `tmp`+`rename` write.

## For the human reviewer

1. **The compatibility gate is the design's hinge.** If `HARPER_SET_CONFIG` / `HARPER_CONFIG` / `HARPER_DEFAULT_CONFIG` also names `models`, hot reload stays off — the watcher sees the raw file, env layers compose over it at boot, so the file alone is not authoritative for the block. This is deliberately also the migration story: a host-manager still injecting models through `HARPER_SET_CONFIG` keeps today's restart semantics untouched; one that writes the file instead (the follow-up HM change) gets live reload. Mixed fleets need no version negotiation.

2. **Boot and reload deliberately differ in one respect.** Boot overwrites whatever occupies a slot (the documented `bootstrapModels` contract); a reload swaps only against the instance the config projection itself installed, so a programmatic `registerBackend` that shadows a config entry keeps its documented precedence and a reload can never clobber it. Removal is equally guarded (`removeIfCurrent`).

3. **Change detection is per-entry raw-JSON equality.** An unrelated config edit (logging level, ports) re-runs no factories; a key-reorder inside an entry would spuriously rebuild it — harmless for built-ins (pure constructors), and module factories are only re-invoked when their own entry text changes.

4. **Carried, not fixed — two ledger items for your judgment.** (a) Helper (`extras`) ownership is publish-order and history dependent — inherent to conditional-swap semantics; the concrete failure modes found in review are fixed and tested, the abstract property stands. (b) `bootstrapModels` no longer rejects at boot on a wholesale failure (it logs at error and continues) — this matches the file's documented per-entry design, and the only reachable systemic throw is a non-YAML poisoned block, but it is a behavior change from `main`.

5. **Review history, for the record.** Seven implementation review rounds (codex/gemini/harper-domain via the pre-push CLI, plus two standalone Cursor-Grok legs after the CLI's per-branch Cursor cap was spent on the abandoned design). They produced 5 majors, ~20 minors, and one Grok blocker (partial-write adoption), all fixed with mutation-verified tests or declined with a stated contract rationale.

6. **Planning history, for the record.** Eight planning-review rounds on earlier designs (secrets-store credential rotation) never cleared the framing gate; the owner then clarified the actual goal — replace `HARPER_SET_CONFIG` injection with file writes that hot-reload — and that design was abandoned. What survived from it are the registry primitives (`constructBackend` / `replaceIfCurrent` / `removeIfCurrent`), which are exactly the staged-apply mechanics this needs.

7. **Docs**: user-facing behavior (the models block hot-reloads; env-layer pinning). Companion documentation PR to follow together with the host-manager change, since the operator guidance (write via tmp+rename, move models out of `HARPER_SET_CONFIG`) is about the orchestrator side.

## Verification

- **2026-09-09 — `claude[bot]`'s review of the merge head surfaced a real bookkeeping leak** (the "suppress-before-confirm" nit the pre-push rounds had carried, made concrete): `claimHelperOccupant` suppresses a factory helper before the claiming entry's own swap is known to win, and the removal path restored the helper only `if (slot.backend)` — so an entry that lost its swap to an application override left the helper suppressed with no way back once the override retired. Fixed in `03682700f` by restoring unconditionally (a no-op while the name is held or nothing is suppressed), pinned by a test that plays the exact sequence; reverting the gate fails exactly that test. Still open by design (decision ledger, `helper-suppression-bookkeeping`): if an override holds the name at the moment the claimant is removed, the restore's compare-and-set loses and the record stays suppressed — the override keeps serving, and nothing re-runs the restore when it retires.
- **2026-09-09 — merged `origin/main` (120 commits, `1a5067e1b`).** One conflict, `validation/configValidator.ts`: this branch hoists the models schema to module scope so `validateModelsBlock` can reuse it, and #2484 appended its `sql` schemas right after the old in-function copy. Resolved by keeping main's `sql` block inside `configValidator()` and the hoisted models schema at module scope; the delta against main is exactly the hoist plus `validateModelsBlock`. `tsc` build clean; `configValidator` + `setConfigurationSql` (100) and the four models suites (137) pass at the merge head; full unit gates at the merge head, Node 24.19 on macOS: `test:unit:resources` 2293 passing / 30 pending / 0 failing; `test:unit:main` (run with `--timeout 30000`, because two watcher/process files otherwise hang the gate under mocharc's `timeout: 0`) 5432 passing / 207 pending / 20 failing — every one of the 20 is a macOS platform class, not this branch: tmpdir realpath mismatches (`/private/var` vs `/var`) in the module-tracker and preload tests, the `applicationSpawn` process-tree wait, chokidar/FSEvents timeouts including `RootConfigWatcher`'s atomic-rename case (which fails identically on pristine `origin/main` here, 2/2, and passes on CI's Linux and Windows unit jobs at the pre-merge head), a native-addon load, and one EPIPE. Pristine `origin/main` under the same command: 5421 passing / 207 pending / 31 failing, covering 19 of the branch's 20 (main additionally fails the `testJSWithDeps` isolation group and a `configValidator` rootPath-warning case). The one branch-only name, `test rsa_keys is defined` in `unitTests/security/tokenAuthentication.test.js` (file unchanged from main), passes 3/3 when the file runs alone, so it is an order-dependent flake of the full run rather than a regression — main's run instead failed a different keys-loading case in `keys.test.js`. The macOS-only hangs are filed as harper#2538. (Local note: a worktree needs `@harperfast/rocksdb-js-darwin-arm64@2.8.0` installed explicitly and Node ≥ 24 for mocha, or the suite crashes at load in `databases.ts` — the binding otherwise resolves to an older binary in the main checkout.)

**Live, on a running instance** — `harper dev` 5.2.5 from this branch, `threads_count=2`, isolated scratch HOME/rootPath; a component with the documented shape (`import { models } from 'harper'`, a `Resource.get()` calling `models.embed`) and a stub provider recording every `Authorization` header Harper sends:

- **Add** the `models:` block to the live config file → `/Embed/` went 500 `ModelBackendNotFoundError` → 200 in ~1s, stub saw the configured key
- **Rotate** `apiKey` (plain write) → requests carry the new bearer in ~1s
- **Rotate** again via tmp+`rename` (how an orchestrator should write) → ~1s
- 12 consecutive requests post-rotation: **12/12 on the new credential** — both worker threads reprojected
- **Remove** the block → 500 `ModelBackendNotFoundError` in ~1s
- One pid throughout — the process never restarted

That live run predates the final review rounds; behaviors added after it (reload validation, `models: null` semantics, whole-key-absence no-op, the 150ms settle window, helper lifecycle) are covered by the unit suite below, and the file-watch path itself is exercised end-to-end by a real-chokidar test. Happy to re-run the live smoke on request before merge.

**Unit** — `unitTests/resources/models/modelsConfigReload.test.js` (28 tests): rewritten credential reaches a real `embed()` through `resolveEmbedding` (fetch-stubbed); unchanged entries keep their instance; removed entries stop serving while a present-but-malformed entry retains its last applied backend; application overrides survive reload and removal; a failed rebuild keeps the previous backend *and* its fallback routing; fallback routing updates with the block; rapid applies coalesce latest-wins (asserted with no settling re-apply, so a dropped queued update fails it); a poisoned block neither rejects into the watcher's fire-and-forget call nor blocks the next apply; the env-pin gate stays off under `HARPER_SET_CONFIG`; and a real-file chokidar end-to-end (temp config, `getConfigFilePath` patched without sinon, the shared `waitFor` helper). `unitTests/resources/models/backendRegistry.test.js` covers the capture scope (concurrent constructions stay separate; an unrelated registration mid-capture installs normally; capture survives a throwing factory) and both conditional primitives.

**Mutation checks (against `dist`, which the tests actually load)** — unconditional swap, disabled removal pass, disabled change-diff, disabled env gate, malformed-treated-as-absent, drain-catch removal, and dropped-queued-update each fail their targeted tests; baseline restored green.

**Boot parity** — the full models suite (431 tests, including the pre-existing `bootstrap.test.js` boot contract: overwrite semantics, fallback clear-on-rebootstrap, module specifier resolution) passes unmodified.

**Suites that could not run locally**: `test:unit:components` / `test:unit:config` fail at load in this checkout on pre-existing ESM-only dependency issues (`chai`, `@harperfast/skills`) — verified pre-existing by running untouched files; `package-lock.json` is unmodified. CI runs them.

Complexity: moderate

<sub>Review-Coverage: authored=claude; ran=gemini,codex; declined=cursor-grok,cursor-composer,domain; rounds=22 @ 03682700f1a8</sub>

<sub>Human-Review-Need: 3 @ 03682700f1a8</sub>


resources/models/bootstrap.ts
```diff
@@ -79,36 +93,111 @@ const FACTORIES: Record<string, BackendRegisterFn> = {
...
+export function applyModelsConfig(models: ModelsConfig | undefined): Promise<void> {
+	return queueModelsApply(models, false);
+}
+
+// Interleaved applies could install an older credential last; serialized + coalesced, they cannot.
+let pendingApply: { block: ModelsConfig | undefined; isBoot: boolean } | undefined;
+let applyChain: Promise<void> | undefined;
+
+function queueModelsApply(block: ModelsConfig | undefined, isBoot: boolean): Promise<void> {
+	pendingApply = { block, isBoot };
```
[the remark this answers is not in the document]
Devin-Holland: Fixed in 358fc8912 — and the fix went one level deeper than the suggestion. A sticky flag turned out to trade this bug for a worse one (the coalesced watcher block would run under boot semantics, bypassing reload validation and the missing-key partial-write guard), so boot and reload now coalesce in **separate lanes**: each latest-wins on its own, a pending boot drains first with its own block, a boot supersedes reloads queued before it (and invalidates a pending settle timer, which is pre-boot content), and a reload queued after still refines under reload rules. Deterministic race tests cover reload-erases-boot, laundering, and stale-reload-after-boot — each mutation-verified. Thanks for confirming this one; it was flagged in the review-need footer and deserved the promotion to a fix.

— Claude (devain)

resources/models/bootstrap.ts
```diff
@@ -132,19 +281,196 @@ async function registerKind(kind: ModelKind, entries: Record<string, ModelEntry>
...
+/** Stop watching the config file. The current projection keeps serving. */
+export function stopModelsConfigHotReload(): void {
+	clearTimeout(pendingSettle);
+	pendingSettle = undefined;
+	modelsConfigWatcher?.close();
+	modelsConfigWatcher = undefined;
+}
+
+function envLayerNamingModels(): string | undefined {
+	for (const name of ['HARPER_SET_CONFIG', 'HARPER_CONFIG', 'HARPER_DEFAULT_CONFIG']) {
```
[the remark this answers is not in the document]
Devin-Holland: Fixed in feef22877 — the gate test is now parametrized over all three `ENV_LAYERS`, each asserted independently. Mutation-verified: misspelling the two previously-untested names fails exactly those two tests. Good catch — the save/restore harness covering all three while only one had an assertion is exactly the kind of gap that reads as coverage.

— Claude (devain)

resources/models/bootstrap.ts
```diff
@@ -79,36 +93,129 @@ const FACTORIES: Record<string, BackendRegisterFn> = {
...
+	if (isBoot) {
+		pendingBoot = { block: block ?? undefined };
+		// A boot supersedes any reload queued before it — draining that older snapshot after the
+		// newer boot would refine it backwards. A reload queued after the boot still refines. The
+		// same applies one layer down: a watcher snapshot OBSERVED before this boot whose settle
+		// timer has not fired yet is also pre-boot content — boot read the same file, so dropping
+		// the timer loses nothing newer.
+		pendingReload = undefined;
+		clearTimeout(pendingSettle);
+		pendingSettle = undefined;
```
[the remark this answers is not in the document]
Devin-Holland: Done in c9e4c7896 — you were right that this deserved better than a disclosure. The missing piece was knowing WHEN the watcher observed the snapshot (my earlier sentinel-watcher attempt hung the suite — two chokidar instances on one file), so the fix is a 2-line `onSnapshotObserved` test seam: the test arms the settle timer deterministically, boots before it fires, and asserts the pre-boot content is discarded. Mutation-verified: removing the clearTimeout guard fails exactly this test. The disclosure in the PR body is superseded.

— Claude (devain)

resources/models/bootstrap.ts
```diff
@@ -132,19 +281,202 @@ async function registerKind(kind: ModelKind, entries: Record<string, ModelEntry>
 			// (env var unset) pass through unchanged — backend's required-field
 			// validation catches them with a meaningful error.
 			const config = expandEnvVarsDeep(entry);
-			const builtin = FACTORIES[entry.backend];
-			if (builtin) {
-				await builtin({ logicalName, kind, config });
-			} else {
-				// Not a built-in name: treat `backend` as a module specifier (#1471).
-				await registerFromModule(kind, logicalName, entry.backend, config);
+			const { backend, extras } = await constructBackend(kind, logicalName, async () => {
```
kriszyp, line 381: Staging every changed entry before publishing any changes the existing module-factory semantics. Previously entries were installed in config order, so a custom factory for `cached` could resolve and wrap an earlier `base`. Here all factories run under `constructBackend` first: on fresh boot the later factory sees no `base`; on reload it sees the old instance and can retain its revoked credential after `base` is swapped. Please preserve sequential boot semantics for module factories, provide an isolated draft registry/return-valued construction API, or restrict hot reload to built-ins whose factories are known to be independent.

resources/models/bootstrap.ts
```diff
@@ -79,36 +93,129 @@ const FACTORIES: Record<string, BackendRegisterFn> = {
...
+}
+
+const installedSlots = new Map<string, InstalledSlot>();
+
+/** Release a name held by a projection-installed helper, so a config entry can claim it. */
+function claimHelperOccupant(kind: ModelKind, logicalName: string): ModelBackend | undefined {
+	for (const slot of installedSlots.values()) {
+		const index = slot.extras?.findIndex((e) => e.kind === kind && e.logicalName === logicalName) ?? -1;
+		if (index >= 0) {
+			const [extra] = slot.extras!.splice(index, 1);
```
kriszyp: Splicing this helper permanently removes it from its parent’s ownership record. Sequence: a parent factory registers helper `h`; an explicit config entry named `h` claims it; a later reload removes that explicit entry while leaving the parent unchanged. The parent is skipped as unchanged, removal deletes `h`, and nothing restores the parent’s helper. The live registry then differs from a restart with the same final config. Please retain suppressed helper candidates and restore the next owner when the winner is removed, or immediately rebuild the affected parent.

components/componentLoader.ts
```diff
@@ -630,7 +630,10 @@ export async function loadComponent(
 		// methods. Per-entry errors are logged and skipped by `bootstrapModels`.
 		// Awaited so module-backed entries (#1471) finish importing before the
 		// per-component iteration below; built-in entries register synchronously.
-		if (isRoot) await bootstrapModels(config);
+		if (isRoot) {
+			await bootstrapModels(config);
+			startModelsConfigHotReload();
```
kriszyp, line 794: This starts another `RootConfigWatcher` for every normal root load/worker, while logging already creates one at `utility/logging/harper_logger.ts:148`. Deployments without a `models` block therefore still double their per-worker native watcher instances and duplicate synchronous file reads/YAML parses on each config write. At high worker counts this also consumes the finite inotify/FD budget and can force polling fallback. Please share one root-config watcher per isolate between logging and models, rather than opening a second watcher in every worker.

kriszyp (commented):
This looks good, and I approve of moving forward with this. However, I will note that this PR inherently creates an interesting situation: Currently it is targeting 5.3, and I think that is the right target for a PR of this size that is adding functionality. But that means you will still need to employ (thread) restarts on 5.2 servers to adopt backend config changes.

resources/models/bootstrap.ts
Devin-Holland: Fixed in 144a1e524, taking two of your offered options together. Boot now constructs and publishes each entry sequentially in config order — exact prior semantics, so a `cached` factory resolves its `base` (covered by a new fixture test that records whether the base was visible at factory time). Reload keeps the staged atomic publish but runs factories for **built-ins only**: module-backed entries are restart-managed in both directions — add, change, AND remove are refused with a warn (removal too, or a rename would half-apply as a bare removal). Each behavior is mutation-verified. The practical loss is small: the hot-reload driver (Fabric credential rotation) is openai, and module factories keep exactly the semantics they had before this PR.

— Claude (devain)

resources/models/bootstrap.ts
Devin-Holland: Fixed in 144a1e524 with your first option: the claim now **suppresses** the parent's helper record instead of splicing it, and the removal pass restores the suppressed helper the moment the claiming entry is removed — `replaceIfCurrent(expected: undefined)` into the vacated name, so an app override that grabbed the name in between still wins. Test asserts the restored instance is the factory helper, not the removed entry, i.e. the live registry matches a restart with the same final config. Mutation-verified (dropping the restore fails the test).

— Claude (devain)

components/componentLoader.ts
Devin-Holland: Fixed in 144a1e524 — you were right to push on this after I'd twice deferred it as out of scope. `config/RootConfigWatcher.ts` now exports `getSharedRootConfigWatcher()`, `harper_logger` rides it instead of constructing its own, and models subscribes to the same instance — so a worker has one native watcher for the root config regardless of consumers. `stopModelsConfigHotReload` removes its listeners without closing the shared instance; only a test-supplied private path gets a module-owned watcher that stop() closes. Tested via listener accounting on the shared instance, mutation-verified against a private-always regression.

— Claude (devain)

Devin-Holland:
Re the 5.2 note in Kris's review: agreed, and the host-manager follow-up is scoped exactly that way — HM only switches from HARPER_SET_CONFIG injection to file writes for instances whose core version advertises this capability (5.3+); 5.2 instances keep today's compose-recreate path. Mixed fleets need no negotiation because the env-layer gate in this PR keeps hot reload off wherever HM still injects models through HARPER_SET_CONFIG.

— Claude (devain)

resources/models/bootstrap.ts
```diff
@@ -132,19 +375,196 @@ async function registerKind(kind: ModelKind, entries: Record<string, ModelEntry>
...
+
+	for (const { desiredEntry, backend, extras } of staged.values()) {
+		publishEntry(desiredEntry, backend, extras, isBoot);
+	}
+	const desiredKeys = new Set(desired.map((d) => slotKey(d.kind, d.logicalName)));
+	for (const [key, slot] of [...installedSlots]) {
+		if (presentKeys.has(key)) continue;
+		// Module-backed entries are restart-managed on reload in BOTH directions: refusing an added
+		// rename target while removing its old name would turn the rename into a bare removal.
+		if (!isBoot && slot.builtin === false) {
```
kriszyp, line 421: Keeping the module-backed slot here does not preserve its full projection. `clearFallbackGroups()` runs below, and the routing rebuild skips this retained-but-absent key because it is neither in `desiredKeys` nor `presentKeys`. Removing or renaming a module-backed primary configured with `fallback: ['backup']` therefore leaves the primary serving but silently disables failover—the half-apply this restart-only path is intended to prevent. Please track retained module keys and restore their recorded fallback groups after the clear; the removal/rename tests should also assert the router still returns the previous candidates.

kriszyp (approved):
Looks like one fix, but not a blocker. Sounds like we are in agreement on target.

resources/models/bootstrap.ts
Devin-Holland: Fixed in 40052f3ba with your first option. A module-backed slot removed/renamed out of config is already retained here (restart-managed removal), but `clearFallbackGroups()` below wiped its routing and neither rebuild loop could see it again — it's absent from both `desiredKeys` and `presentKeys`. Now the removal pass tracks retained module keys and restores their recorded `fallback` group after the clear, so the router still returns the previous candidates. Mutation-verified: without the restore the test drops from 2 candidates to 1.

While here, the pre-push review surfaced the symmetric hole on *change*: the restart-only guard checked only the incoming backend, so a module→built-in rewrite live-replaced the module and retired its helpers. Extended the guard to also refuse when the currently-installed slot is module-backed (40052f3ba), matching the "restart to add, change, OR remove" invariant. Both cases have mutation-verified tests.

— Claude (devain)

resources/models/bootstrap.ts
```diff
@@ -132,19 +378,209 @@ async function registerKind(kind: ModelKind, entries: Record<string, ModelEntry>
...
+	const applyFromFile = (config: unknown) => {
+		const models = (config as RootConfig | undefined)?.models;
+		options?.onSnapshotObserved?.();
+		clearTimeout(pendingSettle);
+		pendingSettle = setTimeout(() => {
+			void queueModelsApply(models, false);
+		}, debounceMs);
+		pendingSettle.unref?.();
+	};
+	modelsConfigWatcher.on('ready', applyFromFile);
```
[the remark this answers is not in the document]
Devin-Holland: Fixed in 1707912bc — took your second option, made deterministic: a snapshot the watcher already holds is applied synchronously right after subscribing, so the shared singleton's pre-fired one-time 'ready' can no longer hide a rewrite that landed before models subscribed. Tested by injecting a caller-owned pre-warmed watcher (its ready consumed before subscription) and asserting the snapshot applies with no further write; mutation-verified by dropping the sync check.

— Claude (devain)

resources/models/bootstrap.ts
```diff
@@ -132,19 +378,216 @@ async function registerKind(kind: ModelKind, entries: Record<string, ModelEntry>
...
+			continue;
+		}
+		if (slot.backend) removeIfCurrent(slot.kind, slot.logicalName, slot.backend);
+		for (const extra of slot.extras ?? []) {
+			if (!extra.suppressed) removeIfCurrent(extra.kind, extra.logicalName, extra.backend);
+		}
+		installedSlots.delete(key);
+		// A removed entry may have been shadowing a factory helper of the same name; put the
+		// helper back, so the live registry matches a restart with this final config.
+		if (slot.backend) restoreSuppressedHelper(slot.kind, slot.logicalName);
```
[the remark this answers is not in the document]
Devin-Holland: Taken — fixed in 03682700f. The restore now runs unconditionally on removal (`restoreSuppressedHelper` is already a no-op while the name is held or nothing matching is suppressed), and a test plays your exact sequence: a factory helper displaced by an application override, a config entry that claims the name and loses its swap, the override retiring, then the entry being removed — the helper comes back. Reverting the gate fails exactly that test.

The remaining corner you also named — the override still holding the name at the moment the claimant is removed — stays as documented in the decision ledger (`helper-suppression-bookkeeping`): the compare-and-set loses, the override keeps serving, and nothing re-runs the restore when it retires.

<sub>— devain (Claude Fable 5.1)</sub>

kriszyp (approved):
Looks good

kriszyp (approved):
Looks good

PR #2378, opened 2026-08-28
Unpin Node 26 in integration CI now that both regressions behind the pin are fixed

### LINKED ISSUE #2025: Node v26.5.1 (llhttp 9.4.3 + undici 8.9.0) breaks integration tests across harper and harper-pro on Node 26

kriszyp:
The integration-test matrix [floats to the latest Node 26.x again](https://github.com/HarperFast/harper/pull/2378/changes?w=1#diff-4761c5069b46ce96f49d74476da61f43ced3abc73d22254cd95a2be0e250a593R51). The 26.5.0 pin covered two defects and both are fixed: undici 8.9.0's idle-socket `fetch()` stall (#2025) is fixed upstream in undici 8.10.0, bundled into Node from 26.8.0; and the mTLS revocation fail-open that Node 26.8.x exposed (#2380) is fixed in Harper by #2457. With the nightly and push-to-main matrix covering latest 26.x directly, the standalone [`node26-canary.yml`](https://github.com/HarperFast/harper/pull/2378/changes?w=1#diff-5ccfc8438111a0d32c4e2901f6eccd0768070e8e3b072323c627f2585570457fL1) has nothing left to watch and is deleted. The two integration suites whose headers still called the undici fix unreleased now [say where it shipped and why they keep `node:http`](https://github.com/HarperFast/harper/pull/2378/changes?w=1#diff-8c218845954e7d752dfa0527eb5fd9e252b715e6f9b44bcc7ecd5a7b7aac4794R34) (the engine range still admits 26.5.1 to 26.7.x). Fixes #2025.

This replaces the PR's earlier shape, which kept the pin for #2380 and re-targeted the canary at it; #2457 removed that reason.

## For the human reviewer

1. **Delete the canary rather than keep it as a regression watch.** The canary existed because the pinned matrix no longer exercised any 26.x newer than the pin; unpinned, the nightly full matrix runs every integration shard on the latest 26.x, which is a superset of the canary's two suites, and its inverted green/red convention was a standing footgun for nightly triage. Cost of "no": keeping ~130 lines of inverted-conclusion workflow that would now be red every night for the wrong reason.
2. **Evidence standard for unpinning: fixes in-tree plus one green floating-26 dispatch, not a nightly soak.** Both defects behind the pin were timing-sensitive, so one green run does not by itself distinguish "fixed" from "did not reproduce this time"; what does is that each has a root-cause fix with its own regression test (#2457's resumed-session OCSP/CRL tests; the TTL-reset and cache-header suites on `node:http`). Reversible in one line if the nightly 26 leg goes red, at the cost of nightly-gate triage noise until then.
3. **The two timing suites keep `node:http` instead of returning to `fetch()`.** The engine range (`^22.18.0 || >=24.0.0`) still admits 26.5.1 to 26.7.x, where undici 8.9.0 stalls `fetch()` on an idle loop, so those suites would flake on any such runtime. Cost: the undici request path is not exercised by those two suites until the engine floor moves past 26.7.
4. **A broken 26.x patch release now fails `build` and skips every dependent job.** The pin shielded CI from that; floating `26` shares the exposure `22` and `24` always had (raised by the Claude review bot, non-blocking). Alternative: gate downstream jobs on per-version artifact presence instead of the whole `build` job. Left as a follow-up because it touches every dependent job; [`check-latest` on the 26 leg](https://github.com/HarperFast/harper/pull/2378/changes?w=1#diff-4761c5069b46ce96f49d74476da61f43ced3abc73d22254cd95a2be0e250a593R79) keeps the leg on the newest release rather than a cached one either way.
5. **Milestone v5.3, no backport.** The `v5.2` branch carries the same pin and canary, but unpinning it would put its Node 26 leg on 26.8.x without #2457 (v5.3 only), which is exactly the failing combination. Backport this only together with #2457.

## Verification

- `integration-tests.yml` dispatched on this branch with `node-version=26` (which floats to the latest 26.x): [run 33585794061](https://github.com/HarperFast/harper/actions/runs/33585794061) resolved Node 26.8.1 and passed Build plus all six Linux `Node.js v26` shards and all six Windows shards. Its uWS and Bun shards failed only because they download the `node-24` build artifact, which a 26-only dispatch never builds (`Artifact not found for name: harper-build-artifacts-node-24`); those shards are green on `main`'s push runs and are unaffected by this change.
- PR CI runs the Node 24 shard only (by design for pull requests); the floating 26 entry is exercised on push to `main` and nightly.
- Locally, the OCSP/CRL suites that failed on 26.8.1 before #2457 pass 16/16 on `origin/main` at 6a42ded92 with Node 26.8.1.

Complexity: easy

<sub>Review-Coverage: authored=claude; ran=codex; blocked=gemini(auth); declined=cursor-grok,cursor-composer,domain; rounds=2 @ b004c3ec1f11</sub>

<sub>Human-Review-Need: 3 @ b004c3ec1f11</sub>


kriszyp:
Reviewers by measured affinity: @heskew (top blame/recency on `.github/workflows/integration-tests.yml`) and @cb1kenobi, who [identified the upstream revert this PR acts on](https://github.com/HarperFast/harper/issues/2025#issuecomment-5430681484) — the undici 8.10.0 change that made the pin removable.

Note for whoever lands this: [Restore floating Node 26 integration coverage #2342](https://github.com/HarperFast/harper/pull/2342) covers the same ground and is still open. Its blocker (`actions/node-versions` resolving only 26.7.0) has cleared — 26.8.1 is in the manifest now — but its freshness floor is set to **26.8.0**, and that release cannot be used: its published artifact self-reports `26.8.0-alpha.0.0.0` and npm's `devEngines` check rejects it outright (`EBADDEVENGINES`, reproduced locally). This PR sets the floor to 26.8.1 for that reason. One of the two should be closed rather than both landing.

🤖 Claude Opus 5 (dispatch agent), via [Claude Code](https://claude.com/claude-code).

kriszyp:
**Direction changed since this PR opened — it no longer removes the pin.**

Dispatching this branch's own workflow on floating `26` ([run 33190835468](https://github.com/HarperFast/harper/actions/runs/33190835468)) resolved Node 26.8.1 and failed integration shards 1/6 and 6/6: a revoked client certificate is accepted, because Harper's mTLS revocation check cannot see the issuer certificate and its `no-issuer-cert` branch returns `valid: true` regardless of `failureMode: 'fail-closed'`. Filed as [#2380](https://github.com/HarperFast/harper/issues/2380) (P1, epic #1674).

Reproduced locally, deterministic: the OCSP and CRL suites pass 12/12 on 26.5.0, 26.5.1 and 26.7.0, and fail the same four assertions on 26.8.1 across two runs. OpenSSL is 3.5.7 on both sides of that flip.

So 26.5.0 is the only Node 26 release clean on both defects, and the PR now keeps the pin, re-documents why, and re-targets the canary at #2380 instead of deleting it. The evidence for harper#2025 itself being fixed is unchanged and is in the description.

@cb1kenobi — your call on undici 8.10.0 was right; it just isn't sufficient any more.

The overlap note with [#2342](https://github.com/HarperFast/harper/pull/2342) still stands, but inverted: neither PR should unpin right now. #2342 in its current form restores floating `26` with a 26.8.0 floor, which would land the revocation failures on `main`.

🤖 Claude Opus 5 (dispatch agent), via [Claude Code](https://claude.com/claude-code).

.github/workflows/node26-canary.yml
```diff
@@ -86,47 +69,55 @@ jobs:
...
+          # marks tests cancelled by a throwing `before` hook with the same ✖ <name> lines as a real
+          # assertion failure, so a harness failure would otherwise read as "the defect reproduced"
+          # and keep the pin alive on a false green. `cancelled 0` is what separates the two, and
+          # requiring an explicit ✔ on both, against an otherwise-clean run, keeps a renamed test or
+          # an unrelated regression from being reported as "the pin is removable".
+          cancelled="$(sed -n 's/^.*ℹ cancelled \([0-9][0-9]*\).*$/\1/p' "$output" | tail -1)"
+          failures="$(sed -n 's/^.*ℹ fail \([0-9][0-9]*\).*$/\1/p' "$output" | tail -1)"
+          reproduced=false; grep -Fq '✖ should reject revoked certificate with OCSP check' "$output" \
+            && grep -Fq '✖ should reject revoked certificate with CRL check' "$output" && reproduced=true
+          cleared=false; grep -Fq '✔ should reject revoked certificate with OCSP check' "$output" \
```
cb1kenobi: **[P1] Canary treats unrelated failures as proof of #2380** — Severity: blocking. This sets `reproduced=true` from the two test names alone, so after #2380 is fixed, a shared transport, server, or assertion failure in both test bodies still produces two `✖` markers with `cancelled 0` and makes the workflow green, silently preserving the pin and hiding the new regression. Require each failure to match the specific #2380 diagnostic (revoked certificate accepted, expected 401 but received 404), or have the tests emit a dedicated machine-readable reproduction signal.

—
_Generated by Barber AI_

kriszyp: Good catch, and it is now moot in the best way: #2457 fixed #2380, so this PR was rewritten to remove the 26.5.0 pin and delete node26-canary.yml entirely (the floating nightly matrix covers latest 26.x directly). The classifier this thread is about no longer exists. Thanks for the careful read of the inversion logic.

— Claude Fable 5.1

.github/workflows/integration-tests.yml
```diff
@@ -48,19 +54,8 @@ jobs:
...
-            # TEMPORARY: pinned to 26.5.0 instead of floating "26" — 26.5.1's llhttp/undici bump
-            # (deps: llhttp 9.4.2->9.4.3, undici 8.7.0->8.9.0) exposed a genuine Harper-side defect
-            # (TTL-reset-on-write silently loses to the eviction sweep on the operations-API/SQL/
-            # custom-resource write surfaces; confirmed NOT an undici-client artifact via a curl-based
-            # discriminator, and NOT a request-latency race via direct commit timing) plus a
-            # pre-existing per-thread config-reload race (#1747) made more visible by the same bump.
-            # Remove this pin once the underlying race is fixed — see harper#2025, which stays open
-            # until then. Do not extend this pin to unit-test.yml or other workflows without new evidence.
-            echo "node-versions=[22, 24, \"26.5.0\"]" >> $GITHUB_OUTPUT
+            echo "node-versions=[22, 24, 26]" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
kriszyp: Agreed that a broken 26.x patch release would now fail `build` and skip every dependent job, where the pin could not. Still out of scope for this PR (it is the same exposure 22 and 24 have carried all along, and the gating change touches every downstream job), but it is recorded as an open decision in the PR description so it is not lost. Happy to do the per-version artifact gating as a follow-up if we want floating 26 to fail soft.

— Claude Fable 5.1

cb1kenobi:
Reviewed b004c3ec and found no blocking issues. The workflow safely restores floating Node 26 coverage and removes the now-redundant canary. No confirmed blocking defect was found on changed lines.

—
_Generated by Barber AI_

kriszyp:
**[Restore floating Node 26 integration coverage #2342](https://github.com/HarperFast/harper/pull/2342) is now closed in this PR's favor** — @kriszyp's call on the overlap the two PRs had. This one is the surviving change; no rebase or content merge is needed, since #2342's only unique content was the deletion of `node26-canary.yml`, which this PR deliberately keeps and re-targets at [#2380](https://github.com/HarperFast/harper/issues/2380).

One corroborating data point from #2342 worth having here: dispatching **that** branch on floating `26` ([run 33560176415](https://github.com/HarperFast/harper/actions/runs/33560176415)) resolved Node 26.8.1, passed the freshness guard and the build, then failed integration shards 2 and 3 on the revocation assertions (`Expected 401 for revoked cert, got: 404` — CRL on shard 2, OCSP on shard 3). That is an independent reproduction, on a different branch, of the #2380 behavior this PR's evidence table documents — so the case for holding the pin does not rest on a single run.

🤖 Claude Opus 5 (dispatch agent), via [Claude Code](https://claude.com/claude-code).

Issue #2382, opened 2026-08-28
A cert renewal whose table write misses the rebuild window downgrades TLS to the self-signed cert and never retries


DavidCockerill:
## Summary

A certificate renewal can leave a node serving its **self-signed internal certificate** for every SNI — including all customer hostnames — indefinitely, with a healthy process and a perfectly matching key/cert pair on disk. Observed in production on harper-pro 5.1.22 (two-node cluster, both nodes independently, at their respective LE renewals; healed only by manual file touches + restarts days later).

The trigger is a lost race between the private-key hot-reload rebuild and the cert-table write, but the severity comes from two `updateTLS` behaviors that are **still on `main`**: a failed rebuild pass downgrades serving *below* last-good, and nothing ever retries it.

## Observed sequence (5.1.22, identical on both nodes)

```
04:51:49.489 [main/0] [warn]  [tls]: Reloading private key: .../keys/privkey.pem
04:51:49.498 [main/0] [warn]  [tls]: Reloading certificate: .../keys/fullchain.pem
04:51:51.017 [main/0] [error] [tls]: Error applying TLS for <node-fqdn>
             Error: error:05800074:x509 certificate routines::key values mismatch
    at tls.createSecureContext (.../dist/core/security/keys.js:775:21)
    at updateTLS (.../keys.js:878:53)
    at Timeout._onTimeout (.../keys.js:935:21)      <- the 1500ms rebuild debounce
(error repeated once = second selector, and on worker threads; then 4+ days of silence)
```

From then on the node answered every TLS handshake with the self-signed cert (`localhost` + internal node name SANs). The on-disk pair in `keys/` matched itself the whole time (`openssl x509 -pubkey` == `openssl pkey -pubout`) — the renewal install was fine.

Both host-manager install variants reproduce it (the legacy `cp -p` + 1s sleep, and the current staged-copy + back-to-back renames), so this is not an installer-timing regression: the race is against a table write that can lag far more than a second (see below).

## Mechanism

1. The key watcher (`loadAndWatch` → `handlePrivateKeyReload`) synchronously updates `privateKeys` and arms the shared `TLS_REBUILD_DEBOUNCE_MS` (1500ms) rebuild.
2. The cert watcher callback ends in an **un-awaited** `certificateTable.put(...)` (`security/keys.ts` ~274). If that write is not durable-and-delivered within 1500ms, step 3 loses.
[...]

PR #2384, opened 2026-08-28
Publish TLS state transactionally so a failed rebuild never downgrades below last-good

DavidCockerill:
A cert renewal whose `hdb_certificate` write misses the 1500ms rebuild debounce made `updateTLS` pair the old table cert with the new on-disk key, throw `ERR_OSSL_X509_KEY_VALUES_MISMATCH`, and — because the pass had already cleared the shared SNI map and treated a per-record failure as log-and-skip — drop the record's hostnames to the self-signed default. Production nodes served the wrong certificate for days, silently, with nothing retrying (#2382, live incident on harper-pro 5.1.22, both HM install variants).

`updateTLS` now builds the entire replacement state into [pass-local candidates](https://github.com/HarperFast/harper/pull/2384/changes?w=1#diff-ea5297f5d9097656312d08f0a1534f0426fb6ef893a2f5bf3a456b18f81a6841R1154) and reconciles the live maps in place only after the pass completes, so a failure of any shape — one record or the whole pass — leaves the currently-served state intact. A record still in the table whose build failed [retains its live hostname entries and its default candidacy](https://github.com/HarperFast/harper/pull/2384/changes?w=1#diff-ea5297f5d9097656312d08f0a1534f0426fb6ef893a2f5bf3a456b18f81a6841R1308) (a SAN-less record can be serving as default with no hostname entries); a failed pass arms a [self-retry with per-signature backoff](https://github.com/HarperFast/harper/pull/2384/changes?w=1#diff-ea5297f5d9097656312d08f0a1534f0426fb6ef893a2f5bf3a456b18f81a6841R1062) (1.5s doubling to 5min) and signature-throttled logging; and [`loadAndWatch`'s mtime latch now means "last successfully applied"](https://github.com/HarperFast/harper/pull/2384/changes?w=1#diff-ea5297f5d9097656312d08f0a1534f0426fb6ef893a2f5bf3a456b18f81a6841R358) — it rolls back on a synchronous throw or rejected callback promise so the periodic poll can heal a lost table write instead of deduplicating it forever.

**Compatibility note for release notes:** an invalid replacement certificate now means "continue serving the previous credential"; explicit record deletion remains the revocation operation. Retention is trust-aware: on an mTLS listener whose CA set changed since the retained context was built, the retained pair is rebuilt against the current trust material, so revoked client-CA trust is never preserved by a stale context.

## For the human reviewer

1. **Retain-vs-fail-closed** — a failed rebuild retains the previous cert indefinitely instead of falling back to the self-signed default. This is the semantic core of the PR and trades availability for the case where an operator is rotating a cert *because* its key was compromised; mitigations: deletion still revokes immediately, and mTLS trust is rebuilt on every CA-set change. Reversible as one branch.
2. **Empty-publish policy** — when a retention rebuild fails AND nothing else is servable, [the zero-certificate guard retains the old state](https://github.com/HarperFast/harper/pull/2384/changes?w=1#diff-ea5297f5d9097656312d08f0a1534f0426fb6ef893a2f5bf3a456b18f81a6841R1384) rather than publishing a drop: availability outranks the drop in that corner (reaching it requires OpenSSL to reject CA material that already passed X509 parsing). Stated in the code and DESIGN.md rather than hidden; the alternative is clearing the default and failing handshakes.
3. **Trust-awareness is scoped to mTLS** — [non-mTLS listeners retain their existing context object untouched](https://github.com/HarperFast/harper/pull/2384/changes?w=1#diff-ea5297f5d9097656312d08f0a1534f0426fb6ef893a2f5bf3a456b18f81a6841R1334) across CA changes (nothing they consult depends on the CA set); only their mirrored CA bookkeeping refreshes. A rebuilt retention keeps the ORIGINAL leaf + appended chain, which is correct for the old leaf.
4. **Retention is unbounded** — no TTL, no notAfter awareness: a permanently stuck rotation serves last-good until the cert expires, with an error-level log on signature change and an error-level summary at most hourly. Escalating inside the final validity window is cheap to add later.
5. **Default candidacy is retained** — a failing record keeps the non-SNI default it earned (necessary: the incident cluster's cert was default-only). A reviewer could prefer the default to fall through to the best healthy candidate; that would have re-served the self-signed cert in the incident.
6. **Retry lives in the selector** — each selector arms its own backoff timer per thread (1.5s doubling to 5min cap) rather than a shared scheduler; moving this later means extracting closure state.
7. **The e2e skip is env-gated** — `HARPER_TEST_REQUIRE_FILE_WATCHERS=1` is wired into the Linux integration matrix so the cert-key-reload skip cannot silently retire the #2382 end-to-end assertion there; other environments still skip when file-watch delivery is broken.
8. **Carried test nit (final review)** — the corrupt-row throttle assertion bounds log growth to ≤1 per 4s window rather than zero, because suite-accumulated selectors can emit late first-logs; a true re-logging regression adds one line per selector per 1.5-3s cycle and trips it immediately. The reviewer would prefer a fully isolated selector harness; that's a test-infra follow-up.

## Verification

- Unit (changed area): [`unitTests/security/keys.test.js`](https://github.com/HarperFast/harper/pull/2384/changes?w=1#diff-9e36586390ca42fc7960c98e46c11d98d713e7fa63aedb0930ce7d63446df167R981) — 71 passing, including 9 new tests: retain-last-good + heal-on-retry, deletion drops contexts, hostname-set changes drop stale entries, non-mTLS retention keeps its object across CA changes, an mTLS retained context stops trusting a REMOVED CA (asserts `options.ca` contents), corrupt authority row reports through the throttle and never claims recovery, whole-pass throw leaves live state untouched, watcher latch rollback (sync + async + stale-rejection guard). Fails-on-base: the core regression tests fail against `origin/main` (verified by rebuilding base `dist` and rerunning).
- End-to-end: `integrationTests/security/cert-key-reload.test.ts` rewritten off the behavior this PR removes — confirms the renewed cert's arrival through the system table (ops API), [asserts every handshake keeps serving the last-good pair](https://github.com/HarperFast/harper/pull/2384/changes?w=1#diff-a88618c9af1b3c52862cf6e69bfb349502038347c2c4f7a858ff6e173589f4a4R215) while the matching key is absent, then delivers the key and requires convergence on every worker via a deadline-retried handshake hammer. Green on this branch; on base it fails at the retain assertion serving the self-signed serial (the incident symptom).
- Full local gates (macOS; CI on ubuntu is authoritative): `test:integration:all` 1875/1897 — the 4 failures are 3 that reproduce identically on clean `origin/main` in this environment (shutdown-drain A/B, rolling-restart) and 1 flake green on retry (deploy-tracking peer-branch); `test:unit:resources` 1789/0; `test:unit:lmdb` green except the 2 pre-existing `deploy by_ref` git-fixture before-hook failures that reproduce on clean main; `test:unit:main`/`test:unit:apitests` stall in darwin process-group teardown in this environment (before and after the change) — their recorded failures are the same pre-existing pair.
- Incident heal path exercised live on the affected cluster (see #2382); the mechanism this PR fixes was verified from the transaction log and compiled 5.1.22 code.

Complexity: complicated

<sub>Review-Coverage: authored=claude; ran=codex; adjudicated=domain; blocked=cursor-grok(not-installed),gemini(exit-1); declined=cursor-composer; rounds=9 @ de5f428c022a</sub>

<sub>Human-Review-Need: 4 (decisions: tie-breaks-favor-incumbent, session-invalidation-scope, empty-publish-policy, trust-awareness-scoped-to-mtls, ca-bookkeeping-mutation-site, rebuild-keeps-original-chain, suite-selector-teardown, strict-gate-linux-only, retention-unbounded, default-candidacy-retained, retry-lives-in-the-selector, design-doc-as-the-record) @ de5f428c022a</sub>


security/keys.ts
```diff
@@ -1195,22 +1296,92 @@ export function createTLSSelector(type, mtlsOptions?, liveReload = true): any {
...
+							(previous as any).certificateAuthorities = Array.from(candidateCAs);
+							return previous;
+						}
+						if (rebuiltRetentions.has(previous)) return rebuiltRetentions.get(previous);
+						try {
+							const secureOptions = {
+								...(previous as any).options,
+								availableCAs: caCerts,
+								ca: mtlsOptions && Array.from(candidateCAs.values()),
+							};
```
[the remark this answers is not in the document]
DavidCockerill: Applied in ad65b3619 — the rebuild now takes `getTicketKeys()` like every fresh build. Worth noting the retained-unchanged path (and every long-lived context) already carries its build-time ticket keys; rotating those on rebuild is strictly more current, and a broader rotate-contexts-on-ticket-rotation question is pre-existing design outside this PR.

— Claude Fable 5

security/keys.ts
```diff
@@ -1103,14 +1191,14 @@ export function createTLSSelector(type, mtlsOptions?, liveReload = true): any {
...
-							logger.error?.('Skipping unparseable certificate record', cert.name, error);
+							// A pass failure like any other (throttle + retry). Its caCerts entry can't be
+							// retained — the subject is unrecoverable from a PEM that won't parse.
+							failedThisPass.push({ cert, error });
 							continue;
 						}
 						if (cert.is_authority) {
 							(certParsed as any).asString = certificate;
-							caCerts.set(certParsed.subject, certificate);
+							candidateCAs.set(certParsed.subject, certificate);
```
kriszyp, line 1206: This candidate CA map and the serving contexts come from two separate `hdb_certificate.search([])` calls. Each search owns a separate read snapshot, so another worker can commit between them: the pass can publish a renewed leaf with the previous CA chain/trust set. That can break certificate chains or continue trusting a removed client CA until a later subscription rebuild, contradicting the transactional/new-handshake guarantee. Please materialize one records array from a single search snapshot and drive both loops from it.

security/keys.ts
```diff
@@ -1195,55 +1285,133 @@ export function createTLSSelector(type, mtlsOptions?, liveReload = true): any {
...
+						}
+					};
+					for (const { cert } of failedThisPass) {
+						const retain = (previous, hostname?) => {
+							const retained = retainable(previous);
+							if (!retained) return;
+							const previousQuality = (retained as any).quality ?? 0;
+							if (
+								hostname !== undefined &&
+								previousQuality > ((candidateContexts.get(hostname) as any)?.quality ?? 0)
```
kriszyp: The strict `>` comparison does not preserve last-good ownership on ties. With equal-quality certificates A and B, if A is the current default and A's renewed cert fails to match its key, B builds first at quality q; retained A then cannot beat q, so non-SNI traffic abruptly serves B. The same applies to an overlapping hostname A previously owned. Please let the retained context reclaim each hostname it previously owned on a tie, and separately prefer the retained current `defaultContext` on a tie. Avoid changing the shared comparison blindly: a context retained only for a hostname must not become the default. An equal-quality regression case would pin both rules.

security/keys.ts
```diff
@@ -1195,55 +1285,133 @@ export function createTLSSelector(type, mtlsOptions?, liveReload = true): any {
...
 						return;
 					}
+					// Publish: reconcile the live maps in place (their identity is aliased by
+					// server.secureContexts and every context's availableCAs) and advance all default
+					// references together so SNI and non-SNI traffic can't diverge.
+					secureContexts.clear();
+					for (const [hostname, context] of candidateContexts) secureContexts.set(hostname, context);
+					caCerts.clear();
+					for (const [subject, certificate] of candidateCAs) caCerts.set(subject, certificate);
+					if (candidateHasWildcards) hasWildcards = true;
```
kriszyp: This only ever sets `hasWildcards` to true. After the last wildcard certificate is deleted or changed to exact SANs, every unmatched SNI handshake still walks the domain suffixes even though the replacement map cannot contain wildcard keys. Please publish the replacement boolean directly with `hasWildcards = candidateHasWildcards`.

unitTests/security/keys.test.js
```diff
@@ -978,6 +978,402 @@ describe('Test keys module', () => {
...
+			keyPairB = await mkcert.createCert({
+				domains: ['retain-b.harper.test'],
+				validityDays: 1,
+				ca: { key: ca_key, cert: test_ca },
+			});
+		});
+
+		beforeEach(() => {
+			databases = require('#src/resources/databases').databases;
+			liveTLSRebuilders = keys.__get__('liveTLSRebuilders');
```
kriszyp, line 1007: These additions reach module internals through rewire and later add sinon-backed failure injection, despite the repository rule forbidding new rewire/sinon uses. Clearing `liveTLSRebuilders` also does not end each selector's table subscription or retry timer, so the per-case live selectors accumulate; the later test's workaround for “every selector accumulated by this suite” confirms the isolation leak. Please exercise public selector/table behavior with real modules and avoid creating one permanent live selector per case. For branches that lack a non-stub seam, document the missing seam on the issue rather than extending the legacy harness.

kriszyp (commented):
Good work, I think looks like a solid improvement in reliability. I think codex spotted a few race conditions to close up.

> gemini(exit-1);

Do you know why gemini/agy is failing?

security/keys.ts
DavidCockerill: Fixed in a06781340 — both loops now iterate one materialized records array from a single search() snapshot, so the CA set and the leaf records can no longer come from different points in time. (The two-call shape predates this PR, but transactional publication is exactly the place it stops being acceptable.)

— Claude Fable 5

security/keys.ts
DavidCockerill: Fixed in a06781340, scoped as you specified: retention reclaims a hostname the retained context already owned on >=, and the live default wins the default tie only when it IS the retained record — a context retained merely for a hostname still needs strict > to become the default, so it can't steal non-SNI traffic on a tie. The new equal-quality regression test pins both rules (sibling B claims the incumbent's hostname while A fails; A must keep both the hostname and the default, by object identity).

— Claude Fable 5

security/keys.ts
DavidCockerill: Fixed in a06781340 — the replacement flag publishes directly (hasWildcards = candidateHasWildcards). One wrinkle: retention can carry a wildcard-keyed entry forward, and those keys are stored sliced (leading '.'), so the retain path marks candidateHasWildcards for them — otherwise a retained wildcard entry would become unreachable the pass after its record failed.

— Claude Fable 5

unitTests/security/keys.test.js
DavidCockerill: Partly accepted, partly proposing your own fallback: the rewire/__get__ + sinon-sandbox pattern is the established seam throughout this exact file (20+ prior uses), and these tests drive real tables, real selectors, and real chokidar-adjacent pollers wherever a public surface exists — the internals reached are the logger binding, the rebuild trigger, and one fault-injection stub for the whole-pass branch, none of which have a public seam today. The isolation leak you point at is real, though: createTLSSelector offers no way to end a selector's subscription or retry timer, so suite selectors accumulate — that's precisely why the throttle test is publish-anchored. Rather than extending the legacy harness further or inventing a disposal API inside this PR, I've documented the missing seam on #2382 as the test-infra follow-up (a disposable/endable selector handle would also let getReplicationCert stop leaking its transient registrations). If you'd rather the seam land in this PR, say so and I'll scope it.

— Claude Fable 5

DavidCockerill:
Re: gemini/agy — yes, root-caused: agy itself works (1.1.22, verified round-trip locally), but the prepush CLI's gemini adapter pins `gemini-3.1-pro-high`, and the Antigravity endpoint rejects that for this location: `Selected model is not supported in the selected location` (from the leg's log). Deterministic config, not flakiness — the fix is the adapter's model pin (or region), which is why the leg failed identically on all seven rounds and got reported as blocked in the coverage footer rather than silently skipped.

— Claude Fable 5

security/keys.ts
```diff
@@ -1195,55 +1289,139 @@ export function createTLSSelector(type, mtlsOptions?, liveReload = true): any {
...
 						return;
 					}
+					// Publish: reconcile the live maps in place (their identity is aliased by
+					// server.secureContexts and every context's availableCAs) and advance all default
+					// references together so SNI and non-SNI traffic can't diverge.
+					secureContexts.clear();
+					for (const [hostname, context] of candidateContexts) secureContexts.set(hostname, context);
+					caCerts.clear();
+					for (const [subject, certificate] of candidateCAs) caCerts.set(subject, certificate);
+					hasWildcards = candidateHasWildcards;
```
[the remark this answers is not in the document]
DavidCockerill: Declining the test as unwritable today: `hasWildcards` is per-selector closure state (not reachable via `keys.__get__` — that only sees module scope), and toggle-off has no observable behavior through the public callback: with no wildcard keys in the replacement map, the suffix walk finds nothing and serves the default either way — the flag only short-circuits the walk's cost. Pinning it needs the same disposal/inspection seam already documented on #2382 from the harness discussion; the toggle-off line itself is pinned by review here and by the DESIGN.md contract.

— Claude Fable 5

unitTests/security/keys.test.js
```diff
@@ -978,6 +978,450 @@ describe('Test keys module', () => {
...
```
kriszyp, line 1007: These additions reach module internals through rewire and later add sinon-backed failure injection, despite the repository rule forbidding new rewire/sinon uses. Clearing `liveTLSRebuilders` also does not end each selector's table subscription or retry timer, so the per-case live selectors accumulate; the later test's workaround for “every selector accumulated by this suite” confirms the isolation leak. Please exercise public selector/table behavior with real modules and avoid creating one permanent live selector per case. For branches that lack a non-stub seam, document the missing seam on the issue rather than extending the legacy harness.

kriszyp (approved):
> Selected model is not supported in the selected location

You need to login with the "global" region instead of the "us" region.

Anyway, the PR looks good.

unitTests/security/keys.test.js
DavidCockerill: Verbatim re-post of the thread answered at https://github.com/HarperFast/harper/pull/2384#discussion_r3884421718 — short version: the rewire/sinon pattern is this file's established seam (20+ prior uses; the tests drive real tables/selectors wherever a public surface exists), the selector-accumulation leak you describe is real and is documented on #2382 as the missing disposal seam per your own fallback guidance, and since then de5f428c0 made the throttle test attributable (unique listener type) so it no longer depends on foreign selectors at all. If you want the disposal seam in THIS PR rather than as a follow-up, say so in the original thread and I'll scope it.

— Claude Fable 5

kriszyp (approved):
Good

Issue #94, opened 2026-01-14
Add .harperignore for Watched Files & Deploys


dawsontoth:
### Feature Summary

Add support for a new file, `.harperignore`, which will be used to ignore files when watching for changes, and will ignore files when building up for `harper deploy`.

### Problem This Solves

Without this, to do a limited deployment, I have to create a sub-directory and copy artifacts into it, then `harper deploy target=AppName` within it. The watched files doesn't have a workaround.

### Proposed Solution

Create the ignore file.

### Priority/Impact

None

PR #2385, opened 2026-08-28
ci: adopt ai-review-prompts #91 — reusable-owned cancellation + caller permissions ceilings

heskew:
Atomic migration per [ai-review-prompts#91](https://github.com/HarperFast/ai-review-prompts/pull/91): every ai-review-prompts pin bumps to `af81872` (post-#94) **and** the review callers' workflow-level `concurrency` blocks are removed **in the same change** — cancellation now lives in the reusables' review jobs (a job-level group that engages only after the authorize, skip, and key gates pass), so an unauthorized or skipped event can never cancel a legitimate in-flight review. Retaining the legacy caller group would have preserved the old pre-gate cancellation residual, hence the atomicity.

Also: the mention and issue-to-pr calling jobs gain explicit `permissions` ceilings (union of the reusables' job grants — `contents`/`pull-requests`/`issues`/`id-token: write`), so their grants stop depending on the repo's default-workflow-permissions setting, matching the review callers' long-standing pattern.

**Ordering safeguard included** ([ai-review-prompts#94](https://github.com/HarperFast/ai-review-prompts/pull/94), requested by @kriszyp on harper#2385): the reusables now refuse admission to stale events (freshness re-checked as the last act of every admitting path in authorize) and a run superseded after acquiring the concurrency slot fails itself instead of reviewing a stale head — so dropping the caller-side cancellation cannot let an older run cancel a newer admitted review. Residual (a push landing between authorize's final read and queue entry) is documented and tracked in ai-review-prompts#92.

**Expected on this PR**: the claude-review check errors with the workflow-must-match-default-branch validation — normal for workflow-editing PRs.

**After merge**: release-branch sync PRs follow (same content, per branch).

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -37,7 +37,7 @@ jobs:
     # opt-in label, so a draft opted in via `claude-review` resumes
     # review when it flips ready even with CLAUDE_ALWAYS_ON unset.
     if: ${{ (github.event.action == 'labeled' && github.event.label.name == 'claude-review') || (github.event.action != 'labeled' && vars.CLAUDE_ALWAYS_ON == 'true') || (github.event.action == 'ready_for_review' && contains(github.event.pull_request.labels.*.name, 'claude-review')) }}
-    uses: HarperFast/ai-review-prompts/.github/workflows/_claude-review.yml@765a51b545f1b2b6c1ffe3b3731baa1512e44dac # main 2026-08-28 (#91 job-level post-authorization cancellation; #90 cost gates; #89 defaults; #88 lenses)
+    uses: HarperFast/ai-review-prompts/.github/workflows/_claude-review.yml@af81872e56dd63da1160adb5eae82d359a456dde # main 2026-09-01 (#94 freshness-gated admission; #91 job-level cancellation; #90 cost gates; #89 defaults; #88 lenses)
     # Caller-side permissions at the calling-job level (NOT workflow-
     # level — that placement caps the reusable's per-job grants below
     # what they need and breaks the workflow at startup; see
@@ -64,7 +64,7 @@ jobs:
       # introspect their own ref (`github.workflow_ref` resolves to the
       # CALLER's ref in `workflow_call` context), and `uses: …@<ref>`
       # is parsed literally so we can't interpolate a variable.
-      ai-review-prompts-ref: 765a51b545f1b2b6c1ffe3b3731baa1512e44dac
+      ai-review-prompts-ref: af81872e56dd63da1160adb5eae82d359a456dde
       review-layers: |
         universal
         harper/common
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -30,8 +30,9 @@ jobs:
     # always-on toggle". CLAUDE_ALWAYS_ON=true (repo/org variable) → auto-
     # review trusted-author PRs; unset → opt-in via the claude-review
     # label. The reusable's authorize job still owns WHO is admitted.
-    # Note: the `claude-review` label name is matched there too —
-    # `_claude-review.yml`'s authorize `if:`, not in this caller.
+    # Note: the `claude-review` label name is deliberately duplicated —
+    # the caller gate names it (below) AND the reusable's authorize
+    # matches it; renaming the label means changing both.
     # `ready_for_review` is admitted when the PR still carries the
     # opt-in label, so a draft opted in via `claude-review` resumes
     # review when it flips ready even with CLAUDE_ALWAYS_ON unset.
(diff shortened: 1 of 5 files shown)


.github/workflows/claude-review.yml
```diff
@@ -20,17 +20,9 @@ on:
...
-  # `if:` — the cancelling set must be a subset of the running set.
-  # ready_for_review runs review (job gate) but never cancel: their
-  # authorization is author-based, so on a bot-authored PR they could
-  # kill a labeler-authorized run and then be rejected themselves.
-  # Ineligible events (per the review job's opt-in gate) get a unique
-  # run_id group so they can never cancel an in-flight eligible review —
-  # workflow-level cancel-in-progress fires before job `if:` is evaluated.
-  group: claude-review-${{ github.event.pull_request.number }}-${{ ((github.event.action == 'labeled' && github.event.label.name == 'claude-review') || (github.event.action != 'labeled' && github.event.action != 'ready_for_review' && vars.CLAUDE_ALWAYS_ON == 'true')) && 'eligible' || github.run_id }}
-  cancel-in-progress: true
+# No concurrency block: cancellation is owned by the reusable's review
```
kriszyp: Can we add a monotonic ordering safeguard before removing the caller-side cancellation? With authorization now running independently, pushes A then B can reach the reusable review job in B→A order. GitHub does not guarantee concurrency-group ordering, so stale A can enter last, cancel B via `cancel-in-progress`, and either publish a stale review or leave the current head unreviewed. The same risk applies at `.github/workflows/gemini-review.yml:37`. Please ensure an older run cannot cancel a newer admitted run—for example, by serializing admission or queueing admitted runs and freshness-checking each after it acquires the slot.

.github/workflows/claude-review.yml
```diff
@@ -20,31 +20,24 @@ on:
...
```
kriszyp, line 23: Can we add a monotonic ordering safeguard before removing the caller-side cancellation? With authorization now running independently, pushes A then B can reach the reusable review job in B→A order. GitHub does not guarantee concurrency-group ordering, so stale A can enter last, cancel B via `cancel-in-progress`, and either publish a stale review or leave the current head unreviewed. The same risk applies at `.github/workflows/gemini-review.yml:37`. Please ensure an older run cannot cancel a newer admitted run—for example, by serializing admission or queueing admitted runs and freshness-checking each after it acquires the slot.

kriszyp (approved):

.github/workflows/claude-review.yml
heskew: Agreed, and holding this PR until it exists — the same race was flagged by the pre-push lane on the oauth twin and filed as ai-review-prompts#92. Design, matching your slot-acquire option: (1) the reusables' authorize job gains a freshness output (event head SHA vs the PR's live head) and the review job requires it, so a stale run is never admitted to the cancelling group — the only path to cancelling a newer admitted run disappears; (2) the review job's debounce step re-checks freshness after acquiring the slot and cancels its own run as superseded if the head moved in the authorize→queue window, so no stale review publishes either. Lands in ai-review-prompts first (through the pre-push lane), then this wave's pins advance to include it before merging. Will link the PR here.

heskew: Implemented: ai-review-prompts#94 (fixes #92) — pre-queue freshness gate (stale events never admitted to the cancelling group) + post-acquire self-fail (no actions grant, red check lands on the superseded head). Went through the pre-push lane twice; the residual seconds-wide reorder window is documented rather than overclaimed, with #92 kept open for full closure via re-dispatch. This wave's pins advance to include it once #94 merges — holding the merge queue until then, per your gate.

cb1kenobi (commented):
Barbarian reviewed b40869aa and found no blocking issues. No new blocking findings were identified. The concurrency ordering risk is already covered by the existing discussion and was not repeated.

—
_Generated by Barber AI_

.github/workflows/claude-review.yml
heskew: ai-review-prompts#94 is merged (af81872) and this wave's pins are advancing to it now — the safeguard exists before the caller-side cancellation goes away, as you asked. Design as landed: pre-queue freshness gate (assess re-checks the live head as the last act of every admitting path; a stale run is never admitted to the cancelling group) plus a post-acquire re-check that fails the run if superseded after taking the slot — no new token grants, every wire test-pinned, four pre-push lane rounds plus two Codex re-reviews. Residual (a push between authorize's final read and queue entry) is documented as visible/recoverable; ai-review-prompts#92 stays open for closing it via re-dispatch. Resolving this thread once the pin commit is on this branch.

cb1kenobi (commented):
Barbarian reviewed 02a3508a and found no blocking issues. No confirmed blocking findings were found. The new commit consistently advances all reusable-workflow pins to the freshness-gated revision, and the previously raised ordering risk was not repeated.

—
_Generated by Barber AI_

cb1kenobi (approved):

Issue #2158, opened 2026-08-13
5.2.0 data-migration stamp not durably persisted — upgrade process re-ran on consecutive boots (data version stayed 5.1.26)


heskew:
### Symptom

On an instance upgraded in-place from 5.1.26 to 5.2.0, **two consecutive boots** both logged:

```
[notify] Starting upgrade process: applying 1 data migration to bring data from 5.1.26 up to software 5.2.0.
[notify] Applying migration 1 of 1 (introduced in 5.2.0): create system.hdb_secret table for the secrets store
```

i.e. the recorded data version was still 5.1.26 at the second boot — the stamp from the first migration run did not persist. A later boot showed no upgrade lines, so it eventually stuck; the migration itself appears idempotent (`system.hdb_secret` contents survived), but the stamp write is evidently not reliably durable at the time the migration completes.

### Why it matters

A create-table migration re-running on every boot is only safe while every migration happens to be idempotent; the stamp is the invariant. Worth checking whether the data-version write is ordered/synced correctly relative to migration completion (observed on a Fabric-hosted harper-pro 5.2.0 container; timestamps and boot evidence available).

PR #2398, opened 2026-08-30
fix(upgrade): fail the boot when the data-version stamp is not recorded

kriszyp:
An upgrade boot could report success while `system.hdb_info` still held the old data version, so the migration re-ran on every subsequent boot. Two independent paths produced that state; both are closed here, and boot now refuses to continue when the stamp is not confirmed.

**The failure was swallowed.** `runUpgrade` caught everything `insertHdbUpgradeInfo()` threw, logged it, and returned normally — so `bin/run.ts`'s existing "Got an error while trying to upgrade your Harper instance… Exiting Harper." / `process.exit(1)` path never fired. [It now logs what state the instance is in and rethrows](https://github.com/HarperFast/harper/pull/2398/changes?w=1#diff-0eacbdd46d3341512e4b66dda191aa163d06db8b53e41a826bc0d15ca76ad4f6R121).

**The stamp could also no-op with no error at all.** `insertHdbUpgradeInfo` derives the new `info_id` as `max(existing) + 1` from a search whose failures `getAllHdbInfoRecords` swallows into `[]`. On that path `newId` is `1`, which collides with the install record — and `createRecords` sets `requires_no_existing`, so `upsertRecords` pushes a colliding key onto `skipped` rather than writing it. The call then **resolves successfully having written nothing**. [`insertHdbUpgradeInfo` now requires the expected id to come back as inserted](https://github.com/HarperFast/harper/pull/2398/changes?w=1#diff-1d40881acb0308d737478bce63620dd5fac186370bf03bbf6eb529fcfc176c18R126) and [throws `VersionStampNotRecordedError`](https://github.com/HarperFast/harper/pull/2398/changes?w=1#diff-1d40881acb0308d737478bce63620dd5fac186370bf03bbf6eb529fcfc176c18R54) otherwise, which covers all three of its call sites (the upgrade path plus the downgrade-confirmed and no-directives-needed stamps in `getVersionUpdateInfo`). This second path fits the issue report better than the first: the reported boots show the migration notices and no stamp error.

The invariant: **`upgrade()` returns normally only if the new data version is actually recorded**, because the stamp is the only thing stopping a directive re-running against data it has already migrated.

`Framing-Verdict: chosen-approach-sound` (planning review, before implementation).

## Re-deciding the pinned trade-off

[The reversed unit test](https://github.com/HarperFast/harper/pull/2398/changes?w=1#diff-e1f8115a77132ba294cdb599d3da91c2aa2ae0d5a730b8be70126f1370b4b3e1R92) pinned "log and continue" by name. Three facts decided the reversal:

- **The swallow was the outlier, not the policy.** `insertHdbUpgradeInfo`'s other two call sites are already fatal to boot — they run inside `getVersionUpdateInfo`'s try/catch, which `log.fatal`s and rethrows, and `bin/run.ts` exits 1.
- **What was ignored is not bookkeeping.** QA finding F-296 measured that today's re-run is benign (4 record shapes byte-identical by sha256 after a forced re-run, positive control armed) — which is exactly why the swallow survived. It costs nothing until the first non-idempotent directive, at which point it costs data.
- **Failing boot leaves a safe state.** The data is already migrated and the process exits naming the version; booting with a stale stamp leaves an instance that re-migrates on every boot and reports success each time.

## For the human reviewer

Decisions taken during the pre-push review that a reader should weigh rather than assume:

- **An orchestrated restart loop is the accepted cost, and it is not new.** One reviewer called the exit a blocker: a supervisor (systemd, Docker, Kubernetes) will restart Harper immediately, so the "do not restart" advice cannot be honoured unattended. The directives are what mutate data, and they re-run once per boot either way — under today's behaviour the instance *also* comes up and serves traffic while doing it, so this change strictly reduces exposure and makes the loop visible. [The failure message](https://github.com/HarperFast/harper/pull/2398/changes?w=1#diff-0eacbdd46d3341512e4b66dda191aa163d06db8b53e41a826bc0d15ca76ad4f6R117) now tells the operator to stop the supervisor first. Whether a stamp failure should additionally use an exit code or marker that suppresses automatic restart is a design decision left open.
- **A transient search failure now fails boot on a healthy instance.** If `pSearchSearchByValue` throws inside `getAllHdbInfoRecords` during the stamp, `newId` collapses to `1`, the insert is skipped, and boot exits — indistinguishable from a real persistence failure. That is the intended fail-closed direction, but it is a real behaviour change. The latest review round sharpened the consequence: because the guard lives in `insertHdbUpgradeInfo` rather than in `runUpgrade`, it also arms the no-directives stamp in `getVersionUpdateInfo`, where the throw escapes before `bin/run.ts` reaches `upgrade.upgrade()` — so that path gets the generic `UPGRADE_ERR` line, not the supervisor-restart guidance. Making the postcondition state-shaped (re-read the latest record and accept a `data_version_num` that already equals the target) instead of receipt-shaped would remove the false negative; that is a deliberate open decision, not an oversight, and is left for the human reviewer.
- **`insertHdbInstallInfo` was deliberately left alone.** It has the same silent-skip exposure (a re-install where `info_id` 1 exists resolves without writing). Extending the check there needs the install path's re-run semantics traced first, and this change is scoped to the upgrade stamp. Reported separately rather than changed here.
- **The failure path has no end-to-end test, by construction.** Both remaining reviewer findings ask for one. Forcing the stamp to fail requires injecting a storage or search failure at boot, which nothing in the integration harness can do. The unit tests carry that contract; [the integration suite](https://github.com/HarperFast/harper/pull/2398/changes?w=1#diff-996ba1f92c68ff708bbbe38d893ef9984c2f76f1789e1a3201313b714a5435a7R118) anchors the postcondition that had no coverage at all.
- **Two review-bot suggestions were taken, one declined.** Falling back to `packageJson.version` for the stamped version and optional-chaining the test's `package.json` read are in. Adding null guards on `assertVersionRecorded`'s `expectedId`/`newVersionString` was declined: both are computed by the only caller a few lines above, and with the version fallback in place neither can arrive undefined.
- **The original error is not logged twice.** The new message deliberately omits `err`; `bin/run.ts` logs the thrown error on the exit path it reaches.
- **`storage.writeAsync: true` durability is out of scope.** It sets LMDB `noSync`, so a *successfully written* stamp can still be lost to a hard kill — the mechanism the issue title hypothesises. Nothing here syncs the stamp; a storage-engine-specific sync at boot is a separate decision.

## Verification

Re-verified after rebasing onto `main` at `3ea2639ae` (see the rebase note below); the PR's own five commits are unchanged.

- `npm run build` — clean.
- `npm run test:unit:resources` — **1841 passing, 23 pending, 0 failing.** `Caching › Can load cached indexed data` is green again.
- `npm run test:unit:main` — 5129 passing, 195 pending, 1 failing: `configValidator › does not warn when a relative rootPath resolves within the limit`. That case resolves `relative/root` against `process.cwd()`, so it only fails from a deeply nested checkout path; it is green in CI on `main` and on this branch.
- `npx mocha unitTests/bin/upgrade.test.js unitTests/dataLayer/hdbInfoController.test.js unitTests/dataLayer/hdbInfoVersionStamp.test.js` — 10 passing, 19 pending.
- Earlier rounds (pre-rebase, unchanged code): `npm run test:integration -- "integrationTests/upgrade/**/*.test.ts"` — 16 passing, 3 skipped (env-gated on a prior-release install), 0 failing; `integrationTests/server/**` — 189 passing, the 6 failures needing a local Ollama server.
- **Fails on base:** with `bin/upgrade.js` and `dataLayer/hdbInfoController.ts` reverted and `dist` rebuilt, the reversed unit test fails with `expected undefined to equal Error: Oh boy...it is an error`, and the new stamp tests cannot run at all (`assertVersionRecorded is not a function`).

### Why CI was red, and where it stands now

All three `Unit Test (Node.js v22/v24/v26)` jobs failed on the same case, `Caching › Can load cached indexed data` — "the fencing source fill should reach the indexed subscription". That was a **red on `main`**, not a regression here: it is the fence named in the commit message of [`b3235b8c6` "Deliver RocksDB subscription events for source fills whose version differs from the log key"](https://github.com/HarperFast/harper/commit/b3235b8c6e35602421b20a697f44be03269c3b72), which landed after this branch's merge-base. The per-node extras were separate known flakes — `HNSW greedy routing above layer 0` on v22 (open: [#2373](https://github.com/HarperFast/harper/pull/2373)) and an `ENOTEMPTY` teardown on v26 (open: [#2276](https://github.com/HarperFast/harper/pull/2276)).

Checks run against the PR's merge ref, so `main`'s fix reaches this PR whether or not the branch is rebased; the rebase re-triggered CI and makes the local tree match what CI actually tests, which is what the verification above was run against. **All three `Unit Test` jobs now pass at `127d1adbf`.**

**One check is red, and it is `main`'s, not this diff's.** `Unit Test (Windows, Node.js v24)` — the gate added by [#2361](https://github.com/HarperFast/harper/pull/2361) after this branch's base — fails on the `unitTests/config/**` group (1 of 9), on two cases from [#2245](https://github.com/HarperFast/harper/pull/2245): `storage exhaustion during boot (#847) › isStorageExhausted › recognizes EDQUOT by errno when the platform gives no usable code` (`false !== true`) and `› persistConfigDuringBoot › swallows storage exhaustion so startup continues` (`Error: Unknown system error -undefined`) — Windows has no `EDQUOT` errno. The failure is byte-identical on `main`'s own runs ([job on `655787409`](https://github.com/HarperFast/harper/actions/runs/33397431652/job/99505218824)), which have been red on it since the gate landed. Nothing in this diff touches `unitTests/config/` or `config/configUtils.ts`.

`Integration Tests 6/6 (Windows, Node.js v24)` is red for the same reason: `an application with a branched database › keeps the application's own data across a restart` (`a restart must not discard the branch`), from [#2352](https://github.com/HarperFast/harper/pull/2352). It fails on `main`'s own last two Integration runs ([`655787409`](https://github.com/HarperFast/harper/actions/runs/33397431612), [`1db23823b`](https://github.com/HarperFast/harper/actions/runs/33397414407)). Both Windows gates landed on `main` after this branch's base, so this is the first run of #2398 that has ever seen them.

Refs #2158

Complexity: complicated

<sub>Review-Coverage: authored=claude; ran=gemini,cursor-composer,codex; adjudicated=domain; declined=cursor-grok; rounds=5 @ 127d1adbf1fb</sub>

<sub>Human-Review-Need: 3 (decisions: fail-boot-on-unrecorded-stamp, guard-placed-in-controller-not-caller, receipt-vs-state-postcondition, integration-suite-that-passes-on-main) @ 127d1adbf1fb</sub>

### DIFF
--- a/bin/upgrade.js
+++ b/bin/upgrade.js
@@ -112,7 +112,7 @@ async function runUpgrade(upgradeObj) {
 
 	// The stamp is what stops these directives running again on data they have already migrated.
 	try {
-		await hdbInfoController.insertHdbUpgradeInfo(upgradeObj[UPGRADE_VERSION]);
+		await hdbInfoController.insertHdbUpgradeInfo(upgradeObj[UPGRADE_VERSION] ?? packageJson.version);
 	} catch (err) {
 		printToLogAndConsole(
 			`The data upgrade completed, but the new data version could not be recorded in the '${hdbTerms.SYSTEM_TABLE_NAMES.INFO_TABLE_NAME}' table, so Harper will not start. Every restart re-runs the upgrade directives against already-upgraded data, so stop any process supervisor that restarts Harper automatically, then check the logs and contact ${hdbTerms.HDB_SUPPORT_ADDRESS}.`,
--- a/bin/upgrade.js
+++ b/bin/upgrade.js
@@ -110,11 +110,16 @@ async function runUpgrade(upgradeObj) {
 		throw err;
 	}
 
+	// The stamp is the only thing stopping these directives from running again on the next boot,
+	// against data they have already migrated, so a boot that cannot record it must not continue.
 	try {
 		await hdbInfoController.insertHdbUpgradeInfo(upgradeObj[UPGRADE_VERSION]);
 	} catch (err) {
-		hdbLogger.error("Error updating the 'hdb_info' system table.");
-		hdbLogger.error(err);
+		printToLogAndConsole(
+			`The data upgrade completed, but the new data version could not be recorded in the '${hdbTerms.SYSTEM_TABLE_NAMES.INFO_TABLE_NAME}' table, so Harper will not start. Restarting re-runs the upgrade directives against already-upgraded data: please check the logs and contact ${hdbTerms.HDB_SUPPORT_ADDRESS} before doing so.`,
+			hdbTerms.LOG_LEVELS.ERROR
+		);
+		throw err;
 	}
 }
 
--- a/bin/upgrade.js
+++ b/bin/upgrade.js
@@ -111,7 +111,7 @@ async function runUpgrade(upgradeObj) {
 	}
 
 	// The stamp is the only thing stopping these directives from running again on the next boot,
-	// against data they have already migrated, so a boot that cannot record it must not continue.
+	// against data they have already migrated.
 	try {
 		await hdbInfoController.insertHdbUpgradeInfo(upgradeObj[UPGRADE_VERSION]);
 	} catch (err) {
--- a/bin/upgrade.js
+++ b/bin/upgrade.js
@@ -110,13 +110,12 @@ async function runUpgrade(upgradeObj) {
 		throw err;
 	}
 
-	// The stamp is the only thing stopping these directives from running again on the next boot,
-	// against data they have already migrated.
+	// The stamp is what stops these directives running again on data they have already migrated.
 	try {
 		await hdbInfoController.insertHdbUpgradeInfo(upgradeObj[UPGRADE_VERSION]);
 	} catch (err) {
 		printToLogAndConsole(
-			`The data upgrade completed, but the new data version could not be recorded in the '${hdbTerms.SYSTEM_TABLE_NAMES.INFO_TABLE_NAME}' table, so Harper will not start. Restarting re-runs the upgrade directives against already-upgraded data: please check the logs and contact ${hdbTerms.HDB_SUPPORT_ADDRESS} before doing so.`,
+			`The data upgrade completed, but the new data version could not be recorded in the '${hdbTerms.SYSTEM_TABLE_NAMES.INFO_TABLE_NAME}' table, so Harper will not start. Every restart re-runs the upgrade directives against already-upgraded data, so stop any process supervisor that restarts Harper automatically, then check the logs and contact ${hdbTerms.HDB_SUPPORT_ADDRESS}.`,
 			hdbTerms.LOG_LEVELS.ERROR
 		);
 		throw err;
--- a/dataLayer/hdbInfoController.ts
+++ b/dataLayer/hdbInfoController.ts
@@ -48,9 +48,7 @@ export class VersionStampNotRecordedError extends Error {
 
 /**
  * An insert whose key already exists is reported as skipped rather than as a failure, so a resolved
- * insert is not by itself evidence that the version was recorded. Boot reads this record to decide
- * whether the upgrade directives still need to run, so anything short of a confirmed insert of the
- * expected id has to fail closed.
+ * insert is not by itself evidence that the version was recorded.
  */
 export function assertVersionRecorded(insertResult: any, expectedId: number, newVersionString: string) {
 	const insertedHashes = insertResult?.inserted_hashes;
--- a/dataLayer/hdbInfoController.ts
+++ b/dataLayer/hdbInfoController.ts
@@ -47,10 +47,9 @@ export class VersionStampNotRecordedError extends Error {
 }
 
 /**
- * An insert whose key already exists is reported as skipped rather than as a failure, so a resolved
- * insert is not by itself evidence that the version was recorded. Ids are compared as strings: the
- * guard is against a write that did not happen, and a boot must not be refused over a hash the
- * storage layer handed back in a different numeric type.
+ * `insert.insert` reports a key that already exists as skipped, not as a failure, so a resolved
+ * insert is not evidence that the version was recorded. Ids compare as strings: the guard is
+ * against a write that did not happen, not against a hash returned in a different type.
  */
 export function assertVersionRecorded(insertResult: any, expectedId: number, newVersionString: string) {
 	const insertedHashes = insertResult?.inserted_hashes;
--- a/dataLayer/hdbInfoController.ts
+++ b/dataLayer/hdbInfoController.ts
@@ -37,6 +37,32 @@ const DEFAULT_DATA_VERSION_NUM = '2.9.9';
 // This value should change as supported versions change.
 const MINIMUM_SUPPORTED_VERSION_NUM = '3.0.0';
 
+export class VersionStampNotRecordedError extends Error {
+	statusCode: number;
+	constructor(message: string) {
+		super(message);
+		this.name = 'VersionStampNotRecordedError';
+		this.statusCode = 500;
+	}
+}
+
+/**
+ * An insert whose key already exists is reported as skipped rather than as a failure, so a resolved
+ * insert is not by itself evidence that the version was recorded. Boot reads this record to decide
+ * whether the upgrade directives still need to run, so anything short of a confirmed insert of the
+ * expected id has to fail closed.
+ */
+export function assertVersionRecorded(insertResult: any, expectedId: number, newVersionString: string) {
+	const insertedHashes = insertResult?.inserted_hashes;
+	if (Array.isArray(insertedHashes) && insertedHashes.length === 1 && insertedHashes[0] === expectedId) return;
+
+	throw new VersionStampNotRecordedError(
+		`Could not record data version ${newVersionString} in '${hdbTerms.SYSTEM_SCHEMA_NAME}.${hdbTerms.SYSTEM_TABLE_NAMES.INFO_TABLE_NAME}': ` +
+			`the insert of ${HDB_INFO_SEARCH_ATTRIBUTE} ${expectedId} reported inserted ${JSON.stringify(insertedHashes ?? null)} ` +
+			`and skipped ${JSON.stringify(insertResult?.skipped_hashes ?? null)}`
+	);
+}
+
 /**
  * * Insert a row into hdbInfo with the initial version data at install.
  *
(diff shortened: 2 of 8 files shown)


dataLayer/hdbInfoController.ts
```diff
@@ -37,6 +37,37 @@ const DEFAULT_DATA_VERSION_NUM = '2.9.9';
...
+	}
+}
+
+/**
+ * `insert.insert` reports a key that already exists as skipped, not as a failure, so a resolved
+ * insert is not evidence that the version was recorded. Ids compare as strings: the guard is
+ * against a write that did not happen, not against a hash returned in a different type.
+ */
+export function assertVersionRecorded(insertResult: any, expectedId: number, newVersionString: string) {
+	const insertedHashes = insertResult?.inserted_hashes;
```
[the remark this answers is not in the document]
kriszyp: Declining this one: `expectedId` and `newVersionString` are both computed by the only caller a few lines above, and with the `packageJson.version` fallback now in `runUpgrade` neither can arrive undefined. A guard here would only restate that. — Claude Opus

dawsontoth (approved):
Makes sense. Separate note from the PR: it would be really nice to get CI green again :\

bin/upgrade.js
```diff
@@ -110,11 +110,15 @@ async function runUpgrade(upgradeObj) {
 		throw err;
 	}
 
+	// The stamp is what stops these directives running again on data they have already migrated.
 	try {
-		await hdbInfoController.insertHdbUpgradeInfo(upgradeObj[UPGRADE_VERSION]);
+		await hdbInfoController.insertHdbUpgradeInfo(upgradeObj[UPGRADE_VERSION] ?? packageJson.version);
```
DavidCockerill, line 115: **The two consumers of a missing `upgrade_version` disagree, and this fallback makes the disagreement permanent.**

With `upgrade_version` absent, `directivesController.getVersionsForUpgrade` returns `[]` (`upgrade/directives/directivesController.ts:46-57` logs "trying to start the server anyways" and gives up), so **zero** directives run. The new `?? packageJson.version` then stamps the data as fully migrated at the running version. One consumer reads the missing field as "can't decide, run nothing"; the other reads it as "we're current". No later boot re-checks, because the stamp says there is nothing to do.

**Not reachable today** — the only caller is `bin/run.ts:170`, whose object comes from `getVersionUpdateInfo`, which sets `upgrade_version = packageJson.version` after throwing when that is falsy (`dataLayer/hdbInfoController.ts:197-200`). That's why this is a comment and not a change request.

What makes it worth raising anyway is the test below: `unitTests/bin/upgrade.test.js:71-77` now *asserts* this outcome, which pins the fail-open behaviour for whatever future caller hand-builds an `UpgradeObject`. A guard that is currently unreachable is fine; a test that locks in the unsafe answer is what turns it into a trap.

The property the fix needs: **a missing `upgrade_version` must not produce a stamp.** The same condition that makes directive selection abstain should make stamping abstain, or throw.

— DAIvid (Claude Opus 5)

dataLayer/hdbInfoController.ts
```diff
@@ -37,6 +37,37 @@ const DEFAULT_DATA_VERSION_NUM = '2.9.9';
 // This value should change as supported versions change.
 const MINIMUM_SUPPORTED_VERSION_NUM = '3.0.0';
 
+export class VersionStampNotRecordedError extends Error {
+	statusCode: number;
+	constructor(message: string) {
+		super(message);
+		this.name = 'VersionStampNotRecordedError';
+		this.statusCode = 500;
```
DavidCockerill, line 45: **This guard can't tell "the write didn't happen" from "we couldn't read the table", and only the first deserves refusing to boot.**

`expectedId` derives from `getAllHdbInfoRecords()`, which swallows a failed search to `console.error` and returns `[]` (`dataLayer/hdbInfoController.ts:147-153`). So a read failure on a table that *does* have rows yields `newId = 1`, the insert is skipped as an existing key, and `assertVersionRecorded` reports *"the insert … reported inserted [] and skipped [1]"* — telling the operator the write was refused when the actual fault was the read. Harper then refuses to start.

*In plain terms:* one bad read of `hdb_info` becomes a permanent "Harper will not start", with an error pointing at the wrong thing.

Low probability, since the search runs after `checkIfInstallIsSupported` has already proven the table loads — so non-blocking. But the fail-open `catch` in `getAllHdbInfoRecords` is what converts a read fault into a misattributed write fault, and that's the seam worth closing rather than the guard itself.

**Separately, on placement:** the guard sits inside `insertHdbUpgradeInfo`, so it also fires on the downgrade-confirmed path and the "upgraded version, no directives required" path (`:231`, `:260`). On the latter nothing was migrated, so a failed stamp costs only a repeated no-op version check next boot — yet it aborts startup, and via `getVersionUpdateInfo`'s rethrow it does so *without* the operator guidance at `bin/upgrade.js:117` (the caller sees `log.fatal` + `UPGRADE_ERR`). Fail-closed is right where directives ran; it's disproportionate where none did.

— DAIvid (Claude Opus 5)

DavidCockerill (commented):
Commenting — the intent here is right, and I'd take it with one change.

**🧊 In plain terms:** this makes Harper refuse to boot when it can't record which data version it just migrated to, which is the correct instinct — a silent failure there means the next boot thinks the migration already happened. Two rough edges: in one path Harper can now mark the data "fully migrated" after running *zero* migrations, and in another a failed *read* of the version table is reported to the operator as a failed *write*.

The one I'd actually fix before merge is the first, and specifically the test rather than the code. The fallback at `bin/upgrade.js:115` isn't reachable today — `getVersionUpdateInfo` throws before it can produce a missing `upgrade_version` — but `unitTests/bin/upgrade.test.js:71-77` now asserts the fail-open result, so the unsafe answer is pinned for any future caller that hand-builds an `UpgradeObject`. An unreachable branch is harmless; a test that locks in "stamp as current after running nothing" is how it becomes reachable later without anyone noticing.

Both inline threads state the property rather than prescribing a patch, since the right shape depends on whether you'd rather abstain or throw.

Verified and dropped: the version stamp itself is written before the boot proceeds, the refuse-to-boot message at `bin/upgrade.js:117` carries actionable operator guidance on the path that matters, and `checkIfInstallIsSupported` really does run first, which is what keeps the read-failure case improbable rather than likely.

<sub>— DAIvid (Claude Opus 5) · cross-model: Codex (graded) + Gemini + Harper domain adjudication · graded leg called the first item major; domain adjudication downgraded it on reachability, and I agree with the downgrade</sub>

PR #2426, opened 2026-09-01
Give a branched database its own blob store, cloned with hard links

kriszyp:
A branch is a RocksDB checkpoint, and a checkpoint clones the engine and nothing else. So a branch's records referenced blob ids that resolved under the branch's own — **empty** — blob root: any blob the base wrote before the branch was taken was unreadable, and once the branch wrote a blob of its own, `getNextFileId` seeded from that empty directory, started low, and could mint an id an existing record already referenced. That second one returns the wrong bytes rather than failing.

Closes the blob half of #642. Refs #644.

## The design: hard links, so the OS does the reference counting

Materialization reproduces each of the base's blob roots as the branch's own, hard-linking every file. The branch and base share the bytes, each holds its own directory entry, and the data is freed only when the last link goes.

That is what lets the branch keep **its own ID allocator**: `getNextFileId` seeds from the highest filename already in its directory, and the clone carries the base's filenames, so it naturally starts above every id that existed at checkpoint time. No shared allocator, no high-water mark, no gated deletion, no orphan-GC suspension — the filesystem already implements the refcounting those would have built by hand. #644 previously specified the shared-store design and has been rewritten to describe this one.

`copyTree` in `dataLayer/blobBackup.ts` already walked a blob tree hard-linking each file, with the cross-filesystem fallback; it is exported rather than reimplemented, with its marker wording parameterized so a branch does not describe itself as a backup. [It now also counts the fallback copies and ticks a progress callback per file](https://github.com/HarperFast/harper/pull/2426/changes?w=1#diff-ef6081f3b8edd93e7e5b3c8bb011fc94a74142673b26b7b93d2f31718d71fc98R151) — the first because a silent fallback turns a constant-time clone into an O(bytes) one and nothing else in the log would say so, the second because the walk runs inside a window other threads are blocked on.

The clone reads [the roots the base store actually resolves through](https://github.com/HarperFast/harper/pull/2426/changes?w=1#diff-cac868f688668988b75b365b484204e6a54f9fcbb121dab67b3358808b9dacf5R120), not the configured list — a `storage.blobPaths` edit against an already-open base would otherwise clone empty volumes — and iterates the *branch's* root list, so every entry its marker records exists even where the base has no counterpart.

## Why there is a completion marker

A branch's blob roots live on the `storage.blobPaths` volumes while its directory sits beside the base's storage root. The two cannot be published by one atomic rename, and neither parent is fsynced — so a power loss can replay the directory without its roots. Adopting on directory existence alone would then restart the allocator in an empty root and remint ids the branch's own checkpointed rows hold.

So materialization writes a marker inside staging, naming the roots it published, before the single rename that publishes the branch. The marker is what adoption trusts.

**The marker is also what the open branch resolves blobs through.** A row's `storageIndex` is a position in that list, so [the handle is pinned to the recorded roots](https://github.com/HarperFast/harper/pull/2426/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1604) rather than to current configuration — on the thread that published it and equally on [a thread that only adopted what a peer published](https://github.com/HarperFast/harper/pull/2426/changes?w=1#diff-cac868f688668988b75b365b484204e6a54f9fcbb121dab67b3358808b9dacf5R514). That is what makes an **appended** `storage.blobPaths` entry benign: [a recorded list that is an exact prefix of the configured one keeps every index where it was](https://github.com/HarperFast/harper/pull/2426/changes?w=1#diff-cac868f688668988b75b365b484204e6a54f9fcbb121dab67b3358808b9dacf5R286), so the branch loads and simply never uses the new volume, while a removed, reordered or replaced entry is still a hard refusal. Pinning is what keeps those detectable: an unpinned branch would write rows at an index its own marker never recorded.

**A directory that looks like a store but has no usable marker is refused, never deleted.** Materialization cannot produce that state, so anything in it came from somewhere else and is carrying data that exists nowhere else. Only leftovers that are not a store at all — including a directory holding nothing but a malformed marker — are cleared and rebuilt.

## The identity a branch answers to, and for how long

Cloning *removes and replaces* the blob root the branch's store identity resolves to, and `schemaRegex` permits digits, `_` and `.`, so a database may legally be named to collide with it. Two things follow.

**A branch owns a pair of names, not one.** The clone renames in from `<blobRoot>.staging`, and a database literally called `<storeName>.staging` resolves its own root to exactly that path. [Every check, reservation and release covers the pair](https://github.com/HarperFast/harper/pull/2426/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1362), and the on-disk question is [asked from both sides](https://github.com/HarperFast/harper/pull/2426/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1455) — `4_myapp__data.staging` is either the sibling of a branch of `data` or a branch of a database actually named `data.staging`, and whichever was created second must be refused.

**The reservation is held through every deletion, not merely checked before them.** `isBranchIdentity` is what stops another database taking the name, and once the branch directory is gone its on-disk half sees nothing either — so [cleanup retakes the pair on the same turn as the `close()` that released it](https://github.com/HarperFast/harper/pull/2426/changes?w=1#diff-cac868f688668988b75b365b484204e6a54f9fcbb121dab67b3358808b9dacf5R591) and holds it until the last blob root is removed. Removal deletes [only the roots the branch recorded](https://github.com/HarperFast/harper/pull/2426/changes?w=1#diff-cac868f688668988b75b365b484204e6a54f9fcbb121dab67b3358808b9dacf5R640), intersected with this node's configured volumes, so an appended volume that already holds a like-named directory restored from elsewhere is not collateral.

When a root **cannot** be removed, the name is not handed back: it goes into a [quarantine set](https://github.com/HarperFast/harper/pull/2426/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R1373) that refuses it to databases — which would mint fresh ids onto the surviving files — while still letting the branch itself retake it, since materializing replaces those roots wholesale and is the only route that clears the condition without an operator.

Database creation refuses the name too. `create_database` reaches `database()` through its own bridge, so `table()`'s guard never saw it; [`createSchemaStructure` now makes the same refusal](https://github.com/HarperFast/harper/pull/2426/changes?w=1#diff-07a296e386acc3155215669015450d25a7ee06830408a635c73991d1ce3619b2R89), after `transformReq` — because that is what settles whether `database` or `schema` names the database being created.

## What bounds the wait while a branch materializes

The claim's CREATING window now covers a hard-link clone of the base's entire blob tree, so any fixed budget is wrong for a large enough base: every waiting thread's application load fails while the winner is healthily copying, the winner then publishes, and the next deploy succeeds — a real outage reading as a flap. The winner reports each unit of work on a second slot of the shared claim word and [a waiter's budget restarts whenever it does](https://github.com/HarperFast/harper/pull/2426/changes?w=1#diff-cac868f688668988b75b365b484204e6a54f9fcbb121dab67b3358808b9dacf5R94), so what is bounded is a winner that has *stopped* rather than one that is merely slow. Because that has no ceiling by construction, a wait outliving its original budget says once that the winner is still reporting progress, rather than reading as a hang with no thread to point at.

## For the human reviewer

1. **This redefines an on-disk contract that another PR's tests depend on.** #2414's replay tests build branch directories by hand; a published branch now includes a marker, so those fixtures publish one too (via a shared helper rather than special-casing). Worth a look — it's the kind of cross-PR coupling that is easy to get subtly wrong.
2. **Marker substitution is a deliberate trade, and it is permanent for that blob.** Backup uses it because a snapshot is restored once; a branch is long-lived and scope-private, so a blob that was mid-write at clone time is unreadable in that branch forever. The alternative — failing branch creation whenever the base has an in-flight blob — makes branching unreliable on a busy base. Owner-settled, but it is the judgement most worth disagreeing with.
3. **Pinning is the owner-chosen answer to appended volumes**, over extending the marker so an existing branch takes the new capacity. The marker stays immutable proof of what was published atomically with the branch and no new crash window is added; the cost is that a long-lived branch cannot use capacity added after it, and filling its original volumes needs a rebuild. Choosing the other way is a storage migration, not a tweak.
4. **The `create_database` guard sits at the operation's validation layer, not inside `database()`/`ensureDB` where the directory is actually allocated.** Two awaits separate the check from `createSchema`, so a branch materializing under exactly the colliding name inside that window is still possible, and every other route into `database()` — restore, replication, component config — stays unguarded. `table()`'s copy has no such gap because its `database()` call is synchronous. Moving the guard down is contained but touches the load and drop paths the reserved-name comment deliberately excludes.
5. **A `storage.blobPaths` list that shrank is refused before checkpoint work.** If an open base still resolves through more roots than current configuration provides, materialization fails rather than publishing a branch whose higher `storageIndex` rows are unreadable.
6. **A published branch that later reads damaged or unmarked stays offline until an operator acts**, rather than being re-forked from the base. That is deliberate — it is carrying data that exists nowhere else — but on an unattended node it means the application never loads again.
7. **The checkpoint phase reports one progress tick, at the end.** The progress-extended budget therefore covers the clone but not the checkpoint, so a base whose checkpoint degrades to a byte copy (branch directory on a different volume from the base's storage root) and exceeds `CLAIM_TIMEOUT_MS + replayTimeBudgetMs` still flaps every waiter. Finer granularity needs a native callback or a timer.
8. **The quarantine set is per worker thread.** Other workers no longer see the branch directory, so the residual is unreclaimed disk on them rather than an id collision — but it is an operator action, and the log says so.
9. **Reservations and quarantine are worker-local while blob paths are filesystem-global.** The branch-side checks close the in-worker collision paths, but a direct allocator on another worker can still race materialization before anything on disk holds the name. Closing that gap requires moving reservation into the common database-allocation path or introducing a process-shared lease, which is a broader storage-lifecycle change than this review fix.

## Known and deliberate

- **The branch's allocator seeds from a directory scan of its own clone**, so any way that clone can come up short of the base's high-water mark lets it re-mint an id a checkpointed row still holds. Two exist: a blob root that exists but was emptied still reads as `complete` (a base with no blobs legitimately clones an empty root, so emptiness is not a signal), and the base can unlink its highest-id blob in the window between `createCheckpoint` and the clone's `readdir`, so the clone never sees it and no substitute marker is written for it. Detecting either properly needs the marker to record an allocator floor, which was judged more mechanism than the residual risk justifies. This is the one remaining path where a read returns wrong bytes rather than failing.
- **The marker is not fsynced**, and neither is the staging directory or either parent before the publishing rename, so a power loss can leave a store directory visible with a zero-length `.branch-complete` — read as `unmarked` and refused until an operator removes it.
- **`.branch-complete` lives inside the RocksDB directory** so one rename publishes both halves. That couples branch metadata to a directory RocksDB owns; relocating it later needs a migration for every published branch.
- **Branch removal on undeploy is not in this PR.** It follows in a second PR built on an explicit sequence (drop → restart → prove no references remain → remove), rather than being iterated onto this one.

## Verification

`test:unit:resources` **1905 passing, 0 failing**; `test:unit:dataLayer` 261; `test:unit:backup` 88. `tsc --project tsconfig.build.json`, `lint:required`, prettier clean.

The regression tests added for this round each name the behaviour they pin: the staging identity refused from both orderings, `create_database` refused through all three request spellings, the identity held until the blob roots are gone rather than until the directory is, removal touching only recorded roots, a branch that stays loadable when a volume is appended and keeps resolving through what it recorded, continued refusal when recorded roots are dropped or reordered, an empty `storage.blobPaths` refused at publish rather than bricking the next boot, an open base with more roots than current configuration refused before checkpoint work, a directory holding nothing but a malformed marker cleared and rebuilt, a stranded identity refused to databases while the branch retakes it, and the claim budget restarting on progress but running down without it.

Every new behaviour is mutation-checked — removing the clone, and treating markerless stores as disposable, each fail the suite. The inode assertion (`statSync().ino`, `nlink >= 2`) is there so a silent fallback from `link` to `copyFile` fails rather than passing quietly, which would double disk and make first load O(bytes).

One trap worth recording: an early version of these tests used an 18-byte blob, which is stored **inline in the record** and never becomes a file — so every assertion passed with the clone entirely removed. The tests now use a 64KB payload and assert `getFilePathForBlob()` is non-null, so they cannot silently stop testing anything if the externalization threshold moves.

Complexity: medium

<sub>Review-Coverage: authored=codex; ran=claude; blocked=gemini(auth); declined=cursor-grok,cursor-composer,domain; rounds=3 @ 156a25ce6456</sub>

<sub>Human-Review-Need: 4 @ 156a25ce6456</sub>

### DIFF
--- a/dataLayer/schema.ts
+++ b/dataLayer/schema.ts
@@ -79,16 +79,18 @@ export async function createSchemaStructure(schemaCreateObject: any) {
 		})
 	);
 	if (validation) throw new ClientError(validation.message);
-	// The same refusal schema authoring makes in `table()`: a branch resolves its blob root from its
-	// store identity, so a database under that name shares the root -- two allocators minting the same
-	// file paths, and `drop_database` deleting the branch's blobs (harper#644).
+
+	transformReq(schemaCreateObject);
+	// After the normalization, because that is what settles which of `database`/`schema` names the
+	// database being created. The same refusal schema authoring makes in `table()`: a branch resolves
+	// its blob root from its store identity, so a database under that name shares the root -- two
+	// allocators minting the same file paths, and `drop_database` deleting the branch's blobs
+	// (harper#644).
 	if (isBranchIdentity(schemaCreateObject.schema)) {
 		throw new ClientError(
 			`'${schemaCreateObject.schema}' is in use as a branch store identity and cannot be a database name`
 		);
 	}
-
-	transformReq(schemaCreateObject);
 
 	if (!(await schemaMetadataValidator.checkSchemaExists(schemaCreateObject.schema))) {
 		throw handleHDBError(
--- a/dataLayer/schema.ts
+++ b/dataLayer/schema.ts
@@ -14,7 +14,7 @@ import { handleHDBError, ClientError } from '../utility/errors/hdbError.ts';
 import { HDB_ERROR_MSGS, HTTP_STATUS_CODES } from '../utility/errors/commonErrors.ts';
 
 import { SchemaEventMsg } from '../server/threads/itc.js';
-import { getDatabases, dropTableMeta } from '../resources/databases.ts';
+import { getDatabases, dropTableMeta, isBranchIdentity } from '../resources/databases.ts';
 import { transformReq } from '../utility/common_utils.ts';
 import { server } from '../server/Server.ts';
 import { cleanupOrphans } from '../resources/blob.ts';
@@ -79,6 +79,14 @@ export async function createSchemaStructure(schemaCreateObject: any) {
 		})
 	);
 	if (validation) throw new ClientError(validation.message);
+	// The same refusal schema authoring makes in `table()`: a branch resolves its blob root from its
+	// store identity, so a database under that name shares the root -- two allocators minting the same
+	// file paths, and `drop_database` deleting the branch's blobs (harper#644).
+	if (isBranchIdentity(schemaCreateObject.schema)) {
+		throw new ClientError(
+			`'${schemaCreateObject.schema}' is in use as a branch store identity and cannot be a database name`
+		);
+	}
 
 	transformReq(schemaCreateObject);
 
--- a/resources/branchDatabase.ts
+++ b/resources/branchDatabase.ts
@@ -1,10 +1,12 @@
-import { existsSync } from 'node:fs';
-import { mkdir, rename, rm, rmdir } from 'node:fs/promises';
-import { dirname, relative } from 'node:path';
+import { existsSync, readFileSync, readdirSync } from 'node:fs';
+import { mkdir, rename, rm, rmdir, writeFile } from 'node:fs/promises';
+import { basename, dirname, join, relative } from 'node:path';
 import logger from '../utility/logging/harper_logger.ts';
 import { CONFIG_PARAMS } from '../utility/hdbTerms.ts';
 import { commonValidators } from '../validation/common_validators.ts';
 import * as env from '../utility/environment/environmentManager.ts';
+import { copyTree } from '../dataLayer/blobBackup.ts';
+import { getBlobPathsForDatabaseName } from './blob.ts';
 import type { RocksDatabase } from '@harperfast/rocksdb-js';
 import {
 	type BranchDatabase,
@@ -14,6 +16,8 @@ import {
 	hydrateBranchRelationships,
 	isReadOnlyMode,
 	openBranchDatabase,
+	releaseBranchIdentity,
+	reserveBranchIdentity,
 	resolveBranchPath,
 } from './databases.ts';
 import { replayLogs, replayTimeBudgetMs } from './replayLogs.ts';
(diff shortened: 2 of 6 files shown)


resources/branchDatabase.ts
```diff
@@ -244,9 +427,21 @@ async function removeBranchAt(branchPath: string): Promise<void> {
 	await rm(branchPath, { recursive: true, force: true }).catch((error) =>
 		logger.warn?.(`Could not remove branch directory ${branchPath}`, error)
 	);
+	// The clone gives a branch blob roots of its own, outside its directory, so removing only the
+	// directory strands them -- and because they are hard links, the bytes survive the base deleting
+	// its own copies. Only once the directory is actually gone: removing them while it survives (a
+	// held handle, permissions) would leave a branch that still looks adoptable but whose allocator
+	// restarts from an empty root and remints ids its own rows hold. Leaking them the other way round
+	// is recoverable; that is not. Derived from the path, so a branch that never opened is cleaned up.
+	if (!existsSync(branchPath)) await removeBlobRootsFor(branchStoreNameFor(branchPath));
```
cb1kenobi: **Branch cleanup releases its identity too early — Severity: high.** After the branch handle closes and `rm(branchPath)` succeeds, `isBranchIdentity()` considers the name available while `removeBlobRootsFor()` is still asynchronously deleting its blob paths. A concurrent database creation and blob write can reuse that identity and have its new files deleted by this cleanup. Keep every identity whose paths are being removed reserved until all root cleanup finishes, then release it.

resources/branchDatabase.ts
```diff
@@ -70,15 +74,160 @@ function claimStateFor(baseName: string, branchPath: string): BigInt64Array {
...
+	} catch {
+		return looksLikeAStore ? { state: 'unmarked', why: 'it has no readable completion marker' } : { state: 'debris' };
+	}
+	// `JSON.parse('null')` succeeds and yields null, so this cannot go straight to a property access.
+	if (recorded == null || !Array.isArray(recorded.blobRoots) || recorded.blobRoots.some((r) => typeof r !== 'string')) {
+		return { state: 'unmarked', why: 'its completion marker is malformed' };
+	}
+	const configured = getBlobPathsForDatabaseName(storeName);
+	// By index, not by set: `storageIndex` on a row is a position in this list, so a reordered
+	// `storage.blobPaths` would silently resolve every row through the wrong root.
```
cb1kenobi: **Appending a blob volume makes existing branches unloadable — Severity: medium.** Requiring equal array lengths marks every existing branch as damaged when `storage.blobPaths` gains an appended entry, despite all recorded roots retaining their indices; `openOrCreate()` then refuses the branch indefinitely. For a matching recorded prefix, create the new empty branch roots and extend the completion marker before opening, while continuing to reject removals or changes to recorded indices.

cb1kenobi (commented):
Barbarian reviewed 6f5a994b and found 2 blocking issues. Branch cleanup can race with database creation and delete the new database's blobs. Adding a blob path also makes existing branches appear damaged; retain cleanup reservations and safely extend branch markers for appended roots.

resources/branchDatabase.ts
kriszyp: Good catch — fixed in `531c91a`. Cleanup now retakes the identity pair on the same turn as the `close()` that released it, holds it through the directory removal *and* every blob-root removal, and releases it in a `finally`. The dormant path (never opened on this worker) reserves the name outright instead, and if that reservation fails it leaves the blob roots alone rather than deleting paths something else now answers to.

A root that could not be removed does not hand the name back either: it goes into a per-worker quarantine set that refuses the name to databases while still letting the branch itself retake it (materializing replaces those roots wholesale, which is what clears the condition).

Regression test: `keeps the identity reserved until the blob roots are gone, not just the directory`.

— Claude Opus 5

resources/branchDatabase.ts
kriszyp: Fixed, though not quite the way you suggested — worth a second look since the difference is a design choice.

`readBranchState` now accepts a recorded list that is an **exact prefix** of the configured one, so an appended `storage.blobPaths` entry no longer reads as damage; removals, reordering and replacement stay a hard refusal. But rather than creating the new empty roots and extending the marker, the branch is **pinned to what its marker recorded**: `openBranchDatabase` takes the recorded roots and seeds `databasePaths` for the branch's root store, so every blob read, write, allocator seed and removal for that handle resolves through the recorded list, on the adopting thread as well as the publishing one.

Kris made that call between the two options. Pinning keeps the marker as immutable proof of what was published atomically with the branch, preserves every positional `storageIndex` exactly, and adds no crash window; extending the marker in place would mutate durable branch state on a load path and needs a temp-write/fsync/rename protocol plus a refusal path for a pre-existing non-empty suffix root. The cost is real and is now stated in the PR body: an existing branch never uses appended capacity until it is rebuilt from the base, and adoption logs once when that is the case.

Regression tests: `stays loadable when a volume is appended, and keeps resolving through the roots it recorded`, `pins a branch adopted on the waiter path, not only one this thread published`, plus the still-refused dropped/reordered cases.

— Claude Opus 5

dataLayer/schema.ts
```diff
@@ -81,6 +81,16 @@ export async function createSchemaStructure(schemaCreateObject: any) {
 	if (validation) throw new ClientError(validation.message);
 
 	transformReq(schemaCreateObject);
+	// After the normalization, because that is what settles which of `database`/`schema` names the
+	// database being created. The same refusal schema authoring makes in `table()`: a branch resolves
+	// its blob root from its store identity, so a database under that name shares the root -- two
+	// allocators minting the same file paths, and `drop_database` deleting the branch's blobs
+	// (harper#644).
+	if (isBranchIdentity(schemaCreateObject.schema)) {
```
cb1kenobi, line 89: **Identity check is non-atomic — Severity: critical.** This check precedes awaited existence and creation work, so branch materialization can claim the same identity before the database is allocated; direct `database()` routes bypass the check entirely. The database and branch can consequently share blob paths, collide on file IDs, or delete each other’s blobs. Enforce and atomically reserve the name in the common database-allocation path until creation finishes.

—
_Generated by Barber AI_

resources/branchDatabase.ts
```diff
@@ -62,23 +76,266 @@ const branchesByPath = new Map<string, Promise<OpenBranch>>();
...
+ *
+ * A blob that is mid-write or already reclaimed is substituted with a marker rather than linked. A
+ * partially-written blob would otherwise be linked as a whole one, and worse, the abort that follows
+ * stamps PENDING onto that inode IN PLACE -- through the branch's link as well, since it is the same
+ * inode. The marker makes that blob fail loudly in the branch instead.
+ */
+async function cloneBlobRoots(baseName: string, branchRoots: string[], progress: () => void): Promise<void> {
+	// What the base's own rows resolve through, which is not the configured list if `storage.blobPaths`
+	// changed while its store was open -- reading the configured volumes would clone an empty tree.
+	const baseRoots = getRootBlobPathsForDB(database({ database: baseName, table: undefined }) as never);
```
cb1kenobi: **Publishes an unreadable branch after volume shrinkage — Severity: high.** When the open base resolves more roots than current configuration provides, this code only warns and then publishes a complete marker containing the shorter list. Checkpointed rows using a higher `storageIndex` consequently resolve through no root and cannot be read. Fail materialization here, or clone every root used by the base, instead of publishing the incomplete branch.

—
_Generated by Barber AI_

resources/branchDatabase.ts
```diff
@@ -62,23 +76,266 @@ const branchesByPath = new Map<string, Promise<OpenBranch>>();
...
+				`${baseRoots.length}; rows referencing the base's higher-numbered volumes will not be readable ` +
+				`in it. storage.blobPaths shrank while the base was open`
+		);
+	}
+	// A row's `storageIndex` is a position in the branch's own list, and its marker records every entry.
+	for (let index = 0; index < branchRoots.length; index++) {
+		const staging = `${branchRoots[index]}.staging`;
+		await rm(staging, { recursive: true, force: true });
+		// Created up front rather than left to the walk: a base with no blobs at all copies nothing, and
+		// the branch still needs its own (empty) root to exist so its allocator starts where the base's
```
cb1kenobi, line 133: **Checkpointed blob IDs can be reminted — Severity: critical.** The checkpoint is created before this directory walk. If the base unlinks its highest-ID blob between those operations, the checkpoint can still reference that ID while the clone contains neither the file nor a marker; the branch allocator then scans a lower maximum and can reuse the ID, making the old row return unrelated bytes. Persist a checkpoint-time allocator floor in the branch metadata and seed every branch allocator above it.

—
_Generated by Barber AI_

dataLayer/schema.ts
kriszyp: Confirmed: the branch-side reservation and normalized `create_database` guard do not make allocation atomic across every `database()` route. Moving the lease into the common allocator would touch restore, replication, load, and drop lifecycles, so I have not represented that broader design change as mechanical. The residual check-then-act/cross-worker gap is now explicit in “For the human reviewer” items 4 and 9 in the PR body.

resources/branchDatabase.ts
kriszyp: Fixed in `b86c2c20f`. `materializeBranch` now compares the open base’s resolved roots with the current branch-root list and throws before creating/removing staging paths or taking the RocksDB checkpoint when configuration has shrunk. The new `volbaseShrink` regression exercises a base opened on two roots under a one-root configuration and verifies that no branch is published.

resources/branchDatabase.ts
kriszyp: Confirmed and retained as a known design trade rather than silently claiming it is impossible. Closing this window correctly requires persisting a checkpoint-time allocator floor and seeding branch allocation from it, which extends the durable marker format. The PR body’s “Known and deliberate” section now calls out this exact reclaim window and its wrong-bytes consequence; no code fix for that storage-format decision is claimed here.

Issue #2182, opened 2026-08-15
HNSW: graph-size resolution is a lifetime id high-water mark — churn-heavy tables permanently over-provision build/search ef


kriszyp:
Follow-up to #2180 / #2181, from the pre-push review's main finding.

Both HNSW auto-scales (search `ef`, and construction `ef` as of #2181) resolve graph size from the shared id counter — a **lifetime high-water mark of allocated node ids**. Deletes never decrement it, so a churn-heavy table (TTL eviction, delete-and-reinsert ingest) reads permanently high. On the search side that means a generous-but-capped `ef`. On the build side it means a table holding, say, 30K live vectors that has allocated 5M ids over its life resolves 5M → `efConstruction` pinned at the cap → every insert runs a 512-candidate connection search on a 30K-node graph, ~5x the base build CPU, inside the write transaction, forever — with zero recall benefit. Documented as an accepted caveat in DESIGN.md (bounded at `AUTO_EFC_MAX`, wastes CPU, never hurts recall), but the right fix is a count that tracks live nodes.

Sketch: a second slot in the shared `next-id` buffer maintained as a live count (increment on insert, decrement on remove). The hard parts are (a) re-seeding it O(1) after restart — the largest-id reverse seek that re-seeds the id counter has no live-count analogue, and an exact count is a full key scan; (b) mixed-version workers sharing the buffer during a rolling upgrade. A persisted periodically-checkpointed count (exact-on-rebuild, drift-bounded between checkpoints) may be the pragmatic middle.

Also worth considering alongside (from the same review): a "pinned build, auto search" configuration does not currently exist — setting `efConstruction` pins the search-side default too, so the escape hatch for the churn cost trades away search-ef auto-scaling.

Generated by Claude Fable 5.

Issue #895, opened 2026-06-01
HNSW: insertion degrades O(N) with graph size, blocks event loop, disrupts replication


kriszyp:
## Summary

HNSW insertion throughput degrades roughly O(N) as the graph grows — the opposite of the expected O(log N). On a free-tier cluster with a 10k-vector table, ingest slows from **~178 vectors/s** to **~19 vectors/s** as the graph fills, and each insert takes **~53 ms** (should be sub-millisecond). The event loop produces 11 "JavaScript execution too long" warnings per run. On single-CPU machines this causes write-queue stalls (HTTP 503) and nodes going offline.

The slowdown also disrupts replication: the event-loop blockage during bulk inserts triggers TCP keepalive timeouts, killing in-flight clone connections and forcing clone restarts from zero. On large tables the cluster never converges.

## Root cause

HNSW graph traversal visits O(N) nodes per operation (during layer-0 search). On every node visit, Harper decodes the stored vector from msgpackr format — a full round-trip over a boxed `Array` of 768 floats. Profiling shows this decode accounts for **54.5% of search CPU** and **70.8% of insert CPU**; the actual `computeDistance` call is only 3–4%. The per-vector decode cost makes traversal O(N·decode) even when the graph's structural complexity is O(log N).

## Reproduction

1. Create a table with an HNSW index on a 768-dim vector attribute.
2. Bulk-insert 10k vectors while monitoring throughput.
3. Observe: throughput decays from ~180/s at the start to <20/s by 10k; Harper logs "JavaScript execution too long".

Repro script and detailed findings from @chris-nelson attached to [#development Slack thread](https://harperdb.slack.com/archives/C3Z2T1QAZ/p1779806565475909).

## Fix in progress

- **[harper#894](https://github.com/HarperFast/harper/pull/894)** — int8 scalar quantization for HNSW navigational graph. Eliminates per-element float32 decode by storing graph nodes as compact int8 bin + per-vector scale. Benchmarked: **~4.9× search throughput**, p99 latency **9s → 0.5s**, ~3.6× update, ~3× smaller index, recall@10 within noise of float. CI ✅ all passing.
- **[harper-pro#248](https://github.com/HarperFast/harper-pro/pull/248)** — replication keepalive by byte activity + decode-loop yield budget. Makes replication resilient to event-loop bursts from large HNSW inserts.
[...]

Issue #711, opened 2026-05-22
Custom indexing support (HNSW, async)


kriszyp:
## Request

Build support for custom user-defined indexing, enabling use cases like HNSW vector indexing and async indexing workflows where the index is maintained outside the standard synchronous write path.

🤖 Filed by Claude on behalf of Kris.

Issue #693, opened 2026-05-21
Improve HNSW vector index niceness (async searchLayer)


kriszyp:
### 1. Async `searchLayer` to yield event loop

`searchLayer` is a synchronous traversal that can run long enough to starve other I/O. Converting it to an async function that occasionally `await`s (e.g., every N iterations) would improve concurrency characteristics without meaningfully impacting throughput.

PR #2430, opened 2026-09-01
Native HNSW index: file-primary mmap graph as a backend on the shared derived-index runtime

kriszyp:
A **file-primary** native HNSW vector index, as a backend on the shared derived-index runtime from #2567. The memory-mapped `.hnsw` file (`@harperfast/hnsw` 0.2.1, exact-pinned optional dependency) is the only home of graph nodes, adjacency, the entry point and the id allocator; RocksDB keeps the primary records, the `pk ↔ nodeId` mappings and one durable cursor vector. This PR is stacked on #2567 and shows only the HNSW part; it is that runtime's first consumer.

The durable design — file format, concurrency, durability, search path, how the index sits on the runtime, what `nativePlane: true` requires and does not promise — is the **§ Native HNSW plane** section of the root `DESIGN.md`. This description holds the transient part: decisions, measurements, verification.

## Why

Measured at 5M nodes / ef 512, ~85% of a JS search visit is object bookkeeping and a RocksDB `Get` per visit (~1–2 µs warm) dwarfs the ~50 ns SIMD distance it feeds:

| 1M × 768-d int8, ef 512 | JS (main) | native |
|---|---|---|
| search p50 | 7.2 ms | **0.75 ms** |
| per-visit cost | 4.34 µs | 0.33 µs |
| recall@10 (set) | 0.997 | 0.999 |

## What changed in this revision

The branch's own delivery runtime (`resources/DerivedIndexBackend.ts`, 486 lines) is deleted. In its place:

- [`resources/indexes/hnswDerivedIndex.ts`](https://github.com/HarperFast/harper/blob/kris/hnsw-native-plane/resources/indexes/hnswDerivedIndex.ts) — `HnswDerivedIndexBackend`: `deliver()` queues and returns accepted, an applier drains in 5 ms `setImmediate` slices, `plane.flushAsync()` is the barrier, then the pending mappings are published, then the batch's `through` vector is written as the one durable cursor (`Symbol.for('derived-index-cursor')`). Application pauses while a barrier is in flight so the barrier publishes exactly what it covers. `reset(epoch)` removes the cursor before the file and the mappings. `attachDerivedIndexes(Table)` registers a table's post-commit indexes with one runtime per database, on every worker.
- `HierarchicalNavigableSmallWorld.ts` — the commit path only validates (a malformed vector is still the client's 400); nothing is staged on the transaction. Search readiness is the runtime's shared record on every worker (`planeSearchReady`), not `isIndexing`; a failed search detaches and asks the owner for a rebuild instead of unlinking a file a peer may already have replaced. A vector the plane cannot hold at apply time is skipped and counted, never a rebuild.
- Writer backpressure is the runtime's lag policy: `maxLagMilliseconds` (30 s default on a `nativePlane` attribute, settable) → retryable 503 `DERIVED_INDEX_LAGGING` on local user writes to the table; canonical-source applies, replay and replication notifications are never shed.
- `RocksTransactionLogStore.aftercommit` no longer carries staged targets.

Six open items from the earlier design record are closed by the shared runtime: shared cross-worker readiness, a new origin's first write forcing a rebuild, per-key coalescing (the ingest bench's repeated-key shape now measures 265 native applies for 1,000 commits over 50 keys; it was 1,000), undecodable audit headers, and the whole-retention-window replay after a rebuild. An interior corrupt frame inside the committed prefix still fails the attempt closed — a log that cannot be read to its tail cannot be replayed from any anchor.

## What `nativePlane: true` requires

Unchanged: explicit `audit: true` on the table (the transaction log is the recovery source; inheriting the global setting is rejected), RocksDB, `M=16` / `efConstruction=200` / int8 cosine, `@harperfast/hnsw` loadable (absence = 503, not degraded). Not promised: a single total order across concurrent CRDT/source-resolution arrivals, byte-identical graphs across nodes or rebuilds, in-place format upgrades.

## Decisions

- **File-primary, not dual-write.** The phase-1 prototype mirrored the RocksDB graph into the file and cut search over; it never shipped, so there is no deployed population to protect, and keeping a second graph would pay its write and storage cost forever. Rejected alternatives: keep the CF graph as authority (duplicate writes, the cost this exists to remove); ship post-commit delivery first and cut over later (another temporary format, a migration, no compatibility evidence to gain); synchronous native mutation inside the record transaction (an aborted transaction can still publish an mmap write; recovery needs the log protocol anyway).
- **Transaction-log delivery, not an outbox** — raised by planning review as a better alternative and overruled by the repo owner in favour of #2489's single protocol; the reasoning is in #2567.
- **One runtime.** This branch's own `DerivedIndexBackend.ts` implemented the same protocol; the measured backend share of the write-and-index wall (95–97%) put the per-transaction bookkeeping the two runtimes differed on inside the remaining 3–5%, so it could not justify keeping both.
- **Committed-tail rebuild anchor.** The first convergence plan proposed a tighter anchor from staged/uncommitted positions and planning review disqualified it (an uncommitted anchor skips its own transaction; an aborted one may never exist as a boundary). Reading rocksdb-js retired the line of work: `commitFinished()` advances `lastCommittedPosition` to the earliest still-uncommitted write, so the committed tail is already a contiguous-prefix boundary and needs no storage-layer change.
- **Degree cap 128, sparse reservation at create, Apache-2.0 standalone package** (Kris, 2026-08-31), recorded in `DESIGN.md`.
- **Admission control is a lag budget, not a queue depth.** The old runtime rejected vector-changing writes at 65,536 pending keys / 262,144 tickets; the shared runtime's `maxLagMilliseconds` sheds the same writes on time behind, which is the quantity retention actually bounds.

## Measurements

Per-visit cost at 5M nodes / ef 512, 768-d int8: 4.34 µs per warm JS visit, of which the int8 cosine is 0.43 µs and a cold msgpackr node decode 5.57 µs; native per-visit 0.33 µs. 1M × 768-d: search p50 7.2 → 0.75 ms, recall@10 0.997 → 0.999. Package in isolation (N = 10,000): insert 208 µs (128-d) / 315 µs (384-d) / 835 µs (1536-d). Ingest bench on this branch (384-d, graph ≈3,000): foreground put 0.090 ms (11,078/s), `applyDerivedValue` 0.365 ms/call, `flushDerived` 8.96 ms/call, event loop max 11.8 ms while draining; repeated-key shape 265 applies for 1,000 commits over 50 keys. Rebuild insertion floor: 4,684/s at 100k against a 1,000/s CI gate; 1,242/s at 1M from the crate anchor.

## Verification

- `unitTests/resources/vectorIndexPlane.test.js` — **22 passing** against the real native package: mappings-only CF; deterministic recall; aborted transaction publishes nothing; unrelated field change schedules no insert; predicates and full-stack rescoring; wrong-dimension write/query are 400s; f32-range rejection; async result contract; **two workers into one native file; a committed write replayed after its worker dies before delivery; whole-table reload marker rebuilds (live and on a replacement worker); recovery across `resetDatabases`; a cursor outside retention rebuilds from primary records; an empty index answers no results; a populated index that lost its file answers 503; drop removes the file.**
- Full `test:unit:resources` on this branch: 2480 passing, 32 pending, 0 failing.
- `hnswDerivedIngest.bench.js` runs; the `hnsw-plane` CI job still requires the native binding to load rather than accepting a skip.
- `npm run build`, `tsc --noEmit`, `lint:required`, prettier: clean.

Restacking exposed one runtime defect, fixed in #2567: a log that has never written a file reports `oldestSequenceNumber: 0`, which the runtime read as a retention gap on every brand-new database.

Refs #2489, #693, #711, #895, #2182

### DIFF
--- a/resources/indexes/HierarchicalNavigableSmallWorld.ts
+++ b/resources/indexes/HierarchicalNavigableSmallWorld.ts
@@ -1,11 +1,31 @@
+import { closeSync, existsSync, openSync, rmSync, statSync, unlinkSync } from 'node:fs';
 import { cosineDistance, euclideanDistance, dotProductDistance } from './vector.ts';
 import { FLOAT32_OPTIONS } from 'msgpackr';
 import { loggerWithTag } from '../../utility/logging/logger.ts';
-import { ClientError } from '../../utility/errors/hdbError.ts';
+import { ClientError, ServerError } from '../../utility/errors/hdbError.ts';
 import type { Id } from '../../resources/ResourceInterface.ts';
+import type { DerivedIndexReadiness } from '../derivedIndexRuntime.ts';
 import { SKIP } from '@harperfast/extended-iterable';
+import { RocksDatabase } from '@harperfast/rocksdb-js';
+import { createHash } from 'node:crypto';
+import {
+	getPlaneBinding,
+	invalidatePlaneFile as invalidateHnswPlaneFile,
+	planeFilePathFor,
+	planeStalePathFor,
+	PLANE_NO_ID,
+	type HnswPlane,
+} from './hnswPlaneBinding.ts';
 
 const logger = loggerWithTag('HNSW');
+
+/** Element-wise equality of the array-like projections a derived index stores; anything else that failed `===` counts as changed. */
+export function derivedValuesEqual(a: any, b: any): boolean {
+	if (a === b) return true;
+	if (a == null || b == null || typeof a.length !== 'number' || a.length !== b.length) return false;
+	for (let i = 0; i < a.length; i++) if (a[i] !== b[i]) return false;
+	return true;
+}
 
 // int8 scalar quantization of stored graph nodes is ON by default. Each node holds the
 // vector as a compact int8 `bin` plus a per-vector `scale`, roughly a 5x size reduction
@@ -104,6 +124,22 @@ function autoScaleEfConstruction(nodeCount: number): number {
 // ef moves with the square root of the count and is capped, so a slightly stale size is immaterial;
 // this only has to be short enough that a table growing from empty picks up a larger ef promptly.
 const NODE_COUNT_TTL = 10_000;
+
+// Native traversal-plane geometry (DESIGN.md § Native HNSW plane). The layer-0 cap is derived from
+// M/optimizeRouting at creation to cover the JS graph's effective cap (grace overshoot above it
+// truncates by distance); a configuration deriving past this ceiling is refused as ineligible
+// rather than silently truncated. maxNodes is a fixed sparse reservation — pages materialize on
+// write — and ids at or past it are rejected by the crate, which disables the plane.
+const PLANE_LAYER0_CAP_MAX = 1024;
+const PLANE_MAX_NODES = 1 << 24;
+// An existing plane file that cannot be opened is normally another worker mid-create (retry);
+// past this age it is a crashed create and is deleted so the audit-backed runtime can rebuild it.
+const PLANE_STALE_CREATE_MS = 60_000;
+// Retry cadence while another worker holds the create; its header lands shortly after exclusive open.
+const PLANE_ATTACH_RETRY_MS = 250;
+// Marks an error thrown by an app-supplied filter during a plane search: the caller re-raises
+// it as an ordinary query failure instead of disabling the (healthy) plane.
+const NOT_A_PLANE_FAILURE = Symbol('notAPlaneFailure');
 
 class MinHeap {
 	private data: Candidate[] = [];
@@ -235,6 +271,23 @@ export class HierarchicalNavigableSmallWorld {
 	private convertedNodes = new WeakMap<object, any>();
 	private nodeCount = 0;
 	private nodeCountAt = 0;
+	// Native file-primary index. The RocksDB index store holds identity mappings and replay cursors;
+	// graph nodes and adjacency exist only in this file.
+	// undefined = not yet attached (may retry), null = unavailable or disabled for this process.
+	private plane: HnswPlane | null | undefined;
+	private planeEligible = false;
+	private planeRetryAt = 0;
+	private planeDisabledLogged = false;
+	private filePrimary = false;
+	private nativePlaneMaxNodes = PLANE_MAX_NODES;
+	// Installed by attachDerivedIndexes on every worker: shared readiness of the index and the way
+	// to ask its owner for a rebuild. Only the owning worker's runtime ever destroys native state.
+	private derivedHost?: { readiness: () => DerivedIndexReadiness; requestRebuild: () => boolean };
+	private pendingDerivedMappings = new Map<
+		Id,
+		{ id?: number; signature?: string; version?: number; pending?: boolean }
+	>();
+	postCommit?: true;
 	constructor(indexStore: any, options: any) {
 		this.indexStore = indexStore;
 		if (indexStore) {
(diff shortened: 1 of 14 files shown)


native/hnsw-plane/src/graph.rs
```diff
@@ -0,0 +1,650 @@
...
+            idx => idx,
+        };
+        let upper_idx = if upper_levels.is_empty() {
+            existing // keep an existing entry bound (level never shrinks in practice)
+        } else if existing != NO_UPPER {
+            self.rewrite_upper(existing, upper_levels)?;
+            existing
+        } else {
+            self.write_upper(upper_levels)?
+        };
```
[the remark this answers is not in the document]
kriszyp: Fixed, though not with the suggested patch — `free_upper` there runs while the slot still
points at the entry, so a reader that already read the index can follow it after it has been
handed to another node. `remove_node` avoids that by clearing the slot and emptying the entry
under its own lock first.

The fix ended up narrower than "free it". Returning the entry to the shared freelist at all is
not atomic with publishing the slot, so a concurrent mirror that read the index first could
republish a slot pointing at an entry already reallocated — trading a bounded retention for
cross-node corruption. The entry is emptied in place instead: the stale hierarchy stops being
readable (which is the part that produces wrong traversals) and it stays bound to the id,
costing at most one idle entry per id.

Two related leaks went with it: `write_node_raw` read the bound index through
`read_consistent`, whose fallback is `NO_UPPER`, so a peer holding the slot lock past the 20 ms
stale window made a bound node look unbound and minted a duplicate — it reads under the slot
write lock now; and both write paths now free a freshly allocated entry when publication
wedges. All covered by tests that fail on the parent commits.

Still open, deliberately: `slot_sanitizer` discards a bound index the same way on dead-owner
takeover, and two writers first-mirroring the same level>=1 id can both allocate. Both are in
the PR description.

— Claude Opus

cb1kenobi (commented):
Barbarian reviewed fd5c8845 and found 1 blocking issue. Disabling nativePlane can leave an undeletable plane trusted on a later re-enable. Tombstone the file when deletion fails so it must be rebuilt before serving searches.

resources/indexes/HierarchicalNavigableSmallWorld.ts
```diff
@@ -265,7 +346,500 @@ export class HierarchicalNavigableSmallWorld {
...
+	 * live index's plane.
+	 */
+	cleanupDisabledPlane(): void {
+		if (this.planeEligible) return;
+		const filePath = this.planeFilePath();
+		if (!filePath) return;
+		try {
+			unlinkSync(filePath);
+			logger.info?.('deleted the HNSW plane file of an index no longer using nativePlane');
+		} catch (error: any) {
```
cb1kenobi, line 365: Stale plane survives flag disable

Severity: High (blocking). On Windows, or while another process still maps the file, `unlinkSync` can fail; this catch only logs, so mutations made while `nativePlane` is off are never mirrored and a later re-enable can reopen the old nonzero-watermark plane as ready, returning incomplete results. On every non-`ENOENT` failure, create the `.stale` tombstone as `resetDerivedStorage` already does so the file cannot be adopted and is rebuilt.

cb1kenobi (commented):
Barbarian reviewed 7661fd1b and found no blocking issues. The new commit restores MADV_RANDOM at both mmap creation and reopen sites. No new blocking findings were identified; previously raised findings were not repeated.

—
_Generated by Barber AI_

native/hnsw-plane/src/graph.rs
```diff
@@ -0,0 +1,741 @@
...
+    /// and leave a graph lying between its samples invisible permanently, not for one search.
+    ///
+    /// Best-level rather than first-live: a level-0 entry degrades every later search to a
+    /// layer-0-only beam.
+    pub(crate) fn probe_for_entry(&self, limit: u32, skip: u32) -> Option<(u32, u8)> {
+        let hw = self.file.id_high_water().min(self.file.max_nodes) as u32;
+        if hw == 0 || limit == 0 {
+            return None;
+        }
+        let stride = (hw / limit).max(1);
```
cb1kenobi: Repair probe permanently skips low IDs — Severity: High (blocking). When `hw >= limit` and `hw` is not divisible by `limit`, floor division makes `stride * limit < hw`; every rotated probe stops before the lowest `hw % limit` slots. If the entry and hint die while the remaining live nodes occupy that low prefix, every search returns empty indefinitely. Use ceiling division (`hw.div_ceil(limit)`) so a full rotation covers every ID.

—
_Generated by Barber AI_

cb1kenobi (commented):
Reviewed 2fb3a3b7 and found no blocking issues. No confirmed blocking findings remain in the new changes. The repair-probe coverage and predicate-drain fixes are sound.

—
_Generated by Barber AI_

cb1kenobi (commented):
Reviewed 8dd6c16b and found no blocking issues. The new commits introduce no confirmed blocking issues on changed lines. Previously raised findings were not repeated.

—
_Generated by Barber AI_

cb1kenobi:
Reviewed a0f735d5 and found no blocking issues. No new blocking issues were confirmed at this commit. Existing findings were not repeated, and the new commits are rebase and formatting changes.

—
_Generated by Barber AI_

PR #2436, opened 2026-09-01
Make MCP resource subscriptions reliable across HTTP workers

### LINKED ISSUE #1751: MCP resources/subscribe fails to establish 88% of the time under default multi-worker deployment
## Summary

MCP `resources/subscribe` (the native MCP server's live resource-change-notification mechanism) fails to establish **88% of the time (14/16 trials)** under Harper's default multi-worker deployment shape, whenever a client's `initialize`/GET-SSE-stream/`subscribe` calls aren't sticky-routed to the same worker.

`sessionRegistry.ts` documents this as a known, deliberate v1 limitation:

```ts
/**
 * Per-worker only. A session's GET stream is bound to the worker that
 * accepted the GET; cross-worker fan-out isn't attempted in v1.
 */
```

That framing reads like a minor deferred nice-to-have. In practice, under Harper's actual **default** multi-worker topology (no external load-balancer session affinity documented or configured out of the box), a `resources/subscribe` call landing on a worker OTHER than the one holding the client's open SSE GET stream hard-rejects with:

```
-32602 "open the GET SSE stream before subscribing to resources"
```

— even though the stream is genuinely open, just on a sibling worker. Measured empirically at an 88% failure rate for session establishment across 16 trials with `threads.count: 4`. This makes `resources/subscribe` effectively unusable by default for anyone not manually pinning client sessions to a single worker.

## What's clean, for contrast

Once a subscription IS successfully established, cross-worker notification **delivery** is flawless: 20/20 sequential cross-worker writes notified 1:1 (~2ms avg latency), 25/25 concurrent-burst writes to distinct records → exactly 25 notifications, zero loss/duplication. The gap is specifically in *establishing* a subscription across workers, not in delivering to an established one.

## Pattern note

Same shape as two other independently-discovered issues this session: `#1610`'s MCP per-client rate limiter (in-memory per worker, ~4× weaker than configured under the default multi-worker topology) and `#1736`'s `registerOperation()` (per-worker `OPERATION_FUNCTION_MAP`, unreachable across workers). Three separate instances of MCP-layer state assumed process-wide but actually living per-worker, each discovered independently. Might be worth a structural fix (a shared cross-worker session/registry seam) rather than three point patches — flagging for visibility, not proposing scope here.

## Suggested direction

Either implement the cross-worker fan-out the code comment already anticipates ("isn't attempted in v1"), or — if that's out of scope near-term — surface guidance that `resources/subscribe` requires session-affinity routing in front of a multi-worker deployment, since nothing in the default setup currently provides or documents that requirement.

## Repro

Local repro (not yet in the tracked suite): `integrationTests/qa-scratch/qa526-mcp-subscribe-allowread-multiworker.test.ts` — the multi-worker establishment-failure rate is measured/logged there rather than hard-asserted; raw trial log available on request.

kylebernhardy:
MCP resource subscriptions now follow the session's active GET-SSE stream across HTTP workers through a [fenced owner claim](https://github.com/HarperFast/harper/pull/2436/changes?w=1#diff-c5ddd1396c67f379c806e4d4a1a541c5a397a3d0bae065fa7e9f7c5174a6f2e5R142), so `resources/subscribe` and `resources/unsubscribe` no longer depend on both requests landing on the same worker. The [routing path](https://github.com/HarperFast/harper/pull/2436/changes?w=1#diff-c5ddd1396c67f379c806e4d4a1a541c5a397a3d0bae065fa7e9f7c5174a6f2e5R288) keeps the full local principal, projects only the fields needed across ITC, and rejects stale ownership commands. [Storage-native serialization](https://github.com/HarperFast/harper/pull/2436/changes?w=1#diff-5ce7e54ce486dd20d10ae91fce54934359f223bfd40ea9172ad8b1961fbea800R177) prevents concurrent subscription changes from losing durable updates.

Fixes #1751.

## For the human reviewer

1. The active GET worker and a random stream token are persisted on the non-replicating MCP session row when the [SSE stream registers its owner](https://github.com/HarperFast/harper/pull/2436/changes?w=1#diff-e5236f0b6a86eebec569f8afdf37573234af03cef53c3817950a0fdf12a14497R409). POST workers route directly to that owner, and the live token rejects stale commands. When ITC is unavailable, the owner claim is cleared and only the local unwired worker may use its registry entry.
2. Routed operations have an ambiguous outcome at the 30-second timeout boundary, so the [JSON-RPC response asks the client to retry](https://github.com/HarperFast/harper/pull/2436/changes?w=1#diff-e5236f0b6a86eebec569f8afdf37573234af03cef53c3817950a0fdf12a14497R965). Exactly-once operation IDs would add durable protocol state for a rare outcome and can be added later without changing successful requests.
3. Durable subscription mutations use Harper's native per-record-store lock across workers rather than a persisted CAS/version field. The acquire is bounded and failures become retryable JSON-RPC errors. This keeps the session schema and protocol unchanged across RocksDB and LMDB.

## Verification

- `npm run build` — passed.
- Focused MCP unit suite (`session`, `subscriptionRouting`, `subscriptions`, `transport`) — 129 passing.
- Changed-file lint, Prettier, and `git diff --check` — passed.
- Final independent Claude review — no remaining code findings.
- Full repository lint currently reports 14 unrelated pre-existing warnings outside the changed files.
- The local MCP integration harness cannot start because this macOS host lacks the required `127.0.0.2`–`127.0.0.32` loopback aliases. The [cross-worker assertion](https://github.com/HarperFast/harper/pull/2436/changes?w=1#diff-c25120786d28a66e0f070f72886b8d52437371ce190dda660dc4eec2f2a01234R474) passed in the Node integration shard and remains fail-closed there. Bun skips this socket-placement proof because all pinned probes are assigned to one HTTP worker.

Complexity: complicated

<sub>Review-Coverage: authored=codex; ran=claude; declined=gemini,cursor-grok,cursor-composer,domain; rounds=2 @ 5c60fea0e5e9</sub>

<sub>Human-Review-Need: 4 @ 5c60fea0e5e9</sub>

### DIFF
--- a/components/mcp/session.ts
+++ b/components/mcp/session.ts
@@ -148,7 +148,15 @@ export async function createSession({
  */
 export async function loadSession(id: string): Promise<McpSessionRecord | null> {
 	const record = (await (getTable() as any).get(id)) as McpSessionRecord | undefined | null;
-	if (!record) return null;
+	if (
+		!record ||
+		typeof record.protocolVersion !== 'string' ||
+		typeof record.initialized !== 'boolean' ||
+		typeof record.user !== 'string' ||
+		typeof record.createdAt !== 'number' ||
+		typeof record.lastActivity !== 'number'
+	)
+		return null;
 	return record;
 }
 
--- a/components/mcp/session.ts
+++ b/components/mcp/session.ts
@@ -34,6 +34,7 @@ const DEFAULT_IDLE_TIMEOUT_SECONDS = 1800;
  * is identical either way.)
  */
 const EVICTION_WINDOW_SECONDS = 60;
+const DEFAULT_SUBSCRIPTION_LOCK_TIMEOUT_MS = 10_000;
 
 export interface McpSessionRecord {
 	id: string;
@@ -68,6 +69,7 @@ export interface McpSessionRecord {
 }
 
 let _sessionTable: Table | undefined;
+let subscriptionLockTimeoutMs = DEFAULT_SUBSCRIPTION_LOCK_TIMEOUT_MS;
 
 /**
  * Lazily declare the system table. Called by `ensureSessionTable()` at
@@ -112,6 +114,10 @@ export function ensureSessionTable(): Table {
 /** Test seam: allow tests to inject a fake table without touching Harper. */
 export function _setSessionTableForTest(fake: Table | undefined): void {
 	_sessionTable = fake;
+}
+
+export function _setSubscriptionLockTimeoutForTest(value: number | undefined): void {
+	subscriptionLockTimeoutMs = value ?? DEFAULT_SUBSCRIPTION_LOCK_TIMEOUT_MS;
 }
 
 function getTable(): Table {
@@ -171,10 +177,23 @@ function subscriptionLockKey(id: string): string {
 
 function acquireSessionSubscriptionLock(store: Table['primaryStore'], key: string): Promise<void> {
 	return new Promise((resolve, reject) => {
+		let settled = false;
+		const timer = setTimeout(() => {
+			if (settled) return;
+			settled = true;
+			reject(new Error('Timed out acquiring MCP subscription lock'));
+		}, subscriptionLockTimeoutMs);
 		const attempt = () => {
+			if (settled) return;
 			try {
-				if (store.tryLock(key, attempt)) resolve();
+				if (store.tryLock(key, attempt)) {
+					settled = true;
+					clearTimeout(timer);
+					resolve();
+				}
 			} catch (error) {
+				settled = true;
+				clearTimeout(timer);
 				reject(error);
 			}
 		};
--- a/components/mcp/session.ts
+++ b/components/mcp/session.ts
@@ -166,7 +166,6 @@ export async function loadSession(id: string): Promise<McpSessionRecord | null> 
 	return record;
 }
 
-/** Incrementally update session fields without replacing concurrent changes. */
 export async function patchSession(id: string, changes: Partial<Omit<McpSessionRecord, 'id'>>): Promise<void> {
 	await (getTable() as any).patch({ id, ...changes });
 }
--- a/components/mcp/session.ts
+++ b/components/mcp/session.ts
@@ -165,6 +165,43 @@ export async function patchSession(id: string, changes: Partial<Omit<McpSessionR
 	await (getTable() as any).patch({ id, ...changes });
 }
 
+function subscriptionLockKey(id: string): string {
+	return `mcp-subscriptions:${id}`;
+}
+
+function acquireSessionSubscriptionLock(store: Table['primaryStore'], key: string): Promise<void> {
+	return new Promise((resolve, reject) => {
+		const attempt = () => {
+			try {
+				if (store.tryLock(key, attempt)) resolve();
+			} catch (error) {
+				reject(error);
+			}
+		};
+		attempt();
+	});
+}
+
+/** Serialize durable subscription read-modify-writes across HTTP workers. */
+export async function updateSessionSubscriptions(
+	id: string,
+	update: (subscriptions: string[]) => string[] | Promise<string[]>
+): Promise<McpSessionRecord | null> {
+	const store = getTable().primaryStore;
+	const key = subscriptionLockKey(id);
+	await acquireSessionSubscriptionLock(store, key);
+	try {
+		const session = await loadSession(id);
+		if (!session) return null;
+		const subscriptions = session.subscriptions ?? [];
+		const updated = await update(subscriptions);
+		if (updated !== subscriptions) await patchSession(id, { subscriptions: updated });
+		return updated === subscriptions ? session : { ...session, subscriptions: updated };
+	} finally {
+		store.unlock(key);
+	}
+}
+
 export async function deleteSession(id: string): Promise<void> {
 	await (getTable() as any).delete(id);
 	// Tear down ancillary per-session in-memory state — the `tools/list`
--- a/components/mcp/session.ts
+++ b/components/mcp/session.ts
@@ -3,7 +3,7 @@
  *
  * Eviction is delegated to Harper's native TTL (`Table.setTTLExpiration`):
  * every write to a session record updates its `version`, which Harper uses
- * to determine expiration. So calling `saveSession(record)` on each request
+ * to determine expiration. So calling `patchSession(id, changes)` on each request
  * gives sliding-window idle semantics for free — no custom timer, no sweep.
  *
  * Spec: when a request bears an `Mcp-Session-Id` the server doesn't
@@ -63,6 +63,8 @@ export interface McpSessionRecord {
 	 * to clients that declared support. Undefined = client declared none.
 	 */
 	clientCapabilities?: Record<string, unknown>;
+	/** Node-local hint for the worker that owns the current GET-SSE stream. */
+	streamOwner?: { threadId: number; token: string };
 }
 
 let _sessionTable: Table | undefined;
@@ -90,6 +92,7 @@ function declareSessionTable(): Table {
 			{ name: 'logLevel' },
 			{ name: 'subscriptions' },
 			{ name: 'clientCapabilities' },
+			{ name: 'streamOwner' },
 		],
 	});
 }
@@ -149,12 +152,9 @@ export async function loadSession(id: string): Promise<McpSessionRecord | null> 
 	return record;
 }
 
-/**
- * Persist updated session state. Used to bump `lastActivity` (sliding-window
- * idle reset) and to flip `initialized` after `notifications/initialized`.
- */
-export async function saveSession(record: McpSessionRecord): Promise<void> {
-	await (getTable() as any).put(record);
+/** Incrementally update session fields without replacing concurrent changes. */
+export async function patchSession(id: string, changes: Partial<Omit<McpSessionRecord, 'id'>>): Promise<void> {
+	await (getTable() as any).patch({ id, ...changes });
 }
 
 export async function deleteSession(id: string): Promise<void> {
@@ -174,6 +174,6 @@ export async function deleteSession(id: string): Promise<void> {
  */
 export async function touchSession(record: McpSessionRecord): Promise<McpSessionRecord> {
 	const touched: McpSessionRecord = { ...record, lastActivity: Date.now() };
-	await saveSession(touched);
+	await patchSession(record.id, { lastActivity: touched.lastActivity });
 	return touched;
 }
--- a/components/mcp/subscriptionRouting.ts
+++ b/components/mcp/subscriptionRouting.ts
@@ -32,6 +32,7 @@ interface Response {
 }
 
 interface ItcBridge {
+	available?: boolean;
 	sendToThread(threadId: number, event: { type: string; message: unknown }): boolean;
 	onMessageByType(type: string, listener: (event: { message?: unknown }) => void): void;
 }
@@ -59,14 +60,14 @@ function bridge(): ItcBridge {
 	try {
 		const { onMessageByType } = require('../../server/threads/manageThreads.js');
 		if (typeof threads !== 'undefined' && typeof threads.sendToThread === 'function') {
-			return { sendToThread: threads.sendToThread.bind(threads), onMessageByType };
+			return { available: true, sendToThread: threads.sendToThread.bind(threads), onMessageByType };
 		}
 	} catch (error) {
 		harperLogger.trace(`MCP subscription routing is unavailable: ${(error as Error).message}`);
-		return { sendToThread: () => false, onMessageByType: () => {} };
+		return { available: false, sendToThread: () => false, onMessageByType: () => {} };
 	}
 	harperLogger.trace('MCP subscription routing is unavailable: thread bridge is not initialized');
-	return { sendToThread: () => false, onMessageByType: () => {} };
+	return { available: false, sendToThread: () => false, onMessageByType: () => {} };
 }
 
 export function _setSubscriptionItcForTest(fake: ItcBridge | undefined): void {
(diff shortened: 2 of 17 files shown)


kylebernhardy:
Resolved the v5.2 cherry-pick conflict on `cherry-pick/v5.2/pr-2436` at `28feda035`. The cherry-pick workflow re-dispatch regenerated the markers instead of testing the resolved branch, so I restored the resolution and dispatched the underlying release gates directly:

- [v5.2 integration tests](https://github.com/HarperFast/harper/actions/runs/33472603632)
- [v5.2 unit tests](https://github.com/HarperFast/harper/actions/runs/33472604884)

components/mcp/session.ts
```diff
@@ -145,16 +148,58 @@ export async function createSession({
...
+/** Incrementally update session fields without replacing concurrent changes. */
+export async function patchSession(id: string, changes: Partial<Omit<McpSessionRecord, 'id'>>): Promise<void> {
+	await (getTable() as any).patch({ id, ...changes });
+}
+
+function subscriptionLockKey(id: string): string {
+	return `mcp-subscriptions:${id}`;
+}
+
+function acquireSessionSubscriptionLock(store: Table['primaryStore'], key: string): Promise<void> {
```
kriszyp, line 177: This wait is unbounded, and the repo's own precedent treats that as a known trap: every other `tryLock` waiter bounds the wait because a holder that dies uncleanly never unlocks (`components/componentLoader.ts:426-436` — "The timer bounds the wait if the holder dies without unlocking"; the `LOCK_TIMEOUT` force-release in `resources/Table.ts:6111-6113`; DESIGN.md's account of the harper#2251 unbounded-spin incident).

The exposure here is worse than for those callers because `handleGet` now acquires this lock on **every** SSE open (`components/mcp/transport.ts:429`), even for sessions with no subscriptions. If a worker is force-killed mid-hold (OOM restart, drain-deadline kill during a deploy — routine events over a process lifetime), the lock is orphaned for the rest of the process: every later GET-SSE open and subscribe/unsubscribe for that session hangs forever, and the client's hung requests pile up with no error. The remote-routing path at least times out after 30s on the sender, but the owner's `withSessionSubscriptionLock` chain stays wedged behind the stuck acquisition, so every subsequent command for the session returns `timeout` until the worker restarts — and the direct call in `dispatchResourcesUnsubscribe`'s no-live-stream fallback (`components/mcp/transport.ts:977`) has no timeout at all.

Suggest a bounded acquire (a ref'd timer, per the componentLoader note that the unlock wake arrives via an unref'd threadsafe function) that rejects or returns a sentinel after ~10s, mapped to the existing retryable `internal-error` result. Fail-open (proceed without the lock) is also defensible given the operation self-heals on reconnect, but fail-with-retryable-error is simpler to reason about.

components/mcp/subscriptionRouting.ts
```diff
@@ -0,0 +1,286 @@
...
+		)
+		.finally(() => {
+			if (operationChains.get(sessionId) === tail) operationChains.delete(sessionId);
+		});
+	operationChains.set(sessionId, tail);
+	return current;
+}
+
+async function executeLocal(command: Command): Promise<SubscriptionRouteResult> {
+	const registered = getRegisteredSession(command.sessionId);
```
kriszyp, line 232: This liveness check runs when the command *arrives*, but the mutation executes later, behind whatever is already queued on `withSessionSubscriptionLock` — e.g. a multi-URI `restoreResourceSubscriptions` that holds the chain across N audit-log subscribes. If the GET stream closes in that window, the queue's close hook has already run `dropSessionSubscriptions`, and the subscribe that then executes calls `addResourceSubscription` for a stream that no longer exists: the audit-log iterator it starts is never stopped unless the client later reconnects to and disconnects from this same worker. That is exactly the leak the old pre-check guarded against (the removed comment in `transport.ts`: "A subscription opened without a stream would leak its audit-log iterator and drop every update silently") — the pre-existing window was the single `await` inside `addResourceSubscription`; queueing behind the chain widens it substantially.

Fix: repeat the `getRegisteredSession(command.sessionId)?.streamToken === command.streamToken` check *inside* the chained callback (first line of the `withSessionSubscriptionLock` operation) and return `no-live-stream` if it fails. The early check can stay as a fast-path rejection. This also correctly rejects a command whose stream was superseded on this worker while it waited (new registry entry, new token).

components/mcp/transport.ts
```diff
@@ -955,10 +970,20 @@ async function dispatchResourcesUnsubscribe(
...
 	}
-	removeResourceSubscription(session.id, uri);
-	if (session.subscriptions?.includes(uri)) {
-		session.subscriptions = session.subscriptions.filter((u) => u !== uri);
-		await saveSession(session);
+	const result = await routeResourceSubscription({ session, operation: 'unsubscribe', uri });
+	if (result === 'no-live-stream') {
+		// The live owner is already gone. Remove durable state locally so a later
+		// reconnect cannot restore the cancelled subscription.
+		await updateSessionSubscriptions(session.id, (subscriptions) =>
```
kriszyp: This fallback can diverge from the owner's live state in the claim race. Sequence: client reconnects SSE — `handleGet` runs `registerSession` (new token), `claimSubscriptionOwner` (patch), then restore under the lock. An unsubscribe POST that loaded the session just before the claim patch routes to the *old* owner/token, gets `no-live-stream`, and lands here. If the new owner's restore already ran (URI restored live), this removes the durable URI but leaves the live subscription running on the new owner — the client gets a 200 success yet keeps receiving `notifications/resources/updated` for that URI until the next stream close (restore only prunes the durable list; it never stops live subs).

Cheapest robust fix: after `no-live-stream`, re-load the session and, if `streamOwner` changed since this request's snapshot, re-route the command once to the new owner before falling back to the local durable removal. That converges both the claim-race case and the genuine no-stream case.

components/mcp/transport.ts
```diff
@@ -403,6 +407,12 @@ async function handleGet(request: NormRequest): Promise<NormResponse> {
 		};
 	}
 	const record = registerSession(sessionId, request.profile, effectiveUser(request));
+	try {
+		await claimSubscriptionOwner(sessionId, record.streamToken);
+	} catch (error) {
+		unregisterSession(sessionId);
```
kriszyp: `unregisterSession(sessionId)` tears down whatever record is registered *now*, not necessarily the `record` this request created. If a second GET for the same session raced in on this worker between `registerSession` and a failing `claimSubscriptionOwner`, this closes the superseding stream's queue — killing the healthy new connection because the losing one failed to claim. The registry itself guards this exact race in its close hook (`components/mcp/sessionRegistry.ts:120-124`, "Only delete if we're still the live record").

Emitting close on this request's own queue — `record.queue.emit('close')` — gets the correct scoped teardown: the close hook deletes the registry entry only if `record` is still the live one.

components/mcp/subscriptionRouting.ts
```diff
@@ -0,0 +1,286 @@
...
+	await patchSession(sessionId, { streamOwner: { threadId: currentThreadId(), token: streamToken } });
+}
+
+function countPendingForSession(sessionId: string): number {
+	let count = 0;
+	for (const entry of pending.values()) if (entry.sessionId === sessionId) count++;
+	return count;
+}
+
+function subscriptionUser(user: AuthedUser): AuthedUser {
```
kriszyp: Question on the projection scope: because `command` is built before the local/remote branch in `routeResourceSubscription`, this allow-list now applies to *same-worker* subscribes too — so `subscribeToResource` sees a thinner principal than it did before this PR (previously the full `effectiveUser(request)` object). `context.user` flows into `transaction(context, …)` and the resource's `subscribe` (`components/mcp/resources.ts:420-421`), where an application resource's custom `allowRead`/`subscribe` override may consult fields outside this list (e.g. custom claims a token-auth setup attaches to the user object). `role.permission` and the scoped-token fields are preserved, so built-in permission checks are fine, and applying the same projection to both paths is the right call for local/remote consistency — but it's a silent behavior change for custom authorization. Worth a comment here stating the contract (which fields subscribe-time authorization may rely on), and confirming nothing in the built-in auth path reads `user.active` or role identity fields beyond `role.role`/`role.permission`.

integrationTests/mcp/sse-listchanged.test.ts
```diff
@@ -318,4 +470,63 @@ suite('MCP v1 SSE channel + list_changed delivery', (ctx: ContextWithHarper) =>
...
 	});
+
+	test('N5: resource subscription routes from a sibling POST worker to the GET-SSE owner', async (t) => {
+		if (process.platform === 'win32') {
+			t.skip('Harper forces one HTTP worker on Windows');
+			return;
+		}
+		const auth = adminAuth(ctx);
+		const session = await initialize(ctx.harper.httpURL, auth);
+		const firstLocalPort = 30000 + (process.pid % 10000);
```
kriszyp: These pinned local ports (30000 + pid%10000, plus up to 24 successors) overlap the default Linux ephemeral range (32768–60999), so any concurrently allocated outbound socket — including ones this very test suite opens via `fetch` — can already hold the port and the `connect` fails EADDRINUSE. The POST loop tolerates that only by accident (the error rejects `workerIdentity`, failing the test rather than advancing `attempt`), and the GET agent has no retry at all, so this is a straight flake vector on busy CI hosts — the class of timing flakiness AGENTS.md calls out.

Two options: catch EADDRINUSE in `requestWithAgent`/`workerIdentity` and retry with the next candidate port (both for the initial GET port and inside the POST loop), or drop `localPort` entirely if distinct keep-alive sockets alone are enough to reach different workers (each fresh `Agent` opens a new connection; the pinning matters only if Harper's worker distribution hashes on the source port — if it does, keep the ports but pick from a range below 32768, e.g. 20000–29999).

components/mcp/subscriptionRouting.ts
```diff
@@ -0,0 +1,286 @@
...
+	const owner = args.session.streamOwner;
+	if (!owner) return 'no-live-stream';
+	const command = {
+		sessionId: args.session.id,
+		operation: args.operation,
+		uri: args.uri,
+		...(args.user ? { user: subscriptionUser(args.user) } : {}),
+	};
+	if (owner.threadId === currentThreadId()) {
+		return executeLocal({ ...command, requestId: '', originator: currentThreadId(), streamToken: owner.token });
```
kriszyp: Nit: error-surface asymmetry between the two paths. Here a throw from `executeLocal` (e.g. the durable update failing) propagates to `handleMcpRequest`'s catch and becomes an HTTP 500, while the identical failure on a remote owner is caught by `handleCommand` and comes back as `internal-error` → a 200 with JSON-RPC `-32603`. Clients see different shapes for the same fault depending on which worker their POST happened to land on. Wrapping this call in try/catch and returning `'internal-error'` makes the local path match the remote one (and `dispatchResourcesSubscribe`'s rollback already runs inside `executeLocal`, so nothing is lost).

kriszyp (changes requested):
I'm skeptical we want to try to do this with cross-thread communication rather than persisting enough state to be able to pick up the subscription from any thread. Or alternately, we could look at establishing sufficient session affinity in symphony to generally route to the same thread. WDYT?

kylebernhardy:
@kriszyp I traced the persistence-first option. A worker can watch durable subscription state, but a simple state update leaves a gap: POST may return before the GET owner installs the live iterator, and custom `Resource.subscribe` streams are not necessarily replayable. Closing that gap appears to require an applied-revision acknowledgment, owner fencing/lease, and principal handling. Watching `mcp_session` would also audit every `lastActivity` update, so a dedicated subscription table seems cleaner if we take that route.

Bounded ITC looks smaller for this issue, while a durable state controller may be stronger long term. What persisted-state flow did you have in mind? More detail would help guide the implementation.

components/mcp/session.ts
kylebernhardy: Added a ref-counted 10-second acquire timeout with a settled guard, so a late wake cannot acquire after the caller has failed. Lock failures now reach the existing retryable JSON-RPC error paths. Covered by the timeout regression test.

components/mcp/subscriptionRouting.ts
kylebernhardy: Added the stream-token check inside the queued operation. A command that waited behind restore now rejects if the stream closed or was superseded before mutation.

components/mcp/transport.ts
kylebernhardy: Unsubscribe now reloads and reroutes across owner changes for three bounded attempts. If ownership is still moving, it returns a retryable -32603 and preserves durable state instead of deleting behind the new owner.

components/mcp/transport.ts
kylebernhardy: The failure path now closes only this request record queue. The registry identity guard leaves any superseding GET stream intact.

components/mcp/subscriptionRouting.ts
kylebernhardy: Projection now happens only when sending across ITC. Same-worker subscribe and reconnect restore receive the full principal, while the remote payload retains the documented built-in authorization fields and excludes credentials and mutable custom state.

integrationTests/mcp/sse-listchanged.test.ts
kylebernhardy: Pinned ports now use 20000-28999 and both connection paths retry EADDRINUSE. N5 also fails instead of skipping when configured workers cannot be reached, so CI cannot pass without exercising the cross-worker route.

components/mcp/subscriptionRouting.ts
kylebernhardy: Local execution is now contained through the same internal-error result as remote execution, preserving HTTP 200 with correlated JSON-RPC -32603 for equivalent failures.

kylebernhardy:
@kriszyp I addressed the review threads, including the bounded lock, ownership races, local/remote authorization behavior, and fail-closed cross-worker coverage. The final Claude pass found no remaining code issues. When you have a chance, could you take another look?

kylebernhardy:
@kriszyp Following up on the subscription-routing design: the review findings are addressed, and I replied with the tradeoffs I found in the persisted-state approach. Could you re-review the current head or share any further direction on the architecture?\n\nComment generated by kAIle (GPT-5)

kriszyp (changes requested):
Can we consider changing this to using persisted stream state so that we can use standard Harper subscriptions instead of a custom inter-thread communication mechanism?

Issue #483, opened 2026-05-06
Add ability to exclusively lock a record for performing a safe operation


kriszyp:
This will be another method on the Table/Resource interface, with the form of:

`table.lock(primaryKey): Promise<WritableRecord>`

(this is very similar interface to the update method, and returns the same record type).

If a node already has an exclusive lock on a record, it can immediately proceed.

If a node does not have an exclusive lock, but has a shared lock, it needs to broadcast a request to all nodes that it wants to acquire an exclusive lock. Once all other nodes have responded and granted this request, the operation can proceed. If an exclusive lock is held by one other node, the lock can be requested from that node.

We may support local-only locks for some use cases as well.

There may need to be a distinction between a node’s lock ownership (is retained after commit) and the transaction’s lock ownership (released after commit).

Changing a record from locked to unlocked should trigger a version change (but possibly only an audit entry for when a node is   locking), to ensure no concurrent attempts to change.

Locking will be recorded in record metadata, and should delay any other writes that are attempted on that record.

kriszyp:
Cross-referencing exploratory QA that motivates this primitive — the absence of a safe non-commutative claim/CAS shows up concretely across single-node and cluster:

**Single-node safe-write matrix.** Atomic deltas (`addTo`/`subtractFrom`) are the *only* safe concurrent-write primitive — the committed aggregate is exact on both engines because the merge is commit-resolved. Every CAS-shaped idiom leaks under contention because the decision is a snapshot read taken before commit and not re-validated at commit:
- `ifVersion` / version-guarded read-modify-write
- create-only first-write-wins
- naive read-modify-write
- REST `If-Match` (silently ignored)

(LMDB leaks on essentially every contended round; RocksDB more rarely.)

**Consequences customers hit:**
- `addTo` returns void and the read-back is unreliable, so there's no "read your serialized position" → no exactly-once claim or hard limit.
- `subtractFrom` has no floor → inventory oversells to negative under concurrency.
- **Cluster (2-node):** a rate limiter built on `addTo` + read-then-decide over-admits by ~+7 on a 30-request burst with a limit of 10 during the replication-lag window — the count still converges (deltas merge), but each node's allow/deny decision is made on its local, not-yet-converged view. There is no conditional-write / lock anywhere, so a hard global limit (rate limit, inventory floor, exactly-once claim) is not achievable today.

[...]

PR #2462, opened 2026-09-02
Exclusive record locks: table.lock(id) serialized across worker threads (Phase 0 of #483)

kriszyp:
Adds process-wide, per-record RocksDB locks that serialize `table.lock(id)` across worker threads on one node while leaving ordinary writes ungated. This review-feedback update exports the public lock types and makes surviving hold versions monotonic only after a real commit—including [static same-key writes](https://github.com/HarperFast/harper/pull/2462/changes?w=1#diff-2ae62080514cb4ef64bf50088e7ea0587701f558af71f76b119f07e00f3d56beR778)—without adding a record-lock scan or key lookup to ordinary transactions that own no locks.

The update also closes transaction-integrity edges found during review: acquiring a scoped lock no longer aborts a native handle containing earlier writes, scoped-to-hold upgrade detaches the correct staged write through an intervening same-key operation, synchronous sweep failures abort and release sibling locks, failed/skipped writes do not advance the holder floor, and [lock acquisition no longer mints a stale ImmediateTransaction timestamp](https://github.com/HarperFast/harper/pull/2462/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R2582). One unresolved concern needs human judgment: acquisition-time record versions are also transaction-log timestamps, so a delayed holder mutation may fall behind a replication cursor that has already passed that time.

## For the human reviewer

1. **Advisory scope:** this phase serializes only `lock()` participants inside one RocksDB process; ordinary writes and other nodes remain ungated. Gating ordinary writes was rejected because it added a native lock round-trip to every write and introduced deadlock/livelock shapes. Reversing this after release would be a breaking behavior change.
2. **Holder precedence and replication ordering:** holder writes use the acquisition-time version so a later ordinary write wins under LWW. The alternative is commit-time precedence. Please explicitly rule on the consequence that a holder mutation committed later can carry a transaction-log timestamp older than a replica's cursor; a “no” requires separating replication ordering from record conflict ordering before merge.
3. **Immediate scoped lifetime:** a scoped lock outside `transaction()` survives sequential saves until `unlock()` or lease expiry. Requiring an explicit transaction or releasing after one save would reduce lifecycle states but break the tested sequential-save behavior.
4. **Hold-mode API:** `{ hold: true }` returns a resource-attached handle that outlives its acquiring transaction and is explicitly released with `unlock()`. Restricting the API to transaction scope would be simpler but removes the intended long-lived use case and is easiest to decide before shipping.
5. **Platform scope:** Phase 0 is in-process and RocksDB-only; LMDB returns 501 and no protocol authorization surface is added. Engine parity, cluster coordination, and protocol policy are deferred extensions rather than hidden fallbacks.
6. **Known lifecycle boundaries:** overlapping un-awaited saves on one handle, scoped locks sharing an ImmediateTransaction with unrelated instance writes, and sourced/caching-table reload semantics remain unproved. The documented and tested use is awaited saves on local RocksDB tables; expanding those guarantees should be a deliberate follow-up rather than inferred from this phase.

## Verification

- `npm run build`; `npm run typecheck`; `npm run test:types`; and `npm run lint:required` all pass. Full `npm run lint` reaches only 13 pre-existing warnings in untouched files.
- `npx mocha unitTests/resources/recordLock.test.js`: 71 passing, 1 pending after the transaction/version fixes. Targeted reruns also pass the [different-key staged-write preservation](https://github.com/HarperFast/harper/pull/2462/changes?w=1#diff-8f5ad3e1088e03e3aef7ed1c0d5ecd0fc749dbda3fda99c32a3ef86d4737cb11R733), [static same-key floor](https://github.com/HarperFast/harper/pull/2462/changes?w=1#diff-8f5ad3e1088e03e3aef7ed1c0d5ecd0fc749dbda3fda99c32a3ef86d4737cb11R1174), and [owning-link timestamp isolation](https://github.com/HarperFast/harper/pull/2462/changes?w=1#diff-8f5ad3e1088e03e3aef7ed1c0d5ecd0fc749dbda3fda99c32a3ef86d4737cb11R1447) cases. The static same-key regression fails on the pre-fix parent as expected.
- `npm run test:unit:resources`: 2068 passing, 29 pending. `HARPER_STORAGE_ENGINE=lmdb npx mocha unitTests/resources/recordLock.test.js`: 2 passing, 65 pending, confirming the LMDB 501 boundary.
- `npm run test:integration -- "integrationTests/resources/record-lock-concurrency.test.ts"`: 3 passing across four workers. The full integration run completed with only six tests requiring an unavailable live Ollama service failing; Harper's local/in-process suites passed.
- `npm run test:unit:main`: 5286 passing, 196 pending, with five unrelated environment-sensitive failures (application-spawn cleanup, deploy socket observation, harness Git environment, missing RSA fixture, and the long-worktree Unix socket warning).
- Independent review ran five rounds on the fix sequence with Claude, Gemini, Cursor/Grok, and Harper storage/data adjudication. It closed the staged-write abort blocker, tail-only upgrade detachment, failed-floor advancement, default-path bookkeeping, and timestamp-minting findings; the replication-ordering decision above remains intentionally open for human review.

Complexity: complicated

<sub>Review-Coverage: authored=codex; ran=claude,gemini; declined=cursor-grok,cursor-composer,domain; rounds=5 @ 553cfb604793</sub>

<sub>Human-Review-Need: 3 @ 553cfb604793</sub>


kriszyp:
Strong Phase 0 — the single-durable-authority model, the stage-nothing gate, and the conditional UNLOCK replay are exactly the right shape. One concern worth resolving before this merges, because it's cheapest to fix at the choke point this PR already owns:

## The lock descriptor leaves the node on every data write to a locked record

The mixed-version story here covers the LOCK/UNLOCK control entries (`LOCAL_ONLY`, peer never sees action 9/10 — good). But holder writes and every preserved-bit rewrite are ordinary PUT/PATCH entries whose encoded record carries `LOCKED` (0x4) + `lockVersion` + `lockExpiresAt` in the metadata prefix, and DESIGN.md makes that deliberate: *"the record itself replicates as usual."* Replication and base copy forward encoded entries — the send path tests `metadataFlags` precisely so it never re-encodes. Two consequences:

1. **Silent value corruption on pre-Phase-0 peers.** DESIGN.md already states the mechanism — *"an older Harper decodes a LOCKED record's value from the wrong offset"* — but scopes it to local downgrade. The same bytes travel the wire: during a rolling upgrade (or to a v5→v4 bridge route, or a clone target), any replicated write to a locked record hands an older peer a layout it decodes 16 bytes short. That's the silent-corruption class we've spent months eliminating, and "don't downgrade this node" doesn't cover it.

2. **Phantom locks on same-version peers.** A peer applies a holder's write with the bit and live `lockExpiresAt` baked in, and its own local writers then hit the `gateOnLock` path — but the origin's UNLOCK never arrives (`LOCAL_ONLY`), so the peer stays write-gated for the *full residual lease* after the origin released. Base-copy targets inherit the same phantom state. That contradicts the PR's own scope statement ("one node, every worker thread") in the deployment mode that matters most.

## Suggested fix: strip the descriptor from the audit/outbound encoding

Dual-encode locked-record data writes at the `recordUpdater` choke point: the **store** copy keeps `LOCKED` + descriptor; the **audit-log** copy (which is what replication and copy read) drops the bit and both fields. Costs one extra encode only on writes to locked records — same cost class as the lock transitions themselves.

Notably, this is *consistent with the crash-recovery stance this PR already takes*: "A LOCK is not replayed: its holder died with the process." Replay restoring the stripped (unlocked) form is the same semantics, lease-bounded, and the conditional `_writeUnlockReplay` doesn't depend on the descriptor being present in data entries. Worth double-checking the resequencing/out-of-order history walk paths that reload entries from the audit log, since store and log encodings would no longer be byte-identical for locked records.

Phase 1 then replaces the strip with capability-gated propagation (the `recordLocks` capability in the harper-pro protocol-capability registry, harper-pro#440), rather than introducing lock state to the wire accidentally now and having to define its semantics retroactively.

If the dual encoding is judged too invasive for this PR, the minimum bar is widening the documented limitation from "no downgrade" to "no replication/clone of lock-bearing tables in a mixed-version cluster, and expect residual-lease write-gating on same-version peers" plus a release-note warning — but the strip closes both holes for real and keeps the PR's central invariant ("what makes a caller the holder is that its conditional write committed — nothing else, and no one else sees it") true on the wire, not just on the node.

— Claude (Fable 5), reviewing on behalf of Kris

integrationTests/resources/record-lock-concurrency.test.ts
```diff
@@ -0,0 +1,153 @@
...
+		strictEqual(held.status, 200, `held: ${JSON.stringify(held.body)}`);
+		ok(held.body.lockVersion > 0, 'the lock generation is on the record');
+
+		const started = Date.now();
+		const write = await put(id, { n: 2, holders: 1 });
+		ok(write.status === 200 || write.status === 204, `the write landed (${write.status})`);
+		ok(write.elapsed >= LEASE - 200, `the write waited for the lease (${write.elapsed}ms)`);
+		ok(Date.now() - started < LEASE + 5_000, 'and proceeded promptly after it');
+
+		const after = await get(id);
```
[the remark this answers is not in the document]
kriszyp: Renamed to `recordAfter`. — Claude (Fable 5.1)

kriszyp:
Addressed by removing the record write altogether rather than dual-encoding it. The lock is now the rocksdb-js process-wide in-memory key lock (`store.tryLock`/`store.unlock`), so `lock()`/`unlock()` write nothing: no `LOCKED` bit, no `lockVersion`/`lockExpiresAt` fields, no audit actions 9/10, no `LOCAL_ONLY` stamping, and the record's version never moves for a lock. Both consequences you listed disappear with the descriptor: nothing lock-related travels the wire (no pre-Phase-0 decode hazard), and peers cannot inherit a phantom lock because no data write carries one. The out-of-order drop that was the PR's entry 1 goes with it. Phase 1 will carry lock request/grant/release as control transaction-log entries instead of record rewrites, so the record format stays untouched then too. `DESIGN.md` has the new protocol.

— Claude (Fable 5.1)

resources/DatabaseTransaction.ts
```diff
@@ -867,6 +939,33 @@ export class DatabaseTransaction implements Transaction {
...
+		// Lock-write timestamp rules.
+		if (operation.lockHandle) {
+			const handle = operation.lockHandle;
+			if (this.open === TRANSACTION_STATE.CLOSED || this.saveCommits) {
+				// CLOSED path (second+ write per ImmediateTransaction cycle) OR the first write in
+				// an ImmediateTransaction (open=OPEN until commit sets it CLOSED, but saveCommits
+				// signals the per-write-commit semantics):  stamp with nextHolderVersion() so that
+				// each sequential save() gets its own monotonically increasing stamp — scoped or hold.
+				if (!operation.lockStamp) operation.lockStamp = handle.nextHolderVersion();
+				this.timestamp = operation.lockStamp;
```
cb1kenobi: **Lock stamp leaks onto unrelated writes on the same context — High (data integrity)**

`this.timestamp = operation.lockStamp` pins the *link's* clock to the lock's acquisition time, but the reset to `0` only happens once the commit settles (async success handler, sync-path end, or `abort()`). On the async path `clearWrites()` runs first, and `ImmediateTransaction.clearWrites()` re-opens the transaction (`open = OPEN`) — so the context accepts new writes while `_timestamp` still holds the past stamp. Any write staged in that window reads it at `let txnTime = this.timestamp` on the next line, including writes that carry no `lockHandle` at all.

**Failure mode** — in one request handler:

```js
const order = await Order.lock(id, { hold: true });
order.set('status', 'shipped');
await Promise.all([order.save(), Audit.put({ id: logId, /* ... */ })]);
```

`Audit.put` is stamped at `order`'s `acquiredAt` rather than real time. If `logId` already carries a version newer than that, the put is silently dropped by LWW — no error to the caller — and peers drop it for the same reason on a replicated table. The record is unrelated to the lock, so none of the Phase 0 ordering reasoning applies to it.

This is the same leak that the per-write `lockStamp` field fixed for lock writes; non-lock writes on the same link have no equivalent protection because the link clock is still mutated here.

**Simplest fix** — keep the stamp per-write and consume it locally instead of mutating the link clock:

```js
if (!operation.lockStamp) operation.lockStamp = handle.nextHolderVersion();
// ...
let txnTime = operation.lockStamp ?? this.timestamp;
```

leaving `this.timestamp` untouched on this path.

—
Reviewed 8a6490db

kriszyp: Great catch — this was real, and the failure was reachable in a plainer shape than the cross-table example: an *off-key* write through the lock-holding resource itself. `holder.save()` leaves the link pinned to `acquiredAt` for the whole in-flight commit, so `holder.delete(otherId)` staged behind it inherits the lock's version and LWW drops it silently.

Fixed as you suggested: the stamp stays on the operation and `save()` consumes it locally (`let txnTime = operation.lockStamp ?? this.timestamp;`), with the `this.timestamp = operation.lockStamp` assignment gone. That also makes the code match what `DESIGN.md` already claimed — *"each write through a scoped lock is stamped with `nextHolderVersion()` … without pinning the context clock."*

Regression coverage: `unitTests/resources/recordLock.test.js` → *"a lock write does not stamp an unrelated write staged on the same context"*. It fails on the parent commit with the off-key delete dropped, and passes here.

— Claude (Opus 5)

unitTests/resources/recordLock.test.js
```diff
@@ -0,0 +1,1120 @@
...
+			}); // commit; hold survives
+			// Without prime: nextHolderVersion() = T_A (tie with in-transaction write) → dropped.
+			// With prime: nextHolderVersion() = T_A + MIN_STEP → lands.
+			holdRef.set('n', 2);
+			await holdRef.save();
+			await holdRef.unlock();
+			assert.strictEqual((await LockTest.get(recordId)).n, 2, 'post-commit hold write landed (not dropped by LWW tie)');
+		});
+
+		it('minor-2: concurrent lock() calls for same key in one transaction coalesce, not self-block', async function () {
```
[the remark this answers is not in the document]
kriszyp: Added all three, and (3) was not just coverage — it failed.

- **(1) follower deadline** — *"a coalesced follower fails at its own timeout, not the leader's"*: leader waits 3000ms, follower 200ms; the follower must 423 at its own deadline.
- **(2) leader fails** — *"a coalesced follower retries on its own budget when the leader fails"*: leader times out at 150ms, follower re-enters with its remaining budget and acquires once the outside holder releases.
- **(3) transaction closes while parked** — *"a transaction aborted while lock() waiters are parked leaves the key unlocked"*: **failed on the parent commit.** The follower's retry calls `this.lock()`, which re-resolves the context; after the abort that no longer points at the aborted link, so the acquired handle landed on a *fresh* transaction that no commit or abort ever releases. The native key stayed locked for the full 30 s lease. `retryOnRemainingBudget()` now throws 500 when `link.open === CLOSED && !link.saveCommits`, matching the leader's own post-acquisition guard.

— Claude (Opus 5)

resources/defineTable.ts
```diff
@@ -235,6 +237,7 @@ interface TypedVerbs<S extends Shape> {
 	post(record: InsertOf<S>, context?: any): MaybePromise<unknown>;
 	patch(id: IdOf<S>, changes: PatchOf<S>, context?: any): MaybePromise<unknown>;
 	update(id: IdOf<S>, updates?: PatchOf<S>, context?: any): MaybePromise<InstanceOf<S>>;
+	lock(id: IdOf<S>, options?: RecordLockOptions, context?: any): Promise<WritableRecord<InstanceOf<S>>>;
```
cb1kenobi, line 240: **`lock()`'s typed signature accepts a `context` the implementation drops** — Medium

This line declares `lock(id, options?, context?)`, matching `update(id, updates?, context?)` directly above it, so TypeScript accepts a three-argument call. But the runtime entry point is `Table.static lock(target, options)` (`resources/Table.ts`), which takes only two parameters and resolves the context itself:

```js
const context: any = contextStorage.getStore() ?? {};
const resource = new TableResource(id, context);
```

The third argument is silently discarded.

**Failure mode.** A `defineTable`-authored table calls `Orders.lock(id, { hold: true }, context)` from somewhere without an ambient context — a background job, a timer, a subscription callback — exactly the situation where the other verbs' `context` parameter exists. The passed context is dropped, `contextStorage.getStore()` returns nothing, and the bare `{}` gets a fresh transaction through `txnForContext`. Nothing ever commits or aborts that transaction, so `releaseRecordLocks()` never runs for it and the scoped handle is never released: the native key stays locked for the full lease (30 s by default) and every other `lock()` on that record fails 423 until it expires. When there *is* an ambient context that differs from the one the caller passed, the handle registers on the wrong link instead — released by a transaction the caller never wrote to.

Note that `ResourceInterface.ts` declares the same method without a `context` parameter, so the two type declarations added in this PR also disagree with each other.

**Simplest fix.** Either thread the argument through:

```ts
static lock(target?, options?, context?) {
	// ...
	const ctx: any = context ?? contextStorage.getStore() ?? {};
```

or, if `lock()` is deliberately ambient-context-only, drop `context?` from this signature so the type stops advertising it.

—
Reviewed 6cc71905

kriszyp: Threaded it through — `Table.lock(id, options?, context?)` now resolves the passed context ahead of the ambient store, normalized by a new `contextArgument()` that mirrors what `transactional()` already does (anything carrying a context yields it via `getContext()`, a bare `DatabaseTransaction` becomes the `{ transaction }` slot holding it, and the released placeholder counts as absent). The `contextStorage.getStore() ?? {}` fallback is unchanged.

Both halves of the failure mode you described are now regression tests in `unitTests/resources/recordLock.test.js`, and all three fail on the parent commit:

- *static lock() uses the context argument when there is no ambient one* — an `isExplicit` context runs its callback outside `contextStorage`, so there is genuinely nothing to fall back on. On the parent the handle registered on a fresh `ImmediateTransaction`, whose `releaseRecordLocks()` is a no-op for record locks, so the key stayed locked; the test's follow-up `lock(..., { timeout: 150 })` after the scope commits is what fails 423 there.
- *static lock() prefers the passed context over a differing ambient one* — asserts the handle lands in `passed.transaction.recordLocks` and not in the ambient transaction's.
- *static lock() accepts a transaction in the context position* — the `transactional()`-parity form.

On the second half of your note: I left `ResourceInterface.ts` alone, because I think the two declarations describe different surfaces rather than disagreeing. `ResourceInterface.lock?(target?, options?)` sits with the **instance** verbs, and none of those take a context either (`get?(target)`, `put?(record, target)`, `delete?(target)`) — the instance resolves it from `this.getContext()`. `TypedVerbs` in `defineTable.ts` types the **static** surface, where every verb does take the trailing context. Happy to add it to the interface too if you read that differently.

Writing the regression tests turned up four more things on this path, all fixed in the same push and all with tests:

- **`setTimestamp(0)`.** The scoped-lock path called `link.transaction.setTimestamp(link.timestamp)` unguarded when iterators pinned the read txn, unlike `DatabaseTransaction`'s own two call sites. A deferred `update()` write leaves the clock at 0, so `transaction(async () => { await T.update(b); /* open a search iterator */ await T.lock(a); })` threw `Invalid timestamp, expected positive number` out of `lock()`. Guarded.
- **A misleading 409.** All three lock guards said "Record lock lease expired" for a handle that was expired *or* released — and a scoped handle released by its own commit is the common case, which sent the caller after lease timeouts instead of transaction scope. New shared `lockNotHeldError()` names the actual cause.
- **`lock()` threw synchronously.** Neither entry point was `async`, so option and id validation escaped past a caller's `.catch()` despite the declared `Promise` return. Both are `async` now.
- **A dead alias.** `const authorizedData = data` in `transactional()` was left over from the gating revision; reverting it takes `resources/Resource.ts` out of this PR's diff entirely.

`DESIGN.md` records the context rule and the 409 distinction.

— Claude Opus 5

resources/Table.ts
```diff
@@ -2322,10 +2413,257 @@ export function makeTable(options) {
...
+						// excluded (saveCommits=true) — its clock is never pinned in lock();
+						// each save() stamps with nextHolderVersion() instead.
+						if (link.writes.length === 0 && !link.timestamp) {
+							link.timestamp = handle.acquiredAt;
+							if (resolved.hold) {
+								// Prime the nextHolderVersion counter so a later CLOSED-path save
+								// (e.g. after the transaction commits and the hold is used from an
+								// ImmediateTransaction context) gets acquiredAt+MIN_STEP instead of
+								// acquiredAt again (which would be a LWW tie with the first write).
+								handle.nextHolderVersion();
```
cb1kenobi: **Hold write versions go backwards after commit when the transaction clock was not pinned — Blocking (silent write loss)**

The `nextHolderVersion()` prime runs only inside `if (link.writes.length === 0 && !link.timestamp)`. When a hold is acquired in an explicit `transaction()` that has already staged a write, the clock is not pinned *and* the counter is not primed, so the handle's `#lastHolderVersion` stays `null` while the in-transaction hold write is stamped from the link clock — which is set during the commit save loop to real "now", later than `acquiredAt`.

Failure:

```js
let h;
await transaction(async () => {
	await T.put({ id: other, n: 1 }); // stages a write → clock not pinned, counter not primed
	h = await T.lock(id, { hold: true });
	h.set('n', 1);
	await h.save();                   // OPEN + hold → stamped at the link clock T_commit > acquiredAt
});
h.set('n', 2);
await h.save();                       // CLOSED path → first nextHolderVersion() → acquiredAt
```

That second save is the counter's first call, so it returns `acquiredAt`, which precedes the version the record already carries from the in-transaction write. LWW drops it, no error reaches the caller, and `n` stays `1`.

This is the same drop the round-23 `upgradeToHold()` prime was added to close, one branch over. It is also not the documented mixed-transaction ordering trade-off: no concurrent writer is involved — the holder silently loses its own later write to its own earlier one, which is the read-modify-write coherence `lock()` exists to provide.

**Simplest fix:** stop letting the counter fall behind the clock actually used. In the OPEN, non-`saveCommits`, hold branch of `DatabaseTransaction.save()` (currently comment-only), note the version on the handle:

```js
} else if (handle.hold) {
	handle.noteHolderVersion(this.timestamp); // #lastHolderVersion = max(#lastHolderVersion ?? 0, this.timestamp)
}
```

so the first CLOSED-path save resolves to `max(this.timestamp + MIN_STEP, acquiredAt)`. Moving the `handle.nextHolderVersion()` call out of the pinning guard on its own is not sufficient — priming to `acquiredAt` still leaves the counter below an unpinned transaction clock.

**Regression test:** extend the "rule D" case with a post-commit `save()` after a hold taken in a transaction that already put a different record; it should fail on this commit.

—
Reviewed 3f5e72e6

resources/ResourceInterface.ts
```diff
@@ -258,6 +264,21 @@ interface TypedUpdatableRecord<Record extends object, Property extends keyof Rec
...
+export type { RecordLockOptions } from './recordLock.ts';
+
+/**
+ * What `lock()` resolves with: the locked record, loaded after the lock committed and tracking
+ * changes for `save()`. Inside an explicit `transaction()` scope a scoped lock is released when
+ * that transaction commits or aborts. Outside any explicit scope (ImmediateTransaction context),
+ * a scoped lock persists until `unlock()` is called or the lease expires. A held lock
+ * (`{ hold: true }`) is always released by `unlock()` or its lease, regardless of scope.
+ */
+export type WritableRecord<Record extends object = any> = UpdatableRecord<Record> & {
```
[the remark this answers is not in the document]
kriszyp: Addressed in `faab6941e`: the root `index.ts` type export now includes both `WritableRecord` and `RecordLockOptions`, with compile-time coverage that imports them from the package root. — GPT-5 Codex

resources/Table.ts
kriszyp: Addressed across `faab6941e` and `320ed4091`: a surviving hold now advances from the version actually applied by a successful native commit, including direct/static same-key writes. Failed, skipped, and rolled-back writes do not advance the floor, and ordinary transactions that own no record locks incur no write scan or key lookup. Regression coverage includes the mixed-transaction post-commit save and static same-key write cases. — GPT-5 Codex

resources/Table.ts
```diff
@@ -2322,10 +2429,266 @@ export function makeTable(options) {
...
```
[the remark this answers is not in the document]
kriszyp: Addressed across `faab6941e` and `320ed4091`: the holder floor is updated from the actual committed record version after native commit success, rather than relying on acquisition-time priming. The regression now covers a hold acquired after another staged write and a later post-commit save, plus static same-key commits and rollback/skip non-advancement. — GPT-5 Codex

Issue #2063, opened 2026-08-03
Corrupt-log containment in RocksTransactionLogStore.getRange is per-drain, so one bad entry head-of-line-blocks a replication stream forever


heskew:
## Summary

A corrupt transaction-log entry permanently head-of-line-blocks a replication stream, because the corrupt-entry containment added in `RocksTransactionLogStore.getRange` is scoped to a **single drain**. The `failedIterators` `WeakSet` is constructed *inside* `getRange` and keyed on iterator object identity, so every fresh `getRange` builds new iterators, re-reads the same corrupt log, and re-throws. There is no durable exclusion, no recovery, and no health signal — the stream simply never advances past the bad entry.

Observed on 5.1.x: the same corrupt log re-erroring on a fixed cadence for five days while replication for that database made no forward progress.

## Mechanism

`resources/RocksTransactionLogStore.ts` (`getRange`):

```ts
const failedIterators = new WeakSet<IterableIterator<TransactionEntry>>();   // per-call
...
const safeNext = (iterator, log?) => {
    if (failedIterators.has(iterator)) return { value: undefined, done: true };
    try { ... } catch { failedIterators.add(iterator); ... }
};
```

Within one `getRange` this behaves as intended: the corrupt log is marked failed, subsequent retry-polls skip it, other peer logs keep draining, and the worker stays healthy (the original goal — avoiding an `uncaughtException` out of the aggregate iterator).

Across calls it provides nothing. The `WeakSet` and the iterators both die with the call, so the next drain re-opens the same log, hits the same entry, and logs the same error. Combined with the replay guard — which *stops* iteration at a corrupt entry rather than skipping or excluding the log — nothing after the bad entry is ever delivered.

## Impact

- Replication for the affected database stops advancing permanently. Records written after the corrupt entry never reach the peer.
- The failure is `warn`/`error` log spam only. `cluster_status` still shows the connection `connected: true`; there is no "this stream is wedged" signal, so the condition is invisible to monitoring and can persist for days.
- Recovery today is manual and requires knowing to look at the transaction-log files at all.

## Expected

Corrupt-log exclusion should survive beyond a single `getRange`: track the failed `(log, position)` durably for the store so later drains skip it, and/or advance past the corrupt entry rather than stopping. Either way the condition needs to surface as a health signal — an operator should not have to grep logs to discover that a replication stream has been dead for days.

## Prior discussion

[...]

Issue #2016, opened 2026-07-31
Mid-log corrupt transaction-log frame silently truncates replay and replication — acknowledged writes lost


heskew:
## Summary

A **mid-log** corrupt frame in a table's local transaction log silently truncates every reader of that log — crash-recovery replay *and* replication — at the position of the tear. Because tables run WAL-less by default, replay is the only thing standing between an unclean exit and the loss of acknowledged writes, so a single torn frame converts into:

1. **Unbounded, silent loss of acknowledged writes.** Observed: a crash 2.2 days after the tear rolled a table back to the tear position, dropping 33 acknowledged records written over those 2.2 days.
2. **Indefinite replication starvation.** The peer's reader stops at the same frame, so the peer receives nothing written after the tear — for days — while `cluster_status` and component health stay green.
3. **All of it at `warn` level, latched to fire once per log.**

## The design assumption that breaks

`endIteratorOnCorruptFrame` (`resources/replayLogsGuards.ts`) deliberately treats a corrupt frame as **end-of-log**:

> "Logs (once per log) when a corrupt frame ends a query iterator early; see endIteratorOnCorruptFrame in replayLogsGuards.ts for why this is end-of-log, not a crash."

That is sound when the tear is at the **tail** — a torn final write, nothing after it. It is wrong when the tear is **mid-log**: a failed append (ENOSPC/EDQUOT partial write) that the process survives. The process keeps accepting writes and appending valid frames *after* the tear, every one of them acknowledged to clients — and every one of them unreachable to any future reader.

With `options.disableWAL ??= true` (`resources/databases.ts:163`), those post-tear writes have exactly two homes: the memtable and the txnlog-beyond-the-tear. An unclean exit destroys the first and the tear amputates the second.

## Observed incident (5.1.x, two-node hosted cluster)

- **Day 0:** the instance exhausted its filesystem quota. A torn frame landed mid-log in one table's local transaction log. First evidence, 3 seconds into the recovery-restart boot:
  ```
  [main/0] [warn]: Stopping transaction log "local" at a corrupt entry during replay
    RangeError: Corrupt transaction log entry ...
  ```
- **Day 0 → Day 2.2:** the node ran continuously; 33 records were written and acknowledged (memtable + post-tear txnlog appends; low write volume, so no flush ever triggered).
- **Day 2.2:** a component deploy's restart crashed the process (native-addon segfault during teardown — separate issue in the component's repo). Recovery replay walked the log, stopped at the tear:
  ```
[...]

PR #2463, opened 2026-09-02
Report a transaction log's corrupt-frame stop to getRange callers through an onCorruptFrame hook

kriszyp:
`RocksTransactionLogStore.getRange` now exposes an [`onCorruptFrame(error, logName)` hook](https://github.com/HarperFast/harper/pull/2463/changes?w=1#diff-56c43863dc5f281e19328e36b0082523391cfb5373c0bd88a8d3d94bfa1ab46cR40) when a transaction-log iterator stops at a corrupt frame, giving replication consumers the signal they need to recover instead of parking on a dead iterator. The hook is composed with main's corruption stop tracking and telemetry [at the existing latch](https://github.com/HarperFast/harper/pull/2463/changes?w=1#diff-56c43863dc5f281e19328e36b0082523391cfb5373c0bd88a8d3d94bfa1ab46cR298), while aggregate delivery remains safe under re-entrancy and keeps mutable exclusions private to each range. The whole contract — who reports, who owns recovery, and how a hook's `removeLog`/`addLog` interact — is [recorded as a row in `resources/DESIGN.md`'s read-path table](https://github.com/HarperFast/harper/pull/2463/changes?w=1#diff-5a3cac25d02c21deca87f353afcbe03c11dd1601a64cfb947c57219f7334a708R103), which is where the next reader of this code will look.

Independent review also identified a pre-existing `addLog` convergence issue: re-including an excluded peer does not itself invalidate `latestUpdates`, so that peer waits for another log-list update before receiving an iterator. It is outside this PR's change and is recorded as a follow-up finding.

Four later commits deal with CI, which was red for reasons that had nothing to do with the hook. The branch had drifted into a merge conflict with `main`, and GitHub builds no merge commit for a conflicting PR, so no `pull_request` workflow ran at all on it — the red checks were the stale ones from before. It is rebased onto `main` here, and three things came out of that:

- **The Next.js adapter job's install phases are now bounded.** Those three jobs were cancelled at their 30-minute job cap inside `Install fixture dependencies` on a day when npm installs that normally take 20-50 seconds took 3.5-18 minutes. Its two existing step caps (Playwright 10m, tests 15m) left only 5m for four uncapped npm phases ahead of them, so the job cap was always reached first. Each network-bound phase is now capped at roughly 10x its measured healthy time and the [job cap sits clear of their sum](https://github.com/HarperFast/harper/pull/2463/changes?w=1#diff-7d35796e0cec9f05ff7ed8b28267e964a8937befa7e31bd9ca98fa1040457de5R67). The Linux unit job needed the same treatment and got a better one in #2501 while this branch was behind; that fix arrives with the rebase, and nothing here touches `unit-test.yml`.
- **The PR's own tests were stale against #2497's audit-clock split.** That PR made `version` the record's own version and `txnLogKey` its key in the transaction log, and updated the one corrupt-entry drain test it had. These tests predate it and still read `record.version`; their synthetic entries carry only a timestamp, so all thirteen collected `undefined` after the rebase. They [read `txnLogKey`](https://github.com/HarperFast/harper/pull/2463/changes?w=1#diff-044077436ab82ecf6c481205bdf0ac84571b38edf4ce663842e72349c3a7b38cR1491) now.
- **Two of them depended on process-global state.** [Corrupt-frame reports are cleared before each case](https://github.com/HarperFast/harper/pull/2463/changes?w=1#diff-044077436ab82ecf6c481205bdf0ac84571b38edf4ce663842e72349c3a7b38cR1426) so the rejecting-async-hook test no longer needs a sibling to have suppressed the report first, and the logger-failure case counts only the reports naming its own log rather than every logger call, since the decoder's undecodable-header warning is latched per process. Both failed under `--grep` or `.only` before; both pass in isolation now.

Refs #2016, #2063. The paired engine work is HarperFast/rocksdb-js#817; harper-pro owns the policy that turns a mid-log report into a bounded base-copy resync.

## For the human reviewer

1. **Core reports the gap; the consumer repairs it.** The alternative is for core to seek past the break or initiate resync itself. Core cannot assign one recovery policy to boot replay, replication, and subscriptions, while harper-pro has the source and bounded-copy machinery. Reversing this later is invasive because recovery ownership crosses repositories; rejecting it now requires moving the policy and its end-to-end proof into core.
2. **The hook is synchronous inside `next()`.** An event or microtask would be simpler, but the replication sender must learn why the drain stopped before it parks on a future transaction. [Reports wait until the aggregate buffer is stable](https://github.com/HarperFast/harper/pull/2463/changes?w=1#diff-56c43863dc5f281e19328e36b0082523391cfb5373c0bd88a8d3d94bfa1ab46cR371), and [iterator splices wait for the outermost poll](https://github.com/HarperFast/harper/pull/2463/changes?w=1#diff-56c43863dc5f281e19328e36b0082523391cfb5373c0bd88a8d3d94bfa1ab46cR469). Changing this later is behaviorally breaking; rejecting it now requires redesigning the sender notification contract.
3. **Promise-returning hooks are fire-and-forget.** Rejections are contained and logged, but async recovery cannot gate a synchronous iterator. The alternatives are a synchronous-only signature or an asynchronous iterator contract. This is cheapest to change before adoption; rejecting it requires choosing one of those narrower APIs.
4. **Exclusions belong to the range, not the caller's options object.** [`removeLog` records the private exclusion immediately but defers structural mutation](https://github.com/HarperFast/harper/pull/2463/changes?w=1#diff-56c43863dc5f281e19328e36b0082523391cfb5373c0bd88a8d3d94bfa1ab46cR499), so frozen options work and a hook's exclusion is visible to a nested refresh. The alternative is retaining the caller's array as live shared state. Reversal is local but restores hidden mutation and the frozen-options failure.

5. **The Next.js caps are sized from measured step timings, and ride on this branch rather than their own PR.** Network-bound phases get roughly 10x their healthy time because they degrade multiplicatively — a degraded registry stretches an install by a factor, not by a constant. The alternative is tighter fail-fast caps; the cost of the sizing chosen is that a genuinely stuck install burns up to its cap before a named step fails, against a job cap that costs no runner minutes unless it is hit. On placement, the alternative is a standalone workflow PR — cleaner to bisect, but this PR is where the failure was visible. It is one self-contained commit, so say the word and I will cherry-pick it out.
6. **Of the three review findings on the hook, two are fixed and one is left for you.** Still open, and adjudicated to nit: the aggregate path fires the hook in the `next()` that then returns the broken log's last intact entry, while the single-log path fires on the following `next()`. Neither ordering is wrong and the error direction is redelivery rather than a skipped entry, but the contract does not state which, so a hook that checkpoints "everything delivered so far" records a cursor one entry short on the aggregate path only. Fixed instead of carried: [a deferred `removeLog` now replays only while the name is still excluded](https://github.com/HarperFast/harper/pull/2463/changes?w=1#diff-56c43863dc5f281e19328e36b0082523391cfb5373c0bd88a8d3d94bfa1ab46cR491), so `removeLog(X); addLog(X)` inside one hook invocation keeps X rather than silently dropping the rest of that log — [covered by a test](https://github.com/HarperFast/harper/pull/2463/changes?w=1#diff-044077436ab82ecf6c481205bdf0ac84571b38edf4ce663842e72349c3a7b38cR1650) that loses entries 2, 3 and 4 against the previous behavior. And [the `CorruptFrameHook` contract](https://github.com/HarperFast/harper/pull/2463/changes?w=1#diff-56c43863dc5f281e19328e36b0082523391cfb5373c0bd88a8d3d94bfa1ab46cR33) no longer claims "once per log per iterable": it is once per mounted log, and it no longer implies `addLog` on its own remounts a removed log, which it cannot — that is the pre-existing `latestUpdates` gap named above.

## Verification

- `npm run build` — passed.
- `npx mocha unitTests/resources/auditLog.test.js --grep "onCorruptFrame|removeLog remembers|hook exclusion"` — 14 passing, including frozen options across refresh, a future peer excluded before it exists, hook-time exclusion before a nested refresh, sync/async failure containment, and [a real on-disk torn-frame report](https://github.com/HarperFast/harper/pull/2463/changes?w=1#diff-044077436ab82ecf6c481205bdf0ac84571b38edf4ce663842e72349c3a7b38cR1684).
- `npm run test:unit:resources` — 2,203 passing, 0 failing on the pushed head (exit 0). Before the rebase resolution it was 2,190 passing / 13 failing, all thirteen the `record.version` staleness above.
- `npx mocha unitTests/resources/auditLog.test.js` — 94 passing, 8 pending, 0 failing on the pushed head, with `main`'s audit tests and this PR's merged.
- `npx mocha unitTests/resources/auditLog.test.js --grep onCorruptFrame` — **14 passing in isolation, including the on-disk torn-frame test.** No review leg could execute that one (all of them run read-only without a shell), so it is worth saying plainly: the real rocksdb-js 2.8.0 reader does surface `CorruptFrameError` with `resyncPosition` through the hook, both standalone and in file order. The two order-dependent cases also pass alone now; before the fixes they failed with `The input did not match the regular expression /onCorruptFrame hook failed/` and `3 !== 2` respectively.
- Next.js caps derived from step-level timings on real runs, not estimated. Healthy is 4-5m end to end — adapter install 31-47s, harper override 42-67s, fixtures 35-57s, Playwright 23-37s, tests 99-131s — consistent across runs [34170623123](https://github.com/HarperFast/harper/actions/runs/34170623123), [34119563615](https://github.com/HarperFast/harper/actions/runs/34119563615), [33944768904](https://github.com/HarperFast/harper/actions/runs/33944768904) and [33912991544](https://github.com/HarperFast/harper/actions/runs/33912991544). The cancelled run [33845099465](https://github.com/HarperFast/harper/actions/runs/33845099465) spent 297s/869s/>1114s in its three then-unbounded npm phases and never reached Playwright.
- `npx prettier --check` on the workflow and the test file — passed; the workflow parses to the intended caps (job 60 > 8 + 12 + 10 + 10 + 15 = 55).
- `npm run build` on the rebased head — passed.
- **CI at the rebase commit, earlier in this branch's history: 41 pass, 3 fail.** All three Next.js adapter jobs pass in 4m55s-5m32s, the jobs that were being cancelled at 30 minutes. The three red checks are `Unit Test (Node.js v22/24/26)`, and they are `main`'s, not this PR's: [run 34124860863](https://github.com/HarperFast/harper/actions/runs/34124860863) on `main` at 38fcbf1ca — the exact commit this branch is rebased onto — fails `test MQTT connections and commands > subscribe to retained record with patch operations` at `unitTests/apiTests/mqtt-test.mjs:503` with `'update 3' == 'update 2'`, and all three jobs here fail on that same assertion in the same `HARPER_STORAGE_ENGINE=lmdb` leg. This PR touches no MQTT, apiTests, or server code, and its own legs are green in CI: `test:unit:main` 5,413 passing, `test:unit:apitests` 201 passing, `test:unit:resources` 2,203 passing. Worth a separate issue; deliberately not fixed here.
- **Maintenance push, 2026-09-09 (`f416489e4` → `72d80cd9d`, no rebase).** `Format Check` was failing deterministically on `resources/DESIGN.md`: the table row this PR adds was not column-padded, which `prettier --check` rejects (`main`'s copy of the file is clean). Re-run after the fix: `npm run format:check` clean repo-wide, `npm run build`, `npm run test:types`, `npx mocha unitTests/resources/auditLog.test.js` 95 passing / 0 failing, `npm run test:unit:resources` 2,256 passing / 0 failing. **CI on `72d80cd9d` is fully green — 44 pass, 3 skipped, 0 fail**, including the `Unit Test (Node.js v22)` leg and all three Next.js adapter jobs.
- **`Unit Test (Node.js v22)` was red on a flake that predates this branch.** `longLivedTransactions.test.js` — "names a chain link reachable only through the root under its own native id" — expects `/state: [^,]*active/` and gets `state: source-apply`. The same assertion fails on `main`'s own run at this branch's merge-base ([run 34280999623](https://github.com/HarperFast/harper/actions/runs/34280999623) at 3a3c0ef48), and nothing since has touched that test or `DatabaseTransaction.ts`. Root cause is the per-pass report cap, not timing noise: a holder skipped once a pass has emitted its 10 lines is retried on a later tick, but `active` is in the state list only while `monitorTick - link.writeTick <= 1`, so a deferred report has already lost it. Confirmed locally by forcing that cap to 1 — the file passes 3/3 alone at the real cap and fails immediately at 1 with the exact CI message. It reproduced in 1 of 6 full `test:unit:resources` runs here and never when the file runs alone. Filed as a follow-up rather than fixed in this PR.
- Regression controls — the frozen-options test fails against the prior mutating implementation with `TypeError: object is not extensible`; [the nested-refresh test](https://github.com/HarperFast/harper/pull/2463/changes?w=1#diff-044077436ab82ecf6c481205bdf0ac84571b38edf4ce663842e72349c3a7b38cR1509) fails before immediate exclusion registration because the excluded future peer is polled twice.
- harper-pro `integrationTests/cluster/txnlogTearReplication.test.mjs` — prior paired-branch evidence: base parks at 39/60 rows; this hook plus the sender's bounded base copy converges to 60/60 and accepts a post-recovery row.
- `npm run test:integration:all` — began successfully and passed numerous unrelated scenarios, then the runner saturated and multiple independent instances timed out together at 120 seconds; stopped after the synchronized infrastructure failures. `npm run test:unit:main` and `npm run test:unit:windows` likewise stopped producing a summary, with the Windows slice also exposing an unrelated ambient credential-socket failure.

Complexity: complicated

<sub>Review-Coverage: authored=claude; ran=codex,gemini; declined=cursor-grok,cursor-composer,domain; rounds=17 @ 72d80cd9dba7</sub>

<sub>Human-Review-Need: 3 @ 72d80cd9dba7</sub>


resources/RocksTransactionLogStore.ts
```diff
@@ -389,7 +476,7 @@ export class RocksTransactionLogStore extends EventEmitter {
 						logs.splice(index, 1);
 						iterators.splice(index, 1);
 						nextEntries.splice(index, 1);
-						options.excludeLogs.push(logName);
+						(options.excludeLogs ??= []).push(logName);
```
cb1kenobi: **Immutable options can break corrupt-frame containment — Medium.** `removeLog()` now assigns `excludeLogs` onto the caller-owned options object. If a caller passes `Object.freeze({ onCorruptFrame })` and the hook removes the corrupt log, this `??=` throws from `guardedNext()`'s `finally`, replacing the iterator result with an exception and aborting recovery. Keep a mutable per-iterable exclusion array initialized from `options.excludeLogs` instead of adding a property to `options`.

—
Reviewed 85b58485

kriszyp: Addressed in 77765108a, with frozen-options and refresh regression coverage in 82ec7d72e and the re-entrant/future-peer cases in 11ff60cf3 and e6d26bd13. `getRange` now copies the caller exclusions into lazily-created per-iterable mutable state; `removeLog` never mutates the caller options or array, retains exclusions before a peer appears, and exposes the exclusion before a hook can trigger a nested refresh.

cb1kenobi (approved):

Issue #2450, opened 2026-09-01
Outstanding-commit wedge recurs on 5.2.6/5.2.7: native commits never settle and checkOverloaded 503s every application write on the thread until restart


Devin-Holland:
## Summary

On a production Harper Fabric control-plane node running harper-pro **5.2.6**, three worker threads each had a native write commit that **never settled**. `checkOverloaded()` then rejected every new application write on those threads with the 503 `Outstanding write transactions have too long of queue, please try again later` for **~12.5 minutes**, until the process was restarted (an upgrade to 5.2.7). Nothing else was logged: no RocksDB stall, no long-reader release, no event-loop stall warning at onset, no `MDB_`/`ERR_BUSY` error. Host resources were healthy throughout (37% memory, 3.5% CPU, 18% disk).

This is the harper#2001 family recurring **with the #2001 fix present**: `trackOutstandingCommit()` correctly *detected* and *named* the stuck commits (that part of #2001 works), but a commit that never settles is still never aborted, timed out, or diagnosed — the thread stays wedged until a restart. 5.2.6 and 5.2.7 are identical on this path (`rocksdb-js 2.7.1`, `abandonWrites` present in both), so upgrading does not prevent recurrence.

## Timeline (UTC, 2026-09-01)

- **20:25:03** — the three commits are submitted (each is timed from its own submission).
- **20:25:48** — the guard fires on three threads within ~100ms, one stuck commit each:

```
[http/9]  [error]: Rejecting writes on this thread: a commit has been outstanding for 45019ms (exceeds the 45000ms limit), from table: data.Host/ (transaction 1724751), started from Host.patch. Further record updates and publishes from new application requests on this thread will be rejected with 503 until the commit settles or the process is restarted (...)
[http/10] ... from table: data.InstanceMonitoring/ (transaction 1724755), started from InstanceMonitoring.patch ...
[http/8]  ... from table: data.Host/ (transaction 1724822), started from Host.patch ...
```

- **20:25:48 → 20:38:15** — every application write on http/8, /9, /10 is rejected. The rejection surfaced to the app as an `unhandledRejection` per write (the app's own bug, fixed separately), ~14K lines in the hour:

```
[http/10] [error]: unhandledRejection in worker thread 10: Error: Outstanding write transactions have too long of queue, please try again later
    at DatabaseTransaction.checkOverloaded (core/resources/DatabaseTransaction.ts:718:10)
    at Object.validate (core/resources/Table.ts:2343:20)
[...]

Issue #768, opened 2026-05-25
Test123


kriszyp:
body

kriszyp:
Test issue created in error; cleaning up.

PR #2466, opened 2026-09-02
Bump rocksdb-js to 2.8.0 on v5.2 so a commit parked on a never-released write intent stops wedging the worker thread

kriszyp:
rocksdb-js 2.8.0 bounds the coordinated-retry park (rocksdb-js#744): a commit waiting on a write intent whose holder never releases now wakes after `ROCKSDB_JS_PARK_TIMEOUT_MS` (5000ms default, unparseable/zero/negative values fall back to it, no way to disable) and consumes a retry attempt exactly as a real release would, instead of never settling. That never-settling park is the root cause of #2450 — the commit that stayed outstanding for 12.5 minutes on three worker threads until the process was restarted.

This is a dependency-pin change only; no core code changes. It moves two pins together: [rocksdb-js to 2.8.0](https://github.com/HarperFast/harper/pull/2466/changes?w=1#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R182) and [msgpackr to 2.0.6](https://github.com/HarperFast/harper/pull/2466/changes?w=1#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R226). `build-tools/check-shrinkwrap-pins.mjs` requires the root `msgpackr` pin to equal rocksdb-js's own so the tree carries one msgpackr instance (and so one structure/extension registry), and rocksdb-js 2.8.0 depends on 2.0.6; the docker smoke job caught the mismatch on the first push. The lockfile correspondingly loses the nested `msgpackr` copy under rocksdb-js. 2.0.6 over 2.0.5 is one decode-hardening commit — reject array and map lengths that exceed the remaining source data — and is the pairing `main` already runs. v5.2 already carries the ladder the woken commit lands in: `MAX_RETRIES = 40` with `abortChainAfterRetries()`, `abandonWrites()` on a retained read handle, and per-attempt outstanding-commit tracking. Because `trackOutstandingCommit()` stamps each *attempt* at submission, every attempt now settles inside the park timeout, so a stuck intent holder no longer pushes the oldest outstanding commit past `storage.maxTransactionQueueTime` (45s) and `checkOverloaded()` stops 503-ing every unrelated write on the thread. The thread recovers without a restart, which is the part of #2450 that hurt.

## For the human reviewer

1. **The whole 2.8.0, or a 2.7.2 with only rocksdb-js#744?** 2.7.1 → 2.8.0 is 186 commits of mostly native change: the RocksDB prebuild jump 11.1.2 → 11.8.1, transaction-log recovery and framing rework (rocksdb-js#723, #750, #751), latched background-error observation and recovery (#758), the #741 pending-transaction close-path rework (#780), the verification-table non-unique-version change (#766), and dropped-transaction snapshot release (#768). That is more native surface than a patch line usually takes at once, and 2.8.0 has about a week of soak on `main` only. The minimal alternative is cutting rocksdb-js 2.7.2 = 2.7.1 + #744 and pinning that here instead. Either way the decision is a one-line pin and trivially reversible, but it is the decision this PR is asking for — @kriszyp asked for the 2.8.0 route explicitly.
2. **#2459 is deliberately not backported.** It bounds the retry ladder by queue time and returns a distinct 503 `TRANSACTION_COMMIT_CONFLICT_TIMEOUT`. On 2.7.1 it would be inert for #2450 (a commit parked forever never reaches a retry decision, so the deadline is never evaluated), and on 2.8.0 it only shortens the tail. Without it, a permanently-held intent ends the initiating request after roughly 40 × 5s ≈ 200s with the existing generic 500 "After 40 coordinated retries, unable to commit transaction" rather than a bounded, typed 503. If the reporter needs the bounded error too, #2459 can be backported after it merges on `main` — it does not change the wedge outcome.
3. **No companion harper-pro PR.** harper-pro pins `@harperfast/rocksdb-js` and `msgpackr` in its own `package.json`, but `build-tools/sync-core.sh` copies core's `dependencies`, `devDependencies`, `overrides` and `optionalDependencies` plus core's lockfile into harper-pro wholesale, so both pins arrive there from this PR through the core sync. An earlier hand-written pro PR for the same bump (harper-pro#798) was closed as redundant.
4. **Direction:** this targets `v5.2` directly rather than arriving as a cherry-pick — `main` has been on 2.8.0 since #2065, which bumped it for the verification-table non-unique flag.

## Verification

- `npm run build` — clean.
- `npm run test:unit:dataLayer` — 242 passing, 0 failing.
- Dependency alignment, checked directly against the three predicates `verifyRocksDbDependencyAlignment()` asserts: root pin, rocksdb-js's own pin and the installed version all read `2.0.6` for `msgpackr` and `1.0.3` for `@harperfast/extended-iterable`, and no nested copy of either remains under `node_modules/@harperfast/rocksdb-js/`. The checker itself cannot run locally — it reads `npm-shrinkwrap.packed.json`, which only exists inside the packed docker image the smoke job builds.
- `npm run test:unit:resources` — 1698 passing, 0 failing, re-run after the msgpackr pin moved (same counts). A first run of the same suite had one failure, `Caching > Can load cached indexed data` (asserted 3, got 2); the file passes in isolation and the full-suite re-run is green, so it is a pre-existing ordering flake, not a bump regression.
- `npm run test:unit:main` — **not run locally.** The suite stalled on the storage locks shared with two other local Harper suites running concurrently in sibling worktrees (a known local-only contention on this machine, not a product behavior); it was killed rather than reported. CI runs it on this PR.
- **Not observable end-to-end here:** reproducing the wedge requires deliberately leaking a native write intent that is never released. The park timeout itself is covered inside rocksdb-js (`test/fixtures/fork-park-timeout.mts`, asserting both a lower and an upper bound on elapsed time so the timeout branch is distinguishable from the not-parked fast path), and the retry ladder the woken commit lands in is already covered on this branch.

Refs HarperFast/harper#2450

Complexity: easy

<sub>Review-Coverage: authored=unknown; ran=none; rounds=1 @ 6ca418d74197</sub>

<sub>Human-Review-Need: 4 @ 6ca418d74197</sub>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -61,7 +61,7 @@
 				"minimist": "1.2.8",
 				"moment": "2.30.1",
 				"mqtt-packet": "~9.0.1",
-				"msgpackr": "2.0.5",
+				"msgpackr": "2.0.6",
 				"needle": "3.5.0",
 				"node-forge": "^1.3.1",
 				"node-stream-zip": "1.16.0",
@@ -2659,15 +2659,6 @@
 			],
 			"engines": {
 				"node": "^22.18.0 || >=24.0.0"
-			}
-		},
-		"node_modules/@harperfast/rocksdb-js/node_modules/msgpackr": {
-			"version": "2.0.6",
-			"resolved": "https://registry.npmjs.org/msgpackr/-/msgpackr-2.0.6.tgz",
-			"integrity": "sha512-plGul/tqjt9vqWFR9zyqyLZls6gb5KTLvnsRO2B9+TZ8tNiXiVI/CR8LiDWlAX4IUeVkQ9gsmzKA6voC2QOciw==",
-			"license": "MIT",
-			"optionalDependencies": {
-				"msgpackr-extract": "^3.0.4"
 			}
 		},
 		"node_modules/@harperfast/skills": {
@@ -11556,9 +11547,9 @@
 			"license": "MIT"
 		},
 		"node_modules/msgpackr": {
-			"version": "2.0.5",
-			"resolved": "https://registry.npmjs.org/msgpackr/-/msgpackr-2.0.5.tgz",
-			"integrity": "sha512-cef05H/dSYpLpqp3sj/qyZh5vhUYCalnaLO7j1yOmpsR0y/XwLVtK7r5gn+U/F7CTEfMowcGhlUQJDLcLf7jcA==",
+			"version": "2.0.6",
+			"resolved": "https://registry.npmjs.org/msgpackr/-/msgpackr-2.0.6.tgz",
+			"integrity": "sha512-plGul/tqjt9vqWFR9zyqyLZls6gb5KTLvnsRO2B9+TZ8tNiXiVI/CR8LiDWlAX4IUeVkQ9gsmzKA6voC2QOciw==",
 			"license": "MIT",
 			"optionalDependencies": {
 				"msgpackr-extract": "^3.0.4"
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
 				"@fastify/cors": "^11.2.0",
 				"@fastify/static": "^9.1.3",
 				"@harperfast/extended-iterable": "1.0.3",
-				"@harperfast/rocksdb-js": "2.7.1",
+				"@harperfast/rocksdb-js": "2.8.0",
 				"@harperfast/skills": "^1.10.8",
 				"@turf/area": "6.5.0",
 				"@turf/boolean-contains": "6.5.0",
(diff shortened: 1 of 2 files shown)


kriszyp:
@cb1kenobi I believe you felt like 2.8.0 was probably reasonable to apply to 5.2.x. I know we could do a more targeted 2.7.2, but I believe 2.8.0 has a gotten some decent testing, and seems reasonable to adopt, if you still agree.

kriszyp:
I think we need to move ahead with this, Fabric is kind of waiting for this fix.

Issue #2471, opened 2026-09-02
A transaction holding staged writes stays open for hours and is never reaped or identified (the holder behind the #2450 wedge)


Devin-Holland:
## Summary

Behind the #2450 wedge is a second defect that rocksdb-js 2.8.0 (#2466) does not touch: **a transaction with staged writes stayed open for hours on a production CM node, holding a verification-table (VT) write intent that nothing reaped, logged, or identified.** On 5.2.7 that held intent parked other writers indefinitely (the wedge). On 5.2.8 the park is bounded (rocksdb-js#744), so the same holder now costs a ~5 s stall on every colliding write and, if it also pins a RocksDB snapshot, blocks version reclamation until restart (the #2107 shape) — and it is still invisible.

Field evidence, 2026-09-01 and 2026-09-02, two nodes, 5.2.6/5.2.7 (details on #2450): commits parked for 4.5 h and 9 h; disk and CPU idle; no RocksDB `LOG` activity at either onset; replication-applied writes into the same database stalled at the same instant (database-wide, so the holder's slot collided with keys of several tables); **not one `Transaction was open too long and has been aborted` line on either node all day**, so the long-transaction monitor never reaped it.

## Why nothing reaps or names it today

- `resources/DatabaseTransaction.ts` (monitor): a write-bearing transaction past the 30 s idle limit is aborted **except when `sourceApply` or `isReplay`** — those keep the prior force-commit behavior and are never aborted, so a replication-apply transaction stuck mid-apply holds its intents for as long as it is stuck.
- rocksdb-js: a park (2.7.1) or a park timeout (2.8.0) records nothing about the holder. rocksdb-js#744 explicitly declined a `transaction:parkTimeout` event, so on 5.2.8 a permanently held intent surfaces only as the initiating write failing after `MAX_RETRIES × 5 s` with the generic coordinated-retries error.
- `checkOverloaded()` logs the *victim* (thread, table, `startedFrom`) once per stuck commit — never the transaction it is parked behind.
- rocksdb-js#768 (GC-release of dropped transactions) and #780 (reap at worker-env exit) cover two leak shapes; a holder that is **alive and referenced but stuck** (awaiting I/O, a blob stream, a lock) is covered by neither.

## Reproduction

[...]

PR #2473, opened 2026-09-02
Report the long-lived transaction holder behind a wedged commit

kriszyp:
A RocksDB transaction that holds staged writes — and with them verification-table write intents — can stay open indefinitely with nothing in core naming it. On 5.2.7 such a holder parked other writers forever (the #2450 wedge); on 5.2.8 the park is bounded, so the same holder costs ~5s on every colliding write and, if it also pins a snapshot, blocks version reclamation until restart. Two field incidents held for 4.5h and 9h across two nodes with not one log line identifying the holder.

Nothing reported it because each observer has a blind spot: the monitor's `sourceApply`/`isReplay` branch force-commits and logs nothing at all, its commit-phase sparing warns but carries no native transaction id to correlate on, the stuck-commit logs name only the victim commit and never what it is parked behind, and a handle opened outside a supervised transaction is in neither monitor registry at all.

This adds three reporting surfaces joined by the native transaction id and leaves reaping behavior unchanged. The load-bearing fact is that rocksdb-js's registry is **process-global, not thread-local** — verified by probe and now pinned by a test that creates a handle only inside a worker and reports it from the main thread — so [one main-thread sweep](https://github.com/HarperFast/harper/pull/2473/changes?w=1#diff-1b87a1fd0c7e3de0eb9039837e4af1126c18cc3fa04b725fbc12c1548ae43719R118) sees every worker's handles, including ones core never tracked. That is the only surface that can see the untracked class. [The per-thread monitor](https://github.com/HarperFast/harper/pull/2473/changes?w=1#diff-2ae62080514cb4ef64bf50088e7ea0587701f558af71f76b119f07e00f3d56beR1941) adds what the sweep cannot: thread, database/table, `startedFrom`, staged write count, and which state kept it from being reaped — walking the whole `.next` chain, so a link that owns its own handle is named under its own id rather than the root's. Both stuck-commit logs — [`checkOverloaded()`](https://github.com/HarperFast/harper/pull/2473/changes?w=1#diff-2ae62080514cb4ef64bf50088e7ea0587701f558af71f76b119f07e00f3d56beR881) and #2459's [`abandonCommitAfterDeadline()`](https://github.com/HarperFast/harper/pull/2473/changes?w=1#diff-2ae62080514cb4ef64bf50088e7ea0587701f558af71f76b119f07e00f3d56beR1643) — append the live handles that could hold the intent they are parked on.

Threshold is [`storage.longTransactionReportThreshold`](https://github.com/HarperFast/harper/pull/2473/changes?w=1#diff-1b87a1fd0c7e3de0eb9039837e4af1126c18cc3fa04b725fbc12c1548ae43719R92) (default `5m`, `0` disables all three); [changing it clears the accrued backoff](https://github.com/HarperFast/harper/pull/2473/changes?w=1#diff-1b87a1fd0c7e3de0eb9039837e4af1126c18cc3fa04b725fbc12c1548ae43719R71) so a handle already under observation is re-measured against the new value.

The rebase follow-up restores two test-seam docblocks and makes the tests deterministic: [the worker sweep waits for its own database path](https://github.com/HarperFast/harper/pull/2473/changes?w=1#diff-27b6b8d8f6591d28d485ce8379eab39e98a631cb2a1a4be793f665a0d25e950dR698), [the chain fixture sets the actual root transaction's budget before its first write](https://github.com/HarperFast/harper/pull/2473/changes?w=1#diff-27b6b8d8f6591d28d485ce8379eab39e98a631cb2a1a4be793f665a0d25e950dR609), and [the abandonment test asserts the candidate suffix alongside the 503](https://github.com/HarperFast/harper/pull/2473/changes?w=1#diff-b18acce06cb2dd26d6a1cbaa2777a159b8bc98caf33ef7d94e0dfcbcb7d003acR147). The root-budget assignment is where to look hardest: it must precede both writes so the chain link inherits the 5s test budget.

## For the human reviewer

1. **Every handle over the threshold is reported at WARN, including a long read-only scan or export** that pins a snapshot but holds no write intent. The alternative was reporting only write-intent-bearing handles, or dropping read-only ones to INFO (invisible at the default `warn` level). I chose breadth because #2107 is exactly a read-pinned snapshot blocking reclamation, and because a read handle surviving 5m is already anomalous — `storage.maxTransactionOpenTime` is 30s and the monitor commits over-limit read-only transactions to close out the snapshot. **Cost of "no":** steady WARN traffic on a healthy node doing long exports trains operators to ignore the one line this issue exists to make them read. Reversible by changing a predicate.
2. **A new permanent public config key** rather than deriving from `storage.maxTransactionOpenTime` or folding under `storage.debugLongTransactions`. Those two govern *reap policy* and a *debug-only stack capture*; this governs reporting, and needs its own noise/disable semantics — an operator tightening it during an incident must not thereby change what gets aborted. **Cost of "no":** a config key is API and needs a docs-repo entry; removing one later is breaking.
3. **[Holder candidates are drawn from every database, not just the stuck commit's own](https://github.com/HarperFast/harper/pull/2473/changes?w=1#diff-1b87a1fd0c7e3de0eb9039837e4af1126c18cc3fa04b725fbc12c1548ae43719R253)**, ranked with the commit's own database first and a foreign one labelled with its path. The verification table is one process-global slot array whose hash mixes in the database id, so a holder in `system` or `oauth` parks a `data` commit at the same 1-in-131072 rate another `data` key would — and the same-database filter printed nothing at all for that shape. The oldest foreign candidate is given the last of the three slots when this database would otherwise fill all of them — without that, a busy `data` database's own three handles push the actual holder into the "and N more" count, which is exactly the case the cross-database search exists for. **Cost of "no":** a wider list could send an operator to the wrong database. Cheap to narrow; `MAX_HOLDER_CANDIDATES` bounds the line either way.
4. **Backoff doubles the next report to 2× current age** rather than a fixed interval. Keeps a 9h holder to a handful of lines and is what rotates the per-pass cap across the whole set. **Cost of "no":** an operator who starts tailing at hour 3 of an 8-hour wedge may wait hours for the next line. One-line change.
5. **Three test seams ship as production exports** (`setRegistryStatusForTests`, `resetLongLivedTransactionReportsForTests`, and `setMaxOutstandingTxnDuration` on `DatabaseTransaction`). Repo precedent exists (`setTxnExpiration`, `resetReplayedWritesWarning`, `trackOutstandingCommit`). The third is what makes the stuck-commit log testable at all — without it a test must wedge a commit for the real 45s. The review flagged that it lacks the `ForTests` suffix its two neighbours carry: it matches this file's existing `setTxnExpiration` convention but not the new module's, and renaming is free now and awkward after release.
6. **Sweep placement: kept on the main thread, armed from `startHTTPThreads`.** Two reviewers asked for a change and I am declining both, on the code rather than the review text. Barber AI's blocker says job-worker and CLI *processes* never sweep — but `launchJobThread` calls `threadsStart.startWorker()`, which is `new Worker(...)`: job workers are **threads in the process that already armed the sweep**, and the registry is process-global, so their handles are already reported. What is genuinely uncovered is a separate CLI process (`bin/backup.ts`, `bin/copyDb.ts`) holding a handle past the threshold — narrow, and not this issue's shape. Devin's stalled-main-thread argument is the stronger one, and is the reason to revisit if it recurs, but it is self-described as unproven (`disableStdio()` on EPIPE/EIO is the competing explanation) and the per-thread attribution surface runs on every worker regardless, so a stalled main thread costs only the untracked-handle sweep.
7. **Deliberately not done: the reap-policy half of the issue's ask 3**, now tracked as #2495. `abandonWrites()` is documented as barring any later commit, and a `sourceApply`/`isReplay` transaction has no resume path, so releasing its intents drops the write while the cursor advances past it — permanent divergence, which is why harper-pro#348 created the exemption. The only variant that preserves the write is abandon-then-replay onto a fresh transaction, which cannot run while the transaction is parked in its pre-commit await. And #2471 states the production trigger is still unidentified, so this ships the "at minimum they should be *reported*" floor and leaves the policy to be driven by what these logs identify. Linked as `Refs #2471` rather than `Closes` for that reason.
8. **Known noise ceiling I did not cap:** the sweep's `registryStatus()` call materializes the whole snapshot, so enumeration is O(live handles) once per minute on the main thread. Capping the loop after the call saves nothing — the snapshot is the cost — and bounding it properly is a rocksdb-js-side "return only over-age handles" API. The pass no longer allocates per sub-threshold handle, and log formatting is capped at 10 per pass plus an omitted count.
9. **Knowingly declined from the pre-push review:** the stuck-commit tests drop the thread-wide shed limit to 1ms while they poll, so a background write on that thread in the window would get a 503 it has no handler for. Narrowing it needs the log cooldown exposed as another production test seam, which is more surface than the nit is worth. Also declined: passing the monitor's already-read threshold into `reportLongLivedHolder` (reached only for links already past the threshold, and it would put a parameter on a public export for a config-map lookup).
10. **Also declined from review:** the commit-phase sparing warn still carries no native transaction id, so for a `sourceApply` apply parked in pre-commit blob I/O an operator gets table/path-only warns from ~30s until the 5m attribution line appears. That gap is named in this PR's problem statement as motivation, not as something it closes; adding the id there is a separate change to a pre-existing log line. And the "N further handles" summary has no backoff of its own, so a sustained leak re-emits it once a minute — one line per minute during an active leak reads as the intended "there is more" signal rather than noise.
11. **Docs.** `storage.longTransactionReportThreshold` is user-facing and owes a HarperFast/documentation companion before this leaves draft; not opened here.
12. **The green "Cherry-pick onto v5.2" check does not mean the cherry-pick is clean.** That job succeeds by committing the markers to `cherry-pick/v5.2/pr-2473` and gating the release-branch integration tests; the branch currently carries four conflict regions in `resources/DatabaseTransaction.ts` and two in `resources/DESIGN.md`. `origin/v5.2` does not contain #2459, so the `checkOverloaded()` log block conflicts and `abandonCommitAfterDeadline()` — which this PR appends the holder line to — does not exist on that line at all. Resolving it is a `/patch-pr` job after merge, not a fix on this branch.

## Verification

Route: resource unit tests, plus real cross-thread and multi-database transactions as the end-to-end proof. The reporting path has no request-level endpoint; the new configuration key remains a documentation-companion decision while this PR is draft.

- `npm run build` — passed.
- `npm run test:unit:resources` — 2,164 passing, 28 pending (3m), including the full-suite cleanup checks after the tests temporarily replace global logger and registry hooks.
- `npx mocha unitTests/resources/longLivedTransactions.test.js unitTests/resources/commitConflictDeadline.test.js` — 57 passing (4s).
- `npx oxlint --format stylish --deny-warnings unitTests/resources/longLivedTransactions.test.js unitTests/resources/commitConflictDeadline.test.js` — passed.
- `npm run lint` — blocked by 13 existing warnings outside this diff; the scoped lint above passes.
- Fails-on-base: with `resources/DatabaseTransaction.ts` reverted to `origin/main` and everything else in place, the attribution test fails (`Timed out after 2000ms waiting for condition`) and passes with the change. The `state: active` assertion added this round was likewise confirmed to fail with the `monitored &&` guard removed and pass with it.

Rebased onto `origin/main` (the branch was `CONFLICTING`). #2459 landed the `describeCommitIdentity()` / `allowStuckCommitLog()` refactor under the same `checkOverloaded()` log line; the resolution keeps that refactor and threads the holder-candidate suffix through it.

Note: `resources/DESIGN.md` is one edited row; the rest of its diff is Prettier re-padding the markdown table.

Refs #2471
Refs #2495

Complexity: complicated

<sub>Review-Coverage: authored=codex; ran=claude; blocked=gemini(permission-denied); declined=cursor-grok,cursor-composer,domain; rounds=6 @ bb14e674d1c4</sub>

<sub>Human-Review-Need: 3 @ bb14e674d1c4</sub>


resources/longLivedTransactions.ts
```diff
@@ -0,0 +1,275 @@
...
+	}
+}
+
+/**
+ * Start the periodic sweep. Armed unconditionally on the main thread — the pass itself honors a live
+ * threshold change, including one that disables it — and idempotent because the thread startup path
+ * that calls it can run more than once.
+ */
+export function startLongLivedTransactionReporting(): void {
+	if (!isMainThread || sweepTimer) return;
```
Devin-Holland, line 182: On item 6 (placement) — the field data from the two #2450 incidents argues for **not** making the main thread the only sweeper.

On both production nodes, stdout mirroring to `docker logs` went silent from the wedge onset (09:02 and 04:25 UTC) until the restart, while `hdb.log` kept receiving lines from every http worker (450k+ lines in the window, zero reached docker). Worker-thread stdout in Node is drained by the main thread's event loop; file appends are not. That is consistent with the main thread being stalled for the whole incident (unproven — the stdio guard's `disableStdio()` on EPIPE/EIO is the alternative explanation, but nothing killed the pipe reader). If that is what happened, this sweep — the one surface that can see an untracked handle — would not have run once during the 4.5 h and 9 h wedges, exactly when it was needed.

`registryStatus()` is process-global, so any thread can sweep it. Suggestion: also arm it on the last http worker (the thread that already owns TTL cleanup), or make that worker the owner instead of main. Two lines per holder from two threads is a small price; the logger already tags the thread, so an operator can tell which one is alive. Non-blocking, but this is the placement I would change.

— Claude (devain), reviewing for Devin

resources/longLivedTransactions.ts
```diff
@@ -0,0 +1,275 @@
...
+ * Returns '' rather than throwing — the caller's own 503 must survive a failure here.
+ */
+export function describeHolderCandidates(
+	databasePath: string | undefined,
+	excludeNativeId: number | undefined
+): string {
+	if (!databasePath || !registryStatusFn) return '';
+	try {
+		const target = resolve(databasePath);
+		const entry = registryStatusFn().find((database) => resolve(database.path) === target);
```
Devin-Holland: Filtering candidates to the stuck commit's own database can exclude the actual holder. The verification table is **one** process-global slot array: `slotFor(dbId, cfId, key)` mixes `dbId` into the hash and indexes `slots_[h & mask_]`, so a key in `system` or `oauth` collides with a `data` key at the same 1-in-131072 rate as two `data` keys do. In CM the `oauth` (session writes on every login) and `system` (`hdb_user`/`hdb_role`) databases take steady writes, so a long-lived holder there parking a `data.Host` commit is a real shape — and this line would return `''` for it (the `only offers handles from the stuck commit's own database` test pins that).

Suggest enumerating every registry entry, ranking the same database first, then by age, and labelling a foreign-database candidate with its path — the `MAX_HOLDER_CANDIDATES` cap keeps the line short either way. Sketch (the test above would invert to assert the `/db/beta` handle is offered, labelled):

```ts
for (const database of registryStatusFn()) {
	const details = database.transactionDetails;
	if (!details) continue;
	const sameDatabase = database.path !== undefined && resolve(database.path) === target;
	for (const handle of details)
		if (handle.id !== excludeNativeId || !sameDatabase)
			candidates.push({ id: handle.id, ageMs: handle.ageMs, sameDatabase, path: database.path });
}
candidates.sort((a, b) => Number(b.sameDatabase) - Number(a.sameDatabase) || b.ageMs - a.ageMs);
```

(The `database.path !== undefined` guard also covers the `resolve(undefined)` throw the bots flagged on this line.) Non-blocking as a correctness-of-diagnostic point, but it is the difference between naming the holder and printing nothing in the cross-database case.

— Claude (devain), reviewing for Devin

resources/DatabaseTransaction.ts
```diff
@@ -1629,17 +1639,61 @@ function chainStillActive(txn: DatabaseTransaction): boolean {
...
+	if (txn.isReplay) states.push('replay');
+	if (txn.committing) states.push('commit-phase');
+	if (txn.open === TRANSACTION_STATE.CLOSED) states.push('closed-with-open-iterators');
+	// A transaction that keeps writing (or a chain link whose sibling does) re-arms its own idle
+	// limit indefinitely and so never reaches the over-limit branches at all.
+	if (txn.timeout > 0) states.push('active');
+	if (states.length === 0) states.push('over-limit');
+	reportLongLivedHolder({
+		databasePath: (txn.db as any)?.rootStore?.path,
+		nativeId: txn.transaction.id,
```
Devin-Holland: Attribution names only this link's handle, but a blind write to a second database attaches its **own** native handle to a `.next` link (`attachOwnedTransaction` at 866) while only the root is supervised (877) and only reading links enter `trackedTxns` (503). So when the child link is the intent holder, the sweep names its id and no attribution line ever joins to it — the exact join this PR exists to make — and `countPendingWrites` reports the chain total against the root's id.

Suggest walking `for (let link = txn; link; link = link.next)` here, reporting each link that has a live `transaction` under its own `link.transaction.id` (skipping links that are in `trackedTxns`, since the tick visits those itself), with the write count per link or labelled as a chain total. Supervision stays root-only; this is reporting only.

(Surfaced by a Codex pass, verified against the head.) — Claude (devain), reviewing for Devin

resources/longLivedTransactions.ts
```diff
@@ -0,0 +1,275 @@
...
+function observeHandle(
+	reports: Map<string, ReportState>,
+	key: string,
+	ageMs: number,
+	thresholdMs: number
+): ReportState {
+	let state = reports.get(key);
+	if (state && ageMs < state.lastAgeMs) state = undefined;
+	if (!state) {
+		state = { nextReportAgeMs: thresholdMs, lastAgeMs: ageMs, touchedAt: 0 };
```
Devin-Holland, line 55: `nextReportAgeMs` is set from the threshold only when the state is first created, so a live config change does not reach a handle that was already observed: lower 5m → 1m and a 3-minute-old handle stays silent until 5m; raise 1m → 10m and it still fires at 1m. The module comment above `getReportThresholdMs()` promises the opposite ("so a live config reload takes effect"), and item 2 of the description leans on an operator tightening it mid-incident.

Cheapest fix: remember the threshold the maps were built under and clear both when it changes (re-reporting once at the new threshold is the behavior an operator asking for a change would expect), plus a test that lowers and raises the threshold while a handle stays present:

```ts
let thresholdInEffectMs = -1;
function rebaseIfThresholdChanged(thresholdMs: number): void {
	if (thresholdMs === thresholdInEffectMs) return;
	thresholdInEffectMs = thresholdMs;
	sweepReports.clear();
	attributionReports.clear();
}
```

(Surfaced by a Codex pass, verified against the head.) — Claude (devain), reviewing for Devin

resources/longLivedTransactions.ts
```diff
@@ -0,0 +1,275 @@
...
+ */
+export function reportLongLivedHolder(holder: LongLivedHolder): void {
+	try {
+		const thresholdMs = getReportThresholdMs();
+		if (thresholdMs === 0) {
+			if (attributionReports.size > 0) attributionReports.clear();
+			return;
+		}
+		const expiredBefore = Date.now() - ATTRIBUTION_STATE_TTL_MS;
+		for (const [key, state] of attributionReports) if (state.touchedAt < expiredBefore) attributionReports.delete(key);
```
Devin-Holland: This prune walks the whole attribution map on **every** `reportLongLivedHolder` call, and the monitor calls it once per over-threshold transaction per tick — O(N²) map visits per tick on the request-serving worker, in exactly the leak scenario where N grows. Prune at most once per interval instead (a module-level `lastAttributionPruneAt` checked against `SWEEP_INTERVAL_MS`), or prune from the monitor tick rather than per holder. Low severity today, but this is the code that runs during the incident.

(Surfaced by a Codex pass, verified against the head.) — Claude (devain), reviewing for Devin

Devin-Holland (commented):
Reviewed with the #2450/#2471 field data in hand (I ran the two-node investigation this morning and filed #2471). The three-surface design is the right shape: the 45 s `checkOverloaded()` candidates line covers the first five minutes, the 5 m sweep catches the untracked-handle class nothing else can see, and the per-thread monitor supplies `startedFrom`/state. The cross-thread test that pins `registryStatus()` as process-global is the right guard for the whole design — if that ever regresses, everything here goes quiet, and now something fails. Verification route is stated and the fails-on-base check is there. Nice work.

Five inline threads. Two are design points from the incident data; three are findings a Codex pass under Devin's account produced as an unanchored pending draft, which I verified against the head, anchored, and folded in here (the stale draft is deleted so replies do not vanish into it):
- **Sweep placement (your item 6)** — the incident evidence points at a stalled main thread during both wedges, so I'd arm the sweep on the last worker too (or instead).
- **Same-database candidate filter** — the VT is one global slot array, so a `system`/`oauth` holder can park a `data` commit and this filter would print nothing for it.
- **Chain-link attribution gap** — a blind-write `.next` link owns its own native handle but is never attributed, so the sweep's id has nothing to join to when the child is the holder.
- **Threshold changes don't reach observed handles** — `nextReportAgeMs` is fixed at first observation, contradicting the live-reload comment and the item-2 use case.
- **Per-holder prune is O(N²) per tick** — prune once per interval instead.

Process notes, no action needed from reviewers:
- I concur with the bot blocker on `resolve(database.path)` throwing on an undefined path; the sketch in my second thread absorbs that guard.
- The `v5.2` cherry-pick already conflicted (`cc147052`), so the milestone alone will not deliver this to 5.2.x — it needs the conflict resolved on `cherry-pick/v5.2/pr-2473` or a `/patch-pr`.
- `storage.longTransactionReportThreshold` is a new public config key — a HarperFast/documentation companion (or a note here on why not) before this leaves draft, per step 16.
- `Closes #2471`: ask 3 (reap policy for `sourceApply`/`isReplay`) is deliberately deferred in your item 7, which I agree with. As the issue's author I'm fine with closing it on merge **if** the policy half gets its own issue so it isn't lost; otherwise `Refs #2471`.

— Claude (devain), reviewing for Devin

server/threads/socketRouter.ts
```diff
@@ -52,6 +53,9 @@ export async function startHTTPThreads(threadCount = 2, dynamicThreads?: boolean
 	// process-global singleton; see startTransactionLogCooling). Runs for all
 	// thread modes below, including the single-threaded (threadCount === 0) path.
 	startTransactionLogCooling();
+	// Same placement and same reason: the rocksdb-js transaction registry is process-global, so one
+	// main-thread sweep names every worker's long-lived handles instead of each worker naming them all.
+	startLongLivedTransactionReporting();
```
cb1kenobi, line 56: **Sweep is never armed outside HTTP startup — Severity: High (blocking).** `startLongLivedTransactionReporting()` is only called from `startHTTPThreads`, so job-worker and CLI processes that open RocksDB without starting HTTP never enumerate the process-global registry. Untracked long-lived handles—the class only this sweep can detect—remain completely silent; start this singleton from main-thread storage/RocksDB initialization shared by every process that can open a handle.

—
_Generated by Barber AI_

resources/longLivedTransactions.ts
cb1kenobi, line 55: **Live threshold changes retain stale backoff — Severity: Medium (blocking).** `nextReportAgeMs` is initialized only for new state, so an already-observed handle keeps its previous threshold or exponential backoff after `storage.longTransactionReportThreshold` changes. Lowering the threshold during an incident can leave the first report delayed until the old threshold, while raising it can continue reporting too early; track the parsed threshold and clear both report-state maps whenever it changes.

—
_Generated by Barber AI_

resources/longLivedTransactions.ts
```diff
@@ -0,0 +1,275 @@
...
+ * on, oldest first, as a sentence to append to that commit's log. Age only ranks them: a coordinated
+ * retry parks on whichever transaction holds the verification-table slot when the commit reaches it,
+ * which can be younger than the commit itself, so filtering by age could drop the real holder.
+ * Returns '' rather than throwing — the caller's own 503 must survive a failure here.
+ */
+export function describeHolderCandidates(
+	databasePath: string | undefined,
+	excludeNativeId: number | undefined
+): string {
+	if (!databasePath || !registryStatusFn) return '';
```
cb1kenobi, line 287: **Disable setting is ignored by candidate reporting — Severity: Medium (blocking).** Setting `storage.longTransactionReportThreshold` to `0` is documented to disable all three reporting surfaces, but `describeHolderCandidates()` never checks it. Every overloaded commit still queries the registry and appends holder candidates, so operators cannot fully disable reporting; add an early `getReportThresholdMs() === 0` guard and cover it through `checkOverloaded()`.

—
_Generated by Barber AI_

resources/longLivedTransactions.ts
```diff
@@ -0,0 +1,275 @@
...
+			);
+		}
+		if (due.length === 0) return;
+		due.sort((a, b) => b.ageMs - a.ageMs);
+		for (const handle of due.slice(0, MAX_REPORTED_PER_SWEEP)) {
+			markReported(handle.state, handle.ageMs);
+			logger.warn?.(
+				`Long-lived RocksDB transaction handle: id ${handle.id} has been open for ${prettyDuration(handle.ageMs)} on database ${handle.path}. ` +
+					`A handle this old can hold write intents that park other writers' commits, and pins a read snapshot that blocks version reclamation.`
+			);
```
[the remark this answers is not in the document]
kriszyp: Fixed — the sweep now prints `handle.path ?? '?'`, matching `reportKey` and the attribution line. Covered by a new test that sweeps a registry entry with no path.

— Claude Opus 5

resources/longLivedTransactions.ts
```diff
@@ -0,0 +1,275 @@
...
+			logger.warn?.(
+				'Long-lived transaction reporting is degraded: the installed @harperfast/rocksdb-js does not expose registryStatus().transactionDetails (requires 2.8.0)'
+			);
+		}
+		if (due.length === 0) return;
+		due.sort((a, b) => b.ageMs - a.ageMs);
+		for (const handle of due.slice(0, MAX_REPORTED_PER_SWEEP)) {
+			markReported(handle.state, handle.ageMs);
+			logger.warn?.(
+				`Long-lived RocksDB transaction handle: id ${handle.id} has been open for ${prettyDuration(handle.ageMs)} on database ${handle.path}. ` +
```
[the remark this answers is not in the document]
kriszyp: Fixed alongside the identical bot finding — `handle.path ?? '?'`, with a test for a pathless registry entry.

— Claude Opus 5

unitTests/resources/txn-tracking.test.js
```diff
@@ -421,6 +424,43 @@ describe('Write txn timeout', () => {
...
+			});
+			const reported = warnings.filter(([message]) =>
+				String(message).includes('Harper transaction has held RocksDB transaction')
+			);
+			assert.ok(reported.length > 0, 'the un-reaped source-apply holder must be named');
+			assert.match(reported[0][0], /state: [^,]*source-apply/);
+			assert.match(reported[0][0], /IndexedTxnTable/);
+			assert.match(reported[0][0], /transaction \d+/, 'the native id is the join key with the registry sweep');
+			// Attribution must not have changed the exemption it is reporting on.
+			assert.equal((await IndexedResource.get(402))?.t, 8, 'source-apply write should still be preserved');
```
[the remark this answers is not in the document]
kriszyp: Fixed — `assert.strictEqual`.

— Claude Opus 5

resources/longLivedTransactions.ts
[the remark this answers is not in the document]
kriszyp: Fixed, and the surrounding logic changed with it: the lookup now enumerates every registry entry rather than `.find()`-ing one, guarding `database.path != null` before `resolve()`. A new test puts a pathless entry ahead of the target in the registry and asserts the target's candidates still come back — that is the case the throw silently zeroed.

— Claude Opus 5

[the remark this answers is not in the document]
kriszyp: Fixed — good catch on the blast radius being every database rather than the pathless one. The predicate now guards `database.path != null`, and the test you noted was missing exists: a pathless entry sits ahead of the target in the registry and the target's candidates are still returned.

— Claude Opus 5

kriszyp: Adopted, close to your sketch. `describeHolderCandidates` enumerates every registry entry, tags each handle with `sameDatabase`, sorts `sameDatabase` first then by age, and labels a foreign candidate with its path. The `only offers handles from the stuck commit's own database` test inverted as you predicted, and two more went in: a foreign handle is offered and labelled, and an older foreign handle ranks behind a younger same-database one. The `database.path != null` guard your sketch carried also closes the bots' `resolve(undefined)` throw, so that thread is fixed by this change too.

— Claude Opus 5

resources/DatabaseTransaction.ts
kriszyp: Adopted. `reportIfLongLived` now walks `for (let link = txn; link; link = link.next)` and reports each link holding its own handle under that link's own id, skipping links in `trackedTxns` since the tick visits those itself. Write count is per link rather than the chain total, since it is printed beside that link's id.

One thing your note did not cover that the test surfaced: `state: active` was wrong for a chain link. `timeout` is armed by `addWrite` and decremented once per tick only for the link the tick entered on, so a link reached through `.next` kept an armed `timeout` forever and reported `active` for a link idle for hours. It now reads `writeTimeout` — the clock `chainStillActive` actually decays — for chain links, and `timeout` for the entry link. Both directions are pinned by tests.

Also worth recording: for a `put` the second-database link *does* enter `trackedTxns` (the prior-entry read registers it), so the gap is narrower than 'blind write to a second database' — it is any link whose write never reads. The test drops the link from `trackedTxns` to reach that shape deterministically.

— Claude Opus 5

resources/longLivedTransactions.ts
kriszyp: Adopted, essentially your snippet — `rebaseIfThresholdChanged()` remembers the threshold the maps were built under and clears both when it changes. It also subsumes the two special-case clears that existed for the `0` path. Three tests: lower the threshold and a previously-silent handle reports, raise it and a reporting handle goes quiet, and the same for the attribution surface.

— Claude Opus 5

resources/longLivedTransactions.ts
kriszyp: Adopted — the prune now runs at most once a minute (`ATTRIBUTION_PRUNE_INTERVAL_MS`) instead of once per `reportLongLivedHolder` call.

A related one from the same 'code that runs during the incident' angle turned up in pre-push review and is also fixed: the sweep was building a key, a report state and two container entries for *every* live handle in the process on every pass, even on a healthy node with nothing near the threshold. It now skips sub-threshold handles outright — `nextReportAgeMs` is seeded at the threshold and only grows, so that state could never change an outcome.

— Claude Opus 5

resources/longLivedTransactions.ts
kriszyp: Fixed — `rebaseIfThresholdChanged()` records the threshold the report maps were built under and clears both whenever it changes, so an already-observed handle is re-measured against the new value in either direction. Tests cover lowering and raising it, on both the sweep and the attribution surface.

— Claude Opus 5

resources/longLivedTransactions.ts
kriszyp: Fixed — `describeHolderCandidates()` returns '' when `getReportThresholdMs() === 0`, covered both directly and through `checkOverloaded()` as you asked (the stuck-commit log still fires, without the candidates suffix). The check is inside the function's `try`, since a throw here would turn the caller's 503 into a 500.

— Claude Opus 5

server/threads/socketRouter.ts
kriszyp: Not taking this one — the premise does not hold. `launchJobThread` (`server/jobs/jobRunner.ts:146`) calls `threadsStart.startWorker()`, which is `new Worker(...)` at `server/threads/manageThreads.js:408`. Job workers are **threads inside the process that already armed the sweep**, and the rocksdb-js registry is process-global, so their handles are enumerated by the existing main-thread sweep — including untracked ones. The cross-thread test in this PR pins exactly that property.

What is genuinely uncovered is a separate CLI process (`bin/backup.ts`, `bin/copyDb.ts`) that opens RocksDB without `startHTTPThreads` and lives past the threshold. That is narrow and is not the #2471 shape, which is a request-serving node parking other writers. Moving the arming into storage init to cover it was weighed and declined by the PR author; rationale is in item 6 of the description.

— Claude Opus 5

resources/longLivedTransactions.ts
kriszyp: Leaving the placement as-is for now — @kriszyp's call, and I want to record why rather than quietly not doing it, because your field evidence is the strongest argument on this thread.

The other placement thread (Barber AI's, marked blocking) turned out to rest on a false premise: job workers are `new Worker(...)` threads in the process that already armed the sweep (`server/jobs/jobRunner.ts:146` → `server/threads/manageThreads.js:408`), so the registry being process-global means their handles are already reported. Only a separate CLI process is uncovered.

Yours does not have that problem — a stalled main thread really would silence this surface, and I could not find anything that rules out your reading. What it does have is that the cost is asymmetric right now: the per-thread attribution surface runs on every worker regardless, so a stalled main thread costs only the untracked-handle sweep, and the current shape has three rounds of Codex review behind it. Against a hypothesis that you yourself scope as unproven (`disableStdio()` on EPIPE/EIO being the competing explanation), that did not clear the bar for changing where a process-global singleton is armed.

If the wedge recurs and the sweep is silent while `hdb.log` keeps taking worker lines, that is the confirmation, and the change is small: drop the `isMainThread` gate to `isMainThread || workerData?.workerIndex === 0` (job workers get no `workerIndex`, so that is exactly main + HTTP worker 0). Leaving this thread open for you.

— Claude Opus 5

Issue #1877, opened 2026-07-20
Log rotation maxSize not respected properly


Devin-Holland:
### Bug Summary

Log rotation is enabled with a maxSize is not being respected properly

### Steps to Reproduce

On 5.1.14 there is the following config:

```
logging:
  auditAuthEvents:
    logFailed: false
    logSuccessful: false
  auditLog: true
  auditRetention: 3d
  file: true
  level: warn
  root: log
  rotation:
    enabled: true
    compress: true
    interval: null
    maxSize: 512M
    path: log
  stdStreams: true
  external:
    level: info
    path: /home/harperdb/harper/log/hdb.log
    console: true
    stdStreams: false
  colors: false
  path: /home/harperdb/harper/log/system.log
```

And the `hdb.log` file was 200G

### Expected Behavior

maxSize should be respected for the `hdb.log` file

### Actual Behavior

maxSize is not enforced for the `hdb.log` file. 

It looks like because that is an `external` log it doesn't respect the default logging configuration for `rotation`. 

The docs mention these as options for `external` loggers: 

``` Each component logger can have its own path, root, level, tag, and stdStreams settings. All components default to the main logging configuration unless overridden.```

`rotation` is not referenced as an option. It should fall to whatever the default is if not set.

### Are you planning to fix this issue?

No, just reporting the issue

### First-time contributor support

- [ ] I'm new to contributing and would appreciate guidance on the process
- [ ] I'd like help understanding the project structure
- [ ] I need assistance with setting up the development environment
- [ ] I'm comfortable contributing, but new to this project specifically

kriszyp:
Verified against `origin/main` @ e16d9616, plus a stronger candidate mechanism for the "never rotated" observation.

The general claim holds: rotation is driven **only** by a `setInterval(..., auditInterval ?? 60000)` in `utility/logging/logRotator.ts` that checks `fs.stat(logger.path).size >= maxBytes` — there is no write-path/size trip outside that timer, so `maxSize` gives no disk-fill protection between ticks under load.

[...]

Issue #1661, opened 2026-07-07
[Epic] Logging subsystem hardening (FD leak/rotation/TS conversion)


kriszyp:
Tracking epic for logging-infrastructure hardening: log-rotator FD leak/retention, cache-hit/outgoing-request correlation logging, stderr routing, scoped-logger naming, and the JS→TS conversion of the logger module.

---
_Filed by Claude (Opus 4.8), issue-backlog triage & grouping pass._

PR #2475, opened 2026-09-02
Enforce logging.rotation.maxSize on the log write path instead of only on the 60-second audit tick

kriszyp:
`logging.rotation.maxSize` is now enforced by every file-writing isolate, so a busy HTTP worker no longer grows the active log until the main thread's 60-second audit tick. The sink counts the bytes it actually appends and invokes the [write-path rotation guard](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-cde66fc5c81773a66e7af2e438e0730dbc3f44a99cc2835b26726729048b2c11R275) at a fixed `maxSize / 16` checkpoint; an already-oversized file rotates on the next append. The final hot path [encodes each batch once](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-22ec5442298b8e9a003e550f7164aa5bbce8d0395aaef06720baeafd92308baaR972), uses that buffer for the append, and records its exact byte length.

Rotation is split into a synchronous rename/descriptor release and an asynchronous publish step. Before compression or retention can unlink an archive, the [generation coordinator](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-c4e280f5946af493a4ee91dfca05681b61528bca5821e213bc83be4ce2da6f18R128) asks every enumerable in-process peer to close a descriptor still pointing at that inode. An unproven generation remains plain and is retried. Workers that have not received the thread transport cannot claim an empty peer set as proof, pending release requests are bounded, and [only the main thread may destructively publish an archive](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-cde66fc5c81773a66e7af2e438e0730dbc3f44a99cc2835b26726729048b2c11R114) because its worker roster is authoritative.

The follow-up fixes in this pass close the review findings on the existing PR:

- Runtime removal of `logging.rotation` now retracts the guard owned by that configuration source, while explicit component policies keep precedence over inherited main policies. Re-adding either source reclaims the cached sink safely through the [policy-source arbitration](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-22ec5442298b8e9a003e550f7164aa5bbce8d0395aaef06720baeafd92308baaR863).
- Compression catch-up and retention use a deadline rather than a four-file ceiling. One compression failure no longer starves later archives, a busy compression leaves work for the next tick, and the retention cursor resumes fairly when a pass exhausts its budget.
- A failed archive rename no longer diverts all future file logs to daemon stdio. Appends continue to the active file, the cap may temporarily be exceeded, and a parseable in-file diagnostic is prepended on the next successful append.
- Interval rotation notices when another isolate replaced the active generation and restarts its clock, preventing a freshly-created log from being rotated immediately by a stale timer.
- The end-to-end test now asserts [every request marker exactly once](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-b2fb589e3adc69b737bf57e661612c5da54a45e2c7a5e763de1f42227a1a574dR123) across active and compressed generations and separately proves that removing the rotation policy reaches both HTTP workers.

## For the human reviewer

Please focus on the lifecycle and failure-policy choices below.

- The release proof covers peers enumerable through this process's worker mesh; it does not cover another OS process writing the same path. A worker can still briefly miss a newly admitted peer, but its local proof now only prompts descriptor release and leaves the archive plain; the main-thread audit obtains a fresh proof from its authoritative roster before compression or retention. A separate OS process remains outside that roster, so default `compress: false` with no retention is non-destructive while configured compression/retention still carries that cross-process limitation.
- A blocked worker deliberately pauses both compression and retention because destroying an inode it may still hold can lose buffered records. The audit emits a warning every five minutes while stalled and a recovery notice afterward. This preserves logs but can let the archive directory grow despite configured retention.
- Compression publishes via temp-file rename but does not `fsync` the gzip before replacing the authoritative name. A process crash leaves the plain archive intact; an abrupt host/storage failure can theoretically leave a present but non-durable gzip that a later sweep trusts. Crash-left `.tmp` files can also persist indefinitely when retention is unset.
- Invalid values such as `0K`, `-1K`, and `1xK` are now rejected by config validation instead of booting with broken historical behavior. That is intentional strictness, but it can make an upgrade fail fast on an existing typo.
- An explicit component rotation policy for the same file retains the historical last-explicit-owner precedence over inherited main policy. Removing the explicit policy restores main inheritance; removing both clears the guard.
- Retention and its release gate have extensive unit coverage but no full-server deletion assertion. The HTTP integration test covers real two-worker rotation, gzip publication, exactly-once markers, and runtime policy removal. Its fallback archive directory is under the OS temp root; the focused run was same-device, while a runner mounting `/tmp` separately from Harper's data root would exercise the EXDEV refusal instead.
- Two consecutive Gemini passes disagreed about append encoding: the first called the prior string write plus `Buffer.byteLength` a major double-scan and requested one buffer; the final pass requested reverting because of buffer allocation. The buffer is retained because `appendFileSync` must encode the string anyway, while this form reuses that encoding for exact accounting. Claude independently traced this path and found it sound.

The framing gate cleared before implementation: `Framing-Verdict: chosen-approach-sound (22fe112820ec)`. The final exact-head review covered the complete diff with Gemini at `4c0ef25b3bf0`; the Claude grader failed at startup and the domain adjudicator timed out, so neither is claimed as final-head coverage. Gemini's repeated buffer-allocation concern was adjudicated against the write path: retaining the buffer avoids separately scanning with `Buffer.byteLength` and then encoding the same string again inside `appendFileSync`. Its additional memory-stress-test request and aggregated comment-cleanup nit identified no failing behavior in this CI repair.

## Changes

- The analytics task-queue probe in [resources/analytics/write.ts](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-f1e47c8ff8e65d10e6aa737a8eb28cf19219efe000ea2890bedc6b013cb955c4R907) now treats a log-path `ENOENT` race as a completed latency round trip instead of an unhandled rejection when another isolate rotates the file.
- [server/threads/logRotationTransport.ts](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-a0f78930589f1806895163c65f9006df1c5c37caf037b6c50a59e4f30fc28a90R1) adapts the existing thread mesh to the generation coordinator without reversing the logging-to-thread dependency, and [server/threads/manageThreads.js](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-02c28577c2ca63129ed10809ce3db1469c70bef080e04fd273f8cbd510b6521bR1596) installs that adapter only after its thread-exit hooks exist.
- The audit rotator in [utility/logging/logRotator.ts](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-ef748258d2f3272624c4f884139aa041682e9d9eead288206ccd0cdddf578c78R11) now shares the synchronous generation primitives, gates destructive compression/retention on peer release, bounds sweep work by time, resumes retention fairly, and contains tick failures without overlapping passes.
- [validation/configValidator.ts](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-aeee467d8401b59eaffd099dd61547bbf4a63b0198e2be2dc0fe55d06c284cb0R505) uses the runtime's `parseMaxSize` implementation, so accepted configuration cannot become a zero, negative, or `NaN` write-path limit.

## Verification

- `npm run build` at `4c0ef25b3bf0` — passed.
- `npx mocha "unitTests/utility/logging/**/*.test.js"` at `4c0ef25b3bf0` — 190 passing, 14 pending. The added [main-thread publication](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-0740149e0ea8986d460a5dbd25572519433e5c551d0b73d715299ebf656600e8R117), [zero-inode release](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-c4e280f5946af493a4ee91dfca05681b61528bca5821e213bc83be4ce2da6f18R168), [buffered-diagnostic wait](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-5305ab395a92c37278874d49421e9e2258d790ac64f5ab0f7f416a91fa52bda1R323), and [worker fixture transport](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-24dfab84ec07cbd39ba2e3a0d2f1be39b43c1d543b8305b50499b9ef33043818R6) cases all passed.
- `npm run test:integration -- integrationTests/server/log-rotation-write-path.test.ts` at `4c0ef25b3bf0` — 2 passing. The test uses the [platform-observable worker count](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-b2fb589e3adc69b737bf57e661612c5da54a45e2c7a5e763de1f42227a1a574dR24) while retaining two-worker Node/Linux coverage.
- The integration fixture's [component mapping](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-16e74f28ddb987dde0c1b9a07644b2a67e4d545f5f900934bf722f0c32eaa073R1), [private package metadata](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-aad2f218f92a410501eb0873feac43988b0a7cee0dc29e082b5a149d8f7c21a9R1), and [request-driven worker logger](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-c3022fe86a83a21fe667f17103b8a01e9b5ac78f6b7c27cbacc09ad25e5e6f0aR1) drive the real HTTP-worker write path rather than a direct logging helper.
- The earlier same-commit focused Bun run recorded 2 passing. A fresh rerun on this worker could not start because `bun` is unavailable (`spawn bun ENOENT`); pushed-head Bun CI is the authoritative rerun.
- [unitTests/utility/logging/harper_logger.test.js](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-8c33beb56189cc9ff99eae34414ae37a65c1e99da970ef11cd57d30bd24e88e9R937) covers pathless setup and every inherited/explicit rotation-policy ownership transition on isolated paths; [unitTests/utility/logging/logRotator.test.js](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-d4bc187e1a3bd090f79ae354250749642065ca254d7fbb544d2b86a610ac7db9R73) covers compression catch-up, duplicate cleanup, live-log protection, interval resets, failure continuation, and shutdown.
- [unitTests/validation/configValidator.test.js](https://github.com/HarperFast/harper/pull/2475/changes?w=1#diff-bfc0bd3b280a1b03fa7b7c6d83f595f3d56383c4e877b739bd149d588acf68a8R507) covers rejected unusable size values and preserves accepted decimal/exponent forms.
- `npm run test:unit:resources` — 1,943 passing, 28 pending.
- `npm run test:unit:main` — 5,319 passing, 196 pending, 12 unrelated/environment failures: the existing Git credentials assertion, ten component-loader cases blocked by this deep worktree resolving dependencies outside its containment rule, and the domain-socket path-length assertion.
- Prettier and oxlint passed on all six files changed by `4c0ef25b3bf0`; `git diff --check origin/main...HEAD` also passed. Full lint still reports 13 pre-existing warnings outside this diff.
- `npm run test:unit:windows` could not start locally because the gate hard-codes a worktree-local `node_modules/mocha/bin/mocha.js`; this checkout uses the parent install. The focused cross-platform worker coverage passed locally.
- The full integration sweep encountered unrelated harness/service failures (including Ollama import/service setup and pre-existing fixtures); the focused logging integration passed independently.

Refs #1877

Complexity: complicated

Comment generated by Codex (GPT-5)

<sub>Review-Coverage: authored=codex; ran=gemini; blocked=claude(exit-1),domain(timeout); declined=cursor-grok,cursor-composer; rounds=8 @ 4c0ef25b3bf0</sub>

<sub>Human-Review-Need: 3 @ 4c0ef25b3bf0</sub>

### DIFF
--- a/server/threads/logRotationTransport.ts
+++ b/server/threads/logRotationTransport.ts
@@ -7,7 +7,7 @@ import { threadId } from 'node:worker_threads';
 import { setRotationTransport } from '../../utility/logging/logGenerationCoordinator.ts';
 import { broadcast, onMessageByType, onThreadExit } from './manageThreads.js';
 
-// The connected-ports array with `sendToThread`, assigned to the `threads` global by manageThreads.
+// Assigned to the `threads` global by manageThreads.
 declare const threads: {
 	sendToThread(threadId: number, message: any): boolean;
 	[index: number]: { threadId?: number };
--- a/server/threads/logRotationTransport.ts
+++ b/server/threads/logRotationTransport.ts
@@ -27,9 +27,9 @@ setRotationTransport({
 	},
 	onThreadExit,
 	peerThreadIds() {
-		// Job workers included: they write to the log whether or not they take part in ITC gossip. The
-		// deadlock that excludes them from the schema broadcast cannot happen here — the handler only
-		// closes a descriptor, and no caller blocks its event loop on the answer.
+		// Job workers included: they hold log descriptors whether or not they take part in ITC gossip,
+		// and the deadlock that excludes them from the schema broadcast cannot happen here — the
+		// handler only closes a descriptor, and no caller blocks its event loop on the answer.
 		const ids = new Set<number>();
 		for (let i = 0; i < threads.length; i++) {
 			const peer = threads[i].threadId;
--- a/server/threads/logRotationTransport.ts
+++ b/server/threads/logRotationTransport.ts
@@ -29,7 +29,9 @@ setRotationTransport({
 	peerThreadIds() {
 		// Every connected port, job workers included: an isolate that writes to the log holds a
 		// descriptor on it whether or not it takes part in ITC gossip, and the whole point of the
-		// acknowledgement is to know that no such descriptor survives.
+		// acknowledgement is to know that no such descriptor survives. The deadlock that keeps job
+		// workers out of the schema-gossip broadcast cannot happen here — the handler is a stat and a
+		// close with nothing to wait on, and no caller blocks its event loop on the answer.
 		const ids = new Set<number>();
 		for (let i = 0; i < threads.length; i++) {
 			const peer = threads[i].threadId;
--- a/server/threads/logRotationTransport.ts
+++ b/server/threads/logRotationTransport.ts
@@ -1,0 +1,40 @@
+'use strict';
+
+// Hands the log-rotation coordinator the thread mesh it needs, from this side of the dependency
+// edge: manageThreads already imports harper_logger, so logging can never import manageThreads.
+
+import { threadId } from 'node:worker_threads';
+import { setRotationTransport } from '../../utility/logging/logGenerationCoordinator.ts';
+import { broadcast, onMessageByType, onThreadExit } from './manageThreads.js';
+
+// The connected-ports array with `sendToThread`, assigned to the `threads` global by manageThreads.
+declare const threads: {
+	sendToThread(threadId: number, message: any): boolean;
+	[index: number]: { threadId?: number };
+	length: number;
+};
+
+setRotationTransport({
+	threadId,
+	broadcast(message) {
+		broadcast(message);
+	},
+	sendToThread(target, message) {
+		threads.sendToThread(target, message);
+	},
+	onMessage(type, handler) {
+		onMessageByType(type, handler);
+	},
+	onThreadExit,
+	peerThreadIds() {
+		// Every connected port, job workers included: an isolate that writes to the log holds a
+		// descriptor on it whether or not it takes part in ITC gossip, and the whole point of the
+		// acknowledgement is to know that no such descriptor survives.
+		const ids = new Set<number>();
+		for (let i = 0; i < threads.length; i++) {
+			const peer = threads[i].threadId;
+			if (peer !== undefined && peer !== threadId) ids.add(peer);
+		}
+		return ids;
+	},
+});
--- a/server/threads/logRotationTransport.ts
+++ b/server/threads/logRotationTransport.ts
@@ -27,11 +27,9 @@ setRotationTransport({
 	},
 	onThreadExit,
 	peerThreadIds() {
-		// Every connected port, job workers included: an isolate that writes to the log holds a
-		// descriptor on it whether or not it takes part in ITC gossip, and the whole point of the
-		// acknowledgement is to know that no such descriptor survives. The deadlock that keeps job
-		// workers out of the schema-gossip broadcast cannot happen here — the handler is a stat and a
-		// close with nothing to wait on, and no caller blocks its event loop on the answer.
+		// Job workers included: they write to the log whether or not they take part in ITC gossip. The
+		// deadlock that excludes them from the schema broadcast cannot happen here — the handler only
+		// closes a descriptor, and no caller blocks its event loop on the answer.
 		const ids = new Set<number>();
 		for (let i = 0; i < threads.length; i++) {
 			const peer = threads[i].threadId;
--- a/utility/logging/harper_logger.ts
+++ b/utility/logging/harper_logger.ts
@@ -891,12 +891,15 @@ function getFileLogger(path, rotation, isExternalInstance) {
 	}
 	// this is called on a timer, and will write the log buffer to the file
 	function logQueuedData(entry?: any) {
-		openLogFile(undefined);
 		const payload = logBuffer ? logBuffer.join('') : entry;
 		// A file that just refused a write will refuse the next one too, and every attempt costs a
 		// failed syscall plus a thrown error on whatever path is logging - which, on a full volume,
 		// is the request path at full rate. Go straight to stdio until the cooldown expires.
-		if (logFD && !(retryAppendAfter > performance.now()) && (rotationGuard?.beforeAppend() ?? true)) {
+		// The file is opened after the guard, not before: a guard that recovers by rotating closes this
+		// descriptor, and the write that triggered the recovery belongs in the new generation.
+		const mayAppend = !(retryAppendAfter > performance.now()) && (rotationGuard?.beforeAppend() ?? true);
+		openLogFile(undefined);
+		if (logFD && mayAppend) {
 			let startTime = performance.now();
 			// Encoded once, so the guard's accounting reads a length instead of re-scanning the string.
 			const data = rotationGuard ? Buffer.from(payload) : payload;
--- a/utility/logging/harper_logger.ts
+++ b/utility/logging/harper_logger.ts
@@ -844,7 +844,9 @@ function getFileLogger(path, rotation, isExternalInstance) {
 				closeLogFile,
 				report: reportRotationProblem,
 				onRotated(archivePath) {
-					logToFile(`log rotated, old log moved to ${archivePath}`);
+					// Same shape as every other line in the file: written straight to the sink, this
+					// bypasses the prefix createLogger's logPrepend adds, and readLog parses by level.
+					logToFile(`[${SERVICE_NAME}/${threadId}] [notify]: hdb.log rotated, old log moved to ${archivePath}`);
 				},
 			});
 		} catch (error) {
--- a/utility/logging/harper_logger.ts
+++ b/utility/logging/harper_logger.ts
@@ -890,17 +890,14 @@ function getFileLogger(path, rotation, isExternalInstance) {
 	// this is called on a timer, and will write the log buffer to the file
 	function logQueuedData(entry?: any) {
 		const payload = logBuffer ? logBuffer.join('') : entry;
-		// Released before anything can re-enter this sink. The rotation notice is written from inside
-		// the append below, and a re-entrant flush that still saw this batch would either write every
-		// line of it a second time into the new generation, or be swallowed when this call clears the
-		// buffer on its way out.
+		// Released before anything can re-enter: the rotation notice is written from inside the append
+		// below, and a re-entrant flush still holding this batch would write every line of it twice.
 		logBuffer = null;
 		if (payload === undefined) return;
 		// A file that just refused a write will refuse the next one too, and every attempt costs a
 		// failed syscall plus a thrown error on whatever path is logging - which, on a full volume,
 		// is the request path at full rate. Go straight to stdio until the cooldown expires.
-		// The file is opened after the guard, not before: a guard that recovers by rotating closes this
-		// descriptor, and the write that triggered the recovery belongs in the new generation.
+		// Opened after the guard: a guard recovering by rotation closes this descriptor first.
 		const mayAppend = !(retryAppendAfter > performance.now()) && (rotationGuard?.beforeAppend() ?? true);
 		openLogFile(undefined);
 		if (logFD && mayAppend) {
@@ -910,8 +907,7 @@ function getFileLogger(path, rotation, isExternalInstance) {
 				// Both cleared, so a volume that fills again months later reports itself again
 				retryAppendAfter = undefined;
 				loggedAppendError = false;
-				// byteLength, not a Buffer: appendFileSync writes a string through Node's own encoder
-				// without allocating, and rotation is on by default so this runs on every flush.
+				// byteLength, not a Buffer: appendFileSync encodes the string without allocating one.
 				rotationGuard?.recordWrite(Buffer.byteLength(payload));
 			} catch (error) {
 				retryAppendAfter = performance.now() + APPEND_RETRY_COOLDOWN;
--- a/utility/logging/harper_logger.ts
+++ b/utility/logging/harper_logger.ts
@@ -14,6 +14,7 @@ import { _assignPackageExport } from '../../globals.js';
 import { Console } from 'console';
 import { inspect, types } from 'node:util';
 import { createRotationGuard, INVALID_MAX_SIZE_MSG, parseMaxSize, resolveRotatedLogDir } from './logRotation.ts';
+import { registerLogSink } from './logGenerationCoordinator.ts';
 
 const { isNativeError } = types;
 // store the native write function so we can call it after we write to the log file (and store it on process.stdout
@@ -795,6 +796,10 @@ function getFileLogger(path, rotation, isExternalInstance) {
 		logger.installRotationGuard = installRotationGuard;
 		logger.path = path;
 		fileLoggers.set(path, logger);
+		// Registered by the sink, not by the size guard: a thread that writes this file but has no
+		// guard (no maxSize, or rotation driven only by interval) still holds a descriptor, and an
+		// answer from it has to mean "released" rather than only "handler ran".
+		registerLogSink(path, closeDescriptorOnGeneration);
 	}
 	// An undefined rotation means "this caller has no opinion", not "turn rotation off" — that is
 	// what `enabled: false` says. Several loggers are created for one path during startup and the
@@ -923,6 +928,12 @@ function getFileLogger(path, rotation, isExternalInstance) {
 			logTimeUsage = Math.max(endTime, logTimeUsage) + (endTime - startTime) * 50;
 		} else writeToStdioDirectly(process.stdout, payload);
 		if (logBuffer) logBuffer = null;
+	}
+
+	function closeDescriptorOnGeneration(ino, dev) {
+		// No identity to compare (no descriptor open, or a filesystem that cannot report one) means
+		// closing is the only answer that can still be called a release.
+		if (!logFDIdentity || (logFDIdentity.ino === ino && logFDIdentity.dev === dev)) closeLogFile();
 	}
 
 	function closeLogFile(_unused?: any) {
--- a/utility/logging/harper_logger.ts
+++ b/utility/logging/harper_logger.ts
@@ -13,6 +13,7 @@ import { PACKAGE_ROOT } from '../../utility/packageUtils.js';
 import { _assignPackageExport } from '../../globals.js';
 import { Console } from 'console';
 import { inspect, types } from 'node:util';
+import { createRotationGuard, INVALID_MAX_SIZE_MSG, parseMaxSize, resolveRotatedLogDir } from './logRotation.ts';
 
 const { isNativeError } = types;
 // store the native write function so we can call it after we write to the log file (and store it on process.stdout
@@ -109,7 +110,16 @@ export function updateLogger(logger: any, logOptions: any, name?: string, mainLo
 	// rotation entirely (#1877). Excluded when `logger` IS mainLoggerRef: the fallback would just
 	// reassign mainLogger.rotation to itself, making it impossible to ever clear main's rotation.
 	const inheritedRotation = logOptions.rotation ?? (logger === mainLoggerRef ? undefined : mainLoggerRef?.rotation);
-	if (inheritedRotation && inheritedRotation === mainLoggerRef?.rotation) {
+	let path = logOptions.path;
+	if (path) {
+		if (!logOptions.root) logOptions.root = pathModule.dirname(path);
+	} else if (logOptions.root) {
+		path = join(logOptions.root, logName);
+	} else {
+		path = mainLoggerRef.path;
+		if (!logOptions.root) logOptions.root = pathModule.dirname(path);
+	}
+	if (inheritedRotation && inheritedRotation === mainLoggerRef?.rotation && path !== mainLoggerRef?.path) {
 		// This logger has no rotation block of its own and is inheriting main's wholesale — but
 		// inheriting `rotation.path` too would point this logger's archives at wherever main
 		// archives to. If this logger's file lives on a different filesystem/volume than that
@@ -120,20 +130,17 @@ export function updateLogger(logger: any, logOptions: any, name?: string, mainLo
 		// goes through `logOptions.rotation` above, not this fallback. Clone rather than mutate:
 		// `inheritedRotation` is the literal object shared by mainLogger.rotation and every other
 		// component that inherited it.
+		//
+		// Only when the paths differ. A logger writing to main's own file shares main's rotation
+		// state — file loggers are cached per path — so stripping there discards the operator's
+		// configured rotation.path for the main log itself, which no EXDEV risk justifies.
 		const { path: _inheritedPath, ...rotationWithoutPath } = inheritedRotation;
 		logger.rotation = rotationWithoutPath;
 	} else {
 		logger.rotation = inheritedRotation;
 	}
-	let path = logOptions.path;
-	if (path) {
-		if (!logOptions.root) logOptions.root = pathModule.dirname(path);
-	} else if (logOptions.root) {
-		path = join(logOptions.root, logName);
-	} else {
-		path = mainLoggerRef.path;
-		if (!logOptions.root) logOptions.root = pathModule.dirname(path);
-	}
+	// After the rotation above: assigning `path` goes through the setter that rebuilds this path's
+	// file logger, and it reads `logger.rotation` as it stands at that moment.
 	if (path) logger.path = path;
 	else console.error('No path for logger', logOptions);
 	logger.level = LOG_LEVEL_HIERARCHY[logOptions.level] ?? mainLoggerRef?.level ?? LOG_LEVEL_HIERARCHY.info;
@@ -764,6 +771,8 @@ export function createLogger(options: any = {} as any) {
 const LOG_TIME_USAGE_THRESHOLD = 100;
 // How long to write straight to stdio after the log file refuses an append.
 const APPEND_RETRY_COOLDOWN = 5000;
+// Ceiling on how often a failing rotation may report itself; the failure repeats every check point.
+const ROTATION_REPORT_INTERVAL = 60000;
 /**
  * Get the file logger for the given path. If it doesn't exist, create it.
  * @param path
@@ -775,14 +784,31 @@ function getFileLogger(path, rotation, isExternalInstance) {
 	let logFD, loggedFDError, loggedAppendError, logTimer, retryAppendAfter;
 	let logBuffer;
 	let logTimeUsage = 0;
+	let rotationGuard,
+		logFDIdentity,
+		nextRotationReport = 0;
 	if (!logger) {
 		logger = logToFile;
 		logger.closeLogFile = closeLogFile;
+		// Only the first call for a path keeps its closure; later calls get the cached logger back and
+		// their own logFD/logBuffer are dead. The guard has to be installed through the live closure.
+		logger.installRotationGuard = installRotationGuard;
 		logger.path = path;
 		fileLoggers.set(path, logger);
 	}
-	if (isMainThread && JSON.stringify(rotation) !== JSON.stringify(logger.rotation)) {
+	// An undefined rotation means "this caller has no opinion", not "turn rotation off" — that is
+	// what `enabled: false` says. Several loggers are created for one path during startup and the
+	// later ones carry no rotation block of their own, so treating undefined as a reconfiguration
+	// tears down the rotation the earlier, configured caller asked for (the same shape of loss #1880
+	// fixed for updateLogger's inheritance).
+	const reconfigured = rotation !== undefined && JSON.stringify(rotation) !== JSON.stringify(logger.rotation);
+	if (reconfigured) {
 		logger.rotation = rotation;
+		// Every writing thread, and synchronously: request logging is produced by the HTTP workers, and
+		// at the rates that make maxSize matter a deferred guard misses megabytes before it exists.
+		logger.installRotationGuard(rotation);
+	}
+	if (isMainThread && reconfigured) {
 		setTimeout(() => {
 			logger.rotator?.end();
 			if (!rotation) return;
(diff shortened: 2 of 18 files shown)


server/threads/logRotationTransport.ts
```diff
@@ -0,0 +1,40 @@
...
+	broadcast(message) {
+		broadcast(message);
+	},
+	sendToThread(target, message) {
+		threads.sendToThread(target, message);
+	},
+	onMessage(type, handler) {
+		onMessageByType(type, handler);
+	},
+	onThreadExit,
```
dawsontoth, line 28: Might want to save an issue for a future cleanup: this smells like something that will be brittle later. We're wrapping manageThreads.js as a rotation transport with a lot of stuff that the logic in this file doesn't really care about, peerThreadIds is the real customization we're making. Subclassing would probably capture this scenario better without us having to specify every method in the wrapper.

dawsontoth (approved):

dawsontoth:
also, windows seems ticked off in the integration tests. Dunno if that's caused by this PR, or something broken on main.

utility/logging/logRotation.ts
```diff
@@ -0,0 +1,319 @@
...
+ * quantum of this writer's own output.
+ */
+export function createRotationGuard(options: any) {
+	const { logPath, maxBytes, rotatedLogDir, compress, getLogIdentity, closeLogFile, report, onRotated } = options;
+	const checkQuantum = Math.max(1, Math.floor(maxBytes / CHECK_QUANTUM_DIVISOR));
+	mkdirSync(rotatedLogDir, { recursive: true });
+	// Rotation is a rename, and a rename across devices can never succeed. Refusing to build the guard
+	// here turns a misconfiguration that would otherwise fail closed on every write — ending file
+	// logging for the life of the process — into one startup error and today's unrotated behavior.
+	if (statSync(rotatedLogDir).dev !== statSync(dirname(logPath)).dev) {
```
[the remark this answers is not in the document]
kriszyp: Confirmed and fixed in ac2e26be — `createRotationGuard` now `mkdirSync`s `dirname(logPath)` alongside `rotatedLogDir`, and a unit test builds a guard whose log directory does not exist yet and asserts it rotates (it fails on the previous commit with exactly this `ENOENT`).

One correction to the framing, because it changes who is exposed: a *default* clean install is not affected. `logging.rotation.path` is defaulted to `log` by `config/configUtils.ts:622-623` and resolved against `rootPath` by `validation/configValidator.ts:604-605`, and `logging.root` defaults to the same directory — so `mkdirSync(rotatedLogDir, { recursive: true })` already created `dirname(logPath)` on that path. What actually triggers it is `logging.root` moved away from its default while `rotation.path` is left alone: the archive directory is then somewhere else entirely, and the log directory the sink would have created on its first append does not exist yet when the guard stats it.

That same divergence had a louder consequence on Windows CI, where the two defaults land on different volumes: the guard refused to build (correctly — rotation is a rename), and reporting the refusal hit a temporal-dead-zone `ReferenceError` that killed every Harper boot on the platform. Both are fixed in the same commit.

utility/logging/logRotation.ts
```diff
@@ -0,0 +1,303 @@
...
```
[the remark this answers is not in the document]
kriszyp: Confirmed and fixed in ac2e26be — `createRotationGuard` now `mkdirSync`s `dirname(logPath)` alongside `rotatedLogDir`, and a unit test builds a guard whose log directory does not exist yet and asserts it rotates (it fails on the previous commit with exactly this `ENOENT`).

One correction to the framing, because it changes who is exposed: a *default* clean install is not affected. `logging.rotation.path` is defaulted to `log` by `config/configUtils.ts:622-623` and resolved against `rootPath` by `validation/configValidator.ts:604-605`, and `logging.root` defaults to the same directory — so `mkdirSync(rotatedLogDir, { recursive: true })` already created `dirname(logPath)` on that path. What actually triggers it is `logging.root` moved away from its default while `rotation.path` is left alone: the archive directory is then somewhere else entirely, and the log directory the sink would have created on its first append does not exist yet when the guard stats it.

That same divergence had a louder consequence on Windows CI, where the two defaults land on different volumes: the guard refused to build (correctly — rotation is a rename), and reporting the refusal hit a temporal-dead-zone `ReferenceError` that killed every Harper boot on the platform. Both are fixed in the same commit.

utility/logging/logRotation.ts
```diff
@@ -0,0 +1,319 @@
...
+	let files;
+	try {
+		files = await fsProm.readdir(rotatedLogDir);
+	} catch (error) {
+		if (error.code !== 'ENOENT') throw error;
+		return;
+	}
+	let compressed = 0;
+	for (const file of files) {
+		if (!file.endsWith('.log') || !isArchiveName(file) || files.includes(`${file}.gz`)) continue;
```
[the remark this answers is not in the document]
kriszyp: Agreed and fixed in ac2e26be — `files` is converted to a `Set` once before the loop.

Worth stating why it matters more here than the shape suggests: this change makes the archive rate `write-rate / maxSize` by definition, so the listing this scans is unbounded between retention passes rather than the handful of files a 60-second interval rotation used to leave.

utility/logging/logRotation.ts
```diff
@@ -0,0 +1,303 @@
...
+ * Compress archives left plain by any isolate, from a listing taken before quiescence was proven.
+ * Write-path rotations happen mostly in the HTTP workers, whose unproven-archive bookkeeping the
+ * main thread cannot see, so without this a worker's archive would silently stay uncompressed for
+ * the life of the process however the operator configured `compress`.
+ * Only safe to call once quiescence has been proven for exactly this listing.
+ */
+export async function compressPendingArchives(rotatedLogDir: string, files: string[], liveLogPaths: Set<string>) {
+	let compressed = 0;
+	for (const file of files) {
+		if (!file.endsWith('.log') || !isArchiveName(file) || files.includes(`${file}.gz`)) continue;
```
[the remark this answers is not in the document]
kriszyp: Agreed and fixed in ac2e26be — `files` is converted to a `Set` once before the loop.

Worth stating why it matters more here than the shape suggests: this change makes the archive rate `write-rate / maxSize` by definition, so the listing this scans is unbounded between retention passes rather than the handful of files a 60-second interval rotation used to leave.

utility/logging/harper_logger.ts
```diff
@@ -109,7 +111,16 @@ export function updateLogger(logger: any, logOptions: any, name?: string, mainLo
...
-	if (inheritedRotation && inheritedRotation === mainLoggerRef?.rotation) {
+	let path = logOptions.path;
+	if (path) {
+		if (!logOptions.root) logOptions.root = pathModule.dirname(path);
+	} else if (logOptions.root) {
+		path = join(logOptions.root, logName);
+	} else {
+		path = mainLoggerRef.path;
+		if (!logOptions.root) logOptions.root = pathModule.dirname(path);
+	}
```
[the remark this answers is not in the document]
kriszyp: Checked and declining this one.

`path = mainLoggerRef.path` is verbatim from `origin/main` — this PR moved the block that contains it above the rotation-path strip, but did not change the expression. And the suggestion would not achieve what it is asking for: `updateLogger` ends with

```ts
logger.tag = logOptions.tag ?? ((mainLoggerRef.path === logger.path || externalLogger.path === logger.path) && name);
```

which dereferences `mainLoggerRef.path` — and `externalLogger.path` — unconditionally on **every** call, not just the `else` branch. A null `mainLoggerRef` throws there regardless, so adding `?.` at the earlier site would be decoration rather than a guard. Making this function genuinely null-safe is a change to a line this PR does not touch, on a path no caller reaches today; it belongs in its own change if it is worth making at all.

server/threads/logRotationTransport.ts
kriszyp: Good call, and filed: [#2534 — logRotationTransport hand-wraps four manageThreads exports it does not customize](https://github.com/HarperFast/harper/issues/2534), parented to the logging-hardening epic (#1661).

One note captured in the issue: subclassing is not reachable as written — `manageThreads.js` is a CommonJS module rather than a class, and `threads` is the `connectedPorts` array with `sendToThread`/`onMessageByType` bolted onto it. The equivalent cleanup is for `manageThreads` to export the narrow mesh interface itself so this file supplies only `peerThreadIds` and the job-worker eligibility decision behind it. The issue also carries the related `require('./logRotationTransport.ts')` at the bottom of `manageThreads.js`, which is load-order-sensitive for the same dependency-direction reason and would be better as an explicit init call from the server bootstrap.

kriszyp:
> also, windows seems ticked off in the integration tests. Dunno if that's caused by this PR, or something broken on main.

It was this PR. `main` was green on Windows at the time (run 34280999766, all six shards), and here all six failed while Node, Bun and uWS passed — the tell that it was platform-specific rather than flaky.

The mechanism, from the shard logs:

```
ReferenceError: Cannot access 'ROTATION_REPORT_INTERVAL' before initialization
    at reportRotationProblem (utility/logging/harper_logger.ts:863)
    at logToFile.installRotationGuard (utility/logging/harper_logger.ts:856)
    at getFileLogger (utility/logging/harper_logger.ts:812)
    at createLogger (utility/logging/harper_logger.ts:706)
    at initLogSettings (utility/logging/harper_logger.ts:511)
    at Object.<anonymous> (utility/logging/harper_logger.ts:382)   <- module scope
```

`initLogSettings()` is called from `harper_logger`'s own module scope, so it reaches `getFileLogger` while the module is still evaluating — and `ROTATION_REPORT_INTERVAL` was declared 380 lines further down, still in its temporal dead zone. Any configuration in which `createRotationGuard` throws turned its one-line failure report into a `ReferenceError` thrown out of `require`, which is why suites with nothing to do with logging (`static plugin with urlPath`, `MCP application profile row-level RBAC`, `REST WebSocket upgrade`) went down with it: no Harper could boot.

Windows is where the guard throws, and that part is working as designed. `logging.rotation.path` resolves against `rootPath` while the test harness points `logging.root` at `runner.temp` — `C:` and `D:` on the runners — and rotation is a rename, so the guard refuses to build across devices rather than failing every write forever. On Linux both land on one filesystem, the guard builds, and the report never fires.

Fixed in ac2e26be: the three file-sink constants moved above the module-scope call, with a unit test that loads `harper_logger` in a child process against a rotation target that cannot be created (`ENOTDIR`, portable) and asserts the module finishes loading and reports to stderr. It reproduces the `ReferenceError` on the previous commit. The rotation integration test's archive directory also moved off `os.tmpdir()` to beside the runner's log directory, so it is no longer cross-volume there either.

Two other review findings landed in the same commit — the missing `mkdir` of the log directory before the device check, and the quadratic `files.includes` in the compression sweep. Details and the four claims I declined are in the PR description.

utility/logging/logGenerationCoordinator.ts
```diff
@@ -0,0 +1,183 @@
...
+ */
+function requestRelease(message: any): Promise<{ released: boolean; liveLogPaths: Set<string> }> {
+	releaseLocally(message);
+	const liveLogPaths = new Set<string>(sinksByPath.keys());
+	// No transport in a worker isolate means the mesh cannot be enumerated yet, not that there is
+	// nobody to ask: harper_logger installs the size guard at its own module load and manageThreads
+	// installs the transport at the end of its, so a worker rotating in that window would otherwise
+	// call an empty peer set a proof and unlink the generation its siblings are appending to. On the
+	// main thread the same state means no worker has been spawned, where zero peers is the truth.
+	if (!transport) return Promise.resolve({ released: isMainThread, liveLogPaths });
```
DavidCockerill, line 105: **The release proof is process-scoped, but the log file is not.**

The comment above this line reasons about exactly the right shape and then stops one step short: *"On the main thread the same state means no worker has been spawned, where zero peers is the truth."* That's true of **this** process. The inode is shared across processes.

Any process that loads `harper_logger` without `manageThreads` has no transport, so `released` comes back on the strength of `isMainThread` alone — the `harper` CLI, `installer.ts:213` (`hdbLogger.initLogSettings(true)`), any short-lived tool. Those install the write-path guard too, and the guard seeds against the file as it actually is:

```js
// logRotation.ts:241
bytesUntilCheck = Math.min(checkQuantum, Math.max(0, maxBytes - statSync(logPath).size));
// → 0 when the log is already over the cap, so the FIRST write rotates
```

So: run a CLI command against a live instance whose `hdb.log` has just crossed the cap, and that process's first notify line rotates it, certifies "no peers", and — with `compress: true` — unlinks the archive at `logRotation.ts:211` while the running server's threads still hold descriptors on that inode. Their appends land in a deleted file until their own checkpoint or the 10s fd timeout.

**I don't think this blocks the merge**, and I want to be clear about why rather than leave it ambiguous: `compress` defaults to `false`, and today's `moveLogFile` destroys archives with no proof whatsoever, so this is strictly better than main either way. It's a residual gap in a new mechanism.

The reason I'm raising it as the lead anyway is that it's the same claim the module header makes — that an answer from every peer rules out a write that hasn't happened yet — and the answer set is peers-in-this-process. **The question for you:** is `compress: true` a configuration you intend to support? If yes, this wants a cross-process guard (an inode-holder check, or a lock file in the rotated dir) before it ships. If it's effectively internal-only, a line in the header narrowing the claim to "peers in this process" would keep the next reader honest.

— DAIvid (Claude Opus 5)

server/threads/logRotationTransport.ts
```diff
@@ -0,0 +1,40 @@
...
+	peerThreadIds() {
+		// Job workers included: they hold log descriptors whether or not they take part in ITC gossip,
+		// and the deadlock that excludes them from the schema broadcast cannot happen here — the
+		// handler only closes a descriptor, and no caller blocks its event loop on the answer.
+		const ids = new Set<number>();
+		for (let i = 0; i < threads.length; i++) {
```
DavidCockerill, line 34: **The peer set can be snapshotted before a newly spawned worker is in it.** Same class as the coordinator finding, narrower trigger.

`threads` is `connectedPorts` (`manageThreads.js:92` assigns it), and a peer only learns about a new worker when it **processes** the `ADDED_PORT` message (`manageThreads.js:1478`). `addPort()` at `:1467` pushes synchronously, but only for the isolate doing the spawning. Meanwhile `requestRelease` takes `expected` synchronously, inside the rotating thread's append.

So a worker whose event loop is blocked across the newcomer's whole bootstrap rotates, gets acked by every peer it knows about, and can compress a generation the unlisted newcomer just opened.

The main thread is safe here, and worth stating so it doesn't get "fixed" unnecessarily: its set is populated synchronously at spawn, so it's conservative.

This needs the rotating worker's loop blocked for the newcomer's entire startup — plausible only under the log storm this change targets, which is exactly when it would happen. The property a fix needs is that the expected set is a superset of every isolate that could hold a descriptor at snapshot time; the cross-process variant needs only a CLI invocation, so I'd let that one drive the design and let this fall out of it.

Separately, and only because it changed my read of this function: the docblock's "job workers included" reasoning is sound, and I traced the deadlock argument — the handler really does only close a descriptor and no caller blocks on the answer. That part is fine.

— DAIvid (Claude Opus 5)

utility/logging/harper_logger.ts
```diff
@@ -775,19 +788,41 @@ function getFileLogger(path, rotation, isExternalInstance) {
...
+		// Registered by the sink, not by the size guard: a thread that writes this file but has no
+		// guard (no maxSize, or rotation driven only by interval) still holds a descriptor, and an
+		// answer from it has to mean "released" rather than only "handler ran".
+		registerLogSink(path, { identity: () => logFDIdentity, close: closeLogFile });
 	}
-	if (isMainThread && JSON.stringify(rotation) !== JSON.stringify(logger.rotation)) {
+	// An undefined rotation means "no opinion", not "off" — that is what `enabled: false` says.
+	// Several loggers are created for one path during startup and the later ones carry no rotation
+	// block, so treating undefined as a reconfiguration tore down the configured caller's rotation.
+	const reconfigured = rotation !== undefined && JSON.stringify(rotation) !== JSON.stringify(logger.rotation);
```
DavidCockerill: **minor** — deleting the `logging.rotation` block at runtime is now a silent no-op.

`reconfigured` requires `rotation !== undefined`, so a config edit that removes the block leaves both the installed write guard and the audit rotator running with the previous settings. The old path (`logger.rotator?.end(); if (!rotation) return;`) tore them down.

This is reachable rather than theoretical: `RootConfigWatcher` re-applies config on change, and `updateLogger` sets `logger.rotation = undefined` for the main logger when the block is gone, which the `path` setter passes straight through.

The undefined-means-no-opinion rule is what fixes the startup teardown your comment describes, so I'm not suggesting reverting it — and `enabled: false` still works, which is probably what most operators reach for. A carve-out for "was defined, now undefined", or just a line in the rotation docs saying `enabled: false` is the way to turn it off, would close the gap between what an operator expects and what happens.

— DAIvid (Claude Opus 5)

utility/logging/logRotation.ts
```diff
@@ -0,0 +1,338 @@
...
+export async function compressPendingArchives(rotatedLogDir: string, files: string[], liveLogPaths: Set<string>) {
+	const present = new Set(files);
+	let worked = 0;
+	for (const file of files) {
+		if (!file.endsWith('.log') || !isArchiveName(file)) continue;
+		const archivePath = join(rotatedLogDir, file);
+		if (liveLogPaths.has(resolve(archivePath)) || unprovenArchives.has(archivePath)) continue;
+		// Counted after the skips, not before: the archive directory holds the live logs too, and
+		// spending the pass's budget on files it then skips is what leaves a backlog standing.
+		if (worked++ >= MAX_RETRIES_PER_PASS) return;
```
DavidCockerill: **minor** — with `compress: true`, compression can't keep up with its own trigger rate, and the only symptom is the wrong file extension.

`compressArchive` declines rather than queues while a compression is in flight for the directory, and this sweep does at most `MAX_RETRIES_PER_PASS` (4) archives per 60s audit interval. At the default 64M, an instance rotating faster than four times a minute accumulates plain archives permanently. At the 64K-class caps the tests use, essentially every archive stays plain.

Failures are also swallowed (`:176`, `:173`), so there's no operator-visible signal — the observable outcome is `.log` files where the operator configured `.gz`.

Worth saying what this **isn't**: byte volume is no worse than main (same bytes, more files), so it's not a disk-exhaustion path. It's "the setting isn't honored under load", which is a smaller thing but still a promise the config makes.

The budget itself looks deliberate — counting after the skips rather than before is a real fix, per your comment. It's the interaction with the decline-while-in-flight behaviour that leaves the backlog standing.

— DAIvid (Claude Opus 5)

DavidCockerill (commented):
The steady-state mechanism reads well and the reasoning is written down where it needs to be — the synchronous rotate-then-close with nothing awaited between, the fail-closed acknowledgement timeout, and the exactly-once handling of buffered writes all hold up under tracing. Moving the cap onto the write path is clearly the right call; a minute of overshoot at request-log rates is a lot of bytes.

Five notes inline. The lead one is a **cross-process** gap in the release proof, and it's the only one I'd want an answer on before merge — not because I think it blocks, but because it turns on a design question only you can settle.

Nothing here is a regression. The pre-change `moveLogFile` destroyed archives with no proof at all, so every one of these is a residual gap in a new safety mechanism rather than something this PR breaks.

One process note: no linked issue that I can see — `closingIssuesReferences` is empty and there's no `Closes`/`Refs` in the body.

**Coverage, so you can weight this:** the first engine pass lost three of four legs (both graded attempts and the domain lens timed out), leaving one model unadjudicated — and its headline finding was wrong. I re-ran to recover the other two, and that's what produced the lead finding below. Worth knowing that one of my own findings from the first pass didn't survive the re-run: I'd flagged the worker-side `unprovenArchives` map as never drained, and `compressPendingArchives`'s docblock already says exactly why that's handled. My mistake, and the docblock is why it was findable.

— DAIvid (Claude Opus 5)

utility/logging/logRotator.ts
```diff
@@ -67,72 +79,144 @@ function logRotator({ logger, maxSize, interval, retention, enabled, path: rotat
...
+				if (released && (retention || reclamationPriority)) {
+					// remove old logs after retention time
+					// adjust retention time if there is a reclamation priority in place
+					const retentionMs = convertToMS(retention ?? '1M') / (1 + reclamationPriority);
+					reclamationPriority = 0; // reset it after use
+					for (const file of candidates) {
+						try {
+							const archivePath = path.join(rotatedLogDir, file);
+							// The rotated directory defaults to the log directory itself
+							// (`logging.rotation.path` defaults to `log`, as does `logging.root`), so it also
```
DavidCockerill: **nit** — the premise in this comment doesn't hold with default config.

It justifies the live-log protection with "`logging.rotation.path` defaults to `log`, as does `logging.root`", but the default is `null` (`static/defaultConfig.yaml:53`), which `resolveRotatedLogDir` turns into `<log dir>/rotated` — a separate directory. The same wording appears in `logRotator.test.js:105`.

The protection is correct and worth keeping, and the test genuinely exercises it (`runRotator` passes `path: LOG_DIR_TEST`, which does collapse the two). It's only the stated reason that's off: the overlap happens when an operator sets `rotation.path` explicitly to the log dir, not by default. Worth fixing because the comment currently reads as "this always overlaps", which would make someone think the guard is load-bearing on every install.

— DAIvid (Claude Opus 5)

server/threads/logRotationTransport.ts
kriszyp: Confirmed. The main-thread roster is conservative, but an existing worker can snapshot `connectedPorts` before its `ADDED_PORT` handler admits the newcomer. A second local reread does not establish a point-in-time superset; closing this requires an authoritative main-thread relay or membership epoch/barrier. I have not added a partial fix. The limitation and concrete risk are now prominent in the draft's human-review section so the lifecycle design remains an explicit maintainer decision.

utility/logging/logRotation.ts
kriszyp: Fixed in `b2fc8f76f` and `6d6155a0b`. Catch-up now runs to a wall-clock budget derived from the audit interval instead of four archives per minute; retention runs first; a busy directory slot stops the sweep for retry; and one gzip failure is reported while later files in the snapshot continue. Tests cover more than four archives, deadline exhaustion/resume, busy compression, failure continuation, and shutdown.

utility/logging/harper_logger.ts
kriszyp: Fixed in `b12579bc2`. Cached sinks now track the configuration source that owns their guard: authoritative runtime removal clears that source, an explicit component policy is not cleared by inherited-main removal, and removing the explicit policy restores inheritance when main still exists. Unit cases cover all ownership transitions, and the two-worker integration test removes `logging.rotation` at runtime and proves neither worker rotates afterward.

utility/logging/logRotator.ts
kriszyp: Fixed. The comments now say the overlap occurs for an explicitly configured rotation path that contains live logs; the default is correctly described as `<log dir>/rotated`. The live-path protection and the explicit-overlap test remain unchanged.

utility/logging/logGenerationCoordinator.ts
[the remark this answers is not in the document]
kriszyp: Fixed. `unitTests/utility/logging/rotation-worker-for-tests.js` now launches a real worker that calls the coordinator before any transport is installed. The test asserts `requestStaleDescriptorRelease()` returns `released: false`, and separately verifies a publish attempt leaves the plain archive present with no `.gz`. The focused logging suite is green at 189 passing / 14 pending.

kriszyp: Confirmed as a real process-boundary limitation, and answered explicitly rather than papering it over. `compress: true` remains supported for the server's enumerable in-process worker mesh; another OS process writing the same path is outside that proof. A safe fix needs a cross-process lease/inode-holder protocol, not a local transport tweak. The module/PR claims are now narrowed to enumerable in-process peers, and the draft's human-review section calls out the CLI scenario and that the shipped `compress: false` / no-retention default is non-destructive.

server/threads/logRotationTransport.ts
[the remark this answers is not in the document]
kriszyp: Addressed in 4c0ef25b3bf0412ddef69fba5099f09f97d91b15: worker-local release still closes peer descriptors promptly, but `publishArchivedGeneration()` now permits destructive gzip publication only on the main thread, whose worker roster is authoritative. Worker archives stay plain until the main-thread audit takes a fresh directory snapshot and obtains its own stale-descriptor proof; the new real-worker test proves that even a successful worker-local release cannot publish `.gz`.

cb1kenobi (commented):
This PR moves the log size cap onto every writing thread's file sink so a busy HTTP worker can no longer grow the active log for a full minute, and adds a peer-release proof so an archived log is only compressed or deleted once every thread has closed its descriptor on it. I traced the guard closure wiring, the rotation-notice re-entrancy, the write-gate ordering, and the prove-then-destroy set, and found no blocking defect on the changed lines. The remaining gaps the author lists — cross-process release, gzip durability, and stricter maxSize validation — are disclosed design decisions rather than bugs.

—
Reviewed 4c0ef25b

Issue #2315, opened 2026-08-25
Staged component deploys: land build-aside-then-swap as a sequence of small PRs


dawsontoth:
Coordination issue for landing staged (build-aside-then-swap) component deploys as a sequence of small, independently mergeable PRs. Replaces the single large effort in #1849 / #2301, both reference-only.

**The claim the sequence is carrying:** one node resolves, every node installs those same bytes, and no node's live tree changes until a candidate exists, has been checked, and can be swapped in as one transaction that includes the root-config entry. Everything else — retention, revert, deploy-from-an-aside — exists to make that transaction reversible after the fact.

```mermaid
flowchart LR
    S1["1 · Build aside,<br/>then swap"]:::done
    S4["4 · Retention of<br/>built asides"]:::done
    S6["6 · Stage now,<br/>activate later"]:::done
    S3["3 · Root config as an<br/>activation effect"]:::next
    S5["5 · Keep the displaced tree,<br/>and make an id mean<br/>something after the swap"]:::next
    S7["7 · Build once,<br/>replicate the artifact"]:::next
    S2["2 · Isolated-worker<br/>validation"]:::shelved

    S1 --> S4 --> S6 --> S5
    S1 --> S3
    S1 --> S7
    S1 -.-> S2

    classDef done fill:#dcfce7,stroke:#16a34a,stroke-width:2px,color:#14532d
    classDef next fill:#fef3c7,stroke:#d97706,stroke-width:2px,color:#78350f
    classDef shelved fill:#fee2e2,stroke:#dc2626,stroke-width:2px,stroke-dasharray:5 4,color:#7f1d1d
```

## Progress

| Step | Status | PR |
|---|---|---|
| 1. Build aside, then swap | **merged** | #2345 |
| 2. Isolated-worker validation | **shelved** (2026-09-14) | #2476 (closed), branch `claude/deploy-worker-validation-step2` @ `9734beaab` |
| 3. Root config as an activation effect | not started | — |
| 4. Retention of built asides | **merged** | #2531 |
| 5. Keep the displaced tree, and make an id mean something after the swap | not started | — |
| 6. Stage a build now, activate it later | **merged** | #2605 |
| 7. Build once, replicate the artifact | not started | — |

## The steps

**1. Build aside, then swap** — merged, #2345. The candidate is built under `.deploy-staging/<id>`, validated there, and activated as two renames with journalled recovery. The live tree stops being a construction site. Not gapless: the live pathname is briefly absent.

[...]

PR #2476, opened 2026-09-02
Certify a candidate in an isolated validator before it can go live

### LINKED ISSUE #2491: Starting a job worker sets the process-wide workerCount to undefined, silently disabling the rolling-restart throttle
## Summary

Starting a job worker sets the process-wide `workerCount` to `undefined`. After any job has run, `server.workerCount` reads `undefined` for components, and the next `restartWorkers()` computes its throttle as `NaN` — which silently disables the rolling-restart throttle instead of falling back.

## Mechanism

`workerCount` is a module-global in `server/threads/manageThreads.js` (declared at line 229, `let workerCount = 1`) and is assigned in exactly one place — inside `startWorker`:

```js
workerCount: (workerCount = options.threadCount),
```

`jobRunner.launchJobThread` (`server/jobs/jobRunner.ts:146-151`, and the `START_JOB` handler just below it) calls `startWorker` without `threadCount`:

```js
threadsStart.startWorker(join(__dirname, './jobProcess.js'), {
    autoRestart: false,
    name: 'job',
    env: { ... },
});
```

so `options.threadCount` is `undefined` and the global is overwritten. Nothing restores it until the next serving-worker start.

Two consequences:

1. **The rolling-restart throttle is disabled.** `restartWorkers()`'s default parameter is
   `maxWorkersDown = Math.max(Math.floor(workerCount / 8), 1)` (line 530). With `workerCount === undefined` that is `Math.max(NaN, 1)` → `NaN`. The guard below it, `if (maxWorkersDown < 1)` (line 565), does **not** catch this, because `NaN < 1` is `false`. `NaN` then flows into the restart loop's comparisons, which are all false — so the intended "restart 1/8 of the threads at a time" throttle does not apply, and the replacement-failure cutoff that depends on the same value does not either.
2. **A public component API reads `undefined`.** `server.workerCount` is exposed to components via `Object.defineProperty(server, 'workerCount', …)` (line 303). A component that sizes a pool, shards work, or elects a leader from it gets `undefined` after any job has run.

## Reproduction

1. Start Harper with more than 8 HTTP threads (so the throttle would otherwise be > 1).
2. Run any operation that launches a job worker — `csv_data_load`, `export_local`, `create_backup`, a bulk `delete`.
3. Read `server.workerCount` from a component, or trigger a restart (`restart`, or a `deploy_component` with `restart: rolling`).

Expected: `workerCount` still reflects the configured thread count; the restart walks 1/8 of the threads at a time. Actual: `undefined`, and the restart is unthrottled.

## Impact

A rolling restart is specifically the mechanism that is supposed to keep the node serving while workers are replaced. Unthrottled, it can take down substantially more capacity at once than intended, which is a service gap during exactly the operation chosen to avoid one. The trigger is structural rather than rare — any job does it — and it is silent: nothing logs, and `NaN` fails the one guard that looks for a bad value.

## Suggested fix

`workerCount` describes the **serving topology**, so a non-topology worker should not write it. Assign it only when `options.threadCount` is provided (or only for `THREAD_TYPES.HTTP` starts), and treat a missing/`NaN` `maxWorkersDown` as the documented minimum of 1 rather than letting `NaN` through the `< 1` guard. A `Number.isFinite` check on the computed default would close the second half independently of the first.

Worth deciding at the same time whether `startWorker` should reject an ephemeral/non-topology start that would mutate topology state at all — the same footgun applies to any future non-HTTP worker.

## Verification route

Unit: start a job-type worker, then assert `getWorkerCount()` / `server.workerCount` is unchanged and that `restartWorkers()`'s default `maxWorkersDown` is a finite number ≥ 1. A regression test that runs a job and then a rolling restart with > 8 threads, asserting the number of simultaneously-down workers stays within the throttle, covers the observable behavior.

## Provenance

Found while re-framing deploy certification (#2315 step 2, PR https://github.com/HarperFast/harper/pull/2476): a planning-mode review flagged that spawning a certification worker through `startWorker` would corrupt `workerCount`, and checking it showed job workers already do. Pre-existing on `main`; not introduced by that PR.

dawsontoth:
`deploy_component` reaches `prepareApplication()` on the **main thread**, and the only check that a
candidate could actually load was gated on `!isMainThread`. So for an operator deploy it ran nothing: a
candidate that installed cleanly and threw at load was renamed live anyway, and `.complete` — which crash
recovery treats as proof a validation happened — was minted for it.

This PR builds the certification mechanism and **lands it switched off** (`HARPER_CERTIFY_DEPLOYS`).
With the switch off, `deploy_component` behaves exactly as it does today. That is the headline decision and
it needs your agreement — see below.

## Why it lands off

A certification load has two non-negotiable properties, and **no available host has both**:

| Host | Serving-equivalent load | Can be force-killed |
|---|---|---|
| A thread in this process | Yes — shares the process's RocksDB handles | **No** under Bun |
| A separate process | **No** — RocksDB's lock is exclusive | Yes |

- **A thread cannot be killed under Bun.** `terminate()` triggers a NAPI segfault there — `manageThreads`
  and `jobProcess.ts` both avoid it, the latter draining its event loop rather than calling `process.exit` —
  so the parent can only *ask* the thread to exit. A candidate that blocks its event loop, or calls
  `parentPort.removeAllListeners()`, never processes the ask: the thread never exits and its concurrency
  slot is held for the life of the process. Two of those and the node stops deploying.
- **A separate process cannot open the databases.** RocksDB's lock is exclusive per process, so a helper
  died with `IO error: While lock file: … Resource temporarily unavailable` the moment `loadRootPlugins`
  reached `getTables()`. `security/auth.ts` calls `table()` at module scope, so loading fewer plugins does
  not avoid it. Opening `readOnly` takes a shared lock and would work, but would then reject any candidate
  that **writes during load** — creating a table, seeding a record — which is a false-rejection class worse
  than the problem being solved.

Landing it on by default therefore means shipping a guarantee that holds only on some runtimes, or a load
that is not the load a serving worker performs. Landing it off keeps the reviewed mechanism, the tests, and
the four independent bug fixes below, and leaves the host decision to its own change with the constraint
known up front.

## For the human reviewer

**The final review verdict is `BLOCK`, and I did not adopt it.** Its blocker is
`components/Application.ts:2203-2207` — "certification is disabled by default, so ordinary deployments
still publish unloadable candidates." That is accurate, and it is the decision above rather than an
oversight: the alternative is one of the two broken guarantees in that table. The reviewer's own
"no host satisfies both requirements" analysis is what produced the switch. If you disagree, the lever is
one line and the on-by-default behaviour is fully tested.

**The planning gate cleared a design that cannot work.** `--mode plan` returned
`better-alternative-exists` for a job-type worker and I adopted its helper-process alternative, which then
returned `chosen-approach-sound`. It fails on the RocksDB lock. The gate reviews reasoning, not viability —
one `fork` would have answered it in a minute, and neither the reviewer nor I ran one before writing the
design note. Worth knowing when weighing a cleared framing verdict on this branch.

**Carried majors, all specific to the on-by-default path** and therefore inert while the switch is off:

- The concurrency cap is **isolate-local**. `prepareApplication()` can run on any worker, so an N-worker
  node admits 2N validators, each a full module graph. Node-wide admission is unbuilt.
- **Candidate subprocesses are not reclaimed.** A candidate can spawn an unref'd descendant that outlives
  the validator thread and mutates the candidate tree after the verdict. The process-group machinery
  (`registerProcessGroup`, `terminateProcessGroup`) exists and is unused by certification.
- **In-process config overrides are not inherited** beyond what the worker `configOverrides` provider
  already carries, and the preload profile for a validator is deliberately off — reasonable for a thread,
  and wrong for a process, if the host changes.
- **The Windows error-detail race is narrowed, not closed.** A rejection's message can still lose to the
  exit; the 250 ms grace makes it unlikely rather than impossible. The shared flag remains the authority.

**Verification you cannot get from me:** Bun and Windows behaviour. I have no Bun runtime here, and the
Windows evidence in this PR comes from CI logs, not from a machine I can step through.

## What is fixed regardless of the switch

Four defects found along the way, each independent of where certification runs:

1. **`package_component` packaged the whole Harper install.** `symlinkHarperModule` links the running
   install into `node_modules/harper` on every non-root load, and the packer dereferences and recurses into
   symlinked directories — so a certification load left a link that turned packaging into a walk of the
   install: 46s of tarring, then `Maximum response size reached`. `certifyCandidate` now snapshots the
   candidate's `node_modules` and removes only the links its own load created, which matters because a
   `file:<directory>` deploy stages a symlink to the developer's own source tree. The pre-existing half —
   packaging a component a worker *has* loaded — is HarperFast/harper#2487.
2. **A job worker corrupted the rolling-restart throttle.** `workerCount` is a module-global written only
   in `startWorker`, and a start omitting `threadCount` set it to `undefined`, after which
   `maxWorkersDown = Math.max(Math.floor(workerCount / 8), 1)` is `NaN` — which the `maxWorkersDown < 1`
   guard does not catch, because `NaN < 1` is false. An unthrottled rolling restart is a service gap during
   the operation chosen to avoid one. Fixes HarperFast/harper#2491.
3. **Every certification leaked RocksDB handles.** The validator reaches `getTables()` via
   `loadRootPlugins` and never closed them; `resources/databases.ts` documents that this leaks
   process-globally and blocks an online `restore_backup` from confirming a database is closed.
4. **The certification slot was released, not handed over.** `active` was decremented before waking a
   waiter, so a caller arriving in that microtask window took the slot and the woken waiter went to the back
   of its own queue; a release offered to a timed-out waiter was swallowed entirely. `active` now never dips.

## Verification

- **Unit** (`unitTests/components/deployCertification.test.js`, 11 tests): main-thread rejection — which
  fails on `main` with "Missing expected rejection" — clean publish, static-only publish, hang/timeout, slot
  exhaustion, the mint gate, safe mode, branch-configured, and **the default-off path** (a candidate that
  throws at load is published and mints no `.complete`).
- **Integration** (`integrationTests/deploy/certified-deploy.test.ts`, 3 tests): drives the real operations
  API and asserts v1 still **answers requests** after a rejected v2 — which step 1's availability test
  explicitly could not show — that a clean v3 publishes, and that the published tree carries no
  `node_modules/harper` link.
- Both suites **opt into the switch explicitly**. Without that they would pass while proving nothing: an
  uncertified deploy publishes, so only the rejection case would notice.
- `integrationTests/apiTests/components.test.mjs` is 25/25 locally with `package_component` at 22ms, down
  from 46s-then-failure.
- Not covered: Bun reclamation, node-wide admission across real workers, descendant reclamation, and
  config/preload inheritance. All are on the dormant path and listed above.

Closes nothing on its own — #2315 step 2 continues in the host change.

<sub>Review-Coverage: authored=claude; ran=codex; declined=gemini,cursor-grok,cursor-composer,domain; rounds=3 @ afea30b47cdf</sub>

<sub>Human-Review-Need: 4 @ afea30b47cdf</sub>


components/deployValidator.ts
```diff
@@ -0,0 +1,106 @@
...
+	} finally {
+		const closes = await Promise.allSettled(Array.from(scopes, (scope) => scope.close()));
+		const failed = closes.filter((result) => result.status === 'rejected');
+		if (failed.length && !reportedError) {
+			throw new AggregateError(
+				failed.map((result) => (result as PromiseRejectedResult).reason),
+				`${componentName} loaded but ${failed.length} scope(s) failed to tear down`
+			);
+		}
+	}
```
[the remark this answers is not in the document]
dawsontoth: Fixed, and the consequence is worse than losing detail: a `loadComponent` throw — a syntax error, an unreadable file — reaches that same `finally`, so a scope failing to close there replaced the candidate's *real* error with a note about its scopes. The operator would have got the symptom instead of the cause, which for a certification failure is the whole message.

Gated on the load having succeeded, as suggested. The `AggregateError` now only fires for the case it was written for: the candidate loaded, and its own teardown is broken.

components/certifyCandidate.ts
```diff
@@ -0,0 +1,238 @@
...
+	}
+	active++;
+	let released = false;
+	return () => {
+		if (released) return;
+		released = true;
+		active--;
+		waiting.shift()?.();
+	};
+}
```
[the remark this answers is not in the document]
dawsontoth: Real risk, and not acted on — because an earlier commit already closed the queue-jumping window from the other side: the slot is claimed before yielding to a waiter, and the wait is a `while` rather than an `if`, so a waiter that loses a race re-checks instead of proceeding.

Your version is stricter — it makes the handoff FIFO rather than merely correct, which does rule out starvation under sustained load, where mine only rules out overcommit. I have left it alone rather than churn the same lines a second time in one PR; if you think the fairness property is worth having on its own, say so and I will take the suggestion as written.

Worth noting a limit neither version fixes: `active` is per-thread module state, so the cap bounds validators per worker thread rather than per process. That is called out in the PR description.

components/certifyCandidate.ts
```diff
@@ -0,0 +1,238 @@
...
+					// `terminate()` triggers a NAPI segfault under Bun; `manageThreads` asks the worker to exit
+					// itself for the same reason.
+					worker.postMessage({ type: 'force-exit' });
+				} else {
+					await Promise.race([worker.terminate(), grace()]);
+				}
+				const outcome = await Promise.race([
+					(exited ?? Promise.resolve()).then(() => 'exited' as const),
+					grace().then(() => 'still-running' as const),
+				]);
```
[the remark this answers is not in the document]
dawsontoth: Fixed as suggested — one grace for the whole operation rather than one per step. The doubling mattered more than it looks: this runs in a `finally` that a deploy is blocked on while holding the preparation lock, so the wait was up to 10s of a component's deploy for a validator already known to be unresponsive.

Structured slightly differently from the snippet — `terminate()`'s rejection is swallowed rather than left to the race, so a terminate that fails fast cannot resolve the `Promise.all` early and report "exited" for a thread that is still running.

components/Application.ts
```diff
@@ -2064,6 +2150,71 @@ export async function markCandidateComplete(
...
+	const { rootApplicationLoadOptions } = await import('./componentLoader.ts');
+	const loadOptions = rootApplicationLoadOptions(application.name, { forCertification: true });
+	if (loadOptions.ok && loadOptions.branchConfigured) {
+		application.logger.warn(
+			`Deploying ${application.name} without certification: a certification load would open the same ` +
+				`database branch the live version is serving from, so it is skipped until validation-scoped ` +
+				`branch storage exists`
+		);
+		return { certified: false };
+	}
```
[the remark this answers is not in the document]
dawsontoth: Fixed — it now throws before spawning. Worth adding why it is more than an optimisation: the thread it used to start would load Harper's entire module graph before failing on the same condition, so the wasted work was a module graph and a heap, not a thread handle.

components/certifyCandidate.ts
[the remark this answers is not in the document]
dawsontoth: Adopted — you were right and I was wrong to defer it. A later reviewer raised the same starvation mechanism as a blocker, and the code did contradict its own comment: it decremented `active` and then woke a waiter that incremented a microtask later, so the slot was released and re-competed for rather than handed over.

`1f8313075` implements it: `active` never dips, the releaser offers the slot along the queue until a live waiter accepts, and an admitted waiter holds the inherited slot without incrementing.

One difference from the suggestion as written, forced by the deadline added in the same area: a waiter can now lapse, and `waiting.shift()?.()` would hand the slot to a caller that has already given up — swallowing it, so a free slot exists with nobody woken for it. Hence the offer loop and a waiter that reports whether it accepted, rather than a plain shift.

No test. The steal needs a caller arriving inside a synchronous microtask window `certifyCandidate` does not expose, and the swallowed-wake half is not observable through the public surface either, since the next caller reads `active` directly. A test that passed with and without the fix would be worse than none, so this is verified by reading against your mechanism.

The per-thread limitation you noted still stands and is unfixed: `active` is module state, so the cap bounds validators per thread, not per process. It is in the PR description as a carried major, and it is inert while certification is off by default.

components/certifyCandidate.ts
```diff
@@ -0,0 +1,449 @@
...
+			const settle = (outcome: CertificationOutcome) => {
+				if (settled) return;
+				settled = true;
+				if (timer) clearTimeout(timer);
+				resolve(outcome);
+			};
+			const fail = (message: string) => settle({ certified: false, error: new Error(message) });
+
+			try {
+				const started = new Worker(entry, {
```
kriszyp: This bare `Worker` bypasses `collectProvidedWorkerData`, including the `configOverrides` provider added so workers inherit `env.setProperty()` values. The validator can consequently load the on-disk root path, database paths, mounts, or storage settings while serving workers use the parent's overrides; its verdict then describes a different environment. This also makes the unit suite depend on an ambient install—the parent test harness's root/database overrides never reach this child. Please introduce a shared non-topology worker bootstrap that carries required inherited configuration, or use the managed spawn path after separating topology registration from option construction.

components/certifyCandidate.ts
```diff
@@ -0,0 +1,449 @@
...
+	}
+	// The load creates `node_modules` when the candidate has none, and an empty directory left behind is
+	// still a difference from the staged tree.
+	if (before.hadNodeModules) return;
+	try {
+		if ((await readdir(nodeModulesDir)).length === 0) await rmdir(nodeModulesDir);
+	} catch {}
+}
+
+let active = 0;
```
kriszyp: This cap is module-local, but `prepareApplication` may run on any Harper thread. Each worker isolate gets its own `active = 0`, so N request workers can spawn up to 2N heavyweight validators concurrently. That defeats the stated process-memory bound and can create exactly the thread/OOM pressure this limit is intended to prevent. Please broker certification through one owning thread or use a genuinely cross-thread semaphore.

components/certifyCandidate.ts
```diff
@@ -0,0 +1,449 @@
...
+	// The same guard `startWorker` applies. A deploy racing shutdown would otherwise spawn a thread the
+	// shutdown path does not know about, and then wait out its deadline on a process that is leaving.
+	if (isProcessShuttingDown()) {
+		const error: any = new Error(`Cannot certify ${appName} while the Harper process is shutting down`);
+		error.statusCode = 503;
+		return { certified: false, error };
+	}
+	let releaseSlot: () => void;
+	try {
+		releaseSlot = await acquireSlot(timeoutMs);
```
kriszyp: The shutdown check occurs before this potentially long slot wait, and there are further awaited filesystem operations before `new Worker`. If shutdown begins while the caller is queued, a later slot handoff still starts a validator that is absent from `manageThreads.workers`; shutdown therefore neither terminates nor awaits it, and it may load databases/components while teardown is in progress. Please re-check shutdown after acquiring the slot and immediately before spawning, or register the validator with a lifecycle owner that atomically excludes new starts once shutdown begins.

server/threads/manageThreads.js
```diff
@@ -378,43 +486,16 @@ function startWorker(path, options = {}) {
...
-			maxOldGenerationSizeMb: maxOldMemory,
-			maxYoungGenerationSizeMb: maxYoungMemory,
-		},
+		resourceLimits,
 		execArgv,
 		argv: process.argv.slice(2),
 		// pass these in synchronously to the worker so it has them on startup:
 		workerData: {
 			...collectProvidedWorkerData(options),
+			...extraWorkerData,
```
kriszyp: Because `extraWorkerData` is spread after provider output, a caller can silently replace any registered provider property—for example `{ configOverrides: ... }` suppresses the authoritative inherited configuration. Provider registration itself rejects name collisions, so this merge path should preserve the same ownership rule. Please reject keys currently owned by `workerDataProviders` (preferably before any side effects), or otherwise make provider values authoritative.

DESIGN.md
```diff
@@ -454,10 +454,111 @@ with its journal intact. Only the sweep itself is best-effort, because it costs
...
+- **A separate process cannot open the databases.** RocksDB takes an exclusive per-process file lock, so a
+  helper process fails with `IO error: While lock file: … Resource temporarily unavailable` the moment
+  `loadRootPlugins` reaches `getTables()` — and `security/auth.ts` calls `table()` at module scope, so this
+  is not avoidable by loading fewer plugins. Opening `readOnly` takes a shared lock and would work, but then
+  any candidate that writes during load — creating a table, seeding a record — is rejected by certification
+  and fine in production, which is a worse failure than the one being prevented.
+
+Two things that are easy to get wrong about the thread host, learned the expensive way:
+
+- **Use `startWorker`, not a bare `new Worker`.** An earlier draft avoided `startWorker` to stay out of the
```
kriszyp: This says the validator must use `startWorker` and explains that the bare path fails on Windows, but `certifyCandidate` currently constructs a bare `new Worker` and the tests skip Windows for exactly that failure. Please update the design to describe the final architecture, or change the implementation to satisfy this documented invariant; leaving both claims makes future work follow mutually exclusive guidance.

DESIGN.md
```diff
@@ -454,10 +454,111 @@ with its journal intact. Only the sweep itself is best-effort, because it costs
...
+  worker (`name: THREAD_TYPES.JOB`) from broadcasts, and the per-peer `MessageChannel` construction is
+  O(workers) for one slow-path deploy. What the bespoke path actually cost was Windows: the thread died
+  _inside its import graph_, before its first statement, with exit code 0 and no `error` event. A thread
+  created by the standard path does not.
+- **`startWorker` cannot take `workerData`.** `...options` is spread into the `Worker` constructor after the
+  bootstrap `workerData`, so passing it replaces `addPorts`/`addThreadIds` and the thread comes up with no
+  ITC wiring at all. Use `extraWorkerData` + `extraTransferList` (merged, reserved keys refused), and
+  `options.noServerStart` for the reserved key a validator needs.
+
+### Diagnosing a validator that reports nothing
```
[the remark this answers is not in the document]
dawsontoth: Correct, and fixed in `afea30b47`. The paragraph described the Windows fix as done when it isn't — and you're right that the consequence is someone removing the skip on the strength of it.

It was prose that survived a revert: written while the helper-process design was still live, and not taken back out when the implementation was.

DESIGN.md now says what ships — a bare `new Worker` that dies in its import graph on Windows (#2494) — that moving to `startWorker` is intended and unstarted, and that the bootstrap plumbing for that migration (`extraWorkerData`, `extraTransferList`, `noServerStart`) landed here and is currently unused by certification. I also removed the claim that a thread created by the standard path *does* load on Windows: that was never established. The evidence proves the bespoke import graph fails, not that mesh membership or the standard bootstrap is what repairs it, and #2494 names the narrow experiment that would tell those apart.

`certifyCandidate`'s docblock is corrected too — it was the third place, arguing the present-tense case against `startWorker` while DESIGN.md called that reasoning half wrong.

server/threads/manageThreads.js
```diff
@@ -561,7 +654,11 @@ async function restartWorkers(
 			const worker = restarting[index];
 			// Terminal shutdown: stop replacing workers mid-loop — the guard for every replacement start below.
 			if (processShuttingDown && startReplacementThreads) break;
-			if ((name && worker.name !== name) || worker.wasShutdown) continue; // filter by type, if specified
+			// One-shot threads are ephemeral by contract (transferred ports force `autoRestart: false`), so they
+			// are never restarted OR replaced: `startCopy` would re-spawn from spent ports and throw
+			// `DataCloneError` synchronously, inside this loop, taking the rest of the restart with it. Left
+			// alone, such a thread finishes its single task or hits its own deadline.
+			if ((name && worker.name !== name) || worker.wasShutdown || worker.isOneShot) continue; // filter by type, if specified
```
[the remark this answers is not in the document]
dawsontoth: Added in `afea30b47`, but only partly — and the part I could not do is worth stating rather than glossing.

Asserted directly: a port-carrying spawn is marked `isOneShot`, and `worker.startCopy()` refuses by name rather than throwing `DataCloneError` from inside `new Worker`.

Not asserted: the restart loop's filter itself. I wrote that test first — spawn a one-shot worker, call `restartWorkers(name, 1, true)`, assert it survives — and it turns out `restartWorkers` performs a real node restart, including reinstalling applications; it shelled out to `npm pack` before I killed it. Driving that from a unit test to cover a one-line `continue` would be slow and fragile, and the failure mode under the old code is a process-wide crash rather than a clean assertion failure. The test comment says this explicitly so the coverage is not overstated.

kriszyp: This condition is shared by restart and shutdown: `shutdownWorkers()` calls `restartWorkers(name, Infinity, false)`. Unconditionally skipping `isOneShot` therefore prevents the shutdown request from ever being sent and lets `shutdownWorkers()` resolve while the worker is still alive. The comment assumes every caller supplies a deadline, but `startWorker` neither requires nor enforces one, so a blocked one-shot can remain indefinitely. Please skip one-shot workers only when replacements are requested; with `startReplacementThreads === false`, shut down the existing worker without calling `startCopy`.

DESIGN.md
```diff
@@ -454,10 +454,117 @@ with its journal intact. Only the sweep itself is best-effort, because it costs
...
+- **A separate process cannot open the databases.** RocksDB takes an exclusive per-process file lock, so a
+  helper process fails with `IO error: While lock file: … Resource temporarily unavailable` the moment
+  `loadRootPlugins` reaches `getTables()` — and `security/auth.ts` calls `table()` at module scope, so this
+  is not avoidable by loading fewer plugins. Opening `readOnly` takes a shared lock and would work, but then
+  any candidate that writes during load — creating a table, seeding a record — is rejected by certification
+  and fine in production, which is a worse failure than the one being prevented.
+
+Two things that are easy to get wrong about the thread host, learned the expensive way:
+
+- **What ships today is a bare `new Worker`, and it does not load on Windows.** The thread dies _inside its
```
kriszyp: This says the validator must use `startWorker` and explains that the bare path fails on Windows, but `certifyCandidate` currently constructs a bare `new Worker` and the tests skip Windows for exactly that failure. Please update the design to describe the final architecture, or change the implementation to satisfy this documented invariant; leaving both claims makes future work follow mutually exclusive guidance.

components/certifyCandidate.ts
```diff
@@ -0,0 +1,450 @@
...
+			const settle = (outcome: CertificationOutcome) => {
+				if (settled) return;
+				settled = true;
+				if (timer) clearTimeout(timer);
+				resolve(outcome);
+			};
+			const fail = (message: string) => settle({ certified: false, error: new Error(message) });
+
+			try {
+				const started = new Worker(entry, {
```
kriszyp, line 308: This bare `Worker` bypasses `collectProvidedWorkerData`, including the `configOverrides` provider added so workers inherit `env.setProperty()` values. The validator can consequently load the on-disk root path, database paths, mounts, or storage settings while serving workers use the parent's overrides; its verdict then describes a different environment. This also makes the unit suite depend on an ambient install—the parent test harness's root/database overrides never reach this child. Please introduce a shared non-topology worker bootstrap that carries required inherited configuration, or use the managed spawn path after separating topology registration from option construction.

components/certifyCandidate.ts
```diff
@@ -0,0 +1,450 @@
...
+	}
+	// The load creates `node_modules` when the candidate has none, and an empty directory left behind is
+	// still a difference from the staged tree.
+	if (before.hadNodeModules) return;
+	try {
+		if ((await readdir(nodeModulesDir)).length === 0) await rmdir(nodeModulesDir);
+	} catch {}
+}
+
+let active = 0;
```
kriszyp, line 150: This cap is module-local, but `prepareApplication` may run on any Harper thread. Each worker isolate gets its own `active = 0`, so N request workers can spawn up to 2N heavyweight validators concurrently. That defeats the stated process-memory bound and can create exactly the thread/OOM pressure this limit is intended to prevent. Please broker certification through one owning thread or use a genuinely cross-thread semaphore.

components/certifyCandidate.ts
```diff
@@ -0,0 +1,450 @@
...
+	// The same guard `startWorker` applies. A deploy racing shutdown would otherwise spawn a thread the
+	// shutdown path does not know about, and then wait out its deadline on a process that is leaving.
+	if (isProcessShuttingDown()) {
+		const error: any = new Error(`Cannot certify ${appName} while the Harper process is shutting down`);
+		error.statusCode = 503;
+		return { certified: false, error };
+	}
+	let releaseSlot: () => void;
+	try {
+		releaseSlot = await acquireSlot(timeoutMs);
```
kriszyp, line 251: The shutdown check occurs before this potentially long slot wait, and there are further awaited filesystem operations before `new Worker`. If shutdown begins while the caller is queued, a later slot handoff still starts a validator that is absent from `manageThreads.workers`; shutdown therefore neither terminates nor awaits it, and it may load databases/components while teardown is in progress. Please re-check shutdown after acquiring the slot and immediately before spawning, or register the validator with a lifecycle owner that atomically excludes new starts once shutdown begins.

server/threads/manageThreads.js
kriszyp, line 506: Because `extraWorkerData` is spread after provider output, a caller can silently replace any registered provider property—for example `{ configOverrides: ... }` suppresses the authoritative inherited configuration. Provider registration itself rejects name collisions, so this merge path should preserve the same ownership rule. Please reject keys currently owned by `workerDataProviders` (preferably before any side effects), or otherwise make provider values authoritative.

kriszyp (commented):
This is a pretty significant amount of code to commit to maintain. My understanding was that we were trying to pursue a staging mechanism for deployments, not a "certification" mechanism. Why do we need to be responsible for this certification? Why isn't it more appropriate to do testing elsewhere (test envs or CI)? If availability of bad deploys is a concern, why not use rolling deploy in coordination with GTM? Do really think this is adequately mimicking the worker threads (and how far do we want to go with that)? And woudln't forcing rollbacks of "bad" applications potentially hinder the debugging of it? (If something is actively broken, that is the best reproduction of a failure). 

server/threads/manageThreads.js
dawsontoth: Fixed in `2c1d27328`. `extraWorkerData` now rejects any key a registered provider owns, before any side effect — the same rule `registerWorkerDataProvider` already applies to registration, on the other path into `workerData`.

You picked the consequential example: overriding `configOverrides` would leave the thread reading on-disk config while its parent runs on `setProperty()` overrides, so a validator's verdict would describe a different environment than the one serving. Tests cover both the `configOverrides` case and a provider registered at runtime, plus the name becoming free again once it unregisters.

components/certifyCandidate.ts
dawsontoth: Fixed in `2c1d27328` — re-checked immediately before the spawn, as you suggested.

You are right that the entry check was the wrong place on its own: the slot wait can last the full certification timeout, and two awaited filesystem calls follow it, so shutdown beginning in that window still produced a validator absent from `manageThreads.workers` — neither terminated nor awaited while it loads databases and components into a process that is tearing down.

I took the first of your two options. The second — registering the validator with a lifecycle owner that atomically excludes new starts once shutdown begins — is the better fix and is the same shape as the cross-thread brokering your other comment asks for; both point at the validator being managed rather than freelance. Left for that work rather than half-done here.

components/certifyCandidate.ts
dawsontoth: Confirmed, not fixed, and I am leaving this thread open rather than closing it.

You are right about the mechanism and about the consequence. The bare `Worker` bypasses `collectProvidedWorkerData` entirely, so `configOverrides` never reaches the validator and its verdict can describe the on-disk environment rather than the one serving. It also makes the unit suite depend on an ambient install, which is why those tests are skipped on Windows (#2494) rather than passing there.

Codex reached the same finding independently for a helper-process variant, so this is not specific to the current spawn — it is what "not using the managed path" costs.

Both of your suggested fixes are the right shape, and they are the same work: a shared non-topology bootstrap, or separating topology registration from option construction so the managed spawn can be used. `extraWorkerData`/`extraTransferList`/`noServerStart` in this PR are the beginning of exactly that and are currently unused by certification. I have not gone further because your review summary questions whether certification should exist at all, and building the bootstrap out on a premise you may reject is the wrong order.

components/certifyCandidate.ts
dawsontoth: Correct, not fixed, thread left open.

The cap's own docstring claims a process-wide bound it does not have: `active` is module state, `prepareApplication` runs on any thread, so N request workers admit up to 2N validators. Gemini raised the same limitation earlier and it is listed as a carried major in the PR description, which is not the same as fixing it.

Both of your options — brokering through one owning thread, or a genuinely cross-thread semaphore — are the same work as the lifecycle ownership the shutdown-race thread points at, and as the managed-spawn question in the `configOverrides` thread. All three are one change: the validator becomes a managed thread rather than a freelance one. That is the natural next step if certification survives your summary question, and the wrong thing to build before it is answered.

Meanwhile the practical bound is that certification is off by default, so the cap admits zero.

dawsontoth:
@kriszyp I do agree, I think this idea came somewhere during the reviews, pointing out that we weren't validating a candidate before trying to flip it to live. If we don't want this mechanism, that's certainly within the realm of reason. What do you think? My thought is that if we wanted behavior like that, we'd probably do it through the instances: bring something out of GTM, do a deployment on it, verify it (through customer logic), and when it's healthy, bring it back up. **Not** by doing it through the component level.

server/threads/manageThreads.js
```diff
@@ -561,7 +662,11 @@ async function restartWorkers(
...
```
kriszyp: This condition is shared by restart and shutdown: `shutdownWorkers()` calls `restartWorkers(name, Infinity, false)`. Unconditionally skipping `isOneShot` therefore prevents the shutdown request from ever being sent and lets `shutdownWorkers()` resolve while the worker is still alive. The comment assumes every caller supplies a deadline, but `startWorker` neither requires nor enforces one, so a blocked one-shot can remain indefinitely. Please skip one-shot workers only when replacements are requested; with `startReplacementThreads === false`, shut down the existing worker without calling `startCopy`.

dawsontoth:
Back to draft and set aside — the juice may not be worth the squeeze. We may return to it.

Nothing here is abandoned quietly: #2315's step 2 section now carries the full state, including a decision that epic had recorded as settled and which evidence has since contradicted.

**Why it stopped, in order of weight:**

1. **No available host satisfies both requirements.** A certification load must be serving-equivalent, and the validator must be terminable. A thread shares this process's RocksDB handles, so its load is real — but under Bun `terminate()` NAPI-segfaults, so the parent can only *ask* it to exit, which a candidate defeats by blocking its event loop. A separate process is killable but cannot open the databases at all: RocksDB's lock is per process, so it dies with `IO error: While lock file: … Resource temporarily unavailable` the moment `loadRootPlugins` reaches `getTables()`. `readOnly` would open, and would then reject any candidate that writes during load — a worse failure than the one being prevented.
2. **@kriszyp's premise question is unanswered and upstream of everything else here** — particularly whether forcing rollback of a broken application hinders debugging, since a live failure is the best reproduction of itself. Two of his four inline findings are deliberately left open rather than resolved, because answering them with code means building the managed-thread work on a premise that may be rejected.

**Worth lifting out if this is abandoned rather than paused** — four defects that are real today, independent of certification:

- `package_component` follows `node_modules/harper` into the entire Harper install (#2487).
- A validator leaks process-global RocksDB handles by never calling `closeLoadedDatabases()`, which blocks an online `restore_backup` from confirming a database is closed.
- A `startWorker` caller with transferred ports crashes the **process** on restart, because `startCopy` re-spawns from spent ports and the `DataCloneError` escapes an `exit` listener.
- The certification slot was released rather than handed over, so a waiter could be queue-jumped and a lapsed waiter could swallow a free slot.

The `startWorker` bootstrap additions (`extraWorkerData`, `extraTransferList`, `noServerStart`, provider-ownership collision checks, `terminateProcessGroup`) are also independently useful and currently unused by certification.

Remaining CI red on this branch is entirely inherited from `main`: #2275 (MQTT retained-record ordering, bisected to #2497) and #2496 (RocksDB `drop_table` column family, bisected to #2492).

Also filed along the way: #2494 (validator dies in its import graph on Windows), #2491 (job worker corrupts the rolling-restart throttle).

dawsontoth:
Shelved. Closing this PR rather than leaving it open as a draft — the work is preserved on `claude/deploy-worker-validation-step2` at `9734beaab`, which is not deleted, and this PR is the record of how to get back to it.

**Not closed because it was wrong, closed because the premise is unsettled.** @kriszyp's question on this PR is the thing to answer before any of it resumes: why certification is Harper's responsibility rather than something done in test environments or CI, why rolling deploy coordinated with GTM is not the answer if availability of bad deploys is the concern, and whether forcing rollback of a broken application *hinders* debugging, since a live failure is the best reproduction of itself. If that premise does not hold, the honest scope is step 1's staging plus rollback, with no certification at all.

The blocking technical constraint, for whoever picks this up: **no available host satisfies both requirements.** A certification load must be serving-equivalent *and* the validator must be terminable, and nothing is both — a thread shares this process's RocksDB handles so its load is real, but under Bun `terminate()` NAPI-segfaults so the parent can only ask it to exit; a separate process is killable but cannot open the databases at all, because RocksDB's lock is per process. Written up in full in #2315's step 2 section, along with a decision that epic had recorded as settled and which evidence contradicted.

Four defects found along the way were real independent of certification. One is fixed and merged (#2487 → #2585). The rest are filed with analysis: #2577, #2494, #2491. #2494 is closed alongside this, since the code it describes exists only on this branch.

No CI state here is meaningful — its last failures were inherited from `main` (#2275, #2496), not from this branch.

PR #2501, opened 2026-09-04
Take install_node_modules off the npm registry so the Node 22 unit job stops timing out

kriszyp:
`Unit Test (Node.js v22)` has hit its 10-minute `timeout-minutes` cap on every `main` run since 2026-09-03T22:01Z. GitHub reports a job that hits that cap as `cancelled`, not `failure`, so the required check never went red and the entire Node-22 unit suite has been silently skipped since.

## Root cause

The leg stalls in `unitTests/utility/npmUtilities.test.js` › `install_node_modules` › `installs when dry_run is false` — the first case in that suite that runs a real, non-dry-run `installModules()`. That spawns `npm install --force --omit=dev --json` against a fixture whose only dependency is `file:../dependency`.

npm 10.9.x — the npm bundled with Node 22, and only with Node 22 — puts that `file:` link into its audit bulk request and POSTs it to the registry. npm 11 (Node 24/26) sends an empty payload and issues no request at all:

| npm | `silly audit bulk request` | real install |
|---|---|---|
| 10.9.8 (Node 22) | `{ 'local-dependency': [ '1.0.0' ] }` | hangs |
| 11.17.0 (Node 24) | `{}` → `audit report null` | 0.506s |

The registry no longer answers that payload. Measured directly against `POST https://registry.npmjs.org/-/npm/v1/security/advisories/bulk`:

```
{"lodash":["4.17.21"]}          -> 200 in 0.18s
{"local-dependency":["1.0.0"]}  -> no response within 60s
```

An npm debug log from one attempt that did eventually return records `http fetch POST 200 .../security/advisories/bulk 160214ms`. The same install took 467ms on the last green run, so this is a registry-side latency change rather than a code change — the single-commit bisect onto #2493 is a false positive; that commit touches `resources/search.ts` and tests under `unitTests/resources/`, none of which this path reaches. What #2493 *did* contribute is ~2 minutes of new `test:unit:resources` time, which is why the v24/v26 legs are now landing at ~9m50s of a 10-minute budget.

Nothing bounded the wait below the job cap: `.mocharc.json` sets `timeout: 0`, so no individual test can time out, and `nonInteractiveSpawn` defaults to `DEFAULT_COMMAND_TIMEOUT_MS` — one hour.

## The fix

The two npm entry points having separate argument lists is how this happened: the automatic component install has always passed `--no-audit --no-fund`, and `install_node_modules` hand-rolled its own list without them. Rather than adding the flags in a second place, [`installModules` now composes its arguments with the same builder](https://github.com/HarperFast/harper/pull/2501/changes?w=1#diff-aab51787131300406743175a786a0896ca9bdc7ce9b86458c3d1fdc36cf34aebR41), exported from `components/Application.ts` as [`packageManagerInstallArguments`](https://github.com/HarperFast/harper/pull/2501/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R2941), so a registry-avoidance flag can no longer be added to one path and missed by the other. `allowInstallScripts: true` reproduces the operation's existing behaviour exactly — its argv is byte-identical to before plus `--no-audit --no-fund`.

`--no-audit` is the flag that matters; `--no-fund` only suppresses the funding message. The suite goes from wedged-past-10-minutes to 202ms on Node 22.

A [new case pins the composed argv](https://github.com/HarperFast/harper/pull/2501/changes?w=1#diff-f0ada45d0d4b92fce70a6e89934360d60e453be2bd36f99b55c6b4a478699c53R119) by putting an `npm` shim on `PATH` and asserting what the real spawn receives — the idiom `unitTests/components/applicationInstall.test.js` already uses for the automatic path. Removing either flag fails it (verified).

## Budget and loud failures

Every step in the job now carries its own `timeout-minutes`, mirroring what the Windows job already does: a step that exceeds its cap **fails**, so the job's conclusion is `failure` and not `cancelled`. [The job cap becomes a pure backstop](https://github.com/HarperFast/harper/pull/2501/changes?w=1#diff-d4223b44df678cece85f2542c19b9cd060d43b22f753c25214d5942ef2ca5d18R33) — set above the sum of the step caps (41 on the longest leg) so a stall can only ever surface as a [step failure on `Run tests`](https://github.com/HarperFast/harper/pull/2501/changes?w=1#diff-d4223b44df678cece85f2542c19b9cd060d43b22f753c25214d5942ef2ca5d18R88). That is why it moves 10 → 45; the numbers that actually bound the run are the step caps, each roughly 2x its measured cost.

Measured on the v22 leg of run 33835237287: checkout + Node setup 6s, dependency install 2m24s, build 10s, type contract tests 1s, benchmark tests 1s, Harper setup 5s, and the unit run ~7m20s (taken from the v24 leg, which completed). That is ~10m05s against the old 10-minute cap — the leg had no headroom left even without the hang.

The real-npm cases also set their own 60s mocha timeout, so the same class of stall fails the run in the dev loop too, not only under a GitHub step cap.

## For the human reviewer

- **`install_node_modules` responses no longer carry npm's `audit` or funding fields.** The operation is deprecated and passes npm's raw `--json` output through verbatim, so this is a change in what npm reports rather than in Harper's own contract — but it is observable to anyone parsing those fields. `--no-fund` in particular comes along for parity rather than for the bug; splitting it out would leave the two paths diverging again, which is the thing being fixed.
- **`packageManagerInstallArguments` is exported from `components/Application.ts`**, a ~4000-line module, and `utility/` now imports it for pure argument construction. The `nonInteractiveSpawn` import already crosses that boundary so there is no new cycle, but a reviewer who wants this in a small shared module would be reasonable — it is a mechanical move.
- **The DESIGN.md claim is deliberately narrower than "every npm install".** npm's own nested install inside `npm pack <git-ref>` is not composed here and cannot be; binding that would mean `npm_config_audit=false` in the spawn environment instead of argv. Not done here — the argv builder is what the diverging path broke, and it is what the tests can pin.
- **The argv pin is skipped on Windows** (the shim is a `/bin/sh` script) while `windowsGate.mjs` runs the rest of the file. The flag set is platform-independent, so a regression fails the three Linux legs before Windows sees it; a `.cmd` shim was judged more likely to produce a Windows-only flake than to catch anything.
- **Declined: tightening this call's spawn timeout.** A review leg proposed a shorter `timeoutMs` for `installModules` so a regression fails fast everywhere. `DEFAULT_COMMAND_TIMEOUT_MS` is shared with the component install path, where hour-long installs are legitimate; lowering it for one caller trades a known hang for an unknown truncation. The mocha-level bound covers the test path, which is what was wedging.
- Not addressed: `.mocharc.json`'s repo-wide `timeout: 0` still means a hung unit test cannot fail on its own. Changing that is its own blast radius; the step caps and this suite's own bound are the backstop until then.

## Verification

- Reproduced the hang on Node 22.23.1 / npm 10.9.8: `npm run test:unit:main` stalls right after `honors the undocumented camelCase dryRun spelling`, with a live `npm install` child sitting in `ep_poll` on an established socket to the registry — the same signature as the CI job's orphan-process list (`Terminate orphan process: pid (5992) (npm install)`).
- Isolated it away from Harper: the same `npm install --force --omit=dev --json` over the same two-package fixture hangs under npm 10.9.8 and completes in 0.506s under npm 11.17.0; the raw `curl` measurements above confirm which request is stalling.
- Fails-on-base for the new guard: removing `--no-audit --no-fund` fails the argv assertion.
- `npx mocha unitTests/utility/npmUtilities.test.js` on Node 22: 8 passing in ~1s, real installs at ~200ms each. `unitTests/components/applicationInstall.test.js` also passes, covering the renamed builder's other two call sites.
- `npm run test:unit:main` on Node 22: 5411 passing, 196 pending, 2 failing — both failures are local-environment artefacts, not regressions. `configValidator.test.js:394` resolves `'relative/root'` against `process.cwd()`, so it warns from a deep worktree path and passes on CI's short runner cwd; `tokenAuthentication.test.js:247` is the known shared-fixture flake and passes on its own (58 passing). The same two pass on CI's v24 leg, whose 5412 total matches this run's 5413 (5411 + 2) once the new case is counted.
- `npm run test:unit:resources` on Node 22: 2042 passing, 0 failing.
- End-to-end route, executed: [run 33874421548](https://github.com/HarperFast/harper/actions/runs/33874421548) on this PR — all four Unit Test legs report a real `success`: Windows/v24 3m01s, v24 8m13s, v26 8m11s, **v22 10m21s**, its `install_node_modules` real installs down to 165ms from the hang. Note the v22 number: with the stall gone the leg still takes 10m21s, so the old 10-minute cap was already too small on its own — the `Run tests` step used 9m32s of its 15-minute cap. Every other check on the PR is green (41 pass, 4 skipping).

Complexity: medium

<sub>Review-Coverage: authored=claude; ran=codex,gemini; declined=cursor-grok,cursor-composer,domain; rounds=5 @ ceebeaf810a5</sub>

<sub>Human-Review-Need: 3 @ ceebeaf810a5</sub>

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -27,11 +27,11 @@ jobs:
   unit-test:
     name: Unit Test (Node.js v${{ matrix.node-version }})
     runs-on: ubuntu-latest
-    # A job that hits this cap is reported as `cancelled`, not `failure`, so it reads as non-red
-    # on the required-check status — the long steps below carry their own caps for that reason.
-    # This one only has to exceed their sum: dependency install and the unit run measure ~2.5 and
-    # ~7.5 minutes, and the remaining steps under a minute.
-    timeout-minutes: 20
+    # GitHub reports a job that hits this cap as `cancelled`, not `failure`, so it reads as
+    # non-red on the required check — which is how a wedged v22 leg went unnoticed for 9 runs.
+    # Every step below is capped in its own right and this sits above their sum (33 on the
+    # longest leg), so a stall always surfaces as a step failure and this is only a backstop.
+    timeout-minutes: 35
     strategy:
       fail-fast: false
       matrix:
@@ -48,24 +48,29 @@ jobs:
           package-manager-cache: false
 
       - name: Install dependencies
-        timeout-minutes: 10 # measures ~2.5 minutes; bounded so a registry stall fails loudly
+        timeout-minutes: 6
         run: npm install --ignore-scripts
 
       - name: Build
+        timeout-minutes: 3
         run: npm run build || true # we currently have type errors so just ignore that
 
       - name: Type contract tests
         if: matrix.node-version == 22
+        timeout-minutes: 3
         run: npm run test:types # strict tsc over unitTests/types/ — asserts the shipped public types
 
       - name: Benchmark unit tests
+        timeout-minutes: 3
         run: npm run test:bench # pure-logic tests for benchmarks/ converters; doesn't need a running Harper instance
 
       - name: Review-coverage action tests
         if: matrix.node-version == 24
+        timeout-minutes: 3
         run: node --test .github/actions/review-coverage/*.test.mjs
 
       - name: Setup Harper
+        timeout-minutes: 3
         env:
           DEFAULTS_MODE: 'dev'
           HDB_ADMIN_USERNAME: 'admin'
@@ -76,10 +81,8 @@ jobs:
 
         run: node --enable-source-maps ./dist/bin/harper.js install
 
-      # Bounded below the job so a hung suite fails the step rather than cancelling the job.
-      # `.mocharc.json` sets `timeout: 0`, so no individual test can ever time out on its own:
-      # anything that blocks — a registry call, a lock wait — stops the run until something above
-      # it gives up, and before this cap that something was the job timeout, which is silent.
+      # `.mocharc.json` sets `timeout: 0`, so nothing inside the run can time out on its own —
+      # this cap is the only thing that turns a wedged suite into a reported failure.
       - name: Run tests
         timeout-minutes: 15
         run: npm run test:unit:all
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -27,11 +27,10 @@ jobs:
   unit-test:
     name: Unit Test (Node.js v${{ matrix.node-version }})
     runs-on: ubuntu-latest
-    # GitHub reports a job that hits this cap as `cancelled`, not `failure`, so it reads as
-    # non-red on the required check — which is how a wedged v22 leg went unnoticed for 9 runs.
-    # Every step below is capped in its own right and this sits above their sum (33 on the
-    # longest leg), so a stall always surfaces as a step failure and this is only a backstop.
-    timeout-minutes: 35
+    # A job that hits this cap is reported as `cancelled`, not `failure`, so it reads as non-red
+    # on a required check. Every step below is capped in its own right and this sits above their
+    # sum (41 on the longest leg), so a stall surfaces as a step failure and this stays a backstop.
+    timeout-minutes: 45
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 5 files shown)


dawsontoth (approved):
Very nice, great approach.

Ethan-Arrowood (approved):

cb1kenobi (approved):

Issue #2520, opened 2026-09-07
Secondary index opened by a schema-change rescan serves partial results during another worker's backfill


kriszyp:
## Summary

When a worker learns of a new secondary index only through the schema-change rescan (`initStores` in `resources/databases.ts`, run by `resetDatabases` on a `schema-change` ITC message), it opens the index with no `isIndexing` flag derived from the catalog's `indexingPID`. `search.ts` gates on `index.isIndexing` alone, so that worker serves the index while the declaring worker's backfill is still running: queries return a subset of matching rows with no 503.

## Reproduction (expected)

1. Populated table; multi-worker node (the default).
2. `create_attribute` (or an `@table` re-declaration) adding an attribute with `indexed: true`.
3. While the backfill runs, query on a **different** worker filtering on that attribute.

Expected: 503 / not-ready (as the declaring worker does). Observed: partial result set, silently.

## Notes

- Pre-existing on `main`; surfaced while reviewing HarperFast/harper#2264, whose branch reload takes the same `initStores` path and inherits the behaviour unchanged (not a regression there).
- Silent wrong-results for the duration of the backfill, bounded by backfill time (long on large tables).
- Fix direction: when `initStores` opens an index whose catalog entry carries an `indexingPID` for a live process, mark it `isIndexing` and clear it on `indexing-finished`, mirroring what the declaring thread does.

🤖 Filed by Claude on behalf of Kris

PR #2523, opened 2026-09-07
feat(branches): declare tables into a branch through @table, ensureTable and defineTable

### LINKED ISSUE #2264: Branched databases: scope the table factory through GraphQL @table, ensureTable, and defineTable
## Summary

Thread a scope-owned table factory through GraphQL schema processing and the schema-authoring
helpers, so that an application declaring `branchedDatabases` gets its `@table` declarations bound to
its branch rather than to the base database.

Subtask of HarperFast/harper#642. Split out from HarperFast/harper#643 because it is the largest
piece of the binding work and is independently reviewable.

## Why this is required, not optional

#642's premise is that application code is unchanged: the app imports `databases` from `harper` and
the binding resolves to its branch. Scoping the module export alone is not enough, because
**declaring tables in `schema.graphql` with `@table` is how most Harper applications define their
schema**, and that path resolves through the process-global table factory:

- `resources/graphql.ts:317` — `typeDef.tableClass = table(typeDef)` calls the global factory.
- `resources/graphql.ts:326` — the resulting class is registered into the resource registry, which is
  what backs exported HTTP routes.

So a branched application would evolve schema and serve exported routes against the **base**, while
its direct `databases.data` writes went to the **branch**. That is a split brain, and it is worse
than having no isolation at all — the failure is silent and partial.

The same global factory is reached by two smaller paths that belong in this change:

- `components/Scope.ts:242` — `scope.ensureTable()` calls `table(options)` directly.
- `resources/defineTable.ts:402` — `defineTable()` closes over the same factory.

## Scope

- A scope-owned table factory that resolves into the branch's private `Tables` graph (created by
  HarperFast/harper#643) instead of the process-global `databases` map.
- Thread it through GraphQL schema processing, `scope.ensureTable()`, and `defineTable()`.
- Unbranched scopes must keep the **exact existing object identities** — this path is on the request
  path and must not regress for the ~all applications that do not use branching.

## Explicitly out of scope

Deferred by decision (Kris, 2026-08-21) when #642 took option (b), the narrow-surface route:

- **`operation(...)` / `server.operation(...)` and SQL.** These resolve databases by name through the
  operations layer, so a branched app calling them writes the base. This cannot be closed by
  validation either, because the call is dynamic. Documented limitation.
- **Bare globals.** Not closable at all under the default loader: `ApplicationScope` defaults `mode`
  to `vm-current-context` (`components/ApplicationScope.ts:38`), whose branch in `importScoped` is
  annotated "shares intrinsics, no custom globals" (`security/jsLoader.ts:99-101`); the per-scope
  global object built by `getGlobalObject()` (`security/jsLoader.ts:825-847`) is only constructed for
  the compartment and custom-context VM paths. Bare `databases` is therefore the process-wide
  singleton by construction.

The resulting contract for #642 is precise: **imports from `harper` and GraphQL-declared tables
resolve to the branch; bare globals and the operations/SQL layer resolve to the base.**

## Acceptance criteria

- [ ] A branched app declaring `@table` in `schema.graphql` gets tables in its branch; the base
      database is unchanged.
- [ ] Its exported HTTP routes read and write the branch.
- [ ] `scope.ensureTable()` and `defineTable()` from a branched scope resolve to the branch.
- [ ] An unbranched scope receives the identical objects it does today (assert reference identity,
      not just equivalence).
- [ ] A second, unbranched app loaded alongside sees the untouched base.

## Depends on

HarperFast/harper#643 — the scope-private `Tables` graph this factory writes into.

🤖 Filed by Claude on behalf of Kris

kriszyp:
## Summary

Rebased this draft onto current `main` while retaining branch-scoped declarations: [`scopedTableFactory`](https://github.com/HarperFast/harper/pull/2523/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R2303) routes `@table`, `scope.ensureTable()`, and `defineTable()` into an application's private branch without changing the unbranched factory identity. The integration proof exercises the exported route, scoped import, and base isolation for a GraphQL declaration in [the branch-schema integration suite](https://github.com/HarperFast/harper/pull/2523/changes?w=1#diff-778708c583b59a13e8920c579934136d64857f1d0fd6e264b2a1c5cefd64a15fR64).

Closes HarperFast/harper#2264. Part of HarperFast/harper#642.

## For the human reviewer

1. The change reuses the global declaration engine with a branch-bound target rather than creating a second schema implementation. The highest-risk boundary is [`reloadBranch`](https://github.com/HarperFast/harper/pull/2523/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R2327): schema changes must remain private while another worker refreshes its branch catalog. Reverting this choice restores the former refusal of branch declarations; a branch-specific engine would be broader and duplicate lifecycle behavior.
2. Cross-model review confirmed the long-standing incomplete-index readiness gap applies to branch reloads too (HarperFast/harper#2520), but this rebase intentionally preserves the existing narrow issue scope rather than adding a cross-worker indexing redesign. A reviewer may require that follow-up before merging if branch schema evolution with live index backfills must be supported now; otherwise the contained declaration behavior remains independently verified.

## Verification

- `npm run build`, `npm run test:unit:main`, `npm run test:unit:resources`, and `npm run test:integration:all` passed before the rebase push.
- `integrationTests/components/branched-database-schema.test.ts` passed; its route, scoped-import, and base-isolation assertions are linked above.
- Required GitHub Actions checks for the rebased head are passing.
- Full independent pre-push review and Harper-domain adjudication completed for `a33222d47248`; no code was changed after that review.

Complexity: complicated

🤖 Rebase and review coverage refreshed by GPT-5 Codex.

<sub>Review-Coverage: authored=codex; ran=gemini,cursor-grok; adjudicated=domain; blocked=claude(exit-1); declined=cursor-composer; rounds=1 @ a33222d47248</sub>

<sub>Human-Review-Need: 4 (decisions: shared-declaration-engine, unbranched-name-fallback, declaration-only-evolution) @ a33222d47248</sub>


kriszyp:
CI note: the Unit Test failures on this PR are main's, not this change's. Main's own Unit Test run at its current head (38fcbf1ca) fails with the same MQTT `subscribe to retained record with patch operations` assertion (`'update 3' == 'update 2'`, the flake previously addressed in #1347 and #1697), and the second failure seen here (`lookup scans every entry sharing the transaction log key`) is in `unitTests/resources/dualClockAuditRecord.test.js`, added by #2497, which is also red on main from 6d5a4632b. Neither test touches the table factory or branch code in this PR; the branch, deploy and both branched-database integration suites pass here.

— Claude Fable 5.1

Issue #2527, opened 2026-09-08
Request.withNodeAdapter() cannot serve real Node middleware: Next.js 500s, then >16 KB responses hang silently


kriszyp:
`Request.withNodeAdapter()` (added in #408, `server/serverHelpers/Request.ts`) is meant to let a Harper component hand a request to an arbitrary Node HTTP handler. It has **no in-tree consumers**, and the first real one — the `@harperfast/nextjs` plugin, which needs it so a Next.js app mounted at a `urlPath` receives the mount-stripped URL — cannot use it: the fake `IncomingMessage`/`ServerResponse` it constructs is not faithful enough for real Node middleware. Every page and route 500s, and once the obvious members are patched in, responses larger than 16 KB hang forever with no error.

Confirmed on `main` (`6d725818c`), `v5.2.9` and `v5.1.27` — none of them have `appendHeader` or `_implicitHeader`, and all build request headers with `Object.create(null)`.

## Symptoms, in the order you hit them

1. **`reqHeaders` is `Object.create(null)`** (`Request.ts:195`); Node's real `IncomingMessage.headers` carries `Object.prototype`. Next's compiled `RouteModule.prepare` calls `headers.hasOwnProperty(...)` → `TypeError: e.hasOwnProperty is not a function`. Every page and route 500s.
2. **No `appendHeader()`** (Node 18.3+). Next's app-page runtime calls it → `TypeError: q.appendHeader is not a function`.
3. **No `_implicitHeader()`**. Next's bundled `compression` calls it on every write → `TypeError: this._implicitHeader is not a function`.
4. **Neither `finished` nor `complete`.** `on-finished` treats that as ALREADY finished (`if (isFinished(msg) !== false) defer(listener)`), so the `send` library — which Next uses for `/_next/static/*` — runs its cleanup immediately and destroys the file read stream. **Every static asset request hangs forever with no error.**
[...]

PR #2528, opened 2026-09-08
Rebuild withNodeAdapter's response on a real Writable so Node middleware (Next.js, compression, send) runs against it

kriszyp:
`Request.withNodeAdapter()` handed middleware an `EventEmitter` with hand-forwarded `write`/`end`/`drain`, but real Node middleware treats a `ServerResponse` as a Writable and relies on `_header`, `_implicitHeader()` and `writeHead()` ordering. Next.js therefore returned 500s, and larger responses hung once the obvious missing members were patched. The response is now [`NodeAdapterResponse extends PassThrough`](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-7643d1df591c46bbe514d936b4feaffdd6cbbc18ed7c8290013aa6c77693d493R33), and that exact object is the body the adapter resolves with, so backpressure and stream lifecycle events are Node's own.

Headers [commit once through `this.writeHead`](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-7643d1df591c46bbe514d936b4feaffdd6cbbc18ed7c8290013aa6c77693d493R154); [disconnects and handler failures](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-aae8e044b25206008a8c85d3a78499c40a5d5ee0af8bc397b51a101deed394e7R218) close or error the stream instead of leaving it open; and [`Headers.delete`](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-e133943ed3811097d82aa42045886aa3a6f7488f382323b67cf51a04a8f7caabR34) is case-insensitive, so compression can remove `send`'s stale `Content-Length`. Post-review maintenance also makes [`setHeaders()` accept Harper `Headers`](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-7643d1df591c46bbe514d936b4feaffdd6cbbc18ed7c8290013aa6c77693d493R112), preserves an own `__proto__` request header without changing the Node-compatible plain-object prototype, and [forwards explicit and server-default timeout events with their socket argument](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-7643d1df591c46bbe514d936b4feaffdd6cbbc18ed7c8290013aa6c77693d493R63). The public `Promise<{ status, headers, body: PassThrough }>` shape is unchanged.

Implementation and evidence are in [`NodeAdapterResponse.ts`](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-7643d1df591c46bbe514d936b4feaffdd6cbbc18ed7c8290013aa6c77693d493R33), [`Request.ts`](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-aae8e044b25206008a8c85d3a78499c40a5d5ee0af8bc397b51a101deed394e7R193), [`Headers.ts`](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-e133943ed3811097d82aa42045886aa3a6f7488f382323b67cf51a04a8f7caabR34), the [`Request` contract tests](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-cc2af353e8437dee3c0ead778cb8dbf2168546038ab6bb40c13d0d803409c482R920), [`Headers` regression test](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-71679eac86e2b9c222efb09d9c033e36142ca7a12e56be0b6393aa950151a04fR61), and [real-middleware/wire tests](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-98d7a739fd8e353990d28e6a5578db54debd652386231f9f9070877c2f3ad1bfR29). The non-obvious contract is recorded in [`DESIGN.md`](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-3dc5dd454e080eb849ee5efaf79df2585fbe0a06804d69249b4c81da05a63875R2374); the real middleware fixtures are documented in [`dependencies.md`](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-1bdc56ae5bf9cf9448297bcf2651338189804cf56ecd8fa92f8ba8f985cb029cR158) and added through [`package.json`](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R154) plus its [`package-lock.json`](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519R120).

## For the human reviewer

1. **Express is an explicit non-goal.** Express replaces the response prototype with one rooted at `http.ServerResponse.prototype`, which a Writable-derived response cannot survive. Next.js, `compression`, `send`, `on-finished`, `on-headers`, `finalhandler`, `serve-static`, h3 and fastify duck-type the response without replacing its prototype. Supporting Express requires a separate design.
2. **Client hangup semantics are deliberate.** After headers, a disconnect [destroys the body without an error](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-98d7a739fd8e353990d28e6a5578db54debd652386231f9f9070877c2f3ad1bfR441); before headers, [the response promise rejects](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-98d7a739fd8e353990d28e6a5578db54debd652386231f9f9070877c2f3ad1bfR429). This matches Node's dropped-connection behavior and lets `pipeBodyToResponse` treat the premature close as routine. One consequence to weigh: the rejection carries an `AbortError`, which has no `statusCode`, so a consumer that routes it into `server/http.ts`'s `onError` gets a 500 written and `harperLogger.error` for a routine hangup. Giving the rejection a classified status is the one-line alternative.
3. **The adapter owns an `error` listener.** `res.writeHead(200); res.destroy(err)` emits before an awaiting caller can attach one. The error remains in the stream's state for `pipeline()`, `finished()` or async iteration, and the [immediate-destroy regression](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-98d7a739fd8e353990d28e6a5578db54debd652386231f9f9070877c2f3ad1bfR418) proves the process does not crash.
4. **A second `writeHead()` throws `ERR_HTTP_HEADERS_SENT`.** This now matches Node, as do late `setHeader`, `appendHeader` and `removeHeader` calls; the [contract test pins the change](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-cc2af353e8437dee3c0ead778cb8dbf2168546038ab6bb40c13d0d803409c482R580).
5. **A handler failure after `res.end()` is logged at warn, sync or async.** It cannot reach the client or the already-resolved promise, so [both paths use the same operator signal](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-aae8e044b25206008a8c85d3a78499c40a5d5ee0af8bc397b51a101deed394e7R226).
6. **Timeout forwarding follows listener ownership.** The adapter attaches to the real response only while middleware has a timeout listener, so it does not accidentally mark an unhandled timeout as handled. Both [`setTimeout()` forwarding](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-cc2af353e8437dee3c0ead778cb8dbf2168546038ab6bb40c13d0d803409c482R947) and [server-default forwarding without `setTimeout()`](https://github.com/HarperFast/harper/pull/2528/changes?w=1#diff-cc2af353e8437dee3c0ead778cb8dbf2168546038ab6bb40c13d0d803409c482R976) preserve the socket argument and detach on close.
7. **An already-aborted request still invokes the handler against a destroyed response.** Node can invoke a handler after its client is gone; writes then fail through normal stream semantics. Returning early is the one-line alternative if this compatibility choice changes.
8. **Informational responses use the real Node response.** `writeContinue`, `writeProcessing` and `writeEarlyHints` delegate when the transport provides a response and call back without hanging in the response-less fixture path.
9. **The added packages are test-only real middleware.** Next.js vendors compression 1.7.4, which gates on `_header`/`_implicitHeader()`, while 1.8 gates on `headersSent`/`writeHead()`. The matrix runs both; `send` was already a runtime dependency.
10. **The in-repo wire proof is intentionally below a mounted Harper route.** It uses a real `http.createServer`, Node request/response objects, `compression` + `send`, and Harper's `pipeBodyToResponse`. The out-of-tree `@harperfast/nextjs` CI matrix is the mounted consumer this unblocks.
11. **`strictContentLength` is declared but not enforced.** `implements ServerResponse` requires the property, and the adapter's default `false` matches Node exactly (verified on Node v26: a 10-byte `Content-Length` ended with 5 bytes throws `ERR_HTTP_CONTENT_LENGTH_MISMATCH` only when the flag is set to `true`). A middleware that opts in by setting it `true` would get no enforcement here. No supported middleware sets it; enforcing it means counting bytes per `write()`.

Two earlier suggestions remain declined: adding `statusMessage` to the resolved public shape, and routing `write()`/`end()` through `flushHeaders()` instead of `_implicitHeader()`.

## Verification

- **Fails on base:** building `origin/main`'s adapter and running the new middleware suite produced 16 failures / 3 passes: compression 1.7 lacked `_implicitHeader`, `send`/`on-finished` paths timed out, header-ordering assertions failed, and an immediate post-`writeHead` destroy escaped as an uncaught error.
- **Current focused gates:** `npm ci`, `npm run build`, `npm run test:types`, `npm run lint:required`, `npm run format:check`, and `git diff --check` pass. `npx mocha unitTests/server/serverHelpers/Request.test.js unitTests/server/serverHelpers/Headers.test.js unitTests/server/serverHelpers/nodeAdapterMiddleware.test.js` reports 126 passing.
- **Windows gate:** `npm run test:unit:windows` passes all 9 groups / 3,762 tests on the final head.
- **Broader local gates (on the merge):** `npm run test:unit:main` reports 5,582 passing / 197 pending with two failures, both environmental and both confirmed by re-running them alone: the known long-worktree domain-socket path case in `configValidator`, and `stuckWorkerDiagnostics` (a `main` test from `6d725818c`, untouched here) which counts two diagnostic samples and loses the race when the box is loaded — it passes on its own, 5 passing.
- **Review:** the original design gate concluded `chosen-approach-sound`. Four further full cross-model passes ran on this head lineage (claude + gemini + Harper-domain adjudication each time), converging at round 11 with verdict LGTM and adjudicated severity minor. They produced two real fixes, both included here: a duplicate `node_modules/@harperfast/rocksdb-js-linux-x64-musl` key in `package-lock.json` (this branch and `main` had each restored the entry, and the merge kept both, leaving which metadata wins parser-dependent), and a functional bug where `writeHead(302, undefined, { Location })` dropped the third argument's headers because the `else` branch assigned over them unconditionally — Node uses `obj ??= reason`, verified against a real `http.createServer`, and the regression test fails on the previous behaviour. Three reviewer claims were rejected on executed evidence rather than argument: `setHeaders` taking `Headers | Map` **is** Node's contract (Node v26 rejects a plain object with `ERR_INVALID_ARG_TYPE`); `req.headers` having `Object.prototype` matches Node, which also lets a `Constructor:` header shadow it; and `strictContentLength` is enforced by Node only when set to `true`, so the adapter's `false` default matches.
- **CI (head `6bb377753`): 45 checks pass, none fail.** The branch was brought current with `main` by a merge commit rather than a rebase. That cleared four checks that had been red at the previous head and were all reproducing failures from the merge-base `3a3c6f1ff`, each already fixed on `main`: the Docker `smoke` job's shrinkwrap pin comparison (fixed by [#2560](https://github.com/HarperFast/harper/pull/2560); the root pins here are identical to `main`'s, and the same job failed on `main` at the merge-base and two later commits), and the `eviction-secondary-index` 500 plus the `txnlog-restart-reclaim` double-cleanup-pass case on shards 2 and 6 (both fixed by `bf0d17d78`). The Next.js adapter matrix passes on Node 20/22/24, every integration shard passes on Node 24, Bun, uWS and Windows, and the Windows unit gate passes.

Refs #2527

— Claude Fable 5.1

https://claude.ai/code/session_01BL9FvWh4XHTxvPgsdak2Ni

Complexity: complicated

<sub>Review-Coverage: authored=codex; ran=gemini,claude; adjudicated=domain; declined=cursor-grok,cursor-composer; rounds=11; full=4 @ 6bb377753204</sub>

<sub>Human-Review-Need: 4 (decisions: response-is-body, disconnect-settlement, timeout-relay, post-end-handler-failure, middleware-boundary, verification-boundary) @ 6bb377753204</sub>


unitTests/server/serverHelpers/Request.test.js
```diff
@@ -859,6 +1034,20 @@ describe('Request class', function () {
...
+				const responsePromise = request.withNodeAdapter(async (req, res) => {
+					res.write('partial');
+					throw new Error('handler failed mid-stream');
+				});
+
+				const { body } = await responsePromise;
+				const chunks = [];
+				await assert.rejects(async () => {
+					for await (const chunk of body) chunks.push(chunk);
+				}, /handler failed mid-stream/);
```
dawsontoth, line 1073: Should we have a chunk after we find the rejection? Or is that behavior not specified? Wondering if the test can be expanded to capture that too (that we get a `partial` chunk, if relevant)

dawsontoth (approved):
I read through it, and things roughly make sense to me. I imagine this is something for fixing up the Next.js component, or people building more middleware on top of Harper? I had one question about a unit test completeness, but nothing necessary. The failing unit tests look unrelated to the changes made here, probably tackled in a different PR already.

dawsontoth:
(Rebasing or merging the latest from main into this branch may tackle the merge conflict and the failing MQTT tests, though I haven't checked that to make sure.)

server/serverHelpers/NodeAdapterResponse.ts
```diff
@@ -0,0 +1,225 @@
...
+	setHeaders(headers: Headers | Map<string, number | string | readonly string[]>) {
+		let cookies: string[] | undefined;
+		for (const [name, value] of headers) {
+			if (name.toLowerCase() === 'set-cookie')
+				(cookies ??= []).push(...(Array.isArray(value) ? value : [String(value)]));
+			else this.setHeader(name, value);
+		}
+		if (cookies) this.setHeader('set-cookie', cookies);
+		return this;
+	}
```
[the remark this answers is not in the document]
kriszyp: Addressed in 3f1a36339. `NodeAdapterResponse.setHeaders()` now accepts Harper `ResponseHeaders`, and the runtime test covers that input alongside Map cookie grouping. The exact TypeScript call produced TS2345 before the fix and compiles after it.

unitTests/server/serverHelpers/Request.test.js
kriszyp: Conclusive answer: the test should not assert delivery of the buffered `partial` chunk. On Node 26, `PassThrough.write("partial")` followed by `destroy(error)` leaves `readableLength === 7`, but async iteration rejects before yielding any bytes. The adapter guarantees the terminal error after a post-header handler failure, not delivery of buffered data after destruction, so the rejection-only assertion is intentional.

PR #2531, opened 2026-09-08
Bound dormant staged builds per component with deployment.stagingRetention.maxCount

dawsontoth:
Step 4 of #2315. Adds [`deployment.stagingRetention.maxCount`](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-85897aa30843afee9a67fc30b41713bbc7f2e3a3e1e9fa1a0437ecac46a698f9R598) (default `5`, `0` keeps none, [coerced like its sibling](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R1039)) and bounds, per component, how many **dormant** staged builds survive under `<componentsRoot>/.deploy-staging` — a [dormant build](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R1066) being a journal-less deployment directory that holds `.complete` and the owner's tree: built and validated, activated by nobody.

The premise this step was planned on moved during step 1's review: `main` discards a rejected or failed candidate immediately, and boot recovery removed *every* owned, journal-less staging directory. So nothing intentionally leaves a dormant build behind today — only the crash window between `.complete` and the journal write does. Step 6 (deploy from an existing aside) is the producer this bound exists for; this PR lands the bound and the knob so that step has a contract to build on, and makes recovery stop destroying the state step 6 will rely on.

What changes on disk: boot recovery [keeps dormant builds instead of removing them](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R1813), and [bounds them per owner](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R1936) to the newest `maxCount` by `.complete` mtime (ties broken by deployment id, so concurrent worker passes evict the same build). Residue — a partial tree, a tree already moved live, a directory carrying a stale `.unsettled` — is still removed exactly as before. The deploy path applies the same bound at the start of a deploy, [inside the settlement scan](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R1582) `prepareApplication` already runs under the component lock, so a deploy pays one traversal of the staging root rather than two. [`drop_component` reclaims the dropped component's dormant builds](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R2849), since no later deploy of that name will.

## For the human reviewer

1. **Dormant is `.complete` + tree + no journal, with removal decided under the owner's lock — not a new marker.** The planning review proposed an explicit "staged/dormant" marker written by a stage-only operation, arguing `.complete` proves durability but not dormancy because activation writes it moments before the journal. That is true of an *unlocked* read, which is why an unlocked read is only a candidate here and nothing is removed without re-deriving it under the owner's preparation lock, where a complete, journal-less directory cannot be mid-activation: activation holds that lock from `.complete` through the swap. A marker would duplicate `.complete` for the one producer step 6 adds. Reversible at zero migration cost: no dormant builds exist in the wild yet, so step 6 can introduce a marker if its stage-only mode wants one. Cost of a "no" here: step 6 writes and retention reads one more control file.
2. **Dormant builds are catalogued without the lock; only an over-quota owner takes it, once.** Round 1 classified each retained build under the owner's lock on every recovery pass — a lock per directory, forever, since a retained build is never removed — and the domain review showed a healthy component losing the 250 ms probe to its sibling threads at boot and being deferred with nothing in progress. Now an unlocked read is a candidate, the lock is taken once per owner over its bound, and each owner is then [reconciled once](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R1936): if a journal appeared on any catalogued directory since the scan, or the owner is over its bound, the lock is taken and only its catalogued directories — never the whole staging root, which sibling threads are probing — are re-read under it. A journal that appeared is [settled there](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R1982) — kriszyp's finding: a deploy that published one and died mid-swap would otherwise leave the component unloadable until the next start, since workers accept a well-formed journal while the legacy pass refuses to restore against it — and what is still dormant is [re-derived before the kept set is chosen](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R1086). The residue branch [re-classifies under its lock](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R1782) too, his second finding: a `.complete` written while the scan waited for the lock is a retainable build, not residue. A lock a live deploy holds on such an owner is still [recorded as the deferral](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R1999) `componentLoader` already acts on, and no `.unsettled` is written: "do not delete" is not "safe to load". The alternative — skip and log on timeout — is the more available reading; the conservative one was kept because it only arises while an owner is over quota, which the same pass then clears.
3. **Default is 5, not 0.** `0` would preserve today's on-disk behaviour exactly but make step 6's default useless and force a later default change. `5` is the reviewed #1849 default. The only behaviour visible today: a `.complete` build whose journal was never written (crash in that window) is retained up to the bound at boot instead of removed. Cost of a "no": one line and the docs.
4. **The deploy path prunes before building, not after.** It rides the settlement scan that already runs under the lock before the candidate is built, so the guarantee is "at most N dormant builds when a deploy of X starts", not "when it returns" — today the deploy's own candidate never survives, so the two are the same; step 6's stage-only mode decides whether to prune again after staging.
5. **Count, not size; `.complete` mtime as the order; no pin for a build a deploy is about to read.** The domain review's ledger raised all three. Count is what the issue specifies and matches #1849; disk is the constraint that bites, so a `maxSize` sibling may be worth adding once step 6 produces real trees — cheap to add beside this knob, awkward to replace it. `.complete` mtime does not survive a timestamp-normalising restore; a monotonic id or a recorded stamp can replace it before any retained build is something an operator rolls back to. Pruning before the build has nothing to pin today because no deploy reads a retained build; step 6 must pin its source or prune after.
6. **Unowned residue is deliberately out of scope.** A crash while resolving or packing leaves an empty `.deploy-staging/<id>`; nothing names it, so it cannot be told from a build that has not reached extraction yet. The fix — writing the sidecar at the start of the build — would make every respawning worker defer that component for the whole `npm install`, a step-1 decision this step does not reopen. Documented in [DESIGN.md](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-3dc5dd454e080eb849ee5efaf79df2585fbe0a06804d69249b4c81da05a63875R657) rather than fixed.

## Verification

- **Unit** — `unitTests/components/deployStagingRetention.test.js` (18 tests): a complete build and a `file:` symlink build survive boot while a partial tree and a swept-but-not-removed directory are removed; per-component bound to the newest N with a neighbour's builds not counted; completion-time tie broken by id; `maxCount: 0` keeps none; a stale `.unsettled` on a journal-less directory is removed so workers stop refusing it; [a journaled activation is settled rather than retained](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-d07055ed56251c81bb86a1506708541c9747138c0623b03a7a4f0bc966af9039R194) however dormant it looks; [a component within its bound takes no lock](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-d07055ed56251c81bb86a1506708541c9747138c0623b03a7a4f0bc966af9039R219), so a held lock does not defer it; [an over-quota owner whose lock is held is deferred](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-d07055ed56251c81bb86a1506708541c9747138c0623b03a7a4f0bc966af9039R240), with nothing deleted and no verdict written; [`dropComponentDirectory` reclaims the dropped component's dormant builds](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-d07055ed56251c81bb86a1506708541c9747138c0623b03a7a4f0bc966af9039R413) and leaves a neighbour's alone; [the eviction re-check keeps a build that acquired a journal since the scan](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-d07055ed56251c81bb86a1506708541c9747138c0623b03a7a4f0bc966af9039R373); [a journal published after the scan on a catalogued build is settled and rolls forward](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-d07055ed56251c81bb86a1506708541c9747138c0623b03a7a4f0bc966af9039R264), or defers the component while a live deploy holds the lock; two held-lock interleavings, each waiting on the scan's lock ticket rather than a timer — [a journal published plus live moved aside between the scan and the end of the pass](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-d07055ed56251c81bb86a1506708541c9747138c0623b03a7a4f0bc966af9039R314), and [a `.complete` landing while the scan waits for the lock](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-d07055ed56251c81bb86a1506708541c9747138c0623b03a7a4f0bc966af9039R346) — recover and retain respectively; an unremovable build is logged, not thrown; the deploy path bounds its own component, leaves its neighbours alone, and still surfaces the deploy's own error; getter coercion (unset/blank/boolean/array/NaN/negative → 5; numeric string, float floor, explicit 0). The 43 existing `deployActivation` tests still pass unchanged.
- **Mutation-checked**: removing the dormant branch fails 3 tests, removing the journal re-check fails exactly the re-check test, removing the bound fails 5 tests, removing the under-lock residue re-classification fails exactly the `.complete`-landed test, and making reconciliation ignore late journals fails exactly the two late-journal tests.
- **End-to-end, route (b)** — [new `integrationTests/deploy/staging-retention.test.ts`](https://github.com/HarperFast/harper/pull/2531/changes?w=1#diff-bc5dd1fd687e402423c9a294e548eefe01d09f5f5264d79f6cecca23ea1153d1R54) starts Harper with `deployment.stagingRetention.maxCount: 1` through the harness's config override, plants two dormant builds beside the running instance, deploys the component once through the operations API, and asserts the deploy landed and only the newest planted build survives. Passes locally. The whole `integrationTests/deploy/` folder also passes locally.
- **Suites**: `unitTests/components/*.test.js` run as one suite locally: 829 pass; the 7 failures (`applicationSpawn` process-group timing, `RuntimeModuleTracker` macOS `/private/var` realpath) reproduce identically on a clean `origin/main` worktree and touch nothing in this diff. `tsc --noEmit`, `oxlint`, `prettier --check` clean on the touched files; the full unit and integration matrix is CI's.
- **Docs**: companion PR HarperFast/documentation#668 adds a `deployment` section to the configuration reference (this knob plus the previously undocumented `payloadRetention.maxSize`) and a 5.3 release note.

Complexity: medium

<sub>Review-Coverage: authored=claude; ran=codex,gemini; adjudicated=domain; declined=cursor-grok,cursor-composer; rounds=6 @ c941cb60854e</sub>

<sub>Human-Review-Need: 3 (decisions: lock-tickets-as-test-oracle) @ c941cb60854e</sub>

### DIFF
--- a/components/Application.ts
+++ b/components/Application.ts
@@ -1709,30 +1709,17 @@ export async function recoverInterruptedActivations(componentsRootDirPath: strin
 		throw error;
 	}
 	const dormant = new Map<string, DormantBuild[]>();
+	const catalogue = (owner: string, build: DormantBuild) => {
+		const builds = dormant.get(owner);
+		if (builds) builds.push(build);
+		else dormant.set(owner, [build]);
+	};
 
 	for (const deployment of deployments) {
 		if (!deployment.isDirectory()) continue;
 		const deploymentDirPath = join(stagingRoot, deployment.name);
 		const journalPath = join(deploymentDirPath, ACTIVATION_JOURNAL);
-		const fail = async (component: string, error: unknown) => {
-			const failure = error instanceof Error ? error : new Error(String(error));
-			if (!failures.has(component)) failures.set(component, failure);
-			logger.error(`Could not settle the interrupted activation of ${component}:`, errorForLog(failure));
-			// A DEFERRAL is not a verdict, and only verdicts go on disk. A held lock means a live deploy, which
-			// settles its own journal; a marker written here would outlive that deploy and have
-			// `unsettleableComponentsFromDisk` read it as an authoritative "cannot be settled", failing a
-			// healthy component closed on every worker. The failure is already recorded above, so this thread
-			// still defers — it just leaves nothing behind.
-			if (failure instanceof ComponentPreparationLockTimeoutError) return;
-			// Everything else IS a verdict, recorded so workers reach the same one. An unreadable journal is
-			// self-evident, but a well-formed journal this pass could not settle looks exactly like a deploy
-			// in flight, and a worker would otherwise load the component over state nobody reconciled.
-			// Best-effort: the alternative to a missing marker is today's behavior, not a worse one.
-			await writeFile(join(deploymentDirPath, UNSETTLED_MARKER), failure.message, { mode: 0o600 }).catch(
-				(markerError) =>
-					logger.warn(`Could not record the unsettled activation of ${component}: ${errorMessage(markerError)}`)
-			);
-		};
+		const fail = (component: string, error: unknown) => recordUnsettled(failures, component, error, deploymentDirPath);
 
 		let journal: ActivationJournal | undefined;
 		try {
@@ -1792,6 +1779,13 @@ export async function recoverInterruptedActivations(componentsRootDirPath: strin
 					activationToFail = undefined;
 					return;
 				}
+				// Re-classified UNDER the lock: the unlocked read that routed this here can predate the `.complete`
+				// a deploy wrote before dying, and that is a retainable build, not residue.
+				const build = await dormantBuildAt(deploymentDirPath, owner!);
+				if (build) {
+					catalogue(owner!, build);
+					return;
+				}
 				// Cleanup, not settlement. There was no activation here — this is most often the residue a
 				// SUCCESSFUL settlement leaves when its own sweep failed — so a sweep that fails again cannot
 				// make anything unsettled, and recording it would refuse a live component on every worker
@@ -1830,9 +1824,7 @@ export async function recoverInterruptedActivations(componentsRootDirPath: strin
 					continue;
 				}
 				if (build) {
-					const builds = dormant.get(owner);
-					if (builds) builds.push(build);
-					else dormant.set(owner, [build]);
+					catalogue(owner, build);
 					continue;
 				}
 			}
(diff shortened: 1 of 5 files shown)


unitTests/components/deployStagingRetention.test.js
```diff
@@ -0,0 +1,310 @@
...
+const {
+	withComponentPreparationLock,
+	ComponentPreparationLockTimeoutError,
+} = require('#src/components/componentPreparationLock');
+const env = require('#src/utility/environment/environmentManager');
+const { CONFIG_PARAMS } = require('#src/utility/hdbTerms');
+
+async function newRoot(label) {
+	return fs.mkdtemp(path.join(os.tmpdir(), `staging-retention-${label}-`));
+}
```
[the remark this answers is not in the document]
dawsontoth: Keeping the existing pattern: each root is a fresh directory under os.tmpdir() and the neighbouring recovery suite (deployActivation.test.js) removes it the same way at the end of the test. A failing assertion leaves one uniquely named tmp directory behind, which the OS reclaims; wrapping 15 tests in try/finally trades that for noise in every test body.

components/Application.ts
```diff
@@ -1689,6 +1816,26 @@ export async function recoverInterruptedActivations(componentsRootDirPath: strin
...
+						`Leaving deploy staging ${deploymentDirPath} in place; it could not be read:`,
+						errorForLog(error)
+					);
+					continue;
+				}
+				if (build) {
+					const builds = dormant.get(owner);
+					if (builds) builds.push(build);
+					else dormant.set(owner, [build]);
+					continue;
```
kriszyp, line 1828: Please preserve activation recovery when taking this shortcut. A concrete interleaving is: the initial journal read returns ENOENT; the deploying worker writes `.complete`, publishes its journal, moves the live tree aside, and dies before moving the candidate live. `dormantBuildAt()` still returns this candidate because it does not check the journal. This `continue` skips settlement; within the default retention limit there is no subsequent lock/check, and above the limit the pruner merely excludes journaled entries without settling them.

The later read-only verdict accepts a well-formed journal, while legacy extraction recovery refuses to restore against it (`components/Application.ts:2591-2598`). The component therefore fails to load despite having everything needed to recover. Previously, `removeResidue` re-read and settled the journal under the lock.

Keep pruning optional, but preserve a per-owner recovery/deferral check that settles journals published after the initial scan. Add a regression covering journal publication and an interrupted swap between the initial journal read and dormant classification.

components/Application.ts
```diff
@@ -1689,6 +1816,26 @@ export async function recoverInterruptedActivations(componentsRootDirPath: strin
 				);
 				continue;
 			}
+			// Catalogued WITHOUT the lock and left alone: a retained build is never removed here, so a per-directory
+			// lock would recur on every pass and contend with sibling threads for a component nothing is deploying.
+			if (owner) {
+				let build: DormantBuild | undefined;
+				try {
+					build = await dormantBuildAt(deploymentDirPath, owner);
```
kriszyp, line 1818: Please also re-check dormancy inside `removeResidue`, after acquiring the preparation lock and before deleting the directory. This unlocked check can see no `.complete` and route a build to residue cleanup. While recovery waits, the deploying worker can finish writing `.complete` and die before publishing the journal—the exact crash window this PR intends to retain. Once recovery acquires the abandoned lock, `removeResidue` checks only for a journal and unconditionally removes the now-complete build (`components/Application.ts:1799`). That loses a retainable build even when the component is below its limit.

Reclassify the entry under the lock and feed a newly complete build into retention instead of deleting it as residue. A regression should publish `.complete` between the unlocked classification and lock acquisition, leaving the journal absent.

kriszyp (approved):
Looks good, I think this will be nice. One finding to consider.

components/Application.ts
dawsontoth: Fixed in c941cb608 (with 569569532). After the unlocked scan, each owner's catalogued builds are reconciled: if any acquired a journal since the scan (or the owner is over its bound), the owner lock is taken once and only those directories are re-read under it; a journal that appeared is settled via `settleInterruptedActivation` with the same split-attribution and unreadable-journal handling as the scan, a held lock defers the component, and a non-timeout lock failure with a journal in view records `.unsettled` like the scan does. Regressions: a direct `reconcileDormantBuilds` test rolling a planted interrupted swap forward, and a whole-pass test that holds the lock, waits for the scan's lock ticket, then publishes the journal and moves live aside before releasing. Kept the unlocked probe so a healthy within-bound owner still takes no lock; that is recorded as a decision in the PR body.

components/Application.ts
dawsontoth: Fixed in 569569532: `removeResidue` re-classifies under the lock after the journal re-read, and a build whose `.complete` landed meanwhile is catalogued for retention instead of removed. Regression holds the lock, waits for the scan's lock ticket, writes `.complete`, releases, and asserts the build survives with no deferral; removing the re-classification fails exactly that test.

PR #2532, opened 2026-09-08
Expose accurate MCP operation schemas across workers

### LINKED ISSUE #878: MCP: refactor operations tool registration so input schemas are introspectable/validatable JSON Schema

kylebernhardy:
Makes operation input schemas [first-class registry metadata](https://github.com/HarperFast/harper/pull/2532/changes?w=1#diff-71babc9060a8fb11d1ac1fc9feb2b3aa60f32f66fa6a3bb39d3cbcb8f4ec2fe8R230) so MCP advertises accurate JSON Schema contracts for built-in and component-registered operations, including schemas [mirrored from HTTP workers](https://github.com/HarperFast/harper/pull/2532/changes?w=1#diff-3ed03c13dc77b511d4689671bb8d375166f177b10f42e4fb7cebddeabbf7193fR181). Schema-less operations are withheld by default, with an explicit compatibility setting for deployments that need the prior permissive contract.

Closes #878.

## For the human reviewer

1. This makes schema-backed tools the v5.3 default and requires existing explicitly allow-listed schema-less operations to also appear in [`mcp.operations.allowSchemaless`](https://github.com/HarperFast/harper/pull/2532/changes?w=1#diff-c70d594c1d7b8ae86e94cb7100742aec23003565a09c8201f321893d30da7c41R437). Keeping the permissive fallback globally would avoid an upgrade change but continue advertising contracts that provide clients no argument guidance. The choice is reversible through configuration; rejecting it means changing the rollout policy or adding schemas for every operation before this can ship.

## Verification

- `npm run build` — passed.
- `npx mocha unitTests/server/serverHelpers/serverUtilities.test.js` — 105 passing.
- `npx mocha unitTests/components/mcp/tools/operations.test.js unitTests/agent/mcpTools.test.js` — 39 passing.
- `npx mocha unitTests/validation/configValidator.test.js --grep "mcp config"` — 5 passing.
- End-to-end route (new integration coverage): [the registered-operation integration test](https://github.com/HarperFast/harper/pull/2532/changes?w=1#diff-98cdf02e424516cac1913a7a8106ec44f1b1ac0424bd43ee8c76fad6ab927ad4R109) exercises a real two-worker registration through MCP `tools/list`, including the schema-less negative case. Both local invocation modes were blocked before Harper startup because this host lacks the required `127.0.0.2+` loopback aliases; CI must supply the executed end-to-end result.
- `npm run test:unit:main` progressed through the affected MCP, server, config, and agent coverage without a failure, then stopped producing output in a later no-timeout suite and was interrupted.
- Independent Claude review: LGTM with no open findings at `282586b3f770` after six review rounds. Gemini was unavailable because its local credentials are expired.

Complexity: complicated

<sub>Review-Coverage: authored=codex; ran=claude; declined=gemini,cursor-grok,cursor-composer,domain; rounds=6 @ 282586b3f770</sub>

<sub>Human-Review-Need: 4 @ 282586b3f770</sub>

### DIFF
--- a/components/mcp/tools/operations.ts
+++ b/components/mcp/tools/operations.ts
@@ -45,7 +45,6 @@ import {
 	type ToolDef,
 	type ToolResult,
 } from '../toolRegistry.ts';
-import { OPERATION_INPUT_SCHEMAS, PERMISSIVE_SCHEMA } from './schemas/operations.ts';
 import { OPERATION_DESCRIPTIONS } from './schemas/operationDescriptions.ts';
 
 // Resolved from Harper's server-helpers graph on demand. The map is built at
@@ -53,7 +52,8 @@ import { OPERATION_DESCRIPTIONS } from './schemas/operationDescriptions.ts';
 // `startOnMainThread` — the provider below re-reads it per request rather than
 // snapshotting (#1562).
 type OperationFunction = (json: object) => unknown | Promise<unknown>;
-type OperationFunctionMap = Map<string, { operation_function: OperationFunction }>;
+type OperationFunctionEntry = { operation_function: OperationFunction; inputSchema?: object };
+type OperationFunctionMap = Map<string, OperationFunctionEntry>;
 
 type ChooseOperation = (body: object) => OperationFunction;
 type ProcessLocalTransaction = (req: { body: object }, fn: OperationFunction) => Promise<unknown>;
@@ -68,9 +68,11 @@ interface OperationsConfig {
 let _opMapOverride: OperationFunctionMap | undefined;
 let _chooseOperationOverride: ChooseOperation | undefined;
 let _processLocalTransactionOverride: ProcessLocalTransaction | undefined;
+const warnedMissingSchemas = new Set<string>();
 
 export function _setOperationFunctionMapForTest(m: OperationFunctionMap | undefined): void {
 	_opMapOverride = m;
+	warnedMissingSchemas.clear();
 }
 export function _setChooseOperationForTest(fn: ChooseOperation | undefined): void {
 	_chooseOperationOverride = fn;
@@ -386,23 +388,39 @@ export function makeOperationToolHandler(operationName: string) {
  * predicate, and handler don't depend on the allow/deny config, which only
  * decides *whether* the op is exposed, checked per request in the provider).
  * `tools/list` isn't a hot path, so the provider rebuilds defs per call rather
- * than caching (no module-level state to leak or stale-cache across tests).
+ * than caching them. The only module state deduplicates missing-schema warnings
+ * and is cleared with the operation-map test seam.
  */
-function buildOperationToolDef(operationName: string): ToolDef {
-	const inputSchema = OPERATION_INPUT_SCHEMAS[operationName] ?? PERMISSIVE_SCHEMA;
+function buildOperationToolDef(operationName: string, inputSchema: object): ToolDef {
 	const annotations: { readOnlyHint?: boolean; destructiveHint?: boolean; idempotentHint?: boolean } = {};
 	if (isReadOnly(operationName)) annotations.readOnlyHint = true;
 	if (isDestructive(operationName)) annotations.destructiveHint = true;
 	if (isIdempotent(operationName)) annotations.idempotentHint = true;
 	return {
 		name: operationName,
-		description: buildDescription(operationName, operationName in OPERATION_INPUT_SCHEMAS),
+		description: buildDescription(operationName, true),
 		inputSchema,
 		profile: 'operations',
 		...(Object.keys(annotations).length > 0 ? { annotations } : {}),
 		visibleTo: (user) => canRoleInvokeOperation(user, operationName),
 		handler: makeOperationToolHandler(operationName),
 	};
+}
+
+function buildRegisteredOperationToolDef(
+	operationName: string,
+	operation: OperationFunctionEntry
+): ToolDef | undefined {
+	if (!operation.inputSchema) {
+		if (!warnedMissingSchemas.has(operationName)) {
+			warnedMissingSchemas.add(operationName);
+			harperLogger.warn(
+				`MCP operations profile: '${operationName}' is allowed but has no inputSchema; register one to expose this tool`
+			);
+		}
+		return undefined;
+	}
+	return buildOperationToolDef(operationName, operation.inputSchema);
 }
 
 /**
@@ -419,17 +437,20 @@ const operationsToolProvider: ProfileToolProvider = {
 		}
 		const config = getOperationsConfig();
 		const defs: ToolDef[] = [];
-		for (const operationName of opMap.keys()) {
+		for (const [operationName, operation] of opMap) {
 			if (!isOperationAllowed(operationName, config)) continue;
-			defs.push(buildOperationToolDef(operationName));
+			const def = buildRegisteredOperationToolDef(operationName, operation);
+			if (def) defs.push(def);
 		}
 		return defs;
 	},
 	get(operationName: string): ToolDef | undefined {
 		const opMap = getOperationFunctionMap();
-		if (!opMap || !opMap.has(operationName)) return undefined;
+		if (!opMap) return undefined;
+		const operation = opMap.get(operationName);
+		if (!operation) return undefined;
 		if (!isOperationAllowed(operationName, getOperationsConfig())) return undefined;
-		return buildOperationToolDef(operationName);
+		return buildRegisteredOperationToolDef(operationName, operation);
 	},
 };
 
--- a/components/mcp/tools/operations.ts
+++ b/components/mcp/tools/operations.ts
@@ -385,9 +385,7 @@ export function makeOperationToolHandler(operationName: string) {
  * function of the operation name (its schema, description, annotations, RBAC
  * predicate, and handler don't depend on the allow/deny config, which only
  * decides *whether* the op is exposed, checked per request in the provider).
- * `tools/list` isn't a hot path, so the provider rebuilds defs per call rather
- * than caching them. The only module state deduplicates missing-schema warnings
- * and is cleared with the operation-map test seam.
+ * `tools/list` isn't a hot path, so the provider rebuilds defs per call.
  */
 function buildOperationToolDef(operationName: string, inputSchema: object): ToolDef {
 	const annotations: { readOnlyHint?: boolean; destructiveHint?: boolean; idempotentHint?: boolean } = {};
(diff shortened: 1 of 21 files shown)


kylebernhardy:
@kriszyp CI is green and the automated review found no issues. While this remains draft, could you take a look at the v5.3 rollout choice around withholding schema-less operations by default versus a broader compatibility period?\n\nComment generated by kAIle (GPT-5)

components/mcp/tools/operations.ts
```diff
@@ -405,6 +427,33 @@ function buildOperationToolDef(operationName: string): ToolDef {
 	};
 }
 
+function buildRegisteredOperationToolDef(
+	operationName: string,
+	operation: OperationFunctionEntry,
+	config: OperationsConfig,
+	issue: RemoteOperationSchema['issue'] = 'missing'
+): ToolDef | undefined {
+	if (!operation.inputSchema) {
```
kriszyp: This is the behavior change I'd most like to see reconsidered, because it silently breaks a working configuration on upgrade.

Concrete scenario: an operator running today with `mcp.operations.allow: ['insert', 'update', 'search_by_conditions']`. `insert` and `update` have no entry in `OPERATION_INPUT_SCHEMAS` (the catalog covers 26 names; the built-in map has ~106), so before this PR they were advertised with `PERMISSIVE_SCHEMA` and the LLM could call them, with the server validating and returning a structured error. After upgrading, `buildRegisteredOperationToolDef` returns `undefined` for both: they vanish from `tools/list`, `tools/call` reports an unknown tool, and the only signal is one `harperLogger.warn` per name per process. The operator's agent quietly loses the ability to write data until someone reads the log and edits YAML.

The goal — don't advertise a contract we can't describe — is right for the *default* surface, where a glob like `list_*` can sweep in an op nobody vetted. But an operator who typed the operation name into `allow` has already made that judgment explicitly; that's the same opt-in `allowSchemaless` now asks them to make a second time, in a second list.

Suggestion that keeps the guarantee and removes the new config key: treat a *literal* (non-glob) entry in `config.allow` as the schemaless opt-in, and withhold only when the op reached the surface via a glob or the default allow list. `matchesAny` already distinguishes them cheaply (`!pattern.includes('*') && pattern === operationName`). That preserves every existing explicit opt-in, still refuses to invent a contract for glob-swept ops, and leaves `mcp.operations` with two lists instead of three.

If you'd rather keep `allowSchemaless`, then the upgrade path deserves more than a per-name warn — at minimum a single startup/`tools/list` warning naming all withheld-but-allowed ops together, since the current warn is emitted inside a `tools/list` walk that an operator may never look at.

The `issue !== 'inconsistent'` carve-out on line 437 is the right call and reads clearly — no concern there.

server/serverHelpers/serverUtilities.ts
```diff
@@ -219,12 +227,21 @@ server.registerOperation = (operationDefinition: OperationDefinition) => {
 		opAuth.registerOperationPermission(name, { requiresSu: requiresSuperUser });
 		declaredPermissionNames.add(name);
 	}
-	OPERATION_FUNCTION_MAP.set(name as any, new OperationFunctionObject(handler));
+	const schemaMetadata =
+		inputSchema ??
+		(!BUILT_IN_OPERATION_NAMES.has(name as any) && Object.hasOwn(OPERATION_INPUT_SCHEMAS, name)
```
kriszyp: This name-keyed inheritance is doing real work — it's what keeps `get_metrics` / `list_deployments` (registered from outside this repo, since neither is in `initializeOperationFunctionMap`) advertised without touching the registering package — but as written it's an implicit shim with two rough edges:

1. **Asymmetry.** A component that re-registers a *built-in* correctly gets no inherited schema (the `BUILT_IN_OPERATION_NAMES` guard, with a test at `unitTests/server/serverHelpers/serverUtilities.test.js:1133`). But a component that registers `get_metrics` with a completely different payload silently inherits Harper's schema and MCP then advertises the wrong argument shape to the LLM. The safety argument for the built-in case applies identically here; the only difference is which map the name happens to live in.
2. **Coverage.** The shim only covers the 26 catalog names. Any other operation registered from another package that matches the default `describe_*` / `list_*` / `search_*` globs still drops off `tools/list` after this change, with only a warn — so the wrapper package needs auditing regardless. Has harper-pro's operation registration been checked against this?

Given (2) means the registering package has to be touched anyway for full coverage, the cleaner shape is to have it pass `inputSchema` explicitly at its `server.registerOperation()` call and drop the name-based inheritance entirely — one source of truth per operation, and no way for a same-named component op to pick up someone else's contract. If the inheritance has to stay for the release, a comment here explaining which registrations depend on it would help the next reader, since nothing in core exercises that branch outside tests.

package.json
```diff
@@ -191,6 +191,7 @@
 		"@turf/distance": "6.5.0",
 		"@turf/helpers": "6.5.0",
 		"@turf/length": "6.5.0",
+		"ajv": "8.18.0",
```
kriszyp: Pinning `ajv` to an exact `8.18.0` at the root changes resolutions beyond this feature, in the opposite direction from what `dependencies.md` describes.

The lockfile diff shows `node_modules/fastify/node_modules/ajv` (8.20.0) and `node_modules/@modelcontextprotocol/sdk/node_modules/ajv` (8.20.0) both being deleted, with everything now de-duping onto the new root 8.18.0. Fastify 5.12.1 pulls ajv through `@fastify/ajv-compiler ^4.0.5` (`ajv: ^8.12.0`), so 8.18.0 satisfies it and npm hoists — meaning the compiler that validates operations-API request schemas is now two minor versions *older* than what shipped before this PR. The `dependencies.md` entry claims the opposite ("declaring Fastify's existing 8.18.0 resolution directly ... without changing its schema compiler version"); 8.18.0 was `@fastify/ajv-compiler`'s and `ajv-formats`' resolution, not Fastify's own.

Suggest declaring `"ajv": "8.20.0"` so the tree consolidates upward onto what Fastify already shipped, and correcting that sentence in `dependencies.md` — the honest version is "consolidates the four existing ajv 8.x copies onto one, at the highest version already present."

Minor side effect worth a line in the same entry: eslint's nested ajv moves 6.14.0 → 6.15.0 as a consequence of the re-resolution (dev-only, no runtime impact).

server/serverHelpers/validateOperationInputSchema.ts
```diff
@@ -0,0 +1,64 @@
...
+		const schema = JSON.parse(serialized);
+		if (!schema || typeof schema !== 'object' || Array.isArray(schema)) {
+			return { error: 'inputSchema must be a JSON object' };
+		}
+		if (schema.type !== 'object') return { error: "inputSchema must declare type: 'object'" };
+		const dialect = getDialect(schema.$schema);
+		if ('error' in dialect) return dialect;
+		const validator = getValidator(dialect);
+		const schemaToValidate = dialect.schemaId ? { ...schema, $schema: dialect.schemaId } : schema;
+		if (!validator.validateSchema(schemaToValidate)) {
```
kriszyp: `validateSchema` only checks the schema against its meta-schema; it does not resolve `$ref`. So a component registering `{ type: 'object', properties: { x: { $ref: '#/definitions/Missing' } } }` — or a `$ref` to an external `https://` schema we'll never fetch — passes here, is stored, mirrored to main, and advertised over `tools/list`. An MCP client that compiles the advertised schema (most SDKs do) then errors on the whole tool list, not just that one tool, which is exactly the "malformed component schema breaking `tools/list`" failure this module exists to prevent.

Note the new built-in lint in `unitTests/components/mcp/tools/operations.test.js:453` holds shipped schemas to `ajv.compile`, the stricter bar — so the code we control is checked more thoroughly than the code we don't.

Since this runs once per registration and never on dispatch, `compile()` inside the existing try/catch would close the gap at no meaningful cost and give component authors the same guarantee. If you want to keep `validateSchema` for a reason I'm not seeing (compile caching a schema in the ajv instance? `addUsedSchema: false` already prevents that), a short comment on the deliberate difference would help.

server/serverHelpers/operationInputSchemas.ts
```diff
@@ -1,34 +1,18 @@
 /**
- * Hand-curated JSON Schemas for the operations-profile MCP tools.
+ * JSON Schemas attached to built-in operation registrations.
```
kriszyp: Design/root-cause question rather than a defect, since this is the piece the whole PR rests on.

Moving this catalog out of `components/mcp/` and attaching it to live `OPERATION_FUNCTION_MAP` entries promotes it from "MCP presentation metadata" to "the operation's declared input contract" — `server.registerOperation`'s new `inputSchema` and this map are now the same field. But the thing that actually rejects a request is still the Joi/legacy validator in `validation/`, and nothing ties the two together. When they drift, an MCP client is told a call is well-formed and the server rejects it — the failure mode is a confidently wrong LLM, which is worse than the pre-PR permissive posture where the client knew it was guessing.

The header says the projection can't be done mechanically "currently", which reads like the real root cause is acknowledged. Two things would make the interim state safer:

- A drift test for the ops that do have both a Joi validator and a catalog entry — even a one-directional check (every `required` name in the JSON Schema is a key the Joi schema knows about) would catch the common rename.
- A note here on the intended end state, so the next person adding an operation knows whether they're maintaining a permanent hand-written catalog or a temporary bridge.

Relatedly: `OPERATION_DESCRIPTIONS` stayed in `components/mcp/tools/schemas/`, so the two halves of the same per-operation catalog now live in different layers with two separate coverage-lint tests. Worth either moving descriptions alongside this file or documenting why they belong on opposite sides of the boundary.

components/mcp/tools/operations.ts
```diff
@@ -268,7 +292,7 @@ function matchesAny(operation: string, patterns: readonly string[] | undefined):
 	return false;
 }
 
-function isOperationAllowed(operation: string, config: OperationsConfig): boolean {
+export function isOperationAllowed(operation: string, config: OperationsConfig): boolean {
```
kriszyp: Nit: this is exported only so `unitTests/server/serverHelpers/serverUtilities.test.js:35` can assert schema coverage over the live op map. Everything else in this file that exists for tests uses the `_…ForTest` prefix, which makes the intent obvious and keeps the module's real surface small. Either rename it to follow that convention, or keep the export and let the coverage lint live next to the other catalog lints in `unitTests/components/mcp/tools/operations.test.js`, where the module is already imported.

PR #2533, opened 2026-09-09
Implement the shared derived-index transaction-log runtime

### LINKED ISSUE #2489: Implement the shared transaction-log runtime for derived indexes
## Goal

Provide one Harper-owned delivery and recovery protocol for non-transactional derived indexes such as full-text and a future asynchronous HNSW plane. Primary records remain authoritative. A derived index is either proven current through its durable cursor or explicitly unavailable pending rebuild.

This issue covers the shared Harper runtime. Tantivy storage, schema declarations, query APIs, blob extraction, generation activation, and migration of the current synchronous HNSW index are separate work.

## Architecture

Each registered backend has an independent runner, transaction-log iterator, cursor vector, and backpressure state.

- Any Harper worker may write records and append to the shared physical transaction logs.
- `RocksDatabase.tryLock()` elects one temporary runner owner per backend id across workers. There is no permanent worker-0 writer.
- A root-store `committed` event is only a wake-up. The elected runner reads committed transaction-log entries; the write path never waits for derived-index work.
- Different indexes use different lock keys, so their runners and native writers can make progress in parallel. One deferred or failed backend does not block another.
- The owner is sticky while work or non-durable accepted progress remains, then releases after a short idle grace period.
- Schema activation installs the same registration on every worker before that worker can accept table work or evict. Registration is worker-local; only runner ownership is elected.

This deliberately does not use the formerly proposed same-thread `aftercommit` delivery. That path retains audit objects on every write, makes every worker a producer into backend state, and does not provide one serial replay path after worker loss.

## Cursor and replay contract

The backend durably owns this versioned cursor:

```ts
type DerivedIndexCursor = {
	format: 1;
	logs: Record<string, number>; // physical log name -> completed transaction timestamp
};
```

The current rocksdb-js transaction-log iterator exposes `{ timestamp, data, endTxn }`, not physical file offsets. Harper therefore uses its existing transaction timestamp identity rather than inventing or requiring a new rocksdb-js cursor primitive.

On startup, owner handoff, or accepted-work loss, the aggregate reader exact-seeks every saved log boundary. It validates and consumes exactly one complete anchor transaction from each same iterator before returning subsequent physical entries. This closes a validation/resume race and still returns a later physical entry whose timestamp is lower than the anchor. Missing, incomplete, or duplicate boundaries require rebuild.

`oldestSequenceNumber` is used for diagnostics and to prove that a newly discovered log still retains its beginning. It cannot be compared with a timestamp cursor. A missing exact timestamp can require a full-log scan during rare owner reconstruction; sticky ownership keeps that cost off the steady-state path.

Harper retains every complete cursor vector offered to a backend until its durability barrier advances. A reported durable cursor must equal one whole offered vector; per-log positions from different batch boundaries cannot be combined. The backend must publish the cursor atomically with the index state covered by that barrier.

Accepted-but-not-durable progress is bounded. At the configured cap, the runner stops reading and waits for a backend state-change notification rather than repeatedly probing the backend on unrelated database commits.

## Delivery contract

```ts
const DERIVED_INDEX_ACCEPTED = 1;
const DERIVED_INDEX_DEFERRED = 0;
const DERIVED_INDEX_FAILED = -1;

interface DerivedIndexBackend {
	readonly id: string;
	getDurableCursor(): DerivedIndexCursor | undefined;
	deliver(batch: DerivedIndexBatch): 1 | 0 | -1;
	onStateChange(
		wake: (change?: 'changed' | 'accepted-work-lost' | 'failed') => void
	): () => void;
}
```

- `deliver()` is synchronous and non-blocking. It may enqueue into native work but may not wait for a writer mutex, merge, or durability barrier.
- `accepted` means the backend owns the batch. It does not mean the cursor is durable.
- `deferred` retains the exact batch and stops only that runner until a backend wake.
- `accepted-work-lost` discards process-local offered progress and reconstructs from the durable cursor.
- `failed` makes that backend unavailable pending rebuild.

Drain limits apply between complete source transactions. One backend call receives each bounded batch, including cursor-only progress through unrelated transactions.

## Authoritative state and projection

The log identifies records that may need revisiting. Before delivery, Harper reads current committed state directly from the primary store, never through `Resource.get()` and never through a caching source. Repeated references to one `(tableId, recordId)` in a batch share one primary read and one schema-compiled projection.

The backend receives only declared derived-index attributes. A present local record produces its current version and projection; a durable removal fact produces absence. This latest-state protocol is idempotent by primary key and safe under replay overlap.

`put`, `patch`, `delete`, `invalidate`, `relocate`, and local-only `evict` are eligible record actions. Whole-table `reload`, decode failures, transaction framing failures, primary read failures, and projection failures are fail-closed.

## Cache eviction

Cache eviction is local residency state, not a canonical delete. For a table with a derived-index registration, direct and batched RocksDB eviction stage a bodyless, `LOCAL_ONLY` eviction marker in the same native transaction as the fresh-version-guarded removal. An aborted or conflicting eviction commits neither the removal nor the marker.

The derived runtime consumes this marker and resolves the authoritative row as absent. Boot replay treats it as a successful control entry without replaying or counting a record. Live subscriptions ignore it, subscription replay and customer history filter it, and the `LOCAL_ONLY` bit prevents replication. Raw transaction-log readers retain it for the derived runtime; the customer `read_audit_log` and `read_transaction_log` operations route through the filtered Table history methods.

Registration counts are scoped by audit-store identity and table id. Tables without a derived-index registration pay only an O(1) guard during eviction and add no transaction-log writes. LMDB derived indexes are not part of this release.

## Failure behavior

The backend transitions to `needs-rebuild` without cursor advancement when Harper cannot prove continuity or authoritative state, including:

- missing saved log or exact boundary;
- newly discovered log whose first sequence was already purged;
- incomplete or duplicate transaction boundary;
- corrupt or undecodable log entry;
- unexpected per-log iterator failure;
- removed physical log;
- malformed, regressed, or unoffered durable cursor vector;
- primary-read or projection failure; or
- permanent backend failure.

Transaction-log retention is not pinned in this phase. A backend that falls behind retained history rebuilds rather than silently skipping work.

## Current implementation unit

The feature branch now contains:

- opt-in physical log names and exact-resume failure metadata on `RocksTransactionLogStore.getRange()`;
- the lock-elected `DerivedIndexRuntime` with bounded batches, backpressure, durable/offered cursor reconciliation, owner epochs, and isolated failure states;
- reference-counted table registration used only by the RocksDB eviction paths;
- atomic local-only markers for direct and batched cache eviction, including customer-history and boot-replay isolation;
- focused fake-backend coverage; and
- an end-to-end audited RocksDB table test for put, patch, delete, eviction, current-state projection, and commit wake-up.

Database/schema lifecycle registration, activation/rebuild generation management, metrics, and a real backend remain follow-on units.

## Acceptance

- No derived-index work is awaited from a primary record commit.
- At most one runner owns one backend id across Harper workers; independent indexes can run in parallel.
- Restart and worker handoff resume from an exact durable boundary with no silent cursor leap.
- Deferred and accepted-but-not-durable work is retried without blocking other indexes.
- Every relevant committed mutation converges to current primary state, including replicated/source-version writes and local eviction.
- Any unprovable gap makes only the affected backend unavailable pending rebuild.
- Focused resource tests, cross-worker ownership/crash tests, replay/reopen tests, and write-path performance benchmarks pass before promotion.

Detailed design: `docs/derived-index-runtime-stage-1.md` on the implementation branch.

kylebernhardy:
## Outcome

Adds Harper’s shared post-commit delivery and exact-replay foundation for non-transactional derived indexes. Each backend has an independent lock-elected runner, cursor vector, bounded backlog, and failure state; record commits never await derived-index work.

## What changed

- exposes physical log identity and exact-resume failure metadata from transaction-log ranges
- adds the `DerivedIndexRuntime` delivery, cursor, ownership, and backpressure contract
- resolves projections from authoritative committed records rather than staged audit bodies
- emits atomic, local-only cache-eviction markers only for registered derived-index tables
- keeps internal eviction markers out of customer history, subscriptions, replication, and boot replay
- documents the per-worker registration invariant and the remaining generation/schema work

Closes #2489

## Validation

- `npm run build`
- `npm run lint:required`
- 75 focused derived-index/eviction tests passed
- 38 adjacent subscription/reload/audit tests passed; 9 pending
- full resource gate reaches the existing native abort in `Audit cleanup retirement`, reproduced on unmodified `main`
- local Claude and Gemini review completed; no accepted findings remain

### DIFF
--- a/docs/derived-index-runtime-stage-1.md
+++ b/docs/derived-index-runtime-stage-1.md
@@ -123,6 +123,9 @@ Stage 1 adds an opt-in `includeLogName` value to `RocksTransactionLogStore.getRa
 reuses its existing aggregate iterator, per-log corrupt-frame tracking, new-log discovery, and
 timestamp merge. `AuditRecord.logName` is initialized to `undefined` on both decoded and sentinel
 entries so existing hot consumers keep one stable object shape; it is assigned only when requested.
+The returned iterable also exposes `failedLogs`, naming any physical iterator that ended on an
+unexpected non-corruption error; a derived-index runner treats either that signal or the existing
+corrupt-frame signal as an availability failure and never advances its cursor through it.
 A runner starts the aggregate with its backend's `startByLog` vector and preserves the physical log
 name on every result. Entries from one physical log are assembled through `endTxn`; a drain budget
 is checked only between complete transactions, never in the middle of one.
--- a/docs/derived-index-runtime-stage-1.md
+++ b/docs/derived-index-runtime-stage-1.md
@@ -30,9 +30,9 @@ API, blob extraction, generation activation, or HNSW migration.
 - Conflict retries can re-resolve a record after its transaction-log payload was first staged. The
   audit entry therefore provides durable mutation identity and version evidence, but its embedded
   record is not authoritative for derived-index projection.
-- `Table.evict()` removes the primary row and its transactional custom indexes without writing an
-  audit entry. A derived index cannot mirror cache eviction until Harper adds a local-only eviction
-  marker to the same RocksDB transaction, limited to tables with derived-index registrations.
+- `Table.evict()` ordinarily removes the primary row and its transactional custom indexes without
+  writing an audit entry. A registered derived index adds a local-only eviction marker to that same
+  RocksDB transaction so asynchronous indexing can mirror the removal.
 - Transaction-log retention removes whole files. `TransactionLog.getStats()` exposes the oldest
   retained file sequence, while `exactStart` proves whether a saved transaction boundary remains
   readable.
@@ -231,14 +231,14 @@ in the existing physical transaction log before removing the primary row. It use
 transaction as the version-guarded removal and is `LOCAL_ONLY`: cache residency is local and must not
 replicate. If the transaction conflicts or aborts, neither removal nor marker commits.
 
-The internal action is a no-op in boot replay and does not increment replayed-record counts. Audit
-and transaction-log read APIs filter it from customer output, and replication continues to reject it
-through `LOCAL_ONLY`. The runtime consumes it before those presentation filters.
+The internal action is a no-op in boot replay and does not increment replayed-record counts.
+Customer-facing history and subscription replay filter it, while raw transaction-log readers retain
+it for the derived runtime. Replication continues to reject it through `LOCAL_ONLY`.
 
 This is the only Stage 1 producer change. It is necessary because HNSW and ordinary secondary
 indexes are currently removed synchronously by `updateIndices`, while an asynchronous derived
 index otherwise has no durable evidence that the cached row disappeared. Tables without a derived
-backend keep the current eviction path and cost.
+backend pay only an O(1) registration guard and retain the existing storage writes.
 
 ### Cursor validation and retention gaps
 
@@ -468,8 +468,8 @@ the cursor and batch contracts do not prevent adding cohorts after comparative b
   presenting a clean end-of-log.
 - Prove direct, expires-at sweep, and `createEvictionBatcher` eviction while delivery is deferred all
   remove the document, while a failed eviction transaction emits neither removal nor marker.
-- Prove internal eviction entries are hidden from audit/log read APIs, ignored and uncounted by boot
-  replay, and never sent to a peer.
+- Prove internal eviction entries are hidden from customer history and subscription APIs, ignored
+  and uncounted by boot replay, and never sent to a peer.
 - Prove an out-of-band reload marker triggers rebuild rather than cursor advancement.
 - Reopen after an unclean shutdown and prove the installed rocksdb-js exposes recovered durable
   entries to the ordinary committed reader; never substitute `readUncommitted` in the runtime.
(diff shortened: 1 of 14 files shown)


docs/derived-index-runtime-stage-1.md
```diff
@@ -0,0 +1,520 @@
+# Derived-index runtime: committed log delivery and exact replay foundation
```
kriszyp, line 1: Putting design docs for a PR into the repository goes against our coding conventions. Design information that needs to be persisted for future agents should go in DESIGN.md files (with tight budgeting, more words is extremely costly and degrading for future agent's context), or for the temporary purpose of PR design, it should go in the PR description (never lost, all PRs histories are retained). Most of this is the latter.

kriszyp:
@kylebernhardy I am trying to optimize this runtime to be safe, efficient, and maximize simplicity with https://github.com/HarperFast/harper/pull/2548. However, https://github.com/HarperFast/harper/pull/2430 exists as the only consumer of the runtime to verify its functionality. Without any Tantivy consumer, the simplification work in 2548 is likely to make changes to this PR that will significantly impact (or break) Tantivy use of this. It seems like you would probably want to have a real consumer drive the contract of this PR rather than just self-assertion of the PR itself?

kriszyp:
Superseded by #2548, which carries this branch's commits and retargets the whole derived-index stack at `main` for a single review and merge. Closing so the stack no longer blocks that retarget. — Claude Fable 5.1

PR #2535, opened 2026-09-09
Add Harper RocksDB storage for derived indexes

kylebernhardy:
## Summary

- add an internal binary RocksDB store for native derived indexes without adding or changing rocksdb-js primitives
- apply each host-storage mutation batch in one WAL-enabled RocksDB transaction and fail closed unless the transaction reports a definite commit
- use Harper's existing database-wide RocksDB flush as the durability barrier, with the cost called out for coalescing and measurement
- reserve derived column-family names so opaque Tantivy bytes cannot alias Harper tables, secondary indexes, or metadata
- register derived handles with Harper's existing close/drop paths while leaving native drain ordering for the later lifecycle integration
- support explicit derived-index retirement, concurrent worker drops, stale-root rejection, and retryable close failures

This is stacked on [Implement the shared derived-index transaction-log runtime #2533](https://github.com/HarperFast/harper/pull/2533). It does not activate `@fullText`, add a fulltext package dependency, or expose a customer storage-provider API. Those remain gated on the full Harper/fulltext lifecycle integration.

## Validation

- `npm run build`
- focused oxlint on the changed TypeScript and test files
- 19 focused Harper tests covering the storage adapter, derived runtime, and RocksDB handle release
- local cross-repo proof: the fulltext Tantivy Directory conformance and lifecycle harness passed through the native host transport into this adapter on a Harper-owned RocksDB database
- Gemini CLI and Claude CLI review; findings were adjudicated and the applicable transaction, namespace, liveness, close/drop, and error-preservation fixes are included
- final Gemini remediation-delta check: no findings

## Known performance boundary

rocksdb-js currently provides the required durability barrier as a database-wide flush over every registered column family. The fulltext publication scheduler must coalesce and rate-limit that barrier, and Phase 0 must measure foreground-write stalls, flush latency, compaction pressure, and multi-index behavior before schema activation.

<sub>Review-Coverage: authored=unknown; ran=none; rounds=1 @ 541339df0187</sub>

<sub>Human-Review-Need: 4 @ 541339df0187</sub>

### DIFF
--- a/resources/RocksDerivedIndexStorage.ts
+++ b/resources/RocksDerivedIndexStorage.ts
@@ -1,0 +1,104 @@
+import { RocksDatabase, type Transaction } from '@harperfast/rocksdb-js';
+import {
+	closeDerivedIndexStore,
+	dropDerivedIndexStore,
+	openDerivedIndexStore,
+	type RootDatabaseKind,
+} from './databases.ts';
+
+export type DerivedIndexWritePolicy = 'wal' | 'no-wal';
+
+export type DerivedIndexStorageMutation = { type: 'put'; key: Buffer; value: Buffer } | { type: 'delete'; key: Buffer };
+
+/**
+ * Synchronous host-storage surface used by native derived indexes. It deliberately exposes only
+ * operations Harper can satisfy with its existing RocksDB ownership and durability primitives.
+ */
+export class RocksDerivedIndexStorage {
+	#closed = false;
+	#rootStore: RootDatabaseKind & RocksDatabase;
+	#store: RocksDatabase;
+
+	constructor(rootStore: RootDatabaseKind, storeName: string) {
+		if (!(rootStore instanceof RocksDatabase)) throw new Error('Derived indexes require RocksDB storage');
+		this.#rootStore = rootStore;
+		this.#store = openDerivedIndexStore(rootStore, storeName);
+	}
+
+	read(key: Buffer): Buffer | undefined {
+		this.#assertOpen();
+		if (!Buffer.isBuffer(key)) throw new TypeError('Derived index storage keys must be Buffers');
+		const value = this.#store.getSync(key);
+		if (value === undefined) return undefined;
+		if (!Buffer.isBuffer(value)) throw new Error('Derived index storage contains a non-binary value');
+		return value;
+	}
+
+	write(mutations: Array<DerivedIndexStorageMutation>, policy: DerivedIndexWritePolicy): void {
+		this.#assertOpen();
+		if (policy !== 'wal') throw new Error('Harper derived index writes require the RocksDB WAL');
+		validateMutations(mutations);
+		if (mutations.length === 0) return;
+
+		const committed = this.#rootStore.transactionSync(
+			(transaction: Transaction) => {
+				for (const mutation of mutations) {
+					if (mutation.type === 'put') {
+						this.#store.putSync(mutation.key, mutation.value, { transaction });
+					} else {
+						this.#store.removeSync(mutation.key, { transaction });
+					}
+				}
+				return true;
+			},
+			{ retryOnBusy: true }
+		);
+		if (committed !== true) throw new Error('Derived index storage transaction did not commit');
+	}
+
+	/**
+	 * Establish durability for preceding WAL writes. rocksdb-js DBDescriptor::flush flushes every
+	 * registered column family regardless of the calling handle. This can stall unrelated writes,
+	 * so the fulltext publication scheduler must coalesce and rate-limit calls.
+	 */
+	sync(): void {
+		this.#assertOpen();
+		this.#rootStore.flushSync({ allowWriteStall: true });
+	}
+
+	/** Close only this derived column-family handle; the caller retains ownership of the root store. */
+	close(): void {
+		if (this.#closed) return;
+		closeDerivedIndexStore(this.#rootStore, this.#store);
+		this.#closed = true;
+	}
+
+	/** Drop this derived column family after its native readers and writer have drained. */
+	drop(): void {
+		this.#assertOpen();
+		dropDerivedIndexStore(this.#rootStore, this.#store);
+		this.#closed = true;
+	}
+
+	#assertOpen(): void {
+		if (this.#closed || !this.#rootStore.isOpen() || !this.#store.isOpen()) {
+			throw new Error('Derived index storage is closed');
+		}
+	}
+}
+
+function validateMutations(mutations: Array<DerivedIndexStorageMutation>): void {
+	if (!Array.isArray(mutations)) throw new TypeError('Derived index mutations must be an array');
+	for (const mutation of mutations) {
+		if (!mutation || !Buffer.isBuffer(mutation.key)) {
+			throw new TypeError('Derived index storage keys must be Buffers');
+		}
+		if (mutation.type === 'put') {
+			if (!Buffer.isBuffer(mutation.value)) {
+				throw new TypeError('Derived index storage values must be Buffers');
+			}
+		} else if (mutation.type !== 'delete') {
+			throw new TypeError('Unknown derived index storage mutation');
+		}
+	}
+}
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -305,6 +305,9 @@ interface RocksRootDatabase extends RocksDatabaseEx {
 	rootStore?: RocksRootDatabase;
 }
 
+const DERIVED_INDEX_STORES = Symbol('derived-index-stores');
+const DERIVED_INDEX_STORE_PREFIX = '`derived-index`/';
+
 export type RootDatabaseKind = LMDBRootDatabase | RocksRootDatabase;
 
 export type DatabaseWatcherEventMap = {
@@ -351,7 +354,7 @@ export function toRocksCompression(compression: unknown): unknown {
 	return compression;
 }
 
-function openRocksDatabase(path: string, options: RocksDatabaseOptions & { dupSort?: boolean }) {
+function openRocksDatabase(path: string, options: RocksDatabaseOptions & { dupSort?: boolean }, raw: boolean = false) {
 	options.disableWAL ??= true;
 	const legacyOptions = options as { compression?: unknown };
 	// A configured codec applies to every column family, overriding whatever per-table metadata
@@ -396,6 +399,8 @@ function openRocksDatabase(path: string, options: RocksDatabaseOptions & { dupSo
 	let db: RocksRootDatabase;
 	if (options.dupSort) {
 		db = new RocksIndexStore(path, options).open() as any;
+	} else if (raw) {
+		db = new RocksDatabase(path, options).open() as RocksRootDatabase;
 	} else {
 		db = new PrimaryRocksDatabase(path, options).open() as unknown as RocksRootDatabase;
 		// the RocksDB put and remove return promises, which masks thrown errors in non-awaiting calls to put/remove,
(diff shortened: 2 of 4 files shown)


resources/RocksDerivedIndexStorage.ts
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	/**
+	 * Establish durability for preceding WAL writes. rocksdb-js DBDescriptor::flush flushes every
+	 * registered column family regardless of the calling handle. This can stall unrelated writes,
+	 * so the fulltext publication scheduler must coalesce and rate-limit calls.
+	 */
+	sync(): void {
+		this.#assertOpen();
+		this.#rootStore.flushSync({ allowWriteStall: true });
```
kriszyp: This barrier is much heavier than the durability it needs, and it lands on machinery Harper tunes carefully.

As your own docblock notes, `flushSync` on the descriptor flushes *every* registered column family, not just this one. Per `DESIGN.md` ("Harper's normal RocksDB open path disables RocksDB's native WAL for primary and index column families"), those families are exactly the ones that hold their data in memtables on purpose — durability for them comes from the rocksdb-js transaction logs, not from a flush. So a derived index that calls `sync()` once a second on a database with, say, 20 tables and 40 secondary indexes cuts up to 60 SSTs per second, most of them nearly empty, purely to make a handful of derived-index WAL bytes durable. That is sustained L0 churn and compaction pressure caused entirely by the derived index, and with `allowWriteStall: true` the flush is also allowed to block foreground writers rather than fail fast. Worth noting there is no other production caller of `flushSync` in the repo today (only `unitTests/resources/auditLog.test.js`), so this introduces whole-database flushes to the steady state.

The primitive that actually matches the semantics is a WAL fsync (`FlushWAL(sync=true)` / `SyncWAL`) or a column-family-scoped flush — these writes go through the WAL, so fsyncing it is O(bytes written) instead of O(database size). That's a rocksdb-js change rather than a Harper one; I'd rather see it asked for upstream than have core commit to the whole-DB hammer as the contract. (There may also be something usable in the existing async barrier the transaction layer relies on for `options.flush` — `this.writes[0].store.flushed` at `resources/DatabaseTransaction.ts:1326` — though that would mean an async `sync()`.)

If the flush is genuinely the only barrier available right now, please at least make it self-limiting here rather than deferring the whole mitigation to a scheduler in another package: track a dirty flag set in `write()` and cleared in `sync()`, so a `sync()` with no intervening write is a no-op. That alone removes the common idle case, and it belongs with the class that knows whether anything was written.

resources/RocksDerivedIndexStorage.ts
```diff
@@ -0,0 +1,104 @@
...
+	close(): void {
+		if (this.#closed) return;
+		closeDerivedIndexStore(this.#rootStore, this.#store);
+		this.#closed = true;
+	}
+
+	/** Drop this derived column family after its native readers and writer have drained. */
+	drop(): void {
+		this.#assertOpen();
+		dropDerivedIndexStore(this.#rootStore, this.#store);
```
kriszyp: `drop()` only invalidates the handle it is called on. Harper's existing model for dropping a column family is a broadcast — see the comment on `ignoreAlreadyDropped` at `resources/Table.ts:250-258`: "Drops are broadcast to every worker thread and each holds its own handle to the same underlying family." That broadcast is what makes the `Column family already dropped` tolerance correct rather than a papering-over.

There is no equivalent here. Concrete sequence: workers A and B each construct a `RocksDerivedIndexStorage` for the same index. A calls `drop()`. B's `#store` handle is untouched — `store.isOpen()` still returns true and `#assertOpen()` passes — so B's `read()` keeps serving from the dropped family and B's `write()` keeps applying mutations into it. If the index is then recreated (as the test itself does at `unitTests/resources/rocksDerivedIndexStorage.test.js:198`), B is writing to the old family and reading data the new one will never see, with no error anywhere.

Two ways out: route derived drops through the same per-worker broadcast the table drop uses so every worker drops its own handle, or — if the contract really is that the native owner drains and closes all other workers' handles before any worker drops — state that precondition in the docblock, because `#assertOpen()` cannot detect the violation. Right now the docblock says "after its native readers and writer have drained", which reads as a single-worker statement.

Relatedly, the test at `unitTests/resources/rocksDerivedIndexStorage.test.js:191-201` calls the second handle `otherWorker`, but it is a second handle in the same thread; it demonstrates `ignoreAlreadyDropped`, not cross-worker safety. Renaming it would stop it reading as coverage for the case above.

resources/RocksDerivedIndexStorage.ts
```diff
@@ -0,0 +1,104 @@
+import { RocksDatabase, type Transaction } from '@harperfast/rocksdb-js';
+import {
+	closeDerivedIndexStore,
+	dropDerivedIndexStore,
+	openDerivedIndexStore,
+	type RootDatabaseKind,
+} from './databases.ts';
+
+export type DerivedIndexWritePolicy = 'wal' | 'no-wal';
```
kriszyp: `'no-wal'` is a member of the exported union that `write()` unconditionally rejects at line 39, so the type advertises a capability the implementation refuses. A caller who reads the type as permission gets a runtime error instead of a compile error — the worst ordering.

If the two-valued policy is dictated by the native index's host-storage interface (a tantivy-style `Directory`, say), say so in the docblock and keep the parameter as the adapter point. Otherwise I'd drop the parameter entirely — `write(mutations)` — or narrow the type to `'wal'`, so the type system carries the invariant rather than a throw. Either way it's one less thing to explain in the external surface.

resources/RocksDerivedIndexStorage.ts
```diff
@@ -0,0 +1,104 @@
...
+
+	/** Drop this derived column family after its native readers and writer have drained. */
+	drop(): void {
+		this.#assertOpen();
+		dropDerivedIndexStore(this.#rootStore, this.#store);
+		this.#closed = true;
+	}
+
+	#assertOpen(): void {
+		if (this.#closed || !this.#rootStore.isOpen() || !this.#store.isOpen()) {
```
kriszyp: `read()` is the hot path for an index — a fulltext query is thousands of block lookups — and every one pays `#closed` plus two `isOpen()` calls before the `getSync`. If `isOpen()` reads a JS-side status field this is cheap; if it crosses into native, it's two extra crossings per lookup on top of the get.

The polling is only there because `#closed` is not set when the root goes away underneath the instance (`closeDatabase`/`dropDatabase`/`closeBranchHandles` close the handle without telling the wrapper). That's solvable without polling: have the teardown path in `databases.ts` flip an invalidation flag on the registered storages — the `DERIVED_INDEX_STORES` set already gives it the list — and let `#assertOpen()` check a single boolean. Reads then cost one branch, and the invalidation becomes push rather than pull, which also removes the drop-related staleness in my comment on `drop()`.

If `isOpen()` is confirmed to be a plain property read, this is a non-issue — worth measuring before changing anything.

resources/databases.ts
```diff
@@ -412,6 +417,84 @@ function openRocksDatabase(path: string, options: RocksDatabaseOptions & { dupSo
 	return db;
 }
 
+/**
+ * Open an unencoded RocksDB column family for a derived index. The returned handle participates in
+ * Harper's existing RocksDB descriptor, memory configuration, compression policy, and backup
+ * boundary, but its lifecycle remains with the derived-index owner until that owner is registered
+ * in the database close graph.
+ */
+export function openDerivedIndexStore(rootStore: RootDatabaseKind, dbiKey: string): RocksDatabase {
```
kriszyp: `openDerivedIndexStore`, `closeDerivedIndexStore` and `dropDerivedIndexStore` are all exported, but the only non-test callers are the three methods of `RocksDerivedIndexStorage`, which is itself the intended entry point. That's two parallel surfaces for one capability, and the free functions are the more dangerous of the two — they hand out a raw `RocksDatabase` handle with no `#closed` tracking and no `#assertOpen()` guard, so a consumer that reaches for them gets none of the invalidation the class exists to provide.

Suggest making all three module-private and exporting only `RocksDerivedIndexStorage`; the plural `closeDerivedIndexStores` stays internal as it already is. The one test that uses `dropDerivedIndexStore` directly (`unitTests/resources/rocksDerivedIndexStorage.test.js:203-222`, the injected drop-then-close failure) can drive the same path through `RocksDerivedIndexStorage.drop()` by stubbing on the instance's store, so nothing is lost in coverage.

resources/databases.ts
```diff
@@ -412,6 +417,84 @@ function openRocksDatabase(path: string, options: RocksDatabaseOptions & { dupSo
...
+	}
+	if (typeof dbiKey !== 'string' || dbiKey.length === 0) {
+		throw new TypeError('Derived index store name must be a non-empty string');
+	}
+	const store = openRocksDatabase(
+		rootStore.path,
+		{
+			// Backticks are forbidden in customer schema names, so this can never alias a table,
+			// secondary index, or existing Harper internal column family.
+			name: `${DERIVED_INDEX_STORE_PREFIX}${dbiKey}`,
```
kriszyp: The store name is entirely opaque to core — `dbiKey` is only checked for being a non-empty string, and nothing records which table (or which index definition) a derived family belongs to. That leaves no retirement path for the case the derived index is attached to a table: `dropTable` drops the table's primary and index families (`resources/Table.ts:1777-1799`) and knows nothing about `` `derived-index`/<key> ``, so the derived family survives its table, invisible to the schema, still occupying space and still opened by every `DB::Open` for the life of the database. `dropDatabase` is fine — `destroy()` takes the directory — but table-level drop is not.

Since there is no consumer in-tree yet, the question is what the intended protocol is: does the owner subscribe to `databaseEventsEmitter`'s `dropTable` and call `drop()` itself, or should the name encode the table so core can sweep? Whichever it is, it belongs in this docblock — otherwise the first orphan shows up as unexplained disk usage long after the code that created it.

resources/databases.ts
```diff
@@ -351,7 +354,7 @@ export function toRocksCompression(compression: unknown): unknown {
 	return compression;
 }
 
-function openRocksDatabase(path: string, options: RocksDatabaseOptions & { dupSort?: boolean }) {
+function openRocksDatabase(path: string, options: RocksDatabaseOptions & { dupSort?: boolean }, raw: boolean = false) {
```
kriszyp: `raw` as a third positional boolean is a boolean trap on a function with eight call sites: `openRocksDatabase(path, opts, true)` gives the reader nothing at the call site, and what it actually selects is substantial — no `PrimaryRocksDatabase`, no `put`/`remove` sync aliasing, no encoder naming. Consider folding it into the options object (`{ raw: true }`, alongside the existing `dupSort` extension right above) so the intent travels with the call. Not blocking.

resources/databases.ts
```diff
@@ -412,6 +417,84 @@ function openRocksDatabase(path: string, options: RocksDatabaseOptions & { dupSo
...
+
+function closeDerivedIndexStores(rootStore: RootDatabaseKind, description: string): void {
+	const stores: Set<RocksDatabase> | undefined = (rootStore as any)[DERIVED_INDEX_STORES];
+	if (!stores) return;
+	for (const store of stores) {
+		try {
+			if (store.isOpen()) store.close();
+			stores.delete(store);
+		} catch (error) {
+			// Retain a handle whose close failed; removing it would make the live native claim
```
kriszyp: The rationale here doesn't hold for any of this function's three callers. `closeBranchHandles` closes the root and drops the branch immediately after (`resources/databases.ts:1760-1765`), `dropDatabase` follows with `close()` + `destroy()` (`resources/databases.ts:2040`), and `closeDatabase` deletes the root from `rocksdbDatabaseEnvs` on the next lines (`resources/databases.ts:2116-2120`). In every case the `rootStore` object — and with it the `DERIVED_INDEX_STORES` set that supposedly keeps the handle discoverable — is unreachable by the time any "later teardown attempt" could run.

The retention is harmless, but the comment claims a recovery path that doesn't exist. Either reword it to what it actually is (retain rather than silently forget, so a leaked native claim is at least logged once) or, if a real retry is wanted, hang the failed handles somewhere that outlives the root store. The singular `closeDerivedIndexStore` at line 455 is a different story — there the caller keeps the instance and can genuinely retry, which the test at `unitTests/resources/rocksDerivedIndexStorage.test.js:233-249` covers.

kriszyp:
Superseded by #2548, which carries this branch's commits and retargets the whole derived-index stack at `main` for a single review and merge. Closing so the stack no longer blocks that retarget. — Claude Fable 5.1

Issue #2537, opened 2026-09-09
Reads served from a partially-built index return 200 with missing rows: isIndexing diverges across threads


kriszyp:
## Summary

The `isIndexing` guard that makes a rebuilding index return `503 "… is not indexed yet"` instead of
serving partial results is **not consistent across threads in the same process**. On a live 5.2.7
cluster, for the same table and the same attribute at the same moment:

- worker threads had `indices.<attr>.isIndexing === true` → queries correctly threw `IndexRebuildingError`;
- the main thread (and the operations thread) had `isIndexing === false` → queries used the
  **partially-built** index and returned **200 with rows silently missing**.

The guard exists precisely to prevent the second behaviour, so wherever the flag doesn't propagate,
a rebuilding index becomes a silent-wrong-results surface.

## Evidence

Against a table whose index was ~9% built, on a thread with `isIndexing === false` — one record,
two lookups:

```
search_by_value  matchUrl="<value>"   ->  []                    # 200 OK, row missing
search_by_hash   id="<the record id>" ->  [{ id, matchUrl }]    # the record is present
```

The record exists; its indexed lookup says it doesn't. No error, no warning, no indication the index
is incomplete.

Sampling both reachable inspector threads on **all 15 nodes** of the cluster gave a completely
uniform split — main thread `isIndexing: false`, worker thread `isIndexing: true`, on every node.
So this is a systematic difference between thread roles, not a race that happened once.

## Where it comes from

`resources/databases.ts` propagates the flag to newly-opened dbis with:

```js
if (attributeDescriptor?.indexingPID) dbi.isIndexing = true;
```

and the failure paths additionally re-stamp both the attribute's dbi and the currently-active one:

```js
attribute.dbi.isIndexing = true;
const activeDbi = Table.indices[attribute.name];
if (activeDbi) activeDbi.isIndexing = true;
```

but the *trigger* path sets it only on the dbi it just opened. A thread that opened the table before
the descriptor gained `indexingPID`, or that holds a different dbi object than the one stamped, keeps
`isIndexing === false` and will happily read through a partial index.

Note the guard is also consulted only at the point a condition is *executed* as the driving index
lookup (`resources/search.ts`, the `!index || index.isIndexing || …` branch). `estimateCondition`
does not consult it at all, so the planner will happily choose a lead condition whose index the
executor is about to refuse.

## Impact

Any read against a rebuilding index can return incomplete results with a 200, non-deterministically
[...]

PR #2539, opened 2026-09-09
Index backfills resume from their checkpoint and yield the event loop, so large tables converge (harper#2536)

### LINKED ISSUE #2536: Index backfill never converges: resume checkpoint is discarded (start=undefined) and the loop never yields
## Summary

When a secondary-index backfill is interrupted, `runIndexing` deliberately leaves the index in
`isIndexing = true` and relies on a later "crash-recovery" trigger to resume it. **That trigger
cannot fire when Harper runs as PID 1 — i.e. in every container deployment.** The index is then
stuck mid-build permanently: it survives process restarts, logs nothing at the default log level,
and sets no `indexingFailed` marker, so nothing in `describe_all`, `get_status` or the logs
indicates the table is only partially indexed.

Observed on a live 15-node 5.2.7 cluster after an LMDB→RocksDB `migrate on start`: a 19.1M-row
table's five indices have been frozen mid-build for hours, at between **8.6% and 91%** complete
depending on the node.

## Root cause

`resources/databases.ts` re-triggers an incomplete backfill on exactly three conditions:

```js
attributeDescriptor?.indexingFailed ||
(attributeDescriptor?.indexingPID && attributeDescriptor?.indexingPID !== process.pid) ||
attributeDescriptor?.restartNumber < currentRestartGeneration
```

All three are structurally unreachable after a process restart in a container:

1. **`indexingPID !== process.pid`** — Harper is PID 1 in a container, and the descriptor was also
   written by a PID 1. The check is a tautology; it can never detect a restart.
2. **`restartNumber < currentRestartGeneration`** — this field exists specifically to cover (1)
   ("re-detect an incomplete index after a worker restart even when the new process reuses the old
   PID"). But the generation it compares against is **in-memory only** —
   `server/threads/manageThreads.ts`: `restartNumber: workerData?.restartNumber || 1` — so a fresh
   process starts at 1 while the persisted value is ≥2. `2 < 1` is false.
3. **`indexingFailed`** — never written on this path. Both exits that produce this state skip it:
   the `interrupted` early `return` inside the backfill loop, and the outer `catch`'s
   store-shutdown branch, which returns after a `logger.debug` (suppressed at the default
   `logging.level: warn`) with an explicit comment that recovery is expected from the trigger above.

So the mechanism that is *supposed* to resume the build is the one that can't run. Within a single
process the worker-generation counter does increment, so a worker restart recovers correctly — the
bug is specific to **process** restarts, which is what a container restart, redeploy, or
`Restarting Harper` always is.

## Evidence

Same table, same cluster, three nodes, sampled repeatedly over several minutes — all frozen, at
different points, which rules out a single bad record:

| node | persisted `lastIndexedKey` | `restartNumber` | index entries | of 19,122,820 |
|---|---|---|---|---|
| A | `12a3a089…` | 2 | 1,726,607 | 9.0% |
| B | `1167c12b…` | 2 | 1,636,343 | 8.6% |
| C | `595396cd…` | 8 | 17,394,907 | 91.0% |

Descriptors carry `indexingPID: 1`, no `indexingFailed`. Probing the primary-key space confirms the
frontier matches the entry count and does not advance. Node C was restarted after the stall and came
back still stuck, confirming restarts do not recover it.

The interruptions themselves were ordinary operator restarts during a migration procedure (no
crashes, no OOM — `OOMKilled=false`, `ExitCode=0`, nothing in `dmesg`). A long backfill spanning an
operator restart is a routine event, not an exotic one.

## Impact

Queries on the affected attribute are wrong in one of two ways, depending on the thread that serves
them (see the companion issue on the per-thread `isIndexing` divergence):

- threads with `isIndexing = true` throw `IndexRebuildingError` (503);
- threads with `isIndexing = false` serve from the partial index and return **200 with missing rows**.

Neither self-heals, and neither is visible in logs at the default level.

## Reproduction

1. Create a RocksDB-backed table large enough that the index backfill takes longer than a restart cycle.
2. Add an indexed attribute so a backfill starts.
3. Restart the Harper **process** mid-backfill (twice, so the persisted `restartNumber` exceeds 1).
4. Observe `isIndexing` stays `true` forever; no further progress; no `indexingFailed`; nothing logged
   above `debug`.

## Suggested direction

The recovery trigger needs a liveness signal that survives a process restart and does not depend on
PID identity:

- persist a per-process boot id (or a durable monotonic generation in the database rather than an
  in-memory counter) and treat a mismatch as crash recovery; **or**
- treat "descriptor has `indexingPID` set, but this process has no live `indexingOperation` for the
  table" as the recovery condition, which is directly observable and PID-independent.

Separately, an index parked mid-build should be visible without a debugger — the store-shutdown exit
should leave a durable marker and log above `debug`, and `describe_all` / `get_status` should report
an attribute whose index is incomplete.

kriszyp:
A secondary-index backfill on a large table could never converge: `runIndexing` discarded its own resume checkpoint (ordered-binary sorts `undefined` lowest, so the running-minimum guard never fired and every retrigger rescanned from the first record), and on a plain RocksDB index it never yielded the event loop (`RocksIndexStore.put` is synchronous, so the `outstanding` counter the yields were keyed to never left 0 and a 19M-row build ran as one ~18-minute turn until the worker was terminated). Backfills now resume from [the minimum persisted checkpoint across the attributes being built](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R3172) ([used as the scan's `start`](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R3206), [clearing every index when a full scan is needed](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R3207)), [yield every 100 scanned entries](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R3322) regardless of write-completion timing ([deletion entries count too](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R3245) and [still pace the checkpoints](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R3257)). Because the checkpoint is now actually consumed, it has to be trustworthy: [`persistCheckpoint`](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R3221) writes it only after the index writes it covers have settled and the RocksDB store has been flushed ([`flushIndexStores`](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R3156): index stores have no WAL), [at most once per 5s and never before 10k more records](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R3311), never advances after any index write fails ([every put's rejection is tracked](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R3275), not only a record's last), and stamps each checkpoint with its own key so [the trigger ignores a checkpoint left by an earlier release](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R2880) (which advanced past failed and unflushed writes) and rebuilds instead of resuming past a gap. [The completion path flushes the tail before persisting the ready descriptor and parks the index if that flush fails](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R3345); [the interrupt path awaits the in-flight checkpoint and persists synchronously](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R3301). Where to look hardest: `persistCheckpoint`, `flushIndexStores` and the trigger's `uncertifiedCheckpoint` — everything else is plumbing around them. Fixes #2536 (the per-thread `isIndexing` divergence and the dead-backfill marker are #2537, whose PR #2543 edits lines adjacent to the trigger change here; expect a trivial rebase on whichever lands second).

## For the human reviewer

Step-6 planning review returned `Framing-Verdict: better-alternative-exists` twice and both were **adopted**: round 1 made the checkpoint certify a successfully indexed prefix; after the task owner's ruling on the two open questions (1b, 2a) and a teammate's rig repro, the recheck accepted the layer and all four rejected alternatives (WAL-on index column families, colocating the checkpoint in the index CF, a trigger-side discard on process change, and accept-and-document) and asked for three corrections, all in: never join a flush already in flight, bind the stamp to its key, flush before readiness. Nine pre-push rounds; the last two produced no new actionable finding.

1. **Flush-gated, rate-limited checkpoints (ruling 1b).** RocksDB data and index stores open with `disableWAL: true` but the descriptor store with the WAL on, so a checkpoint written every 100 records could outlive the entries it certifies after a SIGKILL/OOM (proven on a child-process kill: 99 of the first 100 entries lost, checkpoint intact). A checkpoint is now persisted only after a flush, [at most once per 5s and never before 10k more records](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R3142) (`setIndexingCheckpointPeriod()` is the test seam) — the flush seals every column family in the database (`atomic_flush`), so the record floor keeps a slow backfill from imposing the 5s rate on unrelated tables. [At most one flush in flight and one queued per root store](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R3156); a caller never joins one already running because a flush only covers writes issued before it started. Consequences to weigh: a table under 10k rows never checkpoints mid-scan (a crash costs a sub-second rescan; the final barrier still flushes), each interruption reworks ≤10k records or ≤5s, and a crash loop shorter than that cannot converge. LMDB commits one environment in order and needs no flush.
2. **Only a key-stamped checkpoint resumes (ruling 2a, in the form the teammate's comment asked for).** `checkpointCertified` repeats the checkpoint key; [the trigger resumes only when it equals `lastIndexedKey`](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R2880) and otherwise rebuilds ([logged as reindex reason `uncertified-checkpoint`](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R2925)). That covers both legacy exits — `indexingFailed` and the marker-less `interrupted` return — so no field cleanup is needed before deploying; the one-time cost is that an in-progress legacy backfill restarts from record 0 on upgrade. This is the one descriptor schema addition in the PR (a field the old trigger drops, since it rebuilds the descriptor from the schema definition).
3. **The completion barrier.** [A flush before the ready descriptor](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R3345) closes a hole that exists on `main` too: a kill right after completion lost 0 of 10,000 index entries with it and all 10,000 without it (child-process test). A flush failure parks the index (`indexingFailed`) rather than announcing it complete.
4. **Checkpoint safety rests on index writes settling in dispatch order**: RocksDB puts throw synchronously, LMDB puts settle FIFO in one environment, so an earlier rejection is always visible before a later checkpoint fires. An explicit per-interval barrier over every put is a contained change if an engine ever settles out of order. The per-put rejection handler allocates one derived promise per LMDB put (bounded by the promise lmdb-js already creates per put; the RocksDB path never allocates).
5. **Checkpoint freezes at the first per-record error for the rest of the pass** (follow-up): a table with recurring transient errors advances one error-gap per retrigger — monotone, strictly better than main's full rescan, not single-pass convergence. An in-process retry of the failed keys would close it.
6. **One `start` at the minimum checkpoint across attributes**, re-putting already-indexed rows for attributes that were further ahead: index puts are idempotent and the redundant range is at most one interval. Reversible.
7. **Custom (HNSW) indexes still yield once per record** — unchanged from `main`'s `didSynchronousIndexing` branch; reviewers flagged it as a throughput ceiling, but it is not this change's to alter.
8. **`resumeStartKey` and `setIndexingCheckpointPeriod` are exported** solely for the unit tests; internal, trivially reversible.

## Verification

Route (b): new in-process end-to-end tests through `table()` → `runIndexing` → the real primary store on both engines, plus two real process kills. `unitTests/resources/indexBackfillConvergence.test.js` (with `indexBackfillConvergence-crash.js` as the child fixture): [the four `resumeStartKey` cases](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-7c2107e8bc4d28f75d46aab61a3e874430788e811bc434a1dd7a7b10d6b930c6R114); [an interrupted two-attribute backfill resumes with its scan opened at the persisted, key-stamped checkpoint](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-7c2107e8bc4d28f75d46aab61a3e874430788e811bc434a1dd7a7b10d6b930c6R146) (asserted on the `start` option and the first key visited, and that the stamp is cleared on completion); [a record whose index write throws, rejects asynchronously on a non-last value, or rejects while a later attribute's put resolves freezes the checkpoint before it and the retry revisits and indexes that record](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-7c2107e8bc4d28f75d46aab61a3e874430788e811bc434a1dd7a7b10d6b930c6R257); [unequal persisted checkpoints resume at the lower one, an absent one forces a full scan, and an unstamped legacy checkpoint is rebuilt from the first record](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-7c2107e8bc4d28f75d46aab61a3e874430788e811bc434a1dd7a7b10d6b930c6R336); [a child process SIGKILLs itself at its first persisted checkpoint](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-59f8e0017fc73f7982e4cc21e99259e7c429cf68970ee6b0da9d0302232af1e3R52) and [the parent resumes from that key through the PID-mismatch trigger](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-7c2107e8bc4d28f75d46aab61a3e874430788e811bc434a1dd7a7b10d6b930c6R429); [a child killed right after the ready descriptor](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-59f8e0017fc73f7982e4cc21e99259e7c429cf68970ee6b0da9d0302232af1e3R56) leaves [an index every entry of which survives](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-7c2107e8bc4d28f75d46aab61a3e874430788e811bc434a1dd7a7b10d6b930c6R508); [a rejecting flush parks the index and the retry completes](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-7c2107e8bc4d28f75d46aab61a3e874430788e811bc434a1dd7a7b10d6b930c6R561); [checkpoints are never closer than the record floor](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-7c2107e8bc4d28f75d46aab61a3e874430788e811bc434a1dd7a7b10d6b930c6R613); and [a 2000-row plain-index backfill yields every 100 records on RocksDB](https://github.com/HarperFast/harper/pull/2539/changes?w=1#diff-7c2107e8bc4d28f75d46aab61a3e874430788e811bc434a1dd7a7b10d6b930c6R671) (a `setImmediate` ticker stamps every visited key).

Fails-on-base (base `dist` built from `origin/main`'s `databases.ts`): the resumed scan's `start` was `undefined` (expected the checkpoint) and all 2000 records were visited in one event-loop turn; on the parent commit of the completion barrier, the kill-after-complete test kept 0 of 10,000 index entries. With the fix: 14/14 on RocksDB and 13/13 (+1 skipped) under `HARPER_STORAGE_ENGINE=lmdb`.

Gates on the final head: `npm run test:unit:resources` RocksDB 2266 passing / 0 failing; LMDB 1763 passing / 0 failing; `npm run test:unit:main` 5461 passing / 2 failing (the `configValidator` domain-socket path-length case, a local worktree-path artifact, and the `rsa_keys` token-authentication flake, both known on this box and unrelated). Earlier runs on this branch hit the pre-existing `longLivedTransactions` chain-link flake once per run; it passes alone. `oxlint` and `prettier --check` clean on the changed files. Measured: 0.8µs per `setImmediate` yield; a 100k-row RocksDB backfill ran at 216k rec/s on base and 219–244k rec/s with the fix; an `await` at the checkpoint instead of the deferred write made the LMDB backfill 8x slower, which is why the checkpoint write is attached to the last put's settlement.

https://claude.ai/code/session_01EVCC8xu4zk7eirYMg1ZHPb

Complexity: complicated

<sub>Review-Coverage: authored=claude; ran=codex,gemini; declined=cursor-grok,cursor-composer,domain; rounds=9 @ 763ba6f4d068</sub>

<sub>Human-Review-Need: 3 @ 763ba6f4d068</sub>

### DIFF
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -3124,6 +3124,19 @@ export function canonicalizeIndexOptions(value: any): any {
 }
 const MAX_OUTSTANDING_INDEXING = 1000;
 const MIN_OUTSTANDING_INDEXING = 10;
+// Records scanned between event-loop yields, and between resume checkpoints.
+const INDEXING_YIELD_INTERVAL = 100;
+const yieldEventTurn = () => new Promise((resolve) => setImmediate(resolve));
+// The primary-store key a resumed backfill scans from: the minimum persisted checkpoint across the
+// attributes being built, or undefined (scan everything) when any attribute has none. Exported for tests.
+export function resumeStartKey(attributes: { lastIndexedKey?: any }[]): any {
+	let start: any;
+	for (const attribute of attributes) {
+		if (attribute.lastIndexedKey == undefined) return undefined;
+		if (start === undefined || compareKeys(attribute.lastIndexedKey, start) < 0) start = attribute.lastIndexedKey;
+	}
+	return start;
+}
 async function runIndexing(Table, attributes, indicesToRemove, branchPath?: string) {
 	try {
 		logger.info(`Indexing ${Table.tableName} attributes`, attributes);
@@ -3141,10 +3154,8 @@ async function runIndexing(Table, attributes, indicesToRemove, branchPath?: stri
 		const attributesLength = attributes.length;
 		await new Promise((resolve) => setImmediate(resolve)); // yield event turn, indexing should consistently take at least one event turn
 		if (attributesLength > 0) {
-			let start: any;
+			const start = resumeStartKey(attributes);
 			for (const attribute of attributes) {
-				// if we are resuming, we need to start from the last key we indexed by all attributes
-				if (compareKeys(attribute.lastIndexedKey, start) < 0) start = attribute.lastIndexedKey;
 				if (attribute.lastIndexedKey == undefined) {
 					// if we are starting from the beginning, clear out any previous index entries since we are rewriting
 					if (attribute.dbi.clearAsync) {
@@ -3163,7 +3174,12 @@ async function runIndexing(Table, attributes, indicesToRemove, branchPath?: stri
 				versions: true,
 				snapshot: false, // don't hold a read transaction this whole time
 			})) {
-				if (!record) continue; // deletion entry
+				const atInterval = ++indexed % INDEXING_YIELD_INTERVAL === 0;
+				if (!record) {
+					// deletion entry
+					if (atInterval) await yieldEventTurn();
+					continue;
+				}
 				// TODO: Do we ever need to interrupt due to a schema change that was not a restart?
 				//if (Table.schemaVersion !== schemaVersion) return; // break out if there are any schema changes and let someone else pick it up
 				outstanding++;
@@ -3219,18 +3235,31 @@ async function runIndexing(Table, attributes, indicesToRemove, branchPath?: stri
 				if (workerData && workerData.restartNumber !== manageThreads.restartNumber) {
 					interrupted = true;
 				}
-				if (++indexed % 100 === 0 || interrupted) {
-					// occasionally update our progress so if we crash, we can resume
-					for (const attribute of attributes) {
-						attribute.lastIndexedKey = key;
-						Table.dbisDB.put(attribute.key, attribute);
-					}
+				if (atInterval || interrupted) {
+					// Checkpoint our progress so a crash can resume. A resumed scan starts at the checkpoint, so
+					// it must only ever name a key whose every predecessor was indexed: wait for the writes it
+					// covers to settle, and stop advancing it once any record has failed so the retry re-covers
+					// that record.
+					when(
+						lastResolution,
+						() => {
+							if (hadIndexingErrors) return;
+							try {
+								for (const attribute of attributes) {
+									attribute.lastIndexedKey = key;
+									Table.dbisDB.put(attribute.key, attribute);
+								}
+							} catch (error) {
+								// a lost checkpoint only costs the retry a rescan of this stretch
+								logger.debug(`Could not persist the indexing checkpoint for ${Table.tableName}`, error);
+							}
+						},
+						() => {} // already counted and logged by the rejection handler above
+					);
 					if (interrupted) return;
 				}
 				if (outstanding > MAX_OUTSTANDING_INDEXING) await lastResolution;
-				else if (outstanding > MIN_OUTSTANDING_INDEXING)
-					await new Promise((resolve) => setImmediate(resolve)); // yield event turn, don't want to use all computation
-				else if (didSynchronousIndexing) await new Promise((resolve) => setImmediate(resolve)); // custom indexes (e.g. HNSW) index synchronously and never raise `outstanding`; without this yield a large backfill runs in a single event-loop turn, starving keepalive/replication and queries and never letting the isIndexing flag be observed
+				else if (outstanding > MIN_OUTSTANDING_INDEXING || didSynchronousIndexing || atInterval) await yieldEventTurn(); // custom indexes (e.g. HNSW) index synchronously and a RocksDB put resolves synchronously, so neither raises `outstanding`; without this yield a large backfill runs in a single event-loop turn, starving keepalive/replication and queries and never letting the isIndexing flag be observed
 			}
 		}
 		// Await the last pending put. If it rejects, that is also an indexing error.
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -3180,13 +3180,14 @@ export function resumeStartKey(attributes: { lastIndexedKey?: any }[]): any {
 async function runIndexing(Table, attributes, indicesToRemove, branchPath?: string) {
 	let checkpointing;
 	let hadIndexingErrors = false;
-	let asyncRejectionReported = false;
-	const onIndexPutRejected = (error) => {
+	const attributeErrorReported = {};
+	const onIndexPutRejected = (property, error) => {
 		hadIndexingErrors = true;
-		if (asyncRejectionReported) return;
-		asyncRejectionReported = true;
-		logger.error(`Error indexing ${Table.tableName}`, error);
+		if (attributeErrorReported[property]) return;
+		attributeErrorReported[property] = true;
+		logger.error(`Error indexing attribute ${property}`, error);
 	};
+	const putRejectionHandlers = attributes.map((attribute) => (error) => onIndexPutRejected(attribute.name, error));
 	try {
 		logger.info(`Indexing ${Table.tableName} attributes`, attributes);
 		await signalling.signalSchemaChange(
@@ -3195,17 +3196,15 @@ async function runIndexing(Table, attributes, indicesToRemove, branchPath?: stri
 		let lastResolution;
 		for (const index of indicesToRemove) {
 			lastResolution = index.drop();
-			if (lastResolution?.then) lastResolution.then(undefined, onIndexPutRejected);
+			if (lastResolution?.then) lastResolution.then(undefined, (error) => onIndexPutRejected(index.name, error));
 		}
 		let interrupted;
-		const attributeErrorReported = {};
 		let indexed = 0;
 		const attributesLength = attributes.length;
 		await new Promise((resolve) => setImmediate(resolve)); // yield event turn, indexing should consistently take at least one event turn
 		if (attributesLength > 0) {
 			const start = resumeStartKey(attributes);
 			if (start === undefined) {
-				// a full scan rewrites every index it builds, so clear them all first
 				for (const attribute of attributes) {
 					if (attribute.dbi.clearAsync) {
 						// LMDB, note that we don't need to wait for this to complete, just gets enqueued in front of the other writes
@@ -3260,6 +3259,7 @@ async function runIndexing(Table, attributes, indicesToRemove, branchPath?: stri
 						const attribute = attributes[i];
 						const property = attribute.name;
 						const index = attribute.dbi;
+						const onPutRejected = putRejectionHandlers[i];
 						try {
 							const resolver = attribute.resolve;
 							const value = record && (resolver ? resolver(record) : record[property]);
@@ -3272,7 +3272,7 @@ async function runIndexing(Table, attributes, indicesToRemove, branchPath?: stri
 							if (values) {
 								for (let i = 0, l = values.length; i < l; i++) {
 									lastResolution = index.put(values[i], key);
-									if (lastResolution?.then) lastResolution.then(undefined, onIndexPutRejected);
+									if (lastResolution?.then) lastResolution.then(undefined, onPutRejected);
 								}
 							}
 						} catch (error) {
@@ -3293,7 +3293,7 @@ async function runIndexing(Table, attributes, indicesToRemove, branchPath?: stri
 				when(
 					lastResolution,
 					() => outstanding--,
-					() => outstanding-- // counted and logged by onIndexPutRejected
+					() => outstanding--
 				);
 				if (workerData && workerData.restartNumber !== manageThreads.restartNumber) {
 					interrupted = true;
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -3139,8 +3139,18 @@ export function setIndexingCheckpointPeriod(ms: number): number {
 	return previous;
 }
 const yieldEventTurn = () => new Promise((resolve) => setImmediate(resolve));
-// The primary-store key a resumed backfill scans from: the minimum persisted checkpoint across the
-// attributes being built, or undefined (scan everything) when any attribute has none.
+// Index stores have no WAL, so a flush is what makes a checkpoint's entries durable; one flush per root
+// store at a time, shared by every backfill running on that database.
+const indexingFlushes = new WeakMap<object, Promise<void>>();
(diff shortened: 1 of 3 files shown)


heskew:
**TL;DR:** this PR fixes convergence (confirmed on a rig), but as written it can bake a **permanent silent index gap** into the exact nodes it's meant to recover — and there's a **second, independent** WAL durability gap. Both matter before this recovers a stuck cluster.

**✔ Convergence fix — confirmed.** 1M-row SIGKILL crash-loop: base resumes at record 0 every cycle; with this PR the checkpoint advances each cycle and the final run gate-verifies complete.

**Blocker — resuming a legacy checkpoint can silently drop rows.**
- **What:** the new persist path is correct going forward, but checkpoints already on disk were written by the old path, which advanced `lastIndexedKey` every 100 records with **no `hadIndexingErrors` gate** — so a checkpoint can sit past a record whose `put` threw.
- **Effect:** resume starts past that record → completes → clears markers → row stays unindexed, index reads "complete." The exact production failure, baked in.
- **Confirmed 3 ways:** code trace + Codex + rig (checkpoint at record 299 despite a failure at 20 → this PR leaves 20 missing; clearing `lastIndexedKey` first recovers 300/300).
- **Fix:** stamp new-path checkpoints (`checkpointCertified`), treat unstamped legacy as `undefined`. The `indexingFailed → undefined` guard alone misses the `interrupted`-return path (no marker set there).
- **Field recovery:** clear `lastIndexedKey` on all affected nodes unconditionally before deploy.

**Separate issue — the index store is WAL-off.**
- **What:** `openRocksDatabase` defaults `disableWAL: true`; the checkpoint (descriptor) store overrides to `false` at every call site, the index store open (~L2213) doesn't.
- **Effect:** a hard kill can durably advance the checkpoint past index writes never fsync'd → a gap **independent of** the error-gating fix.
- **Fix:** flush-gate the checkpoint (persist after `flush()`), or open the index store WAL-on.

Both surfaced this session; rig harness available if useful.

_🤖 Findings from a session rig repro + cross-model (Codex) review._

kriszyp:
Thanks — the rig repro was exactly the evidence this needed, and both findings are now addressed on this branch.

**Legacy checkpoints.** Adopted your stamp, bound to the key rather than a boolean: every checkpoint written by the new path carries `checkpointCertified = lastIndexedKey`, and the trigger resumes only when the two match. That covers both legacy exits — the `indexingFailed` path and the marker-less `interrupted` return you pointed at — so no `lastIndexedKey` cleanup is needed before deploying; an unstamped checkpoint is a full rebuild (logged as reindex reason `uncertified-checkpoint`). The key binding also means an older binary that round-trips the field while advancing the key cannot produce a false certificate. Covered by the unstamped-legacy case in `indexBackfillConvergence.test.js`.

**WAL-off index store.** Went with the flush gate: a checkpoint is persisted only after `rootStore.flush()` on RocksDB, at most once per 5s and never before 10k more records (one flush seals every column family, so a slow backfill must not impose the rate on unrelated tables), with at most one flush in flight and one queued per database (a flush only covers writes issued before it started, so a caller never joins one already running). The completion path flushes the tail before persisting the ready descriptor and parks the index if that flush fails — a kill right after completion lost every unflushed entry before, on `main` too. Both are proven by child-process SIGKILL tests with no manual flush.

Your harness would be a good cross-check on the shipping policy (5s / 10k); the unit tests exercise it only at a 10k floor.

— Claude Fable 5.1

PR #2564, opened 2026-09-10
Re-assert @table settings on tables without a declared primary key

### LINKED ISSUE #2480: @table directive changes are silently ignored on existing tables without a declared primary key
## Summary

On an existing table whose GraphQL type declares no `@primaryKey` field, changing `@table` directive settings (`audit`, `expiration`, `eviction`, and likewise `sealed` / `replicate`) has no effect. The new values are silently dropped on every reload and restart, with no warning or log line. The same declaration is honored at table creation, so schema behavior is inconsistent between create and update.

## Mechanism

In `table()`, the only code that persists changed settings for an existing table sits inside the attributes loop under [`if (attribute.isPrimaryKey)`](https://github.com/HarperFast/harper/blob/0db6cabe6bd04a6e6af8e774227d5a879848d561/resources/databases.ts#L2554), with the [audit comparison](https://github.com/HarperFast/harper/blob/0db6cabe6bd04a6e6af8e774227d5a879848d561/resources/databases.ts#L2568) below it. A type with no `@primaryKey` field produces an attributes list containing no `isPrimaryKey` entry, and the graphql loader does not synthesize one ([graphql.ts uses `'id'` only as a display fallback in a log line](https://github.com/HarperFast/harper/blob/0db6cabe6bd04a6e6af8e774227d5a879848d561/resources/graphql.ts#L384)). The settings comparison therefore never evaluates.

At creation the new-table branch does honor the directive ([audit resolution](https://github.com/HarperFast/harper/blob/0db6cabe6bd04a6e6af8e774227d5a879848d561/resources/databases.ts#L2380), expiration/eviction adjacent), and such a table is created with an unnamed primary key (`primaryKeyAttribute = {}`). On boot every worker reads the persisted descriptor ([load path](https://github.com/HarperFast/harper/blob/0db6cabe6bd04a6e6af8e774227d5a879848d561/resources/databases.ts#L1145)), so the creation-time value wins forever.

## Reproduction

1. Deploy a component schema and write a few records:

```graphql
type CacheEntry @table(table: "cache_entry", audit: false) {
	value: String
}
```

2. Edit the directive to `audit: true`, restart.
3. `describe_table` still reports `"audit": false`, no audit records are produced, and replication (which rides the audit stream) never ships this table's writes.

Observed live on 5.2.7 across two full restarts (`describe_table` polled 12 times per restart, `false` every time). Mechanism confirmed against current main; the permalinks above are at 0db6cab.

## Sharp edges around the obvious workarounds

- Declaring a primary key after the fact does not help and is actively dangerous: with records present, the [primary-key-change guard](https://github.com/HarperFast/harper/blob/0db6cabe6bd04a6e6af8e774227d5a879848d561/resources/databases.ts#L2305) throws mid schema load ("Cannot change the primary key ... because it already contains records"), and the declared name can never match an unnamed key. The throw lands during component loading.
- No operations API alternative exists: [`create_table` rejects existing tables](https://github.com/HarperFast/harper/blob/0db6cabe6bd04a6e6af8e774227d5a879848d561/dataLayer/schema.ts#L137) and there is no alter operation.
- `logging.auditLog` cannot override the descriptor's persisted explicit value (same load path as above).

A table in this shape therefore has no supported route to ever change these settings; the only recourse is a manual write to the internal catalog row.

## Suggested fix

Run the settings re-assert even when the declaration carries no primary-key attribute, either by hoisting it out of the `isPrimaryKey` iteration and addressing the `<table>/` primary catalog row directly, or by synthesizing the primary-key attribute entry for PK-less declarations before the loop. While there, consider comparing `audit` / `sealed` / `replicate` against the persisted descriptor rather than the live `Table` values, matching the `schemaDefined` fix described in the comment just above the comparison, which addresses the same stale-disk-value class of bug.

Lavinia, via Claude

ldt1996:
Fixes #2480.

## Problem

On an existing table whose GraphQL type declares no `@primaryKey` field, changing `@table` directive settings (`audit`, `expiration`, `eviction`, `sealed`, `replicate`) had no effect. The only code that re-asserted changed settings on an existing table lived inside the attributes loop in `table()` under `if (attribute.isPrimaryKey)`. A PK-less type yields no `isPrimaryKey` attribute (the graphql loader does not synthesize one), so the comparison never evaluated and the creation-time persisted value won on every reload, silently. Creation honored the same declaration, so create and update disagreed.

## Fix

`resources/databases.ts`, in `table()`:

- The settings reconcile is hoisted out of the per-attribute loop and runs once per table, against the primary catalog row, resolved through the existing `primaryDescriptorKey()` helper (declared primary key name, falling back to the unnamed `tableName + '/'` key, which is exactly where a PK-less table's row lives). The PK branch of the loop now just `continue`s; the reconcile logic itself is the same block, moved.
- All of the block's recent semantics are preserved unchanged: skipped on the create path (`deferredPrimaryRow` publishes the row itself), gated off for cluster-origin callers, and the write happens against a descriptor re-read under the exclusive lock with the `tableIsDropping` check. Create and update resolve the same declaration, and every non-create declare now goes through one comparison regardless of PK-ness, so the paths cannot diverge again.
- `audit` / `sealed` / `replicate` are now compared against the persisted descriptor value instead of the live `Table` statics. The live values can be stale relative to disk in the same way as the `schemaDefined` case documented in the comment above the comparison (the descriptor is what every worker's next disk-load reads), so the descriptor is the value that decides whether a persist is needed. Directive values that are absent (`undefined`) still never trigger a persist; only an explicit changed value does.

Not changed, deliberately:

- The primary-key-change guard earlier in `table()` (declaring a PK on a populated table throws) is untouched and is evaluated before the reconcile; the existing guard test still passes.
- Post-persist semantics are exactly what the PK-declared path already did: `Table.enableAuditing()` when audit turns on, `hasChanges` driving `schemaVersion++` / `updatedAttributes()` / schema-change signalling, and `setTTLExpiration` at the end of `table()` for expiration/eviction. No store is reopened.

## Tests

`unitTests/resources/tableSettingsReassert.test.js`, all driven through `loadGQLSchema` (the real directive path):

1. PK-less table created with `audit: false`, records written (zero audit entries), directive changed to `audit: true`, reloaded: the catalog row and the live table both report `true`, and a subsequent write produces an audit entry. Fails on main, passes here.
2. PK-less table gains `expiration`/`eviction` on reload: persisted to the catalog row and applied (`expirationMS` set). Fails on main, passes here.
3. The same audit flip on a PK-declared table: unchanged behavior, passes on main and here.
4. A PK-less table reloaded with an unchanged directive does not rewrite its catalog row and does not bump `schemaVersion`.

Also run: `unitTests/resources/update-schema.test.js` (includes the populated-table primary-key-change guard test), `databases.test.js`, `auditLog.test.js`, and the full `test:unit:resources` suite.

## Claims from #2480 re-verified against current main

- The settings comparison exists only under `if (attribute.isPrimaryKey)` in the attributes loop (now at databases.ts ~1791, the issue's permalinks are at 0db6cab) and compares `audit`/`sealed`/`replicate` against live `Table` statics: confirmed.
- The graphql loader synthesizes no primary-key attribute for PK-less types (`'id'` appears only as a display fallback): confirmed.
- PK-less tables are created with `primaryKeyAttribute = {}` and their catalog row keyed at `tableName + '/'` with no name: confirmed (the boot loader's "table attribute for a table with no primary key" branch reads that row back as `tableDef.primary`).
- Boot reads audit/expiration/eviction/sealed/replicate from the persisted descriptor, so the creation-time value wins on every worker until the row is rewritten: confirmed.
- The primary-key-change guard throws for a populated table: confirmed by the existing update-schema test.

Lavinia, via Claude

### DIFF
--- a/resources/databases.ts
+++ b/resources/databases.ts
@@ -2711,41 +2711,38 @@ function declareTable<TableResourceType>(target: TableTarget, tableDefinition: T
 				}
 			}
 		}
-		// TODO: If we have attributes and the schemaDefined flag is not set, turn it on
-		// iterate through the attributes to ensure that we have all the dbis created and indexed
-		for (const attribute of attributes || []) {
-			if (attribute.relationship) {
-				refreshRelationshipAttributes = true;
-				continue;
-			}
-			if (attribute.computed) hasChanges = true;
-			let dbiKey = tableName + '/' + (attribute.name || '');
-			Object.defineProperty(attribute, 'key', { value: dbiKey, configurable: true });
-			let attributeDescriptor = attributesDbi.getSync(dbiKey);
-			if (attribute.isPrimaryKey) {
-				if (deferredPrimaryRow) continue;
-				attributeDescriptor = attributeDescriptor || attributesDbi.getSync((dbiKey = tableName + '/')) || {};
-				// Persist schemaDefined when the explicit live value disagrees with disk. Without this,
-				// a stale `false` (from a v4-era write or replicated event) survives every reload: the
-				// in-memory re-assert in the existing-Table branch only fixes the worker that ran @table,
-				// but other workers' next disk-load re-reads the stale value. The whole settings update is
-				// gated off for cluster-origin callers: their values come from this worker's (possibly
-				// stale) snapshot, so a rewrite could revert a newer local declaration already on disk.
-				const schemaDefinedMismatch = schemaDefinedExplicit && attributeDescriptor.schemaDefined !== schemaDefined;
-				// primary key can't change indexing, but settings can change
-				if (
-					origin !== 'cluster' &&
-					(schemaDefinedMismatch ||
-						(audit !== undefined && audit !== Table.audit) ||
-						(sealed !== undefined && sealed !== Table.sealed) ||
-						(replicate !== undefined && replicate !== Table.replicate) ||
-						(+expiration || undefined) !== (+attributeDescriptor.expiration || undefined) ||
-						(+eviction || undefined) !== (+attributeDescriptor.eviction || undefined) ||
-						attribute.type !== attributeDescriptor.type)
-				) {
-					exclusiveLock();
-					const currentPrimaryAttribute = attributesDbi.getSync(dbiKey);
-					if (!currentPrimaryAttribute || tableIsDropping(currentPrimaryAttribute, dbiKey)) continue;
+		// The settings reconcile runs once per table against the primary catalog row, not per
+		// attribute: a table declared without a primary key has no isPrimaryKey attribute (its
+		// catalog row is keyed at `tableName + '/'` with no name), and it must still pick up
+		// directive changes. Skipped on the create path (deferredPrimaryRow publishes the row
+		// itself), so create and update resolve the same declaration.
+		if (!deferredPrimaryRow) {
+			const declaredPrimaryAttribute = (attributes || []).find((attribute) => attribute.isPrimaryKey);
+			const primaryCatalogKey = primaryDescriptorKey();
+			const primaryDescriptor = attributesDbi.getSync(primaryCatalogKey) || {};
+			// Persist schemaDefined when the explicit live value disagrees with disk. Without this,
+			// a stale `false` (from a v4-era write or replicated event) survives every reload: the
+			// in-memory re-assert in the existing-Table branch only fixes the worker that ran @table,
+			// but other workers' next disk-load re-reads the stale value. The whole settings update is
+			// gated off for cluster-origin callers: their values come from this worker's (possibly
+			// stale) snapshot, so a rewrite could revert a newer local declaration already on disk.
+			const schemaDefinedMismatch = schemaDefinedExplicit && primaryDescriptor.schemaDefined !== schemaDefined;
+			// primary key can't change indexing, but settings can change; compared against the persisted
+			// descriptor (not the live Table) so a stale disk value is corrected too, same class of bug
+			// as the schemaDefined re-assert above
+			if (
+				origin !== 'cluster' &&
+				(schemaDefinedMismatch ||
+					(audit !== undefined && audit !== primaryDescriptor.audit) ||
+					(sealed !== undefined && sealed !== primaryDescriptor.sealed) ||
+					(replicate !== undefined && replicate !== primaryDescriptor.replicate) ||
+					(+expiration || undefined) !== (+primaryDescriptor.expiration || undefined) ||
+					(+eviction || undefined) !== (+primaryDescriptor.eviction || undefined) ||
+					(declaredPrimaryAttribute && declaredPrimaryAttribute.type !== primaryDescriptor.type))
+			) {
+				exclusiveLock();
+				const currentPrimaryAttribute = attributesDbi.getSync(primaryCatalogKey);
+				if (currentPrimaryAttribute && !tableIsDropping(currentPrimaryAttribute, primaryCatalogKey)) {
 					const updatedPrimaryAttribute = { ...currentPrimaryAttribute };
 					if (typeof audit === 'boolean') {
 						if (audit) Table.enableAuditing();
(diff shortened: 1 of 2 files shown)


resources/databases.ts
```diff
@@ -2711,41 +2711,38 @@ function declareTable<TableResourceType>(target: TableTarget, tableDefinition: T
 				}
```
kriszyp: Could this reconciliation be made symmetric before landing? Start with an existing no-primary-key table declared `@table(audit: true, expiration: 3600, eviction: 7200)`, then reload it with `audit: false, expiration: 0, eviction: 0`—zero is documented as disabling expiration. The mismatch is detected, but these truthy guards retain the old TTL values in the cloned descriptor; `Table.enableAuditing()` has no false path; and the later `setTTLExpiration()` call is skipped when every value is zero. `resetDatabases()` reuses the existing Table class, so the schema-change broadcast does not repair its captured audit/TTL state. Writes therefore remain audited and expiring, while the stale TTL also survives restart. `sealed` and `replicate` have the same live-state problem because they are only written to metadata. Please apply settings in both directions—including clearing durable TTL fields and updating the live Table closures/statics—and add reverse-direction tests for tables without a declared primary key.

<sub>(`resources/databases.ts:1813` is not part of this PR's diff — anchored to the nearest line this PR's diff can hold)</sub>

resources/databases.ts
```diff
@@ -2755,14 +2752,25 @@ function declareTable<TableResourceType>(target: TableTarget, tableDefinition: T
 					if (eviction) updatedPrimaryAttribute.eviction = +eviction;
```
kriszyp, line 2752: Could this reconciliation be made symmetric before landing? Start with an existing no-primary-key table declared `@table(audit: true, expiration: 3600, eviction: 7200)`, then reload it with `audit: false, expiration: 0, eviction: 0`—zero is documented as disabling expiration. The mismatch is detected, but these truthy guards retain the old TTL values in the cloned descriptor; `Table.enableAuditing()` has no false path; and the later `setTTLExpiration()` call is skipped when every value is zero. `resetDatabases()` reuses the existing Table class, so the schema-change broadcast does not repair its captured audit/TTL state. Writes therefore remain audited and expiring, while the stale TTL also survives restart. `sealed` and `replicate` have the same live-state problem because they are only written to metadata. Please apply settings in both directions—including clearing durable TTL fields and updating the live Table closures/statics—and add reverse-direction tests for tables without a declared primary key.

<sub>(`resources/databases.ts:2751` is not part of this PR's diff — anchored to the nearest line this PR's diff can hold)</sub>

kriszyp (commented):
Nice! I think one thing to consider with eviction/expiration.

kriszyp (changes requested):
Also, I still don't see any cross-model-review receipts in the PR body. These are required going forward and should be included by the harper-engineering-guidelines skill.

resources/databases.ts
kriszyp: Could this reconciliation be made symmetric before landing? Start with an existing no-primary-key table declared `@table(audit: true, expiration: 3600, eviction: 7200)`, then reload it with `audit: false, expiration: 0, eviction: 0`—zero is documented as disabling expiration. The mismatch is detected, but these truthy guards retain the old TTL values in the cloned descriptor; `Table.enableAuditing()` has no false path; and the later `setTTLExpiration()` call is skipped when every value is zero. `resetDatabases()` reuses the existing Table class, so the schema-change broadcast does not repair its captured audit/TTL state. Writes therefore remain audited and expiring, while the stale TTL also survives restart. `sealed` and `replicate` have the same live-state problem because they are only written to metadata. Please apply settings in both directions—including clearing durable TTL fields and updating the live Table closures/statics—and add reverse-direction tests for tables without a declared primary key.

<sub>(`resources/databases.ts:2751` is not part of this PR's diff — anchored to the nearest line this PR's diff can hold)</sub>

PR #2567, opened 2026-09-10
Shared derived-index runtime for native backends

kriszyp:
The shared post-commit derived-index runtime for RocksDB tables: lock-elected delivery of committed transaction-log mutations to a native index backend, exact cursor resume, rebuild from primary records, cross-worker readiness, and opt-in writer backpressure. This PR carries the whole stack — #2533 (runtime foundation), #2535 and the native-backend additions — so one review covers the derived index end to end and one merge lands it. [#2430](https://github.com/HarperFast/harper/pull/2430) (native HNSW) is rebuilt on this branch and is the first consumer; nothing here has a production caller until it lands.

The durable design is the **§ Derived-index runtime** section of the root `DESIGN.md` (invariants, ownership, cursor semantics, backend contract, bounded delivery, cadence, rebuild, fencing, shared readiness, lag policy). This description holds what is transient: scope, decisions, measurements, verification.

## What is here

**Runtime** (`resources/derivedIndexRuntime.ts`). One `DerivedIndexRuntime` per database wakes on the root store's `committed` event; each registered backend has its own process-wide lock, runner, cursor vector (`log name → completed transaction timestamp`) and readiness. The elected runner reads committed entries from every physical log, collects identities per transaction, resolves each changed record once against the primary store after its last occurrence, projects only the registered attributes, and delivers bounded batches. Resume proves every saved cursor with `exactStart`; a boundary the log cannot resolve, a corrupt frame, a removed log or a `reload` marker condemns the generation and rebuilds. Nothing on the commit path is added beyond the log entry that already exists and a local-only eviction marker for registered caching tables.

**Backend contract.** One interface: `attach(host)` (the owner-epoch fence), `deliver(batch)` (accept/defer/fail), `flush(reason)` (barrier request), `shutdown(epoch)` (quiescence), `getDurableCursor()`, `onStateChange()`, and optional `reset(epoch)`. `batch.records` is the last-write-wins view over distinct keys; `batch.through` is the cursor vector the batch completes and is withheld until an oversized transaction's closing chunk. Every hook is required — a backend that completes inside `deliver()` implements them trivially — because work that survives a method return is what an ownership handoff has to fence.

**Rebuild.** `rebuilding` is published, the previous epoch is quiesced, `reset(newEpoch)` runs, the committed tail of every log is captured, primary records are scanned and delivered in bounded chunks, the final chunk carries the tail as `through`, and replay resumes after it. Bounded retry with backoff; `unavailable` after the budget, honoured by peers.

**Shared readiness.** State, reason code, attempt count, rebuild request, lag flag and the owner-epoch counter in one `getUserSharedBuffer` allocation per index, read with plain `Atomics.load`s on any worker (`readDerivedIndexReadiness`). Same dependency as primary-key allocation, blob holds and HNSW node ids already carry.

**Lag policy** (opt-in, `maxLagMilliseconds`). The owner measures cursor distance, time parked, unread-log age and undurable-work age; past the budget every worker rejects local user writes to the index's tables with a retryable 503 (`DERIVED_INDEX_LAGGING`) at the staging layer (`_writeUpdate` / `_writeDelete` / `_writeInvalidate` / `_writeRelocate`), never canonical-source applies, replay or replication notifications. Cleared with hysteresis once the owner proves catch-up, and on every park the runtime cannot leave on its own.

## Simplifications made in review

The design was cut down before asking for review, each item verified against source rather than inferred:

- **One backend contract** instead of a synchronous/asynchronous split with a discriminant, registration-time hook validation for one side, a runtime check that a "synchronous" `flush` returned no promise, and a quiescence path for that undeclared promise. No shipped or planned backend is synchronous.
- **Plain words instead of a sequence lock.** Nothing read the free-form reason string the seqlock existed to publish; the shared record carries a `DerivedIndexReadinessReason` code, the epoch counter moved into the same buffer, and the 512-byte record became 32 bytes with no spin. This removed the only construct in the runtime that was novel with respect to Harper's existing use of `Atomics` over rocksdb-js shared buffers.
- **Committed-tail rebuild anchor** instead of the oldest retained entry. rocksdb-js advances `lastCommittedPosition` only to the earliest still-uncommitted write (`TransactionLogStore::commitFinished`, `uncommittedTransactionPositions.front()`), so a committed read is a contiguous physical prefix and nothing committed after the capture can sit behind the tail. The whole-retention-window replay after every rebuild, the capture-time reload-marker suppression, its `Date.now()` comparison against transaction timestamps, the shared reload word and the backward-clock residual are gone; a reload marker is met exactly once.
- **Lag latch clears on every terminal park**, not only on `unavailable`: a backend without `reset`, a runtime without `scanRecords`, a condemnation marker that could not be written, and entering a rebuild (no cursor to guard; readers act on `rebuilding`). Before this, three paths released ownership with the latch set and nothing scheduled to clear it, so writes stayed 503 indefinitely; one of them needed a commit wake to retry, and commits were what was being shed.
- `partial` dropped from `DerivedIndexTransaction` (nothing consumed it; the withheld cursor already says what is certified), and the per-log cursor checks that could not fail after the whole-vector match removed.
- **`RocksDerivedIndexStorage` (from #2535) removed.** It existed to back Tantivy's `KvDirectory` with a Harper-owned column family. The HNSW plane persists in its own mmap file, and #2568's benchmark shows the hosted path failing its own gates (~26 GB through ~108k JS read callbacks for 500 searches; a root-wide flush per publication) while Tantivy's native `MmapDirectory` passes, so Tantivy will persist the same way. The runtime never depended on it: the cursor is backend-owned. The derived-index column-family open/close/drop hooks in `databases.ts` go with it.
- **A log that has never written a file retains its beginning.** rocksdb-js reports `oldestSequenceNumber: 0` with `fileCount: 0` for it; the `=== 1` test read every brand-new database as a retention gap and failed every rebuild attempt. Found by restacking #2430 and defining a fresh table — the first real consumer exercising the runtime.

## Decisions and alternatives

- **Transaction log, not an `aftercommit` stream or a permanent worker-0 drainer.** Same-thread `aftercommit` dispatch has no durable cursor for logs shared by all workers and retains every audit object in a transaction; a fixed drainer pins all indexes to one worker. A lock-elected runner per backend gives each index one serial stream, its own cursor, and parallelism across workers. `transactionBroadcast`'s subscription registry was considered as the base and rejected: its cursors are live-subscription state, its lifecycle is customer-facing, and its aggregate results omit the log identity a cursor vector needs.
- **Not a transactional dirty-key outbox** (raised twice by planning review, overruled by the repo owner). It survives retention and works with auditing off, but adds a second durable write, a compaction stream and a cleanup protocol to every indexed mutation, a new column family and therefore a storage-format migration for every audited table, and still cannot commit an engine-specific native file atomically with RocksDB — the cursor protocol would be needed regardless.
- **Not in-transaction native mutation.** Neither Tantivy nor the mmap plane shares RocksDB rollback: pre-commit application leaves phantoms after abort and puts native construction on request latency.
- **Rebuild in the runtime, not per backend.** Reset → scan → project → deliver → replay is identical for vector and full-text and its duplication is how two runtimes came to exist.
- **Resolve after the last collected occurrence, not on first encounter** (adopted from planning review): a writer committing between two occurrences of a key inside one chunk would otherwise have its later state certified by the cursor while the earlier state stayed indexed.
- **Eviction marker rather than keeping evicted documents and filtering candidates against the primary store**, which would make this index the one exception to Harper's index lifecycle and add a primary read to every result.
- **Timer-coalesced idle completion** rather than a barrier at every idle pass, which would cost one barrier per write for arrivals spaced just beyond drain completion.
- **Out of scope:** a shared head reader for indexes whose cursors stay within a cohort (no measured N-scan bottleneck yet; the contracts allow it later), and retention pinning (rocksdb-js has no protected-position registration).

## Measurements

Synthetic native-cost bench (`derivedIndexRuntime.bench.js`, 350 µs/apply, 5 ms barrier): 1,000 occurrences over 50 keys in one window — 1,000 applies / 359 ms through `transactions`, 50 applies / 24 ms through `records`; queued application held event-loop delay to 6.8 ms max versus 963 ms inline at the same ~2,800 mutations/s. At 1,500 arrivals/s, a 100 ms / 512-mutation cadence delivered 1,492/s with p99 write→durable 116 ms and 30 barriers in 3 s; the defaults (1 s / 4096) give p50 1.27 s and 4 barriers. On the native HNSW backend (#2430's ingest bench, 384-d): 265 native applies for 1,000 commits over 50 keys, foreground put 0.09 ms, apply 0.365 ms, barrier ~9 ms.

## Verification

- Derived suites (`derivedIndexRuntime`, `derivedIndexRuntimeNativeBackend`, `derivedIndexRuntimeRocks`, `derivedIndexRegistry`): 81 passing. The native-backend suite covers coalescing, oversized transactions, deferral, cadence, rebuild/retry/exhaustion, handoff with an apply scheduled and with a flush pending, shutdown failure holding the lock, condemnation across a simulated restart, lag trip/clear/hysteresis, the three terminal-park cases (each proven red without the fix), tail-anchored rebuilds across several logs, and a real worker thread reading the owner's publication through the binding.
- End to end on real audited RocksDB tables: commit → runtime → queued backend → barrier → readiness; primary scan and replay; write shedding at the staging layer with canonical-source bypass.
- #2430 restacked on this branch: `vectorIndexPlane.test.js` 22 passing against `@harperfast/hnsw` 0.2.1 — two workers into one native file, a committed write replayed after its worker dies, reload markers, a cursor outside retention, an empty index answering no results, a populated index that lost its file answering 503.
- `npm run build`, `tsc --noEmit`, `lint:required`, prettier: clean.

Refs #2489. Supersedes #2533 and #2535 (whose storage adapter is dropped here, see above).


resources/databases.ts
```diff
@@ -2086,6 +2172,7 @@ export function closeDatabase(databaseName: string): boolean {
 	}
 	for (const rootStore of rootStores) {
 		removeStorageReclamation(rootStore.path);
+		closeDerivedIndexStores(rootStore, `while closing database ${databaseName}`);
```
kriszyp: `closeDerivedIndexStores()` closes raw column-family handles immediately in branch close, database drop, and `closeDatabase()`, but none of those paths know about or await `DerivedIndexRuntime.stop()`. A concrete failure is: a backend accepts a batch, online restore calls `closeDatabase()`, this closes its store while the native apply/shutdown queue is still active, and the later apply targets a closed column family after `closeDatabase()` already returned success. That contradicts the runtime's promise that fulfilled stop means nothing can still write. Please register derived runtime/backend owners in the database close graph and quiesce them before closing these handles. Async drop can await that barrier; synchronous close/branch/job paths need an explicit design that either guarantees owners were already stopped or changes their orchestration. A delayed-backend teardown test should prove the store stays open until shutdown settles.

resources/derivedIndexRuntime.ts
```diff
@@ -0,0 +1,1970 @@
...
+		}
+		this.#discardProgress();
+		const backend = this.#registration.backend;
+		const epoch = this.#ownerEpoch;
+		this.#releasingSince = this.#options.now();
+		const unlock = () => {
+			this.#releasing = undefined;
+			this.#releasingSince = undefined;
+			try {
+				this.#logStore.unlock(this.#lockKey);
```
kriszyp: If `unlock()` throws, this catch only logs it. `#releasing` therefore resolves, `stop()` fulfills and unregisters the tables, even though the process-wide runner lock may remain held. A replacement runner then waits indefinitely, while the caller is allowed to close backend storage believing teardown completed safely. Please treat an unlock failure like a failed release: retain/reach the runner, set a release failure, and reject the cached stop promise so an explicit retry can resolve the lock state. An injected unlock failure test would cover the promised stop/handoff contract.

resources/derivedIndexRuntime.ts
```diff
@@ -0,0 +1,1970 @@
...
+			this.#admitWrites();
+		};
+		let flushed: void | Promise<void>;
+		try {
+			flushed = backend.flush('shutdown');
+		} catch (error) {
+			logger.warn?.(`Derived index '${backend.id}' shutdown flush request threw`, error);
+		}
+		const settling = Promise.allSettled([this.#resetting, flushed]).then(() => undefined);
+		this.#releasing = settling.then(() => (epoch === undefined ? undefined : this.#quiesce(epoch))).then(unlock, hold);
```
kriszyp: This can call `shutdown(epoch)` twice after one successful quiescence. If a rebuild already awaits `#quiesce(oldEpoch)`, then `stop()` enters `#release()` and waits here for a slower `flush('shutdown')`. The rebuild's shutdown may resolve and clear `#quiescing` while that flush is pending; once the flush settles, this line calls `#quiesce(oldEpoch)` again. Successful shutdown is not specified as idempotent, and a second call that rejects falsely holds the lock and fails stop despite the epoch already being quiescent. Please capture/reuse the in-flight quiescence promise before waiting for flush/reset, or memoize successful epochs. A test where the rebuild shutdown resolves before the shutdown flush would pin the race.

resources/derivedIndexRuntime.ts
```diff
@@ -0,0 +1,1970 @@
...
+	 * write), so nothing committed after this point can sit behind it: the scan covers everything up to
+	 * the tail and the replay from it covers everything after, including any reload marker committed
+	 * during the scan, which then demands its own rebuild. A log with no committed transaction is left
+	 * out of the cursor and read from its beginning, which it must still retain.
+	 */
+	#captureBoundary(): DerivedIndexCursor {
+		const boundary: DerivedIndexCursor = { format: 1, logs: {} };
+		for (const logName of this.#logStore.rootStore.listLogs()) {
+			let tail: number | undefined;
+			const range = this.#logStore.getRange({ log: logName, start: 0 });
```
kriszyp: Every rebuild exhausts each physical log from `start: 0` synchronously to discover its tail. A newly registered backend normally has no cursor, so its first activation enters this path; on a busy database with a day of retained audit traffic, this can decode millions of entries in one event-loop turn before the bounded table scan even begins. `maxMillisecondsPerTurn` provides no protection here. The root fix appears to belong in rocksdb-js: expose the committed tail/physical boundary it already maintains (the design references `lastCommittedPosition`) and snapshot that in O(1). Please use such an API here, or otherwise acquire a fixed boundary incrementally with yields; simply following a live iterator could chase the head indefinitely.

PR #2568, opened 2026-09-11
Prepare the Harper full-text benchmark for native storage

kylebernhardy:
## Purpose

Prepare the existing benchmark for Harper's native Tantivy integration. The revised [integration and benchmark plan](https://github.com/HarperFast/harper/blob/codex/fulltext-derived-index-slice/docs/fulltext-derived-index-vertical-slice.md) uses the actual derived runtime and Fulltext backend with local native index files.

This PR remains draft. The executable harness still measures the earlier hosted-storage experiment; it has not yet been converted, and its results do not qualify native Harper indexing.

## Current contents and next work

The harness currently compares no-index, native filesystem, WAL-only hosted, and root-flush hosted paths with fixed-arrival foreground/search load, callback and byte counts, publication cost, reopen parity, and explicit failure gates.

Adapt it to measure standalone native Fulltext, Harper without a full-text index, and Harper's real derived runtime/backend using the same native wrapper. Keep the existing workload validation and measurement helpers. Remove hosted-storage comparisons from release gates and preserve prior results as historical evidence.

Native checkpoint publication/readback is the prerequisite. The integrated arm must prove replay, worker handoff, generation identity, reader visibility and source restore before timing claims. Internal queries do not establish Table.search/REST product latency.

This PR is stacked on [Add a full-text derived-index backend](https://github.com/HarperFast/harper/pull/2569). The native-only architecture is documented in [Native Tantivy storage and Harper derived indexes](https://github.com/HarperFast/fulltext/blob/codex/native-storage-design/docs/native-storage-integration.md).

## Evidence and verification

Earlier hosted diagnostic runs exposed large callback volume and foreground/search latency under root flush. Those measurements apply to the retired hosted path and do not predict native integration performance.

The September 14 revision changes documentation only. Its rebase preserves all nine existing benchmark commits unchanged, verified by range-diff. No native integration benchmark was run for this revision.

Claude and Gemini CLI reviewed the native-only design. This is planning review, not new implementation coverage. Existing code/test evidence predates the native conversion.

## For the human reviewer

The remaining review target is the planned native conversion and its real integration evidence. No hosted durability policy or RocksDB Directory decision is needed. Keep this draft until the native path is implemented and measured.

Comment generated by kAIle (GPT-6)

<sub>Review-Coverage: authored=unknown; ran=none; rounds=1 @ 093afbc4beb8</sub>

<sub>Human-Review-Need: 4 @ 093afbc4beb8</sub>


benchmarks/fulltext-hosted/run.cjs
```diff
@@ -0,0 +1,594 @@
...
+			const scheduled = next;
+			next += interval;
+			const query = queries[sequence % queries.length];
+			const bucket = control.indexing ? latencies.duringIndexing : latencies.afterIndexing;
+			const request = index.search({ text: query, limit: 10 }).then((result) => {
+				assert(result.hits.length > 0, `full-text query returned no hits: ${query}`);
+				bucket.push(performance.now() - scheduled);
+			});
+			pending.add(request);
+			request.then(
```
kriszyp: Rejected requests are dropped without a trace. `request.then(..., () => pending.delete(request))` handles the rejection and removes the promise from `pending` one microtask later. Unless the loop happens to be waiting in `Promise.race(pending)` at that exact moment, nothing ever sees the error. The loop is normally parked in `await delay(wait)`, so the promise is already gone by the time the final `Promise.all(pending)` runs.

What goes wrong: if Fulltext search fails during a merge or publication (a native error, a transport `readTimeoutMs` expiry, or the `hits.length > 0` assert), that sample just goes missing. `search.duringIndexing` then reports the p99 of the searches that *succeeded*, and the gate can pass. For example, 400 of 500 searches could fail and the ≥100-sample floor would still be met. The foreground loop at `benchmarks/fulltext-hosted/run.cjs:319-323` has the same problem. There, a failed `Table.put` only shows up later, as a confusing soak-read `assert` failure or a sample-count gate failure.

Suggestion: record the first rejection, e.g. `request.catch((error) => { control.error ??= error; })`. Then check `control.error` after `await searches` / `await foreground` and throw it, or count errors and fail the gate on any. With errors captured there, the `.catch(() => undefined)` guards at lines 123-124 can stay as unhandled-rejection guards only.

benchmarks/fulltext-hosted/run.cjs
```diff
@@ -0,0 +1,594 @@
...
+
+	async function driveSearch(index, control, latencies) {
+		if (!index) return;
+		const queries = ['waterproof trail shoes', 'wireless headphones', 'cotton blue shirt', 'outdoor product'];
+		const interval = 1_000 / options.queryRate;
+		let sequence = 0;
+		const pending = new Set();
+		await control.searchReady.promise;
+		let next = performance.now();
+		while (!control.stop && sequence < options.queryCount) {
```
kriszyp: The search gate only covers the first ~second of indexing. This loop stops after `options.queryCount` dispatches. With the defaults (and the README sweep), that is 500 queries at 500/s, so every search runs within about 1 s of the first publication (`start === 0`). At that point the reader holds only the first `--batch-size` documents (1,000). With `--publication-ms 30000`, no further publication happens during the entire search window, so every sample hits a 1,000-document generation.

The result: `search p99 < 50ms` is judged on a tiny index. It never runs while Tantivy is merging large segments, while a publication or reload is running, or while `root-flush` stalls the event loop later in the run. Those are the conditions this gate exists to catch. `afterIndexing` is also almost always empty.

Suggestion: drive searches at `--query-rate` for as long as `control.indexing` is true, optionally followed by a fixed after-indexing window. Treat `--queries` as a minimum or after-indexing count rather than a total cap. Memory stays bounded by the existing 256 in-flight cap.

benchmarks/fulltext-hosted/run.cjs
```diff
@@ -0,0 +1,594 @@
...
+			eventLoop.enable();
+			const workloadStarted = performance.now();
+			const foreground = driveForeground(tables, control, workloadStarted, foregroundLatencies);
+			const searches = driveSearch(indexContext?.index, control, searchLatencies);
+			foreground.catch(() => undefined);
+			searches.catch(() => undefined);
+			const indexing = await indexDocuments(indexContext, publicationMilliseconds, control);
+			control.indexing = false;
+			await searches;
+			const remaining = options.minimumDurationMs - (performance.now() - workloadStarted);
```
kriszyp: The foreground p99 that gets gated covers the whole window: indexing plus idle padding up to `--minimum-duration-ms`. When indexing finishes in a few seconds, most of the 10 s window is idle. For example, if indexing takes 2 s at the default 100 rps, only ~200 of ~1,000 samples overlap indexing. A root-flush stall that hits fewer than 10 samples then falls below p99 and doesn't show up.

The windows also aren't matched across arms. The no-index arm's window has a different mix of load than an arm whose indexing runs past 10 s, so the "matching no-index control" comparison is weaker than it looks.

Suggestion: bucket foreground samples the way search already does (`duringIndexing` / `afterIndexing`, by scheduled time relative to the end of indexing). Gate on the during-indexing bucket. The no-index arm still runs `indexDocuments` (encoding only), so it has an equivalent window to compare against.

benchmarks/fulltext-hosted/metrics.mts
```diff
@@ -0,0 +1,85 @@
...
+		}
+	}
+	if (input.eventLoopSampleCount < 500) {
+		failures.push(`event-loop delay has ${input.eventLoopSampleCount} samples; at least 500 are required`);
+	} else if (input.eventLoopP99Milliseconds >= 20) {
+		failures.push(`event-loop p99 ${input.eventLoopP99Milliseconds.toFixed(3)}ms is not below 20ms`);
+	}
+	if (input.foregroundSampleCount < 1_000) {
+		failures.push(`foreground writes have ${input.foregroundSampleCount} samples; at least 1000 are required`);
+	} else if (input.foregroundP99Milliseconds > input.baselineForegroundP99Milliseconds * 1.2) {
```
kriszyp: This relative gate is likely to flip on noise rather than on anything architectural:

- Latency is measured from scheduled time after `await delay(wait)`, so every sample includes `setTimeout` lateness (typically ~1 ms). The no-index p99 is therefore mostly timer jitter, and 20% of it is well under a millisecond.
- The baseline is a single run, and it always runs first in a fresh process (`arms[0]` must be `no-index`). Later arms run with more heap growth, a different JIT state, and prior databases' files on disk, so the order bias always points the same way.
- With the defaults (`--foreground-rate 100`, `--minimum-duration-ms 10000`), the sample count lands at ~1000–1001. That is right on the `>= 1000` floor at line 52, and p99 is the 10th-worst sample.

Suggestions:
- Add an absolute tolerance, e.g. `max(baseline * 1.2, baseline + N ms)`.
- Raise the default rate or duration so the sample floor is comfortably exceeded.
- Consider running the no-index control both before and after the indexed arms, and gating against the worse of the two.

The process exits nonzero on this gate, so a noisy false fail or false pass feeds directly into the architecture decision.

benchmarks/fulltext-hosted/run.cjs
```diff
@@ -0,0 +1,594 @@
...
+				p50Milliseconds: eventLoop.percentile(50) / 1e6,
+				p95Milliseconds: eventLoop.percentile(95) / 1e6,
+				p99Milliseconds: eventLoop.percentile(99) / 1e6,
+				maxMilliseconds: eventLoop.max / 1e6,
+			};
+			eventLoop.disable();
+			const storage = indexContext?.storageMeasurements
+				? storageSummary(indexContext.storageMeasurements, indexing.publications)
+				: undefined;
+			if (storage) assert.ok(storage.maxValueBytes <= 256 * 1024, 'stored value exceeded the 256 KiB chunk bound');
```
kriszyp: This `assert.ok`, like any other throw inside an arm (e.g. the parity asserts in `closeAndReopen`), rejects the top-level async function. `FULLTEXT_HOSTED_RESULT` is only printed once the entire sweep finishes. So a failure in the last cell of a `--table-counts 1,16,128 --publication-ms 1000,5000,30000` sweep throws away every earlier arm's measurements. The README also calls the 256 KiB check part of the harness's gating ("rejects a value larger than 256 KiB").

There is a related cleanup gap. If `closeAndReopen` throws after reopening (lines 401 or 407), `reopened` and `reopenedStorage` are never closed: `context.disposed` isn't set, and the `finally` only closes the original handle. `closeDatabase` then closes the derived column family while a live native runtime is still attached to it.

Suggestions:
- Report the value bound as a gate failure, e.g. pass `maxValueBytes` into `evaluateGates`.
- Wrap each `runArm` so an error becomes `{ arm, error }` in `results` and sets `failed`, or print one result line per arm.
- Close the reopened handles in a `try/finally`.

docs/fulltext-derived-index-vertical-slice.md
```diff
@@ -0,0 +1,474 @@
...
+generated foreground RocksDB operation p99 while indexing and search run. Record database-wide
+compaction/stall counters, aggregate compaction/SST properties for the foreground column families,
+and a post-run soak read. Sweep both publication interval and dirty unrelated table count. Run
+no-index, native, and Harper-callback controls at identical offered load on one machine, measuring
+foreground and search latency from scheduled dispatch time so coordinated omission does not hide
+stalls.
+
+The only existing explicit durability barrier is
+`RocksDerivedIndexStorage.sync()` -> `rootStore.flushSync({ allowWriteStall: true })`, which flushes
+every column family and blocks the storage-serving worker. Engineering has declined a new
```
kriszyp: On root cause: I don't think the expensive part of `root-flush` is inherent to the callback bridge. It comes from `RocksDerivedIndexStorage.sync()` having only `flushSync()` to make earlier WAL writes durable. That call flushes the memtable of every column family, with `allowWriteStall`. RocksDB's `FlushWAL(true)` / `SyncWAL()` gives exactly the needed guarantee (fsync the WAL, no memtable flush, no write stall), and `@harperfast/rocksdb-js` is our own package.

The harness brackets the choice with a free sync (`wal-only`) and a root-wide flush (`root-flush`). Nothing measures a WAL fsync, which is likely the policy we would actually want.

I couldn't verify the rocksdb-js API from this checkout (it has no `node_modules`). Could the doc link the decision behind "Engineering has declined" and give its reason? If the concern is API surface area, a WAL-sync arm is probably the cheapest way to revisit it with data. It would also shrink the `wal-replay` vs `root-flush` fault matrix this doc plans.

docs/fulltext-derived-index-vertical-slice.md
```diff
@@ -0,0 +1,474 @@
...
+This is not the same as inheriting one shared WAL durability setting. Harper opens primary table
+column families with native WAL disabled and recovers them through rocksdb-js transaction logs;
+`openDerivedIndexStore` explicitly opens the derived column family with WAL enabled. A root flush
+therefore can make both primary memtables and the derived generation durable, while `wal-replay`
+can recover a derived cursor whose independently stored source-log boundary is unavailable. The
+fault matrix measures that skew and its rebuild rate. The root-flush arm is not eliminated by
+configuration inference because the stores have different recovery mechanisms.
+
+RocksDB's automatic memtable flush is not a third policy because Harper has no notification that
+can atomically define which Tantivy cursor it covered. The harness sweeps 1 s, 5 s, and 30 s
```
kriszyp: This paragraph doesn't match what the harness does. `run.cjs` applies the same `--publication-ms` sweep to every arm, and `wal-only` publishes on that schedule rather than "at #2567's normal cadence". The harness never constructs a `DerivedIndexRuntime`, which the doc's own later paragraph on the backend-integration benchmark also says.

Also, lines 12-18 pin specific SHAs and say #2567 "currently targets `main`". Once #2567 lands, that goes stale in a doc otherwise written as a lasting design. Consider moving the point-in-time details into the PR description, or relying on the result JSON's `revisions` field, which already records them.

benchmarks/README.md
```diff
@@ -38,6 +39,75 @@ On macOS/Windows, set up loopback addresses once (Linux has them by default):
...
+	--queries 500 \
+	--query-rate 500 \
+	--soak-reads 5000 \
+	--foreground-rate 500 \
+  --publication-ms 1000,5000,30000 \
+  --table-counts 1,16,128
+```
+
+The command writes progress to stderr and emits one machine-readable stdout line beginning with
+`FULLTEXT_HOSTED_RESULT`. It exits nonzero when an architecture gate fails: search p99 below 50 ms,
```
kriszyp: Nit: as written, this sounds like the benchmark fails *when* search p99 is below 50 ms. Something like "It exits nonzero unless every architecture gate passes: search p99 below 50 ms, …" would make the list read as pass conditions.

benchmarks/fulltext-hosted/run.cjs
```diff
@@ -0,0 +1,673 @@
...
+			foreground = driveForeground(tables, control, workloadStarted, foregroundLatencies);
+			searches = driveSearch(indexContext?.index, control, searchLatencies);
+			foreground.catch(() => undefined);
+			searches.catch(() => undefined);
+			const indexingStarted = performance.now();
+			const indexing = await indexDocuments(indexContext, publicationMilliseconds, control);
+			indexing.elapsedMilliseconds = performance.now() - indexingStarted;
+			control.indexing = false;
+			await searches;
+			const foregroundCount = await foreground;
```
kriszyp: Thanks for switching to a fixed `--minimum-duration-ms` window with the indexing-coverage gate. Every arm now gets the same arrival window, so the mismatched-window part of this is fixed.

One part remains: the gated p99 is still taken over the whole window, including the idle time after indexing. With the 12 s default at 100 rps, a 12 s window has ~1,200 samples, so p99 is roughly the 12th-worst. Take an arm that finishes indexing in 2 s: only ~200 of its samples overlap indexing. Stalls during indexing that affect fewer than ~12 samples then fall below p99 and don't show up. The faster an arm indexes, the more its window is diluted.

Suggestion (cheap): bucket foreground samples the way search already does, by scheduled time before or after `control.indexing` goes false. Report the during-indexing p99, and consider gating on it, comparing against the no-index arm's own encode-only indexing interval. The whole-window p99 can stay as a secondary number.

benchmarks/fulltext-hosted/metrics.mts
```diff
@@ -0,0 +1,96 @@
...
+		input.indexingElapsedMilliseconds > input.foregroundWindowMilliseconds
+	) {
+		failures.push(
+			`indexing took ${input.indexingElapsedMilliseconds.toFixed(0)}ms and outlasted the ` +
+				`${input.foregroundWindowMilliseconds.toFixed(0)}ms foreground window`
+		);
+	}
+	if (input.foregroundSampleCount < 1_000) {
+		failures.push(`foreground writes have ${input.foregroundSampleCount} samples; at least 1000 are required`);
+	} else if (input.foregroundP99Milliseconds > input.baselineForegroundP99Milliseconds * 1.2) {
```
kriszyp: The 12 s default fixes the sample-count part of this: 100 rps now gives ~1,200 samples, comfortably above the 1000 floor. Two issues remain that can still flip this gate on noise:

- **No absolute floor.** Latency is measured from scheduled time after `await delay(wait)`, so every sample includes `setTimeout` lateness (typically ~1 ms). The no-index p99 is mostly timer jitter, and 20% of it is well under a millisecond.
- **Order bias.** The baseline is one run, and it always runs first in a fresh process (`arms[0]` must be `no-index`). Later arms run with more heap growth, a different JIT state, and prior databases' files on disk, so the bias always points the same way.

Suggestion: add an absolute tolerance, e.g. `max(baseline * 1.2, baseline + N ms)`. Consider also running the no-index control after the indexed arms and gating against the worse of the two runs. The process exits nonzero on this gate, so a noisy result feeds directly into the architecture decision.

kriszyp:
> A final 10,000-document no-index,root-flush verification run delivered 1,200 foreground writes in each 12-second window and passed corpus plus close/reopen parity. 

What do all these invented terms mean? What size are the documents? You are comparing with and without the derived index? What is the throughput ratio of the two? Does WAL refer to RocksDB's WAL or rocksdb-js's WAL?

kylebernhardy:
> What do all these invented terms mean? What size are the documents? You are comparing with and without the derived index? What is the throughput ratio of the two? Does WAL refer to RocksDB's WAL or rocksdb-js's WAL?

@kriszyp 
The benchmark used 10,000 synthetic product records averaging 161 bytes each, with an ID, title, description, and category.

“No index” is the control: no Tantivy index, but the same product encoding and unrelated Harper writes. “Root flush” stores the Tantivy index in Harper’s RocksDB and calls the database-wide flushSync() for each Tantivy sync request.

Both runs completed 1,200 Harper writes over 12 seconds, so throughput was identical at the fixed offered rate. This was not a maximum-throughput test. The meaningful difference was write latency: p99 increased from 8.36 ms without the index to 135.8 ms with root flushes—about 16× worse. We need a separate saturation benchmark to measure the actual indexed-versus-unindexed throughput ratio.

WAL refers to RocksDB’s native write-ahead log, accessed through rocksdb-js, not Harper’s higher-level transaction logs. The correctness check confirmed that search results survived closing and reopening the index. The result shows that the integration works, but mapping every Tantivy sync to a database-wide flush causes unacceptable interference.

Let me know what else would be helpful.

kylebernhardy:
@kriszyp one other thing ommitted in the initial PR description:

These benchmarks were an early architecture check, not a customer performance test. They compare Tantivy’s native storage with Harper’s RocksDB integration to measure indexing cost, storage callback overhead, durability options, and the effect on unrelated Harper traffic.

They confirmed that the RocksDB-backed index can persist, reopen, and search correctly. They also showed that flushing the entire RocksDB database for every Tantivy sync causes unacceptable latency, and that Tantivy generates substantial storage callback traffic.

This gives us direction for the production design: use the derived-index protocol for publication and recovery, avoid root-wide flushing, and reduce callback overhead. End-to-end ingestion, `Table.search`, REST, replication, and large-catalog performance still need separate benchmarks.

These benchmarks are intended to be used internally by us to continue to evaluate overhead and perfomance of the fulltext index so we can make sure we optimize properly as we iterate on development.

kriszyp:
Posted on the other PR, but I think this all aligns and your benchmarking indicates the same thing: Tantivy native storage is the way to go.

benchmarks/fulltext-hosted/run.cjs
```diff
@@ -0,0 +1,673 @@
...
+		let publishMilliseconds = 0;
+		let publications = 0;
+		let sincePublication = performance.now();
+		for (let start = 0; start < options.documents; start += options.batchSize) {
+			const end = Math.min(options.documents, start + options.batchSize);
+			const packed = fulltextNative.encodeMutationBatch({
+				upserts: Array.from({ length: end - start }, (_, offset) => product(start + offset)),
+			});
+			packedBytes += packed.byteLength;
+			if (context) {
```
kriszyp, line 323: This `if (context)` is what makes the `no-index` arm cheap: it still encodes every batch (`fulltextNative.encodeMutationBatch` on line 319) but skips `apply` and `publish`. That is the right control for the *encoding* cost, but it also means the baseline stops generating load far earlier in the window than the arms it is the baseline for.

Concretely, with the defaults: `no-index` runs 100 iterations of encode-1000-products plus a `setImmediate`, which is on the order of a few hundred milliseconds, then `indexDocuments` returns and the arm is idle for the remaining ~11.5 s of the 12 s foreground window. A `root-flush` run over the same 100k documents pushes every one through the JS callback bridge and the root flush, so it stays loaded for a much larger fraction of that window. `evaluateGates` then compares `foreground.p99Milliseconds` from an almost-entirely-idle distribution against one from a partially-loaded distribution and calls the ratio a "20% regression limit" (`metrics.mts:65`).

This compounds, but is distinct from, the whole-window dilution noted on line 154: that one is about sensitivity within an arm, this one is about comparability *between* arms. The bias direction is that the baseline p99 is depressed more than the indexed arm's, which tightens the effective threshold — conservative, but it means neither a pass nor a fail maps to the "identical offered load" comparison the README describes at `benchmarks/README.md:71-73`.

A couple of options, cheapest first:

- Report the actual load duration per arm. `indexing.elapsedMilliseconds` is already computed for every arm (line 151, including `no-index`); surfacing `indexingElapsedMilliseconds / foregroundWindowMilliseconds` in the result line would at least make a 3%-vs-70% comparison visible to whoever reads the JSON.
- Bucket `foregroundLatencies` by `control.indexing` the way `driveSearch` already buckets searches (line 421), gate on the during-indexing bucket, and raise `--foreground-rate` or the window so that bucket clears the 1,000-sample floor. The `no-index` arm's whole-window number stays usable as the baseline since it is uniformly idle anyway.
- Or give `no-index` comparable duration by pacing its encode loop to the observed indexed-arm duration. That is more code and more assumptions; I'd lean toward the second option.

benchmarks/fulltext-hosted/run.cjs
```diff
@@ -0,0 +1,673 @@
...
+			await searches;
+			const foregroundCount = await foreground;
+			control.stop = true;
+			const foregroundWindowMilliseconds = options.minimumDurationMs;
+			const workloadMilliseconds = performance.now() - workloadStarted;
+			const eventLoopSummary = {
+				count: eventLoop.count,
+				p50Milliseconds: eventLoop.percentile(50) / 1e6,
+				p95Milliseconds: eventLoop.percentile(95) / 1e6,
+				p99Milliseconds: eventLoop.percentile(99) / 1e6,
```
kriszyp, line 162: `monitorEventLoopDelay` records the elapsed time between the histogram's timer ticks, and that elapsed time includes the resolution itself. With `resolution: 10` (line 130), an idle event loop records ~10 ms per sample. So `eventLoop.percentile(99) / 1e6` starts at ~10 ms before any stall. The `>= 20` check in `metrics.mts:51` then fails an arm once its p99 stall reaches ~10 ms, half the "event-loop delay p99 below 20 ms" budget in the README and the design doc. (`@fastify/under-pressure` subtracts the resolution for the same reason.) To confirm, look at the no-index arm's `eventLoop.p50Milliseconds` in the result JSON: it should be about 10.

Separately, this histogram runs from workload start until the fixed foreground window closes, so the post-indexing idle dilution described on line 154 applies here too. The two effects push in opposite directions (a tighter effective threshold, a diluted sample), so today's pass/fail doesn't map cleanly to the documented budget.

Suggestion: put the resolution in a named constant and subtract it when summarizing, e.g. `Math.max(0, eventLoop.percentile(99) / 1e6 - resolution)`. Also, take the summary and call `eventLoop.reset()` where `control.indexing = false` is set (line 152). That gives a during-indexing figure to gate on, and the whole-window number can stay in the report.

benchmarks/fulltext-hosted/run.cjs
```diff
@@ -0,0 +1,673 @@
...
+				for (const mutation of mutations) {
+					measurements.writeBytes += mutation.key.byteLength + (mutation.value?.byteLength ?? 0);
+					measurements.maxValueBytes = Math.max(measurements.maxValueBytes, mutation.value?.byteLength ?? 0);
+				}
+				measurements.synchronousCommittedEvents += committedEvents() - eventsBefore;
+			},
+			sync() {
+				measurements.syncs++;
+				const started = performance.now();
+				if (arm === 'root-flush') rawStorage.sync();
```
kriszyp, line 520: Carrying this forward from the previous round as a root-cause note, not a request to change the harness. It was previously anchored to the paragraph in `docs/fulltext-derived-index-vertical-slice.md` that recorded the decision; that paragraph is gone in the rewrite, so re-anchoring it to the line that actually implements the barrier.

Engineering declined a new WAL-sync primitive, so `root-flush` is the only barrier Harper can offer today. `RocksDerivedIndexStorage.sync()` calls `rootStore.flushSync({ allowWriteStall: true })` (`resources/RocksDerivedIndexStorage.ts:66`), which flushes the memtables of every column family in that root store — including the `Noise*` foreground tables this benchmark is measuring. The rocksdb-js typings I could check expose only `flush()`/`flushSync(options?)`. There is no `SyncWAL`/`FlushWAL(sync)` and no per-write `sync` option (the only `sync` flag is on backups). So `root-flush` measures durability achieved through a database-wide memtable flush, not the cost of durability itself. A WAL fsync would give the WAL-enabled derived column family the same machine-loss boundary for its own writes, without flushing other tables, stalling writes, or adding L0 compaction debt.

That distinction matters for the decision this harness feeds. If `root-flush` fails the gates and `wal-only` passes, the conclusion is "`flushSync` is too expensive", not "a durable barrier is too expensive". Suggestion: say in `benchmarks/README.md` — where `root-flush` is described around line 65 — that it is an *upper bound* on barrier cost. If the rocksdb-js decision is ever revisited, a WAL-sync arm is a one-branch addition right here in `measuredStorage.sync()`.

docs/fulltext-derived-index-vertical-slice.md
```diff
@@ -0,0 +1,118 @@
...
+storage path. It must be adapted before its results can qualify the native integration.
+
+## Objective
+
+Measure Harper's real derived-index runtime and Fulltext backend using Tantivy native filesystem
+storage. The wrapper and Harper now share the native datasource. The accepted architecture is
+[Native Tantivy storage and Harper derived indexes](https://github.com/HarperFast/fulltext/blob/codex/native-storage-design/docs/native-storage-integration.md).
+The [backend design](fulltext-derived-index-backend.md) owns the ordered admission/recovery adapter.
+
+Existing hosted Directory, WAL-only, root-flush, and KV transport benchmarks remain historical
```
kriszyp, line 13: New this round, from the doc rewrite. These two lines say the hosted Directory / WAL-only / root-flush arms "remain historical diagnostics" and that "their thresholds are not its release gate" — but `benchmarks/README.md:106` still introduces the same numbers as "The architecture gates", and `benchmarks/README.md:110` states "A failed gate or aborted arm exits nonzero." The README does add "These are diagnostic gates, not published customer SLOs" further down, which is close, but a reader who lands on the README first will reasonably take the 50 ms / 20 ms / 20% thresholds as the bar this work has to clear, which is exactly what this doc is now disclaiming.

Two small things would close the gap:

- Have the README's gate paragraph point at this document and repeat its framing — these thresholds gate *this diagnostic run*, not the native integration and not a product SLO.
- The filename is `fulltext-derived-index-vertical-slice.md` but the title is now "Native Fulltext derived-index integration and benchmark", and the body is a plan for an architecture the shipped harness does not exercise. Either rename the file to match, or add a one-line pointer at the top of the README section saying which document describes the code in `benchmarks/fulltext-hosted/` versus which describes the planned successor. As it stands, someone looking for the design rationale behind the code this PR ships finds a document whose first paragraph tells them the code is not the thing being planned.

kriszyp (commented):
This can now be closed, right?

kylebernhardy:
> This can now be closed, right?
> 
> 🤖 Reviewed with Claude

yep closing now

Issue #2510, opened 2026-09-05
Implement full-text generation lifecycle, rebuild, eviction, backup, and status


kylebernhardy:
## Outcome

Connect the full-text backend to Harper lifecycle and the shared derived-index protocol in HarperFast/harper#2489.

## Acceptance criteria

- Every write-capable worker delivers committed relevant mutations after commit without adding a second source transaction.
- Whole-record cache/TTL eviction removes local derived index eligibility through the shared local-only eviction action.
- Startup resumes exactly from the durable aggregate watermark or marks the generation `NEEDS_REBUILD` when history is unavailable.
- Shadow build, catch-up, durability, atomic activation, rollback, and old-generation deletion are coordinated.
- Backup, checkpoint, restore, close, schema removal, and database drop fence derived operations correctly.
- Status exposes generation, health, lag, queues, searchable checkpoint, rebuild progress, and actionable failures.
- A backend failure cannot suppress change feeds, another derived backend, or an already-durable source commit.

Comment generated by kAIle (Codex GPT-5)

kylebernhardy:
The first backend/lifecycle slice is now in [Add a full-text derived-index backend #2569](https://github.com/HarperFast/harper/pull/2569).

It adds the package-independent Fulltext backend contract, owner acquisition and handoff behavior, bounded queue/backpressure, cursor publication and validation, recovery, fail-closed reset, and a real Harper RocksDB storage route using a deterministic engine. Schema/query activation, the versioned native Fulltext artifact, production generation management, backup integration, and native end-to-end performance remain follow-up work under this issue.

Comment generated by kAIle (GPT-5)

PR #37, opened 2025-11-18
RocksDB storage

kriszyp:
Adding support for running on our new Rocksdb-js storage system.
The major changes in this PR are:
* Updates to interaction with DB stores that normalize appropriately across LMDB and RocksDB
* Ability to find and open LMDB or RocksDB databases
* Updates to auditStore/transaction-log to wrap RocksDB's log writing and querying functionality, and handle its own encoding and deleting
* Blobs are deleted when their record is deleted so that transaction logs can be deleted without scanning before deletion
* New DatabaseTransaction class for handling distinctions of RocksDB transactions; RocksDB can immediately write to transactions, which can be concurrent with other open transactions.
* Resource API updates for v5 that interact with Rocksdb, including new save() method (for saving data in the midst of a transaction).

PR #2569, opened 2026-09-11
Add a full-text derived-index backend

kylebernhardy:
Harper’s derived-index runtime now has a native full-text backend that converts committed record changes into `@harperfast/fulltext` mutation batches and publishes the Harper replay cursor with the Tantivy commit. Harper records and audit logs remain authoritative; the local Tantivy files are reusable derived state.

This is the backend slice of [Implement full-text generation lifecycle, rebuild, eviction, backup, and status #2510](https://github.com/HarperFast/harper/issues/2510). It uses the batch contract from [Add exact native mutation batch partitioning #37](https://github.com/HarperFast/fulltext/pull/37), which is merged. Schema activation, reader attachment, `Table.search`, REST exposure, and status reporting remain follow-up work.

## For the human reviewer

1. Confirm the ownership boundary: Harper coordinates derived-index delivery, backpressure, recovery, and validation; Fulltext owns Tantivy, its single writer, mutation partitioning, native files, and cursor durability.
2. Confirm the failure contract: an apply or publish failure rollback-closes accepted native work, releases ownership, and resumes after the last durable cursor. Missing, incompatible, corrupt, or condemned local files rebuild from Harper’s source records. A cursor that exceeds the native payload limit parks that backend instance until its configuration is replaced; resetting the files cannot make the cursor smaller.
3. Confirm rejected upserts should delete any stale indexed document and increment Harper’s rejected-record count. Preserving old searchable text would be incorrect.
4. Confirm cursor-only progress should use the normal publication cadence. Additional commit coalescing should be based on benchmark data, not a second backend-specific timer.

## Activation gate

A transient synchronous inspection failure currently condemns and rebuilds the local generation. This backend remains inert in this PR. Activation must either add a non-condemning acquisition result to the derived-index protocol or prove that the supported native inspection path cannot throw a recoverable error.

#### Write and storage flow

Harper sends logical record changes across the wrapper boundary. The wrapper owns their exact native representation and the node-local index files.

```mermaid
flowchart LR
    subgraph Harper
        Records[Authoritative records]
        Log[Committed audit log]
        Runtime[DerivedIndexRuntime]
        Backend[Full-text backend]
    end

    subgraph Fulltext
        API[Native mutation API v3]
        Writer[Single Tantivy writer]
        Files[Node-local index files]
    end

    Records --> Runtime
    Log --> Runtime
    Runtime -->|resolved batches| Backend
    Backend -->|logical mutation batch| API
    Backend -->|publish cursor| API
    API --> Writer
    Writer --> Files
    Files -->|inspect durable cursor| Backend
```

#### Boundary contract

- [`fullTextDerivedIndex.ts`](https://github.com/HarperFast/harper/pull/2569/changes?w=1#diff-5179224a4e766e922a50ff7737a6855cc084314decbb63d17831630a1bec3afbR124) implements the Harper queue, bounded slices, publication barrier, rollback, replay, and shutdown behavior.
- [`fullTextNativeBinding.ts`](https://github.com/HarperFast/harper/pull/2569/changes?w=1#diff-2003af761b083176c96faaca97c43360e3210d10b3fad26c25875bd04a405e4aR34) loads the native wrapper lazily and validates the required API and storage capabilities.
- [`nativeFullTextDerivedIndexLifecycle.ts`](https://github.com/HarperFast/harper/pull/2569/changes?w=1#diff-da5b8af7955acf2a3bb45f2df15ebca3d3a9226803a5e38b8ca033040f9f143eR32) owns deterministic local paths, inspection, open, reset, and retired-directory cleanup.

#### Durable cursor flow

Mutations and the Harper cursor become durable in the same native commit. Failed accepted work is discarded before the runtime retries.

```mermaid
sequenceDiagram
    participant Runtime as DerivedIndexRuntime
    participant Backend as Harper backend
    participant Fulltext as Fulltext wrapper
    participant Tantivy

    Runtime->>Backend: deliver resolved batch
    Backend->>Fulltext: applyMutationBatch
    Runtime->>Backend: flush
    Backend->>Fulltext: publish(cursor)
    Fulltext->>Tantivy: commit mutations and cursor
    alt apply or publish fails
        Backend->>Fulltext: close(rollback)
        Backend-->>Runtime: accepted work lost
        Runtime->>Backend: replay after durable cursor
    end
```

#### Restart behavior

On restart, Harper inspects and reuses compatible local files, then replays from the cursor stored in the last Tantivy commit. A new replica, restore without index files, incompatible format, or condemned generation rebuilds locally before full-text queries become available; normal Harper reads and writes remain available.

## Changes

- The implementation and recovery rationale are recorded in [`DESIGN.md`](https://github.com/HarperFast/harper/pull/2569/changes?w=1#diff-3dc5dd454e080eb849ee5efaf79df2585fbe0a06804d69249b4c81da05a63875R2702), the [implementation note](https://github.com/HarperFast/harper/pull/2569/changes?w=1#diff-a90a7ebe7dd85fba97609c67d59d75719610bb4d90717845fe185e07aa5bfa53R60), and the [resource index](https://github.com/HarperFast/harper/pull/2569/changes?w=1#diff-5a3cac25d02c21deca87f353afcbe03c11dd1601a64cfb947c57219f7334a708R28).
- Runtime tests cover [durable-cursor ownership transfer and bounded shutdown](https://github.com/HarperFast/harper/pull/2569/changes?w=1#diff-8b7e708692a233151c7d04a33d1f63051a7a851103717d3d028de4ab9a75c218R141), [Rocks-backed source replay and rebuild](https://github.com/HarperFast/harper/pull/2569/changes?w=1#diff-2541f0d08ae43f2f2def3d2df40d568d638b8fb91e8b57773ef98fc459b83df9R130), and [process restart after an unclean exit](https://github.com/HarperFast/harper/pull/2569/changes?w=1#diff-2b63f926c8f8f1f4031804edd9b81ae260ed413f31e553419d202a9b8697964fR1).
- Backend tests cover [queueing, batching, rollback, replay, cursor limits, reset, and epoch fencing](https://github.com/HarperFast/harper/pull/2569/changes?w=1#diff-6219e3a78fb04c52aaa58cab7fdac4daab2c29eba3229a429346def524fea052R137), while lifecycle tests cover [paths, inspection, reset, reclamation, and binding compatibility](https://github.com/HarperFast/harper/pull/2569/changes?w=1#diff-c804f2475278feb239f8bcdcb7c7dd9f2dcd0c8e585a8839505ccf3240590a3cR110).

## Verification

- `npm run build`
- 161 focused backend, lifecycle, runtime, restart, and Rocks-backed source tests pass.
- The complete resource unit gate passes with 2,963 tests and 46 pending.
- Changed-file Oxlint, Prettier, and `git diff --check` pass.
- The [GitHub test matrix](https://github.com/HarperFast/harper/pull/2569/checks) passes across Linux, Windows, supported Node versions, Bun, uWebSockets, Next.js adapters, and the HNSW native plane.
- The rebased commit reached convergence in independent Claude and Harper-domain review; earlier branch rounds also used Cursor reviewers. The generated coverage receipt below records the full review history.

Complexity: complicated — this adds native writer lifecycle, durable replay coordination, bounded backpressure, rollback/replay, rebuild, and ownership handoff.

<sub>Review-Coverage: authored=codex; ran=cursor-grok,claude,cursor-composer; adjudicated=domain; blocked=gemini(auth); rounds=16; full=5 @ 27280e349e5e</sub>

<sub>Human-Review-Need: 4 (decisions: inert-backend-slice, adapter-local-handoff-fencing, retry-by-default-open-policy, wrapper-owned-field-validation, node-local-native-state) @ 27280e349e5e</sub>


kriszyp (commented):
@kylebernhardy Ok, I think I see how I got confused here. This also kind of a response to https://github.com/HarperFast/harper/pull/2568#issuecomment-5637406761, but this architectural commentary, so belongs here.

Basically: I think Tantivy should actually be using its own persistence/storage capabilities, and not trying to push its segments into RocksDB (into a column family). I think I understand your rationale; unifying backup and stuff, but HNSW is already going down the road of owning its own persistence, and I think that is the intent I had for the derived indexes and how they should be interacting with Harper. I think it is perfectly reasonable to say that (external) indexes have to be re-built on database restore. And/or plan to create fancier checkpoint/coordination mechanisms for built-in backups down the road.

WDYT, is that reasonable?

Anyway, here lengthier version in Claudish, if your agent is interested:

Read against the runtime as it stands on #2567 (one backend contract, committed-tail rebuild, plain-word readiness) and the HNSW backend on #2430. The main point is about where the index lives; the rest follows from it.

**1. Tantivy should persist the way the HNSW plane does: on its own, next to the index store, with no RocksDB round trip.** Nothing in the derived-index runtime needs a backend's state in RocksDB. The contract already makes the cursor backend-owned — `getDurableCursor()` returns whatever the backend made durable with its own barrier — and the runtime's only RocksDB footprint is the transaction log it reads plus a one-byte condemnation marker. Restore, copy-db and replica seed rebuild derived state from records + log regardless of where it lived; that is what the rebuild phase exists for, so "one storage engine for backup" buys nothing the rebuild does not already provide, and a CF snapshot taken mid-generation is not a consistent index either.

HNSW keeps its graph in a `.hnsw` mmap file with its own `msync` barrier and its generation in the file header; only the `pk ↔ nodeId` mappings and the cursor sit in the index CF, and the mappings are there only because the plane addresses nodes by numeric id. Tantivy needs *less* of RocksDB than that, not more: the document carries `<tableId>.<recordKey>`, so there is no mapping to keep, and Tantivy already has an atomic barrier (`meta.json` rewrite) and a commit payload — which is exactly what this backend's `committedPayload` is.

The consistent shape, which Tantivy supports natively (`MmapDirectory`) and the runtime already permits:

- `MmapDirectory` at `<index store path>/<name>.tantivy/`; commit = `meta.json` rewrite; the cursor vector is the commit payload, as here. Generation = a directory name or a payload field; rebuild = build a new directory, swap, delete the old.
- `getDurableCursor()` reads the payload from `meta.json`: a read-only open, no writer, no lock, cheap enough to do at backend construction. That is what #2568's slice doc said the plan was — *"the backend is constructed and its asynchronous Fulltext open is awaited before `register()`; `getDurableCursor()` is valid synchronously from the first owner acquisition. This integration requires no change to `derivedIndexRuntime.ts`."*
- Gone with it: `acquire()` and the ~190 runtime lines that serve it (`#acquireBackend`, `#acquisitionFailed`, the `#acquiring` gates in wake/drain/startRebuild/requestRebuild, release waiting on acquisition, the generation-preserving `#discardProgress`), the new readiness reason and attempt budget, `RocksDerivedIndexStorage`, the `KvDirectory` transport and its per-I/O NAPI callbacks, the root-wide `flushSync` barrier stalling unrelated tables, and the derived-index writes that emit root `committed` events and wake the runners themselves (reviewer note 5). #2568's benchmark then measures a path we would not ship; the number that matters becomes Tantivy's own commit cost, which is already known.

The trade is the one `DESIGN.md § Native HNSW plane` already states for HNSW: node-local derived state on disk that backup includes after a barrier or marks rebuild-on-restore. The two backends should make the same statement.

**2. `cursorAtOrAfter` fails the backend closed on a legitimate cursor.** `deliver()` rejects a `through` whose per-log timestamp is numerically lower than the last accepted one ("cursor moved backward" → `DERIVED_INDEX_FAILED` → condemnation). Transaction timestamps are unique per log but **not monotone in physical order** — `TransactionLogStore::writeBatch` only advances `latestTimestamp` when the batch's is greater — so a later physical transaction with a lower timestamp is normal, and the runtime's cursor design deliberately never compares timestamps (exact-start resume, repeat detection by set). The runtime already guarantees `through` is an offered vector delivered in order; the backend should not re-derive order from the numbers. This will condemn a healthy index under ordinary concurrent writes. Independent of point 1.

**3. `recordKey` on every mutation.** Fine and cheap (it is the key the collector already computes), and the collection-time filter on non-string keys matches the scan rule.

**4. Docs.** `docs/fulltext-derived-index-backend.md` here and `docs/fulltext-derived-index-vertical-slice.md` on #2568 describe the PR and its stages. Repo docs carry only durable design (the engine/lifecycle interfaces, the document-id scheme, the ordered command state machine) as a section next to `DESIGN.md § Derived-index runtime`; alternatives, verification and staging belong in the PR body.

**5. Base.** The diff of #2568's branch against `feat/derived-index-native-backend-runtime` removes `integrationTests/.../choose-operation-authz.test.ts` and ~160 lines of `serverUtilities.ts` that exist on both `main` and the runtime branch. That looks like a base mismatch rather than intent — worth a rebase and a look before either PR merges.

Two facts from the runtime work this backend depends on: a committed transaction-log read is a contiguous physical prefix (`commitFinished()` advances `lastCommittedPosition` to the earliest still-uncommitted write), which is why the committed tail is a safe rebuild anchor; and a log that has never written a file reports `oldestSequenceNumber: 0`, so "retains its beginning" is `fileCount === 0 || oldestSequenceNumber === 1`.

— Claude Fable 5.1

resources/FullTextDerivedIndexBackend.ts
```diff
@@ -0,0 +1,796 @@
...
+
+	async #applyRecords(command: ApplyCommand, records: DerivedIndexBatch['records']): Promise<boolean> {
+		let packed: Uint8Array;
+		try {
+			packed = this.#encodeMutationBatch(toFullTextMutationRecords(records));
+			if (!(packed instanceof Uint8Array)) throw new TypeError('Full-text batch encoder must return a Uint8Array');
+		} catch (error) {
+			if (errorCodeInChain(error) === 'E_BATCH_TOO_LARGE' && records.length > 1) {
+				const middle = Math.ceil(records.length / 2);
+				if (!(await this.#applyRecords(command, records.slice(0, middle)))) return false;
```
cb1kenobi, line 413: Partial split failures can wedge shutdown — Severity: P1 (high). Once the first recursive half has successfully called `engine.apply`, a persistent encoding failure in the second half, such as an oversized delete key, escapes without recovery; `#lastAppliedSequence` remains unchanged, so shutdown selects `require-clean` despite staged native mutations and can fail quiescence. Track native dirtiness after every successful apply and force rollback/reopen before propagating a later split error, or pre-encode all shards before applying any of them.

—
Reviewed 1b6c1e82

cb1kenobi (commented):
Split batch handling can leave native mutations staged when a later shard is permanently unencodable. Shutdown then attempts a clean close instead of rollback and can leave the derived index unavailable. Track partial native writes and roll them back before surfacing the encoding failure.

—
Reviewed 1b6c1e82

resources/NativeFullTextDerivedIndexLifecycle.ts
```diff
@@ -0,0 +1,461 @@
...
+
+	async open(_ownerEpoch: bigint): Promise<FullTextDerivedIndexEngine> {
+		await this.#closeStranded();
+		this.#ensureLayout();
+		const selector = this.#readSelector();
+		if (!selector) throw new FullTextGenerationInvalidError('Full-text generation selector is missing');
+		if (selector.sourceIdentity !== this.#sourceIdentity)
+			throw new FullTextGenerationInvalidError('Full-text source generation does not match the selected index');
+		const generationPath = this.#generationPath(selector.generationId);
+		this.#assertDirectory(generationPath, 'selected full-text generation');
```
cb1kenobi, line 105: **Missing selected generation directory fails as transient, so the index goes permanently `unavailable` instead of rebuilding** — Severity: high.

`#assertDirectory()` calls `lstatSync()` directly, so when `CURRENT` selects a generation whose directory is gone, the failure surfaces as a bare `ENOENT` `Error` rather than a `FullTextGenerationInvalidError`.

Failure path: `FullTextDerivedIndexBackend.#open()` only short-circuits on `invalidGenerationError()`, so the ENOENT consumes all `openAttempts` and is rewrapped as `FullTextDerivedIndexError('Full-text generation could not be opened')`. `acquire()` rethrows it (the cause chain contains no `FullTextGenerationInvalidError`), and `DerivedIndexRunner.#acquisitionFailed()` classifies it as transient — retrying with backoff until `maxAcquisitionAttempts`, then publishing `unavailable` / `acquisition-failed`. A restart repeats the identical sequence, so the generation is never condemned, `reset()`/`replace()` is never called, and the index never recovers without manual intervention.

This contradicts `docs/fulltext-derived-index-backend.md` ("a missing selected directory ... condemn the generation and enter Harper's rebuild protocol without consuming transient acquisition retries") and the PR's claim that missing derived state is replaced and rebuilt locally. Note the asymmetry with the line just above: a missing `CURRENT` is already routed correctly through the `exists()` guard in `#readSelector()`, and `#ensureLayout()` guards its `lstatSync` the same way — this call site is the one that does not.

Simplest fix — make the missing case explicit before stat-ing:

```ts
if (!exists(generationPath))
	throw new FullTextGenerationInvalidError('Selected full-text generation directory is missing');
this.#assertDirectory(generationPath, 'selected full-text generation');
```

or have `#assertDirectory()` catch `ENOENT` and rethrow as `FullTextGenerationInvalidError`. Worth a lifecycle test that removes the selected generation directory and asserts `open()` rejects with `FullTextGenerationInvalidError`, alongside the existing source-identity-mismatch case.

—
Reviewed 1b6c1e82

cb1kenobi (commented):
The native full-text lifecycle uses a raw lstatSync to check the selected generation directory, so a missing directory surfaces as ENOENT rather than FullTextGenerationInvalidError. The backend and runtime then treat it as a transient open failure, burn all acquisition retries, and park the index in a permanent unavailable state instead of rebuilding it — the opposite of the self-heal behavior this PR documents. The fix is to detect the missing directory explicitly and throw FullTextGenerationInvalidError so it enters the rebuild path, as a missing CURRENT selector already does. The rest of the state machine, epoch fencing, and cursor publication logic held up under review.

—
Reviewed 1b6c1e82

resources/FullTextDerivedIndexBackend.ts
```diff
@@ -0,0 +1,804 @@
...
+		}
+		if (applied !== expected)
+			throw new FullTextDerivedIndexError(`Full-text engine applied ${applied} of ${expected} mutations`);
+		return true;
+	}
+
+	async #publish(command: BarrierCommand): Promise<void> {
+		if (command.horizon <= this.#lastPublishedSequence) return;
+		this.#assertCommandEpoch(command.epoch);
+		if (this.#lastAppliedSequence < command.horizon)
```
cb1kenobi, line 451: **Rejected apply can be misclassified as clean — Severity: high.** `#hasStagedMutations` is set only after `apply()` resolves, although a rejection may occur after native mutations were staged. If recovery's rollback close also fails, the retained engine reaches shutdown with this flag false; shutdown selects `require-clean`, which cannot discard the dirty writer and can leave the index unavailable with its runner lock held. Set `#hasStagedMutations = true` before awaiting `apply()` (or in the catch before recovery); successful reopen or publication already clears it.

—
Reviewed e25a9fa6

cb1kenobi (commented):
A rejected native apply can leave mutations staged while the backend records the engine as clean. If rollback close also fails, shutdown can leave the index unavailable with its runner lock held. Mark the engine dirty before awaiting apply so shutdown continues using rollback.

—
Reviewed e25a9fa6

resources/derivedIndexRuntime.ts
```diff
@@ -909,19 +937,107 @@ class DerivedIndexRunner {
...
+					this.#fail('failed to handle backend acquisition rejection', handlerError);
+				}
+			}
+		);
+	}
+
+	#acquisitionFailed(generation: number, error: unknown) {
+		if (!this.#live(generation)) return;
+		this.#acquisitionFailures++;
+		if (this.#acquisitionFailures >= this.#options.maxAcquisitionAttempts) {
```
kriszyp, line 1008: The acquisition attempt budget is the one piece of this policy I think is set too tight, and the escalation it reaches is terminal.

Retries here use a flat `rebuildBackoffMilliseconds` (default 1000 ms), so the default `maxAcquisitionAttempts: 8` gives roughly **eight seconds** of tolerance. Rebuild failures, by contrast, get exponential backoff up to `maxRebuildBackoffMilliseconds` (5 min) over the same 8 attempts — minutes of tolerance for a strictly more destructive operation.

Concrete failure: a worker hits EMFILE (or the native open returns `E_STORAGE`, or a peer that just lost the lock is still closing the writer) for ten seconds. Eight acquisitions fail, and this branch publishes `unavailable` / `acquisition-failed` into the shared readiness words. From there:

- this runner's `wake()` returns immediately at `derivedIndexRuntime.ts:614-619` because the shared state reads `unavailable`;
- any other worker that wins the lock lands in `#acquired()` at `derivedIndexRuntime.ts:912-921`, sees the shared `unavailable`, and releases without ever calling `acquire()` — even though its own open would have succeeded.

So a transient, worker-local condition takes the index down for the whole process until an operator calls `requestRebuild` or the process restarts. Two suggestions, either of which would do:

1. use the same escalating backoff as `#scheduleRebuild()` (`derivedIndexRuntime.ts:1777-1778`) for acquisition retries, so 8 attempts span minutes rather than seconds; and/or
2. don't let a *local* acquisition failure publish a shared terminal state — publish `unknown`/`acquisition-failed` and let a contender that can actually open the generation take ownership, reserving `unavailable` for a failure the generation itself carries (which the `FullTextGenerationInvalidError` path already routes to rebuild instead).

The existing test `publishes unavailable after persistent backend acquisition failures` would still hold under (1) with a tuned `rebuildBackoffMilliseconds`.

resources/FullTextDerivedIndexBackend.ts
```diff
@@ -0,0 +1,804 @@
...
+		if (!this.#engine || this.#activeEpoch !== batch.ownerEpoch || this.#shutdown) return DERIVED_INDEX_FAILED;
+		let through: DerivedIndexCursor | undefined;
+		try {
+			through = batch.through && normalizedCursor(batch.through);
+		} catch (error) {
+			this.#markFailed(error);
+			return DERIVED_INDEX_FAILED;
+		}
+		const bytes = Math.max(1, batch.bytes);
+		if (!Number.isSafeInteger(bytes) || bytes > this.#maxQueuedBytes) {
```
kriszyp, line 252: This guard fails the backend permanently for a batch that is merely too big, and the failure is not self-clearing.

`DERIVED_INDEX_FAILED` here sends the runtime to `#needsRebuild(...'backend-failed')`, the rebuild re-scans the same records, hits the same oversized record, and after `maxRebuildAttempts` the index becomes `unavailable`. With the defaults (runtime `maxChunkBytes` 4 MB vs. `maxQueuedBytes` 64 MB) this only triggers for a single record whose size estimate exceeds 64 MB — but that is exactly the case the encoder path a few lines below (`resources/FullTextDerivedIndexBackend.ts:416-426`) already handles gracefully, by turning the record into a removal and calling `host.noteUnindexable(...)`.

Worth aligning the two: when the over-limit batch carries a single record, convert it to a removal and count it unindexable rather than condemning the index; keep the hard failure for the genuinely unserviceable case (a multi-record batch larger than an empty queue, i.e. a misconfiguration of `maxQueuedBytes` relative to `maxChunkBytes`). A one-line comment noting the intended relationship between those two limits would also help whoever configures them at schema activation.

resources/NativeFullTextDerivedIndexLifecycle.ts
```diff
@@ -0,0 +1,470 @@
...
+
+	#writeSelector(selector: GenerationSelector): void {
+		this.#writeControlFile(this.#selectorPath, JSON.stringify(selector));
+	}
+
+	#writeControlFile(filePath: string, content: string): void {
+		const tempPath = `${filePath}.${process.pid}.${randomUUID()}.tmp`;
+		const descriptor = openSync(tempPath, constants.O_CREAT | constants.O_EXCL | constants.O_WRONLY, 0o600);
+		try {
+			writeFileSync(descriptor, content, 'utf8');
```
kriszyp: The temp file is only cleaned up on the rename/fsync path; if `writeFileSync` or `fsyncSync` throws, the `finally` closes the descriptor and the exception propagates with `${filePath}.<pid>.<uuid>.tmp` left on disk.

That is precisely the ENOSPC/EDQUOT case, where leaving debris is worst. And unlike stray generations, these never get reclaimed: `#sweepExcept()` only walks `generations/` and skips non-UUID names, so root-level `*.tmp` files accumulate one per failed activation across restarts.

Suggest wrapping the write+fsync in the same try and `rm`-ing the temp on failure, and — since a SIGKILL between `openSync` and the rename leaves one behind regardless — sweeping `<root>/*.tmp` during `#ensureLayout()`.

resources/FullTextDerivedIndexBackend.ts
```diff
@@ -0,0 +1,804 @@
...
+
+function logWarning(message: string, error: unknown): void {
+	log('warn', message, error);
+}
+
+function logError(message: string, error: unknown): void {
+	log('error', message, error);
+}
+
+function log(level: 'warn' | 'error', message: string, error: unknown): void {
```
kriszyp, line 670: Every warning and error goes through a fresh dynamic `import()` plus two promise hops, and the trailing `.catch(() => undefined)` means a logging failure is indistinguishable from a successful log.

The practical cost is that diagnostics are delivered asynchronously and out of order relative to the failure, and a message emitted during shutdown (`#markFailed` from a failed close, for instance) can be dropped entirely if the process exits before the microtask runs — the cases where you most want the line in the log. `resources/derivedIndexRuntime.ts:6` imports `loggerWithTag` statically for the same subsystem, so there is precedent and no dependency-cycle hazard that I can see from `import type`-only usage of the runtime here.

Either switch to the static import (mirroring the runtime and `NativeFullTextDerivedIndexLifecycle.ts:466-470`), or leave a short comment stating what the lazy load is protecting — right now a reader can't tell whether it is deliberate.

docs/fulltext-derived-index-backend.md
```diff
@@ -0,0 +1,284 @@
...
+`fsync` where the platform supports directory handles. The rename retry is capped at 15 ms so index
+activation cannot park a worker for the general configuration writer's multi-second retry budget.
+
+Before selecting a new generation, the lifecycle syncs the native generation directory and its
+parent. A crash before the selector rename leaves the old generation selected. A crash after the
+rename selects the new cursorless generation, so recovery rebuilds it from Harper. Replacement never
+deletes or modifies the selected generation first. Cleanup accepts only strict UUID child paths and
+runs after selection; the next open also sweeps non-selected generations, bounding crash leftovers.
+
+Missing `CURRENT` is a rebuild signal; `open()` stays read-only and the existing condemn/reset path
```
kriszyp, line 134: Two doc points, both small:

1. "`open()` stays read-only" isn't quite what the code does — `open()` calls `#ensureLayout()` (`resources/NativeFullTextDerivedIndexLifecycle.ts:97-98`), which `mkdirSync`s the index root and `generations/` and writes `STORE.json` when it is absent (`resources/NativeFullTextDerivedIndexLifecycle.ts:218-226`). It doesn't create or select a generation, which is the invariant that matters, so it'd read more accurately as something like "`open()` never creates or selects a generation" — the directory/metadata creation is worth stating since it means a read-only filesystem fails here rather than later.

2. Carrying over the doc-scope point from the last round: this file still holds PR-scoped material — the staging note in the Objective, `## Verification` (line 250), and `## Approaches considered` (line 275). Those belong in the PR body; what the repo wants durably is the engine/lifecycle interfaces, the document-id scheme, and the ordered command state machine, ideally as a section next to `DESIGN.md § Derived-index runtime` (the `resources/DESIGN.md` rows added here already point at this file, so the link survives either way).

kylebernhardy:
Thanks for the careful review. All three cases are now covered:

- Partial split-batch work is tracked so shutdown rolls it back.
- A missing selected generation is classified as rebuildable rather than exhausting acquisition retries.
- The engine is marked dirty before awaiting native apply, so a rejected apply followed by a failed recovery close is retried in rollback mode during shutdown.

Each path has a regression test, including a mechanical check that the rejected-apply test fails with the old dirty-marker ordering.

cb1kenobi (commented):
No new blocking defect was confirmed at d191a5ca. The latest commit fixes rollback-state tracking, and the earlier blocking findings are addressed.

—
Reviewed d191a5ca

kriszyp (changes requested):
Sorry for the long comment, but wanted to take a review looking at how we can keep this as simple as possible, so this is here for consideration:
Re-read at `e25a9fa6e`. The move to native Tantivy storage (`9f8cb64fe`, `NativeFullTextDerivedIndexLifecycle`) and dropping the numeric `cursorAtOrAfter` check both resolve what I raised on the 11th — thank you. What follows is about size: the PR is +4.2k lines for a backend whose contract with the runtime is seven methods, and Chris's three high-severity findings today all sit inside machinery that can come out.

**1. `acquire()` is still here, and native storage removes its reason.** The doc's "Do less" row now justifies it as: *the same registration reacquires after every idle release, so a synchronous cursor read would force a rebuild after the first close.* That is true of this backend as written — `shutdown()` closes the engine **and** clears `#durableCursor`, so the only way to learn the cursor again is to reopen the writer, which is async. It is not a property of the problem. Two ways to keep `getDurableCursor()` synchronous with no runtime change:

- **Read the committed payload from the selected generation without a writer.** The cursor is Tantivy's own commit payload in that generation's `meta.json`; a synchronous read (file, or a sync native "committed payload" call) returns it. The writer opens lazily on the first `deliver()`, inside the backend, where an open failure is already `failed` / `accepted-work-lost` with its own retry.
- **Or keep the last published cursor across `shutdown()`.** It is by definition what was made durable. On reopen, validate it against the generation's payload; mismatch → `accepted-work-lost` (replay from the on-disk value) or `failed` (rebuild).

The "Cursor outside Fulltext" rejection in the alternatives table is about a *second copy in RocksDB* needing reconciliation; neither option above is that — both read Tantivy's own committed value. Removing the hook takes with it the 219-line runtime diff (`#acquireBackend`, `#acquisitionFailed`, `#acquiring` gates in wake/drain/startRebuild/requestRebuild, release waiting on acquisition, the generation-preserving `#discardProgress`), `maxAcquisitionAttempts`, the `acquisition-failed` readiness reason, and the "never-settling acquisition holds the lock" rule. Chris's "missing generation directory burns all acquisition retries into permanent `unavailable`" finding is in exactly this path.

**2. The recursive `E_BATCH_TOO_LARGE` split re-implements a bound the runtime already enforces.** `#applyRecords` halves the record list on encoder overflow and special-cases a single oversized record into a removal + `host.noteUnindexable()` (a new host method). The runtime already bounds every chunk by `maxChunkRecords` and `maxChunkBytes` (estimated from stored record sizes) — register with `maxChunkBytes` at or below the native batch cap and the split never triggers. A single record too large to index belongs to the **projection**: have it reject fields above the encoder limit with a `ClientError`, and the runtime delivers `unindexable` → the backend removes, exactly as HNSW does for a wrong-dimension vector. That removes the recursion, the removal re-encode, `noteUnindexable` from the host contract, and Chris's "partial split leaves staged mutations" finding.

**3. `#recoverAcceptedWork` is a second recovery path with its own state.** Rollback-close, reopen in place, revalidate the payload, then `accepted-work-lost` — with `#recovering`, `#lossPendingEpoch`, and the shutdown interplay that produced Chris's "rejected apply misclassified as clean" finding. With the lazy writer open from (1), recovery collapses to: rollback-close, discard the queue, report `accepted-work-lost`. The reopen happens on the next `deliver()`, through the one open path.

**4. The generation lifecycle carries two atomic commit points of its own on top of Tantivy's.** `CURRENT` selector (JSON, UUID-validated, path-escape-checked), `STORE.json` metadata (storeName check, repair mode), `generations/` sweep, per-directory `fsync`, and `renameWithRetry`. Tantivy's `meta.json` rewrite is already the atomic commit; the directory *is* the generation. Simpler: one `<name>.tantivy/`; a rebuild writes `<name>.tantivy.next/` and renames it over (atomic on POSIX; on Windows use the plane's `.stale` tombstone rather than retrying), so the selector, the metadata file and the sweep go — roughly 200 of the 470 lines. `renameWithRetry` also sleeps with `Atomics.wait` on the event loop (up to 12 × 500 ms); if it stays, it should be async.

**5. Cursor payload validation duplicates the runtime's.** `normalizedCursor`, `RESERVED_LOG_NAMES`, sorted-key re-encoding, `maxCursorPayloadBytes` — the runtime validates every cursor it receives from `getDurableCursor()` (`isValidCursor`) and rebuilds on an invalid one. `JSON.parse` and hand it over.

**6. Docs.** `docs/fulltext-derived-index-backend.md` (284 lines) describes the PR and its decisions; the durable part — engine/lifecycle interfaces, document-id scheme, ordered command state machine — belongs as a section next to `DESIGN.md § Derived-index runtime`, and the rest in this description (same rule applied to #2567 and #2430).

**7. Base.** One commit behind `main` (#2191, the synchronous config read); the diff shows it as a deletion of `configReadRetry.ts` / `OptionsWatcher` work. A rebase clears it and also picks up #2574 once merged.

Taken together, (1)–(5) leave the backend as: encode `batch.records`, `apply`, `publish` with the cursor as payload, lazy open, rollback-close-and-report on failure — and leave `derivedIndexRuntime.ts` untouched except `recordKey`, which is the one addition every document-keyed backend wants.

cb1kenobi (commented):
No new blocking defects on changed lines at d191a5ca. The three previously reported issues are genuinely fixed: staged-mutation tracking now precedes the native apply so shutdown rolls back, a missing generation directory throws FullTextGenerationInvalidError and enters rebuild, and control-file temp files are cleaned up on write failure. Two non-line items remain: Unit Test (Node.js v24) is failing while v22/v26 pass (the new crash-matrix test spawns child processes on a 5s deadline and is the likely flake), and kriszyp's architectural CHANGES_REQUESTED about the PR's size is still unresolved.

—
Reviewed d191a5ca

kylebernhardy:
Kris, thanks for pushing on the size here. I rechecked each proposed removal against current Fulltext `5c26daf` and the Harper runtime, then had Claude and the actual Gemini CLI independently review the current head. Both returned LGTM, and I don't think points 1–5 can be removed safely under the current contracts:

1. Fulltext acquires Tantivy's exclusive writer before it reads the committed payload because another writer may publish until that lock is held. Harper needs the durable cursor before it chooses where transaction-log replay starts, so lazy open during `deliver()` is too late. Retaining the cursor in memory also does not cover cold restart or crash recovery.
2. `maxChunkBytes` estimates source-record bytes. It cannot bound the projected and natively encoded batch, and it cannot isolate a single record over Fulltext's exact encoder limit. The split runs only after `E_BATCH_TOO_LARGE`; the single-record case removes any stale indexed document.
3. `accepted-work-lost` causes the runtime to read the durable cursor immediately. Reopening before that notification proves Tantivy rolled back staged work and gives the runtime the actual on-disk cursor.
4. Tantivy's `meta.json` commit is atomic within one generation. It does not atomically replace a populated directory, and rename-over-directory is not portable across POSIX and Windows. `CURRENT` is the whole-generation publication point; `STORE.json` prevents a hashed path from being accepted for the wrong logical store.
5. The backend validation is the on-disk format boundary: payload size, envelope version, allowed keys, and reserved log names. The runtime's validator is intentionally generic and does not enforce that storage format.

The branch is now fully current with `main` (`0` commits behind), and the Node 22/24/26 and Windows jobs pass. More than half of the added lines are tests; the standalone document records the durable engine, generation, and recovery contracts rather than PR staging history.

I updated the PR description to put these tradeoffs and the human decision point up front. Would you take another look and let me know if the current Fulltext writer/payload contract changes your recommendation?

resources/FullTextDerivedIndexBackend.ts
```diff
@@ -0,0 +1,805 @@
...
+
+export function toFullTextMutationBatch(batch: DerivedIndexBatch): FullTextMutationBatch {
+	return toFullTextMutationRecords(batch.records);
+}
+
+function toFullTextMutationRecords(records: DerivedIndexBatch['records']): FullTextMutationBatch {
+	const upserts: FullTextMutationBatch['upserts'] = [];
+	const deletes: string[] = [];
+	for (const record of records) {
+		const id = `${record.tableId}.${Buffer.from(record.recordKey, 'latin1').toString('base64url')}`;
```
cb1kenobi, line 634: **High: `latin1` document ids collide for non-Latin-1 primary keys**

`writeKeyId()` returns string primary keys unchanged — `unitTests/resources/derivedIndexRuntimeRocks.test.js` asserts the document id equals `Buffer.from('p1').toString('base64url')`, so `recordKey` is the raw key string. `Buffer.from(str, 'latin1')` then truncates each code unit to its low byte, so within a single table the keys `'日'` (U+65E5) and `'å'` (U+00E5) both encode to byte `0xE5` and share one Fulltext document id: indexing one silently replaces the other's text, and deleting one removes the other's document. This id scheme is written into persisted Tantivy generations, so correcting it after release forces a full index rebuild.

Suggested fix — use an injective encoding:

```suggestion
		const id = `${record.tableId}.${Buffer.from(record.recordKey, 'utf16le').toString('base64url')}`;
```

`utf16le` is lossless for every JS string, including lone surrogates; `utf8` also works if lone surrogates can be ruled out. `unitTests/resources/fullTextDerivedIndexBackend.test.js` and `unitTests/resources/derivedIndexRuntimeRocks.test.js` both hard-code the current encoding with ASCII-only keys, so neither would catch this — worth adding a case that asserts `'日'` and `'å'` produce distinct ids.

—
Reviewed a71635f0

cb1kenobi (commented):
The full-text document id is built by encoding the record key with 'latin1', which truncates every character above U+00FF to a single byte. Since string primary keys pass through writeKeyId() unchanged, two different keys in the same table (for example '日' and 'å') collapse to the same document id, so one record's text silently overwrites the other's and deleting one removes the other. Switching that one line to a lossless encoding such as 'utf16le' fixes it, and a test with a non-ASCII key should be added. Everything else in this head, including the three previously reported issues and the rename-retry extraction, checked out.

—
Reviewed a71635f0

cb1kenobi (commented):
No new blocking defect was confirmed on changed lines. Existing discussion already covers the outstanding blocker, so no duplicate comment is returned.

—
Reviewed a71635f0

kylebernhardy:
Chris, thanks for flagging this. I traced the value into the derived runtime: `recordKey` is always produced by `writeKeyId()`, which first encodes the Harper key with `toBufferKey()` and then uses Latin-1 as a lossless byte carrier. `Buffer.from(recordKey, 'latin1')` recovers those encoded bytes; string primary keys do not pass through unchanged.

I added a regression using `日` and `å` to prove that both the canonical Harper keys and resulting Fulltext document IDs remain distinct. The focused backend suite passes all 38 tests at `ae0a6c5f1`.

cb1kenobi (commented):
The new real-RocksDB integration test passes its 10-second timeout as a bare number where waitFor expects an options object, so it silently falls back to the much shorter default and can fail on a loaded CI runner for timing rather than a real regression. Its wait condition is also exact-match on a publication count, which can be missed between polls. Fixing both is a two-token change. No new blocking defects were confirmed in the product code at this commit; the previously reported blockers are genuinely fixed, and the outstanding architectural and acquisition-budget threads remain open from earlier reviews.

—
Reviewed ae0a6c5f

unitTests/resources/derivedIndexRuntimeRocks.test.js
```diff
@@ -116,9 +134,216 @@ describe('DerivedIndexRuntime with an audited RocksDB table', () => {
...
+			options: { flushAfterMutations: 1, maxFlushAgeMilliseconds: 10 },
+		});
+		try {
+			await Product.put('p1', { title: 'first product' });
+			const p1Cursor = [...Product.auditStore.getRange({ start: anchor })]
+				.filter((entry) => entry.tableId === Product.tableId && entry.recordId === 'p1')
+				.at(-1).txnLogKey;
+			await waitFor(
+				() => binding.states.size > 0 && [...binding.states.values()].some((state) => state.publications === 1),
+				10_000
```
cb1kenobi, line 202: **Test timeout is silently ~5x shorter than intended** — Severity: medium (test-only, CI flake risk).

`waitFor` takes an options object everywhere else in this file — `{ timeout: 10_000 }` in the rebuild test added just below, and `{ timeout: 2000 }` / `{ timeout: 5000 }` in the existing suites. Passing the bare number `10_000` destructures to nothing, so the helper falls back to its default timeout and the intended 10-second budget is never applied.

Failure mode: this is the slowest new test in the PR — it drives a real RocksDB audit log through registration, an invalid-generation rebuild, `replace()`, a scan, and a native publish on a 10 ms flush cadence. On a loaded CI runner that sequence can exceed the default window and fail for timing rather than for a real regression, which is exactly the kind of noise the currently-red unit job makes expensive to triage.

The predicate compounds it: `some((state) => state.publications === 1)` is true only during the window in which exactly one publication has landed. If a second cursor-only publication lands between polls, the condition can never become true again and the test waits out the (already shortened) timeout.

Simplest fix: pass `{ timeout: 10_000 }` instead of `10_000`, and relax the predicate to `state.publications >= 1`.

—
Reviewed ae0a6c5f

resources/FullTextDerivedIndexBackend.ts
```diff
@@ -0,0 +1,805 @@
...
+		}
+		setImmediate(() => {
+			if (this.#activeEpoch !== activeEpoch) return;
+			if (this.#wake !== wake || (change === 'accepted-work-lost' && this.#shutdown)) {
+				if (change === 'accepted-work-lost' && this.#lossPendingEpoch === activeEpoch)
+					this.#lossPendingEpoch = undefined;
+				return;
+			}
+			try {
+				wake(change);
```
cb1kenobi, line 615: **Recovery notification can stall replay indefinitely — Severity: high.** `#lossPendingEpoch` is still set while `wake('accepted-work-lost')` runs. The runtime handles that callback synchronously by resetting to the recovered cursor and immediately draining, so its replay delivery returns `DERIVED_INDEX_DEFERRED`; the subsequent fence clear emits no `changed` notification. With no later source commit, the index remains stalled behind its recovered cursor. Clear `#lossPendingEpoch` immediately before invoking the accepted-work-lost callback, or schedule a `changed` notification immediately after clearing it.

—
Reviewed ae0a6c5f

cb1kenobi (commented):
Recoverable native failures can leave replay indefinitely deferred because the delivery fence remains set during the runtime's recovery callback. Clear the fence before invoking the callback, or wake the runtime again after clearing it.

—
Reviewed ae0a6c5f

kylebernhardy:
Kris, I agree that the current implementation puts too much Fulltext lifecycle machinery into Harper and changes the shared runtime more than this backend warrants. I moved the PR back to draft and am reworking it around your proposed shape:

- remove the runtime acquisition hook and its retry/readiness state;
- keep only the generic document key addition in the shared runtime;
- add synchronous read-only cursor inspection and coordinated reset to the Fulltext wrapper;
- lazy-open the Tantivy writer on first delivery;
- collapse failure recovery to rollback, discard, and replay through the same open path;
- remove recursive batch splitting and the new host hook; and
- reduce the repository documentation to durable architecture.

For generation replacement, I plan to move ownership into the wrapper but use rename-to-a-unique-stale-sibling rather than rename-over a non-empty directory, which is not portable. Harper’s persisted condemnation marker remains the crash-recovery guard.

I will post the reduced diff after the wrapper prerequisite and recovery tests are in place.

kriszyp (commented):
Ok, if I am pushing too hard on simplification over performance, feel free to push back; especially if you have numbers to push back with :). I think the acquire() mechanism is related to async vs sync setup, but that decision can be heavily influenced by how expensive and frequent the sync activities are.
<sub>🤖 Reviewed with Gemini</sub>

resources/derivedIndexRuntime.ts
```diff
@@ -912,6 +924,13 @@ class DerivedIndexRunner {
 			this.#resetFromDurableCursor();
 			if (this.#owned && !this.#rebuilding) this.#drain();
 		} catch (error) {
+			if (error instanceof DerivedIndexBackendRetryError) {
```
kriszyp, line 1035: This retry branch has no terminal state, and it replaced one that did.

A `DerivedIndexBackendRetryError` is raised by `FullTextDerivedIndexBackend.getDurableCursor()` for *any* throw out of `lifecycle.inspect()` — including permanently broken ones: a lifecycle whose binding was never initialized (`#requireBinding()` throws a plain `Error`), EACCES on the index directory, or a native module that throws on every inspection. Concretely: make `inspectNativeFullTextIndex` throw unconditionally and the runner loops `own → inspect → release → armLockRetry(5s)` forever, logging a warn every five seconds, while `getReadiness()` keeps reporting whatever the previous owner published.

Two consequences the removed `maxAcquisitionAttempts` path used to cover (both had dedicated tests in `derivedIndexRuntimeNativeBackend.test.js`, deleted in this delta):

1. Readiness never reaches `unavailable`, so nothing observable distinguishes "no owner can read this index" from "transiently between owners". `#acquisitionFailed` capped this and published `unavailable`.
2. This branch does not call `#admitWrites()`. If the lag policy is enabled and a previous owner set `READINESS_LAG_EXCEEDED`, source-table writes stay rejected indefinitely because no one clears the flag — the deleted test `admits writes while no owner can acquire a lagging backend` pinned exactly this. Every other release-without-progress path here (`resources/derivedIndexRuntime.ts:1631-1632`, `:1540-1541`, `:902-903`) calls `#admitWrites()` first.

Suggest: add `this.#admitWrites();` to this branch unconditionally, and count consecutive retry-classified inspection failures so that after a bounded number the runner publishes `unavailable` with a reason, as the acquisition path did. Setting `this.status = { state: 'idle' }` here also drops `ownerEpoch`, unlike the other idle transitions.

resources/FullTextDerivedIndexBackend.ts
```diff
@@ -0,0 +1,757 @@
...
+		if (command.batch.records.length === 0) {
+			this.#lastAppliedSequence = command.sequence;
+			return true;
+		}
+		let packed: Uint8Array;
+		try {
+			packed = this.#encodeMutationBatch(toFullTextMutationRecords(command.batch.records));
+			if (!(packed instanceof Uint8Array)) throw new TypeError('Full-text batch encoder must return a Uint8Array');
+		} catch (error) {
+			throw new FullTextDerivedIndexError('Failed to encode a full-text mutation batch', error);
```
kriszyp: Every encoder failure is now permanent, which turns a single unpackable batch into an index-wide `unavailable` and eight full table rescans.

This throw propagates out of `#apply` to the drain's catch → `#failAndNotify('failed')` → the runtime's `#needsRebuild('backend reported a permanent failure')`. The rebuild rescans the table and re-delivers the same records, hits the same encoder rejection, and repeats until `maxRebuildAttempts` (default 8) is spent and the index becomes `unavailable` — eight full scans of the source table to arrive there.

The delta removed two behaviours that kept this bounded:
- a multi-record batch rejected with `E_BATCH_TOO_LARGE` was split in half and retried in order, so a batch whose *packed* size exceeded `maxBatchBytes` (something the runtime's byte-based chunking cannot predict, since it measures Harper batch bytes, not Tantivy packed bytes) simply succeeded in two pieces;
- a single record too large to pack became a removal plus a `noteUnindexable` count, so one bad document could not poison the index.

The design doc states the new behaviour deliberately ("fails the backend rather than being split or silently converted into a deletion"), so if that is the intended trade I am not asking for the splitter back. But the failure mode deserves to be cheaper than it currently is. Two options that keep the simplification:
- classify `E_BATCH_TOO_LARGE` (and `E_INVALID_ARGUMENT`) so the runtime goes straight to `unavailable` instead of spending the rebuild budget re-deriving the same rejection, since a rebuild provably cannot fix it; and/or
- enforce a per-field/per-document size cap at the projection boundary during schema activation, which is where the doc says unindexable records should be counted — note that `noteUnindexable` was removed from `DerivedIndexBackendHost` in this delta, so the backend can no longer report one even if it wanted to.

kriszyp (commented):
This is looking great! Just noting the progress (I see it is still in draft).

kriszyp (commented):
This is looking good, good progress. I will note that this is still in draft and conflicted, so I assume you are still working on it.
<sub>🤖 Reviewed with Gemini</sub>

resources/indexes/fullTextDerivedIndex.ts
```diff
@@ -0,0 +1,1143 @@
...
+					await delay(this.#openRetryMilliseconds);
+					continue;
+				}
+			}
+		}
+		this.#scheduleOpenRetry(ownerEpoch, lastError);
+	}
+
+	#scheduleOpenRetry(ownerEpoch: bigint, error: unknown): void {
+		if (this.#openRetryTimer || this.#shutdown || this.#failed) return;
```
kriszyp: The `this.#shutdown` term in this guard turns a shutdown that overlaps a failing writer open into a hot retry loop that burns the entire handoff bound.

Trace: `shutdown()` queues a final barrier and schedules a drain. `#ensureEngine()` calls `#open()`, which exhausts its immediate attempt budget and calls `#scheduleOpenRetry()`. Because a shutdown is pending, this line returns without arming a timer and without growing `#openRetryDelayMilliseconds`. `#ensureEngine()` returns false, `#drain()` returns, and its `finally` (`resources/indexes/fullTextDerivedIndex.ts:483-485`) sees `!#failed && !#openRetryTimer && !#writerRetryTimer && #commands.length > 0`, so it takes the `#scheduleDrain()` branch instead of `#closeForShutdown()`. The next turn repeats. With the defaults that is roughly three native `open()` calls every ~20 ms, with no backoff, for the full 70 s `shutdownTimeoutMilliseconds`, after which `#boundedShutdown` rejects.

Concrete scenario: a worker stops (or a rebuild calls `#quiesce`) while `lifecycle.open()` keeps rejecting with a retryable code such as `E_LOCK_BUSY` or `E_STORAGE` and no writer has been installed yet for this epoch. Note the epoch stays current through a release — `isOwnerEpoch` compares against the shared epoch word, which is only re-minted on a fresh acquisition — so `#assertSharedEpoch` does not break the loop. The result is a 70 s stall of worker shutdown followed by `hold()` in `derivedIndexRuntime.ts`: runner lock retained, readiness `unavailable`/`shutdown-failed` node-wide, recoverable only by `requestRebuild()` or a process restart.

The painful part is that this is the one case where quiescence is trivially provable: no writer was ever opened, nothing was applied natively, and `#closeForShutdown` would resolve immediately (`engine` is undefined). The accepted-but-unapplied work is already safe to drop — the runtime replays it from the durable cursor on the next acquisition, which is exactly what `#loseAcceptedWork` relies on elsewhere.

Suggested fix, narrow enough not to disturb the bounded-slice behavior: in `#ensureEngine()`, when `#open()` returns undefined and `this.#shutdown` is pending, discard the queued commands and rewind accepted work before returning false. `#commands.length` is then 0, the drain's `finally` falls through to `#closeForShutdown()`, and the handoff completes promptly instead of at the bound. Please avoid flipping the order of the two branches in that `finally` — the `#scheduleDrain()` branch is what continues a partially applied batch, and the 'drains every bounded slice before completing shutdown' test depends on it.

resources/indexes/fullTextDerivedIndex.ts
```diff
@@ -0,0 +1,1143 @@
...
+		return DERIVED_INDEX_ACCEPTED;
+	}
+
+	flush(_reason: DerivedIndexFlushReason = 'threshold'): void {
+		if (this.#activeEpoch === undefined || this.#failed || this.#shutdown) return;
+		this.#queueBarrier(this.#activeEpoch);
+	}
+
+	shutdown(ownerEpoch: bigint): Promise<void> {
+		this.#cursorInspectedForAcquisition = false;
```
kriszyp: This unconditional clear undoes the cursorless caching that `reset()` deliberately installs, so the invariant stated at `resources/indexes/fullTextDerivedIndex.ts:385` ("An unknown reset outcome must never rediscover and trust the pre-reset checkpoint") does not actually hold across a `shutdown()`.

Path: `reset(e)` sets `#durableCursor = undefined` and `#cursorInspectedForAcquisition = true`. The reset rejects or hits `#boundedReset`'s timeout, so the outcome is unknown. The runner then calls `shutdown(e)` — this line sets the flag back to `false` before the `#resetOperation` branch below is even considered. On the next acquisition, `getDurableCursor()` re-inspects native state and, if the wrapper had not yet retired the live generation, adopts the pre-reset checkpoint and resumes replay from it. The design note in `docs/design/fulltext-derived-index-main-rebase.md` describes the opposite behavior.

In the supported case this is still safe, because the shared protocol requires the wrapper's first durable reset action to invalidate the cursor so an interrupted reset inspects as cursorless (`resources/derivedIndexRuntime.ts:149-153`). So this is defense-in-depth that silently stops applying, not a live corruption path — which is why I am calling it minor. But the existing coverage does not catch it: 'clears its cached cursor before a native reset that fails' shuts down *before* the reset, never after.

Suggest moving the clear below the `#resetOperation` branch (or skipping it when a reset for this epoch has not settled), and adding a case that calls `shutdown(e)` after a failed `reset(e)` and asserts `getDurableCursor()` is still `undefined` without a second `inspect()`.

resources/indexes/fullTextDerivedIndex.ts
```diff
@@ -0,0 +1,1143 @@
...
+				logError('Full-text derived index state notification failed', error);
+			} finally {
+				if (change === 'accepted-work-lost' && this.#lossPendingEpoch === activeEpoch)
+					this.#lossPendingEpoch = undefined;
+			}
+		});
+	}
+}
+
+export function toFullTextMutationBatch(batch: DerivedIndexBatch): FullTextMutationBatch {
```
kriszyp: Nit: `toFullTextMutationBatch` has no production caller — `#apply` goes straight to `toFullTextMutationSlice`, and the only references are in `unitTests/resources/fullTextDerivedIndex.test.js`. It is exported API on a module that activation, schema and query slices will all import, so it will read as supported.

Since the tests genuinely want the whole-batch conversion, the cheapest options are to export `toFullTextMutationSlice` instead and let the tests pass `records.length`/`Infinity` explicitly, or to keep this wrapper but mark it as test-only in a comment so a later reader does not build on it. Either is fine; I would rather not see a second conversion entry point accumulate callers.

resources/indexes/fullTextDerivedIndex.ts
```diff
@@ -0,0 +1,1151 @@
...
+				const mode =
+					this.#hasStagedMutations || this.#lastAppliedSequence > this.#lastPublishedSequence
+						? 'rollback'
+						: 'require-clean';
+				await this.#closeEngine(engine, { mode });
+			}
+			this.#engine = undefined;
+			this.#activeEpoch = undefined;
+			this.#cursorInspectedForAcquisition = false;
+			if (this.#shutdownFailure) {
```
kriszyp: `#failed` is only cleared here when the failure came from shutdown itself (`#shutdownFailure`), or later by `reset()`. A failure from a terminal open code, a protocol violation, or a thrown drain error stays set on this backend object after a successful shutdown. The native files are shared across workers, though, so the flag can outlive the state it described.

Concrete sequence:
1. Worker A hits a terminal open error on each attempt and exhausts the rebuild budget. The runtime calls `#becomeUnavailable`, which releases, and A's backend keeps `#failed = true`.
2. The operator fixes the cause and calls `requestRebuild`.
3. Worker B acquires, resets through its own backend object, rebuilds, and publishes `ready`.
4. Later A acquires. `getDurableCursor()` returns B's valid checkpoint, and the first `deliver()` returns `DERIVED_INDEX_FAILED` at line 275.
5. The runtime condemns and resets a healthy generation, so full-text is unavailable for a second full rebuild.

Since every shutdown already invalidates the cursor cache so that each acquisition re-derives state from disk, could the successful-close branch clear `#failed` unconditionally? If the files are still bad, the next acquisition's inspection or lazy open will rediscover that. A shutdown that cannot close still leaves `#failed` set through the catch branch.

resources/indexes/fullTextDerivedIndex.ts
```diff
@@ -0,0 +1,1151 @@
...
+	log('warn', message, error);
+}
+
+function logError(message: string, error?: unknown): void {
+	log('error', message, error);
+}
+
+function log(level: 'warn' | 'error', message: string, error?: unknown): void {
+	const code = nativeErrorCode(error);
+	const detail = error instanceof FullTextDerivedIndexError ? error.message : code;
```
kriszyp, line 950: For an error that is neither a `FullTextDerivedIndexError` nor code-bearing, this logs only the class name. For example, `#failAndNotify` from the drain `catch` would emit `Full-text derived index backend failed (TypeError)`. That failure condemns the index and triggers a full rebuild, and the operator gets no message and no stack to tell an adapter bug from a bad record id from a projector problem.

I understand the intent is to keep native paths and record content out of the logs. The shared runtime handles the same concern differently. It keeps the message in the local log and status and sends only the code to shared memory (`resources/derivedIndexRuntime.ts:1726-1733`). Could this follow that split? One option is to keep native messages suppressed for errors that carry a native `code`, but log message and stack for plain JS errors, or log them at `debug`/`trace`. As written, the first production incident here will be very hard to diagnose.

resources/indexes/fullTextDerivedIndex.ts
```diff
@@ -0,0 +1,1151 @@
...
+		}
+		this.#rewindAcceptedWork();
+		this.#assertCommandEpoch(ownerEpoch);
+		this.#hasStagedMutations = false;
+		this.#stagedUnindexableRecords = 0;
+		if (options.notify === false) {
+			this.#lossPendingEpoch = undefined;
+			return;
+		}
+		if (options.retry !== false) this.#scheduleWriterRetry(ownerEpoch, cause);
```
kriszyp, line 667: Outside a rebuild this works as documented: rollback, `accepted-work-lost`, backoff, replay from the last payload. During a rebuild the runtime treats `accepted-work-lost` as a failed rebuild attempt (`resources/derivedIndexRuntime.ts:1704-1708`). It then quiesces, mints a new epoch, and calls `reset()`. So one transient `E_STORAGE` or `E_QUEUE_FULL` from apply or publish partway through a large table discards the whole rebuild, retires the generation, and spends one of the eight attempts. The writer-retry timer armed here is cleared by the following shutdown before it fires.

That makes the DESIGN.md statement that an ordinary apply/publish failure "does not condemn a structurally valid generation" true only for replay. This does not need to block an inert slice. Please either state the rebuild-phase behaviour in the DESIGN.md section, or add it to the activation open items next to the non-condemning stalled state. Large rebuilds are exactly where transient native errors are most likely.

resources/indexes/nativeFullTextDerivedIndexLifecycle.ts
```diff
@@ -0,0 +1,180 @@
...
+	await lifecycle.initialize({ reclaimRetired: false });
+	if (maxCursorPayloadBytes > lifecycle.maxCommitPayloadBytes)
+		throw new RangeError('Full-text maxCursorPayloadBytes exceeds the native commit payload capacity');
+	const backend = new FullTextDerivedIndexBackend({
+		...options,
+		id: options.id,
+		lifecycle,
+		maxCursorPayloadBytes,
+	});
+	lifecycle.startRetiredStorageReclamation();
```
kriszyp, line 167: Question rather than finding: `createNativeFullTextDerivedIndexBackend` runs in every worker, so every worker queues a reclamation sweep at construction regardless of who owns the runner lock. The per-lifecycle promise chain serializes only within one worker. A non-owner's sweep can overlap the owner's `reset()` plus targeted reclaim, or another worker's sweep of the same retired directory.

Does the wrapper's `reclaimRetiredNativeFullTextIndexes` tolerate concurrent callers on one path? It would need to treat an already-removed entry as success, and it must never touch a directory another thread is in the middle of retiring. If so, a one-line comment saying so would help. If not, the construction-time sweep probably belongs behind ownership, for example on the first acquisition's `getDurableCursor()`. Otherwise the `failed > 0` warning will fire spuriously on multi-worker startup.

docs/design/fulltext-derived-index-main-rebase.md
```diff
@@ -0,0 +1,404 @@
...
+   unproven state observable instead of hanging worker shutdown indefinitely.
+
+### Review-response investigation
+
+The final review exposed three lifecycle cases that the earlier tests did not model. The owning
+invariant is: each elected owner must discover the current native checkpoint before replay, temporary
+writer unavailability must not destroy a valid derived index, and an ownership handoff must settle
+within a bound without unlocking an unproven writer.
+
+- **Current behavior:** at Harper branch commit `28f970d25`, the two focused tests named “retains its
```
kriszyp: Nit: this note cites branch commit `28f970d25` as its fails-on-base evidence, but the branch history has been rewritten, so that SHA will not resolve for a future reader. The "Review-response investigation" section and the "remains the recovery reference" prerequisite at line 380 are also PR-process narrative rather than design. The filename `…-main-rebase` has the same problem.

Since root `DESIGN.md` is declared authoritative, consider trimming this file to the rejected alternatives, the verification record, and the activation gates. The investigation narrative could move to the PR description, where it will not go stale in the tree.

kriszyp (changes requested):
Awesome, this looks great!
There are a few minor findings to consider and can you remove the PR doc, that doesn't belong in this repo.

cb1kenobi (commented):
No new blocking defects were confirmed on changed lines. Existing discussion already covers the outstanding issues, so no duplicate comments are included.

—
Reviewed 27280e34

resources/indexes/fullTextDerivedIndex.ts
kylebernhardy: Addressed in 33b7a64ac. DESIGN.md now distinguishes replay from rebuild-attempt consumption: failure during rebuild condemns that attempt, and the next ownership cycle starts a fresh rebuild. It also documents the deliberate fail-closed inspection and terminal cursor-park availability tradeoffs.

resources/indexes/fullTextDerivedIndex.ts
kylebernhardy: Addressed in 3f75c2077. A clean owner shutdown now clears only worker-local failure state after native close and quiescence succeed. The new regression proves a successor owner re-inspects shared native state and resumes delivery.

resources/indexes/nativeFullTextDerivedIndexLifecycle.ts
kylebernhardy: Verified against the wrapper implementation. Fulltext retires a generation with an atomic rename to a unique sibling and sweeps only generated paths with idempotent force removal. Concurrent worker sweeps are safe, Harper keeps reclamation off the ownership critical path, and the invariant is now documented in the lifecycle adapter.

resources/indexes/fullTextDerivedIndex.ts
kylebernhardy: Addressed in 33b7a64ac. Native-boundary warnings and errors now log only stable Harper text plus an optional native error code; raw native errors, paths, and record content are never passed to the logger. Harper-origin JavaScript errors retain their message and stack for local diagnosis.

docs/design/fulltext-derived-index-main-rebase.md
kylebernhardy: Addressed in 3f75c2077. The PR-specific implementation note was deleted. The durable contract remains in root DESIGN.md, while verification and review history stay in the PR description.

DESIGN.md
```diff
@@ -2699,6 +2699,85 @@ cannot be written issues no reset. The cursor's atomic durability mechanism is t
...
+full-text queries until rebuild and catch-up publish `ready`.
+
+Tantivy files are not an opaque encrypted cache. They contain the document-id term dictionary,
+analyzed term dictionaries, postings, frequencies and optionally positions; `surfaceTerms: true`
+also stores the original projected strings needed for surface-term features. Operators must protect
+the full-text directory with the same filesystem controls as Harper data. Removing source records
+does not erase old segment bytes immediately; normal Tantivy merge/reclamation governs physical
+removal, and destroying an index uses the wrapper's retirement protocol.
+
+The binding remains unloaded until a full-text declaration is activated. Before activation can
```
kriszyp, line 2775: Removing the design note was the right call, but some activation gates were recorded only there, and now they're nowhere in the tree. This paragraph still covers the exact pin, `dependencies.md`, and native CI on three platforms. These gates did not carry over:

- Feed `getUnindexableRecords()` (`resources/indexes/fullTextDerivedIndex.ts:270`) into the runner's operator-visible unindexable-record metric. The shared runtime is unchanged in this PR, so nothing reads the count yet.
- Show prolonged writer-open and writer-work retry in status before a lag rejection budget is enabled. Unknown open codes retry indefinitely by design (`DESIGN.md:2730-2733`). With the lag policy on, that turns into rejected source writes with no visible cause. The new sentence at `DESIGN.md:2743-2746` spells this out for the oversized-cursor park, and open retry has the same shape.
- In the packaged-native CI gate, check mutation-batch v3 `processed` counting, which the strict equality at `resources/indexes/fullTextDerivedIndex.ts:596` depends on. Also check representative terminal and retryable open codes against `isTerminalOpenError`.
- Run the HNSW/full-text coexistence matrix on one table before activation merges.

The inspection-failure gate from the note is now a deliberate, documented fail-closed trade at `DESIGN.md:2754-2757`, and that decision is recorded. For the rest, a few lines here or a tracking issue linked from the PR would save the activation author from rediscovering them.

resources/indexes/fullTextDerivedIndex.ts
```diff
@@ -0,0 +1,1165 @@
...
+	#closeOperations = new WeakMap<FullTextDerivedIndexEngine, Promise<{ cleanupError?: unknown }>>();
+	#shutdownFailure?: FullTextDerivedIndexError;
+	#openRetryDelayMilliseconds: number;
+	#openRetryWarned = false;
+	#writerRetryDelayMilliseconds: number;
+	#writerRetryWarned = false;
+	#terminalFailure?: FullTextDerivedIndexConfigurationError;
+	#resetOperation?: Promise<void>;
+	#resetEpoch?: bigint;
+	#failedDeliveryObserved = false;
```
kriszyp, line 172: Nit: the delta dropped the comment on `#failedDeliveryObserved`. The flag isn't self-explanatory. It stops the delayed `'failed'` callback (`:879`) from reporting a fault the runtime already saw as a `DERIVED_INDEX_FAILED` return from `deliver()` (`:276`). Could the one-liner come back?

unitTests/resources/fullTextDerivedIndex.test.js
```diff
@@ -0,0 +1,1324 @@
...
+		source.inspection = { state: 'checkpointed', committedPayload: encodeFullTextCursorPayload(cursor(20)) };
+		assert.deepStrictEqual({ ...backend.getDurableCursor().logs }, cursor(20).logs);
+		assert.strictEqual(source.inspectCalls, 2);
+		await backend.shutdown(2n);
+	});
+
+	it('does not invalidate the current cursor cache for a stale epoch shutdown', async () => {
+		const source = lifecycle({ state: 'checkpointed', committedPayload: encodeFullTextCursorPayload(cursor(10)) });
+		const { backend, setEpoch } = makeBackend(source);
+		backend.getDurableCursor();
```
kriszyp, line 1156: Nit: `makeBackend` starts at epoch 1, so this `getDurableCursor()` fills the cache under epoch 1. The assertion then checks that epoch 2 reuses it without re-inspecting. That pins the opposite of the documented rule that the first read in each acquisition refreshes (`DESIGN.md:2750-2752`).

The real runtime can't produce this sequence, because epoch 2 is minted only after `shutdown(1)` resolves. But if the cache is ever keyed by epoch, this test will fail for the wrong reason. Moving this call after `setEpoch(2n)` would test what the name says: a late `shutdown(1n)` doesn't discard epoch 2's fresh cache. The assertion still holds, because `isOwnerEpoch(1n)` is false at `resources/indexes/fullTextDerivedIndex.ts:342`.

cb1kenobi (commented):
No new blocking defect was confirmed on changed lines. Existing discussion already covers the earlier issues, and this head addresses the relevant code paths.

—
Reviewed 970de771

resources/indexes/fullTextDerivedIndex.ts
```diff
@@ -0,0 +1,804 @@
...
+		}
+		if (applied !== expected)
+			throw new FullTextDerivedIndexError(`Full-text engine applied ${applied} of ${expected} mutations`);
+		return true;
+	}
+
+	async #publish(command: BarrierCommand): Promise<void> {
+		if (command.horizon <= this.#lastPublishedSequence) return;
+		this.#assertCommandEpoch(command.epoch);
+		if (this.#lastAppliedSequence < command.horizon)
```
kylebernhardy: Addressed on the current head. The backend now marks native state dirty before awaiting applyMutationBatch, so a rejection after staging can never make shutdown choose require-clean. Rollback or publication is the only path that clears the dirty flag.

resources/indexes/fullTextDerivedIndex.ts
```diff
@@ -0,0 +1,805 @@
...
+		}
+		setImmediate(() => {
+			if (this.#activeEpoch !== activeEpoch) return;
+			if (this.#wake !== wake || (change === 'accepted-work-lost' && this.#shutdown)) {
+				if (change === 'accepted-work-lost' && this.#lossPendingEpoch === activeEpoch)
+					this.#lossPendingEpoch = undefined;
+				return;
+			}
+			try {
+				wake(change);
```
kylebernhardy: Verified against the current runtime scheduling. The backend callback invokes wake(true), but wake only schedules the drain with setImmediate; the notification finally block clears lossPendingEpoch before that scheduled drain executes. Replay therefore does not re-enter deliver while the loss fence is set.

kriszyp (approved):
Great job!

Issue #2577, opened 2026-09-11
package_component still walks the install through a loaded component's node_modules/harperdb, and skipping the path breaks fresh deploys


dawsontoth:
## Summary

A legacy component that has been **loaded** still makes `package_component` walk the entire Harper install, through `node_modules/harperdb`. This is the residue of #2487, which fixed the same problem for `node_modules/harper`.

Skipping the path is **not** the fix, and that is what makes this its own issue.

## Why the obvious fix is wrong

`symlinkHarperModule` treats the two names differently (`components/componentLoader.ts`):

- `harper` — created unconditionally: `rmSync(harperModule, { recursive: true, force: true })` then a symlink to `PACKAGE_ROOT`, whether or not the path existed.
- `harperdb` — only **repaired when the path already exists** (`harperdbModulePresent && !harperdbModuleLinked`).

So the sequence that breaks is:

1. A legacy component ships a real npm-installed `node_modules/harperdb`.
2. It is loaded, so the loader replaces that directory with a symlink to the install.
3. `package_component` packages it. With the link excluded, the archive contains `node_modules/` but **no** `harperdb`.
4. The archive is deployed to a **fresh** node. `installApplication` returns early because `node_modules` exists (`components/Application.ts`, "already has node_modules; skipping install"), so nothing installs `harperdb`.
5. On load, the loader's present-gate does not fire, because the path does not exist. Nothing recreates it.
6. A native-mode component importing `harperdb` fails. `jsLoader`'s synthetic alias (`security/jsLoader.ts`) only covers VM-loader modes, so bundled output or a natively-resolved dependency is not covered.

Both available choices are therefore wrong on their own:

| Treatment | Failure |
|---|---|
| Don't exclude it (today) | Packaging a loaded legacy component packages the whole install — the #2487 symptom, ~700MB and `Maximum response size reached` |
| Exclude it (by name or by resolved identity) | Fresh target loses `harperdb` with no recovery path; native `import from 'harperdb'` fails |

An earlier revision of the #2487 branch implemented the identity-based exclusion and introduced exactly the second failure; it was backed out rather than shipped.

## Suggested fix

Pack a **placeholder** at `node_modules/harperdb` rather than skipping it — an empty directory entry is enough, because the loader's gate is presence, not content. The target then extracts a path that exists, the present-gate fires on the next load, and the loader repairs it into the install link. That preserves both properties: the archive does not carry the install, and the target can still resolve `harperdb`.

[...]

PR #2585, opened 2026-09-13
Stop package_component packaging the Harper install through a loaded component's node_modules/harper

### LINKED ISSUE #2487: package_component follows node_modules/harper into the install, so packaging a loaded component packages Harper
## Summary

`package_component` on a component that a worker has actually loaded packages the **entire Harper install**, not the component. It either fails with `Maximum response size reached` after tens of seconds, or succeeds and hands back a tarball containing the install.

## Mechanism

Two behaviours that are individually correct combine badly:

1. `symlinkHarperModule` (`components/componentLoader.ts:534`) links the running install into `<component>/node_modules/harper` (and `harperdb`) on **every non-root component load**. That link is what makes `import { tables } from 'harper'` inside component code resolve to the same module instance Harper is running, so it is load-time infrastructure rather than part of the component's source.

2. The packer dereferences symlinks by default (`dereference = !options.skip_symlinks`) and **recurses into valid symlinked directories** — see the walk in `scanPackageDirectory`, `components/packageComponent.ts:60-97`, and the matching `tar.pack({ dereference })` in `streamPackagedDirectory`.

So packaging a loaded component walks `node_modules/harper` into the install root: `dist/`, the install's own `node_modules/`, and in a development checkout whatever else lives under it. `scanPackageDirectory`'s own docblock already notes that it inherits tar-fs's lack of symlink-cycle protection under `dereference`, and calls fixing that out of scope for the change that added it.

## Reproduction

1. `deploy_component` a component with `restart: true`, so a serving worker loads it and creates the link.
2. `package_component` that project.

Observed on a development checkout: 46s of packing, then `Maximum response size reached`. With `estimate: true` the scan does not return at all in a layout where the install root contains further checkouts.

## Why it has not been noticed

The obvious test path never creates the link. `add_component` builds and installs a project but nothing loads it, so `integrationTests/apiTests/components.test.mjs` packages a component whose `node_modules/harper` does not exist yet. The failure needs a component that has been *loaded*, which is the normal state of anything actually deployed and serving.

## Impact

- Packaging a running component does not work. Workaround: `skip_symlinks: true`, which packs the link literally.
- `package_component` runs the walk inline on the shared event loop server-side (noted in `streamPackagedDirectory`), so the failing case also occupies that loop for the duration.
- When the install is small enough not to trip the response cap, the operation *succeeds* and returns a package containing the install — wrong output rather than an error.

## Suggested fix

Treat a `node_modules/harper` / `node_modules/harperdb` entry that is a symlink resolving to `PACKAGE_ROOT` as excluded, in `isExcluded` (`components/packageComponent.ts`), so both the scan and the pack skip it. It is the running install, never component source, and a load recreates it wherever the component lands. Worth deciding at the same time whether this issue closes the broader symlink-cycle gap the docblock names, or just this special case, and whether the CLI's pre-deploy size scan needs the same exclusion.

## Verification route

Integration test: deploy a component with `restart: true` (so a worker loads it and the link exists), then `package_component` with `estimate: true` and assert the total size is the component's, not the install's.

## Provenance

Pre-existing on `main`; not introduced by any open PR. Surfaced while landing deploy certification (#2315 step 2, PR #2476), where a certification load created the link inside a candidate tree that was then swapped live — making a freshly added component reach `package_component` already carrying it. That PR removes its own load's footprint and deliberately does not touch the packer.

dawsontoth:
Packaging a component that a worker has **loaded** packaged the whole Harper install — ~700MB for a component of a few small files, seen in the wild as ~46s of tarring followed by `Maximum response size reached`.

Two individually correct behaviours combine badly. `symlinkHarperModule` links the running install into `<component>/node_modules/harper` on every non-root component load — that is what makes `import { tables } from 'harper'` resolve to the live module instance — and the packer dereferences symlinks and recurses into linked directories. So any component actually running gets packaged as Harper.

## The fix

`node_modules/harper` is excluded by **path**, not by resolving the link's identity. The loader does not merely add that link, it *replaces* whatever is there — `rmSync(harperModule, { recursive: true, force: true })` then a symlink to the install — and it does so whether or not the path existed. Component content there survives no load, so preserving it would protect something Harper destroys anyway.

That choice also buys three things an identity check could not:

- **No race.** Resolving identity needs I/O, so the set must be snapshotted before the walk — and a `deploy_component {restart: true}` landing mid-package lets the loader plant the link afterwards, so the walk recurses the install again. A path test has nothing to snapshot.
- **No per-package I/O.** No `lstat`/`realpath`, and the basename is tested before any path parse, so the common case pays a string compare inside tar-fs's synchronous `ignore` callback — which runs inline on the shared event loop under `package_component`.
- **Correct under `skip_symlinks`.** Gated on dereference, the link is instead packed *literally*, shipping an absolute path to the packaging host's install root that arrives dangling. The exclusion is unconditional.

## For the human reviewer

**`harperdb` is deliberately untouched, and it is not the same gap** — this is the judgement call worth checking. The loader only *repairs* that path when it already exists, unlike `harper` which it creates unconditionally, so a legacy component's real `node_modules/harperdb` is its own content. But excluding it — by name *or* by resolved identity — breaks fresh deployments: the archive omits the path, `installApplication` returns early because `node_modules` exists, the loader's present-gate never fires, and a native-mode `import from 'harperdb'` fails on the target.

An earlier revision of this branch implemented the identity-based exclusion and introduced exactly that regression; it was backed out rather than shipped. Closing it properly means packing a *placeholder* at the path so the gate can fire, which needs an injected tar entry rather than an `ignore` predicate. Tracked as HarperFast/harper#2577 with both options written up.

So a loaded legacy component that ships `harperdb` still packages the install. That is narrower than what this fixes and is not made worse by it.

Other things a reviewer may want to push on:

- The integration suite **skips Windows**, matching its `deploy-dangling-symlink` sibling: `restart: true` restarts there miss even a 45s readiness deadline (harper#1813 — `canPreStartReplacement` is false for win32), and `symlinkHarperModule`'s symlink needs developer mode, with its EPERM caught and logged. Without the skip the suite would report the packer broken when it is fine.
- A symlink *alias* that reaches the owned link under a different lexical path is still followed. The lexical test does not see it; the same limitation already applies to `scanPackageDirectory`'s dangling-link handling, which documents its lack of cycle protection.

## Verification

Both tests fail on base for the defect itself, not for a missing field — the assertions are ordered so the size check trips first:

| | On base |
|---|---|
| Unit — plants a link to `PACKAGE_ROOT` directly | 698,255,261 bytes |
| Integration — `restart: true`, so a **worker** plants it | 698,253,558 bytes |

- Unit runs across `{}`, `skip_symlinks` and `skip_node_modules`, and asserts the estimate stays under 100KB, that no install content reaches the extracted archive, and that both the component's own dependency and a **real** `node_modules/harperdb` directory still ship. That last case also caught a latent flaw in the archive assertion: `node_modules/harper` is a prefix of `node_modules/harperdb`, so the check is segment-exact rather than `startsWith`.
- Integration deploys with `restart: true` so a serving worker plants the link the way production does — the half a packer-only test cannot show — then asserts both the estimate and the real payload.
- 105 unit tests pass across the packaging and CLI suites. `bin/cliOperations.ts` reads only `totalSize` and `danglingSymlinks`, so nothing in the scan contract changed for it.

Closes #2487


kriszyp (approved):

PR #2605, opened 2026-09-14
Stage a component build with activate:false and deploy it later by deployment_id

dawsontoth:
`deploy_component` gains two optional fields. `activate: false` builds, installs and certifies a release and then stops, leaving a dormant artifact under `.deploy-staging`; `deployment_id` makes that exact artifact live later, resolving, fetching and installing nothing. The work that takes time and can fail happens when the operator chooses, and the cut-over itself is two directory renames. Both calls replicate as ordinary deploys, so every node stages and every node activates — no cluster barrier, and #2301's stage/activate protocol stays retired.

```mermaid
flowchart TB
  REQ["deploy_component"]
  REQ -->|"package / payload"| DEP["mode = deploy<br/>publish root config, build, certify"]
  REQ -->|"package / payload<br/>+ activate: false"| STG["mode = stage (new)<br/>same build, but no file: directory sources<br/>and no links that leave the tree"]
  REQ -->|"deployment_id"| ACT["mode = activate (new)<br/>verify the artifact and its descriptor<br/>nothing is resolved, fetched or installed"]

  STG --> CMP2["write .artifact.json, then .complete"]
  CMP2 --> DORM(["dormant artifact, nothing swapped<br/>.deploy-staging/&lt;deployment_id&gt;/&lt;component&gt;<br/>returns status staged + deployment_id"])
  DORM -.->|"later, surviving restarts"| ACT

  DEP --> JRN
  ACT --> JRN

  subgraph SWAP["swap"]
    JRN["write .activation.json"] --> B1["1. live moved to .deploy-aside"]
    B1 --> PUB1["publish root config (activation only)"]
    PUB1 --> B2["2. candidate renamed to live — commit point"]
  end

  B2 --> DONE["returns status success + deployment_id<br/>(an activation's row also records activated_from)"]

  classDef added stroke:#2da44e,stroke-width:3px
  class STG,ACT,CMP2,DORM,PUB1 added
```

Green marks what this PR adds; everything else is the path a deploy already took.

The [mode is chosen from the request](https://github.com/HarperFast/harper/pull/2605/changes?w=1#diff-99fa57601c9a33cdc212f18a0b2b88cbe1d2e4e2cd5ad8d7dd627e529c3df899R596), [validated so an activation carries no build inputs](https://github.com/HarperFast/harper/pull/2605/changes?w=1#diff-074ccf6833fa41042cf09889ef2c061ad132cff07e79c68f06c13da82eaeb277R613), and the [activate path branches around every payload, credential and install step](https://github.com/HarperFast/harper/pull/2605/changes?w=1#diff-99fa57601c9a33cdc212f18a0b2b88cbe1d2e4e2cd5ad8d7dd627e529c3df899R719) — without that branch a replicated activation looks exactly like a blob-backed deploy and every peer waits out `deployment_timeout` for a payload its row never had.

A staged artifact outlives an ordinary deploy of the same component, so staging is also how a release becomes revertible: deploy V1, stage V2, deploy V3, and activating V2's id later puts V2 back. What cannot be recovered is a version that was only ever deployed — its artifact is consumed by the swap and the tree it displaced is swept — which is what #2315 step 5 is for.

Every refusal on these paths carries a status code rather than reaching the caller as a 500, through the same `ClientError` the rest of the deployment operations use: 400 for a component that cannot be staged at all (a `file:` directory source), and 409 for a conflict with what is on disk — a deployment id already taken, an artifact whose descriptor this build cannot activate, one whose tree stopped being what was certified, an incomplete build, an unsettled verdict, an in-flight journal, or another component's artifact. **404 now means one thing only: nothing on this node answers to that id**, so a caller can tell "never existed, or already used" from "present, but not activatable".

This is step 6 of #2315. Step 4 bounded how many unactivated staged builds accumulate; nothing on `main` produced one, so its integration test had to plant them. This is the producer.

Four things make the delay safe, and each exists only because an immediate deploy did not need it. The artifact directory is named by the **public** deployment id while the deploy-lifecycle token stays per-invocation, because `DeployLifecycle` de-duplicates starts by the announced id and releases watcher suppression on the first matching end — two overlapping activations of one artifact sharing that id would count as one owner. A [**mandatory versioned `.artifact.json`**](https://github.com/HarperFast/harper/pull/2605/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R1335) is written before `.complete`, carrying the root-config entry the build would have published (explicitly `null` for a payload deploy), `installationIsOpaque`, and the admitted isolation intent; an optional record could not tell a payload build from a package build whose record was lost. [**Claiming an id is exclusive**](https://github.com/HarperFast/harper/pull/2605/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R1660), because a public id can be repeated where a minted UUID could not. [**Claiming an id publishes ownership**](https://github.com/HarperFast/harper/pull/2605/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R1733) as part of the claim rather than at certification, and an [unattributed directory is refused, never reclaimed](https://github.com/HarperFast/harper/pull/2605/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R1685): a build can hold an empty directory for minutes while it resolves and packs, so an empty `readdir` cannot tell a claim in flight from an abandoned one. And [**staging owns its bytes**](https://github.com/HarperFast/harper/pull/2605/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R1397): a `file:` directory source is refused, and so is any symlink whose walk leaves the built tree at any prefix — certification follows no links, and the post-swap relocation repair runs past the commit point and can only warn.

Certification moved out of [the swap primitive](https://github.com/HarperFast/harper/pull/2605/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R2787) into `prepareApplication`, so a delayed activation does not re-walk and re-fsync a `node_modules` it certified at build time while holding the preparation lock. An activation that fails before it commits [returns its artifact to dormant](https://github.com/HarperFast/harper/pull/2605/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R2850) — journal removed, its parent synced, and the published root-config entry taken back — because otherwise the next preparation reads live-plus-candidate-plus-journal as an abandoned activation and deletes the artifact, or rolls a first deploy forward unasked.

## For the human reviewer

The planning gate never cleared on one axis, and it is the first thing to rule on.

1. **Mixed-version clusters: detection, not prevention — settled, not open.** A stage replicates as an ordinary `deploy_component`, so a peer running a pre-5.3 build ignores `activate: false` and deploys the release immediately. [The origin reports it afterwards](https://github.com/HarperFast/harper/pull/2605/changes?w=1#diff-99fa57601c9a33cdc212f18a0b2b88cbe1d2e4e2cd5ad8d7dd627e529c3df899R1022) from a `staged: true` marker each node returns; nothing in core or in harper-pro's replicator carries a peer version or capability, so it cannot refuse in advance. The reviewer's alternative — replicate the stage as a distinct operation an old peer rejects as unknown — was declined by @dawsontoth: Harper clusters are kept at consistent versions, the UI will steer operators away from the mixed case, the docs will state the prerequisite, and adding a second wire path to cover a case that does not occur in practice would encumber the 99.9% path with more code that can break. **The planning gate's `better-alternative-exists` verdict on this axis is recorded as overruled by the owner, not carried as open.** Reversing it later is a wire-format change, not an API change: the two request fields stay as they are.

2. **Where the root-config publish sits, and the window that remains.** A package artifact's entry is [published between B1 (the live tree moved aside) and the commit rename](https://github.com/HarperFast/harper/pull/2605/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R2926) — the only point where the on-disk state recovery would find rolls *forward* to the certified artifact. Publishing any earlier, journal or no journal, leaves live present and the candidate present with no rollback record, which settlement reads as an activation that never started: it deletes the artifact and the next boot re-resolves that package from the registry instead. Three review rounds landed on this from different angles before it was right, so it is the first thing to check if you disagree. The window that remains is the inverse — a crash after the roll-forward state exists but before the publish leaves the certified artifact live under the *previous* release's config, because `rollForward()` publishes nothing. **That includes its isolation intent**, which is a containment boundary rather than a version string: a component staged to run isolated comes back non-isolated after an ordinary crash, with nothing in the operation reporting it. Closing it needs config to be an effect of the journal, which is step 3 — and that, or gating delayed activation to payload artifacts until step 3 lands, is the open call on @kriszyp's [thread](https://github.com/HarperFast/harper/pull/2605#discussion_r4009459237).

3. **A component with absolute or escaping internal symlinks cannot be staged.** [Staging refuses any link whose walk leaves the built tree at any prefix](https://github.com/HarperFast/harper/pull/2605/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R1397), including absolute ones under `node_modules` — where npm writes them on Windows for `file:`/workspace dependencies. The alternative was to trust `repairRelocatedDependencyLinks`, but that runs past the commit point and only warns (it logs and continues on a failed re-point, and skips a whole subtree on `EACCES`/`EMFILE`), so a component could go live holding a link to a path that no longer exists while the operation reported success. The cost is that such a component deploys immediately but cannot be staged until its links are relative; the error says so. The same rule is [re-run at activation](https://github.com/HarperFast/harper/pull/2605/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R4464) rather than trusted from `.complete`, which is a durability marker over the bytes and not a seal on them — a dormant artifact can be edited while it waits.

4. **Staged artifacts share `deployment_stagingRetention_maxCount` with incidental dormant builds**, and only the in-flight request is pinned — so an operator's staged release can be aged out by unrelated activity. One filter in `pruneDormantBuilds` if you want them to have their own lifetime.

5. **A dormant artifact's contents are not sealed, only its links are re-checked.** Activation re-runs the link rule and the load validation before the swap, but `validateComponentLoads` is a no-op on the main thread and the operations API deploys there, so an artifact whose files were edited while it sat dormant still activates. Detecting that means binding `.complete` to a manifest or digest of the tree — a whole-tree hash at stage time and again at activation, on a path where that cost is affordable. @kriszyp asked for it; I did the link and load re-checks he named as the minimum and left the manifest out, because what it should cover (every file? the installed tree? the descriptor too?) and how it interacts with `repairRelocatedDependencyLinks` rewriting links past the commit point are design questions, not a patch.

6. **A deployment id names one artifact only while that artifact exists.** Activation consumes it (the swap is a rename) and retention can prune it, after which the id is free to name different bytes. Durable identity after removal would need a tombstone store with its own retention and recovery. Cheap to add later; impossible to remove once operators script against the stronger promise.

7. **A redelivered stage is refused, not answered idempotently.** A replayed replication hits the `.complete` refusal and that peer records a *failed* deploy for an artifact that is staged and usable. Idempotent success was available and is what the reviewer preferred. I kept the refusal because `_deploymentId` is operator-forgeable, and answering "success" for bytes that were never staged is the worse failure. Backward-compatible to change later.

8. **`staged` is a terminal status on its own row, and activation gets a second row** linked by `activated_from`, rather than one row transitioning staged→success. This is what makes `delete_deployment_payload` and SSE tails behave, but it is the shape operators will script against, and it is hard to reverse after release.

9. **The config undo is best-effort, and for an existing component a failed undo strands config.** `addConfig` does not fsync and there is no publication lock, so a failed activation restores the previous entry without durability guarantees — that is #2315 step 3's work, deliberately not started here. It [takes back only what it published](https://github.com/HarperFast/harper/pull/2605/changes?w=1#diff-99fa57601c9a33cdc212f18a0b2b88cbe1d2e4e2cd5ad8d7dd627e529c3df899R829), re-reading from disk first because `set_configuration` writes the file without refreshing this process's config object, so a change acknowledged while the swap was retrying is not overwritten by a stale snapshot. That narrows the window rather than closing it. An undo that *fails* keeps the journal only for a first-ever deploy — the one shape recovery can still roll forward from, because the live path is absent. For a component that already had a tree, compensation has just put that tree back and taken the rollback record with it, so the next settle returns the artifact to dormant whatever the journal says: config names the staged release, the previous one is live, and an operator has to resolve it.

10. **A partially successful clustered activation cannot converge by retrying the same id.** The origin activates locally before peer fanout, and the swap consumes its staged directory — so if this node activates and a peer fails, retrying the same `deployment_id` is refused here (nothing is staged under it any more) before the retry ever reaches the peer that still holds the dormant artifact. The operator's route back is a fresh deploy, not a retry. Making the retry converge means either deferring the local swap until peers confirm, which gives up "every node activates the same way it deploys", or answering a `deployment_id` this node already activated as success, which needs the deployment row to be authoritative for what is live. Neither is small; both are cheaper before release than after. A local run turned up the same gap without a cluster: crash a first-ever activation after its journal lands, and recovery rolls the artifact forward at the next boot — the component comes up live from exactly those bytes, and the operator's retry of that `deployment_id` then answers 404, because nothing is staged under it any more. Their activation succeeded; what they are told is that it never existed.

Smaller carried items: concurrent deploys of different components can over-admit the global isolated-worker budget (adjudicated pre-existing — the same race `beforePrepare` already had); `activated_from` is registered for fresh installs only, matching every other attribute on `hdb_deployment`; an activation [re-walks the artifact tree](https://github.com/HarperFast/harper/pull/2605/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R4464) under the preparation lock, measured by a reviewer at 11,615 directories and 148 links in 380ms warm; every refusal of a `deployment_id` is a 404, so a caller cannot tell "never existed" from "already used"; and a crash inside the claim window — between the exclusive `mkdir` and the sidecar write, microseconds with only an `lstat` between them — burns that deployment id and leaves an empty directory nothing reclaims, since refusing an unattributed directory is what removed the only self-heal that path had. A claim that cannot [record who owns it](https://github.com/HarperFast/harper/pull/2605/changes?w=1#diff-4a096c65aa8965b002f6b76f448830f942dfda699311e0dd6bfc22587c938763R1733) now takes back the directory it made, so only a process death in that window — not an ordinary write failure — burns the id. Closing what remains needs claims serialized on the staging root; a staleness threshold would contradict this code's rule against wall-clock verdicts.

## Verification

Unit tree — run per directory rather than as one `test:unit:main`, because `unitTests/components/applicationSpawn.test.js` wedges on this machine at a process-group termination test (it wedges identically at the merge base, and the diff touches zero lines of that code), and `.mocharc.json` sets `timeout: 0`, so one hang swallows the whole run. Excluding that file: resources 2500 pass, components 1662 pass / 5 fail, server 960 pass / 10 fail, utility 759 pass, config 314 pass / 1 fail, validation 260 pass / 1 fail, bin 247 pass, sqlTranslator 67, upgrade 19.

Every one of those failures is pre-existing. The same three directories were run at the merge base (`3ece37294`) for comparison: `server` and `config` have **identical** failure sets, and `components` fails **16** at base against 5 here — every branch failure is also a base failure, none the other way. The named failures are `RuntimeModuleTracker`, `nonInteractiveSpawn git credential scoping`, `uWS UDS adapter`, process-group reclaim ordering, package-specifier resolution and `RootConfigWatcher`; two are explained outright by this environment — `validation`'s is `getDomainSocketPathLengthWarning`, which this worktree's own path length trips, and `dataLayer`'s suite fails to load at all without `HARPER_STORAGE_ENGINE=lmdb`.

**Live server, not only tests.** A Harper 5.3.0-alpha.1 instance on an isolated data root, driven through the operations API, with three builds of a component serving `{"build":"V1|V2|V3"}` over REST plus a real component (`three` as a dependency). A normal deploy returned its `deployment_id`; `activate: false` left V1 serving and wrote `.artifact.json`/`.complete`/`.component` beside the tree; the artifact survived a full Harper restart; activating it made V2 live carrying a sentinel file planted in the dormant tree, which is what proves the staged bytes were swapped rather than rebuilt; `restart: true` on the activation brought the server back serving V3. On the real component the split was **1.7s to stage, 0.1s to activate**, with a sentinel written inside the stage-time `node_modules/three` present in the live tree.

Refusals were exercised the same way: a consumed id, an unknown id, another component's artifact (preserved), `restart` with `activate: false`, `deployment_id` alongside `package`/`payload`, a non-boolean `activate`, a non-UUID id, an id already holding a completed build, an id another component owns, an empty unattributed claim directory (refused and left untouched — the concurrent-claim fix), an artifact whose tree was removed, and five malformed descriptors including one whose isolation intent disagreed with what it publishes. An artifact tampered with after staging — `ln -s /etc` planted in the dormant tree — was refused and preserved. Retention settled at maxCount + the pinned in-flight build and stayed there across further stages; `drop_component` reclaimed that component's artifacts and left its neighbour's alone; a stage whose package could not resolve left no residue; `delete_deployment_payload` on a staged row succeeded and the artifact still activated afterwards, which is the point of certifying at stage time. SSE phases are symmetric for both calls (`prepare/start → prepare/done → replicate → staged|success`).

Recovery was driven on a real boot by planting crash states and restarting: an interrupted activation returned its artifact to dormant (*"Returned the staged build … to dormant after an activation that never moved its live tree aside"*), and a described artifact carrying a stale `.unsettled` with no journal had the verdict cleared (*"Cleared a stale unsettled verdict … its activation was already settled"*). Both artifacts activated successfully afterwards. Two rounds of this exercise are what produced the status codes above; nothing else it found needed a code change.

Docs: HarperFast/documentation#670.

Complexity: complicated

<sub>Review-Coverage: authored=claude; ran=gemini,codex,cursor-grok,cursor-composer; adjudicated=domain; rounds=20; full=8 @ e7e1a9171ca8</sub>

<sub>Human-Review-Need: 4 (decisions: exported-for-testability, caught-failures-reclaim-crashes-do-not, one-404-class-for-missing-and-consumed, isolation-loss-documented-not-closed, artifact-consumed-by-rename, failed-clear-fails-closed, fail-open-peer-confirmation, dangling-internal-links-fail-closed) @ e7e1a9171ca8</sub>


components/Application.ts
```diff
@@ -3824,7 +4228,92 @@ export async function prepareApplication(application: Application, options: Prep
...
+): Promise<void> {
+	const deploymentDirPath = candidateDeploymentDirPath(application.dirPath, artifactId);
+	const candidateDirPath = candidateApplicationPath(application.dirPath, artifactId);
+	const unavailable = (why: string) =>
+		new Error(`Cannot deploy ${application.name} from deployment ${artifactId}: ${why}`);
+
+	const owner = await candidateComponentName(deploymentDirPath);
+	if (owner === undefined) throw unavailable('there is no staged build with that id on this node');
+	if (owner !== application.name) throw unavailable(`that staged build belongs to '${owner}'`);
+	if (!(await presentOrAbsent(join(deploymentDirPath, CANDIDATE_COMPLETE_MARKER)))) {
```
kriszyp, line 4439: `.complete` is only an empty durability marker; it is not bound to the tree state read here. The integration path itself stages and then writes `not-rebuilt.txt` after certification (`integrationTests/deploy/stage-then-activate.test.ts:117-139`), and activation accepts that modified tree. Replace that write with an invalid `index.js` or an external symlink and this request still reaches the commit rename without `validateCandidate` or `assertOwnedArtifactTree`, so it can report success for bytes that never passed the stage-time probe or ownership check. Please bind the marker to a manifest/digest and verify it here, or make dormant artifact storage actually immutable. At minimum, repeat the load and link checks before the rename. This is a deployment path, not a request hotspot, so the check need not add per-request overhead.

components/Application.ts
```diff
@@ -2400,72 +2669,137 @@ export async function activateCandidateApplication(application: Application, dep
...
+		// reads as an activation that never started: it deletes the deployment directory, and the next boot
+		// re-resolves the published package identifier from the registry instead — the substitution this step
+		// exists to prevent.
+		//
+		// A crash in the remaining window — after the roll-forward state exists but before this publish — is
+		// the inverse and smaller hazard: the certified artifact goes live while config still names the
+		// PREVIOUS release. Closing that needs config to be an effect of the journal itself, which is #2315
+		// step 3.
+		pendingEffect = 'publish the root configuration';
+		await options.afterJournal?.();
```
kriszyp: At this await, the previous live tree and rollback record have already been made durable. Killing the process immediately before the callback makes recovery take the roll-forward branch at `components/Application.ts:2416-2419`, but the journal records only the component and candidate ID and recovery never publishes `descriptor.rootConfig`. The candidate therefore becomes live while package and isolation configuration remains on the previous release; for example, an artifact staged to become isolated can boot non-isolated. `addConfig` also returns after a non-fsynced rename (`config/configUtils.ts:141-168`), so power loss can produce the same mismatch after this callback returns. Please include desired root-config state as a journaled effect and replay/flush it during recovery before roll-forward, or keep delayed activation unavailable for package artifacts until that work lands. Moving this callback elsewhere in the current rename window cannot close both crash sides.

components/Application.ts
```diff
@@ -1439,6 +1616,69 @@ async function ensureSecureStagingDirectory(stagingDir: string): Promise<void> {
...
+			throw error;
+		});
+		if (owner === undefined && entries.length > 0) {
+			throw new Error(
+				`Deployment id ${basename(deploymentDirPath)} is already in use by a build that has not named its ` +
+					`component yet`
+			);
+		}
+		// Nothing certified it, so nothing is lost by rebuilding over it.
+		await rm(deploymentDirPath, { recursive: true, force: true, maxRetries: 3, retryDelay: 100 });
```
kriszyp: An empty directory is still a live claim. Component A can complete the `mkdir` at `components/Application.ts:1631`, then spend minutes resolving or packing at `components/Application.ts:3035` before any candidate tree or `.component` exists. Component B using the same `_deploymentId` holds a different component lock, sees zero entries here, deletes A's directory, and recreates it. Both builds can then share the replacement directory; their EEXIST-tolerant descriptor/marker writes can let one report `staged` with the other's ownership, or either failure cleanup can delete both builds. Recovery already treats this exact unowned state as in-flight at `components/Application.ts:2154-2158`. Please publish ownership atomically with the claim—such as preparing a sidecar-bearing directory and renaming it into the deployment ID—or lock by deployment ID. A single empty `readdir` result should never authorize removal.

components/Application.ts
```diff
@@ -1296,6 +1311,151 @@ function activationJournalPath(componentDirPath: string, deploymentId: string):
...
+			// `node_modules` on Windows for a `file:`/workspace dependency, so such a component deploys
+			// immediately but cannot be staged until its links are relative. A refusal naming the path is a
+			// better answer than a release that loads on POSIX and not on Windows.
+			if (isAbsolute(await readlink(entryPath))) {
+				throw new Error(
+					`Cannot stage ${componentName}: ${entryPath} names its target inside the build by absolute path ` +
+						`(${await readlink(entryPath)}), which activation moves. Re-link it relatively — on Windows, npm ` +
+						`writes absolute junctions for 'file:' and workspace dependencies, so those have to be relative ` +
+						`before the component can be staged.`
+				);
```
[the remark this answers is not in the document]
dawsontoth: Already the case at the current HEAD (1331446ab) — `readlink` was hoisted into `linkTarget` in the commit that added the escape-and-re-enter depth check, and the absolute-path error interpolates that variable. The duplicate read you're looking at was in the earlier commit this review ran against; no further change needed.

components/operations.js
```diff
@@ -918,12 +988,42 @@ async function deployComponent(req) {
...
+					const unconfirmedError = new ServerError(
+						`Component '${application.name}' was staged on the origin node, but ${unconfirmed.length} peer ` +
+							`node(s) did not confirm staging: ${detail}. Either they are unreachable, or they run a build ` +
+							`that predates staged deploys — which treats this request as an ordinary deploy, so the ` +
+							`release is already serving there. Check those nodes before activating deployment ` +
+							`${recorder.deploymentId}, or pass ignore_replication_errors: true to accept the difference.`
+					);
+					throw unconfirmedError;
+				}
 			}
```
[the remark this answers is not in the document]
dawsontoth: Good catch — that check is the only safety net for a hazard the PR explicitly records as unpreventable, so an untested branch there is worse than usual. Fixed in 58041c9d3.

Rather than stub `replicateOperation` and drive the whole operation, I split the predicate out as `unconfirmedStagingPeers(replicated)` so it can be tested directly, and covered the three things that matter: a fully-upgraded cluster is not reported as unconfirmed whichever shape harper-pro's replicator returns (flat, `value`, or `body` — assuming flat alone would misreport every peer), an old peer and a failed peer both read as "did not confirm", and a missing or non-array aggregate reports nothing rather than throwing.

kriszyp (changes requested):

components/Application.ts
dawsontoth: Re-ran both checks before the swap: `activateStagedArtifact` now calls `assertOwnedArtifactTree` (the link rule) and `validateCandidate` on the dormant tree, with a fails-on-base unit test that plants an escaping symlink into a staged artifact and asserts the activation is refused and the artifact preserved. Content tampering — your invalid `index.js` — is still undetected: the load validation is a no-op on the main thread until #2315 step 2, and binding `.complete` to a manifest/digest is a design change I've raised with @dawsontoth rather than slipping into this PR.

components/Application.ts
dawsontoth: Fixed: `.component` is now written as part of the claim, immediately after the exclusive `mkdir`, and an unattributed directory is refused rather than reclaimed — a single empty `readdir` no longer authorizes removal, which is the reading recovery already gave it. Two fails-on-base unit tests cover the empty-directory refusal and that the claim names its component before any tree exists to infer one from.

components/Application.ts
dawsontoth: Not changed, and you're right that moving the callback cannot close both sides — the PR already carries this as the remaining window. Closing it is #2315 step 3 (root config as a journaled, durable effect); as you note, `addConfig` returning after a non-fsynced rename means journaling the intent alone would not be enough either. Your alternative — keeping delayed activation unavailable for package artifacts until that lands — is a scope call I've put to @dawsontoth rather than taking unilaterally.

components/Application.ts
```diff
@@ -1439,6 +1633,72 @@ async function ensureSecureStagingDirectory(stagingDir: string): Promise<void> {
...
+		}
+		// Ownership has to be POSITIVE to reclaim, and EMPTINESS IS NOT A VERDICT. The sidecar below is
+		// written as part of the claim, so a directory that names nobody is another component inside the
+		// window between its own mkdir and that write — a live claim, not an abandoned one. Nothing on disk
+		// separates it from a claim that got no further, and only the claimant's preparation lock, which is
+		// not this one, serializes it: a deploy can hold an empty directory for minutes while it resolves and
+		// packs, so reclaiming on an empty read deletes a build that is running. Refusing costs a retry under
+		// a fresh id; reclaiming costs someone else's deploy. ENOENT is the exception — the directory is gone,
+		// so there is no claim left to take.
+		if (owner === undefined && (await presentOrAbsent(deploymentDirPath))) {
```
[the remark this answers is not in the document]
dawsontoth: Confirmed and deliberate — thank you, this is the trade I took. Two reasons I'm not closing it with a staleness threshold: an mtime/`timeoutMs` bound cuts against the rule this code states explicitly for the preparation lock ("a preparation caller never steals a lock from a known-live owner based on elapsed wall time: installs can be long-running and clocks can jump"), and the window it would cover is now `mkdir` → `writeControlFileDurably` — microseconds, with no I/O between them beyond an `lstat`. Since ids are per-deploy UUIDs, a burned one is unreachable in practice and the residue is a single empty directory per crash inside that window.

The clock-free way to actually close it is to serialize claims on the staging root, so "empty and unattributed" can be evaluated at a moment when no claim can be in flight — a new cross-component lock on the deploy path, which is not worth it for this. Recorded in the PR description's carried items instead. (Related: claiming no longer removes a directory it found *absent* either — the exclusive `mkdir` decides, since another component can claim it between that read and the removal.)

components/operations.js
```diff
@@ -781,8 +801,54 @@ async function deployComponent(req) {
...
+				// Read BACK rather than compared against `entry`: the undo below has to tell "nobody has touched
+				// this since" from "somebody has", and only a value that went through the same write-and-parse
+				// round trip is comparable to one read later. Comparing the entry as passed would make any
+				// serialization difference read as a concurrent change and silently skip every undo.
+				const published = configUtils.getConfigObj()?.[req.project];
+				return async () => {
+					// Only what this activation published is this activation's to take back. The preparation lock
+					// does not serialize root-config writers, so a `set_configuration` acknowledged while the swap
+					// was retrying would otherwise be overwritten by a snapshot taken before it.
+					if (!isDeepStrictEqual(configUtils.getConfigObj()?.[req.project], published)) return;
```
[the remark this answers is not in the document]
dawsontoth: Fair, and it's the same precedent you name — extracted as `publishedEntryStillStands` and exported the way `unconfirmedStagingPeers` was. Four tests pin both regressions that would fail silently: comparing against the entry as passed rather than as read back (covered by a key-order case), and reading without refreshing first, which compares against a value that predates the very change the guard exists to protect (the refresh and the read are passed in, so the test asserts their order). No behavior change; the refresh is load-bearing because `set_configuration` writes the config file without updating this process's config object.

components/Application.ts
```diff
@@ -2400,72 +2763,138 @@ export async function activateCandidateApplication(application: Application, dep
...
+		// reads as an activation that never started: it deletes the deployment directory, and the next boot
+		// re-resolves the published package identifier from the registry instead — the substitution this step
+		// exists to prevent.
+		//
+		// A crash in the remaining window — after the roll-forward state exists but before this publish — is
+		// the inverse and smaller hazard: the certified artifact goes live while config still names the
+		// PREVIOUS release. Closing that needs config to be an effect of the journal itself, which is #2315
+		// step 3.
+		pendingEffect = 'publish the root configuration';
+		undoAfterJournal = await options.afterJournal?.();
```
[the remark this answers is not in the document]
dawsontoth: Right on both counts, and I've made the isolation consequence explicit rather than leaving it inside "names the previous release" — in the comment at this window and in DESIGN.md: the staged entry carries the isolation the build admitted, `rollForward()` publishes nothing, so a component staged to run isolated comes back non-isolated after an ordinary crash with nothing in the operation reporting it. That is a containment boundary, and the window should be weighed by it rather than by the version mismatch.

No behavior change here — the fix-or-gate call (journaled durable config now, or keeping delayed activation unavailable for package artifacts until #2315 step 3 lands) is with @dawsontoth, and https://github.com/HarperFast/harper/pull/2605#discussion_r4009459237 stays open deliberately so it is not merged past. Ledger item 2 in the description carries it as unresolved.

components/Application.ts
```diff
@@ -2483,7 +2914,36 @@ export async function activateCandidateApplication(application: Application, dep
...
+						undoError
+					);
+					return false;
+				}
+			);
+		}
+		// Kept only where it changes the outcome. Config is stranded either way for a component that already
+		// had a tree — the durable-config window #2315 step 3 closes — and a journal that recovery will only
+		// settle back to dormant buys nothing for holding it.
+		if (configRestored || !recoveryCanRollForward) await returnToDormant();
```
kriszyp, line 2994: For an existing component, `recoveryCanRollForward` is false. If B2 fails after `components/Application.ts:2893` published the staged package config, compensation restores the previous live tree, and the returned config undo rejects, this condition still calls `returnToDormant()` because `!recoveryCanRollForward` is true. That removes the journal while leaving old bytes live, the certified candidate dormant, and root config naming the staged release; the next restart can then re-resolve that package from the registry rather than use the certified artifact. The new fixture at `unitTests/components/deployStageActivate.test.js:731-754` checks only the trees and journal, so it preserves this mismatch. Please make root config a journaled, durable effect that recovery can restore/replay, or keep delayed package activation unavailable until that transaction exists.

components/Application.ts
```diff
@@ -1439,6 +1633,74 @@ async function ensureSecureStagingDirectory(stagingDir: string): Promise<void> {
...
+			logger.warn(
+				`Could not restrict component deploy staging permissions for ${deploymentDirPath}:`,
+				errorForLog(error)
+			)
+		);
+	}
+	// Published as part of the claim, not at certification: resolving and packing can take minutes, and until
+	// a tree exists to infer an owner from, a directory that answers to nobody is one another component will
+	// take for abandoned.
+	await writeControlFileDurably(join(deploymentDirPath, CANDIDATE_COMPONENT_FILE), componentName);
```
kriszyp: After this invocation wins the exclusive `mkdir`, publishing `.component` can still fail before the final sidecar link exists—for example, an ENOSPC/EIO during the temp-file write or sync. `buildCandidateApplication` calls `claimDeploymentDirectory` before its cleanup `try` at `components/Application.ts:3156-3157`, so that failure leaves an unattributed directory. Recovery preserves it, and every retry of the same deployment ID reaches the `owner === undefined` branch, gets EEXIST from the second `mkdir`, and refuses it permanently. Please clean up only when this invocation created the directory and its sidecar publication failed; do not broaden cleanup to an EEXIST directory another claimant may own. A focused check can inject a pre-link control-file failure and retry the same ID.

components/Application.ts
```diff
@@ -1296,6 +1311,168 @@ function activationJournalPath(componentDirPath: string, deploymentId: string):
...
+				);
+			}
+			// Where the link ENDS UP is not enough: a target that leaves the candidate and comes back resolves
+			// inside it today and somewhere else once activation renames the tree, because the same relative
+			// expression is then evaluated from `components/<component>/…`. Counting `..` segments does not
+			// catch it either, since an intermediate symlink (`up -> ..`) reduces depth without spelling it.
+			// So every PREFIX of the walk is resolved, with symlinks followed as the filesystem will follow
+			// them, and each one has to still be inside the candidate.
+			let prefix = dirname(entryPath);
+			for (const segment of linkTarget.split(/[\\/]/)) {
```
kriszyp: On POSIX, `\` is a legal filename character rather than a path separator. A root-level symlink targeting the literal filename `..\asset` resolves inside the candidate and remains valid after relocation, but this split produces `..` followed by `asset`, drops the synthetic depth below zero, and refuses the stage. I confirmed on Linux with the current Node runtime that `path.resolve('/candidate', String.raw`..\asset`)` remains `/candidate/..\asset` while this loop reports an escape. Please split on both separators only on Windows (and only `/` on POSIX), so the guard rejects actual traversal without excluding valid package contents.

components/Application.ts
dawsontoth: @kriszyp What do you think? Out of my depth on that question.

components/Application.ts
```diff
@@ -1296,6 +1311,176 @@ function activationJournalPath(componentDirPath: string, deploymentId: string):
...
+	action: 'stage' | 'activate' = 'stage'
+): Promise<void> {
+	// The operator supplied a component that cannot be staged (400); or the artifact they named exists and is
+	// theirs but is no longer what was certified (409). Neither is a server fault, and both reached the
+	// operations handler as a bare 500 until a live run showed what that looks like to a caller.
+	const refuse = (message: string) => {
+		const error: Error & { statusCode?: number } = new Error(message);
+		error.statusCode = action === 'stage' ? 400 : 409;
+		return error;
+	};
```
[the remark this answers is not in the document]
dawsontoth: Taken — and further than suggested. All five hand-rolled `Error & { statusCode?: number }` helpers in this file are now `ClientError`, including the 404 family in `activateStagedArtifact` that predates the status-code work, so the convention is uniform here rather than mixed. One import replaces the five casts.

No behavior change: re-verified against a live server — 409 for a taken deployment id and for every descriptor rejection, 409 for an artifact whose tree stopped being what was certified, 400 for a `file:` directory source, 404 unchanged for an id that is not on this node. 882 component unit tests and the stage/activate integration suite pass.

kriszyp (commented):

components/Application.ts
dawsontoth: Confirmed and fixed. Link targets are now split by the rules of the platform they are on — `/` only on POSIX, both on Windows — so `..\asset` stays the one legal entry it is. The new test fails on the previous commit with exactly the escape you predicted: `alias reaches .../.deploy-staging/a1 on its way to ..\asset`.

components/Application.ts
dawsontoth: Fixed as scoped: the claim takes back the directory *it created* when it cannot name it, and only that one — a directory found by EEXIST may be another claimant's, and removing it is the race the refusal exists to prevent. Ownership publication is now `publishClaimOwnership(dir, name, write = writeControlFileDurably)`, with the writer as a parameter so the failure is testable without a filesystem that fails on exactly that write and nothing else (same shape as `publishedEntryStillStands`). Two tests: an injected ENOSPC leaves no directory, so the id can be claimed again; a successful write keeps it and records the owner.

components/Application.ts
dawsontoth: You're right about the end state, and I want to be precise about one step: keeping the journal does not avoid it. On the next start, settlement reads live-plus-candidate-with-no-rollback-record — compensation has already restored the tree and taken the record with it — and for a described artifact that branch returns it to dormant and retires the journal anyway. So holding the journal defers the identical mismatch by one boot and leaves the artifact unusable until then; three review legs landed on that independently. `returnToDormant()` here is not what creates the divergence.

What creates it is that root config is not a journaled, durable effect — your earlier thread, https://github.com/HarperFast/harper/pull/2605#discussion_r4009459237, which is still open and is the one deliberately unresolved item on this PR. The fix-or-gate call between doing #2315 step 3 now and keeping delayed activation unavailable for package artifacts is @dawsontoth's, not mine to take.

What I have changed is the fixture you flagged: it no longer checks only trees and journals. It now pins the divergence — the entry this activation published is asserted as still standing while the previous release is live — so the hazard is stated in the test rather than left to be inferred. Whoever closes step 3 should find that assertion inverted. Ledger item 9 in the description carries it, and item 2 records that the same window costs the artifact's recorded isolation, not just a version string.

components/Application.ts
```diff
@@ -1439,6 +1647,108 @@ async function ensureSecureStagingDirectory(stagingDir: string): Promise<void> {
...
+	}
+	if (process.platform !== 'win32' && (claimed.mode & 0o777) !== 0o700) {
+		await chmod(deploymentDirPath, 0o700).catch((error) =>
+			logger.warn(
+				`Could not restrict component deploy staging permissions for ${deploymentDirPath}:`,
+				errorForLog(error)
+			)
+		);
+	}
+	await publishClaimOwnership(deploymentDirPath, componentName);
```
kriszyp, line 1715: Cleanup only runs when the ownership write rejects in a live process. If the process exits after the exclusive `mkdir` at `components/Application.ts:1665` and before `writeControlFileDurably` publishes the final `.component` entry, no catch runs and the directory remains unattributed. Retrying that deployment ID then reads no owner, receives EEXIST from the second `mkdir` at `components/Application.ts:1683`, and refuses the ID permanently—the same wedge through a crash rather than ENOSPC. Please make claiming the ID and publishing its durable owner one recoverable transaction, such as an atomically published ownership record in the staging parent, or add recovery that can distinguish a dead claim. A focused subprocess check can pause before the final sidecar link, terminate the claimant, restart, and retry the same ID.

components/Application.ts
dawsontoth: @dawsontoth has ruled: defer to step 3 rather than gate delayed activation, which matches the multi-PR intent you read correctly.

#2315's step 3 now carries both halves explicitly — the failed-undo divergence, and that the same window costs the artifact's recorded **isolation** and not just a version string, so a component staged to run isolated can come back non-isolated after an ordinary crash. The fixture asserts the divergence rather than checking only trees and journals, and whoever closes step 3 should find that assertion inverted.

Step 5 also grew from this: local experiments showed the clustered-retry gap is not "the deployment row becomes authoritative for what is live" — rows are origin-only, so a peer cannot use them — but a provenance marker written into the candidate at stage time, which rides the commit rename atomically and is absent rather than stale after an ordinary deploy. That is recorded there with the evidence.

Thank you for these — the backslash one in particular would have made legitimate packages unstageable on POSIX.

components/Application.ts
dawsontoth: @dawsontoth has ruled: defer to step 3 rather than gate delayed activation, which matches the multi-PR intent you read correctly.

#2315's step 3 now carries both halves explicitly — the failed-undo divergence, and that the same window costs the artifact's recorded **isolation** and not just a version string, so a component staged to run isolated can come back non-isolated after an ordinary crash. The fixture asserts the divergence rather than checking only trees and journals, and whoever closes step 3 should find that assertion inverted.

Step 5 also grew from this: local experiments showed the clustered-retry gap is not "the deployment row becomes authoritative for what is live" — rows are origin-only, so a peer cannot use them — but a provenance marker written into the candidate at stage time, which rides the commit rename atomically and is absent rather than stale after an ordinary deploy. That is recorded there with the evidence.

Thank you for these — the backslash one in particular would have made legitimate packages unstageable on POSIX.

kriszyp (approved):
Excellent!

components/Application.ts
dawsontoth: Right — this delta only closes the *caught* failure; a process death in that window still burns the id, and the review flagged the same boundary.

@dawsontoth has decided to defer the remaining fixes to later steps rather than hold #2605, so this is recorded in #2315 step 5 with both of your designs — an atomically published ownership record in the staging parent, or recovery that can distinguish a dead claim — plus a third worth weighing first, since it needs neither a new record nor a liveness test: build the claim at `.deploy-staging/.claiming-<uuid>/`, write `.component` into it, fsync, and `rename` it onto `<id>`. The directory then only ever appears at its id already attributed, so the window does not exist; a crash before the rename leaves a `.claiming-*` directory no id refers to. The wrinkles are that POSIX `rename` onto an *empty* directory succeeds (harmless under that scheme, but it means the two schemes cannot both be live) and that recovery scans would have to sweep `.claiming-*`.

One thing shipped for it in the meantime: the refusal now says what it costs and what clears it, rather than reading like a transient conflict — no retry of the same id can clear it, so it points at minting a fresh one. That landed after #2605 merged, so it is sitting on a branch rather than on `main`; I'll surface it to @dawsontoth as a follow-up.

PR #2612, opened 2026-09-15
Size describe record counts with a bounded estimate instead of a second full key scan

kriszyp:
`describe_table` / `describe_all` sizes a table's record count by scanning its primary store, and when that scan blew its time budget it answered by starting a second, unbounded exact `getKeysCount()` over the same key space — the escape cost more than the scan it was escaping. The base now comes from a calibrated rocksdb-js [`CountEstimator`](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R6178), the forward scan is [bounded whatever the base says](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R6210), and `estimated_record_range` is derived from [evidence the samples actually have](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R6279) rather than from a statistical interval around a rate the unsampled middle may not share.

Measured on rocksdb-js 2.9.0, warm page cache, 73-byte values, no SST I/O — the best case for the scan:

| keys | `getKeysCount()` | `estimateCount()` |
|---|---|---|
| 1,000,000 | 92 ms | < 0.05 ms (confidence 0.9) |
| 10,000,000 | 934 ms | < 0.05 ms (confidence 0.9) |

Linear at ~93 ns/key, so a 100M-key store is ~9.3 s warm and worse cold. This is the describe-path counterpart to #2478, which moved *query planning* off the same exact count.

## For the human reviewer

**The planning gate did not clear, and I resolved it by adopting the reviewer's alternative rather than overruling it.** My design note claimed three defective sites; two did not exist. `openAuditStore()` returns `RocksTransactionLogStore` (which `extends EventEmitter`), never `RocksDatabase`, so the `auditStore instanceof RocksDatabase` branch in the REST describe response was dead on both engines, and `getAuditSize()` was already returning a correct byte total rather than the `NaN` I claimed. Both of my proposed "fixes" would have thrown. Confirmed at runtime before cutting scope. What remains is the one real site plus [removal of that dead branch](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R2013), which is behavior-preserving.

1. **Estimated base over exact count.** A calibrated `CountEstimator` [replaces the exact `getKeysCount()`](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R6178) as the extrapolation base, trading a guaranteed-correct base for a bounded one on the path that exists *because* scanning is already too slow. Reversible per-engine, but the whole `baseMin`/`baseMax` construction unwinds with it. A "no" means describe stays O(keys) on large RocksDB tables.
2. **`estimated_record_range` is a heuristic interval, not a guaranteed bound.** Both [endpoints](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R6280) are estimates on RocksDB — the upper one is the [uncalibrated physical remainder widened by its own reported confidence](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R6218). It holds across every distribution tested, but it is not a proof, and the code says so rather than implying otherwise. The expensive part of revisiting this is that Studio consumes the field.
3. **Point-estimate accuracy is traded for range honesty on non-uniform churn.** Where churn sits outside the sampled ends, prefix calibration cannot see it and `record_count` over-reports — 6000 for 3000 live rows in the middle-churn fixture — while the range still contains the truth. The alternative is reporting no count at all when the base cannot be trusted. `record_count` is what most callers read; the range is what the tests defend.
4. **End-only sampling.** The estimate still extrapolates from a head and a tail sample, so any distribution that differs in the middle is invisible by construction. Sampling the middle too would cost another seek per checkpoint and is the natural follow-up if the over-report in (3) turns out to matter in practice.
5. **Checkpoint ceiling bounds latency at the cost of exactness on prefix-churned tables.** [`MAX_ESTIMATE_CHECKPOINTS`](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R219) [caps the forward scan](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R6210) at ~20 budget intervals whatever the base says, because a base that keeps undershooting would otherwise hold the halfway test false forever and walk the whole table. The cost is real: head-concentrated churn returns 2038 for 3000 live rows where the pre-ceiling code returned an exact 3000. Reviewers proposed tying the ceiling to `entriesScanned >= entryCount` so well-calibrated scans finish exactly; that restores exactness where calibration works and reopens the unbounded case where it does not. One constant, trivially reversible.
6. **No usable base degrades to the exact scan.** If both the range estimator and [`getEstimatedKeyCount()`](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R6195) come back unusable, the scan runs to completion — slow but correct — and logs once. The alternative, letting the ceiling fire with no base, publishes a small number alongside a range claiming to contain it, which is worse than a slow correct answer.
7. **Sample floor at 1000 makes `timeLimit` advisory below it.** Tables under [`MIN_ESTIMATOR_SAMPLE`](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R217) entries always complete exactly regardless of the caller's budget. This keeps small, heavily rewritten tables — whose physical key statistics are least trustworthy — off an estimated base at all, and it is why the existing `describe_table` integration assertions (exact counts of 3 and 350) still hold. A small but genuinely slow table (large values, cold cache) has no escape.
8. **Test seams are monkey-patched onto the live store.** The [estimator-undershoot](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-9252a0adb45941f1db239f98274a1028094bec7ac3b95243c5271bca11495fdaR328), estimator-failure, [zero-estimate](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-9252a0adb45941f1db239f98274a1028094bec7ac3b95243c5271bca11495fdaR293) and [physical-base-inflation](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-9252a0adb45941f1db239f98274a1028094bec7ac3b95243c5271bca11495fdaR262) cases stub `createCountEstimator` / `estimateCount` / `getEstimatedKeyCount` on `primaryStore` and restore them in `finally`. That is how the existing suite already spies on the store, and the alternative is an injection seam in production code that exists only for tests.

## Verification

End-to-end route: existing integration coverage. `integrationTests/apiTests/terminology.test.mjs` and `integrationTests/apiTests/northwind.test.mjs` assert exact `record_count` values (3 and 350) and `skip_record_count` behavior through the operations API; both tables sit below the sample floor, so they exercise the exact path this change must preserve.

Two files change. [`resources/Table.ts`](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R6178) carries the estimator loop and the interval; [`unitTests/resources/getRecordCount.test.js`](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-9252a0adb45941f1db239f98274a1028094bec7ac3b95243c5271bca11495fdaR120) carries the coverage. Within `Table.ts`, beyond the decisions linked above: [`usableCount`](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R222) validates `count` and `confidence` together so a store answering with an unexpected shape degrades rather than poisoning the result; the reverse sample [stops at the forward scan's own last key](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R6249), which makes the two samples disjoint by construction and returns an exact count when they meet; and the [significant-unit division loops](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R6285) so a physical base orders of magnitude above the calibrated estimate cannot round it to zero.

Gates run in this worktree, post-merge with `origin/main`:

- `npm run test:unit:resources` — 2680 passing, 0 failing
- `HARPER_STORAGE_ENGINE=lmdb` over the same suite — 2098 passing, 0 failing
- `npm run test:unit:main` — exit 0
- `prettier --check` and `oxlint --quiet` clean on both changed files

Fails-on-base, with `resources/Table.ts` reverted to the merge-base: three of the new assertions fail, including the base reporting **2000 records for a 1500-row table** where this change returns an exact 1500.

Behavior across churn distributions (3000 live rows unless noted, estimator forced with `timeLimit: -1`). The contract under test is that the range contains the true count and the reported count lies inside its own range:

| scenario | `record_count` | `estimated_record_range` | contains truth |
|---|---|---|---|
| no churn | 3000 | [2000, 3755] | yes |
| uniform ×10 | 3000 | [2000, 12283] | yes |
| uniform ×30 | 3000 | [2000, 12351] | yes |
| middle-only ×30 | 6000 | [2000, 7244] | yes |
| tail-only ×30 | 6000 | [2000, 7241] | yes |
| head-only ×30 | 2038 | [2038, 3800] | yes |
| 20,000 clean | 20000 | [2000, 22759] | yes |
| 1500 rows | 1500 exact | — | yes |
| 30 rows | 30 exact | — | yes |

Every defect found during review is covered by a test verified to fail without its fix: LMDB sampling disabled ([3 cases](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-9252a0adb45941f1db239f98274a1028094bec7ac3b95243c5271bca11495fdaR120) fail), [cumulative rather than incremental `advance`](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-9252a0adb45941f1db239f98274a1028094bec7ac3b95243c5271bca11495fdaR150) (2001 reported entries on a 1500-row table), [the deletion-entry range collapse](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-9252a0adb45941f1db239f98274a1028094bec7ac3b95243c5271bca11495fdaR210) (5000 entries / 2000 live reported `[0, 6]`), [a physical base rounding the estimate away](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-9252a0adb45941f1db239f98274a1028094bec7ac3b95243c5271bca11495fdaR262) (2000 reported for ~3000 live), and [a zero range estimate never reaching the whole-store fallback](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-9252a0adb45941f1db239f98274a1028094bec7ac3b95243c5271bca11495fdaR293). [The sample-floor boundary case](https://github.com/HarperFast/harper/pull/2612/changes?w=1#diff-9252a0adb45941f1db239f98274a1028094bec7ac3b95243c5271bca11495fdaR140) pins that a table just above the floor still returns an exact count.

## Docs

No companion documentation PR. The documented `describe_table` / `describe_all` contract is preserved exactly: counts above a threshold are estimated, `estimated_record_range` accompanies an estimate, and `exact_count: true` still forces a full exact scan (the `exactCount` short-circuit runs before any budget or estimator logic). What changed is the mechanism behind the estimate, not its published contract. The one refinement worth making later is that `reference/operations-api/operations.md` does not say the range is a heuristic interval rather than a bound, or that a non-uniformly churned table can over-report `record_count` — worth a docs note if this lands, but not a contract change this PR creates.

Complexity: complicated

<sub>Review-Coverage: authored=claude; ran=gemini,cursor-composer,codex,cursor-grok; adjudicated=domain; rounds=6; full=2 @ 68f8863cf817</sub>

<sub>Human-Review-Need: 4 @ 68f8863cf817</sub>


kriszyp:
## Suggested resolution for the `v5.2` cherry-pick

Verified on a worktree off `origin/v5.2` with this patch applied — not a code-trace suggestion.

Three conflicts, none semantic. `getRecordCount` itself applied cleanly: v5.2's copy of that method is byte-identical to this PR's merge-base.

| conflict | cause | resolution |
|---|---|---|
| `resources/Table.ts` constants | `MIN_ESTIMATOR_SAMPLE` / `MAX_ESTIMATE_CHECKPOINTS` / `usableCount` were added next to the `Prefer: count=` guardrails (`MAX_EXACT_COUNT_SCAN`, `MAX_COUNT_PAGE`, `COUNT_YIELD_INTERVAL`), which exist only on `main` | keep the three additions, drop the `Prefer: count=` block — it is not part of this change and does not belong on `v5.2` |
| `unitTests/resources/getRecordCount.test.js` ×2 | v5.2's copy predates #2159, so it lacks the `AsyncLocalStorage` call-attribution and `WITHIN_BUDGET_TIME_LIMIT` that the new cases build on | take this PR's file wholesale — it is a superset that carries #2159's CI-timing fix with it |

The dependency this change needs is already present: `v5.2` pins `@harperfast/rocksdb-js` `2.9.1`, the same exact version as `main`, so `createCountEstimator` and `estimateCount` are available. Worth stating explicitly because if they were not, `canEstimate` would be false on RocksDB and the escape would fall through to `primaryStore.getStats().entryCount` — which `RocksDatabase` does not implement — leaving no base, so every describe would full-scan instead of escaping.

### Verification with the patch applied, on `v5.2`

- `npm run build` clean (`harper@5.2.12`), `prettier --check` clean
- `unitTests/resources/getRecordCount.test.js` — 16 passing, 0 failing
- `npm run test:unit:resources` — 1781 passing, 0 failing
- `HARPER_STORAGE_ENGINE=lmdb` over the same suite — 1507 passing, 0 failing

### Patch

Apply to `cherry-pick/v5.2/pr-2612`:

```diff
diff --git a/resources/Table.ts b/resources/Table.ts
index 86ad80217..caee9fb5a 100644
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -139,33 +139,11 @@ const EVICTION_BATCH_SIZE = 100;
 // letting an unbounded number of open transactions (and their snapshots) accumulate.
 const MAX_INFLIGHT_EVICTION_BATCHES = 4;
 const CACHEABLE_STATUS_CODES = new Set([200, 203, 204, 206, 300, 301, 308, 404, 405, 410, 414, 501]);
-<<<<<<< HEAD
-=======
-// Guardrails for `Prefer: count=exact`: once the requested page has been collected, counting the rest
-// of the match set is bounded by BOTH a row cap and a wall-clock budget, so a paginated read can't turn
-// into an unbounded scan. Exceeding either reports an unknown total (Content-Range `.../*`) rather than
-// truncating the page. These bound the count tail, not the page itself; a genuinely expensive query
-// (large filtered full-scan, in-memory sort) should still be gated by config before broad exposure.
-const MAX_EXACT_COUNT_SCAN = 1_000_000;
-const MAX_EXACT_COUNT_MS = 1_000;
-// Largest page a `Prefer: count=` request will materialize. A request whose limit exceeds this (or is
-// not a finite, non-negative integer, e.g. `limit(Infinity)`/`limit(foo)`) falls through to the normal
-// streaming path with no count, so a count request can't be coerced into buffering an unbounded page.
-const MAX_COUNT_PAGE = 10_000;
-// How often the exact-count drain yields to the macrotask queue (must be a power of two for the bit-mask
-// check). Keeps a large scan from monopolizing the event loop without adding a yield per row.
-const COUNT_YIELD_INTERVAL = 2_048;
 // Smallest forward sample `getRecordCount` will extrapolate a record rate from; below it the scan runs
 // to completion and reports an exact count.
 const MIN_ESTIMATOR_SAMPLE = 1_000;
-<<<<<<< HEAD
->>>>>>> 921a635c2 (Size describe record counts with a calibrated estimate instead of a second full key scan)
-=======
 // Budget intervals the forward scan may spend before it must estimate rather than keep scanning.
 const MAX_ESTIMATE_CHECKPOINTS = 20;
-<<<<<<< HEAD
->>>>>>> fb4a43794 (Bound the forward scan and enforce sample disjointness on the estimated base)
-=======
 // A store estimate's `count`, or 0 when the store answered with a shape that cannot be trusted --
 // DESIGN.md's invariant for this API family is that such an answer degrades rather than poisons.
 function usableCount(estimate: any): number {
@@ -173,7 +151,6 @@ function usableCount(estimate: any): number {
 		? estimate.count
 		: 0;
 }
->>>>>>> 5cdd9e5c8 (Restore bounded sampling on LMDB and make the checkpoint-delta test discriminating)
 envMngr.initSync();
 const LMDB_PREFETCH_WRITES = envMngr.get(CONFIG_PARAMS.STORAGE_PREFETCHWRITES);
 const LOCK_TIMEOUT = 10000;
diff --git a/unitTests/resources/getRecordCount.test.js b/unitTests/resources/getRecordCount.test.js
index df62a9fb4..03b14f381 100644
--- a/unitTests/resources/getRecordCount.test.js
+++ b/unitTests/resources/getRecordCount.test.js
@@ -1,12 +1,11 @@
 require('../testUtils');
 const assert = require('node:assert');
+const { AsyncLocalStorage } = require('node:async_hooks');
 const { setupTestDBPath } = require('../testUtils');
 const { table } = require('#src/resources/databases');
 const { setMainIsWorker } = require('#js/server/threads/manageThreads');
 
 describe('Table.getRecordCount', () => {
-<<<<<<< HEAD
-=======
 	// Used instead of the ambient 500ms default in the within-budget cases below, so a contended
 	// CI runner can't blow the budget and flip them into the estimator path.
 	const WITHIN_BUDGET_TIME_LIMIT = Infinity;
@@ -17,7 +16,6 @@ describe('Table.getRecordCount', () => {
 	// mirrors MIN_ESTIMATOR_SAMPLE in resources/Table.ts
 	const MIN_ESTIMATOR_SAMPLE = 1000;
 
->>>>>>> 921a635c2 (Size describe record counts with a calibrated estimate instead of a second full key scan)
 	let RecordCountTable;
 	let EstimatorTable;
 
@@ -63,7 +61,7 @@ describe('Table.getRecordCount', () => {
 	});
 
 	it('returns the exact count when the loop completes within the time budget', async function () {
-		const result = await RecordCountTable.getRecordCount();
+		const result = await RecordCountTable.getRecordCount({ timeLimit: WITHIN_BUDGET_TIME_LIMIT });
 		assert.equal(result.recordCount, 30);
 		assert.equal(result.estimatedRange, undefined);
 	});
@@ -428,33 +426,59 @@ describe('Table.getRecordCount', () => {
 		// a within-budget scan returns the exact count directly and must not pay for it. The source is
 		// engine-dependent -- getKeysCount() (a full key scan) on RocksDB, getStats().entryCount on LMDB
 		// (the resources suite runs under both) -- so spy on whichever the store exposes.
+		// `primaryStore` is shared and the scan yields once per entry, so an absolute call count is not a
+		// count of what this call did: the analytics aggregation timer sweeps every table's getStats()
+		// every half aggregate period (storeRocksDBStatsMetrics, or Table.getSize via
+		// storeTableSizeMetrics under LMDB) and lands inside the window at random.
 		const store = RecordCountTable.primaryStore;
+		const underTest = new AsyncLocalStorage();
 		let calls = 0;
-		const origKeys = typeof store.getKeysCount === 'function' ? store.getKeysCount.bind(store) : undefined;
-		const origStats = typeof store.getStats === 'function' ? store.getStats.bind(store) : undefined;
-		if (origKeys) store.getKeysCount = (...args) => (calls++, origKeys(...args));
-		if (origStats) store.getStats = (...args) => (calls++, origStats(...args));
+		let foreignCalls = 0;
+		const restore = [];
+		for (const name of ['getKeysCount', 'getStats']) {
+			if (typeof store[name] !== 'function') continue;
+			const original = store[name];
+			const owned = Object.hasOwn(store, name);
+			store[name] = function (...args) {
+				if (underTest.getStore()) calls++;
+				else foreignCalls++;
+				return original.apply(this, args);
+			};
+			restore.push(() => {
+				if (owned) store[name] = original;
+				else delete store[name];
+			});
+		}
 		try {
-			const completed = await RecordCountTable.getRecordCount({ timeLimit: 60_000 });
+			// Positive control for the attribution, in the shape of the analytics sweep: queued before the
+			// scan's first `await rest()`, so it runs while the scan is suspended. Contained and reported
+			// here rather than left to escape the immediate, which would abort the run.
+			let foreignError;
+			setImmediate(() => {
+				try {
+					store.getStats();
+				} catch (error) {
+					foreignError = error;
+				}
+			});
+			const completed = await underTest.run(true, () =>
+				RecordCountTable.getRecordCount({ timeLimit: WITHIN_BUDGET_TIME_LIMIT })
+			);
+			assert.equal(foreignError, undefined, 'the interleaved foreign call must not throw');
 			assert.equal(completed.recordCount, 30);
 			assert.equal(completed.estimatedRange, undefined);
+			assert.ok(foreignCalls >= 1, 'the interleaved foreign call must reach the spy, or it proves nothing');
 			assert.equal(calls, 0, 'entry-count source should not be consulted when the scan finishes within budget');
 
 			calls = 0;
-<<<<<<< HEAD
-			await RecordCountTable.getRecordCount({ timeLimit: -1 }); // force timeout -> escape to the exact count
-			assert.ok(calls >= 1, 'entry-count source should be consulted once the scan exceeds the budget');
-=======
 			// Below the sample floor the budget no longer matters: the scan runs to an exact count and
 			// still consults nothing, which is what keeps a small churned table off an estimated base.
 			const belowFloor = await underTest.run(true, () => RecordCountTable.getRecordCount({ timeLimit: -1 }));
 			assert.equal(belowFloor.recordCount, 30);
 			assert.equal(belowFloor.estimatedRange, undefined);
 			assert.equal(calls, 0, 'a table below the sample floor must not consult the entry-count source');
->>>>>>> 921a635c2 (Size describe record counts with a calibrated estimate instead of a second full key scan)
 		} finally {
-			if (origKeys) store.getKeysCount = origKeys;
-			if (origStats) store.getStats = origStats;
+			for (const undo of restore) undo();
 		}
 	});
 });
```

Note this branch is a preview: on merge the workflow re-picks from the final squash SHA, so the same three conflicts will recur once and need the same resolution.

— Claude Opus 5

cb1kenobi (commented):
A zero remainder estimate can collapse the published range even when many live records remain. Reuse the whole-store estimate as a conservative fallback for the remainder.

—
Reviewed 68f8863c

resources/Table.ts
```diff
@@ -6136,82 +6147,149 @@ export function makeTable(options) {
...
+					entryCount > 0 &&
+					(checkpoints >= MAX_ESTIMATE_CHECKPOINTS || entriesScanned < Math.floor(entryCount / 2))
+				) {
+					if (canEstimate) {
+						try {
+							const remaining = primaryStore.estimateCount({ start: lastKey, exclusiveStart: true });
+							// widened by its own reported untrustworthiness: block-granular, so it can land below
+							// the live count it is meant to bound
+							const remainingCount = usableCount(remaining);
+							remainderPhysical = remainingCount > 0 ? remainingCount * (2 - remaining.confidence) : 0;
```
cb1kenobi, line 6232: **Zero remainder estimate collapses the result — Severity: High.** RocksDB range estimates can validly return zero for live entries still in the memtable, as handled by the whole-store fallback above. If the prefix estimator returns any positive underestimate, that fallback is skipped; a zero remainder then makes `baseMax` depend only on the sampled ends and can report thousands of records as roughly `2 * limit`, with a degenerate range excluding the remaining live rows. When the remainder estimate is zero or unusable, use `getEstimatedKeyCount()` as a conservative floor for the total physical base (or derive the remainder from it).

—
Reviewed 68f8863c

cb1kenobi (commented):
The estimator work is sound and well covered by tests, but the new checkpoint ceiling lets the forward scan run up to 20 budget intervals, and the reverse sample is still sized to whatever the forward scan covered. On a large table with head-concentrated churn (the shape the PR itself measures) that makes describe_table cost roughly 20 seconds where main answered in about 2. Capping the reverse sample independently of the forward scan fixes it without touching the ceiling. Separately, the Unit Test job is currently failing on Node 22, 24, and 26 and needs resolving before merge.

—
Reviewed 68f8863c

resources/Table.ts
```diff
@@ -6136,82 +6147,149 @@ export function makeTable(options) {
...
+						} catch {
+							remainderPhysical = 0;
+						}
 					}
-					// Past the halfway point already: finishing the scan for an exact count is cheaper
-					// than estimating. Set the flag so we stop re-evaluating the budget on each remaining iteration.
-					completeForExact = true;
+					limit = entriesScanned;
```
cb1kenobi, line 6248: **Reverse sample is sized by the forward scan, which the checkpoint ceiling now lets run 20 budget intervals — Severity: High**

`limit = entriesScanned` sizes the reverse sample at whatever the forward scan happened to cover. Before this change the forward scan always broke at the first checkpoint — the exact `entryCount` made the halfway test reliable — so `limit` was one budget interval's worth of entries. With `MAX_ESTIMATE_CHECKPOINTS`, a base that undershoots (head-concentrated churn: the shape this PR's own fixture measures at 2038 for 3000 live rows, i.e. the ceiling fired because halfway never did) holds `entriesScanned < Math.floor(entryCount / 2)` false at every checkpoint, so the forward scan runs the full 20 intervals — ~10s at the default 500ms `TIME_LIMIT` — and the reverse sample is then sized to match it.

**Failure mode:** a 10M-key RocksDB table with head-concentrated churn. The forward scan covers ~10s of entries, the two samples are nowhere near meeting (`sampledWholeTable` requires each sample to cover half the table), so the reverse loop reads that same count again: ~20s inside `describe_table`, against ~2s on `main` (one 500ms interval plus a 0.93s `getKeysCount()`). `describe_all` pays it per table. This is the one shape where the change is slower than the code it replaces, on the path it exists to bound.

**Simplest fix:** decouple the reverse sample from the forward scan's length instead of retuning the ceiling.

```ts
const reverseLimit = Math.min(limit, MIN_ESTIMATOR_SAMPLE);
```

Use `reverseLimit` for the reverse `getRange`'s `limit` option and for the `reverseScanned >= reverseLimit` break. `recordRate` already divides by `limit + reverseScanned` and `baseMin` sums the same actual counts, so both stay correct with an asymmetric sample; the `compareKeys(key, firstKey) <= 0` meeting check and the exact-count return are unaffected.

—
Reviewed 68f8863c

resources/Table.ts
```diff
@@ -212,6 +212,18 @@ const MAX_COUNT_PAGE = 10_000;
...
+const MIN_ESTIMATOR_SAMPLE = 1_000;
+// Budget intervals the forward scan may spend before it must estimate rather than keep scanning.
+const MAX_ESTIMATE_CHECKPOINTS = 20;
+// A store estimate's `count`, or 0 when the store answered with a shape that cannot be trusted --
+// DESIGN.md's invariant for this API family is that such an answer degrades rather than poisons.
+function usableCount(estimate: any): number {
+	return Number.isFinite(estimate?.count) && estimate.count >= 0 && estimate.confidence >= 0 && estimate.confidence <= 1
+		? estimate.count
+		: 0;
+}
```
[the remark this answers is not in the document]
kriszyp: Fixed in c3a363932 — `usableCount` now checks `Number.isFinite(confidence)` alongside the range, so `confidence: null` no longer coerces through. Verified with a stubbed estimator returning `{count: 30000, confidence: null}` against a 3,000-row table: 30,000 reported before, ~3,000 after; regression test added. I kept `Number.isFinite` on `count` rather than the suggested `== null` check, since `count: "5"` and `count: Infinity` should be rejected too and a nullish check would pass them.

resources/Table.ts
```diff
@@ -212,6 +212,18 @@ const MAX_COUNT_PAGE = 10_000;
...
+// Smallest forward sample `getRecordCount` will extrapolate a record rate from; below it the scan runs
+// to completion and reports an exact count.
+const MIN_ESTIMATOR_SAMPLE = 1_000;
+// Budget intervals the forward scan may spend before it must estimate rather than keep scanning.
+const MAX_ESTIMATE_CHECKPOINTS = 20;
+// A store estimate's `count`, or 0 when the store answered with a shape that cannot be trusted --
+// DESIGN.md's invariant for this API family is that such an answer degrades rather than poisons.
+function usableCount(estimate: any): number {
+	return Number.isFinite(estimate?.count) && estimate.count >= 0 && estimate.confidence >= 0 && estimate.confidence <= 1
+		? estimate.count
```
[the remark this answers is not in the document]
kriszyp: Fixed in c3a363932 — mirrored `resources/search.ts:1280` exactly: `!Number.isFinite(count) || count < 0 || !Number.isFinite(confidence) || confidence < 0 || confidence > 1`. As you noted, that closes the direct `remaining.confidence` use too, since the remainder probe only multiplies when `usableCount` returned non-zero. Regression test added: a stubbed `{count: 30000, confidence: null}` reported 30,000 for a 3,000-row table before the fix.

resources/Table.ts
```diff
@@ -6136,82 +6147,149 @@ export function makeTable(options) {
...
+						logger.debug?.('Count estimator unavailable, falling back to an exact scan', error);
+						estimatorFailed = true;
+						estimator = undefined;
+						entryCount = 0;
 					}
-					if (entriesScanned < halfway) {
-						// it is taking too long, so we will just take this sample and a sample from the end to estimate
-						limit = entriesScanned;
-						break;
+				} else if (!canEstimate) entryCount = primaryStore.getStats().entryCount;
```
cb1kenobi: **`getStats()` is the one unguarded store call on the degrade path — Severity: Medium**

Every estimator call in this block is feature-gated or try/caught so an unexpected store degrades to the exact scan rather than poisoning the result. `primaryStore.getStats()` here is not. `canEstimate` is false for any store missing `createCountEstimator`/`estimateCount`, which is not the same set as "LMDB" — a RocksDB store whose native module predates those methods (version-skewed `@harperfast/rocksdb-js`, or a release line that hasn't picked up the 2.9.x bump) lands here too. `main` branched on `isRocksDB` precisely to avoid calling `getStats()` on a RocksDB store, and the v5.2 cherry-pick analysis on this PR states `RocksDatabase` does not implement it.

**Failure mode:** `describe_table` / `describe_all` throws `TypeError: primaryStore.getStats is not a function` out of the operations API at the first checkpoint past `MIN_ESTIMATOR_SAMPLE`, instead of degrading to the exact scan the comment four lines below promises. In the softer variant — `getStats()` exists but returns no `entryCount` — `entryCount` becomes `undefined`, `entryCount > 0` is false at every checkpoint, and describe silently walks the whole table (the outcome this PR exists to remove) with no `warnedNoBase` log, because that branch is gated on `canEstimate`.

**Simplest fix:** apply the same shape check used for `getEstimatedKeyCount` a few lines down.

```ts
} else if (!canEstimate) {
	const stats = typeof primaryStore.getStats === 'function' ? primaryStore.getStats() : undefined;
	entryCount = Number.isFinite(stats?.entryCount) && stats.entryCount > 0 ? stats.entryCount : 0;
}
```

Widening the `!entryCount` no-base warning to cover this branch as well would also mean a store with no usable base logs once instead of quietly full-scanning.

—
Reviewed 6dc0ab0f

cb1kenobi (commented):
The new checkpoint block guards every estimator call but calls primaryStore.getStats() unguarded when a store lacks the estimator API, so a RocksDB store without those native methods throws out of describe_table instead of degrading to the exact scan it promises. Applying the same shape check used for the other store calls fixes it in one line. Separately, the two High findings already open from the human reviewer are unaddressed by this commit and Integration Tests 3/6 (Bun) is failing.

—
Reviewed 6dc0ab0f

resources/Table.ts
kriszyp: Good catch, and it reproduced exactly as you described. Fixed in f241196b9 — the remainder now falls back to `getEstimatedKeyCount()` when the range estimate is zero or unusable, as you suggested.

Measured before the fix with a stubbed `{count: 2100}` prefix estimate and a zero remainder against a 3,000-row table: `record_count` 2100, range `[2000, 2100]` — excluding the 3,000 live rows, while the whole-store property reported 3,000 the whole time. Regression test added and verified to fail without the fallback.

resources/Table.ts
kriszyp: You are right, and this was the more serious of the two — the ceiling I added to bound the forward scan quietly doubled the call by sizing the tail sample to match it. Fixed in f241196b9: the reverse sample is now capped at `MIN_ESTIMATOR_SAMPLE` independently of `entriesScanned`, which also restores the "first thousand and last thousand" the surrounding comment always claimed.

The pooled rate does not require equal samples, and disjointness is still enforced against the forward scan last key rather than by size, so nothing else in the math changes. Regression test asserts the requested reverse limit stays capped no matter how far the forward scan ran (it requested 1019 before the fix).

Thanks for tracing it through to the wall-clock number — that framing is what made the severity obvious.

resources/Table.ts
kriszyp: Right on both variants, and the `canEstimate` != LMDB distinction is the part I had missed — I checked that v5.2 pins 2.9.1 and concluded the risk was gone, rather than making the code degrade regardless of pin. Fixed in e1ca0ef09: the statistics branch now applies the same shape check as the estimator calls.

New test strips `createCountEstimator`, `estimateCount` and `getStats` from a real store and asserts an exact count comes back — it raises exactly the `TypeError: primaryStore.getStats is not a function` you described without the guard.

For the other two: both were fixed in f241196b9 (which landed after the 6dc0ab0f you reviewed) — zero remainder now falls back to `getEstimatedKeyCount()`, and the reverse sample is capped at `MIN_ESTIMATOR_SAMPLE` independently of the forward scan. Replies on those threads have the measurements.

On Integration Tests 3/6: that shard is failing on `main` as well, same test and line (`integrationTests/server/qa702-sse-event-data.test.ts:423`) across all four runtimes, so this branch inherited it through the merge rather than causing it.

kriszyp:
## Updated resolution for the `v5.2` cherry-pick — supersedes my earlier patch

The earlier patch was generated against `68f8863c`; five commits have landed since (the fixture-cost cut plus four review fixes), so the cherry-pick branch was recreated at `8fa0b3444` and that patch no longer applies. This one is regenerated against the current branch and re-verified on `v5.2`.

The conflict shape is unchanged — same three sites, same resolutions:

| conflict | cause | resolution |
|---|---|---|
| `resources/Table.ts` constants | `MIN_ESTIMATOR_SAMPLE` / `MAX_ESTIMATE_CHECKPOINTS` / `usableCount` sit next to the `Prefer: count=` guardrails (`MAX_EXACT_COUNT_SCAN`, `MAX_COUNT_PAGE`, `COUNT_YIELD_INTERVAL`), which exist only on `main` | keep the three additions, drop the `Prefer: count=` block — not part of this change and not on `v5.2` |
| `unitTests/resources/getRecordCount.test.js` ×2 | v5.2's copy predates #2159, so it lacks the `AsyncLocalStorage` call-attribution and `WITHIN_BUDGET_TIME_LIMIT` the cases build on | take this PR's file wholesale — a superset that carries #2159's CI-timing fix with it |

`getRecordCount` itself still applies cleanly: v5.2's copy of that method is byte-identical to this PR's merge-base.

`v5.2` pins `@harperfast/rocksdb-js` `2.9.1`, the same exact version as `main`, so `createCountEstimator` and `estimateCount` are present. Note that as of `e1ca0ef09` the code no longer depends on that being true — the statistics branch degrades to an exact scan on a store lacking the estimator API rather than throwing — but the fast path does want the 2.9.x pin.

### Verification with this patch applied, on `v5.2`

- `npm run build` clean (`harper@5.2.12`), `prettier --check` clean
- `unitTests/resources/getRecordCount.test.js` — 20 passing, 0 failing
- `npm run test:unit:resources` — 1785 passing, 0 failing
- `HARPER_STORAGE_ENGINE=lmdb` over the same suite — 1508 passing, 0 failing

### Patch

Apply to `cherry-pick/v5.2/pr-2612` at `8fa0b3444`:

```diff
diff --git a/resources/Table.ts b/resources/Table.ts
index d03642c33..eeaef928b 100644
--- a/resources/Table.ts
+++ b/resources/Table.ts
@@ -139,33 +139,11 @@ const EVICTION_BATCH_SIZE = 100;
 // letting an unbounded number of open transactions (and their snapshots) accumulate.
 const MAX_INFLIGHT_EVICTION_BATCHES = 4;
 const CACHEABLE_STATUS_CODES = new Set([200, 203, 204, 206, 300, 301, 308, 404, 405, 410, 414, 501]);
-<<<<<<< HEAD
-=======
-// Guardrails for `Prefer: count=exact`: once the requested page has been collected, counting the rest
-// of the match set is bounded by BOTH a row cap and a wall-clock budget, so a paginated read can't turn
-// into an unbounded scan. Exceeding either reports an unknown total (Content-Range `.../*`) rather than
-// truncating the page. These bound the count tail, not the page itself; a genuinely expensive query
-// (large filtered full-scan, in-memory sort) should still be gated by config before broad exposure.
-const MAX_EXACT_COUNT_SCAN = 1_000_000;
-const MAX_EXACT_COUNT_MS = 1_000;
-// Largest page a `Prefer: count=` request will materialize. A request whose limit exceeds this (or is
-// not a finite, non-negative integer, e.g. `limit(Infinity)`/`limit(foo)`) falls through to the normal
-// streaming path with no count, so a count request can't be coerced into buffering an unbounded page.
-const MAX_COUNT_PAGE = 10_000;
-// How often the exact-count drain yields to the macrotask queue (must be a power of two for the bit-mask
-// check). Keeps a large scan from monopolizing the event loop without adding a yield per row.
-const COUNT_YIELD_INTERVAL = 2_048;
 // Smallest forward sample `getRecordCount` will extrapolate a record rate from; below it the scan runs
 // to completion and reports an exact count.
 const MIN_ESTIMATOR_SAMPLE = 1_000;
-<<<<<<< HEAD
->>>>>>> 921a635c2 (Size describe record counts with a calibrated estimate instead of a second full key scan)
-=======
 // Budget intervals the forward scan may spend before it must estimate rather than keep scanning.
 const MAX_ESTIMATE_CHECKPOINTS = 20;
-<<<<<<< HEAD
->>>>>>> fb4a43794 (Bound the forward scan and enforce sample disjointness on the estimated base)
-=======
 // A store estimate's `count`, or 0 when the store answered with a shape that cannot be trusted --
 // DESIGN.md's invariant for this API family is that such an answer degrades rather than poisons.
 function usableCount(estimate: any): number {
@@ -175,7 +153,6 @@ function usableCount(estimate: any): number {
 		return 0;
 	return count;
 }
->>>>>>> 5cdd9e5c8 (Restore bounded sampling on LMDB and make the checkpoint-delta test discriminating)
 envMngr.initSync();
 const LMDB_PREFETCH_WRITES = envMngr.get(CONFIG_PARAMS.STORAGE_PREFETCHWRITES);
 const LOCK_TIMEOUT = 10000;
diff --git a/unitTests/resources/getRecordCount.test.js b/unitTests/resources/getRecordCount.test.js
index 9f82d9eb5..47a515781 100644
--- a/unitTests/resources/getRecordCount.test.js
+++ b/unitTests/resources/getRecordCount.test.js
@@ -1,12 +1,11 @@
 require('../testUtils');
 const assert = require('node:assert');
+const { AsyncLocalStorage } = require('node:async_hooks');
 const { setupTestDBPath } = require('../testUtils');
 const { table } = require('#src/resources/databases');
 const { setMainIsWorker } = require('#js/server/threads/manageThreads');
 
 describe('Table.getRecordCount', () => {
-<<<<<<< HEAD
-=======
 	// Used instead of the ambient 500ms default in the within-budget cases below, so a contended
 	// CI runner can't blow the budget and flip them into the estimator path.
 	const WITHIN_BUDGET_TIME_LIMIT = Infinity;
@@ -21,7 +20,6 @@ describe('Table.getRecordCount', () => {
 	// estimate stops climbing well before this), and the suite's CI step has under a minute of slack.
 	const CHURN_ROUNDS = 5;
 
->>>>>>> 921a635c2 (Size describe record counts with a calibrated estimate instead of a second full key scan)
 	let RecordCountTable;
 	let EstimatorTable;
 
@@ -67,7 +65,7 @@ describe('Table.getRecordCount', () => {
 	});
 
 	it('returns the exact count when the loop completes within the time budget', async function () {
-		const result = await RecordCountTable.getRecordCount();
+		const result = await RecordCountTable.getRecordCount({ timeLimit: WITHIN_BUDGET_TIME_LIMIT });
 		assert.equal(result.recordCount, 30);
 		assert.equal(result.estimatedRange, undefined);
 	});
@@ -546,33 +544,59 @@ describe('Table.getRecordCount', () => {
 		// a within-budget scan returns the exact count directly and must not pay for it. The source is
 		// engine-dependent -- getKeysCount() (a full key scan) on RocksDB, getStats().entryCount on LMDB
 		// (the resources suite runs under both) -- so spy on whichever the store exposes.
+		// `primaryStore` is shared and the scan yields once per entry, so an absolute call count is not a
+		// count of what this call did: the analytics aggregation timer sweeps every table's getStats()
+		// every half aggregate period (storeRocksDBStatsMetrics, or Table.getSize via
+		// storeTableSizeMetrics under LMDB) and lands inside the window at random.
 		const store = RecordCountTable.primaryStore;
+		const underTest = new AsyncLocalStorage();
 		let calls = 0;
-		const origKeys = typeof store.getKeysCount === 'function' ? store.getKeysCount.bind(store) : undefined;
-		const origStats = typeof store.getStats === 'function' ? store.getStats.bind(store) : undefined;
-		if (origKeys) store.getKeysCount = (...args) => (calls++, origKeys(...args));
-		if (origStats) store.getStats = (...args) => (calls++, origStats(...args));
+		let foreignCalls = 0;
+		const restore = [];
+		for (const name of ['getKeysCount', 'getStats']) {
+			if (typeof store[name] !== 'function') continue;
+			const original = store[name];
+			const owned = Object.hasOwn(store, name);
+			store[name] = function (...args) {
+				if (underTest.getStore()) calls++;
+				else foreignCalls++;
+				return original.apply(this, args);
+			};
+			restore.push(() => {
+				if (owned) store[name] = original;
+				else delete store[name];
+			});
+		}
 		try {
-			const completed = await RecordCountTable.getRecordCount({ timeLimit: 60_000 });
+			// Positive control for the attribution, in the shape of the analytics sweep: queued before the
+			// scan's first `await rest()`, so it runs while the scan is suspended. Contained and reported
+			// here rather than left to escape the immediate, which would abort the run.
+			let foreignError;
+			setImmediate(() => {
+				try {
+					store.getStats();
+				} catch (error) {
+					foreignError = error;
+				}
+			});
+			const completed = await underTest.run(true, () =>
+				RecordCountTable.getRecordCount({ timeLimit: WITHIN_BUDGET_TIME_LIMIT })
+			);
+			assert.equal(foreignError, undefined, 'the interleaved foreign call must not throw');
 			assert.equal(completed.recordCount, 30);
 			assert.equal(completed.estimatedRange, undefined);
+			assert.ok(foreignCalls >= 1, 'the interleaved foreign call must reach the spy, or it proves nothing');
 			assert.equal(calls, 0, 'entry-count source should not be consulted when the scan finishes within budget');
 
 			calls = 0;
-<<<<<<< HEAD
-			await RecordCountTable.getRecordCount({ timeLimit: -1 }); // force timeout -> escape to the exact count
-			assert.ok(calls >= 1, 'entry-count source should be consulted once the scan exceeds the budget');
-=======
 			// Below the sample floor the budget no longer matters: the scan runs to an exact count and
 			// still consults nothing, which is what keeps a small churned table off an estimated base.
 			const belowFloor = await underTest.run(true, () => RecordCountTable.getRecordCount({ timeLimit: -1 }));
 			assert.equal(belowFloor.recordCount, 30);
 			assert.equal(belowFloor.estimatedRange, undefined);
 			assert.equal(calls, 0, 'a table below the sample floor must not consult the entry-count source');
->>>>>>> 921a635c2 (Size describe record counts with a calibrated estimate instead of a second full key scan)
 		} finally {
-			if (origKeys) store.getKeysCount = origKeys;
-			if (origStats) store.getStats = origStats;
+			for (const undo of restore) undo();
 		}
 	});
 });
```

This branch remains a preview: on merge the workflow re-picks from the final squash SHA, so the same three conflicts recur once and take the same resolution.

— Claude Opus 5

cb1kenobi (commented):
No confirmed blocking defects remain on the changed lines. Previously raised issues are fixed at this commit, and the remaining CI failures are reported as inherited from main.

—
Reviewed e1ca0ef0

cb1kenobi (commented):
No blocking problems found on the changed lines at this commit. The estimator path now guards, shape-validates and try/catches every store call, so an unexpected store falls back to the exact scan instead of throwing or publishing a bad count, and the published range cannot invert or exclude its own reported count. The three earlier findings (confidence validation, zero remainder, reverse-sample sizing) are fixed here and were not re-raised. One low-confidence test-hygiene note about stubbing getStats to undefined on a shared store is mentioned in the review body but is not blocking.

—
Reviewed e1ca0ef0

cb1kenobi (commented):
No confirmed blocking defects remain on changed lines. Earlier findings are fixed at this commit, and all reported checks pass.

—
Reviewed 0767e061

Issue #2513, opened 2026-09-05
Add @fullText schema declaration, validation, activation, and introspection


kylebernhardy:
## Outcome

Add a first-class schema declaration for a full-text field derived from record attributes, following Harper embed-field conventions where they apply.

## Acceptance criteria

- `@fullText` declares weighted source fields and English analyzer settings without overloading `@indexed` or `@computed`.
- Positions and surface terms are configurable; agreed defaults and feature consequences are validated.
- Synonym expansion and highlighting storage are configurable and disabled by default.
- Invalid analyzers, source fields, weights, limits, and capability combinations fail schema activation clearly.
- Multiple full-text indexes per table are supported.
- RocksDB capability and wrapper compatibility preflight occurs before activation.
- LMDB rejects the index type with an actionable error.
- Introspection and generated API metadata describe the declared index and enabled capabilities.

Comment generated by kAIle (Codex GPT-5)

PR #2615, opened 2026-09-15
Add @fullText schema declaration and validation

kylebernhardy:
Adds Harper's schema and catalog contract for native full-text indexes. A table declares one or more named indexes with repeatable `@fullText` directives. Declarations are table metadata, not record attributes, and are stored on the table's canonical primary catalog descriptor.

This is the schema layer of [Add @fullText schema declaration, validation, activation, and introspection](https://github.com/HarperFast/harper/issues/2513), stacked on [Add a full-text derived-index backend](https://github.com/HarperFast/harper/pull/2569). Runtime attachment and query execution remain in [Activate native full-text derived indexes](https://github.com/HarperFast/harper/pull/2616). The feature must remain release-gated until the runtime and harper-pro schema-wire follow-up land.

## Design

- `@fullText(name:, fields:)` is repeatable on `@table` types. Index names occupy a table-level namespace and never enter `Table.attributes`, record validation, storage projection, or generated record schemas.
- Each declaration compiles to canonical metadata with weighted `String` and `[String]` sources, the versioned `english@1` analyzer, configurable positions and surface terms, synonyms, and opt-in highlighting.
- Local declarations require audit logging because the transaction log is the recovery source. GraphQL is the initial authoring surface; `defineTable` and legacy operations are unchanged.
- Declarations persist only on the primary catalog descriptor. Source-row and primary-row writes are ordered so every crash prefix leaves declarations absent or valid. Safe partial catalog progress is allowed when a schema update fails.
- Cluster updates merge additively by index name against durable catalog state. Local definitions win conflicts. Invalid persisted or peer declarations are isolated per index so valid siblings remain available.
- Catalog reloads reuse validated declarations and refresh existing worker classes. A missing `fullTextIndexes` field on an existing descriptor is an authoritative empty list, preventing stale workers from resurrecting removed indexes.

## Implementation

- [`schema.graphql`](https://github.com/HarperFast/harper/pull/2615/changes?w=1#diff-efc1675187620595b0844197a25c35eac9b6df752f9182e5cffddee6f27a8489R225) defines the table directive and configuration inputs. [`resources/graphql.ts`](https://github.com/HarperFast/harper/pull/2615/changes?w=1#diff-72e8b09df8fa4bd777558edfcb08b247ac3965b75d83c9c8c6c7f81cbe2374c8R294) parses repeatable declarations and waits for schema propagation.
- [`resources/fullTextSchema.ts`](https://github.com/HarperFast/harper/pull/2615/changes?w=1#diff-443a94868425b7453a14ef9b58b5d12f07348641da574dfdc64de9ff402758c9R1) owns strict authoring validation, canonical defaults, deterministic ordering, and tolerant per-definition catalog recovery.
- [`resources/databases.ts`](https://github.com/HarperFast/harper/pull/2615/changes?w=1#diff-2a3520ef673512c350042faa431f24956f148f2e19889470bd00a79b5da339c5R2714) owns persistence, durable audit checks, source-removal guards, crash ordering, worker reloads, and local-wins peer reconciliation.
- [`resources/Table.ts`](https://github.com/HarperFast/harper/pull/2615/changes?w=1#diff-4ff51263bd99981165fc179149c3195ea0c0ca3e2f360bafee2a6bb356964377R6444) exposes `Table.fullTextIndexes`, forwards additive peer declarations, suppresses repeated rejected-state churn, and preserves generic computed and relationship accessor cleanup.
- The namespace and ownership contract is documented in [`docs/full-text-index-namespace.md`](https://github.com/HarperFast/harper/pull/2615/changes?w=1#diff-c2a0cc70567d86081515e1c4817830697ee79d61ac05c1f8cd174aad28de4693R1), with supporting notes in [`DESIGN.md`](https://github.com/HarperFast/harper/pull/2615/changes?w=1#diff-3dc5dd454e080eb849ee5efaf79df2585fbe0a06804d69249b4c81da05a63875R2815) and [`resources/DESIGN.md`](https://github.com/HarperFast/harper/pull/2615/changes?w=1#diff-5a3cac25d02c21deca87f353afcbe03c11dd1601a64cfb947c57219f7334a708R31).

## Verification

- `npm run build` passed.
- Changed-file Oxlint, Prettier, and `git diff --check` passed.
- Focused RocksDB tests: 51 passing, 1 LMDB-specific pending.
- Focused LMDB tests: 49 passing, 3 RocksDB-only pending.
- Full resource suite: 2,836 passing and 46 pending. Its three failures are the same condition-delete and read-your-writes failures reproduced on the parent branch; no new failure was introduced.
- [`unitTests/resources/fullTextSchema.test.js`](https://github.com/HarperFast/harper/pull/2615/changes?w=1#diff-262ffbba697a911e19eceece3e94ac194576fd0d0942e1072905ad6ea3e77f51R1) covers restart/reopen recovery, durable removals, malformed metadata, audit authority, accessor cleanup, and RocksDB crash points. [`unitTests/resources/clusterSchemaMerge.test.js`](https://github.com/HarperFast/harper/pull/2615/changes?w=1#diff-20b444841b3df86be516d50126be68e0d1d789cc41b0c393b72fca2b2cfb710cR424) covers stale-worker merges, peer conflicts, durable removals, and hidden-only changes. [`unitTests/resources/fullTextDerivedIndexBackend.test.js`](https://github.com/HarperFast/harper/pull/2615/changes?w=1#diff-c1dca03b72189850d38e9145fd96d98815865418cbbd6017623c281804858f9fR138) pins projected-key encoding behavior.
- Claude and Cursor completed repository-aware review with Harper adjudication. Direct Gemini CLI review of the adjudication fixes returned clean.

Complexity: complicated

<sub>Review-Coverage: authored=codex; ran=claude,cursor-grok,cursor-composer; adjudicated=domain; blocked=gemini(auth); rounds=22; full=10 @ a9673e9b4a04</sub>

<sub>Human-Review-Need: 4 @ a9673e9b4a04</sub>


schema.graphql
```diff
@@ -216,6 +216,45 @@ directive @indexed(
...
+"""
+input FullTextHighlighting {
+	maxFragments: Int! = 3
+	fragmentLength: Int! = 160
+}
+
+"""
+Create a native full-text derived index over the listed source attributes.
+"""
+directive @fullText(
```
kriszyp, line 247: This directive publishes `synonyms`, `highlighting`, and per-field `highlight` (validated in detail in `resources/fullTextSchema.ts` and persisted to the primary catalog row), but nothing in the repo reads them: `Table.fullTextIndexes` has no consumer yet, and `NativeFullTextIndexConfiguration` in `resources/indexes/fullTextNativeBinding.ts:8-22` only has `fields{name,weight}`, `analyzer`, `stopWords`, `positions`, `surfaceTerms`.

Concrete outcome today: a user writes `@fullText(name: "search", fields: [{ name: "title" }], highlighting: { maxFragments: 2 })`, gets no error and no warning, and no highlighting. Once these are accepted and persisted they become a compatibility commitment (the persisted shape is what `readPersistedFullTextDefinitions` re-validates on every load, and `assertKnownKeys` rejects anything it does not recognise), so changing their semantics later means a catalog migration.

Suggest trimming the directive and `FullTextDefinition` to exactly what the native configuration accepts, and adding synonym/highlight options in the PR that implements query-side behaviour. If you want to keep the SDL as a forward declaration, at least reject the unimplemented options with a clear error rather than persisting them.

resources/fullTextSchemaLifecycle.ts
```diff
@@ -0,0 +1,109 @@
...
+	if (!Array.isArray(incomingValues)) {
+		warn('Ignoring peer @fullText declarations because the incoming value is not a list');
+		return {
+			values: persistedValues,
+			definitions: readPersistedFullTextDefinitions(persistedValues, attributes, warn),
+			changed: false,
+		};
+	}
+
+	const merged = readPersistedFullTextDefinitions(persistedValues ?? [], attributes, warn);
```
kriszyp, line 77: `mergePeerFullTextDefinitions` rebuilds `values` from only the definitions this node could compile, then reports `changed` against the raw persisted list. Combined with `assertKnownKeys` rejecting any unrecognised option, this is not forward-compatible:

1. Node A (newer build) persists a definition with a future option.
2. Any `define_schema` event with a `fullTextIndexes` list reaches node B (this build).
3. `readPersistedFullTextDefinitions` warns and drops A's definition, `changed` is true, and the pending block in `resources/databases.ts:3465` rewrites B's primary row without it.

B could not run that index anyway, but it has now erased the declaration from its own catalog, so upgrading B does not bring it back until the schema is redeployed. The malformed-metadata test relies on this same behaviour as a "heal", so it is deliberate for garbage, but it also applies to valid-but-newer entries.

Suggest carrying unparseable persisted entries through in `values` (exclude them from `definitions` only), or treating unknown keys on *persisted* values as a warning while keeping them strict for local declarations. That keeps "local schema is authoritative" without letting the peer path rewrite state it cannot interpret.

resources/Table.ts
```diff
@@ -1387,16 +1390,19 @@ export function makeTable(options) {
 												hasChanges = true;
 											}
 										}
-										if (hasChanges) {
-											table({
+										if (hasChanges || event.fullTextIndexes !== undefined) {
```
kriszyp, line 1393: Two things here.

1. `event.fullTextIndexes !== undefined` now sends every `define_schema` event that carries the field through `table({ origin: 'cluster' })`, even `[]` with nothing persisted and no attribute changes. That path always takes the RocksDB cross-thread update-attributes lock (`resources/databases.ts:2716`) and runs the attribute merge/reconcile. Because the cluster path is additive-only, an empty incoming list can never change anything, so gating on a non-empty list (and letting a non-array through only if you want the warning) avoids a lock per schema event if the sender attaches the field unconditionally.

2. The `define_schema` event is produced outside this repo (harper-pro replication). Is there a companion change that attaches `fullTextIndexes` to the event? Without it this branch is dead code and full-text declarations only converge via per-node schema deploys, which is fine per `resources/DESIGN.md:655` but worth stating in the PR.

resources/fullTextSchemaLifecycle.ts
```diff
@@ -0,0 +1,109 @@
...
+	warn: FullTextWarning
+): FullTextDefinition[] {
+	if (values === undefined) return [];
+	return compileValidFullTextDefinitions(values, attributes, (value, error) => {
+		const name = typeof (value as any)?.name === 'string' ? ` '${(value as any).name}'` : '';
+		warn(`Ignoring invalid persisted @fullText declaration${name}: ${error.message}`);
+	});
+}
+
+/**
```
kriszyp, line 37: The primary catalog row gains a new field (`fullTextIndexes`) with non-obvious rules: interim write before attribute rows, final write after, audit pinned on first local declaration, LMDB keeps peer metadata durable but inactive, peer merge is additive by name. The neighbouring `relationships` field has a section in `resources/DESIGN.md` (around line 313) describing exactly this kind of ordering; this feature should get one too so the next person touching `declareTable` knows why the primary row is written twice.

resources/databases.ts
```diff
@@ -2684,6 +2884,7 @@ function declareTable<TableResourceType>(target: TableTarget, tableDefinition: T
 					);
 				if (existingPrimary) attributes = [existingPrimary, ...attributes];
 			}
+			armFullTextLiveStateRestore?.();
```
kriszyp, line 2887: Nit on the rollback: `armFullTextLiveStateRestore` is called here for every declaration that entered the full-text block, including cluster-origin and retain-only calls that never wrote full-text state, so any later throw (index open, HNSW validation) now rebuilds live attributes from the catalog where it previously left them as-is. That is arguably an improvement, but the restore also replaces `Table.properties` with `projectAttributesToProperties(restored)` (line 2755) rather than the GraphQL-supplied value that was live before, because `originalMetadata` (line 2740) does not capture it. Capturing `Table.properties` alongside the other metadata and restoring it when the restored attribute set matches `originalAttributes` keeps MCP/OpenAPI output identical after a failed reload.

kriszyp (approved):
This looks good to me, although at this point I'm a little confused about the ordering of PRs and what is remaining?

PR #2618, opened 2026-09-15
test(server): align the SSE throw-path legs with #2614's structured stream-error contract

kriszyp:
`main` went green→red at [`91b58a124`](https://github.com/HarperFast/harper/commit/91b58a124e0366c293a5c92c8dca60e52b2846ea) — [Return structured errors from streaming REST responses (#2614)](https://github.com/HarperFast/harper/pull/2614) — with **every** leg of `Integration Tests 3/6` failing (Node 22/24/26, Bun, uWS). That PR made a generator that rejects after the stream has committed to a 200 emit a terminal, format-valid `harper-error` SSE frame and then end the stream cleanly, instead of destroying the response mid-transfer. It updated the two suites whose subject is the throw path (`sse-throw-midstream.test.ts`, `stream-error-contract.test.ts`) but not the two suites that carry a throw-path leg as a secondary case. Both count `data:` lines, so the new error frame read as a third data event:

```
data: {"n":0}

data: {"n":1}

event: harper-error
data: {"error":"Error","message":"QA537-intentional-throw-partway"}
```

The product behavior is correct and documented (`server/DESIGN.md`, "Streaming startup errors"); the two expectations are stale. This is a test-only change — no product code is touched.

Both legs now assert the whole frame sequence rather than a count in a range. The old `1-2` lower bound existed only because the pre-#2614 abrupt destroy could drop the last chunk before it drained; the clean close removes that race, so the assertions get **stronger**, not looser.

## For the human reviewer

- **Why not revert #2614.** The dispatch brief allows a revert when the culprit is a single merge and the fix is not obvious. Here it is obvious and mechanical, with an in-repo pattern to copy (#2614's own update to `sse-throw-midstream.test.ts`). Reverting would trade two stale assertions for losing the fix and re-opening the QA-886/F-272 silent-truncation defect.
- **QA-702's pinned Node-vs-uWS divergence is gone, deliberately.** That leg used to branch on `HARPER_UWS_HTTP` to pin two different close shapes: Node closed abruptly without the terminal `0\r\n\r\n` chunk, uWS ended cleanly and so presented a mid-stream failure as success. #2614 carries the truncation signal in the error frame instead, and both servers now end cleanly — verified on both locally. [One converged `r.ended` expectation](https://github.com/HarperFast/harper/pull/2618/changes?w=1#diff-9c554cae3b0da7f9cfc63f5a7e4775f42e13f5584a1cb7faeda5cfef881483feR484) replaces the branch. If a server ever diverges again, this suite goes red only on that runner and re-adding the branch is a one-line revert.
- **[`strictEqual(r.events.length, 3)`](https://github.com/HarperFast/harper/pull/2618/changes?w=1#diff-7728b9344c34e5343955df6f0bc926a899b478e8fa3478fc22f9468a61956c58R174) trades flake tolerance for exactness.** Sound only because the clean end removes the truncation race. If some environment still truncates, this becomes an intermittent red; loosening back is one line.
- **Three suites now pin the same contract.** QA-537's and QA-702's own comments say the throw contract belongs to `sse-throw-midstream.test.ts` / `stream-error-contract.test.ts`, yet both now assert #2614's error-frame payload. A future contract change edits three suites. Consolidating is easy later, but the duplication is a call you may want to make now.
- **Declined — SSE reconnect after a clean end (Gemini, major → adjudicated outside-this-diff).** A browser `EventSource` auto-reconnects after any clean end on a 200, so a client must hook `harper-error` and call `.close()`. That is inherent SSE semantics that already applies to a generator finishing legitimately, it targets #2614's merged server-side contract rather than this diff, and `server/DESIGN.md:220` states the client obligation explicitly ("clients must inspect SSE for the `harper-error` event"). Not fixed here.
- **Declined — CRLF brittleness in the raw-bytes assertion (Gemini, minor).** `serializeSSEData` (`server/serverHelpers/contentTypes.ts:167-170`) and the SSE `serialize` frame builder emit `'\n'` string literals only, and actively normalize `\r\n`/`\r` in string payloads to `\n`. No CRLF path exists, and `parseSseBlocks` in the same suite already splits on `'\n\n'`. Refuted by code trace.
- **Declined — "the last commit changes assertion logic" (Gemini, major).** Mis-scoped to the branch diff rather than the delta: `14e8171` is comment-only (`git show 14e8171` has no non-comment changed line); the assertion changes landed in `72a4d8a` and were adjudicated `fixed` in round 2. The graded lens returned LGTM with no findings on the same round.

Framing-Verdict: chosen-approach-sound (planning review, gemini-3.1-pro-high)

## Changes

- `integrationTests/server/sse-finite-generator.test.ts` — case 5 (`ThrowGen`) now requires [`terminatedBy === 'end'`](https://github.com/HarperFast/harper/pull/2618/changes?w=1#diff-7728b9344c34e5343955df6f0bc926a899b478e8fa3478fc22f9468a61956c58R169), exactly 3 events, the `[0, 1]` prefix, and [the raw stream ending with the named `event: harper-error` frame](https://github.com/HarperFast/harper/pull/2618/changes?w=1#diff-7728b9344c34e5343955df6f0bc926a899b478e8fa3478fc22f9468a61956c58R181) and its exact record. That raw-bytes check is there because `parseEvents` keeps only `data: ` lines, so the frame's name and terminal position are not visible through `r.events` at all — without it, an error record emitted as an ordinary unnamed frame still passed. The [suite header comment](https://github.com/HarperFast/harper/pull/2618/changes?w=1#diff-7728b9344c34e5343955df6f0bc926a899b478e8fa3478fc22f9468a61956c58R10) is updated from #1789's teardown to #2614's contract.
- `integrationTests/server/qa702-sse-event-data.test.ts` — case b now [bounds the stream to exactly 3 blocks](https://github.com/HarperFast/harper/pull/2618/changes?w=1#diff-9c554cae3b0da7f9cfc63f5a7e4775f42e13f5584a1cb7faeda5cfef881483feR470), asserts the ordered `{"n":0}`/`{"n":1}` prefix, requires [`blocks[2].event === 'harper-error'`](https://github.com/HarperFast/harper/pull/2618/changes?w=1#diff-9c554cae3b0da7f9cfc63f5a7e4775f42e13f5584a1cb7faeda5cfef881483feR476) with its exact record, and replaces the `HARPER_UWS_HTTP` close-shape branch with [the single converged `r.ended` expectation](https://github.com/HarperFast/harper/pull/2618/changes?w=1#diff-9c554cae3b0da7f9cfc63f5a7e4775f42e13f5584a1cb7faeda5cfef881483feR484). Bounding `blocks.length` also closes a gap where a trailing data-less frame after the terminal error frame would have been invisible to both filters.

## Verification

Reproduced the CI failure locally at the current `main` head `ce08a9898` before changing anything — the same two assertions, the same raw bytes as [run 34970917284](https://github.com/HarperFast/harper/actions/runs/34970917284): 20 pass / 2 fail.

Executed route — existing integration tests, extended:

| run | result |
|---|---|
| `sse-finite-generator` + `qa702-sse-event-data` + `sse-throw-midstream` + `stream-error-contract`, Node | 44 pass / 0 fail |
| same four suites, `HARPER_UWS_HTTP=1` (the leg CI runs) | 42 pass / 0 fail / 2 skipped |
| `prettier --check` on both files | pass |
| `npm run lint:required` (what CI runs) | pass |

Fails-on-base is inverted here and is the point: both legs fail on `ce08a9898` and pass with this change. The two failing legs were confirmed identical across CI's Node 24, Bun and uWS jobs (`fail 2`, same two test names in each), and `Integration Tests` is the only red workflow at `ce08a9898` — `Unit Test`, `Lint Source Code` and `Format Check` are green there.

Full `test:integration:all` was not run locally: the change edits only two test files' assertions and cannot reach any other suite, and this box was already running another agent's integration suite against the shared `/tmp` loopback-address pool.

Refs #2614

Complexity: easy

<sub>Review-Coverage: authored=claude; ran=gemini,cursor-composer,codex; adjudicated=domain; declined=cursor-grok; rounds=3; full=1 @ 14e817185e99</sub>

<sub>Human-Review-Need: 4 @ 14e817185e99</sub>


kriszyp:
Closing as a duplicate of [Pin the terminal `harper-error` frame in the two SSE anchors #2614 missed](https://github.com/HarperFast/harper/pull/2617), which was opened before this one and already has green CI.

This PR came from the second shepherd `main-red` card for the same fingerprint (dispatched at `ce08a9898`, the push head after the one #2617's card fired on). Same root cause, same two files. #2617 is the better fix: it shares `parseSseBlocks` into `integrationTests/utils/sseStream.ts` and asserts the whole frame sequence with a single `deepStrictEqual` in both suites, where this branch reached for a raw-byte `endsWith` in QA-537 to see the `event:` name that `parseEvents` drops.

Nothing here needs carrying over. `main` goes green when #2617 merges.

— Claude Opus 5

Issue #2636, opened 2026-09-16
Backup repository operations require a loaded database; add a directory-only mode, a per-repository management lock, and restore pins


cb1kenobi:
Part of #2632 (item 4). Design note §5.5, §5.6 and §7.3 in `docs/proposals/archive-restore.md` on branch `design/archive-restore`. Both restore routes (items 5 and 6) and the importer (item 7) depend on this.

## Problems

1. **Repository operations require a loaded database.** `list_backups`, `delete_backup`, `purge_backups` and `verify_backup` all call `requireRocksRootStore` (`dataLayer/rocksdbBackup.ts:308-337`, `:473`), which needs a loaded database with at least one table. A repository for a database that does not exist yet (an import into a new name) or that is blocked by a restore marker has no loaded root, so none of these work — which means recovery from a failed restore on Fabric would end in a filesystem intervention. The offline functions (`listBackupsOffline`, `verifyBackupOffline`, …) already operate on the directory alone.
2. **A restore's source can be deleted out from under it.** Between a restore request and its execution (especially across the restart route), `delete_backup` or `purge_backups keep_count` can remove the very backup being restored, or the pre-restore backup a rollback depends on. A pin check that is not atomic with deletion is check-then-act: delete reads "unpinned", pauses, the pin lands, delete resumes.
3. **The binding's lock covers only engine files.** `backups.restore` holds `.backup.lock` shared and writers hold it exclusive, but Harper's blob snapshot and manifest phases run outside it (`:318-358`, `:374-392`) — this is the minimum slice of #2031 the restore routes cannot do without.

## What to build

- The operation forms of `list`/`verify`/`delete`/`purge` take the directory-only path whenever the database is absent or blocked, behind the same super-user check.
- A **per-repository management lock** (`tryFileLock` on a file in the repository directory, exclusive) held across: installing a pin; a delete/purge from admission through its last unlink; and the blob/manifest finalization of `create_backup` and (later) `import_backup`.
- **Pins**: `delete_backup` and `purge_backups` — operation and offline CLI — refuse an id any pending, applying or terminal restore intent names, or that an in-flight online restore holds. Pins are installed inside the management lock.
- `getStorageSpaceStats()` gains the sample's `updatedAt` in its return value (needed by the importer's disk ledger; `server/storageReclamation.ts:160-185`).

## Acceptance

- `list_backups`/`verify_backup`/`delete_backup`/`purge_backups` work against a repository whose database is absent, and against one blocked by a restore marker.
[...]

Issue #2632, opened 2026-09-16
[Epic] Restore a get_backup archive into a running Harper (incl. Fabric)


cb1kenobi:
Design note: `docs/proposals/archive-restore.md` on branch `design/archive-restore` (four planning-review rounds, the last two by codex `gpt-6-astra` at extra-high effort; both returned `chosen-approach-sound`). This epic tracks the work the note breaks down in its §9. It supersedes the rationale on #995 (closed 2026-07-31, "we don't want to introduce restoring backups while Harper is running"); #1831 landed `restore_backup` doing exactly that the next day.

## The problem

`get_backup` streams a full point-in-time snapshot of a database as `.tar`/`.tar.gz` (`createBackupStream`, `dataLayer/rocksdbBackup.ts:689`). Nothing consumes it: the archive's embedded README tells the operator to stop Harper and hand-copy files into the database directory and each blob root. On Fabric that is unusable three ways — the operator has no filesystem access, `target_database` is refused while Harper runs (`:500`, `:537`), and "stop the server" is a host-manager action. The third point is not specific to archives: any database a component declares tables in is component-held, `verifyDatabaseClosed` (`:624`) returns 409 for it, and the 409 points at a CLI the Fabric operator cannot run. So `restore_backup` itself cannot restore the databases people actually use on Fabric today.

## The approach

A `get_backup` archive and a managed backup differ only in format, and the conversion is two existing calls: an extracted stream backup "unpacks into a directory that opens directly as a RocksDB database" (rocksdb-js `docs/backups.md`), and `db.backup(repo)` on an opened directory produces a managed backup — which is exactly what `createBackupOffline` already does (`dataLayer/rocksdbBackup.ts:959-990`). Everything else an archive restore needs is already the managed restore path: lock + marker, `verifyDatabaseClosed`, `backups.restore(purgeAllFiles)`, index-aware blob restore, `verify_backup`, and the offline CLI. What that path lacks — a way to run for a held database on Fabric — it lacks for managed backups too.

So:

1. **`import_backup`** converts an archive (multipart upload or a pull from S3/HTTPS) into a verified managed backup id. No restart.
2. **A shared publication protocol** under which a restored database becomes visible: the restore lock as a real name reservation, private open, expected-table and column-family inventory checks, a restore-strict elected transaction-log replay, a blob-completeness decision, durability barriers, then publication. It carries **restore into a new name online** (create-if-absent) and upgrades in-place `restore_backup` for unheld databases.
[...]

Issue #2639, opened 2026-09-16
import_backup: convert a get_backup archive (upload or pull) into a managed backup


cb1kenobi:
Part of #2632 (item 7). Design note §5.2–§5.5 and §8 in `docs/proposals/archive-restore.md` on branch `design/archive-restore`. Depends on #2633, #2636, #2640. This is the operation #995 asked for; it is filed fresh rather than reopening #995 because the design has changed shape since.

## Problem

Nothing consumes a `get_backup` archive. The only route is the README's "stop Harper and hand-copy files into the database directory and each blob root", which requires reproducing the blob root-index mapping by hand (records persist the root index, not the path) — a silent-mis-addressing failure mode — and is impossible on Fabric.

## What to build

`import_backup` converts an archive into a **verified managed backup id** in the database's repository. No restore, no restart. Two sources, one pipeline after the bytes are on disk:

- **Push**: multipart streaming upload (the `deploy_component` machinery; `server/serverHelpers/multipartParser.ts`). The declared uncompressed size is a field part and is **advisory** (early reject only — it is client-supplied). The HTTP worker streams to staging; when the upload is complete the **main thread adopts** the staging area (owner record rewritten to the import id, fsynced) *before* the job row is inserted, so a worker that dies between enqueue and a post-enqueue rewrite cannot have its staging swept from under the job.
- **Pull**: `url` (https) or the `s3` object `import_from_s3` takes. A standard job (scalar request); resumable with `Range`; never crosses the Fabric ingress. Primary Fabric form.

**Staging** is one area beside the repository (`<backupsRoot>/<db>/.import/<importId>/`) so the blob snapshot can hard-link; entry filtering is explicit (reject `..`, absolute paths, symlinks, hardlinks, device nodes) and bounded by entry count, metered uncompressed bytes, per-entry size, path length, duplicate names — the multipart parser deliberately enforces no size cap (`multipartParser.ts:15-17`), so inode exhaustion is otherwise unbounded. Failure removes staging eagerly; the main thread sweeps worker-owned staging on worker exit; a boot sweep removes staging no live job owns.

[...]

PR #2643, opened 2026-09-16
docs: design note for restoring a get_backup archive into a running Harper

### LINKED ISSUE #2451: restore_backup rolls a database back in time with no epoch, so audit cursors, record versions, and replication sequence state all read as valid afterwards
`restore_backup` rolls a database back in time, and nothing tells a consumer holding state from after the backup point that its world changed. Several independent pieces of resumable state are affected, so each one currently looks valid on its own:

- **Audit retention floor.** `restore_backup` replaces the database directory wholesale (`dataLayer/rocksdbBackup.ts:574` → `backups.restore(..., { mode: 'purgeAllFiles' })`), so the restored database carries the *backup's* floor. A consumer that had advanced past the backup point compares its cursor against that older floor, reads it as safe, resumes — and waits for entries that no longer exist. Found reviewing #2447, which added the floor.
- **Record versions.** Restored records carry the backup's versions, so a peer or cache that saw newer versions now holds state the restored database will not converge to on its own.
- **Per-node replication sequence state.** `Symbol.for('seq')` records (`resources/nodeIdMapping.ts`) come back at their backup values, so sequence positions a peer already acknowledged are reissued.

The audit floor could be stamped with a fresh epoch during restore in isolation, and #2447 deliberately did not: making one of these three honest while the other two stay stale gives a consumer a *more* confusing answer, not a safer one — it would be told to resync the audit stream while replication and cached versions silently disagree.

## What this needs

A **database-level generation/epoch** that increments on restore, and that resumable state can be compared against:

- established at a single point in the restore sequence, after engine restoration and before `completeRestore` / worker republication (`dataLayer/rocksdbBackup.ts:574-605`), so no thread can observe the restored database at the old generation;
- readable by the things that hold cursors, so "your state predates the current generation" is answerable without each subsystem inventing its own marker;
- applied to at least the three cases above, with the audit floor's treatment being "the generation changed, so the floor is unknown" rather than a fabricated timestamp.

Worth deciding as part of the design: whether an offline (CLI) restore and an online `restore_backup` must behave identically here, and whether a restore should be *visible* to clients (an error on resume) or merely force a resync.

## Acceptance

- After a restore, a consumer resuming from a cursor recorded before the restore but after the backup point is told its state is stale rather than receiving a silently empty or short stream.
- An integration test covering backup → advance a consumer's cursor → restore → reopen, asserting the resume is refused.
- The three state carriers above either consult the generation or are explicitly documented as out of scope with the reason.

## Related

- #2447 — added the audit retention floor; documents this gap in `getAuditFloor`'s contract rather than papering over it.
- #2448 — consumes the floor inside `Table.subscribe`; whatever this issue lands will need to be visible on the same resume path.

cb1kenobi:
Design note for restoring a `get_backup` archive into a running Harper, including Fabric, where the operator has no filesystem access and cannot stop the server. This is documentation only; it is the base of the implementation stack that follows and the input to the nine issues filed under #2632.

The note went through four planning-review rounds — two by me, two by codex `gpt-6-astra` at extra-high effort — and its history table records what each changed. Round 2 replaced the chosen approach outright: the first draft designed a bespoke rename-swap protocol, and an extracted archive already *is* a database directory that two existing calls turn into a managed backup, so the design became `import_backup` into the managed repository plus a restart route on `restore_backup`. Rounds 3 and 4 both returned `chosen-approach-sound` and produced 22 findings against the protocol rather than the architecture, all adopted. Round 5 is operator review on this PR: it asked where an uploaded archive's bytes actually land, which the note had left implicit. That produced §5.4's staging-volume rules, corrected the note's account of what happens to the old database, and surfaced a new §7.7 defect.

## For the human reviewer

1. **The approach changed between drafts, and the rejected one is kept in §4 with its cost table.** The rename swap needs ~1N of writes against the managed route's ~3N, and peaks at 2N occupancy against 3N. Under a Fabric quota that difference is real, so the note states where the fallback design would win and what it would have to provide. Ruling wanted on whether that tradeoff is acceptable before the feature PRs build on it.
2. **Restore is refused on any node with peers** (§5.9), computed in core from `system.hdb_nodes` rows other than self and `replication.routes` entries. That deliberately does not close #2451: local resumable consumers still read a rolled-back database as valid, so the note says the work must either land after #2451 or mint a generation itself.
3. **A new pre-existing defect, §7.7 (work item 10):** re-pointing `storage.backupPath` at a newly attached volume leaves the old repository unmigrated and unreferenced, and `list_backups` reports `[]` rather than "your backups are on the other volume". #2636's management lock and restore pins live inside that directory too, so a silent re-point also abandons the pins protecting an in-flight restore's only rollback source. Proposed fix is a `.repository.json` stamp and a refusal naming both paths.
4. **The archive manifest records role names** (§5.10, §7.1). That is authorization metadata crossing an instance boundary in an artifact that leaves the host. It is names only, never definitions, and a restore never creates a role — but the field is hard to withdraw once consumed.
5. **Attached storage is a mount at `.import`, not a new setting** (§5.4). Uploaded bytes never reach `os.tmpdir()` — the multipart part is a `Readable` streamed into gunzip + tar-fs with no temp spool — but they do land under `storage.backupPath`, whose default is `<hdb_root>/backup`, the Harper home directory. I first proposed a `storage.importStagingPath` setting and **withdrew it**: no config parameter is live-reloadable, and the staging path does not need to be, because `<backupsRoot>/<db>/.import` is already a stable per-database location an operator can mount attached storage onto while Harper runs. The repository stays on its existing volume, so the two no longer have to be sized as one ~2N space.
6. **A mount there breaks "under the repository path" ⇒ "on the repository's filesystem", and the note now carries the invariants that follow** (§5.4): Harper creates `.import` if absent and never removes it (only `<importId>` subdirectories), the disk ledger groups by `statfs` device rather than path prefix, the staged-blob hard link degrades to a copy across `EXDEV`, and the staging device is recorded at adoption and re-checked so an unmounted mount point cannot quietly fill the OS volume. Each is a place the obvious implementation is wrong in a way nothing reveals until someone actually mounts.
7. **Pull is demoted; an archive already on the staging mount becomes the primary source** (§5.2). The objection is fair: nobody publishes a database backup at an unauthenticated URL, and a protected one needs credentials. Harper does have somewhere to put them — the `hdb_secret` store (#1550, `set_secret`, ciphertext-only, per-node custody), the same mechanism `deploy_component` uses for private npm and git tokens, carrying a **reference** rather than a token because a reference is safe to persist in `hdb_job` and a token is not. But the better answer is to need no credential at all: if the operator can mount a volume at `.import`, they can mount one that already holds the archive, which also removes the fetch policy, the resume logic and the second copy on disk. Presigned URLs stay as the weaker option, with their real limitation stated — the signature's lifetime is bounded by the signing session and can expire mid-transfer, breaking a `Range` resume.
8. **`get_backup` archives stay restorable under an explicit provenance override** (§7.1). Refusing unidentified archives outright would make the feature useless for every archive already on an operator's disk; accepting them silently would defeat the manifest. The note takes the middle road and reports the archive as unidentified.

## Changes

- [`docs/proposals/archive-restore.md`](https://github.com/HarperFast/harper/pull/2643/changes?w=1#diff-5ad6f18a5ff771b70556383487cfbb60aeec8de37ad792c40a63a1adbb2a1462R1) — the whole note: the problem and why Fabric makes it acute (§1), what already exists (§2), the invariant an acknowledged restore must satisfy (§3), the option set with the withdrawn rename-swap design and its cost table (§4), the design itself (§5), why create-if-absent is online (§6), the independently shippable prerequisites (§7), concurrency (§8), the work breakdown linked to its issues, now ten items after §7.7 (§9), outcome reporting (§10), open questions (§11) and the verification route (§12).

## Verification

Route (d) — documentation only, not observable end-to-end. The note's own behavior claims are code-traced in `origin/main`, with anything I could not trace marked `verify:` and excluded from load-bearing decisions.

Complexity: easy

---


docs/proposals/archive-restore.md
```diff
@@ -0,0 +1,1074 @@
...
+
```
cb1kenobi: Empty line is breaking the table rendering.

cb1kenobi: Good catch, thank you — the blank line was orphaning the round 4 row into its own one-row table. Removed in cf75c12, and I checked the rest of the note for the same pattern (a blank line between two table rows) and this was the only one.

The fix landed on `design/archive-restore` and the other six branches in the stack were rebased forward onto it, so every PR now carries the corrected table.

Generated by Barber AI

docs/proposals/archive-restore.md
```diff
@@ -0,0 +1,1299 @@
...
+**What being a mount point changes, and the invariants that follow.** A mount breaks the usual
+implication that a path _under_ `<backupsRoot>/<db>/` is _on_ the same filesystem as it, and every
+consequence below is a place where the obvious implementation is wrong in a way nothing reveals
+until someone actually mounts:
+
+- **Harper creates `.import` if absent and never removes it.** Only `<importId>` subdirectories are
+  created and deleted. §5.4's eager intake cleanup, §8's worker-exit sweep and the startup sweep all
+  operate one level down; an `rm -rf` of `.import` itself would fail with `EBUSY` against a live
+  mount, and would silently destroy the mount point when the volume happens not to be mounted.
+- **The disk ledger groups by device, not by path prefix.** §5.3 already says "grouped by the quota
```
kriszyp: Please make the accounting-domain change explicit here; the current helper does not become mount-aware merely by grouping callers. `getStorageSpaceStats()` returns the root quota for every resolved path beneath `ROOTPATH` (`server/storageReclamation.ts:160-185`). A filesystem mounted at `<backupsRoot>/<db>/.import` still has such a path, so a root quota reporting 500 GB available can admit extraction onto a 10 GB mount and fail only at `ENOSPC`; the reverse can reject usable attached capacity. Node's `statfs` result also has no device identifier—the filesystem key comes from `stat().dev`. Extend the stats contract to expose stable accounting keys and debit every applicable constraint (tenant quota and target filesystem, with sample time), rather than assuming the current quota-or-filesystem result models this mount.

docs/proposals/archive-restore.md
```diff
@@ -0,0 +1,1299 @@
...
+1. **Peers** (§5.9), read by core from `system.hdb_nodes` and `replication.routes`. Peers present →
+   `terminal`, not a restore.
+2. **Pre-restore backup, exactly once, while still `pending`**, unless `retain_previous: false` or
+   the intent is a supersede/rollback: `createBackupOffline` of the old directory, run in the
+   engine child (§5.5). Taken here, after the drain and with no writers, it captures the exact
+   final state — a backup taken at request time would miss every write acknowledged between it and
+   the exit. It must also be taken _only_ here: `createBackupOffline` refuses a database whose
+   marker is present (`dataLayer/rocksdbBackup.ts:965-969`), so a retry that reached `applying`
+   cannot capture again — and must not try, or it would replace the rollback source with a partial
+   generation. The backup is made durable (`db.backup` syncs by default) and its id pinned into the
```
kriszyp: The pre-restore backup is not yet exactly-once across this boundary. If `createBackupOffline` finishes and the process dies before its returned ID is written into the intent, the durable state is still `pending` with no pre-restore ID. The next boot therefore runs step 2 again; it cannot “reuse the pinned id” as the kill-injection expectation later claims. Repeated failures in this window can create multiple full blob snapshots and exhaust the space reserved for one copy. Please add a durable capture substate plus an operation identity recorded with the backup, so recovery can discover and pin the completed backup, or explicitly make this at-least-once and define orphan cleanup/accounting.

docs/proposals/archive-restore.md
```diff
@@ -0,0 +1,1299 @@
...
+`worker.terminate()` runs no `finally`. So the import lock is **owned by the main thread**: the
+worker asks over ITC, main takes `tryFileLock`, and main releases it on the worker's `exit` event.
+`restartWorkers` treats an in-flight import the way it already treats a component preparation and
+waits on it (`withComponentPreparationLock` is the precedent, and its lease-with-owner-liveness
+file is the alternative implementation if ITC ownership proves awkward).
+
+**Staging changes hands once, before the job exists.** The HTTP worker creates
+`<repo>/.import/<importId>/` and writes an owner record beside it (worker thread id, process
+instance id). When the upload completes, the handler asks the main thread — over the same ITC
+channel as the lock — to **adopt** the staging area: main rewrites the owner record to the import
```
kriszyp: This handoff leaves two ownership gaps. First, after main fsyncs the import-ID owner but before `jobs.addJob()` persists the row, termination of the HTTP worker is not cleaned by the described exit sweep—the owner no longer names that worker, and no job exists—so a fully staged archive remains until a process restart. Second, the instance-wide import lock is still described as owned for the HTTP worker's lifetime, although conversion runs in a separate job worker: releasing on HTTP-worker exit permits another import during conversion, while retaining it gives no job-completion release path. Please make adoption, durable job creation/activation, staging ownership, and lock ownership one main-thread state machine, with a recoverable pre-enqueue state and an explicit transfer to/release by the job.

docs/proposals/archive-restore.md
```diff
@@ -0,0 +1,1299 @@
...
+return (`:190-191`), and `restore_backup` fails to find ids that still exist on the old volume.
+The failure is silent in the way §5.3 and §6 care about — an operator reads "no backups" as "there
+were none", and a retention policy that counts what `list_backups` returns will happily conclude
+the database has never been backed up.
+
+This predates the design, but the design leans on the repository far harder than `create_backup`
+alone did: #2636's management lock (`.management.lock`) and the restore pins that keep a recovery
+source alive both live _inside_ `backupDir`, so a re-point also silently abandons the pins
+protecting an in-flight restore's only rollback source. Fix: record the resolved root in the
+repository (a small `.repository.json` written when the directory is created), and on mismatch
```
kriszyp: A stamp inside the repository cannot detect the re-pointing scenario this section is meant to fix. After changing `/old` to `/new`, the only `.repository.json` is under `/old/<db>`; operations derive `/new/<db>`, see no directory or stamp, and can still report `[]` or initialize a fresh repository without ever learning the old path. Consequently the proposed error cannot name both roots, and old pins remain undiscoverable. Persist the database→repository-root binding in a stable location outside `storage.backupPath` (for example restore metadata beside the database), or require an explicit repository migration/initialization operation that carries and verifies both roots.

docs/proposals/archive-restore.md
```diff
@@ -0,0 +1,1299 @@
...
+  applies to entries inside the archive.
+- **Push** — multipart streaming upload, reusing the `deploy_component` machinery. The declared
+  uncompressed size is a **field part**, not an `x-` header: the part-ordering contract guarantees
+  fields arrive before the file, so the size is available before the first archive byte reaches
+  disk. The upload is consumed in the request process (a `Readable` cannot be persisted into a job
+  row — `jobs.addJob()` stores the request, `server/jobs/jobs.ts:200`); once the archive is fully
+  staged the handler records the staging area's owner (§8), enqueues the validation/conversion job
+  with the staging path as its scalar request, and returns the job id.
+- **Pull** — `url` (https) or the same `s3` object `import_from_s3` takes. The whole operation is a
+  standard job: the request is scalar, it persists into the row, the download can resume with
```
kriszyp: Please pin the downloaded representation before promising resumable `Range` fetches. If a URL or S3 key is overwritten after a partial download, an unconditioned range retry can append bytes from a different archive. A structurally valid mixed result is possible—for example, the old engine portion plus changed blob entries—and `sha256` is optional, so later validation does not establish source identity. Capture a strong ETag or S3 version ID on the first response, send `If-Match`/the version on every retry, validate `206` and `Content-Range`, and restart from byte zero when the source provides no stable validator.

docs/proposals/archive-restore.md
```diff
@@ -0,0 +1,1299 @@
...
+Harper-side timeouts do not bound a long upload either: `requestTimeout` is unset (Fastify default 0) and `connectionTimeout` is socket inactivity. `verify:` what the Fabric ingress does with a
+multi-GB request body — the reason the pull source exists.
+
+**Where the pull credentials live.** Nobody publishes a database backup at an unauthenticated URL,
+so "just give us a URL" is not a real answer and the design does not offer one. Harper already
+solved this exact problem for private npm registries and git hosts: the **`hdb_secret` store**
+(#1550), written by `set_secret`, which stores ciphertext only and never the submitted plaintext
+(`components/secretOperations.ts:5`), gated by per-node custody. `deploy_component` ingests a
+credential into it and then carries only a **reference** through the operation, resolving it to a
+token at the point of use (`components/operations.js:639-645`, `:736-746`;
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in `4c3043c93579`.

docs/proposals/archive-restore.md
```diff
@@ -0,0 +1,1299 @@
...
+`waf` on Pro and is empty on OSS core. It carries no Pro semver, because none exists — recording
+the names is the most that is true today, and a Pro version stamp would be a change to harper-pro,
+not to this manifest.
+
+The settings block is an **allowlist, never a config dump**: an archive leaves the host, so it
+carries no path, hostname, or credential. `storage.path` and `storage.blobPaths` are excluded for
+that reason — in Fabric a storage path can embed tenant identifiers, and `blob_root_count` already
+records the only part of that layout a reader can act on. An external compression dictionary is
+recorded as a boolean, because the data depends on a file the archive does not carry
+(`resources/databases.ts:3934-3939`) and the reader needs to know that without being told where it
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in `4c3043c93579`.

docs/proposals/archive-restore.md
```diff
@@ -0,0 +1,1325 @@
...
+
+The manifest therefore carries: archive-schema version, producing Harper version (for logging),
+rocksdb-js major, transaction-log format version, blob-format features present in the archive
+(compression, marker types), source database name, whether blobs are included, the blob root
+count, and the names of the roles that held grants on the database (§5.10). The reader refuses when
+the target lacks any capability the archive declares.
+
+**An external compression dictionary is a capability, not provenance.** It was first written into
+the `source` block as a boolean, which was the wrong half of the manifest: the data cannot be read
+without a file the archive does not carry (`resources/databases.ts:3934-3939`), so a reader that
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in `4c3043c93579`.

docs/proposals/archive-restore.md
```diff
@@ -0,0 +1,1325 @@
...
+  extraction reads from it in place rather than needing room for a fetched copy _and_ its
+  expansion. The filename is the only input, resolved against the staging directory and rejected if
+  it escapes it (no absolute paths, no `..`, symlink resolved and re-checked) — the same rule §5.4
+  applies to entries inside the archive.
+
+  **Delivering the archive and extracting it are two different contracts, and an object-store mount
+  satisfies only the first.** `mountpoint-s3` is read-only for existing objects and supports only
+  sequential writes to new ones; `gcsfuse` and `blobfuse` carry similar limits. Extraction needs an
+  ordinary writable directory — it creates a tree, fsyncs it, and `db.backup()` later reads it back.
+  So the read location and the extraction target are separated: `import_backup` accepts a source
```
kriszyp, line 234: This separation cannot be represented by the surrounding path contract. The staged-path source accepts only a filename resolved under `<backupsRoot>/<db>/.import/` (`docs/proposals/archive-restore.md:216-228`), while extraction is fixed under that same `.import/<importId>` here and in the work item (`docs/proposals/archive-restore.md:1217`). If `.import` is a read-only object-store mount, the target remains beneath that mount and cannot be created; if `.import` is the repository filesystem, the filename cannot name the separate mount. Please give source and target distinct roots in the API and persisted state, validating and accounting for both, or require the object to be copied into writable staging and remove the no-copy claim.

docs/proposals/archive-restore.md
```diff
@@ -0,0 +1,1325 @@
...
+The manifest therefore carries: archive-schema version, producing Harper version (for logging),
+rocksdb-js major, transaction-log format version, blob-format features present in the archive
+(compression, marker types), source database name, whether blobs are included, the blob root
+count, and the names of the roles that held grants on the database (§5.10). The reader refuses when
+the target lacks any capability the archive declares.
+
+**An external compression dictionary is a capability, not provenance.** It was first written into
+the `source` block as a boolean, which was the wrong half of the manifest: the data cannot be read
+without a file the archive does not carry (`resources/databases.ts:3909-3914`), so a reader that
+lacks one must refuse rather than note it. It is therefore a `requires` token —
```
kriszyp, line 1018: A presence-only `external-compression-dictionary` token does not establish compatibility. If the archive was produced with dictionary D1 and the target has a different dictionary D2 configured, the target satisfies this token even though the restored data cannot be decoded correctly. Please record a stable dictionary identity—such as its algorithm and SHA-256 digest—in `requires` and compare it with the configured dictionary before admission; otherwise these archives must be refused whenever an exact match cannot be established.

kriszyp (changes requested):
I think this should go in issues, not in source.

cb1kenobi:
Agreed — closing this rather than committing the note.

The repo backs you up and I should have checked it first: every `docs/design/*.md` on `main` lands in the same commit as the implementation it documents (`analytics-aggregation-resume-marker.md` with #2692, `audit-walk-retention-floor.md` with #2684). Those record design that shipped. This was a plan for work that is mostly unbuilt, in a PR with no code, under a `docs/proposals/` path with no precedent in the tree.

It also decayed exactly the way an in-tree plan does. Three of its `file:line` citations went stale as `main` moved under the branch — `components/operations.js:639-645` and `:736-746` no longer point at `resolveCredentials`, and `resources/databases.ts:3934-3939` is past the end of a 3,927-line file. All three were correct when written.

**Where the content went:**

- **Both of your line findings are design defects, not doc issues, so they moved to the issues that own them.** The staged-path / extraction-target path contract is on [#2639](https://github.com/HarperFast/harper/issues/2639) with the two resolutions you named (distinct roots in the API and persisted state, or copy into writable staging and drop the no-copy claim). The presence-only `external-compression-dictionary` token is on [#2633](https://github.com/HarperFast/harper/issues/2633), with your point that it needs an algorithm + SHA-256 identity compared before admission, and refusal when no exact match can be established.
- **The architecture rationale is on the epic, [#2632](https://github.com/HarperFast/harper/issues/2632)** — the invariant each clause of the work is measured against, the option set with why the rename-swap approach was withdrawn and what it would have to restore if we ever need it back, and the review history. The per-item scope was already on the nine child issues.

**The implementation stack is unaffected.** #2644 is now based on `main` instead of this branch, and #2645–#2649 chain from it; all six are mergeable and their tests pass. Nothing from the note was carried into them — it only ever existed here.

Generated by Barber AI

PR #2644, opened 2026-09-16
Publish the restoring marker by rename so a torn write cannot unblock a half-purged database

cb1kenobi:
`beginRestore()` opened the restoring marker with `openSync(markerPath, 'w')`, which truncates before the database name is written, and `scanBlockedRestores()` skips any marker whose first line is empty. On a recovery attempt — over a directory an earlier restore may have half-purged — that marker is the only thing keeping the directory from loading, so a crash in the truncate-to-write window left a half-purged database that loads as healthy on the next boot.

The marker is now published temp → fsync → rename → parent fsync, so the path only ever holds the previous content or the complete new content, and an intact marker is not rewritten at all: the content a recovery attempt would write is the content already there, so the write was pure risk.

## For the human reviewer

1. **An intact marker is reused rather than rewritten, so its timestamp stops tracking the latest attempt.** Nothing parses the second line, and the original interrupted restore's time is arguably the more useful one, but it is a visible change for anyone reading the file by hand. The alternative — rewriting it — is the defect.
2. **"Intact" means the first line equals the database directory name, not merely non-empty.** A write torn mid-name leaves a prefix that resolves to a different metadata key and therefore blocks nothing, so a prefix is repaired rather than preserved. This also repairs debris left by the buggy version on upgrade.
3. **The temp name is fixed rather than unique.** The caller holds the exclusive per-database restore lock, so a temp left by an earlier crash is this database's own debris and is simply overwritten. It deliberately does not end in `.restoring`, which is how `scanBlockedRestores` selects markers.

## Changes

- [`dataLayer/restoreMarker.ts`](https://github.com/HarperFast/harper/pull/2644/changes?w=1#diff-02157616c3fc5adf480dbb09c948e4edd608fe590232798be6fc7fe700eb3b97R11) — `publishRestoringMarker` writes a temp sibling, fsyncs it, renames it over the marker path and fsyncs the directory; `markerIsIntact` compares the marker's first line against the database directory name; and `beginRestore` publishes only when there is no intact marker, flushing the directory either way.

## Verification

Route (a) — extended `unitTests/dataLayer/restoreMarker.test.js`, the suite that already covers this module, and ran it through `npm run test:unit:backup` (the repo's focused backup suite: 128 passing). Fails-on-base check performed: with the source reverted and the tests kept, `leaves an intact marker byte-identical on a recovery attempt` and `overwrites a temp file a crashed write left behind` both fail, and pass on the branch.

The test file is [`unitTests/dataLayer/restoreMarker.test.js`](https://github.com/HarperFast/harper/pull/2644/changes?w=1#diff-0a5b5d9ee381a52084f3177175d1031328dd54c217a983d2842d3e55f60ae82bR4), which gains cases for the byte-identical recovery marker, the torn-name repair, the empty-marker repair, and the absence of temp debris.

Complexity: medium

---


dataLayer/restoreMarker.ts
```diff
@@ -199,9 +221,41 @@ export function beginRestore(dbPath: string): RestoreLock {
...
+ * restore may have half-purged, so the marker it finds is the only thing keeping that directory
+ * from loading as healthy; rewriting it buys nothing (the content it would write is the content
+ * already there) and risks everything.
+ */
+export function beginRestore(dbPath: string): RestoreLock {
+	const markerPath = restoringMarkerPath(dbPath);
+	const preexisting = existsSync(markerPath);
+	const lock = acquireRestoreLock(dbPath);
+	try {
+		if (!preexisting || !markerIsIntact(markerPath, dbPath)) publishRestoringMarker(dbPath);
```
kriszyp: Could malformed `.restoring` entries fail closed before relying on this repair? The added tests show that both empty and prefix-only markers make `scanBlockedRestores()` return `[]` (`unitTests/dataLayer/restoreMarker.test.js:216-237`). Those are exactly the legacy states an interrupted recovery can leave while the target directory is already half-purged. Harper can therefore boot and load that directory before the operator retries; a concurrent process scanning between lock acquisition and this rename has the same window because the malformed content does not resolve to the correct lock path. Please stop discovery on unreadable, empty, or filename/content-mismatched markers—or make the database identity recoverable independently of the content—before repairing them under the lock.

dataLayer/restoreMarker.ts
```diff
@@ -180,17 +190,29 @@ export function releaseRestoreLock(lock: RestoreLock): void {
...
+ * left by an earlier crash is this database's own debris and is simply overwritten.
+ */
+function publishRestoringMarker(dbPath: string): void {
+	const metaDir = restoreMetaDir(dbPath);
+	const tempPath = join(metaDir, restoreMetaKey(dbPath) + MARKER_TEMP_SUFFIX);
+	try {
+		const fd = openSync(tempPath, 'w');
 		try {
 			// first line is the database directory name so the startup scan can map this marker back to
 			// the database it blocks without reversing the hashed key
```
kriszyp: Please make completion of this write part of the publication check. `writeSync()` returns the number of bytes written and may legally short-write, notably under storage or quota exhaustion, but the return value is ignored. A database-name prefix can therefore be synced and renamed as the final marker. If the restore then purges and fails under the same storage pressure, the next startup resolves that prefix to a different metadata key and can load the partial target. Use `writeFileSync(fd, contents)` or loop until every byte is written, and do not rename on a zero or incomplete write.

<sub>(`dataLayer/restoreMarker.ts:219` is not part of this PR's diff — anchored to the nearest line this PR's diff can hold)</sub>

dataLayer/restoreMarker.ts
```diff
@@ -39,18 +40,27 @@ import { tryFileLock, fileLockRelease } from '@harperfast/rocksdb-js';
...
+ *   fsync) after the lock is acquired and before the destructive restore begins; deleted only after
+ *   the restore completes successfully, while still holding the lock. Its *existence* means "a
+ *   restore started and has not finished successfully". Its first line records the database
+ *   directory name so the startup scan can map a marker back to the database it blocks without
+ *   decoding the hashed key.
+ *
+ *   The rename is what makes the marker trustworthy: the marker path only ever holds the previous
+ *   content or the complete new content, so a torn write can never replace a valid marker with one
+ *   the scan reads as empty (and therefore honors as no block). An intact marker is also never
+ *   rewritten at all — see `beginRestore`.
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in `b29e30da5c13`.

kriszyp (approved):
Looks great!

PR #2645, opened 2026-09-16
Settle jobs left unfinished by a Harper process that is gone

cb1kenobi:
A job runs in a worker thread and dies with its process, but nothing in the jobs subsystem revisited the row afterwards: `jobRunner` writes `IN_PROGRESS` before launching the worker and the worker writes `COMPLETE`/`ERROR` in its own `finally`. A crash, a `restart`, or an orchestrator replacing the container therefore left the row at `CREATED` or `IN_PROGRESS` permanently, and `get_job` reported a job that no longer exists as still running. `create_backup` and `restore_backup` already leave such rows behind under `HARPER_EXIT_ON_RESTART`.

Job rows now carry the identity of the Harper process that accepted them, and boot settles every unfinished row whose owner is not this process.

## For the human reviewer

1. **Ownership is a per-process id, not a pid.** Pids are reused, and a reused pid would make a dead job look alive. The id is minted on the main thread and inherited by every worker through the environment, so all threads of one process agree while any restart produces a new one. This mirrors `componentPreparationLock`'s process-instance pattern.
2. **Settled rows are reported as `ERROR`, not a new `INTERRUPTED` status.** Every existing `get_job` consumer already handles `ERROR`, and a new enum value would break anything switching on status. The distinction lives in the message. Say so if you would rather have the new status and take the compatibility cost.
3. **The owner is stamped at creation, not at `IN_PROGRESS`.** A row that never reached its worker is stuck in exactly the same way as one whose worker died, so the sweep covers `CREATED` too.
4. **The sweep is idempotent and non-fatal.** A row this process owns is skipped and a row it does not own becomes terminal, so a second call finds nothing; a failure to reconcile logs rather than blocking boot. An unreconciled job row is misleading, but it is not a reason to refuse to start.
5. **The owner attributes are internal and stripped from `get_job` responses.** They are added to the `hdb_job` system schema so a fresh install declares them; existing installs create them dynamically on first write.

## Changes

- [`server/jobs/jobOwnership.ts`](https://github.com/HarperFast/harper/pull/2645/changes?w=1#diff-482099e08b5bebf1b7a7434d57425ecc76fdcad7107497611e4172ff15300104R1) — new: the per-process owner id, `stampJobOwner`, and `reconcileInterruptedJobs`, which collects `CREATED`/`IN_PROGRESS` rows this process does not own before writing, because the search walks the `status` attribute the updates change.
- [`server/jobs/jobs.ts`](https://github.com/HarperFast/harper/pull/2645/changes?w=1#diff-7f63a5ee4642fbf985b7d8d810951a65c4ced90c40793d12f6fa9f312fffe552R29) — `addJob` stamps the owner on the new row, and both `get_job` surfaces strip the owner attributes from responses.
- [`server/loadRootComponents.js`](https://github.com/HarperFast/harper/pull/2645/changes?w=1#diff-5f78e88867af94a1408b2c3873addfa46be50f2cf1d025520f424457893ad3f9R70) — runs the pass on the main thread right after `getTables()`, wrapped so a failure logs rather than blocking boot; loading the module here also mints the owner id before any worker is spawned, so every worker inherits it.
- [`json/systemSchema.json`](https://github.com/HarperFast/harper/pull/2645/changes?w=1#diff-2879124913433504393c76e35dcf0b8c07418d03555c3ed00fbec1f639b6a091R92) — declares `owner_instance` and `owner_pid` on `hdb_job` so a fresh install has them; existing installs create them on first write.

## Verification

Route (b) — new `unitTests/server/jobs/jobOwnership.test.js`, run as part of the whole `unitTests/server/jobs` suite (55 passing) rather than on its own, since the module mints a process-level id. The behavior did not exist on base at all, so there is no fails-on-base variant: the module and its entry point are both new.

The test file is [`unitTests/server/jobs/jobOwnership.test.js`](https://github.com/HarperFast/harper/pull/2645/changes?w=1#diff-d8cec121d223e85d7357cd446faf95f10493e5866b035552532ff61213e01049R1), covering the stamp, each settle case, the untouched live and finished rows, idempotence, and the redaction.

Complexity: medium

---


cb1kenobi:
Both addressed in `b48edde`.

**`owner_pid` loose equality** — adopted. `== null` is the house idiom and it also covers a pid stored as `null` rather than absent, which the strict check would have rendered as `Harper process null`.

**`stampJobOwner` guard** — adopted. It is an exported function that assigns onto its argument, so verifying it has an object to assign to belongs with it rather than at the single current call site.

Generated by Barber AI

cb1kenobi:
Cross-model verification of the two adopted fixes, plus the CI question.

**Verified by codex `gpt-6-astra` (xhigh) and cursor-grok**, run against this PR's own diff rather than the cumulative stack. Both fixes confirmed correct: the `stampJobOwner` guard accepts mutable job objects while ignoring null and primitives, and `owner_pid == null` covers both a missing attribute and one stored as `null`.

Codex executed the ownership initializer in real worker and process checks rather than only reading it — a worker inherited the same UUID and pid, and a fresh main process minted a different UUID, which is the property the whole design rests on. It also traced that job workers get the id explicitly through `env: { ...process.env }` at `server/jobs/jobRunner.ts:149`, so propagation does not depend on implicit inheritance. cursor-grok returned no blockers, no significant concerns, and no suggestions.

One nit was raised — redundant commentary on the two constants — and is fixed in `c4b37c1f`. What is **not** executed, and worth a human's attention: the repo Mocha suite, native database persistence, and full crash/restart recovery. The new tests call reconciliation directly; they do not exercise worker inheritance or a real restart end to end.

**On the earlier red checks:** the failure was `unitTests/resources/analytics/aggregationCycle.test.js`, which arrived with #2692 and is unrelated to this change — this branch touches no analytics file. It passed 3/3 locally on both this branch and clean `main`, and re-running the CI jobs with no code change turned them green. Flagging it rather than burying it: that test is timing-sensitive, and `main` itself has failed 3 of its last 8 unit-test runs.

Also moved a comment trim from #2649 down into this PR, where the file actually lives — #2649 was editing `jobOwnership.ts` three levels up the stack. The stack tip is byte-identical after the move.

Generated by Barber AI

server/jobs/jobOwnership.ts
```diff
@@ -0,0 +1,80 @@
...
+ */
+export async function reconcileInterruptedJobs(): Promise<number> {
+	const jobTable = (getDatabases() as any).system?.[hdbTerms.SYSTEM_TABLE_NAMES.JOB_TABLE_NAME];
+	if (!jobTable) return 0;
+
+	// Collect before writing: the search walks the `status` attribute these updates change.
+	const interrupted: Array<{ id: any; owner_pid: any }> = [];
+	for (const status of UNFINISHED_JOB_STATUSES) {
+		for await (const job of jobTable.search([{ attribute: 'status', value: status }])) {
+			if (job.owner_instance !== JOB_OWNER_INSTANCE_ID) interrupted.push({ id: job.id, owner_pid: job.owner_pid });
```
kriszyp: Could we also settle the row when its job worker exits? A concrete failure is a job worker exceeding its `resourceLimits`: Node terminates that Worker, while the main Harper process survives. The exit path in `server/threads/manageThreads.js:600-622` removes the worker, and this launch uses `autoRestart: false`; because the row still has the live process's `JOB_OWNER_INSTANCE_ID`, this condition skips it on every reconciliation and `get_job` remains `IN_PROGRESS` until Harper itself restarts. Please attach finalization to the returned job Worker (or carry the job id into the worker lifecycle) and conditionally mark a still-unfinished row `ERROR` on exit. A focused check would terminate a real job Worker after `IN_PROGRESS` and assert the row becomes terminal without restarting Harper; the current seeded tests do not cross that production boundary.

server/jobs/jobOwnership.ts
```diff
@@ -0,0 +1,80 @@
...
+ * replacing the container leaves the row at CREATED or IN_PROGRESS permanently. `get_job` then
+ * reports a job that no longer exists as still running.
+ *
+ * Ownership is a per-process id rather than a pid: pids are reused, and a reused pid would make a
+ * dead job look alive. The id is minted once on the main thread and inherited by every worker
+ * through `process.env`, so all threads of one Harper process agree on it, while any restart —
+ * self-relaunch or orchestrator — produces a new process and therefore a new id. The pid is carried
+ * alongside it for diagnostics only.
+ */
+const JOB_OWNER_INSTANCE_ENV = 'HARPER_JOB_OWNER_INSTANCE';
```
kriszyp: The thread fabric already defines this exact PID-reuse-safe concept: `server/threads/manageThreads.js:192-196` exports `processIncarnation`, and `server/threads/manageThreads.js:562-575` propagates it to every worker through `workerData`. Could this reuse that identity, or factor it into a small shared module if importing `manageThreads` here is too coupled? A second UUID plus environment propagation channel creates two definitions of a Harper process incarnation that can drift as worker startup evolves; `owner_pid` can remain separate diagnostic metadata.

server/loadRootComponents.js
```diff
@@ -67,6 +67,18 @@ async function loadRootComponents(isWorkerThread = false) {
 	getTables();
 	resources.isWorker = isWorkerThread;
 
+	// Settle jobs whose owning process is gone, before anything can start a new one. Required on the
+	// main thread only, and deliberately not fatal: an unreconciled job row is misleading, but failing
+	// to reconcile it is no reason to refuse to boot. Loading the module here also mints the owner id
+	// before any worker is spawned, so every worker inherits it.
+	if (isMainThread) {
+		try {
+			await require('./jobs/jobOwnership.ts').reconcileInterruptedJobs();
```
kriszyp: `loadRootComponents()` is also awaited by in-process worker/component reloads, not only process boot (`server/threads/manageThreads.js:688-705`). Each call therefore runs the two `status` searches in `reconcileInterruptedJobs`; `status` is unindexed and completed job history is retained, so a node with one million historical jobs makes a deploy/restart walk roughly two million rows before replacement workers start, even though current-owner rows are skipped. Please gate successful reconciliation to once per process startup (while allowing retry after failure), or provide a bounded/indexed lookup; collapsing the two full scans would also halve the unavoidable startup work.

server/jobs/jobOwnership.ts
kriszyp: Because every job worker in a live Harper process shares this value, this comparison cannot distinguish an active job from one whose worker thread has exited. `server/jobs/jobRunner.ts:146-150` starts these workers with `autoRestart: false`; if one is terminated or crashes before its final update, its row remains `IN_PROGRESS` with the current instance ID and is skipped here on every reload. `get_job` therefore reports it as running until the entire Harper process restarts. Please bind the row to the job worker and settle unfinished state from the thread fabric’s exit notification or the returned worker’s `exit` event, with a conditional terminal update so normal completion cannot be overwritten.

server/jobs/jobOwnership.ts
kriszyp: The thread fabric already exposes exactly this process-incarnation contract at `server/threads/manageThreads.js:192-196` and propagates it to every managed worker at `server/threads/manageThreads.js:555-574`. Minting a second UUID and propagation channel creates two identities that lifecycle code can accidentally compare or carry differently. Please reuse `processIncarnation`, or move that existing identity into a dependency-neutral shared module if direct import order is a concern.

server/loadRootComponents.js
kriszyp: `loadRootComponents()` also runs during component reloads, so this is not only a cold-boot cost. Each call executes separate equality searches for `CREATED` and `IN_PROGRESS` at `server/jobs/jobOwnership.ts:58-61`, while `status` is not indexed in `json/systemSchema.json:74-76`; both searches therefore scan the entire retained `hdb_job` history before worker replacement proceeds. Please make reconciliation a once-per-process boot step and either scan once while filtering both states or index `status` so the cost is proportional to unfinished rows.

server/jobs/jobOwnership.ts
kriszyp: Same-process job-worker deaths remain nonterminal until the entire Harper process restarts. Job workers are launched with `autoRestart: false` and no job-specific exit callback in `server/jobs/jobRunner.ts:143-165`; the generic handler in `server/threads/manageThreads.js:605-622` only removes the worker. If a worker is terminated or runs out of memory before `jobProcess` reaches its `finally`, its row still carries this process's owner ID, so this condition skips it during every component reload. `get_job` therefore continues reporting `CREATED`/`IN_PROGRESS` until a full process restart changes the owner ID. Please associate the returned Worker with the job ID and, on exit, conditionally transition a still-unfinished row to `ERROR` (without racing a terminal write); keep boot reconciliation as the process-crash backstop.

server/jobs/jobOwnership.ts
kriszyp: The new owner identifier duplicates the thread fabric’s existing process-incarnation identity. `server/threads/manageThreads.js:191-196` already mints `processIncarnation`, and `server/threads/manageThreads.js:555-575` propagates it to every worker, including job workers. Maintaining a second UUID and environment-variable propagation path gives the same process two independently defined identities and invites future lifecycle disagreement. Please reuse `processIncarnation`, or factor the shared process identity into a neutral module if importing the thread manager here would create an undesirable dependency.

server/loadRootComponents.js
kriszyp: Every root-component reload scans the unindexed, retained job history twice. Reconciliation performs one equality search for each unfinished status, but `status` is not indexed in `json/systemSchema.json:74-76`, so both searches fall back to primary-store scans through `resources/search.ts:486-508`. `restartWorkers()` also calls `loadRootComponents()` during ordinary reloads, while no job-retention path bounds this table. Please index `status` with an upgrade path for existing tables, or collapse this to one scan and avoid rerunning a successful process-level reconciliation on component reloads.

kriszyp (approved):
This looks good, but only really covers process termination, a thread termination doesn't have any real settlement (at least not until process restart, as codex pointed out as well). That's fine, still worthwhile, but just noting that this is kind of half the solution, I think.

cb1kenobi:
You are right, and that was the gap — thanks for naming it precisely. #2645 only settles a row whose owning *process* is gone: the boot sweep keys on a per-process owner id, so a row abandoned by a worker *thread* inside a still-running process still carries this process's id and the sweep skips it by design. It sits at `IN_PROGRESS` until the next restart, exactly as you and codex described.

Closed in [#2763 — Settle a job whose worker thread dies while Harper keeps running](https://github.com/HarperFast/harper/pull/2763), stacked on this branch so this one can merge first and the follow-up reviews as its own diff.

The short version of what is there: the settlement moves onto a new `startWorker` `onUnexpectedExit` option that the jobs layer supplies, so a thread death with nothing else accounting for it settles the row immediately instead of waiting for a boot. A deliberate stop is left alone on purpose — a restart already replaces the worker via `startCopy()`, and a teardown hands the row to this PR's sweep — so the two mechanisms partition the cases rather than racing.

Two things from that work worth knowing here. First, the cross-model review caught me attaching the hook as a listener on the returned worker, which `startCopy()` silently drops when it restarts through `startWorker(path, options)`; making it an option is what gets replacements covered too, and there is now a test that kills a real worker to prove all three cases. Second, that same path means a dev-mode component reload re-runs an in-flight job from its persisted row — pre-existing, out of scope there, and not yet filed; say the word and I will open an issue.

The comment-trimming nit this PR deferred to "a later change" is also done in #2763, now that `server/DESIGN.md` carries the contract durably.

<sub>Generated by Barber AI</sub>

server/jobs/jobOwnership.ts
```diff
@@ -0,0 +1,70 @@
...
+ */
+export async function reconcileInterruptedJobs(): Promise<number> {
+	const jobTable = (getDatabases() as any).system?.[hdbTerms.SYSTEM_TABLE_NAMES.JOB_TABLE_NAME];
+	if (!jobTable) return 0;
+
+	// Collect before writing: the search walks the `status` attribute these updates change.
+	const interrupted: Array<{ id: any; owner_pid: any }> = [];
+	for (const status of UNFINISHED_JOB_STATUSES) {
+		for await (const job of jobTable.search([{ attribute: 'status', value: status }])) {
+			if (job.owner_instance !== JOB_OWNER_INSTANCE_ID) interrupted.push({ id: job.id, owner_pid: job.owner_pid });
```
kriszyp: A job worker can exit unexpectedly while Harper’s main process stays up. Job workers use `autoRestart: false` in `server/jobs/jobRunner.ts:143-150`, and the exit handler in `server/threads/manageThreads.js:605-622` does not settle their jobs. This reconciliation skips the row because its owner still matches the live process, including on a root reload, so `get_job` remains IN_PROGRESS until the whole process restarts. Could the job worker’s exit path mark its row ERROR, with a status check so it cannot overwrite an already completed result?

server/jobs/jobOwnership.ts
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Ownership is a per-process id rather than a pid, because pids are reused and a reused pid would
+ * make a dead job look alive. It is minted on the main thread and inherited by workers through
+ * `process.env`, so every thread of one Harper process agrees on it while any restart produces a
+ * new one. The pid rides along for diagnostics only.
+ */
+const JOB_OWNER_INSTANCE_ENV = 'HARPER_JOB_OWNER_INSTANCE';
+if (isMainThread) process.env[JOB_OWNER_INSTANCE_ENV] = randomUUID();
+export const JOB_OWNER_INSTANCE_ID = process.env[JOB_OWNER_INSTANCE_ENV] ?? randomUUID();
```
kriszyp: The thread fabric already mints a process-incarnation ID on the main thread and passes it to every worker through `workerData` (`server/threads/manageThreads.js:192-196` and `server/threads/manageThreads.js:563-574`). This adds a second ID and a separate environment propagation path for the same lifetime. Could job ownership use `processIncarnation` so recovery and the thread fabric share one definition of process identity?

server/loadRootComponents.js
kriszyp: This hook also runs on root-component reloads, including worker restarts (`server/threads/manageThreads.js:688-695`). Each call searches CREATED and IN_PROGRESS separately in `server/jobs/jobOwnership.ts:48-51`; `status` has no index in `json/systemSchema.json:74-76`. Thus each reload walks retained job history twice even when every row belongs to this live process. Could reconciliation run once per process boot, or use an indexed lookup that bounds the work to unfinished rows?

kriszyp (approved):

PR #2646, opened 2026-09-16
Refuse an engine-only restore over a database that still has blobs

cb1kenobi:
`restoreBlobSnapshot` leaves the live blob roots untouched and logs a warning when a backup has no blob snapshot. That is right in isolation — purging them would strip blobs the restored records still reference — but it means an in-place restore of an `exclude_blobs` backup silently produces a mixed generation: rolled-back records addressing whichever blobs are on disk now. The common outcome is a dangling reference to a blob deleted since the backup point; the worse one is a blob id reissued after those deletions plus a restart, which resolves a restored record onto unrelated bytes. The operator got a log line and a successful restore.

That restore is now refused unless the caller passes `allow_engine_only`, which is recorded in the result.

## For the human reviewer

1. **This breaks existing scripted restores of engine-only backups.** Anything running `restore_backup` against an `exclude_blobs` backup over a database that has blobs now fails until the flag is added. That is the point — the previous behavior was silent corruption — but the default and the flag name are both hard to change after a release.
2. **The refusal only fires when the database actually has blob files.** An empty root has nothing to disagree with, so the common engine-only workflow on a blobless database is unaffected.
3. **`target_database` is never affected**, since a new database has no pre-existing blobs.
4. **One rule, applied in the online operation and the offline CLI**, so the two cannot drift.

## Changes

- [`dataLayer/blobBackup.ts`](https://github.com/HarperFast/harper/pull/2646/changes?w=1#diff-ef6081f3b8edd93e7e5b3c8bb011fc94a74142673b26b7b93d2f31718d71fc98R338) — `blobRootsHaveFiles` stops at the first file it finds across every configured root, and `assertEngineOnlyRestoreAllowed` refuses the in-place case, with the blob-id reuse mechanism recorded as the reason purging the roots instead is not the answer.
- [`dataLayer/rocksdbBackup.ts`](https://github.com/HarperFast/harper/pull/2646/changes?w=1#diff-bbfcea2dd873a1ccd771306cbcf3cc20d95f4c7a01bb171b92f13683195fcb9eR30) — both restore paths call the guard before anything destructive, `validateRestoreBackup` type-checks `allow_engine_only`, the results record the opt-in, and the repository README documents the refusal.
- [`bin/backup.ts`](https://github.com/HarperFast/harper/pull/2646/changes?w=1#diff-704a8de08ce4323c9a25356ef71f54d6bfc110667d1b231f021d137528b9c946R80) — forwards `allow_engine_only` to the offline restore.

## Verification

Route (a) — extended `unitTests/dataLayer/blobBackup.test.js` and `unitTests/dataLayer/rocksdbBackup.test.js`, run through `npm run test:unit:backup` (128 passing). Fails-on-base check performed: with the source reverted and the tests kept, `refuses an in-place restore of an engine-only backup while the database still has blobs` fails on base and passes on the branch.

The test files are [`unitTests/dataLayer/blobBackup.test.js`](https://github.com/HarperFast/harper/pull/2646/changes?w=1#diff-b0412778c1b85da425ad68cce79648858c1a8d7bcc2bb972c74e35639d982591R8), covering `blobRootsHaveFiles` and the admission matrix directly, and [`unitTests/dataLayer/rocksdbBackup.test.js`](https://github.com/HarperFast/harper/pull/2646/changes?w=1#diff-ede16e467225ab11c1aa3231129e337571cd4fd93c86c0b2d17b8b44fa3f1c6fR424), covering the refusal, the opt-in, the empty-roots case and `target_database`.

Complexity: medium

---


dataLayer/rocksdbBackup.ts
```diff
@@ -558,7 +564,14 @@ export async function restoreBackup(request: any) {
 			: resolveDatabasePath(databaseName);
 	// reject a backup with more blob roots than the current config *before* anything destructive —
 	// restoring it would mis-address blobs (records persist their root index)
-	await assertBlobSnapshotRestorable(backupDir, backupId, getBlobPathsForDatabaseName(databaseName));
+	const blobRoots = getBlobPathsForDatabaseName(databaseName);
+	await assertBlobSnapshotRestorable(backupDir, backupId, blobRoots);
+	const allowEngineOnly = requireBooleanOption(request.allow_engine_only, 'allow_engine_only');
+	await assertEngineOnlyRestoreAllowed(databaseName, blobRoots, {
```
kriszyp: This safety decision can race with writes because `dataLayer/rocksdbBackup.ts:570-574` runs before both the restore marker and the worker-close broadcast at `dataLayer/rocksdbBackup.ts:575-585`. Concrete sequence: an engine-only backup references blob id 1; that live blob is later deleted and a restart leaves the root empty; this check passes; another request worker creates a new id-1 blob before receiving the close event; then the engine restore resurrects the old record, which resolves to the new unrelated bytes without `allow_engine_only`. Please retain the early rejection if useful, but re-check after worker closure/`verifyDatabaseClosed()` and immediately before destructive restore—or establish an equivalent blob-write barrier and perform the check inside it. A deterministic regression can insert the blob between the initial check and the close acknowledgement.

cb1kenobi:
Cross-model review — codex `gpt-6-astra` (xhigh), Gemini 3.1 Pro, and cursor-composer, run against this PR's own base so only its diff was in scope. Verdict was CHANGES, and it found a real hole. Fixed in `d7c950fc`.

**The `target_database` exemption was unsound, and the test encoded the bug.** The guard skipped when the restore named a different database, on the reasoning that a new target has no blobs to disagree with. But a destination's blob roots live *outside* its database directory, so "no database of that name" does not establish "no blobs of that name" — a dropped database leaves reclaimable blobs behind for a retention window. Restoring an engine-only backup into that name reproduced exactly the mixed generation this PR exists to prevent. The existing test hid it by only ever targeting a fresh `${BLOB_DB}-copy`.

The fix removes the `inPlace` parameter entirely and always asks the destination's blob roots. A genuinely new target passes naturally because its roots are missing or empty, so the guard got simpler as well as correct. The misleading test is replaced by one that asserts a named target with files present is refused.

**Second fix: the online check ran before writers were stopped.** It sat ahead of the close broadcast, so a blob written between check and purge would survive it. There is now a preflight at admission — which can only refuse, never clear, and exists so an operator is not handed a database taken down for a restore that was never going to be allowed — and the sound check after `verifyDatabaseClosed`. The offline path already had the equivalent ordering.

**Two findings I did not take, with the evidence:**

- *CLI `allow_engine_only=true` parsed as a string, silently ignoring the opt-in.* Not reachable: `buildRequest` runs `JSON.parse` on every `key=value` whose field is not in `RAW_STRING_FIELDS` (`bin/cliOperations.ts:634-646`), and that set is `{'ref'}` — so the value arrives as a boolean.
- *`requireBooleanOption` making `allow_engine_only` mandatory and breaking existing restore scripts.* Not reachable: it returns false for `undefined` and only throws on a non-boolean (`dataLayer/rocksdbBackup.ts:169-174`).

**Caveat worth stating:** the domain adjudication leg was killed by SIGKILL, so these findings were not adjudicated — I triaged them myself, which is weaker. Both unadjudicated dismissals above are backed by the cited lines rather than by judgement.

174 tests passing across the stack's suites.

Generated by Barber AI

dataLayer/blobBackup.ts
```diff
@@ -335,14 +335,65 @@ export async function assertBlobSnapshotRestorable(
...
+				throw error;
+			}
+			for (const entry of entries) {
+				if (entry.isFile()) return true;
+				if (entry.isDirectory()) stack.push(join(dir, entry.name));
+			}
+		}
+	}
+	return false;
+}
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in `1526d82c8810`.

dataLayer/blobBackup.ts
```diff
@@ -335,14 +335,70 @@ export async function assertBlobSnapshotRestorable(
...
+ * Restoring engine files while leaving the live blob roots alone produces a mixed generation:
+ * rolled-back records addressing whichever blobs happen to be on disk now. The common outcome is a
+ * dangling reference to a blob deleted since the backup point; the worse one is a blob id reissued
+ * after the deletions plus a restart (ids are a per-database counter re-seeded from a directory scan
+ * — `resources/blob.ts` `getNextFileId`), which resolves a restored record onto unrelated bytes.
+ *
+ * Purging the roots instead is not the answer — that strips blobs the restored records still
+ * reference — so the operator has to choose, and the choice is recorded. A restore into a new
+ * database name is unaffected: it has no pre-existing blobs to disagree with.
+ */
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in `8a5870de67f7`.

dataLayer/blobBackup.ts
```diff
@@ -101,6 +101,25 @@ const BACKUP_MARKER_REASONS: CaptureMarkerReasons = {
...
 
+export async function* walkBlobFiles(root: string): AsyncGenerator<string> {
+	const stack: string[] = [root];
+	while (stack.length > 0) {
+		const dir = stack.pop() as string;
+		let entries;
+		try {
+			entries = await readdir(dir, { withFileTypes: true });
+		} catch (error: any) {
+			if (error.code === 'ENOENT') continue;
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in `8a5870de67f7`.

dataLayer/blobBackup.ts
```diff
@@ -335,14 +336,56 @@ export async function assertBlobSnapshotRestorable(
...
+ *
+ * Restoring engine files while leaving the live blob roots alone produces a mixed generation:
+ * rolled-back records addressing whichever blobs happen to be on disk now. The common outcome is a
+ * dangling reference to a blob deleted since the backup point; the worse one is a blob id reissued
+ * after the deletions plus a restart (ids are a per-database counter re-seeded from a directory scan
+ * — `resources/blob.ts` `getNextFileId`), which resolves a restored record onto unrelated bytes.
+ *
+ * Purging the roots instead is not the answer — that strips blobs the restored records still
+ * reference — so the operator has to choose, and the choice is recorded. A restore into a new
+ * database name is unaffected: it has no pre-existing blobs to disagree with.
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in `8a5870de67f7`.

dataLayer/rocksdbBackup.ts
```diff
@@ -570,6 +583,14 @@ export async function restoreBackup(request: any) {
 		// purging — restoring under an open instance would corrupt it. If handles remain, fail
 		// with a clear pointer to the offline CLI path rather than purging.
 		await verifyDatabaseClosed(databaseDir, databaseName);
+		// Re-check now that no writer is left: the preflight above ran while the database was still
+		// serving, so a blob written between the two would otherwise survive the purge and produce the
+		// mixed generation this guard exists to prevent. This is the check that is actually sound; the
+		// preflight is there to fail a doomed request before the database is taken down.
+		await assertEngineOnlyRestoreAllowed(databaseName, blobRoots, {
```
kriszyp, line 594: Worker closure is not a blob-write barrier. `saveBlob` allocates the path and starts an asynchronous file pipeline, while the schema handler only closes database handles; it does not cancel or await those pipelines. Concrete sequence: a request allocates/reuses id 1 and yields while its file open is pending; the restore signal closes RocksDB and acknowledges; this walk sees an empty root; the delayed open then creates id 1 while `backups.restore` installs the old record that references id 1. That produces the mixed generation without `allow_engine_only` (or abort cleanup can later leave the restored record dangling). Please stop new blob saves and await each worker's in-flight saves before acknowledging closure, then perform this check while that barrier is held; replication receives should join the same core barrier. A focused regression can delay blob-file creation until after close acknowledgement and assert that restore never reaches `backups.restore`.

dataLayer/rocksdbBackup.ts
```diff
@@ -1014,7 +1023,12 @@ export async function restoreBackupOffline(databaseName: string, backupId?: numb
 	}
 	// reject a backup with more blob roots than the target's current config before anything
 	// destructive (records persist their root index, so collapsing would mis-address blobs)
-	await assertBlobSnapshotRestorable(backupDir, backupId, getBlobPathsForDatabaseName(targetDatabase ?? databaseName));
+	const blobRoots = getBlobPathsForDatabaseName(targetDatabase ?? databaseName);
+	await assertBlobSnapshotRestorable(backupDir, backupId, blobRoots);
+	await assertEngineOnlyRestoreAllowed(targetDatabase ?? databaseName, blobRoots, {
```
kriszyp, line 1038: The offline safety decision still runs before both `beginRestoreForDatabase` and the RocksDB lock probe. Concrete sequence: the CLI takes the offline branch during a restart's PID-file gap; an engine-only backup once referenced blob id 1, whose live file has since been reclaimed, so this check passes; the overlapping Harper process creates a new id-1 blob and exits before the probe; the probe then succeeds and the engine restore resurrects the old record onto the new unrelated bytes without `allow_engine_only`. Please retain this early preflight if useful, but repeat it after the probe at `dataLayer/rocksdbBackup.ts:1056` and immediately before destructive restore. At that point the marker blocks new loaders and the probe has excluded an existing holder. A deterministic regression can insert the blob between the first check and marker/probe and assert that restore rejects.

dataLayer/blobBackup.ts
```diff
@@ -101,6 +101,25 @@ const BACKUP_MARKER_REASONS: CaptureMarkerReasons = {
...
 
+export async function* walkBlobFiles(root: string): AsyncGenerator<string> {
+	const stack: string[] = [root];
+	while (stack.length > 0) {
+		const dir = stack.pop() as string;
+		let entries;
+		try {
+			entries = await readdir(dir, { withFileTypes: true });
+		} catch (error: any) {
+			if (error.code === 'ENOENT' || error.code === 'ENOTDIR') continue;
```
kriszyp: `walkBlobFiles` is also the enumerator used by `copyTree` and streamed backup archives, so swallowing `ENOTDIR` here changes capture behavior, not just the presence guard. With the newly tested shape—a configured `<blobPath>/<database>` replaced by a regular file—a normal `create_backup` now completes an empty snapshot and publishes a manifest with `blobs: true`; `verify_backup` only checks that the snapshot directory exists, so the incomplete backup verifies and later restores records without their blob bytes. `get_backup` likewise silently omits the root. Previously both capture walks propagated `ENOTDIR`. Please keep any intentional `ENOTDIR` handling local to `blobRootsHaveFiles`, or otherwise make capture/archive consumers reject a non-directory root, and cover this at the snapshot/create boundary so no completion manifest is published.

kriszyp (commented):
Looks good, a few findings related to timing, but I think this sounds correct.

dataLayer/rocksdbBackup.ts
```diff
@@ -1014,7 +1023,12 @@ export async function restoreBackupOffline(databaseName: string, backupId?: numb
...
+		backupHasBlobs: manifest.blobs,
+		allowEngineOnly,
+	});
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in `5f6f10a8bb4d`.

dataLayer/rocksdbBackup.ts
cb1kenobi: Addressed in `5f6f10a8bb4d`.

dataLayer/rocksdbBackup.ts
cb1kenobi: Addressed in `5f6f10a8bb4d`.

dataLayer/blobBackup.ts
cb1kenobi: Addressed in `5f6f10a8bb4d`.

resources/blob.ts
```diff
@@ -1776,6 +1776,43 @@ export function createBlobFromStoredBody(
...
+	if (!saves) inFlightBlobSaves.set(databaseName, (saves = new Set()));
+	saves.add(saving);
+	const remove = () => {
+		saves.delete(saving);
+		if (saves.size === 0) inFlightBlobSaves.delete(databaseName);
+	};
+	saving.then(remove, remove);
+}
+
+export async function blockBlobSavesForRestore(databaseName: string): Promise<void> {
```
kriszyp: This fences only writes represented in `inFlightBlobSaves`; generation-specific deletion work remains active. Concrete sequence: backup B contains blob id 1; the current database supersedes that record, so `deleteBlob` queues id 1 for delayed reclamation; restore B before the retention timer fires. The close phase drains saves and restores B's id-1 file, then the old timer unlinks that newly restored path. Restore reports success, but the restored record is unreadable. An in-progress orphan sweep has the same shape. Please make the restore barrier cover all blob-root mutations and invalidate old-generation reclamation/sweep work after a successful restore, while resuming valid work after a pre-destructive abort. A deterministic regression can queue deletion behind a held timer, restore a backup referencing that file, release the timer after reload, and assert the bytes remain.

server/threads/itc.js
```diff
@@ -35,6 +35,13 @@ function sendItcEvent(event) {
 	// The main thread's threadId is 0 (worker_threads convention); parentPort.threadId
 	// is set to 0 in workers, so sendToThread(0, ...) routes back to main.
 	if (event.message) event.message.originator = threadId;
+	if (
+		event.type === hdbTerms.ITC_EVENT_TYPES.SCHEMA &&
+		event.message?.operation === hdbTerms.OPERATIONS_ENUM.RESTORE_BACKUP &&
+		event.message.restorePhase === 'close'
+	) {
+		return broadcastWithAcknowledgement(event, 0);
```
kriszyp: Passing `0` turns one stalled save into an unbounded restore/database wedge. Generic file-backed streams intentionally have no idle deadline, so a client can start a save and never end it; its handler then waits forever in `Promise.allSettled`, this broadcast never resolves, and restore never reaches its catch path to remove a fresh marker and send `reload`. New blob saves remain rejected until the process restarts. Please keep close strict—never proceed destructively without every acknowledgement—but give it a deadline that rejects and aborts the restore rather than the existing best-effort timeout that resolves. The pre-destructive abort can then clear the marker and broadcast `reload`. A focused check can hold a `PassThrough` open and assert restore fails without entering `backups.restore` and blob saves are admitted again.

dataLayer/rocksdbBackup.ts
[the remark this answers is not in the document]
cb1kenobi: Addressed in `59b43ddd0630`.

dataLayer/rocksdbBackup.ts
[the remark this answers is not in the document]
cb1kenobi: Addressed in `59b43ddd0630`.

dataLayer/blobBackup.ts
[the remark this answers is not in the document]
cb1kenobi: Addressed in `59b43ddd0630`.

dataLayer/rocksdbBackup.ts
```diff
@@ -558,18 +565,30 @@ export async function restoreBackup(request: any) {
...
 		// A live component (or the system database) can hold its own handle on the database that
 		// Harper does not track and cannot close, so verify actual process-wide closure before
 		// purging — restoring under an open instance would corrupt it. If handles remain, fail
 		// with a clear pointer to the offline CLI path rather than purging.
 		await verifyDatabaseClosed(databaseDir, databaseName);
+		// Re-check while every worker's blob-save barrier is held. The preflight above ran while the
+		// database was still serving, so it can only fail early; this is the sound admission check.
+		await assertEngineOnlyRestoreAllowed(databaseName, blobRoots, {
+			backupHasBlobs: manifest.blobs,
+			allowEngineOnly,
```
kriszyp: The restore barrier ignores queued deletion work that can remove a newly restored blob. Concrete sequence: backup B contains blob id 1; the current database supersedes that record, so `deleteBlob` queues id 1 for delayed reclamation; restore B before the retention timer fires. Closing the database does not cancel that queue, so the restore reinstalls B's id-1 file and the old timer can later unlink the newly restored path. Restore reports success, but the restored record becomes unreadable. A fire-and-forget `cleanupOrphans` sweep has the same generation-crossing shape. Please make the restore barrier cover every blob-root mutator and invalidate old-generation reclamation/sweep work after a successful restore, while resuming valid work after a pre-destructive abort. A deterministic regression can hold queued deletion work, restore a backup referencing that file, release the work after reload, and assert the restored bytes remain.

resources/blob.ts
```diff
@@ -1776,6 +1776,43 @@ export function createBlobFromStoredBody(
...
+const inFlightBlobSaves = new Map<string, Set<Promise<void>>>();
+
+function blobSaveDatabaseName(store: any): string | undefined {
+	return typeof store?.databaseName === 'string' ? store.databaseName : undefined;
+}
+
+function assertBlobSaveAllowed(store: any): void {
+	const databaseName = blobSaveDatabaseName(store);
+	if (databaseName && blockedBlobSaveDatabases.has(databaseName)) {
+		throw new Error(`Cannot save a blob for database '${databaseName}' while it is being restored`);
```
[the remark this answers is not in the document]
cb1kenobi: Addressed in `21d3b0a2b2b2`.

resources/blob.ts
```diff
@@ -1776,6 +1777,45 @@ export function createBlobFromStoredBody(
...
+		if (saves.size === 0) inFlightBlobSaves.delete(databaseName);
+	};
+	saving.then(remove, remove);
+}
+
+export async function blockBlobSavesForRestore(databaseName: string): Promise<void> {
+	blockedBlobSaveDatabases.add(databaseName);
+	const saves = inFlightBlobSaves.get(databaseName);
+	if (saves) await Promise.allSettled([...saves]);
+}
```
kriszyp: `resumeBlobSavesAfterRestore` is not scoped to the restore that acquired the block. Concrete sequence: restore A calls `completeRestore` at `dataLayer/rocksdbBackup.ts:622`, releasing the per-database lock before its reload broadcast has reached every worker; restore B can then acquire the lock and deliver its `close` phase to a worker; a delayed reload from A reaches that worker afterward and this unconditional `delete` clears B's active block. B can now pass its blob-root check and destructive restore while a normal save writes into the roots, recreating mixed-generation corruption. Please give close/reload a restore-generation token and only clear a matching block, or keep a separate serialization lock through reload acknowledgement. A deterministic regression can order the messages as A-close, B-close, A-reload and assert saves remain blocked until B-reload.

dataLayer/blobBackup.ts
```diff
@@ -335,16 +336,63 @@ export async function assertBlobSnapshotRestorable(
...
+	blobRoots: string[],
+	{ backupHasBlobs, allowEngineOnly }: { backupHasBlobs: boolean; allowEngineOnly: boolean }
+): Promise<void> {
+	if (backupHasBlobs || allowEngineOnly) return;
+	// Deliberately no in-place/target distinction: what makes a restore unsafe is blob files at the
+	// destination, and a destination's blob roots live outside its database directory, so "this name
+	// has no database directory" does not establish "this name has no blobs" — a dropped database
+	// leaves reclaimable blobs behind for a retention window. Asking the roots answers it directly,
+	// and a genuinely new target passes because its roots are missing or empty.
+	if (!(await blobRootsHaveFiles(blobRoots))) return;
```
kriszyp: An empty root does not establish that an engine-only backup has no file-backed references; it only means those referenced IDs are currently available for reuse. Concrete sequence: an engine-only backup contains a record referencing blob id 1, id 1 is reclaimed, and the backup is restored into a new/empty target without `allow_engine_only`; when that target first saves a blob, `getNextFileId` scans the empty roots and allocates id 1 (`resources/blob.ts:2730-2766`), so the restored record now resolves to unrelated new bytes. This is the same mixed-generation corruption the guard is intended to prevent, merely deferred until the next write. Please require explicit opt-in for every engine-only restore unless the backup is proven to contain no external blob references, or persist/restore a non-reusable blob generation/high-water mark. A regression can restore a backup containing id 1 into an empty target, insert one blob, and verify the restored row never reads the new bytes.

dataLayer/rocksdbBackup.ts
```diff
@@ -558,18 +565,30 @@ export async function restoreBackup(request: any) {
...
+		// Each thread also rescans, and the restoring marker keeps it from reloading mid-restore.
+		await signalling.signalSchemaChange(restoreSchemaEvent(databaseName, 'close'));
 		// A live component (or the system database) can hold its own handle on the database that
 		// Harper does not track and cannot close, so verify actual process-wide closure before
 		// purging — restoring under an open instance would corrupt it. If handles remain, fail
 		// with a clear pointer to the offline CLI path rather than purging.
 		await verifyDatabaseClosed(databaseDir, databaseName);
+		// Re-check while every worker's blob-save barrier is held. The preflight above ran while the
+		// database was still serving, so it can only fail early; this is the sound admission check.
+		await assertEngineOnlyRestoreAllowed(databaseName, blobRoots, {
```
kriszyp: The new barrier only drains promises in `inFlightBlobSaves`; it does not suspend or reconcile the existing `pendingReclamation` queue (`resources/blob.ts:1481-1629`) or other already-scheduled unlinks. Concrete sequence: a write supersedes blob id 1 and queues its path, restore closes the workers before the retention deadline, the blob snapshot restores id 1, and the old timer then unlinks that newly restored path. Restore reports success while the restored engine record has lost its blob. Please make the close phase quiesce every blob-root mutator, including reclamation/cleanup, and cancel or reconcile pre-restore deletion intents before reload. A deterministic regression can queue id 1 with delayed reclamation, restore a snapshot containing id 1, advance the deletion, and assert the restored bytes remain.

dataLayer/blobBackup.ts
kriszyp: An empty root does not establish that an engine-only backup has no file-backed references; it only means those referenced IDs are currently available for reuse. Concrete sequence: an engine-only backup contains a record referencing blob id 1, id 1 is reclaimed, and the backup is restored into a new/empty target without `allow_engine_only`; when that target first saves a blob, `getNextFileId` scans the empty roots and allocates id 1 (`resources/blob.ts:2730-2766`), so the restored record now resolves to unrelated new bytes. This is the same mixed-generation corruption the guard is intended to prevent, merely deferred until the next write. Please require explicit opt-in for every engine-only restore unless the backup is proven to contain no external blob references, or persist/restore a non-reusable blob generation/high-water mark. A regression can restore a backup containing id 1 into an empty target, insert one blob, and verify the restored row never reads the new bytes.

dataLayer/rocksdbBackup.ts
kriszyp: The new barrier only drains promises in `inFlightBlobSaves`; it does not suspend or reconcile the existing `pendingReclamation` queue (`resources/blob.ts:1481-1629`) or other already-scheduled unlinks. Concrete sequence: a write supersedes blob id 1 and queues its path, restore closes the workers before the retention deadline, the blob snapshot restores id 1, and the old timer then unlinks that newly restored path. Restore reports success while the restored engine record has lost its blob. Please make the close phase quiesce every blob-root mutator, including reclamation/cleanup, and cancel or reconcile pre-restore deletion intents before reload. A deterministic regression can queue id 1 with delayed reclamation, restore a snapshot containing id 1, advance the deletion, and assert the restored bytes remain.

dataLayer/rocksdbBackup.ts
cb1kenobi: Addressed — there is a real blob-write barrier now, not just handle closure.

`blockBlobSavesForRestore(database)` adds the database to a blocked set so `assertBlobSaveAllowed` rejects new saves, then awaits every promise in `inFlightBlobSaves` for that database (`resources/blob.ts:1806-1810`). The schema handler awaits it *before* `closeDatabase` (`server/itc/serverHandlers.js:54-57`), and the close broadcast goes out through `broadcastWithAcknowledgement` (`server/threads/itc.js:39-44`), so the restore does not proceed until every worker has both stopped accepting saves and drained the pipelines it had already started. `resumeBlobSavesAfterRestore` lifts the block on the `reload` phase.

That closes the sequence you described: the delayed `open` can no longer land after the walk, because the save is either rejected up front or awaited before the walk runs.

Generated by Barber AI

dataLayer/rocksdbBackup.ts
cb1kenobi: Addressed — the offline path now checks twice, and the decisive check is after the probe.

`assertEngineOnlyRestoreAllowed` runs at `dataLayer/rocksdbBackup.ts:1032` as a preflight (it can only refuse, never clear, so an operator is not handed a downed database for a restore that was never going to be allowed), and again at `:1064` — after `beginRestoreForDatabase` (`:1038`) has published the marker and after the RocksDB `LOCK` probe (`:1044-1057`), immediately before `destructionStarted = true` at `:1068`.

So in your sequence the overlapping process that creates a new id-1 blob during the PID-file gap is caught: the marker is published before the probe, the probe establishes nobody else holds the database, and the re-check then sees that blob. This mirrors the online ordering, where the equivalent re-check runs once the close broadcast and blob-save drain have completed.

Generated by Barber AI

dataLayer/blobBackup.ts
cb1kenobi: Agreed, and the shared walk does not swallow it. `walkBlobFiles` tolerates only `ENOENT` (a directory removed mid-walk) and rethrows everything else, `ENOTDIR` included (`dataLayer/blobBackup.ts:109-114`).

So the shape you named — a configured `<blobPath>/<database>` replaced by a regular file — now fails the enumeration rather than producing an empty snapshot that `create_backup` would publish with `blobs: true`. That is the right direction for exactly the reason you gave: the enumerator is shared with `copyTree` and the streamed archive, so tolerating it there would change capture behavior and not just the presence guard, and a blob root that is a file is a misconfiguration worth failing loudly.

Worth noting one consequence, since it is a behavior change: `blobRootsHaveFiles` now propagates that error instead of answering false, so an engine-only restore against such a root fails with the filesystem error rather than a clean refusal. That seems right — the operator has to fix the root either way — but flagging it rather than leaving it implicit.

Generated by Barber AI

resources/blob.ts
```diff
@@ -1776,6 +1776,48 @@ export function createBlobFromStoredBody(
...
+
+export function resumeBlobSavesAfterRestore(databaseName: string): void {
```
kriszyp: `resumeBlobSavesAfterRestore` deletes a database-wide Boolean with no association to the restore that established it. Concrete sequence: restore A releases its lock before awaiting its `reload` broadcast (`dataLayer/rocksdbBackup.ts:618-619`); restore B acquires the lock, completes its `close` phase, and adds the database to this set; then a delayed A reload reaches this worker and deletes B's block. New saves are accepted during B's post-close check and destructive restore. Please make the barrier generational or lease-based—carry a restore token in both phases and release only the matching token, or reference-count ownership—and add an interleaving regression where A's reload lands after B's close and saves remain rejected until B reloads.

server/threads/itc.js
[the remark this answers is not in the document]
cb1kenobi: Addressed in `05b891fea9d1`.

dataLayer/blobBackup.ts
kriszyp: An empty destination root is not evidence that an engine-only backup is blob-free. Concrete sequence: the backup's records reference blob id 1, it was created with `exclude_blobs`, and the destination roots are empty, so this returns; restore installs those records; after reload, `getNextFileId` seeds from the empty roots and allocates id 1 (`resources/blob.ts:2731-2769`), so the next unrelated blob is written at the path now referenced by the restored record. This also applies to a new target. Please require `allow_engine_only` for every backup whose manifest excludes blobs unless the engine can be proven to contain no file-backed references, or persist/restore a monotonic blob-id high-water mark that cannot fall behind restored references. A regression should restore a record referencing id 1 into empty roots, save a new blob, and verify the restored record never resolves to the new bytes.

dataLayer/rocksdbBackup.ts
```diff
@@ -558,18 +565,36 @@ export async function restoreBackup(request: any) {
...
 	const lock = beginRestoreForDatabase(databaseDir, databaseName);
 	let destructionStarted = false;
 	try {
-		// close the database across all worker threads (each thread also rescans, and the
-		// restoring marker keeps the scan from reloading it mid-restore)
-		await signalling.signalSchemaChange(new SchemaEventMsg(process.pid, OPERATIONS_ENUM.RESTORE_BACKUP, databaseName));
+		// Block new blob saves, drain in-flight saves, and close the database across all worker threads.
+		// Each thread also rescans, and the restoring marker keeps it from reloading mid-restore.
+		try {
+			await signalling.signalSchemaChange(restoreSchemaEvent(databaseName, 'close'));
```
kriszyp: The close signal here waits only for `inFlightBlobSaves`; it does not fence deletions. Concrete sequence: an update queues old file id 1 in `pendingReclamation` (`resources/blob.ts:1480-1512`), restore receives every save acknowledgement and restores backup blob id 1, then the old timer calls `unlink(filePath)` (`resources/blob.ts:1622-1628`) and deletes the newly restored bytes. A fire-and-forget `cleanup_orphan_blobs` walk can similarly outlive the close acknowledgement (`dataLayer/schema.ts:381-387`). Please make the restore barrier cover every blob-root mutation: stop new deletions/cleanup, drain or invalidate active work before acknowledging, and discard old-generation reclamation entries when the generation changes. Add a regression that queues deletion of an id restored by the backup, advances the reclamation delay after reload, and verifies the restored file remains.

resources/blob.ts
kriszyp: `resumeBlobSavesAfterRestore` deletes a database-wide Boolean with no association to the restore that established it. Concrete sequence: restore A releases its lock before awaiting its `reload` broadcast (`dataLayer/rocksdbBackup.ts:628-630`); restore B acquires the lock, completes its `close` phase, and adds the database to this set; then a delayed A reload reaches this worker and deletes B's block. New saves are accepted during B's post-close check and destructive restore. Please make the barrier generational or lease-based—carry a restore token in both phases and release only the matching token, or reference-count ownership—and add an interleaving regression where A's reload lands after B's close and saves remain rejected until B reloads.

kriszyp (approved):
A good safety measure.

Issue #2031, opened 2026-07-31
create_backup: serialize backup management ops to close a finalize/purge race


cb1kenobi:
Follow-up from #1831 review (flagged by @kriszyp on `dataLayer/rocksdbBackup.ts`).

## Problem
`create_backup` is two-phase: `rootStore.backup()` writes the RocksDB engine backup, then `finalizeBackup()` copies the blob snapshot and publishes the completion manifest.

The rocksdb-js backup-directory **writer lock is released when `rootStore.backup()` resolves** — *before* `finalizeBackup()` copies blobs and writes the manifest. So a concurrent `purge_backups` / `delete_backup` on the same database can acquire that lock, see and purge the just-created engine backup, and return — while `create_backup` keeps copying blobs and then writes a manifest (and reports a `backup_id`) for an engine backup that no longer exists.

Net: `create_backup` can report success with a phantom `backup_id`, and a manifest can outlive its engine backup.

## Scope / severity
Only bites when two backup **admin** operations run concurrently on the same database — uncommon, but a real correctness gap.

## Suggested fix
Hold a per-database backup-management lock (an OS flock like the restore lock in `dataLayer/restoreMarker.ts`) across `create_backup` / `delete_backup` / `purge_backups` so they serialize through completion, i.e. the whole engine-backup + blob-snapshot + manifest sequence is exclusive. This is the 'serialize management operations through completion' aspect deferred during #1831.

Add a regression that races a create (during its blob-copy phase) against a purge.

PR #2647, opened 2026-09-16
Manage a backup repository without a loaded database, under one lock, with pins

cb1kenobi:
Three gaps on the backup-repository surface, each of which turns a recoverable situation into a filesystem intervention an operator on Fabric cannot perform.

`list_backups`, `verify_backup`, `delete_backup` and `purge_backups` all required a loaded database with at least one table, but a repository outlives its database: a failed restore leaves the database blocked by its restoring marker with the repository intact, and a backup can exist for a name that does not exist yet. Those operations now apply the engine gate only when there is a loaded database to gate, and otherwise run against the directory. rocksdb-js's `.backup.lock` covers only the engine files, so a purge could still remove a blob snapshot or completion manifest out from under a `create_backup` that was finalizing (#2031); every Harper-managed mutation of a repository now takes a Harper-level management lock. And backups that something depends on can now be pinned, so a delete or purge that would remove one is refused.

## For the human reviewer

1. **Online `list`/`verify`/`delete`/`purge` now delegate to the directory-only implementations and drop the RocksDB engine gate when the database is unloaded.** This widens who can mutate a repository with no loaded database to vouch for it. A name that is neither a loaded database nor a repository is still a 404, so a typo does not answer with an empty list. Easy to re-narrow if you would rather keep today's refusal.
2. **`purge_backups` no longer delegates "keep the newest N" to the binding.** Harper picks the oldest ids and deletes them individually, because `db.backup()` does not take the management lock and a create landing in between shifted which ids the binding kept — including, possibly, one whose pin had just been checked. Harper now owns retention semantics (id order, not engine timestamp order), which is behavioural and visible to operators.
3. **A pin counts only while the database it names still carries a restoring marker.** That is the whole lifetime rule: no owner record, no liveness probe, no boot sweep. A process killed mid-restore leaves both marker and pin and the rerun clears both; a process killed between clearing the marker and releasing the pin leaves a pin that is ignored and swept lazily. The alternative — an owner-instance record like the job rows use — is a larger on-disk contract to change later.
4. **A pin file that cannot be parsed blocks every id.** What it was protecting is exactly what cannot be determined, so it fails closed. A pin unlinked between the readdir and the read is a released claim, not a torn file.
5. **`finalizeBackup` re-checks the engine backup still exists before writing the manifest.** The engine backup is created outside the management lock, so a purge can remove it in between; publishing a manifest for engine files that are gone is the #2031 false-green shape. The create now fails loudly instead.
6. **This takes only the minimum of #2031**, not its full end-to-end serialization, which can still ship separately.

## Changes

- [`dataLayer/backupRepository.ts`](https://github.com/HarperFast/harper/pull/2647/changes?w=1#diff-4775ffbf445997f53395fcf246980711e9271d6a37477c21af1885cf2f5c59efR1) — new: the waiting management lock, and pins with the lifetime rule that a pin counts only while the database it names still carries a restoring marker, including the lazy sweep of one that has lapsed.
- [`utility/durableFile.ts`](https://github.com/HarperFast/harper/pull/2647/changes?w=1#diff-3588ac9e8a0f862e24cbeb98181af8186eb73d71b883f019822df78626febf67R1) — new: `writeFileDurably`, `removeFileDurably` and `fsyncDirectory`, the temp-fsync-rename-fsync pattern the marker, the intent and the pins all need.
- [`dataLayer/rocksdbBackup.ts`](https://github.com/HarperFast/harper/pull/2647/changes?w=1#diff-bbfcea2dd873a1ccd771306cbcf3cc20d95f4c7a01bb171b92f13683195fcb9eR5) — the engine gate applies only when a database is loaded; the online operations delegate to the directory-only implementations; delete and purge run under the lock, refuse pinned ids, delete exactly the ids they admitted, and reconcile Harper-managed files in a `finally`; `finalizeBackup` re-checks the engine backup still exists; and both restore paths claim and release their source.
- [`dataLayer/restoreMarker.ts`](https://github.com/HarperFast/harper/pull/2647/changes?w=1#diff-02157616c3fc5adf480dbb09c948e4edd608fe590232798be6fc7fe700eb3b97R3) — drops its private directory-fsync helper for the shared one.

## Verification

Route (b) — new `unitTests/dataLayer/backupRepository.test.js` and `unitTests/utility/durableFile.test.js`, plus extensions to `unitTests/dataLayer/rocksdbBackup.test.js`, all run through `npm run test:unit:backup` (128 passing) and `npx mocha "unitTests/utility/*.js"` (435 passing). Fails-on-base check performed: with `dataLayer/rocksdbBackup.ts` reverted and the tests kept, all four new behavioral tests fail on base and pass on the branch.

The test files are [`unitTests/dataLayer/backupRepository.test.js`](https://github.com/HarperFast/harper/pull/2647/changes?w=1#diff-0093866d92522c475a7579fb20d287005a3a16078cac8b1f80cb1cb1acaaf4f6R1) for lock serialization and the pin lifetime, [`unitTests/utility/durableFile.test.js`](https://github.com/HarperFast/harper/pull/2647/changes?w=1#diff-4a2813785821f988499bb4ef1c8ee4c169b8d627dbab7f4cea3d712ceb8b3060R1) for the write barriers, and [`unitTests/dataLayer/rocksdbBackup.test.js`](https://github.com/HarperFast/harper/pull/2647/changes?w=1#diff-ede16e467225ab11c1aa3231129e337571cd4fd93c86c0b2d17b8b44fa3f1c6fR27) for maintenance without a loaded database, the pin refusals, the exact-id purge and the orphan sweep.

Complexity: complicated

---


cb1kenobi:
All three addressed.

**`readBackupPins` parsing a non-object** — real, fixed in `0595ea9`. `JSON.parse` returns `null`, a number or a string without throwing, and reading `backup_id` off one of those threw past the catch, failing the delete with a 500 rather than failing closed the way a torn pin does. The shape is now validated inside the `try`, so an invalid pin takes the same fail-closed path.

**`fsyncDirectory` not guarding `fsyncSync`** — correct, and already fixed earlier on this branch in `d84cf60`; the review ran against the previous head. The Windows unit gate caught it first: Windows opens a directory happily and rejects the flush with EPERM, so every durable write failed there. Both limbs now tolerate EPERM/EISDIR/ENOTSUP/EINVAL.

**Throwing from the `finally`** — real, fixed in `0595ea9` at both the delete and purge sites. The reconcile is best-effort cleanup, so it now logs rather than replacing the engine error that triggered it.

Generated by Barber AI

